PR #434, opened 2016-02-19
Show auth window when performing import project from private repo or use factory with private repo

mshaposhnik:
Show auth window when performing import project from private repo or use factory with private repo


mshaposhnik:
@skabashnyuk  @vparfonov

vparfonov:
ok

skabashnyuk:
ок

PR #443, opened 2016-02-19
improve exception messages in MachineManager

garagatyi:
Signed-off-by: Alexander Garagatyi agaragatyi@codenvy.com

@skabashnyuk


skabashnyuk:
ок

PR #446, opened 2016-02-19
CHE-50: Fix NPE when loading GitHub repositories list

vinokurig:
NPE appeared when loading GitHub repositories list and the projects list is empty in project explorer
@vparfonov please review


skabashnyuk:
ok

vparfonov:
ok

PR #350, opened 2016-02-16
IDEX-4222: Change model of machine & workspace

garagatyi:
Machine model changes:
- MachineState was removed, Machine should be used instead
- Machine states API was removed
- Machine includes MachineConfig as field instead of extending it
- All runtime information of machine is in runtime field of Machine.
- Now API call GET _/api/machine?workspace=ws-id123456_ returns all existing machines of workspace,
  including machines with status CREATING, DESTROYING
- Machine channels was removed from. To get channels clients should check links of WorkspaceConfigDto.

Workspace model changes:
- EnvironmentState was removed
- UsersWorkspace includes WorkspaceConfig as field instead of extending it
- Start tmp workspace API returns RuntimeWorkspace instead of UsersWorkspace
- Machine channels are included in links of MachineConfigDto of WorkspaceConfig. Don't try to find them
  in runtime machines of RuntimeWorkspaceDto.
- Field activeEnvName in RuntimeWorkspace was renamed to envName in accordance with defaultEnv field
  of WorkspaceConfig.

Also:
- add static method cloneDto to DtoFactory
- add getParameter(paramName) method to DTO Link to simplify parameter retrieval

Signed-off-by: Alexander Garagatyi agaragatyi@codenvy.com

@skabashnyuk @vparfonov @evoevodin @dimasnurenko Please, review.

PR #449, opened 2016-02-19
CHE-494: fix machine and workspace models

olexii4:
**_1 Upvote**_ Signed-off-by: Oleksii Orel oorel@codenvy.com

### DIFF
--- a/dashboard/src/app/dashboard/welcome-back/welcome-back.controller.js
+++ b/dashboard/src/app/dashboard/welcome-back/welcome-back.controller.js
@@ -69,7 +69,7 @@ export class WelcomeBackCtrl {
     // there are workspaces
     if (workspaces.length > 0) {
       // select the last one
-      this.lastWorkspace = workspaces[0].name;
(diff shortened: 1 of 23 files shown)


olexii4:
@ashumilova @benoitf  WDYT?

benoitf:
I would say that this commit should be part of IDEX-4222 and https://github.com/eclipse/che/pull/350 so client and model changes will be merged together

benoitf:
(but I'm already ok with content)

ashumilova:
do we need to change here: https://github.com/eclipse/che/blob/CHE-494/dashboard/src/app/dashboard/welcome-back/welcome-back.controller.js#L72

ashumilova:
and in this method:  https://github.com/eclipse/che/blob/CHE-494/dashboard/src/app/projects/create-project/create-project.controller.js#L346

ashumilova:
+1

olexii4:
Fixed

PR #454, opened 2016-02-22
Fall back to unix socket on Linux if DOCKER_HOST is malformed

ghost:
**_1 Upvote**_ Use of DEFAULT_DOCKER_MACHINE_URI on Linux makes little sense when DOCKER_HOST is malformed, so to us better to use unix socket in this case.

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
@@ -212,9 +212,9 @@ public class DockerConnectorConfiguration {
             try {
                 userURI = new URI(host);
             } catch (URISyntaxException e) {
-                LOG.error(String.format("Unable to parse %s property with value %s", DOCKER_HOST_PROPERTY, host), e);
+                LOG.error(String.format("Unable to parse environment variable %s with the following value - %s", DOCKER_HOST_PROPERTY, host), e);
                 // unable to use given property, fallback to default URL
-                return DEFAULT_DOCKER_MACHINE_URI;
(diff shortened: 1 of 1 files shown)


plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
```diff
@@ -44,7 +44,7 @@
     /**
      * Default ip of docker host (Linux system).
      */
-    protected static final String DEFAULT_LINUX_DOCKER_HOST_IP = "172.17.42.1";
+    protected static final String  DEFAULT_LINUX_DOCKER_HOST_IP = "172.17.42.1";
```
benoitf: extra space

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
```diff
@@ -212,9 +212,9 @@ protected static URI dockerDaemonUri(final boolean isLinux, @NotNull final Map<S
             try {
                 userURI = new URI(host);
             } catch (URISyntaxException e) {
-                LOG.error(String.format("Unable to parse %s property with value %s", DOCKER_HOST_PROPERTY, host), e);
+                LOG.error(String.format("Unable to parse %s value of environment variable %s", DOCKER_HOST_PROPERTY, host), e);
```
benoitf: I'm not sure that order of formatted string args are then correct with the syntax change
It will display 
    Unable to parse DOCKER_HOST value of environment variable http://foo
while I think you expect
    Unable to parse http://foo value of environment variable DOCKER_HOST

garagatyi:
LGTM

garagatyi:
@skabashnyuk ?

skabashnyuk:
ок

benoitf:
+1 by squashing the two commits

PR #467, opened 2016-02-22
Adopt tree widget according to changes in project service

vzhukovs:
Project service is returning all registered projects with sub-projects when getProjects method is called.
So in this case root project is filtered and the other one is cached on the client side for performance improve.

Also were added helpful class from jdt to work with paths, which will be need in future to work with new client side project model.

Related issue: CHE-72

@vparfonov @azatsarynnyy @gazarenkov

### DIFF
--- a/core/commons/che-core-commons-gwt/pom.xml
+++ b/core/commons/che-core-commons-gwt/pom.xml
@@ -26,6 +26,10 @@
     </properties>
     <dependencies>
         <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+        </dependency>
+        <dependency>
             <groupId>com.google.gwt</groupId>
             <artifactId>gwt-elemental</artifactId>
         </dependency>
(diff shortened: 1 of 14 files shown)


azatsarynnyy:
Have you seen already existing class [Path](https://github.com/eclipse/che/blob/project_api_4/core/platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/Path.java)?
May be we can avoid using yet another **Path** class?

vzhukovs:
I'm not sure if `vfs/Path.java` will work on the client side, because it requires Pattern, which is not listed in JRE emulated classes. And I suppose that `vfs/Path.java` uses only on the server side, that's why I re-used and adopted jdt's path for the client side purpose. Moreover, to use `vfs/Path.java` we need to add server side dependency or move `vfs/Path.java` to common module.

azatsarynnyy:
Sorry. I haven't seen that this is for client

gazarenkov:
To me Path interface makes no sense. Lets have it as utility class only,

azatsarynnyy:
This utility class may be useful for extesnions developers also so I propose to move it somewhere to https://github.com/eclipse/che/tree/master/core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide

vzhukovs:
@gazarenkov done, interface was replaced with the implementation.
@azatsarynnyy done, the class has moved into GWT commons module ([Path.java](https://github.com/eclipse/che/blob/CHE-72/core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/resource/Path.java)).

vparfonov:
ok

PR #468, opened 2016-02-22
CHE-314. Exclude che-core-api-analytics from packaging into assembly-machine-war

RomanNikitenko:


vparfonov:
ok

PR #470, opened 2016-02-22
CODENVY-31: Replaced messages from git importer on error codes

akorneta:
Also added git checkout event which goes through by websocket connection

@skabashnyuk @vparfonov

### DIFF
--- a/core/platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitProjectImporter.java
+++ b/core/platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitProjectImporter.java
@@ -18,6 +18,7 @@ import org.eclipse.che.api.core.ForbiddenException;
 import org.eclipse.che.api.core.ServerException;
 import org.eclipse.che.api.core.UnauthorizedException;
 import org.eclipse.che.api.core.model.project.SourceStorage;
+import org.eclipse.che.api.core.notification.EventService;
 import org.eclipse.che.api.core.util.FileCleaner;
 import org.eclipse.che.api.core.util.LineConsumerFactory;
 import org.eclipse.che.api.git.shared.Branch;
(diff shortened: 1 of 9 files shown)


skabashnyuk:
ok for me. @vparfonov @sleshchenko ?

core/platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitWebSocketMessenger.java
```diff
@@ -0,0 +1,63 @@
...
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+/**
+ * @author Anton Korneta.
+ */
+@Singleton
+public class GitWebSocketMessenger implements EventSubscriber<GitCheckoutEvent> {
+    private static final Logger LOG = LoggerFactory.getLogger(GitWebSocketMessenger.class);
+    private static final String CHANNEL = "git:events:channel";
```
sleshchenko: You should use different channels for each project in each workspace

core/platform-api/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitProjectImporter.java
```diff
@@ -323,11 +325,20 @@ private void sparsecheckout(GitConnection git,
...
         new File(project, ".gitignore").delete();
     }
+
+    private String getRemoteBranch(DtoFactory dtoFactory, GitConnection git, String branchName) throws GitException {
+        final List<Branch> remotes = git.branchList(dtoFactory.createDto(BranchListRequest.class)
+                                                              .withListMode(BranchListRequest.LIST_REMOTE));
+        final Optional<Branch> first = remotes.stream()
+                                              .filter(br -> branchName.equals(br.getDisplayName()))
+                                              .findFirst();
+        return first.get().getName();
```
sleshchenko, line 350: Think about NoSuchElementException here. Is it possible here?

sleshchenko:
other OK

skabashnyuk:
ок

PR #481, opened 2016-02-23
CHE-127: Load repo button on import from Github takes too much time

tolusha:
Signed-off-by: Anatoliy Bazko abazko@codenvy.com

### DIFF
--- a/plugins/plugin-github/che-plugin-github-ext-github/src/main/java/org/eclipse/che/ide/ext/github/client/GitHubClientService.java
+++ b/plugins/plugin-github/che-plugin-github-ext-github/src/main/java/org/eclipse/che/ide/ext/github/client/GitHubClientService.java
@@ -174,6 +174,8 @@ public interface GitHubClientService {
 
     /**
      * Get the list of the organizations, where authorized user is a member.
+     *
+     * Use {@link #getOrganizations()}.
      */
     @Deprecated
     void getOrganizations(@NotNull AsyncRequestCallback<List<String>> callback);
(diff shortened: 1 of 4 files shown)


skabashnyuk:
ok. @sleshchenko ?

plugins/plugin-github/che-plugin-github-ext-github/src/main/java/org/eclipse/che/ide/ext/github/shared/GitHubRepository.java
```diff
@@ -208,6 +208,16 @@
 
     void setHasDownloads(boolean isHasDownloads);
 
+
+    /**
+     * Get the repository parent.
+     *
+     * @return the parent repository.
+     */
+    GitHubRepository getParent();
```
sleshchenko, line 221: Add comment that parent can be nullable

plugins/plugin-github/che-plugin-github-ext-github/src/main/java/org/eclipse/che/ide/ext/github/client/GitHubClientService.java
```diff
@@ -175,11 +175,23 @@ void createPullRequest(@NotNull String user, @NotNull String repository, @NotNul
     /**
      * Get the list of the organizations, where authorized user is a member.
      */
+    @Deprecated
```
sleshchenko, line 180: Add link to new method that can be used instead of this one

sleshchenko:
Other OK for me

PR #491, opened 2016-02-24
CHE-567: add directive for formating output

ashumilova:
**_2 Upvotes**_

### DIFF
--- a/dashboard/src/components/attribute/format-output/che-format-output.directive.js
+++ b/dashboard/src/components/attribute/format-output/che-format-output.directive.js
@@ -1,0 +1,51 @@
+/*
+ * Copyright (c) 2015-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ */
+'use strict';
+
+/**
+ * Defines a directive for formatting output.
+ * @author Ann Shumilova
+ */
+export class CheFormatOutput {
+
+  /**
+   * Default constructor that is using resource
+   * @ngInject for Dependency injection
+   */
+  constructor(jsonOutputColors, $compile) {
+    this.restrict = 'A';
+    this.outputColors = angular.fromJson(jsonOutputColors);
+    this.$compile = $compile;
+  }
+
+  /**
+   * Keep reference to the model controller
+   */
+  link($scope, element, attr) {
+    $scope.$watch(attr.ngModel, (value) => {
+      if (!value || value.length === 0) {
+        return;
+      }
+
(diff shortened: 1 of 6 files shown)


ashumilova:
output

dashboard/src/components/attribute/format-output/che-format-output.directive.js
```diff
@@ -0,0 +1,60 @@
...
+      let lines = value.split('\n');
+      let result = '';
+
+      lines.forEach((line) => {
+
+        for (var i = 0; i < this.outputColors.length; i++ ) {
+          let outputColor = this.outputColors[i];
+          line = line.replace('[' + outputColor.type + ']', '[<span style=\"color: ' + outputColor.color + '\">' + outputColor.type + '</span>]');
+
+          //Additional formatting cause may have output like this - [INFO ]
```
benoitf: could we use a regexp to match`[<any space>type<any space>]`?

also replace by default works only on the first occurence
so I think we have [INFO] extracting [INFO] 10% only first occurence will be matched

dashboard/src/components/attribute/format-output/che-format-output.directive.js
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * Keep reference to the model controller
+   */
+  link($scope, element, attr) {
+    $scope.$watch(attr.ngModel, (value) => {
+      if (!value || value.length === 0) {
+        return;
+      }
+
+      let lines = value.split('\n');
```
benoitf: I think this code is a good candidate to have unit tests so we can check if with all our input values, the result is expected

dashboard/src/components/attribute/format-output/che-format-output.directive.js
```diff
@@ -0,0 +1,60 @@
...
+          let outputColor = this.outputColors[i];
+          line = line.replace('[' + outputColor.type + ']', '[<span style=\"color: ' + outputColor.color + '\">' + outputColor.type + '</span>]');
+
+          //Additional formatting cause may have output like this - [INFO ]
+          if ('INFO' === outputColor.type) {
+            line = line.replace('[' + outputColor.type + ' ]', '[<span style=\"color: ' + outputColor.color + '\">' + outputColor.type + '</span>]');
+          }
+        }
+
+        result += '<br/>' + line;
```
benoitf: I think this code always add a blank carriage return at the beginning of the output while output should not have a "blank line" first

benoitf:
+1

olexii4:
+1

riuvshin:
why this was merged? build fail with those  changes.

garagatyi:
I supose that answer is because we don't have trigger that shows if PR brake the build

ashumilova:
build failed cause PR contained adding new auto-generated file. it was not failing locally, cause that file was on file system

benoitf:
yes it was issue only when we were running tests without having previously running app (which is only matching jenkins, not real life)
but yes as @garagatyi said, I would be a lot easier to have PR being checked automatically

PR #493, opened 2016-02-24
Fix equality bug in paste action

vzhukovs:
Also was performed little refactoring to make code more readable.
Also was added utility class which works with path's. This is client adopted utility class which is used on the server side jdt implementation. Can be useful for third-paty extensions. _(This class has created in Project Api 4 branch and anyway will be merged into master, so we can add it to the master to start work with it)_

Relevant issue: CHE-572

@vparfonov


vparfonov:
ok

PR #495, opened 2016-02-24
Updates per issue

TylerJewell:
1. Update to use latest /assembly format.
2. Add in the logback logger to eliminate unnecessary errors
3. Fix error in IF logic around skipping maven check.
4. Add in printing of all used directories in debug mode.

### DIFF
--- a/assembly/assembly-main/src/assembly/assembly.xml
+++ b/assembly/assembly-main/src/assembly/assembly.xml
@@ -79,6 +79,24 @@
         </dependencySet>
         <dependencySet>
             <useProjectArtifact>false</useProjectArtifact>
+            <unpack>false</unpack>
+            <outputDirectory>sdk</outputDirectory>
+            <outputFileNameMapping>che-plugin-sdk-logger.jar</outputFileNameMapping>
+            <includes>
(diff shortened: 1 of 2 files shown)


TylerJewell:
@skabashnyuk Please ok this - needed to fix some issues with users building plug-ins.

riuvshin:
@TylerJewell you have to sign your commit.

TylerJewell:
@riuvshin Actually, we got confirmation from Eclipse Foundation that committers do not need to sign. Only contributors.  The "All checks have failed" issue is a bug with the Eclipse integration and they said that we can ignore it if we know a committer has issued the PR.

riuvshin:
@TylerJewell yeah I know but this looks like a bug and it's better to sign to pass this validation.

TylerJewell:
@riuvshin Actually, the previous commit was signed:
https://github.com/eclipse/che/commit/14b3ff81532cbecbf067474092a000b81f4994af and the error still exists.  I think we are safe to merge.

riuvshin:
@TylerJewell yeah, I see, looks like a bug

PR #502, opened 2016-02-24
CODENVY-88: change secure random source to urandom

riuvshin:
Signed-off-by: Roman Iuvshin riuvshin@codenvy.com

@skabashnyuk @vparfonov WDYT?


skabashnyuk:
-1 
Set this value in che.sh

garagatyi:
-1
I believe we should make it configurable

riuvshin:
Agree, we need to manage this in other way by providing JAVA_OPTS for dev machine.

PR #513, opened 2016-02-25
CHE-141: Add client side validation for packages' and class' names

dimashnurenko:
@vparfonov

### DIFF
--- a/plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/JavaUtils.java
+++ b/plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/JavaUtils.java
@@ -10,16 +10,47 @@
  *******************************************************************************/
 package org.eclipse.che.ide.ext.java.client;
 
+import com.google.gwt.regexp.shared.RegExp;
+
 import org.eclipse.che.ide.runtime.IStatus;
-import org.eclipse.che.ide.runtime.Status;
+
+import java.util.Arrays;
+import java.util.List;
 
 /**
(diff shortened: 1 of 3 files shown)


vparfonov:
ok

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/JavaUtils.java
```diff
@@ -92,28 +139,6 @@ public static void checkPackageName(String name) throws IllegalStateException {
...
-            case Status.WARNING:
-            case Status.OK:
-                return true;
-            default:
-                return false;
-        }
-    }
-
-    private static IStatus validateCompilationUnitName(String name) {
-//        return JavaConventions.validateCompilationUnitName(name, JavaCore.getOption(JavaCore.COMPILER_SOURCE),
```
azatsarynnyy, line 106: May be we can keep using JavaConventions class by porting it to client?
It provides really useful methods for checking Java-specific conventions corresponding to the Java Language Specification. Also it provides informative hints why some name isn't valid.

dimashnurenko, line 106: The porting JavaConventions to client complicates client side validation code. Based on this comments  
https://github.com/eclipse/che/pull/513/files#diff-092710f973f752e67aa777e2cec6c3bcL115

we've already had JavaConventions validation on client side, but it was complicated.
In my implementation i use regex to validate packages and class names in according to java name convention.

vparfonov:
ok

PR #514, opened 2016-02-25
CHE-395 Add visual feedback to clarify whether machine process is finished or still running

vitaliy-guliy:


vparfonov:
ok

PR #518, opened 2016-02-25
CHE-22: check environment parameter on workspace start

garagatyi:
Signed-off-by: Alexander Garagatyi agaragatyi@codenvy.com

@skabashnyuk @evoevodin Please review.

### DIFF
--- a/core/platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/DefaultWorkspaceConfigValidator.java
+++ b/core/platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/DefaultWorkspaceConfigValidator.java
@@ -30,7 +30,7 @@ import static java.lang.String.format;
 @Singleton
 public class DefaultWorkspaceConfigValidator implements WorkspaceConfigValidator {
     /* should contain [3, 20] characters, first and last character is letter or digit, available characters {A-Za-z0-9.-_}*/
-    private static final Pattern WS_NAME = Pattern.compile("[\\w][\\w\\.\\-]{1,18}[\\w]");
+    private static final Pattern WS_NAME = Pattern.compile("[a-zA-Z0-9][-_.a-zA-Z0-9]{1,18}[a-zA-Z0-9]");
 
     /**
      * Checks that workspace configuration is valid.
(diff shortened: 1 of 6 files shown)


skabashnyuk:
ок

core/platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/DefaultWorkspaceConfigValidator.java
```diff
@@ -30,7 +30,7 @@
 @Singleton
 public class DefaultWorkspaceConfigValidator implements WorkspaceConfigValidator {
     /* should contain [3, 20] characters, first and last character is letter or digit, available characters {A-Za-z0-9.-_}*/
-    private static final Pattern WS_NAME = Pattern.compile("[\\w][\\w\\.\\-]{1,18}[\\w]");
+    private static final Pattern WS_NAME = Pattern.compile("[a-zA-Z0-9][-a-zA-Z0-9.]{1,18}[a-zA-Z0-9]");
```
voievodin: You made a small mistake here as `\\w` includes underscore `_` as well, so the second `[-a-zA-Z0-9.]` should include underscore.

core/platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/DefaultWorkspaceConfigValidator.java
```diff
@@ -116,4 +127,14 @@ private void requiredNotNull(Object object, String message) throws BadRequestExc
             throw new BadRequestException(message);
         }
     }
+
+    /**
+     * Checks that first argument is not null nor empty string, throws {@link BadRequestException}
+     * in the case of null or empty {@code string} with given {@code message}.
+     */
+    private void requiredNeitherNullNorEmpty(String string, String message) throws BadRequestException {
```
voievodin: I don't think it is a good to make such checks.
The motivation is that what if we will need another check e.g. _non-null and non-empty collection_ should we then add a new method? My proposal is to change this method to something like `checkArgument(boolean condition, String message)` and reuse it throughout the validator.
The good example is [Preconditions](http://guava-libraries.googlecode.com/svn-history/r14/trunk/javadoc/com/google/common/base/Preconditions.html)

core/platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/DefaultWorkspaceConfigValidator.java
garagatyi: yes, 10x

core/platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -588,6 +592,21 @@ UsersWorkspaceImpl performAsyncStart(UsersWorkspaceImpl workspace,
         }
         workspace.setTemporary(false);
         workspace.setStatus(WorkspaceStatus.STARTING);
+
+        if (envName != null) {
+            if (!workspace.getConfig()
```
voievodin: second `if` statement is an excessive one, use 1 statement for these checks

core/platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/DefaultWorkspaceConfigValidator.java
garagatyi: Would you like me to replace method requiredNotNull too?

core/platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
garagatyi: ok

core/platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/DefaultWorkspaceConfigValidator.java
voievodin: No, i think that this method should leave, as we use it too often

garagatyi:
@skabashnyuk @evoevodin  I reworked the PR. Please take a look.

skabashnyuk:
ok

core/platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -588,6 +592,19 @@ UsersWorkspaceImpl performAsyncStart(UsersWorkspaceImpl workspace,
         }
         workspace.setTemporary(false);
         workspace.setStatus(WorkspaceStatus.STARTING);
+
+        if (envName != null && !workspace.getConfig()
```
voievodin, line 596: It seems that it is the same but shorter

``` java
!workspace.getConfig()
          .getEnvironments()
          .stream()
          .anyMatch(env -> env.getName().equals(envName);
```

core/platform-api/che-core-api-workspace/src/test/java/org/eclipse/che/api/workspace/server/WorkspaceManagerTest.java
```diff
@@ -106,10 +109,13 @@ public User getUser() {
 
     @Test
     public void shouldBeAbleToCreateWorkspace() throws Exception {
+        // given
```
voievodin: Why did you do that?
As you can see _given, when, then_ sections are separated with an empty line which is enough to understand the test, i try to follow this pattern and avoid explicit comments.

core/platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
garagatyi, line 596: lokks better, thanks

core/platform-api/che-core-api-workspace/src/test/java/org/eclipse/che/api/workspace/server/WorkspaceManagerTest.java
garagatyi: it looked more clear to me. I can delete comments if you want.

voievodin:
ok

PR #520, opened 2016-02-25
Poject & Factory & GitServiceClient improvements

voievodin:
Publish `CurrentProjectChangedEvent` when root project is already set.
Exclude projects without location when creating factory from workspace.
Force `AttributesFilter` to use empty list instead of `null`.
Add `status` method to `GitServiceClient` which returns promise.

@vparfonov, @skabashnyuk  please review

### DIFF
--- a/core/platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/AttributeFilter.java
+++ b/core/platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/AttributeFilter.java
@@ -10,6 +10,7 @@
  *******************************************************************************/
 package org.eclipse.che.api.project.server;
 
+import com.google.common.base.MoreObjects;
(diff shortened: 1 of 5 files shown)


core/platform-api/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/AttributeFilter.java
```diff
@@ -179,7 +180,7 @@ private ProjectTypes getProjectTypes(FolderEntry module, ProjectConfig moduleCon
             if (factory == null) {
                 List<String> value = projectConfig.getAttributes().get(attribute.getName());
 
-                persistentAttributes.put(variable.getName(), value);
+                persistentAttributes.put(variable.getName(), value == null ? new ArrayList<>() : value);
```
skabashnyuk: firstNotNull?

PR #523, opened 2016-02-25
Exclude non-source directories from java project

vzhukovs:
If java project (not maven) has language attribute, then we search source folder definition in attributes and if no one was found then dont mark folder as source folder.

Relevant issue: **CHE-589** _(Broken project tree)_.

@vparfonov

### DIFF
--- a/plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/project/interceptor/AbstractJavaContentRootInterceptor.java
+++ b/plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/project/interceptor/AbstractJavaContentRootInterceptor.java
@@ -75,12 +75,12 @@ public abstract class AbstractJavaContentRootInterceptor implements NodeIntercep
         final ProjectConfigDto projectConfig = folderNode.getProjectConfig();
 
         String srcFolder = _getSourceFolder(projectConfig, getSrcFolderAttribute());
-        if (folderNode.getStorablePath().endsWith(srcFolder)) {
+        if (srcFolder != null && folderNode.getStorablePath().endsWith(srcFolder)) {
             return ContentRoot.SOURCE;
         }
 
         String testSrcFolder = _getSourceFolder(projectConfig, getTestSrcFolderAttribute());
-        if (folderNode.getStorablePath().endsWith(testSrcFolder)) {
+        if (testSrcFolder != null && folderNode.getStorablePath().endsWith(testSrcFolder)) {
             return ContentRoot.TEST_SOURCE;
         }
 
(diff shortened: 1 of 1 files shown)


plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/project/interceptor/AbstractJavaContentRootInterceptor.java
```diff
@@ -88,30 +88,17 @@ private ContentRoot getSourceType(Node node) {
     }
 
     private String _getSourceFolder(ProjectConfigDto projectConfig, String srcAttribute) {
-        Map<String, List<String>> attributes = projectConfig.getAttributes();
+        final Map<String, List<String>> attributes = projectConfig.getAttributes();
         if (!attributes.containsKey(srcAttribute)) {
-            return "";
+            return null;
```
sleshchenko, line 93: Why do you return "" and null when srcAttribute doesn't have value? 
https://github.com/eclipse/che/pull/523/files#diff-6063db316be1e2733d888f9686ae3133R98

sleshchenko:
Also, Why do you not add number of issue to commit message?

vzhukovs:
I've added the issue number, see the description.

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/project/interceptor/AbstractJavaContentRootInterceptor.java
vzhukovs, line 93: Because maven project (which is java simultaneously) should have source folder definition, in this case it will be defined or empty, if empty, then we'll mark root folder as source. If project is java (not maven), then source folder may not be defined, if so, then we don't mark any folder as source.

sleshchenko:
Oh, Yeah. Sorry. It is unusually to see number of issue in body of commit message. Earlier we added it to subject of commit message

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/project/interceptor/AbstractJavaContentRootInterceptor.java
vzhukovs, line 93: It's a temporary solution, which I think will be refactored after GA.

sleshchenko, line 93: Thanks for explanation

vparfonov:
ok

PR #527, opened 2016-02-25
CHE-12: Fix NPE when there is no template

dimashnurenko:
@vparfonov @RomanNikitenko @benoitf


vparfonov:
ok

benoitf:
we have Collections.emptyList() instead of Arrays.asList()

PR #536, opened 2016-02-26
Silky smooth factory workflow

benoitf:


mshaposhnik:
It's a part of moving factory acceptance to the dashb ?
So the perUser/perClick/perAccount policies will be managed in other place ?

benoitf:
yes
https://github.com/codenvy/dashboard/blob/master/src/app/factories/load-factory/load-factory.controller.js#L80-L95

mshaposhnik:
ok

skabashnyuk:
@vparfonov ?

vparfonov:
ok

skabashnyuk:
ok

PR #538, opened 2016-02-26
CHE-514: fix jdt tests

svor:


vparfonov:
ok

plugins/plugin-java/che-plugin-java-ext-jdt/che-jdt-ext-machine/src/test/resources/projects/test/src/main/java/com/codenvy/test/MyClass.java
```diff
@@ -9,7 +9,7 @@
  *   Codenvy, S.A. - initial API and implementation
  *******************************************************************************/
 
-package org.eclipse.che.test;
+package com.codenvy.test;
```
skabashnyuk, line 13: why?

svor, line 13: I've already fixed it.

skabashnyuk:
ok

PR #540, opened 2016-02-26
Add promise based getRepository method to the GithubServiceClient

voievodin:
@skabashnyuk please review


mkuznyetsov:
ok

skabashnyuk:
ок

plugins/plugin-github/che-plugin-github-ext-github/src/main/java/org/eclipse/che/ide/ext/github/client/GitHubClientService.java
```diff
@@ -41,10 +41,22 @@
...
 
     /**
+     * Get given repository information.
+     *
+     * @param user
+     *         the owner of the repository.
+     * @param repository
+     *         the repository name.
+     */
+    Promise<GitHubRepository> getRepository(String user, String repository);
```
vzhukovskii: Describe return with `@return`

PR #544, opened 2016-02-26
Disappear JsOauthWindow after action is performed

voievodin:
**_2 Upvotes**_ The problem is that if url was `/ws/wss` then it wouldn't work.

@vparfonov, @azatsarynnyy, @vzhukovskii  please review


vzhukovs:
:+1:

azatsarynnyy:
Brilliantly :fire:

vparfonov:
+1

PR #546, opened 2016-02-26
CHE-622 [dashboard] Do not import templates commands if user is providing its own location

benoitf:
**_1 Upvote**_ Change-Id: I603080bb67ef9bc8b64c80c06ef47ada3dfb6bfe


ashumilova:
+1

PR #557, opened 2016-02-28
Fix use of the wrong gson adapter

AndrienkoAleksandr:
@sleshchenko @vparfonov @skabashnyuk


skabashnyuk:
Ok

sleshchenko:
OK

PR #570, opened 2016-02-29
Add TomEE 1.7.3 stack and sample application

jgallimore:
**_2 Upvotes**_ Signed-off-by: Jonathan Gallimore jgallimore@tomitribe.com

This PR adds TomEE 1.7.3 as a stack, and a Tomitribe AngularJS / JAX-RS sample application as an example. Please let me know if you have any issues.

### DIFF
--- a/core/ide/che-core-ide-stacks/src/main/resources/predefined-stacks.json
+++ b/core/ide/che-core-ide-stacks/src/main/resources/predefined-stacks.json
@@ -696,5 +696,87 @@
       "name": "type-java.svg",
       "mediaType": "image/svg+xml"
     }
-  }
+  },
+  {
+    "id": "tomee-default",
+    "creator": "tomitribe",
+    "name": "TomEE",
+    "description": "Default Java Stack with JDK 8, Maven and Apache TomEE 1.7.3.",
+    "scope": "general",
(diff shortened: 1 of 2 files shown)


slemeur:
+1: Nice contribution. Glad to see TomEE in Che's default stacks/samples

core/ide/che-core-ide-stacks/src/main/resources/predefined-stacks.json
```diff
@@ -696,5 +696,87 @@
...
       "mediaType": "image/svg+xml"
     }
-  }
+  },
+  {
+    "id": "tomee-default",
+    "creator": "tomitribe",
+    "name": "TomEE",
+    "description": "Default Java Stack with JDK 8, Maven and Apache TomEE 1.7.3.",
+    "scope": "general",
```
benoitf: it should be "advanced"  instead of "general" where we only have one per technology : java, php, node, etc

jgallimore: Hi @benoitf, Thanks for the feedback. I have updated!

benoitf:
+1

PR #587, opened 2016-03-01
CHE-661 unzip in quiet mode

benoitf:
Change-Id: If3a51cc6e01e243c47033d4d891858e3997ab725


riuvshin:
if we need that in codenvy hosted we need to change it here aswell:
https://github.com/codenvy/deployment/blob/master/puppet/modules/all_in_one/templates/machine.properties.erb#L26

https://github.com/codenvy/deployment/blob/master/puppet/modules/multi_server/templates/api_instance/machine.properties.erb#L29

benoitf:
I can issue PR on erb files

garagatyi:
Why you want to make it quiet?

riuvshin:
@garagatyi on a meeting we've decided to get rid of that useless output when ws is starting.

benoitf:
@garagatyi because it looks ugly to see that when we start agent and if unzip fails we will have error BTW

garagatyi:
than ok

garagatyi:
LGTM

benoitf:
I will use -q and not -qq param

PR #590, opened 2016-03-01
fix NPE because of incorrect usage of firstNonNull

garagatyi:
**_1 Upvote**_ Signed-off-by: Alexander Garagatyi agaragatyi@codenvy.com

@skabashnyuk @benoitf


benoitf:
+1

skabashnyuk:
ок

PR #611, opened 2016-03-02
CHE-676. Keep focus on dialog when user uses tab button

RomanNikitenko:
1. Keep focus on dialog when user uses tab button
2. Change behavior for focus() for org.eclipse.che.ide.ui.window.Window: https://github.com/eclipse/che/blob/master/core/ide/che-core-ide-ui/src/main/java/org/eclipse/che/ide/ui/window/Window.java#L190
   Now it set focus on last focused element if such exist: https://github.com/eclipse/che/blob/master/core/ide/che-core-ide-ui/src/main/java/org/eclipse/che/ide/ui/window/View.java#L298
   But if you display dialog, last focused elemet = null, so it's impossible to set focus in this use case.
   After my changes it will set focus on the first child of content panel if such exists. Otherwise it will set focus on first child of footer


vparfonov:
ok

PR #620, opened 2016-03-02
CHE-435 [dashboard] Introduce meaningful warning if browser is not being able ping ws agent

benoitf:
**_2 Upvotes**_ (also we decrease the number of loops because now server side only send running event if ws agent is pingable from workspace master)

Change-Id: Icec5dc28c20643d87f39236a60300706c678c761


ashumilova:
+1

akurinnoy:
+1

PR #622, opened 2016-03-03
CHE-653: Change value fails with NullPointerException

mmorhun:
**_1 Upvote**_ Fix change value error when value located inside a array


mmorhun:
@vparfonov, check please

mmorhun:
We need to check this additional condition, because we must add array index(es) in the same path cell

vparfonov:
+1

PR #630, opened 2016-03-03
CHE-642. Correct handling buttons on dialogs when Enter is pushed

RomanNikitenko:
1. Create class which allows to track the focus for widgets.
2. Fix incorrect behavior when user uses "Enter" button to confirm his choice for:
   - Full text search dialog
   - Choice dialog
   - Confirm dialog
   - Input dialog

### DIFF
--- a/core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/ui/WidgetFocusTracker.java
+++ b/core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/ui/WidgetFocusTracker.java
@@ -1,0 +1,82 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.ide.ui;
+
+import com.google.gwt.event.dom.client.BlurEvent;
+import com.google.gwt.event.dom.client.BlurHandler;
+import com.google.gwt.event.dom.client.FocusEvent;
+import com.google.gwt.event.dom.client.FocusHandler;
+import com.google.gwt.user.client.ui.FocusWidget;
+import com.google.inject.Singleton;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static com.google.common.base.MoreObjects.firstNonNull;
+
+/**
+ * The class contains business logic which allows to track the focus for widgets.
+ *
+ * @author Roman Nikitenko
+ */
+@Singleton
+public class WidgetFocusTracker {
+
+    private Map<FocusWidget, Boolean> focusStates;
+
+    /**
+     * Add widget to track the focus
+     *
+     * @param widget
+     *         the widget to track
+     */
+    public void subscribe(final FocusWidget widget) {
+        if (focusStates == null) {
+            focusStates = new HashMap<>();
+        }
+
+        focusStates.put(widget, false);
+        widget.addFocusHandler(new FocusHandler() {
+            @Override
+            public void onFocus(FocusEvent event) {
+                focusStates.put(widget, true);
+            }
+        });
+
+        widget.addBlurHandler(new BlurHandler() {
+            @Override
+            public void onBlur(BlurEvent event) {
+                focusStates.put(widget, false);
+            }
+        });
+    }
+
+    /**
+     * Unsubscribe widget from tracking the focus
+     *
+     * @param widget
+     *         the widget to unsubscribe from tracking the focus
+     */
+    public void unSubscribe(FocusWidget widget) {
+        if (widget == null) {
+            return;
+        }
+        focusStates.remove(widget);
+    }
+
+    /**
+     * Returns {@code true} if widget is in the focus and {@code false} - otherwise.
+     * Note: this method returns {@code false} if widget hasn't subscribed to track the focus.
+     */
+    public boolean isWidgetFocused(FocusWidget widget) {
+        return widget != null && firstNonNull(focusStates.get(widget), false);
+    }
+}
--- a/core/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/search/FullTextSearchView.java
+++ b/core/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/search/FullTextSearchView.java
@@ -31,6 +31,9 @@ public interface FullTextSearchView extends View<FullTextSearchView.ActionDelega
 
         /** Set focus to root window. */
         void setFocus();
+
+        /** Called when Enter clicked */
+        void onEnterClicked();
     }
 
     /** Shows error message on the view */
@@ -42,6 +45,17 @@ public interface FullTextSearchView extends View<FullTextSearchView.ActionDelega
     /** Set focus to root window. */
     void setFocus();
 
+    /** Returns {@code true} if accept button is in the focus and {@code false} - otherwise. */
+    boolean isAcceptButtonInFocus();
+
+    /** Returns {@code true} if Cancel button is in the focus and {@code false} - otherwise. */
+    boolean isCancelButtonInFocus();
+
+    /** Returns {@code true} if 'select path' button is in the focus and {@code false} - otherwise. */
+    boolean isSelectPathButtonInFocus();
+
+    void showSelectPathDialog();
+
     /** Close dialog. */
     void close();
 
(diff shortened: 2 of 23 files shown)


core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/ui/WidgetFocusTracker.java
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * The class contains business logic which allows to track the focus for widgets.
+ *
+ * @author Roman Nikitenko
+ */
+@Singleton
+public class WidgetFocusTracker {
+
+    private Map<FocusWidget, Boolean> storage;
```
sleshchenko: WDYT about naming this variable as widgetsToFocusStates?

core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/ui/WidgetFocusTracker.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * Unsubscribe widget from tracking the focus
+     *
+     * @param widget
+     *         the widget to unsubscribe from tracking the focus
+     */
+    public void unSubscribe(FocusWidget widget) {
+        if (widget == null || !storage.containsKey(widget)) {
```
sleshchenko: I think you can remove this check at all

core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/ui/WidgetFocusTracker.java
```diff
@@ -0,0 +1,78 @@
...
+            return;
+        }
+        storage.remove(widget);
+    }
+
+    /** Returns is the widget in the focus */
```
sleshchenko: I think it would be better to describe that method return false if it widget is not subscribed

core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/ui/WidgetFocusTracker.java
```diff
@@ -0,0 +1,78 @@
...
+    public boolean isWidgetFocused(FocusWidget widget) {
+        return widget == null || !storage.containsKey(widget) ? false : storage.get(widget);
```
sleshchenko: it looks better for me

```
Boolean focusState = storage.get(widget);
return focusState == null ? false : focusState;
```

core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/ui/WidgetFocusTracker.java
RomanNikitenko: I don't think so - map throws NullPointerException if the specified key is null and this map does not permit null keys

core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/ui/WidgetFocusTracker.java
RomanNikitenko: a bit confusing for me, WDYT about focusStates?

core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/ui/WidgetFocusTracker.java
RomanNikitenko: What about usecase when widget is null? map throws NullPointerException if the specified key is null

core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/ui/WidgetFocusTracker.java
RomanNikitenko: agree, I'll fix it

core/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/search/FullTextSearchView.java
```diff
@@ -42,6 +45,17 @@
     /** Set focus to root window. */
     void setFocus();
 
+    /** Returns is accept button in the focus */
+    boolean isAcceptButtonInFocus();
+
+    /** Returns is cancel button in the focus */
```
azatsarynnyy: I think that this sentence (and other similar sentences in this PR) grammatically is incorrect.
WDYT about something like:

> Returns true if cancel button is in the focus and false - otherwise.

core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/ui/WidgetFocusTracker.java
sleshchenko: For now it is hard to understand function of this variable

core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/ui/WidgetFocusTracker.java
sleshchenko: But you use HashMap. It doesn't throw NPE in this case

core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/ui/WidgetFocusTracker.java
sleshchenko: You know your implementation of Map. But OK about check widget on null. But what about !storage.containsKey(widget)?

core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/ui/WidgetFocusTracker.java
RomanNikitenko: I use interface Map and we can have  NPE if somebody change implementation of subscribe() method

core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/ui/WidgetFocusTracker.java
RomanNikitenko: So - do you mean focusStates - hard to understand?

sleshchenko: No, I mean It is hard to understand Map<FocusWidget, Boolean> storage;
What does this storage store? FocusWidget and Boolean... What does Boolean store?

RomanNikitenko: Did you see my comment: WDYT about focusStates?

sleshchenko: Sorry. I missed it. BTW focusStates is more better than storage :smile:

core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/ui/WidgetFocusTracker.java
RomanNikitenko: I use interface Map and we can have NPE if somebody change implementation of subscribe() method
About !storage.containsKey(widget)  - agree - this check is unnecessary - I'll fix it

core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/ui/WidgetFocusTracker.java
voievodin: I have several remarks.

@sleshchenko suggested a variant where `null` widget conisdered as _non-focused_.

```
Boolean focusState = storage.get(widget);
return focusState == null ? false : focusState;
```

If it is right -  method behaviour i would say that the following variant is the best here:

```
 return widget != null && firstNonNull(map.get(widget), false);
```

RomanNikitenko: I'm sorry, but  according to the original variant  widget conisdered as non-focused if it is null: 
return widget == null || !storage.containsKey(widget) ? false : storage.get(widget);

core/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/search/FullTextSearchView.java
RomanNikitenko: thank you, I've fixed it.

RomanNikitenko:
@vparfonov WDYT?

vparfonov:
ok

PR #648, opened 2016-03-07
CHE-656 Fix switching between editor tabs by hotKeys after rename opened file

ghost:


vinokurig:
@pankorez thank you for making a pull request, but we have already fixed this problem by using List instead of Map for collecting opened editors.

PR #650, opened 2016-03-07
CHE-503: Add ability to save edited changes from compare widget

vinokurig:
@vparfonov please review


plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/CompareViewImpl.java
```diff
@@ -100,10 +120,10 @@ public void show(String oldContent, String newContent, String revision, String f
 
         super.show();
 
-        this.revision.setText(revision);
+        this.revision.setText(revision + " (Read-only)");
```
vzhukovs: You can move this string to the locale constants.

vinokurig: done

vparfonov:
ok

PR #652, opened 2016-03-07
CHE-666 Refine icons in project configuration wizard

vitaliy-guliy:


ashumilova:
+1

PR #659, opened 2016-03-09
CHE-176. Set 'accept' button in the focus for Choice dialog and Confirm dialog

RomanNikitenko:
1. Set 'accept' button in the focus for Choice and Confirm dialogs.
2. Fix method show() of org.eclipse.che.ide.ui.window.Window 
   for correct setting focus on focusable element.


vparfonov:
ok

PR #663, opened 2016-03-09
CHE-673. Correct handling buttons of Commands dialog when Enter is clicked

RomanNikitenko:
1. Correct handling buttons of Commands dialog when Enter is clicked
2. Fix incorrect definition workspace ID in after start new workspace:
   Now workspace ID is  determined when presenter is created, 
   after my changes it will be determined at showing of dialog.


vparfonov:
ok

PR #666, opened 2016-03-09
Avoid NPE if event don't has EventOrigin annotation

vparfonov:
Signed-off-by: Vitalii Parfonov vparfonov@codenvy.com
@skabashnyuk


garagatyi:
LGTM
But must be changed to throwing runtime exception when dto will handle annotations correctly

PR #654, opened 2016-03-08
Upgrade Apache Commons Collections to v3.2.2

ProgramMax:
Version 3.2.1 has a CVSS 10.0 vulnerability. That's the worst kind of
vulnerability that exists. By merely existing on the classpath, this
library causes the Java serialization parser for the entire JVM process
to go from being a state machine to a turing machine. A turing machine
with an exec() function!

https://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2015-8103
https://commons.apache.org/proper/commons-collections/security-reports.html
http://foxglovesecurity.com/2015/11/06/what-do-weblogic-websphere-jboss-jenkins-opennms-and-your-application-have-in-common-this-vulnerability/

PR #667, opened 2016-03-09
Upgrade Apache Commons Collections to v3.2.2

ProgramMax:
Signed-off-by: Chris Blume cblume@google.com


benoitf:
New dependency needs to be approved by IP team first
we need to get a PB on https://dev.eclipse.org/ipzilla/show_bug.cgi?id=10385 first

skabashnyuk:
We don't in this case. See more https://github.com/eclipse/che/pull/654#issuecomment-194307588

benoitf:
Here is PB https://dev.eclipse.org/ipzilla/show_bug.cgi?id=11073

riuvshin:
@benoitf it's resources, why we need CQ for that?

benoitf:
@skabashnyuk @riuvshin you are right, we don't need CQ

benoitf:
BTW we can still merge the update of the test example

PR #676, opened 2016-03-10
Remove unused properties. Rename project_template.location.dir

ghost:

@vparfonov @skabashnyuk please take a look

I have removed an unused property that used to make it possible to load project templates from a local file system. Also, port range properties were removed as they are heritage of 3.x.


vparfonov:
ok

PR #685, opened 2016-03-11
CODENVY-26: Fix stream closing before it is direct use

akorneta:
@skabashnyuk @garagatyi 

Why do we need this fix:
since this method does not use alone it is used together with [this](https://github.com/eclipse/che/blob/master/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java#L692) method so we do not need to close connection until the file from one machine will not be fully written in another machine.


garagatyi:
LGTM @skabashnyuk WDYT?

skabashnyuk:
ok

skabashnyuk:
ok

PR #687, opened 2016-03-11
CHE-457: add servers conf and env vars to model of machine config.

garagatyi:
This PR adds new fields to machine config: servers, envVariables. So it will be possible to add more configuration information directly into 
EnvVariables is map of strings where key is envVariable name and value is env variable value.
Servers is list of objects that represent config of servers in machine. Its structure is:

```
{
    "ref" : "some reference",
    "port" : "8080/tcp",
    "protocol" : "http"
}
```

Ref is text reference to specific server.
Port is port of the server that should be exposed in machine. It includes transport protocol suffix: `/tcp` or `/udp`.
Protocol is protocol that will be used to build url of server.

After merging this PR API will use port in form 8080/tcp or 8080/udp instead of current variant 8080 or 8080/udp.

@skabashnyuk @evoevodin @gazarenkov @sleshchenko @vparfonov @ashumilova @benoitf @dimasnurenko @vzhukovskii


skabashnyuk:
ok

gazarenkov:
OK

sleshchenko:
ok

dimashnurenko:
ok

vparfonov:
ok

vzhukovs:
ok

PR #692, opened 2016-03-12
CODENVY-43. Create Profile menu for Che

RomanNikitenko:
1. Create Profile menu for Che
2. Move RedirectToDashboardProjects and RedirectToDashboardWorkspaces actions from plugins project


vparfonov:
ok

PR #701, opened 2016-03-13
add support for VMware Fusion

zaps:
This is a small modification to the `che.sh` startup script that adds an optional `--vmware` flag to have docker-machine use the vmwarefusion driver instead of the virtualbox driver, for those users that don't have or want to install VirtualBox.

### DIFF
--- a/assembly/assembly-main/src/assembly/bin/che.sh
+++ b/assembly/assembly-main/src/assembly/bin/che.sh
@@ -31,6 +31,7 @@ init_global_variables () {
   USAGE="
 Usage:
   che [OPTIONS] [COMMAND]
+     -v,        --vmware             Use the docker-machine VMware driver (instead of VirtualBox)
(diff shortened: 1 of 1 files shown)


TylerJewell:
@zaps - so awesome that you did this.  I think if you check the archives there have been a few requests for this.

Two thoughts:
1. I'd like to have a more generic option like --machine:vbox, --machine:fusion with the default being vbox.
1. Also, since we are an eclipse project, we need you to fill out a contributor agreement (CLA).  You can see what you have to do in the CONTRIBUTING.md file that is in the root of the project.  It's a simple form.

zaps:
@TylerJewell: 
1. You read my mind that was the next thing I was planning :)
2. I'm on it! Having some troubles amending my git commits with the right email but I'll get it!

zaps:
I really seem to have messed up this commit amending real good-- I'm going to close this PR out and refork 
Sorry!!

PR #702, opened 2016-03-14
add VMware Fusion support

zaps:
This is a small modification to the che.sh startup script that adds an optional --vmware flag to have docker-machine use the vmwarefusion driver instead of the virtualbox driver, for those users that don't have or want to install VirtualBox.

### DIFF
--- a/assembly/assembly-main/src/assembly/bin/che.sh
+++ b/assembly/assembly-main/src/assembly/bin/che.sh
@@ -31,30 +31,31 @@ init_global_variables () {
   USAGE="
 Usage:
   che [OPTIONS] [COMMAND]
-     -v,        --vmware             Use the docker-machine VMware driver (instead of VirtualBox)
-     -m:name,   --machine:name       For Win & Mac, sets the docker-machine VM name; default=default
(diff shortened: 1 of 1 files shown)


TylerJewell:
@zaps CLA looks good.  
Let's go ahead and change this to --machine:fusion and --machine:vbox as the configured option with vbox as the default for mac / windows.  We will then accept it.

zaps:
Will do.
I noticed there already is a `--machine` switch for specifying the name of the machine so right now, running for example `che.sh --machine:vbox` will create a docker-machine with name 'vbox'.

Instead, could we add another option, maybe `--machine-driver` to specify the driver type? 
I.e. `--machine-driver:fusion`. If the option is not given it would default to vbox.

Otherwise, we'd have to do something like `--machine:name=myCheMachine,driver=fusion`. And again if the argument isn't provided it would default to default/vbox.

TylerJewell:
Ahhh didn't notice that --machine was already used.  Yes `--machine-driver:vbox` and `--machine-driver:fusion` is better approach.

zaps:
ok all set!

TylerJewell:
Looks good to me. Past all my tests.  Congrats on the first merge.

PR #706, opened 2016-03-14
CHE-656: Refactor getOpenEditors method from Map to List

vinokurig:
@vparfonov please review

### DIFF
--- a/core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/EditorAgent.java
+++ b/core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/EditorAgent.java
@@ -14,10 +14,10 @@ import com.google.gwt.user.client.rpc.AsyncCallback;
 
 import org.eclipse.che.commons.annotation.Nullable;
 import org.eclipse.che.ide.api.project.tree.VirtualFile;
+import org.eclipse.che.ide.resource.Path;
 
 import javax.validation.constraints.NotNull;
 import java.util.List;
-import java.util.Map;
 
 /**
  * Editor Agent manages Editors, it allows to open a new editor with given file,
@@ -63,10 +63,19 @@ public interface EditorAgent {
     /**
      * Get all opened editors
      *
-     * @return map with all opened editors
+     * @return list with all opened editors
      */
     @NotNull
-    Map<String, EditorPartPresenter> getOpenedEditors();
+    List<EditorPartPresenter> getOpenedEditors();
+
+    /**
+     * Get opened editor by related file path
+     *
+     * @param path path of the file opened in editor
+     * @return opened editor or null if it does not exist
+     */
+    @Nullable
+    EditorPartPresenter getOpenedEditor(Path path);
 
     /**
      * Saves all opened files whose content have changed since the last save operation
--- a/core/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/actions/SaveAllAction.java
+++ b/core/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/actions/SaveAllAction.java
@@ -24,8 +24,6 @@ import org.eclipse.che.ide.api.editor.EditorPartPresenter;
 import org.eclipse.che.ide.api.editor.EditorWithAutoSave;
 import org.eclipse.che.ide.util.loging.Log;
 
-import java.util.ArrayList;
-import java.util.Collection;
 import java.util.List;
 
 /** @author Evgen Vidolob */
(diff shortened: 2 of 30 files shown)


core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/EditorAgent.java
```diff
@@ -63,10 +62,19 @@
...
+    List<EditorPartPresenter> getOpenedEditors();
+
+    /**
+     * Get opened editor by related file path
+     *
+     * @param path path of the file opened in editor
+     * @return opened editor or null if it does not exist
+     */
+    @Nullable
+    EditorPartPresenter getOpenedEditor(String path);
```
vzhukovs: I think that passing `VirtualFile` as param to find the concrete editor will be better. Then to find specific editor you can compare virtual files with given one.

core/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/actions/SaveAllAction.java
```diff
@@ -84,7 +80,7 @@ public void onSuccess(EditorInput result) {
     public void updateProjectAction(ActionEvent e) {
 //        e.getPresentation().setVisible(true);
```
vzhukovs: Remove unneccessary commented code, plz.

core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/EditorAgent.java
vitaliy-guliy: I prefer using the path. Why not?

vzhukovs: Because you can pass as string anything, when you're expecting specific file, then you couldn't pass any rubbish. It's easier to maintain the code.

vitaliy-guliy: Using the string is easier and understandable. Using the VirtualFile needs to write the special comparator to find. And what shall I do when I need to find an editor by the file path?

vzhukovs: Then use [Path.java](https://github.com/eclipse/che/blob/master/core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/resource/Path.java). We have specialized class to work with paths.

vitaliy-guliy: -1

vzhukovs: Arguments?

`Path path = Path.valueOf("/some/path")` no?

vitaliy-guliy: Why do we need Path.java ? Where is it used? Why did you decide to write so heavy class and used it only once ?

ashumilova: +1 for String path parameter

vzhukovs: > Why did you decide to write so heavy class and used it only once ?

I use it in my branch everywhere, where it possible. It provides usefull support for paths. With less code writing. IMO.

For example, if you have smth like `/path/to/file` and you want to get project from this path, you should create bycicles with substrings, but your path may contains leading slash or not, may contains trailing slash or not. With this utility class you can simple do this: `String project = Path.valueOf("path/to/file").segment(1);` and get need value. Or if you want to get parent of the specified path, you just do this: `Path.valueOf("/some/very/huge/path/to/file").parent()` and get `/some/very/huge/path/to`.

Take a look on eclipse sources, how the developsers use it, and you'll understand. It really helpful for maintaining the resources.

_Pros from using Path is that there is a basic path validation for input value and you wont be able to pass invalid value as file path_

-1 for String path parameter

vitaliy-guliy: Ok.
+1 for String path parameter

core/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/actions/SaveAllAction.java
vinokurig: done

core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/EditorAgent.java
vinokurig: I prefer to use  String path parameter. There are many methods where getOpenedEditor(String path) used and this path directly received from their parameters.
@vparfonov what do you think?

vparfonov:
I like idea with Path, looks useful (similar to java.nio.file). But i don't insist.

vparfonov:
ok

PR #708, opened 2016-03-14
CHE-740 Connect via SSH button in Consoles has no tooltip

vitaliy-guliy:


vparfonov:
+1

Issue #678, opened 2016-03-10
Unable to create project (project already exists)


rhopp:
I'm unable to create project in che. I'm running it natively (bin/che.sh). After some digging i found that I'm unable to read/write into /projects folder in workspace container. I thought that it would be UID mismatch between host and workspace container, but it does not seem like that:

In workspace container:

```
user@62f51e408cb5:/$ id                                                                                                                                                                                                                        
uid=1000(user) gid=1000(user) groups=1000(user),27(sudo),100(users)                                                                                                                                                                            
user@62f51e408cb5:/$ ls -lan                                                                                                                                                                                                                   
total 80                                                                                                                                                                                                                                       
drwxr-xr-x.  23    0    0 4096 Mar 10 13:45 .                                                                                                                                                                                                  
drwxr-xr-x.  23    0    0 4096 Mar 10 13:45 ..                                                                                                                                                                                                 
-rwxr-xr-x.   1    0    0    0 Mar 10 13:45 .dockerenv                                                                                                                                                                                         
-rwxr-xr-x.   1    0    0    0 Mar 10 13:45 .dockerinit                                                                                                                                                                                        
drwxr-xr-x.   2    0    0 4096 Feb 26 19:22 bin                                                                                                                                                                                                
drwxr-xr-x.   2    0    0 4096 Apr 10  2014 boot                                                                                                                                                                                               
drwxr-xr-x.   5    0    0  360 Mar 10 13:45 dev                                                                                                                                                                                                
drwxr-xr-x.  73    0    0 4096 Mar 10 13:45 etc                                                                                                                                                                                                
drwxr-xr-x.   3    0    0 4096 Mar  1 14:00 home                                                                                                                                                                                               
drwxr-xr-x.  12    0    0 4096 Mar  1 13:57 lib                                                                                                                                                                                                
drwxr-xr-x.   2    0    0 4096 Feb 26 19:21 lib64                                                                                                                                                                                              
drwxrwxr-x.   2 1000 1000 4096 Mar 10 13:51 local-storage                                                                                                                                                                                      
drwxr-xr-x.   2    0    0 4096 Feb 26 19:21 media                                                                                                                                                                                              
drwxr-xr-x.   3    0    0 4096 Mar 10 13:45 mnt                                                                                                                                                                                                
drwxr-xr-x.   3    0    0 4096 Mar  1 13:58 opt                                                                                                                                                                                                
[...]

PR #712, opened 2016-03-15
Add :Z suffix when mounting workspace project sources

ghost:
A user has reported an issue with creating projects in Che - https://github.com/eclipse/che/issues/678

It turned out that SELinux has caused that on RHEL. Adding :Z suffix fixes the issue (similarly to what's been done to mount ws agent and terminal)

@garagatyi @skabashnyuk please take a look.


garagatyi:
LGTM

skabashnyuk:
ok

PR #720, opened 2016-03-15
Fix NPE at cleaning selections in processes panel

RomanNikitenko:
We have NPE when workspace is stopped and we clear selections in processes panel.
@vparfonov 


vparfonov:
ok

PR #721, opened 2016-03-15
CHE-731. Close all opened files when workspace is stopped

RomanNikitenko:
@vparfonov 


vparfonov:
ok

PR #723, opened 2016-03-15
Add an ability to get pull requests by repo, owner and head

voievodin:
Provides an ability to get a pull request from `user1/repo` to `user2/repo` instead of getting pull request by number. The client of this method(such as contribution plugin) needs this improvement because the only way of getting the pull request from one repository to another is getting all the pull requests and then matching them. In that case when repository has 10+ open pull requests it takes some time to get all of the pull requests and parse those into dtos, on the other hand `GitService` client makes the same request to get all the pull requests returned by server. As a result we had 2 huge requests.

I think here is the only one way for adding additional query parameter to the `kohsuke` github client -  it's custom `HttpConnector` which adds `head` query parameter directly to the url and delegates connect call to the `DEFAULT` implementation.

Here is [api ref](https://developer.github.com/v3/pulls/#list-pull-requests), koksuke library doesn't allow to put head param.

@skabashnyuk please review

### DIFF
--- a/plugins/plugin-github/che-plugin-github-ext-github/src/main/java/org/eclipse/che/ide/ext/github/client/GitHubClientService.java
+++ b/plugins/plugin-github/che-plugin-github-ext-github/src/main/java/org/eclipse/che/ide/ext/github/client/GitHubClientService.java
@@ -151,6 +151,18 @@ public interface GitHubClientService {
     Promise<GitHubPullRequestList> getPullRequests(@NotNull String owner, @NotNull String repository);
 
     /**
+     * Get pull requests for given repository.
+     *
+     * @param owner
+     *         the repository owner.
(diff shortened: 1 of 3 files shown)


skabashnyuk:
For now Ok.
1. Add please comment to remove this code after upgrade to 1.73 github-api
2. Create CQ for upgrade to 1.73 github-api
3. Create issue to remove this patch after 1.73 approval
See more 
https://github.com/kohsuke/github-api/commit/2440a676bdb9e880c3752312bb45ff42ff3f8e6a

voievodin:
1. done
2. [cq request](https://dev.eclipse.org/ipzilla/show_bug.cgi?id=11098)
3. [issue](https://jira.codenvycorp.com/browse/CHE-758)

PR #729, opened 2016-03-16
CHE-741: change model of machine server conf and runtime server

garagatyi:
Add field 'path' to machine server conf.
Add fields 'path', 'protocol' to runtime machine server.
Field 'url' of runtime machine server now includes path if path is
not null.

@skabashnyuk @evoevodin @vparfonov


core/platform-api/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/ServerConfImpl.java
```diff
@@ -21,17 +21,20 @@
     private final String ref;
     private final String port;
     private final String protocol;
+    private final String path;
 
-    public ServerConfImpl(String ref, String port, String protocol) {
+    public ServerConfImpl(String ref, String port, String protocol, String path) {
```
voievodin, line 26: Did you consider adding `Builder` for this class?

garagatyi, line 26: I think it is still too small to add builder. At least I haven't found a place where I need it.

voievodin:
Ok.

garagatyi:
@ashumilova Can you take a look if something should be changed in dashboard.

skabashnyuk:
ок

ashumilova:
ok

PR #733, opened 2016-03-17
CODENVY-262: Delete containers which left after build with fail

mmorhun:
Deletes broken containers which left after build workspace with fail


mmorhun:
@garagatyi  @skabashnyuk @vparfonov check please

garagatyi:
LGTM

skabashnyuk:
up to you

PR #734, opened 2016-03-17
Fix url encoding in DefaultHttpJsonRequest

akorneta:
@skabashnyuk @evoevodin 
UriBuilder.build()  performs encode it was necessary to get rid of second encoding of query params

### DIFF
--- a/core/platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequest.java
+++ b/core/platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequest.java
@@ -42,13 +42,12 @@ import java.io.OutputStream;
 import java.io.Reader;
 import java.net.HttpURLConnection;
 import java.net.URL;
-import java.net.URLEncoder;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
-import static java.util.Objects.requireNonNull;
 import static com.google.common.base.Strings.isNullOrEmpty;
+import static java.util.Objects.requireNonNull;
 
 /**
  * Simple implementation of {@link HttpJsonRequest} based on {@link HttpURLConnection}.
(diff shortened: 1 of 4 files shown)


core/platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequest.java
```diff
@@ -34,7 +34,6 @@
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriBuilder;
-
```
voievodin, line 37: Why did you remove this line ? [spec](https://github.com/codenvy/codenvy-dev-resources/wiki/Source-file#imports)

core/platform-api/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequest.java
```diff
@@ -201,11 +199,7 @@ DefaultHttpJsonResponse doRequest(int timeout,
 
             if (hasQueryParams) {
                 for (Pair<String, ?> parameter : parameters) {
-                    String name = URLEncoder.encode(parameter.first, "UTF-8");
-                    String value = parameter.second == null ?
-                                   null :
-                                   URLEncoder.encode(String.valueOf(parameter.second), "UTF-8");
-                    ub.queryParam(name, value);
+                    ub.queryParam(parameter.first, parameter.second == null ? null : String.valueOf(parameter.second));
```
voievodin: If you put `parameter.second` to the `ub.queryParam` the behaviour will be the same as with ternary operator

skabashnyuk:
ок

garagatyi:
@akorneta Please run selenium test to ensure everything will work as expected.

voievodin:
ok

PR #740, opened 2016-03-18
Fix year in header of DTO genereted classes and in some project classes

mshaposhnik:
Signed-off-by: Max Shaposhnik mshaposhnik@codenvy.com


skabashnyuk:
ok

garagatyi:
LGTM

PR #743, opened 2016-03-18
CHE-781 Che structural refactoring

skabashnyuk:
Moved  che-core-api-core che-core-api-dto che-core-api-dto-maven-plugin to upper level


vparfonov:
ok

mshaposhnik:
Ok

PR #754, opened 2016-03-21
Add chown to projects dir to make sure a current container user has write access to it

ghost:
There have been a number of user complaints about not being able to create projects in Che if it has been started by root user or any user with UID != 1000.

Proposed solution - run chown on projects dir in the exec so that a current user in a container has write access to `/projects` directory.


ghost:
@garagatyi please take a look

garagatyi:
LGTM
@skabashnyuk WDYT?

skabashnyuk:
ok

PR #755, opened 2016-03-21
CHE-730. Fix problems related to refusing WebSocket connection when machine is restarted.

RomanNikitenko:
1. We create new instance of  MachineMessageBus and try to connect via WebSocket connection when machine is started.
   But old instance tries to reconnect at this moment by timer, so we need to cancel all attempts to reconnect.
2. We have double method call and code execution when we restart machine https://github.com/eclipse/che/blob/master/plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/machine/MachineManagerImpl.java#L277:
   - from machine component
   - from handler in machine manager

It causes to create new WebSocket connection twice and creates bugs related to refusing connection.
So we need to check it to avoid these bugs.


vparfonov:
ok

PR #763, opened 2016-03-22
CHE-657: fix generation on docker container name

garagatyi:
Fixes generation of docker container name. Upper-case characters are not allowed.


skabashnyuk:
ок

voievodin:
Ok

PR #764, opened 2016-03-22
CHE-260 Rename some factory parameters

mkuznyetsov:
Rename the following project source parameters:
"project.source.parameters.keepDirectory" to "project.source.parameters.keepDir".
"project.source.parameters.remoteOriginFetch" to "project.source.parameters.fetch"

The old parameter names are not going to work anymore, so all existing workspaces with such projects should be updated

### DIFF
--- a/plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/importer/page/GitImporterPagePresenter.java
+++ b/plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/importer/page/GitImporterPagePresenter.java
@@ -127,12 +127,12 @@ public class GitImporterPagePresenter extends AbstractWizardPage<ProjectConfigDt
         view.enableDirectoryNameField(keepDirectory);
 
         if (keepDirectory) {
-            projectParameters().put("keepDirectory", view.getDirectoryName());
+            projectParameters().put("keepDir", view.getDirectoryName());
             dataObject.withType("blank");
             view.highlightDirectoryNameField(!NameUtils.checkProjectName(view.getDirectoryName()));
             view.focusDirectoryNameFiend();
         } else {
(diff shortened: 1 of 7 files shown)


garagatyi:
what about existing factories? will they work as expected?

skabashnyuk:
AFAIK we are going to provide mongo script for this cas?

garagatyi:
for codenvy mongo script is OK, but Che users may have serialized workspaces. Will this PR broke these workspaces?

mkuznyetsov:
Mongo script is added to Codenvy issue.
however, there is no migration for serialized workspaces.

TylerJewell:
What is the documentation that needs to be updated on docs.codenvy.com or eclipse.org/che?
What are the parameters that are being changed?  

The GitHub PR needs to have a complete set of documents for the history so that future readers of this issue can understand what is being changed and the impact to their system.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/importer/page/GitImporterPagePresenter.java
```diff
@@ -127,12 +127,12 @@ public void keepDirectorySelected(boolean keepDirectory) {
         view.enableDirectoryNameField(keepDirectory);
 
         if (keepDirectory) {
-            projectParameters().put("keepDirectory", view.getDirectoryName());
+            projectParameters().put("keepDir", view.getDirectoryName());
```
vzhukovs, line 130: You forgot to adapt `GithubImporterPagePresenter`, be carefully about that.

mkuznyetsov, line 130: thx, fixed

PR #768, opened 2016-03-22
CHE-339 Review and rework the popups that do not extend Window

vitaliy-guliy:


ashumilova:
Ok

PR #769, opened 2016-03-22
Delete project from workspace's configuration in case project doesn't…

azatsarynnyy:
… contains sources

Signed-off-by: Artem Zatsarynnyi azatsarynnyy@codenvy.com

@vparfonov


vparfonov:
ok

PR #770, opened 2016-03-22
Vadidate username on it's creation

mkuznyetsov:
Username should be usable in URL as path param, so it's being validated by matching a regex

### DIFF
--- a/master/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
+++ b/master/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
@@ -58,6 +58,7 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.core.SecurityContext;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.regex.Pattern;
 
 import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.base.Strings.isNullOrEmpty;
@@ -88,6 +89,9 @@ import static org.eclipse.che.dto.server.DtoFactory.newDto;
 @Api(value = "/user", description = "User manager")
 @Path("/user")
 public class UserService extends Service {
+
+    private static final Pattern VALID_USER_NAME = Pattern.compile("[a-zA-Z0-9][-_.a-zA-Z0-9]*[a-zA-Z0-9]");
+
     @VisibleForTesting
     static final String USER_SELF_CREATION_ALLOWED = "user.self.creation.allowed";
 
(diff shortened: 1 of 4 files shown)


master/che-core-api-user/src/test/java/org/eclipse/che/api/user/server/UserServiceTest.java
```diff
@@ -4,13 +4,12 @@
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-v10.html
- *
+ * <p>
```
sleshchenko: fix license

mkuznyetsov: fixed

master/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
```diff
@@ -472,6 +474,15 @@ private User fromEntity(UserDescriptor userDescriptor) throws BadRequestExceptio
         return user;
     }
 
+    private void checkUserName(String name) throws BadRequestException {
+        if (isNullOrEmpty(name)) {
+            throw new BadRequestException("User name required");
+        }
+        if (!VALID_USER_NAME.matcher(name.toLowerCase()).matches()) {
```
garagatyi, line 481: In that case name could not contain uppercase letters, but regexp includes it

master/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
```diff
@@ -472,6 +474,15 @@ private User fromEntity(UserDescriptor userDescriptor) throws BadRequestExceptio
...
+            throw new BadRequestException("User name must consist of URL allowed characters");
```
garagatyi, line 482: Error message should explain which characters are allowed

master/che-core-api-user/src/test/java/org/eclipse/che/api/user/server/UserServiceTest.java
```diff
@@ -585,6 +617,31 @@ public void shouldBeAbleToGetSettingOfUserService() throws Exception {
...
+                              {"test.123"},
+                              {"TeSt123"}
+        };
+    }
+
+    @DataProvider(name = "invalidUserNames")
+    public static Object[][] invalidUserNames() {
+        return new Object[][]{{"TE ST"},
+                              {""},
+                              {"(test)"},
```
garagatyi, line 635: check for name length is needed too

master/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
sleshchenko, line 481: It's problem of variable name because name will be converted to lower case. But maybe It would be better to convert username to lower case before using instead of checking with uppercase letters and converting after that

PR #773, opened 2016-03-22
CHE-866: wrap fileWatcher.startup() with try..catch.

vparfonov:
Avoid failing ProjectManager startup


skabashnyuk:
ок

PR #775, opened 2016-03-22
CHE-774:fix perspective id & move Action to the better code base

vparfonov:


vparfonov:
ok

PR #776, opened 2016-03-22
Move test according to https://github.com/eclipse/che/pull/775

vparfonov:


vparfonov:
ok

PR #778, opened 2016-03-22
CHE-867: Fix set commands after create template project on dashboard

dimashnurenko:
@vparfonov @ashumilova


vparfonov:
ok @benoitf @ashumilova

benoitf:
why commands need to be added after project is updated ? and not when once we import it ?

dimashnurenko:
Actions after project importing:
1. Project service get workspace clone from storage in order to add projects in workspace (this workspace clone without commands yet)
2. Workspace service get workspace clone from storage in order to add commands in workspace
3. Workspace service add commands and update workspace in storage
4. Project service configured project and update workspace in storage (after that workspace in storage is overridden by workspace from project service which has no commands)

So we have to add commands in workspace after projects will be added and configured.

dimashnurenko:
The issue will be fixed on server side.

PR #782, opened 2016-03-23
CHE-766. Correct handling of pressing 'enter' on some dialogs

RomanNikitenko:
1. Add method isWidgetFocused(FocusWidget widget) in https://github.com/eclipse/che/blob/master/core/ide/che-core-ide-ui/src/main/java/org/eclipse/che/ide/ui/window/Window.java#L43
2. Add checking which button in the focus when 'Enter' is pressed on some dialogs


vparfonov:
ok

PR #783, opened 2016-03-23
CHE-796: browser die when creating a project with a new ws

akurinnoy:
**_2 Upvotes**_ Signed-off-by: Oleksii Kurinnyi okurinnyi@codenvy.com


ashumilova:
+1 nice fix

olexii4:
+1

PR #784, opened 2016-03-23
Remove 'Create Maven module' feature

azatsarynnyy:
@vparfonov


vzhukovs:
What would be instead of this functionality?

vparfonov:
ok

PR #787, opened 2016-03-23
Exclude tests

vparfonov:


riuvshin:
LGTM, but we need to investigate why it's fails on CI

PR #788, opened 2016-03-23
Schedule annotation documentation

skabashnyuk:


voievodin:
Use `java` keyword when you define javacode with triple quote blocks

skabashnyuk:
> Use java keyword when you define javacode with triple quote blocks

done

voievodin:
ok

PR #791, opened 2016-03-23
CHE-15 'Close' button on find/replace panel should be square

vitaliy-guliy:


riuvshin:
LGTM

vparfonov:
ok

PR #801, opened 2016-03-24
CHE-525: Fix unexpected error message after refactoring

dimashnurenko:
**_1 Upvote**_ @vparfonov


dimashnurenko:
@svor

vparfonov:
ok

svor:
:+1:

TylerJewell:
Nice job.

I had a heart attack during the keynote demo as I was never sure what would trigger this bug.  Sometimes it happened in practice.  Other times it didn't!

PR #829, opened 2016-03-25
CHE-553: add ability to view logs on start workspace

akurinnoy:
**_2 Upvotes**_ screenshot

### DIFF
--- a/dashboard/src/app/ide/ide.controller.js
+++ b/dashboard/src/app/ide/ide.controller.js
@@ -20,7 +20,7 @@ class IdeCtrl {
    * Default constructor that is using resource
    * @ngInject for Dependency injection
    */
-  constructor (ideSvc, $routeParams, ideLoaderSvc, ideIFrameSvc, $rootScope, cheWorkspace, $timeout, $location, routeHistory) {
+  constructor(ideSvc, $routeParams, ideLoaderSvc, ideIFrameSvc, $rootScope, cheWorkspace, $timeout, $location, routeHistory) {
     this.ideSvc = ideSvc;
     this.ideIFrameSvc = ideIFrameSvc;
     this.$rootScope = $rootScope;
@@ -36,7 +36,7 @@ class IdeCtrl {
     // search the selected workspace
     let workspace = $routeParams.workspaceName;
     if (!workspace) {
-        this.selectedWorkspaceName = null;
+      this.selectedWorkspaceName = null;
     } else {
       this.selectedWorkspaceName = workspace;
     }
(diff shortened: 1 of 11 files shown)


akurinnoy:
@ashumilova @benoitf  wdyt?

dashboard/src/app/ide/ide.controller.js
```diff
@@ -81,6 +81,14 @@ class IdeCtrl {
 
       let promise = cheWorkspace.fetchWorkspaces();
 
+      if ($routeParams.showLogs) {
+        // remove action from path
+        $location.url('/ide/' + this.selectedWorkspaceName, false);
```
benoitf: I don't think it could be hardcoded. (the value /ide) 
it should be a property like https://github.com/eclipse/che/blob/master/dashboard/src/assets/branding/product.json#L6

dashboard/src/app/ide/ide.controller.js
```diff
@@ -122,8 +129,14 @@ class IdeCtrl {
...
 
     if (this.selectedWorkspace) {
+      if (this.ideSvc.getPreventRedirection() && 'STARTING' === this.selectedWorkspace.status) {
+        this.$rootScope.hideIdeLoader = false;
+        this.ideLoaderSvc.addLoader();
+      }
+
       // now, check if workspace has been started or not
-      if ('RUNNING' === this.selectedWorkspace.status) {
+      else if ('RUNNING' === this.selectedWorkspace.status) {
```
benoitf: I think else should be on same line than previous closing bracket

dashboard/src/app/ide/ide.controller.js
olexii4: If I understand corectly, we should add geter which returns the URL depends on workspace name (it can be near this one https://github.com/eclipse/che/blob/master/dashboard/src/components/api/che-workspace.factory.js#L294) to factory and use it. Because we have a several places are needed  relative IDE url. 

Then, for contextPath(https://github.com/eclipse/che/blob/master/dashboard/src/app/ide/ide.service.js#L258
), we should add a property like https://github.com/eclipse/che/blob/master/dashboard/src/assets/branding/product.json#L6

@benoitf Am I right?

benoitf: @olexii4  correct

ashumilova:
+1 after fixes of mentions above

olexii4:
+1

ashumilova:
OK (logs fixes should also be applied to factory loading)

PR #831, opened 2016-03-25
CHE-746. Split github plugin to several modules client/server/shared

RomanNikitenko:


plugins/plugin-github/che-plugin-github-server/pom.xml
```diff
@@ -0,0 +1,201 @@
...
+            <groupId>org.kohsuke</groupId>
+            <artifactId>github-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
+    </dependencies>
+    <build>
+        <sourceDirectory>src/main/java</sourceDirectory>
```
sleshchenko: Why do you need it?

pom.xml
```diff
@@ -361,6 +361,16 @@
             </dependency>
             <dependency>
                 <groupId>org.eclipse.che.plugin</groupId>
+                <artifactId>che-plugin-github-server</artifactId>
+                <version>${project.version}</version>
```
sleshchenko: remove version

plugins/plugin-github/che-plugin-github-server/pom.xml
RomanNikitenko: We should to write it when we are overriding the default place for source code.
You're right - it isn't this use case, but I prefer to define it clearly.
WDYT?

sleshchenko: I prefer to remove it but up to you

RomanNikitenko: I've removed it.

vparfonov:
ok

PR #851, opened 2016-03-28
CHE-883: cleanup websocket channels when ide is opened

ashumilova:


riuvshin:
Excellent thanks @ashumilova !!!

PR #757, opened 2016-03-21
CHE-233:Rework Project API with new VFS

vparfonov:
Signed-off-by: Vitaly Parfonov vparfonov@codenvy.com

PR #892, opened 2016-03-29
Fixing wrong merge:

vparfonov:
during merging https://github.com/eclipse/che/pull/757 in dame way will be broken FactoryWorkspaceComponent.java. Rollback it to corect state


vparfonov:
ok

PR #914, opened 2016-03-29
CHE-901: use ConcurrentHashMap for projects in ProjectRegistry

vparfonov:


vparfonov:
ok

PR #930, opened 2016-03-30
CHE-884: Add Node.js plugin

dimashnurenko:
@vparfonov @sleshchenko


plugins/plugin-nodejs/che-plugin-nodejs-lang-ide/src/main/java/org/eclipse/che/plugin/nodejs/ide/NodeJsLocalizationConstant.java
```diff
@@ -0,0 +1,27 @@
...
+package org.eclipse.che.plugin.nodejs.ide;
+
+import com.google.gwt.i18n.client.Messages;
+
+/**
+ * @author Dmitry Shnurenko
+ */
+public interface NodeJsLocalizationConstant extends Messages {
+
+    @Key("action.new.node.js.file.title")
```
azatsarynnyy, line 20: these key annotations are never used
so it may be removed

plugins/plugin-nodejs/che-plugin-nodejs-lang-ide/src/main/resources/org/eclipse/che/plugin/nodejs/ide/NodeJsLocalizationConstant.properties
```diff
@@ -0,0 +1,12 @@
+#
+# Copyright (c) 2012-2016 Codenvy, S.A.
```
azatsarynnyy, line 2: it's empty so may be removed

plugins/plugin-nodejs/che-plugin-nodejs-lang-server/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>che-plugin-nodejs-parent</artifactId>
+        <groupId>org.eclipse.che.plugin</groupId>
+        <version>4.1.0-RC1-SNAPSHOT</version>
+    </parent>
+    <artifactId>che-plugin-nodejs-lang-server</artifactId>
+    <name>Che Plugin :: NodeJs :: Extension Server</name>
+    <properties>
+        <findbugs.failonerror>false</findbugs.failonerror>
```
sleshchenko: Does findbugs fail in this module?

vparfonov:
remove findbugs.failonerror = false and ok

PR #941, opened 2016-03-31
Remove unused docker version property

ghost:
We do not check Docker version, so this property isn't used anymore and it causes confusion.  


ghost:
@garagatyi @skabashnyuk please take a look

azatsarynnyy:
I think that [DockerVersionVerifier](https://github.com/eclipse/che/blob/master/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerVersionVerifier.java) should be removed as well

garagatyi:
-1
We still have [class](https://github.com/eclipse/che/blob/867ec652424cd23a587f164f6acf3e8e44158bdf/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerVersionVerifier.java) that uses that property even though it is not used anymore.
If you want to delete that property you should delete this class too.

ghost:
Updated the PR

garagatyi:
LGTM

garagatyi:
@skabashnyuk @vparfonov Please take a look

vparfonov:
ok

PR #944, opened 2016-03-31
CHE-912 returning no value instead of throwing an exception when project is not a SVN repository

stour:
@skabashnyuk @vparfonov


skabashnyuk:
why?

stour:
As described in issue currently you cannot import a Git project if it does not contain a '.svn' subfolder. This is a not an expected behavior, I guess you agree ?
So when the project is not a SVN project we need to return an empty list of values and not throw an exception.

It's the same logic as in [plugin-git](https://github.com/eclipse/che/blob/master/wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitValueProviderFactory.java#L54).

skabashnyuk:
LGTM

vparfonov:
ok

PR #952, opened 2016-04-01
CHE-253: Refactor Che docker client

mmorhun:
**_2 Upvotes**_ Refactor Che docker client to be able to add new parameters without breaking code dependent on docker client

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
@@ -10,12 +10,15 @@
  *******************************************************************************/
 package org.eclipse.che.plugin.docker.client;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.io.CharStreams;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import com.google.gson.reflect.TypeToken;
 
 import org.apache.commons.codec.binary.Base64;
 import org.eclipse.che.api.core.util.FileCleaner;
 import org.eclipse.che.api.core.util.ValueHolder;
+import org.eclipse.che.commons.annotation.Nullable;
 import org.eclipse.che.commons.json.JsonHelper;
 import org.eclipse.che.commons.json.JsonNameConvention;
 import org.eclipse.che.commons.json.JsonParseException;
@@ -27,13 +30,12 @@ import org.eclipse.che.plugin.docker.client.connection.DockerConnection;
 import org.eclipse.che.plugin.docker.client.connection.DockerConnectionFactory;
 import org.eclipse.che.plugin.docker.client.connection.DockerResponse;
 import org.eclipse.che.plugin.docker.client.dto.AuthConfigs;
-import org.eclipse.che.plugin.docker.client.json.ContainerCommited;
+import org.eclipse.che.plugin.docker.client.json.ContainerCommitted;
 import org.eclipse.che.plugin.docker.client.json.ContainerConfig;
 import org.eclipse.che.plugin.docker.client.json.ContainerCreated;
 import org.eclipse.che.plugin.docker.client.json.ContainerExitStatus;
 import org.eclipse.che.plugin.docker.client.json.ContainerInfo;
 import org.eclipse.che.plugin.docker.client.json.ContainerProcesses;
-import org.eclipse.che.plugin.docker.client.json.ContainerResource;
 import org.eclipse.che.plugin.docker.client.json.Event;
 import org.eclipse.che.plugin.docker.client.json.ExecConfig;
 import org.eclipse.che.plugin.docker.client.json.ExecCreated;
@@ -45,6 +47,28 @@ import org.eclipse.che.plugin.docker.client.json.Image;
 import org.eclipse.che.plugin.docker.client.json.ImageInfo;
 import org.eclipse.che.plugin.docker.client.json.ProgressStatus;
 import org.eclipse.che.plugin.docker.client.json.Version;
+import org.eclipse.che.plugin.docker.client.params.AttachContainerParams;
+import org.eclipse.che.plugin.docker.client.params.BuildImageParams;
+import org.eclipse.che.plugin.docker.client.params.CommitParams;
+import org.eclipse.che.plugin.docker.client.params.CreateContainerParams;
+import org.eclipse.che.plugin.docker.client.params.CreateExecParams;
+import org.eclipse.che.plugin.docker.client.params.GetEventsParams;
+import org.eclipse.che.plugin.docker.client.params.GetExecInfoParams;
+import org.eclipse.che.plugin.docker.client.params.GetResourceParams;
+import org.eclipse.che.plugin.docker.client.params.InspectContainerParams;
+import org.eclipse.che.plugin.docker.client.params.InspectImageParams;
+import org.eclipse.che.plugin.docker.client.params.KillContainerParams;
+import org.eclipse.che.plugin.docker.client.params.PullParams;
+import org.eclipse.che.plugin.docker.client.params.PushParams;
+import org.eclipse.che.plugin.docker.client.params.PutResourceParams;
+import org.eclipse.che.plugin.docker.client.params.RemoveContainerParams;
+import org.eclipse.che.plugin.docker.client.params.RemoveImageParams;
+import org.eclipse.che.plugin.docker.client.params.StartContainerParams;
+import org.eclipse.che.plugin.docker.client.params.StartExecParams;
+import org.eclipse.che.plugin.docker.client.params.StopContainerParams;
+import org.eclipse.che.plugin.docker.client.params.TagParams;
+import org.eclipse.che.plugin.docker.client.params.TopParams;
+import org.eclipse.che.plugin.docker.client.params.WaitContainerParams;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -63,7 +87,6 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -82,6 +105,7 @@ import static javax.ws.rs.core.Response.Status.OK;
  * @author andrew00x
  * @author Alexander Garagatyi
  * @author Anton Korneta
+ * @author Mykola Morhun
  */
 @Singleton
 public class DockerConnector {
@@ -109,14 +133,14 @@ public class DockerConnector {
      *
      * @return system-wide information
      * @throws IOException
+     *          when a problem occurs with docker api calls
      */
     public org.eclipse.che.plugin.docker.client.json.SystemInfo getSystemInfo() throws IOException {
         try (DockerConnection connection = connectionFactory.openConnection(dockerDaemonUri)
                                                             .method("GET")
                                                             .path("/info")) {
             final DockerResponse response = connection.request();
-            final int status = response.getStatus();
-            if (OK.getStatusCode() != status) {
+            if (OK.getStatusCode() != response.getStatus()) {
                 throw getDockerException(response);
             }
             return parseResponseStreamAndClose(response.getInputStream(), org.eclipse.che.plugin.docker.client.json.SystemInfo.class);
@@ -130,14 +154,14 @@ public class DockerConnector {
      *
      * @return information about version docker
      * @throws IOException
+     *          when a problem occurs with docker api calls
      */
     public Version getVersion() throws IOException {
         try (DockerConnection connection = connectionFactory.openConnection(dockerDaemonUri)
                                                             .method("GET")
                                                             .path("/version")) {
             final DockerResponse response = connection.request();
-            final int status = response.getStatus();
-            if (OK.getStatusCode() != status) {
+            if (OK.getStatusCode() != response.getStatus()) {
                 throw getDockerException(response);
             }
             return parseResponseStreamAndClose(response.getInputStream(), Version.class);
@@ -151,17 +175,17 @@ public class DockerConnector {
      *
      * @return list of docker images
      * @throws IOException
+     *          when a problem occurs with docker api calls
      */
-    public Image[] listImages() throws IOException {
+    public List<Image> listImages() throws IOException {
         try (DockerConnection connection = connectionFactory.openConnection(dockerDaemonUri)
                                                             .method("GET")
                                                             .path("/images/json")) {
             final DockerResponse response = connection.request();
-            final int status = response.getStatus();
-            if (OK.getStatusCode() != status) {
+            if (OK.getStatusCode() != response.getStatus()) {
                 throw getDockerException(response);
             }
-            return parseResponseStreamAndClose(response.getInputStream(), Image[].class);
+            return parseResponseStreamAsListAndClose(response.getInputStream());
         } catch (JsonParseException e) {
             throw new IOException(e.getLocalizedMessage(), e);
         }
@@ -171,7 +195,9 @@ public class DockerConnector {
      * The same as {@link #buildImage(String, ProgressMonitor, AuthConfigs, boolean, long, long, File...)} but without memory limits.
      *
      * @see #buildImage(String, ProgressMonitor, AuthConfigs, boolean, long, long, File...)
+     * @deprecated use {@link #buildImage(BuildImageParams, ProgressMonitor)} instead
      */
+    @Deprecated
     public String buildImage(String repository,
                              ProgressMonitor progressMonitor,
                              AuthConfigs authConfigs,
@@ -200,7 +226,9 @@ public class DockerConnector {
      * @throws IOException
      * @throws InterruptedException
      *         if build process was interrupted
+     * @deprecated use {@link #buildImage(BuildImageParams, ProgressMonitor)} instead
      */
+    @Deprecated
     public String buildImage(String repository,
                              ProgressMonitor progressMonitor,
                              AuthConfigs authConfigs,
@@ -233,7 +261,9 @@ public class DockerConnector {
      * @throws IOException
      * @throws InterruptedException
      *         if build process was interrupted
+     * @deprecated use {@link #buildImage(BuildImageParams, ProgressMonitor)} instead
      */
+    @Deprecated
     protected String buildImage(String repository,
                                 File tar,
                                 final ProgressMonitor progressMonitor,
@@ -251,16 +281,34 @@ public class DockerConnector {
      * @throws IOException
      */
     public ImageInfo inspectImage(String image) throws IOException {
-        return doInspectImage(image, dockerDaemonUri);
+        return inspectImage(InspectImageParams.from(image));
     }
 
+    /**
+     * @deprecated use {@link #inspectImage(InspectImageParams)} instead
+     */
+    @Deprecated
     protected ImageInfo doInspectImage(String image, URI dockerDaemonUri) throws IOException {
+        return inspectImage(InspectImageParams.from(image), dockerDaemonUri);
+    }
+
+    /**
+     * Gets detailed information about docker image.
+     *
+     * @return detailed information about {@code image}
+     * @throws IOException
+     *          when a problem occurs with docker api calls
+     */
+    public ImageInfo inspectImage(InspectImageParams params) throws IOException {
+        return inspectImage(params, dockerDaemonUri);
+    }
+
+    private ImageInfo inspectImage(InspectImageParams params, URI dockerDaemonUri) throws IOException {
         try (DockerConnection connection = connectionFactory.openConnection(dockerDaemonUri)
                                                             .method("GET")
-                                                            .path("/images/" + image + "/json")) {
+                                                            .path("/images/" + params.getImage() + "/json")) {
             final DockerResponse response = connection.request();
-            final int status = response.getStatus();
-            if (OK.getStatusCode() != status) {
+            if (OK.getStatusCode() != response.getStatus()) {
                 throw getDockerException(response);
             }
             return parseResponseStreamAndClose(response.getInputStream(), ImageInfo.class);
@@ -269,12 +317,21 @@ public class DockerConnector {
         }
     }
 
+    /**
+     * @deprecated use {@link #removeImage(RemoveImageParams)} instead
+     */
+    @Deprecated
     public void removeImage(String image, boolean force) throws IOException {
-        doRemoveImage(image, force, dockerDaemonUri);
+        removeImage(RemoveImageParams.from(image)
+                                     .withForce(force));
     }
 
+    /**
+     * @deprecated use {@link #tag(TagParams)} instead
+     */
+    @Deprecated
     public void tag(String image, String repository, String tag) throws IOException {
-        doTag(image, repository, tag, dockerDaemonUri);
+        tag(TagParams.from(image, repository).withTag(tag), dockerDaemonUri);
     }
 
     /**
@@ -290,10 +347,12 @@ public class DockerConnector {
      *         ProgressMonitor for images creation process
      * @return digest of just pushed image
      * @throws IOException
-     *         when problems occurs with docker api calls
+     *          when a problem occurs with docker api calls
      * @throws InterruptedException
      *         if push process was interrupted
+     * @deprecated use {@link #push(PushParams, ProgressMonitor)}
      */
+    @Deprecated
     public String push(String repository,
                        String tag,
                        String registry,
@@ -308,7 +367,9 @@ public class DockerConnector {
      *
      * @throws IOException
      * @throws InterruptedException
+     * @deprecated use {@link #pull(PullParams, ProgressMonitor)} instead
      */
+    @Deprecated
     public void pull(String image,
                      String tag,
                      String registry,
@@ -316,10 +377,18 @@ public class DockerConnector {
         doPull(image, tag, registry, progressMonitor, dockerDaemonUri);
     }
 
+    /**
+     * @deprecated use {@link #createContainer(CreateContainerParams)} instead
+     */
+    @Deprecated
     public ContainerCreated createContainer(ContainerConfig containerConfig, String containerName) throws IOException {
-        return doCreateContainer(containerConfig, containerName, dockerDaemonUri);
+        return createContainer(CreateContainerParams.from(containerConfig).withContainerName(containerName), dockerDaemonUri);
     }
 
+    /**
+     * @deprecated use {@link #startContainer(StartContainerParams)} instead
+     */
+    @Deprecated
     public void startContainer(String container, HostConfig hostConfig) throws IOException {
         doStartContainer(container, hostConfig, dockerDaemonUri);
     }
@@ -334,19 +403,30 @@ public class DockerConnector {
      * @param timeunit
      *         time unit of the timeout parameter
      * @throws IOException
+     * @deprecated use {@link #stopContainer(StopContainerParams)} instead
      */
+    @Deprecated
     public void stopContainer(String container, long timeout, TimeUnit timeunit) throws IOException {
-        final List<Pair<String, ?>> headers = new ArrayList<>(2);
-        headers.add(Pair.of("Content-Type", MediaType.TEXT_PLAIN));
-        headers.add(Pair.of("Content-Length", 0));
+        stopContainer(StopContainerParams.from(container)
+                                         .withTimeout(timeout, timeunit));
+    }
+
+    /**
+     * Stops container.
+     *
+     * @throws IOException
+     *          when a problem occurs with docker api calls
+     */
+    public void stopContainer(final StopContainerParams params) throws IOException {
+        final Long timeout = (params.getTimeout() == null) ? null :
+                             (params.getTimeunit() == null) ? params.getTimeout() : params.getTimeunit().toSeconds(params.getTimeout());
+
         try (DockerConnection connection = connectionFactory.openConnection(dockerDaemonUri)
                                                             .method("POST")
-                                                            .path("/containers/" + container + "/stop")
-                                                            .query("t", timeunit.toSeconds(timeout))
-                                                            .headers(headers)) {
+                                                            .path("/containers/" + params.getContainer() + "/stop")) {
+            addQueryParamIfNotNull(connection, "t", timeout);
             final DockerResponse response = connection.request();
-            final int status = response.getStatus();
-            if (!(NO_CONTENT.getStatusCode() == status || NOT_MODIFIED.getStatusCode() == status)) {
+            if (response.getStatus() / 100 != 2) {
                 throw getDockerException(response);
             }
         }
@@ -360,20 +440,27 @@ public class DockerConnector {
      * @param signal
      *         code of signal, e.g. 9 in case of SIGKILL
      * @throws IOException
+     * @deprecated use {@link #killContainer(KillContainerParams)} instead
      */
+    @Deprecated
     public void killContainer(String container, int signal) throws IOException {
-        final List<Pair<String, ?>> headers = new ArrayList<>(2);
-        headers.add(Pair.of("Content-Type", MediaType.TEXT_PLAIN));
-        headers.add(Pair.of("Content-Length", 0));
+        killContainer(KillContainerParams.from(container).withSignal(signal));
+    }
 
+    /**
+     * Sends specified signal to running container.
+     * If signal not set, then SIGKILL will be used.
+     *
+     * @throws IOException
+     *          when a problem occurs with docker api calls
+     */
+    public void killContainer(final KillContainerParams params) throws IOException {
         try (DockerConnection connection = connectionFactory.openConnection(dockerDaemonUri)
                                                             .method("POST")
-                                                            .path("/containers/" + container + "/kill")
-                                                            .query("signal", signal)
-                                                            .headers(headers)) {
+                                                            .path("/containers/" + params.getContainer() + "/kill")) {
+            addQueryParamIfNotNull(connection, "signal", params.getSignal());
             final DockerResponse response = connection.request();
-            final int status = response.getStatus();
-            if (NO_CONTENT.getStatusCode() != status) {
+            if (NO_CONTENT.getStatusCode() != response.getStatus()) {
                 throw getDockerException(response);
             }
         }
@@ -387,7 +474,7 @@ public class DockerConnector {
      * @throws IOException
      */
     public void killContainer(String container) throws IOException {
-        killContainer(container, 9);
+        killContainer(KillContainerParams.from(container));
     }
 
     /**
@@ -400,16 +487,30 @@ public class DockerConnector {
      * @param removeVolumes
      *         if {@code true} removes volumes associated to the container
      * @throws IOException
+     *          when a problem occurs with docker api calls
+     * @deprecated use {@link #removeContainer(RemoveContainerParams)} instead
      */
+    @Deprecated
     public void removeContainer(String container, boolean force, boolean removeVolumes) throws IOException {
+        removeContainer(RemoveContainerParams.from(container)
+                                             .withForce(force)
+                                             .withRemoveVolumes(removeVolumes));
+    }
+
+    /**
+     * Removes docker container.
+     *
+     * @throws IOException
+     *          when a problem occurs with docker api calls
+     */
+    public void removeContainer(final RemoveContainerParams params) throws IOException {
         try (DockerConnection connection = connectionFactory.openConnection(dockerDaemonUri)
                                                             .method("DELETE")
-                                                            .path("/containers/" + container)
-                                                            .query("force", force ? 1 : 0)
-                                                            .query("v", removeVolumes ? 1 : 0)) {
+                                                            .path("/containers/" + params.getContainer())) {
+            addQueryParamIfNotNull(connection, "force", params.isForce());
+            addQueryParamIfNotNull(connection, "v", params.isRemoveVolumes());
             final DockerResponse response = connection.request();
-            final int status = response.getStatus();
-            if (NO_CONTENT.getStatusCode() != status) {
+            if (NO_CONTENT.getStatusCode() != response.getStatus()) {
                 throw getDockerException(response);
             }
         }
@@ -424,17 +525,22 @@ public class DockerConnector {
      * @throws IOException
      */
     public int waitContainer(String container) throws IOException {
-        final List<Pair<String, ?>> headers = new ArrayList<>(2);
-        headers.add(Pair.of("Content-Type", MediaType.TEXT_PLAIN));
-        headers.add(Pair.of("Content-Length", 0));
+        return waitContainer(WaitContainerParams.from(container));
+    }
 
+    /**
+     * Blocks until container stops, then returns the exit code
+     *
+     * @return exit code
+     * @throws IOException
+     *          when a problem occurs with docker api calls
+     */
+    public int waitContainer(final WaitContainerParams params) throws IOException {
         try (DockerConnection connection = connectionFactory.openConnection(dockerDaemonUri)
                                                             .method("POST")
-                                                            .path("/containers/" + container + "/wait")
-                                                            .headers(headers)) {
+                                                            .path("/containers/" + params.getContainer() + "/wait")) {
             final DockerResponse response = connection.request();
-            final int status = response.getStatus();
-            if (OK.getStatusCode() != status) {
+            if (OK.getStatusCode() != response.getStatus()) {
                 throw getDockerException(response);
             }
             return parseResponseStreamAndClose(response.getInputStream(), ContainerExitStatus.class).getStatusCode();
@@ -452,16 +558,35 @@ public class DockerConnector {
      * @throws IOException
      */
     public ContainerInfo inspectContainer(String container) throws IOException {
-        return doInspectContainer(container, dockerDaemonUri);
+        return inspectContainer(InspectContainerParams.from(container));
     }
 
+    /**
+     * @deprecated use {@link #createContainer(CreateContainerParams)} instead
+     */
+    @Deprecated
     protected ContainerInfo doInspectContainer(String container, URI dockerDaemonUri) throws IOException {
+        return inspectContainer(InspectContainerParams.from(container), dockerDaemonUri);
+    }
+
+    /**
+     * Gets detailed information about docker container.
+     *
+     * @return detailed information about {@code container}
+     * @throws IOException
+     *          when a problem occurs with docker api calls
+     */
+    public ContainerInfo inspectContainer(final InspectContainerParams params) throws IOException {
+        return inspectContainer(params, dockerDaemonUri);
+    }
+
+    private ContainerInfo inspectContainer(final InspectContainerParams params, URI dockerDaemonUri) throws IOException {
         try (DockerConnection connection = connectionFactory.openConnection(dockerDaemonUri)
                                                             .method("GET")
-                                                            .path("/containers/" + container + "/json")) {
+                                                            .path("/containers/" + params.getContainer() + "/json")) {
+            addQueryParamIfNotNull(connection, "size", params.isReturnContainerSize());
             final DockerResponse response = connection.request();
-            final int status = response.getStatus();
-            if (OK.getStatusCode() != status) {
+            if (OK.getStatusCode() != response.getStatus()) {
                 throw getDockerException(response);
             }
             return parseResponseStreamAndClose(response.getInputStream(), ContainerInfo.class);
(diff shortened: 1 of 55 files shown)


mmorhun:
@garagatyi check, please

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/BuildImageParams.java
```diff
@@ -0,0 +1,97 @@
...
+ * Arguments holder for {@code buildImage} method of {@link org.eclipse.che.plugin.docker.client.DockerConnector}.
+ *
+ * @author Mykola Morhun
+ */
+public class BuildImageParams {
+    /** full repository name to be applied to newly created image */
+    private String repository;
+    /** authentication configuration for private registries. Can be null */
+    private AuthConfigs authConfigs;
+    /**  */
```
sleshchenko: take a look

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/GetSystemInfoParams.java
```diff
@@ -0,0 +1,20 @@
...
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.plugin.docker.client.params;
+
+/**
+ * Arguments holder for {@code getSystemInfo} method of {@link org.eclipse.che.plugin.docker.client.DockerConnector}.
+ *
+ * @author Mykola Morhun
+ */
+public class GetSystemInfoParams {
```
sleshchenko: Do we need this new unused empty class?

mmorhun: Yes

mmorhun:
@garagatyi check please

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/ContainerInfo.java
```diff
@@ -106,12 +107,12 @@ public void setNetworkSettings(NetworkSettings networkSettings) {
         this.networkSettings = networkSettings;
     }
 
-    public String getResolvConfPath() {
-        return resolvConfPath;
+    public String getResolveConfPath() {
```
garagatyi: Why you've changed this?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/AttachContainerParams.java
```diff
@@ -0,0 +1,44 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.plugin.docker.client.params;
+
+/**
+ * Arguments holder for {@code attachContainer} method of {@link org.eclipse.che.plugin.docker.client.DockerConnector}.
```
garagatyi: Method reference is more preferable than reference to class and name of method

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/AttachContainerParams.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @author Mykola Morhun
+ */
+public class AttachContainerParams {
+    /** id of container */
+    private String container;
+    /**
+     * if {@code true} gets 'live' stream from container.
+     * Note, that 'live' stream blocks until container is running.
+     */
```
garagatyi, line 37: Docs for private members is less usable than docs for public methods

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/AttachContainerParams.java
```diff
@@ -0,0 +1,44 @@
...
+    private Boolean stream;
+
+    public AttachContainerParams withContainer(String container) {
+        this.container = container;
+        return this;
+    }
+
+    public AttachContainerParams withStream(Boolean stream) {
```
garagatyi: why you need wrapper in args? wrapper is needed in filed to avoid setting default boolean, but what is the purpose to use it here?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/ContainerInfo.java
mmorhun: I fixed typo

garagatyi: It is not a typo. We used the same wording Docker use itself. You can check it by inspecting docker container.

mmorhun:
@garagatyi I corrected the code according to your comments

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/AttachContainerParams.java
```diff
@@ -0,0 +1,51 @@
...
+ * @author Mykola Morhun
+ */
+public class AttachContainerParams {
+
+    private String container;
+    private Boolean stream;
+
+    /**
+     * @param container
+     *         id of container
```
garagatyi: are you sure that only id is valid value?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/BuildImageParams.java
```diff
@@ -0,0 +1,111 @@
...
+     *         full repository name to be applied to newly created image
+     */
+    public BuildImageParams withRepository(String repository) {
+        this.repository = repository;
+        return this;
+    }
+
+    /**
+     * @param authConfigs
+     *         authentication configuration for private registries. Can be null
```
garagatyi: Not only for private registries

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/BuildImageParams.java
```diff
@@ -0,0 +1,111 @@
...
+     */
+    public BuildImageParams withAuthConfigs(AuthConfigs authConfigs) {
+        this.authConfigs = authConfigs;
+        return this;
+    }
+
+    /**
+     * @param doForcePull
+     *         is pull with force
```
garagatyi: Try to play with wording to make it more clear.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/AttachContainerParams.java
```diff
@@ -0,0 +1,51 @@
...
+     */
+    public AttachContainerParams withContainer(String container) {
+        this.container = container;
+        return this;
+    }
+
+    /**
+     * @param stream
+     *         if {@code true} gets 'live' stream from container.
```
garagatyi: please explain what is live stream

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/BuildImageParams.java
```diff
@@ -0,0 +1,111 @@
...
+     */
+    public BuildImageParams withDoForcePull(boolean doForcePull) {
+        this.doForcePull = doForcePull;
+        return this;
+    }
+
+    /**
+     * @param memoryLimit
+     *         memory limit for build in bytes
```
garagatyi: make it more clear that it is only RAM limit, not something else

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/BuildImageParams.java
```diff
@@ -0,0 +1,111 @@
...
+        this.memorySwapLimit = memorySwapLimit;
+        return this;
+    }
+
+    /**
+     * @param files
+     *         files that are needed for creation docker images (e.g. file of directories used in ADD instruction in Dockerfile).
+     *         One of them must be Dockerfile.
+     */
+    public BuildImageParams withFiles(File... files) {
```
garagatyi: I would say we can improve that by addition of syntax sugar methods: addFile(File file), addFiles(File... files)
And even addDockerfile(Dockerfile dockerfile)
WDYT?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/CreateContainerParams.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.eclipse.che.plugin.docker.client.json.ContainerConfig;
+
+/**
+ * Arguments holder for {@link org.eclipse.che.plugin.docker.client.DockerConnector#createContainer(CreateContainerParams)}.
+ *
+ * @author Mykola Morhun
+ */
+public class CreateContainerParams {
+    /** configuration of future container */
```
garagatyi: move docs to public methods

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/CreateContainerParams.java
```diff
@@ -0,0 +1,44 @@
...
+    private ContainerConfig containerConfig;
+    /** assign the specified name to the container. Must match /?[a-zA-Z0-9_-]+ */
```
garagatyi: Can you point me where you found this rule of matching?

garagatyi: I found it in Docker API docs.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/CreateExecParams.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+public class CreateExecParams {
+
+    private String   container;
+    private Boolean  detach;
+    private String[] cmd;
+
+    /**
+     * @param container
+     *         id of container
```
garagatyi: Id or name https://docs.docker.com/engine/reference/api/docker_remote_api_v1.21/#exec-create

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/GetResourceParams.java
```diff
@@ -0,0 +1,49 @@
...
+ * @author Mykola Morhun
+ */
+public class GetResourceParams {
+
+    private String container;
+    private String sourcePath;
+
+    /**
+     * @param container
+     *         container id
```
garagatyi: id or name https://docs.docker.com/engine/reference/api/docker_remote_api_v1.21/#get-an-archive-of-a-filesystem-resource-in-a-container

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/GetResourceParams.java
```diff
@@ -0,0 +1,49 @@
...
+     */
+    public GetResourceParams withContainer(String container) {
+        this.container = container;
+        return this;
+    }
+
+    /**
+     * @param sourcePath
+     *         path to file or directory inside specified container
```
garagatyi: please point that it is required param

garagatyi: also add description of path's value. you can summarize https://docs.docker.com/engine/reference/api/docker_remote_api_v1.21/#get-an-archive-of-a-filesystem-resource-in-a-container

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/GetSystemInfoParams.java
```diff
@@ -0,0 +1,20 @@
...
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.plugin.docker.client.params;
+
+/**
+ * Arguments holder for {@link org.eclipse.che.plugin.docker.client.DockerConnector#getSystemInfo(GetSystemInfoParams)}.
+ *
+ * @author Mykola Morhun
+ */
+public class GetSystemInfoParams {
```
garagatyi: I know it is me who convince you to add these parameters but now I doubt it is needed. WDYT?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/InspectContainerParams.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @author Mykola Morhun
+ */
+public class InspectContainerParams {
+
+    private String container;
+
+    /**
+     * @param container
+     *         id of container
```
garagatyi: id or id /name?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/InspectContainerParams.java
```diff
@@ -0,0 +1,35 @@
...
+package org.eclipse.che.plugin.docker.client.params;
+
+/**
+ * Arguments holder for{@link org.eclipse.che.plugin.docker.client.DockerConnector#inspectContainer(InspectContainerParams)}.
+ *
+ * @author Mykola Morhun
+ */
+public class InspectContainerParams {
+
+    private String container;
```
garagatyi: Please add size parameter https://docs.docker.com/engine/reference/api/docker_remote_api_v1.21/#inspect-a-container

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/KillContainerParams.java
```diff
@@ -0,0 +1,35 @@
...
+package org.eclipse.che.plugin.docker.client.params;
+
+/**
+ * Arguments holder for {@link org.eclipse.che.plugin.docker.client.DockerConnector#killContainer(KillContainerParams)}.
+ *
+ * @author Mykola Morhun
+ */
+public class KillContainerParams {
+
+    private String container;
```
garagatyi: Please add signal https://docs.docker.com/engine/reference/api/docker_remote_api_v1.21/#kill-a-container

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/ListImagesParams.java
```diff
@@ -0,0 +1,20 @@
...
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.plugin.docker.client.params;
+
+/**
+ * Arguments holder for {@link org.eclipse.che.plugin.docker.client.DockerConnector#listImages(ListImagesParams)}.
+ *
+ * @author Mykola Morhun
+ */
+public class ListImagesParams {
```
garagatyi: Now I doubt we need that

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/PullParams.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+public class PullParams {
+
+    private String image;
+    private String tag;
+    private String registry;
+
+    /**
+     * @param image
+     *         full repository name to be applied to newly created image
```
garagatyi: AFAIK `full repository name` includes registry, but you added separate field for it

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/PullParams.java
```diff
@@ -0,0 +1,65 @@
...
+     *         tag of the image
+     */
+    public PullParams withTag(String tag) {
+        this.tag = tag;
+        return this;
+    }
+
+    /**
+     * @param registry
+     *         registry url
```
garagatyi: Are you sure that URL is OK here? As far as I understand it should be only host, port. Am I correct?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/PushParams.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+public class PushParams {
+
+    private String repository;
+    private String tag;
+    private String registry;
+
+    /**
+     * @param repository
+     *         full repository name
```
garagatyi: AFAIK `full repository name` includes registry, but you added separate field for it

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/PutResourceParams.java
```diff
@@ -0,0 +1,80 @@
...
+public class PutResourceParams {
+
+    private String container;
+    private String targetPath;
+    private InputStream sourceStream;
+    private Boolean noOverwriteDirNonDir;
+
+    /**
+     * @param container
+     *         container id
```
garagatyi: Only id or id|name?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/PutResourceParams.java
```diff
@@ -0,0 +1,80 @@
...
+     */
+    public PutResourceParams withContainer(String container) {
+        this.container = container;
+        return this;
+    }
+
+    /**
+     * @param targetPath
+     *         path to file or directory inside specified container
```
garagatyi: please describe value. Is it absolute or no, what if no, etc.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/PutResourceParams.java
```diff
@@ -0,0 +1,80 @@
...
+     */
+    public PutResourceParams withTargetPath(String targetPath) {
+        this.targetPath = targetPath;
+        return this;
+    }
+
+    /**
+     * @param sourceStream
```
garagatyi, line 83: Please describe how it this stream was obtained

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/PutResourceParams.java
```diff
@@ -0,0 +1,80 @@
...
+     *         stream of files from source container
+     */
+    public PutResourceParams withSourceStream(InputStream sourceStream) {
+        this.sourceStream = sourceStream;
+        return this;
+    }
+
+    /**
+     * @param noOverwriteDirNonDir
+     *         If "false" then it will be an error if unpacking the given content would cause an existing
```
garagatyi: I believe that _then it will be an error_ is a bad wording. Please rephrase.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/RemoveImageParams.java
```diff
@@ -0,0 +1,49 @@
...
+ * @author Mykola Morhun
+ */
+public class RemoveImageParams {
+
+    private String  image;
+    private Boolean force;
+
+    /**
+     * @param image
+     *         image name
```
garagatyi: Can you check if it is possible to use image id here

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/RemoveImageParams.java
```diff
@@ -0,0 +1,49 @@
...
+     */
+    public RemoveImageParams withImage(String image) {
+        this.image = image;
+        return this;
+    }
+
+    /**
+     * @param force
+     *         force removal of the image
```
garagatyi: please describe as you did it in RemoveContainerParams

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/SendSignalToContainerParams.java
```diff
@@ -0,0 +1,49 @@
...
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.plugin.docker.client.params;
+
+/**
+ * Arguments holder for {@link org.eclipse.che.plugin.docker.client.DockerConnector#sendSignalToContainer(SendSignalToContainerParams)}.
+ *
+ * @author Mykola Morhun
+ */
+public class SendSignalToContainerParams {
```
garagatyi: Why you decided to add send signal endpoint to docker client? Docker API lacks this endpoint AFAIK

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/StartContainerParams.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.plugin.docker.client.params;
+
+import org.eclipse.che.plugin.docker.client.json.HostConfig;
+
+/**
+ * Arguments holder for {@link org.eclipse.che.plugin.docker.client.DockerConnector#startContainer(StartContainerParams)}.
```
garagatyi, line 20: You can see [here](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#start-a-container) that HostConfig is supported for backward compatibility only.
I think you should not add HostConfig to StartContainer params because we don't need such support.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/StartExecParams.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.eclipse.che.plugin.docker.client.MessageProcessor;
+
+/**
+ * Arguments holder for {@link org.eclipse.che.plugin.docker.client.DockerConnector#startExec(StartExecParams, MessageProcessor)}.
+ *
+ * @author Mykola Morhun
+ */
+public class StartExecParams {
+    private String id;
```
garagatyi: Please add Tty, detach fields. [API](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#exec-start)

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/StopContainerParams.java
```diff
@@ -0,0 +1,68 @@
...
+     *         container identifier, either id or name
+     */
+    public StopContainerParams withContainer(String container) {
+        this.container = container;
+        return this;
+    }
+
+    /**
+     * @param timeout
+     *         time to wait for the container to stop before killing it
```
garagatyi: You must describe in javadoc default value of timeunit

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/TopParams.java
```diff
@@ -0,0 +1,49 @@
...
+ * @author Mykola Morhun
+ */
+public class TopParams {
+
+    private String   container;
+    private String[] psArgs;
+
+    /**
+     * @param container
+     *         container id
```
garagatyi: id or name

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/TopParams.java
```diff
@@ -0,0 +1,49 @@
...
+    public TopParams withContainer(String container) {
+        this.container = container;
+        return this;
+    }
+
+    /**
+     * @param psArgs
+     *         ps arguments to use
+     */
+    public TopParams withPsArgs(String[] psArgs) {
```
garagatyi: Why not varargs?

garagatyi:
I believe that such a big class as DockerConnector is a problem. Even Gihub doesn't want to show diff of that class! :)

garagatyi:
You must add yourself to authors list in javadoc when you do significant changes in class

mmorhun:
@garagatyi , done

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/AttachContainerParams.java
```diff
@@ -33,8 +33,11 @@ public AttachContainerParams withContainer(String container) {
 
     /**
      * @param stream
-     *         if {@code true} gets 'live' stream from container.
-     *         Note, that 'live' stream blocks until container is running.
+     *         if {@code true} gets stream from container.
```
garagatyi: Stream of what?
Don't forget to add html line breaks if needed

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/BuildImageParams.java
```diff
@@ -40,7 +43,7 @@ public BuildImageParams withRepository(String repository) {
 
     /**
      * @param authConfigs
-     *         authentication configuration for private registries. Can be null
+     *         authentication configuration for registries. Can be null
```
garagatyi: It doesn't make sense to write here that it can be null, so I suggest we need to move javadocs to getter. In that case description will fit purpose of method better.
Also use @Nullable instead of **Can be null** in that case (with getter).

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/BuildImageParams.java
```diff
@@ -75,12 +78,34 @@ public BuildImageParams withMemorySwapLimit(long memorySwapLimit) {
...
+     */
+    public BuildImageParams addFiles(File... files) {
+        this.files.addAll(Arrays.asList(files));
+        return this;
+    }
+
+    /**
+     * The same as {@link #addFiles(File...)}, but for single file.
+     */
+    public BuildImageParams addFile(File file) {
```
garagatyi: I think it is OK to have only 1 method addFiles(), WDYT?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/BuildImageParams.java
```diff
@@ -75,12 +78,34 @@ public BuildImageParams withMemorySwapLimit(long memorySwapLimit) {
...
+
+    /**
+     * Adds files to the file list.
+     * @see {@link #withFiles(File...)}
+     *
+     * @param files
+     *         files to add to image
+     */
+    public BuildImageParams addFiles(File... files) {
+        this.files.addAll(Arrays.asList(files));
```
garagatyi: This method allows to pass 0 arguments! Change signature to avoid such a case.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/GetResourceParams.java
```diff
@@ -31,7 +31,9 @@ public GetResourceParams withContainer(String container) {
 
     /**
      * @param sourcePath
-     *         path to file or directory inside specified container
+     *         resource in the container’s filesystem to archive. Required.
```
garagatyi, line 66: Don't forget that new line in javadocs doesn't make new line in resulting html

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/InspectContainerParams.java
```diff
@@ -17,19 +17,33 @@
...
     public InspectContainerParams withContainer(String container) {
         this.container = container;
         return this;
     }
 
+    /**
+     * @param getContainerSize
+     *         if {@code true} it will return container size information
+     */
+    public InspectContainerParams withGetContainerSize(boolean getContainerSize) {
```
garagatyi: returnContainerSize?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/PullParams.java
```diff
@@ -43,7 +43,7 @@ public PullParams withTag(String tag) {
 
     /**
      * @param registry
-     *         registry url
+     *         host and port of registry, e.g. localhost:5000
```
garagatyi: Please mention that it is optional in case of official docker registry - hub.docker.com

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/BuildImageParams.java
garagatyi: Or as parameters of docker mostly optional we can consider all parameters nullable and add required annotation to those that are not. If so we need package docs.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/TopParams.java
```diff
@@ -33,7 +33,7 @@ public TopParams withContainer(String container) {
      * @param psArgs
      *         ps arguments to use
      */
-    public TopParams withPsArgs(String[] psArgs) {
+    public TopParams withPsArgs(String... psArgs) {
```
garagatyi, line 67: Don't allow empty params list

garagatyi:
Please do not add doSmth(args, dockerDaemonUri) if it is not needed.
For example doGetSystemInfo

garagatyi:
```
@throws IOException
     *         when problems occurs with docker api calls
```

doesn't make sense if it is only one throws annotation

mmorhun:
@garagatyi check please

mmorhun:
@evoevodin , @skabashnyuk , @sleshchenko ?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -109,14 +132,14 @@ public DockerConnector(DockerConnectorConfiguration connectorConfiguration,
      *
      * @return system-wide information
      * @throws IOException
+     *         when problems occurs with docker api calls
```
voievodin: typo, `problems` - plural

voievodin:
-1 so far.

The idea of the changes is good and brings vertical growing of the API but:
- Those changes are not consistent:
  e.g.

``` java
BuildImageParams.from() // <- why empty from allowed ?
                .withMemoryLimit(long memoryLimit) // <- why class holds Long wrapper
                .withRepository(String repository)

```

I don't like mixed API style.
If you use `with*` methods for setting values then you should use `get*` methods for getting values, 
because it looks really weird when you get repository with `.repository()` method while you set it by `withRepository(repository)` method.

So you should either use getters or change with method: 
e.g.

``` java
// either
.repository() // <- gets value
.repository(String repository) // <- sets value

// or
.getRepository();
.withRepository(String repository);

// BUT NOT
.repository();
.withRepository(String repository)
```
-  Don't deprecate existing methods which are convenient.
  e.g. `killContainer(container)` is really useful. So, i think that you shouldn't remove the methods, or even add new methods which work with a small(only required) set of parameters, without `*Params` objects.
- Review params objects. Some of them won't even work 
  e.g.

```
    public BuildImageParams withFiles(@NotNull File... files) {
        requireNonNull(files);
        this.files = Arrays.asList(files); <- this.files is unmodifiable so addFiles will throw runtime exception
        return this;
    }

    public BuildImageParams addFiles(@NotNull File... files) {
        requireNonNull(files);
        this.files.addAll(Arrays.asList(files)); 
        return this;
   }
```

Which means that you running out of required tests.
I don't like the documentation style you provided for `*Params` objects, if `@param` field is documented the method description should provided as well, even if it looks useless, try to generate javadoc and see what you get.

I did a general review, so let me know if PR is ready for the picky review =)

mmorhun:
@evoevodin thank you for brief review
- Where are you found it? All `from(~)` methods get a set of arguments

> BuildImageParams.from() // <- why empty from allowed ?
- Each nullable parameter has reference type. In this case we can exactly say, was this parameter set by user or it was left for its default value.

> .withMemoryLimit(long memoryLimit) // <- why class holds Long wrapper
- I want to make it like sentence (good example here is Mockito). Look:

```
PullParams.from(registry, image).withForce(true).withSmth(smth);
// VS
PullParams p = PullParams.from(registry, image);
p.setForce(true);
p.setSmth(smth);
```

```
parameters.image()
// VS
parameters.getImage();
```
- I agree, that killContainer(container) is really useful. But adding methods with only required parameters is unacceptable, I think. What if I decide to add some optional parameter? In this case, I must change method that I call. In case with parameters object I'll just add `with*` to existing code.
- Tests is important, no doubts. May be I add they, if I have time.
- What is wrong in javadoc like this?

```
    withMemoryLimit

    public BuildImageParams withMemoryLimit(long memoryLimit)

    Parameters:
        memoryLimit - RAM memory limit for build in bytes
```

P.S. I'll mention you, when PR will be ready, thx)

voievodin:
@mmorhun i think that you missed some of my points.

> Where are you found it? All from(~) methods get a set of arguments

I didn't say that you used it somewhere i said that it is weird that empty `.from` is allowed.

> I want to make it like sentence (good example here is Mockito). Look:

It's good that you want to do it, but the style is mixed.

> I agree, that killContainer(container) is really useful. But adding methods with only required parameters > is unacceptable, I think. What if I decide to add some optional parameter? In this case, I must change method that I call. In case with parameters object I'll just add with\* to existing code.

Well, my point is that you MUST add `Params*` like method, and you MUST leave existing methods with required parameters(both variants).

> What is wrong in javadoc like this?

I don't like the approach of writing such javadoc i'd say that it is much better to write a description which will copy method name somehow, because you wrote obvious things on `from` methods but you didn't on `withMethods`.

``` java
+    /**
 +     * Creates arguments holder with required parameters.
 +     *
 +     * @param files
 +     *         info about this parameter see {@link #withFiles(File...)}
 +     * @return arguments holder with required parameters
 +     */
 +     public static BuildImageParams from(@NotNull File... files) {
```

Why do you document this method fully but you don't want to document `with*` methods in the same way?

mmorhun:
We have discussion, which name convention use for getters in `*Params` classes.
Typical use case:

```
// creation of arguments holder
TagParams.from(image, repository)
         .withTag(tag)
         .withForce(false);
```

```
// using tag parameters 
 try (DockerConnection connection = connectionFactory.openConnection(dockerDaemonUri)
                                                     .method("POST")
                                                     .path("/images/" + params.image() + "/tag")
                                                     .query("repo", params.repository())) {
            addQueryParamIfNotNull(connection, "force", params.force());
            addQueryParamIfNotNull(connection, "tag", params.tag());
```

Note, that getters and setters are used separately.
So, the question is in the names of getters. Now, we have two variants:

```
params.tag()                    params.getTag()
params.image()       VS         params.getImage()
params.force()                  params.isForce()
```

@skabashnyuk @garagatyi @sleshchenko @AndrienkoAleksandr WDYT?

voievodin:
Again, it is not a discussion about getters, you can pick any style you want the thing which is important  is a couple of getters and setters. So i would rather say that this is more about:

```
params.withTag(tag)                                     params.tag(tag)
params.getTag()                       VS                params.tag()
```

skabashnyuk:
+1 for 

```
params.withTag(tag)
params.getTag() 
```

garagatyi:
+0

AndrienkoAleksandr:
+1 for
params.withTag(tag)
params.getTag()

mmorhun:
@evoevodin @skabashnyuk check please

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -334,19 +401,31 @@ public void startContainer(String container, HostConfig hostConfig) throws IOExc
...
                                                             .method("POST")
-                                                            .path("/containers/" + container + "/stop")
-                                                            .query("t", timeunit.toSeconds(timeout))
-                                                            .headers(headers)) {
+                                                            .path("/containers/" + params.getContainer() + "/stop")) {
+            addQueryParamIfNotNull(connection, "t", timeout);
             final DockerResponse response = connection.request();
             final int status = response.getStatus();
-            if (!(NO_CONTENT.getStatusCode() == status || NOT_MODIFIED.getStatusCode() == status)) {
+            if (NO_CONTENT.getStatusCode() != status && NOT_MODIFIED.getStatusCode() != status) {
```
garagatyi: I think that `status / 100 != 2` will be better in that case

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -387,7 +473,7 @@ public void killContainer(String container, int signal) throws IOException {
      * @throws IOException
      */
     public void killContainer(String container) throws IOException {
-        killContainer(container, 9);
+        killContainer(KillContainerParams.from(container));
```
garagatyi, line 477: Why you haven't set signal 9 here?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -711,32 +869,46 @@ public InputStream getResource(String container, String sourcePath) throws IOExc
...
              DockerConnection connection = connectionFactory.openConnection(dockerDaemonUri)
                                                             .method("PUT")
-                                                            .path("/containers/" + container + "/archive")
-                                                            .query("path", targetPath)
-                                                            .query("noOverwriteDirNonDir", noOverwriteDirNonDir ? 0 : 1)
-                                                            .headers(headers)
+                                                            .path("/containers/" + params.getContainer() + "/archive")
+                                                            .query("path", params.getTargetPath())
+                                                            .header("Content-Type", ExtMediaType.APPLICATION_X_TAR)
+                                                            .header("Content-Length", length)
```
garagatyi, line 911: @mmorhun according to [http spec](http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.4) it is possible to send request without specifying content length if chunked encoding header is set. So I assume that we don't have to save this tar file on disk! Please can you check if it is true with Docker API.

garagatyi, line 911: Sorry, it shouldn't be part of this task as it purpose is absolutely separated. Can you add TODO in that code with explanation that we should check that improvement

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -1038,66 +1323,120 @@ public void run() {
...
                           String tag,
                           String registry,
                           final ProgressMonitor progressMonitor,
                           URI dockerDaemonUri) throws IOException, InterruptedException {
-        final List<Pair<String, ?>> headers = new ArrayList<>(3);
-        headers.add(Pair.of("Content-Type", MediaType.TEXT_PLAIN));
-        headers.add(Pair.of("Content-Length", 0));
-        headers.add(Pair.of("X-Registry-Auth", initialAuthConfig.getAuthConfigHeader()));
+        pull(PullParams.from(image)
+                         .withTag(tag)
```
garagatyi: Please align this code

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -1240,4 +1628,31 @@ public String toJavaName(String jsonName) {
...
+     *
+     * @param connection
+     *         connection to docker service
+     * @param queryParamName
+     *         name of query parameter
+     * @param paramValue
+     *         value of query parameter
+     */
+    private void addQueryParamIfNotNull(DockerConnection connection, String queryParamName, Object paramValue) {
+        if (paramValue != null && queryParamName != null && !queryParamName.equals("")) {
```
garagatyi: are you sure that check of param name is useful here?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/connection/DockerConnection.java
```diff
@@ -61,6 +62,11 @@ public DockerConnection query(String name, Object... values) {
         return this;
     }
 
+    public DockerConnection header(String key, Object value) {
+        this.headers.add(Pair.of(key, value));
+        return this;
+    }
+
     public DockerConnection headers(List<Pair<String, ?>> headers) {
         this.headers = headers;
```
garagatyi: I think that your changes introduced a bug. What if argument of this method is immutable list and method `header(key, value)` will be called after this method. In that case you will get exception

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/BuildImageParams.java
```diff
@@ -0,0 +1,178 @@
...
+     * @return this params instance
+     * @throws NullPointerException
+     *         if {@code files} is null
+     * @throws IllegalArgumentException
+     *         if {@code files} is empty array
+     */
+    public BuildImageParams withFiles(@NotNull File... files) {
+        requireNonNull(files);
+        requireNonEmptyArray(files);
+        this.files = new LinkedList<>();
```
garagatyi: I think you can make prediction of list size based on array size

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/CreateContainerParams.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * Adds container name to this parameters.
+     *
+     * @param containerName
+     *         assign the specified name to the container. Must match /?[a-zA-Z0-9_-]+
+     * @return this params instance
+     */
+    public CreateContainerParams withContainerName(String containerName) {
```
garagatyi: NotNull

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/CreateExecParams.java
```diff
@@ -0,0 +1,107 @@
...
+     * @throws NullPointerException
+     *         if {@code cmd} is null
+     * @throws IllegalArgumentException
+     *         if {@code cmd} is empty
+     */
+    public CreateExecParams withCmd(@NotNull String[] cmd) {
+        requireNonNull(cmd);
+        requireNonEmptyArray(cmd);
+        if (cmd[0].isEmpty()) {
+            throw new IllegalArgumentException();
```
garagatyi: Add message

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/GetEventsParams.java
```diff
@@ -0,0 +1,85 @@
...
+public class GetEventsParams {
+
+    private Long    sinceSecond;
+    private Long    untilSecond;
+    private Filters filters;
+
+    /**
+     * Creates arguments holder.
+     */
+    public static GetEventsParams create() {
```
garagatyi: I think that from is better to align API of all params classes

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/GetResourceParams.java
```diff
@@ -0,0 +1,85 @@
...
+     * Adds container to this parameters.
+     *
+     * @param container
+     *         container id or name
+     * @return this params instance
+     * @throws NullPointerException
+     *         if {@code container} is null
+     */
+    public GetResourceParams withContainer(@NotNull String container) {
+        requireNonNull(container);
```
garagatyi, line 57: NPE without any message is not developer friendly API. I think it is better to add message as second parameter in this and all other places like that.

garagatyi, line 57: The same for string/array emtyness check

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
mmorhun, line 477: Because it is default signal according to docker API: https://docs.docker.com/engine/reference/api/docker_remote_api_v1.21/#kill-a-container

garagatyi, line 477: So why do you want to set default behavior in API client instead of letting API server do this

mmorhun:
@garagatyi ?

garagatyi:
Code looks really good!
But we still need tests.

garagatyi:
LGTM
👍

PR #957, opened 2016-04-01
CHE-926 Hidding not implemented SVN actions from menu

stour:
@skabashnyuk @vparfonov


skabashnyuk:
Can you remove it at all? If we need it we can find it in history.

stour:
@skabashnyuk done

skabashnyuk:
ок

PR #964, opened 2016-04-04
Separate logic of showing dialog on workspace stop

mkuznyetsov:
Dialog showing logic is moved to separate workspace stop handler, so it can be overriden in non-local environment.

### DIFF
--- a/plugins/plugin-nodejs/che-plugin-nodejs-lang-shared/pom.xml
+++ b/plugins/plugin-nodejs/che-plugin-nodejs-lang-shared/pom.xml
@@ -11,18 +11,15 @@
       Codenvy, S.A. - initial API and implementation
 
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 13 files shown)


plugins/plugin-nodejs/che-plugin-nodejs-lang-shared/pom.xml
```diff
@@ -30,4 +27,4 @@
             </resource>
         </resources>
     </build>
-</project>
\ No newline at end of file
+</project>
```
skabashnyuk, line 30: Nice. But how it's related to this issue?

skabashnyuk:
ok. @vparfonov ?

vparfonov:
ok

plugins/plugin-nodejs/che-plugin-nodejs-lang-shared/pom.xml
mkuznyetsov, line 30: It's because of applying sortpom:sort on whole che project.
I decided to commit fixed file, because it is annoying to see this file changing in VCS after sorting.

akorneta:
ok

PR #971, opened 2016-04-05
CHE-98: Restyle check-boxes position in 'git commit' view

vinokurig:
gitcommit
@vparfonov please review


TylerJewell:
@vinokurig - it looks like the check boxes are not vertically spaced properly?

vinokurig:
I have separated with space 'Amend previous commit' check-box because this operation differs from those that are above, @TylerJewell what do you think?

TylerJewell:
@vinokurig - thank you for the explanation.  If you hadn't said that, i would have not understood it.  I think to the average user, it will look weird, though.  It would need to be separated by a separator to make it clear that we mean something different.  I think it's safer to keep it all evenly spaced only because it is not so intuitive.

vinokurig:
gitcommit
Aligned all checkboxes in common way

vparfonov:
ok

PR #976, opened 2016-04-06
CHE-928: fix bug with project creation from project wizard with prese…

olexii4:
**_1 Upvote**_ …lected name

@ashumilova  WDYT?


ashumilova:
+1

PR #978, opened 2016-04-06
CHE-605: Extend Workspace API with new post/pre operations

voievodin:
Add post workspace create operation which allows to start workspace immediately after workspace is created, `start-after-create=true` query parameter is responsible for that.
Add new `auto-snapshot` and `auto-restore` query parameters which go with stop and start requests, if its value is `true` then:
- A snapshot of the workspace will be created before workspace stop is performed.
  If snapshot creation failed then workspace will be stopped anyway.
- Workspace will be recovered from the snapshot if starting with `auto-restore=true` query parameter and snapshot exists. If snapshot doesn't exist the wokrpsace will be regularly started.

@skabashnyuk, @akorneta, @garagatyi please review

### DIFF
--- a/core/platform-api-client-gwt/che-core-client-gwt-workspace/src/main/java/org/eclipse/che/api/workspace/gwt/client/WorkspaceServiceClient.java
+++ b/core/platform-api-client-gwt/che-core-client-gwt-workspace/src/main/java/org/eclipse/che/api/workspace/gwt/client/WorkspaceServiceClient.java
@@ -38,7 +38,7 @@ public interface WorkspaceServiceClient {
      * @param account
      *         the account id related to this operation
      * @return a promise that resolves to the {@link WorkspaceDto}, or rejects with an error
-     * @see WorkspaceService#create(WorkspaceConfigDto, List, String)
+     * @see WorkspaceService#create(WorkspaceConfigDto, List, Boolean, String)
      */
     Promise<WorkspaceDto> create(WorkspaceConfigDto newWorkspace, String account);
 
@@ -106,7 +106,7 @@ public interface WorkspaceServiceClient {
      * @param envName
      *         the name of the workspace environment that should be used for start
      * @return a promise that resolves to the {@link WorkspaceDto}, or rejects with an error
-     * @see WorkspaceService#startById(String, String, String)
+     * @see WorkspaceService#startById(String, String, Boolean, String)
      */
     Promise<WorkspaceDto> startById(String id, String envName);
 
@@ -116,7 +116,7 @@ public interface WorkspaceServiceClient {
      * @param wsId
      *         workspace ID
      * @return a promise that will resolve when the workspace has been stopped, or rejects with an error
-     * @see WorkspaceService#stop(String)
+     * @see WorkspaceService#stop(String, Boolean)
      */
     Promise<Void> stop(String wsId);
 
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
@@ -52,6 +52,7 @@ import static java.lang.String.format;
 import static java.lang.System.currentTimeMillis;
 import static java.util.Collections.emptyMap;
 import static java.util.Objects.requireNonNull;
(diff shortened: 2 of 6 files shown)


core/platform-api-client-gwt/che-core-client-gwt-workspace/src/main/java/org/eclipse/che/api/workspace/gwt/client/WorkspaceServiceClient.java
```diff
@@ -116,7 +116,7 @@
      * @param wsId
      *         workspace ID
      * @return a promise that will resolve when the workspace has been stopped, or rejects with an error
-     * @see WorkspaceService#stop(String)
+     * @see WorkspaceService#stop(String, Boolean)
      */
```
skabashnyuk, line 120: do you have here 1 or 2 parameters?

voievodin, line 120: `WorkspaceService#stop` has an optional boolean query parameter, i didn't change client-side implementation, 
that's why the method takes 1 parameter, but it works as it worked before

akorneta:
ok

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -952,6 +977,16 @@ private SnapshotDto injectLinks(SnapshotDto snapshotDto) {
         return snapshotDto.withLinks(asList(machineLink, workspaceLink, workspaceSnapshotLink));
     }
 
+    /** Returns true if snapshot exists and false otherwise. */
```
garagatyi: Do you need that docs as method's name says exactly the same

skabashnyuk:
ок

benoitf:
it seems default value when stopping workspace is "no snapshot" while reading issue 

when a user is working in a workspace, we need to:
auto-snapshot the state of the workspace where we are shutting it down
auto-restore the latest workspace's snapshot when the user come back to his work.

voievodin:
Read the [comments](https://jira.codenvycorp.com/browse/CHE-605?focusedCommentId=273604&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-273604) below

skabashnyuk:
AFAIK that is expected behavior according to the issue. @evoevodin ?

benoitf:
@evoevodin I've commented JIRA issue, Also I don't see how client could manage per workspace setting as it should store somewhere the settings. So it would mean two clients may have different logic for the same workspace

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -418,7 +419,36 @@ public WorkspaceImpl recoverWorkspace(String workspaceId,
...
+     * @throws ConflictException
+     *         when workspace status is different from the {@link WorkspaceStatus#RUNNING}
+     * @throws ServerException
+     *         when any server error occurs during the workspace getting
+     */
+    public void stopWorkspace(String workspaceId, boolean createSnapshot) throws NotFoundException,
+                                                                                 ServerException,
+                                                                                 ConflictException {
+        requireNonNull(workspaceId, "Required non-null workspace id");
+        final WorkspaceImpl workspace = normalizeState(workspaceDao.get(workspaceId));
```
garagatyi: inline as in previous method

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -444,35 +474,12 @@ public void stopWorkspace(String workspaceId) throws ServerException, NotFoundEx
      * @throws ServerException
      *         when any other error occurs
      */
-    public void createSnapshot(String workspaceId) throws NotFoundException, ServerException {
+    public void createSnapshot(String workspaceId) throws NotFoundException, ServerException, ConflictException {
```
garagatyi, line 478: Please add possible cause of conflict exception into javadoc

garagatyi:
LGTM

PR #1005, opened 2016-04-10
Fixing usage of DataOutputStream and ObjectOutputStream 

Wajihulhassan:
When a `DataOutputStream` or an `ObjectOutputStream` instance wraps an underlying `ByteArrayOutputStream` instance, it is recommended to flush or close the DataOutputStream and ObjectOutputStream before invoking the  underlying instances's toByteArray(). Also, it is a good practice to call flush/close explicitly as mentioned for example [here](http://stackoverflow.com/questions/2984538/how-to-use-bytearrayoutputstream-and-dataoutputstream-simultaneously-java).
This pull request adds a `flush` method before calling toByteArray().

Signed-off-by: Wajih ul hassan wajih.lums@gmail.com


garagatyi:
Thank you for the contribution!
In my opinion flush is not mandatory here. 
DataOutputStream doesn't use bufferization at all and ObjectOutputStream use buffer for primitives only, but test doesn't write primitives into the stream.
On the other hand lack of flushing is not obvious and may introduce bug if someone who doesn't know this behavior changes the code.
So I would say it is better to add flushing than not.

garagatyi:
LGTM
@skabashnyuk Can you take a look.

skabashnyuk:
ok

PR #1010, opened 2016-04-11
Polishing che-core-commons-schedule

alexVengrovsk:

### DIFF
--- a/core/commons/che-core-commons-schedule/src/main/java/org/eclipse/che/commons/schedule/executor/CronExpression.java
+++ b/core/commons/che-core-commons-schedule/src/main/java/org/eclipse/che/commons/schedule/executor/CronExpression.java
@@ -19,16 +19,19 @@ package org.eclipse.che.commons.schedule.executor;
 
 import java.io.Serializable;
 import java.text.ParseException;
+import java.time.LocalDate;
+import java.time.temporal.TemporalAdjusters;
+import java.util.Collection;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.Locale;
 import java.util.Map;
 import java.util.SortedSet;
 import java.util.StringTokenizer;
 import java.util.TimeZone;
 import java.util.TreeSet;
+import java.util.stream.Collectors;
 
 /**
  * Provides a parser and evaluator for unix-like cron expressions. Cron 
@@ -211,8 +214,8 @@ public final class CronExpression implements Serializable, Cloneable {
     protected static final int NO_SPEC_INT = 98; // '?'
     protected static final Integer ALL_SPEC = ALL_SPEC_INT;
     protected static final Integer NO_SPEC = NO_SPEC_INT;
-    
-    protected static final Map<String, Integer> monthMap = new HashMap<String, Integer>(20);
+
+    protected static final Map<String, Integer> monthMap = new HashMap<>(20);
     protected static final Map<String, Integer> dayMap = new HashMap<String, Integer>(60);
     static {
         monthMap.put("JAN", 0);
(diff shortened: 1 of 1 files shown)


alexVengrovsk:
@skabashnyuk I also have some doubts about fielsd and methods scopes in this class. Many of them have _protected_  access modifiers, but could be changed to private. Let me know is it OK.

alexVengrovsk:
@skabashnyuk One more thing, the _findNextWhiteSpace()_ method doesn't work as it should. Is it need to be fixed?

garagatyi:
Please do not use wildcard imports. We don't use it.

core/commons/che-core-commons-schedule/src/main/java/org/eclipse/che/commons/schedule/executor/CronExpression.java
```diff
@@ -846,101 +840,35 @@ public String getCronExpression() {
...
-        buf.append("\n");
-        buf.append("lastdayOfMonth: ");
-        buf.append(lastdayOfMonth);
-        buf.append("\n");
-        buf.append("years: ");
-        buf.append(getExpressionSetSummary(years));
-        buf.append("\n");
-
-        return buf.toString();
+        return "seconds: " + getExpressionCollectionSummary(seconds) + "\n"
```
garagatyi, line 852: Can you explain why you've changed that? AFAIK usage of StringBuilder is more convinient

alexVengrovsk:
@garagatyi Ok. I will change this wildcard imports. Is it something else wrong in my PR?

core/commons/che-core-commons-schedule/src/main/java/org/eclipse/che/commons/schedule/executor/CronExpression.java
```diff
@@ -846,101 +840,35 @@ public String getCronExpression() {
...
-                buf.append(",");
-            }
-            buf.append(val);
-            first = false;
-        }
-
-        return buf.toString();
+        return collection.stream()
+                .map(Object::toString)
+                .collect(Collectors.joining(","));
```
garagatyi, line 876: Yeah, stream looks much better here!

core/commons/che-core-commons-schedule/src/main/java/org/eclipse/che/commons/schedule/executor/CronExpression.java
alexVengrovsk, line 852: @garagatyi I did it becouse string concatenation will be represented by compiler with a help os StringBuilder any way. So, it seams to me, that it was unnecessary code (creation StringBuilder object and all _append()_ methods and _toString()_ method).

alexVengrovsk:
@garagatyi Here is one more thing about this class, but I am not shure about this.

```
public static boolean isValidExpression(String cronExpression) {

        try {
            new CronExpression(cronExpression);
        } catch (ParseException pe) {
            return false;
        }

        return true;
    }
```

This kind of check is very coastly. Like we need to ceate whole object here. Maybe it is a good idea to think how to improve it?

alexVengrovsk:
@garagatyi  I have chended import block.

skabashnyuk:
Hello @myArea51 
Is this pr add some new feature. Or this is more code cleanup?
You might not notice but this code originally from quartz. I would like not to introduce changes without strong need.

alexVengrovsk:
Hi @skabashnyuk!  This PR is more about polishing. But take a close look at  the _findNextWhiteSpace()_ method. I have changed it, becouse in its previous state it worked wrong. Instead of returning an index of the next whitespace it returned a string length from the start position to the end of a string. I dont think, that it is ok. But it is up to you to decide.

skabashnyuk:
Can you do pr to this file https://github.com/quartz-scheduler/quartz/blob/master/quartz-core/src/main/java/org/quartz/CronExpression.java first

alexVengrovsk:
@skabashnyuk Ok. I will do it. So, do I needto close this one?

skabashnyuk:
Yes I think. When quartz approve your changes. We can accept this changes back with new pr.

alexVengrovsk:
Ok @skabashnyuk

PR #1011, opened 2016-04-11
CHE-941 temporarily disable copy to local path

stour:

@skabashnyuk @vparfonov


vparfonov:
ok

skabashnyuk:
ok

PR #1012, opened 2016-04-11
CHE-964 empty credentials at wizard opening and avoid storing empty credentials

stour:
@skabashnyuk @vparfonov


skabashnyuk:
ok

vparfonov:
ok

PR #1015, opened 2016-04-11
CHE-750 Remove snapshots on workspace removal

mkuznyetsov:
@evoevodin @garagatyi review please

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
@@ -279,6 +279,9 @@ public class WorkspaceService extends Service {
                                                                                         ConflictException,
                                                                                         ForbiddenException {
         ensureUserIsWorkspaceOwner(id);
+        if (!workspaceManager.getSnapshot(id).isEmpty()) {
+            machineManager.removeSnapshots(getCurrentUserId(), id);
+        }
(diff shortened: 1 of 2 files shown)


garagatyi:
LGTM

skabashnyuk:
ok

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -279,6 +279,9 @@ public void delete(@ApiParam("The workspace id") @PathParam("id") String id) thr
                                                                                         ConflictException,
                                                                                         ForbiddenException {
         ensureUserIsWorkspaceOwner(id);
+        if (!workspaceManager.getSnapshot(id).isEmpty()) {
```
sleshchenko, line 282: I understand that it is out scope of current issue. But maybe it would be better to rename getSnapshot to getSnapshots?

voievodin, line 282: I think no, the main idea of this method name is `get workspace snapshot` which is actually the list of machine snapshots, we also planed to create separate object describing workspace snapshot, that's why this name fits our needs

sleshchenko, line 282: :+1: Thanks for explanation

PR #1020, opened 2016-04-12
fixup! CHE-942 temporarily disable move to path

stour:
@skabashnyuk @vparfonov


sleshchenko:
Do you want to merge fixup to master?

stour:
Yes I discovered a bug caused by a commit previously merged into master: https://github.com/eclipse/che/pull/1003/commits/601a572586b98615ef0853222c0d05a7425584f1

sleshchenko:
But fixups are intended for joining theirs changes to original commit. For master it is impossible. I think It would better to make this commit as normal commit and describe fix of bug that you discovered. WDYT?

stour:
done

skabashnyuk:
ок

PR #1024, opened 2016-04-12
CHE-605: Use auto_snapshot/auto_restore attributes instead of params

voievodin:
@skabashnyuk, @akorneta i made these changes according to the [comment](https://jira.codenvycorp.com/browse/CHE-605? focusedCommentId=274077&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-274077) please review.

Workspace attributes contain `auto_snapshot=true` workspace will be snapshoted before stop operation is perfmored, Workspace attributes contain `auto_restore=true & workspace snapshot exists`, then workspace will be restored from that snapshot.

Note that these changes won't affect default behavior of the Workspace API, and also these changes do not add general configuration parameters, as mentioned in [this comment](https://jira.codenvycorp.com/browse/CHE-605?focusedCommentId=274034&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-274034).


skabashnyuk:
ok

akorneta:
ok

PR #1026, opened 2016-04-12
Optimizing "if" checks

alexVengrovsk:
There are few changes, which should spead up a little bit _if_ checks. 
There is, also, a substitution of _isNumber()_ method's body. Now method is protected from _NullPointerException_ and it doesn't need to create an arrey of _Class_ .


skabashnyuk:
ok @vparfonov ?

vparfonov:
ok

PR #1027, opened 2016-04-12
CHE-896 cleaning custom colors in SVN plugin

stour:

@skabashnyuk @vparfonov

### DIFF
--- a/core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/theme/Theme.java
+++ b/core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/theme/Theme.java
@@ -1459,10 +1459,12 @@ public interface Theme {
 
     /********************************************************************************************
      *
-     * Git output console
+     * VCS output console
      *
      ********************************************************************************************/
-    String gitConsoleStagedFilesColor();
-    String gitConsoleUnstagedFilesColor();
-    String gitConsoleErrorColor();
+    String vcsConsoleStagedFilesColor();
+    String vcsConsoleUnstagedFilesColor();
+    String vcsConsoleErrorColor();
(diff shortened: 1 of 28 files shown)


skabashnyuk:
ok. @vparfonov ?

core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/theme/Theme.java
```diff
@@ -1465,4 +1465,12 @@
     String gitConsoleStagedFilesColor();
     String gitConsoleUnstagedFilesColor();
     String gitConsoleErrorColor();
+
```
vparfonov: Looks here problem with styles, not good add SVN colors to the IDE API need find other solution. The same problem with GIT. @ashumilova what you think about it?

stour: What about having styles specific to VCS and not to Git and Subversion ?
So in this case, replace above and below lines with:

``` java
String vcsConsoleStagedFilesColor();
String vcsConsoleUnstagedFilesColor();
String vcsConsoleErrorColor();
String vcsConsoleModifiedFilesColor();
String vcsConsoleChangesLineNumbersColor();
```

vparfonov: better for me

stour: ok I will do this change

stour:
@vparfonov PR updated

vparfonov:
ok

PR #1029, opened 2016-04-12
CHE-721: Fix docker authentication error log on create workspace

AndrienkoAleksandr:


AndrienkoAleksandr:
Method "getInitialAuthConfig()" is unused, I deleted it. @garagatyi please review this pull request.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/InitialAuthConfig.java
```diff
@@ -11,6 +11,7 @@
 package org.eclipse.che.plugin.docker.client;
 
 import org.apache.commons.codec.binary.Base64;
+import org.eclipse.che.commons.annotation.Nullable;
```
garagatyi: It is not needed

garagatyi:
LGTM
@skabashnyuk Can you take a look

skabashnyuk:
ok

PR #1035, opened 2016-04-13
CHE-896 Fixing test errors due to status colors static block

stour:

@skabashnyuk @vparfonov


skabashnyuk:
ok

plugins/plugin-svn/che-plugin-svn-ext-ide/src/main/java/org/eclipse/che/plugin/svn/ide/diff/DiffPresenter.java
```diff
@@ -40,6 +38,7 @@
     private final NotificationManager                      notificationManager;
     private final SubversionClientService                  service;
     private final SubversionExtensionLocalizationConstants constants;
+    private final StatusColors                             statusColors;
```
sleshchenko, line 41: I think you can remove this variable

vparfonov:
ok

PR #1045, opened 2016-04-14
CHE-728

mmorhun:
CHE-728: Fix an error that occurred while writing line to the line consumer during che shutting down

### DIFF
--- a/wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
+++ b/wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
@@ -1011,8 +1011,23 @@ public class MachineManager {
     private void cleanup() {
         eventService.unsubscribe(machineCleaner);
 
-        boolean interrupted = false;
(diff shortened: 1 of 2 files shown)


mmorhun:
@evoevodin check please

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
```diff
@@ -1011,24 +1011,14 @@ private void createLogsDir() {
...
-            interrupted = true;
-            executor.shutdownNow();
-        }
 
         try {
             for (MachineImpl machine : machineRegistry.getMachines()) {
                 try {
                     destroy(machine.getId(), false);
+                } catch (NotFoundException ignore) {
+                    // if machine was destroyed before, it is ok
```
voievodin: I'd say that it's better to use comment like `it is ok, machine has been already destroyed`

voievodin:
Please change the comment and :+1:  good job

garagatyi:
LGTM

PR #1046, opened 2016-04-14
CHE-995: mark architecture field as beta

garagatyi:
@skabashnyuk @evoevodin Please review

### DIFF
--- a/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/MachineConfig.java
+++ b/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/MachineConfig.java
@@ -10,6 +10,8 @@
  *******************************************************************************/
 package org.eclipse.che.api.core.model.machine;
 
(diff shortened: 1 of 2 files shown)


core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/MachineConfig.java
```diff
@@ -60,6 +60,9 @@
 
     /**
      * Architecture of target machine. Default is 'linux_amd64'.
+     *
+     * <p/>This method is in beta. And can be removed in next versions.
```
sleshchenko, line 66: com.google.common.annotations.Beta ?

garagatyi, line 66: done

skabashnyuk:
ok

sleshchenko:
OK

PR #1055, opened 2016-04-15
The Workspace validator should allow the 'ssh' machine type

kaloyan-raev:


garagatyi:
Thank you for that fix! I've missed it.
But I think it is better to avoid hardcoding machine types at all.
To do that validator should inject machine instance providers and check if machine type is registered on validation.
Are you willing to contribute that?

kaloyan-raev:
It would be a good opportunity to learn more about Che.
Could you educate me what does it mean to "inject machine instance providers"?

garagatyi:
Sure.
Other classes are injected in CHE using Dependency injection. We use Google Guice for that.
[Here](https://github.com/eclipse/che/blob/ea6af309d67fa96d61d6ef84f86476078ebca824/wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java#L108-L108) we injects machine impl providers. Constructor should have annotation `@Inject` from javax.
Then you can get provider 
`final InstanceProvider instanceProvider = machineInstanceProviders.getProvider(machineConfig.getType());`
If provider of machine implementation with given type is not found method throws NotFoundException.

Feel free to ask for details if it is needed.

kaloyan-raev:
Thanks for the instructions. I believe I've done it correctly.

kaloyan-raev:
@garagatyi is there anything else necessary to have this PR merged?

garagatyi:
Yes, you did exactly what I meant!
But I have several improvements to this PR.
Return immutable collection from **getProviderTypes** method. Than you can use Guava's Joiner to join collection's elements Joiner.on(',').join(valuesList)
In tests use more declarative constructions:
instead of 

``` java
wsValidator = new DefaultWorkspaceValidator(machineInstanceProviders);
MachineInstanceProviders machineInstanceProviders = Mockito.mock(MachineInstanceProviders.class);
```

use fields 

``` java
@Mock
MachineInstanceProviders machineInstanceProviders
@InjectMocks
private DefaultWorkspaceValidator wsValidator;
```

kaloyan-raev:
So be it. Please, check my latest commit.

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineInstanceProviders.java
```diff
@@ -44,9 +47,28 @@ public MachineInstanceProviders(Set<InstanceProvider> instanceProviders) {
...
     public InstanceProvider getProvider(String machineType) throws NotFoundException {
-        if (instanceProviders.containsKey(machineType)) {
+        if (hasProvider(machineType)) {
             return instanceProviders.get(machineType);
         }
         throw new NotFoundException(String.format("Can't find machine provider for unsupported machine type '%s'", machineType));
     }
+    
+    /**
+     * Checks if an {link InstanceProvider} implementation of the given machine type exists
```
garagatyi: add at `@` sign to before **link** word

wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/workspace/server/DefaultWorkspaceValidatorTest.java
```diff
@@ -32,19 +40,27 @@
...
  * @author Alexander Reshetnyak
  */
+@Listeners(MockitoTestNGListener.class)
 public class DefaultWorkspaceValidatorTest {
-
-    private WorkspaceValidator wsValidator;
-
-    @BeforeClass
+	
+	@Mock
```
garagatyi: please fix indentation

kaloyan-raev:
link annotation and indentation are fixed.

garagatyi:
LGTM
@skabashnyuk Can you take a look?

skabashnyuk:
ok

skabashnyuk:
wait for 4.3 in master before merge

PR #1056, opened 2016-04-15
Allow factory to contain projects with keepDir function

mkuznyetsov:
Projects that are created with keepDir option have no source type or location.
So the source validation or project in factory will be omitted, if such project is located in another project

### DIFF
--- a/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectManager.java
+++ b/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectManager.java
@@ -333,6 +333,12 @@ public final class ProjectManager {
         List<? extends ProjectConfig> projects = workspaceConfig.getProjects();
         for (ProjectConfig project : projects) {
             if (normalizePath.equals(project.getPath())) {
+                // TODO Needed for factory project importing with keepDir. It needs to find more appropriate solution
+                List<String> innerProjects = projectRegistry.getProjects(normalizePath);
+                for (String innerProject : innerProjects) {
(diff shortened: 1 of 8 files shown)


benoitf:
it seems commit is missing related issue in the message

mkuznyetsov:
will fix that during merge

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectManager.java
```diff
@@ -333,6 +333,12 @@ public RegisteredProject importProject(String path, SourceStorage sourceStorage,
         List<? extends ProjectConfig> projects = workspaceConfig.getProjects();
         for (ProjectConfig project : projects) {
             if (normalizePath.equals(project.getPath())) {
+                // TODO Needed for factory project importing with keepDir. It needs to find more appropriate solution
+                List<String> projects1 = projectRegistry.getProjects(normalizePath);
+                for (String s : projects1) {
```
skabashnyuk: projects1  - nasty name. Can you come up new one?

skabashnyuk:
other ok

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectManager.java
sleshchenko: s -> project?

mkuznyetsov: fixed

PR #1062, opened 2016-04-18
Fixing 'StaticAccessedFromInstance' warnings

Wajihulhassan:
According to [Error Prone](http://errorprone.info/bugpattern/StaticAccessedFromInstance) website

> A static variable or method should never be accessed from an instance. This hides the fact that the variable or method is static and does not depend on the value of the object instance on which this variable or method is being invoked.

This pull request fixes this issue/warning.

Signed-off-by: Wajih ul hassan wajih.lums@gmail.com


garagatyi:
LGTM
@vparfonov Please take a look

PR #1064, opened 2016-04-18
CHE-362. Add different display modes for StatusNotification

RomanNikitenko:
Define next statuses for StatusNotification:
1. FLOAT_MODE -  notification will be float as balloon during the notification is in progress status 
2.  EMERGE_MODE - notification will be emerge as balloon only when status is changed
3. NOT_EMERGE_MODE - Notification won't be emerge at all


vparfonov:
ok

PR #1068, opened 2016-04-18
CHE-999 SSH machines: handle exceptions on connections: properly subscribe to events

vitaliy-guliy:


vparfonov:
ok

PR #1071, opened 2016-04-18
Fix websocket terminal launching in ssh machine

garagatyi:
@skabashnyuk @vparfonov @riuvshin Please review


skabashnyuk:
How?

garagatyi:
If you ment how this PR fix the problem then answer is my mistake when I implemented that feature. At that point it worked because of several coincedences and I decided that it works, but later I discovered that it doesn't work mostly.

garagatyi:
So this PR changes permissions setting for websocket terminal files

skabashnyuk:
Ok

riuvshin:
LGTM

PR #1073, opened 2016-04-19
CHE-946: [dashboard] improve workspace details page with error status

akurinnoy:
**_1 Upvote**_ - improve workspace details page with error status
- fix loading multiple workspaces simultaneously

### DIFF
--- a/dashboard/src/app/ide/ide.service.js
+++ b/dashboard/src/app/ide/ide.service.js
@@ -40,7 +40,7 @@ class IdeSvc {
     this.ideParams = new Map();
 
     this.currentStep = 0;
-    this.selectedWorkspace = null;
+    this.lastWorkspace = null;
 
     this.listeningChannels = [];
 
@@ -60,6 +60,9 @@ class IdeSvc {
       step.hasError = false;
     });
 
+    if (this.lastWorkspace) {
+      this.cleanupChannels(this.lastWorkspace.id);
+    }
   }
 
   getStepText(stepNumber) {
@@ -84,10 +87,6 @@ class IdeSvc {
     return this.$rootScope.ideIframeLink && (this.$rootScope.ideIframeLink !== null);
   }
 
-  setSelectedWorkspace(selectedWorkspace) {
-    this.selectedWorkspace = selectedWorkspace;
-  }
-
   handleError(error) {
     if (error.data.message) {
       this.steps[this.currentStep].logs += '\n' + error.data.message;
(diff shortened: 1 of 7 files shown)


dashboard/src/app/ide/ide.service.js
```diff
@@ -227,26 +240,26 @@ class IdeSvc {
     return startWorkspacePromise;
   }
 
-  connectToExtensionServer(websocketURL, workspaceId) {
+  connectToExtensionServer(workspaceId, websocketURL) {
```
benoitf: maybe we shouldn't change the order of params to avoid potential failures

dashboard/src/app/ide/ide.service.js
```diff
@@ -96,38 +99,48 @@ class IdeSvc {
     this.$log.error(error);
   }
 
-  startIde(noIdeLoader) {
+  startIde(workspace, noIdeLoader) {
```
benoitf, line 98: I don't get why there is a new workspace parameter as there is already a class attribute selectedWorkspace

dashboard/src/app/workspaces/workspace-details/workspace-details.controller.js
```diff
@@ -132,20 +131,24 @@ export class WorkspaceDetailsCtrl {
 
   runWorkspace() {
     this.showShowMore = true;
+    delete this.errorMessage;
 
     this.ideSvc.init();
     this.ideSvc.setSelectedWorkspace(this.workspaceDetails);
     this.$rootScope.loadingIDE = false;
-    let promise = this.ideSvc.startIde(true);
+    let promise = this.ideSvc.startIde(this.workspaceDetails, true);
```
benoitf, line 138: why do we need to give parameter that is given 4 lines before ?

dashboard/src/app/ide/ide.service.js
akurinnoy: you're right, I'll revert it

dashboard/src/app/ide/ide.service.js
akurinnoy, line 98: selectedWorkspace contains only the lastest workspace which is starting. I need workspace parameter to use it in chain of promises as current workspace.

dashboard/src/app/workspaces/workspace-details/workspace-details.controller.js
akurinnoy, line 138: setSelectedWorkspace isn't needed now, so I'll remove it

ashumilova:
+1

PR #1082, opened 2016-04-19
CHE-995: discover ssh machines architecture automatically

garagatyi:
Discover ssh machines architecture automatically.
Remove beta field in machine conf that provides architecture.
Move server-side classes to appropriate package.
Removed workarounds in ssh client and properly wait until command finishes.


garagatyi:
@skabashnyuk @vparfonov Please review

plugins/plugin-ssh-machine/src/main/java/org/eclipse/che/plugin/machine/ssh/SshMachineImplTerminalLauncher.java
```diff
@@ -61,39 +72,106 @@ public String getMachineType() {
...
+                                                               null);
+        ListLineConsumer lineConsumer = new ListLineConsumer();
+        getUnameOutput.start(lineConsumer);
+        String unameOutput = lineConsumer.getText();
+        Matcher matcher = UNAME_OUTPUT.matcher(unameOutput);
+        if (matcher.matches()) {
+            String os = matcher.group("os").toLowerCase();
+            String arch = matcher.group("architecture").toLowerCase();
+            StringBuilder result = new StringBuilder();
+            if (os.contains("linux")) {
```
skabashnyuk, line 131: Shall we add this string constants to some enum?

plugins/plugin-ssh-machine/src/main/java/org/eclipse/che/plugin/machine/ssh/jsch/JschSshClient.java
```diff
@@ -236,64 +237,47 @@ private String getAbsolutePath(String path) throws MachineException {
...
-            // todo fix that
-            // this method is used to run commands such as "pwd" or "mkdir -p /some/path"
-            // And we need to wait to check if it return 0. If command is still running method returns -1.
-            try {
-                // workaround exit code -1
-                Thread.sleep(1000);
-            } catch (InterruptedException ignored) {
-            }
-            return exec.getExitStatus();
+            IoUtil.readStream(inStream);
```
skabashnyuk, line 250: why we need to read this streams?

skabashnyuk:
other ok

plugins/plugin-ssh-machine/src/main/java/org/eclipse/che/plugin/machine/ssh/SshMachineImplTerminalLauncher.java
garagatyi, line 131: I don't see the reason to add constant with value equal to name

PR #1085, opened 2016-04-19
CHE-243: Add paging support

voievodin:
Add `Page` object + util methods for managing _Link_ header.

@skabashnyuk, @akorneta, @garagatyi please review

### DIFF
--- a/core/che-core-api-core/pom.xml
+++ b/core/che-core-api-core/pom.xml
@@ -43,6 +43,10 @@
             <artifactId>guice-multibindings</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.jayway.restassured</groupId>
+            <artifactId>rest-assured</artifactId>
+        </dependency>
+        <dependency>
             <groupId>javax.annotation</groupId>
             <artifactId>javax.annotation-api</artifactId>
         </dependency>
--- a/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/Page.java
+++ b/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/Page.java
@@ -1,0 +1,312 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.api.core;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Function;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.toList;
+
+/**
+ * Defines paged result of data selection, it is rather dynamic data window than
+ * regular page, as it refers to the specified selection region based on the items before
+ * and the page size, but not on the page number and page size.
+ *
+ * <p>Examples:
+ *
+ * <p>Regular page.<br>
+ * For page input arguments {@code skipItems = 3}, {@code pageSize = 3} and {@code totalCount = 10}
+ * the page is normalized and will refer to the second page which is(item4, item5, item6):
+ * <pre>
+ *     item1.                      <- previous page start      <- first page start
+ *     item2.
+ *     item3.                      <- previous page end        <- first page end
+ *     item4. <- page start
+ *     item5.
+ *     item6. <- page end
+ *     item7.                      <- next page start
+ *     item8.
+ *     item9.                      <- next page end
+ *     item.10                                                  <- last page start/end
+ * </pre>
+ *
+ * <ul>Result page:
+ * <li>Non-empty</li>
+ * <li>Contains 3 items(item4, item5, item6)</li>
+ * <li>Has the previous page(item1, item2, item3)
+ * <li>Has the next page(item7, item8, item9)</li>
+ * </ul>
+ *
+ *
+ * <p>Data window.<br>
+ * For page input arguments {@code skipItems = 2}, {@code pageSize = 3} and {@code totalCount = 7}
+ * the page will refer to the following data window:
+ * <pre>
+ *     item1.                   <- first page start
+ *     item2.
+ *     item3. <- page start     <- first page end
+ *     item4.
+ *     item5. <- page end
+ *     item6.
+ *     item7.                    <- last page start/end
+ * </pre>
+ *
+ * <ul>Result page:
+ * <li>Non-empty</li>
+ * <li>Contains 3 items(item3, item4, item5)</li>
+ * <li>Doesn't have the previous page(because page refers to the elements which
+ * are partially present in the first page(item3) and the second page(item4, item5)</li>
+ * <li>Doesn't have the next page(the reason is the same to previous statement)</li>
+ * </ul>
+ *
+ * <p>Note that all {@code Page} instances perform reference
+ * calculations based on the given {@code itemsBefore} and {@code pageSize}
+ * values which means that implementor is responsible for providing correct bounds
+ * and data management.
+ *
+ * <p>The instances of this class are <b>NOT thread safe</b>.
+ *
+ * @param <ITEM_T>
+ *         the type of the page items
+ * @author Yevhenii Voevodin
+ */
+public class Page<ITEM_T> {
+
+    private final int          pageSize;
+    private final long         itemsBefore;
+    private final long         totalCount;
+    private final List<ITEM_T> items;
+
+    /**
+     * Creates a new page.
+     *
+     * @param items
+     *         page items
+     * @param itemsBefore
+     *         items count before this page
+     * @param pageSize
+     *         page size
+     * @param totalCount
+     *         count of all the items
+     * @throws NullPointerException
+     *         when {@code items} collection is null
+     * @throws IllegalArgumentException
+     *         when {@code itemsBefore} is negative
+     * @throws IllegalArgumentException
+     *         when {@code pageSize} is non-positive
+     * @throws IllegalArgumentException
+     *         when {@code totalCount} is negative
+     */
+    public Page(Collection<? extends ITEM_T> items, long itemsBefore, int pageSize, long totalCount) {
+        requireNonNull(items, "Required non-null items");
+        this.items = new ArrayList<>(items);
+        checkArgument(itemsBefore >= 0, "Required non-negative value of items before");
+        this.itemsBefore = itemsBefore;
+        checkArgument(pageSize > 0, "Required positive value of page size");
+        this.pageSize = pageSize;
+        checkArgument(totalCount >= 0, "Required non-negative value of total items");
+        this.totalCount = totalCount;
+    }
+
+    /** Returns true whether this page doesn't contain items, returns false if it does. */
+    public boolean isEmpty() {
+        return items.isEmpty();
+    }
+
+    /**
+     * Returns true when the current page has the next page,
+     * otherwise when the page is the last page false will be returned.
+     */
+    public boolean hasNextPage() {
+        return getNumber() != -1 && itemsBefore + pageSize < totalCount;
+    }
+
+    /**
+     * Returns true when this page has the previous page,
+     * otherwise when the page is the first page false will be returned.
+     */
+    public boolean hasPreviousPage() {
+        return getNumber() != -1 && itemsBefore != 0;
+    }
+
+    /**
+     * Returns an {@link Optional} describing reference to the next page,
+     * or the empty {@code Optional} when this page is the first one.
+     */
+    public Optional<PageRef> getNextPageRef() {
+        if (!hasNextPage()) {
+            return Optional.empty();
+        }
+        return Optional.of(new PageRef(itemsBefore + pageSize, pageSize));
+    }
+
+    /**
+     * Returns an {@link Optional} describing reference to the previous page,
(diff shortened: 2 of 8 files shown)


core/che-core-api-core/pom.xml
```diff
@@ -43,6 +43,11 @@
             <artifactId>guice-multibindings</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.jayway.restassured</groupId>
+            <artifactId>rest-assured</artifactId>
+            <version>2.4.0</version>
```
skabashnyuk: remove version

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/Page.java
```diff
@@ -0,0 +1,307 @@
...
+ * and data management.
+ *
+ * <p>The instances of this class are <b>NOT thread safe</b>.
+ *
+ * @param <ITEM_T>
+ *         the type of the page items
+ * @author Yevhenii Voevodin
+ */
+public class Page<ITEM_T> {
+
```
skabashnyuk, line 87: I think there is a common way to call such parameters with one letter. For instance Page<I> wdyt?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/Page.java
```diff
@@ -0,0 +1,307 @@
...
+     *         when {@code items} collection is null
+     * @throws IllegalArgumentException
+     *         when {@code itemsBefore} is negative
+     * @throws IllegalArgumentException
+     *         when {@code pageSize} is non-positive
+     * @throws IllegalArgumentException
+     *         when {@code totalCount} is negative
+     */
+    public Page(Collection<? extends ITEM_T> items, long itemsBefore, int pageSize, long totalCount) {
+        this.items = new ArrayList<>(requireNonNull(items, "Required non-null items"));
```
skabashnyuk: can we put checks before assignments?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/PagingUtil.java
```diff
@@ -0,0 +1,101 @@
...
+import static java.util.stream.Collectors.joining;
+
+/**
+ * Provides useful methods for working with {@link Page} instances
+ * and pageable uris management.
+ *
+ * @author Yevhenii Voevodin
+ */
+public final class PagingUtil {
+
```
skabashnyuk, line 37: Can we move this functionality to org.eclipse.che.api.core.Page class?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/PagingUtil.java
```diff
@@ -0,0 +1,101 @@
...
+    private static final String  LINK_HEADER_SEPARATOR = ", ";
+    private static final Pattern LINK_HEADER_REGEX     = Pattern.compile("<(.+)>;.*rel=\"(.+)\".*");
```
skabashnyuk: Do we have  test for this regexp?

garagatyi: please add comment that describes pattern content also

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/Page.java
```diff
@@ -0,0 +1,307 @@
...
+     * Gets the page items and maps them with given {@code mapper}.
+     *
+     * @param mapper
+     *         items mapper
+     * @param <R>
+     *         the type of the result items
+     * @return the list of mapped items
+     */
+    public <R> List<R> getItems(Function<? super ITEM_T, ? extends R> mapper) {
+        return items.stream()
```
akorneta, line 242: what if mapper is null

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/Page.java
```diff
@@ -0,0 +1,307 @@
...
+     *
+     * @param container
+     *         collection which is used to fill result into
+     * @param <COL_T>
+     *         collection type
+     * @return given collection instance {@code container} with items filled
+     * @throws NullPointerException
+     *         when {@code container} is null
+     */
+    public <COL_T extends Collection<? super ITEM_T>> COL_T fill(COL_T container) {
```
akorneta, line 268: we should document the case when immutable collections are as argument

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/PagingUtil.java
voievodin, line 37: I think no, because it is strongly REST related, while page is data related.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/PagingUtil.java
voievodin: I tested `PagingUtil.parseLinkHeader` along with this regex

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/Page.java
voievodin, line 87: i don't think it's a problem actually, i used `<ITEM_T>` parameter name because it is simpler to recognize the type in the following methods:

``` java
    public <R> List<R> getItems(Function<? super ITEM_T, ? extends R> mapper) {
        return items.stream()
                    .map(mapper::apply)
                    .collect(toList());
    }
// ...

    public <COL_T extends Collection<? super ITEM_T>> COL_T fill(COL_T container) {
        requireNonNull(container, "Required non-null items container");
        container.addAll(items);
        return container;
    }
```

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/Page.java
```diff
@@ -0,0 +1,312 @@
...
+    }
+
+    /**
+     * Returns page number starting from 1.
+     *
+     * <p>If the page is not regular page(it refers rather to the
+     * data window than to the certain page(e.g. skip=2, pageSize=4))
+     * then this method returns -1.
+     */
+    public long getNumber() {
```
vzhukovs, line 199: I'd rather call this method smth like `getPageNumber()`, because you already have _getPrevious**Page**Ref_, _hasNext**Page**_. Look, in the code you'll have smth like:

```
int index = someVar.getNumber();
```

with this, code becomes clearer:

```
int index = someVar.getPageNumber();
```

wdyt?

akorneta:
ok

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/Page.java
voievodin, line 199: i think it is clearer in your examples because you used `someVar` name for variable, it looks okay when variable name somehow related to the _Page_ e.g.`page.getNumber()`.

voievodin, line 199: I used _Page_ word in methods which are related to the pages different from the current one, 
e.g. it is still `getItems` not `getPageItems` it is still `getTotalItemsCount` not `getTotalPageItemsCount`, so it makes sense for me.

voievodin:
@garagatyi, @skabashnyuk so what is the result of the review?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/Page.java
vzhukovs, line 199: Okay, thx for the explanation.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/Service.java
```diff
@@ -70,7 +79,66 @@ protected ServiceDescriptor createServiceDescriptor() {
...
+     * @param qpMap
+     *         query parameters map, if multiple query params needed then
+     *         {@link #createLinkHeader(Page, String, ListMultimap, Object...)}
+     *         method should be used instead
+     * @param pathParams
+     *         path param values like {f@link UriBuilder#build(Object...)} method arguments
+     */
+    protected String createLinkHeader(Page<?> page,
+                                      String method,
+                                      Map<String, Object> qpMap,
```
garagatyi: type already says that it is map, but qp is not obvious. I would prefer querieParams

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/Service.java
```diff
@@ -70,7 +79,66 @@ protected ServiceDescriptor createServiceDescriptor() {
...
+        return createLinkHeader(page, method, emptyMap(), pathParams);
+    }
+
+    /**
+     * This method is the same to {@link #createLinkHeader(Page, String, Map, Object...)}
+     * except of receiving multivalued query parameters.
+     */
+    protected String createLinkHeader(Page<?> page,
+                                      String method,
+                                      ListMultimap<String, Object> qpMap,
```
garagatyi: I would prefer parameter name queryParams

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/PagingUtil.java
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * Provides useful methods for working with {@link Page} instances
+ * and pageable uris management.
+ *
+ * @author Yevhenii Voevodin
+ */
+public final class PagingUtil {
+
+    private static final String  LINK_HEADER_SEPARATOR = ", ";
```
garagatyi: 2 spaces after type

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/Page.java
skabashnyuk, line 87: ок

skabashnyuk:
ок

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/PagingUtil.java
```diff
@@ -0,0 +1,106 @@
...
+     * @param uri
+     *         the uri which is used for adding {@code skipCount} & {@code maxItems} query parameters
+     * @return 'Link' header value
+     * @throws NullPointerException
+     *         when either {@code page} or {@code uri} is null
+     */
+    public static String createLinkHeader(Page<?> page, URI uri) {
+        requireNonNull(page, "Required non-null page");
+        requireNonNull(uri, "Required non-null uri");
+        final ArrayList<Pair<String, Page.PageRef>> pageRefs = new ArrayList<>(4);
```
garagatyi: As it is short living collection and in case when there is next and previous pages it is unneeded to create collection that will grow. So please increase the size of this collection. In case when there is no next/prev pages this collection will be ripped fast enough.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/PagingUtil.java
```diff
@@ -0,0 +1,106 @@
...
+     */
+    public static Map<String, String> parseLinkHeader(String linkHeader) {
+        if (isNullOrEmpty(linkHeader)) {
+            return emptyMap();
+        }
+        final Map<String, String> res = new HashMap<>();
+        for (String part : linkHeader.split(LINK_HEADER_SEPARATOR)) {
+            final Matcher matcher = LINK_HEADER_REGEX.matcher(part);
+            if (matcher.matches()) {
+                res.put(matcher.group(2), matcher.group(1));
```
garagatyi: have you considered named groups? They looks better for reading of code.

garagatyi:
@evoevodin Please describe how this functionality can be used in GWT/Angular clients. Specifically I'm interesting in snippet that shows how to parse response in page object and then send requests to next/prev pages.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/PagingUtil.java
voievodin: nice hint, i will change the code to use named groups

voievodin:
@garagatyi i followed [web linking spec](https://tools.ietf.org/html/rfc5988), so i think that it is not a problem to parse it even without using external libraries, but anyway here is  [angular tool](https://github.com/igorissen/angular-link-header-parser) so code snippet may be taken from there, or even first public [gist i found](https://gist.github.com/niallo/3109252), i think that gwt client may reuse `PagingUtil.parseLinkHeader` by including it somewhere in gwt configuration.

>  Specifically I'm interesting in snippet that shows how to parse response in page object and then send requests to next/prev pages.

I think it is just native javascript which retrieves response header and parses it, i'm sorry but i won't provide any code here, i think it is kind of common task for client, and it is out of the issue scope.

voievodin:
@garagatyi i applied changes related to your comments, can you please review again

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/Page.java
```diff
@@ -0,0 +1,312 @@
...
+     *         when {@code container} is null
+     */
+    public <COL_T extends Collection<? super ITEM_T>> COL_T fill(COL_T container) {
+        requireNonNull(container, "Required non-null items container");
+        container.addAll(items);
+        return container;
+    }
+
+    /** Represents page reference as a combination of {@code skipItems & pageSize}. */
+    public static class PageRef {
```
garagatyi, line 275: Why you decided to make that class inner?

voievodin, line 275: Because it is strongly related to the `Page` + there is no reason to create `PageRef` from the other place.

garagatyi:
LGTM

riuvshin:
why this merged to 4.2.0 ? it was planned for 4.3.0

PR #1089, opened 2016-04-20
CODENVY-350: Set up workspaces to work behind the proxy

mmorhun:


garagatyi:
LGTM
@skabashnyuk Can you take a look

skabashnyuk:
ок

riuvshin:
should not be ok, DockerProxyModule class does not have license.

PR #1099, opened 2016-04-20
Code cleanup

alexVengrovsk:
This is just a code cleanup. But it helps to reduse verbosity of this huge class.

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
@@ -114,11 +114,7 @@ public class DockerConnector {
         try (DockerConnection connection = connectionFactory.openConnection(dockerDaemonUri)
                                                             .method("GET")
                                                             .path("/info")) {
-            final DockerResponse response = connection.request();
-            final int status = response.getStatus();
-            if (OK.getStatusCode() != status) {
-                throw getDockerException(response);
-            }
+            final DockerResponse response = getDockerResponse(connection);
             return parseResponseStreamAndClose(response.getInputStream(), org.eclipse.che.plugin.docker.client.json.SystemInfo.class);
         } catch (JsonParseException e) {
             throw new IOException(e.getLocalizedMessage(), e);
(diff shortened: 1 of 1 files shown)


skabashnyuk:
@garagatyi ?

garagatyi:
Hi Alex. 
Thank you for that contribution!
Your PR forces me to take a look at Docker client one more time and think more about its design.
I think we can improve part of client which you addresses in this PR, but this change is not what I would want to accept.
The problem with this PR is that method getResponse does some things implicitly.
It gets response from connection, it is OK, it check response code, parse error message and throws exception, which is not OK.
Method name says nothing about response code neither parsing and throwing exception.

What I would like to introduce instead is:
- add default response status exception filter to [connection factory](https://github.com/eclipse/che/blob/867ec652424cd23a587f164f6acf3e8e44158bdf/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/connection/DockerConnectionFactory.java).
- add ability to set another status code exception filter in connection factory (by inheritance?) and docker connection.
  This filter should automatically check status code of connection on response retrieval and throw exception if status is not 2xx.
  In that case code will look like:

``` java
final DockerResponse response = connection.request();
return parseResponseStreamAndClose(response.getInputStream(), ContainerCommited.class).getId();
```

So private methods won't be needed.

@skabashnyuk @evoevodin @sleshchenko What do you think about such an idea?

garagatyi:
By the way @alexVengrovsk we have another PR with enhancements in docker client. If you are interested take a look at #952

skabashnyuk:
>  What do you think about such an idea?

+100

sleshchenko:
Sounds good

voievodin:
+0, i don't think that response status filter is a good solution, because then connection behaves the same for all the methods we have, so my point is that response should bring its status and status may be different from the 2xx, so it will be simple to build custom logic based on the response code without creating another connection/factory which won't filter 2xx responses. Also i think that status check is implementation logic and private method will be exactly what we need here, e.g:

``` java
final DockerResponse response = checkIsOk(connection.request());
return parseResponseStreamAndClose(response.getInputStream(), ContainerCommited.class).getId();
```

garagatyi:
@evoevodin I don't like neither **getResponse** method from this PR nor **checkIsOk** method for suggested.

garagatyi:
@evoevodin We can add method to docker connection **disableResponseCodeValidation**.
If developer relies on non 200 response codes he can use that method to disable default behavior.
In most cases validation of response code and parsing error is useless and boilerplate.

alexVengrovsk:
@garagatyi I will think about your idea about adding a filter to the connection factory. But it will take a while.

alexVengrovsk:
@garagatyi I will think about your ideas. And I think that I can implement some suitable solution. But it will take a while

PR #1104, opened 2016-04-21
CODENVY-212: Add DockerContainerCleaner for clean up docker containers

AndrienkoAleksandr:

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
@@ -34,6 +34,7 @@ import org.eclipse.che.plugin.docker.client.json.ContainerCommitted;
 import org.eclipse.che.plugin.docker.client.json.ContainerConfig;
 import org.eclipse.che.plugin.docker.client.json.ContainerCreated;
 import org.eclipse.che.plugin.docker.client.json.ContainerExitStatus;
+import org.eclipse.che.plugin.docker.client.json.ContainerListEntry;
 import org.eclipse.che.plugin.docker.client.json.ContainerInfo;
 import org.eclipse.che.plugin.docker.client.json.ContainerProcesses;
 import org.eclipse.che.plugin.docker.client.json.Event;
@@ -69,6 +70,7 @@ import org.eclipse.che.plugin.docker.client.params.StopContainerParams;
 import org.eclipse.che.plugin.docker.client.params.TagParams;
 import org.eclipse.che.plugin.docker.client.params.TopParams;
 import org.eclipse.che.plugin.docker.client.params.WaitContainerParams;
+import org.eclipse.che.plugin.docker.client.params.ListContainersParams;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -81,6 +83,7 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
+import java.lang.reflect.Type;
 import java.net.URI;
 import java.net.URLEncoder;
 import java.nio.file.Files;
@@ -106,6 +109,7 @@ import static javax.ws.rs.core.Response.Status.OK;
  * @author Alexander Garagatyi
  * @author Anton Korneta
  * @author Mykola Morhun
+ * @author Alexander Andrienko
  */
 @Singleton
 public class DockerConnector {
@@ -185,7 +189,48 @@ public class DockerConnector {
             if (OK.getStatusCode() != response.getStatus()) {
                 throw getDockerException(response);
             }
-            return parseResponseStreamAsListAndClose(response.getInputStream());
+            return parseResponseStreamAsListAndClose(response.getInputStream(), new TypeToken<List<Image>>() {}.getType());
+        } catch (JsonParseException e) {
+            throw new IOException(e.getLocalizedMessage(), e);
+        }
+    }
+
+    /**
+     * Method returns list of docker containers, include non-running ones.
+     *
+     * @throws IOException
+     *         when problems occurs with docker api calls
+     */
+    public List<ContainerListEntry> listContainers() throws IOException {
+        return listContainers(ListContainersParams.create().withAll(true));
+    }
+
+    /**
+     * Method returns list of docker containers which was filtered by {@link ListContainersParams}
+     *
+     * @throws IOException
+     *         when problems occurs with docker api calls
+     */
+    public List<ContainerListEntry> listContainers(ListContainersParams params) throws IOException {
+        final Filters filters = params.getFilters();
+
+        try (DockerConnection connection = connectionFactory.openConnection(dockerDaemonUri)
+                                                            .method("GET")
+                                                            .path("/containers/json")) {
+            addQueryParamIfNotNull(connection, "all", params.isAll());
+            addQueryParamIfNotNull(connection, "size", params.isSize());
+            addQueryParamIfNotNull(connection, "limit", params.getLimit());
+            addQueryParamIfNotNull(connection, "since", params.getSince());
+            addQueryParamIfNotNull(connection, "before", params.getBefore());
+            if (filters != null) {
+                connection.query("filters", urlPathSegmentEscaper().escape(JsonHelper.toJson(filters)));
+            }
+            DockerResponse response = connection.request();
+            final int status = response.getStatus();
+            if (OK.getStatusCode() != status) {
+                throw getDockerException(response);
+            }
+            return parseResponseStreamAsListAndClose(response.getInputStream(), new TypeToken<List<ContainerListEntry>>() {}.getType());
         } catch (JsonParseException e) {
             throw new IOException(e.getLocalizedMessage(), e);
         }
@@ -1610,11 +1655,11 @@ public class DockerConnector {
 
     @VisibleForTesting
     @SuppressWarnings("unchecked")
-    <T> List<T> parseResponseStreamAsListAndClose(InputStream inputStream) throws IOException, JsonParseException {
+    <T> List<T> parseResponseStreamAsListAndClose(InputStream inputStream, Type type) throws IOException, JsonParseException {
         try (InputStream responseStream = inputStream) {
             return (List<T>)JsonHelper.fromJson(responseStream,
                                                 List.class,
-                                                new TypeToken<List<T>>() {}.getType(),
+                                                type,
                                                 FIRST_LETTER_LOWERCASE);
         }
     }
--- a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/ContainerListEntry.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/ContainerListEntry.java
@@ -1,0 +1,173 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.plugin.docker.client.json;
+
+import java.util.Arrays;
+import java.util.Map;
+
+/**
+ * Defines information about container from list containers.
+ *
+ * @author Alexander Andrienko
+ */
+public class ContainerListEntry {
+    private String              id;
+    private String[]            names;
+    private String              image;
+    private String              imageID;
+    private String              command;
+    private long                created;
+    private String              status;
+    private ContainerPort[]     ports;
+    private Map<String, String> labels;
+    private int                 sizeRw;
+    private int                 sizeRootFs;
+
+    /**
+     * Returns unique container identifier
+     */
+    public String getId() {
+        return id;
+    }
+
+    public void setId(String id) {
+        this.id = id;
+    }
+
+    /**
+     * Returns docker container name. Earlier docker api allowed creation additional names, that why this method return String[].
+     * But for now container name is the first element of this array. Read more https://github.com/docker/docker/issues/12538
+     */
+    public String[] getNames() {
+        return names;
+    }
+
+    public void setNames(String[] names) {
+        this.names = names;
+    }
+
+    /**
+     * Returns name for image, which was used for creation the container
+     */
+    public String getImage() {
+        return image;
+    }
+
+    public void setImage(String image) {
+        this.image = image;
+    }
+
+    /**
+     * Returns id for image, which was used for creation the container
+     */
+    public String getImageID() {
+        return imageID;
+    }
+
+    public void setImageID(String imageID) {
+        this.imageID = imageID;
+    }
+
+    /**
+     * Returns time creation of the container in Unix time format(number of milliseconds since Unix epoch January 1 1970 to data
+     * creation container)
+     */
+    public long getCreated() {
+        return created;
+    }
+
+    public void setCreated(long created) {
+        this.created = created;
+    }
+
+    /**
+     * Returns command which container performs
+     */
+    public String getCommand() {
+        return command;
+    }
+
+    public void setCommand(String command) {
+        this.command = command;
+    }
+
+    /**
+     * Returns container status in declarative form, f.e "Up 6 minutes", "Exit 0"
+     */
+    public String getStatus() {
+        return status;
+    }
+
+    public void setStatus(String status) {
+        this.status = status;
+    }
+
+    /**
+     * Returns published a container's port(s) to the host. See more {@link ContainerPort}
+     */
+    public ContainerPort[] getPorts() {
+        return ports;
+    }
+
+    public void setPorts(ContainerPort[] ports) {
+        this.ports = ports;
+    }
+
+    /**
+     * Returns metadata of the container
+     */
+    public Map<String, String> getLabels() {
+        return labels;
+    }
+
+    public void setLabels(Map<String, String> labels) {
+        this.labels = labels;
+    }
+
+    /**
+     * Returns the real size of the container.
+     */
+    public int getSizeRw() {
+        return sizeRw;
+    }
+
+    public void setSizeRw(int sizeRw) {
+        this.sizeRw = sizeRw;
+    }
+
+    /**
+     *  Returns the virtual size of the container
+     */
+    public int getSizeRootFs() {
+        return sizeRootFs;
+    }
+
+    public void setSizeRootFs(int sizeRootFs) {
+        this.sizeRootFs = sizeRootFs;
+    }
+
+    @Override
+    public String toString() {
+        return "ContainerListEntry{" +
+               "id='" + id + '\'' +
+               ", names=" + Arrays.toString(names) +
+               ", image='" + image + '\'' +
+               ", imageID='" + imageID + '\'' +
+               ", command='" + command + '\'' +
+               ", created=" + created +
+               ", status='" + status + '\'' +
+               ", ports=" + Arrays.toString(ports) +
+               ", labels=" + labels +
+               ", sizeRw=" + sizeRw +
+               ", sizeRootFs=" + sizeRootFs +
+               '}';
+    }
+}
--- a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/Image.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/Image.java
@@ -10,31 +10,33 @@
  *******************************************************************************/
 package org.eclipse.che.plugin.docker.client.json;
 
+import java.util.Arrays;
+import java.util.Map;
+
 /** @author andrew00x */
 public class Image {
-    private String repository;
-    private String tag;
-    private String id;
-    private long   created;
-    private long   size;
-    private long   virtualSize;
+    private String[]            repoTags;
+    private String              parentId;
+    private String              id;
+    private long                created;
+    private long                size;
+    private long                virtualSize;
+    private Map<String, String> labels;
 
-    public String getRepository() {
-        return repository;
+    /**
+     * Return a collection of tags grouped under a common prefix (the name component before ':').
+     */
+    public String[] getRepoTags() {
+        return repoTags;
     }
 
-    public void setRepository(String repository) {
-        this.repository = repository;
+    public void setRepoTags(String[] repoTags) {
+        this.repoTags = repoTags;
     }
 
-    public String getTag() {
-        return tag;
-    }
-
-    public void setTag(String tag) {
-        this.tag = tag;
-    }
-
+    /**
+     * Return uniquely identifies the image.
+     */
     public String getId() {
         return id;
     }
@@ -43,6 +45,9 @@ public class Image {
         this.id = id;
     }
 
+    /**
+     * Return ISO-8601 formatted combined date and time at which the image was created.
+     */
     public long getCreated() {
         return created;
     }
@@ -51,6 +56,9 @@ public class Image {
         this.created = created;
     }
 
+    /**
+     * Return the size in bytes of the filesystem changeset associated with the image layer.
+     */
     public long getSize() {
         return size;
     }
@@ -59,6 +67,9 @@ public class Image {
         this.size = size;
     }
 
+    /**
+     *  Return the virtual size of the image
+     */
     public long getVirtualSize() {
         return virtualSize;
     }
@@ -67,15 +78,38 @@ public class Image {
         this.virtualSize = virtualSize;
     }
 
+    /**
+     * Return parent IDs of image layers to find the root ancestor
+     */
+    public String getParentId() {
+        return parentId;
+    }
+
+    public void setParentId(String parentId) {
+        this.parentId = parentId;
+    }
+
+    /**
+     * Returns metadata of the image
+     */
+    public Map<String, String> getLabels() {
+        return labels;
+    }
+
+    public void setLabels(Map<String, String> labels) {
+        this.labels = labels;
+    }
+
     @Override
     public String toString() {
         return "Image{" +
-               "repository='" + repository + '\'' +
-               ", tag='" + tag + '\'' +
+               "repoTags=" + Arrays.toString(repoTags) +
+               ", parentId='" + parentId + '\'' +
                ", id='" + id + '\'' +
                ", created=" + created +
                ", size=" + size +
                ", virtualSize=" + virtualSize +
+               ", labels=" + labels +
                '}';
     }
 }
--- a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/ListContainersParams.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/ListContainersParams.java
@@ -1,0 +1,175 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.plugin.docker.client.params;
+
+import org.eclipse.che.plugin.docker.client.json.Filters;
+
+import java.util.Objects;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Arguments holder for {@link org.eclipse.che.plugin.docker.client.DockerConnector#listContainers(ListContainersParams)}.
+ *
+ * @author Alexander Andrienko
+ */
+public class ListContainersParams {
+    private Boolean all;
+    private Integer limit;
+    private String  since;
+    private String  before;
+    private Boolean size;
+    private Filters filters;
+
+    private ListContainersParams() {
+    }
+
+    /**
+     * Creates and returns arguments holder.
+     */
+    public static ListContainersParams create() {
+        return new ListContainersParams();
+    }
+
+    /**
+     * Adds parameter show all containers. Only running containers are shown by default.
+     *
+     * @param all
+     *         if true show all containers, if false show only running containers
+     */
+    public ListContainersParams withAll(boolean all) {
+        this.all = all;
+        return this;
+    }
+
+    /**
+     * Adds parameter show limit last created containers, include non-running ones.
+     *
+     * @param limit
+     *         amount elements of the list containers
+     */
+    public ListContainersParams withLimit(int limit) {
+        this.limit = limit;
+        return this;
+    }
+
+    /**
+     * Adds parameter show only containers created since container with {@code id}.
+     *
+     * @param id
+     *         container id
+     */
+    public ListContainersParams withSince(String id) {
+        requireNonNull(id);
+        this.since = id;
+        return this;
+    }
+
+    /**
+     * Adds parameter show only containers created before container with {@code id}.
+     *
+     * @param id
+     *         container id
+     */
+    public ListContainersParams withBefore(String id) {
+        requireNonNull(id);
+        this.before = id;
+        return this;
+    }
+
+    /**
+     * Adds parameter show docker container size information.
+     *
+     * @param size
+     *         if size = true then api add container size information, otherwise hide this information.
+     *         Warning: if size = true docker api need more time for calculation container size
+     */
+    public ListContainersParams withSize(boolean size) {
+        this.size = size;
+        return this;
+    }
+
+    /**
+     * Adds parameter filters for filter list containers. See more {@link Filters}
+     *
+     * @param filters
+     *         parameters for filter list containers.
+     *         Filter values:
+     *         <li>exited=<int>; -- containers with exit code of <int>;</li>
+     *         <li>status=(created|restarting|running|paused|exited|dead)</li>
+     *         <li>label=key or label="key=value" of a container label</li>
+     *         <li>isolation=(default|process|hyperv) (Windows daemon only)</li>
+     *         <li>ancestor=(<image-name>[:<tag>], <image id> or <image@digest>)</li>
+     *         <li>before=(<container id> or <container name>)</li>
+     *         <li>since=(<container id> or <container name>)</li>
+     *         <li>volume=(<volume name> or <mount point destination>)</li>
+     */
+    public ListContainersParams withFilters(Filters filters) {
+        requireNonNull(filters);
+        this.filters = filters;
+        return this;
+    }
+
+    public Boolean isAll() {
+        return all;
+    }
+
+    public Integer getLimit() {
+        return limit;
+    }
+
+    public String getSince() {
+        return since;
+    }
+
+    public String getBefore() {
+        return before;
+    }
+
+    public Boolean isSize() {
+        return size;
+    }
+
+    public Filters getFilters() {
+        return filters;
+    }
+
+    @Override
+    public String toString() {
+        return "ListContainersParams{" +
+               "all=" + all +
+               ", limit=" + limit +
+               ", since='" + since + '\'' +
+               ", before='" + before + '\'' +
+               ", size=" + size +
+               ", filters=" + filters +
+               '}';
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) return true;
+        if (!(obj instanceof ListContainersParams)) return false;
+        ListContainersParams another = (ListContainersParams)obj;
+
+        return Objects.equals(isAll(), another.isAll()) &&
+               Objects.equals(getLimit(), another.getLimit()) &&
+               Objects.equals(getSince(), another.getSince()) &&
+               Objects.equals(getBefore(), another.getBefore()) &&
+               Objects.equals(isSize(), another.isSize()) &&
+               Objects.equals(getFilters(), another.getFilters());
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(isAll(), getLimit(), getSince(), getBefore(), isSize(), getFilters());
+    }
+}
--- a/plugins/plugin-docker/che-plugin-docker-client/src/test/java/org/eclipse/che/plugin/docker/client/DockerConnectorTest.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/test/java/org/eclipse/che/plugin/docker/client/DockerConnectorTest.java
@@ -11,6 +11,7 @@
 package org.eclipse.che.plugin.docker.client;
 
 import com.google.common.io.CharStreams;
+import com.google.common.reflect.TypeToken;
 
 import org.eclipse.che.commons.json.JsonParseException;
 import org.eclipse.che.commons.lang.ws.rs.ExtMediaType;
@@ -27,10 +28,12 @@ import org.eclipse.che.plugin.docker.client.json.ContainerConfig;
 import org.eclipse.che.plugin.docker.client.json.ContainerCreated;
 import org.eclipse.che.plugin.docker.client.json.ContainerExitStatus;
 import org.eclipse.che.plugin.docker.client.json.ContainerInfo;
+import org.eclipse.che.plugin.docker.client.json.ContainerListEntry;
 import org.eclipse.che.plugin.docker.client.json.ContainerProcesses;
 import org.eclipse.che.plugin.docker.client.json.Event;
 import org.eclipse.che.plugin.docker.client.json.ExecCreated;
 import org.eclipse.che.plugin.docker.client.json.ExecInfo;
+import org.eclipse.che.plugin.docker.client.json.Filters;
 import org.eclipse.che.plugin.docker.client.json.Image;
 import org.eclipse.che.plugin.docker.client.json.ImageInfo;
 import org.eclipse.che.plugin.docker.client.json.SystemInfo;
@@ -46,6 +49,7 @@ import org.eclipse.che.plugin.docker.client.params.GetResourceParams;
 import org.eclipse.che.plugin.docker.client.params.InspectContainerParams;
 import org.eclipse.che.plugin.docker.client.params.InspectImageParams;
 import org.eclipse.che.plugin.docker.client.params.KillContainerParams;
+import org.eclipse.che.plugin.docker.client.params.ListContainersParams;
 import org.eclipse.che.plugin.docker.client.params.PullParams;
 import org.eclipse.che.plugin.docker.client.params.PushParams;
 import org.eclipse.che.plugin.docker.client.params.PutResourceParams;
@@ -78,8 +82,12 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import static java.util.Collections.emptyMap;
+import static java.util.Collections.singletonList;
+import static java.util.Collections.singletonMap;
 import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.anyInt;
+import static org.mockito.Matchers.anyObject;
 import static org.mockito.Matchers.anyString;
 import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.doNothing;
@@ -97,11 +105,11 @@ import static org.testng.Assert.assertEquals;
 @Listeners(MockitoTestNGListener.class)
 public class DockerConnectorTest {
 
-    private static final String EXCEPTION_ERROR_MESSAGE    = "Error response from docker API, status: 500, message: Error";
-    private static final int    RESPONSE_ERROR_CODE        = 500;
-    private static final int    RESPONSE_SUCCESS_CODE      = 200;
-    private static final int    RESPONSE_NO_CONTENT_CODE   = 204;
-    private static final int    RESPONSE_CREATED_CODE      = 201;
+    private static final String EXCEPTION_ERROR_MESSAGE  = "Error response from docker API, status: 500, message: Error";
+    private static final int    RESPONSE_ERROR_CODE      = 500;
+    private static final int    RESPONSE_SUCCESS_CODE    = 200;
+    private static final int    RESPONSE_NO_CONTENT_CODE = 204;
+    private static final int    RESPONSE_CREATED_CODE    = 201;
 
     private static final String REQUEST_METHOD_GET    = "GET";
     private static final String REQUEST_METHOD_POST   = "POST";
@@ -119,8 +127,8 @@ public class DockerConnectorTest {
     private static final String   STREAM_DATA           = "stream data";
     private static final String   DOCKER_RESPONSE       = "stream";
     private static final String   ERROR_MESSAGE         = "some error occurs";
-    private static final String[] CMD_WITH_ARGS         = { "command", "arg1", "arg2" };
-    private static final String[] CMD_ARGS              = { "arg1", "arg2" };
+    private static final String[] CMD_WITH_ARGS         = {"command", "arg1", "arg2"};
+    private static final String[] CMD_ARGS              = {"arg1", "arg2"};
     private static final byte[]   STREAM_DATA_BYTES     = STREAM_DATA.getBytes();
     private static final byte[]   DOCKER_RESPONSE_BYTES = DOCKER_RESPONSE.getBytes();
 
@@ -149,7 +157,7 @@ public class DockerConnectorTest {
     private File                         dockerfile;
 
     @Captor
-    private ArgumentCaptor<Object>       captor;
+    private ArgumentCaptor<Object> captor;
 
     @BeforeMethod
     public void setup() throws IOException, URISyntaxException {
@@ -229,7 +237,7 @@ public class DockerConnectorTest {
     public void shouldBeAbleToGetListImages() throws IOException, JsonParseException {
         List<Image> images = new ArrayList<>();
 
-        doReturn(images).when(dockerConnector).parseResponseStreamAsListAndClose(inputStream);
+        doReturn(images).when(dockerConnector).parseResponseStreamAsListAndClose(eq(inputStream), any());
 
         List<Image> returnedImages =
                 dockerConnector.listImages();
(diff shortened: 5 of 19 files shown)


plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -168,6 +171,41 @@ public Version getVersion() throws IOException {
     }
 
     /**
+     * Method returns list docker containers which wos filtered by query parameters from {@link ListContainersParams}
```
garagatyi: Typo: wos

garagatyi:
Add yourself as author

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleanerTest.java
```diff
@@ -0,0 +1,148 @@
...
+
+        cleaner.run();
+
+        verify(dockerConnector, never()).killContainer(containerId1);
+        verify(dockerConnector, never()).removeContainer(containerId1, true, true);
+
+        verify(dockerConnector, never()).killContainer(containerId2);
+        verify(dockerConnector, never()).removeContainer(containerId2, true, true);
+    }
+}
```
garagatyi: Please add trailing new line at the end of file

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/ListContainersParams.java
```diff
@@ -0,0 +1,81 @@
...
+
+import org.eclipse.che.plugin.docker.client.json.Filters;
+
+/**
+ * Arguments holder for {@link org.eclipse.che.plugin.docker.client.DockerConnector#listContainers(ListContainersParams)}.
+ *
+ * @author Alexander Andrienko
+ */
+public class ListContainersParams {
+    private boolean all;
```
garagatyi: Align fields formatting

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/ListContainersParams.java
```diff
@@ -0,0 +1,81 @@
...
+    private int limit;
```
garagatyi: use wrappers for primitives. It will allow to omit setting default values and allow to rely on docker API

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/ListContainersParams.java
```diff
@@ -0,0 +1,81 @@
...
+package org.eclipse.che.plugin.docker.client.params;
+
+import org.eclipse.che.plugin.docker.client.json.Filters;
+
+/**
+ * Arguments holder for {@link org.eclipse.che.plugin.docker.client.DockerConnector#listContainers(ListContainersParams)}.
+ *
+ * @author Alexander Andrienko
+ */
+public class ListContainersParams {
```
garagatyi, line 24: Add java docs to parameters of that object

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/ListContainersParams.java
```diff
@@ -0,0 +1,81 @@
...
+        this.all = all;
+        return this;
+    }
+
+    public ListContainersParams withLimit(int limit) {
+        this.limit = limit;
+        return this;
+    }
+
+    public ListContainersParams withSince(String since) {
```
garagatyi: Use **id** instead of **since** as argument name.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/ListContainersParams.java
```diff
@@ -0,0 +1,81 @@
...
+        this.since = since;
+        return this;
+    }
+
+    public ListContainersParams withBefore(String before) {
```
garagatyi: Use **id** instead of **before** as argument name.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/ListContainersParams.java
```diff
@@ -0,0 +1,81 @@
...
+        this.before = before;
+        return this;
+    }
+
+    public ListContainersParams withSize(boolean size) {
```
garagatyi, line 95: Explain in docs and param name that it is containers size.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/ListContainersParams.java
garagatyi, line 24: Add equals, hachcode, toString methods

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -168,6 +171,41 @@ public Version getVersion() throws IOException {
...
+     * @throws IOException
+     *         in case error parsing response from docker api
+     */
+    public ContainerFromList[] listContainers(ListContainersParams params) throws IOException {
+        try (DockerConnection connection = connectionFactory.openConnection(dockerDaemonUri)
+                                                            .method("GET")
+                                                            .path("/containers/json")
+                                                            .query("all", params.isAll())
+                                                            .query("size", params.isSize())) {
+            if (params.getLimit() > 0) {
```
garagatyi: Use wrapper to avoid such code

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -168,6 +171,41 @@ public Version getVersion() throws IOException {
...
+                connection.query("limit", params.getLimit());
+            }
+            if (!isNullOrEmpty(params.getSince())) {
```
garagatyi: Do not check for emptyness

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -168,6 +171,41 @@ public Version getVersion() throws IOException {
...
+                connection.query("since", params.getSince());
+            }
+            if (!isNullOrEmpty(params.getBefore())) {
```
garagatyi: Do not check for emptyness

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -168,6 +171,41 @@ public Version getVersion() throws IOException {
     }
 
     /**
+     * Method returns list docker containers which wos filtered by query parameters from {@link ListContainersParams}
+     *
+     * @throws IOException
+     *         in case error parsing response from docker api
+     */
+    public ContainerFromList[] listContainers(ListContainersParams params) throws IOException {
```
garagatyi: Please add method with no argument because it will be used mostly. So simplified code of its execution will be appreciated.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/ContainerFromList.java
```diff
@@ -0,0 +1,138 @@
...
+
+import java.util.Arrays;
+import java.util.Map;
+
+/**
+ * Defines information about container from list containers.
+ *
+ * @author Alexander Andrienko
+ */
+public class ContainerFromList {
```
garagatyi: Rename class - now it looks like it is collection, but it is not

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/Port.java
```diff
@@ -0,0 +1,55 @@
...
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.plugin.docker.client.json;
+
+/**
+ * Defines information about container port
+ *
+ * @author Alexander Andrienko
+ */
+public class Port {
```
garagatyi: Name is too general. Please make it more obvious where it is used.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerContainerNameGenerator.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * This class used for generation docker container name or parse important information from docker container name.
+ *
+ * @author Alexander Andrienko
+ */
+public class DockerContainerNameGenerator {
+    private static final String WORKSPACE_ID_REGEX      = "workspace[\\da-z]{16}";
```
garagatyi: As underscore is not allowed in the IDs you can rewrite regexps to **workspace[\da-z]+** to avoid problems if id side is changed.

garagatyi: I think 0-9 is better than \d because it is more informative for those who doesn't remember all special groups

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerContainerNameGenerator.java
```diff
@@ -0,0 +1,133 @@
...
+    private static final String MACHINE_ID_REGEX        = "machine[\\da-z]{16}";
+
+    private static final String  CONTAINER_NAME_REGEX   = "^" + WORKSPACE_ID_REGEX + '_' + MACHINE_ID_REGEX + "_[a-z0-9_-]+_[a-z0-9_-]+$";
```
garagatyi: [a-z0-9_-]+_[a-z0-9_-]+ is the same as [a-z0-9_-]+
Isn't it?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerContainerNameGenerator.java
```diff
@@ -0,0 +1,133 @@
...
+    private static final Pattern CONTAINER_NAME_PATTERN = Pattern.compile(CONTAINER_NAME_REGEX);
+
+    /**
+     * Return generated name for docker container. Method generate docker container name in format:
+     * workspaceId + "_" + machineId + "_" + userName +"_" + machineName
```
garagatyi: Use `<p/>` and `<br/>` to format javadocs. Regular line feeds doesn't affect javadocs

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerContainerNameGenerator.java
sleshchenko: I think no. [a-z0-9_-]_+[a-z0-9-]+ doesn't match to 'qwerty' and [a-z0-9_-]+ matches it.
Maybe it is not required is this case. But it is not the same.
_[a-z0-9_-]_+[a-z0-9-]+ is the same to something like (_[a-z0-9_-]+){2}

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineModule.java
```diff
@@ -24,6 +25,9 @@
 public class DockerMachineModule extends AbstractModule {
     @Override
     protected void configure() {
+        install(new org.eclipse.che.commons.schedule.executor.ScheduleModule());
+        bind(DockerContainerCleaner.class);
```
garagatyi: Use FQN

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerContainerNameGenerator.java
garagatyi: Yes, @sleshchenko you're right.
But I think we don't need such regexp at all because we won't use that part, only WSID, MachineID. So @AndrienkoAleksandr we can replace it with `_.+`

AndrienkoAleksandr: I mean that userName can contains at least 1 char and machine name can contains at leas 1 char. And `_` between them. Yes looks `_[a-z0-9_-]_+[a-z0-9-]+` equal to `(_[a-z0-9_-]+){2}`

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/ListContainersParams.java
AndrienkoAleksandr, line 24: Java equals, hashcode, toString, java docs for parameters added

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleanerTest.java
AndrienkoAleksandr: Trailing new line added

AndrienkoAleksandr:
Pull request updated. @garagatyi @sleshchenko review please.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleaner.java
```diff
@@ -0,0 +1,114 @@
...
+                                  DockerConnector dockerConnector,
+                                  DockerContainerNameGenerator nameGenerator) {
+        this.machineManager = machineManager;
+        this.dockerConnector = dockerConnector;
+        this.nameGenerator = nameGenerator;
+    }
+
+    @ScheduleRate(period = 5, initialDelay = 5, unit = TimeUnit.MINUTES)//it will be changed to one per hour. Maybe it should be set in property ?
+    @Override
+    public void run() {
```
sleshchenko, line 59: I would recommend add catch (Exception ...) to avoid suppressing of next subsequent executions in case throwing RuntimeException

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -168,6 +171,55 @@ public Version getVersion() throws IOException {
     }
 
     /**
+     * Method returns list docker containers, include non-running ones.
+     *
+     * @throws IOException
```
garagatyi, line 201: And in case of communication error with Docker API, am I right?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleaner.java
AndrienkoAleksandr, line 59: Ok, thank you.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
AndrienkoAleksandr, line 201: Ok, I''ll fix it

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -168,6 +171,55 @@ public Version getVersion() throws IOException {
...
+     *         in case error parsing response from docker api
+     */
+    public ContainerListEntry[] listContainers() throws IOException {
```
garagatyi: I think list is better in that case than array

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/ContainerListEntry.java
```diff
@@ -0,0 +1,142 @@
...
+    private String              imageID;
+    private String              command;
+    private long                created;
+    private String              status;
+    private ContainerPort[]     ports;
+    private Map<String, String> labels;
+    private int                 sizeRw;
+    private int                 sizeRootFs;
+
+    public String getId() {
```
garagatyi, line 37: Javadocs

AndrienkoAleksandr:
Pull request updated. Moved job period to the properties file.

assembly/assembly-ide-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -70,6 +70,7 @@ docker.registry.auth.password=NULL
 docker.registry.auth.email=NULL
 docker.connection.tcp.connection_timeout_ms=600000
 docker.connection.tcp.read_timeout_ms=600000
+docker.clean.up.unused.containers.period_min=5
```
garagatyi: It is good that you've added unit to property name. But naming is not the best. We use dots to separate components and underscores to separate words of one component.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -168,6 +171,55 @@ public Version getVersion() throws IOException {
     }
 
     /**
+     * Method returns list of docker containers, include non-running ones.
+     *
+     * @throws IOException
+     *         when problems occurs with docker api calls
+     */
+    public ContainerListEntry[] listContainers() throws IOException {
```
garagatyi: Use kist instead of array

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/ListContainersParams.java
```diff
@@ -0,0 +1,160 @@
...
+    private String  since;
+    private String  before;
+    private Boolean size;
+    private Filters filters;
+
+    /**
+     * Adds parameter show all containers. Only running containers are shown by default.
+     *
+     * @param all
+     *         if all == true show all containers, if all = false show only running containers
```
garagatyi: `if true show all containers, if false show only running containers` looks better

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/ListContainersParams.java
```diff
@@ -0,0 +1,160 @@
...
+     */
+    public ListContainersParams withAll(Boolean all) {
```
garagatyi: Avoid wrappers in parameters to prevent passing of nulls. Use wrappers in fields, getters only instead

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/ListContainersParams.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /**
+     * Adds parameter show only containers created since container with {@code id}.
+     *
+     * @param id
+     *         container id
+     */
+    public ListContainersParams withSince(String id) {
+        this.since = id;
```
garagatyi, line 72: check nullness

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/ListContainersParams.java
garagatyi, line 24: As far as I understand you've tried to align this PR with #952. So it is better to add private constructor and factory method `from()` to this class to avoid different API styles for different docker methods.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/ListContainersParams.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /**
+     * Adds parameter show only containers created before container with {@code id}.
+     *
+     * @param id
+     *         container id
+     */
+    public ListContainersParams withBefore(String id) {
+        this.before = id;
```
garagatyi, line 84: check nullness

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/ListContainersParams.java
```diff
@@ -0,0 +1,160 @@
...
+     *         <li>status=(created|restarting|running|paused|exited|dead)</li>
+     *         <li>label=key or label="key=value" of a container label</li>
+     *         <li>isolation=(default|process|hyperv) (Windows daemon only)</li>
+     *         <li>ancestor=(<image-name>[:<tag>], <image id> or <image@digest>)</li>
+     *         <li>before=(<container id> or <container name>)</li>
+     *         <li>since=(<container id> or <container name>)</li>
+     *         <li>volume=(<volume name> or <mount point destination>)</li>
+     */
+    public ListContainersParams withFilters(Filters filters) {
+        this.filters = filters;
```
garagatyi, line 117: check nullness

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerContainerNameGenerator.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * This class is used for generation docker container name or parsing important information from docker container name.
+ *
+ * @author Alexander Andrienko
+ */
+public class DockerContainerNameGenerator {
+    private static final String WORKSPACE_ID_GROUP = "(?<workspaceId>workspace[0-9a-z]{16})";
```
garagatyi: I think it is better to avoid strict limit to random chars number to prevent fail of this regexp if ID size will be increased

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerContainerNameGenerator.java
```diff
@@ -0,0 +1,128 @@
...
+    private static final String MACHINE_ID_GROUP   = "(?<machineId>machine[0-9a-z]{16})";
+
+    private static final String  CONTAINER_NAME_REGEX   = "^" + WORKSPACE_ID_GROUP + "_" + MACHINE_ID_GROUP + "(_.*){2}$";
```
garagatyi: I think it doesn't matter which chars are in container name after valuable for you parts. So you can simplify regexp.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerContainerNameGenerator.java
```diff
@@ -0,0 +1,128 @@
...
+    /**
+     * Parse machine's {@code containerName} to get information about this container (like workspaceId, machineId).
+     * Notice: method doesn't parse information about userName or machineName,because we do not give guarantees
+     * about the integrity of this data(see more {@link #generateContainerName(String, String, String, String)})
+     *
+     * @param containerName
+     *         name of the container
+     * @return information about container
+     */
+    public ContainerNameInfo parse(@Nullable String containerName) {
```
garagatyi: it can't be nullable because you will get NPE if argument is null. Return value is nullable instead.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerContainerNameGenerator.java
```diff
@@ -0,0 +1,128 @@
...
+        containerName = containerName.replace("/", "");
```
garagatyi: As I can see from method `generateContainerName` it is not possible that slash will be present in a container name.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerContainerNameGenerator.java
```diff
@@ -0,0 +1,128 @@
...
+
+        Matcher matcher = CONTAINER_NAME_PATTERN.matcher(containerName);
+        if (matcher.matches()) {
+            String workspaceId = matcher.group("workspaceId");
+            String machineId = matcher.group("machineId");
+            return new ContainerNameInfo(workspaceId, machineId);
+        }
+        return null;
```
garagatyi: I think that it is better to return Optional in that case than return null and mark method as those which returns nullable

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -214,7 +216,11 @@ public Instance createInstance(Recipe recipe,
                                    LineConsumer creationLogsOutput) throws MachineException, UnsupportedRecipeException {
         final Dockerfile dockerfile = parseRecipe(recipe);
 
-        final String machineContainerName = generateContainerName(machine.getWorkspaceId(), machine.getConfig().getName());
+        String userName = EnvironmentContext.getCurrent().getUser().getName();
```
garagatyi: You can inline this into `containerNameGenerator.generateContainerName`

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -234,7 +240,11 @@ public Instance createInstance(InstanceKey instanceKey,
 
         pullImage(dockerInstanceKey, creationLogsOutput);
 
-        final String machineContainerName = generateContainerName(machine.getWorkspaceId(), machine.getConfig().getName());
+        String userName = EnvironmentContext.getCurrent().getUser().getName();
```
garagatyi: You can inline this into `containerNameGenerator.generateContainerName`

AndrienkoAleksandr:
Pull request updated. Added machine name to output logs, renamed property, simplified pattern for search unused containers.

AndrienkoAleksandr:
Pull request updated. Improved parsing docker container name for deleting node host (in case using docker  Swarm).

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -168,6 +173,55 @@ public Version getVersion() throws IOException {
...
+            }
+            if (params.getFilters() != null) {
+                connection.query("filters", urlPathSegmentEscaper().escape(JsonHelper.toJson(params.getFilters())));
+            }
+            DockerResponse response = connection.request();
+            final int status = response.getStatus();
+            if (OK.getStatusCode() != status) {
+                throw getDockerException(response);
+            }
+            return stream(parseResponseStreamAndClose(response.getInputStream(), ContainerListEntry[].class)).collect(toList());
```
garagatyi: It is not efficient to parse json as array and then stream and collect it into list. Use parsing directly into list

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerContainerNameGenerator.java
```diff
@@ -0,0 +1,129 @@
...
+     * Parse machine's {@code containerName} to get information about this container (like workspaceId, machineId).
+     * Notice: method doesn't parse information about userName or machineName,because we do not give guarantees
+     * about the integrity of this data(see more {@link #generateContainerName(String, String, String)})
+     *
+     * @param containerName
+     *         name of the container
+     * @return information about container
+     */
+    public Optional<ContainerNameInfo> parse(String containerName) {
+        String[] containerNameParts = containerName.split("/");
```
garagatyi: please add comment that explains why you cat of the slash.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerContainerNameGenerator.java
```diff
@@ -0,0 +1,129 @@
...
+        containerName = containerNameParts[containerNameParts.length - 1];
```
garagatyi: why you assign it back to the argument? you can use it directly in matching

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleaner.java
```diff
@@ -0,0 +1,123 @@
...
+    @ScheduleRate(periodParameterName = "machine.docker.unused_containers_cleanup_period_min",
+                  initialDelayParameterName = "machine.docker.unused_containers_cleanup_period_min",
+                  unit = TimeUnit.MINUTES)
+    @Override
+    public void run() {
+        try {
+            List<ContainerListEntry> allDockerContainers = dockerConnector.listContainers();
+            List<MachineImpl> machines = machineManager.getMachines();
+
+            List<ContainerListEntry> unUsedContainers = findUnUsedContainers(allDockerContainers, machines);
```
garagatyi: Unused is one word

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleaner.java
```diff
@@ -0,0 +1,123 @@
...
+    private List<ContainerListEntry> findUnUsedContainers(List<ContainerListEntry> containers, List<MachineImpl> machines) {
+        List<ContainerListEntry> unUsedContainers = new ArrayList<>();
+        for (ContainerListEntry container : containers) {
+            Optional<ContainerNameInfo> optional = nameGenerator.parse(container.getNames()[0]);
+            if (!optional.isPresent()) {
+                continue;
+            }
+            final ContainerNameInfo containerNameInfo = optional.get();
+            boolean containerIsUsed = machines.stream()
+                                              .anyMatch(machine -> machine.getId().equals(containerNameInfo.getMachineId())
```
garagatyi: Machine Id is unique enough to use only it

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleaner.java
```diff
@@ -0,0 +1,123 @@
...
+
+    private List<ContainerListEntry> findUnUsedContainers(List<ContainerListEntry> containers, List<MachineImpl> machines) {
+        List<ContainerListEntry> unUsedContainers = new ArrayList<>();
+        for (ContainerListEntry container : containers) {
+            Optional<ContainerNameInfo> optional = nameGenerator.parse(container.getNames()[0]);
+            if (!optional.isPresent()) {
+                continue;
+            }
+            final ContainerNameInfo containerNameInfo = optional.get();
+            boolean containerIsUsed = machines.stream()
```
garagatyi: You stream all machines inside of loop which actually means that it is nested loop. Instead of that you can check if machine exists in machine manager by getting it or adding method that says if machine with provided id exists.

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/DockerContainerNameGeneratorTest.java
```diff
@@ -0,0 +1,177 @@
...
+    private DockerContainerNameGenerator nameGenerator;
+
+    @BeforeMethod
+    public void setUp() {
+        EnvironmentContext.getCurrent().setUser(new UserImpl(USER_NAME, "stub", "stub", emptyList(), false));
+    }
+
+    @Test
+    public void containerNameShouldBeGenerated() {
+        String expectedResult = "workspacebbbx2ree3iykn8gc_machineic131ppamujngv6y_some_user_ws-machine";
```
garagatyi, line 43: Please add tests for usernames, machine names with invalid characters and upper cased characters

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleanerTest.java
```diff
@@ -0,0 +1,151 @@
...
+        Optional<ContainerNameInfo> optional2 = Optional.of(containerNameInfo2);
+        when(nameGenerator.parse(containerName1)).thenReturn(optional1);
+        when(nameGenerator.parse(containerName2)).thenReturn(optional2);
+
+        when(containerNameInfo1.getMachineId()).thenReturn(machineId1);
+        when(containerNameInfo1.getWorkspaceId()).thenReturn(workspaceId1);
+    }
+
+    @Test
+    public void dockerContainerShouldBeKilledAndRemoved() throws MachineException, IOException {
```
garagatyi: Names of tests say nothing about usecases. Please fix that

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleanerTest.java
```diff
@@ -0,0 +1,151 @@
...
+        cleaner.run();
+
+        verify(dockerConnector).listContainers();
+        verify(machineManager).getMachines();
+
+        verify(container1).getNames();
+        verify(container2, times(3)).getNames();
+
+        verify(nameGenerator).parse(containerName1);
```
garagatyi: Are you sure that tests should include so many checks in each usecase?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleaner.java
garagatyi: For example you have 2k machine and 2k + 1 containers. You try to find machine for container which is not used by machines anymore. You have to perform ~ 2k \* 2k = 4kk opertaions.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleaner.java
```diff
@@ -0,0 +1,123 @@
...
+            if (!containerIsUsed) {
+                unUsedContainers.add(container);
+            }
+        }
+        return unUsedContainers;
+    }
+
+    private void killContainer(ContainerListEntry container) {
+        String containerId = container.getId();
+        String containerName = container.getNames()[0];
```
garagatyi, line 76: Why you assign here variable for the sake of logging? just use this expression in logging operation.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleaner.java
```diff
@@ -0,0 +1,123 @@
...
+        try {
+            if (container.getStatus().startsWith("Up")) {
+                dockerConnector.killContainer(containerId);
+                LOG.warn("Container with 'id': '{}' and 'name': '{}' was killed ", containerId, containerName);
+            }
+        } catch (IOException e) {
+            LOG.error("Failed to kill unused container with 'id': '{}' and 'name': '{}'", containerId, containerName, e);
```
garagatyi: In that case exception will be converted to string. You have to use String.format here.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleaner.java
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }
+
+    private void removeContainer(ContainerListEntry container) {
+        String containerId = container.getId();
+        String containerName = container.getNames()[0];
```
garagatyi: Why you assign here variable for the sake of logging? just use this expression in logging operation.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleaner.java
```diff
@@ -0,0 +1,123 @@
...
+        try {
+            //remove force with volumes
+            dockerConnector.removeContainer(containerId, true, true);
+            LOG.warn("Container with 'id': '{}' and 'name': '{}' was removed", containerId, containerName);
+        } catch (IOException e) {
+            LOG.error("Failed to delete unused container with 'id': '{}' and 'name': '{}'", containerId, containerName, e);
```
garagatyi: In that case exception will be converted to string. You have to use String.format here.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
AndrienkoAleksandr: Ok.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/ListContainersParams.java
AndrienkoAleksandr, line 24: Added

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerContainerNameGenerator.java
AndrienkoAleksandr: ok

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerContainerNameGenerator.java
AndrienkoAleksandr: ok

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleaner.java
AndrienkoAleksandr: Ok, I agree with you.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleaner.java
AndrienkoAleksandr: Ok, it's make sense.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -168,6 +173,55 @@ public Version getVersion() throws IOException {
...
+     *
+     * @throws IOException
+     *         when problems occurs with docker api calls
+     */
+    public List<ContainerListEntry> listContainers() throws IOException {
+        return listContainers(ListContainersParams.from().withAll(true));
+    }
+
+    /**
+     * Method returns list of docker containers which was filtered by query parameters from {@link ListContainersParams}
```
garagatyi: ListContainersParams says nothing about query params, so don't use query keyword in explanation

garagatyi:
This PR can't be merged to master. please rebase it

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -185,7 +189,48 @@ public Version getVersion() throws IOException {
...
+        try (DockerConnection connection = connectionFactory.openConnection(dockerDaemonUri)
+                                                            .method("GET")
+                                                            .path("/containers/json")) {
+            addQueryParamIfNotNull(connection, "all", params.isAll());
+            addQueryParamIfNotNull(connection, "size", params.isSize());
+            addQueryParamIfNotNull(connection, "limit", params.getLimit());
+            addQueryParamIfNotNull(connection, "since", params.getSince());
+            addQueryParamIfNotNull(connection, "before", params.getBefore());
+            if (filters != null) {
+                connection.query("filters", urlPathSegmentEscaper().escape(JsonHelper.toJson(params.getFilters())));
```
garagatyi: Either use local variable you've previously set or don't create that variable at all

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/Image.java
```diff
@@ -10,31 +10,33 @@
...
     }
 
-    public void setTag(String tag) {
-        this.tag = tag;
+    public void setRepoTags(String[] repoTags) {
+        this.repoTags = repoTags;
     }
 
+    /**
+     * Return randomly generated, 256-bit, hexadecimal encoded uniquely identifies the image.
```
garagatyi: Describe is as docker image id. Current explanation is too precise as for client of Docker API

plugins/plugin-docker/che-plugin-docker-client/src/test/java/org/eclipse/che/plugin/docker/client/DockerConnectorTest.java
```diff
@@ -1175,4 +1232,95 @@ public void shouldUseFirstLetterLowercaseWhenParseResponseStreamAndClose() throw
         assertEquals(parsedVersion, version);
     }
 
+    @Test
+    public void shouldAbilityParseResponseStreamListImagesAndClose() throws IOException, JsonParseException {
```
garagatyi: shouldBeAbleToParseResponseStreamAsListOfImagesAndClose

plugins/plugin-docker/che-plugin-docker-client/src/test/java/org/eclipse/che/plugin/docker/client/DockerConnectorTest.java
```diff
@@ -1175,4 +1232,95 @@ public void shouldUseFirstLetterLowercaseWhenParseResponseStreamAndClose() throw
...
+        assertEquals(actualImage2.getId(), "b750fe79269d2ec9a3c593ef05b4332b1d1a02a62b4accb2c21d589ff2f5f2dc");
+        assertEquals(actualImage2.getParentId(), "27cf784147099545");
+        assertEquals(actualImage2.getCreated(), 1364102658);
+        assertEquals(actualImage2.getSize(), 24653);
+        assertEquals(actualImage2.getVirtualSize(), 180116135);
+        assertEquals(actualImage2.getLabels(), singletonMap("com.example.version", "v1"));
+    }
+
+    @Test
+    public void shouldAbilityParseResponseStreamListContainersAndClose() throws IOException, JsonParseException {
```
garagatyi: shouldBeAbleToParseResponseStreamAsListOfContainersAndClose

plugins/plugin-docker/che-plugin-docker-client/src/test/java/org/eclipse/che/plugin/docker/client/DockerConnectorTest.java
```diff
@@ -1175,4 +1232,95 @@ public void shouldUseFirstLetterLowercaseWhenParseResponseStreamAndClose() throw
...
+        assertEquals(actualContainer1.getImage(), "ubuntu:latest");
+        assertEquals(actualContainer1.getCommand(), "echo 1");
+        assertEquals(actualContainer1.getCreated(), 1367854155);
+        assertEquals(actualContainer1.getStatus(), "Exit 0");
+        assertEquals(actualContainer1.getSizeRw(), 12288);
+        assertEquals(actualContainer1.getSizeRootFs(), 0);
+    }
+
+    @Test
+    public void shouldAbilityParseResponseStreamListAndClose() throws IOException, JsonParseException {
```
garagatyi: shouldBeAbleToParseResponseStreamAsEmptyListOfContanersAndClose

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerContainerNameGenerator.java
```diff
@@ -0,0 +1,128 @@
...
+     *
+     * @param workspaceId
+     *         unique workspace id, see more (@link WorkspaceConfig#getId)
+     * @param machineId
+     *         unique machine id, see more {@link Machine#getId()}
+     * @param machineName
+     *         name of the workspace machine, see more {@link MachineConfig#getName()}
+     */
+    public String generateContainerName(String workspaceId, String machineId, String machineName) {
+        String userName = EnvironmentContext.getCurrent().getUser().getName();
```
garagatyi: Don't you think that it is better to pass username as an argument to avoid reliance on EnvironmentContext in that component?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerContainerNameGenerator.java
```diff
@@ -0,0 +1,128 @@
...
+     * Notice: method doesn't parse information about userName or machineName, because we do not give guarantees
+     * about the integrity of this data(see more {@link #generateContainerName(String, String, String)})
+     *
+     * @param containerName
+     *         name of the container
+     * @return information about container
+     */
+    public Optional<ContainerNameInfo> parse(String containerName) {
+        // Split container name, in case it contains node host (e.g. "/node-host.dev.box/workspacebbx2_machineic3_user321_ws-machine"
+        String[] containerNameParts = containerName.split("/");
```
garagatyi: Why not include that in the regexp?

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/DockerContainerNameGeneratorTest.java
```diff
@@ -0,0 +1,195 @@
...
+        assertEquals(containerNameInfoParser.get().getMachineId(), MACHINE_ID);
+        assertEquals(containerNameInfoParser.get().getWorkspaceId(), WORKSPACE_ID);
+    }
+
+    @DataProvider(name = "validContainerNames")
+    public static Object[][] validContainerNames() {
+        return new Object[][] {{"/workspacep2bivvctac5ciwoh_machineri6bxnoj5jq7ll9j_user-name_ws-machine-name",
+                                new Pair<>("machineri6bxnoj5jq7ll9j", "workspacep2bivvctac5ciwoh")},
+
+                               {"/workspacep2bivvctac5ciwoh_machineri6bxnoj5jq7ll9j_user-name_ws-machine-name",
```
garagatyi: include example with Swarm's node name at the beggining

AndrienkoAleksandr:
Updated. @garagatyi review please.

garagatyi:
Please perform final tests with docker and swarm

garagatyi:
LGTM

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -185,7 +189,48 @@ public Version getVersion() throws IOException {
             if (OK.getStatusCode() != response.getStatus()) {
                 throw getDockerException(response);
             }
-            return parseResponseStreamAsListAndClose(response.getInputStream());
+            return parseResponseStreamAsListAndClose(response.getInputStream(), new TypeToken<List<Image>>() {}.getType());
```
voievodin, line 192: There is no need to create new instance of  `TypeToken` each time when you list the images

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleaner.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    @ScheduleRate(periodParameterName = "machine.docker.unused_containers_cleanup_period_min",
+                  initialDelayParameterName = "machine.docker.unused_containers_cleanup_period_min",
+                  unit = TimeUnit.MINUTES)
+    @Override
+    public void run() {
+        try {
+            List<ContainerListEntry> dockerContainers = dockerConnector.listContainers();
+            for (ContainerListEntry container : dockerContainers) {
```
mshaposhnik: why not simplify to
`for (ContainerListEntry container : dockerConnector.listContainers())`

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/ListContainersParams.java
```diff
@@ -0,0 +1,175 @@
...
+    private Boolean size;
+    private Filters filters;
+
+    private ListContainersParams() {
+    }
+
+    /**
+     * Creates and returns arguments holder.
+     */
+    public static ListContainersParams from() {
```
voievodin: well, i don't think that empty `from()` is a really good choice

garagatyi: At least it is aligned with other docker client's params. If it should be changed it should be changed  in all params

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
AndrienkoAleksandr, line 192: If we move creation this TypeToken to the field we should do that for all methods, which used method  parseResponseStreamAsListAndClose (for now we have two such methods, but it can be more). I think method parseResponseStreamAsListAndClose should be ease of use, so we should not create new fields for TypeToken in the docker connector.

skabashnyuk:
ok

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/ListContainersParams.java
voievodin: the API is really new and as far as i know the version with it hasn't been released yet, so we can consider a different name for all the methods like this, lets say it is `.create`. WDYT @mmorhun, @garagatyi ?

garagatyi: +1 to change the name from **from** to **create** but -1 to do that in this PR because it is not a purpose of that PR

mmorhun: Ok, let be so

voievodin: Imho the classes added in this PR should use `.create` method instead and then we will be able to create a new issue and refer to those classes.

garagatyi: agree

AndrienkoAleksandr: Ok, @evoevodin. I think it's a great idea.

PR #1105, opened 2016-04-21
CHE-1039: add ability to bind additional folder to docker dev machine

garagatyi:
@skabashnyuk @eivantsov @vparfonov @evoevodin Please review

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
@@ -121,11 +121,13 @@ public class DockerInstanceProvider implements InstanceProvider {
         this.supportedRecipeTypes = Collections.singleton("dockerfile");
         this.projectFolderPath = projectFolderPath;
 
+        allMachinesSystemVolumes = filterEmptyAndNullValues(allMachinesSystemVolumes);
+        devMachineSystemVolumes = filterEmptyAndNullValues(devMachineSystemVolumes);
         if (SystemInfo.isWindows()) {
-            allMachinesSystemVolumes = escapePaths(allMachinesSystemVolumes);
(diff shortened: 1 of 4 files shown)


plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -121,11 +121,13 @@ public DockerInstanceProvider(DockerConnector docker,
         this.supportedRecipeTypes = Collections.singleton("dockerfile");
         this.projectFolderPath = projectFolderPath;
 
+        allMachinesSystemVolumes = filterEmptyAndNullValues(allMachinesSystemVolumes);
+        devMachineSystemVolumes = filterEmptyAndNullValues(devMachineSystemVolumes);
         if (SystemInfo.isWindows()) {
-            allMachinesSystemVolumes = escapePaths(allMachinesSystemVolumes);
-            devMachineSystemVolumes = escapePaths(devMachineSystemVolumes);
+            allMachinesSystemVolumes = escapePaths(filterEmptyAndNullValues(allMachinesSystemVolumes));
```
voievodin: Why did you add `filterEmptyAndNullValues` to the `escapePaths` method, as far as i understand they are already filtered 2 lines above + i think that it is better to name method `removeEmptyAndNullValues` which is more obvious against filter.

skabashnyuk:
ok

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
garagatyi: Oh, its a mistake. I forgot to remove that. Tnx

garagatyi: done

voievodin:
ok

PR #1107, opened 2016-04-21
Fix error 'removal of the recipe file' in the console log

AndrienkoAleksandr:
When file was delete succesfully then file.delete() == true.
@garagatyi @skabashnyuk review please


garagatyi:
LGTM

skabashnyuk:
ок

PR #1108, opened 2016-04-21
CHE-1005: properly display machine type icons in console tree and targets view, fix logo size

ashumilova:
**_1 Upvote**_ Signed-off-by: Ann Shumilova ashumilova@codenvy.com


ashumilova:
artik_console_panel

TylerJewell:
@ashumilova - it looks weird to have the DEV icon and the artik icon different sizes.  I think they should both be the same size / alignment as the ARTIK icon.  So perhaps we should shrink the DEV icon so that everything looks aligned?

vitaliy-guliy:
+1 to decrease DEV icon size

ashumilova:
reduced a little
artik_console_panel1

vitaliy-guliy:
Looks better

TylerJewell:
Still looks weird to me.   I think we should leave it as ART instead.  That was cleaner than this.  It's because one is a box and another is a logo.  So they don't present cleanly.

PR #1109, opened 2016-04-21
CHE-1002 Add target UX improvements

vitaliy-guliy:


ashumilova:
+1

PR #1113, opened 2016-04-21
Optimazing builder's implementation

alexVengrovsk:

Maybe I get it all wrong, but it seems to me, that when builder is provided for creation an instance of the  _DockerImageIdentifier_  it is not a very good idea to make a constructor of this class public. Otherwise, what is a reason to do all this overhead work with builder implementation?

With changes from this PR, the instance of the _DockerImageIdentifier_  can be created this way:

```
DockerImageIdentifier dockerImageIdentifier = DockerImageIdentifier.builder()
                                                                   .setDigest("digest")
                                                                   .setRegistry("registry")
                                                                   .setRepository("repository")
                                                                   .setTag("tag")
                                                                   .build(); 
```

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/parser/DockerImageIdentifier.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/parser/DockerImageIdentifier.java
@@ -29,7 +29,7 @@ import java.util.Objects;
  * <br>my_private_registry:15800/my_image1:latest
  * <br>my_private_registry:15800/my_image1@sha256:6b019df8c73bb42e606225ef935760b9c428521eba4ad2519ef3ff4cdb3dbd69
  *
- * @author Alexander Garaagtyi
+ * @author Alexander Garagatyi
  */
 public class DockerImageIdentifier {
     private final String  registry;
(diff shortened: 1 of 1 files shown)


plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/parser/DockerImageIdentifier.java
```diff
@@ -29,22 +29,19 @@
  * <br>my_private_registry:15800/my_image1:latest
  * <br>my_private_registry:15800/my_image1@sha256:6b019df8c73bb42e606225ef935760b9c428521eba4ad2519ef3ff4cdb3dbd69
  *
- * @author Alexander Garaagtyi
+ * @author Alexander Garagatyi
```
garagatyi, line 32: nice catch :)

garagatyi:
generally OK, but pass parameters in constructor instead of instance of builder.

alexVengrovsk:
@garagatyi I chenged everithing, you requested

garagatyi:
looks good, but not all of your commits are signed. Also please squash you commits into one.

alexVengrovsk:
@garagatyi I heve corrected all mistakes

garagatyi:
LGTM
@skabashnyuk Can you take a look?

skabashnyuk:
ok. this PR can be merged after 4.2 release.

PR #1114, opened 2016-04-22
CHE-1050: show console output on starting workspace runtime

akurinnoy:
**_1 Upvote**_ - fix console output
- add che-steps-container directive


ashumilova:
+1

PR #1119, opened 2016-04-22
CHE-1002 Add target UX improvements

vitaliy-guliy:

### DIFF
--- a/plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/targets/TargetsPresenter.java
+++ b/plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/targets/TargetsPresenter.java
@@ -340,35 +340,75 @@ public class TargetsPresenter implements TargetsView.ActionDelegate {
             return;
         }
 
+        // Update text of Connect / Disconnect button
         if (selectedTarget.isConnected()) {
             view.setConnectButtonText("Disconnect");
         } else {
             view.setConnectButtonText("Connect");
         }
-        view.enableConnectButton(!selectedTarget.isDirty());
 
-        view.enableCancelButton(selectedTarget.isDirty());
-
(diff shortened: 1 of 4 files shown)


plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/targets/TargetsPresenter.java
```diff
@@ -700,24 +740,32 @@ protected void onErrorReceived(Throwable exception) {
      * Ensures machine is started.
      */
     private void onConnected(final String machineId) {
-        machineService.getMachine(machineId).then(new Operation<MachineDto>() {
+        // There is a little bug in machine service on the server side.
+        // The machine info is updated with a little delay after running a machine.
+        // Using timer must fix the problem.
+        new Timer() {
```
vparfonov, line 746: Not clear about timer, how it solve problem? Why you set 500 not 600? Can accept it as urgent fix but must be fixed ASAP. Other ok for me

garagatyi, line 746: If you say that there is a bug in machine API than create an issue and assign it to me, please.

vitaliy-guliy, line 746: I'm not sure where is a problem. But when I receive a RUNNING state via websocket I ask the machine service about the machine and get CREATING state. Only waiting for a while gives the result.

vitaliy-guliy, line 746: Sure it's a temporary desigion and has to be fixed ASAP.

ashumilova:
+1

PR #1121, opened 2016-04-24
Code cleanup

alexVengrovsk:

### DIFF
--- a/plugins/plugin-svn/che-plugin-svn-ext-ide/src/main/java/org/eclipse/che/plugin/svn/ide/common/SubversionActionPresenter.java
+++ b/plugins/plugin-svn/che-plugin-svn-ext-ide/src/main/java/org/eclipse/che/plugin/svn/ide/common/SubversionActionPresenter.java
@@ -205,8 +205,9 @@ public class SubversionActionPresenter {
      */
     private void printOutput(List<String> output, SubversionOutputConsole console) {
         for (final String line : output) {
-            if (!line.trim().isEmpty()) {
-                String prefix = line.trim().substring(0, 1);
+            String trimLine = line.trim();
+            if (!trimLine.isEmpty()) {
+                String prefix = trimLine.substring(0, 1);
 
                 final String color = statusColors.getStatusColor(prefix);
                     if (color != null) {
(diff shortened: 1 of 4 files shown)


skabashnyuk:
ok to merge in 4.3

plugins/plugin-svn/che-plugin-svn-ext-ide/src/main/java/org/eclipse/che/plugin/svn/ide/common/SubversionOutputConsolePresenter.java
```diff
@@ -58,8 +59,7 @@ public void go(final AcceptsOneWidget container) {
     }
 
     public void print(@NotNull final String text) {
-        final String[] lines = text.split("\n");
-        for (String line : lines) {
+        for (String line : Splitter.on("\n").split(text)) {
```
garagatyi: Can you explain the purpose usage of Splitter instead of JRE API?

alexVengrovsk: @garagatyi I made this change, because of increasing of productivity (according to some benchmarks http://demeranville.com/battle-of-the-tokenizers-delimited-text-parser-performance/ and https://dzone.com/articles/guava-splitter-vs-stringutils), which gives Splitter's split() compare to String's split(). As Guava was already used in this project, there is no additional overhead of using it here.

garagatyi: I don't think we should use that as 
- it is not obvious and others can replace it back with default splitting later
- It will be compiled into JavaScript so these benchmarks are useless in this case.

skabashnyuk: +

alexVengrovsk:
@garagatyi Hi. I excluded Guava's _Splitter_ from _JavaFqnResolver_

garagatyi:
LGTM

PR #1126, opened 2016-04-25
Added new box + proxy config

TylerJewell:
**_1 Upvote**_ 1. Adds in proxy variables at the top of the vagrantfile for user.
2. Updates .bashrc inside VM
3. Updates Docker configuration to have daemon use proxy.
4. Configures che.properties to allow workspaces to use proxy.
5. Changes out the vbox to an image with java in it to simplify installation


riuvshin:
+1

PR #1132, opened 2016-04-26
CHE-450 do not fail if impossible to resolve variable in environment props

mshaposhnik:
Signed-off-by: Max Shaposhnik mshaposhnik@codenvy.com

### DIFF
--- a/core/commons/che-core-commons-inject/src/main/java/org/eclipse/che/inject/CheBootstrap.java
+++ b/core/commons/che-core-commons-inject/src/main/java/org/eclipse/che/inject/CheBootstrap.java
@@ -23,6 +23,8 @@ import org.eclipse.che.inject.lifecycle.DestroyModule;
 import org.eclipse.che.inject.lifecycle.Destroyer;
 import org.eclipse.che.inject.lifecycle.InitModule;
 import org.everrest.guice.servlet.EverrestGuiceContextListener;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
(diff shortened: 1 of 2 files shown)


core/commons/che-core-commons-inject/src/main/java/org/eclipse/che/inject/CheBootstrap.java
```diff
@@ -239,11 +243,16 @@ protected void bindProperties(String prefix, Map<String, String> properties) {
                             if (actual == null) {
                                 actual = System.getenv(name);
                             }
-                            if (actual == null) {
+                            if (!skipUnresolved && actual == null) {
```
sleshchenko: We can rewrite it to avoid double check actual value

```
if (actual != null) {
    buf.append(actual);
} else if (skipUnresolved) {
    buf.append(pValue.substring(i, j));
    LOG.warn("Environment variable " + name + " cannot be resolved, leaving as is.");
} else {
    throw new ConfigurationException(...);
}
```

WDYT?

mshaposhnik: Ok

skabashnyuk:
ok @garagatyi ?

core/commons/che-core-commons-inject/src/main/java/org/eclipse/che/inject/CheBootstrap.java
```diff
@@ -239,11 +243,16 @@ protected void bindProperties(String prefix, Map<String, String> properties) {
                             if (actual == null) {
                                 actual = System.getenv(name);
                             }
-                            if (actual == null) {
+                            if (actual != null) {
```
garagatyi, line 246: you can use **else** instead of **if (actual != null)**

sleshchenko, line 246: I think I will have +1 level nesting in your case. Can you provide full example?

garagatyi, line 246: Sorry, you can't

garagatyi:
LGTM

PR #1135, opened 2016-04-26
add logging of OOM of not PID1 processes in docker containers

garagatyi:
This PR helps to detect OOM of wsagent
@skabashnyuk @riuvshin


skabashnyuk:
ок

PR #1146, opened 2016-04-27
Decouple shared modules or user, machine, workspace, factory

mshaposhnik:


wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/model/impl/stack/StackSourceImpl.java
```diff
@@ -4,13 +4,14 @@
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-v10.html
- *
+ * <p>
```
azatsarynnyy: something wrong with copyright formatting

core/platform-api-client-gwt/che-core-client-gwt-workspace/pom.xml
```diff
@@ -75,6 +75,11 @@
             <artifactId>gwt-user</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-workspace</artifactId>
```
sleshchenko, line 80: Can you explain why we need this dependency in provided scope?

mshaposhnik, line 80: For javadocs. Example https://github.com/eclipse/che/blob/master/core/platform-api-client-gwt/che-core-client-gwt-workspace/src/main/java/org/eclipse/che/api/workspace/gwt/client/WorkspaceServiceClient.java#L41

skabashnyuk:
ок

akorneta:
ok

PR #1156, opened 2016-04-29
Dashboard fixes

akurinnoy:
**_1 Upvote**_ - disable menu items workspaces list after all workspaces were deleted
- add checks for uniqueness of workspace's and project's name on updating
- fix button's borders


ashumilova:
+1

PR #1157, opened 2016-04-29
Che 1076 Refactoring of che-core-api-infrastructure-local

skabashnyuk:
**_1 Upvote**_ -  Split che-core-api-infrastructure-local to  /assembly/wsagent-local /assembly/wsmaster-local
-  moved to /assembly/wsagent-local /assembly/wsmaster-local classes from wars
-  Removed unused or deprecated code


benoitf:
it looks like strange to have many classes as part of an assembly module.
Assembly in maven is a result of taking dependencies, http://maven.apache.org/plugins/maven-assembly-plugin/ , not defining new modules

skabashnyuk:
What do you propose?

benoitf:
it may be use existing https://github.com/eclipse/che/tree/master/wsagent and https://github.com/eclipse/che/tree/master/wsmaster modules ?

skabashnyuk:
ok

benoitf:
+1 on my side

vparfonov:
Ok

PR #1164, opened 2016-04-30
Fix auto extension work

TylerJewell:
@riuvshin @eivantsov -- please OK.  If you run the InstallExtension utility twice, and the second utility does not have an extension to write, it will erase the results of the first execution.  The PR cleans up the issue.


ghost:
@TylerJewell it worked for me
+1

PR #1168, opened 2016-05-01
Fix source storage field naming

akorneta:
**_1 Upvote**_ @evoevodin @skabashnyuk @gazarenkov


gazarenkov:
OK
And remove "modules" field and related methods, it is not needed anymore

voievodin:
:+1:

PR #1176, opened 2016-05-04
Add VM mount to backup workspaces and prefs on host file system

ghost:
**_1 Upvote**_ @TylerJewell @riuvshin please take a look. This way we auto mount a current dir into `~/che-storage` which is configured as storage for Che workspaces and prefs inside a VM. As a result, even when a box is destroyed, workspaces, projects and prefs are there.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -17,6 +17,7 @@ Vagrant.configure(2) do |config|
   config.vm.box_download_insecure = true
   config.ssh.insert_key = false
   config.vm.network :private_network, ip: "192.168.28.30"
+  config.vm.synced_folder ".", "/home/vagrant/.che"
   config.vm.define "che" do |che|
   end
   config.vm.provider "virtualbox" do |vb|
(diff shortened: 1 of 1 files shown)


riuvshin:
LGTMN

TylerJewell:
Three basic changes proposed.  
1.  Let's name the directory /home/vagrant/.che.
2.  Let's copy /conf/che.properties to .che/che.properties.
3.  Let's update the properties in that file instead of appending.

riuvshin:
+1 on this

ghost:
@TylerJewell @riuvshin I have updated the PR. che.properties is copied and updated now in ~/.che which is set to be the location of CHE_LOCAL_CONF_DIR

TylerJewell:
@eivantsov LGTM.  Question - did you not need to source .bashrc after you modify the file?  I am surprised that the modification to the environment variable was picked up.

ghost:
It's picked up since I both export it in the command line and in .bashrc. So, it works for Vagrant because of export in the command line, and for user (when logged in) in .bashrc.

TylerJewell:
Great - will you consider doing a similar PR for the ARTIK IDE Vagrantfile?

ghost:
yes, sure

PR #1184, opened 2016-05-04
CODENVY-151: Stop all running workspaces within 1 minute

mmorhun:
**_1 Upvote**_ Signed-off-by: Mykola Morhun mmorhun@codenvy.com

### DIFF
--- a/wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
+++ b/wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
@@ -1011,7 +1011,7 @@ public class MachineManager {
         executor.shutdown();
 
         try {
-            for (MachineImpl machine : machineRegistry.getMachines()) {
+            machineRegistry.getMachines().parallelStream().forEach(machine -> {
                 try {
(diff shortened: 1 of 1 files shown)


mmorhun:
@garagatyi check please

garagatyi:
@evoevodin @sleshchenko @skabashnyuk WDYT?

sleshchenko:
LGTM

AndrienkoAleksandr:
+1

garagatyi:
-1

voievodin:
-1 

I don't think that it is a right way of destroying the machines.
1. You don't control the count of threads(so it is not guaranteed that you  shutdown each machine in the separate thread)
2. `.parallelStream` uses `ForkJoinPool.commonPool()` -> you are not waiting for machines to be destroyed. See `commonPool` documentation for more information. 

I suggest you to shutdown machines asynchronously each in the separate thread, to achieve this goal you can either reuse an existing executor + use `CountDownLatch`, or create a separate executor to perform machines destroy.

A few things to respect:
1. Each machine must be destroyed in the separate thread
2. You must wait until all the machines destroyed

garagatyi:
@mmorhun since @evoevodin and me agreed that machine destroying should be implemented in a different way, please close this issue.

PR #1190, opened 2016-05-05
CODENVY-151: Stop all running workspaces within 1 minute

mmorhun:

### DIFF
--- a/wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
+++ b/wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
@@ -1009,23 +1009,41 @@ public class MachineManager {
     private void cleanup() {
         eventService.unsubscribe(machineCleaner);
 
+        boolean interrupted = false;
+
         executor.shutdown();
 
+        final ExecutorService destroyMachinesExecutor =
+                Executors.newFixedThreadPool(2 * Runtime.getRuntime().availableProcessors(),
+                                             new ThreadFactoryBuilder().setNameFormat("DestroyMachine-%d")
+                                                                       .setDaemon(false)
+                                                                       .build());
         try {
             for (MachineImpl machine : machineRegistry.getMachines()) {
-                try {
-                    destroy(machine.getId(), false);
-                } catch (NotFoundException ignore) {
-                    // it is ok, machine has been already destroyed
-                } catch (Exception e) {
-                    LOG.warn(e.getMessage());
+                destroyMachinesExecutor.execute(() -> {
+                    try {
+                        destroy(machine.getId(), false);
+                    } catch (NotFoundException ignore) {
+                        // it is ok, machine has been already destroyed
+                    } catch (Exception e) {
+                        LOG.warn(e.getMessage());
+                    }
+                });
+            }
+            destroyMachinesExecutor.shutdown();
+            if (!destroyMachinesExecutor.awaitTermination(50, TimeUnit.SECONDS)) {
+                destroyMachinesExecutor.shutdownNow();
+                if (!destroyMachinesExecutor.awaitTermination(10, TimeUnit.SECONDS)) {
+                    LOG.warn("Unable terminate destroy machines pool");
                 }
             }
+        } catch (InterruptedException e) {
+            interrupted = true;
+            destroyMachinesExecutor.shutdownNow();
         } catch (MachineException e) {
             LOG.error(e.getLocalizedMessage(), e);
         }
 
-        boolean interrupted = false;
         try {
             if (!executor.awaitTermination(10, TimeUnit.SECONDS)) {
                 executor.shutdownNow();


mmorhun:
@garagatyi take a look please

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
```diff
@@ -1010,21 +1010,37 @@ private void cleanup() {
...
                 }
             }
+        } catch (InterruptedException e) {
+            interrupted = true;
+            destroyMachinesExecutor.shutdownNow();
         } catch (MachineException e) {
             LOG.error(e.getLocalizedMessage(), e);
         }
 
-        boolean interrupted = false;
```
garagatyi: Why you've removed that line?

garagatyi: Sorry. I see

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
```diff
@@ -1010,21 +1010,37 @@ private void cleanup() {
...
+                                          .build());
         try {
             for (MachineImpl machine : machineRegistry.getMachines()) {
-                try {
-                    destroy(machine.getId(), false);
-                } catch (NotFoundException ignore) {
-                    // it is ok, machine has been already destroyed
-                } catch (Exception e) {
-                    LOG.warn(e.getMessage());
+                destroyMachinesExecutor.execute(ThreadLocalPropagateContext.wrap(() -> {
```
garagatyi: Why you wrap to threadlocalpropagate context?

garagatyi:
@evoevodin @sleshchenko Please take a look

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
mmorhun: You are right, it is redundant here

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
```diff
@@ -1010,21 +1010,37 @@ private void cleanup() {
 
         executor.shutdown();
 
+        boolean interrupted = false;
+        final ExecutorService destroyMachinesExecutor = Executors.newCachedThreadPool(
```
voievodin: Can you explain the choice of `CachedThreadPool` executor service?

mmorhun: I used `CachedThreadPool`, because it 

> creates new threads as needed, but will reuse previously constructed threads when they are available.

Do you have a suggestion?

voievodin: The suggestion is to use `FixedThreadPool` executor service instead, as you can control the count of threads(btw it caches them as well). What if you have 100 machines running ?  Also, i think that it is necessary to increase termination timeout, or at least make it configurable along with count of threads.

mmorhun: According to [this](https://github.com/eclipse/che/pull/1184):

> Each machine must be destroyed in the separate thread

I see no difference here between `FixedThreadPool` and `CachedThreadPool`.
It even doesn't depends on number of machines: 100, 1000 ... all the same. (from `*ThreadPool` point of view)

mmorhun: It is logically calculate termination timeout using number of machines, but note, that task is terminate it all in one minute! (otherwise tomcat just be killed).
@evoevodin @garagatyi @sleshchenko WDYT?

garagatyi:
What if we just use some fixed, and maybe configurable, amount of threads? It can be high enough number, lets say 20.
@evoevodin @sleshchenko @mmorhun WDYT?

voievodin:
Well that was my point, @garagatyi +1

sleshchenko:
Also it would be better to make configurable timeout that equals to one minute for now

mmorhun:
@garagatyi review please

garagatyi:
LGTM
@sleshchenko @evoevodin WDYT?

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
```diff
@@ -1037,6 +1023,37 @@ private void cleanup() {
             executor.shutdownNow();
         }
 
+        final ExecutorService destroyMachinesExecutor =
+                Executors.newFixedThreadPool(10 * Runtime.getRuntime().availableProcessors(),
```
voievodin: I think that `10 * Runtime.getRuntime().availableProcessors()` is too many :open_mouth: :gun: , `2 * Runtime.getRuntime().availableProcessors()` or even  `Runtime.getRuntime().availableProcessors()` will be more than enough.

mmorhun:
@evoevodin @skabashnyuk ?

skabashnyuk:
ok

voievodin:
But the time of `awaitTermination` is still not configurable, is it okay @mmorhun ?

mmorhun:
@evoevodin, it is ok for now. I think `awaitTermination` time is redundant here, because even admin not know total quantity of workspaces and/or machines running simultaneously. The system should be able to stop without such parameters. It means that we need some refactoring to stop all workspaces regardless of number of running machines.

voievodin:
well, if you say so, then okay.

LGTM

PR #1195, opened 2016-05-06
Don't store WorkspaceRuntime in local implementation

voievodin:
Workspace API responds `runtime` instance along with workspace instance when workspace instance is requested even if workspace is stopped, it happens because runtime is not ignored by inmemory implementation while it must be.

@skabashnyuk please review


skabashnyuk:
ок

Issue #366, opened 2016-02-17
che project maven install error


blingeee:
l inserted 'mvn clean install' in terminal for building che project.
but error occurred like below..

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.18.1:test (default-test) on project che-core-api-machine: There are test failures.
[ERROR] 
[ERROR] Please refer to /home/ubuntu/che/core/platform-api/che-core-api-machine/target/surefire-reports for the individual test results.
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
[ERROR] 
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <goals> -rf :che-core-api-machine

```

ddementieva:
Can you share an entire build log with us?

ghost:
@blingeee can you try it with a current master of the latest release?

bepurushoth:
I have the same issue. Tried with tag 4.0.0-RC6. Any solution to this issue?

bepurushoth:
tried with the master also and got the same issue

ghost:
@bepurushoth  What is your OS and Maven version?

bepurushoth:
Windows 10 64 bit and Maven version is apache-maven-3.3.9

logs below:
Running TestSuite
Configuring TestNG with: TestNG652Configurator
2016-02-25 15:52:29,581[main]             [INFO ] [org.eclipse.jetty.util.log 186]      - Logging initialized @1251ms
2016-02-25 15:52:29,699[main]             [INFO ] [o.eclipse.jetty.server.Server 345]   - jetty-9.3.1.v20150714
2016-02-25 15:52:29,864[main]             [INFO ] [o.e.j.s.handler.ContextHandler 737]  - Started o.e.j.s.ServletContextHandler@7cc0cdad{/,null,AVAILABLE}
2016-02-25 15:52:29,936[main]             [INFO ] [o.e.j.server.ServerConnector 270]    - Started ServerConnector@7c6908d7{HTTP/1.1,[http/1.1]}{0.0.0.0:10162}
[...]

PR #1204, opened 2016-05-09
createMachineLogsDir fix for issue #366

ndlu2:
The following change fixes one of the failing tests for issue #366. The mkdirs() function in File returns false when the file already exists which caused the test to throw an exception. To fix this problem I added a check to see if the directory already exists before throwing the exception.

This change does not fix all of the issues when trying to build Eclipse Che on Windows.

Signed-off-by: Nicholas Lu ndlu2@illinois.edu


garagatyi:
LGTM
@skabashnyuk Please upvote

PR #1214, opened 2016-05-10
Codenvy 426 show dedicated message when clone fails due to lack of SSH key

mshaposhnik:


mshaposhnik:
@skabashnyuk @vparfonov pls check

skabashnyuk:
ок

vparfonov:
ok

PR #1221, opened 2016-05-10
Improvements

TylerJewell:
1. Fixed errors with command substitution of the che properties file.
2. Added in logic to reboot che server if VM was poorly suspended.
3. Allow users to choose latest vs. nightly vs. tagged version with flag in Vagrantfile
4. Allow users to change the IP address used in the Vagrantfile

@eivantsov - please check.  Authored in response to GitHub issue CHE-1207

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -24,7 +24,7 @@ Vagrant.configure(2) do |config|
   end
   config.vm.provider "virtualbox" do |vb|
     vb.memory = "4096"
-    vb.name = "eclipse-che-vm2"
+    vb.name = "eclipse-che-vm"
   end
 
   $script = <<-SHELL
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -138,7 +138,7 @@ Vagrant.configure(2) do |config|
         export JAVA_HOME=/usr &>/dev/null
         echo vagrant | sudo -S -E -u vagrant /home/vagrant/eclipse-che-*/bin/che.sh --remote:${IP} --skip:client -g start &>/dev/null
       fi
-      # If we are not awake after 180 seconds, restart server
+      # If we are not awake after 180 seconds, exit with failure
       if [ $counter == "35" ]; then
         echo "---------------------------------------------"
         echo "."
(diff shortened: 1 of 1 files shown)


ghost:
@TylerJewell worked for me

Vagrantfile
```diff
@@ -61,49 +65,91 @@ Vagrant.configure(2) do |config|
...
+        echo "ECLIPSE IDE: SERVER NOT RESPONSIVE -- REBOOTING"
+        echo "."
+        echo "-----------------------------------------------"
+        export JAVA_HOME=/usr &>/dev/null
+        echo vagrant | sudo -S -E -u vagrant /home/vagrant/eclipse-che-*/bin/che.sh --remote:$IP --skip:client -g start &>/dev/null
+      fi
+
+      let counter=counter+1
+      sleep 5
+    done
```
riuvshin, line 153: what if server will not be able to boot at all? there should be exit after some amount of tries.
Also on some environments server may start longer than 30 secs and in that case you will be in a endless loop with wait starting & reboot server. I think we have to let server few mins to start and than check is it running, try restart and if restart didn;t help we have to exit that loop.

TylerJewell, line 153: Yep +1 - added.

Vagrantfile
```diff
@@ -61,49 +61,101 @@ Vagrant.configure(2) do |config|
...
+        echo "-----------------------------------------------"
+        echo "."
+        echo "ECLIPSE CHE: SERVER NOT RESPONSIVE -- REBOOTING"
+        echo "."
+        echo "-----------------------------------------------"
+        export JAVA_HOME=/usr &>/dev/null
+        echo vagrant | sudo -S -E -u vagrant /home/vagrant/eclipse-che-*/bin/che.sh --remote:${IP} --skip:client -g start &>/dev/null
+      fi
+      # If we are not awake after 180 seconds, restart server
+      if [ $counter == "35" ]; then
```
TylerJewell, line 142: fixed - thanks

riuvshin:
LGTM

PR #1222, opened 2016-05-11
CHE-909: Stop additional machines when stopping workspace

mmorhun:


mmorhun:
@garagatyi @evoevodin check please

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/event/MachineStateListener.java
```diff
@@ -62,6 +70,12 @@ public void onEvent(MachineStatusEvent event) {
             } catch (NotFoundException | ServerException | ConflictException exception) {
                 LOG.error(exception.getLocalizedMessage(), exception);
             }
+        } else if (CREATING.equals(event.getEventType()) && !event.isDev()) {
+            try {
+                workspaceManager.getWorkspace(workspaceId).getRuntime().addMachine(machineManager.getMachine(event.getMachineId()));
```
voievodin: It will not affect the runtime instance, see [javadoc](https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java#L95-L100)

voievodin:
-1. It seems that proposed solution doesn't work. Have you tested it?

mmorhun:
@evoevodin, of course, I tested this solution. It works!

garagatyi:
test-ci-build

mmorhun:
@evoevodin I fixed it, check please

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -245,6 +245,34 @@ public void stop(String workspaceId) throws NotFoundException, ServerException,
...
+     *         id of workspace to which runtime machine should be added
+     * @param machine
+     *         machine to add to specified runtime
+     * @throws ConflictException
+     *         if given machine already exists in specified runtime
+     * @throws NotFoundException
+     *         if workspace with specified id not exists
+     */
+    public void addMachineIntoRuntime(String workspaceId, MachineImpl machine) throws ConflictException, NotFoundException {
+        final RuntimeDescriptor descriptor = descriptors.get(workspaceId);
```
voievodin: please use the appropriate lock to get the descriptor.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -245,6 +245,34 @@ public void stop(String workspaceId) throws NotFoundException, ServerException,
...
+     * @param workspaceId
+     *         id of workspace to which runtime machine should be added
+     * @param machine
+     *         machine to add to specified runtime
+     * @throws ConflictException
+     *         if given machine already exists in specified runtime
+     * @throws NotFoundException
+     *         if workspace with specified id not exists
+     */
+    public void addMachineIntoRuntime(String workspaceId, MachineImpl machine) throws ConflictException, NotFoundException {
```
voievodin: - I don't like the method name, `addMachine`(`injectMachine` maybe) is more than enough to understand what it does, for what reason `workspaceId` is used (p.s. see machine#getWorkspaceId)? 
- Locks used wrong(or event not used at all)
- The method doesn't check the workspace runtime state while it MUST do it, as what is the purpose of adding the machine to the starting workspace? 
- The method doesn't respect the current state of the `isPreDestroyInvoked` flag.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -245,6 +245,34 @@ public void stop(String workspaceId) throws NotFoundException, ServerException,
...
+        final RuntimeDescriptor descriptor = descriptors.get(workspaceId);
+        if (descriptor == null) {
+            throw new NotFoundException("No runtime with \"" + workspaceId + "\" id");
```
voievodin: What is wrong with the message in the `get(...)` method?, or you want the same component to respond differently on the same use cases?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -245,6 +245,34 @@ public void stop(String workspaceId) throws NotFoundException, ServerException,
...
+        }
+
+        rwLock.readLock().lock();
```
voievodin: Why do you use read lock?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/model/impl/WorkspaceRuntimeImpl.java
```diff
@@ -94,6 +95,13 @@ public void setMachines(List<MachineImpl> machines) {
         this.machines = machines;
     }
 
+    public synchronized void addMachine(MachineImpl machine) throws ConflictException {
```
voievodin: 1. This synchronization is ugly as long as you have another methods related to `this.machines` not synchornized, but even if you have those synchronizations it is still ugly. Remove synchronized from the method.
2. `ConflictException` ? did you see any model implementation which throws checked exceptions? i guess no.
   Please remove `throws ConflictException`. I don't think that validations should be performed on this object level, as we have certain places for it(btw validation is also ugly as it may be performed in more lightweight way, and is it weak as it doesn't check all the things(e.g. that machine is not dev one)).

So basically you can add the `addMachine` method to the `WorkspaceRuntimeImpl` but all what this method should do is to add the machine to the list of the machines.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -245,6 +245,34 @@ public void stop(String workspaceId) throws NotFoundException, ServerException,
...
+        try {
+            WorkspaceRuntimeImpl workspaceRuntime = descriptor.getRuntime();
+            workspaceRuntime.addMachine(machine);
+            descriptors.put(workspaceId, new RuntimeDescriptor(workspaceRuntime));
```
voievodin: Why do you put the copy of the workspace runtime intance?

voievodin:
I reviewed `WorkspaceRuntimes` and `WorkspaceRuntime` objects. **-1 again**.
Except of the commens i wrote you should consider adding the same method to the `WorkspaceManager` as it is the place where we perform business logic validation + `WorkspaceRuntimes` is the internal component which should not be used across the app. 

`WorkspaceRuntimes` is one of the most important and central classes in the workspace infrastructure its tests coverage is `90%+`, when you change something, please write the tests.

mmorhun:
@evoevodin take a look please

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -487,6 +487,29 @@ public void createSnapshot(String workspaceId) throws NotFoundException, ServerE
...
+     *
+     * @param machine
+     *         machine to add to specified runtime
+     * @throws ConflictException
+     *         if given machine already exists in specified runtime
+     *         or if machine is dev-machine
+     * @throws NotFoundException
+     *         if machine's workspace hasn't runtime
+     */
+    public void injectMachine(MachineImpl machine) throws ConflictException, NotFoundException, ServerException {
```
voievodin: I remember that i suggested you to consider either `addMachine` or `injectMachine` now i see that, `addMachine` is actually better name, just because we will need another method for removing machine from the running workspace, so there is no good opposite word for `inject`, but there is a good opposite word for `add`. Can you please rename it to the `addMachine`.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -487,6 +487,29 @@ public void createSnapshot(String workspaceId) throws NotFoundException, ServerE
...
+        if (machine.getConfig().isDev()) {
```
voievodin: What if machine is `null`? See the other methods of the `WorkspaceManager`.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -487,6 +487,29 @@ public void createSnapshot(String workspaceId) throws NotFoundException, ServerE
...
+            throw new ConflictException("Cannot add another dev-machine");
```
voievodin: The message is not informative enough,  you should identify machine along with the workspace.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -487,6 +487,29 @@ public void createSnapshot(String workspaceId) throws NotFoundException, ServerE
...
+        }
+        String workspaceId = machine.getWorkspaceId();
+        if (runtimes.get(workspaceId).getRuntimeStatus() != WorkspaceStatus.RUNNING) {
```
voievodin: You can't guarantee that it is true unless you are in the lock, so this is not the right place for such check, move it to the `WorkspaceRuntimes` method.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -245,6 +245,34 @@ public void stop(String workspaceId) throws NotFoundException, ServerException,
...
+     * @throws ServerException
+     *         if application server is stopping
+     */
+    public void injectMachine(MachineImpl machine) throws NotFoundException, ServerException {
+        ensurePreDestroyIsNotExecuted();
+
+        String workspaceId = machine.getWorkspaceId();
+
+        rwLock.writeLock().lock();
+        try {
```
voievodin: Repeat `ensurePreDestroyIsNotExecuted` check

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -487,6 +487,29 @@ public void createSnapshot(String workspaceId) throws NotFoundException, ServerE
...
+    /**
+     * Adds machine into workspace runtime.
+     *
+     * @param machine
+     *         machine to add to specified runtime
+     * @throws ConflictException
+     *         if given machine already exists in specified runtime
+     *         or if machine is dev-machine
+     * @throws NotFoundException
+     *         if machine's workspace hasn't runtime
```
voievodin: This sentence is quiet confusing, rephrase it. btw try to follow the class style in the case of documentation.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -487,6 +487,29 @@ public void createSnapshot(String workspaceId) throws NotFoundException, ServerE
         return machineManager.getSnapshots(sessionUser().getId(), workspaceId);
     }
 
+    /**
+     * Adds machine into workspace runtime.
```
voievodin: I think that you should add more useful information in the javadoc. e.g. What if workspace is stopped, will this machine be stopped with the workspace. Will this method affect the workspace configuration?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -487,6 +487,29 @@ public void createSnapshot(String workspaceId) throws NotFoundException, ServerE
...
+     * @throws NotFoundException
+     *         if machine's workspace hasn't runtime
+     */
+    public void injectMachine(MachineImpl machine) throws ConflictException, NotFoundException, ServerException {
+        if (machine.getConfig().isDev()) {
+            throw new ConflictException("Cannot add another dev-machine");
+        }
+        String workspaceId = machine.getWorkspaceId();
+        if (runtimes.get(workspaceId).getRuntimeStatus() != WorkspaceStatus.RUNNING) {
+            throw new ConflictException("Cannot add machine to not running workspace");
```
voievodin: See the [comment](https://github.com/eclipse/che/pull/1222/files#r63170113)

voievodin:
Please cover the added to `WorkspaceManage` code with tests.
Please add yourself as author to the `WorkspaceRuntimes` and `WorkspaceManager`.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
voievodin: `ServerException` is not documented

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/event/DestroyedMachineEventListener.java
```diff
@@ -30,37 +30,38 @@
...
-    public MachineStateListener(WorkspaceManager workspaceManager, EventService eventService) {
+    public DestroyedMachineEventListener(WorkspaceManager workspaceManager, EventService eventService) {
         this.workspaceManager = workspaceManager;
         this.eventService = eventService;
     }
 
     @Override
     public void onEvent(MachineStatusEvent event) {
-        String workspaceId = event.getWorkspaceId();
+        if (DESTROYED.equals(event.getEventType())) {
```
garagatyi: Why you have separated these checks? Inner conditional operator doesn't make code clearer

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/event/CreateMachineEventListener.java
```diff
@@ -0,0 +1,86 @@
...
+                                      WorkspaceManager workspaceManager) {
+        this.eventService = eventService;
+        this.machineManager = machineManager;
+        this.workspaceManager = workspaceManager;
+    }
+
+    @Override
+    public void onEvent(MachineStatusEvent event) {
+        if (CREATING.equals(event.getEventType())) {
+            if (!event.isDev()) {
```
garagatyi: merge checks

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/event/CreateMachineEventListener.java
```diff
@@ -0,0 +1,86 @@
...
+
+import static org.eclipse.che.api.machine.shared.dto.event.MachineStatusEvent.EventType.CREATING;
+
+/**
+ * The class listens changing of machine status and perform some actions when new machine is creating.
+ *
+ * @author Mykola Morhun
+ */
+@Singleton
+public class CreateMachineEventListener implements EventSubscriber<MachineStatusEvent> {
```
voievodin: I think that event listener must say something about the action which it performs.
I this case it does the only one certain action and i think class name should display it.
wdyt @garagatyi ?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/event/CreateMachineEventListener.java
```diff
@@ -0,0 +1,86 @@
...
+                                      MachineManager machineManager,
+                                      WorkspaceManager workspaceManager) {
+        this.eventService = eventService;
+        this.machineManager = machineManager;
+        this.workspaceManager = workspaceManager;
+    }
+
+    @Override
+    public void onEvent(MachineStatusEvent event) {
+        if (CREATING.equals(event.getEventType())) {
```
garagatyi: Why you add machine in CREATING state? It is quite possible that machine creation will fail.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/event/CreateMachineEventListener.java
```diff
@@ -0,0 +1,86 @@
...
+            if (!event.isDev()) {
+                try {
+                    MachineImpl machine = machineManager.getMachine(event.getMachineId());
+                    workspaceManager.injectMachine(machine);
+                } catch (NotFoundException | ServerException | ConflictException exception) {
+                    LOG.error(exception.getLocalizedMessage(), exception);
+                    try {
+                        machineManager.destroy(event.getMachineId(), true);
```
voievodin: What is the purpose of `machineManager.destroy` invocation if `machineManager.getMachine` threw an exception?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/event/CreateMachineEventListener.java
```diff
@@ -0,0 +1,86 @@
...
+        this.workspaceManager = workspaceManager;
+    }
+
+    @Override
+    public void onEvent(MachineStatusEvent event) {
+        if (CREATING.equals(event.getEventType())) {
+            if (!event.isDev()) {
+                try {
+                    MachineImpl machine = machineManager.getMachine(event.getMachineId());
+                    workspaceManager.injectMachine(machine);
```
garagatyi: Have you checked what will happen if machine is stopped (non-dev) while workspace is in RUNNING state?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/model/impl/WorkspaceRuntimeImpl.java
```diff
@@ -94,6 +94,10 @@ public void setMachines(List<MachineImpl> machines) {
         this.machines = machines;
     }
 
+    public void addMachine(MachineImpl machine) {
+        this.machines.add(machine);
```
voievodin: Can you please rework the [code in constructor](https://github.com/eclipse/che/blob/6ba34531215c09e98cb470937bdb32edaaff8f54/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/model/impl/WorkspaceRuntimeImpl.java#L50-L52) to be npe safe.

``` java
if (machines != null) {
    this.machines = machines.stream()
                            .map(MachineImpl::new)
                            .collect(toList());
}
```

and then reuse the getter `getMachines().add(machine)`, in this way the implementation becomes fully npe safe .

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/event/DestroyedMachineEventListener.java
mmorhun: I separated that checks here because this event listener should process only `DESTROYED` machine state. The fact of processing only dev-machine is internal logic of listener.
The same for `CreateMachineEventListener`

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/event/CreateMachineEventListener.java
garagatyi: +1

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/event/CreateMachineEventListener.java
mmorhun: It will be covered in another issue.

mmorhun:
@garagatyi take a look please

assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterModule.java
```diff
@@ -71,7 +73,8 @@ protected void configure() {
         bind(org.eclipse.che.api.workspace.server.WorkspaceValidator.class)
                 .to(org.eclipse.che.api.workspace.server.DefaultWorkspaceValidator.class);
 
-        bind(org.eclipse.che.api.workspace.server.event.MachineStateListener.class).asEagerSingleton();
+        bind(StopWorkspaceOnDestroyDevMachine.class).asEagerSingleton();
+        bind(AddMachineIntoWorkspaceRuntime.class).asEagerSingleton();
```
garagatyi: When it adds machine?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
garagatyi: +1

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/event/AddMachineIntoWorkspaceRuntime.java
```diff
@@ -0,0 +1,94 @@
...
+                                          WorkspaceManager workspaceManager) {
+        this.eventService = eventService;
+        this.machineManager = machineManager;
+        this.workspaceManager = workspaceManager;
+    }
+
+    @Override
+    public void onEvent(MachineStatusEvent event) {
+        if (RUNNING.equals(event.getEventType())) {
+            if (!event.isDev()) {
```
garagatyi: What if non-dev machine is included in workspace configuration?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -245,6 +245,42 @@ public void stop(String workspaceId) throws NotFoundException, ServerException,
...
+     * @param machine
+     *         machine to add to specified runtime
+     * @throws NotFoundException
+     *         when workspace with specified id not exists
+     * @throws ServerException
+     *         when application server is stopping
+     * @throws ConflictException
+     *         when workspace is not running
+     */
+    public void addMachine(MachineImpl machine) throws NotFoundException, ServerException, ConflictException {
```
garagatyi: Consider addition yourself to the authors

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -487,6 +487,31 @@ public void createSnapshot(String workspaceId) throws NotFoundException, ServerE
...
+     * @param machine
+     *         machine to add to specified runtime
+     * @throws ConflictException
+     *         when given machine already exists in specified runtime
+     *         or when machine is dev-machine
+     * @throws NotFoundException
+     *         when workspace in which machine is adding doesn't have runtime
+     * @throws ServerException
+     *         when application server is stopping
+     * @throws NullPointerException
```
garagatyi: Consider addition yourself to the authors

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/model/impl/WorkspaceRuntimeImpl.java
```diff
@@ -47,9 +47,11 @@ public WorkspaceRuntimeImpl(String activeEnv,
         if (devMachine != null) {
             this.devMachine = new MachineImpl(devMachine);
         }
-        this.machines = machines.stream()
-                                .map(MachineImpl::new)
-                                .collect(toList());
+        if (machines != null) {
```
garagatyi, line 50: Is it possible that collection will be null? It is quite unusual for Che code base

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/event/AddMachineIntoWorkspaceRuntime.java
```diff
@@ -0,0 +1,94 @@
...
+                MachineImpl machine;
+                try {
+                    machine = machineManager.getMachine(event.getMachineId());
+                } catch (NotFoundException | MachineException exception) {
+                    LOG.error(exception.getLocalizedMessage(), exception);
+                    return;
+                }
+                try {
+                    workspaceManager.addMachine(machine);
```
garagatyi: What will happen with workspace stop if user stop that machine over API before stopping the workspace

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/model/impl/WorkspaceRuntimeImpl.java
mmorhun, line 50: May be, but evoevodin asked me to add this

garagatyi, line 50: @evoevodin can you comment?

voievodin, line 50: It is possible to pass the null to the POJO constructor or setter where collection is required, BUT the getter of the collection must return non-null collection anyway [example](https://github.com/eclipse/che/blob/21d59d72cd8f6c7636f7f679d8208ca8fa913475/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/model/impl/EnvironmentImpl.java#L42-L46) that's how we avoid npes is such usecases.
This change is a good one and should come with a getter changes as well.

garagatyi, line 50: yep, sorry I missed that it is a constructor

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -245,6 +246,74 @@ public void stop(String workspaceId) throws NotFoundException, ServerException,
...
+        try {
+            final RuntimeDescriptor descriptor = descriptors.get(workspaceId);
+            if (descriptor == null) {
+                throw new NotFoundException("Workspace with id '" + workspaceId + "' is not running.");
+            }
+
+            List<MachineImpl> machines = descriptor.getRuntime().getMachines();
+            for (int i = 0; i < machines.size(); i++) {
+                if (machine.getId().equals(machines.get(i).getId())) {
+                    machines.remove(i);
```
voievodin: reuse `rmFirst`

NicholasMorhun: @evoevodin what if machine isn't first?

mmorhun: @evoevodin what if machine isn't first?

voievodin: it will be removed!!!
which is obvious from the method body!

``` java
    private <T> T rmFirst(List<? extends T> elements, Predicate<T> predicate) {
        T element = null;
        for (final Iterator<? extends T> it = elements.iterator(); it.hasNext() && element == null; ) {
            final T next = it.next();
            if (predicate.apply(next)) {
                element = next;
                it.remove();
            }
        }
        return element;
    }
```

Probably `rmFirstMatching` is better name, next time please check out the code before asking.

BTW iterating through the `List` by using indexes is a bad practice as you don't know what is the instance, so for linked list it will take _N_ for each operation, so overall complexity is `1/2 * N * (1 + N)` which is obviously more than N, while for ArrayList it will be still _N_.

dkulieshov: Hi,

Indeed it is quite clear that in this particular case it is not a good idea to walk through the List using index. 

However the explanation looks to me a bit confusing so I would like to clarify a few things. The main argument here is a worst-case complexity of `get` method for different implementations of `List`, using asymptotic notation for `LinkedList` we have _O(n)_ while for `ArrayList` we have _O(1)_. This gives us total estimations: _O(n*n)_ for `LinkedList` and _O(n)_ for `ArrayList` for our code snippet. At the same time using iterator you will have _O(n)_ for both implementations. 

P.S. Still can't figure out where did you get such formula: _1/2 \* N \* (1 + N)_. Does that correspond to some average-case complexity measurement?

voievodin: Okay let me explain you my point of view:

``` java
for (int i = 0; i < machines.size(); i++) {
    if (machine.getId().equals(machines.get(i).getId())) {
        machines.remove(i);
        return;
    }
}
```

From my comment: 
For ArrayList in the worst case(when the target element is the last one) get method will be invoked N times where N is the size of the list, so as the get from the ArrayList is O(1), then the search complexity is O(N).

For LinkedList  in the worst case(when the target element is the last one) get method will be performed N times as well, but as you mentioned get for LinkedList is O(N), so for the list from 4 elemenets we will have (1 + 2 + 3 + 4) operations, which is arithmetic progression with _1_ as a step, so to get the complexity we use the arithmetic progression formula which is  `N * (A1 + An) / 2`, so in our case it can be converted to the `N(1+N)/2` which is actually exactly that count of operations we perform using linkedlist(in wost case), but yes according to the O notation it is O(n^2).

> P.S. Still can't figure out where did you get such formula: 1/2 \* N \* (1 + N). Does that correspond to some average-case complexity measurement?

I didn't say anything about average case complexity measurement.
There are were 2 examples for ArrayList and for LinkedList iteration.

dkulieshov: No need in such detailed explanation. Your point is almost correct! However next time please use common terminology to avoid misunderstanding: [Big O notation](https://en.wikipedia.org/wiki/Big_O_notation), [Worst-case complexity](https://en.wikipedia.org/wiki/Worst-case_complexity), etc.

voievodin: :ok_hand:

mmorhun:
@garagatyi @evoevodin review please

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -495,6 +502,55 @@ public void createSnapshot(String workspaceId) throws NotFoundException, ServerE
...
+     * @param machineConfig
+     *         configuration of new machine
+     * @param workspaceId
+     *         id of workspace this machine belongs to
+     * @param activeEnv
+     *         current machine environment
+     * @return starting machine instance
+     * Throws the same exceptions as @{@link MachineManager#createMachineAsync(MachineConfig, String, String)}
+     */
+    public MachineImpl createMachineAsyncInRuntime(MachineConfig machineConfig, String workspaceId, String activeEnv)
```
garagatyi: Don't like the name. Consider `createMachineAsync`

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -495,6 +502,55 @@ public void createSnapshot(String workspaceId) throws NotFoundException, ServerE
...
+            throws MachineException,
+                   BadRequestException,
+                   SnapshotException,
+                   NotFoundException,
+                   ConflictException {
+        new AddNewMachineInRuntimeEventSubscriber();
```
garagatyi: This line looks really strange for Java. Name of class looks like name of method. Please refactor

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -495,6 +502,55 @@ public void createSnapshot(String workspaceId) throws NotFoundException, ServerE
...
+     *         when workspace doesn't exist
+     * @throws ConflictException
+     *         when machine is dev-machine
+     * @throws MachineException
+     *         when another error occurs
+     * @throws NullPointerException
+     *         when {@code machineId} is null
+     */
+    public void removeMachineFromRuntime(String machineId) throws NotFoundException, ConflictException, MachineException {
+        requireNonNull(machineId, "Require non-null machine id");
```
garagatyi: In method `createMachineAsyncInRuntime` you don't validate arguments. In this method you validate them. Looks strange

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -495,6 +502,55 @@ public void createSnapshot(String workspaceId) throws NotFoundException, ServerE
...
+        MachineImpl machine = machineManager.getMachine(machineId);
+        if (machine.getConfig().isDev()) {
+            throw new ConflictException("Cannot remove dev-machine " + machine.getId() +
+                                        " from " + machine.getWorkspaceId() + " workspace");
```
garagatyi: Add description how to remove dev machine in the exception message

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -495,6 +502,55 @@ public void createSnapshot(String workspaceId) throws NotFoundException, ServerE
...
+     *         when workspace from which machine is removing doesn't have runtime or
+     *         when workspace doesn't exist
+     * @throws ConflictException
+     *         when machine is dev-machine
+     * @throws MachineException
+     *         when another error occurs
+     * @throws NullPointerException
+     *         when {@code machineId} is null
+     */
+    public void removeMachineFromRuntime(String machineId) throws NotFoundException, ConflictException, MachineException {
```
garagatyi: I think we should make this method package private to prevent incorrect usage by outer components. @evoevodin WDYT?

mmorhun: @garagatyi I disagree. I think, that we should have mechanism to remove machines from runtime.
As about wrong usage - it is responsibility of developer to use it correctly. We can broke something in many classes by invoking something wrong, isn't it?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -708,6 +764,59 @@ private WorkspaceImpl getByKey(String key) throws NotFoundException, ServerExcep
...
+                LOG.error(exception.getLocalizedMessage(), exception);
+                return;
+            }
+            try {
+                runtimes.addMachine(machine);
+            } catch (ServerException | ConflictException | NotFoundException exception) {
+                try {
+                    machineManager.destroy(event.getMachineId(), true);
+                } catch (NotFoundException | MachineException e) {
+                    e.printStackTrace();
```
benoitf: shouldn't use this instruction --> logger

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -245,6 +246,72 @@ public void stop(String workspaceId) throws NotFoundException, ServerException,
...
+     * @param machine
+     *         machine to add to specified runtime
+     * @throws NotFoundException
+     *         when workspace with specified id not running or not exists
+     * @throws ServerException
+     *         when application server is stopping
+     * @throws ConflictException
+     *         when workspace is not running
+     */
+    public void addMachine(MachineImpl machine) throws NotFoundException, ServerException, ConflictException {
```
benoitf: do we need implementation class or only interface (Machine ? )

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -245,6 +246,72 @@ public void stop(String workspaceId) throws NotFoundException, ServerException,
...
+     * Removes machine from running workspace.
+     * This method do not touch workspace configuration and is opposite to {@link #addMachine(MachineImpl)}
+     * Does nothing if remove nonexistent machine
+     *
+     * @param machine
+     *         machine to remove from specified runtime
+     * @throws NotFoundException
+     *         when workspace with specified id not running or not exists
+     */
+    public void removeMachine(MachineImpl machine) throws NotFoundException {
```
benoitf: MachineImpl --> Machine ?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/event/CleanUpNonDevMachineOnStop.java
```diff
@@ -0,0 +1,74 @@
...
+    @Inject
+    public CleanUpNonDevMachineOnStop(WorkspaceManager workspaceManager, EventService eventService) {
+        this.workspaceManager = workspaceManager;
+        this.eventService = eventService;
+    }
+
+    @Override
+    public void onEvent(MachineStatusEvent event) {
+        if (DESTROYING.equals(event.getEventType()) && !event.isDev()) {
+            try {
```
benoitf: for me this try () empty catch block looks odd

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
mmorhun: Does it may sense? Can workspace 'live' without dev-machine?

garagatyi: It can't. But someone tried to to that in that case, so you can explain in the exception message that  user have to stop WS to stop dev machine

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
garagatyi: you added that for out internal component. it is not supposed that this method will be called by some other code

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
garagatyi: sure, we neither use e.printStackTrace() nor System.out

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
garagatyi: This is internal component that works with implementations

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
garagatyi: This is internal component that works with implementations

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/event/CleanUpNonDevMachineOnStop.java
garagatyi: it is not empty, there is an explanation why it is OK to suppress exceptions there

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
mmorhun: Sorry, I missed it

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
benoitf: I never saw that in architecture design

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/event/CleanUpNonDevMachineOnStop.java
benoitf: I have problem of code design.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
voievodin: Implementation class is the right choice here.
Explanation:
Lets start from the [Machine](https://github.com/eclipse/che/blob/master/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/Machine.java) interface.

``` java
public interface Machine {
   // ...
   MachineConfig getConfig();
   // ...
}
```

This interface defines `MachineConfig`(which is also model interface) as configuration type.
So we have several outcomes:
-  Dto can't use non-dto classes, which means that MachineDto implementation must return `MachineConfigDto` instead, which means that all the classes used in MachineDto are dtos. So we override the method and define 
  [MachineConfigDto](https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-machine-shared/src/main/java/org/eclipse/che/api/machine/shared/dto/MachineDto.java#L28) as return type of the `config()` method
- `MachineImpl` follows the same strategy and uses Impl objects, instead of model objects in composition.
  - It allows to use setters(not defined by model), which is really important
  - There is no any reason to define the composition of the object from the same family using interfaces, because
    it is not profitable at all, it complicates process of changing values and doesn't give the benefits from the abstraction, 
    and it is supposed that objects family is used as a one integrated piece.

But why don't we use the interfaces for such classes like managers or daos?
Explanation:
1. We don't have data object factories or abstract builders, which means that we create objects directly by using `new`. In this case managers and daos are automatically dependent from the `*Impl` classes.
2. We need to modify(use setters) data returned by daos and managers, so in the case of model interfaces we need to recreate object over and over again adding new modifications to it, but what is the profit?

So we decided to use _Impl_ objects throughout the same workspaces infrastructure, but use model objects in dependent parts e.g. `Account`. It allows us to avoid allocations and clears relationship between workspace internals. 

For DTOs to Impl objects transformation we define copy constructors in `Impl` classes based on model interfaces, which allows to switch between implementations easily.

In this case Model object is rather for integrity and consistency of client and server, than for something else.

benoitf: My issue is about using classes for a contract and not interface. (whiteboard pattern)
Machine is like 'user' view interface and in your case MachineImpl the inner model. But usually I see contracts with interface. Like Machine and any name like "ModelMachine", MachineImpl implementing this interface.

so you can use setters, contract is clear, mocking is easy. you can plug any implementation "even replace internal stuff"

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/event/StopWorkspaceOnDestroyDevMachine.java
```diff
@@ -30,28 +30,27 @@
 import static org.eclipse.che.api.machine.shared.dto.event.MachineStatusEvent.EventType.DESTROYED;
 
 /**
- * The class listens changing of machine status and perform some actions when status is changed.
+ * The class listens changing of machine status and perform stop workspace if dev-machine is destroyed.
```
benoitf: is it dev-machine or ws-machine ?

mmorhun: @benoitf yes, now it is ws-machine, tnx

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -495,6 +502,61 @@ public void createSnapshot(String workspaceId) throws NotFoundException, ServerE
...
+     * If machine creation fails, then that machine will be destroyed.
+     *
+     * @param machineConfig
+     *         configuration of new machine
+     * @param workspaceId
+     *         id of workspace this machine belongs to
+     * @param activeEnv
+     *         current machine environment
+     * @return starting machine instance
+     * Throws the same exceptions as @{@link MachineManager#createMachineAsync(MachineConfig, String, String)}
```
benoitf: i'm not sure this kind of Throws sentence is compliant with javadoc

mmorhun:
@evoevodin @garagatyi check please

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -72,6 +78,7 @@
  * @author gazarenkov
  * @author Alexander Garagatyi
  * @author Yevhenii Voevodin
+ * @author Mykola Morhun
```
garagatyi: Remove added lines please

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -279,6 +293,49 @@ public boolean hasRuntime(String workspaceId) {
         }
     }
 
+    @Override
+    public void onEvent(MachineStatusEvent event) {
+        if (hasRuntime(event.getWorkspaceId())) {
```
garagatyi: HasRuntime is expensive, use it as a last resort only

wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/workspace/server/WorkspaceManagerTest.java
```diff
@@ -80,7 +80,7 @@
 @Listeners(value = {MockitoTestNGListener.class})
 public class WorkspaceManagerTest {
 
-    private static final String USER_ID = "user123";
+    private static final String USER_ID      = "user123";
```
garagatyi: Why you changed this

voievodin:
Can you please create another pull request for the new solution(when it is ready), 
because it is not about _Stop additional machines when stopping workspace_ anymore, at least it is just a half of the problem + provide good description.
I am seek of these jenkins notification after each fixup you make + there are to many comments and -1 things.

mmorhun:
#1367

PR #1223, opened 2016-05-11
fix workspace name uniqueness validation

akurinnoy:


benoitf:
if we're testing against workspaceId I would change the name of the directive then (it's id and not anymore name)

akurinnoy:
The directive should test name of workspace to be unique so I would let it's name. Instead I've changed directive's attribute value. It's needed for current workspace not to compare to itself.

benoitf:
what about merging it ?

benoitf:
also I don't see JIRA issue in commit message

ashumilova:
Jira issue is https://jira.codenvycorp.com/browse/CHE-1163, is not completed and @akurinnoy is ill at the moment

PR #1224, opened 2016-05-11
CHE-416: Make Git branches window higher and narrower

vinokurig:
**_1 Upvote**_ branchlist

@vparfonov please review


vparfonov:
+1

PR #1231, opened 2016-05-11
Simplification changes for running Che in a Docker container

TylerJewell:
Per the specification of CHE-1230.  Make default storage location in /storage.

@riuvshin please approve
@eivantsov we will need to update docs and verify che.sh


garagatyi:
LGTM
Please squash all your commits

PR #1236, opened 2016-05-12
CHE-787; Make it possible to change log level when starting Che

mshaposhnik:
**_1 Upvote**_

### DIFF
--- a/assembly/assembly-main/src/assembly/bin/che.sh
+++ b/assembly/assembly-main/src/assembly/bin/che.sh
@@ -35,7 +35,7 @@ Usage:
      -m:name,   --machine:name          For Win & Mac, sets the docker-machine VM name; default=che
      -a:driver, --machine-driver:driver For Win & Mac, specifies the docker-machine driver to use; default=vbox
      -p:port,   --port:port             Port that Che server will use for HTTP requests; default=8080
-     -ll:level, --logs_level:level      Logging level in Che; default= INFO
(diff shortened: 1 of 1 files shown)


mshaposhnik:
@vparfonov @eivantsov @skabashnyuk

vparfonov:
+1

TylerJewell:
1. Do not use two letter short code options - only use a single letter.
2. Rebase against master - the che.sh was heavily rewritten and merged yesterday.
3. What are the full list of log levels that are possible.  Are these Tomcat log levels or che log levels.  If they are tomcat log levels, should they be exposed in che.sh script?

mshaposhnik:
@TylerJewell 
1. Which letter do you propose?
2. Seems changes fits ok
3. It's a logback log levels.

TylerJewell:
1. Just use "-l" if it is available.
2. Make it clear in the log description that it's logback log levels.  Otherwise average person will not know, I think.

mshaposhnik:
-l used for copy-lib. We can change in to -c, and use -l to logs then
-l|--copy-lib
      COPY_LIB=true
    ;;

TylerJewell:
If you make this change, also change the Dockerfile in the root of master.  The ENTRYPOINT passes in the -l option and needs to be updated to use the correct option.

mshaposhnik:
https://github.com/eclipse/che/pull/1242/files

PR #1237, opened 2016-05-12
Remove a trailing slash from the base URL to the ws agent REST services

vparfonov:
@garagatyi

### DIFF
--- a/core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/DevMachine.java
+++ b/core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/DevMachine.java
@@ -82,12 +82,16 @@ public class DevMachine {
 
     /**
      *
-     * @return return base URL to the ws agent REST services. URL will be always without trailing slash 
+     * @return return base URL to the ws agent REST services. URL will be always without trailing slash
      */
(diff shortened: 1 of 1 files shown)


core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/DevMachine.java
```diff
@@ -80,10 +80,14 @@ public String getWsAgentWebSocketUrl() {
...
 
+    /**
+     *
+     * @return return base URL to the ws agent REST services. URL will be always without trailing slash 
+     */
     public String getWsAgentBaseUrl() {
         DevMachineServer server = getServer(Constants.WSAGENT_REFERENCE);
         if (server != null) {
-            return server.getUrl();
+            return server.getUrl().replaceAll("/$", ""); //here we remove trailing slash if exist
```
garagatyi: I think that you can improve speed and clarity of this operation with

``` java
String url = server.getUrl();
if (url.endsWith("/")) {
    url = url.substring(0, url.length() - 1);
}
return url;
```

vparfonov: ok

PR #1238, opened 2016-05-12
[WIP] CHE-1144: add fixes of multithreaded code in DockerConnector

garagatyi:
@skabashnyuk @evoevodin @sleshchenko Can you take a look?

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
@@ -34,8 +34,8 @@ import org.eclipse.che.plugin.docker.client.json.ContainerCommitted;
 import org.eclipse.che.plugin.docker.client.json.ContainerConfig;
 import org.eclipse.che.plugin.docker.client.json.ContainerCreated;
 import org.eclipse.che.plugin.docker.client.json.ContainerExitStatus;
-import org.eclipse.che.plugin.docker.client.json.ContainerListEntry;
 import org.eclipse.che.plugin.docker.client.json.ContainerInfo;
+import org.eclipse.che.plugin.docker.client.json.ContainerListEntry;
 import org.eclipse.che.plugin.docker.client.json.ContainerProcesses;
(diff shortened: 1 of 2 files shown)


plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -1147,41 +1147,22 @@ private String buildImage(final BuildImageParams params,
...
-                    public void run() {
-                        try {
-                            ProgressStatus progressStatus;
-                            while ((progressStatus = progressReader.next()) != null) {
-                                final String buildImageId = getBuildImageId(progressStatus);
-                                if (buildImageId != null) {
-                                    imageIdHolder.set(buildImageId);
-                                }
-                                progressMonitor.updateProgress(progressStatus);
+                executeInSeparateThread(() -> {
```
voievodin, line 1151: I think that this name is not really good, as it doesn't say anything about waiting.
My suggestions `runAsyncAndWait`, `executeAndWait`

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -1720,4 +1656,51 @@ private void addQueryParamIfNotNull(DockerConnection connection, String queryPar
...
+
+    /**
+     * This method runs provided runnable in a separate thread in a such a way that that thread can be interrupted.
+     * It is supposed that it runs long-running task and at some point it can be stopped.
+     * <br/>Provided task should throw unchecked exception if exceptional situation happens.
+     * In that case {@link IOException#IOException(String, Throwable)} will be thrown
+     * with {@link Throwable#getLocalizedMessage()} and unchecked exception as arguments.
+     */
+    private void executeInSeparateThread(Runnable task) throws InterruptedException, IOException {
+        final ValueHolder<Throwable> errorHolder = new ValueHolder<>();
```
skabashnyuk, line 1676: What about to use 
https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/Future.html
instead of this method?

PR #1247, opened 2016-05-13
CODENVY-502 allow to create factories by providing custom parameters

benoitf:
Change-Id: I0c9554ac6d767e110f9bc39c30e2fb2712ad413f

### DIFF
--- a/wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryParametersResolver.java
+++ b/wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryParametersResolver.java
@@ -1,0 +1,46 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.api.factory.server;
+
+import org.eclipse.che.api.core.BadRequestException;
+import org.eclipse.che.api.factory.shared.dto.Factory;
+
+import javax.validation.constraints.NotNull;
+import java.util.Map;
+
+/**
+ * Defines a resolver that will produce factories for some parameters
+ *
+ * @author Florent Benoit
+ */
+public interface FactoryParametersResolver {
+
+    /**
+     * Resolver acceptance based on the given parameters.
+     *
+     * @param factoryParameters
+     *         map of parameters dedicated to factories
+     * @return true if it will be accepted by the resolver implementation or false if it is not accepted
+     */
+    boolean accept(@NotNull Map<String, String> factoryParameters);
+
+    /**
+     * Create factory object based on provided parameters
+     *
+     * @param factoryParameters
+     *         map containing factory data parameters provided through URL
+     * @throws BadRequestException
+     *         when data are invalid
+     */
+    Factory createFactory(@NotNull Map<String, String> factoryParameters) throws BadRequestException;
+
+
+}
--- a/wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryService.java
+++ b/wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryService.java
@@ -63,9 +63,12 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.UnsupportedEncodingException;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.function.Predicate;
 
@@ -82,12 +85,33 @@ import static org.eclipse.che.dto.server.DtoFactory.newDto;
  * Defines Factory REST API.
  *
  * @author Anton Korneta
+ * @author Florent Benoit
  */
 @Api(value = "/factory",
      description = "Factory manager")
 @Path("/factory")
 public class FactoryService extends Service {
     private static final Logger LOG = LoggerFactory.getLogger(FactoryService.class);
+
+    /**
+     * Error message if there is no plugged resolver.
+     */
+    public static final String ERROR_NO_RESOLVER_AVAILABLE = "Cannot build factory with any of the provided parameters.";
+
+    /**
+     * If there is no parameter.
+     */
+    public static final String ERROR_NO_PARAMETERS = "Missing parameters";
+
+    /**
+     * Validate query parameter. If true, factory will be validated
+     */
+    public static final String VALIDATE_QUERY_PARAMETER = "validate";
+
+    /**
+     * Set of resolvers for factories. Injected through an holder.
+     */
+    private final Set<FactoryParametersResolver> factoryParametersResolvers;
 
     private final FactoryStore           factoryStore;
     private final FactoryEditValidator   factoryEditValidator;
@@ -106,6 +130,7 @@ public class FactoryService extends Service {
                           LinksHelper linksHelper,
                           FactoryBuilder factoryBuilder,
                           WorkspaceManager workspaceManager,
+                          FactoryParametersResolverHolder factoryParametersResolverHolder,
                           UserDao userDao) {
         this.factoryStore = factoryStore;
         this.createValidator = createValidator;
@@ -114,6 +139,7 @@ public class FactoryService extends Service {
         this.linksHelper = linksHelper;
         this.factoryBuilder = factoryBuilder;
         this.workspaceManager = workspaceManager;
+        this.factoryParametersResolvers = factoryParametersResolverHolder.getFactoryParametersResolvers();
         this.userDao = userDao;
     }
 
@@ -569,6 +595,74 @@ public class FactoryService extends Service {
                        .build();
     }
 
+
+    /**
+     * Resolve parameters and build a factory for the given parameters
+     *
+     * @param parameters
+     *         map of key/values used to build factory.
+     * @param uriInfo
+     *         url context
+     * @return a factory instance if found a matching resolver
+     * @throws NotFoundException
+     *         when no resolver can be used
+     * @throws ServerException
+     *         when any server errors occurs
+     * @throws BadRequestException
+     *         when the factory is invalid e.g. is expired
+     */
+    @POST
+    @Path("/resolver")
+    @Consumes(APPLICATION_JSON)
+    @Produces(APPLICATION_JSON)
+    @ApiOperation(value = "Create factory by providing map of parameters",
+                  notes = "Get JSON with factory information.")
+    @ApiResponses({@ApiResponse(code = 200, message = "OK"),
+                   @ApiResponse(code = 400, message = "Failed to validate factory"),
+                   @ApiResponse(code = 500, message = "Internal server error")})
+    public Factory resolveFactory(
+            @ApiParam(value = "Parameters provided to create factories")
+            final Map<String, String> parameters,
+            @ApiParam(value = "Whether or not to validate values like it is done when accepting a Factory",
+                      allowableValues = "true,false",
+                      defaultValue = "false")
+            @DefaultValue("false")
+            @QueryParam(VALIDATE_QUERY_PARAMETER)
+            final Boolean validate,
+            @Context
+            final UriInfo uriInfo) throws NotFoundException, ServerException, BadRequestException {
+
+        // Check parameter
+        if (parameters == null) {
+            throw new BadRequestException(ERROR_NO_PARAMETERS);
+        }
+
+        // search matching resolver
+        Optional<FactoryParametersResolver> factoryParametersResolverOptional = this.factoryParametersResolvers.stream().filter((resolver -> resolver.accept(parameters))).findFirst();
+
+        // no match
+        if (!factoryParametersResolverOptional.isPresent()) {
+            throw new NotFoundException(ERROR_NO_RESOLVER_AVAILABLE);
+        }
+
+        // create factory from matching resolver
+        final Factory factory = factoryParametersResolverOptional.get().createFactory(parameters);
+
+        // Apply links
+        try {
+            factory.setLinks(linksHelper.createLinks(factory, uriInfo, null));
+        } catch (UnsupportedEncodingException e) {
+            throw new ServerException(e.getLocalizedMessage(), e);
+        }
+
+        // time to validate the factory
+        if (validate) {
+            acceptValidator.validateOnAccept(factory);
+        }
+
+        return factory;
+    }
+
     /**
      * Creates factory links.
      *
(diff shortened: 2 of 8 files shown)


benoitf:
@skabashnyuk @vparfonov

benoitf:
FYI, in the JIRA issue there is a video

skabashnyuk:
-1 For the Factory Service. 
I don't like idea to add to generic Factory service github specific. 
May be it's part of Github Service or GithubFactoryService.

benoitf:
@skabashnyuk it can't be in github service as github is only deployed on workspace agent

skabashnyuk:
ok GithubFactoryService.
Also FactoryServiceClient should not contains Github sprcific

benoitf:
GithubFactoryService with /factory/github mapping is ok ?

mshaposhnik:
Dont forget that github is a plugin. Some custom builds theoretically may not support it.

UPD: we just allowed to create factories with any source type (and gihub too). See https://github.com/eclipse/che/pull/1248

skabashnyuk:
Not sure yet. BTW why you need that rest method? Why don't you create factory on  client ?

benoitf:
@skabashnyuk because logic is server side

skabashnyuk:
Can you describe a bit more desired parameters and desired behaviour of your rest method? What configuration of factory will be?

benoitf:
@mshaposhnik it's related to pure git, not github. You don't need github plugin in the agent.

benoitf:
@skabashnyuk parameter is listed in PR. It's to create a factory based on a github url

garagatyi:
-1 to include github specific features in common API

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryService.java
```diff
@@ -639,4 +654,143 @@ private void processDefaults(Factory factory) {
...
+        // decode value
+        final String decodedGithubUrl;
+        try {
+             decodedGithubUrl = URLDecoder.decode(githubUrl, "UTF-8");
+        } catch (UnsupportedEncodingException e) {
+            throw new BadRequestException(String.format("Unable to decode value %s", githubUrl));
+        }
+
+        // use regexp to find repository details (repository name, project name and branch and subfolder)
+        final Pattern GITHUB_PATTERN = Pattern.compile("^(?:http)(?:s)?(?:\\:\\/\\/)github.com/(.*?)/(.*?)(?:/tree/(.*?)(?:/(.*?))?)?$");
```
garagatyi: Compilation of pattern can be improved by making it static

skabashnyuk:
BTW it should be POST

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryService.java
benoitf: BTW it will be moved to another class

benoitf:
@skabashnyuk I don't see why it should be more POST

skabashnyuk:
I thought it create new factory, NO?

garagatyi:
I don't think that we should allow such API for factory. This task should be done by client, not API. It can be done by Java client or Js, but it definitely should not be in the API

benoitf:
@skabashnyuk it is indempotent

skabashnyuk:
-1 for this service. 
The whole use case what you are trying to solve can help to understand the way better way of implementation.

benoitf:
I have reworked to allow "pluggable" factory resolver. It's now based on plugins
to sum-up : no github introduced directly but allow to plug github, etc through extension mechanisms

skabashnyuk:
I don't get it. The whole picture how it will work is unclear for me.

benoitf:
@skabashnyuk 
the whole picture is quite easy and described in the specification document
the idea is to have factory model based without creating/storing them. Like the previous "non-encoded factories".
Factories data will be stored inside the repository, not in codenvy database.

skabashnyuk:
> the whole picture is quite easy and described in the specification document

Sorry. Not for me.
It's looks like your vision of technical solution for some sort of user-story.
But initial use case is not described. The missing part is the problem what we are trying to solve.
BTW it  may looks like we may need to resurect "non-encoded " factory.

benoitf:
I can explain everything you want.
The referenced specification document is written by Tyler
Usecase is the title of this pull request: create factories "on demand" based on a github repository for example if I have a link to a github repository.
This PR enable extensibility which is not available for now by allowing to plug external resolvers and then enhance factories through plugins. So we could extend this for other stuff.

skabashnyuk:
I think solution should be more general like non-encoded factory.

TylerJewell:
We do not have non encoded factory.  We have this resolver.  The comments should focus on the code readiness.  What is left to be improved?

benoitf:
it's a general solution. And it allows to plug server side logic which is required

skabashnyuk:
> I can explain everything you want.

`
Objective is to make it simple to add a factory URL to open source project.  
`
Need more details of user-story (How user will use it,  etc.) 
Are you sure you've added all code you want?

benoitf:
> Need more details of user-story (How user will use it, etc.) 

For example, user will add on its github README a link like http://beta.codenvy.com/f/?github=<encoded url of github repository>

Then by clicking on this link it will clone this github repository and use as runtime the .codenvy.dockerfile provided in the repository if there is one, else it will stick to a default image.

The idea is to be able to plug other resolvers. So github is one but we could have many others. We could transform "named factories" as well as factory resolver. It's just another way to provide factories.

> Are you sure you've added all code you want?

About the code, for che repository, as idea was to externalize it to be plugin-based, for now that's all for che repository. (There is few change on this repository. It's linked to the fact that while factory API code is in che, it's not used at all in che and only in codenvy) The remaining stuff is part of codenvy PR

skabashnyuk:
As I understand you would like to have, factory url.
Where user can provide source url and somehow docker file (it can be a link to docker file that stored in repository   as query parameter). 
For example 
`
http://beta.codenvy.com/f/?source.location=https://github.com/skabashnyuk/castest.git&dockerfile=https://dockerfiles.codenvycorp.com/templates-4.0/factory/factory-dockerfile
`

benoitf:
idea is more that data are "inside" the repository, so we provide a link to the repository. (non-encoded factory was a way to give all data with the url (but there was a limit on url length), encoded factory we give only the id, here it's that we give a key to a repository and it's the repository that has the data)

about source.location it's obviously a too long parameter and is not fulfilling the goal. Because if repository contains a json file, it may include other sources

If you don't like ?github= it could be replaced by ?url= (then depending of the URL we can have/plug different url resolver, one for github, one for some other repositories, etc)

skabashnyuk:
> If you don't like ?github= it could be replaced by ?url= 

yes. I don't like github. url - more or less ok.

> idea is more that data are "inside" the repository

What are the data format and how it gets there?

benoitf:
> If you don't like ?github= it could be replaced by ?url=
> ok.

OK I will rework PR

> idea is more that data are "inside" the repository
> What are the data format and how it gets there?

It's step #2.
It should be a .codenvy.factory.json (so keep same format that the current one) but that can be incomplete. In the spec document for example it's said that repository URL can be computed automatically without being defined inside the file. So using like an "override mode"
I'm now only focusing on step #1 which is about specifying a custom dockerfile: (which can cover many projects to start)

skabashnyuk:
Ok. So you would like to have ability to accept factory

> http://beta.codenvy.com/f/?url=sourcodeurl.

And repository must contains .codenvy.factory.json where source code repository is replaced with "sourcodeurl"  during factory acceptance. 
Is that all? Anything else? 

I think in general this concept can introduce more cons when pros in future.
For instance user anyway need to know somehow  .codenvy.factory.json format, need to modify his repository, we lose ability to update format with just simple update of mongo database. 

Are you going to introduce support of old .codenvy.factory.json format in case if something will be changed?

benoitf:
this is for the next step, for now json is not required.
about factory content, there is a version field so this version allow to define the version of the format.
If format is changed, it implies a new version of factory format. If not then it's a mistake to have the same version with different attributes/model.

Today user can have a factory file on his computer and then he may want to create or re-create factories by providing the content of this file as well. So "simple update of mongo database" is a hack and do not work because you can't update files on user computer as well. While handling "version" inside the factory allow to do it.

skabashnyuk:
> this is for the next step, for now json is not required.

Ok so for now this two https://github.com/codenvy/codenvy/pull/195
https://github.com/eclipse/che/pull/1247  PR only needed to accept factory in format 

> http://beta.codenvy.com/f/?url=sourcodeurl.

Is that ok if project type will be blank at the end?

benoitf:
yes but I need to rework PR for now to use "url"

skabashnyuk:
And this functionality has no github specific?

benoitf:
it has specific part for github that will be provided by a plugin.
(like extracting branch, subfolder, etc from a github url)

?url=https://github.com/ --> handle by github plugin (if enabled else it will provide error like today)
?url=http://foo.com --> error as no plugin is handling (but may work if someone write a plugin for this)

skabashnyuk:
I would like to understand the scope of this two PR

https://github.com/codenvy/codenvy/pull/195
https://github.com/eclipse/che/pull/1247

Acceptance of 
http://beta.codenvy.com/f/?url=sourcodeurl
Without of Github specific?

A couple of thougts.
Puting codenvy.factory.json in repository and using it may intruduce some complexity.
Like we need to have it before workspace creating and can't use cloning, for instance
some vcs providers may not introduce web interface. Other question authentication on 
vcs hosting. Are we going to provide this fucntionality for public projects?

What if we provide 3 types of factory?
 Current. Most advanced
1. http://beta.codenvy.com/f/?id=i-123123

 Realy realy simple. Jut to play with because with project type blank it's almoust useles. 
2. http://beta.codenvy.com/f/?source=sourcodeurl

  Allow to users specify external factory json with source authoreplacemet. Almoust identical to 1.
3. http://beta.codenvy.com/f/?factory.json=linktojson

In 2-3 cases during acceptance we will crate persistant factory and continue acceptance as usual.
I think we can't miss this step. AFAIK some part of ide may reuse factory id during workspace liftime

benoitf:
The goal is that we don't need persistent factory IN codenvy for 2 and 3. (if 100 users click on a url factory link it would create 100 dummy factories for nothing as persistent state is on the repository)
IDE is receiving factory data without id (this is already done today in my PR)

skabashnyuk:
ok. So 2-3 are just sending to dashboard and dashboard do the business. Correct?

benoitf:
no business is done server side

skabashnyuk:
Not clear for me why. 
AFAIK we are accepting factory in dashboard - no?

benoitf:
FactoryServlet checks if factory is valid before redirecting to dashboard so factory is checked first server side.
then FactoryServlet redirect user to dashboard which create/start workspace which then delegates to IDE the projects acceptance and IDE acceptance
(dashboard and IDE don't do any logic about factory parameters, they just execute the data from the factory object returned from the server that they share)

skabashnyuk:
So we just missing the part on dashboard how to get from 2-3 format to real Factory object ?

benoitf:
it asks server side

skabashnyuk:
from concept point of view
So we just missing the part on dashboard how to get from 2-3 format to real Factory object ?
Correct?

benoitf:
there is no missing part as it's implemented in my PR

skabashnyuk:
Not yet. I'm not get the whole picture. And you propose only github specific service.
Let's talk about why we need such kind of service and its obligation and parameters.

gazarenkov:
As I stated in email thread I need the big picture for factories, so we need a specs after that we discuss and consider this PR.

benoitf:
Pull request rewritten
please review

in addition, diagram to understand details
image

core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/factory/ParametersFactoryServiceClient.java
```diff
@@ -0,0 +1,38 @@
...
+     * Get factory object based on user parameters
+     *
+     * @param factoryParameters
+     *         map containing factory data parameters provided through URL
+     * @param validate
+     *         indicates whether or not factory should be validated by accept validator
+     * @param callback
+     *         callback which return valid JSON object of factory or exception if occurred
+     */
+    void getFactory(@NotNull Map<String, String> factoryParameters, boolean validate, @NotNull AsyncRequestCallback<Factory> callback);
```
garagatyi: Have you considered usage of promises here? AFAIK it is default codestyle for IDE now

benoitf: I used the same pattern than FactoryServiceClient.

skabashnyuk: For new code we prefer to use promises.

azatsarynnyy: +1

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryParametersResolver.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Defines a resolver that will produce factories for some parameters
+ * @author Florent Benoit
+ */
+public interface FactoryParametersResolver {
+
+    /**
+     * Resolver acceptance based on the given parameters.
+     * @param factoryParameters map of parameters dedicated to factories
```
azatsarynnyy: not formatted code

core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/factory/ParametersFactoryServiceClient.java
benoitf: maybe you prefer but this rule is not written anywhere
so, it can't be mandatory

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryParametersResolver.java
benoitf: ok

core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/factory/ParametersFactoryServiceClient.java
benoitf: @azatsarynnyy @garagatyi could you tell me in which class this pattern is used

garagatyi: https://github.com/eclipse/che/blob/826353e6ab575c8aca0e215d656409fb4fd43bda/core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/importer/AbstractImporter.java#L57-L57

azatsarynnyy: [MachineServiceClient](https://github.com/eclipse/che/blob/master/core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/MachineServiceClient.java)
[MachineServiceClientImpl](https://github.com/eclipse/che/blob/master/core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/MachineServiceClientImpl.java)
for example

benoitf: ok thanks

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryParametersResolver.java
benoitf: done

core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/factory/ParametersFactoryServiceClient.java
benoitf: done

benoitf: I also upgrade the other getFactory method to promise

benoitf:
so we're good now ?

garagatyi:
LGTM

skabashnyuk:
ок

PR #1248, opened 2016-05-13
CODENVY-301; do not check source storage types on creating factory

mshaposhnik:
In general, at the moment of factory creation or accepting, we can't know all source types we support, because they are pluggable and deployed on agent and also can be subject to change.


mshaposhnik:
@vparfonov  @skabashnyuk

vparfonov:
ok

skabashnyuk:
ok

PR #1251, opened 2016-05-13
CHE-1105: Avoid NPE in case current project null.

vparfonov:
Unregister Create Maven Module action because this feature outdated, need to rework it in future


skabashnyuk:
ok

PR #1268, opened 2016-05-16
CHE-1081: remove workspace id from services that deployed on wsagent

vparfonov:
@skabashnyuk @sleshchenko @gazarenkov


benoitf:
this is a breaking change for dashboard and I don't see any UD file modified in this PR

core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/project/ProjectServiceClient.java
```diff
@@ -33,8 +33,8 @@
     /**
      * Get all projects in current workspace.
      *
-     * @param workspaceId
-     *         id of current workspace
+     * @param devMachine
+     *         of current devMachine
```
sleshchenko, line 37: mb just something like `instance of current dev machine`? Or maybe we can use not current dev machine here

sleshchenko:
It would easier to review to have separated issues for removing id from services' path and removing classes from project and maven plugins.
P.S. In issue and pull request description nothing about changes in project and maven plugins

vparfonov:
@benoitf i talk with @ashumilova about that we will add  changes later

vparfonov:
@skabashnyuk @garagatyi

wsagent/wsagent-local/src/main/java/org/eclipse/che/WorkspaceIdProvider.java
```diff
@@ -8,21 +8,20 @@
...
+package org.eclipse.che;
 
-import com.google.gwt.resources.client.ClientBundle;
-
-import org.eclipse.che.ide.ui.Styles;
 
 /**
- * @author Evgen Vidolob
+ * //
+ *
```
skabashnyuk, line 18: javadoc

skabashnyuk:
other ok

garagatyi:
LGTM

gazarenkov:
OK for me

PR #1269, opened 2016-05-16
CHE-770: Update Orion editor to eleventh version

azatsarynnyy:
@vparfonov


plugins/plugin-orion/che-plugin-orion-editor/src/main/java/org/eclipse/che/ide/editor/orion/client/inject/OrionCodeEditWidgetProvider.java
```diff
@@ -0,0 +1,44 @@
+/*******************************************************************************
+ * Copyright (c) 2014-2015 Codenvy, S.A.
```
skabashnyuk: Is this file excluded from license check?

azatsarynnyy: This was due to this property in pom.xml
`<license_years>2014-2015</license_years>`
I've fixed it
Thx

vparfonov:
ok

PR #1270, opened 2016-05-16
fix usage of wsagent websocket url in case of trailing slash

garagatyi:
@skabashnyuk @vparfonov Please take a look


vparfonov:
ok

skabashnyuk:
ok

PR #1273, opened 2016-05-17
CHE-1155: use correct wsagent websocket link

ashumilova:

### DIFF
--- a/dashboard/src/components/api/che-workspace.factory.js
+++ b/dashboard/src/components/api/che-workspace.factory.js
@@ -334,26 +334,17 @@ export class CheWorkspace {
    */
   getWebsocketUrl(workspaceId) {
     let workspace = this.workspacesById.get(workspaceId);
-    if (!workspace || !workspace.runtime) {
+    if (!workspace || !workspace.runtime || !workspace.runtime.links) {
       return '';
     }
-    let runtimeData = workspace.runtime;
 
-    // extract the Websocket URL of the runtime
-    let servers = runtimeData.devMachine.runtime.servers;
-
-    var wsagentServerAddress;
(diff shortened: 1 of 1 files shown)


dashboard/src/components/api/che-workspace.factory.js
```diff
@@ -334,26 +334,17 @@ export class CheWorkspace {
...
-
-    var wsagentServerAddress;
-    for (var key in servers) {
-      let server = servers[key];
-      if ('wsagent' === server.ref) {
-        wsagentServerAddress = server.address;
+    for (let i = 0; i < workspace.runtime.links.length; i++) {
+      let link = workspace.runtime.links[i];
+      if (link.rel === 'wsagent.websocket') {
+        return link.href + '/ws/' + workspaceId;
```
garagatyi: Please take into account that link.href can contain trailing slash

dashboard/src/components/api/che-workspace.factory.js
```diff
@@ -334,26 +334,18 @@ export class CheWorkspace {
...
-    var wsagentServerAddress;
-    for (var key in servers) {
-      let server = servers[key];
-      if ('wsagent' === server.ref) {
-        wsagentServerAddress = server.address;
+    for (let i = 0; i < workspace.runtime.links.length; i++) {
+      let link = workspace.runtime.links[i];
+      if (link.rel === 'wsagent.websocket') {
+        let href = link.href.endsWith('/') ? link.href : link.href + '/';
+        return href + 'ws/' + workspaceId;
```
garagatyi, line 345: I think that client must not know the path to the websocket endpoint. Please move that into link on serverside

PR #1276, opened 2016-05-17
Replace internal everrest cache with guava cache

skabashnyuk:
https://github.com/codenvy/everrest/pull/23 is going to remove HelperCache. So this change reuse guava cache instead of everrest.


garagatyi:
LGTM

vparfonov:
ok

PR #1279, opened 2016-05-17
CHE-1160: Update Vagrantfile

TylerJewell:
Moved the Vagrantfile to now launch the Che Docker container instead of Che as a server.  Simpler syntax, more reliable, and operates as a Daemon by default "--restart=always" syntax.  Also adds in better checks for Vagrant plugin configuration.  Maps Che workspaces and projects into the Vagrant image sync'd folder.  Also now lets the user set the port that the image is listening on.

Updated 4.3 documentation:  https://eclipse-che.readme.io/v4.3/docs/usage-vagrant
@eivantsov @riuvshin - please approve.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -10,14 +10,30 @@
 # Set to "<proto>://<user>:<pass>@<host>:<port>"
 $http_proxy  = ""
 $https_proxy = ""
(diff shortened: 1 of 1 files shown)


riuvshin:
LGTM

PR #1281, opened 2016-05-18
Add parameters for auto-snapshot and auto-restore of the workspace state

akorneta:
This feature is planned for the 4.3
@evoevodin, @skabashnyuk, @garagatyi


voievodin:
Good, LGTM

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -143,3 +143,10 @@ machine.ssh.connection_timeout_ms=3000
 # Suffix helps differentiate archive for different architectures/OSes
 machine.server.terminal.path_to_archive.linux_amd64=${che.home}/lib/linux_amd64/terminal
 machine.server.terminal.path_to_archive.linux_arm7=${che.home}/lib/linux_arm7/terminal
+
+# During the stop of the workspace automatically creates a snapshot if the value is {true},
+# otherwise just stops the workspace.
+workspace.runtime.auto_snapshot=false
```
benoitf, line 149: is true shouldn't be the default value for Eclipse Che ?

akorneta, line 149: probably not because in che we do not have default registry

skabashnyuk:
ok

PR #1283, opened 2016-05-18
CHE-1120 disable binding maven project type if M2_HOME env property n…

evidolob:
…ot set, code clean up


plugins/plugin-maven/che-plugin-maven-server/src/main/java/org/eclipse/che/plugin/maven/server/inject/MavenModule.java
```diff
@@ -39,7 +39,10 @@
     protected void configure() {
         newSetBinder(binder(), ValueProviderFactory.class).addBinding().to(MavenValueProviderFactory.class);
 
-        newSetBinder(binder(), ProjectTypeDef.class).addBinding().to(MavenProjectType.class);
+        //bind maven project type only if maven installed on dev machine
+        if(System.getenv("M2_HOME") != null) {
```
azatsarynnyy, line 43: AFAIK usually we inject named variable for getting environment variable's value
`@Named("env.M2_HOME") String m2Home`

vparfonov:
ok

PR #1284, opened 2016-05-18
[WIP] CODENVY-439 Make update file an asynchronous operation

mkuznyetsov:
Updating file might take a long time if Lucene searcher has not been initialized yet.
We need to make it asynchronous, so it will not hold the IDE waiting it to complete.

### DIFF
--- a/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
+++ b/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
@@ -16,6 +16,8 @@ import io.swagger.annotations.ApiParam;
 import io.swagger.annotations.ApiResponse;
 import io.swagger.annotations.ApiResponses;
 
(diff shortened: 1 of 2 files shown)


benoitf:
missing referencing JIRA issue in commit message or pull request name

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
```diff
@@ -474,13 +492,21 @@ public Response updateFile(@ApiParam(value = "Full path to a file", required = t
             throw new NotFoundException("File not found for " + path);
         }
 
-        file.updateContent(content);
+        executor.execute(() -> {
+            try {
+                file.updateContent(content);
+            } catch (ForbiddenException e) {
+                LOG.warn(e.getLocalizedMessage());
+            } catch (ServerException e) {
```
benoitf: multicatch ? 
catch (ForbiddenException | ServerException e)

PR #1295, opened 2016-05-19
CODENVY-439 Initialize Lucene searcher asynchronously

mkuznyetsov:
**_1 Upvote**_ Lucene searcher initializes on first file updating. If the project is big, it may take a long time while the IDE is being blocked, waiting for response.
Whe should make this process asynchronous

### DIFF
--- a/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/search/impl/AbstractLuceneSearcherProvider.java
+++ b/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/search/impl/AbstractLuceneSearcherProvider.java
@@ -10,6 +10,8 @@
  *******************************************************************************/
 package org.eclipse.che.api.vfs.search.impl;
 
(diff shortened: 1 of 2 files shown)


skabashnyuk:
ok  @evoevodin @garagatyi ?

benoitf:
it will be a great "first experience" improvement for factories

slemeur:
+1

garagatyi:
LGTM

voievodin:
Well my thoughts: 
I don't think that changing behaviour from the synchronous to the asynchronous in such a low level is really good solution. What if someone relies on the synchronous initialization of the searcher ? BUT on the other hand [doc](https://github.com/eclipse/che/blob/a3a864ee7e84109670b3db9a9a5137d649e68efd/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/search/SearcherProvider.java#L22-L32) doesn't say anything about the nature of the initialization, so i am not really into this field and can't say anything. So it is okay for me as long as it is okay for @vparfonov and @skabashnyuk.

PR #1301, opened 2016-05-19
CODENVY-538:Fix checking of recipe location: now will match only host…

vparfonov:
… wihout scheme.

Add method for download file in case server send redirect(in some reason setInstanceFollowRedirects(true) don't work so need to do one more request)

@skabashnyuk @garagatyi @riuvshin

### DIFF
--- a/core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/IoUtil.java
+++ b/core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/IoUtil.java
@@ -13,12 +13,25 @@ package org.eclipse.che.commons.lang;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.*;
+import javax.ws.rs.HttpMethod;
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.FilenameFilter;
+import java.io.IOException;
+import java.io.InputStream;
 import java.net.HttpURLConnection;
 import java.net.URL;
 import java.net.URLConnection;
 import java.nio.channels.FileChannel;
-import java.nio.file.*;
+import java.nio.file.FileSystems;
+import java.nio.file.FileVisitOption;
+import java.nio.file.FileVisitResult;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.PathMatcher;
+import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.security.DigestInputStream;
 import java.security.MessageDigest;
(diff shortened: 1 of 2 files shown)


wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/util/RecipeDownloader.java
```diff
@@ -58,24 +59,27 @@ public RecipeDownloader(@Named("api.endpoint") String apiEndpoint) {
...
         URL recipeUrl;
         File file = null;
+        final String location = machineConfig.getSource().getLocation();
         try {
-            UriBuilder targetUriBuilder = UriBuilder.fromUri(machineConfig.getSource().getLocation());
+            UriBuilder targetUriBuilder = UriBuilder.fromUri(location);
             // add user token to be able to download user's private recipe
-            if (machineConfig.getSource().getLocation().startsWith(apiEndpoint)) {
+            final String apiEndPointHost = apiEndpoint.getHost();
+            final String host = UriBuilder.fromUri(location).build().getHost();
```
garagatyi: Why not use targetUriBuilder instead of creating new uribuilder

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/IoUtil.java
```diff
@@ -231,6 +242,62 @@ public static File downloadFile(File parent, String prefix, String suffix, URL u
...
+     *         prefix of temporary file name, may not be <code>null</code> and must be at least three characters long
+     * @param suffix
+     *         suffix of temporary file name, may be <code>null</code>
+     * @param url
+     *         URL for download
+     * @return downloaded file
+     * @throws java.io.IOException
+     *         if any i/o error occurs
+     */
+    public static File downloadFileWithRedirect(File parent, String prefix, String suffix, URL url) throws IOException {
```
garagatyi, line 263: Have you tried to allow redirects in the connection instead of maling this custom logic?

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/util/RecipeDownloader.java
riuvshin: yeah, we tried but it does not work, that is why @vparfonov added custom method with redirect.

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/IoUtil.java
riuvshin, line 263: yeah, we tried but it does not work, that is why @vparfonov added custom method with redirect.
As @vparfonov mentioned above "setInstanceFollowRedirects(true) don't work"

vparfonov:
http://stackoverflow.com/questions/1884230/urlconnection-doesnt-follow-redirect

skabashnyuk:
ОК

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/util/RecipeDownloader.java
vparfonov: @garagatyi  check my fix pls

garagatyi:
LGTM

PR #1305, opened 2016-05-20
CHE-962: Add ability to store and use SSH keys for Subversion hosts

tolusha:


plugins/plugin-svn/che-plugin-svn-ext-server/src/main/java/org/eclipse/che/plugin/svn/server/utils/SvnUrl.java
```diff
@@ -0,0 +1,29 @@
...
+    private SvnUrl() { }
+
+    /**
+     * Indicates if uri represents ssh connection.
+     *
+     * @param uri
+     *      the url to svn repository
+     */
+    public static boolean isSSH(String uri) {
+        return uri != null && uri.startsWith("svn+ssh://");
```
sleshchenko: You use this method in this class with one method once :) Let's remove this utils class WDYT?

tolusha: sure.

vparfonov:
ok

PR #1306, opened 2016-05-20
CHE-956 Move che-jdt-ext-machine to java-ext-lang-server module 

evidolob:
And change REST paths from jdt to java


evidolob:
@vparfonov

vparfonov:
ok

PR #1310, opened 2016-05-20
CODENVY-427: add temporary workaround to log illegal response from swarm

garagatyi:
@skabashnyuk @riuvshin WDYT?


skabashnyuk:
ок

PR #1313, opened 2016-05-20
CHE-1139 Renaming in Create Project wizard in IDE

vitaliy-guliy:


vparfonov:
ok

PR #1315, opened 2016-05-20
refactor GitAskPassScript directory generation

gazarenkov:
@skabashnyuk


wsagent/che-core-git-impl-native/src/main/java/org/eclipse/che/git/impl/nativegit/GitAskPassScript.java
```diff
@@ -71,9 +71,7 @@ public File build(UserCredential credentials) throws GitException {
 
     public void remove() {
 
-        File askScriptDirectory = new File(System.getProperty("java.io.tmpdir")
-                                           + "/" + EnvironmentContext.getCurrent().getSubject().getUserName());
-        if (askScriptDirectory.exists()) {
+        if (askScriptDirectory != null && askScriptDirectory.exists()) {
```
skabashnyuk: askScriptDirectory looks like can't be null. Shall we omit this statement.

skabashnyuk:
ок

PR #1319, opened 2016-05-20
CHE-1060: configure 'java.source.folder' attribute from the project wizard

svor:
@vparfonov take a look please. 
[Yo can see the result here! ](https://youtu.be/hHDK9usXoAM)

### DIFF
--- a/plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-ide/src/main/java/org/eclipse/che/plugin/java/plain/client/wizard/PlainJavaPageViewImpl.java
+++ b/plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-ide/src/main/java/org/eclipse/che/plugin/java/plain/client/wizard/PlainJavaPageViewImpl.java
@@ -1,0 +1,108 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.plugin.java.plain.client.wizard;
+
+import com.google.gwt.core.client.GWT;
+import com.google.gwt.event.dom.client.ClickEvent;
+import com.google.gwt.event.dom.client.ClickHandler;
+import com.google.gwt.event.dom.client.KeyUpEvent;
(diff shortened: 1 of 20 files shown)


TylerJewell:
1. Project type should be labled "Java" not "Plain Java Project" or "Plain Java".
2. We should not have any default package for projects created from this wizard.  Empty project, not "com.company".

benoitf:
what if I have multiple source folders in my project that I import ?

like
api/src
client/src

slemeur:
@TylerJewell : it has been handled into a separate PR: https://github.com/eclipse/che/pull/1313

plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-ide/src/main/java/org/eclipse/che/plugin/java/plain/client/wizard/PlainJavaPageViewImpl.java
```diff
@@ -0,0 +1,108 @@
...
+    public void changeBrowseBtnVisibleState(boolean isVisible) {
+        browseButton.setVisible(isVisible);
+    }
+
+    @Override
+    public void changeSourceFolderFieldState(boolean isEnable) {
+        sourceFolderField.setEnabled(isEnable);
+    }
+
+    interface MavenPageViewImplUiBinder extends UiBinder<FlowPanel, PlainJavaPageViewImpl> {
```
azatsarynnyy: PlainJavaPageViewImplUiBinder

benoitf:
So can I work with
https://github.com/santanche/java2learn 
only one src folder but there is a lib folder to configure as well when we import project

these should work 
https://github.com/moleksyuk/Java-Design-Patterns
https://github.com/winterbe/java8-tutorial
(only one src folder)

here are two src folders
https://github.com/Nirman-Rathod/Java

vparfonov:
@slemeur in case several source folder we need to your UI mock up

benoitf:
I may have to specify as well all .jar files to get as libraries as well (or specify lib folder which will include all .jars from this directory) (multiple jars and/or multiple lib folder)

gazarenkov:
As for the goal of the issue LGTM

vparfonov:
ok

benoitf:
it's pretty useless without being able to specify multi folder and without being able to pickup java libraries as well even.
While issue is only targeting one folder I think epic https://jira.codenvycorp.com/browse/CHE-915 could not be closed (add remaining part as new issue) until libraries and multiple source folders are handled with the wizard

svor:
@benoitf I've created two issues for the configuring [library](https://jira.codenvycorp.com/browse/CHE-1207) and [source folders](https://jira.codenvycorp.com/browse/CHE-1207)

benoitf:
@svor that's perfect

PR #1320, opened 2016-05-20
CHE-648. Try to restore websocket connection when this one has closed

RomanNikitenko:
1. The first problem is - we have double method call and code execution for initialization of WsAgentStateController: 
   - when a workspace has started: https://github.com/eclipse/che/blob/master/core/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/client/BootstrapController.java#L105
   - when a dev machine has started: https://github.com/eclipse/che/blob/master/plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/machine/MachineManagerImpl.java#L298
     This causes incorrect behavior to tracking of websocket connection state.
2. Next problem - Che loses connection with WS agent after the several hours of inactivity. My suggestion is try to restore connection when connection has closed or when we have some errors related to websocket connection. So we will have 5 attempts to restore connection (reconnection period is 2ms) and notify corresponding listeners only after these attempts. For example, I've tested it on the use case when session timeout has been reached: connection was restored in 2 ms without some problems.


vparfonov:
ok

PR #1325, opened 2016-05-20
refactor GitAskPassScript directory generation

gazarenkov:
@skabashnyuk


riuvshin:
@gazarenkov your initial PR was reverted by https://github.com/eclipse/che/pull/1322 so you need to restore it otherwise your fix is incompatible with current master state

PR #1327, opened 2016-05-20
Git ask pass script

gazarenkov:
Removing user sensitive information from folder names. Since it can contains some special characters.


benoitf:
there is no description or issue in this pull request

benoitf:
thanks

PR #1329, opened 2016-05-22
Update che.sh

TylerJewell:
**_1 Upvote**_ CHE-1202 - Add in checks in case 'docker ps' fails. There are two additional checks.  First, verify that group and owner have rw permissions for /var/run/docker.sock.  Second, verify that the docker client version and the docker server version match.  In situations where docker client is in a Che container and that container is using a host's docker daemon, it is possible for the client and server to not match.

@garagatyi, @eivantsov what do you think?

I don't have a way to generate good tests for this.  I did the best that I could in a VM, but all the VMs I create generally don't have either of these issues.  Do we have good environments to test for this improvement?

### DIFF
--- a/assembly/assembly-main/src/assembly/bin/che.sh
+++ b/assembly/assembly-main/src/assembly/bin/che.sh
@@ -410,6 +410,11 @@ get_docker_ready () {
                    `Consider updating docker engine to have the versions match."
         return
       fi
+      
+      error_exit "Running 'docker' succeeded, but 'docker ps' failed. \n`
+                 `/var/run/docker.sock is ok and your docker client and server have matching versions. \n`
+                 `Run 'docker ps' and inspect the output for additional clues."
+      return
     fi
   fi
 
--- a/assembly/assembly-main/src/assembly/bin/che.sh
+++ b/assembly/assembly-main/src/assembly/bin/che.sh
@@ -379,32 +379,32 @@ get_docker_ready () {
       # CHE-1202: Improve error messages in case of docker ps failure
       # Verify that /var/run/docker.sock has owner and group read / write permissions
       PERMS=$(stat -c %A /var/run/docker.sock)
-      OWNERREAD=$(cut -c2 <(echo $PERMS))
-      OWNERWRITE=$(cut -c3 <(echo $PERMS))
-      GROUPREAD=$(cut -c5 <(echo $PERMS))
-      GROUPWRITE=$(cut -c6 <(echo $PERMS))
+      OWNER_READ=$(cut -c2 <(echo $PERMS))
+      OWNER_WRITE=$(cut -c3 <(echo $PERMS))
+      GROUP_READ=$(cut -c5 <(echo $PERMS))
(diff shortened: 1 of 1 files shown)


assembly/assembly-main/src/assembly/bin/che.sh
```diff
@@ -375,11 +375,45 @@ get_docker_ready () {
...
     else
-      # If Linux and docker ps fails, then this likely a file permissions issue.
-      error_exit "Running 'docker' succeeded, but 'docker ps' failed. `
-                 `This usually means that docker cannot reach its daemon. `
-                 `On Mac and Linux, check the read / write permissions on '/var/run/docker.sock'. `
-                 `Consider running 'sudo chmod 777 /var/run/docker.sock'."
+
+      # CHE-1202: Improve error messages in case of docker ps failure
+      # Verify that /var/run/docker.sock has owner and group read / write permissions
+      PERMS=$(stat -c %A /var/run/docker.sock)
```
garagatyi, line 381: 1. Even if docker.sock has owner/group rw permission it doesn't mean that current user is owner or included in needed group.
2. docker.sock can have rw permission for others.
3. `sudo chmod 660 /var/run/docker.sock` will help only when user has needed id or gid

assembly/assembly-main/src/assembly/bin/che.sh
```diff
@@ -375,11 +375,45 @@ get_docker_ready () {
...
+      OWNERREAD=$(cut -c2 <(echo $PERMS))
+      OWNERWRITE=$(cut -c3 <(echo $PERMS))
```
garagatyi: Hard to read, use camelcase or underscore

assembly/assembly-main/src/assembly/bin/che.sh
```diff
@@ -375,11 +375,45 @@ get_docker_ready () {
...
+                   `OWNER WRITE: ${OWNERWRITE} \n`
+                   `GROUP READ:  ${GROUPREAD} \n`
+                   `GROUP WRITE: ${GROUPWRITE} \n`
+                   `Run 'sudo chmod 660 /var/run/docker.sock' to give the right permissions."
+        return
+      fi
+
+      # CHE-1202: Improve error messages in case of docker ps failure
+      # Verify that docker client and server versions match
+      DOCKERSERVERVERSION=$(docker version --format '{{.Server.Version}}')
```
garagatyi: Hard to read, use camelcase or underscore

assembly/assembly-main/src/assembly/bin/che.sh
TylerJewell, line 381: We perform the uid check elsewhere in the script.  It is checked before getting to this step.

TylerJewell:
@garagatyi Added formatting.

garagatyi:
LGTM

garagatyi:
Please use squash when you will be ready to merge

TylerJewell:
@eivantsov Added check for other group.

ghost:
+1

PR #1331, opened 2016-05-23
Restore CORS support on WS master

vparfonov:
@skabashnyuk @benoitf @ashumilova


skabashnyuk:
ok

PR #1333, opened 2016-05-23
CHE-797: correct switch to mobile view

akurinnoy:
CHE-797: correct switch to mobile view for
- list of workspaces
- list of projects
- panels on workspace details page
- che-input
- che-label
- che-text-info


benoitf:
need to be reviewed by @slemeur as it's purely UI stuff

ashumilova:
The workspace list and inputs look much more better.
But now have problem with projects list alignment (its name is centered in all cases now) and hover state of the line.
project_list_align

PR #1336, opened 2016-05-23
CODENVY-480 Remove user's roles

sleshchenko:
@skabashnyuk @vparfonov


skabashnyuk:
ok

benoitf:
I suppose PR should get rid of flag "WiP" before being considered as "ok" ?

sleshchenko:
@benoitf Of course. Thanks for remark

PR #1343, opened 2016-05-24
CHE-1143: fixed synchronization issue during removal of several projects

dkulieshov:
**_1 Upvote**_ Hi,

Recently I was working on some bugs in Eclipse Che project management infrastructure. The report itself looks like:

When we delete several projects simultaneously (in the same running workspace with project api):
1. We get `204` status on each request however projects are not deleted
2. In some cases we get `500` status combined with `java.util.ConcurrentModificationException`

The first point is a separate question and there are several ideological problems that should be solved first. As far as I know, the discussion is not yet complete.

And for the second item I created a dedicated test to cover the bug. Though the bug is not reproduced anymore most likely thanks to [this commit](https://github.com/eclipse/che/commit/658410556f8d0d5c91316890111ff7dc9d57cba1#diff-9414b806e2941cc99a2d071cd5b982fc), I've spotted another problem. We rarely get exception:  

```
java.lang.AssertionError: Error: Comparison method violates its general contract! expected [204] but found [500]
        at org.testng.Assert.fail(Assert.java:94)
        at org.testng.Assert.failNotEquals(Assert.java:494)
        at org.testng.Assert.assertEquals(Assert.java:123)
        at org.testng.Assert.assertEquals(Assert.java:370)
        at org.eclipse.che.api.project.server.ProjectServiceTest.testDeleteProjectsConcurrently(ProjectServiceTest.java:981)
```

when we try to simultaneously remove 100 projects (in 100 threads), while it is always reproduced for 1k+ threads.

I assume that the problem is the following. At some point of project deletion we are to get its filesystem parent's sorted children list, the process is a set of two operations: get and sort the list. These operations are not synchronized so sometimes it happens to change somehow state of children collections by another thread. This results in a following exception: 

```
java.lang.IllegalArgumentException: Comparison method violates its general contract!
    at java.util.ComparableTimSort.mergeHi(ComparableTimSort.java:835)
    at java.util.ComparableTimSort.mergeAt(ComparableTimSort.java:453)
    at java.util.ComparableTimSort.mergeForceCollapse(ComparableTimSort.java:392)
    at java.util.ComparableTimSort.sort(ComparableTimSort.java:191)
    at java.util.ComparableTimSort.sort(ComparableTimSort.java:146)
    at java.util.Arrays.sort(Arrays.java:472)
    at java.util.Collections.sort(Collections.java:155)
```

That basically happens when arrays of children that are to be merged are indeed not sorted (because of impact of the other thread: there are changes of some elements' state or elements themselves after they are already sorted) or elements are sorted but children's compareTo method cannot fulfill its contract (e.g. does not provide transitivity).

My suggestion is to synchronize those two operations, which should not have great impact on performance (taking into account the kind of operations that we have here) and at the same time we will not impact general logic of such a low level component as file system.

### DIFF
--- a/wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
+++ b/wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
@@ -92,11 +92,17 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Scanner;
 import java.util.Set;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
 
@@ -219,7 +225,7 @@ public class ProjectServiceTest {
         FileWatcherNotificationHandler fileWatcherNotificationHandler = new DefaultFileWatcherNotificationHandler(vfsProvider);
         FileTreeWatcher fileTreeWatcher = new FileTreeWatcher(root, new HashSet<>(), fileWatcherNotificationHandler);
 
-        pm = new ProjectManager(vfsProvider, null, ptRegistry, projectRegistry, phRegistry,
+        pm = new ProjectManager(vfsProvider, new EventService(), ptRegistry, projectRegistry, phRegistry,
                                 importerRegistry, fileWatcherNotificationHandler, fileTreeWatcher, workspaceHolder);
         pm.initWatcher();
 
@@ -227,30 +233,14 @@ public class ProjectServiceTest {
 
         //List<ProjectConfigDto> modules = new ArrayList<>();
 
-        final ProjectConfigDto testProjectConfigMock = mock(ProjectConfigDto.class);
-        when(testProjectConfigMock.getPath()).thenReturn("/my_project");
-        when(testProjectConfigMock.getName()).thenReturn("my_project");
-        when(testProjectConfigMock.getDescription()).thenReturn("my test project");
-        when(testProjectConfigMock.getType()).thenReturn("my_project_type");
-//        when(testProjectConfigMock.getModules()).thenReturn(modules);
-        when(testProjectConfigMock.getSource()).thenReturn(DtoFactory.getInstance().createDto(SourceStorageDto.class));
-//        when(testProjectConfigMock.findModule(anyString())).thenReturn(testProjectConfigMock);
-
-        Map<String, List<String>> attr = new HashMap<>();
-        for (Attribute attribute : myProjectType.getAttributes()) {
-            attr.put(attribute.getName(), attribute.getValue().getList());
-        }
-        when(testProjectConfigMock.getAttributes()).thenReturn(attr);
-
         projects = new ArrayList<>();
-        projects.add(testProjectConfigMock);
+        addMockedProjectConfigDto(myProjectType, "my_project");
+
         when(httpJsonRequestFactory.fromLink(any())).thenReturn(httpJsonRequest);
         when(httpJsonRequest.request()).thenReturn(httpJsonResponse);
         when(httpJsonResponse.asDto(WorkspaceDto.class)).thenReturn(usersWorkspaceMock);
         when(usersWorkspaceMock.getConfig()).thenReturn(workspaceConfigMock);
         when(workspaceConfigMock.getProjects()).thenReturn(projects);
-
-        pm.createProject(testProjectConfigMock, null);
 
 //        verify(httpJsonRequestFactory).fromLink(eq(DtoFactory.newDto(Link.class)
 //                                                             .withHref(apiEndpoint + "/workspace/" + workspace + "/project")
@@ -288,6 +278,27 @@ public class ProjectServiceTest {
         env = org.eclipse.che.commons.env.EnvironmentContext.getCurrent();
     }
 
+    private void addMockedProjectConfigDto(org.eclipse.che.api.project.server.type.ProjectTypeDef myProjectType, String projectName)
+            throws ForbiddenException, ServerException, NotFoundException, ConflictException {
+        final ProjectConfigDto testProjectConfigMock = mock(ProjectConfigDto.class);
+        when(testProjectConfigMock.getPath()).thenReturn("/" + projectName);
+        when(testProjectConfigMock.getName()).thenReturn(projectName);
+        when(testProjectConfigMock.getDescription()).thenReturn("my test project");
+        when(testProjectConfigMock.getType()).thenReturn("my_project_type");
+        when(testProjectConfigMock.getSource()).thenReturn(DtoFactory.getInstance().createDto(SourceStorageDto.class));
+        //        when(testProjectConfigMock.getModules()).thenReturn(modules);
+        //        when(testProjectConfigMock.findModule(anyString())).thenReturn(testProjectConfigMock);
+
+        Map<String, List<String>> attr = new HashMap<>();
+        for (Attribute attribute : myProjectType.getAttributes()) {
+            attr.put(attribute.getName(), attribute.getValue().getList());
+        }
+        when(testProjectConfigMock.getAttributes()).thenReturn(attr);
+
+        projects.add(testProjectConfigMock);
+
+        pm.createProject(testProjectConfigMock, null);
+    }
 
     @Test
     @SuppressWarnings("unchecked")
@@ -925,6 +936,50 @@ public class ProjectServiceTest {
         assertEquals(response.getStatus(), 204, "Error: " + response.getEntity());
 
         pm.getProject("my_project");
+    }
+
+    @Test
+    public void testDeleteProjectsConcurrently() throws Exception {
+        int threadNumber = 100;
+        ExecutorService executor = Executors.newFixedThreadPool(threadNumber);
+        CountDownLatch countDownLatch = new CountDownLatch(threadNumber);
+        List<Future<ContainerResponse>> futures = new LinkedList<>();
+
+
+        for (int i = 0; i < threadNumber; i++) {
+            addMockedProjectConfigDto(ptRegistry.getProjectType("my_project_type"), "my_project_name" + i);
+        }
+
+        IntStream.range(0, threadNumber).forEach(
+                i -> {
+                    futures.add(executor.submit(() -> {
+                        countDownLatch.countDown();
+                        countDownLatch.await();
+
+                        try {
+                            return launcher.service(DELETE,
+                                                    "http://localhost:8080/api/project/my_project_name" + i,
+                                                    "http://localhost:8080/api", null, null, null);
+                        } catch (Exception e) {
+                            throw new IllegalStateException(e);
+                        }
+                    }));
+                }
+                                                );
+
+        boolean isNotDone;
+        do {
+            isNotDone = false;
+            for (Future<ContainerResponse> future : futures) {
+                if (!future.isDone()) {
+                    isNotDone = true;
+                }
+            }
+        } while (isNotDone);
+
+        for (Future<ContainerResponse> future : futures) {
+            assertEquals(future.get().getStatus(), 204, "Error: " + future.get().getEntity());
+        }
     }
 
     @Test
--- a/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/impl/file/LocalVirtualFileSystem.java
+++ b/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/impl/file/LocalVirtualFileSystem.java
@@ -243,7 +243,7 @@ public class LocalVirtualFileSystem implements VirtualFileSystem {
         return newArrayList(path.elements()).contains(".vfs");
     }
 
-    List<VirtualFile> getChildren(LocalVirtualFile parent, VirtualFileFilter filter) throws ServerException {
+    synchronized List<VirtualFile> getChildren(LocalVirtualFile parent, VirtualFileFilter filter) throws ServerException {
         if (parent.isFolder()) {
             final List<VirtualFile> children = doGetChildren(parent, DOT_VFS_DIR_FILTER, filter);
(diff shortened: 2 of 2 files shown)


TylerJewell:
I like the approach. Welcome back @dkuleshov.  Are you a contributor to the eclipse foundation?  Have you signed the cla?   The signed off check failed.

gazarenkov:
+1
As I understand and that's what we discussed with @evoevodin the big problem is that operations have not been syched.

And yes great approach to explain

dkulieshov:
@TylerJewell It's nice to be back 😄. I'm not a contributor yet, just a member. The CLA is signed but there was a git misconfiguration on my side. Looks like now the signed off check is okay.

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
```diff
@@ -928,6 +939,50 @@ public void testDeleteProject() throws Exception {
...
+
+
+        for (int i = 0; i < threadNumber; i++) {
+            addMockedProjectConfigDto(ptRegistry.getProjectType("my_project_type"), "my_project_name" + i);
+        }
+
+        IntStream.range(0, threadNumber).forEach(
+                i -> {
+                    futures.add(executor.submit(() -> {
+                        countDownLatch.countDown();
```
voievodin, line 956: hmm, it seems that you don't have to `countDown()` each time, you can create a `CountDownLatch` with 1 as count and perform `.countDown()` once after `forEach`.

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
```diff
@@ -928,6 +939,50 @@ public void testDeleteProject() throws Exception {
     }
 
     @Test
+    public void testDeleteProjectsConcurrently() throws Exception {
+        int threadNumber = 100;
+        ExecutorService executor = Executors.newFixedThreadPool(threadNumber);
+        CountDownLatch countDownLatch = new CountDownLatch(threadNumber);
+        List<Future<ContainerResponse>> futures = new LinkedList<>();
```
voievodin, line 946: As you know the size of the list i'd say that it is better to choose `ArrayList` instead of `LinkedList`.

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
```diff
@@ -928,6 +939,50 @@ public void testDeleteProject() throws Exception {
     }
 
     @Test
+    public void testDeleteProjectsConcurrently() throws Exception {
+        int threadNumber = 100;
+        ExecutorService executor = Executors.newFixedThreadPool(threadNumber);
```
voievodin, line 944: Are you sure about non-daemon executor which is never shut down?

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
```diff
@@ -928,6 +939,50 @@ public void testDeleteProject() throws Exception {
...
+
+        boolean isNotDone;
+        do {
+            isNotDone = false;
+            for (Future<ContainerResponse> future : futures) {
+                if (!future.isDone()) {
+                    isNotDone = true;
+                }
+            }
+        } while (isNotDone);
```
voievodin, line 978: I think you can do it clearer and more gracefully by using another `CountDownLatch(numThreads)` and awaiting with timeout.

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
dkulieshov, line 956: I'm not sure that I got your idea, could you explain more?

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
dkulieshov, line 946: I see no significant benefit of `ArrayList` over `LinkedList` here, do you?

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
```diff
@@ -928,6 +939,50 @@ public void testDeleteProject() throws Exception {
...
+        ExecutorService executor = Executors.newFixedThreadPool(threadNumber);
+        CountDownLatch countDownLatch = new CountDownLatch(threadNumber);
+        List<Future<ContainerResponse>> futures = new LinkedList<>();
+
+
+        for (int i = 0; i < threadNumber; i++) {
+            addMockedProjectConfigDto(ptRegistry.getProjectType("my_project_type"), "my_project_name" + i);
+        }
+
+        IntStream.range(0, threadNumber).forEach(
```
benoitf, line 953: is there a reason why here you use forEach and just before you used a for (int i =0; loop)

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
dkulieshov, line 978: I have deep concerns that using 

> another `CountDownLatch(numThreads)` and awaiting with timeout

 is clearer than a simple loop. Unless you can prove it?

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
voievodin, line 956: The idea: await 100 times, count down once.

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
garagatyi: I think count down latch was designed exactly for such use cases. So +1 that it will be much clearer

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
```diff
@@ -288,6 +278,27 @@ public void setUp() throws Exception {
...
 
+    private void addMockedProjectConfigDto(org.eclipse.che.api.project.server.type.ProjectTypeDef myProjectType, String projectName)
+            throws ForbiddenException, ServerException, NotFoundException, ConflictException {
+        final ProjectConfigDto testProjectConfigMock = mock(ProjectConfigDto.class);
+        when(testProjectConfigMock.getPath()).thenReturn("/" + projectName);
+        when(testProjectConfigMock.getName()).thenReturn(projectName);
+        when(testProjectConfigMock.getDescription()).thenReturn("my test project");
+        when(testProjectConfigMock.getType()).thenReturn("my_project_type");
+        when(testProjectConfigMock.getSource()).thenReturn(DtoFactory.getInstance().createDto(SourceStorageDto.class));
+        //        when(testProjectConfigMock.getModules()).thenReturn(modules);
```
garagatyi, line 289: Why have you added commented code?

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
garagatyi, line 946: Array list is almost always faster

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
garagatyi, line 953: +1 foreach looks odd here.

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
voievodin, line 946: well it is not a significant one but yes i think i do, it is all about the heap and objects location in it.

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
garagatyi, line 956: Please consider usage of barrier here instead of countdown latch

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
dkulieshov, line 944: Well you're right, though in this very case it should not necessarily be shut down except for the reason of reclamation of its resources, but I guess I will add a call to make the code more clear.

garagatyi, line 944: You can make them daemons

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
dkulieshov, line 946: Still don't see the point. In this particular case we are adding and iterating. For these operations `LinkedList` and `ArrayList` have the similar complexity, noting that `ArrayList` has constant size of course. 

Not sure I can get what you mean saying 

>  it is all about the heap and objects location in it
> unless you provide reliable arguments.

Generally speaking if we talk about optimization I find it quite unusual that you did not mention using java arrays for example. Looks like you are arguing just for a dispute.

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
dkulieshov, line 953: Yes, the reason that stands by using of `forEach` (and stream api in general) here is that it is not quite possible to access non final variables from lambda context:
`addMockedProjectConfigDto(ptRegistry.getProjectType("my_project_type"), "my_project_name" + i);`

This situation would happen if I used simple loop.

Do you see how we can make it less ugly?

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
benoitf, line 946: unless you clearly want to use some features of a LinkedList, ArrayList do the work in a less complex way
so basically, you should use ArrayList and as you know the size of the elements you can set in in the constructor

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
dkulieshov, line 289: I just extracted the sources into a separate method and those comments already were there. Unfortunately there is no information about the author of that ugly code so I'm still not sure if that was left for some reason or by a mistake.

As I'm quite a new person to this very project I feel some responsibility for the legacy sources, so I think that it is yet not of my competence to remove someones comments.

Anyway if you feel confident and recommend to remove it - that should not be a problem at all ;-)

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
benoitf, line 953: my concern was only to unify the two loops into a single loop. (because we iterate on the same numbers but in a different way)
and yes to avoid hacks like int j = i; you've to use IntStream

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
dkulieshov, line 956: > One major difference is that CyclicBarrier takes an (optional) Runnable task which is run once the common barrier condition is met.
> 
> It also allows you to get the number of clients waiting at the barrier and the number required to trigger the barrier. Once triggered the barrier is reset and can be used again.
> 
> For simple use cases - services starting etc... a CountdownLatch is fine. A CyclicBarrier is useful for more complex co-ordination tasks. An example of such a thing would be parallel computation - where multiple subtasks are involved in the computation - kind of like MapReduce.

http://stackoverflow.com/questions/4168772/java-concurrency-countdown-latch-vs-cyclic-barrier

There is no need to run `Runnable`
There are no cycles
So please could you make me understand what is the point of using [CyclicBarrier](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CyclicBarrier.html) instead?

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
dkulieshov, line 946: Okay, obviously for me it is not a big deal to replace a `LinkedList` with an `ArrayList`. But if you indeed distinguish velocity of ArrayList and LinkedList and optimization critically matters here then tell me why do you insist on using an `ArrayList` while java array will be even more efficient?

I guess if I know the size of an array I definitely should use simple java array (taking into account that I need only set element and get iterator).

I need to know how do you see the difference between well-optimized and over-optimized code, how do you see where optimization matters more than convenience. Where can I find any rules or conventions or you just impose your preferences? 😆

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
dkulieshov, line 944: Sure

wsagent/che-core-api-project/src/test/java/org/eclipse/che/api/project/server/ProjectServiceTest.java
dkulieshov, line 953: Exactly! As we can see in [this article](https://jaxenter.com/java-performance-tutorial-how-fast-are-the-java-8-streams-118830.html) in general cases java streams are a bit slower than `ArrayList`, so I would obviously use simple java loop here because it would also improve readability but the limitation is that it is not possible to refer to a non final variable.

PR #1345, opened 2016-05-24
CHE-1151: update project tree after changing classpath

svor:
Signed-off-by: Valeriy Svydenko vsvydenko@codenvy.com
- Update project after changing classpath's entries;
- Remove package name from the Java sample and from Java command template.

@vparfonov take a look please


vparfonov:
ok

PR #1344, opened 2016-05-24
CODENVY-457: fix NPE when swarm doesn't detect container death

garagatyi:
@skabashnyuk Please review

PR #1346, opened 2016-05-24
CODENVY-457: fix NPE when swarm doesn't detect container death

garagatyi:
This PR is port of #1344 to 4.3.x
@skabashnyuk @benoitf Please take a look


skabashnyuk:
ok

PR #1351, opened 2016-05-25
CHE-398: Add ability to create local machine snapshots without registry

akorneta:
@garagatyi, @skabashnyuk, @evoevodin


plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
```diff
@@ -183,40 +190,58 @@ public InstanceProcess createProcess(Command command, String outputChannel) thro
...
+        } catch (IOException ioEx) {
+            throw new MachineException(ioEx);
+        } catch (InterruptedException ex) {
             Thread.currentThread().interrupt();
-            throw new MachineException(e.getLocalizedMessage(), e);
+            throw new MachineException(ex.getLocalizedMessage(), ex);
         }
     }
 
+    @VisibleForTesting
```
benoitf, line 226: why VisibleForTesting and not only protected access instead of package access ?
also there is no test calling this method so I don't see why it should be "VisibleForTesting"

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
```diff
@@ -183,40 +190,58 @@ public InstanceProcess createProcess(Command command, String outputChannel) thro
...
+    String saveInRegistry(String owner, String repository, String tag) throws MachineException,
+                                                                                IOException,
+                                                                                InterruptedException {
+        final String repositoryName = registry + '/' + repository;
+        createImage(owner, repositoryName, tag);
+        //TODO fix this workaround. Docker image is not visible after commit when using swarm
```
benoitf: then there is // FIXME: if it requires a fix

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -93,6 +93,7 @@
     private final Set<String>                      commonMachineEnvVariables;
     private final String[]                         allMachinesExtraHosts;
     private final String                           projectFolderPath;
+    private final boolean                          locallySnapshot;
```
benoitf: do we need adjective or noun like localSnapshot
other booleans of this constructor are "doForce", "privilegeMode"

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/DockerInstanceTest.java
```diff
@@ -0,0 +1,239 @@
...
+    @Mock
+    private DockerConnector            dockerConnectorMock;
+    @Mock
+    private DockerInstanceStopDetector dockerInstanceStopDetectorMock;
+
+    private DockerInstance dockerInstance;
+
+    @BeforeMethod
+    public void setUp() throws IOException {
+        dockerInstance = getDockerInstance();
```
benoitf, line 86: we can't use InjectMocks on DockerInstance ? and mock as well Machine ?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -269,6 +270,8 @@ public Instance createInstance(InstanceKey instanceKey,
                               creationLogsOutput);
     }
 
+
```
benoitf: looks like new extra space is added

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -110,9 +111,8 @@ public DockerInstanceProvider(DockerConnector docker,
                                   @Named("machine.docker.pull_image") boolean doForcePullOnBuild,
                                   @Named("machine.docker.privilege_mode") boolean privilegeMode,
                                   @Named("machine.docker.dev_machine.machine_env") Set<String> devMachineEnvVariables,
-                                  @Named("machine.docker.machine_env") Set<String> allMachinesEnvVariables)
-            throws IOException {
-
+                                  @Named("machine.docker.machine_env") Set<String> allMachinesEnvVariables,
+                                  @Named("machine.save.snapshot.locally") boolean locallySnapshot) throws IOException {
```
benoitf: name of property is not containing docker. is it expected ?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
```diff
@@ -183,40 +190,58 @@ public InstanceProcess createProcess(Command command, String outputChannel) thro
...
-            return new DockerInstanceKey(repository, "latest", registry, digest);
-        } catch (IOException e) {
-            throw new MachineException(e);
-        } catch (InterruptedException e) {
+        } catch (IOException ioEx) {
+            throw new MachineException(ioEx);
+        } catch (InterruptedException ex) {
             Thread.currentThread().interrupt();
-            throw new MachineException(e.getLocalizedMessage(), e);
+            throw new MachineException(ex.getLocalizedMessage(), ex);
```
benoitf: why localizedMessage for this one but not for the previous one. Multicatch ?

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/DockerInstanceTest.java
akorneta, line 86: I prefer to explicitly construct objects in tests, do you have an point on this?
Sure we can use mocks instead of Machine and MachineConfig, but I purposefully use constructors because if we change one of this complex objects constructor we will have an obvious error, and it will simplify Troubleshooting.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
akorneta: I divided the logic of the saveToSnapshot method without changing comments to the existing code, but maybe you're right and in should be FIXME instead of TODO.

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/DockerInstanceTest.java
benoitf, line 86: my point are
1. when a new parameter is added in constructor, then we need to adjust constructor
so if I create object in each method, I need to add this new dummy object in every calls (adding many lines) even if I don't need this object in the test method.
2. By only testing the class matching <ClassName>Test.class and mocking the other dependencies we're sure to test the behavior of this object and to reach pretty close 100% of code coverage while by using some real objects.
3. I don't see the point of "changing complex object constructors" because maybe if you follow TDD approach you would have modify the tests before modifying the code. about troubleshooting, if a required parameter is not injected then instance won't be created and you'll get error

benoitf, line 86: I forgot also
4. by using only framework "injection", then you can use field injection instead of constructor injection (often Constructor injection was to be able to test the object) . With field injection, when you've class inheritance, adding a new parameter in top level class doesn't imply to change every constructor in subclasses.

benoitf, line 86: also notice it's a way to exchange, I don't require these changes at all

akorneta, line 86: Yes, I understand, nice points.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -150,3 +149,6 @@ workspace.runtime.auto_snapshot=false
 # During the start of the workspace automatically restored it from a snapshot if the value is {true},
 # otherwise just creates the new workspace.
 workspace.runtime.auto_restore=false
+# Allows to use registry for workspace snapshots, you should set this property in {true},
+# otherwise workspace snapshots would be saved locally.
+workspace.snapshot_use_registry=false
```
benoitf: missing end of line

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProviderTest.java
```diff
@@ -70,14 +70,15 @@
...
-    private static final int    MEMORY_LIMIT_MB     = 64;
+    private static final String  PROJECT_FOLDER_PATH = "/projects";
+    private static final String  CONTAINER_ID        = "containerId";
+    private static final String  WORKSPACE_ID        = "wsId";
+    private static final String  MACHINE_ID          = "machineId";
+    private static final String  MACHINE_NAME        = "machineName";
+    private static final String  USER_TOKEN          = "userToken";
+    private static final String  USER_NAME           = "user";
+    private static final int     MEMORY_LIMIT_MB     = 64;
+    private static final boolean locallySnapshot     = true;
```
benoitf: new name should be snapshotUseRegistry to be more accurate with field rename in tested class

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
```diff
@@ -187,36 +194,56 @@ public InstanceProcess createProcess(Command command, String outputChannel) thro
...
             final String repository = generateRepository();
-            String comment = format("Suspended at %1$ta %1$tb %1$td %1$tT %1$tZ %1$tY", System.currentTimeMillis());
-            if (owner != null) {
-                comment = comment + " by " + owner;
+            final String tag = "latest";
+            if (snapshotUseRegistry) {
+                final String digest = saveInRegistry(owner, repository, tag);
+                return new DockerInstanceKey(repository, tag, registry, digest);
+            } else {
+                createImage(owner, repository, tag);
```
garagatyi: I think it is better to name that method in the same way as `saveInRegistry` - `saveLocallly`

akorneta: :+1:  ok

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -361,8 +363,16 @@ public void removeInstanceSnapshot(InstanceKey instanceKey) throws SnapshotExcep
         // use registry API directly because docker doesn't have such API yet
         // https://github.com/docker/docker-registry/issues/45
         final DockerInstanceKey dockerInstanceKey = new DockerInstanceKey(instanceKey);
-        String registry = dockerInstanceKey.getRegistry();
-        String repository = dockerInstanceKey.getRepository();
+        if (!snapshotUseRegistry) {
```
garagatyi: Value of this property can be changed between calls of snaspshot saving and removal.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -239,9 +241,9 @@ public Instance createInstance(InstanceKey instanceKey,
                                    Machine machine,
                                    LineConsumer creationLogsOutput) throws NotFoundException, MachineException {
         final DockerInstanceKey dockerInstanceKey = new DockerInstanceKey(instanceKey);
-
-        pullImage(dockerInstanceKey, creationLogsOutput);
-
+        if (snapshotUseRegistry) {
```
garagatyi: Value of this property can be changed between calls of snapshot saving and recovering.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -361,8 +363,16 @@ public void removeInstanceSnapshot(InstanceKey instanceKey) throws SnapshotExcep
         // use registry API directly because docker doesn't have such API yet
         // https://github.com/docker/docker-registry/issues/45
         final DockerInstanceKey dockerInstanceKey = new DockerInstanceKey(instanceKey);
-        String registry = dockerInstanceKey.getRegistry();
-        String repository = dockerInstanceKey.getRepository();
+        if (!snapshotUseRegistry) {
+            try {
+                docker.removeImage(RemoveImageParams.create(dockerInstanceKey.getFullName()));
```
garagatyi, line 368: This code is not covered with tests

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -239,9 +241,9 @@ public Instance createInstance(InstanceKey instanceKey,
                                    Machine machine,
                                    LineConsumer creationLogsOutput) throws NotFoundException, MachineException {
         final DockerInstanceKey dockerInstanceKey = new DockerInstanceKey(instanceKey);
-
-        pullImage(dockerInstanceKey, creationLogsOutput);
-
+        if (snapshotUseRegistry) {
+            pullImage(dockerInstanceKey, creationLogsOutput);
```
garagatyi, line 245: This code is not covered with tests

garagatyi:
LGTM

voievodin:
ok

skabashnyuk:
ок

PR #1357, opened 2016-05-26
Remove wrong required parameters in DockerConnector methods params

mmorhun:
We had parameters in some methods of `DockerConnector` which is used in API requests even if they wasn't really required. Now we should remove they from required list.

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/CommitParams.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/CommitParams.java
@@ -39,10 +39,25 @@ public class CommitParams {
      * @return arguments holder with required parameters
      * @throws NullPointerException
      *         if {@code container} or {@code repository} is null
+     * @deprecated use {@link CommitParams#create(String)} instead
      */
+    @Deprecated
     public static CommitParams create(@NotNull String container, @NotNull String repository) {
         return new CommitParams().withContainer(container)
                                  .withRepository(repository);
+    }
+
+    /**
+     * Creates arguments holder with required parameters.
+     *
+     * @param container
+     *         id or name of container
+     * @return arguments holder with required parameters
+     * @throws NullPointerException
+     *         if {@code container} is null
+     */
+    public static CommitParams create(@NotNull String container) {
+        return new CommitParams().withContainer(container);
     }
 
     private CommitParams() {}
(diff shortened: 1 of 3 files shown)


plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/CommitParams.java
```diff
@@ -39,12 +39,27 @@
      * @return arguments holder with required parameters
      * @throws NullPointerException
      *         if {@code container} or {@code repository} is null
+     * @deprecated use
```
garagatyi: Use what?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/CommitParams.java
```diff
@@ -68,11 +83,8 @@ public CommitParams withContainer(@NotNull String container) {
      * @param repository
      *         full repository name
      * @return this params instance
-     * @throws NullPointerException
-     *         if {@code repository} is null
      */
-    public CommitParams withRepository(@NotNull String repository) {
```
garagatyi, line 74: Why it is OK to pass null here now?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/CommitParams.java
mmorhun: I saw and fixed it, while I was reviewing my code in diff here

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/CommitParams.java
mmorhun, line 74: Because `repository` is not required parameter and its default value is null

benoitf, line 74: maybe there is a missing notice explaining in the PR or commit why current status is wrong
it contains only the "how you fix it" but not "why it requires a fix"
Also I suppose some tests are missing

garagatyi, line 74: +1

mmorhun:
@garagatyi @benoitf is it looks good for you now?

garagatyi:
LGTM

PR #1359, opened 2016-05-26
Fix WorkspaceService to be able to share workspaces

sleshchenko:
@skabashnyuk  Can you please review my changes?
@evoevodin Can you please take a look changes in WorkspaceManager?


wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/WorkspaceDao.java
```diff
@@ -133,13 +133,13 @@
     List<WorkspaceImpl> getByNamespace(String namespace) throws ServerException;
 
     /**
-     * Gets list of workspaces where user is worker
+     * Gets list of workspaces where given user is worker and has read permission
```
skabashnyuk: WorkspaceDao should check permissions?

sleshchenko: It should get all workspaces by relation between user and workspace. Take a look implementation in codenvy project https://github.com/codenvy/codenvy/blob/workspacePermissions/core/codenvy-hosted-platform-api-impl/src/main/java/com/codenvy/api/dao/mongo/WorkspaceDaoImpl.java#L264-264

skabashnyuk:
ok

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -246,22 +246,22 @@ public WorkspaceImpl getWorkspace(String name, String namespace) throws NotFound
...
      *         when {@code owner} is null
      * @throws ServerException
-     *         when any server error occurs while getting workspaces with {@link WorkspaceDao#getByNamespace(String)}
+     *         when any server error occurs while getting workspaces with {@link WorkspaceDao#getWorkspaces(String)}
      */
-    public List<WorkspaceImpl> getWorkspaces(String namespace) throws ServerException {
-        requireNonNull(namespace, "Required non-null workspace namespace");
-        final List<WorkspaceImpl> workspaces = workspaceDao.getByNamespace(namespace);
+    public List<WorkspaceImpl> getWorkspaces(String user) throws ServerException {
+        requireNonNull(user, "Required non-null workspace namespace");
```
voievodin: fix namespace

sleshchenko: Fixed. Thx

voievodin:
The `WorkspaceManager` looks ok.

skabashnyuk:
ok

PR #1362, opened 2016-05-26
CHE-1230: return default test source dir if not set other

vparfonov:
@evidolob


evidolob:
LGTM

PR #1363, opened 2016-05-26
CHE-1244: Add checking of the workspace status during asynchronous stop

akorneta:
Fix the situation when directly invoke the stop of the workspace which status is not Running. Not added new tests, because the current test set completely covers the functionality.
@evoevodin, @skabashnyuk


voievodin:
good, lgtm

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -627,12 +631,9 @@ boolean createSnapshotSync(WorkspaceRuntimeImpl runtime, String namespace, Strin
         return devMachineSnapshotFailMessage == null;
     }
 
-    private void checkWorkspaceBeforeCreatingSnapshot(WorkspaceImpl workspace) throws ConflictException {
+    private void checkWorkspaceIsRunning(WorkspaceImpl workspace, String msg) throws ConflictException {
```
benoitf: as all parameters of checkWorkspaceIsRunning method are the same except the formatted string in both calls, maybe we should only provide the "unformatted string", and keep 
workspace.getNamespace(),
workspace.getConfig().getName(),
workspace.getStatus()) at a single place (there) to avoid copy/paste

skabashnyuk:
ок

PR #1366, opened 2016-05-27
CODENVY-560 : new types of docker recipes, remove InstanceKey, rework InstanceProvider

benoitf:
### 1. new docker recipe type

currently we have type:"dockerfile", location: "http://path-to-recipe"

now we could provide
type:"dockerfile", content: "FROM codenvy/foo\nENV FLORENT=TRUE\"

and
type:"image", location : "codenvy/foo"
### 2. InstanceKey

Up to now, InstanceKey was used to perform snapshot recovery.
But machine source is a way to provide this information.
So remove InstanceKey and replace it by MachineSource (and DockerMachineSource instead of DockerInstanceKey)

InstanceProvider:
void removeInstanceSnapshot(InstanceKey instanceKey)
--> void removeInstanceSnapshot(MachineSource machineSource)

Instance:
InstanceKey saveToSnapshot(String owner)
--> MachineSource saveToSnapshot(String owner)
### 3. InstanceProvider model

To avoid also that MachineManager "knows" the inner type, the recipe handling is moved to the instance provider implementation
And as the snapshot handling is with MachineSource (included in MachineConfig included in Machine), no need to give extra InstanceKey parameter

Replace two previous methods

Instance createInstance(Recipe recipe,
                            Machine machine,
                            LineConsumer creationLogsOutput)

 Instance createInstance(InstanceKey instanceKey,
                            Machine machine,
                            LineConsumer creationLogsOutput) throws NotFoundException, InvalidInstanceSnapshotException, MachineException;

by only one:
   createInstance(Machine machine,
                            LineConsumer creationLogsOutput)

Change-Id: Ia7ea97bc1a44059b4892f5db387f54f2e1709fa3


gazarenkov:
ok with me

@skabashnyuk @garagatyi ?

garagatyi:
This PR can't be merged, please rebase

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/MachineSource.java
```diff
@@ -24,4 +24,25 @@
...
+     * @return content of the machine source. No need to use an external link.
+     */
+    String getContent();
+
+    /**
+     * Defines the new content to use for this machine source.
+     * Alternate way is to provide a location
+     * @param content the content instead of an external link like with location
+     */
+    void setContent(String content);
```
garagatyi: We don't use setters in model interfaces

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/MachineSource.java
```diff
@@ -24,4 +24,25 @@
...
+
+    /**
+     * Defines the new content to use for this machine source.
+     * Alternate way is to provide a location
+     * @param content the content instead of an external link like with location
+     * @return the current intance of the object
+     */
+    MachineSource withContent(String content);
```
garagatyi: We don't use with methods in model interfaces

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
```diff
@@ -208,7 +211,7 @@ public InstanceKey saveToSnapshot(String owner) throws MachineException {
 
             docker.removeImage(registry + '/' + repository, false);
 
-            return new DockerInstanceKey(repository, "latest", registry, digest);
+            return new DockerMachineSource(repository).setRegistry(registry).setDigest(digest).setTag(LATEST_TAG);
```
garagatyi: We don't use fluent style setters, please consider usage of with methods instead

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -212,10 +228,50 @@ public String getType() {
...
+        MachineConfig machineConfig = machine.getConfig();
+        MachineSource machineSource = machineConfig.getSource();
+        String type = machineSource.getType();
+
+        // get recipe
+        // - it's a dockerfile type:
+        //    - location defined : download this location and get script as recipe
+        //    - content defined  : use this content as recipe script
+        // - it's an image:
+        //    - use either location or content as image ([registry:port]/<repository-image>[:tag][@digest])
```
garagatyi: I think we agreed that image should be placed in location field, not content. @gazarenkov confirmed to me that it is what he wanted. Please remove another case.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -234,21 +290,20 @@ public Instance createInstance(Recipe recipe,
...
-        final DockerInstanceKey dockerInstanceKey = new DockerInstanceKey(instanceKey);
 
-        pullImage(dockerInstanceKey, creationLogsOutput);
+    protected Instance doCreateInstanceImage(final Machine machine,
+                                        final LineConsumer creationLogsOutput) throws NotFoundException, MachineException {
+        final DockerMachineSource dockerMachineSource = new DockerMachineSource(machine.getConfig().getSource());
+
+        pullImage(dockerMachineSource, creationLogsOutput);
 
         final String userName = EnvironmentContext.getCurrent().getSubject().getUserName();
```
garagatyi: This code is duplicated. WDYT if we avoid duplication here?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -335,15 +390,15 @@ protected void buildImage(Dockerfile dockerfile,
         }
     }
 
-    private void pullImage(DockerInstanceKey dockerInstanceKey, final LineConsumer creationLogsOutput) throws MachineException {
-        if (dockerInstanceKey.getRepository() == null) {
+    private void pullImage(DockerMachineSource dockerMachineSource, final LineConsumer creationLogsOutput) throws MachineException {
+        if (dockerMachineSource.getRepository() == null) {
             throw new MachineException("Machine creation failed. Snapshot state is invalid. Please, contact support.");
```
garagatyi: With new source type `image` this situation can happen not only with snapshot

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
benoitf: it looks strange to not support same stuff by using location and content as it is the same value for the user.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -335,15 +390,15 @@ protected void buildImage(Dockerfile dockerfile,
...
         }
         try {
             final ProgressLineFormatterImpl progressLineFormatter = new ProgressLineFormatterImpl();
-            docker.pull(dockerInstanceKey.getRepository(),
-                        dockerInstanceKey.getTag(),
```
garagatyi, line 347: I think we should add latest tag in case tag field is null. AFAIK docker/swarm can download all tags otherwise

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -335,15 +390,15 @@ protected void buildImage(Dockerfile dockerfile,
...
 
-    private void pullImage(DockerInstanceKey dockerInstanceKey, final LineConsumer creationLogsOutput) throws MachineException {
-        if (dockerInstanceKey.getRepository() == null) {
+    private void pullImage(DockerMachineSource dockerMachineSource, final LineConsumer creationLogsOutput) throws MachineException {
+        if (dockerMachineSource.getRepository() == null) {
             throw new MachineException("Machine creation failed. Snapshot state is invalid. Please, contact support.");
         }
         try {
             final ProgressLineFormatterImpl progressLineFormatter = new ProgressLineFormatterImpl();
-            docker.pull(dockerInstanceKey.getRepository(),
```
garagatyi, line 346: Avoid usage of deprecated methods. If you edit that code, please use new methods.

wsmaster/wsmaster-local/src/test/java/org/eclipse/che/api/local/LocalSnapshotDaoTest.java
```diff
@@ -80,7 +83,7 @@ private SnapshotImpl createSnapshot() {
         return SnapshotImpl.builder()
                            .generateId()
                            .setType("docker")
-                           .setInstanceKey(new DummyInstanceKey())
+                           .setMachineSource(machineSource)
```
voievodin, line 86: What is the purpose of mock serialization ?
This test is for checking whether objects correctly serialized/deserialised because we had problems with it.

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/MachineSourceImpl.java
```diff
@@ -25,15 +27,28 @@
...
+
+    protected MachineSourceImpl() {
+
+    }
 
     public MachineSourceImpl(String type, String location) {
+        this(type, location, null);
+    }
+
+    public MachineSourceImpl(String type, String location, String content) {
```
garagatyi: Don't get why we need this constructor as location and content are mutually exclusive

wsmaster/che-core-api-machine-shared/src/main/java/org/eclipse/che/api/machine/shared/dto/MachineSourceDto.java
```diff
@@ -30,10 +31,33 @@
     MachineSourceDto withType(String type);
 
     @Override
-    @FactoryParameter(obligation = MANDATORY)
+    @FactoryParameter(obligation = OPTIONAL)
```
garagatyi, line 34: In that case it will be possible to save factory with neither location nor content. Does it seem OK to you @benoitf ?

plugins/plugin-ssh-machine/src/main/java/org/eclipse/che/plugin/machine/ssh/SshMachineInstanceProvider.java
```diff
@@ -62,41 +67,57 @@ public String getType() {
...
+     */
     @Override
-    public Instance createInstance(Recipe recipe,
-                                   Machine machine,
-                                   LineConsumer machineLogsConsumer) throws MachineException {
+    public Instance createInstance(Machine machine, LineConsumer lineConsumer)
+            throws UnsupportedRecipeException, InvalidRecipeException, NotFoundException, MachineException {
         requireNonNull(machine, "Non null machine required");
-        requireNonNull(machineLogsConsumer, "Non null logs consumer required");
+        requireNonNull(lineConsumer, "Non null logs consumer required");
```
garagatyi, line 92: Consider usage of source.getContent() here

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
```diff
@@ -0,0 +1,203 @@
...
+
+    /**
+     * Returns full name of docker image.
+     * <p>
+     * It consists of registry, userspace, repository name, tag, digest.
+     * E.g. docker-registry.company.com:5000/userspace1/my-repository:some-tag
+     * E.g. docker-registry.company.com:5000/userspace1/my-repository@some-digest
+     * @param includeDigest if digest needs to be included or not
+     */
+    public String getLocation(boolean includeDigest) {
```
garagatyi, line 187: Is there a need to evaluate this if you can use origin string you have parsed?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
```diff
@@ -0,0 +1,203 @@
...
+
+import static org.eclipse.che.plugin.docker.machine.DockerInstanceProvider.DOCKER_IMAGE_TYPE;
+
+/**
+ * Set of helper methods that identifies docker image properties
+ *
+ * @author Sergii Kabashnyuk
+ * @author Florent Benoit
+ */
+public class DockerMachineSource extends MachineSourceImpl {
```
garagatyi, line 27: Looks like you can use next classes instead of implementing pretty the same in that class
https://github.com/eclipse/che/blob/1dbb9d7c3ee96fc04cdfe04a33f2ea850d3038e3/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/parser/DockerImageIdentifier.java#L34-L34
https://github.com/eclipse/che/blob/342248f90936e31826942edfbfa601ad7036f8f5/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/parser/DockerImageIdentifierParser.java#L27-L27

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
gazarenkov: @benoitf  Content of what you want to support in case of "image"? 
Like downloading image binary?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
benoitf, line 187: yes it seems for recovering from snapshot you're using image name without digest (while the stored DockerMachineSource contains this digest)

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
benoitf, line 27: as I need to extends MachineSourceImpl using a wrapper/delegate won't reduce much the required lines of code
Also DockerInstanceKey was not previously using these classes at all as well

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/MachineSource.java
benoitf: ok

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/MachineSource.java
benoitf: ok

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
benoitf: ok

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
garagatyi, line 187: Yep, digest is needed for removal of image from registry

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
benoitf: @gazarenkov no it was the same content used for the location
"content": "codenvy/ubuntu_jdk8"
or
"location" :"codenvy/ubuntu_jdk8"

user pickup the property he wants, it doesn't matter.

BTW I may remove content support for "image" if it looks strange for you

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
benoitf: yes, I was reusing previous method call with this duplicated code

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
benoitf: ok

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
benoitf, line 347: I didn't change the existing logic so ?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
benoitf, line 346: ok, I was reusing existing code to avoid too many changes

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
garagatyi, line 27: I think duplication of parsing is a problem. We won't support both these parsing codes in the same state and it can introduce a bug. So it is better to use only 1 of them.
Note that you modified origin DockerInstanceKey logic, so I believe it is OK to use separate class that provides parsing logic.

wsmaster/wsmaster-local/src/test/java/org/eclipse/che/api/local/LocalSnapshotDaoTest.java
benoitf, line 86: yes for sure, but there is a dedicated test class for MachineSource serialization , MachineSourceAdapter + MachineSourceAdapterTest class

so this test will reuse MachineSourceAdapter which is tested

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/MachineSourceImpl.java
benoitf: ok

wsmaster/che-core-api-machine-shared/src/main/java/org/eclipse/che/api/machine/shared/dto/MachineSourceDto.java
benoitf, line 34: I think I've added a condition in WorkspaceService but I might add a new check in Factory Validator

I've not found an annotation @FactoryParameter with a "either content, either location" entry
Like "mutually exclusive"

plugins/plugin-ssh-machine/src/main/java/org/eclipse/che/plugin/machine/ssh/SshMachineInstanceProvider.java
benoitf, line 92: you want to add support for content or avoid it ?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
benoitf, line 187: https://github.com/eclipse/che/blob/45251ee275ae3390b45c1afd650b5e5d339a4a00/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java#L253-265

digest is not used

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
benoitf, line 27: ok so in this case I would get rid of DockerImageIdentiferParser as well (as MachineSource implementation is required by API while DockerImageIndentifierParser is an utility class)

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
garagatyi, line 347: You will change it, but it is more correct that old behavior

wsmaster/che-core-api-machine-shared/src/main/java/org/eclipse/che/api/machine/shared/dto/MachineSourceDto.java
garagatyi, line 34: We don't have a good way to handle such parameters in a factory now. Yes, I saw check in WSservice, but I'm saying about saving of factory, not accepting.
Probably we need some new features in factory validation framework to validate different types of implementation specific configs.
So I think it is ok for now

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
garagatyi, line 187: It is used here https://github.com/eclipse/che/blob/45251ee275ae3390b45c1afd650b5e5d339a4a00/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java#L384

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
garagatyi, line 27: I'm not OK with that decision. Parsing is docker specific feature and should be implemented in docker client framework

benoitf, line 27: BTW in all case this kind of work is part of another JIRA issue as it's far beyond the original scope of this issue. I may work on it but it won't be included with this JIRA/PR scope.
At first, refactoring InstanceKey was also not part of this issue.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
gazarenkov: yes, please

benoitf: Ok

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
garagatyi, line 27: Sorry, but it doesn't make sense to me. I've offered to reuse existing component instead of creating Regexp and other logic yourself and test it. And you answered that it is not part of this issue.

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/MachineSource.java
benoitf: fixed

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/MachineSource.java
benoitf: fixed

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
benoitf: I've removed 

but it's funny because here I clearly see this usage
https://github.com/eclipse/che/blob/a8dfb6bf541940c636590119ef2fb45c1fa659c9/wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/MachineSourceImpl.java#L44-56

also this kind of stuff can't be a rule maybe like a current practice but it's not true as I saw code using it. But clearly the with() usage is anti pattern as set() can of course return the instance of the object. It is more "contributor specific"
It avoids two methods and only change the return type. Also with() is clearly more seen with a fluent builder pattern than when there is no associated builder.

I may as well say that with fluent pattern, the usage of 'set' or 'get' is deprecated in favor of a more readable pattern
new MachineSource("type).location("my-location")
vs
new MachineSource("type").setLocation("my-location")
or new MachineSource("type").withLocation("my-location")

this kind of old stuff introduces too much verbosity which is not what tends to be "fluent " so I think it may be a good idea to let contributors pickup what they want

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
benoitf: fixed

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
benoitf: fixed

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
benoitf: fixed

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
benoitf, line 347: fixed

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
benoitf, line 346: it's funny because here https://github.com/eclipse/che/commit/e868d806498a806f3f4213922d8727e593abe230 (same class just one day before mine) you didn't say to not use anymore deprecated methods. So please keep the same review rules for everyone.

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/MachineSourceImpl.java
benoitf: the problem is how do you differentiate MachineSourceImpl(String type, String location) from MachineSourceImpl(String type, String content)

I've removed the location or content from constructor.

wsmaster/che-core-api-machine-shared/src/main/java/org/eclipse/che/api/machine/shared/dto/MachineSourceDto.java
benoitf, line 34: fixed, I've added check in validation of machinesource in WorkspaceValidator implementation (used in factory validation)

plugins/plugin-ssh-machine/src/main/java/org/eclipse/che/plugin/machine/ssh/SshMachineInstanceProvider.java
benoitf, line 92: fixed

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
benoitf, line 187: yes as I said, I don't see why the digest is not reused for a snapshot recovery as it clearly identify an image. BTW I keep same returned data.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -236,24 +298,18 @@ public Instance createInstance(Recipe recipe,
...
-        final DockerInstanceKey dockerInstanceKey = new DockerInstanceKey(instanceKey);
-        if (snapshotUseRegistry) {
-            pullImage(dockerInstanceKey, creationLogsOutput);
-        }
-        final String userName = EnvironmentContext.getCurrent().getSubject().getUserName();
-        final String machineContainerName = containerNameGenerator.generateContainerName(machine.getWorkspaceId(),
-                                                                                         machine.getId(),
-                                                                                         userName,
-                                                                                         machine.getConfig().getName());
+    protected Instance doCreateInstanceImage(final Machine machine, String machineContainerName,
```
garagatyi: This method returns Instance which is representation of the container. Please rename this method.

benoitf: It's creating instance for type image si method name is accurate

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
garagatyi, line 346: I have the same review standards  for all, but I'm not computer to find all the things in all PRs I'm reviewing.
BTW it is you who asked to create new API of docker client that will help to avoid backward incompatible changes. So I don't understand why you are against usage of it.

benoitf, line 346: First i'm not against it as I've updated code. 
I will also make sure this check is performed on all new incoming pull requests to match the same expectations towards my pull requests which is great to code quality

benoitf:
are we good now ?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
garagatyi: It looks to me as `create instance of image` instead of `create instance from image`. But if you don't want to change then ok.

benoitf: createInstanceFromImage() ?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
garagatyi, line 187: Because docker introduced digest later than we implemented that code and only latest versions of docker supports digest.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
```diff
@@ -358,30 +422,39 @@ private void pullImage(DockerInstanceKey dockerInstanceKey, final LineConsumer c
...
             return;
         }
 
-        final String registry = dockerInstanceKey.getRegistry();
-        final String repository = dockerInstanceKey.getRepository();
+        final String registry = dockerMachineSource.getRegistry();
+        final String repository = dockerMachineSource.getRepository();
         if (registry == null || repository == null) {
-            LOG.error("Failed to remove instance snapshot: invalid instance key: {}", instanceKey);
+            LOG.error("Failed to remove instance snapshot: invalid instance key: {}", dockerMachineSource);
```
garagatyi: It is not instance key anymore

benoitf: ok

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
benoitf, line 187: ok thx for explanation

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProviderTest.java
```diff
@@ -110,6 +113,9 @@
     @Captor
     private ArgumentCaptor<ContainerConfig> containerConfigArgumentCaptor;
 
+    @Mock
+    private RecipeRetriever recipeRetriever;
```
garagatyi, line 118: Have you tested instance creation from both types of dockerfile - content and location?

plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/machine/MachineManagerImpl.java
```diff
@@ -189,11 +190,11 @@ public void onMachineRunning(MachineStateEvent event) {
             @Override
             public void onMachineDestroyed(MachineStateEvent event) {
                 if (isMachineRestarting) {
-                    final String recipeUrl = machineState.getConfig().getSource().getLocation();
+                    final MachineSource machineSource = machineState.getConfig().getSource();
                     final String displayName = machineState.getConfig().getName();
                     final boolean isDev = machineState.getConfig().isDev();
 
-                    startMachine(recipeUrl, displayName, isDev, RESTART, "dockerfile", "docker");
+                    startMachine(asDto(machineSource), displayName, isDev, RESTART, "dockerfile", "docker");
```
garagatyi: please check that "dockerfile" is correct value of fifth argument in all cases

benoitf: hmm what I don't see is why it's hardcoded "dockerfile" and "docker" in machine manager ?
(ok here it's in IDE)

plugins/plugin-ssh-machine/src/main/java/org/eclipse/che/plugin/machine/ssh/SshMachineInstanceProvider.java
```diff
@@ -62,41 +67,58 @@ public String getType() {
...
+        requireNonNull(machine.getConfig().getSource().getLocation(), "Location in machine source is required");
```
garagatyi, line 93: What if content of recipe in inlined in content field?

benoitf, line 93: You didn't reply on my question https://github.com/eclipse/che/pull/1366#discussion_r64888388 so I keep original behavior as I'm not changing ssh as part of this PR.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
```diff
@@ -0,0 +1,226 @@
...
+    public DockerMachineSource(MachineSource machineSource) {
+        super();
+
+        // check type
+        if (!DOCKER_IMAGE_TYPE.equals(machineSource.getType())) {
+            throw new IllegalArgumentException("Docker machine source can only be built with '" + DOCKER_IMAGE_TYPE + "' type");
+        }
+        setType(DOCKER_IMAGE_TYPE);
+
+        // parse either content or location
```
garagatyi: I think you forgot to remove retrieval from content here.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
```diff
@@ -0,0 +1,226 @@
...
+     */
+    public String getLocation() {
+        return getLocation(true);
+    }
+
+    /**
+     * Returns full name of docker image.
+     * <p>
+     * It consists of registry, userspace, repository name, tag, digest.
+     * E.g. docker-registry.company.com:5000/userspace1/my-repository:some-tag
```
garagatyi: AFAIK it is not absolutely correct to say that this url has userspace in the given format. Docker doesn't declare userspace - only repository. Instead of that each provider of registry implement repository details as he wants. E.g. it can be absolutely correct that userspace is **/company/employee** in image **docker-registry.company.com:5000/codenvy/agaragatyi/my-repository:some-tag**

garagatyi: just FYI

benoitf: It comes from https://github.com/eclipse/che/blob/45251ee275ae3390b45c1afd650b5e5d339a4a00/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceKey.java#L57-62

benoitf: (as DockerMachineSource is the new name of previous class DockerInstanceKey)

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
benoitf: yes good catch

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
```diff
@@ -0,0 +1,226 @@
...
+        if (includeDigest && getDigest() != null) {
+            fullRepoId.append('@').append(getDigest());
+        }
+        return fullRepoId.toString();
+    }
+
+    /**
+     * Returns full name of docker image (without digest)
+     */
+    public String getFullName() {
```
garagatyi: Honestly it is quite confusing to me to see methods getLocation() and getFullName(). It is hard to remember what returns each of these methods if you don't read javadocs on each usage. Please consider usage getLocation(includeDigest) only.

benoitf: it was to keep backward compliance

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
```diff
@@ -0,0 +1,226 @@
...
+ * @author Florent Benoit
+ */
+public class DockerMachineSource extends MachineSourceImpl {
+
+    /**
+     * Regexp used to parse a location of Docker image.
+     * it's based on optional registry followed by repository name followed by a digest or a tag
+     * Examples are available in the test.
+     */
+    public static final Pattern IMAGE_PATTERN = Pattern.compile("^((?<registry>[^/]++)(?:\\/))?(?<repository>.*?)((?:\\:)(?<tag>.*?))?((?:@)(?<digest>.*))?$");
```
garagatyi: As far as I can see this regexp is not tested. Please review your tests.
Also it looks like 
**user/repo**  will be treated as registry **user** and repository **repo**

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/MachineSourceImpl.java
garagatyi: ok

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
garagatyi, line 346: Thx, you help in PR reviewing is appreciated

wsmaster/che-core-api-machine-shared/src/main/java/org/eclipse/che/api/machine/shared/dto/MachineSourceDto.java
garagatyi, line 34: ok

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
garagatyi, line 347: thx

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
garagatyi: +1

plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/machine/MachineManagerImpl.java
garagatyi: yes, it is IDE part, and IDK the answer to your question

plugins/plugin-ssh-machine/src/main/java/org/eclipse/che/plugin/machine/ssh/SshMachineInstanceProvider.java
garagatyi, line 93: I saw that you commented fixed, so I misunderstood you. I mean that now it is possible to provide either location or content of ssh-machine recipe so we can check that exact one of the fields (location or content) is not null.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
garagatyi: Ok, then I send this comment to Sergey too.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
garagatyi: But you changed the name of the class so no backward compliance is supposed I think.

plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/machine/MachineManagerImpl.java
benoitf: test added on MachineSource added
- "dockerfile" as been removed on this line

plugins/plugin-ssh-machine/src/main/java/org/eclipse/che/plugin/machine/ssh/SshMachineInstanceProvider.java
benoitf, line 93: it is possible but implementation may pickup what it wants. By default ssh-machine will stay as before (no content option) as it's not part of my issue. (same behavior that today)

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
benoitf: fixed

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
benoitf: I've removed "userspace" from existing comment

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
benoitf: I drop getFullName()

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
benoitf: There is a current class DockerMachineSourceTest with REGEXP being tested by constructor calls 
Also now it's delegating to DockerMachineIdentifierParser

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
benoitf: fixed

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
benoitf: fixed

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
benoitf, line 27: I've reused as plugin-docker-machine is based on plugin-docker (and not the opposite)

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
garagatyi: thx

plugins/plugin-ssh-machine/src/main/java/org/eclipse/che/plugin/machine/ssh/SshMachineInstanceProvider.java
garagatyi, line 93: ok

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineSource.java
garagatyi: Thx

garagatyi:
LGTM

plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/machine/MachineManagerImpl.java
```diff
@@ -208,6 +209,14 @@ public void apply(Void arg) throws OperationException {
         });
     }
 
+    /**
+     * Converts {@link MachineSource} to {@link MachineSourceDto}.
+     */
+    public MachineSourceDto asDto(MachineSource source) {
+        return this.dtoFactory.createDto(MachineSourceDto.class).withType(source.getType()).withLocation(source.getLocation()).withContent(source.getContent());
```
skabashnyuk: wrap lines? Would it be more readable?

skabashnyuk:
ok

plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/machine/MachineManagerImpl.java
benoitf: done

PR #1367, opened 2016-05-27
CHE-909, CHE-386: Synchronize machine stop/start with related runtime workspace environment

mmorhun:
We do not track temporary machines in API. As a consequence this additional machines is not stopped after stop of their workspace.

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
@@ -21,11 +21,13 @@ import org.eclipse.che.api.core.model.machine.MachineConfig;
 import org.eclipse.che.api.core.model.workspace.WorkspaceRuntime;
 import org.eclipse.che.api.core.model.workspace.WorkspaceStatus;
 import org.eclipse.che.api.core.notification.EventService;
+import org.eclipse.che.api.core.notification.EventSubscriber;
 import org.eclipse.che.api.machine.server.MachineManager;
 import org.eclipse.che.api.machine.server.exception.MachineException;
 import org.eclipse.che.api.machine.server.exception.SnapshotException;
 import org.eclipse.che.api.machine.server.model.impl.MachineConfigImpl;
 import org.eclipse.che.api.machine.server.model.impl.MachineImpl;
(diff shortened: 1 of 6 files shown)


mmorhun:
@garagatyi @evoevodin take a look please

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -279,6 +293,70 @@ public boolean hasRuntime(String workspaceId) {
         }
     }
 
+    @Override
+    public void onEvent(MachineStatusEvent event) {
+        if (!event.isDev()) {
```
garagatyi, line 298: Add comment why you add this clause

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -279,6 +293,70 @@ public boolean hasRuntime(String workspaceId) {
...
+                    WorkspaceStatus status = descriptor.getRuntimeStatus();
+                    if (status != WorkspaceStatus.RUNNING) {
+                        LOG.warn("Attempt to add machine {} to workspace {} which is {}",
+                                 machineId,
+                                 event.getWorkspaceId(),
+                                 status);
+                        destroyMachine(event.getMachineId());
+                    }
+
+                    Queue<MachineConfigImpl> queue = startQueues.get(event.getWorkspaceId());
```
garagatyi, line 321: I think this code should be synced

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -279,6 +293,70 @@ public boolean hasRuntime(String workspaceId) {
...
+                        }
+                    } else {
+                        try {
+                            addMachine(machineId);
+                        } catch (NotFoundException | ServerException | ConflictException e) {
+                            destroyMachine(machineId);
+                        }
+                    }
+                    break;
+                case DESTROYING:
```
garagatyi, line 341: Why not DESTROYED?

voievodin:
Will be fixed by #1384

PR #1375, opened 2016-05-27
CODENVY-470: Add user profile updated event

AndrienkoAleksandr:

### DIFF
--- a/wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/notification/ProfileUpdatedEvent.java
+++ b/wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/notification/ProfileUpdatedEvent.java
@@ -1,0 +1,45 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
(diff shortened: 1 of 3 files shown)


skabashnyuk:
Are you sure you need whole profile in event?

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/notification/ProfileUpdatedEvent.java
```diff
@@ -0,0 +1,45 @@
...
+      * Returns copy updated profile
+     */
+    public Profile getProfile() {
+        return profile;
+    }
+
+    @Override
+    public String toString() {
+        return "ProfileUpdatedEvent{" +
+               "profile=" + profile +
```
benoitf, line 42: Profile class has no toString() method so I'm not sure it is what is expected there

wsmaster/che-core-api-user/src/test/java/org/eclipse/che/api/user/server/UserProfileServiceTest.java
```diff
@@ -315,6 +320,7 @@ public void shouldBeAbleToUpdateCurrentProfileAttributes() throws Exception {
 
         assertEquals(response.getStatus(), OK.getStatusCode());
         verify(profileDao, times(1)).update(profile);
+        verify(eventService).publish(any());
```
benoitf, line 323: we should verify the profile is inside the event no ? call to getProfile() on the event

skabashnyuk:
Consider to remove Profile from event. 
It can be replaced with userid for instance.
No GO in current state.

AndrienkoAleksandr:
Code will be reworked.

PR #1382, opened 2016-05-30
CODENVY-441: Add get method to retrieve all ws-agents

akorneta:
Needed for CODENVY-441 see [here](https://github.com/codenvy/codenvy/pull/188/commits/5ce0e376e148dc4e5da9c4a303f689ff2025584c)
@benoitf please take a look

### DIFF
--- a/dashboard/src/components/api/che-workspace.factory.js
+++ b/dashboard/src/components/api/che-workspace.factory.js
@@ -89,6 +89,14 @@ export class CheWorkspace {
     return null;
   }
 
+/**
+ * Gets all workspace agents of this remote
+ * @returns {Map}
+ */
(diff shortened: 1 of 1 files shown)


dashboard/src/components/api/che-workspace.factory.js
```diff
@@ -89,6 +89,10 @@ export class CheWorkspace {
     return null;
   }
 
+  getWorkspaceAgents() {
```
benoitf, line 96: Could you please add javascriptdoc (especially the return type) (array, map, etc)

akorneta, line 96: sure

PR #1384, opened 2016-05-30
CHE-386: Synchronize machine start/stop with corresponding runtime

voievodin:
This PR provides fixes for `WorkspaceRuntimes` plus solves the problem described by @mmorhun in #1367, the solution is based on the approach provided by @mmorhun in the the mentioned pr.

The major changes:
- `WorkspaceRuntimes` uses distributed lock for the locking, this increases component performace and allows to perform more actions simultaneously.
- `WorkspaceRuntimes` listens for `RUNNING` and `DESTROYED` machine events and manages
  the corresponding workspace runtime states based on the incoming machine properties.
- The synchronization policy and code flow documented better
- Fix of some concurrency issues:
  - Use copy of the workspace runtime when stopping the workspace to prevent stale data effect, and
    inconsistent workspace runtime state as result. 
  - Use write lock instead of read lock when polling queue

@sleshchenko, @garagatyi, @akorneta and @mmorhun please review the changes

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
@@ -11,19 +11,19 @@
 package org.eclipse.che.api.workspace.server;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Predicate;
+import com.google.common.util.concurrent.Striped;
 
 import org.eclipse.che.api.core.BadRequestException;
 import org.eclipse.che.api.core.ConflictException;
 import org.eclipse.che.api.core.NotFoundException;
 import org.eclipse.che.api.core.ServerException;
 import org.eclipse.che.api.core.model.machine.MachineConfig;
+import org.eclipse.che.api.core.model.machine.MachineStatus;
 import org.eclipse.che.api.core.model.workspace.WorkspaceRuntime;
 import org.eclipse.che.api.core.model.workspace.WorkspaceStatus;
 import org.eclipse.che.api.core.notification.EventService;
 import org.eclipse.che.api.machine.server.MachineManager;
 import org.eclipse.che.api.machine.server.exception.MachineException;
-import org.eclipse.che.api.machine.server.exception.SnapshotException;
 import org.eclipse.che.api.machine.server.model.impl.MachineConfigImpl;
 import org.eclipse.che.api.machine.server.model.impl.MachineImpl;
 import org.eclipse.che.api.workspace.server.model.impl.EnvironmentImpl;
@@ -38,7 +38,6 @@ import javax.annotation.PreDestroy;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 import java.util.ArrayDeque;
-import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
@@ -46,7 +45,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Queue;
 import java.util.concurrent.locks.ReadWriteLock;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.Predicate;
 
 import static java.lang.String.format;
 import static org.eclipse.che.dto.server.DtoFactory.newDto;
@@ -54,14 +53,21 @@ import static org.eclipse.che.dto.server.DtoFactory.newDto;
 /**
  * Defines an internal API for managing {@link WorkspaceRuntimeImpl} instances.
  *
- * <p>This component implements {@link WorkspaceStatus} spec.
+ * <p>This component implements {@link WorkspaceStatus} spec,
+ * except of the {@link WorkspaceStatus#STOPPED} status, if workspace is STOPPED
+ * then {@link #get(String)} will thrown {@link NotFoundException}.
  *
  * <p>All the operations performed by this component are synchronous.
  *
- * <p>The implementation is thread-safe and guarded by {@link ReentrantReadWriteLock rwLock}.
+ * <p>The implementation is thread-safe and guarded by
+ * eagerly initialized readwrite locks produced by {@link WorkspaceRuntimes#STRIPED}.
+ * The component doesn't expose any api for client-side locking.
+ * All the instances produced by this component are copies of the real data.
  *
- * <p>The implementation doesn't validate parameters.
- * Parameters should be validated by caller of methods of this class.
+ * <p>The component doesn't check if the incoming objects are in application-valid state.
+ * Which means that it is expected that if {@link #start(WorkspaceImpl, String)} method is called
+ * then {@code WorkspaceImpl} argument is a application-valid object which contains
+ * all the required data for performing start.
  *
  * @author Yevhenii Voevodin
  * @author Alexander Garagatyi
@@ -69,9 +75,9 @@ import static org.eclipse.che.dto.server.DtoFactory.newDto;
 @Singleton
 public class WorkspaceRuntimes {
 
-    private static final Logger LOG = LoggerFactory.getLogger(WorkspaceRuntimes.class);
+    private static final Logger                 LOG     = LoggerFactory.getLogger(WorkspaceRuntimes.class);
+    private static final Striped<ReadWriteLock> STRIPED = Striped.readWriteLock(16);
 
-    private final ReadWriteLock                         rwLock;
     private final Map<String, RuntimeDescriptor>        descriptors;
     private final Map<String, Queue<MachineConfigImpl>> startQueues;
     private final MachineManager                        machineManager;
@@ -85,7 +91,6 @@ public class WorkspaceRuntimes {
         this.eventService = eventService;
         this.descriptors = new HashMap<>();
         this.startQueues = new HashMap<>();
-        this.rwLock = new ReentrantReadWriteLock();
     }
 
     /**
@@ -93,20 +98,19 @@ public class WorkspaceRuntimes {
      * workspace runtime.
      *
      * <p>Note that the {@link RuntimeDescriptor#getRuntime()} method
-     * returns {@link Optional} which describes just a snapshot copy of
-     * a real {@code WorkspaceRuntime} object, which means that any
-     * runtime copy modifications won't affect the real object and also
-     * it means that copy won't be affected with modifications applied
+     * returns a copy of a real {@code WorkspaceRuntime} object,
+     * which means that any runtime copy modifications won't affect the
+     * real object and also it means that copy won't be affected with modifications applied
      * to the real runtime workspace object state.
      *
      * @param workspaceId
      *         the id of the workspace to get its runtime
      * @return descriptor which describes current state of the workspace runtime
      * @throws NotFoundException
-     *         when workspace with given {@code workspaceId} doesn't have runtime
+     *         when workspace with given {@code workspaceId} is not running
      */
     public RuntimeDescriptor get(String workspaceId) throws NotFoundException {
-        rwLock.readLock().lock();
+        acquireReadLock(workspaceId);
         try {
             final RuntimeDescriptor descriptor = descriptors.get(workspaceId);
             if (descriptor == null) {
@@ -114,7 +118,7 @@ public class WorkspaceRuntimes {
             }
             return new RuntimeDescriptor(descriptor);
         } finally {
-            rwLock.readLock().unlock();
+            releaseReadLock(workspaceId);
         }
     }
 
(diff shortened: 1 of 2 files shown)


wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -85,36 +98,42 @@ public WorkspaceRuntimes(MachineManager machineManager, EventService eventServic
         this.eventService = eventService;
         this.descriptors = new HashMap<>();
         this.startQueues = new HashMap<>();
-        this.rwLock = new ReentrantReadWriteLock();
+        this.addMachineEventSubscriber = new AddMachineEventSubscriber();
+        this.removeMachineEventSubscriber = new RemoveMachineEventSubscriber();
+    }
+
+    @PostConstruct
+    private void subscribe() {
```
garagatyi: Please move private methods below public ones

garagatyi:
LGTM

mmorhun:
good!

sleshchenko:
LGTM

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -85,36 +98,36 @@ public WorkspaceRuntimes(MachineManager machineManager, EventService eventServic
...
      * @param workspaceId
      *         the id of the workspace to get its runtime
      * @return descriptor which describes current state of the workspace runtime
      * @throws NotFoundException
-     *         when workspace with given {@code workspaceId} doesn't have runtime
+     *         when workspace with given {@code workspaceId} is not running
      */
     public RuntimeDescriptor get(String workspaceId) throws NotFoundException {
-        rwLock.readLock().lock();
+        acquireReadLock(workspaceId);
```
benoitf, line 123: as you're introducing a new paradigm, you should benefit with try() resources with something autocloseable. (no need to forget the release call)

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -34,48 +37,58 @@
...
-    private static final Logger LOG = LoggerFactory.getLogger(WorkspaceRuntimes.class);
+    private static final Logger                 LOG     = LoggerFactory.getLogger(WorkspaceRuntimes.class);
+    private static final Striped<ReadWriteLock> STRIPED = Striped.readWriteLock(16);
 
-    private final ReadWriteLock                         rwLock;
-    private final Map<String, RuntimeDescriptor>        descriptors;
-    private final Map<String, Queue<MachineConfigImpl>> startQueues;
-    private final MachineManager                        machineManager;
-    private final EventService                          eventService;
+    @VisibleForTesting
```
benoitf, line 84: you shouldn't need this kind of annotation as it is like there is a design flaw if you're requiring it.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -34,48 +37,58 @@
...
  *
  * @author Yevhenii Voevodin
  * @author Alexander Garagatyi
  */
 @Singleton
 public class WorkspaceRuntimes {
 
-    private static final Logger LOG = LoggerFactory.getLogger(WorkspaceRuntimes.class);
+    private static final Logger                 LOG     = LoggerFactory.getLogger(WorkspaceRuntimes.class);
+    private static final Striped<ReadWriteLock> STRIPED = Striped.readWriteLock(16);
```
benoitf, line 82: comment about why 16 value is coming from

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -306,199 +372,246 @@ void publishEvent(EventType type, String workspaceId, String error) {
...
-        }
 
-        // Clean up the start queue when all the machines successfully started
-        rwLock.writeLock().lock();
-        try {
-            final Queue<MachineConfigImpl> queue = startQueues.get(wsId);
-            if (queue != null && queue.isEmpty()) {
-                startQueues.remove(wsId);
+            // According to WorkspaceStatus specification the workspace start
+            // is failed when dev-machine start is failed, so if any error
```
benoitf, line 452: dev-machine does not exist anymore (it's ws-machine)

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -306,199 +372,246 @@ void publishEvent(EventType type, String workspaceId, String error) {
...
                 if (machine != null) {
-                    final WorkspaceRuntimeImpl runtime = descriptors.get(wsId).getRuntime();
-                    if (config.isDev()) {
-                        runtime.setDevMachine(machine);
-                        publishEvent(EventType.RUNNING, wsId, null);
-                    }
-                    runtime.getMachines().add(machine);
+                    machineManager.destroy(machine.getId(), false);
                 }
+                throw new ConflictException(format("Workspace '%s' start interrupted. " +
```
benoitf: why do we need a string concatenation ?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -510,44 +623,208 @@ private void ensurePreDestroyIsNotExecuted() throws ServerException {
...
+        }
+    }
+
+    /**
+     * Tries to add the machine to the WorkspaceRuntime.
+     *
+     * @return true if machine is added or will be added to the corresponding WorkspaceRuntime,
+     * returns false when incoming machine can't be added and should be destroyed.
+     */
+    @VisibleForTesting
```
benoitf, line 682: There is no need to have package access and VisibleForTesting

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
voievodin, line 452: Well it is dev-machine throughout all the code and [model](https://github.com/eclipse/che/blob/master/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/MachineConfig.java#L36), so as long as model is not changed it is dev-machine.

benoitf, line 452: It's not "all the code" as it's new name
it is also stored in the workspaces
https://github.com/eclipse/che/blob/ad5d3d9d7f084c8fa6441a279c42bc82d4dc4cb1/core/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/workspace/create/CreateWorkspacePresenter.java#L230

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
voievodin, line 84: This field is package published, which means that clients can use it, when we put this annotation on the field/method it is still package visible but, it says that field shouldn't be used directly as it is exposed for tests. 

> Annotates a program element that exists, or is more widely visible than
> otherwise necessary, only for use in test code.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
voievodin, line 452: It will be confusing if ws-machine mentioned in comments while the checks are still `machine.isDev` like. It will be changed in couple with model

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
voievodin, line 682: Okay, then can you please describe the use-case when `VisibleForTesting` is necessary, i see it only in couple with package private elements

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
benoitf, line 84: I know about this annotation but package access is still not required and all the VisibleForTesting are polluting class more than anything else

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
benoitf, line 682: why do you want to keep package access ? and not for example protected for methods ?

benoitf, line 682: about the use-case for "VisibleForTesting", I don't see any valid use-case in Eclipse Che.

voievodin, line 682: Because `protected` is more wide, and methods are not designed to be accessible in subclasses(if any)

benoitf, line 682: and I still don't get why anyone couldn't subclasses and override this existing code. With the "keep everything private" it's tending to avoid something pluggable, overridable and extensible.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
benoitf, line 123: try with resources:
try (acquireReadLock(workspaceId)) {

} // do not need to release the lock

voievodin, line 123: Well the idea is good, but i will not change the code as it requires an infrastructure under it.
The design of the solution should be moved to the another place like commons and reused later by another clients. The short alias methods are quiet simple and look good for me right now, so i will not change it, but will keep in mind. Thank you

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
voievodin, line 82: ok

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
voievodin: no we don't, thank you!

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -229,17 +267,28 @@ public void stop(String workspaceId) throws NotFoundException, ServerException,
                 throw new NotFoundException("Workspace with id '" + workspaceId + "' is not running.");
             }
             if (descriptor.getRuntimeStatus() != WorkspaceStatus.RUNNING) {
-                throw new ConflictException(format("Couldn't stop '%s' workspace because its status is '%s'",
+                throw new ConflictException(format("Couldn't stop '%s' workspace because its status is '%s'. " +
```
benoitf: String concatenation here as well ?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -306,199 +372,246 @@ void publishEvent(EventType type, String workspaceId, String error) {
...
+     * @throws ServerException
+     *         only if pre destroy has been invoked before peek config retrieved
+     */
+    private MachineConfigImpl queuePeekOrFail(String workspaceId) throws ConflictException, ServerException {
+        acquireReadLock(workspaceId);
+        try {
+            ensurePreDestroyIsNotExecuted();
+            final Queue<MachineConfigImpl> queue = startQueues.get(workspaceId);
+            if (queue == null) {
+                throw new ConflictException(format("Workspace '%s' start interrupted. " +
```
benoitf: - concatenation

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -306,199 +372,246 @@ void publishEvent(EventType type, String workspaceId, String error) {
...
-        } catch (BadRequestException brEx) {
-            // TODO fix this in machineManager
-            throw new IllegalArgumentException(brEx.getLocalizedMessage(), brEx);
-        }
-    }
-
-    private void ensurePreDestroyIsNotExecuted() throws ServerException {
-        if (isPreDestroyInvoked) {
-            throw new ServerException("Could not perform operation because application server is stopping");
+        } catch (ConflictException x) {
```
benoitf, line 584: note: why x instead of 'usual' e ?

voievodin, line 584: i like x, i saw it alot in jdk internals, any problems with x?

benoitf, line 584: no problem at all. just something "homogeneous"

voievodin:
Thank you for the review guys

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -510,44 +623,208 @@ private void ensurePreDestroyIsNotExecuted() throws ServerException {
...
+                         machineId,
+                         workspaceId,
+                         machineName);
+            }
+        } finally {
+            releaseWriteLock(workspaceId);
+        }
+    }
+
+    private static <T> T removeFirstMatching(List<? extends T> elements, Predicate<T> predicate) {
```
benoitf, line 793: shouldn't it belong to a commons utility class ?

PR #1390, opened 2016-05-30
Add Push to Gerrit functionality

offer8:
1. Fix storing issue in JGitConfigImpl
2. Commit command scenario - Get gerrit changid property from configuration
   and set the commit command to set insert change ID to work with Gerrit
3. Push command scenario - Create PushResponse with details regarding updates
   and refs and use it in push command

Signed-off-by: Offer Shostak offershostak@gmail.com

### DIFF
--- a/wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConfigImpl.java
+++ b/wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConfigImpl.java
@@ -16,6 +16,7 @@ import org.eclipse.che.api.git.GitException;
 import org.eclipse.jgit.lib.Repository;
 import org.eclipse.jgit.lib.StoredConfig;
 
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -29,6 +30,7 @@ import static java.lang.System.lineSeparator;
  * @author Tareq Sharafy (tareq.sha@sap.com)
  */
 class JGitConfigImpl extends Config {
+
     private final Repository repository;
 
     JGitConfigImpl(Repository repository) throws GitException {
@@ -96,6 +98,11 @@ class JGitConfigImpl extends Config {
     public Config set(String name, String value) throws GitException {
         ConfigKey key = parseName(name);
         repository.getConfig().setString(key.section, key.subsection, key.name, value);
+        try {
+            this.repository.getConfig().save();
+        } catch (IOException e) {
+            throw new GitException(e.getMessage(), e);
+        }
         return this;
     }
 
@@ -108,6 +115,11 @@ class JGitConfigImpl extends Config {
     public Config unset(String name) throws GitException {
         ConfigKey key = parseName(name);
         repository.getConfig().unset(key.section, key.subsection, key.name);
+        try {
+            this.repository.getConfig().save();
+        } catch (IOException e) {
+            throw new GitException(e.getMessage(), e);
+        }
         return this;
     }
 
--- a/wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
+++ b/wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
@@ -125,6 +125,7 @@ import org.eclipse.jgit.transport.PushResult;
 import org.eclipse.jgit.transport.RefSpec;
 import org.eclipse.jgit.transport.RemoteConfig;
 import org.eclipse.jgit.transport.RemoteRefUpdate;
+import org.eclipse.jgit.transport.TrackingRefUpdate;
 import org.eclipse.jgit.transport.SshSessionFactory;
 import org.eclipse.jgit.transport.SshTransport;
 import org.eclipse.jgit.transport.Transport;
@@ -150,6 +151,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashSet;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
@@ -175,6 +177,21 @@ class JGitConnection implements GitConnection {
     private static final String REBASE_OPERATION_CONTINUE = "CONTINUE";
     private static final String REBASE_OPERATION_ABORT    = "ABORT";
     private static final String ADD_ALL_OPTION            = "all";
+
+    // Push Response Constants
+    private static final String BRANCH_REFSPEC_SEPERATOR = " -> ";
+    private static final String REFSPEC_COLON = ":";
+    private static final String KEY_COMMIT_MESSAGE = "Message";
+    private static final String KEY_RESULT = "Result";
+    private static final String KEY_REMOTENAME = "RemoteName";
+    private static final String KEY_LOCALNAME = "LocalName";
+    private static final String INFO_PUSH_ATTEMPT_IGNORED_UP_TO_DATE    = "Could not push because the repository is up-to-date.";
+    private static final String ERROR_PUSH_ATTEMPT_FAILED_WITHMSG       = "Could not push refspec {0} to the branch {1}. Try to merge " +
+                                                                          "the remote changes using pull, and then push again. \n Error " +
+                                                                          "Status: {2}, Error message {3}. \n";
+    private static final String ERROR_PUSH_ATTEMPT_FAILED_WITHOUTMSG    = "Could not push refspec {0} to the branch {1}. Try to merge " +
+                                                                          "the remote changes using pull, and then push again. \n " +
+                                                                          "Error Status: {2}. \n";
 
     private static final String ERROR_UPDATE_REMOTE_NAME_MISSING   = "Update operation failed, remote name is required.";
     private static final String ERROR_ADD_REMOTE_NAME_MISSING      = "Add operation failed, remote name is required.";
@@ -491,6 +508,12 @@ class JGitConnection implements GitConnection {
                                                   .setAll(request.isAll())
                                                   .setAmend(request.isAmend());
 
+            // Check if repository is configured with Gerrit Support
+            String gerritSupportConfigValue = repository.getConfig().getString(
+                    ConfigConstants.CONFIG_GERRIT_SECTION, null,
+                    ConfigConstants.CONFIG_KEY_CREATECHANGEID);
+            boolean isGerritSupportConfigured = gerritSupportConfigValue != null ? Boolean.valueOf(gerritSupportConfigValue) : false;
+            commitCommand.setInsertChangeId(isGerritSupportConfigured);
             RevCommit result = commitCommand.call();
             GitUser gitUser = newDto(GitUser.class).withName(committerName).withEmail(committerEmail);
 
@@ -957,8 +980,8 @@ class JGitConnection implements GitConnection {
     public PushResponse push(PushRequest request) throws GitException, UnauthorizedException {
         String remoteName = request.getRemote();
         String remoteUri = getRepository().getConfig().getString(ConfigConstants.CONFIG_REMOTE_SECTION, remoteName,
-                                                                 ConfigConstants.CONFIG_KEY_URL);
-
+                ConfigConstants.CONFIG_KEY_URL);
+        PushResponse pushResponseDto = newDto(PushResponse.class);
         try {
             PushCommand pushCommand = getGit().push();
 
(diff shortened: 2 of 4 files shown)


offer8:
@vinokurig

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConfigImpl.java
```diff
@@ -96,6 +97,11 @@ private void addConfigValues(String section, String subsection, Set<String> name
     public Config set(String name, String value) throws GitException {
         ConfigKey key = parseName(name);
         repository.getConfig().setString(key.section, key.subsection, key.name, value);
+        try {
+            this.repository.getConfig().save();
+        } catch (IOException e) {
+            throw new GitException("Failed to save to config file");
```
benoitf: as exception is dropped, should we at least log it ?

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConfigImpl.java
```diff
@@ -108,6 +114,11 @@ public Config add(String name, String value) throws GitException {
     public Config unset(String name) throws GitException {
         ConfigKey key = parseName(name);
         repository.getConfig().unset(key.section, key.subsection, key.name);
+        try {
+            this.repository.getConfig().save();
+        } catch (IOException e) {
+            throw new GitException("Failed to save to config file");
```
benoitf: as exception is dropped, should we at least log it ?

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -981,26 +1003,70 @@ public PushResponse push(PushRequest request) throws GitException, UnauthorizedE
...
-                throw new GitException(ERROR_NO_REMOTE_REPOSITORY, exception);
+            return newDto(PushResponse.class).withCommandOutput(commandOutput).withUpdates(updates);
+        }
+        return newDto(PushResponse.class);
+    }
+
+    private String buildPushFailedMessage(final PushRequest request, final RemoteRefUpdate remoteRefUpdate,
+                                          final String currentBranch) {
+        String message;
+        if (remoteRefUpdate.getStatus().equals(RemoteRefUpdate.Status.UP_TO_DATE)) {
```
benoitf: if (RemoteRefUpdate.Status.UP_TO_DATE.equals(remoteRefUpdate.getStatus())
or 
replace equals by ==

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -981,26 +1003,70 @@ public PushResponse push(PushRequest request) throws GitException, UnauthorizedE
...
+            PushResult pushResult = list.iterator().next();
+            String currentBranch = getCurrentBranch();
+            return generatePushResponse(request, pushResult, currentBranch);
+        } catch (GitAPIException exception) {
+            return newDto(PushResponse.class).withCommandOutput(exception.getMessage());
+        }
+    }
+
+    private PushResponse generatePushResponse(final PushRequest request, final PushResult result, final String currentBranch) {
+        if (result != null) {
```
benoitf: can the variable result be null ?

offer8: Yes

offer8:
I fixed the comments.

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConfigImpl.java
```diff
@@ -108,6 +121,12 @@ public Config add(String name, String value) throws GitException {
     public Config unset(String name) throws GitException {
         ConfigKey key = parseName(name);
         repository.getConfig().unset(key.section, key.subsection, key.name);
+        try {
+            this.repository.getConfig().save();
+        } catch (IOException e) {
+            LOG.debug(ERROR_FAIL_SAVE_CONFIG);
```
benoitf: but we're still not getting e root cause , no ?

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
benoitf: In which case do we have null ?

offer8:
Fixed.

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
offer8: Line 1005: "Iterable<PushResult> list = (Iterable<PushResult>)executeRemoteCommand(remoteUri, pushCommand);"
the list can return empty hence:
PushResult pushResult = list.iterator().next();
Will be null.

benoitf: ok so we tend to provide something null from something which is non null : the list (empty)
currently we have:
iterator: empty or not
get iterator which can be null
then check for null or not-null
null--> return PushResponse DTO
non null --> return PushResponse populated with commandOutput and updates

could it be more readable if :
we always generate PushResponse DTO
if list is not empty then we add commandOutput/updates

```
create dto
if iterator hasNext
  addCommandAndOutput(dto, request, iterator.next/pushResult)
return dto
```

also I notice that currently we do
String currentBranch = getCurrentBranch();
generatePushResponse(request, pushResult, currentBranch);

we give currentBranch which is a class method so we could get rid of currentBranch parameter as generatePushResponse method is able to call it.

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConfigImpl.java
```diff
@@ -96,6 +103,12 @@ private void addConfigValues(String section, String subsection, Set<String> name
     public Config set(String name, String value) throws GitException {
         ConfigKey key = parseName(name);
         repository.getConfig().setString(key.section, key.subsection, key.name, value);
+        try {
+            this.repository.getConfig().save();
+        } catch (IOException e) {
+            LOG.debug(String.format(ERROR_FAIL_SAVE_CONFIG, e.getCause().getMessage()), e.getCause());
```
benoitf: note: as now you're sending the exception in the GitException (before it was dropped) the logger call is optional

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConfigImpl.java
```diff
@@ -108,6 +121,12 @@ public Config add(String name, String value) throws GitException {
     public Config unset(String name) throws GitException {
         ConfigKey key = parseName(name);
         repository.getConfig().unset(key.section, key.subsection, key.name);
+        try {
+            this.repository.getConfig().save();
+        } catch (IOException e) {
+            LOG.debug(String.format(ERROR_FAIL_SAVE_CONFIG, e.getCause().getMessage()), e.getCause());
```
benoitf: note: as now you're sending the exception in the GitException (before it was dropped) the logger call is optional

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
vinokurig: How about 
`if (result == null) {
    throw new GitException("No push results received")
}
...` ?
You will get rid of one nesting level in this case and exception will be thrown instead of empty response

benoitf: @vinokurig ok as well
I was keeping the PR flow which was always returning PushResponse, not an exception

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -981,26 +1003,70 @@ public PushResponse push(PushRequest request) throws GitException, UnauthorizedE
 
             @SuppressWarnings("unchecked")
             Iterable<PushResult> list = (Iterable<PushResult>)executeRemoteCommand(remoteUri, pushCommand);
```
vinokurig: Please rename this variable `list` to `pushResults` for example

offer8:
Fixed all the comments.

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -980,27 +1003,72 @@ public PushResponse push(PushRequest request) throws GitException, UnauthorizedE
...
-                            return newDto(PushResponse.class).withCommandOutput("Everything up-to-date");
-                        } else {
-                            throw new GitException(remoteRefUpdate.getStatus().toString());
-                        }
+            Iterable<PushResult> pushResults = (Iterable<PushResult>)executeRemoteCommand(remoteUri, pushCommand);
+            if (pushResults.iterator().hasNext()) {
+                PushResult pushResult = pushResults.iterator().next();
+                pushResponseDto = addCommandOutputUpdates(pushResponseDto, request, pushResult);
+            }
+            return pushResponseDto;
```
vinokurig: `pushResponseDto` can be empty here, may be it would be better to throw exception if it is not updated?

offer8: I prefer not to because it is not an error.

offer8:
Can you merge the code?

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
vinokurig: In what cases `pushResults` can be empty list?

offer8: PushCommand.call()->Transport.openall()->if cfg not exist goes to Transport.openAll()->Transport.getURIs->can be empty in CTOR of RemoteConfig when Config.getStringList returns EMPTY_STRING_ARRAY

vinokurig: Strange, what version of JGit do you use? My is 4.3.1.201605051710-r
I have this in openAll method in JGit Transport class:

public static List<Transport> openAll(final Repository local,
            final String remote, final Operation op)
            throws NotSupportedException, URISyntaxException,
            TransportException {
        final RemoteConfig cfg = new RemoteConfig(local.getConfig(), remote);
        if (doesNotExist(cfg)) {
            final ArrayList<Transport> transports = new ArrayList<Transport>(1);
            transports.add(open(local, new URIish(remote), null));
            return transports;
        }
        return openAll(local, cfg, op);
    }

So if cfg does not exist the method returns List with one Transport that is never null

offer8: Yes it's the same code. I will remove the "if (pushResults.iterator().hasNext())" check entirely.

offer8:
Fixed.

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -957,8 +980,8 @@ public PullResponse pull(PullRequest request) throws GitException, UnauthorizedE
     public PushResponse push(PushRequest request) throws GitException, UnauthorizedException {
         String remoteName = request.getRemote();
         String remoteUri = getRepository().getConfig().getString(ConfigConstants.CONFIG_REMOTE_SECTION, remoteName,
-                                                                 ConfigConstants.CONFIG_KEY_URL);
-
+                ConfigConstants.CONFIG_KEY_URL);
+        PushResponse pushResponseDto = newDto(PushResponse.class);
```
vinokurig: You can get rid of this line by generating newDto in line 1049

vinokurig:
OK for me, @skabashnyuk ?

offer8:
?

skabashnyuk:
ok.

offer8:
Can someone merge the request?

vinokurig:
PushTest failed. please fix it and rebase with latest master

offer8:
Which test?

offer8:
Fix the API test - does not return the exception in the JGIT flow

vinokurig:
Actually it returns exception, add to line 1010 old flow instead of removing test:
`if ("origin: not found.".equals(exception.getMessage())) {`
`throw new GitException(ERROR_NO_REMOTE_REPOSITORY, exception);`
`} else {`
`throw new GitException(exception.getMessage(), exception);`
`}`

offer8:
But we do not want to throw exception here.
This is part of our change - we build the entire message and scenanrio for this.

vinokurig:
Native git throws exception if no remote repository specified. If we will not throw exception the response status will be 200 OK, but actually it is not because push was not performed @skabashnyuk what do you think?

offer8:
OK, I will change the logic to throw an exception.

offer8:
Changed.

vinokurig:
I will check soon

vinokurig:
Please merge latest master to your branch and perform rebase, then update me, I will merge it

offer8:
It is rebased against master

vinokurig:
merged manually to master

PR #1391, opened 2016-05-30
Update Vagrantfile

TylerJewell:
Fixes issue where vagrant suspend, vagrant up fails to restart the container.  Discovered as part of CHE-1328 issue with ARTIK IDE.


TylerJewell:
@eivantsov

PR #1393, opened 2016-05-31
CHE-1078: UserService structural refactorings

voievodin:
I separated changes on different commits which can be reviewed separately:
- Profile & User [models](https://github.com/eclipse/che/commit/a1d1be312e3eb2d882090ffc77f3d1623edc8a37)
- Components [separation](https://github.com/eclipse/che/commit/5410340ddfd290c1f724b0808ee950c6ca9853a2) 
  - Extract preferences API from the ProfileService
  - Add Manager layer for profile/user/preferences
  - Rewrite REST services tests using RestAssured
- Dashboard [refactorings](https://github.com/eclipse/che/commit/62a9cb6c47dc059d1768105a3a6a0cca7274a2ab)
- Infrastructure [adaptation](https://github.com/eclipse/che/commit/7d4c20a4794805596302488f248744f4e962e99d) to the refactorings

@skabashnyuk, @benoitf, @akorneta please review the pull request.
Please pay attention on classes `User`, `Profile`, `UserManager`, `ProfileManager`, `PreferencesManager`, `UserService`, `ProfileService`, `PreferencesService`


wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/PreferencesService.java
```diff
@@ -0,0 +1,128 @@
...
+                                                      "all the user preferences"),
+                   @ApiResponse(code = 400, message = "Request doesn't contain preferences update"),
+                   @ApiResponse(code = 500, message = "Couldn't update preferences due to internal server error")})
+    public Map<String, String> update(Map<String, String> preferences) throws ServerException, BadRequestException {
+        if (preferences == null) {
+            throw new BadRequestException("Required non-null preferences update");
+        }
+        return preferencesManager.update(userId(), preferences);
+    }
+
```
benoitf, line 104: what about a update(String key, String value) and remove(String key) (no bulk mode)

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/PreferencesManager.java
```diff
@@ -0,0 +1,174 @@
...
+     */
+    public void remove(String userId) throws ServerException {
+        requireNonNull(userId, "Required non-null user id");
+        preferenceDao.remove(userId);
+    }
+
+    /**
+     * Removes the preferences with the given {@code names}.
+     *
+     * @param names
```
benoitf: missing @param userId

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/user/Profile.java
```diff
@@ -0,0 +1,45 @@
...
+ * @see User
+ */
+public interface Profile {
+
+    /**
+     * Returns the identifier of the user {@link User#getId()}
+     * whom this profile belongs to, which must be also considered
+     * as profile identifier.
+     */
+    String getId();
```
benoitf: getUserId() then ?

dashboard/src/app/index.module.js
```diff
@@ -39,11 +39,11 @@ let initModule = angular.module('userDashboard', ['ngAnimate', 'ngCookies', 'ngT
 initModule.config(['$routeProvider', ($routeProvider) => {
   $routeProvider.accessWhen = (path, route) => {
     route.resolve || (route.resolve = {});
-    route.resolve.app = ['cheBranding', '$q', 'cheProfile', 'cheUser', (cheBranding, $q, cheProfile, cheUser) => {
+    route.resolve.app = ['cheBranding', '$q', 'cheUser', 'chePreferences', (cheBranding, $q, cheUser, chePreferences) => {
       var deferred = $q.defer();
 
       cheUser.fetchUser().then(() => {
-        let profilePreferences = cheProfile.getPreferences();
+        let profilePreferences = chePreferences.getPreferences();
```
benoitf: please rename profilePreferences as it's not anymore in profile

dashboard/src/app/index.module.js
```diff
@@ -65,11 +65,11 @@ initModule.config(['$routeProvider', ($routeProvider) => {
 
   $routeProvider.accessOtherWise = (route) => {
     route.resolve || (route.resolve = {});
-    route.resolve.app = ['$q', 'cheProfile', 'cheUser', ($q, cheProfile, cheUser) => {
+    route.resolve.app = ['$q', 'cheUser', 'chePreferences', ($q, cheUser, chePreferences) => {
       var deferred = $q.defer();
 
       cheUser.fetchUser().then(() => {
-        let profilePreferences = cheProfile.getPreferences();
+        let profilePreferences = chePreferences.getPreferences();
```
benoitf: please rename profilePreferences as it's not anymore in profile

dashboard/src/app/projects/list-projects/list-projects.controller.js
```diff
@@ -49,7 +49,7 @@ export class ListProjectsCtrl {
         }
       });
 
-    let profilePreferences = cheAPI.getProfile().getPreferences();
+    let profilePreferences = cheAPI.getPreferences().getPreferences();
```
benoitf: please rename profilePreferences as it's not anymore in profile

dashboard/src/app/workspaces/workspace-details/workspace-projects/workspace-details-projects.controller.js
```diff
@@ -27,7 +27,7 @@ export class WorkspaceDetailsProjectsCtrl {
     this.$mdMedia = $mdMedia;
     this.workspaceId = $route.current.params.workspaceId;
 
-    let profilePreferences = cheAPI.getProfile().getPreferences();
+    let profilePreferences = cheAPI.getPreferences().getPreferences();
```
benoitf: please rename profilePreferences as it's not anymore in profile

plugins/plugin-java/che-plugin-java-ext-jdt/org-eclipse-jdt-ui/src/main/java/org/eclipse/jdt/internal/core/JavaProject.java
```diff
@@ -1839,7 +1839,7 @@ public IEclipsePreferences getEclipsePreferences() {
         // Init project preferences
         IScopeContext context = new ProjectScope(getProject());
         final IEclipsePreferences eclipsePreferences = context.getNode(JavaCore.PLUGIN_ID);
-//        updatePreferences(eclipsePreferences);
+//        update(eclipsePreferences);
```
benoitf: ?

wsmaster/che-core-api-user-shared/src/main/java/org/eclipse/che/api/user/shared/dto/ProfileDto.java
```diff
@@ -21,32 +22,29 @@
  * @author andrew00x
  */
 @DTO
-public interface ProfileDescriptor {
+public interface ProfileDto extends Profile {
```
benoitf, line 31: missing some javadoc (DTO is used by clients developing plugins)

wsmaster/che-core-api-user-shared/src/main/java/org/eclipse/che/api/user/shared/dto/UserDto.java
```diff
@@ -10,56 +10,56 @@
...
 
 import java.util.List;
 
 /**
- * @author andrew00x
+ * @author Yevhenii Voevodin
  */
 @DTO
-public interface UserDescriptor {
+public interface UserDto extends User {
```
benoitf, line 30: missing javadoc (DTO is used by clients developing plugins)

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserManager.java
```diff
@@ -10,158 +10,204 @@
...
+                                           firstNonNull(newUser.getPassword(), generate("", PASSWORD_LENGTH)),
+                                           newUser.getAliases());
         try {
-            preferenceDao.setPreferences(user.getId(), preferences);
-        } catch (NotFoundException e) {
-            LOG.warn(format("Cannot set creation time preferences for user %s.", user.getId()), e);
+            userDao.create(user);
+            profileManager.create(new ProfileImpl(user.getId(), newUser.getEmail()));
+            preferencesManager.save(user.getId(), ImmutableMap.of("temporary", Boolean.toString(isTemporary),
+                                                                  "codenvy:created", Long.toString(currentTimeMillis())));
```
benoitf: why "codenvy" ?

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
```diff
@@ -58,293 +59,187 @@
...
+                                                           ServerException {
+        final User createdUser;
+        if (token == null) {
+            checkUser(userDto);
+            createdUser = userManager.create(userDto, isTemporary);
+        } else {
+            if (!userSelfCreationAllowed) {
+                throw new BadRequestException("User self creation is not allowed");
+            }
+            createdUser = userManager.create(fromToken(token), isTemporary);
```
benoitf: could the manager do all the work ?
keeping only createdUser = userManager.create(token, userDto, isTemporary) there
all checks performed by the manager including userSelfCreationAllowed

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
```diff
@@ -58,293 +59,187 @@
...
-                   @ApiResponse(code = 500, message = "Internal Server Error")})
-    public UserDescriptor getByAlias(@ApiParam(value = "User alias", required = true)
-                                     @QueryParam("alias")
-                                     @Required String alias) throws NotFoundException,
-                                                                    ServerException,
-                                                                    BadRequestException {
-        if (alias == null) {
-            throw new BadRequestException("Missed parameter alias");
+    @GenerateLink(rel = LINK_REL_USER)
+    @RolesAllowed({"user", "system/admin", "system/manager"})
```
benoitf: roles are still used and not permissions ?

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
```diff
@@ -391,84 +286,30 @@ public UserInRoleDescriptor inRole(@Required @Description("role inside a scope")
...
-     */
-    @GET
-    @Path("/settings")
-    @Produces(APPLICATION_JSON)
-    public Map<String, String> getSettings() {
-        return ImmutableMap.of(USER_SELF_CREATION_ALLOWED, Boolean.toString(userSelfCreationAllowed));
+    private User fromToken(String token) throws UnauthorizedException, ConflictException {
+        final String email = tokenValidator.validateToken(token);
+        final int atIdx = email.indexOf('@');
+        // Getting all the characters before '@' e.g. user@codenvy.com -> user
```
benoitf: user@eclipse.org :-)

benoitf: as I said previously the checks around token/user may not be handled by class managing token and users ?
or by using validators

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/model/impl/ProfileImpl.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+        final ProfileImpl that = (ProfileImpl)obj;
+        return Objects.equals(id, that.id)
+               && Objects.equals(email, that.email)
+               && getAttributes().equals(that.getAttributes());
+    }
+
+    @Override
+    public int hashCode() {
+        int hash = 7;
```
benoitf, line 75: java.util.Objects.hash(id, email, getAttributes()) ?

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/model/impl/UserImpl.java
```diff
@@ -75,48 +92,53 @@ public void setPassword(String password) {
...
     }
 
     @Override
     public int hashCode() {
         int hash = 7;
         hash = 31 * hash + Objects.hashCode(id);
         hash = 31 * hash + Objects.hashCode(email);
-        hash = 31 * hash + Objects.hashCode(password);
         hash = 31 * hash + Objects.hashCode(name);
+        hash = 31 * hash + Objects.hashCode(password);
```
benoitf, line 129: java.util.Objects.hash(Object...) ?

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/spi/ProfileDao.java
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * Creates user profile.
+     *
+     * @param profile
+     *         new profile
+     * @throws NullPointerException
+     *         when {@code profile} is null
+     * @throws ServerException
+     *         when any error occurs
+     */
```
benoitf, line 38: missing @throws ConflictException

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/spi/UserDao.java
```diff
@@ -0,0 +1,166 @@
...
+package org.eclipse.che.api.user.server.spi;
+
+import org.eclipse.che.api.core.ConflictException;
+import org.eclipse.che.api.core.NotFoundException;
+import org.eclipse.che.api.core.ServerException;
+import org.eclipse.che.api.core.UnauthorizedException;
+import org.eclipse.che.api.user.server.model.impl.UserImpl;
+
+/**
+ * Defies data access object contract for {@link UserImpl}.
```
benoitf: typo on Defi**N**es

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/PreferencesService.java
voievodin, line 104: it may be useful

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
voievodin: Yes, @sleshchenko will remove the roles processing  after this pr is merged.

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
```diff
@@ -391,84 +286,30 @@ public UserInRoleDescriptor inRole(@Required @Description("role inside a scope")
...
-     *
-     * @param password
-     *         user's password
-     * @throws BadRequestException
-     *         when password violates any rule
-     */
-    private void checkPassword(String password) throws BadRequestException {
+    private static void checkPassword(String password) throws BadRequestException {
         if (password == null) {
             throw new BadRequestException("Password required");
```
benoitf: not covered by test

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
voievodin: Well, i think it is not a good approach.
- `token` and `userDto` are mutual exclusive
-  we don't always have session user instance in the manager, while we always do in the service
  which excludes `userSelfCreationAllowed` check from the manager
- token and userDto should be valid, _bean valid_ and all the bean validations are performed in the service, i agree that we can extract this validation to the validators but, it is still service level
- Imho manager should keep clear and context independent operations which may be called from any place + contain all the business logic checks and operations, like `UserManager` creates profile along with a user when you call `.create(User user)`

What if in the other place we need a different parameter should manager be extended with a new create operation? I think no, while we can extend rest service with a new operations/parameters we should do so with manager

wsmaster/che-core-api-user/src/test/java/org/eclipse/che/api/user/server/UserManagerTest.java
```diff
@@ -10,58 +10,133 @@
  *******************************************************************************/
 package org.eclipse.che.api.user.server;
```
benoitf, line 11: missing lot of code coverage

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/TokenValidator.java
```diff
@@ -21,7 +21,7 @@
 public interface TokenValidator {
 
     /**
-     * Validates {@param token}
+     * Validates {@code token}.
```
benoitf, line 25: is this interface is used ?

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserManager.java
voievodin: In some historical reasons =)
@skabashnyuk do you have any thoughts?
i think we can remove `codenvy:` prefix and let only `created` part

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/ProfileManager.java
```diff
@@ -0,0 +1,115 @@
...
+ * easily reusable throughout the system.
+ *
+ * <p>The manager doesn't perform any bean validations and it
+ * is expected that all the incoming objects are valid, nevertheless
+ * this exactly the right place for performing business validations.
+ *
+ * @author Yevhenii Voevodin
+ */
+@Singleton
+public class ProfileManager {
```
benoitf, line 38: I don't see any test for this class

wsmaster/che-core-api-user/src/test/java/org/eclipse/che/api/user/server/PreferencesManagerTest.java
```diff
@@ -0,0 +1,96 @@
+/*******************************************************************************
```
benoitf: missing half of methods not tested

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/model/impl/ProfileImpl.java
voievodin, line 75: Is anything wrong with the current approach?

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/model/impl/UserImpl.java
```diff
@@ -8,52 +8,64 @@
  * Contributors:
  *   Codenvy, S.A. - initial API and implementation
  *******************************************************************************/
-package org.eclipse.che.api.user.server.dao;
+package org.eclipse.che.api.user.server.model.impl;
```
benoitf, line 11: missing tests on UserImpl (like equals)

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/model/impl/ProfileImpl.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+        return attributes;
+    }
+
+    public void setAttributes(Map<String, String> attributes) {
+        this.attributes = attributes;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
benoitf, line 62: it is not tested

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/user/Profile.java
voievodin: yes, it will be clearer i think

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
benoitf: about new parameter, if there was a class parameter it's quite easy to add new settings
userManager.create(UserCreateParameter)

About role of manager it's up to you. Either it is a dummy manager or smart manager.

so as it's not the role you want to give to the manager I think that it would be better to delegate to service then (not being the user manager)

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/model/impl/UserImpl.java
voievodin, line 11: haven't tested such cases before, but okay

dashboard/src/components/api/che-profile.spec.js
```diff
@@ -113,7 +111,7 @@ describe('CheProfile', function () {
       factory.setAttributes(testAttributes);
 
       // expecting a POST
-      httpBackend.expectPOST('/api/profile');
+      httpBackend.expecPUT('/api/profile/attributes');
```
sleshchenko: expec**T**?

dashboard/src/components/api/che-profile.spec.js
```diff
@@ -113,7 +111,7 @@ describe('CheProfile', function () {
       factory.setAttributes(testAttributes);
 
       // expecting a POST
```
benoitf: also here we do not expect a POST but a PUT

voievodin: Yes thank you  :+1: 
i forgot to change this useless comment

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserManager.java
skabashnyuk: Maybe we can. Are you sure it's not used? Also if we change this we need to migrate existed data.

sleshchenko: We had 'codenvy:' prefix to deny users edit system attributes. User should not be able to update created field. I am not sure that we need system attributes for now, but BTW we can't use just 'created'

benoitf: also do we have concept of "read-only" properties ? because created shouldn't be writeable

voievodin: @sleshchenko as far as i remember we had codenvy prefix check for workspace attributes but not for preferences [take a look](https://github.com/codenvy/platform-api/blob/master/codenvy-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java#L1036-L1040)

voievodin: > also do we have concept of "read-only" properties ? because created shouldn't be writeable

We don't have such a concept

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/model/impl/UserImpl.java
voievodin, line 11: @benoitf btw should i test setters and getters ?

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/spi/PreferenceDao.java
```diff
@@ -31,21 +30,19 @@
...
-    void setPreferences(String userId, Map<String, String> preferences) throws ServerException, NotFoundException;
+    void setPreferences(String userId, Map<String, String> preferences) throws ServerException;
 
     /**
      * Gets user preferences.
      *
      * <p>Note that this method must always return upgradable map, thus it may be used as:
      * <pre>{@code
-     *      Map<String, String> prefs = dao.getPreferences("user123");
+     *      Map<String, String> prefs = spi.getPreferences("user123");
```
sleshchenko: Is it error renaming of package? Mb it would be better to revert to dao here

voievodin: yes, thank you

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/model/impl/UserImpl.java
benoitf, line 11: it's always good even if usually we see quickly when we fail in these methods

but for example with copy/paste we could return another field for a given method or set the wrong field

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserManager.java
gazarenkov: 1/ I do not actually think we need to create User's preferences right here. I'd say creating Preferences should be on demand operation - i.e. if some preference is tried to be saved and there is no such a record - it should be silently created. We discussed it with @skabashnyuk  a a couple weeks ago I think. 
2/ We "probably" need to have the same read-only concept as for attributes but not sure, at all. Thoughts why we would need it?
3/ (related to 1) What would mean "temporary" Preference? Is it related to User?

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserManager.java
```diff
@@ -10,158 +10,204 @@
...
+            } catch (ConflictException | ServerException rollbackEx) {
+                LOG.error(format("An attempt to clean up resources due to user creation failure was unsuccessful." +
+                                 "Now the system may be in inconsistent state. " +
+                                 "User with id '%s' must not exist",
+                                 user.getId()),
+                          rollbackEx);
+            }
+            throw x;
         }
+        return user;
```
gazarenkov, line 115: Not sure I support data integrity operations on this level (I take Manager as business level entity) especially calling one Manager from another.
Could you at least replace  profileManager/preferencesManager.remove() with profileDao/preferencesDao.remove() ?

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserManager.java
```diff
@@ -10,158 +10,204 @@
...
      *         when any other error occurs
      */
-    public void remove(String id) throws NotFoundException, ServerException, ConflictException {
+    public void remove(String id) throws ServerException, ConflictException {
+        requireNonNull(id, "Required non-null id");
+        profileManager.remove(id);
+        preferencesManager.remove(id);
         userDao.remove(id);
     }
 }
```
gazarenkov, line 227: Not sure I support data integrity operations on this level (I take Manager as business level entity) especially calling one Manager from another.
Could you at least replace profileManager/preferencesManager.remove() with profileDao/preferencesDao.remove() ?

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
gazarenkov: I would probably move token/ userSelfCreationAllowed check to the Permission check filter.
@sleshchenko WDYT?

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserManager.java
voievodin: @gazarenkov i have nothing to say about the `temporary` preference as far as i know it was used to indicate whether user is temporary, because we don't provide such kind of information in user model, but again i am not sure about it. I think @skabashnyuk should know, + about the 1st statement, how will on-demand creation work for `codenvy:created`?

gazarenkov: Well, depending on _what_ codenvy:created 
I guess it is in fact _user created_ no?
We save it on Prefs b/c could not invent other place to store it?

So:
- if it is actually _user create_ May be better to discuss what we can do here to make it better since user != preferences
- if it is _preferences created_ then let it be created when actually Preferences created

@skabashnyuk  WDYT?

voievodin: It is 100% _user created_, i am okay with another place.
Profile lifecycle is the same to user's what do you think about using profile attributes for such purpose? Then we can set it directly in `ProfileManager`.

gazarenkov: yes, it sounds better

And I think writing this attribute should be actually a job of Data Storage layer, and not Manager.
At least because it should not try to rewrite this value if our system is adopted to existed user database (enterprise LDAP).
Correct?

voievodin: I didn't get the idea, the storage level is the same, i mean `LdapProfileDao` is the same for all the ldaps, like ours or customers, how do you want to determine whether this attribute should be stored or not?

gazarenkov: I do not know how to determine but it could be even physically impossible to write something on existed corporate LDAP, just because of policy (not OK for sys admins to let our app to modify corporate LDAP) so in some case it should be taken as normal situation.

skabashnyuk: >  I guess it is in fact user created no? 

AFAIK It is date and time when user was created. It should exists on all users

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/model/impl/ProfileImpl.java
benoitf, line 75: well it could remove many lines of code as it does the same stuff
http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/8-b132/java/util/Arrays.java#Arrays.hashCode%28java.lang.Object%5B%5D%29

voievodin, line 75: I don't like array creation actually, i think that such place as `hashCode` is a thin place, which should avoid unnecessary allocations. The implementation fits method contract and it is based on the approach described by _Josh Bloch in Effective Java 2nd ed item8_ (with some modifications according to Objects statis methods or java8 wrapper classes methods such as `Long.hashCode(..)`).

benoitf, line 75: I'm just pointing that there are alternatives since this book.

benoitf, line 75: here is an article with Java7 features
http://baptiste-wicht.com/posts/2010/04/java-7-the-new-java-util-objects-class.html

voievodin:
I added tests and fixes.
As _created:created_ preference moving to attributes requires migration i decided to create a separate [issue](https://jira.codenvycorp.com/browse/CHE-1270) for that 

@skabashnyuk, @benoitf are you okay with the changes?

benoitf:
@evoevodin ok for the changes, thx! 
reviewing code, it seems also that now ServiceClient methods should use Promise instead of AsyncRequestCallback

 void createUser(@NotNull String token, boolean isTemporary, AsyncRequestCallback<UserDto> callback);
--> Promise<UserDto> createuser(@NotNull String token, boolean isTemporary)

skabashnyuk:
ok

PR #1394, opened 2016-05-31
CHE-1159: Add to Jgit implementation ability to view cloning process

vinokurig:
This will show process of cloning project in imprt project notification
@skabashnyuk please review

### DIFF
--- a/wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
+++ b/wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
@@ -22,6 +22,7 @@ import org.apache.commons.io.filefilter.TrueFileFilter;
 import org.eclipse.che.api.core.ErrorCodes;
 import org.eclipse.che.api.core.ServerException;
 import org.eclipse.che.api.core.UnauthorizedException;
+import org.eclipse.che.api.core.util.LineConsumer;
 import org.eclipse.che.api.core.util.LineConsumerFactory;
 import org.eclipse.che.api.git.Config;
 import org.eclipse.che.api.git.CredentialsLoader;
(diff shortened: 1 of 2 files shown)


wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -439,6 +442,30 @@ public void clone(CloneRequest request) throws GitException, UnauthorizedExcepti
                 cloneCommand.setBranchesToClone(request.getBranchesToFetch());
             }
 
+            LineConsumer lineConsumer = lineConsumerFactory.newLineConsumer();
+            cloneCommand.setProgressMonitor(new BatchingProgressMonitor() {
+                @Override
+                protected void onUpdate(String taskName, int workCurr) {
```
sleshchenko, line 467: when this method should be invoked?

vinokurig, line 467: This method invokes when the total work isn't known. I've added an output inside this method too

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -439,6 +442,35 @@ public void clone(CloneRequest request) throws GitException, UnauthorizedExcepti
...
+                    try {
+                        lineConsumer.writeLine(taskName + ": " + workCurr + " completed");
```
benoitf, line 469: where is the test that check this progress monitor ?

vinokurig, line 469: We have one TCK test suite for both Git implementations. I think it is not good to test there this progress monitor because it is a part of implementation and it doesn't perform any git operations.

benoitf, line 469: so you mean we should let untested code ? or you want to test it elsewhere or to move out the progress monitor out of this implementation along the test of the progress monitor ?

vinokurig, line 469: I've added `CloneTest` that checks `lineConsumer.writeLine`,  in JGit implementation it called from ProgressMonitor

skabashnyuk:
ok

PR #1395, opened 2016-05-31
CHE-1219: add heartbeat messages to keep websocket connection alive

ashumilova:

### DIFF
--- a/dashboard/src/components/api/che-websocket.factory.js
+++ b/dashboard/src/components/api/che-websocket.factory.js
@@ -22,9 +22,10 @@ export class CheWebsocket {
    * Default constructor that is using resource
    * @ngInject for Dependency injection
    */
-  constructor ($websocket, $location, proxySettings, userDashboardConfig) {
+  constructor ($websocket, $location, $interval, proxySettings, userDashboardConfig) {
 
     this.$websocket = $websocket;
+    this.$interval = $interval;
 
     var wsUrl;
     let inDevMode = userDashboardConfig.developmentMode;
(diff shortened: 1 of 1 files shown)


dashboard/src/components/api/che-websocket.factory.js
```diff
@@ -22,9 +22,10 @@ export class CheWebsocket {
    * Default constructor that is using resource
    * @ngInject for Dependency injection
    */
-  constructor ($websocket, $location, proxySettings, userDashboardConfig) {
+  constructor ($websocket, $location, $interval, proxySettings, userDashboardConfig) {
```
garagatyi, line 25: I think it is better to explain what `interval` means. You can rename it to pingInterval

dashboard/src/components/api/che-websocket.factory.js
```diff
@@ -145,15 +156,45 @@ class MessageBuilder {
 
 class MessageBus { // jshint ignore:line
 
-  constructor(datastream) {
+  constructor(datastream, $interval) {
```
garagatyi, line 159: I think it is better to explain what interval means. You can rename it to pingInterval

dashboard/src/components/api/che-websocket.factory.js
ashumilova, line 25: it's AngularJs common wrapper for JS Window.setInterval ()

benoitf, line 25: yes these are angularjs variables

dashboard/src/components/api/che-websocket.factory.js
benoitf, line 159: it can't https://docs.angularjs.org/api/ng/service/$interval

PR #1396, opened 2016-05-31
CHE-1058: Debugger connection failure: Indicate reasons

tolusha:


tolusha:
@vparfonov

vparfonov:
ok

plugins/plugin-gdb/che-plugin-gdb-server/src/main/java/org/eclipse/che/plugin/gdb/server/GdbDebugger.java
```diff
@@ -110,8 +112,16 @@ public static GdbDebugger newInstance(String host,
                                           String file,
                                           String srcDirectory,
                                           DebuggerCallback debuggerCallback) throws DebuggerException {
-        Gdb gdb;
+        Gdb gdb = null;
         try {
+            if (!exists(Paths.get(file))) {
```
benoitf, line 117: do we need to throw IOException and to check path in the try { block ?
we could just throw DebuggerException before instantiating Gdb (and not throw IOException to be catched after to be rethrown in DebuggerException and try to stop gdb as obviously it is not started in that case

PR #1397, opened 2016-05-31
CODENVY-584: Add terminal link into each machine instance

akorneta:
Since the websocket terminal link now used for each machine separately, so it is nessesary to add the formation mechanism for these links like we did for the dev machine.
@skabashnyuk, @garagatyi

### DIFF
--- a/wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineService.java
+++ b/wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineService.java
@@ -16,7 +16,6 @@ import io.swagger.annotations.ApiParam;
 import io.swagger.annotations.ApiResponse;
 import io.swagger.annotations.ApiResponses;
 
-import com.google.common.collect.Lists;
 import com.google.common.io.CharStreams;
 
 import org.eclipse.che.api.core.BadRequestException;
@@ -25,15 +24,10 @@ import org.eclipse.che.api.core.NotFoundException;
 import org.eclipse.che.api.core.ServerException;
 import org.eclipse.che.api.core.model.machine.Machine;
 import org.eclipse.che.api.core.rest.Service;
-import org.eclipse.che.api.core.rest.ServiceContext;
-import org.eclipse.che.api.core.rest.shared.dto.Link;
-import org.eclipse.che.api.core.rest.shared.dto.LinkParameter;
 import org.eclipse.che.api.machine.server.exception.MachineException;
 import org.eclipse.che.api.machine.server.model.impl.SnapshotImpl;
 import org.eclipse.che.api.machine.server.spi.Instance;
-import org.eclipse.che.api.machine.shared.Constants;
 import org.eclipse.che.api.machine.shared.dto.CommandDto;
-import org.eclipse.che.api.machine.shared.dto.MachineConfigDto;
 import org.eclipse.che.api.machine.shared.dto.MachineDto;
 import org.eclipse.che.api.machine.shared.dto.MachineProcessDto;
 import org.eclipse.che.api.machine.shared.dto.NewSnapshotDescriptor;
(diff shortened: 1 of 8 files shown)


garagatyi:
LGTM

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineService.java
```diff
@@ -82,9 +64,12 @@
 public class MachineService extends Service {
     private MachineManager machineManager;
 
+    private final MachineServiceLinksInjector linksInjector;
+
     @Inject
-    public MachineService(MachineManager machineManager) {
```
benoitf, line 86: I don't see related tests of MachineService

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineServiceLinksInjector.java
```diff
@@ -0,0 +1,215 @@
...
+                             Constants.LINK_REL_GET_PROCESSES));
+        final URI getLogsUri = uriBuilder.clone()
+                                         .path(MachineService.class, "getMachineLogs")
+                                         .build(machine.getId());
+        links.add(createLink(HttpMethod.GET, getLogsUri.toString(), TEXT_PLAIN, Constants.LINK_REL_GET_MACHINE_LOGS));
+
+        injectTerminalLink(machine, serviceContext, links)
+        ;
+
+        // add links to websocket channels
```
benoitf, line 121: where these links are tested ?

skabashnyuk:
ok

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineServiceLinksInjector.java
```diff
@@ -0,0 +1,215 @@
...
+                                       .toString(),
+                             APPLICATION_JSON,
+                             Constants.LINK_REL_GET_PROCESSES));
+        final URI getLogsUri = uriBuilder.clone()
+                                         .path(MachineService.class, "getMachineLogs")
+                                         .build(machine.getId());
+        links.add(createLink(HttpMethod.GET, getLogsUri.toString(), TEXT_PLAIN, Constants.LINK_REL_GET_MACHINE_LOGS));
+
+        injectTerminalLink(machine, serviceContext, links)
+        ;
```
benoitf: ; belongs to the upper line

benoitf:
how can it be approved without tests ? while some PR are required to provide tests while modifying REST service ?

garagatyi:
AFAIK this PR doesn't change MachineService behavior at all - just move links evaluation to another class.
So in this case I don't ask @akorneta to add tests but would appreciate if he added them.
There are no tests for this class at all. It is my fault. When this class was created we were at this stage of new architecture that it was just an experiment and was not obvious whether we use that or not. Later we  accepted this code but I had no time to add tests. Sorry guys!

benoitf:
rules are pretty simple: moving code is introducing changes (also this is what is being said on my PR)
MachineServiceLinksInjector is a new class so it needs to be tested.
MachineService is modified --> what is being changed needs to be tested.

garagatyi:
Be honest you've changed code, not just moved. There was no regexp in origin DockerInstanceKey

garagatyi:
I'm not against adding tests, but looks like you really think that we use double standarts in PR reviewing. It is not true.

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineService.java
akorneta, line 86: I add tests for MachineServiceLinksInjector

benoitf:
btw thanks @akorneta and without the tests it would have introduced wrong links, as during the move, getClass() was not replaced by MachineService.class

PR #1402, opened 2016-05-31
CODENVY-590 Add factory editing ownership check

mkuznyetsov:
Allow factory updating only if user id is matched to author id.

### DIFF
--- a/wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/impl/FactoryEditValidatorImpl.java
+++ b/wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/impl/FactoryEditValidatorImpl.java
@@ -41,12 +41,16 @@ public class FactoryEditValidatorImpl implements FactoryEditValidator {
      */
     @Override
     public void validate(Factory factory) throws ForbiddenException, ServerException {
-        // ensure user has the correct permissions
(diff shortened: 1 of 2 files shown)


skabashnyuk:
ok @sleshchenko ?

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/impl/FactoryEditValidatorImpl.java
```diff
@@ -41,13 +41,17 @@
...
-        final String userId = EnvironmentContext.getCurrent().getSubject().getUserId();
         // Checks if there is an author from the factory (It may be missing for some old factories)
         Author author = factory.getCreator();
         if (author == null || author.getUserId() == null) {
             throw new ServerException(format("Invalid factory without author stored. Please contact the support about the factory ID '%s'",
                                              factory.getId()));
         }
+        // ensure user has the correct permissions
+        final String userId = EnvironmentContext.getCurrent().getSubject().getUserId();
+        if (userId == null || !userId.equals(author.getUserId())) {
```
benoitf: as author is not null and author.getUserId() not null due to previous block check
it can be simplified to 
if (!author.getUserId().equals(userId)) {

mkuznyetsov: ok

PR #1414, opened 2016-06-02
[Issue #1413] register multiple callbacks (fixes #1413)

### LINKED ISSUE #1413: Callbacks send to WsAgentStateController#getMessageBus ignored
Callbacks send are assigned to a field in the singleton. On multiple calls previous handlers get overwritten and never called. I will provide a PR in sec.

svenefftinge:


core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/WsAgentStateController.java
```diff
@@ -132,12 +132,16 @@ private void started() {
...
         loader.hide();
 
-        if (messageBusCallback != null) {
-            messageBusCallback.onSuccess(messageBus);
-        }
-        if (devMachineCallback != null) {
-            devMachineCallback.onSuccess(devMachine);
+        for (AsyncCallback<MessageBus> callback : messageBusCallbacks) {
+        	callback.onSuccess(messageBus);
+		}
```
benoitf, line 137: formatting issue

core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/WsAgentStateController.java
```diff
@@ -132,12 +132,16 @@ private void started() {
...
+        messageBusCallbacks = null;
```
benoitf, line 138: should we only clear list ?

core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/WsAgentStateController.java
```diff
@@ -132,12 +132,16 @@ private void started() {
...
+        
+        for (AsyncCallback<DevMachine> callback : devMachineCallbacks) {
+        	callback.onSuccess(devMachine);
         }
+        devMachineCallbacks = null;
```
benoitf, line 143: should't we only clear list ?

core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/WsAgentStateController.java
```diff
@@ -39,14 +39,14 @@
...
-import static org.eclipse.che.ide.ui.loaders.initialization.OperationInfo.Status.IN_PROGRESS;
-import static org.eclipse.che.ide.ui.loaders.initialization.OperationInfo.Status.SUCCESS;
+import com.google.gwt.json.client.JSONArray;
+import com.google.gwt.json.client.JSONObject;
+import com.google.gwt.json.client.JSONParser;
+import com.google.gwt.user.client.Timer;
+import com.google.gwt.user.client.rpc.AsyncCallback;
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+import com.google.web.bindery.event.shared.EventBus;
```
benoitf, line 49: it seems there is a fight in import order 
(static are last while here static import are first)

core/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/WsAgentStateController.java
svenefftinge, line 138: It should not be used multiple times. So I set it to null.

benoitf, line 138: with null, you're sure that WsAgentStateController.this.messageBusCallbacks.add(callback); can't throw any NPE ?

svenefftinge, line 138: It should. 'null' is the protocol for "no more callbacks". In case there is a bug elsewhere we will have an NPW here. Much better than silently ignoring callbacks which will never be called.

benoitf, line 138: well then before calling
 WsAgentStateController.this.messageBusCallbacks.add(callback); it means you've to introduce a check with an invalid state error report as npe may not indicate the correct line as it's transformed into javascript

PR #1415, opened 2016-06-02
CHE-1275: Add api endpoint into each non dev machine environment

akorneta:
Since the terminal is not only dev machine feature so we need api edpoint env variable for non dev machines too.
@skabashnyuk, @garagatyi please review.


garagatyi:
LGTM

skabashnyuk:
ok

PR #1417, opened 2016-06-03
CODENVY-566 make workspace export working again

benoitf:
- It has been broken when MachineServlet proxy has been removed as cors has been removed
- it has been broken when machine token was required but UD code was not updated


skabashnyuk:
@evoevodin @mshaposhnik  wdyt about such cors configuration?

mshaposhnik:
"*" is a worst thing, but if we can't avoid it for now - ok.

benoitf:
for sure, the only way to remove it is if "exporting a workspace + create remote projects" is moved to server side and not kept client side else I don't see how it is possible.

skabashnyuk:
ок

vparfonov:
ok

PR #1418, opened 2016-06-03
CHE-1196: Artik: Debugger hangs when debugging binaries using Artik SDK

tolusha:
- Fix synchronization issue
- Respect gdb termination
- Fix UI bugs

### DIFF
--- a/plugins/plugin-gdb/che-plugin-gdb-server/src/main/java/org/eclipse/che/plugin/gdb/server/Gdb.java
+++ b/plugins/plugin-gdb/che-plugin-gdb-server/src/main/java/org/eclipse/che/plugin/gdb/server/Gdb.java
@@ -12,19 +12,21 @@ package org.eclipse.che.plugin.gdb.server;
 
 import org.eclipse.che.api.debugger.server.exceptions.DebuggerException;
 import org.eclipse.che.commons.annotation.Nullable;
+import org.eclipse.che.plugin.gdb.server.exception.GdbException;
+import org.eclipse.che.plugin.gdb.server.exception.GdbTerminatedException;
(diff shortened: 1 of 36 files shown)


tolusha:
@vparfonov 
@sleshchenko 
@benoitf 
pls review

plugins/plugin-gdb/che-plugin-gdb-server/src/main/java/org/eclipse/che/plugin/gdb/server/Gdb.java
```diff
@@ -142,129 +147,155 @@ public GdbInfoLine next() throws IOException, InterruptedException, GdbParseExce
...
 
     /**
      * `info locals` command.
      */
-    public GdbInfoLocals infoLocals() throws IOException, InterruptedException, GdbParseException {
-        sendCommand("info locals");
-        return GdbInfoLocals.parse(outputs.take());
+    public GdbInfoLocals infoLocals() throws IOException, InterruptedException, DebuggerException {
+        GdbOutput gdbOutput = sendCommand("info locals");
+        return GdbInfoLocals.parse(gdbOutput);
```
sleshchenko, line 254: WDYT about inlining these methods as

```
return GdbInfoLocals.parse(sendCommand("info locals"));
```

?

sleshchenko:
LGTM

plugins/plugin-gdb/che-plugin-gdb-server/src/main/java/org/eclipse/che/plugin/gdb/server/Gdb.java
tolusha, line 254: I personally prefer splitting.
In case of debugging it will help.

vparfonov:
ok

PR #1430, opened 2016-06-06
CHE-394: Add ability to specify branch when cloning from GitHub

vinokurig:
@vparfonov is it ok to merge when 4.4 will be released?


vparfonov:
ok

PR #1383, opened 2016-05-30
Resource management

vzhukovs:
Provided the new client side resource management which is responsible for loading, processing resources (files, folders and projects).
## New and noteworthy

There is a proposal to create an API to access and operate with files and folders. Benefits from this dedicated API will be isolating and separating the project explorer and creating dedicated mechanism which would be used by all core components and third party extensions.
#### Proposed API

There is a base interface `org.eclipse.che.ide.api.resources.Resource` which represents a certain resource. It may be a file, folder or project.
On higher abstraction there is a `org.eclipse.che.ide.api.resources.Container` interface which can contains more than one resource.
Folder `org.eclipse.che.ide.api.resources.Folder` extends `org.eclipse.che.ide.api.resources.Container`.
Project `org.eclipse.che.ide.api.resources.Project` also extends `org.eclipse.che.ide.api.resources.Container`, and `org.eclipse.che.api.core.model.project.ProjectConfig`
File `org.eclipse.che.ide.api.resources.File` extends `org.eclipse.che.ide.api.resources.Resource` and `org.eclipse.che.ide.api.resources.VirtualFile`

Each operation on resources produce event into general event bus: `org.eclipse.che.ide.api.resources.ResourceChangedEvent`. Event provides delta which describe a part of change which applied to the specific resource.

There is a simple snippet of code which describes how to handle resource changed event:

```
      eventBus.addHandler(ResourceChangedEvent.getType(), new ResourceChangedHandler{
          void onResourceChanged(ResourceChangedEvent event) {
              ResourceDelta delta = event.getDelta();
              Resource resource = delta.getResource();

              if (delta.getKind() == ResourceDelta.ADDED) {
                  //handle resource creation
              }

              if (delta.getKind() == ResourceDelta.REMOVED) {
                  //handle resource removal
              }

              if (delta.getKind() == ResourceDelta.UPDATED) {
                  //handle resource update
              }
          }
      });
```

At this moment we separate project explorer from injecting into other components and later it would be moved into dedicated plugin.

There is method in application context which allow to access the workspace root: 
`org.eclipse.che.ide.api.app.AppContext#getWorkspaceRoot`

Also there are three methods:
`org.eclipse.che.ide.api.app.AppContext#getResource` - retrieves current resource from the context;
`org.eclipse.che.ide.api.app.AppContext#getResources` - retrieves current resources from the context (if more than one provided);
`org.eclipse.che.ide.api.app.AppContext#getRootProject` - retrieves the root project from the current provided resource;
[...]

PR #1437, opened 2016-06-06
Add utility class to operate with arrays

vzhukovs:
There is a proposal to add an utility class which helps easily operate with arrays.

This PR is part of https://github.com/eclipse/che/pull/1383.

@vparfonov @skabashnyuk please, review it.

Related issue: CHE-907


vparfonov:
ok

PR #1440, opened 2016-06-06
CHE-1113: Add ability to clone project with submodules

vinokurig:
import-recursively
@vparfonov is it ok to merge when 4.4 will be released?


wsagent/che-core-api-git-shared/src/main/java/org/eclipse/che/api/git/shared/CloneRequest.java
```diff
@@ -49,6 +49,13 @@
     
     CloneRequest withRemoteName(String remoteName);
 
+    /** @return true if 'Recursive' parameter enabled */
+    boolean getRecursiveEnabled();
```
sleshchenko: As for me name is a little bit strange for DTO. Maybe something like isRecursive would be better

vinokurig: done

vparfonov:
ok

PR #1441, opened 2016-06-06
CODENVY-583: Fix javadoc navigation in https mode

akorneta:
Moved JavadocService  from dynamodule for overriding in codenvy assembly.
@skabashnyuk, @vparfonov


vparfonov:
ok

skabashnyuk:
ok

PR #1443, opened 2016-06-06
Add utility class to operate with arrays

vzhukovs:
There is a proposal to add an utility class which helps easily operate with arrays.

This PR is part of #1383.

@vparfonov @skabashnyuk please, review it.

Related issue: CHE-907

### DIFF
--- a/core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/util/Arrays.java
+++ b/core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/util/Arrays.java
@@ -1,0 +1,174 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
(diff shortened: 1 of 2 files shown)


core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/util/Arrays.java
```diff
@@ -0,0 +1,175 @@
...
+     * @throws IllegalArgumentException
+     *         in case if given {@code array} is null
+     * @since 4.3.0
+     */
+    public static <T> int indexOf(T[] array, T o) {
+        checkArgument(array != null, "Input array is null");
+
+        for (int index = 0; index < array.length; index++) {
+            if (array[index] == o) {
+                return index;
```
skabashnyuk: can you make one statement instead of two?

skabashnyuk:
ok

core/commons/che-core-commons-gwt/src/main/java/org/eclipse/che/ide/util/Arrays.java
vzhukovs: Comparison has changed to use with guava method `com.google.common.base.Objects#equal`

vparfonov:
ok

PR #1444, opened 2016-06-06
CODENVY-610: add possibility to set no_proxy ENV var in machine

riuvshin:
**_1 Upvote**_ Signed-off-by: Roman Iuvshin riuvshin@codenvy.com
@garagatyi @skabashnyuk  please review this.


TylerJewell:
+1

garagatyi:
LGTM

skabashnyuk:
ок

riuvshin:
:bow:

TylerJewell:
@riuvshin - Is it possible to configure different no_proxy options for Codenvy installation vs. Codenvy machine / workspace?

I know that we have the --no-proxy option.  And the workspace environment variable no_proxy should (by default) be set to the value of --no-proxy. However, if a customer wants to install with a no-proxy, but have a different no_proxy for machines, is that possible?

PR #1453, opened 2016-06-07
CHE-1168: determine source folders with generated artifacts

svor:
Determine folders with generated artifacts and add them to the project attributes as source/test folders when project is resolving


vparfonov:
ok

PR #1454, opened 2016-06-07
CHE-1180: add ability to configure private docker registries

olexii4:
@ashumilova WDYT?

in profilePreferences

{dockerCredentials: credentialsBase64}

credentials[registryUrl] = {
      name: userName,
      email: userEmail,
      password: userPassword
};

selection_015
selection_013
selection_012

### DIFF
--- a/dashboard/src/app/navbar/navbar.html
+++ b/dashboard/src/app/navbar/navbar.html
@@ -32,7 +32,9 @@
               </md-button>
             </md-list-item>
             <md-list-item flex class="navbar-subsection-item">
-              <md-button nav-bar-selected flex che-reload-href che-on-long-touch="navbarCtrl.openLinkInNewTab('{{navbarCtrl.menuItemUrl.workspaces}}')" ng-href="{{navbarCtrl.menuItemUrl.workspaces}}" layout-align="left">
+              <md-button nav-bar-selected flex che-reload-href
+                         che-on-long-touch="navbarCtrl.openLinkInNewTab('{{navbarCtrl.menuItemUrl.workspaces}}')"
+                         ng-href="{{navbarCtrl.menuItemUrl.workspaces}}" layout-align="left">
                 <div class="navbar-item" layout="row" layout-align="start center">
                   <md-icon md-font-icon="navbar-icon chefont cheico-workspace"></md-icon>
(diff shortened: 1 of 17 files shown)


TylerJewell:
It should be "Docker Registries" -- missing capitalization.

TylerJewell:
I am not understanding the user scenario.  
1. If a user has many registries, how does the system know which registry to push / pull from?
2. What is the trigger in our system that causes a "docker login <user> <pass>"?

mmorhun:
Sorry, I accidentally closed it

mmorhun:
Hi Tyler
1) So, for now we have three operations which may require authorization: pull, push and build.
- push is used for pushing workspace's snapshots only. Registry to which snapshots will be pushed is defined in properties.
- pull. When user create workspace from custom image he references it in the Docker recipe:
  `FROM myregistry:5000/myRepo`
  If this referenced repo is a private one, a user will then have to save credentials to user prefs in User Dashboard. It is also possible to save registry configuration in che.properties.

2) System do not perform login at all. It sends credentials on each query if it needed (i.e. login for one query).

mmorhun:
@olexii4 please use `AuthConfig` dto format for credentials

olexii4:
selection_007

dashboard/src/app/navbar/navbar.html
```diff
@@ -50,6 +54,17 @@
...
               </md-button>
             </md-list-item>
+            <md-list-item flex class="navbar-subsection-item">
+              <md-button nav-bar-selected flex che-reload-href
+                         che-on-long-touch="navbarCtrl.openLinkInNewTab('{{navbarCtrl.menuItemUrl.administration}}')"
+                         ng-href="{{navbarCtrl.menuItemUrl.administration}}" layout-align="left">
+                <div class="navbar-item" layout="row" layout-align="start center">
+                  <md-icon md-font-icon="navbar-icon material-design icon-ic_settings_24px"></md-icon>
+                  <span>Administration</span>
+                  <span class="navbar-number" ng-show="navbarCtrl.getProjectsNumber()">&nbsp;({{navbarCtrl.getProjectsNumber()}})</span>
```
akurinnoy: it seems copy-paste error

PR #1460, opened 2016-06-08
CHE-459 Associate a set of commands with provided stacks

ddementieva:
@eivantsov @ashumilova please, take a look

### DIFF
--- a/core/ide/che-core-ide-stacks/src/main/resources/stacks.json
+++ b/core/ide/che-core-ide-stacks/src/main/resources/stacks.json
@@ -59,7 +59,7 @@
       "commands": [
         {
           "commandLine": "mvn clean install -f ${current.project.path}",
-          "name": "newMaven",
+          "name": "build",
           "type": "mvn"
         }
       ]
@@ -184,7 +184,14 @@
       ],
       "name": "default",
       "defaultEnv": "default",
-      "description": null
+      "description": null,
+      "commands": [
+        {
+          "commandLine": "mvn clean install -f ${current.project.path}",
+          "name": "build",
+          "type": "mvn"
+        }
+      ]
     },
     "acl": [
       {
@@ -253,7 +260,14 @@
       ],
       "name": "default",
       "defaultEnv": "default",
-      "description": null
+      "description": null,
+      "commands": [
+        {
+          "name": "run",
+          "type": "custom",
+          "commandLine": "cd ${current.project.path} && make && ./a.out"
+        }
+      ]
     },
     "acl": [
       {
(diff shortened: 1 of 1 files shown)


core/ide/che-core-ide-stacks/src/main/resources/stacks.json
```diff
@@ -184,7 +184,14 @@
       ],
       "name": "default",
       "defaultEnv": "default",
-      "description": null
+      "description": null,
+      "commands": [
+        {
+          "commandLine": "mvn clean install -f ${current.project.path}",
+          "name": "newMaven",
```
benoitf: Name shouldn't be more descriptive ? 
Like 
mvn clean install
Or
Build current project

core/ide/che-core-ide-stacks/src/main/resources/stacks.json
```diff
@@ -402,7 +434,17 @@
       ],
       "name": "default",
       "defaultEnv": "default",
-      "description": null
+      "description": null,
+      "commands": [
+        {
+          "name": "run",
```
benoitf: I'm not sure that this command is useful for every projects as name of app is different and we may use other tools like grunt gulp etc

core/ide/che-core-ide-stacks/src/main/resources/stacks.json
ghost: @benoitf we use short command names so that they are nicely displayed in the widget. But agree, newMaven isn't good.

core/ide/che-core-ide-stacks/src/main/resources/stacks.json
```diff
@@ -402,7 +434,17 @@
       ],
       "name": "default",
       "defaultEnv": "default",
-      "description": null
+      "description": null,
```
benoitf: I'm not sure that this command is useful for every projects as name of app is different and we may use other tools like grunt gulp etc

core/ide/che-core-ide-stacks/src/main/resources/stacks.json
```diff
@@ -545,7 +613,17 @@
...
+      "commands": [
+        {
+          "name": "run",
```
benoitf, line 609: I'm not sure this command is generic

ghost, line 609: @benoitf your suggestion of a generic python command please?

core/ide/che-core-ide-stacks/src/main/resources/stacks.json
ghost: @benoitf yes, it is not easy to find a command that works for all users. But at least this is smth to start with and the command can be edited if needed. I think it's better to remove it at all for node as there can be plenty of options here.

core/ide/che-core-ide-stacks/src/main/resources/stacks.json
benoitf, line 609: I would say there is nothing generic (except calling only make)

benoitf, line 609: there is no generic python command for me

PR #1463, opened 2016-06-08
CODENVY-27; allow to open workspaces by using /ide/namespace/ws_name schema

mshaposhnik:


assembly/assembly-ide-war/src/main/java/org/eclipse/che/DashboardRedirectionFilter.java
```diff
@@ -0,0 +1,48 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che;
+
+
+import javax.servlet.*;
```
sleshchenko: We should not use \* in imports

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -752,6 +750,10 @@ private static String getCurrentUserId() {
         return EnvironmentContext.getCurrent().getSubject().getUserId();
     }
 
+    private static String getCurrentNamespace() {
```
skabashnyuk: I would say this method should be inline

wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/workspace/server/WorkspaceManagerTest.java
```diff
@@ -120,7 +121,7 @@ public void setUp() throws Exception {
         EnvironmentContext.setCurrent(new EnvironmentContext() {
             @Override
             public Subject getSubject() {
-                return new SubjectImpl("Test User", USER_ID, "token", new ArrayList<>(), false);
+                return new SubjectImpl(NAMESPACE, USER_ID, "token", new ArrayList<>(), false);
```
benoitf: do we need arraylist or an empty list ?

mshaposhnik: Do you imply Collections.emptyList() here ?

mshaposhnik: We can simply pass null and new emptySet ill be created

skabashnyuk:
Other ok

wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/workspace/server/WorkspaceManagerTest.java
benoitf: Ok

vparfonov:
ok

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -247,7 +245,7 @@ public void delete(@ApiParam("The workspace id") @PathParam("id") String id) thr
                                                                                         ConflictException,
                                                                                         ForbiddenException {
         if (!workspaceManager.getSnapshot(id).isEmpty()) {
-            machineManager.removeSnapshots(getCurrentUserId(), id);
+            machineManager.removeSnapshots(EnvironmentContext.getCurrent().getSubject().getUserId(), id);
```
sleshchenko, line 248: https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/dao/SnapshotDao.java#L78-78
namespace should be equal to user name, no?

PR #1464, opened 2016-06-08
CHE-1239 Smooth transition when restarting an existing workspace

vitaliy-guliy:


ashumilova:
+1

PR #1471, opened 2016-06-08
CHE-1245: Update docker client to use new format of X-Registry-Config header

mmorhun:


plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -1406,8 +1422,8 @@ private String commit(final CommitParams params, URI dockerDaemonUri) throws IOE
         // TODO: add option to pause container
         try (DockerConnection connection = connectionFactory.openConnection(dockerDaemonUri)
                                                             .method("POST")
-                                                            .path("/commit")
-                                                            .query("container", params.getContainer())) {
+                                                            .path("/commit")) {
+            addQueryParamIfNotNull(connection, "container", params.getContainer());
```
garagatyi: Commit is mandatory parameter. And it should not be null. Why we need this check?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -1477,14 +1493,17 @@ public void pull(final PullParams params, final ProgressMonitor progressMonitor)
...
-        final String image = params.getImage();
         final String registry = params.getRegistry();
+        final AuthConfigs authConfigs = params.getAuthConfigs();
 
         try (DockerConnection connection = connectionFactory.openConnection(dockerDaemonUri)
                                                             .method("POST")
                                                             .path("/images/create")
-                                                            .query("fromImage", registry != null ? registry + '/' + image : image)
-                                                            .header("X-Registry-Auth", initialAuthConfig.getAuthConfigHeader())) {
+                                                            .query("fromImage", params.getImageFqn())
```
garagatyi: Where you have found such a term as docker image FQN?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerRegistryAuthManager.java
```diff
@@ -0,0 +1,170 @@
...
+import org.apache.commons.codec.binary.Base64;
+import org.eclipse.che.commons.annotation.Nullable;
+import org.eclipse.che.commons.json.JsonHelper;
+import org.eclipse.che.plugin.docker.client.dto.AuthConfig;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Class for managing
```
garagatyi: Please improve description of this class

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -114,18 +117,20 @@
 @Singleton
 public class DockerConnector {
     private static final Logger LOG = LoggerFactory.getLogger(DockerConnector.class);
+    private static final String DEFAULT_REGISTRY = "";
```
garagatyi: Why we need default registry here and in auth manager?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -1477,14 +1493,17 @@ public void pull(final PullParams params, final ProgressMonitor progressMonitor)
...
         try (DockerConnection connection = connectionFactory.openConnection(dockerDaemonUri)
                                                             .method("POST")
                                                             .path("/images/create")
-                                                            .query("fromImage", registry != null ? registry + '/' + image : image)
-                                                            .header("X-Registry-Auth", initialAuthConfig.getAuthConfigHeader())) {
+                                                            .query("fromImage", params.getImageFqn())
+                                                            .header("X-Registry-Auth",
+                                                                    authManager.getXRegistryAuthHeaderValue(
+                                                                            registry != null ? registry : DEFAULT_REGISTRY,
+                                                                            authConfigs != null ? authConfigs.getConfigs() : null))) {
```
garagatyi: Why not pass AuthConfigs object and let auth manager do everything else?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -1133,10 +1141,14 @@ private String buildImage(final BuildImageParams params,
                                                             .header("Content-Type", "application/x-compressed-tar")
                                                             .header("Content-Length", tar.length())
                                                             .header("X-Registry-Config",
-                                                                    Base64.encodeBase64String(JsonHelper.toJson(authConfigs)
-                                                                                                        .getBytes()))
+                                                                    authManager.getXRegistryConfigHeaderValue(
```
garagatyi: Why not pass AuthConfigs object and let auth manager do everything else?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/PullParams.java
```diff
@@ -25,9 +26,11 @@
  */
 public class PullParams {
 
-    private String image;
-    private String tag;
-    private String registry;
+    private String      image;
+    private String      tag;
+    private String      registry;
+    private String      namespace;
```
garagatyi: What is namespace?

garagatyi: Where you have found it in docker and why you need it?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/PullParams.java
```diff
@@ -96,19 +124,49 @@ public String getRegistry() {
...
+
+    public AuthConfigs getAuthConfigs() {
+        return authConfigs;
+    }
+
+    /**
+     * Returns FQN of image for pull.
+     * It has following format: [registry/][namespace/]image
+     */
+    public String getImageFqn() {
```
garagatyi: Are you sure that you need that method? Why docker connector doesn't concat strings instead?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/PushParams.java
```diff
@@ -95,19 +126,36 @@ public String getRegistry() {
...
 
+    public AuthConfigs getAuthConfigs() {
+        return authConfigs;
+    }
+
+    /**
+     * Returns FQN of image for push.
+     * It has following format: [registry/][namespace/]image
+     */
+    public String getImageFqn() {
```
garagatyi: Are you sure we need that method?

garagatyi:
Please ensure that you've covered all changes with tests

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
mmorhun: I used general meaning of FQN. May be it is not good idea because tag is excluded. `fullImagePath` WDYT?

garagatyi: Yes, I think it is not good idea

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
mmorhun: You are right. It is redundant here.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
mmorhun: Because the map has format that similar with `X-Registry-Config`

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerRegistryAuthManager.java
```diff
@@ -0,0 +1,170 @@
...
+    public DockerRegistryAuthManager(InitialAuthConfig initialAuthConfig) {
+        this.initialAuthConfig = initialAuthConfig;
+    }
+
+    public String getDefaultUsername() {
+        return initialAuthConfig.getDefaultUsername();
+    }
+
+    /**
+     * Looks for auth header for specified registry.
```
mmorhun: Says in @return

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/PullParams.java
mmorhun: After last fixes we needn't it any more. In case of docker.io namespace is the same as username.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/PullParams.java
mmorhun: I think it useful. Just method invocation instead of `if` or ternary operator

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
garagatyi: so? code will be much cleaner if you pass the whole object

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/PullParams.java
garagatyi: Please do not use username/namespace at all. Docker doesn't declare that in its code.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/PullParams.java
garagatyi: There is no such a thing as FQN.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerRegistryAuthManager.java
```diff
@@ -0,0 +1,170 @@
...
+     * First searches in the params and then in the initial auth config.
+     * If nothing found empty json will be returned.
```
garagatyi: Here javadocs says empty json will be returned. Nothing about encoding. It is possible that devs will not read return description. So it is better to duplicate explanation of base 64 encoding in the main part of javadocs

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/PullParams.java
mmorhun: So, in case of docker.io it is mandatory, in case of another registry - optional. I removed the term `namespace`, but it content is included in full repo name

mmorhun:
@garagatyi check again please. I fixed bug with snapshot recovering from registry.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/PullParams.java
```diff
@@ -96,19 +110,36 @@ public String getRegistry() {
         return registry;
     }
 
+    public AuthConfigs getAuthConfigs() {
+        return authConfigs;
+    }
+
+    /**
+     * Returns full repo.
+     * It has following format: [registry/][namespace/]image
```
garagatyi: Remove namespace mentioning

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerRegistryAuthManager.java
mmorhun: ok

mmorhun:
@garagatyi ?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerRegistryAuthManager.java
```diff
@@ -0,0 +1,166 @@
...
+import org.apache.commons.codec.binary.Base64;
+import org.eclipse.che.commons.annotation.Nullable;
+import org.eclipse.che.commons.json.JsonHelper;
+import org.eclipse.che.plugin.docker.client.dto.AuthConfig;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Class for preparing docker auth data.
```
garagatyi: Headers?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerRegistryAuthManager.java
```diff
@@ -0,0 +1,166 @@
...
+    private final InitialAuthConfig initialAuthConfig;
+
+    @Inject
+    public DockerRegistryAuthManager(InitialAuthConfig initialAuthConfig) {
+        this.initialAuthConfig = initialAuthConfig;
+    }
+
+    /**
+     * Looks for auth header for specified registry and encode it in base64.
+     * First searches in the params and then in the initial auth config.
```
garagatyi: Replace `then in the initial auth config` with `then in the preconfigured system wide auth configs`

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerRegistryAuthManager.java
```diff
@@ -0,0 +1,166 @@
...
+     * If nothing found empty encoded json will be returned.
+     *
+     * @param registry
+     *         registry to which API call will be applied
+     * @param paramsAuthConfig
+     *         auth data for current API call
+     * @return base64 encoded X-Registry-Auth header value
+     */
+    public String getXRegistryAuthHeaderValue(String registry, @Nullable Map<String,AuthConfig> paramsAuthConfig) {
+        if (DEFAULT_REGISTRY.equals(registry)) {
```
garagatyi: Why you ignore provided auth configs in that case?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerRegistryAuthManager.java
```diff
@@ -0,0 +1,166 @@
...
+import java.util.Map;
+
+/**
+ * Class for preparing docker auth data.
+ *
+ * @author Mykola Morhun
+ */
+public class DockerRegistryAuthManager {
+
+    public static final String DEFAULT_REGISTRY = "";
```
garagatyi: Why you need that default registry? You can declare in method that if null is provided as registry parameter authconfig of default registry is returned

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerRegistryAuthManager.java
```diff
@@ -0,0 +1,166 @@
...
+            XRegistryAuthUnit auth = new XRegistryAuthUnit(authConfig.getUsername(), authConfig.getPassword());
+            return Base64.encodeBase64String(JsonHelper.toJson(auth).getBytes());
+        }
+
+        return Base64.encodeBase64String("{}".getBytes());
+    }
+
+    /**
+     * Looks for auth header for specified registry and encode it in base64.
+     * First searches in the params and then in the initial auth config.
```
garagatyi: Replace `then in the initial auth config` with `then in the preconfigured system wide auth configs`

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerRegistryAuthManager.java
```diff
@@ -0,0 +1,166 @@
...
+                                new XRegistryAuthUnit(value.getUsername(), value.getPassword()));
+            }
+        }
+
+        return Base64.encodeBase64String(JsonHelper.toJson(authConfigs).getBytes());
+    }
+
+    /** This class is used for generate X-Registry-Auth and X-Registry-Config */
+    // protected is needed for JsonHelper
+    protected static class XRegistryAuthUnit {
```
garagatyi: I think name **DockerCredentials** is better

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/PushParams.java
```diff
@@ -25,15 +26,16 @@
...
+    private String      tag;
+    private String      registry;
+    private AuthConfigs authConfigs;
 
     /**
      * Creates arguments holder with required parameters.
      *
      * @param repository
-     *         repository name
+     *         repository name in format: [username/]image
```
garagatyi: Remove mention of username

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/PushParams.java
```diff
@@ -48,7 +50,7 @@ private PushParams() {}
      * Adds repository to this parameters.
      *
      * @param repository
-     *         repository name
+     *         repository name in format: [username/]image
```
garagatyi: Remove mention of username

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/params/PushParams.java
```diff
@@ -25,15 +26,16 @@
...
 
     /**
      * Creates arguments holder with required parameters.
      *
      * @param repository
-     *         repository name
+     *         repository name in format: [username/]image
      * @return arguments holder with required parameters
      * @throws NullPointerException
      *         if {@code registry} is null
```
garagatyi: `if {@code registry} is null` mistake

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/InitialAuthConfig.java
```diff
@@ -43,14 +43,17 @@
 
     AuthConfig predefinedConfig;
 
+    private String serverAddress = "https://index.docker.io/v1/";
```
garagatyi: It should be configurable with properties

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/InitialAuthConfig.java
```diff
@@ -95,4 +98,13 @@ public AuthConfigs getAuthConfigs() {
         }
         return authConfigs;
     }
+
```
garagatyi: remove getAuthConfigHeader method if it is not needed anymore

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/InitialAuthConfig.java
```diff
@@ -95,4 +98,13 @@ public AuthConfigs getAuthConfigs() {
...
+    public String getDefaultUsername() {
```
garagatyi: I don't understand why we need these methods

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
```diff
@@ -98,6 +101,7 @@
     @Inject
     public DockerInstance(DockerConnector docker,
                           @Named("machine.docker.registry") String registry,
+                          @Named("docker.registry.auth.username") @Nullable String username,
```
garagatyi, line 106: Why you've added that?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerRegistryAuthManager.java
```diff
@@ -0,0 +1,166 @@
...
+     */
+    public String getXRegistryAuthHeaderValue(String registry, @Nullable Map<String,AuthConfig> paramsAuthConfig) {
+        if (DEFAULT_REGISTRY.equals(registry)) {
+            XRegistryAuthUnit auth = new XRegistryAuthUnit(initialAuthConfig.getDefaultUsername(), initialAuthConfig.getDefaultPassword());
+            return Base64.encodeBase64String(JsonHelper.toJson(auth).getBytes());
+        }
+
+        AuthConfig authConfig = null;
+        if (paramsAuthConfig != null) {
+            for(Map.Entry<String, AuthConfig> entry : paramsAuthConfig.entrySet()) {
```
benoitf: the two following loops (except the set of data) are the same, maybe we could "factorize" this code ?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/InitialAuthConfig.java
mmorhun: Yes, it is default value for that parameter

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
mmorhun, line 106: Because in case of docker.io namespace (i.e. username) is mandatory

garagatyi: Then it should be inside method that generates repository. It should not be username because it there is no username in repo. There are only parts of repo and some of them could implement some namespace. But it is not mandatory, it can include not 1 part of repo (separated by slashes) and it should not be called username.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerRegistryAuthManager.java
mmorhun: Note, that entry sets of that maps are obtained in two different ways and I don't like to make from foreach loop another method in this case.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerRegistryAuthResolver.java
```diff
@@ -0,0 +1,138 @@
...
+    public String getXRegistryAuthHeaderValue(String registry, @Nullable AuthConfigs paramsAuthConfigs) {
+        Map<String, AuthConfig> authConfigsMap = null;
+        if (paramsAuthConfigs != null) {
+            authConfigsMap = paramsAuthConfigs.getConfigs();
+        }
+
+        AuthConfig authConfig = null;
+        if (authConfigsMap != null) {
+            for(Map.Entry<String, AuthConfig> entry : authConfigsMap.entrySet()) {
+                AuthConfig value = entry.getValue();
```
garagatyi: Also check case with default regsitry

garagatyi: So you can move check from line 70 above

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerRegistryAuthResolver.java
```diff
@@ -0,0 +1,138 @@
...
+                if (value.getServeraddress().contains(registry)) {
+                    authConfig = value;
+                    break;
+                }
+            }
+        }
+
+        if (authConfig == null) {
+            if (DEFAULT_REGISTRY.equals(registry)) {
```
garagatyi: I think it can be good improvement to have a set with default registry names: docker.io, index.docker.io, http://indev.docker.io/v1/, "" so we will allow all of these as a default resitry identifier from auth conf provided as a parameter of this method

garagatyi:
LGTM

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -1297,14 +1305,17 @@ public String push(final PushParams params, final ProgressMonitor progressMonito
...
+        final String fullRepo = params.getFullRepo();
 
+        final ValueHolder<String> digestHolder = new ValueHolder<>();
         try (DockerConnection connection = connectionFactory.openConnection(dockerDaemonUri)
                                                             .method("POST")
                                                             .path("/images/" + fullRepo + "/push")
-                                                            .header("X-Registry-Auth", initialAuthConfig.getAuthConfigHeader())) {
+                                                            .header("X-Registry-Auth",
+                                                                    authManager.getXRegistryAuthHeaderValue(
+                                                                            registry != null ? registry : DEFAULT_REGISTRY_VALUE,
```
skabashnyuk: firstNotNull?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -1477,14 +1488,16 @@ public void pull(final PullParams params, final ProgressMonitor progressMonitor)
...
 
         try (DockerConnection connection = connectionFactory.openConnection(dockerDaemonUri)
                                                             .method("POST")
                                                             .path("/images/create")
-                                                            .query("fromImage", registry != null ? registry + '/' + image : image)
-                                                            .header("X-Registry-Auth", initialAuthConfig.getAuthConfigHeader())) {
+                                                            .query("fromImage", params.getFullRepo())
+                                                            .header("X-Registry-Auth",
+                                                                    authManager.getXRegistryAuthHeaderValue(
+                                                                            registry != null ? registry : DEFAULT_REGISTRY_VALUE,
```
skabashnyuk: firstNotNull

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerRegistryAuthResolver.java
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * Class for preparing docker auth data.
+ *
+ * @author Mykola Morhun
+ */
+public class DockerRegistryAuthResolver {
+
+    public static final ImmutableSet<String> DEFAULT_REGISTRY = ImmutableSet.of("", "docker.io", "index.docker.io",
+                                                                                "https://index.docker.io", "https://index.docker.io/v1",
```
skabashnyuk: wrap. Make each in separate line

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerRegistryAuthResolver.java
```diff
@@ -0,0 +1,142 @@
...
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.plugin.docker.client;
+
+import com.google.common.collect.ImmutableSet;
+import com.google.inject.Inject;
+
+import org.apache.commons.codec.binary.Base64;
```
skabashnyuk, line 16: https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html

skabashnyuk:
Other ok

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerRegistryAuthResolver.java
benoitf, line 16: on client side (gwt) this java.util class is not yet available

PR #1473, opened 2016-06-09
Set memory_swap_multiplier -1 to disable swap limit

riuvshin:
@garagatyi @vparfonov @skabashnyuk  Please take a look.
We should disable swap limits to workaround docker bug on some debian systems.
More info: https://github.com/docker/docker/issues/16256


garagatyi:
LGTM

vparfonov:
ok

PR #1478, opened 2016-06-10
CODENVY-641: Remove user verification during getting factory json

akorneta:
Since we add permission mechanism for access management, we need rework existing services 
@skabashnyuk, @sleshchenko


skabashnyuk:
ok

sleshchenko:
LGTM

PR #1479, opened 2016-06-10
CODENVY-557: remove user role from dashboard

olexii4:


ashumilova:
Also need to remove following things:
https://github.com/eclipse/che/blob/CODENVY-557/dashboard/src/components/api/builder/che-api-builder.factory.js#L21
https://github.com/eclipse/che/blob/CODENVY-557/dashboard/src/components/api/test/che-http-backend.js#L40
https://github.com/eclipse/che/ablob/CODENVY-557/dashboard/src/components/api/builder/che-user-builder.js

Other is Ok, +1

olexii4:
Forgot about httpBackend.
Done

PR #1487, opened 2016-06-13
CODENVY-619 Handle normalization of fully illegal usernames

mkuznyetsov:
In case where username in existing LDAPs consists of illegal characters only, it will be replaced with generated username.
Also, ensure username is unique.

### DIFF
--- a/wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserNameValidator.java
+++ b/wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserNameValidator.java
@@ -10,6 +10,13 @@
  *******************************************************************************/
 package org.eclipse.che.api.user.server;
 
+import org.eclipse.che.api.core.NotFoundException;
+import org.eclipse.che.api.core.ServerException;
+import org.eclipse.che.commons.lang.NameGenerator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.inject.Inject;
 import java.util.regex.Pattern;
 
 /**
@@ -18,8 +25,17 @@ import java.util.regex.Pattern;
  * @author Mihail Kuznyetsov
  */
 public class UserNameValidator {
+    private static final Logger LOG = LoggerFactory.getLogger(UserNameValidator.class);
+
     private static final Pattern ILLEGAL_USERNAME_CHARACTERS = Pattern.compile("[^a-zA-Z0-9]");
     private static final Pattern VALID_USERNAME              = Pattern.compile("^[a-zA-Z0-9]*");
+
+    private final UserManager userManager;
+
+    @Inject
+    public UserNameValidator (UserManager userManager) {
+        this.userManager = userManager;
+    }
 
     /**
      * Validate name, if it doesn't contain illegal characters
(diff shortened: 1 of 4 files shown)


wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserNameValidator.java
```diff
@@ -18,28 +25,53 @@
...
      * @return true if valid name, false otherwise
      */
-    public static boolean isValidUserName(String name) {
+    public boolean isValidUserName(String name) {
         return VALID_USERNAME.matcher(name).matches();
     }
 
     /**
      * Remove illegal characters from username, to make it URL-friendly.
+     * If all characters are illegal, return username.
```
mshaposhnik: you mean, "return auto-generated username" ?

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserNameValidator.java
```diff
@@ -18,28 +25,53 @@
...
+        int i = 1;
+        try {
+            while (true) {
+                userManager.getByName(candidate);
+                candidate = normalized.isEmpty() ? NameGenerator.generate("username", 4) : normalized + String.valueOf(i++);
+            }
+        } catch (NotFoundException e) {
+            return candidate;
+        } catch (ServerException e) {
+            LOG.warn("Error occured during username normalization", e);
```
sleshchenko, line 70: Maybe we should throw exception here instead of returning potential existing name

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserNameValidator.java
mkuznyetsov: fixed

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserNameValidator.java
mkuznyetsov, line 70: ok

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserNameValidator.java
```diff
@@ -18,28 +25,53 @@
...
      */
-    public static String normalizeUserName(String name) {
-        return ILLEGAL_USERNAME_CHARACTERS.matcher(name).replaceAll("");
+    public String normalizeUserName(String name) {
+        String normalized = ILLEGAL_USERNAME_CHARACTERS.matcher(name).replaceAll("");
+        String candidate = normalized.isEmpty() ? NameGenerator.generate("username", 4) : normalized;
+
+        int i = 1;
+        try {
+            while (true) {
```
mkuznyetsov: @evoevodin @skabashnyuk WDYT?

voievodin: I would rework it to the following:

``` java
while (userExists(userName)) {
    // generate new name
}
```

similar approach [from 3.x](https://github.com/codenvy/che-core/blob/3.x/platform-api/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java#L1086-L1088)

wsmaster/che-core-api-user/src/test/java/org/eclipse/che/api/user/server/UserNameValidatorTest.java
```diff
@@ -10,32 +10,47 @@
...
 public class UserNameValidatorTest {
+
+    @Mock
+    private UserManager userManager;
+
+    private UserNameValidator userNameValidator;
+
     @BeforeMethod
     public void setUp() {
+        userNameValidator = new UserNameValidator(userManager);
```
voievodin: What about `@InjectMocks`?

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserNameValidator.java
mkuznyetsov: :+1:

wsmaster/che-core-api-user/src/test/java/org/eclipse/che/api/user/server/UserNameValidatorTest.java
mkuznyetsov: ok

skabashnyuk:
ok

mshaposhnik:
ok

PR #1488, opened 2016-06-13
CHE-1320 Fix redirection from IDE to dashboard - workspace is not selected

vitaliy-guliy:


ashumilova:
+1

PR #1489, opened 2016-06-13
CODENVY-645: add checking on empty list of paths

vparfonov:
start re-import Maven model only for maven project 


riuvshin:
OK

PR #1491, opened 2016-06-13
CHE-1317: fix import project while creating a workspace

olexii4:


akurinnoy:
OK, as a temporary fix

benoitf:
AFAIK it's not a fix, as if I import a project from git containing a pom.xml it will be imported as blank project and not as a maven project

ashumilova:
by "fix" mean - fixing project import, cause now doesn't work at all. Imported projects will be blank, and then will provide ASAP fix for project type estimation, cause we detect project type base on it, and estimation doesn't work properly.

slemeur:
I disagree with this quickfix. We need to fix the project type detection - this workaround can't be merged.

gazarenkov:
https://jira.codenvycorp.com/browse/CHE-1331

PR #1492, opened 2016-06-13
CHE-1316: Temporary fix for to avoid cyclic dependencies of components

akorneta:
During the initialization of the projects we have :
IllegalStateException with message "Projects are not initialized yet", to avoid this situation we comment the code that depends on component which initialize project because the workspace agent must always start even if there are any problems with projects.
@vparfonov, @skabashnyuk please review


skabashnyuk:
Ok

PR #1495, opened 2016-06-14
[WIP] add log output to debug build fail on jenkins

evidolob:
This pull request to debug build failing on jectkins, do not accept it!!!

### DIFF
--- a/plugins/plugin-maven/che-plugin-maven-server/src/main/java/org/eclipse/che/plugin/maven/server/MavenServerManager.java
+++ b/plugins/plugin-maven/che-plugin-maven-server/src/main/java/org/eclipse/che/plugin/maven/server/MavenServerManager.java
@@ -86,6 +86,7 @@ public class MavenServerManager extends RmiObjectWrapper<MavenRemoteServer> {
         return () -> {
             JavaParameters parameters = buildMavenServerParameters();
             CommandLine command = parameters.createCommand();
+            LOG.debug("Creating new Executor for: " + parameters.getJavaExecutable());
             return new ProcessHandler(command.createProcess());
         };
     }
--- a/plugins/plugin-maven/che-plugin-maven-server/src/main/java/org/eclipse/che/plugin/maven/server/execution/ProcessHandler.java
+++ b/plugins/plugin-maven/che-plugin-maven-server/src/main/java/org/eclipse/che/plugin/maven/server/execution/ProcessHandler.java
@@ -91,6 +91,7 @@ public class ProcessHandler implements Executor {
             @Override
             public void onStart(ProcessEvent event) {
                 try {
+                    LOG.debug("Maven server JVM Started");
                     OutputReader stdOutReader = createStdOutReader();
                     stdOutReader.start();
 
@@ -160,6 +161,7 @@ public class ProcessHandler implements Executor {
     }
 
     private void notifyOnText(String text, ProcessOutputType type) {
+        LOG.debug(type.name() + " " + text);
         listenerNotifier.onText(new ProcessEvent(this, text), type);
     }
 
--- a/plugins/plugin-maven/che-plugin-maven-server/src/main/java/org/eclipse/che/plugin/maven/server/rmi/RmiClient.java
+++ b/plugins/plugin-maven/che-plugin-maven-server/src/main/java/org/eclipse/che/plugin/maven/server/rmi/RmiClient.java
@@ -97,6 +97,7 @@ public abstract class RmiClient<Remote> {
         try {
             ProcessExecutor executor = getExecutor();
             handler = executor.execute();
+            LOG.debug("Maven Server JVM process executed: " + handler.toString());
         } catch (ExecutionException e) {
             removeProcessInfo(key, null, e.getMessage());
             return;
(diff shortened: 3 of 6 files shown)


skabashnyuk:
ok

evidolob:
ci-build

plugins/plugin-maven/che-plugin-maven-server/src/main/java/org/eclipse/che/plugin/maven/server/MavenServerManager.java
```diff
@@ -86,6 +86,7 @@ private ProcessExecutor createExecutor() {
         return () -> {
             JavaParameters parameters = buildMavenServerParameters();
             CommandLine command = parameters.createCommand();
+            LOG.debug("Creating new Executor for: " + parameters.getJavaExecutable());
```
benoitf, line 89: for log debug, we should not use string concatenation but object parameters

plugins/plugin-maven/che-plugin-maven-server/src/main/java/org/eclipse/che/plugin/maven/server/execution/ProcessHandler.java
```diff
@@ -160,6 +161,7 @@ private void notifyOnTerminating() {
     }
 
     private void notifyOnText(String text, ProcessOutputType type) {
+        LOG.debug(type.name() + " " + text);
```
benoitf, line 164: string concatenation shouldn't be used in DEBUG traces (give object as parameters and use kind of MessageFormat

plugins/plugin-maven/che-plugin-maven-server/src/main/java/org/eclipse/che/plugin/maven/server/rmi/RmiClient.java
```diff
@@ -97,6 +97,7 @@ private void startProcess(Pair<Object, Object> key) {
         try {
             ProcessExecutor executor = getExecutor();
             handler = executor.execute();
+            LOG.debug("Maven Server JVM process executed: " + handler.toString());
```
benoitf, line 100: concatenation shouldn't be used for debug trace

plugins/plugin-maven/che-plugin-maven-server/src/test/java/org/eclipse/che/plugin/maven/server/rmi/MavenServerManagerTest.java
```diff
@@ -53,7 +54,8 @@
 
 
     @BeforeMethod
-    public void setUp() throws Exception {
+    public void setUp(Method method) throws Exception {
+        System.out.println("TEST NAME:  "+method.getName() + "!!!!!!!!!!!!!!!!!");
```
benoitf, line 58: I would say that even in a test println shouldn't be used

plugins/plugin-maven/maven-server/maven-server-api/src/main/java/org/eclipse/che/rmi/RmiServer.java
```diff
@@ -33,9 +33,9 @@ protected static void start(Remote remote) throws RemoteException {
         if (RmiServer.remote != null) {
             throw new AssertionError("This server is already started!");
         }
-
         RmiServer.remote = remote;
 
+        System.err.println("Maven Server try to select port");
```
benoitf, line 38: is it an error message ?
we can't use Logger to print ?

plugins/plugin-maven/maven-server/maven-server-api/src/main/java/org/eclipse/che/rmi/RmiServer.java
```diff
@@ -51,14 +51,16 @@ protected static void start(Remote remote) throws RemoteException {
             }
 
         }
-
+        System.err.println("Maven Server port selected: " + port);
```
benoitf, line 54: is it an error message ?
we can't use Logger to print ?

plugins/plugin-maven/maven-server/maven-server-api/src/main/java/org/eclipse/che/rmi/RmiServer.java
```diff
@@ -51,14 +51,16 @@ protected static void start(Remote remote) throws RemoteException {
...
         Remote exportObject = UnicastRemoteObject.exportObject(remote, port);
         String exportName = remote.getClass().getSimpleName() + Integer.toHexString(exportObject.hashCode());
+        System.err.println("Maven Server remote object exported.");
```
benoitf, line 57: it should use DEBUG logger

plugins/plugin-maven/maven-server/maven-server-api/src/main/java/org/eclipse/che/rmi/RmiServer.java
```diff
@@ -51,14 +51,16 @@ protected static void start(Remote remote) throws RemoteException {
...
         try {
             registry.bind(exportName, exportObject);
             String portName = port + "/" + exportName;
             System.out.println("Port/Name:" + portName);
             int twoMinutes = 2 * 60 * 1000;
+            System.err.println("Maven Server port and name printed");
```
benoitf, line 63: it should use DEBUG logger

benoitf:
I've added WIP as it is only for jenkins :-)

PR #1500, opened 2016-06-14
Export proxy object instead of real maven terminal object. …

vparfonov:
Avoid NPE in case jar lab's are not set for javac project


vparfonov:
@riuvshin

PR #1501, opened 2016-06-14
CHE-1317: add a second attempt with type blank if we have an error du…

olexii4:
**_1 Upvote**_ …ring project import


gazarenkov:
+1

PR #1503, opened 2016-06-14
CHE-1314 Add loader during workspace snapshotting process

mkuznyetsov:
Prevent user from doing any actions in IDE while snapshot is being created.


skabashnyuk:
Can you add some screenshots, how its looks like?

skabashnyuk:
@vparfonov ?

mkuznyetsov:
loader
It will be shown while snapshotting is in progress on server

skabashnyuk:
ok

PR #1506, opened 2016-06-14
CHE-1337: fix import Maven project form dashboard

olexii4:
**_1 Upvote**_ Signed-off-by: Oleksii Orel oorel@codenvy.com

### DIFF
--- a/dashboard/src/app/projects/create-project/create-project.controller.js
+++ b/dashboard/src/app/projects/create-project/create-project.controller.js
@@ -570,23 +570,42 @@ export class CreateProjectCtrl {
         if (estimateTypes.length > 0) {
           // wait estimate are all finished
           let waitEstimate = this.$q.all(estimatePromises);
+          let attributesByMatchingType = new Map();
 
           waitEstimate.then(() => {
-            var firstMatchingType;
-            var firstMatchingResult;
+            let firstMatchingType;
             estimateTypes.forEach((type) => {
               let resultEstimate = projectService.getEstimate(projectName, type);
               // add attributes
-              // there is a matching estimate
-              if (Object.keys(resultEstimate.attributes).length > 0 && 'java' !== type && !firstMatchingType) {
-                firstMatchingType = type;
-                firstMatchingResult = resultEstimate.attributes;
+              if (Object.keys(resultEstimate.attributes).length > 0) {
+                attributesByMatchingType.set(type, resultEstimate.attributes);
+              }
+            });
+            
+            attributesByMatchingType.forEach((attributes, type) => {
+              if (!firstMatchingType) {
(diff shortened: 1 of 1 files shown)


slemeur:
+1 - for the time being and resolution of import maven project

dashboard/src/app/projects/create-project/create-project.controller.js
```diff
@@ -570,23 +570,42 @@ export class CreateProjectCtrl {
...
+                attributesByMatchingType.set(type, resultEstimate.attributes);
+              }
+            });
+            
+            attributesByMatchingType.forEach((attributes, type) => {
+              if (!firstMatchingType) {
+                let projectType = projectTypesByCategory.get(type);
+                if (projectType && projectType.parents) {
+                  projectType.parents.forEach((parentType) => {
+                    if (parentType === 'java') {
```
vparfonov, line 590: can we avoid hard-coding here?

vparfonov:
@benoitf

dashboard/src/app/projects/create-project/create-project.controller.js
benoitf, line 590: I think we have the same in IDE as java is not shown

benoitf:
OK only for 4.3.2 this has to be written correctly and work smoothly from server

riuvshin:
OK, thanks!

dashboard/src/app/projects/create-project/create-project.controller.js
olexii4, line 590: When I try to get project types:     /wsagent/ext/project-type (GET)
Response:
0: {persisted: false, parents: ["java"], mixable: false, primaryable: true,…id: "javac"...}
1: {persisted: false, parents: ["blank"], mixable: false, primaryable: true,…id: "cpp"...}
2: {persisted: false, parents: ["blank"], mixable: false, primaryable: true, ... id: "python",…}
3: {persisted: false, parents: [], mixable: false, primaryable: true, ancestors: [], id: "blank",…}
4: {persisted: false, parents: ["blank"], mixable: false, primaryable: true, ... id: "java",…}
5: {persisted: false, parents: ["blank"], mixable: true, primaryable: false, ... id: "git",…}
6: {persisted: false, parents: ["blank"], mixable: false, primaryable: true, ... id: "node-js",…}
7: {persisted: false, parents: ["blank"], mixable: false, primaryable: true, ... id: "c",…}
8: {persisted: false, parents: ["blank"], mixable: true, primaryable: false, ... id: "subversion",…}
9: {persisted: false, parents: ["java"], mixable: false, primaryable: true, ... id: "maven",  ...}

@vparfonov: What should we do for 4.4 version with this parent type in case when we have more than one valid type with attributes( like 'maven' and 'jjavac')?

PR #1509, opened 2016-06-15
CHE-1326: avoid creation recipe on workspace creation from stack and overriding environment

ashumilova:

Partly depends on https://jira.codenvycorp.com/browse/CODENVY-651

### DIFF
--- a/dashboard/src/app/projects/create-project/create-project.controller.js
+++ b/dashboard/src/app/projects/create-project/create-project.controller.js
@@ -581,7 +581,7 @@ export class CreateProjectCtrl {
                 attributesByMatchingType.set(type, resultEstimate.attributes);
               }
             });
-            
+
             attributesByMatchingType.forEach((attributes, type) => {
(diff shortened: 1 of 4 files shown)


dashboard/src/app/projects/create-project/create-project.controller.js
```diff
@@ -831,34 +831,17 @@ export class CreateProjectCtrl {
     }
     // check workspace is selected
     if (option === 'create-workspace') {
+      let source = {};
```
benoitf: is that source should be only in the else block ?

ashumilova: yes, will move it there

ashumilova:
@benoitf @evoevodin @AndrienkoAleksandr  please check

voievodin:
serverside is ok

AndrienkoAleksandr:
LGTM

PR #1510, opened 2016-06-15
CHE-1328: fix checking on equals in case checking null and empty string

vparfonov:
Code clean up


core/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/workspace/DefaultWorkspaceComponent.java
```diff
@@ -54,7 +54,7 @@
  *
  */
 @Singleton
-public class DefaultWorkspaceComponent extends WorkspaceComponent implements Component {
+public class DefaultWorkspaceComponent extends WorkspaceComponent  {
```
benoitf, line 57: it seems unrelated with PR or commit message, no ?

benoitf, line 57: ok thx would have been cleaner with two commits in the PR

PR #1513, opened 2016-06-15
CHE-1340: Move ide related modules to /che/ide

vparfonov:
move:
core/ide -> ide
core/commons/che-core-commons-gwt -> ide/commons-gwt


vparfonov:
@skabashnyuk @gazarenkov @benoitf @vzhukovskii @TylerJewell

benoitf:
2,457 changes ^^

ide/che-core-ide-api/pom.xml
```diff
@@ -18,6 +18,7 @@
         <groupId>org.eclipse.che.core</groupId>
         <version>4.4.0-RC1-SNAPSHOT</version>
     </parent>
+    <groupId>org.eclipse.che.core</groupId>
```
benoitf, line 21: groupId should be org.eclipse.che.ide now as it's no longer in the core part

benoitf, line 21: (could be done in a next commit)
to be sure groupId are aligned with the location of modules

benoitf, line 21: artifact id or module name are still using "core" inside their name as well
che-core-ide-api vs che-ide-api

vparfonov:
I think here we don't nee review.
Just moving directory, any changes inside classes, mention you to be sure you are informed because you not in our office

ide/che-core-ide-api/pom.xml
vparfonov, line 21: agree it will be done in other PR

vzhukovs:
Okay, I'll apply the same movement in my branches.

skabashnyuk:
ok

PR #1515, opened 2016-06-16
CHE-1297: Git service refactoring

vinokurig:
The goal of this refactoring is to set correct type of HTTP methods in `GitService`. Now almost all HTTP methods in this service are POST methods, but there are methods that are used to get some content, so they should be marked as GET method. Also I had to change such methods to receive query parameters instead of body.
I have added parameter objects to use them instead of using DTO as parameter in `GitConnection` methods.
@skabashnyuk @sleshchenko please review


wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/params/CheckoutParams.java
```diff
@@ -0,0 +1,111 @@
...
+
+    /** @see CheckoutRequest#withTrackBranch(String) */
+    public CheckoutParams withTrackBranch(String trackBranch) {
+        this.trackBranch = trackBranch;
+        return this;
+    }
+
+    /** @see CheckoutRequest#getFiles() */
+    public List<String> getFiles() {
+        return filesToCheckout == null ? Collections.emptyList() : filesToCheckout;
```
sleshchenko: Should we return immutable list here?

wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/params/CloneWithSparseCheckoutParams.java
```diff
@@ -0,0 +1,74 @@
...
+        return this;
+    }
+
+    /** @return branch to checkout */
+    public String getBranch() {
+        return branch;
+    }
+
+    public CloneWithSparseCheckoutParams withBranch(@NotNull String branch) {
+        this.branch = branch;
```
sleshchenko: requireNonNull?

skabashnyuk:
Can you describe what is going on here?

vinokurig:
Updated description of this PR

skabashnyuk:
@gazarenkov  @evoevodin  please review GitService REST service

wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/params/CheckoutParams.java
vinokurig: reworked all parameters classes to return `new ArrayList`

wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/params/CloneWithSparseCheckoutParams.java
vinokurig: I have separated checkout logic from cloneWithSparceCheckout method, added a check for null directory inside each implementation, added a test  for clone with sparce-checkout

voievodin:
The approach is a bit incorrect 
e.g. 

```
@DELETE
@Path("delete-repository")
```

should be replaced with 

```
@DELETE
@Path("repository")
```

and similar mistakes.
@vinokurig [this](https://wiki.codenvycorp.com/display/PSR/REST+services+links) documentation may be helpful for you. Try to avoid usage of verbs as much as possible e.g. i am not sure that you can avoid `checkout` in path but at least you can avoid `delete-repository`-like paths

voievodin:
@vinokurig what motivation behind renaming `*Request` classes to `*Params` classes?

vinokurig:
@evoevodin It is not renaming. I used new `*Params` classes instead of using request DTO as parameter in `GitConnection` methods. If we will need to create any method from `GitConnection` in server side we will use new parameter object instead of creating DTO as parameter.

voievodin:
i see, thank you

wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitService.java
```diff
@@ -169,224 +191,281 @@ public RepoInfo clone(final CloneRequest request) throws URISyntaxException, Api
...
             return gitConnection.status(format);
         }
     }
 
-    @Path("tag-create")
     @POST
+    @Path("tag")
     @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
     public Tag tagCreate(TagCreateRequest request) throws ApiException {
```
iamleshiy: In other services we name methods in another way. WorkspaceService#createMachine, WorkspaceService#createSnapshot. Mb it would be better to rename this method to createTag. WDYT?

sleshchenko, line 416: In other services we name methods in another way. WorkspaceService#createMachine, WorkspaceService#createSnapshot. Mb it would be better to rename this method to createTag. WDYT?

wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitService.java
```diff
@@ -169,224 +191,281 @@ public RepoInfo clone(final CloneRequest request) throws URISyntaxException, Api
...
     public void mv(MoveRequest request) throws ApiException {
         try (GitConnection gitConnection = getGitConnection()) {
-            gitConnection.mv(request);
+            gitConnection.mv(request.getSource(), request.getTarget());
         }
     }
 
-    @Path("pull")
     @POST
+    @Path("rm")
```
sleshchenko: rm - is unclear name for me

wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitConnection.java
```diff
@@ -434,38 +407,32 @@
...
-     * @return list of tags matched to request, see {@link TagListRequest#getPattern()}
+     * @param pattern
+     *         tag's names pattern
+     * @return list of tags matched to request
      * @throws GitException
      *         if any error occurs
-     * @see TagListRequest
      */
-    List<Tag> tagList(TagListRequest request) throws GitException;
+    List<Tag> tagList(String pattern) throws GitException;
```
iamleshiy: getTags?

wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitService.java
```diff
@@ -90,76 +100,88 @@
...
-    @Consumes(MediaType.APPLICATION_JSON)
-    @Produces({MediaType.APPLICATION_JSON, MediaType.TEXT_PLAIN})
-    public GenericEntity<List<Branch>> branchList(BranchListRequest request) throws ApiException {
+    @GET
+    @Path("branch")
+    @Produces({MediaType.APPLICATION_JSON})
+    public GenericEntity<List<Branch>> branchList(@QueryParam("listMode") String listMode) throws ApiException {
         try (GitConnection gitConnection = getGitConnection()) {
-            return new GenericEntity<List<Branch>>(gitConnection.branchList(request)) {
+            return new GenericEntity<List<Branch>>(gitConnection.branchList(listMode == null ? null : BranchListMode.valueOf(listMode))) {
```
sleshchenko: BranchListMode.valueOf(listMode) can throw IllegalArgumentException. Mb we should wrap it to BadRequestException?

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/git/GitServiceClientImpl.java
```diff
@@ -445,69 +419,60 @@ public void push(DevMachine devMachine,
...
-    /** {@inheritDoc} */
     @Override
     public Promise<List<Remote>> remoteList(DevMachine devMachine, ProjectConfig project, @Nullable String remoteName, boolean verbose) {
-        RemoteListRequest remoteListRequest = dtoFactory.createDto(RemoteListRequest.class).withVerbose(verbose);
-        if (remoteName != null) {
-            remoteListRequest.setRemote(remoteName);
-        }
-        String url = devMachine.getWsAgentBaseUrl() + REMOTE_LIST + "?projectPath=" + project.getPath();
-        return asyncRequestFactory.createPostRequest(url, remoteListRequest)
+        String params = "?projectPath=" + project.getPath() + (remoteName != null ? "&remoteName=" + remoteName : "") +
```
sleshchenko: I would move this chech `(remoteName != null ? "&remoteName=" + remoteName : "")` to if

```
if (remoteName != null) {
  params += &remoteName=" + remoteName;
}
```

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/git/GitServiceClientImpl.java
```diff
@@ -445,69 +419,60 @@ public void push(DevMachine devMachine,
...
+                        "&verbose=" + String.valueOf(verbose);
+        String url = appContext.getDevMachine().getWsAgentBaseUrl() + REMOTE + params;
+        return asyncRequestFactory.createGetRequest(url)
                                   .loader(loader)
                                   .send(dtoUnmarshallerFactory.newListUnmarshaller(Remote.class));
     }
 
-    /** {@inheritDoc} */
     @Override
     public void branchList(DevMachine devMachine,
```
sleshchenko: Make sure that you add @Deprecated annotation to all methods for which you provide similar method with promise

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/git/GitServiceClientImpl.java
```diff
@@ -445,69 +419,60 @@ public void push(DevMachine devMachine,
...
-        String url = devMachine.getWsAgentBaseUrl() + BRANCH_LIST + "?projectPath=" + project;
-        return asyncRequestFactory.createPostRequest(url, branchListRequest).send(dtoUnmarshallerFactory.newListUnmarshaller(Branch.class));
+    public Promise<List<Branch>> branchList(DevMachine devMachine, Path project, BranchListMode listMode) {
+        String url = appContext.getDevMachine().getWsAgentBaseUrl() + BRANCH + "?projectPath=" + project.toString() + "&listMode=" +
+                     listMode;
+        return asyncRequestFactory.createGetRequest(url).send(dtoUnmarshallerFactory.newListUnmarshaller(Branch.class));
     }
 
     @Override
     public Promise<Status> status(DevMachine devMachine, ProjectConfig project) {
```
sleshchenko: I think that `getStatus` and `status` methods can reuse common code and should be named the same

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/git/GitServiceClientImpl.java
```diff
@@ -806,129 +760,137 @@ public void pull(DevMachine devMachine,
...
                      String commitA,
                      String commitB, @NotNull AsyncRequestCallback<String> callback) {
-        DiffRequest diffRequest = dtoFactory.createDto(DiffRequest.class)
-                                            .withFileFilter(fileFilter)
-                                            .withType(type)
-                                            .withNoRenames(noRenames)
-                                            .withCommitA(commitA)
-                                            .withCommitB(commitB)
-                                            .withRenameLimit(renameLimit);
+        StringBuilder params = new StringBuilder().append("?projectPath=").append(project.getPath());
```
sleshchenko: Can we reuse code for method with callback and with promise? mb we can create some private method `AsyncRequest diff(...)` which will construct needed request. And then method with promise will have following code

```
    return diff(...).send(new StringUnmarshaller())
```

And method with callback:

```
  diff(...).send(callback);
```

wsagent/che-core-api-git-shared/src/main/java/org/eclipse/che/api/git/shared/BranchListMode.java
```diff
@@ -10,19 +10,21 @@
  *******************************************************************************/
 package org.eclipse.che.api.git.shared;
 
-import org.eclipse.che.dto.shared.DTO;
-
 /**
- * Request for calling git ls-remote.
+ * Set of branch list modes.
```
sleshchenko: IMHO Word `set` is a little bit confusing in this case

wsagent/che-core-api-git-shared/src/main/java/org/eclipse/che/api/git/shared/BranchListMode.java
```diff
@@ -10,19 +10,21 @@
...
-    /** @return url of remote repository. */
-    String getRemoteUrl();
-
-    void setRemoteUrl(String remoteUrl);
-
-    LsRemoteRequest withRemoteUrl(String remoteUrl);
+public enum BranchListMode {
+    /**
+     * Show both remote and local branches. <br/>
+     * Corresponds to -a option in C git.
```
sleshchenko: Can you explain meaning of `C git`?

wsagent/che-core-api-git-shared/src/main/java/org/eclipse/che/api/git/shared/BranchListMode.java
```diff
@@ -10,19 +10,21 @@
...
+     */
+    LIST_ALL,
+    /**
+     * Show only remote branches. <br/>
+     * Corresponds to -r option in C git.
+     */
+    LIST_REMOTE,
+    LIST_LOCAL
```
sleshchenko, line 35: I would be better to write doc for this enum value if you provide doc for another values

wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitConnection.java
```diff
@@ -135,110 +114,122 @@
...
 
     /**
      * List branches.
      *
-     * @param request
-     *         list branches request
-     * @return list of branch
+     * @param listMode
+     *         specifies what branches to list.
+     *         "List all" mode will be set if parameter is not specified
```
sleshchenko: You can use {@link } here instead of `"List all"`

JamesDrummond:
Please provide status.

wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitService.java
vinokurig: I totally agree with you but I think it will be another issue with separate pull request as it will contain such renaming of almost half methods in `GitService`, `GitConnection` and `JGitConnection`

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/git/GitServiceClientImpl.java
vinokurig: All such methods are deprecated in the interface

wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitService.java
vinokurig: renamed to `remove`

wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitService.java
vinokurig: done

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/git/GitServiceClientImpl.java
vinokurig: done

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/git/GitServiceClientImpl.java
vinokurig: `status` method is deprecated. I think it will be removed soon.

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/git/GitServiceClientImpl.java
vinokurig: done

wsagent/che-core-api-git-shared/src/main/java/org/eclipse/che/api/git/shared/BranchListMode.java
vinokurig: changed to `Branch list modes`

wsagent/che-core-api-git-shared/src/main/java/org/eclipse/che/api/git/shared/BranchListMode.java
vinokurig: It is old javadoc from `BranchListRequest`. Changed `C` to `console`

wsagent/che-core-api-git-shared/src/main/java/org/eclipse/che/api/git/shared/BranchListMode.java
vinokurig: done

wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitConnection.java
vinokurig: done

sleshchenko (approved):

skabashnyuk (approved):

PR #1519, opened 2016-06-17
Cleanup in maven module build

skabashnyuk:
Move wsmaster and wsagent maven module declaration to upper level from core module


garagatyi:
LGTM

PR #1520, opened 2016-06-17
CODENVY-667 Adapt predefined recipes and stacks to new format

sleshchenko:
@skabashnyuk Please review my changes


skabashnyuk:
ok

PR #1521, opened 2016-06-17
CODENVY-669: Catch exception on init projects for avoiding crash wsagent

vparfonov:
@gazarenkov @skabashnyuk @riuvshin 

### DIFF
--- a/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/RegisteredProject.java
+++ b/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/RegisteredProject.java
@@ -17,6 +17,7 @@ import org.eclipse.che.api.core.model.project.SourceStorage;
 import org.eclipse.che.api.core.model.project.type.Attribute;
 import org.eclipse.che.api.core.model.project.type.Value;
 import org.eclipse.che.api.project.server.type.AttributeValue;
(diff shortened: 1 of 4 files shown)


gazarenkov:
LGTM

just add comment here to make it clear like:

// need that to make "problematic" project and not break the workspace
try {
              putProject(projectConfig, projectFolder, false, false);

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/RegisteredProject.java
```diff
@@ -95,6 +96,40 @@
...
+                                                                      ValueStorageException {
+        problems = new ArrayList<>();
+        attributes = new HashMap<>();
+
+        this.folder = folder;
+        this.config = new NewProjectConfig(folder.getPath());
+        this.updated = updated;
+        this.detected = detected;
+        types = new ProjectTypes(folder.getPath().toString(), BaseProjectType.ID, null, projectTypeRegistry);
+        problems.add(problem);
```
skabashnyuk: problems  = ImmutableList.of(problem)
wdyt?

vparfonov: ok

PR #1525, opened 2016-06-17
CHE-1234 Allow only URL safe characters in command names

mkuznyetsov:


benoitf:
I don't see any test for the regexp

"Workspace command '%s' must only consist of letters, digits, and '-' '_' '.' in the middle",
also does it mean space is not allowed ?

because I've tons of command names including space in it

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -484,6 +484,8 @@ public void deleteCommand(@ApiParam("The id of the workspace")
         final WorkspaceImpl workspace = workspaceManager.getWorkspace(id);
         if (workspace.getConfig().getCommands().removeIf(command -> command.getName().equals(commandName))) {
             workspaceManager.updateWorkspace(id, workspace);
+        } else {
+            throw new NotFoundException(String.format("Failed to delete command '%s'", commandName));
```
sleshchenko: > Command with name '%s' was not found in workspace with id '%s'?

I think this message would be better for NotFoundException
Also add swagger annotation for this case of 404 response code

mkuznyetsov: ok

PR #1526, opened 2016-06-17
CHE-1234 Encode command name when making request to service

mkuznyetsov:


ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/workspace/WorkspaceServiceClientImpl.java
```diff
@@ -269,7 +271,11 @@ private void updateCommand(@NotNull final String wsId,
                                @NotNull final CommandDto commandUpdate,
                                final String commandName,
                                @NotNull AsyncCallback<WorkspaceDto> callback) {
-        final String url = baseHttpUrl + '/' + wsId + "/command/" + commandName;
+        String url = null;
+        try {
+            url = baseHttpUrl + '/' + wsId + "/command/" + URLEncoder.encode(commandName, "UTF-8");
+        } catch (UnsupportedEncodingException ignored) {
```
benoitf: I'm not sure silent catch is good as after that url is null and it will create exceptions

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/workspace/WorkspaceServiceClientImpl.java
```diff
@@ -290,7 +296,11 @@ public void makeCall(AsyncCallback<WorkspaceDto> callback) {
     private void deleteCommand(@NotNull final String wsId,
                                @NotNull final String commandName,
                                @NotNull AsyncCallback<WorkspaceDto> callback) {
-        final String url = baseHttpUrl + '/' + wsId + "/command/" + commandName;
+        String url = null;
+        try {
+            url = baseHttpUrl + '/' + wsId + "/command/" + URLEncoder.encode(commandName, "UTF-8");
+        } catch (UnsupportedEncodingException ignored) {
+        }
```
benoitf: I'm not sure silent catch is good as after that url is null and it will create exceptions

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/workspace/WorkspaceServiceClientImpl.java
garagatyi: Do you think it is possible that "UTF-8" is not supported?

skabashnyuk:
ci-build

skabashnyuk:
ci-build

skabashnyuk:
Can you make a screenshot what user will have?

plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/command/edit/EditCommandsPresenter.java
```diff
@@ -142,6 +142,11 @@ public void apply(WorkspaceDto arg) throws OperationException {
                 fetchCommands();
                 fireConfigurationUpdated(selectedConfiguration);
             }
+        }).catchError(new Operation<PromiseError>() {
+            @Override
+            public void apply(PromiseError arg) throws OperationException {
+                dialogFactory.createMessageDialog("Error", arg.toString(), null).show();
```
azatsarynnyy: plz, use arg.getMessage() instead of arg.toString()
because toString() returns nothing interesting for the user

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -484,6 +484,8 @@ public void deleteCommand(@ApiParam("The id of the workspace")
         final WorkspaceImpl workspace = workspaceManager.getWorkspace(id);
         if (workspace.getConfig().getCommands().removeIf(command -> command.getName().equals(commandName))) {
             workspaceManager.updateWorkspace(id, workspace);
+        } else {
+            throw new NotFoundException(String.format("Command with name '%s' was not found in workspace with id %s", commandName, id));
```
benoitf: we may provide also the name of the workspace as the id is less "user friendly"

plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/command/edit/EditCommandsPresenter.java
mkuznyetsov: ok

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
mkuznyetsov: ok

mkuznyetsov:
selection_009

In case if unable to delete selected command (in this case, command was removed in other session)

skabashnyuk:
ок

mkuznyetsov:
Note that slashes and backslashes are still not allowed, there is a separate issue to handle that (CHE-1386)

PR #1528, opened 2016-06-18
CHE-1303: add .swf path location using ngClipProvider

olexii4:

@benoitf  WDYT?

```
https://github.com/zeroclipboard/ZeroClipboard/blob/master/docs/instructions.md#a-note-on-the-allowscriptaccess-option
-
ZeroClipboard was intentionally configured to not allow the SWF to be served from a secure domain (HTTPS) but scripted by an insecure domain (HTTP).
...
1.Serve the SWF over HTTP instead of HTTPS. If the page's protocol can vary (e.g. authorized/unauthorized, staging/production, etc.), you should include add the SWF with a relative protocol (//s3.amazonaws.com/blah/ZeroClipboard.swf) instead of an absolute protocol (https://s3.amazonaws.com/blah/ZeroClipboard.swf).
2.Serve the page over HTTPS instead of HTTP. If the page's protocol can vary, see the note on the previous option (1).
3. Fork ZeroClipboard and update "src/flash/ZeroClipboard.as" to call the allowInsecureDomain method as needed, then recompile the SWF with your custom changes.
-
```

### DIFF
--- a/dashboard/gulp/build.js
+++ b/dashboard/gulp/build.js
@@ -102,6 +102,12 @@ gulp.task('brandingassets', function () {
     .pipe(gulp.dest(conf.paths.dist + '/assets/branding/'));
 });
 
+gulp.task('zeroclipboardassets', function () {
+  return gulp.src($.mainBowerFiles().concat('bower_components/zeroclipboard/dist/**/*'))
+    .pipe($.filter('**/*.swf'))
+    .pipe($.flatten())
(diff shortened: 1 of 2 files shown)


benoitf:
@olexii4 is that this swf file is also available on production build ?

olexii4:
Yes it is.

selection_018

olexii4:
Yes it is
(I have updated)

olexii4:
selection_012

benoitf:
ok

PR #1532, opened 2016-06-20
1531: Parameterized Vagrantfile

creckord:
This pull request parameterizes the Che Vagrantfile in order to be able to configure the VM through environment parameters for vagrant up.

All parameters from the Vagrantfile can be set through environment variables. The network setup has been further parameterized to support DHCP network configuration and separate host and guest port configuration. FInally, an option for more verbose progress output has been added.

The behavior of the resulting Vagrantfile is unchanged if none of the environment variables in question have been set, i.e. a plain `vagrant up` will behave exactly as without these changes.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -8,12 +8,13 @@
 #   Codenvy, S.A. - initial API and implementation
 
 # Set to "<proto>://<user>:<pass>@<host>:<port>"
-$http_proxy  = ""
-$https_proxy = ""
-$no_proxy    = "localhost,127.0.0.1"
-$che_version = "nightly"
-$ip          = "192.168.28.100"
-$port        = 8080
+$http_proxy    = ENV['HTTP_PROXY'] || ""
+$https_proxy   = ENV['HTTPS_PROXY'] || ""
+$no_proxy      = ENV['NO_PROXY'] || "localhost,127.0.0.1"
+$che_version   = ENV['CHE_VERSION'] || "nightly"
+$ip            = ENV['CHE_IP'] || "192.168.28.100"
+$port          = (ENV['CHE_PORT'] || 8080).to_i
+$user_data     = ENV['CHE_DATA'] || "."
 
 Vagrant.configure(2) do |config|
   puts ("ECLIPSE CHE: VAGRANT INSTALLER")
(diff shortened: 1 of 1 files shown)


TylerJewell:
Nice improvements, especially the use of dhcp.  Two things:
1. What are these files for:

```
echo "${CHE_URL}" > /home/user/che/.che_url
echo "${MAPPED_PORT}" > /home/user/che/.che_host_port
```
1. Can you apply this basic template to the [ARTIK IDE custom assembly](https://github.com/codenvy/artik-ide/blob/master/Vagrantfile)?  We need a better way to abstract out the Dockerfile and Vagrantfile for custom assemblies, so that projects that are based off of the base Che can inherit changes made to Vagrantfile / Dockerfile base.  Until that time, any time we have major restructuring to these files in the core repository, we have to propagate those changes out.

creckord:
1. Those files are for accessing the final URL where Che will be available. `/home/user/che/.che_url` contains the URL for the virtual machine as reported by the "BOOTED AND REACHABLE" output. `/home/user/che/.che_host_port` contains the port on the host machine that is bound to the Che server. We could track those out-of-channel without these files if you prefer  (by keeping track of the port parameter and by parsing the provisioning output for the IP in case of dhcp). But this way, the information is ensured to be in sync with the latest provisioning run.
2. I'll submit a separate pull request for the ARTIK Vagrantfile. One question: This bit is intentionally silent:

> ```
> echo "----------------------------------------"
> echo "."
> echo "ARTIK IDE: DOWNLOADING ARTIK RUNTIME SDK"
> echo "           950MB: SILENT OUTPUT         "
> echo "."
> echo "----------------------------------------"
> docker pull codenvy/artik &>/dev/null
> ```

Should I just leave it this way, or also enable (normal and extended) progress here?

TylerJewell:
@creckord - let's go ahead and extend it so that it's consistent with normal / extended progress.

creckord:
@TylerJewell I've created ARTIK PR #69 to apply these changes.

creckord:
I confirm that I have signed the Eclipse CLA.

Not sure if there is an Eclipse URL showing the CLA status, but you can see here my committer status, which implies the CLA: https://projects.eclipse.org/user/362

TylerJewell:
Thanks @creckord for the great contribution.

PR #1534, opened 2016-06-21
Add Z suffix to override ws agent conf on RHEL systems

ghost:
Discovered by RedHat while working with OpenShift Che Assembly. It was impossible to override ws agent properties since .properties files never got correctly mounted into the container. Adding Z suffix, similar to all must have container mounts (terminal, ws-agent, projects), fixes the problem.


ghost:
@garagatyi please take a look

garagatyi:
LGTM

skabashnyuk:
ok

PR #1536, opened 2016-06-21
CODENVY-659: remove unnecessary data from workspace config

akurinnoy:
which is used in export workspace widget


benoitf:
why are we not using the export workspace API anymore ?

akurinnoy:
because this API provides full workspace config including unnecessary properties

benoitf:
if there are runtime properties, then server side should have option to disable it, I don't see why client should do tons of parsing to cleanup server response

ashumilova:
it's not only properties, links as well are not needed there

benoitf:
ok as workaround before proper way provided through CHE-1355

ashumilova:
Need to export workspace config and omit other workspace instance stuff.
For future - https://jira.codenvycorp.com/browse/CHE-1355

PR #1537, opened 2016-06-21
Temporary exclude MavenServerManagerTest …

vparfonov:
because it fail time-to-time and we don't know reason

@skabashnyuk @riuvshin


skabashnyuk:
ok

PR #1538, opened 2016-06-21
CHE-1349: Throw exception if nothing to commit with Jgit implementation

vinokurig:
Empty message appears when commit with unstaged changes performed.
@skabashnyuk please review

### DIFF
--- a/wsagent/che-core-api-git/src/test/java/org/eclipse/che/git/impl/CommitTest.java
+++ b/wsagent/che-core-api-git/src/test/java/org/eclipse/che/git/impl/CommitTest.java
@@ -27,9 +27,11 @@ import org.testng.annotations.Test;
 import java.io.File;
 import java.io.IOException;
 
+import static java.nio.file.Files.write;
 import static org.eclipse.che.dto.server.DtoFactory.newDto;
 import static org.eclipse.che.git.impl.GitTestUtil.addFile;
 import static org.eclipse.che.git.impl.GitTestUtil.cleanupTestRepo;
+import static org.eclipse.che.git.impl.GitTestUtil.connectToGitRepositoryWithContent;
 import static org.eclipse.che.git.impl.GitTestUtil.connectToInitializedGitRepository;
 import static org.testng.Assert.assertEquals;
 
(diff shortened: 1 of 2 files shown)


wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -485,6 +485,12 @@ public Revision commit(CommitRequest request) throws GitException {
             if (committer == null) {
                 throw new GitException("Committer can't be null");
             }
+            Status status = status(StatusFormat.SHORT);
```
skabashnyuk, line 492: Add some comments about conditions. It's not clear what is going on here.

skabashnyuk:
other ok. @sleshchenko ?

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
vinokurig, line 492: done

wsagent/che-core-api-git/src/test/java/org/eclipse/che/git/impl/CommitTest.java
```diff
@@ -108,4 +111,33 @@ public void testAmendCommit(GitConnectionFactory connectionFactory) throws GitEx
         assertEquals(revision.getMessage(), commitRequest.getMessage());
         assertEquals(beforeCount, afterCount);
     }
+
+    @Test(dataProvider = "GitConnectionFactory", dataProviderClass = org.eclipse.che.git.impl.GitConnectionFactoryProvider.class,
+          expectedExceptions = GitException.class)
```
sleshchenko, line 115: add expectedExceptionsMessageRegExp

sleshchenko:
other OK for me

wsagent/che-core-api-git/src/test/java/org/eclipse/che/git/impl/CommitTest.java
vinokurig, line 115: The problem is that Native Git and JGit are giving different messages in this testcase

sleshchenko, line 115: OK

PR #1547, opened 2016-06-22
CHE-1232. Add mapper for RuntimeException

RomanNikitenko:
The mapper will allow to get readable error messages on client side for RuntimeException
@vparfonov please review

### DIFF
--- a/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/RuntimeExceptionMapper.java
+++ b/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/RuntimeExceptionMapper.java
@@ -24,7 +24,10 @@ import javax.ws.rs.ext.Provider;
 import static com.google.common.base.Strings.isNullOrEmpty;
 
 /**
+ * Exception mapper that wrap RuntimeException message if it exists in to json.
+ *
  * @author Roman Nikitenko
+ * @author Sergii Kabashniuk
  */
 @Provider
 @Singleton
@@ -35,13 +38,16 @@ public class RuntimeExceptionMapper implements ExceptionMapper<RuntimeException>
     public Response toResponse(RuntimeException exception) {
         String errorMessage = exception.getLocalizedMessage();
         if (!isNullOrEmpty(errorMessage)) {
-            LOG.error(errorMessage);
+            LOG.error(errorMessage, exception);
             ServiceError serviceError = DtoFactory.newDto(ServiceError.class).withMessage(errorMessage);
             return Response.serverError()
                            .entity(DtoFactory.getInstance().toJson(serviceError))
                            .type(MediaType.APPLICATION_JSON)
                            .build();
+        } else {
+            LOG.error("RuntimeException occurred", exception);
         }
+
         return Response.serverError().build();
     }
 }
--- a/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/RuntimeExceptionMapper.java
+++ b/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/RuntimeExceptionMapper.java
@@ -31,7 +31,7 @@ public class RuntimeExceptionMapper implements ExceptionMapper<RuntimeException>
     public Response toResponse(RuntimeException exception) {
         String errorMessage = exception.getLocalizedMessage();
         if (!isNullOrEmpty(errorMessage)) {
-            ServiceError serviceError = DtoFactory.getInstance().createDto(ServiceError.class).withMessage(errorMessage);
+            ServiceError serviceError = DtoFactory.newDto(ServiceError.class).withMessage(errorMessage);
             return Response.serverError()
                            .entity(DtoFactory.getInstance().toJson(serviceError))
                            .type(MediaType.APPLICATION_JSON)
--- a/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/RuntimeExceptionMapper.java
+++ b/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/RuntimeExceptionMapper.java
@@ -36,7 +36,7 @@ public class RuntimeExceptionMapper implements ExceptionMapper<RuntimeException>
 
     @Override
     public Response toResponse(RuntimeException exception) {
-        String errorMessage = exception.getLocalizedMessage();
+        final String errorMessage = exception.getLocalizedMessage();
         if (!isNullOrEmpty(errorMessage)) {
             LOG.error(errorMessage, exception);
             ServiceError serviceError = DtoFactory.newDto(ServiceError.class).withMessage(errorMessage);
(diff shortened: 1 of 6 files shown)


benoitf:
do we have a test checking these runtime exceptions are correctly handle (so that a mapper does the job) ?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/RuntimeExceptionMapper.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * @author Roman Nikitenko
+ */
+@Provider
+@Singleton
+public class RuntimeExceptionMapper implements ExceptionMapper<RuntimeException> {
+    @Override
+    public Response toResponse(RuntimeException exception) {
+        String errorMessage = exception.getMessage();
```
garagatyi: Findbug usually recommends usage of `getlocalizedmessage` instead of `getMessage`

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/RuntimeExceptionMapper.java
```diff
@@ -0,0 +1,43 @@
...
+        if (!isNullOrEmpty(errorMessage)) {
+            DtoFactory dtoFactory = DtoFactory.getInstance();
+            ServiceError serviceError = dtoFactory.createDto(ServiceError.class).withMessage(errorMessage);
```
garagatyi: it is more convinient to use static DtoFactory.newDto()

garagatyi:
@benoitf I suppose it is hard to test such thing because it require Everrest. And to make it really reliable it should launch whole Che server.

garagatyi:
@skabashnyuk Please take a look

benoitf:
@garagatyi AFAIK EverrestJetty.class listener allow to test REST components and we already test some filters, REST Service, etc so I don't see the issue of being hard to be tested ?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/RuntimeExceptionMapper.java
RomanNikitenko: thank you, fixed

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/RuntimeExceptionMapper.java
RomanNikitenko: fixed

skabashnyuk:
EverrestJetty can handle that I think.

garagatyi:
LGTM

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/RuntimeExceptionMapper.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * @author Roman Nikitenko
+ */
+@Provider
+@Singleton
+public class RuntimeExceptionMapper implements ExceptionMapper<RuntimeException> {
+    @Override
+    public Response toResponse(RuntimeException exception) {
+        String errorMessage = exception.getLocalizedMessage();
```
sleshchenko: Maybe it would be better to log this error because we should not throw RuntimeException in services at all and usually it is possible in case of bug
@garagatyi @skabashnyuk @vparfonov WDYT?

RomanNikitenko: @sleshchenko @vparfonov Done!

benoitf:
I still don't see this server code being tested ?

skabashnyuk:
-1 to merge it without test

RomanNikitenko:
@benoitf @skabashnyuk I don't merge it. I need to finish my current issue, so - don't worry - I will do it :-)

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/RuntimeExceptionMapper.java
```diff
@@ -0,0 +1,47 @@
...
+@Provider
+@Singleton
+public class RuntimeExceptionMapper implements ExceptionMapper<RuntimeException> {
+    private static final Logger LOG = LoggerFactory.getLogger(RuntimeExceptionMapper.class);
+
+    @Override
+    public Response toResponse(RuntimeException exception) {
+        String errorMessage = exception.getLocalizedMessage();
+        if (!isNullOrEmpty(errorMessage)) {
+            LOG.error(errorMessage);
```
sleshchenko: I would say that `LOG.warn(errorMessage, exception);` would be better. And put it before check

skabashnyuk: I would say that `LOG.error(errorMessage, exception)`

garagatyi: +1 for log error with stacktrace. RuntimeException here means bug in code, so we should findout how to fix that ASAP and stacktrace will definitely help

skabashnyuk:
@benoitf  FYI server code is tested

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/RuntimeExceptionMapper.java
```diff
@@ -0,0 +1,53 @@
...
+ * @author Sergii Kabashniuk
+ */
+@Provider
+@Singleton
+public class RuntimeExceptionMapper implements ExceptionMapper<RuntimeException> {
+    private static final Logger LOG = LoggerFactory.getLogger(RuntimeExceptionMapper.class);
+
+    @Override
+    public Response toResponse(RuntimeException exception) {
+        String errorMessage = exception.getLocalizedMessage();
```
mshaposhnik: final ?

garagatyi:
LGTM

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/RuntimeExceptionMapper.java
```diff
@@ -0,0 +1,53 @@
...
+            ServiceError serviceError = DtoFactory.newDto(ServiceError.class).withMessage(errorMessage);
+            return Response.serverError()
+                           .entity(DtoFactory.getInstance().toJson(serviceError))
+                           .type(MediaType.APPLICATION_JSON)
+                           .build();
+        } else {
+            LOG.error("RuntimeException occurred", exception);
+        }
+
+        return Response.serverError().build();
```
sleshchenko: Mb it would be better to send ServiceError with default message to avoid writing client side's logic for expecting service error or empty body ?

RomanNikitenko:
I've updated pull request
User will get error message in the next format when RuntimeException occured:
runtimeexception

skabashnyuk:
ok

vparfonov:
ok

garagatyi:
@skabashnyuk @vparfonov @riuvshin Can we merge it now?

skabashnyuk:
ok

PR #1548, opened 2016-06-22
CODENVY-672: Add invalidation of token when the refresh failed

akorneta:
We have a problem with refreshing of expired msft vsts oauth tokens to solve this problem we suggest to invalidate expired tokens if it can't be refreshed.
@skabashnyuk


skabashnyuk:
ok

mshaposhnik:
Ok

Issue #1482, opened 2016-06-11
I cannot create a fully functional CHE workspace using Docker beta for mac


l0rd:
## The issue

The configuration is illustrated in the following diagram :

![image](https://cloud.githubusercontent.com/assets/606959/15984335/73b12e00-2fc7-11e6-970c-4297a0deb045.png)

Both Che server and its workspaces are running in a Docker container. Both container are running in a Mac [xhyve](https://github.com/mist64/xhyve/) VM.

With Docker for Mac beta, wsagent URL is `localhost` (URL1) when connecting from outside the VM (i.e. the browser) whereas wsmaster need to use the VM IP address to connect to the wsagent (URL2).

However I could not manage to configure two distinct values for URL1 and URL2. Both are set through the same property: `machine.docker.local_node_host`. 
## How to reproduce the issue
1. Install Docker for Mac Beta
2. Retrieve the IP of the xhyve VM. To get a shell into the VM `screen ~/Library/Containers/com.docker.docker/Data/com.docker.driver.amd64-linux/tty`
   and to retrieve the IP `ip a show eth0 | grep 'inet ' |  cut -d/  -f1 | awk '{ print $2}'`
3. Create `che.properties` file:
   
   ```
   export DOCKER_VM_IP=<put the IP here>
   echo "machine.docker.local_node_host=$DOCKER_VM_IP" > che.properties
   ```
4. Run docker from the `che.properties` folder to start a shell in a new Che container:
   
   ``` bash
   docker run --name=che \
          -v /home/user/che/lib:/home/user/che/lib-copy \
          -v /home/user/che/workspaces:/home/user/che/workspaces \
          -v /home/user/che/storage:/home/user/che/storage \
          -v /var/run/docker.sock:/var/run/docker.sock \
          -v $(pwd)/che.properties:/home/user/.che/che.properties \
          -p 8080:8080 \
          -ti \
          --entrypoint=bash \
          codenvy/che:nightly
   # Retrieve the container IP address and add a property to che.properties
   export CHE_SERVER_IP=$(ip address show eth0 | grep 'inet ' |  cut -d/  -f1 | awk '{ print $2}')
   echo "machine.docker.che_api.endpoint=http://${CHE_SERVER_IP}:8080/wsmaster/api" >> /home/user/.che/che.properties
   # Add user to the group with gid=50 (that's the VM docker group)
   sudo usermod -aG 50 user
   sudo su -l user
[...]

Issue #745, opened 2016-03-18
Error starting workspace, WorkspaceManager error


boban74:
**_1 Upvote**_ Hi,

currently I am trying to get Eclipse Che running on Ubuntu 16.04 but with no luck. Starting an Workspace in Che ends up in a exception 

`2016-03-18 13:04:34,490[kspaceManager-0]  [ERROR] [o.e.c.a.w.s.WorkspaceManager 610]    - Failed writing 8192 bytes
org.eclipse.che.api.machine.server.exception.MachineException: Failed writing 8192 bytes
    at org.eclipse.che.api.machine.server.MachineManager.createInstance(MachineManager.java:396) ~[che-core-api-machine-4.0.0-RC11.jar:4.0.0-RC11]
    at org.eclipse.che.api.machine.server.MachineManager.createMachine(MachineManager.java:333) ~[che-core-api-machine-4.0.0-RC11.jar:4.0.0-RC11]
    at org.eclipse.che.api.machine.server.MachineManager.createMachineSync(MachineManager.java:167) ~[che-core-api-machine-4.0.0-RC11.jar:4.0.0-RC11]
    at org.eclipse.che.api.workspace.server.RuntimeWorkspaceRegistry.createMachine(RuntimeWorkspaceRegistry.java:421) ~[che-core-api-workspace-4.0.0-RC11.jar:4.0.0-RC11]
    at org.eclipse.che.api.workspace.server.RuntimeWorkspaceRegistry.startEnvironment(RuntimeWorkspaceRegistry.java:320) ~[che-core-api-workspace-4.0.0-RC11.jar:4.0.0-RC11]
    at org.eclipse.che.api.workspace.server.RuntimeWorkspaceRegistry.start(RuntimeWorkspaceRegistry.java:146) ~[che-core-api-workspace-4.0.0-RC11.jar:4.0.0-RC11]
    at org.eclipse.che.api.workspace.server.WorkspaceManager.performSyncStart(WorkspaceManager.java:637) ~[che-core-api-workspace-4.0.0-RC11.jar:4.0.0-RC11]
    at org.eclipse.che.api.workspace.server.WorkspaceManager.lambda$performAsyncStart$2(WorkspaceManager.java:608) ~[che-core-api-workspace-4.0.0-RC11.jar:4.0.0-RC11]
[...]

PR #1550, opened 2016-06-22
Add HTTP Host header to support Docker v1.12

l0rd:
`Host` is a mandatory header in [HTTP 1.1 specifications](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html). In previous versions of Docker that header wasn't required but release v1.12 use [golang 1.6.2](https://github.com/docker/docker/pull/22840) that makes this header mandatory. 

Therefore Docker v1.12 daemon doesn't accept HTTP 1.1 requests with no `Host` header set (you can find some references to the discussions [here](https://github.com/docker/swarm/issues/2332)).

This works:
`echo -e "GET /containers/json HTTP/1.1\r\nHost: \r\n\r\n"  | netcat -U /var/run/docker.sock`

This doesn't work:
`echo -e "GET /containers/json HTTP/1.1\r\n\r\n"  | netcat -U /var/run/docker.sock`

As a consequence in Che we had this nasty error message whenever we tried to create a new workspace: 

> org.eclipse.che.api.machine.server.exception.MachineException: Failed writing 8192 bytes

This PR is related to #1482 and #745.

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/connection/TcpConnection.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/connection/TcpConnection.java
@@ -67,6 +67,12 @@ public class TcpConnection extends DockerConnection {
         for (Pair<String, ?> header : headers) {
             connection.setRequestProperty(header.first, String.valueOf(header.second));
         }
+        String host = url.getHost();
+        if (url.getPort() != -1) {
+            host += ":" + Integer.toString(url.getPort());
+        }
+        // Host header is mandatory in HTTP 1.1
+        connection.setRequestProperty("Host", host);
         if (entity != null) {
             connection.setDoOutput(true);
             try (OutputStream output = connection.getOutputStream()) {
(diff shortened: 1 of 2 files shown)


TylerJewell:
@mmorhun @garagatyi

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/connection/TcpConnection.java
```diff
@@ -67,6 +67,7 @@ protected DockerResponse request(String method, String path, String query, List<
         for (Pair<String, ?> header : headers) {
             connection.setRequestProperty(header.first, String.valueOf(header.second));
         }
+        connection.setRequestProperty("Host", url.toString());
```
garagatyi: In that case not only host will be put in that header but host + path +query. 
@l0rd Do you think it is OK or only host+port should be put in that header.

l0rd: @garagatyi you are absolutely right. We should set it to host+port only. Although I've test it and it worked, path+query should not be there. I'll fix that.

garagatyi: We still can use empty value. it is valid value due to the spec.

garagatyi:
LGTM

garagatyi:
@skabashnyuk Please take a look

skabashnyuk:
ok

mmorhun:
LGTM

riuvshin:
OK

PR #1552, opened 2016-06-22
CODENVY-630: save a value of the library folder

svor:
Save value of the library folder for the javac project type. This value can change by using configuration wizard.


benoitf:
do you think src/test/resources/projects/project/lib/a.jar could be generated during the test build instead of adding a binary into git repository ?

svor:
@benoitf  this is rubbish :worried: 
I've deleted it. Thanks for this

benoitf:
@svor ok thx

vparfonov:
ok

PR #1553, opened 2016-06-22
CHE-957: fixed a problem with updating content of the compare widget

svor:
Refresh the content of the compare widget when files which are compared are changed


vparfonov:
ok

PR #1559, opened 2016-06-23
CHE-1346: fix project detail view on starting workspace

akurinnoy:


dashboard/src/app/projects/project-details/project-details.controller.js
```diff
@@ -59,22 +59,46 @@ export class ProjectDetailsCtrl {
   }
 
   fetchProjectDetails() {
-    this.projectService = this.cheAPI.getWorkspace().getWorkspaceAgent(this.workspaceId).getProject();
+    if (this.workspace.status === 'STARTING' || this.workspace.status === 'RUNNING') {
```
olexii4: We can use next:

```
    if (this.workspace.status !== 'STARTING' && this.workspace.status !== 'RUNNING') {
      this.loading = false;
      this.noWorkspaceRuntime = true;
      return;
    }
```

WDYT?

akurinnoy: it make sense, will do

PR #1571, opened 2016-06-23
Allow character based key bindings 

AlexTugarev:
We need to be add (or replace) editor-specific key bindings not only for key codes but also for characters.

Therefore I've extended `OrionKeyStrokeOverlay`, `OrionEditorWidget`, `Action`, and `KeyBinding` to be able to add/repalce character based key bindings.

Signed-off-by: Alex Tugarev alex.tugarev@typefox.io

### DIFF
--- a/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/keymap/KeyBinding.java
+++ b/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/keymap/KeyBinding.java
@@ -22,8 +22,12 @@ public class KeyBinding {
     private final boolean alt;
     /** Is the cmd key pressed? */
     private final boolean cmd;
-    /** The key code. */
-    private final int keycode;
+    /** The key code of this key binding. */
+    private final Integer keyCodeNumber;
+    /** The character of this key binding. */
+    private final String character;
+    /** The key event type. */
+    private final String type;
     /** The action taken on the key binding. */
     private final KeyBindingAction action;
 
@@ -33,9 +37,23 @@ public class KeyBinding {
         this.shift = shift;
         this.alt = alt;
         this.cmd = cmd;
-        this.keycode = keycode;
+        this.keyCodeNumber = keycode;
+        this.character = null;
         this.action = action;
+        this.type = "keydown";
     }
+    
+    public KeyBinding(final boolean control, final boolean shift, final boolean alt,
+            final boolean cmd, final char character, final KeyBindingAction action) {
+    	this.control = control;
+		this.shift = shift;
+		this.alt = alt;
+		this.cmd = cmd;
+		this.keyCodeNumber = null;
+		this.character = String.valueOf(character);
+		this.action = action;
+		this.type = "keypress";
+	}
 
     /**
      * Whether the control key is hold during the key binding.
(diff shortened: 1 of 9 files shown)


svenefftinge:
Good job Alex! Also returning boolean in the Action was important to control whether a certain action 'consumed' the event or not.

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/keymap/KeyBinding.java
```diff
@@ -70,12 +88,36 @@ public boolean isCmd() {
...
+    public Integer getKeyCodeNumber() {
+        return this.keyCodeNumber;
     }
+    
+    /**
+     * Returns the keycode of the key binding.
+     * @return the keycode
+     */
+    public String getCharacter() {
+		return this.character;
```
vzhukovs: Code formatting problem.

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/keymap/KeyBinding.java
```diff
@@ -70,12 +88,36 @@ public boolean isCmd() {
...
+    public boolean isCharacterBinding() {
+    	return getCharacter() != null;
+    }
+    
+    /**
+     * Returns the event type of the key binding.
+     * @return the event type
+     */
+    public String getType() {
+		return type;
```
vzhukovs: Code formatting problem.

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/keymap/KeyBindingAction.java
```diff
@@ -16,5 +16,5 @@
 public interface KeyBindingAction {
 
     /** The triggered action. */
-    void action();
+    boolean action();
```
vzhukovs, line 19: What's the benefit of returning boolean value in this case?

AlexTugarev, line 19: That was the tricky part. Have a look at the call chain starting with `OrionEditorWidget.addKeyBinding(KeyBinding, String)` where the `OrionTextViewOverlay.setAction(String, Action, String)` is called, which will then create a `function` delegate in JS to return `Action.onAction()`'s boolean return value at [l168](https://github.com/TypeFox/che/blob/aa5a9368165db032c8568626c41712b11187dd86/plugins/plugin-orion/che-plugin-orion-editor/src/main/java/org/eclipse/che/ide/editor/orion/client/jso/OrionTextViewOverlay.java#L168). 

Passing the boolean result along this call chain was the key to override a bound `autoPairAngleBracket` action in order to _disable_ the auto insert of the closing bracket. Basically, if the action always returns true, the key event handler of `orion.editor.TextView` won't proceed with the default event handling, which makes sure the opening bracket is inserted at all. The overriding `KeyBindingAction` may return `false` to re-enable the default handling.

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/keymap/KeyBinding.java
AlexTugarev: Resolved.

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/keymap/KeyBinding.java
AlexTugarev: Resolved.

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/keymap/KeyBindingAction.java
vzhukovs, line 19: @AlexTugarev thx for the explanation

PR #1574, opened 2016-06-24
CHE-1277 Move machines output to project perpective

vitaliy-guliy:


ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/workspace/event/WorkspaceStartingEvent.java
```diff
@@ -0,0 +1,55 @@
...
+     */
+    public interface Handler extends EventHandler {
+
+        void onWorkspaceStarting(WorkspaceStartingEvent event);
+
+    }
+
+    public static final GwtEvent.Type<WorkspaceStartingEvent.Handler> TYPE = new GwtEvent.Type<>();
+
+    private final WorkspaceDto workspace;
```
azatsarynnyy: What was the motivation for using DTO object here instead of an appropriate model object [Workspace](https://github.com/eclipse/che/blob/master/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/Workspace.java)?

vitaliy-guliy: No special motivation, just the same as in WorkspaceStartedEvent and WorkspaceStoppedEvent

azatsarynnyy: We should use DTO for transport **only**.
And model interfaces in other places like this event class.

vitaliy-guliy: Done

ashumilova:
+1

vparfonov:
ok

PR #1575, opened 2016-06-24
CHE-1359: fix NPE in case recipe set as content but not location URL

vparfonov:
Mark Machine class form ide module as Deprecated. It must be reworked


skabashnyuk:
test?

vparfonov:
Add test in case script loaded as content, then need to load it by URL not done for now. Hard to implement it add TODO for adding it in future

PR #1576, opened 2016-06-24
Improvements to set the right API endpoint URL  

l0rd:
This PR introduce two improvements and is related to #1482: 
- The endpoint URL can be set using an environment variable
- When Che server is running inside a container the alias in wsagents /etc/hosts is set its internal IP

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
@@ -37,9 +37,14 @@ import static java.io.File.separatorChar;
  */
 public class DockerConnectorConfiguration {
     /**
-     * Docker bridge name on Linux.
+     * Default Docker bridge name on Linux.
      */
     protected static final String BRIDGE_LINUX_INTERFACE_NAME = "docker0";
+
+    /**
+     * Default network interface name (Linux system).
+     */
+    protected static final String DEFAULT_LINUX_INTERFACE_NAME = "eth0";
 
     /**
      * Default ip of docker host (Linux system).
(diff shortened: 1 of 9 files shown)


plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
```diff
@@ -44,6 +44,11 @@
     /**
      * Default ip of docker host (Linux system).
```
benoitf: comment becomes invalid

skabashnyuk:
ci-build

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java
l0rd: Thank you @benoitf. I've updated the commit.

TylerJewell:
@garagatyi

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ext/provider/ApiEndpointEnvVariableProvider.java
```diff
@@ -30,6 +32,11 @@
 
     @Override
     public String get() {
+        String apiEndpointEnvVar = System.getenv(DockerInstanceRuntimeInfo.API_ENDPOINT_URL_VARIABLE);
```
garagatyi, line 35: @l0rd Can you describe why you want to add that functionality and how it will be used

l0rd, line 35: Hello @garagatyi, sorry for answering late to your questions but I became father for the second time yesterday and I'm taking a few days off (not completely off though 😄)

By the way to aswer your question:
- Fonctionality: Add the ability to set the property `machine.docker.che_api.endpoint` using an environment variable (`CHE_API_ENDPOINT`). 
- How it will be used: `export CHE_API_ENDPOINT=<...> && bin/che.sh run` or `docker run -e CHE_API_ENDPOINT="<...>" ... codenvy/che`

I think you suggest me to set that variable to resolve problems running Che on Docker for Mac beta. I tried to set that variable but Che didn't parse it. So I've fixed it. It's useful in situations where Che fails retrieving the server IP.

garagatyi, line 35: Congratulations!

garagatyi:
LGTM

garagatyi:
@benoitf Please review

benoitf:
ok

PR #1577, opened 2016-06-24
CHE-1376. Remove 'Stop dev machine' action from 'Machine' menu

RomanNikitenko:
Before:
before
After:
after

@vparfonov please review


vparfonov:
ok

PR #1579, opened 2016-06-24
CHE-1276: Add exception handler during creation machine from snapshot

akorneta:
When docker image is not exist and snapshot is not null then in case of creating   machine instance from this config, docker client will throw an exception, to avoid this situation we should handle this error. It is proposed to create machine instance  form incoming config(recipe).
@skabashnyuk @vparfonov @garagatyi

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
@@ -21,6 +21,7 @@ import org.eclipse.che.commons.annotation.Nullable;
 import org.eclipse.che.commons.json.JsonHelper;
 import org.eclipse.che.commons.json.JsonNameConvention;
 import org.eclipse.che.commons.json.JsonParseException;
+import org.eclipse.che.commons.lang.IoUtil;
 import org.eclipse.che.commons.lang.Pair;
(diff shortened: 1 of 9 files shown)


garagatyi:
In general I'm OK with the solution, but I think you can improve & simplify it. 
Instead of cloning the whole origin MachineConfig you can clone just origin source. Then if there is an exception and origin source is not null you replace source in already created machine instance and try to start machine creation again. 
I believe that in that case this PR will introduce less changes and logic will be clearer.

Also it looks like MachineRegistry.addMachine() should throw ConflictException in that case. Can you check that?

vparfonov:
ok

garagatyi:
LGTM
@skabashnyuk Please take a look.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -1242,21 +1245,27 @@ private void removeImage(final RemoveImageParams params,URI dockerDaemonUri) thr
...
 
     /**
      * Tag the docker image into a repository.
      *
      * @throws IOException
-     *          when a problem occurs with docker api calls
+     *         when a problem occurs with docker api calls
      */
-    public void tag(final TagParams params) throws IOException {
+    public void tag(final TagParams params) throws ImageNotFoundException,
```
evoevodin: Document ImageNotFoundException

voievodin:
ok :+1:

PR #1582, opened 2016-06-27
CHE-699: handle IDE initialization failed error

ashumilova:


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/client/BootstrapController.java
```diff
@@ -218,7 +218,7 @@ public void execute() {
      * Sends a message to the parent frame to inform that IDE application has been loaded.
      */
     private native void notifyIDELoaded() /*-{
```
benoitf: method is still named notifyIDELoaded if we send show-ide ?

ashumilova: fixed

vitaliy-guliy:
LGTM

PR #1586, opened 2016-06-27
CODENVY-703; do not show terminal by default in factories

mshaposhnik:
@vparfonov @vitaliy-guliy

### DIFF
--- a/plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/MachineExtension.java
+++ b/plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/MachineExtension.java
@@ -94,15 +94,6 @@ public class MachineExtension {
             @Override
             public void onWsAgentStarted(WsAgentStateEvent event) {
                 machinePortProvider.get();
-
-                /**
-                 * There is a bug in perspective management and it's unable to add Consoles part in
-                 * OperationsPerspective and ProjectPerspective directly. Following code resolves the issue.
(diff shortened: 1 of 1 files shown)


vitaliy-guliy:
Great!

riuvshin:
Super! Thanks Maxim!

plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/MachineExtension.java
```diff
@@ -94,15 +94,6 @@ public MachineExtension(MachineResources machineResources,
             @Override
             public void onWsAgentStarted(WsAgentStateEvent event) {
                 machinePortProvider.get();
```
sleshchenko, line 96: Do you know why we need invoke provider here?

mshaposhnik, line 96: No (

sleshchenko, line 96: Looks like we can remove it. @vparfonov WDYT?

vparfonov, line 96: Yep, you are right. Looks like we can remove it.

vparfonov:
ok

PR #1587, opened 2016-06-27
Use of TypeScript instead of ES6

benoitf:
Change-Id: I023d80fc736867c3984fb715c0eeb85098d7d168


skabashnyuk:
Can you talk a bit why we are doing this?

benoitf:
it's only a poc ;-)

benoitf:
maybe I shouldn't have created a PR for that, I had this for a long time in my local branches

benoitf:
and to answer the question of : "why we could use TypeScript" ?
--> main reason is to stop to break UD each time core api is updated with no backward compliance

then we could reuse DTO objects (turned into .js code) and then it could throw error at compilation time

TylerJewell:
@ashumilova @benoitf @skabashnyuk - can we get an update on the status of this POC please?  I have started to review the differences between TypeScript and ES6 and find that there are many encouraging qualities of TypeScript - particularly its typing and class management.  If we are going to go forward, we should define specification and have clarity on definition of done.  If this is just a POC we should probably remove the PR.

ashumilova (approved):

PR #1588, opened 2016-06-27
CHE-1393: Fix Tests according to changes in latest Git

vinokurig:
Some tests from Git TCK test-suite failed if Git is updated to latest (2.9.0) version
@skabashnyuk please review

### DIFF
--- a/wsagent/che-core-api-git/src/test/java/org/eclipse/che/git/impl/PullTest.java
+++ b/wsagent/che-core-api-git/src/test/java/org/eclipse/che/git/impl/PullTest.java
@@ -115,15 +115,12 @@ public class PullTest {
         connection.add(newDto(AddRequest.class).withFilepattern(Arrays.asList(".")));
         connection.commit(newDto(CommitRequest.class).withMessage("remote test"));
 
-        GitConnection connection2 = connectToInitializedGitRepository(connectionFactory, remoteRepo);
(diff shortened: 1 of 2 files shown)


skabashnyuk:
Can you explain what it was happening? And what is your solution?

vinokurig:
As for PullTest: 
This is from Git change-log:
- "git merge" used to allow merging two branches that have no common
  base by default, which led to a brand new history of an existing
  project created and then get pulled by an unsuspecting maintainer,
  which allowed an unnecessary parallel history merged into the
  existing project.  The command has been taught not to allow this by
  default, with an escape hatch "--allow-unrelated-histories" option
  to be used in a rare event that merges histories of two projects
  that started their lives independently.
  - "git pull" has been taught to pass the "--allow-unrelated-histories"
    option to underlying "git merge".

So I have changed the repository from which we perform pull to have the same base as in repository from which we pull.
Also I changed `RefSpec` to common git refspec format

As for RemoteUpdateTest:
testDeleteUrl was not correct because `remoteUpdate(addRequest)` in "given" section was not performed so the expacted config key was not set. The test did not check that the config key is removed, Actualy it chacked that initial config key exist. New Git throwed an error when the test tried to remove nonexistent config key.
So I have fixed the behavior of the test

skabashnyuk:
ok

PR #1600, opened 2016-06-29
[Not for merge] CHE-1368: rework machine's output/statuses events subscribing

garagatyi:
This PR is a part of contribution that change the way of streaming logs/statuses of machines of active environment in a workspace. This PR introduces changes on server side and require changes in UD and IDE clients to be merged in master.

Rework output/status events of machine to be per workspace as a
part of implementation of new multi-machine environments.
Get rid of usage of machine manager in workspace service
Refactor code of machine manager.

### DIFF
--- a/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/MessageConsumer.java
+++ b/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/MessageConsumer.java
@@ -14,14 +14,10 @@ import java.io.Closeable;
 import java.io.IOException;
 
 /**
- * Consumes text line by line for analysing, writing, storing, etc.
+ * Consumes messages one by one for analysing, writing, storing, etc.
  *
- * @author andrew00x
- * @see AbstractLineConsumer
+ * @author Alexander Garagatyi
  */
-public interface LineConsumer extends Closeable {
-    /** Consumes single line. */
-    void writeLine(String line) throws IOException;
-
-    LineConsumer DEV_NULL = new AbstractLineConsumer() {};
+public interface MessageConsumer<T> extends Closeable {
+    void consume(T message) throws IOException;
 }
--- a/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/MachineLogMessage.java
+++ b/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/MachineLogMessage.java
@@ -8,20 +8,21 @@
  * Contributors:
  *   Codenvy, S.A. - initial API and implementation
  *******************************************************************************/
-package org.eclipse.che.api.core.util;
-
-import java.io.Closeable;
-import java.io.IOException;
+package org.eclipse.che.api.core.model.machine;
 
 /**
- * Consumes text line by line for analysing, writing, storing, etc.
+ * Represents log message from machine
  *
- * @author andrew00x
- * @see AbstractLineConsumer
+ * @author Alexander Garagatyi
  */
-public interface LineConsumer extends Closeable {
-    /** Consumes single line. */
-    void writeLine(String line) throws IOException;
+public interface MachineLogMessage {
+    /**
+     * Content of log message
+     */
+    String getContent();
 
-    LineConsumer DEV_NULL = new AbstractLineConsumer() {};
+    /**
+     * Machine name
+     */
+    String getMachineName();
 }
--- a/wsmaster/che-core-api-machine-shared/src/main/java/org/eclipse/che/api/machine/shared/dto/MachineLogMessageDto.java
+++ b/wsmaster/che-core-api-machine-shared/src/main/java/org/eclipse/che/api/machine/shared/dto/MachineLogMessageDto.java
@@ -8,20 +8,21 @@
  * Contributors:
  *   Codenvy, S.A. - initial API and implementation
  *******************************************************************************/
-package org.eclipse.che.api.core.util;
+package org.eclipse.che.api.machine.shared.dto;
 
-import java.io.Closeable;
-import java.io.IOException;
+import org.eclipse.che.api.core.model.machine.MachineLogMessage;
+import org.eclipse.che.dto.shared.DTO;
 
 /**
- * Consumes text line by line for analysing, writing, storing, etc.
- *
- * @author andrew00x
- * @see AbstractLineConsumer
+ * @author Alexander Garagatyi
  */
-public interface LineConsumer extends Closeable {
-    /** Consumes single line. */
-    void writeLine(String line) throws IOException;
+@DTO
+public interface MachineLogMessageDto extends MachineLogMessage {
+    void setContent(String content);
 
-    LineConsumer DEV_NULL = new AbstractLineConsumer() {};
+    MachineLogMessageDto withContent(String content);
+
+    void setMachineName(String machineName);
+
+    MachineLogMessageDto withMachineName(String machineName);
 }
--- a/wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
+++ b/wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
@@ -14,6 +14,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 
+import org.eclipse.che.api.core.ApiException;
 import org.eclipse.che.api.core.BadRequestException;
 import org.eclipse.che.api.core.ConflictException;
 import org.eclipse.che.api.core.ForbiddenException;
@@ -135,6 +136,8 @@ public class MachineManager {
      *         id of the workspace the created machine will belong to
      * @param environmentName
      *         environment name the created machine will belongs to
+     * @param outputConsumer
+     *         output consumer of machine
      * @return new machine
      * @throws NotFoundException
      *         if machine type from recipe is unsupported
@@ -153,7 +156,8 @@ public class MachineManager {
      */
     public MachineImpl createMachineSync(MachineConfig machineConfig,
                                          final String workspaceId,
-                                         final String environmentName)
+                                         final String environmentName,
+                                         LineConsumer outputConsumer)
             throws NotFoundException,
                    SnapshotException,
                    ConflictException,
@@ -167,7 +171,8 @@ public class MachineManager {
                                                   workspaceId,
                                                   environmentName,
                                                   this::createInstance,
-                                                  null);
+                                                  null,
+                                                  outputConsumer);
         LOG.info("Machine [ws = {}: env = {}: machine = {}] was successfully created, its id is '{}'",
                  workspaceId,
                  environmentName,
@@ -186,6 +191,8 @@ public class MachineManager {
      *         workspace id
      * @param envName
      *         name of environment
+     * @param outputConsumer
+     *         output consumer of machine
      * @return machine instance
      * @throws NotFoundException
      *         when snapshot doesn't exist
@@ -198,11 +205,15 @@ public class MachineManager {
      * @throws BadRequestException
      *         when either machineConfig or workspace id, or environment name is not valid
      */
-    public MachineImpl recoverMachine(MachineConfig machineConfig, String workspaceId, String envName) throws NotFoundException,
-                                                                                                              SnapshotException,
-                                                                                                              MachineException,
-                                                                                                              ConflictException,
-                                                                                                              BadRequestException {
+    public MachineImpl recoverMachine(MachineConfig machineConfig,
+                                      String workspaceId,
+                                      String envName,
+                                      LineConsumer outputConsumer) throws NotFoundException,
+                                                                          SnapshotException,
+                                                                          MachineException,
+                                                                          ConflictException,
+                                                                          BadRequestException {
+
         final SnapshotImpl snapshot = snapshotDao.getSnapshot(workspaceId, envName, machineConfig.getName());
 
         LOG.info("Recovering machine [ws = {}: env = {}: machine = {}] from snapshot", workspaceId, envName, machineConfig.getName());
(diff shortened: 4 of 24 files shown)


garagatyi:
@skabashnyuk @evoevodin @vparfonov @ashumilova @benoitf 
Please review.

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/MachineLogMessage.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.api.core.model.machine;
+
+/**
+ * Represents log message from machine
+ *
+ * author Alexander Garagatyi
```
voievodin: typo, missed `@` symbol

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/MachineLogMessage.java
```diff
@@ -0,0 +1,28 @@
...
+public interface MachineLogMessage {
+    /**
+     * Content of log message
+     */
+    String getContent();
+
+    /**
+     * Machine name
+     */
+    String getMachine();
```
voievodin: i'd rename it to `getMachineName()` as we have alot of places where `getMachine` returns Machine instance

wsmaster/che-core-api-machine-shared/src/main/java/org/eclipse/che/api/machine/shared/dto/MachineLogMessageDto.java
```diff
@@ -0,0 +1,28 @@
...
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.api.machine.shared.dto;
+
+import org.eclipse.che.api.core.model.machine.MachineLogMessage;
+import org.eclipse.che.dto.shared.DTO;
+
+/**
+ * author Alexander Garagatyi
```
voievodin: typo, again missed `@`

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
```diff
@@ -326,37 +343,42 @@ private MachineImpl createMachine(MachineConfigImpl machineConfig,
...
 
         createMachineLogsDir(machineId);
-        final LineConsumer machineLogger = getMachineLogger(machineId, getMachineChannels(machine.getConfig().getName(),
-                                                                                          machine.getWorkspaceId(),
-                                                                                          machine.getEnvName())
-                .getOutput());
+        final LineConsumer machineLogger = getMachineLogger(machineId, outputConsumer);
 
         try {
+            machineRegistry.addMachine(machine);
```
benoitf, line 352: should we add machine in machineRegistry only in case of success ? (I mean adding it when instance is successfully created, and not before and then check exception and remove)

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
```diff
@@ -326,37 +343,42 @@ private MachineImpl createMachine(MachineConfigImpl machineConfig,
...
                 if (ex.getCause() instanceof SourceNotFoundException) {
-                    final LineConsumer logger = getMachineLogger(machineId,
-                                                                 getMachineChannels(machine.getConfig().getName(),
-                                                                                    machine.getWorkspaceId(),
-                                                                                    machine.getEnvName()).getOutput());
                     LOG.error("Image of snapshot for machine " + machineConfig.getName() + " not found. " +
                               "Machine will be created from origin source");
                     machine.getConfig().setSource(sourceCopy);
+                    try {
+                        machineRegistry.remove(machineId);
```
voievodin, line 364: What do you think about using _not found safe_ remove methods as they are pretty convenient ?

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/MachineLogMessageImpl.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(machine, content);
+    }
+
+    @Override
+    public String toString() {
+        return "MachineLogMessageImpl{" +
```
benoitf: we could avoid some concatenate as it's always starting with "MachineLogMessageImpl{machine=''"
or maybe MachineLogMessageImpl should use class.getName so refactoring will rename automatically the toString

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineServiceLinksInjector.java
```diff
@@ -118,23 +118,29 @@ public MachineDto injectLinks(MachineDto machine, ServiceContext serviceContext)
...
-        injectMachineChannelsLinks(machine.getConfig(),
-                                   machine.getWorkspaceId(),
-                                   machine.getEnvName(),
-                                   machineChannelLink,
-                                   channelParameter);
+        links.add(cloneDto(workspaceChannelLink).withRel(GET_ENVIRONMENT_OUTPUT_CHANNEL)
+                                                .withParameters(singletonList(cloneDto(channelParameter)
+                                                                                      .withDefaultValue("workspace:" +
+                                                                                                        machine.getWorkspaceId() +
+                                                                                                        ":environment_output"))));
```
benoitf: "environment_output" could be a constant ? as it's used at several places
or a constant "workspace:%s:environment_output" then we only use String format with selected workspaceId

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
```diff
@@ -326,37 +343,42 @@ private MachineImpl createMachine(MachineConfigImpl machineConfig,
...
+                    instanceCreator.createInstance(instanceProvider, machine, outputConsumer);
                 }
             }
             return machine;
-        } catch (ConflictException e) {
+        } catch (ApiException e) {
             try {
                 machineLogger.close();
-            } catch (IOException ignored) {
+            } catch (IOException e1) {
```
voievodin: My preference:
when you have nested exceptions it is always good to use significant names, as if you need to reuse nested exception it may be confusing to deal with _e_, _e1_, _e2_ etc. So i would rename `e -> apiEx`, `e1 -> ioEx`.

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/event/MachineStateMessenger.java
```diff
@@ -44,7 +44,7 @@ public MachineStateMessenger(EventService eventService) {
     public void onEvent(MachineStatusEvent event) {
         try {
             final ChannelBroadcastMessage bm = new ChannelBroadcastMessage();
-            bm.setChannel("machine:status:" + event.getWorkspaceId() + ':' + event.getMachineName());
+            bm.setChannel("workspace:" + event.getWorkspaceId() + ":machines_statuses");
```
voievodin: This kind of refactoring always affects client, and you always change it, and usually it is changed in 2 places(ide, dashboard). I think in this case it will be useful to extract the creation of the channel name into separate SHARABLE class which may be reused by GWT, in that way you don't need to refactor ide client when you change the format of the channel. wdyt?

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/MachineLogMessageImpl.java
```diff
@@ -0,0 +1,71 @@
...
+        if (this == o) return true;
+        if (!(o instanceof MachineLogMessageImpl)) return false;
+        MachineLogMessageImpl that = (MachineLogMessageImpl)o;
+        return Objects.equals(machine, that.machine) &&
+               Objects.equals(content, that.content);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(machine, content);
```
voievodin: -1 for using `Objects.hash(...)` for hashcodes, as it will create an array and i think that it is important to make such thin places as cheaper as possible. If you define hashcode for LogMessage, then it is useful to use this message in collections or smth, and based on log message nature you will have to deal with tons of those messages.
Checkout [this](https://github.com/evoevodin/intellij-idea-tools/tree/master/templates/equals-hashcode/collections-never-null) for generating cheaper hashcode.

benoitf: I don't see the issue of using boxing (varargs), it is used like for every log message you send.
While hashCode would be less called. Also your cheaper hashcode is too verbose and can be error prone as well.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/MessageConsumer.java
```diff
@@ -0,0 +1,26 @@
...
+import java.io.Closeable;
+import java.io.IOException;
+
+/**
+ * Consumes messages one by one for analysing, writing, storing, etc.
+ *
+ * @author Alexander Garagatyi
+ */
+public interface MessageConsumer<T> extends Closeable {
+    void write(T message) throws IOException;
```
vzhukovs: Javadoc

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/MachineLogMessage.java
garagatyi: tnx

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/MachineLogMessage.java
garagatyi: I'll do that

wsmaster/che-core-api-machine-shared/src/main/java/org/eclipse/che/api/machine/shared/dto/MachineLogMessageDto.java
garagatyi: tnx

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
garagatyi, line 352: Thanks for that proposal. I do not want to change current behavior since I do believe that we have some code that rely on such behavior. I'm going to cleanup/refactoring that code later and I'll try to change that.

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
garagatyi, line 364: Thanks for that proposal. I do not want to change current behavior since I do believe that we have some code that rely on such behavior. I'm going to cleanup/refactoring that code later and I'll try to change that.

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/MachineLogMessageImpl.java
garagatyi: I think that IDE should be smart enough to change Class name. But I'll reduce number of concatenations.
BTW Probably we should find single policy of toString/hashcode/equals methods writing. @benoitf @evoevodin WDYT?

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
garagatyi: ok, tnx

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineServiceLinksInjector.java
garagatyi: ok

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/event/MachineStateMessenger.java
garagatyi: I will use constant for that. But refactoring is needed in any case.

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/MachineLogMessageImpl.java
garagatyi: For now I prefer usage of auto-generation of these classes since it is tested feature and will not introduce a bug which can be introduced by manual writing of these methods. Also there is a problem of changing fields in a class with manually created equals and hashcode since it is harder and error prone to modify such methods.

skabashnyuk:
ok

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/MessageConsumer.java
garagatyi: It is obvious what this single  method  in this class does. So I would prefer to omit useless javadocs here.

garagatyi:
@evoevodin @benoitf I've updated PR with requested fixes. Please take a look.

voievodin:
lgtm

PR #1603, opened 2016-06-29
Add JPA implementions of UserDao/ProfileDao

voievodin:
Along with JPA implementations i added 2 password encryptors _SHA-512_ and _PBKDF2_.
`JpaUserDao` and `JpaProfileDao` pass all the TCK prepared before. 

@skabashnyuk, @sleshchenko, @garagatyi, @akorneta  can you please review.
Please note that this will be merged into the `jpa-integration` branch which will be merged to the master when all the DAOs are migrated to JPA.

### DIFF
--- a/core/che-core-api-core/src/main/java/org/eclipse/che/security/PBKDF2PasswordEncryptor.java
+++ b/core/che-core-api-core/src/main/java/org/eclipse/che/security/PBKDF2PasswordEncryptor.java
@@ -67,7 +67,7 @@ public class PBKDF2PasswordEncryptor implements PasswordEncryptor {
             return false;
         }
         // retrieve salt, password hash and iterations count from hash
-        final Integer iterations = tryParse(firstNonNull(matcher.group("iterations"), ""));
+        final Integer iterations = tryParse(matcher.group("iterations"));
         final String salt = matcher.group("saltHash");
         final String pwdHash = matcher.group("pwdHash");
         // compute password's hash and test whether it matches to given hash
--- a/core/che-core-api-core/src/main/java/org/eclipse/che/security/PBKDF2PasswordEncryptor.java
+++ b/core/che-core-api-core/src/main/java/org/eclipse/che/security/PBKDF2PasswordEncryptor.java
@@ -1,0 +1,87 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.security;
+
+import com.google.common.hash.HashCode;
+
+import javax.crypto.SecretKeyFactory;
+import javax.crypto.spec.PBEKeySpec;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+import java.security.spec.InvalidKeySpecException;
+import java.security.spec.KeySpec;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static com.google.common.base.MoreObjects.firstNonNull;
+import static com.google.common.primitives.Ints.tryParse;
+import static java.lang.String.format;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Encrypts password using <a href="https://en.wikipedia.org/wiki/PBKDF2">Password-based-Key-Derivative-Function</a>
+ * with <b>SHA512</b> as pseudorandom function.
+ * See <a href="https://www.ietf.org/rfc/rfc2898.txt">rfc2898</a>.
+ *
+ * @author Yevhenii Voevodin
+ */
+public class PBKDF2PasswordEncryptor implements PasswordEncryptor {
+
+    private static final String  PWD_FMT   = "%s:%s:%d";
+    private static final Pattern PWD_REGEX = Pattern.compile("(?<pwdHash>\\w+):(?<saltHash>\\w+):(?<iterations>[0-9]+)");
+
+    private static final String       SECRET_KEY_FACTORY_NAME = "PBKDF2WithHmacSHA512";
+    private static final SecureRandom SECURE_RANDOM           = new SecureRandom();
+    /**
+     * Minimum number of iterations required is 1_000(rfc2898),
+     * pick greater as potentially safer in the case of brute-force attacks .
+     */
+    private static final int          ITERATIONS_COUNT        = 10_000;
+    /** 64bit salt length based on the rfc2898 spec . */
+    private static final int          SALT_LENGTH             = 64 / 8;
+
+    @Override
+    public String encrypt(String password) {
+        requireNonNull(password, "Required non-null password");
+        final byte[] salt = new byte[SALT_LENGTH];
+        SECURE_RANDOM.nextBytes(salt);
+        final HashCode hash = computeHash(password.toCharArray(), salt, ITERATIONS_COUNT);
+        final HashCode saltHash = HashCode.fromBytes(salt);
+        return format(PWD_FMT, hash, saltHash, ITERATIONS_COUNT);
+    }
+
+    @Override
+    public boolean test(String password, String encryptedPassword) {
+        requireNonNull(password, "Required non-null password");
+        requireNonNull(password, "Required non-null encrypted password");
+        final Matcher matcher = PWD_REGEX.matcher(encryptedPassword);
+        if (!matcher.matches()) {
+            return false;
+        }
+        // retrieve salt, password hash and iterations count from hash
+        final Integer iterations = tryParse(firstNonNull(matcher.group("iterations"), ""));
+        final String salt = matcher.group("saltHash");
+        final String pwdHash = matcher.group("pwdHash");
+        // compute password's hash and test whether it matches to given hash
+        final HashCode hash = computeHash(password.toCharArray(), HashCode.fromString(salt).asBytes(), iterations);
+        return hash.toString().equals(pwdHash);
+    }
+
+    private HashCode computeHash(char[] password, byte[] salt, int iterations) {
+        try {
+            final SecretKeyFactory keyFactory = SecretKeyFactory.getInstance(SECRET_KEY_FACTORY_NAME);
+            final KeySpec keySpec = new PBEKeySpec(password, salt, iterations, 512);
+            return HashCode.fromBytes(keyFactory.generateSecret(keySpec).getEncoded());
+        } catch (NoSuchAlgorithmException | InvalidKeySpecException x) {
+            throw new RuntimeException(x.getMessage(), x);
+        }
+    }
+}
--- a/core/che-core-api-core/src/main/java/org/eclipse/che/security/SHA512PasswordEncryptor.java
+++ b/core/che-core-api-core/src/main/java/org/eclipse/che/security/SHA512PasswordEncryptor.java
@@ -1,0 +1,64 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.security;
+
+import com.google.common.hash.HashCode;
+import com.google.common.hash.Hashing;
+import com.google.common.primitives.Bytes;
+
+import java.nio.charset.Charset;
+import java.security.SecureRandom;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * SHA-512 based encryptor {@code hash = sha512(password + salt) + salt}.
+ *
+ * @author Yevhenii Voevodin
+ */
+public class SHA512PasswordEncryptor implements PasswordEncryptor {
+
+    /** 64 bit salt length is based on the <a href="https://www.ietf.org/rfc/rfc2898.txt">source</a>. */
+    private static final int          SALT_BYTES_LENGTH               = 64 / 8;
+    /** SHA-512 produces 512 bits. */
+    private static final int          ENCRYPTED_PASSWORD_BYTES_LENGTH = 512 / 8;
+    private static final SecureRandom SECURE_RANDOM                   = new SecureRandom();
+    private static final Charset      PWD_CHARSET                     = Charset.forName("UTF-8");
+
+    @Override
+    public String encrypt(String password) {
+        requireNonNull(password, "Required non-null password");
+        // generate salt
+        final byte[] salt = new byte[SALT_BYTES_LENGTH];
+        SECURE_RANDOM.nextBytes(salt);
+        // sha512(password + salt)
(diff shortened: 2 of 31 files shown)


core/che-core-api-jdbc-vendor-h2/src/main/java/org/eclipse/che/api/core/jdbc/jpa/vendor/h2/H2ExceptionHandler.java
```diff
@@ -0,0 +1,39 @@
...
+import org.eclipse.persistence.exceptions.ExceptionHandler;
+
+import static org.eclipse.che.api.core.jdbc.DBUtil.extractSqlErrorCode;
+
+/**
+ * @author Yevhenii Voevodin
+ */
+public class H2ExceptionHandler implements ExceptionHandler {
+
+    // TODO move it from here
```
sleshchenko: When do you plan remove it?

pom.xml
```diff
@@ -42,11 +42,19 @@
     <properties>
         <che.lib.version>4.5.0-RC1-SNAPSHOT</che.lib.version>
         <che.version>4.5.0-RC1-SNAPSHOT</che.version>
+        <com.h2database.version>1.4.186</com.h2database.version>
```
sleshchenko, line 45: Do you plan move these dependencies to che-dependencies project?

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/model/impl/ProfileImpl.java
```diff
@@ -21,19 +31,36 @@
...
+    @MapKeyColumn(name = "name")
+    @Column(name = "value")
+    @CollectionTable(joinColumns = @JoinColumn(name = "user_id"),
+                     uniqueConstraints = @UniqueConstraint(columnNames = {"user_id", "name"}))
     private Map<String, String> attributes;
 
-    public ProfileImpl(String id, String email) {
-        this.id = id;
+    // jpa mapping delegated to 'user' field
+    @Transient
```
benoitf: I don't get why we need the field if it's never stored

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/model/impl/UserImpl.java
```diff
@@ -22,14 +28,29 @@
...
  */
+@Entity(name = "User")
 public class UserImpl implements User {
 
-    private String       id;
-    private String       email;
-    private String       name;
-    private String       password;
+    @Id
+    @Column(name = "id")
```
benoitf: why do we set the name of the columns and not let the default ?

benoitf:
few remarks: 
- If there are password encryptors, why they're not part of a separated commits as it's not linked to JPA ?
- why the wheel has been re-invented with TCK modules ? while people are using Arquilian when testing JPA ? http://arquillian.org/guides/testing_java_persistence/
  - che-core-api-jdbc-vendor-h2
    the name is not obvious as it links the module to EclipseLink, while JPA allow to plug any persistence provider. so it should include that it's specific to eclipselink and is not "jdbc" neutral
- persistence.xml shouldn't be in wsmaster/che-core-api-user/src/main/resources/META-INF/
  it's part of the final assembly as it's where we select the persistence provider, and the database 
- Using container managed instead of JavaSE would have helped as here you've to managed transaction instead of only using annotations
  Using an autocloseable (or an interceptor like it's done in container managed world) should help to avoid to have tons of lines doing. Then you could pickup the transaction strategy (REQUIRED, REQUIRES_NEW, etc)

```

try {
            manager.getTransaction().begin();

// do something

           manager.getTransaction().commit();
        } finally {
            if (manager.getTransaction().isActive()) {
               manager.getTransaction().rollback();
            }
            manager.close();
       }
```

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/model/impl/ProfileImpl.java
voievodin: Email is stored by `User` table while it is used in profile as well, there is no need to store email so the mapping is delegated to the User's field, on the other hand we need to create new profile instances from another profile instances or any data, in this way we need email field, which allows to avoid creation of embedded user object.

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/model/impl/UserImpl.java
voievodin: Are you talking about `id` field or about the whole class mapping?
I am wondering that if we change field name the mappings will be also changed which requires
 migration as well. So it looks more strict for me, and i think it is necessary to name the columns explicitly. Your proposal?

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/model/impl/ProfileImpl.java
benoitf: in which case do we need to create a ProfileImpl directly with a userId and a given email instead of getting email through the user object ?

If Profile has a JPA relationship with User object, there is no need to store/wrap email in any case as you can do getProfile().getUser().getEmail() as well

voievodin:
> If there are password encryptors, why they're not part of a separated commits as it's not linked to JPA ?

np, i will extract ecnryptors code to the different commit.

> why the wheel has been re-invented with TCK modules ? while people are using Arquilian when testing JPA ? http://arquillian.org/guides/testing_java_persistence

Current TCK approach allows us use the same tests for different implementations Local/Ldap/Jpa.
Can Arquilian resolve this?

> che-core-api-jdbc-vendor-h2 the name is not obvious as it links the module to EclipseLink, while JPA allow to plug any persistence provider. so it should include that it's specific to eclipselink and is not "jdbc" neutral

_che-core-api-jdbc_ module contains common error codes and error codes mapper astraction used by daos, _che-core-api-jdbc-vendor-h2_ provides an implementation of error codes mapper specific for h2, while error codes are jdbc errors.
I will move code specific to eclipse link to the dedicated _jpa/eclipselink_ package are you okay with that? 

> persistence.xml shouldn't be in wsmaster/che-core-api-user/src/main/resources/META-INF/
> it's part of the final assembly as it's where we select the persistence provider, and the database

okay thank you, i am using it for test purposes only, so i will remove it from there.

> Using container managed instead of JavaSE would have helped as here you've to managed transaction instead of only using annotations
> Using an autocloseable (or an interceptor like it's done in container managed world) should help to avoid to have tons of lines doing. Then you could pickup the transaction strategy (REQUIRED, REQUIRES_NEW, etc)

We will use [this annotation](https://github.com/google/guice/wiki/Transactions), i will introduce it with next pull request, right now alot of things depend on this pull request that's why we decided to merge it asap which allows me and guys to work on jpa implementations separately.

pom.xml
voievodin, line 45: yes

benoitf:
> > If there are password encryptors, why they're not part of a separated commits as it's not linked to JPA ?
> 
> np, i will extract ecnryptors code to the different commit.

thx

> > why the wheel has been re-invented with TCK modules ? while people are using Arquilian when testing JPA ? http://arquillian.org/guides/testing_java_persistence
> 
> Current TCK approach allows us use the same tests for different implementations Local/Ldap/Jpa.
> Can Arquilian resolve this?

It allows to test with different JPA providers and different database as well
also it's a Container Agnostic framework and allow plugability.

> > che-core-api-jdbc-vendor-h2 the name is not obvious as it links the module to EclipseLink, while JPA allow to plug any persistence provider. so it should include that it's specific to eclipselink and is not "jdbc" neutral
> 
> _che-core-api-jdbc_ module contains common error codes and error codes mapper astraction used by >daos, _che-core-api-jdbc-vendor-h2_ provides an implementation of error codes mapper specific for h2, >while error codes are jdbc errors.
> I will move code specific to eclipse link to the dedicated _jpa/eclipselink_ package are you okay with >that? 

sure

> > persistence.xml shouldn't be in wsmaster/che-core-api-user/src/main/resources/META-INF/
> > it's part of the final assembly as it's where we select the persistence provider, and the database
> 
> okay thank you, i am using it for test purposes only, so i will remove it from there.

ok

> > Using container managed instead of JavaSE would have helped as here you've to managed >>transaction instead of only using annotations
> > Using an autocloseable (or an interceptor like it's done in container managed world) should help to >>avoid to have tons of lines doing. Then you could pickup the transaction strategy (REQUIRED, >>REQUIRES_NEW, etc)
> 
> We will use [this annotation](https://github.com/google/guice/wiki/Transactions), i will introduce it with >next pull request, right now alot of things depend on this pull request that's why we decided to merge it >asap which allows me and guys to work on jpa implementations separately.

Yes this Guice stuff or better, CDI, which offer more stuff around Java EE component will allow more possibilities around PersistenceContext

core/che-core-api-core/src/main/java/org/eclipse/che/security/PBKDF2PasswordEncryptor.java
```diff
@@ -0,0 +1,90 @@
...
+        requireNonNull(password, "Required non-null encrypted password");
+        final Matcher matcher = PWD_REGEX.matcher(encryptedPassword);
+        if (!matcher.matches()) {
+            return false;
+        }
+        // retrieve salt, password hash and iterations count from hash
+        final Integer iterations = tryParse(firstNonNull(matcher.group("iterations"), ""));
+        final String salt = matcher.group("saltHash");
+        final String pwdHash = matcher.group("pwdHash");
+        if (salt == null || iterations == null || iterations <= 0 || pwdHash == null) {
```
garagatyi: Encrypted password matches regexp and regexp require 1+ symbols in each group. It looks like NULL checks are not needed.

garagatyi: I mean salt and pwdHash

core/che-core-api-core/src/main/java/org/eclipse/che/security/PBKDF2PasswordEncryptor.java
```diff
@@ -0,0 +1,90 @@
...
+ * Encrypts password using <a href="https://en.wikipedia.org/wiki/PBKDF2">Password-based-Key-Derivative-Function</a>
+ * with <b>SHA512</b> as pseudorandom function.
+ * See <a href="https://www.ietf.org/rfc/rfc2898.txt">rfc2898</a>.
+ *
+ * @author Yevhenii Voevodin
+ */
+public class PBKDF2PasswordEncryptor implements PasswordEncryptor {
+
+    private static final String  PWD_FMT   = "%s:%s:%d";
+    private static final Pattern PWD_REGEX = Pattern.compile("(?<pwdHash>\\w+):(?<saltHash>\\w+):(?<iterations>\\w+)");
```
garagatyi: Are you sure that iterations section may contain letters?

core/che-core-api-core/src/main/java/org/eclipse/che/security/SHA512PasswordEncryptor.java
```diff
@@ -0,0 +1,64 @@
...
+import java.security.SecureRandom;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * SHA-512 based encryptor {@code hash = sha512(password + salt) + salt}.
+ *
+ * @author Yevhenii Voevodin
+ */
+public class SHA512PasswordEncryptor implements PasswordEncryptor {
```
garagatyi, line 27: Why do we need 2 implementations of password encryptor?

core/che-core-api-core/src/main/java/org/eclipse/che/security/PBKDF2PasswordEncryptor.java
```diff
@@ -0,0 +1,90 @@
...
+    @Override
+    public boolean test(String password, String encryptedPassword) {
+        requireNonNull(password, "Required non-null password");
+        requireNonNull(password, "Required non-null encrypted password");
+        final Matcher matcher = PWD_REGEX.matcher(encryptedPassword);
+        if (!matcher.matches()) {
+            return false;
+        }
+        // retrieve salt, password hash and iterations count from hash
+        final Integer iterations = tryParse(firstNonNull(matcher.group("iterations"), ""));
```
garagatyi: Since non-ASCII characters are not allowed here please check that also in encrypt method

core/che-core-api-core/src/test/java/org/eclipse/che/security/PasswordEncryptorsTest.java
```diff
@@ -0,0 +1,41 @@
...
+import static org.testng.Assert.assertTrue;
+
+/**
+ * @author Yevhenii Voevodin
+ */
+public class PasswordEncryptorsTest {
+
+    @Test(dataProvider = "encryptorsProvider")
+    public void testEncryption(PasswordEncryptor encryptor) throws Exception {
+        final String password = "password";
```
garagatyi, line 26: I think that this test doesn't cover several corner cases

core/che-core-api-jdbc/src/main/java/org/eclipse/che/api/core/jdbc/jpa/DetailedRollbackException.java
```diff
@@ -0,0 +1,32 @@
...
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.api.core.jdbc.jpa;
+
+import org.eclipse.che.api.core.jdbc.DBErrorCode;
+
+import javax.persistence.RollbackException;
+
+/**
+ * @author Yevhenii Voevodin
```
garagatyi, line 20: Please add description of class into javadocs

core/che-core-api-core/src/main/java/org/eclipse/che/security/PBKDF2PasswordEncryptor.java
garagatyi: Skip this comment. My error.

core/che-core-api-core/src/main/java/org/eclipse/che/security/SHA512PasswordEncryptor.java
voievodin, line 27: Well, first i created _classic_ `SHA512PasswordEncryptor`  but then found a lot of articles recommending bcrypt/scrypt/PBKDF2 as safer mechanisms for password encryption, so i implemented another `PBKDF2PasswordEncryptor` with sha-512 as pseudorandom function. For `PBKDF2PasswordEncryptor` ecnryption time depends on the count of the iterations but `SHA512PasswordEncryptor` works faster, so i decided to not remove it as it is useful in tests.
Notice time:
image

voievodin:
Pull request updated, @benoitf @skabashnyuk please check

core/che-core-api-core/src/main/java/org/eclipse/che/security/PBKDF2PasswordEncryptor.java
garagatyi: I think group can't be Null in that case

benoitf:
ok

core/che-core-api-core/src/main/java/org/eclipse/che/security/PBKDF2PasswordEncryptor.java
voievodin: yes, thank you!

pom.xml
skabashnyuk, line 45: I think it's good time to start creating CQs

voievodin, line 45: :C ok

skabashnyuk:
ok

garagatyi:
LGTM

PR #1613, opened 2016-06-30
CODENVY-606: fix NPE in DockerInstanceStopDetector

mmorhun:
We had NPE in `DockerInstanceStopDetector` because of new format of docker response to `get events` method and bug in docker swarm which do not filter events.
This PR adapts expected format of `get events` response and adds check for event type in response processing.

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/Actor.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/Actor.java
@@ -1,0 +1,54 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.plugin.docker.client.json;
+
+import com.google.gson.annotations.SerializedName;
+
+import java.util.Map;
+
+/**
+ * Represents Actor field of {@link Event}.
+ *
+ * @author Mykola Morhun
+ */
+public class Actor {
+    @SerializedName("ID")
+    private String             id;
+    @SerializedName("Attributes")
+    private Map<String,String> attributes;
+
+    public String getId() {
+        return id;
+    }
+
+    public Map<String, String> getAttributes() {
+        return attributes;
+    }
+
+    public Actor withId(String id) {
+        this.id = id;
+        return this;
+    }
+
+    public Actor withAttributes(Map<String, String> attributes) {
+        this.attributes = attributes;
+        return this;
+    }
+
+    @Override
+    public String toString() {
+        return "Actor{" +
+               "id='" + id + '\'' +
+               ", attributes=" + attributes +
+               '}';
+    }
+
+}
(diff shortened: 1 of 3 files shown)


plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/Actor.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Map;
+
+/**
+ * Represents Actor field of {@link Event}.
+ *
+ * @author Mykola Morhun
+ */
+public class Actor {
+    private String ID;
+    Map<String,String> Attributes;
```
garagatyi: private

garagatyi: It is not common for java to use names of var or field with leading upper-case symbol

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/Event.java
```diff
@@ -14,11 +14,15 @@
  * Docker event.
  *
  * @author Alexander Garagatyi
+ * @author Mykola Morhun
  */
 public class Event {
     private String status;
     private String id;
     private String from;
+    private String Type;
```
garagatyi: Do not use upper-case leading character

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/Event.java
```diff
@@ -38,6 +42,18 @@ public String getStatus() {
...
 
+    public String getType() {
+        return Type;
+    }
+
+    public String getAction() {
+        return Action;
+    }
+
+    public org.eclipse.che.plugin.docker.client.json.Actor getActor() {
```
sleshchenko: Can you explain why do you use fqn here?

garagatyi: Why FQN?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/Event.java
sleshchenko: Why can't we use camelcase for this variable name?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/Event.java
mmorhun: I do not know, auto-generated getter. thx

mmorhun:
@garagatyi @sleshchenko ?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/Actor.java
```diff
@@ -0,0 +1,50 @@
...
+package org.eclipse.che.plugin.docker.client.json;
+
+import java.util.Map;
+
+/**
+ * Represents Actor field of {@link Event}.
+ *
+ * @author Mykola Morhun
+ */
+public class Actor {
```
garagatyi, line 22: Can you provide an example of JSON that this class represents?

garagatyi:
LGTM

PR #1614, opened 2016-06-30
CHE-1353: Add possibility to configure specific docker api version

AndrienkoAleksandr:
Add possibility to configure api version by che.properties. Set minimum default docker api version 1.20.

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
@@ -134,13 +134,17 @@ public class DockerConnector {
     private final ExecutorService            executor;
     private final DockerConnectionFactory    connectionFactory;
 
+    protected final String apiVersionPathPrefix;
+
     @Inject
     public DockerConnector(DockerConnectorConfiguration connectorConfiguration,
                            DockerConnectionFactory connectionFactory,
-                           DockerRegistryAuthResolver authResolver) {
+                           DockerRegistryAuthResolver authResolver,
+                           DockerApiVersionPathPrefixProvider dockerApiVersionPathPrefixProvider) {
         this.dockerDaemonUri = connectorConfiguration.getDockerDaemonUri();
         this.connectionFactory = connectionFactory;
         this.authResolver = authResolver;
+        this.apiVersionPathPrefix = dockerApiVersionPathPrefixProvider.get();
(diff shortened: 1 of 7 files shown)


garagatyi:
LGTM
@mmorhun @sleshchenko Please take a look

plugins/plugin-docker/che-plugin-docker-client/src/test/java/org/eclipse/che/plugin/docker/client/DockerConnectorTest.java
```diff
@@ -174,8 +177,12 @@ public void setup() throws IOException, URISyntaxException {
         when(dockerResponse.getInputStream()).thenReturn(inputStream);
         when(initialAuthConfig.getAuthConfigs()).thenReturn(authConfigs);
         when(authConfigs.getConfigs()).thenReturn(new HashMap<>());
+        when(dockerApiVersionPathPrefixProvider.get()).thenReturn(API_VERSION_PREFIX);
```
sleshchenko, line 205: It would be better to add some test for ensuring that docker connector uses api version prefix

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -121,14 +121,17 @@
...
     @Inject
     public DockerConnector(DockerConnectorConfiguration connectorConfiguration,
                            DockerConnectionFactory connectionFactory,
-                           DockerRegistryAuthResolver authResolver) {
+                           DockerRegistryAuthResolver authResolver,
+                           DockerApiVersionPathPrefixProvider dockerApiVersionPathPrefixProvider) {
         this.dockerDaemonUri = connectorConfiguration.getDockerDaemonUri();
         this.connectionFactory = connectionFactory;
         this.authResolver = authResolver;
+        this.apiVersionPathPrefix = dockerApiVersionPathPrefixProvider.get();
```
sleshchenko, line 147: Is it OK to use provider here? Mb it would be better to inject named string here and bind this string to provider?

sleshchenko:
other LGTM

skabashnyuk:
ok

garagatyi:
Please rebase PR

plugins/plugin-docker/che-plugin-docker-client/src/test/java/org/eclipse/che/plugin/docker/client/DockerConnectorTest.java
AndrienkoAleksandr, line 205: Ok, I added test for this testcase

PR #1615, opened 2016-06-30
CHE-1350: Set refspec in PushTest to prevent git error

vinokurig:
Git can be configured to throw error on push without refspec


wsagent/che-core-api-git/src/test/java/org/eclipse/che/git/impl/PushTest.java
```diff
@@ -151,6 +151,8 @@ public void testPushWhenLocalRepositoryIsUpToDate(GitConnectionFactory connectio
 
         //when
         PushResponse pushResponse = localConnection.push(newDto(PushRequest.class)
+                                                                 .withRefSpec(
+                                                                         Collections.singletonList("refs/heads/master:refs/heads/master"))
```
voievodin: Use static import for `singletonList` and change it to one line please

voievodin:
The code is ok for me

PR #1623, opened 2016-07-01
Fix integration test on latest ubuntu

garagatyi:
**_1 Upvote**_ @AndrienkoAleksandr Please take a look.
@eivantsov FYI


AndrienkoAleksandr:
+1

PR #1625, opened 2016-07-01
Update according to changes in everrest

mkuznyetsov:


skabashnyuk:
ок

musienko-maxim:

PR #1628, opened 2016-07-01
CHE-1403: Fix clone with http url in JGit implementation

vinokurig:
JGit does not support HTTP 301 redirects yet, it throws exception when
performing commands with remote using http url instead https. Native git can do it.
The solution is to replace http to https in the given url when performing clone.
Also I have added error code to "ssh key" exception and fixed related client message
@skabashnyuk please review


skabashnyuk:
ок

PR #1634, opened 2016-07-02
[WIP] CHE-11: VFS event handling system

dkulieshov:
Hi,

I'm working on a new VFS event handling system, I prepared a pull request with initial implementation and I would like to discuss the approach.
### Reasoning

We need a service to be able to detect virtual file system (and physical file system correspondingly) events and properly react. In this there is a need because now there is a number of problems. For example right now some file system related events are not handled at all or handled partially. Another issue: there is no automated event management on VFS level, that means that a developer has to care by himself to raise an event to handle it later. For example when we implement an editor to be able to properly react on file changes first we have to manually detect (somehow) those changes and then manually raise corresponding event while that could be done on VFS level automatically and a developer would only care about how to react on an event. Finally, there is one more thing that is worth mentioning. The matter is that without proper means of event handling on file system level we have no possibility to detect (and as a consequence to react on) events that have their origin outside of IDE. For example, when a user edits a file (that is already opened in IDE editor) using terminal or MC editor we are unable detect that event thus we can't refresh edited file's content in IDE editor.
### Idea

Generally speaking that should not be a problem to catch file or directory modification event but there are many examples where it is just not enough for our purposes. When we create a project it is not enough to track folder creation and when we check out a branch it is not enough to track `.git/HEAD` file modifications. Often to see the picture we need to track the whole sequence of events or vice versa the sequence leads to the final event that is the only one that is meaningful for us. For example, if we create a java project we track root folder, `src` folder and almost certainly project object model file (let's assume that it is a maven project). One more example, when we edit a file within a few seconds (a sequence of file modifications) there is no real need to track intermediate file states, so we can skip most of events except the last one with the final file state. So the challenge I see here is to implement a clever system that will not only handle events but also will provide means to easily and effectively analyze both single and sequential events.

In the first place, I propose to distinguish events according to their nature and logical level. The low level includes events related to modification of items in context of file system (e.g. creation of a directory, modification of a file). While the high level includes events related to complex logical events that represents a higher level of abstraction (e.g. modification of POM file, compiling project sources).

Next thing is event segments. If we imagine a timeline with events then the event segment will be a segment of that line that contains all events that are close to each other in terms of time. If time interval between two events is less than some defined value, then we consider those events related to each other and we attribute them to a common event segment. And the opposite, if timestamps of two consequent events differs greatly we consider those events as belonging to two different event segments and are not related. Thus we will split an event timeline into segments containing sequences of related events.

In order to simplify analysis of those segments, I assume we can introduce a tree-like structures that will correspond to trees of items of a file system and hold data related to events that we caught. If an item is modified on a file system it is also noted in an event tree in a corresponding place with all related data set.

The last important thing I would like to introduce is so called event detectors. Briefly, they are responsible for event tree analysis and also produce corresponding high level events. How it should work: we apply those detectors to each event tree that we get in real time to see if corresponding event happened, when appropriate conditions are met we raise high level event.
### Implementation

Probably it is yet to be discussed but I have implemented a very basic draft of a system and I would like to provide a brief description of the most valuable components.

`LoEventListener` class is the beginning. The only thing it does is listens for `LoEvent` - low level file system events - and populates a `LoEventQueueHolder`. 

`LoEventService` is responsible for event timeline segmentation and populating of  `EventTreeQueueHolder`.

Generally speaking, a couple of `QueueHolder` classes are used to share data access between entities using dependency injection while queues themselves are used to conform with variation of producer/consumer pattern.

Finally, `HiEventService` uses `HiEventDetectorManager` and `HiEventBroadcasterManager` to correspondingly detect `HiEvent` with `HiEventDetector` implementations and broadcast them.
### Conclusion

In conclusion I will emphasize the benefits that I see:
- A good idea of the overall picture. When we create event trees we have quite a good amount of data to make more accurate decision, in comparison to processing event sequence when we separately and independently analyze events.
- Source of modification does not matter. We have a huge amount of problems because of inability to effectively track events caused by outer sources (e.g. separate SSH). While new approach will handle such issues automatically because it does not care about source of modification, though it may take it into account if necessary.
- Ease of extending. If a developer wants to track a new type of events there is no need to do extra coding, he just need to provide corresponding HiEventDetector. This is an incontestable benefit of dependency injection framework Guice and well-decoupled logics approach.
- Simplification of recognition of complex events. This is achieved thanks to separation of logical (high level) and physical (low level) events.
- Still effective detection of simple events. We can easily detect any simple event inside complex one due to the fact that the complex event is represented by a handy data structure.

And the impediments that I foresee:
- Performance. While the implementation performance has already been tested on staging server, I still believe that there is a risk of performance issues. The thing is that along with good opportunities for an accurate event timeline analysis we get overhead in memory consumption (which is within the normal right now) and a bit more usage of processor time. In comparison to plain real time event handling of course. 
- Depends on low level implementations. I mean that we don't get events directly form file system. Before we know that something happens the event passes several intermediate layers which are not under full our control. It is a potential place for problems and may cause misunderstandings or system malfunctions.
- Real time file system event processing. Any real time processing always makes things more complex because no one clearly know what will happen and when. We assume that events related to each other happen almost simultaneously but in fact that may not always be true.
- More complex event we want to track, more complex event detector we must provide. Some events may require quite a big amount of effort to track, especially if they affect several event segments or require extra data.

### DIFF
--- a/plugins/plugin-maven/che-plugin-maven-server/src/main/java/org/eclipse/che/plugin/maven/server/core/project/PomChangeListener.java
+++ b/plugins/plugin-maven/che-plugin-maven-server/src/main/java/org/eclipse/che/plugin/maven/server/core/project/PomChangeListener.java
@@ -17,10 +17,11 @@ import com.google.inject.name.Named;
 import org.eclipse.che.api.core.notification.EventService;
 import org.eclipse.che.api.core.notification.EventSubscriber;
 import org.eclipse.che.api.project.server.notification.ProjectItemModifiedEvent;
+import org.eclipse.che.api.project.shared.dto.event.PomModifiedEventDto;
 import org.eclipse.che.commons.schedule.executor.ThreadPullLauncher;
+import org.eclipse.che.ide.maven.tools.Model;
 import org.eclipse.che.plugin.maven.server.core.EclipseWorkspaceProvider;
 import org.eclipse.che.plugin.maven.server.core.MavenWorkspace;
-import org.eclipse.che.ide.maven.tools.Model;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IWorkspace;
 import org.eclipse.core.runtime.Path;
@@ -68,6 +69,16 @@ public class PomChangeListener {
                         projectToUpdate.add(new Path(eventPath).removeLastSegments(1).toOSString());
                     }
 //                    }
+                }
+            }
+        });
+
+        eventService.subscribe(new EventSubscriber<PomModifiedEventDto>() {
+            @Override
+            public void onEvent(PomModifiedEventDto event) {
+                String eventPath = event.getPath();
+                if (pomIsValid(eventPath)) {
+                    projectToUpdate.add(new Path(eventPath).removeLastSegments(1).toOSString());
                 }
             }
         });
--- a/wsagent/che-core-api-project-shared/src/main/java/org/eclipse/che/api/project/shared/dto/event/GitBranchCheckoutEventDto.java
+++ b/wsagent/che-core-api-project-shared/src/main/java/org/eclipse/che/api/project/shared/dto/event/GitBranchCheckoutEventDto.java
@@ -1,0 +1,30 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.api.project.shared.dto.event;
+
+import com.google.common.annotations.Beta;
+
+import org.eclipse.che.dto.shared.DTO;
+
+/**
+ * To transfer branch name after git checkout operation
+ *
+ * @author Dmitry Kuleshov
+ *
+ * @since 4.5
+ */
+@Beta
+@DTO
+public interface GitBranchCheckoutEventDto {
+    String getBranchName();
+
+    GitBranchCheckoutEventDto withBranchName(String branchName);
+}
--- a/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectManager.java
+++ b/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectManager.java
@@ -34,18 +34,16 @@ import org.eclipse.che.api.project.server.type.ProjectTypeRegistry;
 import org.eclipse.che.api.project.server.type.ProjectTypeResolution;
 import org.eclipse.che.api.project.server.type.ValueStorageException;
 import org.eclipse.che.api.project.shared.dto.event.FileWatcherEventType;
-import org.eclipse.che.api.project.shared.dto.event.VfsWatchEvent;
 import org.eclipse.che.api.vfs.Path;
 import org.eclipse.che.api.vfs.VirtualFile;
-import org.eclipse.che.api.vfs.VirtualFileFilter;
 import org.eclipse.che.api.vfs.VirtualFileSystem;
 import org.eclipse.che.api.vfs.VirtualFileSystemProvider;
 import org.eclipse.che.api.vfs.impl.file.FileTreeWatcher;
 import org.eclipse.che.api.vfs.impl.file.FileWatcherNotificationHandler;
 import org.eclipse.che.api.vfs.impl.file.FileWatcherNotificationListener;
+import org.eclipse.che.api.vfs.impl.file.event.LoEvent;
 import org.eclipse.che.api.vfs.search.Searcher;
 import org.eclipse.che.api.vfs.search.SearcherProvider;
-import org.eclipse.che.dto.server.DtoFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -109,16 +107,22 @@ public final class ProjectManager {
 
     @PostConstruct
     void initWatcher() throws IOException {
-        FileWatcherNotificationListener defaultListener = new FileWatcherNotificationListener(VirtualFileFilter.ACCEPT_ALL) {
-            @Override
-            public void onFileWatcherEvent(VirtualFile virtualFile, FileWatcherEventType eventType) {
-                LOG.debug("FS event detected: " + eventType + " " + virtualFile.getPath().toString() + " " + virtualFile.isFile());
-                eventService.publish(DtoFactory.newDto(VfsWatchEvent.class)
-                                               .withPath(virtualFile.getPath().toString())
-                                               .withFile(virtualFile.isFile())
-                                               .withType(eventType));
-            }
-        };
+        FileWatcherNotificationListener defaultListener =
+                new FileWatcherNotificationListener(file -> !(file.getPath().toString().contains(".codenvy")
+                                                              || file.getPath().toString().contains(".#"))) {
+                    @Override
+                    public void onFileWatcherEvent(VirtualFile virtualFile, FileWatcherEventType eventType) {
+                        LOG.debug("FS event detected: " + eventType + " " + virtualFile.getPath().toString() + " " + virtualFile.isFile());
+                        eventService.publish(LoEvent.newInstance()
+                                                    .withPath(virtualFile.getPath().toString())
+                                                    .withName(virtualFile.getName())
+                                                    .withItemType(virtualFile.isFile()
+                                                                           ? LoEvent.ItemType.FILE
+                                                                           : LoEvent.ItemType.DIR)
+                                                    .withTime(System.currentTimeMillis())
+                                                    .withEventType(eventType));
+                    }
+                };
         fileWatchNotifier.addNotificationListener(defaultListener);
         try {
             fileWatcher.startup();
--- a/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/impl/file/event/EventTreeHelper.java
+++ b/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/impl/file/event/EventTreeHelper.java
@@ -1,0 +1,93 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.api.vfs.impl.file.event;
+
+import com.google.common.annotations.Beta;
+
+import org.eclipse.che.api.vfs.Path;
+
+import java.util.Optional;
+
+import static java.io.File.separator;
+import static org.eclipse.che.api.vfs.impl.file.event.EventTreeNode.newInstance;
+
+/**
+ * Helper for event tree related operations.
+ *
+ * @see {@link EventTreeNode}
+ *
+ * @author Dmitry Kuleshov
+ *
+ * @since 4.5
+ */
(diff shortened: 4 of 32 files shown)


wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/impl/file/event/VfsEventService.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+@Beta
+abstract class VfsEventService {
+    private static final Logger LOG = getLogger(VfsEventService.class);
+
+    private final ExecutorService executor;
+    private final AtomicBoolean   running;
+
+    public VfsEventService() {
+        this.executor = newSingleThreadExecutor();
```
garagatyi: Please use thread builder factory to set name of the thread. That will help in case we will need to analyze thread dump. Here is an example of the way we usually use it https://github.com/eclipse/che/blob/25e6cc2360133ead8847feccf1f36a5c3f03a2fb/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java#L110-L110

garagatyi: Looks like usage of daemon thread is appropriate here, WDYT?

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/impl/file/event/VfsEventService.java
```diff
@@ -0,0 +1,63 @@
...
+import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import static java.util.concurrent.Executors.newSingleThreadExecutor;
+import static org.slf4j.LoggerFactory.getLogger;
+
+/**
+ * @author Dmitry Kuleshov
```
garagatyi, line 30: Please add description to that class.

plugins/plugin-maven/che-plugin-maven-server/src/main/java/org/eclipse/che/plugin/maven/server/core/project/PomChangeListener.java
```diff
@@ -71,6 +72,16 @@ public void onEvent(ProjectItemModifiedEvent event) {
                 }
             }
         });
+
+        eventService.subscribe(new EventSubscriber<PomModifiedEventDto>() {
```
garagatyi, line 76: FYI: since EventService is synchronous long running subscriber can break file events analysis and reaction for further subscribers.

wsagent/che-core-api-project-shared/src/main/java/org/eclipse/che/api/project/shared/dto/event/GitBranchCheckoutEventDto.java
```diff
@@ -0,0 +1,30 @@
...
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.api.project.shared.dto.event;
+
+import com.google.common.annotations.Beta;
+
+import org.eclipse.che.dto.shared.DTO;
+
+/**
+ * To transfer branch name after git checkout operation
```
garagatyi, line 18: Git checkout can be performed not only with branch name but also commitId, and other links (besides branch name) to commitId. 
Is this class just an example? Will it handle other checkout variants later?

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectManager.java
```diff
@@ -109,16 +107,22 @@ public ProjectManager(VirtualFileSystemProvider vfsProvider,
...
-            public void onFileWatcherEvent(VirtualFile virtualFile, FileWatcherEventType eventType) {
-                LOG.debug("FS event detected: " + eventType + " " + virtualFile.getPath().toString() + " " + virtualFile.isFile());
-                eventService.publish(DtoFactory.newDto(VfsWatchEvent.class)
-                                               .withPath(virtualFile.getPath().toString())
-                                               .withFile(virtualFile.isFile())
-                                               .withType(eventType));
-            }
-        };
+        FileWatcherNotificationListener defaultListener =
+                new FileWatcherNotificationListener(file -> !(file.getPath().toString().contains(".codenvy")
```
garagatyi, line 111: Consider removing of negation of expression in braces - it is hard for reading.

garagatyi, line 111: @dkuleshov @vparfonov  What do you think about moving that component out of ProjectManager class. It is not obvious for me why it is placed here. Git and other components also related to it, not only project manager.

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectManager.java
```diff
@@ -109,16 +107,22 @@ public ProjectManager(VirtualFileSystemProvider vfsProvider,
...
+                                                              || file.getPath().toString().contains(".#"))) {
```
garagatyi, line 112: AFAIK git performs a lot of file operation in .git folder. We had hard time with git on GlusterFS because of that.
So I would consider exclusion of .git folder from tracking (except HEAD file or other files we may need to track).
Also we can improve that by binding filters to track/untrack some files/paths.

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/impl/file/event/EventTreeHelper.java
```diff
@@ -0,0 +1,93 @@
...
+ * Helper for event tree related operations.
+ *
+ * @see {@link EventTreeNode}
+ *
+ * @author Dmitry Kuleshov
+ *
+ * @since 4.5
+ */
+@Beta
+class EventTreeHelper {
```
garagatyi: Is it on purpose that this class is package-private? Does it make sense to use public modifier since some components may add their own event subscribers?

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/impl/file/event/EventTreeNode.java
```diff
@@ -0,0 +1,186 @@
...
+        this.events = new LinkedHashMap<>();
+        this.children = new LinkedList<>();
+        this.type = UNDEFINED;
+    }
+
+    public static EventTreeNode newRootInstance() {
+        return EventTreeNode.newInstance().withName(ROOT_NODE_NAME);
+    }
+
+    public static EventTreeNode newInstance() {
```
garagatyi, line 87: Can you explain why static factory method is more preferable that usage of constructor?

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/impl/file/event/VfsEventService.java
dkulieshov: Well, in general case you could be right, but since we're using dependency injection especially combined with Guice implementation it is quite obvious in this case (and quite enough of course) to benefit of controlling thread behaviour using @PostConstruct and @PreDestroy instead of using daemons.

To name a thread - good idea.

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/impl/file/event/VfsEventService.java
dkulieshov, line 30: Honestly, I don't see the necessity in describing that abstract class containing one trivial operation but if you do really need that description I do not mind if you add it by yourself. I foresee that you will not see the description for a few more classes so consider adding description to `org.eclipse.che.api.vfs.impl.file.event.HiEventDetectorManager`,
`org.eclipse.che.api.vfs.impl.file.event.LoEventListener`,
`org.eclipse.che.api.vfs.impl.file.event.LoEventQueueHolder`, 
and 
`org.eclipse.che.api.vfs.impl.file.event.HiEventBroadcaster` as well.

plugins/plugin-maven/che-plugin-maven-server/src/main/java/org/eclipse/che/plugin/maven/server/core/project/PomChangeListener.java
dkulieshov, line 76: That is really nice of you that you've also noticed that.

wsagent/che-core-api-project-shared/src/main/java/org/eclipse/che/api/project/shared/dto/event/GitBranchCheckoutEventDto.java
dkulieshov, line 18: Yes, that is exactly what they are saying [here](https://git-scm.com/docs/git-checkout). And yes, it is an example, it is not really intended to be production ready. You are right twice in a row! But anyway if you could pay a little attention to the name of the interface you would notice that it is named `GitBranchCheckoutEventDto`, I'm pointing to words **branch** and **checkout**. Thus it seems obvious for me that DTO is related to branch checkout events only. Still I don't mind if you add a more explanatory description.

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectManager.java
dkulieshov, line 111: It is out of scope of this pull request, but indeed this whole section will be moved and rearranged to comply with dependency injection pattern with the help of Guice. That will basically allow us to easily configure filters for `FileTreeWatcher` for example.

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/impl/file/event/EventTreeHelper.java
dkulieshov: I just don't get what you mean

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/impl/file/event/EventTreeNode.java
dkulieshov, line 87: Well dude, there is a lot of information over the internet, just type something like 'static factory method vs constructor java', so I don't think I want to spend time explaining you that, but you can get general idea by reading this [article](https://jlordiales.wordpress.com/2012/12/26/static-factory-methods-vs-traditional-constructors). And just subjectively, I like how it looks in the code.

gazarenkov:
GO for it

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/impl/file/event/GitCheckoutHiEventDetector.java
```diff
@@ -0,0 +1,114 @@
...
+import static java.util.Optional.empty;
+import static java.util.regex.Pattern.compile;
+import static org.eclipse.che.api.project.shared.dto.event.FileWatcherEventType.MODIFIED;
+import static org.eclipse.che.api.vfs.impl.file.event.HiEvent.Category.UNDEFINED;
+import static org.eclipse.che.dto.server.DtoFactory.newDto;
+import static org.slf4j.LoggerFactory.getLogger;
+
+/**
+ * Detects if there was a .git/HEAD file modification, which is a sign of git branch
+ * checkout operation, though in some rare cases it simply shows that the head of
```
garagatyi, line 36: HEAD shows reference to current state of working directory. It can be commit or branch/tag (some reference to commit )

dkulieshov, line 36: You are absolutely right, that is why your words are represented in a description: 

> Detects if there was a `.git/HEAD` file modification, which is a sign of git branch checkout operation, though in some rare cases it simply shows that the head of  current branch is changed. By the moment of this class creation those situations are considered rare enough to ignore false detections.

In general as I told before it is an example class which is planned to be reorganized when things will come to real production usage.

vparfonov:
ok

PR #1639, opened 2016-07-04
CHE-1408:Add check for amend before throwing error

vinokurig:
When we perform commit without staged changes we will get an error.
If it is needed to change the message of last commit we should allow
to do it, even if there are no staged changes. @skabashnyuk please review

### DIFF
--- a/wsagent/che-core-api-git/src/test/java/org/eclipse/che/git/impl/CommitTest.java
+++ b/wsagent/che-core-api-git/src/test/java/org/eclipse/che/git/impl/CommitTest.java
@@ -96,7 +96,7 @@ public class CommitTest {
         addFile(connection, "README.txt", CONTENT);
         connection.add(newDto(AddRequest.class).withFilepattern(ImmutableList.of("README.txt")));
         connection.commit(newDto(CommitRequest.class).withMessage("Initial addd"));
-        int beforeCount = connection.log(newDto(LogRequest.class)).getCommits().size();
+        int beforeCommitsCount = connection.log(newDto(LogRequest.class)).getCommits().size();
 
         //when
         //change existing README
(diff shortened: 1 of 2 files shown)


wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -524,6 +524,8 @@ public Revision commit(CommitRequest request) throws GitException {
         try {
             String message = request.getMessage();
             GitUser committer = getUser();
+            boolean isAmend = request.isAmend();
+            boolean isAll = request.isAll();
```
skabashnyuk: Do not need to extract this variables. You can keep it in request

vinokurig: done

skabashnyuk:
ok. @sleshchenko ?

wsagent/che-core-api-git/src/test/java/org/eclipse/che/git/impl/CommitTest.java
```diff
@@ -111,6 +111,22 @@ public void testAmendCommit(GitConnectionFactory connectionFactory) throws GitEx
...
+        GitConnection connection = connectToGitRepositoryWithContent(connectionFactory, repository);
+        int commitsBefore = connection.log(newDto(LogRequest.class)).getCommits().size();
+
+        //when
+        CommitRequest commitRequest = newDto(CommitRequest.class).withMessage("Changed message").withAmend(true);
+        connection.commit(commitRequest);
+
+        //then
+        int commitsAfter = connection.log(newDto(LogRequest.class)).getCommits().size();
+        assertEquals(commitsBefore, commitsAfter);
```
sleshchenko: Looks like you check equals of commits before and after your commit. But in fact you compare only commits number. As for me it would be better to rename this variable to make this aspect clear

wsagent/che-core-api-git/src/test/java/org/eclipse/che/git/impl/CommitTest.java
```diff
@@ -111,6 +111,22 @@ public void testAmendCommit(GitConnectionFactory connectionFactory) throws GitEx
...
     }
 
+    @Test(dataProvider = "GitConnectionFactory", dataProviderClass = org.eclipse.che.git.impl.GitConnectionFactoryProvider.class)
+    public void testChangeMessageOfLastCommit(GitConnectionFactory connectionFactory) throws GitException, IOException {
+        //given
+        GitConnection connection = connectToGitRepositoryWithContent(connectionFactory, repository);
+        int commitsBefore = connection.log(newDto(LogRequest.class)).getCommits().size();
+
+        //when
+        CommitRequest commitRequest = newDto(CommitRequest.class).withMessage("Changed message").withAmend(true);
```
sleshchenko, line 124: mb it would be better to make some commit in given sesction to make sure that you really change commit message

sleshchenko:
other OK

wsagent/che-core-api-git/src/test/java/org/eclipse/che/git/impl/CommitTest.java
vinokurig: done

wsagent/che-core-api-git/src/test/java/org/eclipse/che/git/impl/CommitTest.java
vinokurig, line 124: done

PR #1643, opened 2016-07-04
CHE-117: Fix viewing some git errors on client

vinokurig:
Throw `GitException` instead `IllegalArgumentException` because `IllegalArgumentException` can not be serialized in the client side. @skabashnyuk please review


vzhukovs:
You can add runtime exception mapper like it did in this PR (https://github.com/eclipse/che/pull/1547).
So, then you'll be able to deserialize the service error. WDYT?

vinokurig:
We need more understandable exceptions instead `GitException`, but it is another question. @skabashnyuk  @sleshchenko What do you think?

sleshchenko:
I think that throwing RuntimeException from our services is unnormal situation. It is OK for to use GitException here

skabashnyuk:
ok

PR #1645, opened 2016-07-05
Reveal new java source file

vzhukovs:
After new java source file creation the last ones will be revealed in project tree. A part of CHE-907.

@vparfonov review it plz.


vparfonov:
ok

PR #1646, opened 2016-07-05
Fix upload file and folder feature

vzhukovs:
Where fixed possible NPE and for now upload file and upload folder is working.

Related issue: CODENVY-721

@vparfonov review it plz.


vparfonov:
ok

PR #1658, opened 2016-07-05
CHE-1248 remove workspace information from EnvironmentContext

mshaposhnik:


wsagent/che-core-api-git/pom.xml
```diff
@@ -87,6 +87,10 @@
             <artifactId>che-core-commons-lang</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>wsagent-local</artifactId>
```
sleshchenko, line 91: che-core-api-git will be used for codenvy application also. Are you sure that it is OK that wsagent-local will be used there?

sleshchenko:
fix java doc here https://github.com/eclipse/che/pull/1658/files#diff-ce3c54a447aeb1135e810a8cfe16f30dR31

wsagent/che-core-api-git/pom.xml
mshaposhnik, line 91: Ok for now.

skabashnyuk:
ok

musienko-maxim:

PR #1659, opened 2016-07-05
Avoid getting possible null pointer exception.

vzhukovs:
In some cases, during the file creation we could get a NPE. This fix avoid such situation.

@vparfonov review it, plz.


vparfonov:
ok

PR #1662, opened 2016-07-05
Fix docker system info model class.

garagatyi:
Also add comment with missing fields. 
I haven't added these fields because it is not the target of this PR.

@skabashnyuk @riuvshin @mmorhun Please review

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/SystemInfo.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/SystemInfo.java
@@ -13,9 +13,11 @@ package org.eclipse.che.plugin.docker.client.json;
 import org.eclipse.che.commons.lang.Size;
 
 import java.util.Arrays;
+import java.util.Objects;
 
 /**
  * @author andrew00x
+ * @author Alexander Garagatyi
  */
 public class SystemInfo {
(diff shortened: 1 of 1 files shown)


plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/SystemInfo.java
```diff
@@ -303,92 +343,150 @@ public void setName(String name) {
...
+               && Objects.equals(operatingSystem, that.operatingSystem)
+               && Objects.equals(systemTime, that.systemTime)
+               && Arrays.equals(systemStatus, that.systemStatus);
     }
 
-    public void setNoProxy(String noProxy) {
-        this.noProxy = noProxy;
+    @Override
+    public int hashCode() {
+        int hash = 7;
```
mmorhun: Why not `Objects.hash()` ?

benoitf: yep looks like half of the class is about equals and hashcode

garagatyi: Changed to less verbose version of equals and hashcode

mmorhun:
+1

AndrienkoAleksandr:
+1

PR #1663, opened 2016-07-05
Fix docker client on some systems

garagatyi:
Fix DockerConnector to avoid failing tests on some systems.
Use Future to prevent undefined order of thread execution.
Also change some exception message.
Also disable useless warnings of IDEs in some tests.


garagatyi:
@skabashnyuk @evoevodin @mmorhun Please review

mmorhun:
+1

PR #1672, opened 2016-07-06
Always set root as parent in create project wizard

vzhukovs:
Now when create project wizard appears, the parent is always set to root instead of selected item in application context.

Related issue: CHE-1455

@vparfonov review it, plz.


vparfonov:
ok

PR #1673, opened 2016-07-06
Allow to configure classpath for the plain java projects

vzhukovs:
After project import project was not appeared in application context and project configuration wizard didn't know everything about imported project, so that's why select source folder wizard wasn't able to be shown. This PR fix this situation, so that wizard is available in anytime.

Related issue: CHE-1446

@vparfonov review it, plz.


vparfonov:
ok

PR #1675, opened 2016-07-06
Fix disappearing a folder after changing content in pom.xml

vzhukovs:
Intermediate buffer was added to collect nodes to load after specific time (batch operation). Several requests may occurred to load the same node at once and to reduce the load counts we perform load ones a time.

Related issue: CHE-1448

@vparfonov review it, plz.


ide/che-core-ide-ui/src/main/java/org/eclipse/che/ide/ui/smartTree/NodeLoader.java
```diff
@@ -62,6 +63,31 @@
...
+        @Override
+        public void onExecute() {
+            for (Map.Entry<Node, Boolean> entry : loadQueue.entrySet()) {
+                if (childRequested.containsKey(entry.getKey())) {
+                    continue;
+                }
+
+                childRequested.put(entry.getKey(), entry.getValue());
+
+                _load(entry.getKey());
```
vzhukovs: Fixed

vparfonov:
thx, ok

PR #1677, opened 2016-07-07
CODENVY-556 Do not allow HTTP recipes if working over HTTPS

mkuznyetsov:
We have problems with accessing HTTP recipes, if Che/Codenvy is working over HTTPS.
So it is proposed to not allow having workspaces with such recipes.

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/DefaultWorkspaceValidator.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/DefaultWorkspaceValidator.java
@@ -130,7 +130,7 @@ public class DefaultWorkspaceValidator implements WorkspaceValidator {
         checkNotNull(machineCfg.getSource(), "Environment " + envName + " contains machine without source");
 
         if (apiEndpoint.startsWith("https://")
-            && machineCfg.getSource().getType() == "recipe"
+            && machineCfg.getSource().getType() == "dockerfile"
             && machineCfg.getSource().getLocation() != null) {
             checkArgument(machineCfg.getSource().getLocation().startsWith("https://"),
                           "Environment '%s' contains source recipe with http location, which is not allowed", envName);
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/DefaultWorkspaceValidator.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/DefaultWorkspaceValidator.java
@@ -16,6 +16,7 @@ import com.google.inject.name.Named;
 import org.eclipse.che.api.core.BadRequestException;
 import org.eclipse.che.api.core.model.machine.Command;
 import org.eclipse.che.api.core.model.machine.MachineConfig;
+import org.eclipse.che.api.core.model.machine.MachineSource;
 import org.eclipse.che.api.core.model.machine.ServerConf;
 import org.eclipse.che.api.core.model.workspace.Environment;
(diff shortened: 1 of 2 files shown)


benoitf:
I don't see why we shouldn't use http to provide workspaces recipe ?
also for this issue, I think client should never reach them directly as well (even https) but through server operation.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/DefaultWorkspaceValidator.java
```diff
@@ -123,6 +127,10 @@ private void validateEnv(Environment environment, String workspaceName) throws B
     private void validateMachine(MachineConfig machineCfg, String envName) throws BadRequestException {
         checkArgument(!isNullOrEmpty(machineCfg.getName()), "Environment %s contains machine with null or empty name", envName);
         checkNotNull(machineCfg.getSource(), "Environment " + envName + " contains machine without source");
+        if (apiEndpoint.startsWith("https://") && machineCfg.getSource().getLocation() != null) {
+            checkArgument(!machineCfg.getSource().getLocation().startsWith("http://"),
```
voievodin: As for me it will be more natural to use 

``` java
 checkArgument(machineCfg.getSource().getLocation().startsWith("https://"), ...)
```

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/DefaultWorkspaceValidator.java
```diff
@@ -123,6 +127,10 @@ private void validateEnv(Environment environment, String workspaceName) throws B
     private void validateMachine(MachineConfig machineCfg, String envName) throws BadRequestException {
         checkArgument(!isNullOrEmpty(machineCfg.getName()), "Environment %s contains machine with null or empty name", envName);
         checkNotNull(machineCfg.getSource(), "Environment " + envName + " contains machine without source");
+        if (apiEndpoint.startsWith("https://") && machineCfg.getSource().getLocation() != null) {
```
voievodin: Check the location only if the type of the machine source is `recipe` [javadoc](https://github.com/eclipse/che/blob/ae4c5521bf2202740fc211ee3d2e94881e414d59/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/MachineSource.java#L19-L19)

mkuznyetsov: ok

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/DefaultWorkspaceValidator.java
mkuznyetsov: ok

mkuznyetsov:
@benoitf good point, @skabashnyuk WDYT about that?

skabashnyuk:
> I don't see why we shouldn't use http to provide workspaces recipe ?

Because it conflicts with client logic that can work on https. And we have no plans to rewrite it. We are not planning to create proxy service to fetch http based recipes also

benoitf:
service is already existing as recipes are fetched on the server side to create the workspaces. (If recipes were not downloaded we couldn't create workspaces with Machine Implementation)
So workspace service is working with http or https. Then I don't see why server should then forbid a valid usecase on the server.

mkuznyetsov:
Ok, created new pr with usage of service https://github.com/eclipse/che/pull/1717

PR #1679, opened 2016-07-07
CODENVY-651 Save machine config with 'dev' instead of 'isDev' field

mkuznyetsov:
change MachineConfigImpl field 'isDev' to 'dev', as it is done in MachineConfigDto.
It should fix consitensy problems.

Also it needs to make migrations of existing workspaces.


garagatyi:
LGTM

skabashnyuk:
ok. @evoevodin ?

voievodin:
ok, don't forget to fix [this](https://github.com/eclipse/che/pull/1509/commits/34d34b70717c15fed9e6b34fe0899d4e70b177ae)

skabashnyuk:
ok. do not merge until 4.6

riuvshin:
agree with @skabashnyuk, do not merge this please to until we release 4.5. And do not forget prepare issue with migration script.

musienko-maxim:

PR #1683, opened 2016-07-07
Use a docker launcher image to simplify start/stop/update/restart commands

l0rd:
The goal of this PR is to simplify the `docker run` command to start/stop/update/restart Che. To test it:
- Build the launcher image: `docker build -t codenvy/che:launcher -f Dockerfile.launcher .`
- Run Che: `docker run --rm -v /var/run/docker.sock:/var/run/docker.sock codenvy/che-launcher start`
- Update Che: `docker run --rm -v /var/run/docker.sock:/var/run/docker.sock codenvy/che-launcher update`
- Restart Che: `docker run --rm -v /var/run/docker.sock:/var/run/docker.sock codenvy/che-launcher restart`
- Stop Che: `docker run --rm -v /var/run/docker.sock:/var/run/docker.sock codenvy/che-launcher stop`

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -106,9 +106,9 @@ Vagrant.configure(2) do |config|
       local progress=$1
       local command=$2
       shift 2
-      
+
       local pid=""
-      
+
       case "$progress" in
         extended)
           # simulate tty environment to get full output of progress bars and percentages
@@ -125,7 +125,7 @@ Vagrant.configure(2) do |config|
           ;;
       esac
     }
-    
+
     echo "------------------------------------"
     echo "ECLIPSE CHE: UPGRADING DOCKER ENGINE"
     echo "------------------------------------"
@@ -136,7 +136,7 @@ Vagrant.configure(2) do |config|
     perform $PROVISION_PROGRESS sudo yum -y update docker-engine
 
     echo $(docker --version)
- 
+
     # Add the 'vagrant' user to the 'docker' group
     usermod -aG docker vagrant &>/dev/null
 
@@ -178,7 +178,7 @@ Vagrant.configure(2) do |config|
               `codenvy/che:${CHE_VERSION} --remote:${IP} --port:${PORT} run &>/dev/null
   SHELL
 
-  config.vm.provision "shell" do |s| 
+  config.vm.provision "shell" do |s|
     s.inline = $script
     s.args = [$http_proxy, $https_proxy, $no_proxy, $che_version, $ip, $containerPort, $provisionProgress]
   end
@@ -198,20 +198,19 @@ Vagrant.configure(2) do |config|
     CHE_URL="http://${IP}:${PORT}"
 
     # Test the default dashboard page to see when it returns a non-error value.
-    # Che is active once it returns success        
+    # Che is active once it returns success
     while [ true ]; do
       printf "#"
-      curl -v ${CHE_URL}/dashboard &>/dev/null
-      exitcode=$?
-      if [ $exitcode == "0" ]; then
+      http_status_code=$(curl -I ${CHE_URL}/api/ -s -o /dev/null --write-out "%{http_code}")
+      if [ "$http_status_code" == "200" ]; then
         echo "${CHE_URL}" > /home/user/che/.che_url
         echo "${MAPPED_PORT}" > /home/user/che/.che_host_port
         echo "---------------------------------------"
         echo "ECLIPSE CHE: BOOTED AND REACHABLE"
         echo "ECLIPSE CHE: ${CHE_URL}      "
         echo "---------------------------------------"
-        exit 0             
-      fi 
+        exit 0
+      fi
       sleep 10
     done
   SHELL
--- a/dockerfiles/che-cli/files-to-install/che.sh
+++ b/dockerfiles/che-cli/files-to-install/che.sh
@@ -1,0 +1,158 @@
+#!/bin/sh
+# Copyright (c) 2012-2016 Codenvy, S.A.
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License v1.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-v10.html
+
+init_logging() {
+  BLUE='\033[1;34m'
+  GREEN='\033[0;32m'
+  RED='\033[0;31m'
+  NC='\033[0m'
+}
+
+init_global_variables() {
+
+  CHE_LAUNCHER_CONTAINER_NAME="che-launcher"
+  CHE_LAUNCHER_IMAGE_NAME="che-launcher"
+
+  # User configurable variables
+  DEFAULT_CHE_VERSION="latest"
+  DEFAULT_CHE_CLI_ACTION="help"
+  
+  CHE_VERSION=${CHE_VERSION:-${DEFAULT_CHE_VERSION}}
+  CHE_CLI_ACTION=${CHE_CLI_ACTION:-${DEFAULT_CHE_CLI_ACTION}}
+  
+  USAGE="
+Usage:
+  che [COMMAND]
+     start                              Starts Che server
+     stop                               Stops Che server
+     restart                            Restart Che server
+     update                             Pull latest version of ${CHE_LAUNCHER_IMAGE_NAME}
+     info                               Print some debugging information
+
+Docs: http://eclipse.org/che/getting-started.
+"
+}
+
+usage () {
+  printf "%s" "${USAGE}"
+}
+
+info() {
+  printf  "${GREEN}INFO:${NC} %s\n" "${1}"
+}
+
+debug() {
+  printf  "${BLUE}DEBUG:${NC} %s\n" "${1}"
+}
+
+error() {
+  printf  "${RED}ERROR:${NC} %s\n" "${1}"
+}
+
+error_exit() {
+  echo  "---------------------------------------"
+  error "!!!"
+  error "!!! ${1}"
+  error "!!!"
+  echo  "---------------------------------------"
+  exit 1
+}
+
+check_docker() {
+  if ! docker ps > /dev/null 2>&1; then
+    output=$(docker)
+    error_exit "Error - Docker not installed properly: ${output}"
+  fi
+}
+
+parse_command_line () {
+  for command_line_option in "$@"; do
+    case ${command_line_option} in
+      start|stop|restart|update|info|-h|--help)
+        CHE_CLI_ACTION=${command_line_option}
+      ;;
+      *)
+        # unknown option
+        error_exit "You passed an unknown command line option."
+      ;;
+    esac
+  done
+}
+
+execute_che_launcher() {
+
+  update_che_launcher
+
+  info "ECLIPSE CHE: LAUNCHING LAUNCHER"
+  docker run -it --rm --name "${CHE_LAUNCHER_CONTAINER_NAME}" \
+    -v //var//run//docker.sock://var//run//docker.sock \
+    "${CHE_LAUNCHER_IMAGE_NAME}":"${CHE_VERSION}" "${CHE_CLI_ACTION}" \
+    # > /dev/null 2>&1
+}
+
+execute_command_with_progress() {
+  local progress=$1
+  local command=$2
+  shift 2
+      
+  local pid=""
+  printf "\n"     
+
+  case "$progress" in
+    extended)
+      $command "$@"  
+      ;;
+    basic|*)
+      $command "$@" &>/dev/null &
+      pid=$!
+      while kill -0 "$pid" >/dev/null 2>&1; do
+        printf "#"
+        sleep 10
+      done
+      wait $pid # return pid's exit code
+      printf "\n"
+    ;;
+  esac
+  printf "\n"     
+}
+
+update_che_launcher() {
+  if [ -z "${CHE_VERSION}" ]; then
+    CHE_VERSION=${DEFAULT_CHE_VERSION}
+  fi
+
+  CURRENT_IMAGE=$(docker images -q ${CHE_LAUNCHER_IMAGE_NAME}:${CHE_VERSION})
+
+  if [ "${CURRENT_IMAGE}" != "" ]; then 
+    info "ECLIPSE CHE: ALREADY HAVE IMAGE ${CHE_LAUNCHER_IMAGE_NAME}:${CHE_VERSION}"
+  else 
+    info "ECLIPSE CHE: PULLING IMAGE ${CHE_LAUNCHER_IMAGE_NAME}:${CHE_VERSION}"
+    execute_command_with_progress extended docker pull ${CHE_LAUNCHER_IMAGE_NAME}:${CHE_VERSION}
+    info "ECLIPSE CHE: IMAGE ${CHE_LAUNCHER_IMAGE_NAME}:${CHE_VERSION} INSTALLED"
+  fi
+}
+
+# See: https://sipb.mit.edu/doc/safe-shell/
+set -e
+set -u
+
+init_logging
+check_docker
+init_global_variables
+parse_command_line "$@"
+
+case ${CHE_CLI_ACTION} in
+  start|stop|restart|update|info)
+    execute_che_launcher
+  ;;
+  update-cli)
+    update_che_launcher
+  ;;
+  help)
+    usage
+  ;;
+esac
--- a/dockerfiles/che-cli/files-to-install/che.sh
+++ b/dockerfiles/che-cli/files-to-install/che.sh
@@ -20,10 +20,10 @@ init_global_variables() {
   # User configurable variables
   DEFAULT_CHE_VERSION="latest"
   DEFAULT_CHE_CLI_ACTION="help"
-  
+
   CHE_VERSION=${CHE_VERSION:-${DEFAULT_CHE_VERSION}}
   CHE_CLI_ACTION=${CHE_CLI_ACTION:-${DEFAULT_CHE_CLI_ACTION}}
-  
+
   USAGE="
 Usage:
   che [COMMAND]
@@ -98,13 +98,13 @@ execute_command_with_progress() {
   local progress=$1
   local command=$2
   shift 2
-      
+
   local pid=""
-  printf "\n"     
+  printf "\n"
 
   case "$progress" in
     extended)
-      $command "$@"  
+      $command "$@"
       ;;
     basic|*)
       $command "$@" &>/dev/null &
@@ -117,7 +117,7 @@ execute_command_with_progress() {
       printf "\n"
     ;;
   esac
-  printf "\n"     
+  printf "\n"
 }
 
 update_che_launcher() {
@@ -127,9 +127,9 @@ update_che_launcher() {
 
   CURRENT_IMAGE=$(docker images -q ${CHE_LAUNCHER_IMAGE_NAME}:${CHE_VERSION})
 
-  if [ "${CURRENT_IMAGE}" != "" ]; then 
+  if [ "${CURRENT_IMAGE}" != "" ]; then
     info "ECLIPSE CHE: ALREADY HAVE IMAGE ${CHE_LAUNCHER_IMAGE_NAME}:${CHE_VERSION}"
-  else 
+  else
     info "ECLIPSE CHE: PULLING IMAGE ${CHE_LAUNCHER_IMAGE_NAME}:${CHE_VERSION}"
     execute_command_with_progress extended docker pull ${CHE_LAUNCHER_IMAGE_NAME}:${CHE_VERSION}
     info "ECLIPSE CHE: IMAGE ${CHE_LAUNCHER_IMAGE_NAME}:${CHE_VERSION} INSTALLED"
--- a/dockerfiles/che-launcher/launcher.sh
+++ b/dockerfiles/che-launcher/launcher.sh
@@ -1,4 +1,13 @@
 #!/bin/sh
+# Copyright (c) 2012-2016 Codenvy, S.A., Red Hat, Inc
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License v1.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-v10.html
+#
+# Contributors:
+#   Mario Loriedo - Initial implementation
+#
 
 init_logging() {
   BLUE='\033[1;34m'
@@ -10,8 +19,8 @@ init_logging() {
 init_global_variables() {
 
   CHE_SERVER_CONTAINER_NAME="che-server"
-  CHE_SERVER_IMAGE_NAME="codenvy/che"
-  CHE_LAUNCER_IMAGE_NAME="codenvy/che-launcher"
+  CHE_SERVER_IMAGE_NAME="codenvy/che-server"
+  CHE_LAUNCHER_IMAGE_NAME="codenvy/che-launcher"
 
   # Possible Docker install types are:
   #     native, boot2docker or moby
@@ -22,7 +31,7 @@ init_global_variables() {
   DEFAULT_CHE_HOSTNAME=$(get_che_hostname)
   DEFAULT_CHE_PORT="8080"
   DEFAULT_CHE_VERSION="latest"
-  DEFAULT_CHE_RESTART_POLICY="always"
+  DEFAULT_CHE_RESTART_POLICY="no"
   DEFAULT_CHE_USER="root"
   DEFAULT_CHE_LOG_LEVEL="info"
   DEFAULT_CHE_DATA_FOLDER="/home/user/che"
@@ -42,9 +51,24 @@ init_global_variables() {
   CHE_CONF_ARGS=${CHE_CONF_FOLDER:+-v ${CHE_CONF_FOLDER}:/conf -e \"CHE_LOCAL_CONF_DIR=/conf\"}
   CHE_LOCAL_BINARY_ARGS=${CHE_LOCAL_BINARY:+-v ${CHE_LOCAL_BINARY}:/home/user/che}
 
+  if is_docker_for_mac || is_docker_for_windows; then
+    CHE_STORAGE_ARGS=${CHE_DATA_FOLDER:+-v ${CHE_DATA_FOLDER}/storage:/home/user/che/storage \
+                                        -e \"CHE_WORKSPACE_STORAGE=${CHE_DATA_FOLDER}/workspaces\" \
+                                        -e \"CHE_WORKSPACE_STORAGE_CREATE_FOLDERS=false\"}
+  else
+    CHE_STORAGE_ARGS=${CHE_DATA_FOLDER:+-v ${CHE_DATA_FOLDER}/storage:/home/user/che/storage \
+                                        -v ${CHE_DATA_FOLDER}/workspaces:/home/user/che/workspaces}
+  fi
+
+  if [ "${CHE_LOG_LEVEL}" = "debug" ]; then
+    CHE_DEBUG_OPTION="--debug --log_level:debug"
+  else
+    CHE_DEBUG_OPTION=""
+  fi
+
   USAGE="
 Usage:
-  docker run -v /var/run/docker.sock:/var/run/docker.sock ${CHE_LAUNCER_IMAGE_NAME} [COMMAND]
+  docker run -v /var/run/docker.sock:/var/run/docker.sock ${CHE_LAUNCHER_IMAGE_NAME} [COMMAND]
      start                              Starts Che server
      stop                               Stops Che server
      restart                            Restart Che server
@@ -115,11 +139,46 @@ print_debug_info() {
   debug "---------------------------------------"
 }
 
-get_docker_install_type() {
+is_boot2docker() {
   if uname -r | grep -q 'boot2docker'; then
+    return 0
+  else
+    return 1
+  fi
+}
+
+has_docker_for_windows_ip() {
+  DOCKER_HOST_IP=$(get_docker_host_ip)
+  if [ "${DOCKER_HOST_IP}" = "10.0.75.2" ]; then
+    return 0
+  else
+    return 1
+  fi
+}
+
+is_docker_for_mac() {
+  if uname -r | grep -q 'moby' && ! has_docker_for_windows_ip; then
+    return 0
+  else
+    return 1
+  fi
+}
+
+is_docker_for_windows() {
+  if uname -r | grep -q 'moby' && has_docker_for_windows_ip; then
+    return 0
+  else
+    return 1
+  fi
+}
+
+get_docker_install_type() {
+  if is_boot2docker; then
     echo "boot2docker"
-  elif uname -r | grep -q 'moby'; then
-    echo "moby"
+  elif is_docker_for_windows; then
+    echo "docker4windows"
+  elif is_docker_for_mac; then
+    echo "docker4mac"
   else
     echo "native"
   fi
@@ -127,8 +186,7 @@ get_docker_install_type() {
 
 get_docker_host_ip() {
   NETWORK_IF="eth0"
-  INSTALL_TYPE=$(get_docker_install_type)
-  if [ "${INSTALL_TYPE}" = "boot2docker" ]; then
+  if is_boot2docker; then
     NETWORK_IF="eth1"
   fi
 
@@ -142,12 +200,9 @@ get_docker_host_ip() {
 
 get_che_hostname() {
   INSTALL_TYPE=$(get_docker_install_type)
-  CHE_IP=$(get_docker_host_ip)
-
-  if [ "${INSTALL_TYPE}" = "boot2docker" ]; then
-    echo "${CHE_IP}"
-  elif [[ "${INSTALL_TYPE}" = "moby" && "${CHE_IP}" = "10.0.75.2" ]]; then
-    echo "${CHE_IP}"
+  if [ "${INSTALL_TYPE}" = "boot2docker" ] ||
+     [ "${INSTALL_TYPE}" = "docker4windows" ]; then
+    get_docker_host_ip
   else
     echo "localhost"
   fi
@@ -200,10 +255,12 @@ wait_until_container_is_running() {
 }
 
 server_is_booted() {
-  if ! curl -v http://"${CHE_HOST_IP}":"${CHE_PORT}"/dashboard > /dev/null 2>&1 ; then
-    return 1
-  else
+  HTTP_STATUS_CODE=$(curl -I http://"${CHE_HOST_IP}":"${CHE_PORT}"/api/  \
+                     -s -o /dev/null --write-out "%{http_code}")
+  if [ "${HTTP_STATUS_CODE}" = "200" ]; then
     return 0
+  else
+    return 1
   fi
 }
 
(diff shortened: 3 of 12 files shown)


Dockerfile.launcher
```diff
@@ -0,0 +1,12 @@
+# To build it:
+#  `docker build -t codenvy/che-launcher -f Dockerfile.launcher .`
+# To use it:
+#  `docker run --rm \
+#       -v /var/run/docker.sock:/var/run/docker.sock \
+#        codenvy/che-launcher [start|stop|restart|update]
+FROM docker:1.11
```
garagatyi: If user has docker 1.10 or 1.12 on his machine/server, will he get exception due to incompatible docker client/server?

Dockerfile.launcher
```diff
@@ -0,0 +1,12 @@
...
+
+COPY /assembly/assembly-main/src/assembly/bin/launcher.sh /bin/launcher.sh
+RUN chmod +x /bin/launcher.sh
```
garagatyi: Is it possible to make that script runnable in git and remove that step in dockerfile?

assembly/assembly-main/src/assembly/bin/launcher.sh
```diff
@@ -0,0 +1,126 @@
+#!/bin/sh
+
+# See: https://sipb.mit.edu/doc/safe-shell/
+set -e
+#set +o posix
```
garagatyi: Uncomment or remove please

assembly/assembly-main/src/assembly/bin/launcher.sh
```diff
@@ -0,0 +1,126 @@
...
+
+# Run the finish function if exit signal initiated
+trap exit SIGHUP SIGINT SIGTERM
```
garagatyi: I think you can move `set -e` and `trap` commands down to other commands. It will make reading easier

assembly/assembly-main/src/assembly/bin/launcher.sh
```diff
@@ -0,0 +1,126 @@
...
+
+  # For coloring console output
+  BLUE='\033[1;34m'
+  GREEN='\033[0;32m'
+  NC='\033[0m'
+
+  ### Define various error and usage messages
+  USAGE="
+Usage:
+  launcher [COMMAND]
```
garagatyi: Script has an extension. Please add extension to usage docs or remove extension from file name.

TylerJewell:
I am such a huge fan of this approach to simplifying the run syntax.  I think we have to develop a strategy that inverses some of the files.
1. We need to move codenvy/che --> codenvy/che-server.
1. We need this image to be codenvy/che (not codenvy/che-launcher)
2. We need to find a location in the Che source repository to hold all of our Dockerfiles.  Maybe we should have:
   /docker/Dockerfile.server
   /Dockerfile (represents the launcher)
   /docker/Dockerfile.<others> for when we add Dockerfiles for version check, volume-from, etc.

One issue with this approach is that it is not easy to build the images. The server Dockerfile assumes that there is a Che assembly in a certain sub-directory.  Also the file naming becomes inconsistent.
1. Subsequently, if there are executing scriptions that need to be packaged into each image, how do we store those?  The launcher.sh is specific only to the outer launcher image.  The server image, for example, has /bin/che.sh inside of it for launching the Tomcat server.

assembly/assembly-main/src/assembly/bin/launcher.sh
```diff
@@ -0,0 +1,126 @@
...
+parse_command_line () {
+
+  if [ $# -eq 0 ]; then
+    usage
+    return
+  fi
+
+  for command_line_option in "$@"
+  do
+  case ${command_line_option} in
```
garagatyi: Personally I prefer leaving `do` in the same line where `for` is placed and addition of indentation of body of cycle.
It makes shell script more similar to java code. Please consider such reformatting.

assembly/assembly-main/src/assembly/bin/launcher.sh
```diff
@@ -0,0 +1,126 @@
...
+    start|stop|restart|update)
+      CHE_SERVER_ACTION=${command_line_option}
+    ;;
+    -h|--help)
+      USE_HELP=true
```
garagatyi: I can not find where this variable is used

assembly/assembly-main/src/assembly/bin/launcher.sh
```diff
@@ -0,0 +1,126 @@
...
+
+error_exit () {
+  echo
+  echo "!!!"
+  echo -e "!!! ${1}"
+  echo "!!!"
+  echo
+}
+
+get_docker_host_ip() {
```
garagatyi: Usually get in method name means that it returns something. Please consider another name for the method. Maybe `find` or `detect`.

assembly/assembly-main/src/assembly/bin/launcher.sh
```diff
@@ -0,0 +1,126 @@
...
+
+update_che_server() {
+  docker pull codenvy/che:$CHE_IMAGE_VERSION
+}
+
+init_global_variables
+parse_command_line "$@"
+get_docker_host_ip
+
+if [ "${CHE_SERVER_ACTION}" == "start" ]; then
```
garagatyi: I think case operator will look better here.

assembly/assembly-main/src/assembly/bin/launcher.sh
```diff
@@ -0,0 +1,126 @@
...
+  NC='\033[0m'
+
+  ### Define various error and usage messages
+  USAGE="
+Usage:
+  launcher [COMMAND]
+     start                              Starts Che server in new console
+     stop                               Stops Che server
+     restart                            Restart Che server
+     update                             Pull latest version of codenvy/che
```
garagatyi: Please add that to complete update che container should be restarted

garagatyi:
I like simplified Che management approach. Good job, @l0rd!

Here are some possible improvements (not obviously for that PR):
- we should consider possibility to launch Che from stable channel, so user should be able to choose 
- also running che as a daemon would help some users. I mean starting container with restart policy and maybe provide possibility to store logs at specified file.

garagatyi:
I'm not sure how launcher container will behave if user call start of the server and then press `CTRL-C` at some point. Will it stop the server? Will it react somehow at all?

@l0rd Can you clarify?

l0rd:
@garagatyi thanks for your feedback! I will made changes accordingly and answer to your questions tomorrow.

assembly/assembly-main/src/assembly/bin/launcher.sh
```diff
@@ -0,0 +1,126 @@
...
+}
+
+get_docker_host_ip() {
+  DOCKER_HOST_IP=$(docker run --rm --net host \
+                    alpine sh -c \
+                    "ip a show eth0" | \
+                    grep 'inet ' | \
+                    cut -d/ -f1 | \
+                    awk '{ print $2}')
+}
```
benoitf: with docker for mac it returns the correct ip (I tried with both implementations)

but it doesn't work with docker toolbox for mac

> $ docker run --rm --net host    alpine sh -c ifconfig
> docker0   Link encap:Ethernet  HWaddr 02:42:BD:D7:32:B8  
>           inet addr:172.17.0.1  Bcast:0.0.0.0  Mask:255.255.0.0
>           inet6 addr: fe80::42:bdff:fed7:32b8%32597/64 Scope:Link
>           UP BROADCAST MULTICAST  MTU:1500  Metric:1
>           RX packets:49 errors:0 dropped:0 overruns:0 frame:0
>           TX packets:39 errors:0 dropped:0 overruns:0 carrier:0
>           collisions:0 txqueuelen:0 
>           RX bytes:4277 (4.1 KiB)  TX bytes:5483 (5.3 KiB)
> 
> eth0      Link encap:Ethernet  HWaddr 08:00:27:A2:AF:A6  
>           inet addr:10.0.2.15  Bcast:10.0.2.255  Mask:255.255.255.0
>           inet6 addr: fe80::a00:27ff:fea2:afa6%32597/64 Scope:Link
>           UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
>           RX packets:3609 errors:0 dropped:0 overruns:0 frame:0
>           TX packets:1934 errors:0 dropped:0 overruns:0 carrier:0
>           collisions:0 txqueuelen:1000 
>           RX bytes:2580871 (2.4 MiB)  TX bytes:159603 (155.8 KiB)
> 
> eth1      Link encap:Ethernet  HWaddr 08:00:27:70:6A:D8  
>           inet addr:192.168.99.100  Bcast:192.168.99.255  Mask:255.255.255.0
>           inet6 addr: fe80::a00:27ff:fe70:6ad8%32597/64 Scope:Link
>           UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
>           RX packets:1170 errors:0 dropped:0 overruns:0 frame:0
>           TX packets:909 errors:0 dropped:0 overruns:0 carrier:0
>           collisions:0 txqueuelen:1000 
>           RX bytes:182053 (177.7 KiB)  TX bytes:265568 (259.3 KiB)
> 
> lo        Link encap:Local Loopback  
>           inet addr:127.0.0.1  Mask:255.0.0.0
>           inet6 addr: ::1%32597/128 Scope:Host
>           UP LOOPBACK RUNNING  MTU:65536  Metric:1
>           RX packets:0 errors:0 dropped:0 overruns:0 frame:0
>           TX packets:0 errors:0 dropped:0 overruns:0 carrier:0
>           collisions:0 txqueuelen:1 
>           RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)

the working ip is 192.168.99.100 (eth1) in my case

l0rd: @benoitf thanks for testing. I'll look at that.

Dockerfile.launcher
l0rd: That's a problem. I've tested it with 1.12rc2 and it worked but I got the exception with v1.10. I can think about 2 possible solutions:
- Mount the docker binary as well:
  `docker run -v $(which docker):/usr/local/bin/docker -v /var/run/docker.sock:/var/run/docker.sock codenvy/che-launcher start`
- Use a tool like [dvm](https://github.com/getcarina/dvm) to install multiple Docker clients in the image and update `launch.sh` to test which docker client works well.

What do you think?

Dockerfile.launcher
l0rd: Done

assembly/assembly-main/src/assembly/bin/launcher.sh
l0rd: Removed because `set +o` does not exist for `sh` (and `bash` is not installed on `docker` image)

Dockerfile.launcher
garagatyi: We can also support several tags of launcher image with different versions of docker. WDYT?

assembly/assembly-main/src/assembly/bin/launcher.sh
l0rd: Done

assembly/assembly-main/src/assembly/bin/launcher.sh
l0rd: I've removed that line

assembly/assembly-main/src/assembly/bin/launcher.sh
l0rd: Renamed to find_docker_host_ip

assembly/assembly-main/src/assembly/bin/launcher.sh
l0rd: I agree and I've changed it

assembly/assembly-main/src/assembly/bin/launcher.sh
l0rd: Added a sentence about it in the Usage message.

l0rd:
Che server container is now started in background (`-d` parameter).

`CTRL+C` works if both Che server and Che launcher containers are run with `-ti` parameter (this way Docker forward signal signals to PID 1 process in container). But running it in the background makes more sense in my opinion.

Let me add a function in `launcher.sh` to retrieve the `tag` (i.e. version) of the Che launcher image and set `CHE_IMAGE_VERSION`. Running `codenvy/che-launcher:v123` will spawn `codenvy/che:v123`.

TylerJewell:
@l0rd - if you launch the server in the background where people cannot get the logs, should we:
1. Add in some sort of test like we do in the vagrant files to print out a message in the launcher when the server is ready, what the URL is for the client to reach it?
1. Should we print out a message on where a user can monitor the server startup (which log file is being streamed)?  This is pretty hard because the log file will be in a container that a user will have to get into.

Dockerfile.launcher
l0rd: How would you tag these images with different Docker version? `codenvy/che-launcher:release-docker1.11` ?

My main concern is that users should check their Docker version before running Che and adapt the `docker run` command accordingly. I would prefer copy/paste a simple command without any editing.

Let me check what Rancher and Docker UCP do because they must have the same problem.

l0rd:
@TylerJewell Would you think a message like the following would help?

> Che server has been started. Run `docker logs -f che` to see Che server logs

TylerJewell:
Yeah, definitely that is super useful.  I think that combined with the test to verify the server is up, and the printout of the client URL is perfect.

TylerJewell:
@l0rd - Based upon this [comment](https://github.com/eclipse/che/issues/1482#issuecomment-231402834), it dawns on me that we have many people who are running into issues because the directory where /lib-copy will get mounted is not writable.  In one case, i saw a windows user who just didn't have his directory activated for mounting.  And in a second case, the directory was mountable but not writable.

Should we put in a check to test to make sure that the directories are writable for the following two locations?  If these values are not provided in a che.properties, then we are using ${che.home}.  If they are provided, then they are overrides.

If you do not have write permissions to these folders, this is when the /mnt errors show up, and they are very mysterious to most people.

```
machine.server.terminal.path_to_archive.linux_amd64=${che.home}/lib/linux_amd64/terminal
machine.server.terminal.path_to_archive.linux_arm7=${che.home}/lib/linux_arm7/terminal
machine.server.ext.archive=${che.home}/lib/ws-agent.zip
```

LeDominik:
@TylerJewell I think you also need a configuration for the machine-workspaces location...

TylerJewell:
@LeDominik - it's a separate configuration.  The location of the /storage and /workspaces and /conf/che.properties are done with mounts, and placed to any directory you want (that is writable).  You don't need additional configuration.  This is for two reasons:
1. The che.properties is going into the container, so a basic mount will work.
2. The container will write workspaces created inside of it to the storage location.

So customizing che.properties is essential for changing how the /lib-copy stuff behaves, but not for the others.

assembly/assembly-main/src/assembly/bin/launcher.sh
```diff
@@ -0,0 +1,126 @@
+#!/bin/sh
+
+# See: https://sipb.mit.edu/doc/safe-shell/
+set -e
+trap exit SIGHUP SIGINT SIGTERM
```
garagatyi: What about moving these lines to line 109, WDYT?

assembly/assembly-main/src/assembly/bin/launcher.sh
```diff
@@ -0,0 +1,126 @@
...
+
+  # For coloring console output
+  BLUE='\033[1;34m'
+  GREEN='\033[0;32m'
+  NC='\033[0m'
+
+  ### Define various error and usage messages
+  USAGE="
+Usage:
+  `basename "$0"` [COMMAND]
```
garagatyi: I think that you've already used more modern syntax for expression evaluation `$(.........)`. Please don't mix styles, use modern syntax.

TylerJewell:
Hi @l0rd - I have started a list of to-dos, and also started to lay down some of the requirements that I think we need to zero in on for this to be mergeable.

Note - this comment has been heavily edited.  We now reference #1697 for the broader checklist of items to complete. 

| Image Name | Container Name | Purpose |
| --- | --- | --- |
| `codenvy/che` | `che` | Launcher container |
| `codenvy/che-server` | `che-server` | Launches Che via Tomcat |
| `codenvy/che-version` | `che-version` | Checks for new versions of Che and downloads images |
| `codenvy/che-sdk` | `che-sdk` | A container that can build extensions and custom assemblies |
| `codenvy/che-file` | `che-file` | A container that will run the Chefile utility |
| `codenvy/che-install` | `che-install` | A container that installs [platform scripts / binaries](https://github.com/TylerJewell/che-install). |

The repository of Che would be restructured as:

```
/
  che.sh   # Utility script to simplify executing "docker run ..."
  che.bat  # Utility script
  dockerfiles/
  dockerfiles/<image-name>/Dockerfile
  dockerfiles/<image-name>/<something-for-build>
```

We would remove the default Dockerfile out of the root of the repository and place it in the right /dockerfiles folder. 

List of environment variables for launcher:
- [x] `CHE_HOST_LOCATION`, default = `${che.home}`
- [x] `CHE_WORKSPACES_LOCATION`, default = `${CHE_HOST_LOCATION}/workspaces`
- [ ] `CHE_TERMINAL_MOUNT_LOCATION`, default = `${CHE_HOST_LOCATION}/lib/`
- [ ] `CHE_AGENT_ZIP_LOCATION`, default = `{CHE_HOST_LOCATION}/lib/ws-agent.zip`
- [x] `CHE_DOCKER_HOST_IP`, default = `find_docker_host_ip()` method
- [x] `CHE_PORT`, default = `8080`
- [x] `CHE_VERSION`, default = `latest`
- [x] `CHE_PROPERTIES_LOCATION`, default = /home/user/che
- [x] `CHE_RESTART_POLICY`, default = always

For the policy on che.properties.  We can allow the user to pass in an environment variable which provides a fully qualified path to a local che.properties file. If that is present, we should mount that into the Che server and use that.

For the environment variables that are involved with changing the mount location, if any of the environment variables requires a che.properties modification, take the file passed in by the user and use sed to alter the values so that the environment variables are overrides.  If no che.properties exists, then we can create a new one and append the appropriate property entries.

assembly/assembly-main/src/assembly/bin/launcher.sh
```diff
@@ -0,0 +1,126 @@
...
+     start                              Starts Che server in new console
```
garagatyi: Container is started as detached docker container. I'm not sure it is precise and meaningful to say that it is launched in a new console

Dockerfile.launcher
TylerJewell: The `$(which docker)` syntax will fail on Docker for Windows.  But you can do a string parse using `docker version`. There is go format syntax accepted as part of the command. But this does seem like an inelegant solution.  All this work to simplify the run syntax to just introduce a complicated mount to handle the docker client binary.

Seems like we need to embed smart logic within the launcher.sh to figure out what docker daemon version is running, test for version compatibility, and then use the right docker client within the launcher.

assembly/assembly-main/src/assembly/bin/launcher.sh
garagatyi: Since you don't have exit function in this script you don't need trapping

garagatyi: And actually script fails because of bad trapping on my machine

garagatyi:
If che container is stopped neither `launcher.sh start` nor `launcher.sh restart` is working. Both returns errors. @l0rd Please ensure these situations are handled properly.

garagatyi:
Doesn't work on my linux because of problems with permissions

```
bin[simpler-docker-command !]$ ./launcher.sh start
cabf1976df166d651006a434c9da916f60e71ed604205003f6afded0c2d9292e
bin[simpler-docker-command !]$ docker ps
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
bin[simpler-docker-command !]$ docker ps -a
CONTAINER ID        IMAGE                 COMMAND                  CREATED             STATUS                      PORTS               NAMES
cabf1976df16        codenvy/che:nightly   "/home/user/che/bin/c"   12 seconds ago      Exited (0) 11 seconds ago                       che
bin[simpler-docker-command !]$ docker logs che

!!!
!!! Running 'docker' succeeded, but 'docker ps' failed.
The file /var/run/docker.sock does not have appropriate permissions.
OWNER READ:  r
OWNER WRITE: w
GROUP READ:  r
GROUP WRITE: w
OTHER READ:  -
OTHER WRITE: -
Run 'sudo chmod 666 /var/run/docker.sock' to give the right permissions.
!!!

bin[simpler-docker-command !]$
```

TylerJewell:
@garagatyi - is that because of the launcher, or is that a flaw in the che.sh script where we do the docker ps checks?  In other words, should we check permissions of /var/run/docker.sock in the outer launcher.sh before we attempt to start the Che container?

garagatyi:
Launcher is executed from host user space. Then docker.sock is mounted into Che container. Che container has another user which is not allowed to use mounted socket.
But launcher script even doesn't check if che started successfully or not. But I think it is in the todo list.

TylerJewell:
So what is the fix for when host user and container users are different but accessing same socket?

LeDominik:
@garagatyi, @TylerJewell same permission error also happened yesterday when starting plain nightly che container on OS X Docker Beta. Permissions on the host machine are correctly 666 but not inside the che container.

TylerJewell:
@l0rd - so it seems there are two approaches to solving the docker client / daemon version mismatch issue. For each tag, we create multiple images that represent the Che version and the docker client version. Then the launcher can check the docker daemon version and launch the correct image that matches.  Or, we add logic within the che-server image to figure out which docker client it needs to use based upon an input parameter, and select that client from within the image.  The latter option will keep the image labels cleaner.

assembly/assembly-main/src/assembly/bin/launcher.sh
l0rd: @garagatyi I agree. Do we have the same issue in `che.sh`?

assembly/assembly-main/src/assembly/bin/launcher.sh
l0rd: @garagatyi I'll do that

l0rd:
@TylerJewell @garagatyi The problem of permissions on Linux has always been there. `chmod 666` on `/var/run/docker.sock` is a workaround but I don't like it. To fix it definitively we could:
- Give the possibility to choose start Che server as particular user (e.g. with a an env variable)
- Start Che server as root when no particular user have been provided (e.g. the env variable is empty)

But I prefer using a distinct PR to fix `/var/run/docker.sock` permissions. The problems are unrelated and keep the PR separated would make it easier to code and review.

However I agree that both PR should be merged if we want Docker container to be the default way to run Che.

The same apply to the improvements about the `volumes`. I prefer use a disticnt PR to address them.

I've opened #1697 to track all these improvements.

TylerJewell:
That is good for me.  It looks like we have a dozen or so problems to solve together to make this eligible for merging into master.  I don't mind the wait as we will end up with a very nice packaging when complete.

TylerJewell:
@l0rd - here is some feedback on what I have tested.

First - WOW WOW WOW.  This is absolutely incredible.  

1: There is an issue with building the image on Docker for Windows. It seems that the issue has to do with end of file line endings. I tested building the Dockerfile with both `\r` and `\n` for line endings and got different results, both failing. With the \n as the configuration, it seems that the container is unable to find the launcher.js and gives a Go docker error.  With the \r configuration, it finds the launcher.sh, but provides a long list of bash errors as if it has syntax errors.

2: I think the usage statement can be simplified.

```
Usage:
  docker run -v /var/run/docker.sock:/var/run/docker.sock [COMMAND]
     start                             Starts Che server
     stop                              Stops Che server
     restart                           Restart Che server
     update                            Pull latest version of codenvy/che 
     info                               Print debugging information

Docs: http://eclipse.org/che/getting-started.
```

3: It seems that on docker for windows, I was unable to use `http://localhost:8080`.  So it seems that maybe we should always print out the IP address for how to access Che instead of the hostname.

4: It would be a good idea to move the "get logs" to be right after the CONTAINER STARTED statement.  You can simplify it to:

```
ECLIPSE CHE: See logs at "docker logs -f che"
```

5: Change "CONTAINER IS STARTING" to "CONTAINER STARTING", and I think you delete "CONTAINER STARTED".

l0rd:
This PR is not WIP anymore. 

@TylerJewell: I've included your last suggestions 2) 4) and 5). For 3) it's not so easy because printing the IP would break Docker for Mac. We need to find a way to know if the container run on Docker for Mac or for Windows. Do you mind if we open another issue for that? 😁

@garagatyi I've included your suggestions. Now it works on Linux too (with the same syntax as on other systems) and with versions 10, 11 and 12 of Docker.  The only exception is starting Che on Docker 10: that's not a launcher problem but an issue with the nightly build docker image of Che :sob:. I'll fix that in a separate issue.

@benoitf I have fixed the boot2docker network interface issue 😌 

To try it (should work on MacOS, Windows and Linux):

``` bash
docker build -t che-launcher dockerfiles/che/
docker run -v /var/run/docker.sock:/var/run/docker.sock -e "CHE_VERSION=nightly" che-launcher
```

Currently supported env variables that affect Che configuration:

```
CHE_HOSTNAME: external hostname of the Che dashboard (default depends by the platform)
CHE_PORT: external TCP port of Che dashboard (default is 8080)
CHE_VERSION: version of the Che server Docker image (default is latest)
CHE_RESTART_POLICY: Docker restart policy for Che server (default is always) 
CHE_USER: Docker container user (default is root) 
DOCKER_HOST_IP: This is the IP used by the Che server to contact the wsagents, and by the browser web sockets to contact the wsagents (default is the Docker host IP)
CHE_CONF_FOLDER: Che properties files folder (default is not set)
```

TylerJewell:
@l0rd - I have opened a new tracking issue for the hostname / IP for Docker for Windows and Docker for Mac here: https://github.com/eclipse/che/issues/1744.

TylerJewell:
Added a bunch of  improvements waiting to be added to the PR.  https://github.com/eclipse/che/compare/simpler-docker-command

dockerfiles/che/launcher.sh
```diff
@@ -0,0 +1,336 @@
...
+  GREEN='\033[0;32m'
+  RED='\033[0;31m'
+  NC='\033[0m'
+}
+
+init_global_variables() {
+
+  CHE_SERVER_CONTAINER_NAME="che"
+  CHE_SERVER_IMAGE_NAME="codenvy/che"
+  CHE_LAUNCER_IMAGE_NAME="codenvy/che-launcher"
```
garagatyi: Missing **h** in variable name

dockerfiles/che/launcher.sh
```diff
@@ -0,0 +1,373 @@
...
+
+  # CHE_CONF_ARGS are the Docker run options that need to be used if users set CHE_CONF_FOLDER:
+  #   - empty if CHE_CONF_FOLDER is not set
+  #   - -v ${CHE_CONF_FOLDER}:/conf -e "CHE_LOCAL_CONF_DIR=/conf" if CHE_CONF_FOLDER is set
+  CHE_CONF_ARGS=${CHE_CONF_FOLDER:+-v ${CHE_CONF_FOLDER}:/conf -e \"CHE_LOCAL_CONF_DIR=/conf\"}
+  CHE_LOCAL_BINARY_ARGS=${CHE_LOCAL_BINARY:+-v ${CHE_LOCAL_BINARY}:/home/user/che}
+
+  USAGE="
+Usage:
+  docker run -v /var/run/docker.sock:/var/run/docker.sock ${CHE_LAUNCER_IMAGE_NAME} [COMMAND]
```
garagatyi: How to run Che under debug?

dockerfiles/che/launcher.sh
```diff
@@ -0,0 +1,373 @@
...
+
+  ELAPSED=0
+  until che_container_is_running || [ ${ELAPSED} -eq "${CONTAINER_START_TIMEOUT}" ]; do
+    sleep 1
+    ELAPSED=$((ELAPSED+1))
+  done
+}
+
+server_is_booted() {
+  if ! curl -v http://"${CHE_HOST_IP}":"${CHE_PORT}"/dashboard > /dev/null 2>&1 ; then
```
garagatyi: please consider check of API war since working dashboard war doesn't mean Che is alive

l0rd: `curl -v http://"${CHE_HOST_IP}":"${CHE_PORT}"/api` is ok? 

Should we fix https://github.com/eclipse/che/blob/master/Vagrantfile#L204 too?

garagatyi: > curl -v http://"${CHE_HOST_IP}":"${CHE_PORT}"/api is ok?

yes

> Should we fix https://github.com/eclipse/che/blob/master/Vagrantfile#L204 too?

I think yes

dockerfiles/che/launcher.sh
l0rd: Thanks. I've fixed that.

dockerfiles/che/launcher.sh
l0rd: I've updated `launcher.sh`. Now Che is started in debug mode if the env variable `CHE_LOG_LEVEL` value is `debug`.

Vagrantfile
```diff
@@ -201,7 +201,7 @@ Vagrant.configure(2) do |config|
     # Che is active once it returns success        
     while [ true ]; do
       printf "#"
-      curl -v ${CHE_URL}/dashboard &>/dev/null
```
garagatyi, line 204: It won't show if API is running. I think you have to parse respond code to be equal to 200. Trailing slash is also needed.

dockerfiles/che-cli/files-to-install/che.sh
```diff
@@ -0,0 +1,158 @@
+#!/bin/sh
```
garagatyi, line 1: @l0rd Can you clarify why we need this script and https://github.com/eclipse/che/pull/1683/files#diff-358fea9821fbcf9770f2820a93c4c5ab. They are pretty similar.

TylerJewell, line 1: It's for the CLI.  `che.bat` and `che.sh` are files that you can install on linux / windows that will simplify the CLI down to `che start` so that you do not have to use docker syntax.  The `che-clie` image installs these files onto your computer. We'll also probably add in some sort of sudo yum install command for these files as well.

assembly/assembly-main/src/assembly/bin/launcher.sh
garagatyi: Yes we have the same bug in che.sh

dockerfiles/che/launcher.sh
```diff
@@ -0,0 +1,380 @@
...
+    info "ECLIPSE CHE: ALREADY HAVE IMAGE ${CHE_SERVER_IMAGE_NAME}:${CHE_VERSION}"
+  else
+    update_che_server
+  fi
+
+  info "ECLIPSE CHE: CONTAINER STARTING"
+  docker run -d --name "${CHE_SERVER_CONTAINER_NAME}" \
+    -v /var/run/docker.sock:/var/run/docker.sock \
+    -v "${CHE_DATA_FOLDER}"/lib:/home/user/che/lib-copy \
+    -v "${CHE_DATA_FOLDER}"/workspaces:/home/user/che/workspaces \
```
garagatyi: Looks like syntax for running che has been changed recently. This mount is not needed. And 2 additional env variable are needed. Please check master.

garagatyi:
Everything else is awesome. What a complex PR. @l0rd super cool contribution!

dockerfiles/che-cli/files-to-install/che.sh
```diff
@@ -0,0 +1,158 @@
...
+Docs: http://eclipse.org/che/getting-started.
+"
+}
+
+usage () {
+  printf "%s" "${USAGE}"
+}
+
+info() {
+  printf  "${GREEN}INFO:${NC} %s\n" "${1}"
```
benoitf, line 45: multiple spaces after printf

dockerfiles/che/launcher.sh
```diff
@@ -0,0 +1,380 @@
...
+  CHE_SERVER_CONTAINER_NAME="che-server"
+  CHE_SERVER_IMAGE_NAME="codenvy/che"
+  CHE_LAUNCHER_IMAGE_NAME="codenvy/che-launcher"
+
+  # Possible Docker install types are:
+  #     native, boot2docker or moby
+  DOCKER_INSTALL_TYPE=$(get_docker_install_type)
+
+  # User configurable variables
+  DEFAULT_DOCKER_HOST_IP=$(get_docker_host_ip)
```
benoitf: is that get_docker_host_ip function could be in its own docker image as well ?
so it can be re-used by other scripts

garagatyi:
And cool improvement from @TylerJewell, good job guys!

dockerfiles/che/launcher.sh
l0rd: I agree. Probably not in this PR but we should try putting shared code in a common che base image to avoid code duplication.

Vagrantfile
l0rd, line 204: You are right. This check only verify if host is responding or not. I've changed it to:  
`curl -I http://"${CHE_HOST_IP}":"${CHE_PORT}"/api/ 2>/dev/null | head -n 1 | cut -d' ' -f2`

dockerfiles/che-cli/files-to-install/che.sh
l0rd, line 45: Trailing spaces removed

dockerfiles/che/launcher.sh
garagatyi: tnx

dockerfiles/che/launcher.sh
benoitf: yes I agree
Also two dockerfiles have the same parent images and install docker in it by duplicating lines
I also think this common stuff should be added in a shared image so that we do not duplicate these layers

garagatyi:
@l0rd @TylerJewell I'm curious how are you going to marge that PR with 30 commits? Are you going to squash it or leave these 30 commits?

dockerfiles/che-launcher/launcher.sh
```diff
@@ -0,0 +1,430 @@
...
+
+  ELAPSED=0
+  until che_container_is_running || [ ${ELAPSED} -eq "${CONTAINER_START_TIMEOUT}" ]; do
+    sleep 1
+    ELAPSED=$((ELAPSED+1))
+  done
+}
+
+server_is_booted() {
+  HTTP_STATUS_CODE=$(curl -I http://"${CHE_HOST_IP}":"${CHE_PORT}"/api/ 2>/dev/null \
```
garagatyi: you can also use specific options of curl instead of manual parsing of output 
`curl http://"${CHE_HOST_IP}":"${CHE_PORT}"/api/ -s -o /dev/null --write-out "%{http_code}"`
where 
-s do not print additional info 
-o /dev/null suppress response body
--write-out "%{http_code}"` print status code

Vagrantfile
garagatyi, line 204: Can you change curl command here to reflect what you've done in launcher.sh

l0rd:
@garagatyi I've updated `curl` command with your suggestions (and I kept option `-I` to just retrieve the header). On both `Vagrantfile` and `launcher.sh`. 

I personally don't mind leaving the 30 commits like that (authors and history are preserved) but if you thinks it's more appropriate I have no problem squashing all of them.

garagatyi:
@skabashnyuk @riuvshin @TylerJewell how do you think is it better to squash commits or we are ok if this goes into master with such a huge history tree?

skabashnyuk:
squash

riuvshin:
--squash

l0rd:
@garagatyi @riuvshin @skabashnyuk done 😃

garagatyi:
@l0rd Thanks!
I think that all commits should be signed because you are not eclipse che committer, but one of them is not signed. Please check that.

l0rd:
The only commit that is not signed is @TylerJewell's commit. Do I need to sign that commit with my signature?

garagatyi:
I believe so. This commit is pushed by you, so it doesn't prove that it is authored by @TylerJewell.

l0rd:
Done

garagatyi:
Thanks

garagatyi:
@l0rd @TylerJewell  is this PR ready or other changes are still needed?

TylerJewell:
Congratulations @l0rd!

PR #1685, opened 2016-07-07
CODENVY-733: add possibility to set JAVA_OPTS for dev machines

riuvshin:
@garagatyi @skabashnyuk @vparfonov please take a look on that.

We have that `JAVA_OPTS` set up in setenv https://github.com/eclipse/che-lib/blob/master/che-tomcat8-slf4j-logback/src/assembly/bin/setenv.sh#L18
Same time we have `JAVA_OPTS` defined in `.bashrc` in our docker base image. According to logic in setenv.sh predefined `JAVA_OPTS` will never be used, that is why ws agent start so long `-Djava.security.egd=file:/dev/./urandom` just ignored.
But main goal of that PR is provide a way to configure http / s proxy configuration for WS agent.

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ext/provider/JavaOptsEnvVariableProvider.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ext/provider/JavaOptsEnvVariableProvider.java
@@ -1,0 +1,35 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.plugin.docker.machine.ext.provider;
+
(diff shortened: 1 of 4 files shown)


plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ext/provider/JavaOptsEnvVariableProvider.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+@Singleton
+public class JavaOptsEnvVariableProvider implements Provider<String> {
+    @Inject
+    @Named("che.machine.java_opts")
+    private String javaOpts;
+
+    @Override
+    public String get() {
+        if (Strings.isNullOrEmpty(javaOpts)) {
```
garagatyi: Field can not be null in that case since it doesn't have Nullable annotation.

garagatyi:
We should consider more general solution,  because it is ugly workaround. 
We should not override whole machine java opts since it can broke something in user's java application inside dev machine. 

for the time being I'm ok with that. LGTM

riuvshin:
I've tried to do that by:
`return DockerInstanceRuntimeInfo.JAVA_OPTS_VARIABLE + "=$JAVA_OPTS " + javaOpts;`
but in that case docker due to some reason not able to properly insert such ENV var, it trying to escape dollar symbol like with `\` and that ENV var becomes not valid.

skabashnyuk:
ok. see @garagatyi  comment in code

riuvshin:
@skabashnyuk @garagatyi I've created issue on docker https://github.com/docker/docker/issues/24441

PR #1687, opened 2016-07-08
Add IDs for consoles buttons

RomanNikitenko:
Add IDs for consoles buttons to have ability to test Consoles panel using Selenium


vparfonov:
ok

PR #1691, opened 2016-07-08
Provide correct project path

vzhukovs:
Now `CurrentProjectPathProvider` provides correct project path based on selected resource in application context.

Related issue: CHE-1475

@vparfonov review it, plz.


vparfonov:
ok

PR #1693, opened 2016-07-08
CHE-1444: fix problems with updating classpath

svor:
@vparfonov take a look pls
Fix for CHE-1444 and CHE-1475


riuvshin:
What a hero! double issue! thanks @svor ! thanks for help on that!

vparfonov:
Ok

PR #1701, opened 2016-07-11
CHE-1412: Fix bug when codenvy/ubuntu_jdk8 image is removed

vinokurig:
When we are creating machine not from recipe, but from local image, the image will be removed after machine is created. So if user creates a machine from codenvy/ubuntu_jdk8 image, the image will be removed.
In the step of creating a snapshot from machine I have added a special prefix to image repository to mark  this image as snapshot.
In the step of creating machine (recovering from snapshot) I have added two checks: 
do not pull if image is a local snapshot;
remove source image only if the image is a snapshot from registry.


garagatyi:
LGTM

mmorhun:
LGTM

PR #1704, opened 2016-07-11
Remove environment before adding it

voievodin:
I think this is refactoring issue as previously environments were stored in map, so _contains + put_ actions were replaced with _contains + add_ for list, which is incorrect.

@garagatyi please review


garagatyi:
LGTM

PR #1707, opened 2016-07-11
CODENVY-739: pass proxy credentials to jgit

dmytro-ndp:
@vinokurig, @evoevodin: please, review this request.

Signed-off-by: Dmytro Nochevnov dnochevnov@codenvy.com

### DIFF
--- a/core/che-core-api-core/src/main/java/org/eclipse/che/commons/proxy/ProxyAuthenticator.java
+++ b/core/che-core-api-core/src/main/java/org/eclipse/che/commons/proxy/ProxyAuthenticator.java
@@ -7,18 +7,30 @@
  *
  * Contributors:
  *   Codenvy, S.A. - initial API and implementation
- *   SAP           - implementation
  *******************************************************************************/
-package org.eclipse.che.git.impl.jgit;
+package org.eclipse.che.commons.proxy;
 
 import com.google.common.base.Strings;
 
 import java.net.Authenticator;
 import java.net.PasswordAuthentication;
 
+import static com.google.common.base.Strings.isNullOrEmpty;
 import static java.lang.String.format;
 
 /**
+ * Set default java.net.Authenticator for requesting through the http(s) proxy.
+ * Proxy credentials are read from the system properties:
+ * http.proxyUser
+ * http.proxyPassword
+ * https.proxyUser
+ * https.proxyPassword
+ *
+ * Usage:
+ * ProxyAuthenticator.initAuthenticator(url)
+ * ... making http(s) request by url ...
+ * ProxyAuthenticator.resetAuthenticator()
+ *
  * @author Dmytro Nochevnov
  */
 public class ProxyAuthenticator extends Authenticator {
@@ -36,14 +48,18 @@ public class ProxyAuthenticator extends Authenticator {
         }
     }
 
-    public enum Protocol {
+    public static void resetAuthenticator() {
+        currentProtocolHolder.remove();
+    }
+
+    private enum Protocol {
         HTTP, HTTPS;
 
-        public PasswordAuthentication passwordAuthentication = createPasswordAuthentication();
+        private PasswordAuthentication passwordAuthentication = createPasswordAuthentication();
 
         private PasswordAuthentication createPasswordAuthentication() {
-            if (! (Strings.isNullOrEmpty(getProxyUserSystemProperty())
-                   || Strings.isNullOrEmpty(getProxyPasswordSystemProperty()))) {
+            if (! (isNullOrEmpty(getProxyUserSystemProperty())
+                   || isNullOrEmpty(getProxyPasswordSystemProperty()))) {
                 return new PasswordAuthentication(getProxyUserSystemProperty(),
                                                   getProxyPasswordSystemProperty().toCharArray());
             } else {
@@ -61,7 +77,7 @@ public class ProxyAuthenticator extends Authenticator {
             return System.getProperty(propertyName);
         }
 
-        public PasswordAuthentication getPasswordAuthentication() {
+        private PasswordAuthentication getPasswordAuthentication() {
             return passwordAuthentication;
         }
     }
--- a/wsagent/che-core-api-git/src/test/java/org/eclipse/che/git/impl/PullTest.java
+++ b/wsagent/che-core-api-git/src/test/java/org/eclipse/che/git/impl/PullTest.java
@@ -137,4 +137,29 @@ public class PullTest {
         PullRequest request = newDto(PullRequest.class);
         connection.pull(request);
     }
+
+    @Test(dataProvider = "GitConnectionFactory", dataProviderClass = GitConnectionFactoryProvider.class,
+        expectedExceptions = GitException.class, expectedExceptionsMessageRegExp = "No remote repository specified.  " +
+                                                                                   "Please, specify either a URL or a remote name from which new revisions should be fetched in request.")
+    public void testWhenThereAreNoAnyRemotesBehindTheProxy(GitConnectionFactory connectionFactory) throws Exception {
+        //given
+        System.setProperty("http.proxyUser", "user1");
+        System.setProperty("http.proxyPassword", "paswd1");
+        System.setProperty("https.proxyUser", "user2");
+        System.setProperty("https.proxyPassword", "paswd2");
+
+        GitConnection connection = connectToInitializedGitRepository(connectionFactory, repository);
+
+        //when
+        PullRequest request = newDto(PullRequest.class);
+        connection.pull(request);
+    }
+
+    @AfterMethod
+    public void tearDown() throws Exception {
+        System.clearProperty("http.proxyUser");
+        System.clearProperty("http.proxyPassword");
+        System.clearProperty("https.proxyUser");
+        System.clearProperty("https.proxyPassword");
+    }
 }
--- a/wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
+++ b/wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
@@ -15,7 +15,6 @@ import com.google.common.io.Files;
 import com.jcraft.jsch.JSch;
 import com.jcraft.jsch.JSchException;
 import com.jcraft.jsch.Session;
-
 import org.apache.commons.io.filefilter.DirectoryFileFilter;
 import org.apache.commons.io.filefilter.IOFileFilter;
 import org.apache.commons.io.filefilter.TrueFileFilter;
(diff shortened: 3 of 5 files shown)


wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -235,6 +238,29 @@
     private final GitUserResolver   userResolver;
     private final Repository        repository;
 
+    // setup authenticators to use JGit behind the proxy
+    private static Authenticator httpsProxyAuth;
+    private static Authenticator httpProxyAuth;
+    static {
+        if (nonNull(System.getProperty("http.proxyUser")) && nonNull(System.getProperty("http.proxyPassword"))) {
```
voievodin: 1. I don't see any reason to use `nonNull(...)` against `!= null`.
2. I don't think that camel cased properties are good choice, i would  preferable change them to `http.proxy_user` and `http.proxy_password`
3. `Authenticator` looks not good as anonymous class to me, i think you  should extract it to the private static _NAMED_ inner class, and it seems that findbugs  will suggest you to do the same.

voievodin:
reviewed, check the comment

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
dmytro-ndp: > (1). I don't see any reason to use nonNull(...) against != null.

Agree.

> (2). I don't think that camel cased properties are good choice, i would preferable change them to http.proxy_user and http.proxy_password

I don't think so. Camel case is good enough here, moreover - there is a convention of Java proxy settings naming which we can't change: http://rolandtapken.de/blog/2012-04/java-process-httpproxyuser-and-httpproxypassword

> (3). Authenticator looks not good as anonymous class to me, i think you should extract it to the private static NAMED inner class, and it seems that findbugs will suggest you to do the same.

Agree.

dmytro-ndp:
Thank you! Please, see my answers.

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
vzhukovs: Why did you use `java.util.Objects#nonNull` instead of `com.google.common.base.Strings#isNullOrEmpty`?
Empty property value may means that this property is not configured, isn't it?

dmytro-ndp: Empty property shouldn't affect the code. But usage of _Strings#isNullOrEmpty_ is definitely great suggestion. Thank you!

voievodin:
Well, then there is nothing to add. LGTM Good Job :+1:

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -1590,13 +1616,22 @@ public void configure(Transport transport) {
                         sshTransport.setSshSessionFactory(sshSessionFactory);
                     }
                 });
+
+                if (nonNull(httpProxyAuth)) {
+                    Authenticator.setDefault(httpProxyAuth);
```
sleshchenko: 1. Please, can you explain why we should use **http**ProxyAuth for **ssh** connections and **https**ProxyAuth for **http/https** ?
2. I guess there can be some problem with concurrency. I mean you can send two requests with ssh and http links. Then there can be situation when before invoking of `command.call()` will be invoked `Authenticator.setDefault(httpsProxyAuth);` for ssh link. Am I wrong?

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -15,7 +15,6 @@
 import com.jcraft.jsch.JSch;
 import com.jcraft.jsch.JSchException;
 import com.jcraft.jsch.Session;
-
```
sleshchenko, line 18: try to use formatter here. I guess it'll revert this change

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
dmytro-ndp: > (1) Please, can you explain why we should use httpProxyAuth for ssh connections and httpsProxyAuth for http/https ?

You found an error, and we even don't need setup proxy for ssh.

> (2) I guess there can be some problem with concurrency. I mean you can send two requests with ssh and http links. Then there can be situation when before invoking of command.call() will be invoked Authenticator.setDefault(httpsProxyAuth); for ssh link. Am I wrong?

You are right, but I would prefer don't work around synchronization of invoking authenticator so as it will lead to redundant delays.

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
dmytro-ndp, line 18: I will try. Thank you.

dmytro-ndp:
@evoevodin, @sleshchenko: please, review my fixup on your notices.

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -235,13 +235,16 @@
...
+    private final ProxyAuthenticator proxyAuthenticator;
+
     @Inject
     JGitConnection(Repository repository, CredentialsLoader credentialsLoader, SshKeyProvider sshKeyProvider,
                    GitUserResolver userResolver) {
         this.repository = repository;
         this.credentialsLoader = credentialsLoader;
         this.sshKeyProvider = sshKeyProvider;
         this.userResolver = userResolver;
+        this.proxyAuthenticator = new ProxyAuthenticator();
```
sleshchenko: Seems that this instance is unused

dmytro-ndp: You are right.

wsagent/che-core-api-git/src/test/java/org/eclipse/che/git/impl/PullTest.java
```diff
@@ -137,4 +137,29 @@ public void testWhenThereAreNoAnyRemotes(GitConnectionFactory connectionFactory)
         PullRequest request = newDto(PullRequest.class);
         connection.pull(request);
     }
+
+    @Test(dataProvider = "GitConnectionFactory", dataProviderClass = GitConnectionFactoryProvider.class,
+        expectedExceptions = GitException.class, expectedExceptionsMessageRegExp = "No remote repository specified.  " +
```
sleshchenko, line 142: Are you sure that this test check your use case?

sleshchenko:
OK for me

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -235,9 +237,8 @@
     private final GitUserResolver   userResolver;
     private final Repository        repository;
 
-    @Inject
-    JGitConnection(Repository repository, CredentialsLoader credentialsLoader, SshKeyProvider sshKeyProvider,
-                   GitUserResolver userResolver) {
+    @Inject JGitConnection(Repository repository, CredentialsLoader credentialsLoader, SshKeyProvider sshKeyProvider,
```
sleshchenko: revert formatting

core/che-core-api-core/src/main/java/org/eclipse/che/commons/proxy/ProxyAuthenticator.java
```diff
@@ -0,0 +1,92 @@
...
+ * https.proxyPassword
+ *
+ * Usage:
+ * ProxyAuthenticator.initAuthenticator(url)
+ * ... making http(s) request by url ...
+ * ProxyAuthenticator.resetAuthenticator()
+ *
+ * @author Dmytro Nochevnov
+ */
+public class ProxyAuthenticator extends Authenticator {
```
dmytro-ndp, line 36: There is a test PullTest.

core/che-core-api-core/src/main/java/org/eclipse/che/commons/proxy/ProxyAuthenticator.java
```diff
@@ -0,0 +1,92 @@
...
+        currentProtocolHolder.remove();
+    }
+
+    private enum Protocol {
+        HTTP, HTTPS;
+
+        private PasswordAuthentication passwordAuthentication = createPasswordAuthentication();
+
+        private PasswordAuthentication createPasswordAuthentication() {
+            if (! (Strings.isNullOrEmpty(getProxyUserSystemProperty())
```
vinokurig: Please use static import of `isNullOrEmpty`

wsagent/che-core-api-git/src/test/java/org/eclipse/che/git/impl/PullTest.java
```diff
@@ -137,4 +137,29 @@ public void testWhenThereAreNoAnyRemotes(GitConnectionFactory connectionFactory)
...
+        System.setProperty("http.proxyUser", "user1");
+        System.setProperty("http.proxyPassword", "paswd1");
+        System.setProperty("https.proxyUser", "user2");
+        System.setProperty("https.proxyPassword", "paswd2");
+
+        GitConnection connection = connectToInitializedGitRepository(connectionFactory, repository);
+
+        //when
+        PullRequest request = newDto(PullRequest.class);
+        connection.pull(request);
```
vinokurig, line 155: Please make this in one line

core/che-core-api-core/src/main/java/org/eclipse/che/commons/proxy/ProxyAuthenticator.java
dmytro-ndp: Thank you for advice.

wsagent/che-core-api-git/src/test/java/org/eclipse/che/git/impl/PullTest.java
dmytro-ndp, line 155: As for me, current variant is more readable.

dmytro-ndp:
@evoevodin: please, review new fixup.

voievodin:
LGTM

dmytro-ndp:
Thank @evoevodin, @sleshchenko, @vinokurig for your helpful notices!

PR #1714, opened 2016-07-12
CODENVY-570: Log warning on recipe exception

vinokurig:
We should not log if failed to create machine because of error in recipe.
@garagatyi please review


wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/MachineManager.java
```diff
@@ -262,7 +263,11 @@ public MachineImpl createMachineAsync(MachineConfig machineConfig,
                                                             machine,
                                                             machineLogger);
                                          } catch (MachineException | NotFoundException e) {
-                                             LOG.error(e.getLocalizedMessage(), e);
+                                             if (e.getCause() instanceof InvalidRecipeException) {
+                                                 LOG.warn(e.getLocalizedMessage());
```
garagatyi: Don't log that situation at all

vinokurig: done

garagatyi:
LGTM

PR #1718, opened 2016-07-12
CHE-1486: fix problem with 'open declaration' action

svor:
Fix for CHE-1486
@vparfonov @vzhukovskii take a look pls


vparfonov:
ok

PR #1720, opened 2016-07-13
CHE-1474: Add update event on close git compare

vinokurig:
When user opens git compare window on modified or new file and makes
some changes in this window, the changes will not be updated in opened
file editor.
Update event solves this problem.
@vzhukovskii @vparfonov please review


vzhukovs:
ok

vparfonov:
Ok

PR #1721, opened 2016-07-13
CHE-1075: Auto select first project after workspace startup

vzhukovs:
This PR fix problem, which could be occurred after factory import, when actions has been performed.

Related issue: CHE-1075

@vparfonov review it, plz.

### DIFF
--- a/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/resources/impl/ResourceManager.java
+++ b/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/resources/impl/ResourceManager.java
@@ -203,9 +203,13 @@ public final class ResourceManager {
                         Project[] tmpProjects = copyOf(projects, projects.length + 1);
                         tmpProjects[projects.length] = project;
                         projects = tmpProjects;
-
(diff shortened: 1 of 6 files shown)


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/resources/impl/ResourceManager.java
```diff
@@ -203,11 +203,15 @@ public ResourceManager(@Assisted DevMachine devMachine,
...
                         tmpProjects[projects.length] = project;
                         projects = tmpProjects;
-
-                        eventBus.fireEvent(new ResourceChangedEvent(new ResourceDeltaImpl(project, ADDED | DERIVED)));
                     }
                 }
 
+                java.util.Arrays.sort(projects);
+
+                for (Project project : projects) {
```
vparfonov, line 212: Need to some coments here, not clear that doing here

vzhukovs, line 212: Added comment.

We need to guarantee that order of returned projects will be sorted according to the `compareTo` method implementation.

vparfonov:
Ok

PR #1727, opened 2016-07-14
CHE-1447: Fix collapse External Libraries after changing pom.xml

vzhukovs:
Were changed mechanism of equality of synthetic nodes. With this PR external libraries don't collapse after modifying the pom.xml.

Related issue: CHE-1447

@vparfonov


vparfonov:
ok

PR #1729, opened 2016-07-14
CHE-1404: make commands widget fixed width, and tooltip for full name, cut name with dots

ashumilova:
CHE-1404: make commands widget fixed width, and tooltip for full name, cut name with dots


vparfonov:
ok

vitaliy-guliy:
Great!

PR #1734, opened 2016-07-14
CHE-1451: Fix collapsing directories while creating new resource

vzhukovs:
This PR fix the problem with collapsing expanded directories while new resource is created

@vparfonov please, review it.


vparfonov:
ok

PR #1747, opened 2016-07-15
CHE-1482: Fix setting up source directory in java project

vzhukovs:
This PR fix possible NPE in case when node transforms from the folder to the source folder.

Related issue: CHE-1482

@vparfonov review it, plz.


vparfonov:
ok

PR #1749, opened 2016-07-15
CHE-1461: Automatically set up project type to blank when keepDir is used

vzhukovs:
This PR automatically provides blank project type to newly created projects which where using keepDir while import.

Related issue: CHE-1461

@vparfonov review it, plz.


vparfonov:
ok

PR #1833, opened 2016-07-18
Downgrade docker installed in Che image

l0rd:
### What does this PR do?

Use Docker version 1.6 in Che image

On a machine with Docker 1.10 or lower Che failed to start in a container.
image
### Previous Behavior

image
### New Behavior

image
### Tests written?

No
### Docs requirements?

No 

### DIFF
--- a/dockerfiles/che-server/Dockerfile
+++ b/dockerfiles/che-server/Dockerfile
@@ -16,13 +16,18 @@ FROM alpine:3.4
 ENV LANG=C.UTF-8 \
     JAVA_HOME=/usr/lib/jvm/default-jvm/jre \
     PATH=${PATH}:${JAVA_HOME}/bin \
-    CHE_HOME=/home/user/che
+    CHE_HOME=/home/user/che \
+    DOCKER_VERSION=1.6.0 \
+    DOCKER_BUCKET=get.docker.com
 
 RUN echo "http://dl-4.alpinelinux.org/alpine/edge/community" >> /etc/apk/repositories && \
     apk upgrade --update && \
-    apk add --update docker openjdk8 sudo bash && \
+    apk add --update ca-certificates curl openssl openjdk8 sudo bash && \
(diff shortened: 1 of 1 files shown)


TylerJewell:
Lgtm

ghost:
@l0rd wouldn't it be possible then that client IP version is lower than the one a user might have locally?

l0rd:
@eivantsov yes that's the goal: docker client version inside the container is lower than docker daemon version on the host. The client does not complain speaking with a daemon that uses a more recent API. And the docker commands we use in the shell scripts works fine.  

That's the same solution used by Rancher agent:  https://github.com/rancher/rancher/blob/master/agent/Dockerfile#L24. They have the same need that we do (a client inside a container that need to speak to daemons of any version) and they solved it using a Docker client version 1.6.

Dockerfile
```diff
@@ -3,13 +3,18 @@ FROM alpine:3.4
...
+    CHE_HOME=/home/user/che \
+    DOCKER_VERSION=1.6.0 \
+    DOCKER_BUCKET=get.docker.com
 
 RUN echo "http://dl-4.alpinelinux.org/alpine/edge/community" >> /etc/apk/repositories && \
     apk upgrade --update && \
-    apk add --update docker openjdk8 sudo bash && \
+    apk add --update ca-certificates curl openssl openjdk8 sudo bash && \
+    curl -sL "https://${DOCKER_BUCKET}/builds/Linux/x86_64/docker-$DOCKER_VERSION" \
+    > /usr/bin/docker; chmod +x /usr/bin/docker && \
```
garagatyi: I think that **> /usr/bin/docker** is an ending of previous command should not be moved to a new line.
Also **; chmod +x /usr/bin/docker** means skip error of previous step. It is better to use **&&** instead

l0rd:
@garagatyi I've updated / splitted the curl command

TylerJewell:
lgtm +1

garagatyi:
LGTM

l0rd:
We should merge this after #1683

TylerJewell:
Thanks, @l0rd for the fix.

Issue #1850, opened 2016-07-19
Opened files list in the editor area does not work properly


AndreyChizhikov:
**Che version:**  
Version :4.5.0-RC1-SNAPSHOT
Revision :f246d649578ea08a
Build Time :2016-07-15 10:42:26

**Additional information:**
For more details see video from the archive:
[openFile.zip](https://github.com/eclipse/che/files/371007/openFile.zip)

PR #1867, opened 2016-07-19
#1850: Remove closed file from the editor's popup list

vzhukovs:
Event calling was reorganized in editor tabs and list item, so after closing file, the last ones also removes from the editor's popup list.

Related issue: https://github.com/eclipse/che/issues/1850

@vparfonov, @RomanNikitenko review it, plz.


vparfonov:
ok

Issue #1819, opened 2016-07-18
Delete multi-module projects in some case holding failure


vparfonov:
Steps to reproduce:
1. Import multi-module project (https://github.com/eclipse/che.git)
2. Try to delete this project without waiting update maven dependencies

[log.txt](https://github.com/eclipse/che/files/369019/log.txt)

riuvshin:
this is still actual

svor:
This bug has already fixed for the localhost, but still reproduced on the cloud version, I've created an issue https://github.com/eclipse/che/issues/1985.

PR #1874, opened 2016-07-20
CHE-1819: fix NPE when unresolved project is deleted

svor:
### What does this PR do?

Fix NPE after deleting unresolved multi-module project. 
Show notification with the number of modules which were deleted.

https://github.com/eclipse/che/issues/1819
### Previous Behavior

After deleting multi-module project the process of resolving dependencies is continued and threw many NPEs
### New Behavior

This content may also be included in the release notes.
After deleting multi-module project: project is deleted without exceptions and the process of resolving dependencies is ended 
### Tests written?

No

Please review @vparfonov

### DIFF
--- a/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/context/AppContextImpl.java
+++ b/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/context/AppContextImpl.java
@@ -414,10 +414,14 @@ public class AppContextImpl implements AppContext,
             }
         }
 
+        if (currentResource == null) {
+            return null;
+        }
+
         Project root = null;
 
         for (Project project : projects) {
-            if (project.getLocation().isPrefixOf(currentResources[0].getLocation())) {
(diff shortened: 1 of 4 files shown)


vparfonov:
ok

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/context/AppContextImpl.java
```diff
@@ -414,6 +414,10 @@ public Project getRootProject() {
             }
         }
 
+        if (currentResources[0] == null) {
```
tolusha: What does it actually mean?

svor: It means that current resource (selected resource) is null. In this class 
`currentResources[0] == currentResource`
I've made some changes and now it's more clear, I hope.

tolusha: It is better now. thx

PR #1877, opened 2016-07-20
CHE-1535: Reveal resources after move and copy operation

vzhukovs:
Now, it is possible to navigate to the copied and moved file after cut/copy/paste operation.

Related issue: CHE-1535

@vparfonov review it, plz.


vparfonov:
ok

PR #1890, opened 2016-07-21
CODENVY-765: Provide correct link to the file content

vzhukovs:
This PR adds the token query part to the file content url.

Related issue: CODENVY-765

@vparfonov review it, plz.

### DIFF
--- a/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/imageviewer/ImageViewer.java
+++ b/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/imageviewer/ImageViewer.java
@@ -31,6 +31,7 @@ import org.eclipse.che.ide.api.editor.EditorAgent.OpenEditorCallback;
 import org.eclipse.che.ide.api.editor.EditorInput;
 import org.eclipse.che.ide.api.event.FileEvent;
 import org.eclipse.che.ide.api.event.FileEventHandler;
+import org.eclipse.che.ide.api.machine.WsAgentURLModifier;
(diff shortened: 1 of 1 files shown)


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/resources/impl/ResourceManager.java
```diff
@@ -820,9 +819,7 @@ private Resource newResourceFrom(final ItemReference reference) {
 
         switch (reference.getType()) {
             case "file":
-                final Link link = reference.getLink(GET_CONTENT_REL);
-
-                return resourceFactory.newFileImpl(path, link.getHref(), this);
+                return resourceFactory.newFileImpl(path, urlModifier.modify(ps.getFileContentUrl(path)), this);
```
vparfonov: we should use link reference provided by server-side

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/project/ProjectServiceClientImpl.java
```diff
@@ -242,6 +242,12 @@ public void apply(PromiseError arg) throws OperationException {
 
     /** {@inheritDoc} */
     @Override
+    public String getFileContentUrl(Path path) {
```
vparfonov: i propose you localize this changes here https://github.com/eclipse/che/blob/master/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/imageviewer/ImageViewer.java#L153-153

vparfonov:
ok

PR #1896, opened 2016-07-21
Fix the ENV about configuration folder to set when using docker image

benoitf:
### What does this PR do?

Fix the environment variable setup on configuration folder
### Previous Behavior

Previously, when using the CHE_CONF_FOLDER to the launcher, there was an invalid argument given to the docker container with quotes
### New Behavior

Variable is given as expected
### Tests written?

No


TylerJewell:
+1 - ok to merge.

PR #1904, opened 2016-07-22
CHE-1361 Artik: Resources monitor (CPU, memory and storage)

vitaliy-guliy:
Display cpu, memory and disk monitors for machine nodes in consoles panel.
Feature was requested for Artik devices. 
https://jira.codenvycorp.com/browse/CHE-1361


plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/processes/monitoring/MachineMonitors.java
```diff
@@ -0,0 +1,110 @@
...
+            widgets = new HashMap<>();
+            monitorWidgets.put(machineID, widgets);
+        }
+
+        MonitorWidget widget = widgets.get(referrer);
+        if (widget == null) {
+            widget = new MonitorWidget();
+            widgets.put(referrer, widget);
+        }
+
```
mshaposhnik, line 51: save the byte - save the nature

vparfonov:
ok

Issue #1825, opened 2016-07-18
After closing the Project wizard widget by ESC select path form still opened


vparfonov:
Step to reproduce:
1. Select some folder
2. In context menu select Convert folder to Project 
3. In opened window select parent path 
4. Push Esc 

Should close Select Path window but closing Project wizard 

![1825](https://cloud.githubusercontent.com/assets/1636592/16978452/98570282-4e64-11e6-8c84-7ec373573cb7.gif)

**Che version:**    4.5.0

TylerJewell:
Please provide reproduction steps + animated GIF.  Cannot be taken into a sprint without this information.

riuvshin:
QA checked. close issue

PR #1916, opened 2016-07-22
CHE-1825: fix problem with the window selection

svor:
### What does this PR do?

Activate last opened window

https://github.com/eclipse/che/issues/1825
### Previous Behavior

Project Wizard window was active when trying to select the path to the parent folder.
### New Behavior

Select Path window is active when it is opened.
### Tests written?

No

@vparfonov please review


vparfonov:
ok

PR #1917, opened 2016-07-22
Allow to continue to start the script even if there is a chown failure

benoitf:
### What does this PR do?

Allow to continue to start the script even if there is a chown failure

Keep docker image booting if chown issue
### Previous Behavior

When mounting a local folder with .git we have the following warnings
chown: changing permission on ....git.idx : permission denied
### New Behavior

we still see warning but we continue to start docker image
### Tests written?

No


l0rd:
+1

TylerJewell:
+1

PR #1926, opened 2016-07-25
CHE-1146 Factory jpa integration and structural refactoring

akorneta:
### What does this PR do?

Adds factory model objects, dtos, impls;
Adds factory manager;
Reworks factory dao interface;
Adds a new FactoryDao implementation based on JPA;
Adds tests for FactoryDao interface contract(TCKs).

https://github.com/codenvy/codenvy/issues/392
### Tests written?

Yes

@skabashnyuk, @evoevodin, @sleshchenko please review


wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryManager.java
```diff
@@ -0,0 +1,286 @@
...
+
+    /**
+     * Updates factory accordance to the new configuration.
+     *
+     * <p>Note: Updating uses replacement strategy,
+     * therefore existing factory would be replaced with given update {@code update}
+     *
+     * @param update
+     *         factory update
+     * @return updated factory with links
```
sleshchenko: are you sure that manager'll return factory with links

skabashnyuk:
1. Logger for find method
2. Rework IDE actions storage

akorneta:
1. Added logger of FactoryDao#findByAttribute.
2. We decide do not change IDE action store mechanism.(the model can be extended in that case we would have to do the same implementation mapping)

skabashnyuk:
ok

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/LegacyConverter.java
```diff
@@ -11,13 +11,13 @@
...
+import org.eclipse.che.api.factory.shared.dto.FactoryDto;
 
 /**
  * Convert legacy factory parameter to new the latest format
  *
  * @author Alexander Garagatyi
  */
 public interface LegacyConverter {
-    void convert(Factory factory) throws ApiException;
+    void convert(FactoryDto factory) throws ApiException;
```
sleshchenko: Maybe it should use FactoryImpl instead of dto?

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/jpa/JpaFactoryDao.java
```diff
@@ -0,0 +1,155 @@
...
+            throw new NotFoundException(format("Could not update factory with id %s because it doesn't exist", update.getId()));
+        }
+        return manager.merge(update);
+    }
+
+    @Transactional
+    protected void doRemove(String id) {
+        final EntityManager manager = managerProvider.get();
+        final FactoryImpl factory = manager.find(FactoryImpl.class, id);
+        if (factory != null) {
```
sleshchenko, line 151: fix doc or add throwing of NotFoundException https://github.com/eclipse/che/pull/1926/files#diff-7c26382a725a0abec59646c16536c728R69

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/factory/Button.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Defines factory button.
+ *
+ * @author Anton Korneta
+ */
+public interface Button {
+
+    enum ButtonType {
+        logo, nologo
```
voievodin: 1. I think just `Type` is better, because it's already `Button.Type` and `Button.ButtonType` looks weird for me.
2. Why lowercase? It doesn't fit our codestyle, change it please.

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/DtoConverter.java
```diff
@@ -0,0 +1,131 @@
...
+            factoryDto.withPolicies(asDto(factory.getPolicies()));
+        }
+        if (factory.getButton() != null) {
+            factoryDto.withButton(asDto(factory.getButton()));
+        }
+        return factoryDto;
+    }
+
+    public static IdeDto asDto(Ide ide) {
+        IdeDto ideDto = newDto(IdeDto.class);
```
voievodin: from the variables definition context it looks like `ideDto` will be modified

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryLinksHelper.java
```diff
@@ -0,0 +1,158 @@
...
+import static org.eclipse.che.api.factory.shared.Constants.SNIPPET_REL_ATT;
+
+/**
+ * Helper class for creation links.
+ *
+ * @author Anton Korneta
+ */
+public class FactoryLinksHelper {
+
+    private static List<String> snippetTypes = ImmutableList.of("markdown", "url", "html", "iframe");
```
voievodin: add `final` and make it `CONSTANT_CASE` please

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryManager.java
```diff
@@ -0,0 +1,283 @@
...
+     *         type of snippet
+     * @param uriInfo
+     *         url context
+     * @return snippet content.
+     * @throws NotFoundException
+     *         when factory with specified id doesn't not found - with response code 400
+     * @throws ServerException
+     *         when any server error occurs during snippet creation
+     */
+    public String getFactorySnippet(String factoryId,
```
voievodin, line 249: Mark the method with `@Nullable` or factory doesn't follow _everything is not-null by default_ pattern ?

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryManager.java
```diff
@@ -0,0 +1,283 @@
...
+                                    String snippetType,
+                                    UriInfo uriInfo) throws NotFoundException,
+                                                            ServerException {
+        requireNonNull(factoryId);
+        final URI baseUri = uriInfo.getBaseUri();
+        final String baseUrl = UriBuilder.fromUri(baseUri)
+                                         .replacePath("")
+                                         .build()
+                                         .toString();
+        switch (snippetType) {
```
voievodin: From the method definition `snippetType` is nullable, so this code may produce unexpected npe, while you can avoid it by requiring non-null value or checking it before making switch.

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryManager.java
```diff
@@ -0,0 +1,283 @@
...
+                final Set<FactoryImage> images = getFactoryImages(factoryId);
+                final String imageId = (images.size() > 0) ? images.iterator().next().getName()
+                                                           : null;
+                try {
+                    return SnippetGenerator.generateMarkdownSnippet(baseUrl, getById(factoryId), imageId);
+                } catch (IllegalArgumentException e) {
+                    throw new ServerException(e.getLocalizedMessage());
+                }
+            default:
+                // when the specified type is not supported
```
voievodin, line 280: it should be defined by method javadoc

PR #1931, opened 2016-07-25
Update IDE resource path after changes for namespace handling (Fix #1909)

### LINKED ISSUE #1909: IDE resources caching from UD is broken
Earlier UD cached IDE resources to speedup loading of IDE. Now this technick is broken. URL of compilation mapping file is not reachable due to changes in URL scheme. Here is the URL used by UD: http://localhost:8080/ide/_app/compilation-mappings.txt?uid=553185

**Reproduction Steps:**
1. Load UD
2. Check error in console of browser dev tools

**Expected behavior:**

UD should be able to get compilation mapping file, get resources URL and fetch it into the cache. 

**Observed behavior:**

Request to retrieve compilation mappings file returns main IDE page.
There is an error in console of the browser: **Unable to find the IDE javascript file to cache**

**Che version:**    4.6.0-SNAPSHOT
**Additional information:**
- Problem can be reliably reproduced, doesn't happen randomly: [Yes]

benoitf:
### What does this PR do?

Update the IDE resource path as it has been changed for namespace handling

it fixes #1909 
### Previous Behavior

IDE JS is not cached
### New Behavior

IDE JS is now cached
### Tests written?

No
### Docs requirements?

No

### DIFF
--- a/dashboard/src/assets/branding/product.json
+++ b/dashboard/src/assets/branding/product.json
@@ -4,7 +4,7 @@
   "logoFile": "che-logo.svg",
   "logoTextFile": "che-logo-text.svg",
   "favicon": "favicon.ico",
-  "ideResources": "/ide/_app/",
+  "ideResources": "/_app/",
   "helpPath": "https://www.eclipse.org/che/",
   "helpTitle": "Community",
   "supportEmail": "wish@codenvy.com",


garagatyi:
LGTM

Issue #1899, opened 2016-07-21
Copy and paste not working in the terminal


ghost:
**Reproduction Steps:**
1. Start the latest Che from master
2. Run a command in a terminal
3. Ctrl + C and then Ctrl + V the output.

**Expected behavior:**

Text is copied and pasted

**Observed behavior:**

An error occurs (see screenshot)

**Che version:**    [4.6.0-SNAPSHOT]   

**Additional information:**
- Problem started happening recently, didn't happen in an older version of Che: [Yes]
- Problem can be reliably reproduced, doesn't happen randomly: [Yes]
  ![terminal](https://cloud.githubusercontent.com/assets/5337267/17033580/834a2666-4f87-11e6-9630-6c3069b946e4.png)

thpglobal:
Still not working today - can you guys roll-back the version on beta.codenvy.com until it's fixed???

TylerJewell:
We have a roll forward policy due to data migrations. It's labeled as a blocker and hot fix investigations are underway.

rajat1saxena:
It seems like this issue is still there in codenvy.io IDE.

cifren:
It seems copy is not working anymore, though paste is working.

PR #1932, opened 2016-07-25
#1899: Add ability to copy and paste text in terminal view

vzhukovs:
This PR adds ability copy and paste selected text in and from the terminal view.

Related issue: https://github.com/eclipse/che/issues/1899

@vparfonov review it, plz.


vparfonov:
ok

Issue #1935, opened 2016-07-25
Add custom user agent header into JGit 


mshaposhnik:
[Short description of problem here]
Some of  git providers have filtration by User-Agent header and disallow agents other than native git client have. Need to set custom header instead of "JGit-4.3.0" current. See codenvy/codenvy#430

**Reproduction Steps:**
1. Try import project from https://sourcegraph.com/sourcegraph/sourcegraph

**Expected behavior:**

Project cloned

**Observed behavior:**

Cloning failed

**Additional information:**

mshaposhnik:
Done.

PR #1936, opened 2016-07-25
CHE-1935; add custom User-Agent header value for JGit;

mshaposhnik:
### What does this PR do?

 Sets custom header instead of "JGit-4.3.0" for git providers have filtration by User-Agent header.
#1935  codenvy/codenvy#430
### Tests written?

No
### Docs requirements?

None

### DIFF
--- a/wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
+++ b/wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
@@ -135,7 +135,6 @@ import org.eclipse.jgit.transport.SshTransport;
 import org.eclipse.jgit.transport.TrackingRefUpdate;
 import org.eclipse.jgit.transport.Transport;
 import org.eclipse.jgit.transport.URIish;
-import org.eclipse.jgit.transport.UserAgent;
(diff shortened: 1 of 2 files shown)


wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -1605,7 +1608,7 @@ public void configure(Transport transport) {
             }
 
             ProxyAuthenticator.initAuthenticator(remoteUrl);
-
+            UserAgent.set(USER_AGENT);
```
benoitf: do we need to call this static method from JGitConnection instance or from a class that is less created like JGitConnectionFactory ?

mshaposhnik: Yes, make sense. Maybe i should move it to connection constructor.

benoitf:
Not in factory ? So we share for all connections

skabashnyuk:
> Not in factory ? Son we share for all connections

@benoitf  Yes. What's wrong with this? How do you propose?

benoitf:
@skabashnyuk with latest change it's ok for me (it's in the factory so we call only once)

Issue #1797, opened 2016-07-18
SnapshotDao implementation on JPA


skabashnyuk:
SnapshotDao implementation on JPA

PR #1937, opened 2016-07-25
CHE-1797: Add JPA based SnapshotDao implementation

voievodin:
### What does this PR do?

Adds a new JPA based implementation of `SnapshotDao`
Removes `namespace` from the `Snapshot` model, as it's stored/retrieved but not used for business logic.
Simplifies implementation of LocalTckRepositories. 
Adds TCKs for SnapshotDao and covers local implementation by them.
#1797
### Tests written?

Yes

@skabashnyuk, @sleshchenko, @akorneta please review


wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/spi/SnapshotDao.java
```diff
@@ -65,17 +65,15 @@
     void saveSnapshot(SnapshotImpl snapshot) throws SnapshotException;
 
     /**
-     * Find snapshots by namespaces, workspace, project
+     * Find snapshots by workspace.
      *
-     * @param namespace
-     *         snapshot namespace(e.g. owner).
      * @param workspaceId
      *         workspace specified in desired snapshot, optional
```
sleshchenko: Optional? https://github.com/eclipse/che/pull/1937/files#diff-ae483626205558e96141b48c9fb29ce8R84

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/SnapshotImpl.java
```diff
@@ -13,40 +13,91 @@
...
+@NamedQueries(
+        {
+                @NamedQuery(name = "Snapshot.getByMachine",
+                            query = "SELECT snapshot " +
+                                    "FROM Snapshot snapshot " +
+                                    "WHERE snapshot.workspaceId = :workspaceId" +
+                                    "  AND snapshot.envName     = :envName" +
+                                    "  AND snapshot.machineName = :machineName"),
+                @NamedQuery(name = "Snapshot.findSnapshots",
+                            query = "SELECT snapshot " +
```
skabashnyuk, line 45: Make sure all necessary indexes for this query exists.

skabashnyuk:
Other ok

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/SnapshotImpl.java
voievodin, line 45: It's [okay](https://devcenter.heroku.com/articles/postgresql-indexes#multi-column-indexes)

sleshchenko:
LGTM

Issue #1814, opened 2016-07-18
Wrong behavior after configure folder with name 'java' as source 


AndreyChizhikov:
**Reproduction Steps:**
1. Create plain java project and add several folders ( /test/java/com/company).
2.  Use the 'java' folder as source folder by context menu.

**Observed behavior:**
Folder 'java' didn't mark as source folder. Need to invoke F5. The same behavior if the 'test' folder unmark as source.

**Che version:**  
4.5.0-RC1-SNAPSHOT

PR #1954, opened 2016-07-26
#1814: Synchronize project instead of parent directory

vzhukovs:
Project synchronization is more reasonable, because we also retrieve project properties and with actual values we can redraw project correctly in the UI.

Additional checks were added to avoid throwing illegal state exceptions.

Related issue: https://github.com/eclipse/che/issues/1814

@vparfonov review it, plz.


vparfonov:
ok

PR #1958, opened 2016-07-27
[WiP] Introduce maven bom plugin that generate pom files

benoitf:
### What does this PR do?

It provides a bom maven plugin and a maven module using this plugin to generate a bom file (Bills Of Materials)
So clients can develop components by importing this pom instead of having to use inheritance
### Tests written?

Yes

### DIFF
--- a/plugins/che-bom-maven-plugin/pom.xml
+++ b/plugins/che-bom-maven-plugin/pom.xml
@@ -1,0 +1,128 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) 2012-2016 Codenvy, S.A.
+    All rights reserved. This program and the accompanying materials
+    are made available under the terms of the Eclipse Public License v1.0
+    which accompanies this distribution, and is available at
+    http://www.eclipse.org/legal/epl-v10.html
+
+    Contributors:
+      Codenvy, S.A. - initial API and implementation
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>che-plugin-parent</artifactId>
+        <groupId>org.eclipse.che.plugin</groupId>
+        <version>4.6.0-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+    <artifactId>che-bom-maven-plugin</artifactId>
+    <version>4.6.0-SNAPSHOT</version>
+    <packaging>maven-plugin</packaging>
+    <name>Che Core :: Bom Maven Plugin</name>
+    <properties>
+        <findbugs.failonerror>false</findbugs.failonerror>
+    </properties>
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.maven</groupId>
+            <artifactId>maven-aether-provider</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.maven</groupId>
+            <artifactId>maven-artifact</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.maven</groupId>
+            <artifactId>maven-core</artifactId>
+            <exclusions>
+                <exclusion>
+                    <artifactId>org.eclipse.sisu.plexus</artifactId>
+                    <groupId>org.eclipse.sisu</groupId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.maven</groupId>
+            <artifactId>maven-model</artifactId>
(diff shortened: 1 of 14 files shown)


skabashnyuk:
How we are going to use this functionality?

benoitf:
people developing plugins may use in their pom

```

<dependencyManagement>
    <dependencies>
        <dependency>
            <groupId>org.eclipse.che</groupId>
            <artifactId>bom</artifactId>
            <version>4.6.0</version>
            <scope>import</scope>
            <type>pom</type>
        </dependency>
    </dependencies>
</dependencyManagement>


```

then they could get the same versions than che but keeping their own pom inheritance

skabashnyuk:
Why they can't include 

``` xml
            <dependency>
                <groupId>org.eclipse.che</groupId>
                <artifactId>che-parent</artifactId>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
```

What is the profit of your approach?

plugins/che-bom-maven-plugin/src/test/projects/project/mymodule/pom.xml
```diff
@@ -0,0 +1,45 @@
...
+                    </exclusion>
+                </exclusions>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <modules>
+        <module>mysubmodule</module>
+    </modules>
+</project>
```
skabashnyuk, line 45: new line

plugins/che-bom-maven-plugin/src/test/projects/project/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+    <packaging>pom</packaging>
+    <name>Test of Eclipse Che Bom plugin with some dependency management</name>
+
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.apache.maven</groupId>
+                <artifactId>maven</artifactId>
+                <version>3.3.3</version>
```
skabashnyuk, line 30: shall we move version to parent pom as we did for regular dependencies?

plugins/che-bom-maven-plugin/src/test/projects/project/mymodule/mysubmodule/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+                <exclusions>
+                    <exclusion>
+                        <groupId>org.eclipse.che.test.gId2</groupId>
+                        <artifactId>org.eclipse.che.test.aId2</artifactId>
+                    </exclusion>
+                </exclusions>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+</project>
```
skabashnyuk, line 41: new line

plugins/che-bom-maven-plugin/src/test/projects/project/pom.xml
benoitf, line 30: it's a test pom it's only used as test entry, it's not launched nor installed by maven

plugins/che-bom-maven-plugin/pom.xml
```diff
@@ -0,0 +1,128 @@
...
+            <artifactId>maven-plugin-annotations</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.sisu</groupId>
+            <artifactId>org.eclipse.sisu.plexus</artifactId>
+            <scope>provided</scope>
+            <exclusions>
+                <exclusion>
+                    <artifactId>jsr250-api</artifactId>
```
skabashnyuk, line 72: shall we exclude this dependency  in parent pom?

plugins/che-bom-maven-plugin/src/test/projects/project/pom.xml
benoitf, line 30: and we need to use <dependencyManagement> as the test validate that when ppl use dependencyManagement, it is working as expected.

skabashnyuk, line 30: ok

TylerJewell:
@benoitf @skabashnyuk - this PR has been open a long time. Are we still going to move forward with this approach?

I was having a very good conversation with @vparfonov and @skabashnyuk about the development workflow. They are both of the mindset that we need to start developing a set of "assembly" tools:
1. Generate a new custom assembly from version of Che.
2. Add new (blank, but compiles) custom ide / ws / master extension to the custom assembly
3. Add new (blank) default stack, template to custom assembly
4. Add new (blank) agent, such as a language server protocol agent

The custom assembly generation would be one-way menaing that you could generate custom assemblies, but once they are generated the utility would not be usable again.  It would:
a) Generate a new git repository
b) Likely use a maven archetype to do the work
c) It would be the proper structure and always compile after generation
d) It would have a set of README files to tell the developer which files they should edit to custom certain branding items, to add / remove new file types for syntax highlighting, to alter existing existings, where your new extensions are at (which files to modify), which files to modify to customize the custom CLI generated for you.   
e) It would have a README file to tell the developer how to build the assembly, how to use it with their custom CLI. This README would also have instructions on how to set up incremental compilation and possible super dev mode.

There may be other componets that I am missing.

Issue #1969, opened 2016-07-28
Add Native Scripts to Launch All Che Commands


TylerJewell:
Provide windows and linux batch + shell scripts that can simplify the invocation of the `che-launcher` and the `che-file` docker containers.
- [x] Add CI systems to push nightly, latest, and tagged version of che.sh and che.bat files.
- [x] Move files into root of the che repository.
- [x] Add docs pages on how to make use of these simpler scripts.
- [x] If boot2docker and directory is not %userprofile%\AppData, then ignore CHE_CONF parameters.
- [x] If docker for mac, then we should automatically call the loopback configuration command.

PR #1970, opened 2016-07-28
WIP: Add Che CLI with native scripts for Windows / Linux

TylerJewell:
### What does this PR do?

The implementation of #1969 - platform native scripts.
1. Adds Windows and Linux scripts.
2. Instructions for how CI systems modified to post these scripts onto hosted servers.
3. Simple single word syntax for launching all commands for che-launcher and che-file.

Usage:
che help
che start
che stop
che restart
che update
che init
che up

Users will install the files (windows requires git bash installed):

```
curl -sL http://eclipse.org/che/che.sh | sh
curl -sL http://eclipse.org/che/che.bat > che.bat
```

### DIFF
--- a/dockerfiles/che-file/Dockerfile
+++ b/dockerfiles/che-file/Dockerfile
@@ -1,0 +1,39 @@
+# Copyright (c) 2016 Codenvy, S.A.
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License v1.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-v10.html
+#
+# Prerequisite : typescript library compiled
+# see ../lib/typescript/README.md folder
+#
+# To build, see README.md
+#
+# To use:
+#  `docker run -v /var/run/docker.sock:/var/run/docker.sock \
+#              -v "$PWD":"$PWD" --rm codenvy/che-file \
+#              /bin/che $PWD <init|up>`
+# 
+#  where [COMMAND]:
+#    init -- Initialize configuration files for this directory
+#    up   -- Launch workspace, starting Che server, if necessary 
+#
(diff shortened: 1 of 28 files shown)


l0rd:
I like it. 

I have a few comments:
- Use https. http is not acceptable. With a MITM attack we could easily take control of the machine where the script will run. 
- For windows why don't you use a [`curl` equivalent for powershell](http://superuser.com/questions/344927/powershell-equivalent-of-curl)? Or, since you ask users to have `git bash`, you could have a unique `.sh` script for both Linux and Windows.
- When I visit `http://eclipse.org/che/che.sh` with Chrome the script is downloaded automatically. If I visit `https://get.docker.com` the script source code is rendered by the browser and not downloaded. I prefer the latter as a default behaviour. Users can have easily a look at the script before actually run it.

TylerJewell:
Yes - definitely on the `https`.

On the curl equivalent for powershell. The issue with powershell is that it requires the user to start a powershell session. Where as, if you have a standard windows command prompt open, then the Git Bash will provide a standard curl command.

On the syntax for preventing a file from being downloaded directly - how can we do that? We have limited controls for how such files are hosted with eclipse.org.

l0rd:
When requesting the script in the header of the http response Content-Type should be set to `text/plain` but is `application/x-shellscript`:
image

benoitf:
I tried to add a custom .htaccess file but it seems the override is not allowed on Eclipse website Server

TylerJewell:
I will start a discussion with the eclipse foundation on any options we may have.

benoitf:
maybe they could add to our che directory on Apache2 configuration the following content
http://git.eclipse.org/c/www.eclipse.org/che.git/plain/.htaccess

TylerJewell:
Added changes

Issue #1785, opened 2016-07-18
Preview URL may be lost after refreshing page


vparfonov:
[Short description of problem here]

**Reproduction Steps:**
1. Start some web app with preview URL
2. Refresh page
3. Preview URL don't display anymore 

**Expected behavior:**
need to keep preview URL after refreshing 

**Che version:**   4.5.0

riuvshin:
this issue is still actual

PR #1978, opened 2016-07-28
CHE-1785: restore commands when WS is started

svor:
### What does this PR do?

Restores all commands when ws-agent is started

https://github.com/eclipse/che/issues/1785
### Tests written?

Yes

Please review @vparfonov


vparfonov:
ok

Issue #1976, opened 2016-07-28
Not able to run Che as a Docker Container on OSX


skabashnyuk:
[Short description of problem here]

**Reproduction Steps:**
1. Set up Docker for OSX beta 1.12.0-rc4,
2. Set up environment as described here https://eclipse-che.readme.io/docs/usage-docker-for-mac
3. Run che as described here https://eclipse-che.readme.io/docs/usage-docker-for-mac

**Expected behavior:**
Che is running

**Observed behavior:**

``` bash
11:13:31]sj:~#: docker run --name=che \
>     -v /home/user/che/lib:/home/user/che/lib-copy \
>     -v /home/user/che/workspaces:/home/user/che/workspaces \
>     -v /home/user/che/storage:/home/user/che/storage \
>     -v /var/run/docker.sock:/var/run/docker.sock \
>     -p 8080:8080 \
>     codenvy/che --remote:$DOCKER_VM_IP
Unable to find image 'codenvy/che:latest' locally
latest: Pulling from codenvy/che
e110a4a17941: Already exists
a3ed95caeb02: Pull complete
0e30e6967f5d: Pull complete
bfa497568536: Pull complete
Digest: sha256:b300245d808f8e788d90658679159f0f1fccd90c9e50f8a39a9a9e6db1b9fb56
Status: Downloaded newer image for codenvy/che:latest
docker: Error response from daemon: Mounts denied:
The paths /home/user/che/lib and /home/user/che/workspaces and /home/user/che/storage
are not shared from OS X and do not belong to the system.
You can configure shared paths from Docker -> Preferences...
```

**Che version:**    che:latest
**OS and version:**  OSX
**Docker version:** 1.12.0-rc4,
**Che install:**    Docker container

ghost:
@l0rd can you comment please?

l0rd:
@skabashnyuk @eivantsov Running it with the new launcher should work. Is it an option here?

l0rd:
Please use codenvy/che-launcher:nightly

l0rd:
If you need latest:

> You can configure shared paths from Docker -> Preferences...

This behaviour has been introduced in the latest release candidate of Docker for Mac beta. The client verify that you are mounting in the container a folder that is actually mounted in the VM.

skabashnyuk:
Looks like it's working. But huge wait with not output on start looks strange

TylerJewell:
You need to add -t to the run command.  We updated the docs yesterday with this requirement.

PR #1983, opened 2016-07-28
Fix get_che_launcher_version function when version is not specified

l0rd:
### What does this PR do?

The launcher try to infer che version from its own version. When a user run the launcher without specifying a version the launcher return a non valid version. 

This issue has been spotted on #1976 
### Previous Behavior

If launcher was used without an explicit version (e.g. `docker run -v /var/run/docker.sock:/var/run/docker.sock codenvy/che-launcher info`) function get_che_launcher_version returned `codenvy/che-launcher`.
### New Behavior

If launcher is used without and explicit version function get_che_launcher_version returns `latest`. That's the default behavior of the Docker client.
### Tests written?

Yes
### Docs requirements?

No


TylerJewell:
Ok

Issue #1989, opened 2016-07-29
Implement Account API


sleshchenko:
Epic link https://github.com/codenvy/codenvy/issues/409

Implement Account API

As result of this issue there should be implemented API that will allow users to:
- create, remove workspace which belong to user (personal account)
- fetch workspaces which belong to some personal account by account name

TylerJewell:
@sleshchenko - the epic link provided doesn't seem to be the right one.  Can you update please.

sleshchenko:
@TylerJewell thanks for remark, fixed.

bmicklea:
@sleshchenko is this closed?  The epic link you've referenced has been closed already.

PR #1990, opened 2016-07-29
CHE-1989 Implement Account API

sleshchenko:
### What does this PR do?
- Added new entity Account that represents abstract resources owner and allows to extend it by User or another entities later
- Adapted Workspace API to it. Removed account id parameters because namespace parameter will contain account name. Added method for getting workspaces by namespace (account name).

https://github.com/eclipse/che/issues/1989
### Previous Behavior

Earlier workspaces were linked to user but without explicit declaring
### New Behavior

Workspaces are linked to account by namespace field. User extends Account and it is personal account.
### Tests written?

Yes


skabashnyuk:
ok. @evoevodin @akorneta ?

akorneta:
ok

wsmaster/che-core-api-account/src/main/java/org/eclipse/che/account/spi/AccountImpl.java
```diff
@@ -0,0 +1,98 @@
...
+import javax.persistence.UniqueConstraint;
+import java.util.Objects;
+
+/**
+ * Data object for {@link Account}.
+ *
+ * @author Sergii Leschenko
+ */
+@Entity(name = "Account")
+@Table(uniqueConstraints = @UniqueConstraint(columnNames = {"name"}))
```
skabashnyuk: duplicate with field annotation

wsmaster/che-core-api-account/src/main/java/org/eclipse/che/account/spi/AccountImpl.java
```diff
@@ -0,0 +1,97 @@
...
+                                    "FROM Account a " +
+                                    "WHERE a.name = :name")
+        }
+)
+public abstract class AccountImpl implements Account {
+    @Id
+    protected String id;
+
+    @Column(unique = true, nullable = false)
+    protected String name;
```
voievodin, line 47: It seems like you need an index for the `name` field

wsmaster/che-core-api-account/src/main/java/org/eclipse/che/account/spi/jpa/JpaAccountDao.java
```diff
@@ -0,0 +1,74 @@
...
+
+import static java.lang.String.format;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * JPA based implementation of {@link AccountDao}.
+ *
+ * @author Sergii Leschenko
+ */
+public class JpaAccountDao implements AccountDao {
```
voievodin, line 36: `@Singleton` ?

voievodin:
Ok

wsmaster/che-core-api-account/src/main/java/org/eclipse/che/account/spi/AccountDao.java
```diff
@@ -0,0 +1,51 @@
...
+     *         account identifier
+     * @return account instance with given id
+     * @throws NullPointerException
+     *         when {@code id} is null
+     * @throws NotFoundException
+     *         when account with given {@code id} was not found
+     * @throws ServerException
+     *         when any other error occurs during account fetching
+     */
+    AccountImpl getById(String id) throws NotFoundException, ServerException;
```
dmytro-ndp, line 35: It should return Account instead of AccountImpl, shouldn't it?

wsmaster/che-core-api-account/src/main/java/org/eclipse/che/account/spi/AccountDao.java
```diff
@@ -0,0 +1,51 @@
...
+     *         account name
+     * @return account instance with given name
+     * @throws NullPointerException
+     *         when {@code name} is null
+     * @throws NotFoundException
+     *         when account with given {@code name} was not found
+     * @throws ServerException
+     *         when any other error occurs during account fetching
+     */
+    AccountImpl getByName(String name) throws ServerException, NotFoundException;
```
dmytro-ndp, line 50: It should return Account instead of AccountImpl, shouldn't it?

wsmaster/che-core-api-account/src/main/java/org/eclipse/che/account/spi/AccountDao.java
sleshchenko, line 35: No. Why should it return Account?

dmytro-ndp, line 35: 1) to program to interface, not to implementation
2) because this is AccountDAO, not AccountImplDAO

dmytro-ndp, line 35: Also, there are several setters in the AccountImpl, which are absent in Account and seems therefore are useless and should be closed.

dmytro-ndp, line 35: If we don't change API, IMHO it's more clear to have the next class tree:

```
Account -> AbstractAccount -> UserAccountImpl
                           -> ...
```

An example of such hierarchy: http://docs.oracle.com/javase/6/docs/api/javax/swing/Action.html

```
Action -> AbstractAction -> BasicDesktopPaneUI.OpenAction
                         -> BasicDesktopPaneUI.CloseAction

```

Moreover: there is a lot of abstract classes with _Abstract_ prefixes in name which implements certain interface: http://stackoverflow.com/a/6861797/2556329

Issue #1919, opened 2016-07-22
When restarting a workspace, open project tree re-appears as closed.


musienko-maxim:
**Reproduction Steps:**
1. Go to the IDE and open some file and directory. 
2. Workspace ->Stop. Start again.

**Expected behavior:** Opened items should be resored.
**Observed behavior:** The project is collapsed

**Expected behavior:** after clicking on the fires item the Compare button should be enabked
**Observed behavior:** after clicking on the fires item the Compare button still disabled
**Che version:**    4.6.0
**OS and version:** Ubuntu OS 14.0
**Docker version:** 1.11.2
**Che install:**  local bundle
**Additional information:**
- Problem started happening recently, didn't happen in an older version of Che: YES

**Related test**: RestoreStateOfProjectTest

riuvshin:
it seems this is same issue as https://github.com/eclipse/che/issues/1902

JamesDrummond:
Please provide a status.

bmicklea:
@vzhukovskii : Was this merged into the 4.6.0 milestone?  The milestone isn't set on the issue but based on the merge time I suspect it's in 4.6...

ashumilova:
Added milestone, was missed

PR #1994, opened 2016-07-29
#1919: Save application state to the user profile correctly

vzhukovs:
This PR fix correct saving the project state. The problem was in AppStateManager, there were two handlers which handled workspace and ws agent stopped event, they override each other and that's why we had wrong state in user preference.

Related issue: https://github.com/eclipse/che/issues/1919

@vparfonov review it, plz.


vparfonov:
ok

Issue #1933, opened 2016-07-25
Wrong behavior after starting workspace from snapshot 


AndreyChizhikov:
**Reproduction Steps:**
1. Create snapshot
2. Stop workspace
3. Start workspace from snapshot

**Observed behavior:**

Workspace starts well but next error appears in browser console:
`Request URL:http://localhost:8080/wsmaster/api/recipe/script/machinedi5xowscpv2tfvgn
Request Method:GET
Status Code:500 Internal Server Error
java.net.URISyntaxException: Illegal character in scheme name at index 7: machine_snapshot_t1eg3vcfcyf0zrc4:latest`

**Che version:**   4.6.0-SNAPSHOT
**OS and version:** Ubuntu 14.04 LTS
**Docker version:** 1.11.2
**Che install:**    local

TylerJewell:
Removed severity blocker after issue update.  It was not clear that the ws starts normally even with the error.

PR #2011, opened 2016-08-01
CHE-1933 Do not show recipe content if type is not Dockerfile

mkuznyetsov:
### What does this PR do?

Fixes an error during restoring workspace from snapshot.

https://github.com/eclipse/che/issues/1933
### Previous Behavior

Tried to fetch recipe which would fail if location is not a valid URL (which is valid, if only type is 'dockerfile' ).
### New Behavior

Do not fetch recipe if type is not 'dockerfile'.
### Tests written?

Yes
### Docs requirements?

No

### DIFF
--- a/plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/perspective/widgets/machine/appliance/recipe/RecipeTabPresenter.java
+++ b/plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/perspective/widgets/machine/appliance/recipe/RecipeTabPresenter.java
@@ -43,23 +43,40 @@ public class RecipeTabPresenter implements TabPresenter {
 
     /**
      * Calls special method on view which updates recipe of current machine.
+     * If case of 'image', it is it's location, in case of 'dockerfile'
+     * - it is content, that is fecthed by location URL
      *
      * @param machine
      *         machine for which need update information
      */
     public void updateInfo(@NotNull final Machine machine) {
-        recipeScriptClient.getRecipeScript(machine).then(new Operation<String>() {
+        if (machine.getRecipeType() == null) {
+            Log.error(RecipeTabPresenter.class, "Recipe type is null for machine '" + machine.getId() + "'");
+            view.setScript("Recipe type is null for machine '" + machine.getId() + "'");
+            return;
+        }
+        switch (machine.getRecipeType()) {
+            case "image":
+                view.setScript("Image location: " + machine.getRecipeLocation());
+                break;
+            case "dockerfile":
+                recipeScriptClient.getRecipeScript(machine).then(new Operation<String>() {
                     @Override
                     public void apply(String recipe) throws OperationException {
                         view.setScript(recipe);
                     }
                 }).catchError(new Operation<PromiseError>() {
-            @Override
-            public void apply(PromiseError error) throws OperationException {
-                Log.error(RecipeTabPresenter.class,
-                          "Failed to get recipe script for machine " + machine.getId() + ": " + error.getMessage());
-            }
-        });
+                    @Override
+                    public void apply(PromiseError error) throws OperationException {
+                        Log.error(RecipeTabPresenter.class,
+                                  "Failed to get recipe script for machine " + machine.getId() + ": " + error.getMessage());
+                        view.setScript("Failed to get recipe script for machine '" + machine.getId() + "'");
+                    }
+                });
+                break;
+            case "default":
+                view.setScript("Recipe type: " + machine.getRecipeType());
+        }
     }
 
     /** {@inheritDoc} */
--- a/plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/machine/Machine.java
+++ b/plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/machine/Machine.java
@@ -75,16 +75,22 @@ public class Machine {
         return descriptor.getConfig().getType();
     }
 
-    /** @return script of machine recipe */
+    /** @return location of machine recipe */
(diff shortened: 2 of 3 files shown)


plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/perspective/widgets/machine/appliance/recipe/RecipeTabPresenter.java
```diff
@@ -43,23 +43,28 @@ public RecipeTabPresenter(RecipeView view, RecipeScriptDownloadServiceClient rec
...
+                }
+            }).catchError(new Operation<PromiseError>() {
+                @Override
+                public void apply(PromiseError error) throws OperationException {
+                    Log.error(RecipeTabPresenter.class,
+                              "Failed to get recipe script for machine " + machine.getId() + ": " + error.getMessage());
+                }
+            });
+        }
+        view.setScript("");
```
mshaposhnik: So we just show empty panel? Maybe write something like "unsupported script type" ?

skabashnyuk: @mkuznyetsov  can   you attach screenshot of this case?

mkuznyetsov: selection_013

As before, it would show an empty 'Recipe' tab

mkuznyetsov: @TylerJewell is it ok to show nothing here if we are unable to get recipe content? (e.g. recipe type is 'image').

tolusha: If I were a user I would be confused why there is nothing on the panel.

benoitf: we couldn't display the image name in that case ?

TylerJewell: @mkuznyetsov if the recipe type is "image", we should still show the name of the image that we pulled.  The "image", in this case, is still the definition of the recipe.  You could say something like, "no user-provided recipe, pulled image "codenvy/ubuntu_jdk8".

mkuznyetsov: Would simple ''Image location: xxx" would do it?
selection_014

mkuznyetsov: also, display ""Failed to get recipe script for machine xxx" if error happens on fetching dockerfile recipe script.
and in case recipe type is not 'image' or 'dockerfile', display "Recipe type: xxx"

plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/perspective/widgets/machine/appliance/recipe/RecipeTabPresenter.java
```diff
@@ -43,23 +43,31 @@ public RecipeTabPresenter(RecipeView view, RecipeScriptDownloadServiceClient rec
...
-            @Override
-            public void apply(PromiseError error) throws OperationException {
-                Log.error(RecipeTabPresenter.class,
-                          "Failed to get recipe script for machine " + machine.getId() + ": " + error.getMessage());
-            }
-        });
+        if (machine.getRecipeType().equals("image")) {
+            view.setScript("Image location: " + machine.getRecipeLocation());
+        }
+        if (machine.getRecipeType().equals("dockerfile")) {
```
benoitf: use "else if" instead of if
also "image".equals(machine.getRecipeType()) is usually a better check for avoiding any NPE

mkuznyetsov: Updated with switch statement, and handled null in recipe type

plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/perspective/widgets/machine/appliance/recipe/RecipeTabPresenter.java
mkuznyetsov: @TylerJewell WDYT^

TylerJewell: ok

skabashnyuk:
ok

mshaposhnik:
ok

PR #2018, opened 2016-08-02
Add che-mount support for desktop IDEs

TylerJewell:
### What does this PR do?

Introduces a new Docker container which can mount Che workspaces to a local host directory and use unison to keep the local directory in sync with the Che workspace.  It also introduces improvements to the `che.sh` CLI to simplify using the container.

Container usage:

```
docker run --rm -it --cap-add SYS_ADMIN --device /dev/fuse 
            -v /var/run/docker.sock:/var/run/docker.sock
            -v <host-dir>:/mnthost codenvy/che-mount <ip> <ws-ssh-port>
```

where `<ip` is the IP address of the Che server and `<ws-ssh-port>` is the ephemeral port of the SSH daemon running within a workspace, retrieved by the user when he opens the SSH entry.

CLI usage:

```
che mount <local-path> <ws-ssh-port>
```

Note that if the Che mount is run using boot2docker and the <local-path> is not %userprofile% then the mount will fail.  The intention of this pull request is to allow the image to get staged on Docker Hub for experimentation and improvement.


dockerfiles/che-mount/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+    cd src && \
+    make && \
+    cp unison /usr/local/sbin/ && \
+    cp unison /usr/bin/ && \
+    cd /opt && \
+    rm -rf src/
+
+COPY sync.sh /bin
+
+ENTRYPOINT ["/bin/sync.sh"]
```
benoitf: missing newline

TylerJewell: fixed

benoitf: ok thx, do you have tried with an alpine based image so we could decrease the size of the docker image ?

TylerJewell: I tried - but couldn't get it to work.  But that is likely my own issues.  I couldn't figure out how to build unison and install all of the necessary packages.  Do you think it's possible to do?

Issue #2022, opened 2016-08-02
Incorrect factory validation behavior when field value matches default for its type


mshaposhnik:
Factory validator fails when field value matches default for its type, e.g. if set `boolean` variable to `false`, ot `intereger` to `0`e.t.c

Issue related to https://github.com/codenvy/codenvy/issues/455

**Reproduction Steps:**
Try create factory with several machines described, when one of them has `"dev" : false` property.

**Expected behavior:**
Factory must be created

**Observed behavior:**
Factory validator throws error saying "dev" property is not set

**Che version:**    4.6.0

**Additional information:**
- Problem started happening recently, didn't happen in an older version of Che: No
- Problem can be reliably reproduced, doesn't happen randomly: Yes

TylerJewell:
@mshaposhnik - should this belong in a different repository - /codenvy/codenvy

mshaposhnik:
@TylerJewell No - factory API is in Che.

bmicklea:
@mshaposhnik was is merged into the 4.7.0 milestone?  If so please add that milestone designation to this issue.

PR #2023, opened 2016-08-02
CHE-2022; fixed validation behavior when field value matches default value for its type

mshaposhnik:
### What does this PR do?

 Fixes validation behavior when field value matches default for its type.
This PR limits checking to just collections and strings, because for other types we have no way to distinguish whether the value wasn't set at all or it was set to value which is matches default.
#2022
### Previous Behavior

Validator failed on fields like `"integer":0` or `"boolean": false`
### New Behavior

No fails on default values
### Tests written?

Yes
### Docs requirements?

None


skabashnyuk:
ок

PR #2024, opened 2016-08-02
CHE-408: Add ramUsage() method to DockerConnector

vinokurig:
### What does this PR do?

Add ability to get 'used ram' and 'total ram' values from docker

https://github.com/codenvy/codenvy/pull/501


vinokurig:
Will be implemented in Codenvy repo

PR #2027, opened 2016-08-02
Move che-mount to alpine-based image

TylerJewell:
### What does this PR do?

Converts che-mount Docker image to use alpine based image, dropping size from 750MB to 14MB.


TylerJewell:
@benoitf  - please check and +1 so I can merge.  This is tested and now works fully on alpine-based image.

benoitf:
OK for me 
That's great for the size of the container

Issue #1803, opened 2016-07-18
Investigate strategies for removing cascading entities


skabashnyuk:
_The problem:_
User removed, but related entities like _Workspaces_, _Factories_ or _Permissions_ are not removed which leaves the system in inconsistent state and causes different problems(e.g. if workspace is not removed than _namespace_ is still aquired by non-existing user, while it is impossible to use that namespace).

_Requirements:_
1. Entity related data must be removed along with the entity
2. If it is impossible to remove some of entity related data, then entity mustn't be removed either and removed data must be restored. In this way we can create application related constraints which will enforce user to perform some actions before removing the entity.
3. Sometimes it is necessary to know that entity is removed, the mechanism of notifying clients based on the _EntityRemoved_ events which usually published after entity is removed.
So the strategy of cascade removing should be rather on application-level than on the DBMS level which allows to control _EntityRemoved_ events.

_What should be done:_
As a result we should investigate the best way for performing cascade remove which fits our application architecture with all the pros and cons of it. 

_Consider using:_
- EventBus
- Jpa Events

TylerJewell:
@skabashnyuk - I think we can close this issue, correct?  This was an investigation issue and we have already completed the implementation and merge.

skabashnyuk:
issues is not closed because code related to it is not  merged in master. We will do that when epic https://github.com/eclipse/che/issues/1790 will be closed

PR #2031, opened 2016-08-03
Add cascade removal for known entities

voievodin:
### What does this PR do?

Introduces cascade removal strategies, which are:
- Use `orphanRemoval` or `@CascadeOnDelete` those strategies require entries to declare
  either bidirectional mapping, or mapping present on owning side, for example `Profile` depends on `User`while `User` doesn't know anything about profile, so for removing profile along with a user we have to declare Profile mapping on User object side, and mark the dependency with `orphanRemoval`, this is not always possible as objects may be in separate modules and it will produce circular dependencies.
- Use `foreignKeyDefinition` for defining foreign keys e.g.:

```
@JoinColumn(foreignKey = 
            @ForeignKey(name = "USER_FK",
                        foreignKeyDefinition = "FOREIGN KEY(user_id) REFERENCES User(id) ON DELETE CASCADE"))
```

While this strategy doesn't require to define dependenies on owning side, it is RDBMS specific, 
another problem is that it won't affect Caches(L2 is a problem here) as all the removal stuff will be performed by the database.
- Use JPA events + `EventService` for removing entities. 
  As `EventService` is synchronous we can publish events and wait while all the notifiers proceed those events. On the other hand we don't use any native queries and always  remove entries using `EntityManager#remove` method, which means that JPA takes care of removal and publishes entity related events. In that case we can connect JPA events and `EventService` events to allow another modules/entities/components to react on the removal. All the database interaction methods are marked with the `@Transactional` annotation which may be nested.
  The example:

```
-> Transaction begins
-> before User#remove() 
       -> @PreRemove(User) event is published by JPA, 
       -> @PreRemove(User) event is handled and `BeforeUserRemoved` event is published to `EventService`
       -> BeforeUserRemoved event is catched by profile subscriber, 
       -> before ProfileDao#remove(profile)
             -> @PreRemove(Profile) event is published
            // ...
       -> ProfileDao#remove(profile) performed
      // ...
-> User#remove() performed
-> Transaction committed/rolled back
```

If any entry removal fails, then transaction will be rolled back and all the data will be in consistent state. 
Another nice thing in this strategy is that we can create application level components which may react on such kind of events and rollback transactions on the application level(which is useful for recipes and stacks their removal will be based on permissions).
It is still possible and preferable to use `orphanRemoval` for those entities which are not root model entities, for instance `Workspace` relies on the `WorkspaceConfig` with `orphanRemoval` attribute.

I implemented the last strategy as it is the most appropriate. Along with that i changed 
current Local dao implementations to  use self monitor for locking, as it allows to extend them with a new functionality which may also be thread-safe as client becomes allowed to use instance locks, this may be needed for migration tools.
#1803
### Previous Behavior

No cascade removals 
### New Behavior

All the entries removed when any of entries which they depend on is removed.
### Tests written?

Yes

@skabashnyuk, @sleshchenko, @akorneta please review

### DIFF
--- a/assembly/assembly-wsmaster-war/pom.xml
+++ b/assembly/assembly-wsmaster-war/pom.xml
@@ -60,6 +60,10 @@
         </dependency>
         <dependency>
             <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-jdbc</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
             <artifactId>che-core-api-machine</artifactId>
         </dependency>
         <dependency>
@@ -73,6 +77,10 @@
         <dependency>
             <groupId>org.eclipse.che.core</groupId>
             <artifactId>che-core-api-ssh</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-ssh-shared</artifactId>
         </dependency>
         <dependency>
             <groupId>org.eclipse.che.core</groupId>
@@ -119,6 +127,10 @@
             <artifactId>che-plugin-ssh-machine</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.eclipse.persistence</groupId>
+            <artifactId>eclipselink</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.everrest</groupId>
             <artifactId>everrest-core</artifactId>
         </dependency>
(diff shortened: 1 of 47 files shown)


assembly/assembly-wsmaster-war/pom.xml
```diff
@@ -131,6 +151,14 @@
             <artifactId>everrest-websockets</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-all</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.testng</groupId>
+            <artifactId>testng</artifactId>
```
sleshchenko: Are you sure that we need testng with compile scope?

assembly/assembly-wsmaster-war/src/test/resources/META-INF/persistence.xml
```diff
@@ -0,0 +1,63 @@
...
+        <class>org.eclipse.che.api.factory.server.model.impl.ButtonAttributesImpl</class>
+        <class>org.eclipse.che.api.factory.server.model.impl.ButtonImpl</class>
+        <class>org.eclipse.che.api.factory.server.model.impl.FactoryImpl</class>
+        <class>org.eclipse.che.api.factory.server.model.impl.IdeImpl</class>
+        <class>org.eclipse.che.api.factory.server.model.impl.OnAppClosedImpl</class>
+        <class>org.eclipse.che.api.factory.server.model.impl.OnProjectsLoadedImpl</class>
+        <class>org.eclipse.che.api.factory.server.model.impl.OnAppLoadedImpl</class>
+        <class>org.eclipse.che.api.factory.server.model.impl.PoliciesImpl</class>
+        <class>org.eclipse.che.api.factory.server.FactoryImage</class>
+        <!--<class>org.eclipse.che.api.factory.server.model.impl.AuthorImpl</class>-->
```
sleshchenko: I think you can remove this commented line because you already have AuthorImpl here https://github.com/eclipse/che/pull/2031/files#diff-2aebeda36e4acb8637396901436b9dd2R37

assembly/assembly-wsmaster-war/pom.xml
```diff
@@ -47,6 +47,10 @@
             <artifactId>guice-servlet</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.h2database</groupId>
+            <artifactId>h2</artifactId>
```
skabashnyuk: not needed here
driver will be a part of tomcat/lib

skabashnyuk:
ok

sleshchenko:
LGTM

akorneta:
ok

Issue #2039, opened 2016-08-03
Could not start workspace <wokespace-id> Reason: Timeout reached


dtoebe:
I loaded Che with the default docker start up explained in your docs: 
    [https://eclipse-che.readme.io/docs/usage-docker#usage](https://eclipse-che.readme.io/docs/usage-docker#usage)

The exact command I used: 

``` Bash
docker run -t -v /var/run/docker.sock:/var/run/docker.sock \                                       [09:54:18]
           codenvy/che-launcher:nightly start
```

Everything downloaded and booted fine. I went to http:localhost:8080 in Google Chrome (Not Chromium) created a Java Console Hello world workspace.  I didn't even change the name. When it IDE i get a popup error message: 

"
Could not start workspace wksp-9b0k. Reason: Timeout reached. The Che server has been unable to verify that your workspace's agent has successfully booted. Either the workspace is unreachable, the agent had an error during startup, or your workspace is starting slowly. You can configure machine.ws_agent.max_start_time_ms in Che properties to increase the timeout.
"

I created a gist with the logs `docker logs -f che-server`:

[https://gist.github.com/dtoebe/746ef514824078044636f271aecc94d3](https://gist.github.com/dtoebe/746ef514824078044636f271aecc94d3)

Here is a screenshot:

![screencapture-localhost-8080-dashboard-1470244400872](https://cloud.githubusercontent.com/assets/390097/17374863/69d03a50-5963-11e6-8f4f-9eda0698d381.png)

**Che version:**    docker:latest  (4.7.0-SNAPSHOT)
**OS and version:** Fedora 24 x86_64  
**Docker version:** 1.12.0 / **Go version** 1.6.3
**Che install:**    Docker   

_EDIT:_
I deleted all the docker images and `rm` the che dir in `/home/users`.  Then I ran the docker command again as stated above.  I created this gist of the CLI output:
[https://gist.github.com/dtoebe/7fd1478b967ad5637b994f855cf3d724](https://gist.github.com/dtoebe/7fd1478b967ad5637b994f855cf3d724)

After that I tried again and got the same results.

The only Issue I saw in the cli output was line 19 in the gist `ip: can't find device 'eth0'`

TylerJewell:
@l0rd @eivantsov @ddementieva - given that we are just now releasing the new launcher, let's dive into this issue.  It's Fedora, so I am unfamiliar with that setup.  But this looks like a networking issue.

TylerJewell:
[...]

PR #2040, opened 2016-08-03
Fix docker_host_ip lookup

TylerJewell:
### What does this PR do?
1. Replaces windows line endings with linux line endings.
2. Updates get_docker_host_ip() function to return docker0 if on native systems.
#2039
### Previous Behavior

On Fedora 24, the lookup method would return an error that eth0 is not discoverable.

### DIFF
--- a/che.sh
+++ b/che.sh
@@ -146,10 +146,17 @@ is_boot2docker() {
 }
 
 get_docker_host_ip() {
-  NETWORK_IF="eth0"
-  if is_boot2docker; then
-    NETWORK_IF="eth1"
-  fi
+  case $(get_docker_install_type) in
+   boot2docker)
+     NETWORK_IF="eth1"
+   ;;
+   native)
+     NETWORK_IF="docker0"
+   ;;
+   *)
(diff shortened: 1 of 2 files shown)


TylerJewell:
@l0rd - FYI, will you check this please.  Only changes were to get_docker_host_ip() method in launcher_funcs.sh.

l0rd:
@TylerJewell looking at the diff it seems like `get_docker_host_ip` is not the only change. You have included `che.sh` too. Was it intentional?

Besides that it's a good workaround. `docker0` is the default name for the docker bridge and should be there 99% of the times (the name of the bridge [can still be customized](https://docs.docker.com/engine/userguide/networking/default_network/custom-docker0/)). In the future we could improve it with something like `ip -f inet address show | grep "scope global" | head -1 | cut -d" " -f8`.

TylerJewell:
I had to make a fix in `che.sh` as well to match the function, as the error exists in both places.  So, it was intention this time.  I'll let you make an improvement to the function for a more global change if you want.  Glad we are able to work on these situations and get users a little more stable.

PR #2041, opened 2016-08-03
Update Vagrantfile to support new docker-launcher

TylerJewell:
### What does this PR do?

It updates the Vagrantfile in the root of the repository to use the che-launcher capability.  It also uses the che.sh CLI to automate the start and stop of the launcher.
### Previous Behavior

Previously, it would call codenvy/che-server directly and then run a loop to detect boot. This logic is now embedded within the launcher.
1. Added storage-overlay driver.
2. Use che-launcher instead of che-server image launche
3. Use the che.sh CLI to execute the start of the program.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -143,6 +143,9 @@ Vagrant.configure(2) do |config|
     # We need write access to this file to enable Che container to create other containers
     sudo chmod 777 /var/run/docker.sock &>/dev/null
 
+    # Setup the overlay storage driver to eliminate errors
+    sudo sed -i '/ExecStart=\/usr\/bin\/dockerd/c\ExecStart=\/usr\/bin\/dockerd --storage-driver=overlay' /lib/systemd/system/docker.service
+
     # Configure Docker daemon with the proxy
     if [ -n "$HTTP_PROXY" ] || [ -n "$HTTPS_PROXY" ]; then
         mkdir /etc/systemd/system/docker.service.d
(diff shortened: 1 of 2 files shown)


TylerJewell:
@l0rd - I have updated this Vagrantfile. I had to use the overlay storage driver to suppress a large number of warnings.  However, when I do this, it takes a longer time for the container to start. And the che-launcher timeout is being triggered even though the server is ultimately able to boot.

```
==> che: INFO: ECLIPSE CHE: ALREADY HAVE IMAGE codenvy/che-server:latest
==> che: INFO: ECLIPSE CHE: CONTAINER STARTING
==> che: INFO: ECLIPSE CHE: SERVER LOGS AT "docker logs -f che-server"
==> che: INFO: ECLIPSE CHE: SERVER BOOTING...
==> che: ---------------------------------------
==> che: ERROR: !!!
==> che: ERROR: !!! ECLIPSE CHE: Timeout waiting Che server to boot. Run "docker logs che-server" to see the logs.
==> che: ERROR: !!!
==> che: ---------------------------------------
==> che: Error: No such container: bf15a8792a3ef91a263bb5350394216c67634eb174b28f113034c4019f09684a
```

Given enough time, I can ping the che-server container and the server does boot.

l0rd:
You can safely increase the [timeout for the `start` command](https://github.com/eclipse/che/blob/master/dockerfiles/che-launcher/launcher_cmds.sh#L49) to 60 seconds. As we did it for the `stop` command last week.

l0rd:
And since you are on centOS you will probably get better performances with [`devicemapper` storage driver](https://docs.docker.com/engine/userguide/storagedriver/device-mapper-driver/).

TylerJewell:
That is the default storage driver - and it's configuration out of the box is throwing a variety of warning messages. So the switch over to a different storage driver didn't have those messages. So it's a bit of a one half dozen or another.  We probably should make the timeout configurable.

PR #2049, opened 2016-08-04
Fixes infinite loop on windows

TylerJewell:
### What does this PR do?
1. No longer uses (get_docker_host_ip) to determine if Docker for windows.
2. Prevents an infinite loop


TylerJewell:
I also added a "che debug" command, which is not displayed to the end user, but can help to provide information about what the CLI is discovering.  This also fixes the same loop introduced last night when we updated the discovery of docker install type.

PR #2053, opened 2016-08-05
CHE-476: Add new methods to GitUrlUtils

vinokurig:
### What does this PR do?

Add methods to get username and password from git url.
This is needed to paste these credentials to Jgit command when performing remote git operation 
with http url with credentials (http://username:password@host.xz/path/to/repo.git).

https://github.com/codenvy/codenvy/issues/476
### Tests written?

Yes

@sleshchenko @skabashnyuk please review

### DIFF
--- a/wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
+++ b/wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
@@ -11,6 +11,7 @@
  *******************************************************************************/
 package org.eclipse.che.git.impl.jgit;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.io.Files;
 import com.jcraft.jsch.JSch;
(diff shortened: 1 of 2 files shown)


wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitUrlUtils.java
```diff
@@ -17,16 +17,64 @@
...
     }
+
+    /**
+     * Check that given url contains credentials.
+     *
+     * @param url
+     *         git url
+     * @return {@code true} if given url contains credentials or {@code false} if not
+     */
+    public static boolean containsCredentials(String url) {
```
sleshchenko: Are you sure that it should be a part of API or maybe it would be better to make it part of jgit implementation?
Then it will looks more simple, just one block

```
if (url != null && GIT_URL_WITH_CREDENTIALS_PATTERN.matcher(url).matches()) {
    String username = url.substring(url.lastIndexOf(":") + 1, url.indexOf("@"));
    String password = url.substring(url.lastIndexOf(":") + 1, url.indexOf("@"));
    command.setCredentialsProvider(new UsernamePasswordCredentialsProvider(username, password));
}
```

vinokurig:
revorked in https://github.com/eclipse/che/pull/2072

PR #2054, opened 2016-08-05
Refactor TCK repositories

mkuznyetsov:
1. Remove unneeded TckRepository implementations, use JpaTckRepository<> instead.
2. Do not manipulate with different entities in one repository, use multiple repositories in tests instead.


voievodin:
Ok

sleshchenko:
LGTM

PR #2057, opened 2016-08-05
Connectivity test + refactoring

TylerJewell:
### What does this PR do?
1. Improves performance by limiting calls to docker.
2. Adds new networking connectivity test
3. Publishes "che debug" command with different options

Docs will be added.

### DIFF
--- a/che.sh
+++ b/che.sh
@@ -16,6 +16,28 @@ init_logging() {
   NC='\033[0m'
 }
 
+error_exit() {
+  echo  "---------------------------------------"
+  error "!!!"
+  error "!!! ${1}"
+  error "!!!"
+  echo  "---------------------------------------"
+  exit 1
+}
+
+check_docker() {
+  if ! docker ps > /dev/null 2>&1; then
+    output=$(docker)
+    error_exit "Error - Docker not installed properly: ${output}"
+  fi
+
+  # Prep script by getting default image
+  if [[ "$(docker images -q alpine 2> /dev/null)" == "" ]]; then
+    info "ECLIPSE CHE: PULLING IMAGE alpine:latest"
+    docker pull alpine > /dev/null 2>&1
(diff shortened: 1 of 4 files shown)


TylerJewell:
@l0rd @dmytro-ndp  - this has an upgraded version of the connectivity test.

garagatyi:
Are these exports really needed?

TylerJewell:
@garagatyi - only the last two exports are required.  The others are not - and they have been removed.

TylerJewell:
Also added:
1. Consolidated `che info` and `che debug` into a single command line offering.
2. Fixed error where docker run was failing, so commands after docker run syntax were skipped by CLI.
3. Moved all environment variables into temporary files and used `--env-file` which respects variables with spaces better.
4. Added support to che-launcher for inheriting environment variables.

l0rd:
I've tested on Mac with Docker 4 Mac where we still have issue #1644. In this case the expected behaviour should be:

> DEBUG: Browser             => Workspace Agent              : Connection failed

Instead the command hangs for some minutes and then return without showing the test results:

image

This is because the curl command silently fails :
image

To fix that I would use curl option `--connect-timout 5` to limit the timeout to 5 seconds and check for curl return value to decide if the connection succeeded or not.

l0rd:
The second test in `run_connectivity_test()`, (TEST 1a: Simulate browser (websockets)) does not add any value IMO.

In other words `run_connectivity_test()` tests Che connectivity at IP/TCP layer not at application layer.

TylerJewell:
@l0rd @riuvshin @tolusha 
1. Removed shell incompatibility from che.sh
2. Removed container self destruct
3. Fixed the get_hostname function to always return localhost unless it is boot2docker
4. Fixed network connectivity tests for mac - added in extra hostname test there as well

PR #2069, opened 2016-08-09
Fix realpath posix issue

TylerJewell:
### What does this PR do?

Replaces `realpath` with posix-compliant syntax to determine the fully qualified path of the directory passed in as a parameter.
### Previous Behavior

Previously, realpath is not installed on all posix-compliant systems.


TylerJewell:
@tolusha - please give this a try - should resolve issues you encountered with che-mount.

TylerJewell:
I'm +1 this after reviewing with @riuvshin - we need this included in the master for users on posix systems otherwise they will encounter issues.

PR #2079, opened 2016-08-09
Set proper dependency version

riuvshin:
### What does this PR do?

Fix dependency version

@skabashnyuk @vparfonov @benoitf


benoitf:
good catch

skabashnyuk:
ok

Issue #1824, opened 2016-07-18
IDE (client) - workspace agent (server) event transport service


dkulieshov:
### Overview

There is a need in a handy mechanics so we could transfer arbitrary information (events, data, etc.) from server side (i.e. workspace agent) to client (i.e. IDE web application) and backward.
### Description

Most likely it would require to implement services both on client and server sides. The idea is to make a transparent service that should listen to events and transfer appropriate data through the transport bus (e.g. `web sockets`) using predefined transport protocol. 

Here are a few examples of such events:
- Client
  - file opened in an editor
  - file directory is expanded in a project explorer
  - project properties are modified
- Server
  - file is modified
  - directory is removed
  - git operation is performed

It is also worth noting that services should not strictly define the way that data is transferred. The best way is to use some transport agnostic protocol to be able to extend or change the service. So we should consider means for convenient change of transport bus (e.g. from `web sockets` to `REST`). 

As a transport protocol most suitable seems to be `Json` `RPC` protocol or something similar. It is `Json` based so it would make it easy to implement and to use. 
### Technical aspects
- Bidirectional web socket conversation
  - Client-server data transfer protocol that are applicable in our application are asymmetric in a way that server cannot start conversation on demand but only a client
  - Web sockets are also asymmetric but there are ways how to overcome this limitation by using workarounds that only web socket protocol provides.
  - Bidirectional behavior requires us to implement both sides for requests and responses processing. Those implementations differs because of the nature of client and server.
- Several agents support
  - There are several use cases when there can be a few IDE's opened and connected to a single server (workspace agent)
    - Several tabs in a single browser
    - Several browsers
    - Several users of a single workspace
  - That requires specific management approach for data transfer (via `web socket`).
  - This will require both server side and client side modifications
  - This will not require any work related to UI 
- High level data transfer protocol (`Json` `RPC`)
  - Transport agnosticism - in order to be able to use `REST` and `web socket` depending on our needs
  - Conformance to the specification
- Effective client side and server side event handling
  - We have a good client side event handling but it does not support new data transfer approach
  - We have `VFS` event handling system implemented so we just need a proper event detector implementation
### Technical tasks
- [x] Client-side service to track client state #1830 
- [x] Detector of VFS high level events related to client side state #1832 
### User stories
- [x] File editor real time content update #1910
- [x] File explorer real time content update #1911
- [x] Git checkout operation notification #1912

dkulieshov:
[...]

PR #2080, opened 2016-08-09
Duplex Json RPC protocol implementation based on web socket transport

dkulieshov:
## Overview

There is a need in a simple, extensible, text-based and as symmetrical as possible approach that we could use to transport relevant data and call for remote procedures (RPC) between two endpoints: client (IDE) and server (workspace agent).
- Simplicity is required on two levels
  - Easy support in future releases
  - No pain for end-users (che developers)
- Extensibility obviously should make it a long living solution - save time and materials
- Using some variation of text-based data transfer formats seems to be suitable here: simple and clear enough to be read by human along with sufficient efficiency. 
- Bidirectional conversation: the client and the server both must have an equal ability to send and receive messages

In order to achieve those goals I have the following considerations:
- To be able to send messages in both directions on demand I think WEB SOCKETS protocol would be the most comfortable.
- As a human readable text-base data transfer format that also can be used for RPC we can use  JSON RPC 2.0 protocol or something similar but simpler.
- Simplicity and extensibility should be a result of proper making use of DI pattern implemented by  Guice and Gin.

So the solution should have two levels of abstraction: web socket based transport (low level) and Json RPC (high level). While a user works with simple enough high level RPC entities, the real magic happens on web socket level. The API of both levels must be independent of each other so it could be possible and easy enough to use them separately (e.g for Json RPC implementation to use another transport or for web socket based transport support other protocols (except Json RPC). 
Next, we will consider brief specification details associated with these abstractions, presented as a set of requirements or short rules.
## Web socket based transport (low level)

_Based on and extends [web socket specification](https://tools.ietf.org/html/rfc6455)._
#### Conversation
- Conversation happens between two endpoints
- Conversation happens via web socket protocol
- Conversation is a consequential sending and receiving of web socket transmission
#### Conversation endpoints
- Conversation has two participants: client (IDE) and server (workspace agent)
- Endpoints are equal and can perform same set of operations:
  - send transmission
  - receive transmission
- Server side endpoint (workspace agent) can take part in several conversations with several clients (IDE)
- Client side endpoint (IDE) can take part in one conversation with server side (workspace agent)
- There can be a several conversations client web socket conversations to a single server (workspace agent)
- Each client side (IDE) endpoint must provide a number to make it possible to be identified by server side during multiple conversations
#### Web socket transmission
- Web socket transmission is a JSON valid message
- Web socket transmission has two properties:
  - Protocol
  - Message
##### Protocol
- Protocol value must be recognized by the system and have a corresponding protocol implementation
- Protocol is a `String` value
- Protocol must not be empty
- Protocol must not be `null`
##### Message
- Message value is a string that a corresponding protocol implementation knows what to do with
- Message is a `String` value
- Message must not be empty
- Message must not be `null`
#### Examples:
- Example of web socket transmission:

``` json
{
    "protocol": "coffeemaker",
    "message": "could  you please make me some coffee?"
}
```

That means:
- Web socket transmission protocol is "coffeemaker"
- There is an implementation of this protocol in the system
- The implementation knows what to do with web socket transmission message
- Which is "could you please make me some coffee?"
### How to use
- If you want to make your own protocol implementation that will receive web socket transmissions you must provide and register an implementation of `WebSocketMessageReceiver` in `Guice` or `Gin` configuration modules: 

``` java
// GWT side
 GinMapBinder<String, WebSocketMessageReceiver> receivers =
GinMapBinder.newMapBinder(binder(), String.class, WebSocketMessageReceiver.class);
receivers.addBinding("coffeemaker").to(CoffeeMakerMessageReceiver.class);
```

``` java
// Server side
 MapBinder<String, WebSocketMessageReceiver> receivers =
MapBinder.newMapBinder(binder(), String.class, WebSocketMessageReceiver.class);
receivers.addBinding("coffeemaker").to(CoffeeMakerMessageReceiver.class);
```

Where "coffeemaker" is the name of the protocol that will be used in web socket transmissions.
- In order to send web socket transmissions you must use provided `WebSocketMessageTransmitter` implementation. Again it can be injected using `Guice`or `Gin` routines.
## Json RPC (high level)

_Based on [Json RPC specification](http://www.jsonrpc.org/specification)_
#### Conversation
- Conversation happens between two endpoints
- Conversation happens via Json RPC protocol
- Conversation is a consequential sending and receiving of Json RPC requests and responses
#### Conversation endpoints
- Conversation has two participants: client (IDE) and server (workspace agent)
-  Endpoints are equal and can perform same set of operations:
  - send Json RPC request
  - send Json RPC response
  - receive Json RPC request
  - receive Json RPC response
- Server side endpoint (workspace agent) can take part in several conversations with several clients (IDE)
- Client side endpoint (IDE) can take part in one conversation with server side (workspace agent)
- There can be a several conversations client web socket conversations to a single server (workspace agent)
- Each client side (IDE) endpoint must provide a number to make it possible to be identified by server side during multiple conversations
#### Json RPC request and response
- Json RPC request and response are defined in  [Json RPC specification](http://www.jsonrpc.org/specification). 
### How to use
- In order to receive Json Rpc requests you must provide and register an implementation of `JsonRpcRequestReceiver`:

``` java
// GWT side
GinMapBinder<String, JsonRpcRequestReceiver> requestReceivers =
GinMapBinder.newMapBinder(binder(), String.class, JsonRpcRequestReceiver.class);
requestReceivers.addBinding("method-name").to(CustomRequestReceiver.class);
```

``` java
// Server side
MapBinder<String, JsonRpcRequestReceiver> requestReceivers =
MapBinder.newMapBinder(binder(), String.class, JsonRpcRequestReceiver.class);
requestReceivers.addBinding("method-name").to(CustomRequestReceiver.class);
```

Where "method-name" is a method property of Json RPC request.
- In order to receive Json Rpc responses you must provide and register an implementation of `JsonRpcResponseReceiver`:

``` java
// GWT side
GinMapBinder<String, JsonRpcResponseReceiver> responseReceiver =
GinMapBinder.newMapBinder(binder(), String.class, JsonRpcResponseReceiver.class);
responseReceiver.addBinding("method-name").to(CustomResponseReceiver.class);
```

``` java
// Server side
MapBinder<String, JsonRpcResponseReceiver> requestReceivers = MapBinder.newMapBinder(binder(), String.class, JsonRpcResponseReceiver.class);
responseReceiver.addBinding("method-name").to(CustomResponseReceiver.class);
```

Where "method-name" is a method property of Json RPC request that this response answers to.
- To send Json RPC request inject provided instance of `JsonRpcRequestTransmitter` interface.
- To send Json RPC response inject provided instance of `JsonRpcResponseTransmitter` interface.
### Examples:
- Example of Json RPC request:

``` json
{ 
   "jsonrpc": "2.0", 
   "method": "subtract", 
   "params": [42, 23], 
   "id": 1
}
```
- Example of Json RPC response

``` json
{
   "jsonrpc": "2.0",
   "result": 19,
   "id": 1
}
```
## Further considerations

There is a number of functionality or improvements that I see useful but not covered by this pull request:
- This implementation and specification does not guarantee message deliverance on web socket level shifting the responsibility to high level protocols. So it may be quite useful if we could provide any mechanism that guarantees such deliverance on low level to simplify high level protocols implementation even more.
- The support of Json RPC protocol is not complete as there is no need for that at the moment though it may become important and required in future:
  - Json RPC errors are supported partially
  - Json RPC validation
  - Json RPC batch requests
- Some of Json RPC routines related to request/response generation can be improved/simplified by generalization
- There is quite a big amount of functional duplication on GWT and server side caused by different approaches in coding and build/compile mechanism but also because of partially asymmetric interaction mechanism between endpoints (several to one conversations). At the moment I don't think that is not an urgent problem but probably this can and should be improved in future.
- As @evoevodin suggested, there are some places where we can automate processes and free end user of extra routines:
  - Json Rpc requests and responses IDs can be automatically read on reception and assigned on transmission
  - Json Rpc request body (params) can be automatically created parsing `String` and instantiating corresponding `DTO`s by specifying appropriate `DTO` class as generic type for receivers
- At the moment there is no guaranties as of the order of transmissions (mostly because of transport algorithms simplification concerns) passed to web socket layer, but that is quite obvious that we can benefit of having such ones in future ( @evoevodin )

### DIFF
--- a/assembly/assembly-wsagent-war/src/main/java/org/eclipse/che/wsagent/server/WsAgentModule.java
+++ b/assembly/assembly-wsagent-war/src/main/java/org/eclipse/che/wsagent/server/WsAgentModule.java
@@ -12,6 +12,7 @@ package org.eclipse.che.wsagent.server;
 
 import com.google.inject.AbstractModule;
 import com.google.inject.Provides;
+import com.google.inject.multibindings.MapBinder;
 import com.google.inject.name.Names;
 
 import org.eclipse.che.ApiEndpointAccessibilityChecker;
@@ -20,18 +21,36 @@ import org.eclipse.che.EventBusURLProvider;
 import org.eclipse.che.UriApiEndpointProvider;
 import org.eclipse.che.UserTokenProvider;
 import org.eclipse.che.api.auth.oauth.OAuthTokenProvider;
+import org.eclipse.che.api.core.jsonrpc.JsonRpcRequestReceiver;
+import org.eclipse.che.api.core.jsonrpc.JsonRpcRequestTransmitter;
+import org.eclipse.che.api.core.jsonrpc.JsonRpcResponseReceiver;
+import org.eclipse.che.api.core.jsonrpc.JsonRpcResponseTransmitter;
+import org.eclipse.che.api.core.jsonrpc.impl.BasicJsonRpcObjectValidator;
+import org.eclipse.che.api.core.jsonrpc.impl.JsonRpcDispatcher;
+import org.eclipse.che.api.core.jsonrpc.impl.JsonRpcObjectValidator;
+import org.eclipse.che.api.core.jsonrpc.impl.WebSocketJsonRpcDispatcher;
+import org.eclipse.che.api.core.jsonrpc.impl.WebSocketJsonRpcRequestDispatcher;
+import org.eclipse.che.api.core.jsonrpc.impl.WebSocketJsonRpcRequestTransmitter;
+import org.eclipse.che.api.core.jsonrpc.impl.WebSocketJsonRpcResponseDispatcher;
+import org.eclipse.che.api.core.jsonrpc.impl.WebSocketJsonRpcResponseTransmitter;
 import org.eclipse.che.api.core.notification.WSocketEventBusClient;
 import org.eclipse.che.api.core.rest.ApiInfoService;
 import org.eclipse.che.api.core.rest.CoreRestModule;
 import org.eclipse.che.api.core.util.FileCleaner.FileCleanerModule;
+import org.eclipse.che.api.core.websocket.WebSocketMessageReceiver;
+import org.eclipse.che.api.core.websocket.WebSocketMessageTransmitter;
+import org.eclipse.che.api.core.websocket.impl.BasicWebSocketTransmissionValidator;
+import org.eclipse.che.api.core.websocket.impl.GuiceInjectorEndpointConfigurator;
+import org.eclipse.che.api.core.websocket.impl.BasicWebSocketMessageTransmitter;
+import org.eclipse.che.api.core.websocket.impl.WebSocketTransmissionValidator;
 import org.eclipse.che.api.git.GitConnectionFactory;
 import org.eclipse.che.api.git.GitUserResolver;
+import org.eclipse.che.api.git.LocalGitUserResolver;
 import org.eclipse.che.api.project.server.ProjectApiModule;
 import org.eclipse.che.api.ssh.server.HttpSshServiceClient;
 import org.eclipse.che.api.ssh.server.SshServiceClient;
 import org.eclipse.che.api.user.server.spi.PreferenceDao;
 import org.eclipse.che.commons.lang.Pair;
-import org.eclipse.che.api.git.LocalGitUserResolver;
 import org.eclipse.che.git.impl.jgit.JGitConnectionFactory;
 import org.eclipse.che.inject.DynaModule;
 import org.eclipse.che.plugin.java.server.rest.WsAgentURLProvider;
@@ -78,6 +97,9 @@ public class WsAgentModule extends AbstractModule {
 
         bind(String.class).annotatedWith(Names.named("wsagent.endpoint"))
                           .toProvider(WsAgentURLProvider.class);
+
+        configureJsonRpc();
+        configureWebSocket();
     }
 
     //it's need for WSocketEventBusClient and in the future will be replaced with the property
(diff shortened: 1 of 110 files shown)


benoitf:
as client is not only the IDE, it could be dashboard, I would like to see JSON RPC client integrated as well for UD.

dkuleshov:
ci-build

dkuleshov:
ci-build

vparfonov:
ok

benoitf:
@dkuleshov you didn't reply to my comment that I left 8 days ago

dkulieshov:
@benoitf I beg your pardon! In fact I did not realize that it was addressed to me directly but not to the audience, because it looked more like a call to general discussion.

As I understand you are interested in a separate javascript client for UD. That seems quite logical and reasonable to me, I see that we can benefit of that so I would like to assist you in achieving that goal. Unfortunately and to my great regret it is out of scope for this very pull request and this task in general because as we can see from the description it was initiated mostly to provide communication between two endpoints: IDE and workspace agent. However I have no doubt that we should discuss your suggestion in some other, more convenient place with a more vast circle of people.

benoitf:
@dkuleshov well I don't see why the scope was limited to only one client while we have different clients connected to the remote agents (IDE, UD, CLI, etc). IDE is only one client among all the others then discussion should be broader before getting into a specific client use case. Then you could write for a specific usecase.
Also I didn't see any github issue related to this pull request so it can be discussed before getting a pull request submitted

dkulieshov:
@benoitf Sorry for the late response! I'm sad to admit but I have no information about the reasoning of scope limitations because in fact I did not participate in those decisions. However if you are indeed interested in the topic why don't you create a dedicated issue so we could shift the discussion there.

Actually the issue related to this PR is https://github.com/eclipse/che/issues/1824. To my shame, there is no direct mentioning of it in this PR but with a little effort you could possibly find it if you had taken a look onto commit messages and the name of the branch related to this PR.

vparfonov:
@benoitf What problem you met with this pull? If you don't want use it in UD so you should not, this is  only our draft that we want to test and will see is it good for our future.

benoitf:
@vparfonov there is not only UD but other clients.
For example maybe cli could use it to detect local folder changes as well

PR #2083, opened 2016-08-09
Fix che.sh issue on linux machines

riuvshin:
### What does this PR do?

Fix che.sh issue on linux machines

@TylerJewell


TylerJewell:
test - verified on win + mac.

Issue #1865, opened 2016-07-19
After select an item in the Git compare widget the Compare button is disabled


musienko-maxim:
**Reproduction Steps:**
1. Clone a project
2. Change some files
3. Select the folder with files and Git -> Compare -> Compare with latest repository version

**Expected behavior:** after clicking on the fires item the Compare button should be enabked
**Observed behavior:** after clicking on the fires item the Compare button still disabled
**Che version:**    4.5.0
**OS and version:** Ubuntu OS 14.0
**Docker version:** 1.11.2
**Che install:**  local bundle
**Additional information:**
- Problem started happening recently, didn't happen in an older version of Che: YES
  Related selenium test: GitCompareTest

PR #2091, opened 2016-08-10
che-1865: Fix disable compare button in the compare widget when we have changed few files

AndrienkoAleksandr:
### What does this PR do?

Fix disable compare button when we have changed few files and open compare widget. By default tree with changes select first element and compare button should be enabled.

https://github.com/eclipse/che/issues/1865
### Tests written?

No
### Docs requirements?

Include the content for all the docs changes required. 
1.  API changes  : without changes
2.  User docs changes  :  without changes

@vinokurig  review please

### DIFF
--- a/plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/changedList/ChangedListPresenter.java
+++ b/plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/changedList/ChangedListPresenter.java
@@ -67,11 +67,13 @@ public class ChangedListPresenter implements ChangedListView.ActionDelegate {
     public void show(Map<String, Status> changedFiles, String revision, Project project) {
         this.changedFiles = changedFiles;
         this.project = project;
+        this.revision = revision;
+
+        view.setEnableCompareButton(false);
+        view.setEnableExpandCollapseButtons(treeViewEnabled);
+
         view.showDialog();
         viewChangedFiles();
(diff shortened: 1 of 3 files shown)


vinokurig:
Ok to me

AndrienkoAleksandr:
@azatsarynnyy @sleshchenko review please

azatsarynnyy:
ok

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/changedList/ChangedListPresenter.java
```diff
@@ -68,9 +68,9 @@ public void show(Map<String, Status> changedFiles, String revision, Project proj
         this.changedFiles = changedFiles;
         this.project = project;
         view.showDialog();
+        view.setEnableCompareButton(false);
```
sleshchenko, line 72: mb it would be better to init state of view and only then show dialog?

sleshchenko:
OK

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/changedList/ChangedListPresenter.java
AndrienkoAleksandr, line 72: ok

sleshchenko:
OK

skabashnyuk:
+0

PR #2096, opened 2016-08-11
Remove *.sh from gitignore

riuvshin:
As we have lot of sh scripts in that repo it is really hard to manage them if it's ignored.

@skabashnyuk @TylerJewell

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,6 @@ test-output/
 *.exe
 *.o
 *.so
-*.sh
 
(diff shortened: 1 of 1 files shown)


skabashnyuk:
+0

benoitf:
with current gitignore I've forget twice to add my .sh script from my current branch
so +1

TylerJewell:
I've missed it a couple times as well.  So I am +1

PR #2099, opened 2016-08-11
Convert nightly to master

TylerJewell:
### What does this PR do?

Moves all images from nightly to latest as default.


riuvshin:
what about https://github.com/eclipse/che/blob/master/che.sh#L460 ?

TylerJewell:
@riuvshin - aye!  good catch.  Fixed.

PR #2100, opened 2016-08-11
update che.sh to get rid of weird issue on linux machines

riuvshin:
### What does this PR do?

Fix che.sh issues on some linux machines

@TylerJewell


TylerJewell:
lgtm

Issue #2102, opened 2016-08-11
Che launcher debug prints wrong port / hostname for running server if variables change


TylerJewell:
The debugging output of che-launcher is wrong if you have a running server whose CHE_HOSTNAME or CHE_PORT values change after the server was started.

**Reproduction Steps:**
1. Start che with launcher.
2. Set CHE_HOSTNAME or CHE_PORT to a value that is not the value of the currently running server.
3. Run `che info`

Gives wrong PORT & HOSTNAME values for the running server.

This is the offending code:
https://github.com/eclipse/che/blob/master/dockerfiles/che-launcher/launcher_cmds.sh#L121-L122

@l0rd - do you have a clever way to interrogate a running che-server container to find out which hostname and port it is using instead of inheriting from environment variables?

l0rd:
I think we have 2 options here:
- Set the hostname to `$DEFAULT_CHE_HOSTNAME` and get the port using the command `docker inspect --format='{{ (index (index .NetworkSettings.Ports "8080/tcp") 0).HostPort }}' ${CHE_SERVER_CONTAINER_NAME}`
- Write a `.che-launcher` file (that could be stored in `CHE_DATA_FOLDER`) where all the launcher options are written whe `che` is running.

The first option is the easiest to implement (and I probably prefer it right now). The second option has the benefit to be more reliable and make the `restart` command consistent too.

PR #2105, opened 2016-08-12
Optimize che-launcher for perf

TylerJewell:
### What does this PR do?

This pull request implements optimizations to minimize the number of times that `docker run` is called within the `che-launcher` utility. Previously, there were numerous helper functions that have embedded `docker run` commands. If these functions were repeatedly called, we were launching unnecessary containers when one invocation will do.

Also fixes a small error where one of the docker run commands did not have a `--rm` leaving a lot of lingering exited containers on the user's system.

Also adds in a build.sh into the directory to perform a build of the image.

### DIFF
--- a/dockerfiles/che-launcher/build.sh
+++ b/dockerfiles/che-launcher/build.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+# Copyright (c) 2012-2016 Codenvy, S.A., Red Hat, Inc
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License v1.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-v10.html
+if [ "latest" = "$1" ]
+then
+  TAG="latest"
+else
+  TAG="nightly"
+fi
+
+docker build -t codenvy/che-launcher:$TAG .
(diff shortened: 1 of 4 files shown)


TylerJewell:
@benoitf @l0rd @riuvshin - FYI - one of you +1.  The speed of a `che info` command is now dramatically faster.

l0rd:
@TylerJewell 👏 👏  the launcher is really faster now. 

I have one comment:

You are calling 3 times `docker run` at the beginning of the launcher to set `ETH0_ADDRESS`, `ETH1_ADDRESS` and `DOCKER0_ADDRESS` but you really just need one of them. You can check if `DEFAULT_DOCKER_HOST_IP has already been set at the very beginning of`get_docker_host_ip()` instead:

```
if [ -n "${DEFAULT_DOCKER_HOST_IP}" ]; then
    echo "${DEFAULT_DOCKER_HOST_IP}"
fi
```

You could do the same in `has_docker_for_windows_ip()` and `get_che_launcher_version()`: if the variables are already set don't run the command, otherwise run it.

As a result you will be able to call the functions directly as many times as you need and you should not need to worry to run all the docker run command at the beginning (but just when you need them).

TylerJewell:
@l0rd - so I studied this for a bit this morning and concluded that the system would not profit. In order to implement variable check within `get_docker_host_ip`, I'd have to reintroduce additional `docker run` command back into is_docker_for_windows to avoid an infinite loop, thus impacting performance again.

TylerJewell:
Added a fix for #2102 as well.

l0rd:
LGTM

PR #2111, opened 2016-08-12
Development and production mode for Artik devices

vitaliy-guliy:
Fixed functionality of broken Targets dialog.

https://github.com/codenvy/artik-ide/issues/94


ashumilova:
+1

Issue #2116, opened 2016-08-13
Iterative improvements to simplify usage of che in che scenarios


TylerJewell:
The goal is to minimize the steps that a user needs to take to get Che in Che working with our current architecture. Che in Che involves setting up a volume folder mount, getting a proper stack with Docker inside of it, having special build commands, and then additional macros in commands to simplify the discovery of the server for execution.
- [x] Add `machine.server.extra.volume` as CHE_ environment variable.  The logic to add would be in. Implemented initially with CLI improvement: #2239.  Will be added into the core Che server as well for #2175.
- [x] Improve `che-launcher` container to allow user to set the name of the `che-server` container that will be launched (and stopped). #2119
- [x] Add Docker  to the base alpine image, so that it is available in all workspaces.  #2119 
- [x] Add in missing macros for commands necessary for doing discovery. #953 
- [x] #2119 Create a Che stack that is optimized with alpine software needed to build Che in Che.  Include it as a ready-to-go stack in the default che distribution.
- [x] #2119 Add Che template - that allows to setup Che from our github repo along with commands to build 
- [x] Change ERROR to WARNING (yellow) message for two scenarios - when pulling a docker image remotely that is not present, and then switching to local.  And also when GitHub oAuth is not configured during ws-agent boot. #2135 
- [x] Extend this capability for artik ide. 

The expected end user scenario would be:
1: start che
2: create ws from che stack
3: select template for building all of che (maybe we hvae template for just extension in future)
4: let them build a module
5: have built-in command to run the new che-in-che with properly displayed preview

Test cases:
1: Start Che using che-launcher with extra volume folder set to /var/run/docker.sock using environment variable instead of setting che.properties file.

2: Launch a Che stack with Che template, get a clone of the Che repository.  Ability to build & run it with the embedded commands.

3: Ability to run two che on the same computer side-by-side by using Che CLI profile - create two profiles, set one, start change, change the profile, and start the second che.

4: Test ability to launch Che in superdev mode using Che-in-Che.

PR #2119, opened 2016-08-14
Changes to support che-in-che

TylerJewell:
### Motivation

There are a variety of things that we need to add to simplify running Che within a Che workspace for new users. The full description of the use case is located in #2116.
### What does this PR do?

Adds in capabilities to make it easier to run che-in-che scenarios.
1: Improved the `che-launcher` and `che.sh` scripts so that the name of the che-server container can be defined by an environment property.  This will allow workspaces starting a second che-server to change the name so that it does not conflict with the one that is already running.

2: Added `CHE_WORKSPACE_VOLUME` environment variable that lets user set an extra volume to be mounted into Che workspaces when they are started. Also introduces `che.workspace.volume` property and uses that value instead of `machine.server.extra.volume` property.

3: Adds a new ready-to-go stack that lists the components that are part of the new Che recipe and launches that container.

4: Adds a new template mapped to the Che stack, that will clone the che repo and add a series of commands to build / run che within Che.

5: Adds fix for Ubuntu where `--env-file` was getting misinterpreted by echo as -e flag.

6: Adds fix to handle Windows directories with spaces in the names for the mount path conversion.

7: Adds enhancement to provide new profile capability into the CLI.  Profile capability lets user store the values of current environment variables into a profile that will be loaded and used by the che-launcher. This allows users to save multiple Che configurations and switch between them.

```
che profile add <name>                 Add a profile to ~/.che/ 
che profile set <name>                 Set this profile as the default for Che CLI
che profile unset                      Removes the default profile - leaves it unset
che profile rm <name>                  Remove this profile from ~/.che/
che profile update <name>              Update profile in ~/.che/
che profile info <name>                Print the profile configuration
che profile list                       List available profiles
```

8: Added fix for CHE_LOCAL_BINARY on boot2docker systems.

9: Added fix for CHE_DATA_FOLDER on boot2docker systems. Before we were only adding in the new CHE_WORSKPACE_STORAGE_CREATE_FOLDERS if it was docker for windows or mac, but this generally needed on all systems now.

10: Adds additional environment variables which can be end user configurable and stored within a Che profile. The benefit of setting these values is that you can have different profiles with different names for CHE_SERVER_CONTAINER name which lets you run multiple Che servers at the same time without the container names conflicting.  Also, this allows us to support assemblies like the ARTIK IDE by letting them reuse the launcher and the CLI just changing a profile to get it to work.

```
  CHE_LAUNCHER_IMAGE_NAME
  CHE_SERVER_IMAGE_NAME
  CHE_FILE_IMAGE_NAME
  CHE_MOUNT_IMAGE_NAME
  CHE_TEST_IMAGE_NAME

  CHE_LAUNCHER_CONTAINER_NAME
  CHE_SERVER_CONTAINER_NAME
  CHE_FILE_CONTAINER_NAME 
  CHE_MOUNT_CONTAINER_NAME
  CHE_TEST_CONTAINER_NAME
```


TylerJewell:
@eivantsov @benoitf @riuvshin - can you please test the new capabilities related to the che profile on Ubuntu?  Sequence for users would be the following.

```
che profile add my-profile
che profile info my-profile
che profile set my-profile
che start  <-- uses values of my-profile
che profile unset 
che start <-- uses defaults or env variables
```

@eivantsov - one action would be to update the ARTIK Vagrant file to use this latest Che CLI. User can set CHE_ environment variables to define the name of the image and the name of the container that will be created.

benoitf:
+1
I will need this PR merged to use the CHE_SERVER_CONTAINER_NAME / CHE_LAUNCHER_CONTAINER_NAME option

che.sh
```diff
@@ -488,15 +514,216 @@ run_connectivity_tests() {
...
+      echo "CHE_TEST_IMAGE_NAME=$CHE_TEST_IMAGE_NAME" >> ~/.che/"${3}"
+
+      echo "CHE_LAUNCHER_CONTAINER_NAME=$CHE_LAUNCHER_CONTAINER_NAME" >> ~/.che/"${3}"
+      echo "CHE_SERVER_CONTAINER_NAME=$CHE_SERVER_CONTAINER_NAME" >> ~/.che/"${3}"
+      echo "CHE_FILE_CONTAINER_NAME=$CHE_FILE_CONTAINER_NAME" >> ~/.che/"${3}"
+      echo "CHE_MOUNT_CONTAINER_NAME=$CHE_MOUNT_CONTAINER_NAME" >> ~/.che/"${3}"
+      echo "CHE_TEST_CONTAINER_NAME=$CHE_TEST_CONTAINER_NAME" >> ~/.che/"${3}"
+
+      env | grep CHE_ >> "${3}"
+      cat ~/.che/"${3}" | sort | uniq > ~/.che/"${3}"
```
benoitf: need to use intermediate file like in add section

benoitf:
I'm eager to have it in master 👍

Issue #1884, opened 2016-07-20
Move "Subversion" menu after to "Git" in project's explorer right-click menu


slemeur:
Move "Subversion" menu after to "Git" in project's explorer right-click menu

<img width="502" alt="screen shot 2016-07-20 at 19 00 38" src="https://cloud.githubusercontent.com/assets/1636769/16995691/09b1fe2a-4eae-11e6-92ee-571312f99459.png">

"Subversion" must not be at the top of the menu - it should be the next entry after "Git".

**Che version:**    4.5.0

PR #2126, opened 2016-08-15
CHE-1884: Move down svn item in the context menu

vinokurig:
### What does this PR do?

Move subversion item down to git in context menu

https://github.com/eclipse/che/issues/1884
### Previous Behavior

svn_1
### New Behavior

svn_2

@vparfonov please review

### DIFF
--- a/plugins/plugin-svn/che-plugin-svn-ext-ide/src/main/java/org/eclipse/che/plugin/svn/ide/SubversionExtension.java
+++ b/plugins/plugin-svn/che-plugin-svn-ext-ide/src/main/java/org/eclipse/che/plugin/svn/ide/SubversionExtension.java
@@ -199,7 +199,7 @@ public class SubversionExtension {
 
         contextGroup.getTemplatePresentation().setSVGResource(resources.svn());
 
-        DefaultActionGroup mainContextMenuGroup = (DefaultActionGroup)actionManager.getAction("resourceOperation");
-        mainContextMenuGroup.add(contextGroup);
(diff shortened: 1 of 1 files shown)


plugins/plugin-svn/che-plugin-svn-ext-ide/src/main/java/org/eclipse/che/plugin/svn/ide/SubversionExtension.java
```diff
@@ -199,8 +199,7 @@ public SubversionExtension(final ActionManager actionManager,
 
         contextGroup.getTemplatePresentation().setSVGResource(resources.svn());
 
-        DefaultActionGroup mainContextMenuGroup = (DefaultActionGroup)actionManager.getAction(IdeActions.GROUP_MAIN_CONTEXT_MENU);
-        mainContextMenuGroup.addSeparator();
+        DefaultActionGroup mainContextMenuGroup = (DefaultActionGroup)actionManager.getAction("resourceOperation");
```
vparfonov: Can we use some Constant here?

vinokurig: This is inner part of main context menu (https://github.com/eclipse/che/blob/24bdc4d09f6435df9b51ea1a6de25bbdd1126f20/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/action/IdeActions.java#L34)
We haven't a constant for this section, so it is defined by string everywhere.

vparfonov:
ok

Issue #1912, opened 2016-07-22
Git checkout operation notification


dkulieshov:
### User story

As a `Che` user I want to be informed if git project that I'm working on in `Che` IDE is modified by git checkout operation irrespective of source of modification.
### Scenario
- User **A** opens a git project **P** in workspace **W** in `Che` IDE
- User **A** or user **B** `ssh` to workspace **W**
- User **A** or user **B** perform `Git` `checkout` operation
  - Checking out of a `branch`
  - Checking out of a `commit`
  - Checking out of a `tag`
- `Git` operation is performed outside of `Che` by
  - Third party `Git` client 
  - CLI
- User **A** is notified by `Che` that `Git` operation is performed
- All relative data is updated automatically (no refresh needed)

JamesDrummond:
@vparfonov Provide status label.

Issue #1910, opened 2016-07-22
File editor real time content update


dkulieshov:
### User story

As a `Che` user I want to have actual content of file in the editor irrespective of source of file modification.
### Scenario
- User **A** opens a file **F** in `Che` IDE in workspace **W**
- User **A** or user **B** `ssh` mounts to workspace **W**
- User **A** or user **B** modifies the file **F** using outside means
  - IDE (Intellij IDEA, Eclipse, NetBeans)
  - CLI
  - Text Editor (Sublime, TextMate, vi, etc.)
- Content of the file **F** is updated automatically (no refresh needed)

JamesDrummond:
@vparfonov Provide status label.

PR #2153, opened 2016-08-17
Editor file content synchronization and git checkout notification

dkulieshov:
### What does this PR do?
- Adds means to synchronize content of file opened in editor when it is being modified outside of che
- Adds means to track outside git check out operations (branch checkout or revision checkout) and inform a user

https://github.com/eclipse/che/issues/1910
https://github.com/eclipse/che/issues/1912
### New Behavior

If file that is opened in che editor tab is modified somewhere else its content is updated
If git repository is checked out outside of che UI a user is notified about that
### Tests written?

Yes


dkulieshov:
ci-build

vparfonov:
ok

PR #2159, opened 2016-08-18
Make WorkspaceImpl usable in JPA

tareksha:
### What does this PR do?

Remove 'final' modifier from fields of WorkspaceImpl

JPA does not support final fields in entity classes. Removing the final modified enables using WorkspaceImpl as a JPA entity
### Previous Behavior

Mapping WorkspaceImpl as a JPA entity failed.
### New Behavior

Mapping WorkspaceImpl as a JPA entity works.
### Tests written?

No

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/model/impl/WorkspaceImpl.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/model/impl/WorkspaceImpl.java
@@ -34,8 +34,8 @@ public class WorkspaceImpl implements Workspace {
         return new WorkspaceImplBuilder();
     }
 
-    private final String id;
-    private final String namespace;
+    private String id;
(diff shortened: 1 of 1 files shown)


tareksha:
@skabashnyuk  @vparfonov please review

voievodin:
Hi @tareqhs!

The code is okay :+1: 
Actually we are currently  working on JPA integration, 
if you are interested check out our [working branch](https://github.com/eclipse/che/tree/jpa-integration) and the [WorkspaceImpl](https://github.com/eclipse/che/blob/jpa-integration/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/model/impl/WorkspaceImpl.java) object itself. 
We are planning to merge it soon, so maybe this will reduce some amount of work you are planning to do, if you have any questions about it - feel free to ask!

tareksha:
@evoevodin that's good news ! I'll look into that branch and catch up ASAP

PR #2162, opened 2016-08-18
Pack ws-agent binaries in tar.gz archive

tolusha:
Tho goal is to reduce number of installing components in the container.

@skabashnyuk 
@sleshchenko 
pls review


sleshchenko:
+1

skabashnyuk:
ок

Issue #953, opened 2016-04-01
Macro for selected files in the project explorer


dhuebner:
It would be handy to have a macro that points to the current selected file (multiple files could be separated by a comma) in the package explorer:

e.g. ${explorer.selected} 

https://eclipse-che.readme.io/docs/commands#macros

TylerJewell:
It seems that the initial set of command macros that have been provided are somewhat limited right now.  Let's go ahead and define a wider list of macros for inclusion.

```
${current.project.file} -- Currently selected file
${current.project.path}  -- Absolute path to the selected file
${current.project.relpath} -- Path relative to the /projects folder
${server.port.<port>} -- Returns protocol, hostname and port of a service registered with an EXPOSE port
${server.port.<port>.protocol} -- Returns protocol of a service registered with an EXPOSE port
${server.port.<port>.hostname} -- Returns hostname of a service registered with an EXPOSE port
${server.port.<port>.port} -- Returns port of a service registered with an EXPOSE port

** Note that <port> is the port that is exposed by the Dockerfile provided by the user.  But the port returned is the ephemeral port assigned by Docker to that server.
```

juampe:
If this is implemented, will be a big ergonomic enhancement for language scripting users

command: cd /projects/python && PYTHONUNBUFFERED=1 python ${current.project.file}
/bin/bash: ${current.project.file}: bad substitution

TylerJewell:
@juampe - we are going to include the analysis and implementation of many of these into our next sprint.

gazarenkov:
better
${server.<ref>} -- Returns server by ref name and so on

TylerJewell:
Proposed new specification:

```
${current.project.file} -- Currently selected file
${current.project.path}  -- Absolute path to the selected file
${current.project.relpath} -- Path relative to the /projects folder
${server.<port>} -- Returns protocol, hostname and port of an internal server
${server.<port>.protocol} -- Returns protocol of a service registered with an EXPOSE port
${server.<port>.hostname} -- Returns hostname of a service registered with an EXPOSE port
${server.<port>.port} -- Returns port of a service registered with an EXPOSE port
${server.<ref>} -- Returns protocol, hostname and port of an internal server
${server.<ref>.protocol} -- Returns protocol of a server registered by name
${server.<ref>.hostname} -- Returns hostname of a server registered by name
${server.<ref>.port} -- Returns port of a server registered by name
```

sunix:
Hi,
[...]

PR #2169, opened 2016-08-19
#953: Move command property value providers to core API

vzhukovs:
Move command property value provider interface and registry to the core API.

Need for: https://github.com/eclipse/che/issues/953

@vparfonov @azatsarynnyy review it, plz.


azatsarynnyy:
+1

vparfonov:
ok

Issue #1706, opened 2016-07-11
Struggling working with Japanese filenames. 


Leward:
Eclipse Che (I use it via codenvy.com) seems not to work very well with Japanese filenames. 

**Reproduction Steps:**
1. Use the "code" button on this repo : https://github.com/Leward/leward-blog 
2. Try to rename a japanese filename via the UI (remove the first character of the filename)
3. Try to rename a japanse filename through the command line (using the `mv` command, remove the first character of the filename)

**Expected behavior:**

The filename with Japanese characters should be an acceptable value.
Renaming via command line should rename the file with the expected value. 

**Observed behavior:**

The filename is not accepted in the UI : 
![rename_jp_file_ui](https://cloud.githubusercontent.com/assets/1108235/16732091/ae98f6f8-477b-11e6-90f6-b80690384bed.png)

Though command line it does not rename the file as expected: 

```
$ ls -l
drwxr-xr-x 2 user user 138 Jul 11 13:20 ./
drwxr-xr-x 3 user user  37 Jul 11 12:54 ../
-rw-r--r-- 1 user user 244 Jul 11 12:45 201-JPブログ.markdown

$ mv 201-JPブログ.markdown 01-JPブログ.markdown
$ ll                                                                                                                                                                                        
drwxr-xr-x 2 user user 137 Jul 11 13:26 ./
drwxr-xr-x 3 user user  37 Jul 11 12:54 ../
-rw-r--r-- 1 user user 244 Jul 11 12:45 201JPブログ.markdown
```

Filenames: 

```
201-JPブログ.markdown # original
01-JPブログ.markdown # expected
201JPブログ.markdown # actual
```

**Che version:**    Codenvy 4.4.2
**OS and version:** Don't know
**Docker version:** Don't know  
**Che install:**    Codenvy 

**Additional information:**
- Problem started happening recently, didn't happen in an older version of Che: I cannot tell
- Problem can be reliably reproduced, doesn't happen randomly: Yes

thyeun:
i have the same issue also for created a Chinese/Korean character name for a new file. 

Che version : 4.4.0
OS and version : windows 10 build 10240

TylerJewell:
Yes - just double checked.  Issue confirmed.

ddementieva:
https://jira.codenvycorp.com/browse/CHE-1485

musienko-maxim:
Cannot reproduce

PR #2176, opened 2016-08-21
CHE-1485 allow special characters in file names and folder names + fix an encoding issue

Leward:
### What does this PR do?

It allows user to create files and folders that contains special (non-ASCII) characters expects the one that are not allowed in Linux or Windows. 

[CHE-1485 Struggling working with Japanese filenames](https://github.com/eclipse/che/issues/1706)
### Previous Behavior

File/Folder names with special characters were marked as invalid
### New Behavior

File/Folder names with special characters are now valid
### Tests written?

Yes
### Docs requirements?

None

image

Note that Japanese names for example makes the terminal to behave a bit weirdly. However by getting into the container it is fine (see screenshot).

### DIFF
--- a/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/project/ProjectServiceClientImpl.java
+++ b/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/project/ProjectServiceClientImpl.java
@@ -264,7 +264,7 @@ public class ProjectServiceClientImpl implements ProjectServiceClient {
     /** {@inheritDoc} */
     @Override
     public Promise<Void> copy(Path source, Path target, String newName, boolean overwrite) {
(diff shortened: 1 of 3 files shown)


Leward:
By the way I think all the file names passed as a query string parameter to the API should be URL Encoded. 

I supposed because of that, some operation may be broken when executed against files containing special chars in their names. 

This is also my first contribution to Che  :)

TylerJewell:
Thanks for the contribution!  @kuleshov @vparfonov should review and comment.

slemeur:
@vparfonov : could you please have your team reviewing the PR.

slemeur:
@vparfonov : still no review planned on this PR?

dkulieshov (approved):
Looks nice

PR #2182, opened 2016-08-22
Fix NPE on factory links injection

mkuznyetsov:
If factory is created in resolver method, it's gonna have null creator. Prevent NPE by checking it


akorneta:
ok

skabashnyuk:
ок

Issue #2136, opened 2016-08-16
User password stores/updates without encoding


akorneta:
After refactoring user infrastructure was added new password encryption mechanism, but hasn't been used. So it is needed to apply new encryption mechanism and to reset the old passwords.

JamesDrummond:
@skabashnyuk Provide status label.

PR #2197, opened 2016-08-23
CHE-2136: Add password encryption during user create/update

akorneta:
### What does this PR do?

Provides user's password ecryption during update/create, erases passwords from each user instance that taken from JpaUserDao
#2136
### Previous behavior

users passwords stored plainly without encryption
### Tests written?

Yes

@skabashnyuk, @evoevodin please take a look


voievodin:
LGTM

mkuznyetsov:
ok

PR #2203, opened 2016-08-24
Add che product name to CLI & launcher

TylerJewell:
### What does this PR do?

Adds the ability to set the name of the product that is used in the output for the Che CLI and the Che launcher. This will help in scenarios with ARTIK IDE where we need to rename all of the outputs, container names, and images.


TylerJewell:
lgtm - adding for ARTIK.

PR #2208, opened 2016-08-25
Fix che-in-che stack

mkuznyetsov:
### What does this PR do?

Fixes broken stack for "che-in-che" (https://github.com/eclipse/che/commit/fb088d041dddaf110d2c0c8a41790f98dc3aa163), that would fail Che startup.


skabashnyuk:
ok

Issue #2179, opened 2016-08-22
Revamp list's header


slemeur:
As part of the work done to simplify the list and grips in UD.
We needt improve the way we are displaying description of UD's sections.

Here is the design:
![2 - workspace list header](https://cloud.githubusercontent.com/assets/1636769/17850582/2775e77c-685e-11e6-8d54-d4eb6c127cc5.png)

Labels of each pages:
_Dashboard:_
Access your recently used workspaces and Factories, or create a new workspace. Clicking on your user name at the bottom will open up account and (with sufficient permissions) administration settings.

_Workspaces:_
A workspace is composed of projects, environments and commands. Create a new workspace or manage an existing workspace (start, stop, reconfigure or delete).  Learn more.

Learn more will redirect to: 
https://eclipse-che.readme.io/docs/environments

_Administration:_
Control the configuration and policies for your installation.

ashumilova:
Resolved

PR #2210, opened 2016-08-25
CHE-2179: add description widget

ashumilova:
### What does this PR do?

Adds description widget to workspace list, administation page and dashboard page.

https://github.com/eclipse/che/issues/2179

### DIFF
--- a/dashboard/src/components/widget/description/che-description.styl
+++ b/dashboard/src/components/widget/description/che-description.styl
@@ -1,0 +1,18 @@
+$description-text-color = #6e6e6e
+$description-border-color = $box-border-color
+$description-background-color = #EDEDED
+
+.che-description
+  padding 5px 52px
+  color $description-text-color
+  background-color $description-background-color
+  border-top 1px solid $description-border-color
+  border-bottom 1px solid $description-border-color
+  font-size 9pt
+
+  .che-link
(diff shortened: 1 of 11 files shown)


slemeur:
+1

dashboard/src/components/widget/description/che-description.styl
```diff
@@ -0,0 +1,18 @@
+$description-text-color = #6e6e6e
+$description-border-color = #cbcbcb
+$description-background-color = #EDEDED
```
benoitf, line 3: should colors bring to the top level .styl ?

ashumilova: those are used only here, so decided not to define them in styl.

benoitf: ok seems #CBCBCB color is also used there https://github.com/eclipse/che/blob/a8dfb6bf541940c636590119ef2fb45c1fa659c9/dashboard/src/components/widget/empty-state/che-empty-state.styl#L2

ashumilova: Fixed, added very-light-grey-color
name was taken from here http://www.htmlcsscolor.com/hex/CBCBCB

PR #2212, opened 2016-08-25
CODENVY-515: Add ability to save workspace snapshots in AWS ECR

mmorhun:
### What does this PR do?

Moves logic of creation and pushing workspace snapshots to remote docker registry into separate method.
Adds old snapshots removing.

https://github.com/codenvy/codenvy/issues/515
### PR type
### Minor change checklist
- [ ] New API required?
- [ ] API updated
- [x] Tests passed
### Major change checklist
- [ ] New API required?
- [ ] API updated
- [ ] Tests passed

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
@@ -219,30 +219,47 @@ public class DockerInstance extends AbstractInstance {
                 return new DockerMachineSource(image).withTag(LATEST_TAG);
             }
 
-            PushParams pushParams = PushParams.create(image)
-                                              .withRegistry(registry)
-                                              .withTag(LATEST_TAG);
-
-            final String fullRepo = pushParams.getFullRepo();
-            commitContainer(fullRepo, LATEST_TAG);
-            //TODO fix this workaround. Docker image is not visible after commit when using swarm
-            Thread.sleep(2000);
-            final ProgressLineFormatterImpl lineFormatter = new ProgressLineFormatterImpl();
-            final String digest = docker.push(pushParams,
-                                              progressMonitor -> {
-                                                  try {
-                                                      outputConsumer.writeLine(lineFormatter.format(progressMonitor));
-                                                  } catch (IOException ignored) {
-                                                  }
-                                              });
-            docker.removeImage(RemoveImageParams.create(fullRepo).withForce(false));
-            return new DockerMachineSource(image).withRegistry(registry).withDigest(digest).withTag(LATEST_TAG);
+           return createAndPushSnapshotToRemoteRegistry(image, registry);
         } catch (IOException ioEx) {
             throw new MachineException(ioEx);
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
             throw new MachineException(e.getLocalizedMessage(), e);
         }
+    }
+
+    /**
+     * Creates snapshot of workspace and push it to remote docker registry.
+     * This method has protected access because it can be extended to improve flexibility.
+     *
+     * @param image
+     *         snapshot image name (may include namespace)
+     * @param registry
+     *         registry to which snapshot image should be pushed
+     * @return Docker machine source
+     * @throws IOException
+     * @throws InterruptedException
+     */
+    protected DockerMachineSource createAndPushSnapshotToRemoteRegistry(String image, String registry) throws IOException,
+                                                                                                              InterruptedException {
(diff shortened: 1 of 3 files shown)


mmorhun:
@akorneta @evoevodin take a look please

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
```diff
@@ -245,6 +228,40 @@ public MachineSource saveToSnapshot() throws MachineException {
...
+     */
+    protected DockerMachineSource createAndPushSnapshotToRemoteRegistry(String image, String registry) throws IOException,
+                                                                                                              InterruptedException {
+        PushParams pushParams = PushParams.create(image)
+                                          .withRegistry(registry)
+                                          .withTag(LATEST_TAG);
+
+        final String fullRepo = pushParams.getFullRepo();
+        commitContainer(fullRepo, LATEST_TAG);
+        //TODO fix this workaround. Docker image is not visible after commit when using swarm
```
voievodin, line 251: Can you fix it in this pr?

mmorhun: It is not obviously why we have such behaviour of swarm here. So, it require investigation of ability to fix it. I think that it is separate issue.

akorneta:
ok

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
voievodin: > I think that it is separate issue.

Is there any issue describing this problem ?

mmorhun: It seems, that we do not have issue for this. But it may be fixed by using docker swarm mode instead of swarm.

voievodin: If you think that it should be done by separate issue, can you please create an issue for that?
Because i don't think that such kind of TODO is really ok + `Thread.sleep(2000)` doesn't look okay either

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
```diff
@@ -245,6 +228,40 @@ public MachineSource saveToSnapshot() throws MachineException {
...
+     *
+     * @param image
+     *         snapshot image name (may include namespace)
+     * @param registry
+     *         registry to which snapshot image should be pushed
+     * @return Docker machine source
+     * @throws IOException
+     * @throws InterruptedException
+     */
+    protected DockerMachineSource createAndPushSnapshotToRemoteRegistry(String image, String registry) throws IOException,
```
voievodin, line 243: It looks like this method is designed to be used in couple with [this](https://github.com/codenvy/codenvy/pull/629/files#diff-2264523513ff681f58e2aee07cd710bfR39) solution am i right ?
If so can you please answer my [question](https://github.com/codenvy/codenvy/pull/629/files#r76377299).

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
mmorhun: ok, I'll create issue. But it should be a rule to create issue for each `TODO` by that person who added it.

voievodin: It depends on TODO actually, in this case it looks like a problem and thread stucks for 2 seconds on each snapshot creation, why 2 seconds, why not 1? what is the problem exactly? i have no idea from the code, so yes it deserves an issue. Thank you

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
mmorhun: Yes, you are right, you could make this conclusion from PR headers.

voievodin:
#2273 will cover this functionality

PR #2220, opened 2016-08-25
Changes to support move to che-dockerfiles repository

TylerJewell:
### What does this PR do?

1: Moves dockerfiles\che-server\Dockerfile to the root of the repo.
2: Removes all other dockerfiles - they are now in che-dockerfiles repo


benoitf:
+1

PR #2225, opened 2016-08-26
Fix errors in che-in-che stack

TylerJewell:
1. Adds an SVG file for the che stack to appear in the dashboard
2. Fixes some errors in the stack.

In order to make this work, user will need to:
1. Checkout github.com/eclipse/che-dockerfiles
2. In the recipes/alpine_jdk8 directory:

docker build -t codenvy/alpine_jdk8 .

You can then run Che and the stack will work.

Note - we are still missing improvements to the commands that appear to help automate the configuration of the commands.  Currently, you need to provide the host mount in the command that points to the location of the source code on the host in order to be able to build it.


riuvshin:
LGTM

Issue #2151, opened 2016-08-17
Consider Adding Paramaterized Custom Commands


rickspencer3:
**Background**
Some development environments receive commands, especially for code generation. For example, in Rails, in order to generate a model, you use this command:

rails generate model NAME [field[:type][:index] field[:type][:index]] [options]

In the case "NAME" is a required field. This command modifies multiple files. There are many such commands in Rails, which are necessary to use Rails effectively.

Another example is generating a project with Revel, where the user should typically specify a path that includes a repository location and a new application name, generally using their github account name to set up a typical Go workspace. 

**The Request:**
We would like to be able to make our development containers easy to use in Che by wrapping obscure or difficult to discover and use commands in Che commands that appear in the IDE. However, the IDE does not currently support passing arguments to commands

For example, we could provide JSON such as:

```
    "commands": [
      {
        "name": "Create Project",
        "prompt": "Provide your github username and a name for your new project",
        "inputs", [{"text":"Github Id","name":"ghid"}, {"text":"Project Name","name":"proj"}]
        "type": "custom",
        "commandLine": "revel new {{ghid}}/{{proj}} && revel run {{ghid}}/{{proj}}",
        "attributes": {}
      }]
```

When run the IDE could pop open a dialog that collects the user input and runs the command (which in this specific case would create and run a properly named revel project). Some thought would need to be put into handling the user canceling, and also error handling.

We could make development frameworks much easier to learn with such functionality.

slemeur:
@vzhukovskii + @ashumilova : We need to study that enhancement request with the work done along #953

TylerJewell:
+1 - I think to be clear the only input type that we would accept would be text. I don't think there is any need for numbers, booleans or other types.

skabashnyuk:
Not relevant anymore. Please consider opening a new issue if you think it is up to date.

PR #2232, opened 2016-08-26
#953: Extend macro list

vzhukovs:
This PR extends existed list of macros, by adding new ones.

Related issue: https://github.com/eclipse/che/issues/953

Created new macros to provide various information from basic part.

It includes:

```
${editor.current.file.name} -- Currently selected file in editor with context
${editor.current.file.path}  -- Absolute path to the selected file in editor with context
${editor.current.file.relpath} -- Path relative to the /projects folder in editor with context
${editor.current.project.name} -- Project name of the file currently selected in editor with context
${editor.current.project.type} -- Project type of the file currently selected in editor with context
${explorer.current.file.name} -- Currently selected file in project tree
${explorer.current.file.path}  -- Absolute path to the selected file in project tree
${explorer.current.file.relpath} -- Path relative to the /projects folder in project tree
${explorer.current.project.name} -- Project name of the file currently selected in explorer
${explorer.current.project.type} -- Project type of the file currently selected in explorer
${server.<name>} -- Returns protocol, hostname and port of an internal server
${server.<name>.protocol} -- Returns protocol of a server registered by name
${server.<name>.hostname} -- Returns hostname of a server registered by name
${server.<name>.port} -- Returns port of a server registered by name
${workspace.name} -- Returns the name of the workspace
```

@vparfonov @ashumilova @azatsarynnyy review it, please.

### DIFF
--- a/ide/che-core-ide-app/src/test/java/org/eclipse/che/ide/editor/macro/AbstractEditorMacroProviderTest.java
+++ b/ide/che-core-ide-app/src/test/java/org/eclipse/che/ide/editor/macro/AbstractEditorMacroProviderTest.java
@@ -1,0 +1,72 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.ide.editor.macro;
+
+import com.google.common.base.Optional;
+
+import org.eclipse.che.api.promises.client.PromiseProvider;
+import org.eclipse.che.ide.api.app.AppContext;
+import org.eclipse.che.ide.api.editor.EditorAgent;
+import org.eclipse.che.ide.api.editor.EditorInput;
+import org.eclipse.che.ide.api.editor.EditorPartPresenter;
+import org.eclipse.che.ide.api.resources.File;
+import org.eclipse.che.ide.api.resources.Project;
+import org.eclipse.che.ide.resource.Path;
+import org.mockito.Mock;
+
+import static org.mockito.Mockito.when;
+
+/**
+ * @author Vlad Zhukovskyi
+ */
+public abstract class AbstractEditorMacroProviderTest {
+
+    public static final String FILE_NAME     = "name";
+    public static final String FILE_PATH     = "/project/name";
+    public static final String PROJECTS_ROOT = "/projects";
+    public static final String PROJECT_NAME  = "project-name";
+    public static final String PROJECT_TYPE  = "type";
+
+    @Mock
+    protected EditorAgent editorAgent;
+
+    @Mock
+    protected PromiseProvider promiseProvider;
+
+    @Mock
+    protected EditorPartPresenter activeEditor;
+
+    @Mock
+    protected File activeFile;
+
+    @Mock
+    protected EditorInput activeEditorInput;
+
+    @Mock
+    protected AppContext appContext;
+
+    @Mock
+    protected Project project;
+
+    protected abstract AbstractEditorMacroProvider getProvider();
+
+    protected void initEditorWithTestFile() {
+        when(editorAgent.getActiveEditor()).thenReturn(activeEditor);
+        when(activeEditor.getEditorInput()).thenReturn(activeEditorInput);
+        when(activeEditorInput.getFile()).thenReturn(activeFile);
+        when(activeFile.getName()).thenReturn(FILE_NAME);
+        when(activeFile.getLocation()).thenReturn(Path.valueOf(FILE_PATH));
+        when(appContext.getProjectsRoot()).thenReturn(Path.valueOf(PROJECTS_ROOT));
+        when(activeFile.getRelatedProject()).thenReturn(Optional.of(project));
+        when(project.getName()).thenReturn(PROJECT_NAME);
+        when(project.getType()).thenReturn(PROJECT_TYPE);
+    }
+}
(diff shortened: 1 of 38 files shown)


benoitf:
IMHO 
`${server.*}` and `${workspace.*}` should be handled server side

vzhukovs:
We can resolve this information on the client and also this is the only one way to do this at the moment.

vzhukovs:
Later, when we'll have mechanism to resolve it on server side, we'll move it. Maybe.

benoitf:
I don't see why it is the only one way as you're bringing new code to handle it, this code could be added on server instead of client

ashumilova:
+1 
@benoitf for the server side command macros - will discuss how better to implement it and will create an issue.

vparfonov:
I think we have separate issue on your idea @benoitf https://github.com/eclipse/che/issues/2151

benoitf:
@vparfonov yes but on this issue it's more how to bring custom resolver by prompting the user
my concern is that a command definition inside a workspace is shared by all clients and then some properties that are only related at the server side (like mapping of ports, workspace info, etc) could be resolved for all clients by the server (and not having each client solving the properties on its own as at the end it may have different result)

tolusha:
@vparfonov @vzhukovskii 
Is there any way to show the whole list of macros for user ?

ide/che-core-ide-app/src/test/java/org/eclipse/che/ide/part/explorer/project/macro/AbstractExplorerMacroProviderTest.java
```diff
@@ -0,0 +1,109 @@
...
+import org.eclipse.che.ide.ui.smartTree.SelectionModel;
+import org.eclipse.che.ide.ui.smartTree.Tree;
+import org.junit.Before;
+import org.mockito.Mock;
+
+import java.util.Collections;
+
+import static org.mockito.Mockito.when;
+
+public class AbstractExplorerMacroProviderTest {
```
azatsarynnyy, line 36: javadoc with author is needed here

vzhukovs:
@tolusha at this moment - no. But we can do the same as for key binding has done. Dedicate window with table of registered bindings. For example.

ide/che-core-ide-app/src/test/java/org/eclipse/che/ide/part/explorer/project/macro/AbstractExplorerMacroProviderTest.java
vzhukovs: Fixed

ide/che-core-ide-app/src/test/java/org/eclipse/che/ide/editor/macro/AbstractEditorMacroProviderTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.eclipse.che.ide.api.editor.EditorInput;
+import org.eclipse.che.ide.api.editor.EditorPartPresenter;
+import org.eclipse.che.ide.api.resources.File;
+import org.eclipse.che.ide.api.resources.Project;
+import org.eclipse.che.ide.resource.Path;
+import org.mockito.Mock;
+
+import static org.mockito.Mockito.when;
+
+public abstract class AbstractEditorMacroProviderTest {
```
azatsarynnyy, line 30: javadoc with author is needed here

vzhukovs: Fixed

azatsarynnyy:
+1 but can't understand what does it mean - **macro provider**

vzhukovs:
Provider for macros

tolusha:
@vzhukovskii Could you pls create the issue, otherwise we will forget it ^)

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/explorer/project/macro/AbstractExplorerMacroProvider.java
```diff
@@ -0,0 +1,126 @@
...
+ * @see CommandPropertyValueProvider
+ * @see ExplorerCurrentFileNameProvider
+ * @see ExplorerCurrentFilePathProvider
+ * @see ExplorerCurrentFileRelativePathProvider
+ * @see ExplorerCurrentProjectNameProvider
+ * @see ExplorerCurrentProjectTypeProvider
+ * @since 4.7.0
+ */
+@Beta
+public abstract class AbstractExplorerMacroProvider implements CommandPropertyValueProvider {
```
vparfonov: not sure we are rely need this abstraction

vparfonov:
not sure about Abstract*Provider's but other ok, 

need ok from @benoitf

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/explorer/project/macro/AbstractExplorerMacroProvider.java
vzhukovs: Removed

TylerJewell:
@vparfonov @vzhukovskii - question for you.  Is there any way we could add one more macro.  It would be `${property.<name>}` where `<name>` is the name of a property from che.properties file?  This would allow for scenarios where a user could look up the value of a property that is in the standard list OR they could provide a custom property in `che.properties`.

For example:

```
# che.properties
my.custom.property=hamburgers

# In the macro:
${property.my.custom.property}
```

If this is something that requires server-side behavior, then I understand - let's add it into the issue tracking the movement of macros into the server.

vzhukovs:
@TylerJewell it requires server side modification. I think, it would be better to do this with the issue tracking the movement of macros into the server.

JamesDrummond:
@TylerJewell Custom properties will help bring in useful information about the host. Helpful for docker commands run in the host if it is possible.

m1L31s:
Are those macros working? 

> ${server.<name>.protocol} -- Returns protocol of a server registered by name
> ${server.<name>.hostname} -- Returns hostname of a server registered by name
> ${server.<name>.port} -- Returns port of a server registered by name

vzhukovskii:
Hi @m1L31s! At this moment only `${server.<servername>}` only work. Those three macros have already unsupported.

PR #2239, opened 2016-08-27
Let Che CLI compile Che

TylerJewell:
### What does this PR do?

1: Adds `che compile <mvn command>` syntax. Will run any mvn command in a Che directory or sub-module. It uses the codenvy/che-dev container to perform the build command.  This is useful for any users that do not have maven or Java setup properly on their computer but want to test doing a build.  The performance is about 4x slower than doing it native - but it's more convenient.

2: Added in ability to convert any environment variable with the format of CHE_PROPERTY_<name>=<value> into a value that will be inserted into a temporary `che.properties` file that will be used for the Che launcher, Che mount, Che dir, Che dev utilities.  If the user provides a CHE_CONF_FOLDER value, then that directory will be used and any CHE_PROPERTY_ values will be ignored.  Variables need to include a double underscore `__` to be interpreted as a single underscore `_` in the property name.  Variables need to provide a single underscore "_" to be interpreted as a period `.`.  For example:

```
# This environment variable:
export CHE_PROPERTY_machine_ws__agent_max__start__time__ms=1000

# Is placed into the temporary che.properties as:
machine.ws_agent.max_start_time_ms=1000
```

3: Refactorings of the Che CLI to make layering in different portions of the docker run syntax so that it's more extensible and reusable by different types of containers.

### DIFF
--- a/che.sh
+++ b/che.sh
@@ -56,6 +56,7 @@ init_global_variables() {
   DEFAULT_CHE_SERVER_CONTAINER_NAME="che-server"
   DEFAULT_CHE_VERSION="latest"
   DEFAULT_CHE_CLI_ACTION="help"
+  DEFAULT_IS_INTERACTIVE="true"
 
   CHE_PRODUCT_NAME=${CHE_PRODUCT_NAME:-${DEFAULT_CHE_PRODUCT_NAME}}
   CHE_MINI_PRODUCT_NAME=${CHE_MINI_PRODUCT_NAME:-${DEFAULT_CHE_MINI_PRODUCT_NAME}}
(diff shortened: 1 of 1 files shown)


skabashnyuk:
che compile will work only for che repository? How about upstream repositories like: che-lib, che-parent, che-dependencies. Will I be able to compile them too?

TylerJewell:
Needs to be tested.  However, the original intent is just the Che repository itself.  If anything is missing it would need to be added into the che-dev image to be available for the container to do its job.

riuvshin:
+1

bmicklea:
@TylerJewell please add a milestone.

PR #2240, opened 2016-08-28
Fix Processes panel behavior

azatsarynnyy:
Fix the behavior of the Processes panel in case it was split on several sub-panels.

@vparfonov


vparfonov:
ok

bmicklea:
@azatsarynnyy please add a milestone.

PR #2257, opened 2016-08-29
fixup - propagate Che server image

TylerJewell:
### What does this PR do?

Propagates values that are needed by the che-launcher as environment variables. They were passed previously as part of a profile, but not passed if no profile set.

Also - adds in additional IS_PSEUDO_TTY variable which will add `-t` to docker run commands if `true`, which is the default.  Set to `false` when using Jenkins or other automated systems like Vagrant.


TylerJewell:
@benoitf

Issue #108, opened 2015-12-07
Data is lost when I restart my che in docker, how to prevent it?


freewind:
I install eclipse-che via docker, following this article: http://eclipse-che.readme.io/docs/install-using-docker, and I start che as:

```
docker run --privileged -it -p 8080:8080 -p 49152-49162:49152-49162 codenvy/che
```

The problem is when I restart my che, all data (including created projects/downloaded dependencies).

How to prevent it? If I need to mount something when I start the docker to store the data outside the docker, which directories do I need to care?

ghost:
Take a look at https://eclipse-che.readme.io/v1.0/docs/install-using-docker#configuration

You need to mount:

-v {local-dir}:/home/user/che/temp/fs-root

freewind:
@eivantsov Shamed that I didn't see that :(

freewind:
I followed the document and start che like this:

```
docker run --privileged -it -p 8080:8080 \
               -p 49152-49162:49152-49162 \
               -v /Users/twer/.m2:/home/user/.m2 \
               -v /Users/twer/che-local/che-conf:/home/user/.che \
               -v /Users/twer/che-local/che-projects:/home/user/temp/fs-root \
               codenvy/che
```

Note: `/Users/twer/che-local` is already `chmod -R 757`

But the problem is after restarting, all data is still lost :(

And there is no new files created in `/Users/twer/che-local`.

Could you help to point it out where is wrong?

ghost:
-v $(pwd):/home/user/che-3.13.4.2/temp/fs-root - missed che base dir in the path. This way it should work. Updated docs.

-v /Users/twer/che-local/che-conf:/home/user/.che - do you have a .properties file locally? If not, the one that is in ~/.che will disappear. You need to mount .che only if you have your own configuration file with properties that you want to override. 

Just to keep projects -v $(pwd):/home/user/che-3.13.4.2/temp/fs-root is OK

freewind:
@eivatsov thanks!

PR #2261, opened 2016-08-30
Artik #108 Cannot open the terminal of ARTIK board, after connect multiple boards

vitaliy-guliy:
Fixes position of buttons in consoles panel


plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/processes/panel/ProcessesPanelPresenter.java
```diff
@@ -711,6 +711,14 @@ private void restoreState(final Machine machine) {
             @Override
             public void apply(List<MachineProcessDto> arg) throws OperationException {
                 for (MachineProcessDto machineProcessDto : arg) {
+                    /**
+                     * Do not show the process if the command line has prefix #hidden
+                     */
+                    if (machineProcessDto.getCommandLine() != null && !machineProcessDto.getCommandLine().isEmpty()
```
vparfonov: Please, use here Strings.isNullOrEmpty() (from Guava library)

vitaliy-guliy: Done

vparfonov:
thank you very much

vparfonov:
ok

PR #2264, opened 2016-08-30
Allow mounting of multiple volumes to mount properties

benoitf:
### What does this PR do?

allow to specify multiples volumes to mount for all mount properties
including`che.properties` `machine.server.extra.volume` property
### New behavior

Volumes names can be separated by semi colon character ;
machine.server.extra.volume=/tmp/foo:/tmp/foo;/tmp/dummy:/tmp.dummy
### PR type
### Minor change checklist
- [ ] New API required? no
- [ ] API updated
- [x] Tests passed

Change-Id: Ic029b47511f784d41e906e0c2b570869466f605e

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceProvider.java
@@ -74,6 +74,7 @@ import java.net.SocketTimeoutException;
 import java.net.URL;
 import java.nio.file.Files;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


TylerJewell:
+1 - this is a huge improvement.  Opens up many use cases for adoption.

benoitf:
about separator, currently it's tab character.

I could replace it with another character but I've seen in mount syntax that
: ; , # were used

so if someone has an idea of a character never used inside mount syntax ?
or if we replace by more than one character

TylerJewell:
I did a search and ext4 file systems allow anything but null and forward-slash `/` in their names.  So using `\t` is as good as anything.  So I either propose a tab or a colon ":" as the separator.  I think colon is easier to read.

garagatyi:
Colon is used for setting read-write access to volumes. I suggest semicolon. Tab is not so easy to recognize because it can be confused with several spaces.

benoitf:
@garagatyi as I said in my initial comment, semicolon is used for some syntax

https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man8/mount_smbfs.8.html
(after domain)

```
mount_smbfs [-N] [-o options] [-d mode] [-f mode] [-h] //[domain;][user[:password]@]server[/share] path
```

benoitf:
what is the status ? I keep tab, I use semi colon ? I merge ?

TylerJewell:
Lets gow ith semicolon.

PR #2269, opened 2016-08-30
CLI: Fix Mac tmp dir access

TylerJewell:
### What does this PR do?

If you install the CLI into /usr/local/bin, which is a directory that requires root access to create & write files, the CLI will fail unless you run it as root. This change writes all temporary files to $PWD, which is where the current user should have write access to.

This allows users to install the CLI into a directory that has a path and use it as another user.


bmicklea:
+1

PR #2288, opened 2016-08-31
backport PR2264 into CHE-1818 branch

benoitf:
### What does this PR do?

Backport changes of PR 2264 into CHE-1818 workspace env refactoring

Without this, change in master will be lost
### PR type

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ComposeMachineProviderImpl.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ComposeMachineProviderImpl.java
@@ -67,8 +67,10 @@ import javax.inject.Named;
 import java.io.IOException;
 import java.net.SocketTimeoutException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


benoitf:
@garagatyi

garagatyi:
I will merge CHE-1818 with squash. Are you ok if I'll squash theses changes too?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ComposeMachineProviderImpl.java
```diff
@@ -162,6 +164,11 @@ public ComposeMachineProviderImpl(DockerConnector docker,
 
         allMachinesSystemVolumes = removeEmptyAndNullValues(allMachinesSystemVolumes);
         devMachineSystemVolumes = removeEmptyAndNullValues(devMachineSystemVolumes);
+
+        Set<String> volumes = new HashSet<>();
+        devMachineSystemVolumes.forEach((volume) -> Arrays.asList(volume.split(";")).stream().forEach((entry) -> volumes.add(entry)));
```
garagatyi: Consider usage of flatMap instead. I think that this operation can be clearer in that case.

skabashnyuk: what if?

```
        devMachineSystemVolumes = devMachineSystemVolumes.stream()
                                                         .map(line -> line.split(";"))
                                                         .flatMap(Arrays::stream)
                                                         .distinct()
                                                         .collect(Collectors.toSet());
```

benoitf:
@garagatyi and @skabashnyuk I'm ok to squash everything or not use this PR at all as long as the change about handling multiple volumes with semi colon is handled.

@skabashnyuk as your solution is being more elegant, I've updated PR.

garagatyi:
AFAIK now it is possible to change base branch of PR. So I think we can wait until CHE-1818 is merged and then change base branch of this PR to master and merge it to master. @benoitf WDYT?

benoitf:
@garagatyi well current master is having the feature
so technically, merging it without merging first this into CHE-1818 will result in a regression
This is why I wanted to have it merged in what will be applied on master

tolusha:
Why not to do the same for `allMachinesSystemVolumes` ?

benoitf:
@tolusha I was not sure user was providing something for this variable

tolusha:
It will be usefull for https://github.com/codenvy/codenvy/issues/499

benoitf:
ok added

PR #2289, opened 2016-08-31
Add support for Che in Che with a Chefile

benoitf:
### What does this PR do?

Allow to boot a workspace with che after cloning repository locally
and compile and run che inside Che
### PR type
### Minor change checklist
- [x] Tests passed

Change-Id: Id94052b580a66296880e207d6fda88758b1e9be6

### DIFF
--- a/Chefile
+++ b/Chefile
@@ -1,20 +1,36 @@
-workspace.runtime.image.location="codenvy/alpine_jdk8" 
-workspace.commands[0].name="2. Build Che in Che"
+# Runtime image to use for booting and compiling Eclipse Che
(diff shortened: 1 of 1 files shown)


TylerJewell:
Cool that we have gotten this far.  I think the commands will need to be improved and will not work out of the box.  The commands need to reference the CHE_HOME, which is a value that a user has to type directly as the command needs to know the absolute path of where the Che server binary is on the host itself.

I will need some time to play with this - but did you try this Chefile by cloning che into different directories?

benoitf:
yes it works as we ask docker to know the host path of the volume mount (for this usecase)

TylerJewell:
+1 - ready to merge

PR #2305, opened 2016-09-01
CODENVY-645; add cascading removal for permissions;

mshaposhnik:
### What does this PR do?

Adds entity listeners and events on removing recipes and stacks, to be able to remove permissions before/

CODENVY-645
### Minor change checklist
- [NO ] New API required?
- [ No] API updated
- [ No] Tests provided / updated
- [YES] Tests passed


skabashnyuk:
ok. @evoevodin ?

PR #2306, opened 2016-09-01
fixup to che.sh

TylerJewell:
### What does this PR do?
1. Fixes issue where if there are no CHE_ environment variables then important CHE_ variables are not passed along to children processes.
2. Removed a duplicate comment.
3. Converts the shell to use bash instead of posix shell.  Tested on Mac, Windows, and Linux.


riuvshin:
LGTM

PR #2307, opened 2016-09-02
Separate CHE_VERSION from CHE_UTILITY_VERSION

TylerJewell:
MUST BE MERGED WITH: https://github.com/eclipse/che-dockerfiles/pull/8
1: Creates a separate configuration for the version of the Che server `CHE_VERSION` which is passed along to the Che launcher apart from the version used for Che utilities `CHE_UTILITY_VERSION`. This allows utilities to be versioned and released differently from the Che server. These were hard coupled before.  

2: Prints out the name of the profile that is loaded, if one is loaded at the beginning, giving users a heads up that their execution is governed by a profile.

3: Fixes an issue where the networking connectivity check was hard coded the Che server image to be used.

4: Adds a `--force` option to the `che update` command which, if provided, will delete the image before doing a docker pull. Solves some issues with caching.

5: Adds `CHE_CLI_DEBUG` option. If true, will print out stack trace as execution goes through.

6: Adds `CHE_CLI_INFO` flag, which if disabled, will not print any INFO messages. This flag is packaged in the environment package sent to utilities.

### DIFF
--- a/che.sh
+++ b/che.sh
@@ -14,6 +14,15 @@ init_logging() {
   GREEN='\033[0;32m'
   RED='\033[0;31m'
   NC='\033[0m'
+
+  # Turns on stack trace
+  DEFAULT_CHE_CLI_DEBUG="false"
+  CHE_CLI_DEBUG=${CHE_CLI_DEBUG:-${DEFAULT_CHE_CLI_DEBUG}}
+
+  # Activates console output
+  DEFAULT_CHE_CLI_INFO="true"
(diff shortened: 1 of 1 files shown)


TylerJewell:
@eivantsov @riuvshin @bmicklea - good additions to the CLI to make it more robust.

benoitf:
about 2. I would like to have a way to disable this kind of traces so if we need command output without extra lines at first, we're sure we get them.
or maybe add profile info into the prefix of log output

INFO : che(profilename) : ...

riuvshin:
LGTM

PR #2315, opened 2016-09-02
CHE-1818: Add new WS environment types, machines attributes

garagatyi:
This PR adds 
- attributes map to machines in environment.
- new environment types `dockerimage` and `dockerfile` with single machine

Attribute `memoryLimitBytes` sets RAM limit for machine. If RAM limit is set in the environment recipe this property overrides limit.


skabashnyuk:
ok

benoitf:
how can you merge PR within minutes and not PR that are here since a couple of days for the same branch
https://github.com/eclipse/che/pull/2288 ?

garagatyi:
They are both merged

benoitf:
thx @garagatyi

Issue #1820, opened 2016-07-18
Make Che with multi-machine environments backward compatible


garagatyi:
#1818 Should introduce new format of environment definition.

We should support usage of workspaces in old format to provide backward compatibility

PR #2317, opened 2016-09-02
Add support for old format of Factory/Stack/Workspace

voievodin:
### What does this PR do?

Adds support for old format of Factories/Stacks/Workspaces.
Adds on fly migration for local data.
It is supposed that all the REST API works with old model objects.
#1820

@garagatyi, @skabashnyuk please review


garagatyi:
LGTM

skabashnyuk:
OK

Issue #2343, opened 2016-09-06
Make username validation to use account name validation


sleshchenko:
In state of [jpa-integration](https://github.com/eclipse/che/tree/jpa-integration) branch user doesn't have own name and uses name of his [personal account](https://github.com/eclipse/che/blob/jpa-integration/wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/model/impl/UserImpl.java#L140).
It is need to rework username validation to use common account name validation

PR #1442, opened 2016-06-06
CODENVY-619 Restrict usernames to letters and digits only

mkuznyetsov:
Users with names, that are not URL-friendly must not be created

PR #2346, opened 2016-09-06
CHE-2343 Make username validation to use account name validation

sleshchenko:
### What does this PR do?

Extract account name validation from username validation to be able to use it for validation of names of other kinds accounts

https://github.com/eclipse/che/issues/2343
### PR type


sleshchenko:
@skabashnyuk @akorneta @evoevodin Please take a look my changes

wsmaster/che-core-api-account/src/main/java/org/eclipse/che/account/spi/AccountValidator.java
```diff
@@ -0,0 +1,86 @@
...
+ * Utils for account validation.
+ *
+ * @author Sergii Leschenko
+ */
+// TODO extract normalization code from the validator as it is not related to the validation at all
+public class AccountValidator {
+    private static final Logger LOG = LoggerFactory.getLogger(AccountValidator.class);
+
+    private static final Pattern ILLEGAL_ACCOUNT_NAME_CHARACTERS = Pattern.compile("[^a-zA-Z0-9]");
+    private static final Pattern VALID_ACCOUNT_NAME              = Pattern.compile("^[a-zA-Z0-9]*");
```
voievodin, line 35: Are you sure that underscore is not valid character?

sleshchenko: I moved this logic from UserValidator. @mkuznyetsov Can you give answer?

wsmaster/che-core-api-account/src/main/java/org/eclipse/che/account/spi/AccountValidator.java
```diff
@@ -0,0 +1,86 @@
...
+import javax.inject.Inject;
+import java.util.regex.Pattern;
+
+/**
+ * Utils for account validation.
+ *
+ * @author Sergii Leschenko
+ */
+// TODO extract normalization code from the validator as it is not related to the validation at all
+public class AccountValidator {
```
voievodin, line 31: Does it have to be a singleton?

voievodin:
ok

wsmaster/che-core-api-account/src/main/java/org/eclipse/che/account/spi/AccountValidator.java
mkuznyetsov: Underscore was not intended to be a valid character for user name.

sleshchenko: https://github.com/eclipse/che/pull/1442#discussion_r66219481

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
```diff
@@ -106,7 +104,10 @@ public Response create(@ApiParam("New user")
                                                        UnauthorizedException,
                                                        ConflictException,
                                                        ServerException {
-        userDto.withId(generate("user", ID_LENGTH));
+        if (userDto != null) {
+            //should be generated by userManager
```
skabashnyuk, line 108: may be if `userDto == null` or `userDto.getId!= null` we should throw exception instead of silently set `null`?

skabashnyuk:
other ok

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserService.java
sleshchenko: No. We always should set id to null because user should not be able to set id during creation and id should be generated by user manager.
Case when userDto is null then user entity should be fetched by token.
So we should not throw exception here

Issue #2291, opened 2016-08-31
Node console template not showing up


bmicklea:
When you select the default Node stack only the node web app template appears. However, there's a node console template that you can see in the full list of templates. (perhaps this is a tagging issue?)

**Reproduction Steps:**
1. Go to the dashboard page for creating a new workspace + project
2. In the Workspace section, select the Node stack
3. In the Templates section, notice only the node web app template is the only one there
4. Go back to the Workspace section and select Custom Stack
5. In the Templates section you'll now see both a web and console app template for node.

**Expected behavior:**
Both web and console templates should be shown for the Node app

**Observed behavior:**
Only the web app template is shown.

**Che version:**    4.6+
**OS and version:** Mac OS 10.11.6
**Che install:**   Docker launcher + beta.codenvy.com (4.7)

ghost:
https://github.com/eclipse/che/pull/2376

Fixed.

Issue #2290, opened 2016-08-31
Fix tag on default template "web-javaee-jaxrs"


bmicklea:
The web-javaee-jaxrs template appears when you select the default Java stack, however, this template actually requires the "TomEE" stack in order to build and run properly.

**Reproduction Steps:**
1. Select Java stack
2. Select web-javaee-jaxrs template
3. Create workspace
4. Try and build app and you'll get an error because $TOMEE_HOME isn't defined because it's not part of this stack.

**Expected behavior:**
I only see templates that work with my stack so this template should only appear if I select the TomEE stack.

**Che version:**    4.6+
**Che install:**   verified on che 4.7, codenvy on-prem 4.6.2 and beta.codenvy.com 4.7

bmicklea:
@ddementieva I spoke to Eugene about this and he said you could retag this template so that it will only show up with the TomEE stack.

ghost:
https://github.com/eclipse/che/pull/2376
 Closing it.

Issue #2316, opened 2016-09-02
Advanced stack configuration form


slemeur:
_Goal:_
- Provide the ability to author a stack with advanced configuration options.
- Handle authoring for multi-containers based stack.
- Provide the ability to test the stack while authoring it.

_Preliminary:_
The scope of this issue is pretty large and will be handled step by step. It is the natural extension of the basic flow that we started with the issue #2207.

_Advanced Form_
The form is going to be extended with more and more parameters that are handled by the Environment Model.
This form will be used similarly to edit a stack.

![6 - stacks - create new](https://cloud.githubusercontent.com/assets/1636769/18203661/a124f6e0-7118-11e6-88a8-8a7bceba846e.png)

1- User provides the name for the stack

2- User can provide the source of the stack. He can either:
- customize a stack already registered in the system
- provide a dockerfile
- provide a composefile (for multi-container)
- start from a blank template.

3- User configures the machines composing the stack's environment. It can be single machine or multi-machine. The list of machines is collapsible/expendable. By expending a machine, its details and configuration options are displayed, 
Each machine have:
- a name
- RAM
- can be defined as dev-machine (meaning it will be the primary machine where Che will injects agents for IDE tooling) (it is a toggle, activating the dev-machine in one machine of the environment will toggle-off in the other machines)
- a list of agents (if selected, they'll be injected in the machine)
- configuration for environment variables
- configuration for ports
- a list of commands

4- User provides meta-datas for the stack:
- the category in which the stack will be displayed to other
- the list of components included in the stack
- the list of tags (which are used to filter the project templates)

5- User is able to see the raw JSON configuration of the stack.

_Test the Stack_
Once the user is done with the configuration of the stack, we provide the ability to test the stack with a temporary workspace, which will be used only for this purpose.
While editing the stack configuration, two buttons are displayed to the user to "test" the stack. Those buttons are displayed in the header and in the footer overlay.

![6 - stacks - testing - step 1](https://cloud.githubusercontent.com/assets/1636769/18206231/fb3327f8-7125-11e6-8958-28593b18a5a7.png)

When the user hits the "test" button, we show a popup where he can select the templates that are matching the tags he provided to the stack. This step is optional, user is not forced to inject project templates in the temporary workspace, he could also import projects from git, subversion sources.

![6 - stacks - testing - step 2](https://cloud.githubusercontent.com/assets/1636769/18206233/00a3c22e-7126-11e6-98c0-467aa0d4c1a5.png)

[...]

PR #2376, opened 2016-09-07
Add Petclinic+MySQL stack, petclinic mysql sample project. Fix dashbo…

ghost:
### What does this PR do?

This PR adds a new stack with a sample app - Petclinic + MySQL in 2 containers

https://github.com/eclipse/che/issues/2290
https://github.com/eclipse/che/issues/2291
### Previous behavior

### New behavior

A new stack is a demo of the new compose syntax.
### PR type


riuvshin:
ci-build

ghost:
@slemeur @TylerJewell the new stack and template is there. I placed it on #1 position

slemeur:
Making it #1 is a good idea - but I'm not sure we should make it the default one. Default should remain the same "Java". 
Is this imply to change the position of the stack, then I'd make it #2 after "Java" and before "Blank"

ghost:
@slemeur the first one means the default in UD.

TylerJewell:
@eivantsov - do you have a milestone that you really want this in?  ie, are you hoping it is M1 or can it wait until M2? Your answer will influence how aggressive I try to get in evaluating this sooner vs. later.

This is exciting to see.

ide/che-core-ide-stacks/src/main/resources/stacks.json
```diff
@@ -1,5 +1,75 @@
...
   {
+    "id": "java-mysql",
+    "creator": "ide",
+    "name": "Java-MySQL",
+    "description": "Multi-machine environment with Default Java Stack and MySQL database",
+    "scope": "general",
+    "tags": [
+      "Java 1.8, Tomcat 8, MySQL 5.7"
+    ],
+    "components": [
```
TylerJewell, line 11: One of the suggestions that I made to Stevan was that we should probably be more complete on the list of components that we have and their versions. We should list all of the libraries that we are installing into the Ubuntu system - so bash, openssl, etc.  Anything that we are explicitly installing with a RUN command should probably be added here.

We also need the mysql reference, and I think the JDK / maven versions are out of date?

ide/che-core-ide-stacks/src/main/resources/stacks.json
```diff
@@ -1,5 +1,75 @@
...
+    ],
+    "source": {
+      "type": "image",
+      "origin": "codenvy/ubuntu_jdk8"
+    },
+    "workspaceConfig": {
+      "environments": {
+        "default": {
+          "machines": {
+            "dev-machine": {
```
TylerJewell, line 33: Is this the most user friendly name we can give it?  Is this the name that shows up in the machine drop down?  I think that this name is super clear for us, but perhaps it's not as clear as it could be for new users experiencing the system the first time.

ide/che-core-ide-templates/src/main/resources/samples.json
```diff
@@ -1,5 +1,74 @@
...
+      "location": "https://github.com/che-samples/web-java-spring-petclinic.git",
+      "parameters": {
+        "branch":"mysql"
+      }
+    },
+    "commands": [
+      {
+        "name": "build",
+        "type": "mvn",
+        "commandLine": "mvn -f ${current.project.path} clean install -DskipTests && cp ${current.project.path}/target/*.war $TOMCAT_HOME/webapps/ROOT.war",
```
TylerJewell: I think we should have a `build` command, a `deploy` command, and then separately a `build and deploy` command, which is this syntax.  I am not sure it's intuitive for users that build, automatically does a deploy - so we should provide all the variants.

ide/che-core-ide-stacks/src/main/resources/stacks.json
ghost: These are tags, smth that will show up in user dashboard. We use tags to match stacks and samples.

So, if I like versions of everything, we just do not have space in the stack box on UD.

Also, this will break all the filtering.

This is a UD issue. Perhaps, we should use components to display in UD but use tags that can be super unique but not super readable to match samples.

@ashumilova can comment - we talked about it.

ide/che-core-ide-stacks/src/main/resources/stacks.json
ghost: `dev-machine` is the only value that dashboard accepts. @ashumilova has a fix

ide/che-core-ide-templates/src/main/resources/samples.json
ghost: ok, makes sense

ghost:
@TylerJewell I do not have any preferences on when this should be released.

ide/che-core-ide-stacks/src/main/resources/stacks.json
```diff
@@ -1,5 +1,75 @@
...
+              "agents": [
+                "ws-agent"
+              ],
+              "servers": {},
+              "attributes" : {}
+            }
+          },
+          "recipe": {
+            "content": "services:\n db:\n  image: codenvy/mysql\n  environment:\n   MYSQL_DATABASE: petclinic\n   MYSQL_USER: petclinic\n   MYSQL_PASSWORD: password\n  mem_limit: 2147483648\n dev-machine:\n  image: codenvy/ubuntu_jdk8\n  mem_limit: 2147483648\n  depends_on:\n    - db\n  links:\n    - db",
```
garagatyi: Why you used both depends_on and links?

ashumilova:
We can change the name of the dev-machine, how to name it in stacks?
About tags - would be nice to display components and tags to be for binding only. In the list of stacks - components will be displayed, maybe need the same for stack boxes on creation, @slemeur wdyt?

ide/che-core-ide-stacks/src/main/resources/stacks.json
evoevodin: The name of machine doesn't have to be `dev-machine` one thing which makes it `dev` is `ws-agent` in agents list.

slemeur:
@ashumilova : Yes agree. it is planned for the advanced stack form https://github.com/eclipse/che/issues/2316

ghost:
ci-build

ide/che-core-ide-stacks/src/main/resources/stacks.json
TylerJewell: @eivantsov - I think there is a difference between components and tags in this document. The tags are used for filtering and they are a different list of labels. But I believe the components are the list that appears in the UD.

If we have a long list of components, are we concerned about the display height of the stack boxes?  @ashumilova - if we have a super long list of components, do we just cut off the list or do we make the box taller?  We probably need to think about what to do here.

But I would think that we should go ahead and make the components list complete and then as a secondary measure work on improving the UX display of the stack box?

TylerJewell:
I do not have a great alternative for dev-machine.  But it always feels like a little bit of a robot to me whenever i see that appear in our product:  "i-am-a-dev-machine," said the metalic robot as he violently waves his robotic arms up and down wishing he had opposable thumbs.

ide/che-core-ide-stacks/src/main/resources/stacks.json
ghost: @TylerJewell Tag appear in UI and they are used to match stacks and filters. Components aren't used at all now in master.

This is a broader question and does not affect this PR only. This is how things work now, you may check with @ashumilova if you have doubts.

ghost:
@TylerJewell all your questions are general questions to how things work in stacks and templates for UD. What should I do with this PR? Everything you have mentioned is not going to be fixed in this PR.

ide/che-core-ide-templates/src/main/resources/samples.json
```diff
@@ -1,5 +1,66 @@
...
+      "parameters": {
+        "branch":"mysql"
+      }
+    },
+    "commands": [
+      {
+        "name": "build",
+        "type": "mvn",
+        "commandLine": "mvn -f ${current.project.path} clean install -DskipTests",
+        "attributes": {
```
TylerJewell, line 28: Can you check this macro with M1 - according to what Vlad produced, this macro is now in a different format.  Should we update it?

https://eclipse-che.readme.io/docs/commands

TylerJewell:
Just need to respond to my comment on components from this evening, and then take a look at the thought on changing the macro to the updated names.

ghost:
ci-build

Issue #1883, opened 2016-07-20
Add ability to preview an HTML file into another browser window


slemeur:
It should be possible to preview an HTML file into another browser window.

The experience expected would be the following:
- The project explorer is showing HTML files
- User can right-click on the HTML files
  - User will get "Preview" option in the right-click menu
- When click on "Preview", the HTML file is opened and rendered into another browser window. 

**Che version:**    4.5.0

TylerJewell:
Or would "Send-To" be better words?  I think either can work.

PR #2378, opened 2016-09-07
Add ability to preview an HTML file into another browser window

azatsarynnyy:
Added ability to preview an HTML file into another browser window.
#1883

@vparfonov


vparfonov:
ok

Issue #2278, opened 2016-08-31
After creation a new workspace the ssh button in the processes widget is absent


musienko-maxim:
**Reproduction Steps:**
Create a new workspace and go to the IDE

**Expected behavior:**
The ssh button should be present
**Observed behavior:**
There is no button. But after refreshing of browser - all ok.

![selection_009](https://cloud.githubusercontent.com/assets/1640058/18121217/2f509a5e-6f6c-11e6-9e84-8a82464a0956.png)
**Che version:**    4.7.0
**OS and version:** Ubuntu OS 14.0
**Docker version:** 1.12
**Che install:**  local bundle

**_I think this problem can be related with  this issue:_**  https://github.com/eclipse/che/issues/2249

bmicklea:
@vparfonov @TylerJewell I think this is a blocker because of how many people use SSH with our workspaces. I'm hoping it's an easy fix...

PR #2387, opened 2016-09-08
CHE-2278: Fixing appearing SSH button. Now it showing only after starting machine 

vparfonov:
Fixing appearing SSH button. Now it showing only after starting machine.
Fix stop workspace behavior: close all related process panel

https://github.com/eclipse/che/issues/2278
### Previous behavior

see https://github.com/eclipse/che/issues/2278
### New behavior

SSH button will be shown only after starting machine 


riuvshin:
+1

Issue #2030, opened 2016-08-03
Che server should connect to ws-agent on internal URL


ghost:
Cannot run Che in Dockert

**Expected behavior:**
Che starts

**Observed behavior:**

```
eugene@eugene:~/projects/artik-ide (master)$ docker run -t -v /var/run/docker.sock:/var/run/docker.sock codenvy/che-launcher start
Unable to find image 'codenvy/che-launcher:latest' locally
latest: Pulling from codenvy/che-launcher

4fe15f8d0ae6: Already exists 
4d8efc227322: Already exists 
5f70bf18a086: Pull complete 
6fb31488f897: Already exists 
5b30f66f120f: Pull complete 
4c092e74c599: Pull complete 
66227230d784: Pull complete 
Digest: sha256:ab282d4f1239610fd2a7f80ac9cb6e7e6edd2d97dabef64608a4083c939de5ad
Status: Downloaded newer image for codenvy/che-launcher:latest
INFO: ECLIPSE CHE: PULLING IMAGE codenvy/che-server:codenvy/che-launcher

FATA[0000] Invalid repository name (che-server:codenvy/che-launcher), only [a-z0-9-_.] are allowed 

```

**Che version:**    [Enter Che version here]  
**OS and version:** Ubuntu 14.04
**Docker version:** 1.11.2

ghost:
@l0rd can you reproduce it?

l0rd:
This has been fixed in #1983 but [codenvy/che-launcher](https://hub.docker.com/r/codenvy/che-launcher/tags/) hasn't been updated.

l0rd:
@riuvshin can you please check?

ghost:
@l0rd I tried it with nightly

 Fail pinging ws agent. Workspace ID:workspace94c7eunv699lec0u. Url:http://172.19.20.16:32773/wsagent/ext/. Timestamp:{}

Server cannot reach agent, disabling ufw solves the problem.

Are we OK with this?

l0rd:
We had an issue with the firewall on centos: the launcher could not verify if the server was up or not. So that's a different issue.

Can you please copy/paste the output of:
`docker run -v /var/run/docker.sock:/var/run/docker.sock codenvy/che-launcher:nightly info`

ghost:
```
eugene@eugene:~$ docker run -v /var/run/docker.sock:/var/run/docker.sock codenvy/che-launcher:nightly info
DEBUG: ---------------------------------------
[...]

PR #2402, opened 2016-09-09
Add property machine.docker.local_node_host.external

l0rd:
### What does this PR do?

Introduce a new property, `machine.docker.local_node_host.external` (along with the environment variable `CHE_DOCKER_MACHINE_HOST_EXTERNAL`), that allow to configure the external URL browsers need to use to connect with containers that are not directly pingable. This happens for example with Docker for Mac.

``` properties
machine.docker.local_node_host.external=localhost
```

When this property is set the browser will use it to connect to the containers. Note that the wsmaster won't use that value to connect to the containers.

If this property is not set Che behaves as before.

This PR introduce a change to the API: [Server class](https://github.com/eclipse/che/blob/master/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/Server.java) has a couple of new attributes: `internalUrl` and `internalAddress`. On the other there no particular change is needed client side and all clients will benefit of this change.

``` json
"servers": {
      "4401/tcp": {
                "url": "http://localhost:32906/wsagent/ext", 
                "internalUrl": "http://192.168.65.2:32906/wsagent/ext", 
                "address": "localhost:32906",
                "internalAddress": "192.168.65.2:32906",
                "path": "/wsagent/ext",
                "protocol": "http",
                "ref": "wsagent"
     }
}
```

This PR takes into account some feedbacks from PR #2004:
- There are no changes on client side
- Works on configurations where the wsagent and docker containers are on different hosts

I will not close PR #2004. I want to continue the work to integrate Traefik in the che-launcher there. But containous/traefik#444 should be fixed first.   
#1644
### Previous behavior

Running Che on Docker for Mac required some networking hacking
### New behavior

Remove the need for the networking hack when running Che on Docker for Mac and make running Che as simple as `docker run -v /var/run/docker.sock:/var/run/docker.sock codenvy/che-launcher start` on all platforms.
### PR type
### Minor change checklist
- [ ] New API required?
- [x] API updated
- [x] Tests updated
- [x] Tests passed

### DIFF
--- a/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/Server.java
+++ b/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/Server.java
@@ -25,7 +25,13 @@ public interface Server {
     String getRef();
 
     /**
-     * Address of the server in form <b>host:port</b>
+     * External address of the server in form <b>hostname:port</b>.
+     * <p>
+     * This address is used by the browser to communicate with the server.
+     * <b>hostname</b> can be configured using property machine.docker.local_node_host.external
+     * or environment variable CHE_DOCKER_MACHINE_HOST_EXTERNAL.
+     * <b>port</b> is the external port and cannot be configured.
+     * If not explicitly configured that address is set using {@link ServerProperties#getInternalAddress()}
      */
     String getAddress();
 
@@ -36,14 +42,15 @@ public interface Server {
     String getProtocol();
 
     /**
-     * Path to access the server.
-     */
-    @Nullable
-    String getPath();
-
-    /**
-     * Url of the server, e.g. http://localhost:8080
+     * Url of the server, e.g.&nbsp;http://localhost:8080
      */
     @Nullable
     String getUrl();
+
+
+    /**
+     * Non mandatory properties of the server.
+     */
+    @Nullable
+    ServerProperties getProperties();
 }
--- a/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/ServerProperties.java
+++ b/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/ServerProperties.java
@@ -1,0 +1,44 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.api.core.model.machine;
+
+import org.eclipse.che.commons.annotation.Nullable;
+
+/**
+ * Not mandatory properties of a {@link Server}
+ *
+ * @author Mario Loriedo
+ */
+public interface ServerProperties {
+
+    /**
+     * Path to access the server.
+     */
+    @Nullable
+    String getPath();
+
+    /**
+     * Internal address of the server in form <b>host:port</b>.
+     * <p>
+     * Used by wsmaster to communicate with the server
+     */
+    @Nullable
+    String getInternalAddress();
+
+
+    /**
+     * Internal Url of the server, e.g.&nbsp;http://localhost:8080.
+     * <p>
+     * Used by wsmaster to comunicate with the server
+     */
+    @Nullable
+    String getInternalUrl();
+}
--- a/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/DevMachineServer.java
+++ b/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/DevMachineServer.java
@@ -11,29 +11,32 @@
 package org.eclipse.che.ide.api.machine;
 
 import org.eclipse.che.api.core.model.machine.Server;
+import org.eclipse.che.api.core.model.machine.ServerProperties;
 import org.eclipse.che.api.machine.shared.dto.ServerDto;
+
+import java.util.Objects;
 
 /**
  * Describe development machine server instance.
- * {@link Server}
  *
+ * @link Server
  * @author Vitalii Parfonov
  */
 public class DevMachineServer implements Server {
 
 
-        private final String path;
-        private final String address;
-        private final String protocol;
-        private final String ref;
-        private final String url;
+    private final String address;
+    private final String protocol;
+    private final String ref;
+    private final String url;
+    private final ServerProperties properties;
 
         public DevMachineServer(Server dto) {
-            path = dto.getPath();
             address = dto.getAddress();
             protocol = dto.getProtocol();
             ref = dto.getRef();
             url = dto.getUrl();
+            properties = dto.getProperties();
         }
 
 
@@ -53,12 +56,45 @@ public class DevMachineServer implements Server {
         }
 
         @Override
-        public String getPath() {
-            return path;
-        }
-
-        @Override
         public String getUrl() {
             return url;
         }
-}
+
+        @Override
+        public ServerProperties getProperties() { return properties; };
+
+        @Override
+        public boolean equals(Object o) {
+            if (this == o) return true;
+            if (!(o instanceof DevMachineServer)) return false;
+            final DevMachineServer other = (DevMachineServer) o;
+            return Objects.equals(ref, other.ref) &&
+                           Objects.equals(protocol, other.protocol) &&
+                           Objects.equals(address, other.address) &&
+                           Objects.equals(url, other.url) &&
+                           Objects.equals(properties, other.properties);
+        }
+
+        @Override
+        public int hashCode() {
+            int hash = 7;
+            hash = hash * 31 + Objects.hashCode(ref);
+            hash = hash * 31 + Objects.hashCode(protocol);
+            hash = hash * 31 + Objects.hashCode(address);
+            hash = hash * 31 + Objects.hashCode(url);
+            hash = hash * 31 + Objects.hashCode(properties);
+            return hash;
+        }
+
+        @Override
+        public String toString() {
+            return "DevMachineServer{" +
+                           "ref='" + ref + '\'' +
+                           ", protocol='" + protocol + '\'' +
+                           ", address='" + address + '\'' +
+                           ", url='" + url + '\'' +
+                           ", properties='" + properties + '\'' +
+                           '}';
+        }
+
+    }
--- a/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/DevMachineServerProperties.java
+++ b/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/DevMachineServerProperties.java
@@ -1,0 +1,79 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.ide.api.machine;
+
+import org.eclipse.che.api.core.model.machine.ServerProperties;
+
+import java.util.Objects;
+
+/**
+ * Describe development machine server instance.
+ *
+ * @link ServerProperties
+ * @author Mario Loriedo
+ */
+public class DevMachineServerProperties implements ServerProperties {
+
+    private final String path;
+    private final String internalAddress;
+    private final String internalUrl;
+
+    public DevMachineServerProperties(ServerProperties properties) {
+        path = properties.getPath();
+        internalAddress = properties.getInternalAddress();
+        internalUrl = properties.getInternalUrl();
+    }
+
+    @Override
+    public String getInternalAddress() {
+        return internalAddress;
+    }
+
+    @Override
+    public String getPath() {
+        return path;
+    }
+
+    @Override
+    public String getInternalUrl() {
+        return internalUrl;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof DevMachineServerProperties)) return false;
+        final DevMachineServerProperties other = (DevMachineServerProperties)o;
+
+        return Objects.equals(path, other.path) &&
+                       Objects.equals(internalAddress, other.internalAddress) &&
+                       Objects.equals(internalUrl, other.internalUrl);
+    }
+
+    @Override
+    public int hashCode() {
+        int hash = 7;
+        hash = hash * 31 + Objects.hashCode(path);
+        hash = hash * 31 + Objects.hashCode(internalAddress);
+        hash = hash * 31 + Objects.hashCode(internalUrl);
+        return hash;
+    }
+
+    @Override
+    public String toString() {
+        return "DevMachineServerProperties{" +
+                       "path='" + path + '\'' +
+                       ", internalAddress='" + internalAddress + '\'' +
+                       ", internalUrl='" + internalUrl + '\'' +
+                       '}';
+    }
+
+}
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
@@ -145,6 +145,7 @@ public class DockerInstance extends AbstractInstance {
             try {
                 final ContainerInfo containerInfo = docker.inspectContainer(container);
                 machineRuntime = new MachineRuntimeInfoImpl(dockerMachineFactory.createMetadata(containerInfo,
+                                                                                                null,
                                                                                                 node.getHost(),
                                                                                                 getConfig()));
             } catch (IOException e) {
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
@@ -17,6 +17,8 @@ import org.eclipse.che.api.core.model.machine.MachineRuntimeInfo;
 import org.eclipse.che.api.core.model.machine.ServerConf;
 import org.eclipse.che.api.machine.server.model.impl.ServerConfImpl;
 import org.eclipse.che.api.machine.server.model.impl.ServerImpl;
+import org.eclipse.che.api.machine.server.model.impl.ServerPropertiesImpl;
+import org.eclipse.che.commons.annotation.Nullable;
 import org.eclipse.che.plugin.docker.client.json.ContainerConfig;
 import org.eclipse.che.plugin.docker.client.json.ContainerInfo;
 import org.eclipse.che.plugin.docker.client.json.ContainerState;
@@ -83,17 +85,21 @@ public class DockerInstanceRuntimeInfo implements MachineRuntimeInfo {
     protected static final String SERVER_CONF_LABEL_PATH_SUFFIX     = ":path";
 
     private final ContainerInfo               info;
-    private final String                      containerHost;
+    private final String                      containerExternalHostname;
+    private final String                      containerInternalHostname;
     private final Map<String, ServerConfImpl> serversConf;
 
     @Inject
     public DockerInstanceRuntimeInfo(@Assisted ContainerInfo containerInfo,
-                                     @Assisted String containerHost,
+                                     @Assisted("externalhost") @Nullable String containerExternalHostname,
+                                     @Assisted("internalhost") String containerInternalHostname,
                                      @Assisted MachineConfig machineConfig,
                                      @Named("machine.docker.dev_machine.machine_servers") Set<ServerConf> devMachineSystemServers,
                                      @Named("machine.docker.machine_servers") Set<ServerConf> allMachinesSystemServers) {
         this.info = containerInfo;
-        this.containerHost = containerHost;
+        this.containerExternalHostname = containerExternalHostname == null ?
+                                                    containerInternalHostname : containerExternalHostname;
+        this.containerInternalHostname = containerInternalHostname;
         Stream<ServerConf> confStream = Stream.concat(machineConfig.getServers().stream(), allMachinesSystemServers.stream());
         if (machineConfig.isDev()) {
             confStream = Stream.concat(confStream, devMachineSystemServers.stream());
@@ -244,7 +250,8 @@ public class DockerInstanceRuntimeInfo implements MachineRuntimeInfo {
         }
         return addDefaultReferenceForServersWithoutReference(
                 addRefAndUrlToServers(
-                        getServersWithFilledPorts(containerHost,
+                        getServersWithFilledPorts(containerExternalHostname,
+                                                  containerInternalHostname,
                                                   ports),
                         labels));
     }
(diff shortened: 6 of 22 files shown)


TylerJewell:
Nice work!  Excited to have this get into master.  I reviewed the changes for the che-launcher and they all look good.

TylerJewell:
@garagatyi - this resolves a big inconvenience that users have with Docker for Mac. This eliminates a painful setup of the loopback alias on Macs. Your +1 here will be helpful.

garagatyi:
In my opinion changing API for that workaround is not a good way to go. It will add pain sooner or later.
I've just figured out that we can try to play with IPTables in Che container. 
For example we can try to redirect traffic inside of Che container from 32000-xxxxx to IP of xhyve machine. And let expose machines ports as localhost. In that case browser will be able to connect to machines on localhost and Che-server will use localhost also. But IPTables rules in che-server container will forward all traffic to another IP.
As I understand we can do that even after start of che-server, so we will be able to do that in che launcher and not change base che-server which works fine in all cases except `che in container on mac`.
Here is an example that I googled:
`iptables -t nat -A OUTPUT -i lo -p tcp --dport 32000:65000 -j DNAT --to <IP OF XHYVE>:32000-65000`
But I didn't try it and have no experience with IPTables, so not sure that I found correct (and complete) set of rules for out case.

l0rd:
@garagatyi can you please be more specific? Where do you want to run this `iptable` command: OSX, xyve VM, che-server container, agent container? What do you mean when you say "let expose machines ports as localhost"?

garagatyi:
Add iptables rules in che-server container.

Configure che-server in such a way that **localhost** is host in servers provided by Che API.
I think we just should not set CHE_DOCKER_MACHINE_HOST variable to achieve that.

So all traffic from che-server container to localhost with ports 32000-xxxxx will go through IPTables redirection to IP of xhyve where ports of all containers are available.

l0rd:
Ok thanks for clarifying @garagatyi. I understand now. It's a smart solution :-) I will test that tomorrow.

l0rd:
Bingo! I've struggled to apply the IP tables rules but I've finally found the following command that work if executed inside `che-sever` container!

```
apk update
apk install iptables
sysctl -w net.ipv4.conf.eth0.route_localnet=1
iptables -t nat -A OUTPUT -p tcp -d localhost -o lo --dport 32000:65000 -j DNAT --to ${XHYVE_VM_IP}:32000-65000
iptables -t nat -A POSTROUTING -o eth0 -m addrtype --src-type LOCAL --dst-type UNICAST -j MASQUERADE
```

l0rd:
I've created a new PR to address the Docker for Mac problem: eclipse/che-dockerfiles#12

@garagatyi I will let you choose which one of these 2 PR is more suitable but let me explain why I believe we should keep this one and close eclipse/che-dockerfiles#12. Apart [some concerns](https://github.com/eclipse/che-dockerfiles/pull/12#issuecomment-246398648) with the iptables solution I also think that changing the Server class, adding an internal url attribute, is not just a patch to solve the Docker for Mac issue but a solution for a recurring problem.

image

In the current model there is only one address to reach the wsagent server. That same address is used by the browser and the wsmaster. In the diagram I've called this address "Post Office". Depending on the platform the "Post Office" can be `docker0`, the IP of the docker host or just `localhost`. However in most scenarios wsmaster and wsagent are close and there is a shorter path to reach the wsagent (the red path in the diagram). Most important this red path avoid some scary mountains that can be tricky to cross (firewalls, routers, IPtables etc...). That's the case with Docker for Mac but that's also the case with Ubuntu (#2030) or Kubernetes or Openshift etc...This PR is to add the address of this red path to the attributes of the Server class.

TylerJewell:
I tend to agree with Mario that having the internal sokution is more optimal for the longer term.  And that we can continue to layer in the iptables solution where necessary.

Great graphic :)

l0rd:
I've updated the PR. I've added a new ServerProperties class where I moved servers `path`, `internalUrl` and `internalAddress` fields. That's the new Server API:

``` json
"servers": {
    "4401/tcp": {
        "ref": "wsagent",
        "protocol": "http",
        "address": "localhost:32929",
        "url": "http://localhost:32929/wsagent/ext",
        "port": "32929",
        "properties": {
            "internalUrl": "http://192.168.65.2:32929/wsagent/ext",
            "internalAddress": "192.168.65.2:32929",
            "path": "/wsagent/ext"
        }
    }
}
```

@garagatyi  please review

benoitf:
I like it also
 I would say that timeframe for updating json format as we're in process for 5.x release and still in milestone,is the right time. After it would be too late

TylerJewell:
I am ok with this approach.  I want to target this for M2 - solving this problem simplifies a lot of scenarios for us with Mac.

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/ServerProperties.java
```diff
@@ -0,0 +1,41 @@
...
+     * Path to access the server.
+     */
+    @Nullable
+    String getPath();
+
+    /**
+     * Internal address of the server in form <b>host:port</b>
+     * Used by wsmaster to communicate with the server
+     */
+    String getInternalAddress();
```
garagatyi, line 34: Looks like all fields in that object are nullable.

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/ServerProperties.java
```diff
@@ -0,0 +1,41 @@
...
+public interface ServerProperties {
+
+    /**
+     * Path to access the server.
+     */
+    @Nullable
+    String getPath();
+
+    /**
+     * Internal address of the server in form <b>host:port</b>
```
garagatyi: Line break by entering 'enter' key doesn't work in javadocs. Please add dot and `<p/>` or `<br/>`

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/Server.java
```diff
@@ -25,25 +25,35 @@
     String getRef();
 
     /**
-     * Address of the server in form <b>host:port</b>
+     * External address of the server in form <b>host:port</b>
```
garagatyi: Line break by entering 'enter' key doesn't work in javadocs. Please add dot and `<p/>` or `<br/>`

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/DevMachineServerProperties.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * Contributors:
+ *   Mario Loriedo - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.ide.api.machine;
+
+import org.eclipse.che.api.core.model.machine.ServerProperties;
+
+/**
+ * Describe development machine server instance.
```
garagatyi, line 18: Line break by entering 'enter' key doesn't work in javadocs. Please add `<p/>` or `<br/>`

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
```diff
@@ -82,17 +83,20 @@
...
     private final ContainerInfo               info;
-    private final String                      containerHost;
+    private final String                      containerExternalHostname;
+    private final String                      containerInternalHostname;
     private final Map<String, ServerConfImpl> serversConf;
 
     @Inject
     public DockerInstanceRuntimeInfo(@Assisted ContainerInfo containerInfo,
-                                     @Assisted String containerHost,
+                                     @Assisted String containerExternalHostname,
```
garagatyi: Not named assisted injection doesn't work when several arguments have the same class

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/Server.java
```diff
@@ -25,25 +25,35 @@
...
+     * This address is used by the browser to communicate with the server and
+     * corresponds to the internal address if not explicitly set.
+     * Can be set using property machine.docker.local_node_host.external or
+     * environment variable CHE_DOCKER_MACHINE_HOST_EXTERNAL
      */
     String getAddress();
 
     /**
+     * Port of the server, e.g. 8080
```
garagatyi: I'm not sure should this port contain `8080` or `8080/tcp`.

garagatyi: And it is external port. Please explain that in javadocs.

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/DevMachineServer.java
```diff
@@ -53,12 +56,13 @@ public String getProtocol() {
...
-        @Override
         public String getUrl() {
             return url;
         }
+
+        @Override
+        public String getPort() { return port; }
+
+        @Override
+        public ServerProperties getProperties() { return properties; };
```
garagatyi: Please add equals/hashcode/toString methods

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/DevMachineServerProperties.java
```diff
@@ -0,0 +1,47 @@
...
+    @Override
+    public String getPath() {
+        return path;
+    }
+
+    @Override
+    public String getInternalUrl() {
+        return internalUrl;
+    }
+}
```
garagatyi, line 79: Please add equals/hashcode/toString methods

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/DevMachineServerProperties.java
```diff
@@ -0,0 +1,47 @@
...
+ *
+ * @author Mario Loriedo
+ */
+public class DevMachineServerProperties implements ServerProperties {
+
+    private final String path;
+    private final String internalAddress;
+    private final String internalUrl;
+
+    public DevMachineServerProperties(ServerProperties dto) {
```
garagatyi: Argument doesn't have to be DTO object. Please rename parameter to not confuse developers.

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/DevMachineServerProperties.java
```diff
@@ -0,0 +1,47 @@
...
+ * Contributors:
+ *   Mario Loriedo - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.ide.api.machine;
+
+import org.eclipse.che.api.core.model.machine.ServerProperties;
+
+/**
+ * Describe development machine server instance.
+ * {@link ServerProperties}
```
garagatyi: Curly brackets are used when link is inlined into description. When it is used as a separate tag it has the same syntax as `@author`

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
garagatyi: Also please check what will happen if someone provide null value as one of these parameters. I believe nullable annotation should be declared somewhere.

plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/perspective/widgets/machine/appliance/server/ServerProperties.java
```diff
@@ -0,0 +1,46 @@
...
+        return descriptor.getPath();
+    }
+
+    @Override
+    public String getInternalAddress() { return descriptor.getInternalAddress(); }
+
+    @Override
+    public String getInternalUrl() {
+        return descriptor.getInternalUrl();
+    }
```
garagatyi, line 46: Please add equals/hashcode/toString methods

garagatyi (changes requested):
Overall solution is good. Code requires some cleanups and fixes.

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/Server.java
```diff
@@ -25,25 +25,35 @@
...
 
     /**
-     * Address of the server in form <b>host:port</b>
+     * External address of the server in form <b>host:port</b>
+     * This address is used by the browser to communicate with the server and
+     * corresponds to the internal address if not explicitly set.
+     * Can be set using property machine.docker.local_node_host.external or
+     * environment variable CHE_DOCKER_MACHINE_HOST_EXTERNAL
      */
     String getAddress();
```
garagatyi, line 36: Now this method returns both host and port of the server. Are you going to change the behavior in this PR?

l0rd: No I'm not:

``` json
"address": "localhost:32929"
```

Is that ok?

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/Server.java
l0rd: It's `8080`, the protocol is not there:

``` json
"port": "32929"
```

garagatyi:
Looks like I'm not fully understand new server object. I'll take a look again tomorrow and ask you if something is not clear.

l0rd:
@garagatyi I've commited the fixes you have suggested:
- Javadoc improved
- Removed `port` field in `Server` class
- Annotated all fields in `ServerProperties` as nullable
- Overwritten methods `equals`, `hashcode` and  `toString` for classes `DevMachineServer` and `DevMachineServerProperties`
- Fixed unnamed assisted injections
- Added some unit tests
- Restored the license header (I'll submit a distinct PR for that)
- Environment variables have now higher priority then properties

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/Server.java
```diff
@@ -36,14 +42,15 @@
     String getProtocol();
 
     /**
-     * Path to access the server.
+     * Url of the server, e.g.&nbsp;http://localhost:8080
```
garagatyi: Please replace `&nbsp;` with a regular backspace.

plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/perspective/widgets/machine/appliance/server/ServerProperties.java
garagatyi: This class still needs equals/hashcode/toString methods

plugins/plugin-machine/che-plugin-machine-ext-client/src/test/java/org/eclipse/che/ide/extension/machine/client/perspective/widgets/machine/appliance/server/ServerTest.java
```diff
@@ -81,4 +80,37 @@ public void refShouldBeReturned() {
...
+        when(descriptor.getProperties()).thenReturn(descriptor2);
+        when(descriptor2.getInternalUrl()).thenReturn(SOME_TEXT);
+
+        assertThat(server.getProperties().getInternalUrl(), equalTo(SOME_TEXT));
+
+        verify(descriptor).getProperties();
+        verify(descriptor2).getInternalUrl();
+    }
+
 }
```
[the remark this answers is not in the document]
garagatyi: Please add trailing new line at the end of file

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/DtoConverter.java
```diff
@@ -10,24 +10,8 @@
  *******************************************************************************/
 package org.eclipse.che.api.machine.server;
 
-import org.eclipse.che.api.core.model.machine.MachineLimits;
```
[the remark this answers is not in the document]
garagatyi: Wildcard imports are forbidden in Che

wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/workspace/server/WorkspaceServiceTest.java
```diff
@@ -24,14 +24,7 @@
 import org.eclipse.che.api.core.rest.shared.dto.ServiceError;
 import org.eclipse.che.api.environment.server.MachineProcessManager;
 import org.eclipse.che.api.environment.server.MachineServiceLinksInjector;
-import org.eclipse.che.api.machine.server.model.impl.CommandImpl;
```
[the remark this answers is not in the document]
garagatyi: Please do not use wildcard imports

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
```diff
@@ -145,6 +145,7 @@ public MachineRuntimeInfoImpl getRuntime() {
             try {
                 final ContainerInfo containerInfo = docker.inspectContainer(container);
                 machineRuntime = new MachineRuntimeInfoImpl(dockerMachineFactory.createMetadata(containerInfo,
+                                                                                                null,
```
garagatyi, line 148: For me it looks like this argument is always null. Did I miss something?

garagatyi (changes requested):

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/Server.java
l0rd, line 45: I think you mean "regular space". If I add a regular space and I try to generate the Javadoc I have the following result:

image

If I use `&nbsp;` instead the sentence is not broken:

image

plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/perspective/widgets/machine/appliance/server/ServerProperties.java
l0rd: Class https://github.com/eclipse/che/blob/master/plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/perspective/widgets/machine/appliance/server/Server.java had no equals method. Should I add equals/hashcode/toString to existing classes too?

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/DtoConverter.java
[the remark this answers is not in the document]
l0rd: Sorry for forgetting that. I've fixed it.

plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/perspective/widgets/machine/appliance/server/ServerProperties.java
garagatyi: At least add it in classes added by you.

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/machine/Server.java
garagatyi: OK, interesting

wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/workspace/server/WorkspaceServiceTest.java
```diff
@@ -24,14 +24,7 @@
...
-import org.eclipse.che.api.machine.server.model.impl.MachineConfigImpl;
-import org.eclipse.che.api.machine.server.model.impl.MachineImpl;
-import org.eclipse.che.api.machine.server.model.impl.MachineLimitsImpl;
-import org.eclipse.che.api.machine.server.model.impl.MachineRuntimeInfoImpl;
-import org.eclipse.che.api.machine.server.model.impl.MachineSourceImpl;
-import org.eclipse.che.api.machine.server.model.impl.ServerImpl;
-import org.eclipse.che.api.machine.server.model.impl.SnapshotImpl;
+import org.eclipse.che.api.machine.server.model.impl.*;
```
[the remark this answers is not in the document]
l0rd: I think I alrady fixed that: https://github.com/l0rd/che/blob/a35ccae37ab7fdecf958beb02acbf0e8177c8a3b/wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/workspace/server/WorkspaceServiceTest.java

You were probably looking at the outdated version?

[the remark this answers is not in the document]
benoitf: yep it was pending request (I forgot to validate it)

benoitf (approved):

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
l0rd: Feel free to add more details to your questions if that was not what you meant. I think you are wondering how `Server.url` is ever set if this parameter is always `null`. It should be set using property `machine.docker.local_node_host.external` but it looks as it's always null.

Property `machine.docker.local_node_host.external` is injected in `LocalDockerModule` objects that are binded to `DockerInstanceRuntimeInfo` in `LocalDockerModule`. 

Therefore method `DockerInstance.getRuntime()` is not used by the API to get the machine runtime. That explains why `Server.url` is correctly returned by the API.

garagatyi: Let me explain how I understand what is happening with these changes. Please correct me if I'm mistaken.
DockerInstance always pass null as external host. DockerInstanceRuntimeInfo doesn't change that. LocalDockerInstanceRuntimeInfo uses assisted injection, so this null is passed here too. 
So I don't see how another than null value can be accepted by DockerInstanceRuntimeInfo.

garagatyi (changes requested):

plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/perspective/widgets/machine/appliance/server/ServerProperties.java
l0rd: I've added for both.

wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/workspace/server/WorkspaceServiceTest.java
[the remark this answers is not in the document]
l0rd: Fixed

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/ServerImpl.java
```diff
@@ -22,22 +23,24 @@
...
         this.url = url;
-        this.protocol = protocol;
-        this.path = path;
+        this.port = port;
+        this.properties = properties;
     }
 
     public ServerImpl(Server server) {
-        this(server.getRef(), server.getProtocol(), server.getAddress(), server.getPath(), server.getUrl());
+        this(server.getRef(), server.getProtocol(), server.getAddress(), server.getUrl(), server.getPort(), (ServerPropertiesImpl) server.getProperties());
```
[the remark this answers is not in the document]
l0rd: Fixed

plugins/plugin-machine/che-plugin-machine-ext-client/src/test/java/org/eclipse/che/ide/extension/machine/client/perspective/widgets/machine/appliance/server/ServerTest.java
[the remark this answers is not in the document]
l0rd: Fixed

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfoTest.java
```diff
@@ -507,30 +574,43 @@ public void shouldAddCommonAndDevSystemServersConfigToDevMachine() throws Except
...
         expectedServers.put("4305/udp", new ServerImpl("devSysServer1-udp",
                                                        null,
                                                        CONTAINER_HOST + ":32103",
-                                                       "some/path4",
-                                                       null));
+                                                       null,
+                                                       "32103",
+                                                       new ServerPropertiesImpl("some/path4",
+                                                           CONTAINER_HOST + ":32103",
+                                                           null)));
```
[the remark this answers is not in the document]
l0rd: Added a couple of tests

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalDockerInstanceRuntimeInfo.java
```diff
@@ -24,32 +24,52 @@
...
+                                          @Nullable @Named("machine.docker.local_node_host") String dockerNodeInternalHostname,
+                                          @Nullable @Named("machine.docker.local_node_host.external") String dockerNodeExternalHostname,
                                           @Named("machine.docker.dev_machine.machine_servers") Set<ServerConf> devMachineServers,
                                           @Named("machine.docker.machine_servers") Set<ServerConf> allMachinesServers) {
+
         super(containerInfo,
-              dockerNodeHost != null ? dockerNodeHost : (System.getenv(CHE_DOCKER_MACHINE_HOST) != null ?
-                                                         System.getenv(CHE_DOCKER_MACHINE_HOST) :
-                                                         containerHost),
+              dockerNodeExternalHostname != null ? dockerNodeInternalHostname : (
```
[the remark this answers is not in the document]
l0rd: Fixed. Env variables are now evaluated first.

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/DevMachineServerProperties.java
l0rd: Done

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/DevMachineServerProperties.java
l0rd: Done

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/DevMachineServer.java
l0rd: Done

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
l0rd: I've changed to named assisted injection and checked if `containerExternalHostname` is null and set it to `containerHost` in that case.

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/DevMachineServerProperties.java
l0rd: Done

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
```diff
@@ -299,9 +315,10 @@ public String projectsRoot() {
             String externalPort = portEntry.getValue().get(0).getHostPort();
             servers.put(portProtocol, new ServerImpl(null,
                                                      null,
-                                                     host + ":" + externalPort,
+                                                     externalHostame + ":" + externalPort,
                                                      null,
-                                                     null));
+                                                     externalPort,
+                                                     new ServerPropertiesImpl(null, internalHostname + ":" + externalPort, null)));
```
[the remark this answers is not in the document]
l0rd: After fixup `externalHostname`, `internalHostname` and `externalPorts` can never be `null`.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstance.java
l0rd: As discussed on Slack I've cleaned-up `LocalDockerInstanceRuntimeInfo` class and made it easier to read.

garagatyi:
@musienko-maxim Please do a QA cycle for this PR

garagatyi (approved):

garagatyi:
ci-build

garagatyi:
QA team has approved this PR. We will merge it when M3 is released

PR #2408, opened 2016-09-10
Move oauth implementation to ide app

vzhukovs:
Reorganize internal API. Move implementation `org.eclipse.che.ide.auth.OAuthServiceClientImpl` from `code-ide-api` to `core-ide-app`, because it isn't a part of exposed API.

@vparfonov review it, plz.


vzhukovs:
These changes has been already merged with https://github.com/eclipse/che/pull/2867

PR #2414, opened 2016-09-12
Change fields mapping from varchar to text

akorneta:
Since we have stored data in text format we consider to change mapping of chosen fields from varchar to text.

@skabashnyuk, @mkuznyetsov please take a look.


skabashnyuk:
ok

Issue #1985, opened 2016-07-28
NPE after deleting multi-module project


svor:
If I try to delete some big project (for example [eclipse/che](https://github.com/eclipse/che/) when the process of resolving dependencies is in progress, many NPEs are thrown (see attached file [log.txt](https://github.com/eclipse/che/files/388608/log.txt)).

**Reproduction Steps:**
1. Import https://github.com/eclipse/che/ 
2. Configure as Maven project
3. Delete this project

**Expected behavior:**

Project should be deleted without exceptions

**Observed behavior:**

Project isn't deleted and an error notification is shoved

**Che version:**    4.5.0  
**OS and version:** Ubuntu 12.04  
**Docker version:** 1.11.1  
**Che install:**   Codenvy

**Additional information:**
Can reproduced only on https://nightly4.codenvy-stg.com

TylerJewell:
Starting this off as a blocker - until we have discussion.

JamesDrummond:
@evidolob Provide status label.

JamesDrummond:
@azatsarynnyy Please remove status label.

PR #2419, opened 2016-09-12
Fix NPE after deleting multi-module project

azatsarynnyy:
Fix NPE on deleting project while it's indexing.
#1985


vparfonov:
ok

PR #2445, opened 2016-09-14
Fix local JPA migration for workspaces and ssh

mkuznyetsov:
### What does this PR do?
- Fix migration for workspace, count that old format uses namespace field instead of account
- Fix migration for SSH keys, count that old format is stored as map<owner, pair>


akorneta:
ok

Issue #2372, opened 2016-09-07
Not all editor tabs are closed after stopping workspace (regression)


artaleks9:
**Reproduction Steps:**
- Expand a project and open two or more files.
- Perform the stop workspace.

**Expected behavior:**
The editor tabs should be closed.

**Observed behavior:**
One editor tab still stays open.
Also we have errors into console of browser.

**Che version:**    5.0.0  
**OS and version:** Ubuntu 14.04  
**Docker version:** 1.11.1  
**Che install:**    local bundle   

**Additional information:**
See the attachment.
![tab-editor-stop-ws](https://cloud.githubusercontent.com/assets/5408906/18307712/43edb3c8-74fa-11e6-8411-8f90af9d58af.gif)

riuvshin:
regression

bmicklea:
Clearing milestone since this is listed as starting in the next sprint.

artaleks9:
Right now this bug is reproduced again.

vparfonov:
can't reproduce

PR #2448, opened 2016-09-14
CHE-1857: don't save the content if ws is stopped

svor:
### What does this PR do?

Fixes problems with closing tabs after stopping the ws

https://github.com/eclipse/che/issues/2372

@vparfonov


plugins/plugin-orion/che-plugin-orion-editor/src/main/java/org/eclipse/che/ide/editor/orion/client/OrionEditorPresenter.java
```diff
@@ -567,7 +565,9 @@ public void onFailure(final Throwable caught) {
 
     @Override
     public void doSave(final AsyncCallback<EditorInput> callback) {
-
+        if (appContext.getDevMachine() == null) {
```
vparfonov, line 569: please add here comments  what means if devMachine is null

svor: done

vparfonov (approved):

PR #2485, opened 2016-09-19
CHE-2484 Remove autostart LS agents

tolusha:
LS agents should be started on demand.

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
@@ -286,9 +286,6 @@ public class WorkspaceRuntimes {
                 if (extendedMachine != null) {
                     List<String> agents = extendedMachine.getAgents();
                     launchAgents(instance, agents);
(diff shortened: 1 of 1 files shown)


vparfonov (approved):

kaloyan-raev:
This change broke the language server support in the master branch.

Trying to open a JSON editor fails because the language server cannot be started.

It seems that on demand is started only the language server, but not the agent. Using the debugger I see a call to launch the language server, but its binaries (node runtime and ls-json folder) are missing under /home/user/che.

tolusha:
@kaloyan-raev 
We are working on the issue when a use will be able to turn on/off agents in his workspace.
By default they are off since workspace staring time grows.

kaloyan-raev:
@tolusha in the current state of the master branch, is there a way to manually turn on an agent (e.g. JSON LS agent)?

Or should I revert this change in my local repo?

tolusha:
Currently just modify `https://github.com/eclipse/che/blob/master/ide/che-core-ide-stacks/src/main/resources/stacks.json` by adding required ls agents into agents section of any stack you need.

Available ls agents are:
- org.eclipse.che.ls.csharp
- org.eclipse.che.ls.json
- org.eclipse.che.ls.php

kaloyan-raev:
@tolusha Thanks! Good enough for now.

PR #2487, opened 2016-09-19
CHE-2366: Fix volumes_from usage in compose environment

### LINKED ISSUE #2366: volumes_from does not work
When using volumes_from in compose file, an error occurred saying that container with a `service` name does not exist.

**Reproduction Steps:**

Create a custom stack with the following recipe:

```
services:
 db:
  image: codenvy/mysql
  environment:
   MYSQL_ROOT_PASSWORD: root
   MYSQL_USER: petclinic
   MYSQL_PASSWORD: password
  mem_limit: 2147483648
 dev-machine:
  image: codenvy/ubuntu_jdk8
  mem_limit: 2147483648
  depends_on:
    - db
  links:
    - db
  volumes_from:
   - db

```

**Expected behavior:**

Workspace starts and dev-machine inherits volumes from db service

**Observed behavior:**

Unable to start workspace agent. Error when trying to start the workspace agent: Start of environment wksp-7vlgfw failed. Error: Error response from docker API, status: 404, message: No such container: db

garagatyi:
### What does this PR do?

Workarounds usage of volumes_from in compose environment.
Also cleanups code.

Fixes #2366 
### Previous behavior

Volumes from didn't work.
### New behavior

Volumes from works. Volumes from are treated as machines dependency also.
### PR type
- [ ] ~~Major change = changes existing features or docs~~
### Minor change checklist
- [ ] ~~New API required?~~
- [ ] ~~API updated~~
- [x] Tests passed


garagatyi:
@skabashnyuk @evoevodin Please review

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/MachineProviderImplTest.java
```diff
@@ -1574,29 +1527,28 @@ public void shouldAddEnvVarsFromMachineConfigToContainerOnDevInstanceCreationFro
...
-                                                  workspaceFolderPathProvider,
-                                                  PROJECT_FOLDER_PATH,
-                                                  false,
-                                                  false,
-                                                  Collections.emptySet(),
-                                                  Collections.emptySet(),
-                                                  SNAPSHOT_USE_REGISTRY,
-                                                  MEMORY_SWAP_MULTIPLIER,
-                                                  emptySet());
+        provider = new MachineProviderImpl(dockerConnector,
```
voievodin: I see the code like this 

``` java
new MachineProviderImpl(dockerConnector,
                        dockerConnectorConfiguration,
                        credentialsReader,
                        dockerMachineFactory,
                        dockerInstanceStopDetector,
                        Collections.emptySet(),
                        Collections.emptySet(),
                        Collections.emptySet(),
                        Collections.emptySet(),
                        null,
                        workspaceFolderPathProvider,
                        PROJECT_FOLDER_PATH,
                        false,
                        false,
                        Collections.emptySet(),
                        Collections.emptySet(),
                        SNAPSHOT_USE_REGISTRY,
                        MEMORY_SWAP_MULTIPLIER,
                        emptySet());
```

in all the tests. It's logical that there is some difference between them but still it is not obvious from object 
creation, i understand that test name may help on this consideration but still i think it is not really nice to have such code.

I suggest to create a helpful class which may produce the needed instance of `MachineProviderImpl` e.g.
`MachineProviderImplBuilder` with a method which may produce default instance of it and than set only the difference. 
The example:

``` java
MachineProviderImplBuilder.initDefault()
                          .setMemorySwapMultiplier(...)
                          .build()
```

In this way you may simplify the code for the readers who are not really into constructor parameters

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/ContainerNameGenerator.java
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Returns generated name for container.
+     *
+     * @param workspaceId
+     *         unique workspace id, see more (@link WorkspaceConfig#getId)
+     * @param machineId
+     *         unique machine id, see more {@link Machine#getId()}
+     * @param namespace
+     *         name of the user who is docker container owner
```
voievodin, line 31: Is it docker related?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/DefaultServicesStartStrategy.java
```diff
@@ -132,7 +156,22 @@ private String getServiceFromMachineLink(String link) throws IllegalArgumentExce
         if (link != null) {
             String[] split = service.split(":");
             if (split.length != 1 && split.length != 2) {
-                throw new IllegalArgumentException(format("Service link %s is invalid", link));
+                throw new IllegalArgumentException(format("Service link '%s' is invalid", link));
+            }
+            service = split[0];
```
evoevodin: Is it possible to get split length equal to `0` ?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/DefaultServicesStartStrategy.java
```diff
@@ -132,7 +156,22 @@ private String getServiceFromMachineLink(String link) throws IllegalArgumentExce
...
+        }
+        return service;
+    }
+
+    /**
+     * Parses volumesFrom content into depends_on field representation - removes column and further chars
+     */
+    private String getServiceFromVolumesFrom(String volumesFrom) throws IllegalArgumentException {
```
voievodin, line 169: The same code is in `getServiceFromMachineLink` method, are you sure it can't be reused?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
```diff
@@ -106,20 +110,55 @@ public ComposeEnvironmentImpl parse(Environment environment) throws IllegalArgum
...
+                                                                                           .entrySet()) {
+            ComposeServiceImpl service = composeServiceEntry.getValue();
+
+            CheServiceBuildContextImpl buildContext = null;
+            if (service.getBuild() != null) {
+                buildContext = new CheServiceBuildContextImpl().withContext(service.getBuild().getContext())
+                                                               .withDockerfilePath(service.getBuild().getDockerfile());
+            }
+
+            CheServiceImpl cheService = new CheServiceImpl().withBuild(buildContext)
```
voievodin, line 133: add method `fromService(service)` to the builder, to make this code reusable

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/model/CheServiceImpl.java
```diff
@@ -0,0 +1,485 @@
...
+                      containerName,
+                      command,
+                      entrypoint,
+                      image,
+                      dependsOn,
+                      environment,
+                      expose,
+                      ports,
+                      labels,
+                      links,
```
voievodin: This may be wrong for 2 equal instances, consider the following example:

``` java
CheServiceImpl s1 = new CheServiceImpl();
CheServiceImpl s2 = new CheServiceImpl();
s1.getLinks();
assert sq.equals(s2); // <- this will fail.
```

Explanation: an empty `ArrayList#hashCode` returns 1, while `Object.hashCode(null)` returns 0
:cry:

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/model/CheServiceImpl.java
```diff
@@ -0,0 +1,485 @@
...
+               Objects.equals(containerName, service.containerName) &&
+               Objects.equals(command, service.command) &&
+               Objects.equals(entrypoint, service.entrypoint) &&
+               Objects.equals(image, service.image) &&
+               Objects.equals(dependsOn, service.dependsOn) &&
+               Objects.equals(environment, service.environment) &&
+               Objects.equals(expose, service.expose) &&
+               Objects.equals(ports, service.ports) &&
+               Objects.equals(labels, service.labels) &&
+               Objects.equals(links, service.links) &&
```
voievodin: The object with empty links and null links should be considered as equal

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/model/CheServicesEnvironmentImpl.java
```diff
@@ -0,0 +1,96 @@
...
+        return this;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (!(o instanceof CheServicesEnvironmentImpl)) return false;
+        CheServicesEnvironmentImpl that = (CheServicesEnvironmentImpl)o;
+        return Objects.equals(version, that.version) &&
+               Objects.equals(services, that.services);
```
voievodin: The same is here, see the comment to `CheServiceImpl` class

voievodin (changes requested):
Generally ok, please answer/fix suggestions.
Do the changes you're proposing related to the way we're storing data ?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/ContainerNameGenerator.java
garagatyi: Yes

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/MachineProviderImplTest.java
garagatyi: I'll take a look

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/DefaultServicesStartStrategy.java
garagatyi: It is more readable for me

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/MachineProviderImplTest.java
garagatyi: It really looks cleaner, thanks.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
garagatyi: Not sure this class needs such a builder. Notice that I build one class from another (very similar) class.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/model/CheServiceImpl.java
garagatyi: ok, I'll use getters for that.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/model/CheServiceImpl.java
garagatyi: I'll fix that by getters

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/model/CheServicesEnvironmentImpl.java
garagatyi: I'll fix that by getters

garagatyi:
@evoevodin I've fixed code that you commented.

voievodin (approved):

Issue #2498, opened 2016-09-20
Improve error handling when read a container logs


mmorhun:
Now we just retrying to get container logs when some error occurs.
We need some stop condition and delay between attempts to get logs.

mmorhun:
Fixed

PR #2499, opened 2016-09-20
CHE-2498: Improve error handling when read a container logs

mmorhun:
### What does this PR do?

Improves error handling while getting logs from a container. Adds more information into error log.
#2498
### Previous behavior

No stop condition when many errors occur
### New behavior

Added parameter to stop trying to get logs and added delay between this attempts


mmorhun:
@AndrienkoAleksandr @garagatyi take a look please

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ComposeMachineProviderImpl.java
```diff
@@ -590,10 +595,22 @@ private void readContainerLogsInSeparateThread(String container,
                 } catch (ContainerNotFoundException e) {
                     isContainerRunning = false;
                 } catch (IOException e) {
-                    LOG.error("Failed to get logs from machine {} of workspace {} backed by container {}",
+                    LOG.error("Failed to get logs from machine {} of workspace {} backed by container {}, because: {}. Cause: {}",
```
AndrienkoAleksandr: Maybe it would be better to print error in the log after 5 five failed attempts...

mmorhun:
@sleshchenko take a look, please

garagatyi:
Have you found the reason of such behavior?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ComposeMachineProviderImpl.java
```diff
@@ -590,10 +593,28 @@ private void readContainerLogsInSeparateThread(String container,
...
                 } catch (IOException e) {
-                    LOG.error("Failed to get logs from machine {} of workspace {} backed by container {}",
+                    LOG.warn("Failed to get logs from machine {} of workspace {} backed by container {}, because: {}.",
                               workspaceId,
                               machineId,
-                              container);
+                              container,
+                              e.getMessage(),
+                              e);
+                    if (++errorsCounter == 5) {
```
sleshchenko: I'd move harcoded value to constant

sleshchenko:
OK for me

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ComposeMachineProviderImpl.java
garagatyi: I think it is Ok to hardcode it

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ComposeMachineProviderImpl.java
```diff
@@ -590,10 +593,28 @@ private void readContainerLogsInSeparateThread(String container,
                 } catch (ContainerNotFoundException e) {
                     isContainerRunning = false;
                 } catch (IOException e) {
-                    LOG.error("Failed to get logs from machine {} of workspace {} backed by container {}",
+                    LOG.warn("Failed to get logs from machine {} of workspace {} backed by container {}, because: {}.",
                               workspaceId,
                               machineId,
```
garagatyi: Arguments order is incorrect

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ComposeMachineProviderImpl.java
```diff
@@ -590,10 +593,28 @@ private void readContainerLogsInSeparateThread(String container,
...
-                              container);
+                              container,
+                              e.getMessage(),
+                              e);
```
garagatyi: Looks like LOG.warn is misused. This methods accepts 1 string and 1 throwable or multiple strings. Did you use it in such a way on purpose?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ComposeMachineProviderImpl.java
```diff
@@ -590,10 +593,28 @@ private void readContainerLogsInSeparateThread(String container,
...
+                    if (++errorsCounter == 5) {
+                        LOG.error("Too many errors while streaming logs from machine {} of workspace {} backed by container {}. " +
+                                  "Logs streaming is closed. Last error: {}.",
+                                  workspaceId,
+                                  machineId,
```
garagatyi: Check order of arguments and usage of multiple arguments + throwable

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ComposeMachineProviderImpl.java
```diff
@@ -590,10 +593,28 @@ private void readContainerLogsInSeparateThread(String container,
...
+                                  container,
+                                  e.getMessage(),
+                                  e);
+                        break;
+                    } else {
+                        try {
+                            sleep(1000);
+                        } catch (InterruptedException ie) {
+                            break;
```
garagatyi: return is more obvious here

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ComposeMachineProviderImpl.java
garagatyi: What if errors happen once an hour? Please respect timeout.

garagatyi (changes requested):

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ComposeMachineProviderImpl.java
mmorhun: thx

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ComposeMachineProviderImpl.java
```diff
@@ -590,10 +594,34 @@ private void readContainerLogsInSeparateThread(String container,
                 } catch (ContainerNotFoundException e) {
                     isContainerRunning = false;
                 } catch (IOException e) {
-                    LOG.error("Failed to get logs from machine {} of workspace {} backed by container {}",
-                              workspaceId,
-                              machineId,
-                              container);
+                    LOG.warn("Failed to get logs from machine {} of workspace {} backed by container {}, because: {}.",
```
garagatyi: Don't you think that we should set **lastProcessedLogDate** here also?

garagatyi (changes requested):

garagatyi (approved):

PR #2507, opened 2016-09-20
CHE-2506: Use tag instead of snapshot of the io.typefox.lsapi project

tolusha:


benoitf (changes requested):
requires https://github.com/eclipse/che-dependencies/pull/18 to be merged

benoitf (approved):
cq validated as parallel IP 
OK for me

PR #2523, opened 2016-09-21
Che CLI: added some parameters to execute che-mount

l0rd:
### What does this PR do?

Adapt Che CLI to the che-mount updates introduced by eclipse/che-dockerfiles#18:
- Fix local synched folder permissions on Linux
- Persist workspace SSH ECDSA key fingerprint in local `~/.ssh/known_hosts`
- eclipse/che-dockerfiles#18

### DIFF
--- a/che.sh
+++ b/che.sh
@@ -756,9 +756,20 @@ execute_che_mount() {
     error "${CHE_MINI_PRODUCT_NAME} mount: Path provided is not a valid directory."
     return
   fi
+  
+  if is_native; then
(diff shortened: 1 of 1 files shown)


benoitf:
how to provide custom ssh key if it's not stored in user home folder ?

for example with `che dir up`, ssh files are stored in the .che directory 
is that an option to provide private key is possible ?

for example sshfs option seems to be
`-o IdentityFile=/path/to/key`

l0rd:
@benoitf This PR do not address that. Using option `IdentityFile` would be a way to do that or we can just run che-mount using volume `-v /path/to/key:${HOME}/.ssh/key`

benoitf (approved):

PR #2526, opened 2016-09-21
Use depends_on instead of links

ghost:
links does not work. depends_on does.


riuvshin:
+1, there is an issue for problems with links: https://github.com/codenvy/codenvy/issues/806

riuvshin (approved):

garagatyi (approved):
LGTM

PR #2529, opened 2016-09-21
CODENVY-502: introduce application notifications and container for them

ashumilova:
### What does this PR do?

Introduces dashboard application notifications.

Is needed for https://github.com/codenvy/codenvy/issues/502

### DIFF
--- a/dashboard/src/components/notification/application-notifications.factory.js
+++ b/dashboard/src/components/notification/application-notifications.factory.js
@@ -1,0 +1,112 @@
+/*
+ * Copyright (c) 2015-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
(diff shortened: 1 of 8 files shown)


dashboard/src/components/notification/application-notifications.factory.js
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  /**
+   * Add error notification.
+   *
+   * @param title notification's title
+   * @param content notification's content
+   * @returns {{notification}} notification
+   */
+  addErrorNotification(title, content) {
```
benoitf, line 57: maybe we could factorize code of addErrorNotification, addWarningNotification and addInfoNotification as only type is changing

ashumilova: Fixed https://github.com/eclipse/che/pull/2529/files#diff-e28604a5d5045aea96555422a3400cd9R58

PR #2535, opened 2016-09-21
CLI: Use new che-mount & adds mechanism for CLI to update itself to new versions

TylerJewell:
### What does this PR do?

1: Alters the syntax so that volume mounts can have windows paths with spaces and dots.
2: Changes the syntax so that always $PWD will be selected as the local mount point.
3: Adds in discovery - if there is only a single ws running, will find the SSH port and use that.
4: If there are multiple workspaces with valid SSH ports, lists all of them and their ports.
5: Adds in new capability that allows users to download specific version of CLI and run it.
6: Uses curl to download initial CLI if it is not already there.

This new logic assumes that SSH:
1: Is running on port 22/tcp within the container, and:
2: Is accessible by the user named 'user'.  

It will be better in the future, if we only exchange keys and automate this.  Will wait for Florent's improvements here.


TylerJewell:
@benoitf @l0rd

che.sh
```diff
@@ -740,38 +740,66 @@ update_che_image() {
...
+  # Port not provided, let's do a simple discovery of running workspaces
+  else 
+    info "${CHE_MINI_PRODUCT_NAME} mount: Searching for running workspaces with open SSH port..."
+
+    CURRENT_WS_INSTANCES=$(docker ps -aq --filter "name=_che_dev-machine")
+    CURRENT_WS_COUNT=$(echo $CURRENT_WS_INSTANCES | wc -w)
+    
+    if [ $CURRENT_WS_COUNT -eq 0 ]; then
+      error "${CHE_MINI_PRODUCT_NAME} mount: We could not find any running workspaces"
+
```
l0rd: You should `return` after printing the error message

che.sh
```diff
@@ -740,38 +740,66 @@ update_che_image() {
...
-            -v /etc/group:/etc/group:ro \
-            -v /etc/passwd:/etc/passwd:ro \
-            -u $(id -u ${USER})"
+    docker_run_with_che_properties --cap-add SYS_ADMIN \
+                                   --device /dev/fuse \
+                                   -v ${HOME}/.ssh:${HOME}/.ssh \
+                                   -v ${HOME}/.unison:${HOME}/.unison \
+                                   -v /etc/group:/etc/group:ro \
+                                   -v /etc/passwd:/etc/passwd:ro \
+                                   -u $(id -u ${USER})
```
l0rd: The backslash at the end of the line is missing

l0rd (changes requested):
I've tested on OSX and Linux. There are couple of things to fixes. Besides that LGTM

che.sh
```diff
@@ -740,38 +740,66 @@ update_che_image() {
...
+                                   -v "${MOUNT_PATH}":/mnthost \
+                                   "${CHE_MOUNT_IMAGE_NAME}":"${CHE_UTILITY_VERSION}" \
+                                        "${GLOBAL_GET_DOCKER_HOST_IP}" $WS_PORT
```
benoitf: indent is wrong

che.sh
```diff
@@ -740,38 +740,66 @@ update_che_image() {
...
+    
   else
-    PLATFORM_SPECIFIC_OPTIONS="-v ${HOME}/.ssh:/root/.ssh"
+    docker_run_with_che_properties --cap-add SYS_ADMIN \
+                                   --device /dev/fuse \
+                                   -v "${HOME_PATH}"/.ssh:/root/.ssh \
+                                   -v "${MOUNT_PATH}":/mnthost \
+                                   "${CHE_MOUNT_IMAGE_NAME}":"${CHE_UTILITY_VERSION}" \
+                                        "${GLOBAL_GET_DOCKER_HOST_IP}" $WS_PORT
```
benoitf: indent

che.sh
```diff
@@ -740,38 +740,66 @@ update_che_image() {
...
+
+    CURRENT_WS_INSTANCES=$(docker ps -aq --filter "name=_che_dev-machine")
+    CURRENT_WS_COUNT=$(echo $CURRENT_WS_INSTANCES | wc -w)
+    
+    if [ $CURRENT_WS_COUNT -eq 0 ]; then
+      error "${CHE_MINI_PRODUCT_NAME} mount: We could not find any running workspaces"
+
+    elif [ $CURRENT_WS_COUNT -eq 1 ]; then
+      RUNNING_WS_PORT=$(docker inspect --format='{{ (index (index .NetworkSettings.Ports "22/tcp") 0).HostPort }}' ${CURRENT_WS_INSTANCES})
+      info "${CHE_MINI_PRODUCT_NAME} mount: Connecting to remote workspace on port $RUNNING_WS_PORT"
```
benoitf: side note
if we're always using info "${CHE_MINI_PRODUCT_NAME} 
"${CHE_MINI_PRODUCT_NAME} " should be inside info function

or we need to add another info commands to avoid to write ${CHE_MINI_PRODUCT_NAME}  in all cals

che.sh
```diff
@@ -740,38 +740,66 @@ update_che_image() {
...
+    info "${CHE_MINI_PRODUCT_NAME} mount: Searching for running workspaces with open SSH port..."
+
+    CURRENT_WS_INSTANCES=$(docker ps -aq --filter "name=_che_dev-machine")
+    CURRENT_WS_COUNT=$(echo $CURRENT_WS_INSTANCES | wc -w)
+    
+    if [ $CURRENT_WS_COUNT -eq 0 ]; then
+      error "${CHE_MINI_PRODUCT_NAME} mount: We could not find any running workspaces"
+
+    elif [ $CURRENT_WS_COUNT -eq 1 ]; then
+      RUNNING_WS_PORT=$(docker inspect --format='{{ (index (index .NetworkSettings.Ports "22/tcp") 0).HostPort }}' ${CURRENT_WS_INSTANCES})
```
benoitf: what if there is no ssh running ?

che.sh
TylerJewell: good catch

che.sh
TylerJewell: yes, another good catch.

che.sh
TylerJewell: I made a number of improvements to handle this:
1. I now get a list of all ws containers.
2. I added in a check to see if they have ssh or not.
3. I then print out a list of all the containers their SSH port & whether they have SSH.

```
INFO: che mount: Setting local mount path to /c/codenvy/tmp/dir 4
INFO: che mount: Searching for running workspaces with open SSH port...
INFO: che mount: Re-run with 'che mount <ssh-port>'
WS CONTAINER ID    HAS SSH?    SSH PORT
3ca5f2d5c302       y           32797
33c0ce87fabf       n
679494b6cae7       y           32788
```

che.sh
TylerJewell: Yes, i agree - will address in another improvement.

che.sh
TylerJewell: The indent was intended that way for readability - because it's the post container params.

che.sh
TylerJewell: The indent was intended that way for readability - because it's the post container params.

benoitf:
cli.sh is being part of this pull request but it seems cli.sh is already in che repository which look strange ? as this PR seems to want to split it in two parts

cli.sh
```diff
@@ -0,0 +1,956 @@
+init_global_variables() {
```
benoitf, line 12: Requires copyright headers + shebang

cli.sh
```diff
@@ -0,0 +1,956 @@
...
+
+  GLOBAL_NAME_MAP=$(docker info | grep "Name:" | cut -d" " -f2)
+  GLOBAL_HOST_ARCH=$(docker version --format {{.Client}} | cut -d" " -f5)
+  GLOBAL_UNAME=$(docker run --rm alpine sh -c "uname -r")
+  GLOBAL_GET_DOCKER_HOST_IP=$(get_docker_host_ip)
+
+  if is_boot2docker && has_docker_for_windows_client; then
+  	if [[ "${CHE_DATA_FOLDER,,}" != *"${USERPROFILE,,}"* ]]; then
+  	  CHE_DATA_FOLDER=$(get_mount_path "${USERPROFILE}/.${CHE_MINI_PRODUCT_NAME}/")
+      warning "Boot2docker for Windows - CHE_DATA_FOLDER set to $CHE_DATA_FOLDER"
```
benoitf, line 53: indent look odd to me

che.sh
```diff
@@ -184,806 +83,79 @@ is_debug() {
...
-  ;;
-  info)
-    load_profile
-    execute_che_info "$@"
-  ;;
-  help)
-    usage
-  ;;
-esac
+execute_cli "$@"
```
benoitf: missing end of line

benoitf:
```
$ ./che.sh start
INFO: Pulling image curl:latest
INFO: Downloading cli-latest
INFO: ECLIPSE CHE: Already have image codenvy/che-server:latest
INFO: ECLIPSE CHE: Starting container...
INFO: ECLIPSE CHE: Server logs at "docker logs -f che-server"
INFO: ECLIPSE CHE: Server booting...
INFO: ECLIPSE CHE: Booted and reachable
INFO: ECLIPSE CHE: http://localhost:8080
```

```

$ export CHE_VERSION=nightly
benoitf@MacBookFlorent:che (che-mount-cli=)$ ./che.sh start
INFO: ECLIPSE CHE: Already have image codenvy/che-server:nightly
INFO: ECLIPSE CHE: Starting container...
INFO: ECLIPSE CHE: Server logs at "docker logs -f che-server"
INFO: ECLIPSE CHE: Server booting...
INFO: ECLIPSE CHE: Booted and reachable
INFO: ECLIPSE CHE: http://localhost:8080

```

cli.sh
TylerJewell: added.

cli.sh
TylerJewell: fixed

che.sh
TylerJewell: fixed

benoitf (approved):

PR #2541, opened 2016-09-22
Add testing stack for Langage Server agents

slemeur:
### What does this PR do?

This PR add a testing stack where the LS Agents are embed. This way, it ease the ability to test those LS until the UI is implemented in the Dashboard
#1281
### Previous behavior

To test LS Agent, developers had to create a new stack and add the LS Agents. 
### New behavior

A default sample stack is available to test LS Agents.
### PR type

Signed-Off-by: Stevan Le Meur stevan.lemeur@gmail.com

### DIFF
--- a/ide/che-core-ide-stacks/src/main/resources/stacks.json
+++ b/ide/che-core-ide-stacks/src/main/resources/stacks.json
@@ -962,6 +962,56 @@
     }
   },
   {
+    "id": "debianlsp",
+    "creator": "ide",
+    "name": "Debian LSP",
+    "description": "Simple stack with Debian (Jessie) to test LSP",
+    "scope": "advanced",
+    "tags": [
+      "Debian",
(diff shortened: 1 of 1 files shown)


ashumilova (approved):

ide/che-core-ide-stacks/src/main/resources/stacks.json
```diff
@@ -962,6 +962,56 @@
...
+          },
+          "recipe": {
+            "location": "codenvy/debian_jre",
+            "type": "dockerimage"
+          }
+        }
+      },
+      "name": "default",
+      "defaultEnv": "default",
+      "description": null
```
benoitf, line 1011: do we need to set null if there is no description ?

benoitf (approved):
Please add Signed-Off

Issue #2435, opened 2016-09-13
Projects are not removed after deleting a workspace


artaleks9:
**Reproduction Steps:**
- Create a workspace, open in the IDE and create a project (e.g.'web-spring').
-  Go to dashboard and delete the workspace.
- Create the workspace with same name and open it In the IDE. 

**Expected behavior:**
in the project explorer should not be any project

**Observed behavior:**
there is the project as no configured.

**Che version:**   5.0.0  
**OS and version:** Ubuntu 14.04  
**Docker version:** 1.11.1  
**Che install:**   local bundle  

**Additional information:**
- Problem can be reliably reproduced, doesn't happen randomly: Yes
- On the hosted che we have the same behavior (the projects are not deleted after delete workspace).
  ![project_after_delete_ws](https://cloud.githubusercontent.com/assets/5408906/18470994/0df088dc-79b8-11e6-81f2-46d23b84e9d2.png)

bmicklea:
@vkuznyetsov can you get a code review completed for this so we can have it merged into M5

riuvshin:
this issue in  progress again after reviews.

PR #2544, opened 2016-09-22
CHE-2435: Add component for clean up workspace folder after remove workspace.

AndrienkoAleksandr:
### What does this PR do?

 Add component for clean up workspace folder after remove workspace.

https://github.com/eclipse/che/issues/2435
### PR type
### Minor change checklist
- [ ] API updated
- [ ] Tests passed
### Major change checklist
- [ ] API updated
- [ ] Tests passed

@garagatyi review please.

### DIFF
--- a/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
+++ b/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
@@ -10,9 +10,19 @@
  *******************************************************************************/
 package org.eclipse.che.api.core.util;
 
+import com.google.common.base.Joiner;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import static java.lang.String.format;
 
 /**
  * Helpers to manage system processes.
@@ -21,6 +31,8 @@ import java.io.InputStreamReader;
  * @author Alexander Garagatyi
  */
 public final class ProcessUtil {
+
+    private static final Logger         LOG             = LoggerFactory.getLogger(ProcessUtil.class);
     private static final ProcessManager PROCESS_MANAGER = ProcessManager.newInstance();
 
     /**
@@ -65,6 +77,54 @@ public final class ProcessUtil {
                 stdout.writeLine(line);
             }
         }
+    }
+
+    /**
+     * Start the process, writing the stdout and stderr to {@code outputConsumer} and terminate process by {@code timeout}.<br>
+     *
+     * @param commandLine
+     *         arguments of process command
+     * @param timeout
+     *         timeout for process. If process duration > {@code timeout} than kill process and throw {@link TimeoutException}.
+     * @param timeUnit
+     *         timeUnit of the {@code timeout}.
+     * @param outputConsumer
+     *         a consumer where stdout and stderr will be redirected
+     * @return the started process
+     * @throws InterruptedException
+     *         in case terminate process
+     * @throws IOException
+     *         in case I/O error
+     * @throws TimeoutException
+     *         if process gets more time then defined by {@code timeout}
+     */
+    public static Process executeAndWait(String[] commandLine, int timeout, TimeUnit timeUnit, LineConsumer outputConsumer)
+            throws TimeoutException, IOException, InterruptedException {
+        ProcessBuilder pb = new ProcessBuilder(commandLine).redirectErrorStream(true);
+
+        Process process = pb.start();
+
+        CompletableFuture.runAsync(() -> {
+            try {
+                // consume logs until process ends
+                process(process, outputConsumer);
+            } catch (IOException e) {
+                LOG.error(format("Failed to complete reading of the process '%s' output due to occurred error",
+                                 Joiner.on(" ").join(commandLine)), e);
+            }
+        });
+
+        if (!process.waitFor(timeout, timeUnit)) {
+            try {
+                ProcessUtil.kill(process);
+            } catch (RuntimeException x) {
+                LOG.error("An error occurred while killing process '{}'", Joiner.on(" ").join(commandLine));
+            }
+            throw new TimeoutException(format("Process '%s' was terminated by timeout %s %s.",
+                                              Joiner.on(" ").join(commandLine), timeout, timeUnit.name().toLowerCase()));
+        }
+
+        return process;
     }
 
     /**
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/LocalWorkspaceFilesCleaner.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/LocalWorkspaceFilesCleaner.java
@@ -1,0 +1,52 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.plugin.docker.machine.cleaner;
+
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+import com.google.inject.name.Named;
+
+import org.eclipse.che.api.core.model.workspace.Workspace;
+import org.eclipse.che.api.workspace.server.WorkspaceFilesCleaner;
+import org.eclipse.che.commons.lang.IoUtil;
+import org.eclipse.che.plugin.docker.machine.local.node.provider.LocalWorkspaceFolderPathProvider;
+
+import java.io.File;
+import java.io.IOException;
+
+/**
+ * Local implementation of the {@link WorkspaceFilesCleaner}.
+ *
+ * @author Alexander Andrienko
+ */
+@Singleton
+public class LocalWorkspaceFilesCleaner implements WorkspaceFilesCleaner {
+
+    private final LocalWorkspaceFolderPathProvider workspaceFolderPathProvider;
+
+    @Inject(optional = true)
+    @Named("host.projects.root")
+    private String hostProjectsFolder;
+
+    @Inject
+    public LocalWorkspaceFilesCleaner(LocalWorkspaceFolderPathProvider workspaceFolderPathProvider) {
+        this.workspaceFolderPathProvider = workspaceFolderPathProvider;
+    }
+
+    @Override
+    public void clear(Workspace workspace) throws IOException {
+        String workspacePath = workspaceFolderPathProvider.getPathByName(workspace.getConfig().getName());
+        File workspaceStorage = new File(workspacePath);
+        if (!workspacePath.equals(hostProjectsFolder) && workspaceStorage.exists()) {
+            IoUtil.deleteRecursive(workspaceStorage);
+        }
+    }
+}
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
@@ -32,7 +32,6 @@ import org.eclipse.che.api.machine.server.model.impl.SnapshotImpl;
 import org.eclipse.che.api.machine.server.spi.Instance;
 import org.eclipse.che.api.workspace.server.WorkspaceRuntimes.RuntimeDescriptor;
 import org.eclipse.che.api.workspace.server.event.WorkspaceCreatedEvent;
-import org.eclipse.che.api.workspace.server.event.WorkspaceRemovedEvent;
 import org.eclipse.che.api.workspace.server.model.impl.WorkspaceConfigImpl;
 import org.eclipse.che.api.workspace.server.model.impl.WorkspaceImpl;
 import org.eclipse.che.api.workspace.server.model.impl.WorkspaceRuntimeImpl;
@@ -326,8 +325,8 @@ public class WorkspaceManager {
         if (runtimes.hasRuntime(workspaceId)) {
             throw new ConflictException("The workspace '" + workspaceId + "' is currently running and cannot be removed.");
         }
+
         workspaceDao.remove(workspaceId);
-        eventService.publish(new WorkspaceRemovedEvent(workspaceId));
         LOG.info("Workspace '{}' removed by user '{}'", workspaceId, sessionUserNameOr("undefined"));
     }
 
@@ -648,7 +647,12 @@ public class WorkspaceManager {
     void performAsyncStop(WorkspaceImpl workspace) throws ConflictException {
         checkWorkspaceIsRunning(workspace, "stop");
         final String autoSnapshotAttr = workspace.getAttributes().get(AUTO_CREATE_SNAPSHOT);
-        final boolean createSnapshot = autoSnapshotAttr == null ? defaultAutoSnapshot : parseBoolean(autoSnapshotAttr);
+        boolean createSnapshot;
+        if (workspace.isTemporary()) {
+            createSnapshot = false;
+        } else {
+            createSnapshot = autoSnapshotAttr == null ? defaultAutoSnapshot : parseBoolean(autoSnapshotAttr);
+        }
         executor.execute(ThreadLocalPropagateContext.wrap(() -> {
             final String stoppedBy = sessionUserNameOr(workspace.getAttributes().get(WORKSPACE_STOPPED_BY));
             LOG.info("Workspace '{}:{}' with id '{}' is being stopped by user '{}'",
--- a/assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterModule.java
+++ b/assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterModule.java
@@ -141,5 +141,7 @@ public class WsMasterModule extends AbstractModule {
         final MessageBodyAdapterInterceptor interceptor = new MessageBodyAdapterInterceptor();
         requestInjection(interceptor);
         bindInterceptor(subclassesOf(CheJsonProvider.class), names("readFrom"), interceptor);
+        bind(org.eclipse.che.api.workspace.server.WorkspaceFilesCleaner.class)
+                .to(org.eclipse.che.plugin.docker.machine.cleaner.LocalWorkspaceFilesCleaner.class);
     }
 }
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineModule.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineModule.java
@@ -28,6 +28,7 @@ public class DockerMachineModule extends AbstractModule {
     @Override
     protected void configure() {
         bind(org.eclipse.che.plugin.docker.machine.cleaner.DockerContainerCleaner.class);
+        bind(org.eclipse.che.plugin.docker.machine.cleaner.RemoveWorkspaceFilesAfterRemoveWorkspaceEventSubscriber.class);
 
         Multibinder<String> devMachineEnvVars = Multibinder.newSetBinder(binder(),
                                                                          String.class,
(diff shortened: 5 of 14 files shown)


wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceProjectStorageCleaner.java
```diff
@@ -0,0 +1,28 @@
...
+ *******************************************************************************/
+package org.eclipse.che.api.workspace.server;
+
+/**
+ * This component removes workspace folder from workspace storage. It's used for "delete workspace" operation
+ * (see more {@link WorkspaceService#delete}).
+ *
+ * @author Alexander Andrienko
+ */
+public interface WorkspaceProjectStorageCleaner {
```
mshaposhnik: `WorkspaceFileStorageCleaner` or `WorkspaceFSStorageCleaner` looks somehow better for me. WDYT?

AndrienkoAleksandr: Ok, it's make sense. I'll renamed it soon. Thank you.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
```diff
@@ -68,6 +70,55 @@ public static void process(Process p, LineConsumer stdout) throws IOException {
...
+     *         a consumer where stdout and stderr will be redirected
+     * @return the started process
+     * @throws InterruptedException
+     *         in case terminate process
+     * @throws IOException
+     *         in case I/O error
+     * @throws TimeoutException
+     *         if process gets more time then defined by {@code timeout}
+     */
+    public static Process execute(String[] commandLine, int timeout, TimeUnit timeUnit, LineConsumer outputConsumer)
```
garagatyi: If process is ended/killed by the end of execution of this method what for it returns process object?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
```diff
@@ -68,6 +70,55 @@ public static void process(Process p, LineConsumer stdout) throws IOException {
...
+            throws TimeoutException, IOException, InterruptedException {
+        ProcessBuilder pb = new ProcessBuilder(commandLine).redirectErrorStream(true);
+
+        // process will be stopped after timeout
+        Watchdog watcher = new Watchdog(timeout, timeUnit);
```
garagatyi: Please use name for watchdog thread - it will help understand thread dump in case of troubles with the server.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
```diff
@@ -68,6 +70,55 @@ public static void process(Process p, LineConsumer stdout) throws IOException {
...
+            final ValueHolder<Boolean> isTimeoutExceeded = new ValueHolder<>(false);
+            watcher.start(() -> {
+                isTimeoutExceeded.set(true);
+                ProcessUtil.kill(process);
+            });
+
+            // consume logs until process ends
+            process(process, outputConsumer);
+
+            process.waitFor();
```
garagatyi: Please use timeout here

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
```diff
@@ -68,6 +70,55 @@ public static void process(Process p, LineConsumer stdout) throws IOException {
...
+
+            if (isTimeoutExceeded.get()) {
+                throw new TimeoutException();
```
garagatyi: This exception would be clearer if it had a message on what has happened.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -326,6 +329,9 @@ public void removeWorkspace(String workspaceId) throws ConflictException, Server
         if (runtimes.hasRuntime(workspaceId)) {
             throw new ConflictException("The workspace '" + workspaceId + "' is currently running and cannot be removed.");
         }
+
+        workspaceProjectStorageCleaner.remove(workspaceId);
```
garagatyi: If it is clearer then method name `clear` looks more natural for me than `remove`

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/LocalWorkspaceProjectStorageCleanerImpl.java
```diff
@@ -0,0 +1,53 @@
...
+
+    @Inject
+    public LocalWorkspaceProjectStorageCleanerImpl(WorkspaceFolderPathProvider workspaceFolderPathProvider) {
+        this.workspaceFolderPathProvider = workspaceFolderPathProvider;
+    }
+
+    @Override
+    public void remove(String workspaceId) {
+        try {
+            File workspaceStorage = new File(workspaceFolderPathProvider.getPath(workspaceId));
```
garagatyi: Have you covered the case when workspace are sharing single folder as a storage of projects?

garagatyi (changes requested):
Overall OK, but needs some fixes

garagatyi:
@evoevodin Please take a look

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
AndrienkoAleksandr: OK

garagatyi:
QA is needed

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
AndrienkoAleksandr: OK

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/LocalWorkspaceProjectStorageCleanerImpl.java
AndrienkoAleksandr: No, I''ll look at this case.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
AndrienkoAleksandr: This class is utility.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
AndrienkoAleksandr: This class is utility.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceFSStorageCleaner.java
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * This component removes workspace folder from workspace storage. It's used for "delete workspace" operation
+ * (see more {@link WorkspaceService#delete}).
+ *
+ * @author Alexander Andrienko
+ */
+public interface WorkspaceFSStorageCleaner {
+
+    /**
+     * Removes workspace project folder with all data by {@code workspaceId}. Note: all user's projects will be deleted.
```
voievodin: From the definition it's not clear for me what is _workspace project folder_? Does the workspace have the only one project?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceFSStorageCleaner.java
```diff
@@ -0,0 +1,28 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.api.workspace.server;
+
+/**
+ * This component removes workspace folder from workspace storage. It's used for "delete workspace" operation
```
voievodin: > It's used for "delete workspace" operation

If this interface exists for certain needs please document this needs better, as it's not clear what the phrase above means. As far as i understand we need this interface for cleaning file system resources after workspace is removed?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/LocalWorkspaceFSStorageCleanerImpl.java
```diff
@@ -0,0 +1,59 @@
...
+        this.workspaceFolderPathProvider = workspaceFolderPathProvider;
+    }
+
+    @Override
+    public void clear(String workspaceId) {
+        try {
+            String workspacePath = workspaceFolderPathProvider.getPath(workspaceId);
+            File workspaceStorage = new File(workspacePath);
+            if (!workspacePath.equals(hostProjectsFolder) && workspaceStorage.exists()) {
+                FileCleaner.addFile(workspaceStorage);
```
voievodin: Are you sure that it's expected behaviour, should it be asynchronous? If yes, do you think it's better to document it?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/LocalWorkspaceFSStorageCleanerImpl.java
```diff
@@ -0,0 +1,59 @@
...
+            }
+        } catch (IOException e) {
```
voievodin: I think it's better throw the exception from the method, document it, and catch it on callers side. The motivation is that if we need to add another cleaner, that cleaner will have to implement a batch of the same logic, like exception handling and it's processing.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
```diff
@@ -68,6 +70,55 @@ public static void process(Process p, LineConsumer stdout) throws IOException {
...
+            final ValueHolder<Boolean> isTimeoutExceeded = new ValueHolder<>(false);
+            watcher.start(() -> {
+                isTimeoutExceeded.set(true);
+                ProcessUtil.kill(process);
+            });
+
+            // consume logs until process ends
+            process(process, outputConsumer);
+
+            process.waitFor(timeout, timeUnit);
```
evoevodin: Is the watchdog needed here?
It looks like you can probably just wait for a process and then fail or continue depending on value returned from the `waitFor`

voievodin: What if instead of using `Watchdog` we consume all the logs in separate thread and wait for pocess to end or reach timeout, i mean we can use the value returned from the `waitFor` method to achive this goal, 
e.g.

``` java
if (!process.waitFor(timeout, timeUnit)) {
    ProcessUtil.kill(process);
}
```

@garagatyi, @AndrienkoAleksandr what do you think?

voievodin (changes requested):
Please comment on my comments/questions

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceFSStorageCleaner.java
AndrienkoAleksandr: This interface need to realize delete workspace folder with projects after remove workspace, I'll fix doc.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/LocalWorkspaceFSStorageCleanerImpl.java
AndrienkoAleksandr: I think remove workspace folder can be potentionally long time operation(if workspace contains a lot of files and server computer is busy to done this task quickly), so yes I think it should be asynchronous operation. Thank you, I will document it.

AndrienkoAleksandr: Documented in the interface WorkspaceFSStorageCleaner

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceFSStorageCleaner.java
AndrienkoAleksandr: Docs changed

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/LocalWorkspaceFSStorageCleanerImpl.java
AndrienkoAleksandr: I don't think that failed clean up workspace folder is reason to cancel delete workspace operation. That why I handle exception here.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
garagatyi: ok

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
garagatyi: I see that it is needed to process return code

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
garagatyi: @evoevodin I think it can be really clean solution in term of code. 
WDYT about usage of ComplletableFuture.runAsync in such cases? Without that addition of executor and everything like that would not make it simpler.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/LocalWorkspaceFSStorageCleanerImpl.java
garagatyi: But it is still better to handle batch of delete operations on higher level and decide there completion of which operation is optional and which not. +1 to what Eugene said

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -326,6 +329,9 @@ public void removeWorkspace(String workspaceId) throws ConflictException, Server
         if (runtimes.hasRuntime(workspaceId)) {
             throw new ConflictException("The workspace '" + workspaceId + "' is currently running and cannot be removed.");
         }
+
+        workspaceFSStorageCleaner.clear(workspaceId);
```
garagatyi: Please cover in tests

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -645,7 +654,12 @@ WorkspaceImpl performAsyncStart(WorkspaceImpl workspace,
     void performAsyncStop(WorkspaceImpl workspace) throws ConflictException {
         checkWorkspaceIsRunning(workspace, "stop");
         final String autoSnapshotAttr = workspace.getAttributes().get(AUTO_CREATE_SNAPSHOT);
-        final boolean createSnapshot = autoSnapshotAttr == null ? defaultAutoSnapshot : parseBoolean(autoSnapshotAttr);
+        boolean createSnapshot;
+        if (workspace.isTemporary()) {
+            createSnapshot = false;
```
garagatyi, line 652: It is very nice that you added that fix but , please, cover it in tests

garagatyi (changes requested):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
AndrienkoAleksandr: Done.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
AndrienkoAleksandr: Done

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/LocalWorkspaceFilesCleaner.java
```diff
@@ -0,0 +1,52 @@
...
+    public LocalWorkspaceFilesCleaner(LocalWorkspaceFolderPathProvider workspaceFolderPathProvider) {
+        this.workspaceFolderPathProvider = workspaceFolderPathProvider;
+    }
+
+    @Override
+    public void clear(Workspace workspace) throws IOException {
+        String workspacePath = workspaceFolderPathProvider.getPathByName(workspace.getConfig().getName());
+        File workspaceStorage = new File(workspacePath);
+        if (!workspacePath.equals(hostProjectsFolder) && workspaceStorage.exists()) {
+            FileCleaner.addFile(workspaceStorage);
```
AndrienkoAleksandr: Maybe it would be better to use not bad alternative: IoUtil.deleteRecursive(workspaceStorage); What do you think @garagatyi @evoevodin ?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/RemoveWorkspaceListener.java
```diff
@@ -0,0 +1,65 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(RemoveWorkspaceListener.class);
+
+    private final WorkspaceFilesCleaner workspaceFilesCleaner;
+    private final ExecutorService       executor;
+
+    @Inject
+    public RemoveWorkspaceListener(EventService eventService, WorkspaceFilesCleaner workspaceFilesCleaner) {
+        this.workspaceFilesCleaner = workspaceFilesCleaner;
+        eventService.subscribe(this);
```
voievodin: subscribe on events in `@PostConstruct`

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/RemoveWorkspaceListener.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @Override
+    public void onEvent(WorkspaceRemovedEvent event) {
+        executor.execute(ThreadLocalPropagateContext.wrap(() -> {
+            Workspace workspace = event.getWorkspace();
+            try {
+                workspaceFilesCleaner.clear(workspace);
+            } catch (IOException | ServerException e) {
+                LOG.error("Failed to delete folder for workspace with id: '{}'. Cause: '{}'", workspace.getId(), e.getMessage());
```
voievodin: consider a different message, less concrete, let's say it's _Failed to remove workspace files_

voievodin (approved):
Good job, please fix some small remarks

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/LocalWorkspaceFilesCleaner.java
AndrienkoAleksandr: For now class LocalWorkspaceFilesCleaner.java works synchronously so I used  IoUtil.deleteRecursive

garagatyi (approved):
Looks good!

Issue #2424, opened 2016-09-13
Cannot start a workspace with alpine base image


ghost:
Although alpine has bash installed, Che fails to inject its agents

**Reproduction Steps:**
1. Create a custom workspace from the following recipe:

```
services:
 dev-machine:
  image: codenvy/alpine_jdk8
  mem_limit: 2147483648
**Expected behavior:**
```

Workspace starts

**Observed behavior:**

```
[dev-machine] [DOCKER] latest: Pulling from codenvy/alpine_jdk8 
[dev-machine] [DOCKER] Digest: sha256:2da6f80e2c0689f3b5d9ddc8788e69287f77eed832b717a4984640a1c4ba52f2 
[dev-machine] [DOCKER] Status: Image is up to date for codenvy/alpine_jdk8:latest 
[dev-machine] [STDOUT] fetch http://dl-cdn.alpinelinux.org/alpine/v3.4/main/x86_64/APKINDEX.tar.gz
[dev-machine] [STDOUT] ssh-keygen: generating new host keys: RSA DSA ECDSA ED25519 
[dev-machine] [STDOUT] fetch http://dl-cdn.alpinelinux.org/alpine/v3.4/community/x86_64/APKINDEX.tar.gz
[dev-machine] [STDOUT] v3.4.3-26-gc506510 [http://dl-cdn.alpinelinux.org/alpine/v3.4/main]
[dev-machine] [STDOUT] v3.4.2-11-g9b41a63 [http://dl-cdn.alpinelinux.org/alpine/v3.4/community]
[dev-machine] [STDOUT] OK: 5970 distinct packages available
[dev-machine] [STDOUT] (1/3) Installing libssh2 (1.7.0-r0)
[dev-machine] [STDOUT] (2/3) Installing libcurl (7.50.1-r0)
[dev-machine] [STDOUT] (3/3) Installing curl (7.50.1-r0)
[dev-machine] [STDOUT] Executing busybox-1.24.2-r11.trigger
[dev-machine] [STDOUT] Executing glibc-bin-2.23-r3.trigger
[dev-machine] [STDOUT] OK: 27 MiB in 30 packages
[dev-machine] [STDERR] ps: unrecognized option: C
[dev-machine] [STDERR] BusyBox v1.24.2 (2016-08-12 14:38:34 GMT) multi-call binary.
[dev-machine] [STDERR] 
[dev-machine] [STDERR] Usage: ps [-o COL1,COL2=HEADER]
[...]

PR #2546, opened 2016-09-22
CHE-2424: Fix agents scripts to work with codenvy/alphine_jdk image

tolusha:
Commands were adopted to work with alpine images:
- ps 
- sed


tolusha:
@skabashnyuk 
@benoitf 
pls review

wsmaster/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/ProcessIsLaunchedChecker.java
```diff
@@ -39,7 +45,7 @@ public ProcessIsLaunchedChecker(String processNameToWait) {
     @Override
     public boolean isLaunched(Agent agent, InstanceProcess process, Instance machine) throws MachineException {
         Command command = new CommandImpl(format("Wait for %s, try %d", agent.getName(), ++counter),
-                                          format("ps -fC %s 1>/dev/null && echo 0 || echo 1", processNameToWait),
+                                          format(CHECK_COMMAND, processNameToWait),
```
benoitf, line 47: multiline command is working ?
I saw many times a very long line with only && in it

wsmaster/che-core-api-agent/src/main/resources/agents/org.eclipse.che.ls.csharp.json
```diff
@@ -3,5 +3,5 @@
...
...n }\n\n command -v nodejs >/dev/null 2>&1 || {\n curl --silent --location https://rpm.nodesource.com/setup_6.x | bash -;\n ${SUDO} yum -y install nodejs;\n }\n\n\n# openSUSE 13.2\n###############\nelif echo ${LINUX_TYPE} | grep -qi \"opensuse\"; then\n test \"${PACKAGES}\" = \"\" || {\n ${SUDO} zypper install -y ${PACKAGES};\n }\n\n command -v dotnet >/dev/null 2>&1 || {\n curl -L -o dotnet.tar.gz https://go.microsoft.com/fwlink/?LinkID=816867;\n ${SUDO} zypper install -y libunwind libicu;\n ${SUDO} mkdir -p /opt/dotnet;\n ${SUDO} tar zxf dotnet.tar.gz -C /opt/dotnet;\n rm dotnet.tar.gz;\n ${SUDO} ln -s /opt/dotnet/dotnet /usr/local/bin;\n }\n\n command -v nodejs >/dev/null 2>&1 || {\n ${SUDO} zypper ar http://download.opensuse.org/repositories/devel:/languages:/nodejs/openSUSE_13.1/ Node.js\n ${SUDO} zypper in nodejs\n }\n\nelse\n >&2 echo \"Unrecognized Linux Type\"\n >&2 cat /etc/os-release\n exit 1\nfi\n\n\n#####################\n### Install C# LS ###\n#####################\n\ncurl -s ${AGENT_BINARIES_URI} | tar xzf - -C ${CHE_DIR}\n\ntouch ${LS_LAUNCHER}\nchmod +x ${LS_LAUNCHER}\necho \"nodejs ${LS_DIR}/node_modules/omnisharp-client/languageserver/server.js\" > ${LS_LAUNCHER}"
```
benoitf: shebang should be present in start of each script file no ?
#!/bin/sh ?

wsmaster/che-core-api-agent/src/main/resources/agents/scripts/org.eclipse.che.ls.json.script.sh
```diff
@@ -21,7 +21,7 @@ LS_DIR=${CHE_DIR}/ls-json
 LS_LAUNCHER=${LS_DIR}/launch.sh
 
 LINUX_TYPE=$(cat /etc/os-release | grep ^ID= | tr '[:upper:]' '[:lower:]')
-LINUX_VERSION=$(cat /etc/os-release | grep ^VERSION=)
+LINUX_VERSION=$(cat /etc/os-release | grep ^VERSION.*=)
```
benoitf: on alpine I have now
$ echo $(cat /etc/os-release | grep ^VERSION.*=)
VERSION_ID=3.4.0

but on ubuntu I have several values now

echo $(cat /etc/os-release | grep ^VERSION=)
VERSION="16.04 LTS (Xenial Xerus)"

user@87e1b3d386f3:/projects$ echo $(cat /etc/os-release | grep ^VERSION.*=)
VERSION="16.04 LTS (Xenial Xerus)" VERSION_ID="16.04"

it doesn't break ?
or it should only check for VERSION_ID ?

wsmaster/che-core-api-agent/src/main/resources/agents/scripts/org.eclipse.che.ssh.script.sh
```diff
@@ -92,7 +92,11 @@ else
     exit 1
 fi
 
-ps -fC sshd && exit
+if echo ${LINUX_TYPE} | grep -qi "alpine"; then
+   pidof sshd >/dev/null 2>&1 && exit
```
benoitf: pidof is not working for all linux ?

wsmaster/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/ProcessIsLaunchedChecker.java
tolusha: Yes, it works

wsmaster/che-core-api-agent/src/main/resources/agents/org.eclipse.che.ls.csharp.json
tolusha: It is added automatically 
https://github.com/eclipse/che/blob/d666fbda275a1a4859e3c5f344791b0293e6cea8/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerProcess.java#L98

wsmaster/che-core-api-agent/src/main/resources/agents/scripts/org.eclipse.che.ssh.script.sh
tolusha: It is absent in fedora:23

wsmaster/che-core-api-agent/src/main/resources/agents/scripts/org.eclipse.che.ls.json.script.sh
tolusha: Agreed. Should be fixed

wsmaster/che-core-api-agent/src/main/resources/agents/org.eclipse.che.ls.csharp.json
benoitf: but for example bash is not included on alpine 

```
/ # /bin/bash -c "echo hello"
/bin/sh: /bin/bash: not found
/ # /bin/sh -c "echo hello"
hello

```

wsmaster/che-core-api-agent/src/main/resources/agents/scripts/org.eclipse.che.ssh.script.sh
benoitf: ok 
do we need to use the check on linux version ? or we could check against the availability of the tool ?
if pidof is there use it else use ps else use another tool, etc ?

wsmaster/che-core-api-agent/src/main/resources/agents/org.eclipse.che.ls.csharp.json
tolusha: It is supposed that bash already has been installed on alpine. It is our requirement.
User is able to create workspace based on `codenvy/alpine_jdk8` after the fix.

wsmaster/che-core-api-agent/src/main/resources/agents/scripts/org.eclipse.che.ls.json.script.sh
tolusha: done

benoitf:
FYI I have also tested update script on macOS and I've

```
$ ./update_agents.sh 
sed: illegal option -- r
usage: sed script [-Ealn] [-i extension] [file ...]
       sed [-Ealn] [-i extension] [-e script] ... [-f script_file] ... [file ...]
(like 10 times)
```

wsmaster/che-core-api-agent/src/main/resources/agents/scripts/org.eclipse.che.ssh.script.sh
tolusha: Agree

wsmaster/che-core-api-agent/src/main/resources/agents/org.eclipse.che.ls.csharp.json
benoitf: but it's not true then that it will work on "default alpine image"

what are the specific bash features that we're using ?

tolusha: In term of agents - none. But we rely on it here:

https://github.com/eclipse/che/blob/d666fbda275a1a4859e3c5f344791b0293e6cea8/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerProcess.java#L98

benoitf: ok but we couldn't replace all /bin/bash by {"/bin/sh", "-c", shellCommand}; instead ?
or to have DockerShProcess and DockerBashProcess (to not break some existing commands that use bash)

tolusha: I think we could but it is out of scope of the issue https://github.com/eclipse/che/issues/2424

wsmaster/che-core-api-agent/src/main/resources/agents/scripts/org.eclipse.che.ssh.script.sh
tolusha: Done

wsmaster/che-core-api-agent/src/main/resources/agents/org.eclipse.che.ls.csharp.json
benoitf: yes maybe add another issue for "pure Alpine" because users will start to use "official docker images" or "based images" that are switching to Alpine

for example openjdk:8-alpine has not bash inside

benoitf (approved):
would be great to have update_agents.sh script unix compliant (does not work on macOs)

tolusha:
Could you pls check if it is work on macOs now ?

benoitf:
@tolusha I can launch the script now but I'm not sure it is working

https://gist.github.com/benoitf/f00b487b596bf01ec065f0879c401512

without modifying any .sh script, all json are modified and it seems that \n are not written but it adds space

Issue #2438, opened 2016-09-13
Stop workspaces before user removal


skabashnyuk:
To be able to remove all user workspaces from database we need to stop all his running workspaces.

akorneta:
At current state we decide to reuse logic from `WorkspaceManager` and throw `ConflictException` when the workspace is `running`, we have a common vision of the solution and partial implementation.

bmicklea:
@skabashnyuk is this a must-have for M5?  If not I'll move to M6.

PR #2566, opened 2016-09-23
[WIP]CHE-2438: Add handler for workspace remove event

akorneta:
### What does this PR do?

Adds cascade removal events, subscribers, context, new unsubscribe method to EventService.
#2438
### Tests written?

Yes

@skabashnyuk, @evoevodin please take a look.

### DIFF
--- a/core/che-core-api-jdbc/src/main/java/org/eclipse/che/api/core/jdbc/jpa/event/CascadeRemovalEvent.java
+++ b/core/che-core-api-jdbc/src/main/java/org/eclipse/che/api/core/jdbc/jpa/event/CascadeRemovalEvent.java
@@ -8,24 +8,18 @@
  * Contributors:
  *   Codenvy, S.A. - initial API and implementation
  *******************************************************************************/
-package org.eclipse.che.api.machine.server.event;
-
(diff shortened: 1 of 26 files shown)


skabashnyuk (approved):

core/che-core-api-jdbc/src/main/java/org/eclipse/che/api/core/jdbc/jpa/event/CascadeRemovalEvent.java
```diff
@@ -0,0 +1,29 @@
...
+package org.eclipse.che.api.core.jdbc.jpa.event;
+
+/**
+ * Special event type which is needed only for
+ * notification in the process of cascade removing.
+ *
+ * @author Anton Korneta
+ */
+public abstract class CascadeRemovalEvent {
+    private final RemovalContext context;
```
voievodin: What do you think about using direct field initialization in this way you can get rid of explicitly defined constructor .

``` java
private final RemovalContext context = new RemovalContext();
```

voievodin (approved):

Issue #2560, opened 2016-09-23
CLI Environment Variables


JamesDrummond:
Can we add an argument to the CLI that would list all available environment variables and a short description of each. Maybe we can include this information in che-launcher and run `docker run che-launcher list env`.

TylerJewell:
Resolved in this PR. 
https://github.com/eclipse/che/pull/2575

TylerJewell:
This was part of M4.

PR #2575, opened 2016-09-23
Fixes space issue with profile variables

TylerJewell:
### What does this PR do?

Fixes issue in CLI where default variables with spaces were not read properly from a default profile. It also adds in a fix that if values from a default profile are loaded and passed into che-launcher, they do so with the appropriate double quotes.


TylerJewell:
@benoitf

TylerJewell:
Added this request.
https://github.com/eclipse/che/issues/2560

benoitf (approved):
It may not work if user is exporting its own CHE_\*  variables but it fixes listed issue with default product name

PR #2578, opened 2016-09-23
improve IDE iframe injection

olexii4:
### What does this PR do?

Improve IDE iframe injection

### DIFF
--- a/dashboard/src/app/ide/ide-iframe/ide-iframe.service.ts
+++ b/dashboard/src/app/ide/ide-iframe/ide-iframe.service.ts
@@ -45,16 +45,6 @@ class IdeIFrameSvc {
     }, false);
   }
 
-  addIFrame() {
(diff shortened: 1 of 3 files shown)


akurinnoy (approved):

ashumilova (approved):

benoitf:
I don't see what is the improvement or the reason of the change ? except hardcoding some iframe ?

olexii4:
Now we have iframe dom injection every time when view content is loaded.  Injector service tries to find an existing element with same id and deletes it. Next step - injector  adds a new one iframe element.   We still have  problem with doubled iframe.  Earlier I tried to fix it twice, but the problem still exists(hardly ever).  @eivantsov showed me the new issue with doubled IDE iframe (I checked the version).
That is why i propose this solution as guaranteed one.

benoitf (approved):
@olexii4 ok thanks for the explanation, PR context is more helpful now

Issue #1754, opened 2016-07-16
Create a Che-specific loader to replace the vertical bars


TylerJewell:
This issue is to replace the current loader we have with vertical bars with another one. We would like to find a glowing one that uses the Che logo.
![screen shot 2016-07-14 at 10 39 31](https://cloud.githubusercontent.com/assets/392458/16896215/f7ddb096-4b41-11e6-8c28-d3344660bfae.png)

olexii4:
@TylerJewell  It is not clear for me. Could you tell me ideas about what it should be?

slemeur:
We are discussing it @olexii4 .

slemeur:
@olexii4 You can replace the vertical bar with the Che logo. we will have a small animation like the following.

[loader-idea2.mp4.zip](https://github.com/eclipse/che/files/484531/loader-idea2.mp4.zip)

olexii4:
![anim](https://cloud.githubusercontent.com/assets/6310786/18914016/ed9fb70a-8593-11e6-9803-fc4c67485b38.gif)

olexii4:
2d 4h

PR #2589, opened 2016-09-26
CHE-1754: improve ide loader

olexii4:
### What does this PR do?

Loader Improving  (it changed loader widget content and its styles).
#1754

anim

### DIFF
--- a/dashboard/src/app/ide/ide.html
+++ b/dashboard/src/app/ide/ide.html
@@ -15,12 +15,8 @@
   <div class="ide-page-loader"
        layout="column" layout-align="center center">
     <div ng-hide="ideCtrl.selectedWorkspace">No workspace selected. Unable to open IDE</div>
-    <div layout="column" class="ide-page-loader-content" ng-show="ideCtrl.selectedWorkspace">
-      <div class="spinner">
-        <div class="rect1"></div>
-        <div class="rect2"></div>
-        <div class="rect3"></div>
-      </div>
+    <div class="ide-page-loader-content" ng-show="ideCtrl.selectedWorkspace">
+      <img ng-src="{{branding.loaderURL}}">
     </div>
   </div>
 </div>
--- a/dashboard/src/app/ide/ide.styl
+++ b/dashboard/src/app/ide/ide.styl
@@ -18,6 +18,16 @@
   right 0
   margin auto
 
-.ide-page-loader-content
-  width 120px
-  height 120px
+.ide-page-loader-content img
+  max-width 140px
+  max-height 140px
+  animation-name opacity
+  animation-duration 2s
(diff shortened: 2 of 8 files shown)


dashboard/src/app/ide/ide.html
```diff
@@ -15,12 +15,13 @@
...
     <div ng-hide="ideCtrl.selectedWorkspace">No workspace selected. Unable to open IDE</div>
-    <div layout="column" class="ide-page-loader-content" ng-show="ideCtrl.selectedWorkspace">
-      <div class="spinner">
-        <div class="rect1"></div>
-        <div class="rect2"></div>
-        <div class="rect3"></div>
-      </div>
+    <div class="ide-page-loader-content" ng-show="ideCtrl.selectedWorkspace">
+        <div>
+          <span>Loader idea</span>
```
benoitf: Could you attach an animated gif ?
Because I'm not sure we should display "loader idea" text

benoitf (changes requested):

dashboard/src/app/ide/ide.html
slemeur: We should definitely not show it.

benoitf (approved):
ok for technical part
for UI I let @slemeur

.gitignore
```diff
@@ -24,6 +24,10 @@ test-output/
 *.o
 *.so
 
+# TypeScript built source #
+###########################
+dashboard/src/**/*.js
```
benoitf: I don't have on my side the typescript built source in the same folder so I'm not sure about this exclude ?

ashumilova (approved):

akurinnoy (approved):

.gitignore
olexii4: Sorry. It needed for some built in compiler. It very useful to go through to resources. I have removed this commit.

olexii4:
@benoitf Sorry. It needed for some built in compiler. It very useful to go through to resources. I have removed this commit.

benoitf:
@olexii4 is that builtin compiler could generate in a separate folder ? because I think it should be difficult to have all files in same folders no ?
Also is that sourceMaps isn't working ?

olexii4:
@benoitf  It is configurable. I have removed this commit.

slemeur:
It looks okay for me.
We need to try to use this loader in other areas - where we has the small vertical bars loader.
In this case, I would think that the Che logo must have to be white.

slemeur:
So looking at the changes in the animation, I'm -1 for changing the vertical bar from other places than the splash-screen. (ie, the left navbar and in the IDE)

olexii4:
anim

slemeur (approved):

PR #2599, opened 2016-09-27
CHE-2470: Fix usage of machine servers in environemnt configuration

### LINKED ISSUE #2470: Servers in machine config are not applied to runtime after workspace restart
The attribute **servers** in machine config in environment config is not applied to runtime, after workspace restart.

```
...
servers: {
 1024/tcp: {
   port: "1024",
   properties: { },
   protocol: "http"
}
...
```

**Reproduction Steps:**
1. Add ports in machine configuration
2. Restart workspace
3. Go to Machines perspective
4. 

**Expected behavior:**

The servers tab contains mentioned port.

**Observed behavior:**

The servers tab doesn't contain mentioned port.

garagatyi:
### What does this PR do?

Fixes exposing of servers in machine that were defined in environment configuration.

Fixes #2470 
### Previous behavior

Servers were not exposed.
### New behavior

Servers are exposed.
### PR type
- [ ] ~~Major change = changes existing features or docs~~
### Minor change checklist
- [ ] ~~New API required?~~
- [ ] ~~API updated~~
- [ ] Tests passed

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
@@ -32,6 +32,7 @@ import java.util.Map;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static java.lang.String.format;
+import static java.util.stream.Collectors.toList;
 
 /**
  * Parses {@link Environment} into {@link CheServicesEnvironmentImpl}.
@@ -40,6 +41,12 @@ import static java.lang.String.format;
  */
 public class EnvironmentParser {
     private static final List<String> types = Arrays.asList("compose", "dockerimage", "dockerfile");
+
+    // TODO move to container related code
+    protected static final String SERVER_CONF_LABEL_PREFIX          = "che:server:";
+    protected static final String SERVER_CONF_LABEL_REF_SUFFIX      = ":ref";
+    protected static final String SERVER_CONF_LABEL_PROTOCOL_SUFFIX = ":protocol";
+    protected static final String SERVER_CONF_LABEL_PATH_SUFFIX     = ":path";
 
     private final ComposeFileParser composeFileParser;
     private final RecipeDownloader  recipeDownloader;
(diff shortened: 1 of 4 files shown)


garagatyi:
@skabashnyuk @evoevodin @sleshchenko @akorneta Please review

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
```diff
@@ -78,36 +85,79 @@ public CheServicesEnvironmentImpl parse(Environment environment) throws IllegalA
...
+                normalizeMachine(name, service, extendedMachine);
             }
         });
 
-        return composeEnvironment;
+        return cheServicesEnvironment;
+    }
+
+    private void normalizeMachine(String name, CheServiceImpl service, ExtendedMachine extendedMachine) {
+        if (extendedMachine.getAttributes() != null &&
```
sleshchenko: As far I as know our POJO and DTO objects should return empty collections instead of null values

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
```diff
@@ -78,36 +85,79 @@ public CheServicesEnvironmentImpl parse(Environment environment) throws IllegalA
...
+            extendedMachine.getAttributes().containsKey("memoryLimitBytes")) {
+
+            try {
+                service.setMemLimit(Long.parseLong(extendedMachine.getAttributes().get("memoryLimitBytes")));
+            } catch (NumberFormatException e) {
+                throw new IllegalArgumentException(
+                        format("Value of attribute 'memoryLimitBytes' of machine '%s' is illegal", name));
+            }
+        }
+        if (service.getExpose() != null) {
```
sleshchenko: the same to https://github.com/eclipse/che/pull/2599/files#r80632123
CheServiceImpl#getExpose returns new ArrayList<String>() instead of null https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/model/CheServiceImpl.java#L285-290

sleshchenko:
Looks like there is typo in your email (agaragatyi@codenvy.comqa) in committer and author information. Please fix it

akorneta:
ok

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
```diff
@@ -78,36 +85,79 @@ public CheServicesEnvironmentImpl parse(Environment environment) throws IllegalA
...
+            service.setExpose(service.getExpose()
+                                     .stream()
+                                     .map(expose -> expose.contains("/") ?
```
voievodin: i'd probably change it to 

``` java
service.getExpose()
       .stream()
       .filter(expose -> !expose.contains("/"))
       .map(expose -> expose + "/tcp")
       .collect(toList())
```

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
```diff
@@ -78,36 +85,79 @@ public CheServicesEnvironmentImpl parse(Environment environment) throws IllegalA
...
+                                                    expose :
+                                                    expose + "/tcp")
+                                     .collect(toList()));
+        }
+        if (extendedMachine.getServers() != null) {
+            extendedMachine.getServers().forEach((serverRef, serverConf) -> {
```
voievodin: Why doesn't the native foreach work here?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
```diff
@@ -78,36 +85,79 @@ public CheServicesEnvironmentImpl parse(Environment environment) throws IllegalA
...
+                String normalizedPort = serverConf.getPort().contains("/") ?
+                                        serverConf.getPort() :
+                                        serverConf.getPort() + "/tcp";
+
+                service.getExpose().add(normalizedPort);
+
+                service.getLabels().put(SERVER_CONF_LABEL_PREFIX +
```
voievodin: You check all the object not to be nulls, should you check labels as well?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
```diff
@@ -78,36 +85,79 @@ public CheServicesEnvironmentImpl parse(Environment environment) throws IllegalA
...
+                                        normalizedPort +
+                                        SERVER_CONF_LABEL_REF_SUFFIX,
+                                        serverRef);
+                if (serverConf.getProperties() != null && serverConf.getProperties().get("path") != null) {
+
+                    service.getLabels().put(SERVER_CONF_LABEL_PREFIX +
```
voievodin: It looks like  `SERVER_CONF_LABEL_PREFIX + normalizedPort` deserves a separate variable as the concatenation will be repeated 3 times in a raw if properties and protocol exist

voievodin (changes requested):

voievodin (approved):

akorneta (approved):
ok

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
garagatyi: Yes, but this object may come from Everrest as a proxy object

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
garagatyi: Yes, but this object may come from Everrest as a proxy object

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
garagatyi: But what if some expose values already contains /tcp. Will they be removed?

voievodin: Oh yes, sorry, i missed that you still need the other exposes

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
garagatyi: Sorry, I didn't understand what you mean.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
garagatyi: labels are part of non-rest code, so it returns empty collections always and proxy doesn't affect it

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
garagatyi: I think it is OK to leave that code with this construct instead of common loop over entries

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
garagatyi: ok

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
garagatyi: Actually this class uses model interface, so it is not guarantied that DTO or default impl is used

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
sleshchenko: As far as I understand you create this instance here https://github.com/eclipse/che/pull/2599/files#diff-0adf8dd72b68fcf12958c9f8602151e0R209

garagatyi: changed

PR #2606, opened 2016-09-27
Publish appropriate event before user is persisted.

mshaposhnik:
### What does this PR do?

Adds event publishing before user is persisted in the DB layer.

Needed for CODENVY-398;
### PR type
- [ X ] Minor change = no change to existing features or docs
- [   ] Major change = changes existing features or docs
### Minor change checklist
- [ – ] New API required
- [ – ] API updated
- [ – ] Tests provided 
- [ + ] Tests passed


skabashnyuk (approved):

skabashnyuk (approved):

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/jpa/UserEntityListener.java
```diff
@@ -33,4 +35,9 @@
     public void preRemove(UserImpl user) {
         eventService.publish(new BeforeUserRemovedEvent(user));
     }
+
+    @PrePersist
+    public void postPersist(UserImpl user) {
```
sleshchenko: I think you should rename it to `prePersist`

skabashnyuk (approved):

Issue #2528, opened 2016-09-21
Failed to restart workspace with renamed environment


ashumilova:
**Reproduction Steps:**
1. Rename environment
2. Restart workspace

**Expected behavior:**

The workspace successfully restarted

**Observed behavior:**

Error message occurs.

![pasted_image_at_2016_09_21_03_16_pm](https://cloud.githubusercontent.com/assets/1611939/18711181/5ec4aec8-8011-11e6-8a59-9828db71bb9e.png)

PR #2613, opened 2016-09-27
CHE-3017: Improve updating of environments

mmorhun:
### What does this PR do?
Handles environment snapshots on environments update.
Adds ability to rename specified environment. 
Fixes bug, when workspace with snapshot fails to start after renaming of environment.

#3017 
#2528

### Previous behavior
On workspace update new configs of environments just overwrites old ones.
Workspace with snapshot fails to start after renaming of the environment.

### New behavior
Updates snapshots on environments rename.
Removes / updates snapshots on environment update.
Always deletes snapshots on environment deletion.
Adds ability to rename specified environment.
Workspace with snapshots can start after renaming of environment.

New tests have been written.


mmorhun:
@garagatyi @skabashnyuk review please

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -839,4 +856,26 @@ private WorkspaceImpl getByKey(String key) throws NotFoundException, ServerExcep
...
+     *         original workspace
+     * @param update
+     *         update workspace request
+     * @throws ConflictException
+     *         if some updates cannot be applied now
+     */
+    private void ensureRuntimeInfoNotChangedIfWorkspaceRunning(String id, Workspace workspace, Workspace update) throws ConflictException {
+        if (runtimes.hasRuntime(id)) { // we can change any parameter in stopped workspace
+            if (!workspace.getConfig().getDefaultEnv().equals(update.getConfig().getDefaultEnv())) {
+                throw new ConflictException("Cannot affect name of runtime when workspace is running");
```
benoitf: it is missing a test of this check

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -760,6 +763,20 @@ private String replaceSnapshot(MachineImpl machine, String namespace) {
         }
     }
 
+    private void updateWorkspaceSnapshots(String id, Workspace workspace, Workspace update) throws SnapshotException {
+        if (!workspace.getConfig().getDefaultEnv().equals(update.getConfig().getDefaultEnv())) {
+            for (SnapshotImpl snapshot : snapshotDao.findSnapshots(id)) {
+                try {
+                    snapshotDao.removeSnapshot(snapshot.getId());
```
garagatyi: I think it should be replaced with update of snapshot. I know that snapshot dao miss that method

garagatyi: Whole new functionality should be well covered by tests

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -760,6 +763,20 @@ private String replaceSnapshot(MachineImpl machine, String namespace) {
         }
     }
 
+    private void updateWorkspaceSnapshots(String id, Workspace workspace, Workspace update) throws SnapshotException {
+        if (!workspace.getConfig().getDefaultEnv().equals(update.getConfig().getDefaultEnv())) {
```
garagatyi: What if workspace started with non-default env?

garagatyi (changes requested):

mmorhun:
ci-build

mmorhun:
@garagatyi @evoevodin review please

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/jpa/JpaSnapshotDao.java
```diff
@@ -132,6 +132,17 @@ public void removeSnapshot(String snapshotId) throws NotFoundException, Snapshot
         }
     }
 
+    @Override
+    public SnapshotImpl updateSnapshot(SnapshotImpl update) throws SnapshotException {
+        requireNonNull(update, "Required non-null snapshot");
+        requireNonNull(update.getId(), "Required non-null snapshot id");
+        try {
+            return doUpdate(update);
```
evoevodin: handle `DuplicateKeyException`

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/jpa/JpaSnapshotDao.java
```diff
@@ -132,6 +132,17 @@ public void removeSnapshot(String snapshotId) throws NotFoundException, Snapshot
...
+        } catch (RuntimeException x) {
```
voievodin, line 146: handle `DuplicateKeyException`

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/jpa/JpaSnapshotDao.java
```diff
@@ -161,4 +172,10 @@ protected void doRemove(String snapshotId) throws NotFoundException {
         newSnapshots.forEach(manager::persist);
         return existing;
     }
+
+    @Transactional
+    protected SnapshotImpl doUpdate(SnapshotImpl update) {
+        return managerProvider.get().merge(update);
```
voievodin: 1. `update` may not be managed by jpa
2. if `update` is managed, it may be either inserted or updated, consider using get + merge
3. Usually update methods throw `NotFoundException` when there is no entity with given identifier

voievodin (changes requested):
- We usually perform been validation on service level
- Can you please explain what is the reason to have 3 additional methods for environment management in `WorkspaceManager`, is it possible to achive the same behaviour in workspace update method?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -523,10 +523,8 @@ public WorkspaceDto addEnvironment(@ApiParam("The workspace id")
         requiredNotNull(newEnvironment, "New environment");
         requiredNotNull(envName, "New environment name");
         relativizeRecipeLinks(newEnvironment);
-        final WorkspaceImpl workspace = workspaceManager.getWorkspace(id);
-        workspace.getConfig().getEnvironments().put(envName, new EnvironmentImpl(newEnvironment));
-        validator.validateConfig(workspace.getConfig());
-        return linksInjector.injectLinks(asDto(workspaceManager.updateWorkspace(id, workspace)), getServiceContext());
+        workspaceManager.addEnvironment(id, envName, newEnvironment);
+        return linksInjector.injectLinks(asDto(workspaceManager.getWorkspace(id)), getServiceContext());
```
garagatyi: please elaborate on why environment validation here and further should not be performed

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -547,20 +545,21 @@ public WorkspaceDto updateEnvironment(@ApiParam("The workspace id")
...
-                                                                        ForbiddenException {
+                                          EnvironmentDto update,
+                                          @ApiParam("New name of the environment")
+                                          @QueryParam("newName")
+                                          String newEnvName) throws ServerException,
+                                                                    BadRequestException,
+                                                                    NotFoundException,
+                                                                    ConflictException,
+                                                                    ForbiddenException {
         requiredNotNull(update, "Environment description");
```
garagatyi, line 569: Please also check new name for emptiness

garagatyi (commented):
Please add tests of WorkspaceService for environment renaming

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -571,6 +787,31 @@ public void removeSnapshots(String workspaceId) throws NotFoundException, Server
...
+     * @param workspaceId
+     *         workspace id to remove environment snapshots
+     * @param envName
+     *         environment name to remove snapshots from
+     * @throws NotFoundException
+     *           when workspace with given id doesn't exists
+     * @throws ServerException
+     *          when any other error occurs
+     */
+    public void removeEnvironmentSnapshots(String workspaceId, String envName) throws NotFoundException, ServerException {
```
garagatyi: Why this method is public?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -571,6 +787,31 @@ public void removeSnapshots(String workspaceId) throws NotFoundException, Server
...
+        getSnapshot(workspaceId).stream()
+                                .filter(snapshot -> snapshot.getEnvName().equals(envName))
```
garagatyi: Is it probable to find more than 1 entry? If not then filter doesn't suit this situation well

garagatyi (changes requested):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
mmorhun: ok, let's make it public when needed

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
mmorhun: Environment can contain few machines and in this case we'll have snapshot per machine.

mmorhun:
@garagatyi @evoevodin you are right validation should be in `WorkspaceService`.

mmorhun:
@evoevodin 
> Can you please explain what is the reason to have 3 additional methods for environment management in WorkspaceManager, is it possible to achive the same behaviour in workspace update method?

The original problem here is to determine one correspondence between old and new environment on a workspace update. To resolve this, I assume, that name of environment will be something like its id during update. And with this assumption we can determine which environments should be added, removed or updated. But if we want to rename environment this strategy won't work. To avoid this, we should have mechanisms to work with specified environment which needn't its name as id or can resolve it somehow. More, with this methods we needn't to perform always deep analyzing of workspace, for example, on environment deletion.

voievodin (changes requested):
Please answer my questions

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -490,6 +539,175 @@ public void stopWorkspace(String workspaceId, @Nullable Boolean createSnapshot)
...
+     * @throws IllegalArgumentException
+     *         when environment configuration is not valid
+     * @throws NotFoundException
+     *         when workspace with specified ID doesn't exist
+     * @throws ConflictException
+     *         when environment with given name already exists
+     * @throws ServerException
+     *         when other error occurs
+     */
+    public void addEnvironment(String workspaceId, String newEnvName, Environment newEnvironment) throws NotFoundException,
```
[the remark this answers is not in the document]
mmorhun: @evoevodin I don't see any reasons to not to have such method here. Lets analyze this.
> It can be simply done by `WorkspaceManager#update`

Yes it can. But we'll have overhead in this case. Look, just for add one environment to workspace we need to validate all existing (and already valid) environments. More, we need to analyze update to determine added, updated and deleted environments - it is expensive.

> It performs validation while it should not

I added validation here because methods are public and theoretically might be invoked from any place. But looking on `WorkspaceManager#update` method I see, that it doesn't perform validation (except checks for null). So, I can remove validation from here and hope that this will be invoked only from `WorkspaceService` as it done for workspace update.

> It makes weird things like, normalizeState(workspaceDao.update(workspace)) what is the reason to inject runtime information into the instance which is not used by anyone ?

You are right, I have to optimize this code. Thx.

> It doesn't refresh workspace attribute `#UPDATED_ATTRIBUTE_NAME`

It can be easily added. I see no problem here.

> the environment object is validated twice

Please see my second replay.

> Moreover the service makes `addEnvironment` call and then immediately calls `WorkspaceManager#getWorkspace`

What is the problem? By the way, what is the reason to return whole workspace on add environment?

> in service it looks simpler and requires less operations and code changes

I doubt, that it requires less operations. And service should validate data and delegate any logic to (in our case) managers, am I right?

mmorhun:
@garagatyi @evoevodin I updated PR. Review please.

wsmaster/wsmaster-local/src/main/java/org/eclipse/che/api/local/LocalSnapshotDaoImpl.java
```diff
@@ -121,6 +122,13 @@ public synchronized void removeSnapshot(String snapshotId) throws NotFoundExcept
     }
 
     @Override
+    public SnapshotImpl updateSnapshot(SnapshotImpl update) throws NotFoundException, ConflictException, SnapshotException {
+        requireNonNull(update, "Required non-null snapshot");
```
voievodin: Please throw unsupported operation exception instead, this dao is not used. We're going to get rid of it.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -555,18 +555,28 @@ public WorkspaceDto updateEnvironment(@ApiParam("The workspace id")
...
                                           String envName,
                                           @ApiParam(value = "The environment update", required = true)
-                                          EnvironmentDto update) throws ServerException,
-                                                                        BadRequestException,
-                                                                        NotFoundException,
-                                                                        ConflictException,
-                                                                        ForbiddenException {
+                                          EnvironmentDto update,
+                                          @ApiParam("New name of the environment")
+                                          @QueryParam("new-name")
```
evoevodin: Can you please explain why do we need this qp, it seems that everything should be fine with incoming environment

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -310,13 +314,67 @@ public WorkspaceImpl updateWorkspace(String id, Workspace update) throws Conflic
...
+                                                         .stream()
+                                                         .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+
+        for (Map.Entry<String, Environment> env : currentDiff.entrySet()) {
+            Environment envValue = env.getValue();
+            if (Collections.frequency(currentDiff.values(), envValue) == 1 &&
+                Collections.frequency(updateDiff.values(), envValue) == 1 &&
+                !updateDiff.containsKey(env.getKey())) {
+                try {
+                    updateEnvironmentSnapshots(workspaceId, env.getKey(), updateDiff.entrySet()
```
voievodin, line 354: It looks like it's possible to get this value while checking "frequency"
```java
entrySet() 
    .stream()
    .filter(e -> e.getValue().equals(envValue))
    .findAny()
    .orElse(null) // should never happens
    .getKey()
```

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -310,13 +314,67 @@ public WorkspaceImpl updateWorkspace(String id, Workspace update) throws Conflic
...
+                                                                                    .stream()
+                                                                                    .filter(e -> e.getValue().equals(envValue))
+                                                                                    .findAny()
+                                                                                    .orElse(null) // should never happens
+                                                                                    .getKey());
+                } catch (ConflictException | SnapshotException | NotFoundException err) {
+                    LOG.error("Failed to update snapshots of '%s' environment in '%s' workspace.", env.getKey(), workspaceId, err);
+                    safeRemoveEnvironmentSnapshots(workspaceId, env.getKey());
```
voievodin: i wouldn't use _safe_ for such kind of method it's usually called called _actionQuietly_ or just _action_.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -571,6 +629,46 @@ public void removeSnapshots(String workspaceId) throws NotFoundException, Server
...
+     * @param envName
+     *         environment name to remove snapshots from
+     * @throws NotFoundException
+     *           when workspace with given id doesn't exists
+     * @throws ServerException
+     *          when any other error occurs
+     */
+    @VisibleForTesting
+    void removeEnvironmentSnapshots(String workspaceId, String envName) throws NotFoundException, ServerException {
+        getSnapshot(workspaceId).stream()
```
voievodin: i wouldn't use stream api in this case, as all you need is a regular foreach with if statement

voievodin (approved):
DB part looks good.
I would't say that i would implement snapshot updating logic in exactly the same manner, but as long as it works and fits the requirements it's fine. 

@skabashnyuk can you please also review the code

Please don't merge it before #3405

wsmaster/wsmaster-local/src/main/java/org/eclipse/che/api/local/LocalSnapshotDaoImpl.java
mmorhun: ok

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
mmorhun: Agree

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
mmorhun: Yes, it is possible, but I think that it will make the code less readable and ugly.
As about performance, usually we have small number of environments, so it won't be a problem. In case of only one environment this code won't be invoked.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -551,6 +612,50 @@ public void removeSnapshots(String workspaceId) throws NotFoundException, Server
...
+     *         workspace id to remove environment snapshots
+     * @param envName
+     *         environment name to remove snapshots from
+     * @throws NotFoundException
+     *           when workspace with given id doesn't exists
+     * @throws ServerException
+     *          when any other error occurs
+     */
+    @VisibleForTesting
+    void removeEnvironmentSnapshots(String workspaceId, String envName) throws NotFoundException, ServerException {
```
garagatyi: Since this method is used in method **removeEnvironmentSnapshotsQuietly** only please move it under that method to ease code navigation

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -555,18 +557,28 @@ public WorkspaceDto updateEnvironment(@ApiParam("The workspace id")
...
+                                                                    ForbiddenException {
         requiredNotNull(update, "Environment description");
         relativizeRecipeLinks(update);
         final WorkspaceImpl workspace = workspaceManager.getWorkspace(id);
-        EnvironmentImpl previous = workspace.getConfig().getEnvironments().put(envName, new EnvironmentImpl(update));
+        EnvironmentImpl previous = workspace.getConfig().getEnvironments().remove(envName);
         if (previous == null) {
             throw new NotFoundException(format("Workspace '%s' doesn't contain environment '%s'", id, envName));
         }
+        if (newEnvName != null && !"".equals(newEnvName)) {
```
garagatyi: isNullOrEmpty?

wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/workspace/server/WorkspaceManagerTest.java
```diff
@@ -812,6 +812,154 @@ public void shouldBeAbleToGetMachineInstanceIfWorkspaceIsStarting() throws Excep
...
+
+        verify(workspaceManager, never()).removeEnvironmentSnapshots(eq(workspace.getId()), eq(workspace.getConfig().getDefaultEnv()));
+        verify(workspaceManager, never()).updateEnvironmentSnapshots(eq(workspace.getId()), eq(workspace.getConfig().getDefaultEnv()), any());
+        verify(workspaceManager).removeEnvironmentSnapshots(eq(workspace.getId()), eq("env1"));
+        verify(workspaceManager).updateEnvironmentSnapshots(eq(workspace.getId()), eq("env2"), eq("envX"));
+        verify(workspaceManager).removeEnvironmentSnapshots(eq(workspace.getId()), eq("env3"));
+        verify(workspaceManager, never()).updateEnvironmentSnapshots(eq(workspace.getId()), any(), eq("envZ"));
+    }
+
+    @Test
```
garagatyi, line 963: Please cover all the code you've changed

wsmaster/wsmaster-local/src/main/java/org/eclipse/che/api/local/LocalSnapshotDaoImpl.java
```diff
@@ -121,6 +122,11 @@ public synchronized void removeSnapshot(String snapshotId) throws NotFoundExcept
     }
 
     @Override
+    public SnapshotImpl updateSnapshot(SnapshotImpl update) throws NotFoundException, ConflictException, SnapshotException {
```
garagatyi, line 125: Tests?

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/jpa/JpaSnapshotDao.java
```diff
@@ -119,6 +120,21 @@ public void removeSnapshot(String snapshotId) throws NotFoundException, Snapshot
     }
 
     @Override
+    public SnapshotImpl updateSnapshot(SnapshotImpl update) throws NotFoundException, ConflictException, SnapshotException {
```
garagatyi, line 123: Tests?

garagatyi (changes requested):

wsmaster/wsmaster-local/src/main/java/org/eclipse/che/api/local/LocalSnapshotDaoImpl.java
mmorhun: Do we need tests here?

TylerJewell:
Pushed out of 5.0.  Requires a rewrite according to architects.

mmorhun:
Taking into account, that new solution will be different, this one is not needed any more :( .

PR #2643, opened 2016-09-29
Build che server Docker image with CentOS base image

dharmit:
### What does this PR do?

Build Che Server Docker image on CentOS using [CentOS Community Container Pipeline](https://github.com/CentOS/container-pipeline-service/).
### New behavior

Che Server image can now be built on top of CentOS base image. 
### PR type
### Minor change checklist
- [ ] New API required?
- [ ] API updated
- [ ] Tests passed
### Major change checklist
- [ ] New API required?
- [ ] API updated
- [ ] Tests passed

### DIFF
--- a/Dockerfile.centos
+++ b/Dockerfile.centos
@@ -1,0 +1,53 @@
+# Copyright (c) 2012-2016 Red Hat, Inc
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License v1.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-v10.html
+#
+# Contributors:
+#   Dharmit Shah  - Initial implementation
+#   Mario Loriedo - Improvements
+#
+# To build it, run in the repository root:
+#  `docker build -t registry.centos.org/eclipse/che-server -f Dockerfile.centos .`
+#
+# To run it:
+#  docker run --net=host \
+#             --name che \
+#             -v /var/run/docker.sock:/var/run/docker.sock \
+#             -v /home/user/che/lib:/home/user/che/lib-copy \
+#             -v /home/user/che/workspaces:/home/user/che/workspaces \
+#             -v /home/user/che/storage:/home/user/che/storage \
+#             registry.centos.org/eclipse/che-server
+#           
+FROM registry.centos.org/centos/centos:latest
+
+ENV LANG=C.UTF-8 \
+    JAVA_HOME=/usr/lib/jvm/jre-1.8.0 \
+    PATH=${PATH}:${JAVA_HOME}/bin \
+    CHE_HOME=/home/user/che \
+    DOCKER_VERSION=1.6.0 \
+    DOCKER_BUCKET=get.docker.com
+
+RUN yum -y update && \
+    yum -y install openssl java sudo && \
+    curl -sSL "https://${DOCKER_BUCKET}/builds/Linux/x86_64/docker-${DOCKER_VERSION}" -o /usr/bin/docker && \
+    chmod +x /usr/bin/docker && \
+    yum -y remove openssl && \
+    yum clean all
+RUN useradd user -d /home/user -s /bin/bash -u 1000 -G root && \
+    groupadd docker -g 101 && \
+    usermod -aG ftp,users,docker user && \
+    echo "%root ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers && \
+    sed -i 's/Defaults    requiretty/#Defaults    requiretty/g' /etc/sudoers && \
+    rm -rf /tmp/* /var/cache/yum
+
+EXPOSE 8000 8080
+
+USER user
+
+ADD assembly/assembly-main/target/eclipse-che-*/eclipse-che-* /home/user/che/
+
+ENTRYPOINT [ "/home/user/che/bin/che.sh", "-c" ]
+
+CMD [ "run" ]
--- a/.cccp.yml
+++ b/.cccp.yml
@@ -1,0 +1,7 @@
+# This is file is needed to include the project in the CentOS Container
+# Pipeline main index. It can be used to set the image name (using job-id), set
+# the test script and/or build script and whether to perform or skip the
+# user-defined tests. More information on cccp.yml file can be found on:
+# https://github.com/CentOS/container-index#the-cccpyml-file
+
+job-id: che-server


dharmit:
cc @l0rd

Dockerfile.centos
```diff
@@ -0,0 +1,58 @@
...
+    usermod -aG ftp,users,docker user && \
+    echo "%root ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers && \
+    sed -i 's/Defaults    requiretty/#Defaults    requiretty/g' /etc/sudoers && \
+    rm -rf /tmp/* /var/cache/yum
+
+EXPOSE 8000 8080
+
+USER user
+
+ENV CHE_ARTIFACT_URL="https://install.codenvycorp.com/che/eclipse-che-nightly.tar.gz"
```
benoitf: as this Dockerfile is not using current assembly binary but downloading it, should this PR be moved to https://github.com/eclipse/che-dockerfiles ?

Dockerfile.centos
```diff
@@ -0,0 +1,58 @@
...
+
+RUN curl -sSL "$CHE_ARTIFACT_URL" -o /home/user/eclipse-che.tar.gz
+RUN tar -xzf /home/user/eclipse-che.tar.gz -C /home/user/
+RUN rm /home/user/eclipse-che.tar.gz
+RUN mv /home/user/eclipse-che-* /home/user/che
```
benoitf: to avoid several layers, RUN instructions should be merged into one (and tar command should be for example piped to curl (no need to remove at the end, etc)

l0rd (commented):
@benoitf @TylerJewell CentOS container pipeline is already setup to build this image. As a result we can now:

``` sh
docker pull http://registry.centos.org/eclipse/che-server
docker pull http://registry.centos.org/eclipse/che-launcher
```

Even if the Dockerfile is not 100% completed (the assembly binary is downloaded and not copied form local folder) we thought it would be better to move it from @dharmit fork to here. What do you think?

TylerJewell:
Can you explain the connection to openjdk and updating this dockerfile to have a build context similar to the default one we currently have?  I don't see the dependency.

The build context just calls for an assembly directory to be present and added into the built image.  The dockerfile shouldn't have an opinion about how that assembly wells previously built.

l0rd:
@TylerJewell we are splitting the CI for CentOS container pipeline in 3 steps:

1) Build of Che from source code (with OpenJDK and nodejs 4 or 6)
2) Build the che-server Docker image and publish it 
3) Deploy Che on an OpenShift cluster for testing 

We decided to start with steps  2) and 3) and leave 1) as last step. That's the reason why we download the binary: we don't have an automated build that generates it.

By the way the more I think about that the more I think we should provide the right Dockerfile from the beginning. It does not make sense to commit a Dockerfile that we know we are going to change in a couple of weeks. We should modify Dockerfile.centos to copy the assembly binary from the local folder. And implement a workaround in our CI until step 1) is completed. A couple of possible workarounds:
- build Che with current default tools/version (i.e. Oracle JDK and nodejs012) or
- download the binary in folder /assembly/assembly-main/ before running `docker-build` 
  @dharmit what do you think? Would it be possible for you do these changes?

TylerJewell:
Yes, that was my request as well.  Let's leave the Dockerfile with the syntax of using local assembly.  So I think both of your options are suitable as a way to solve your goals.  And then this way the Dockerfile for different OS can remain consistent.  I support that as part of this, we should rename the base Dockerfile to also have a .ubuntu extension?

cccp.yml
```diff
@@ -0,0 +1 @@
+job-id: che-server
```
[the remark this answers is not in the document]
dharmit: @l0rd Yes, this is required to build the image in CentOS pipeline. It's referenced [here](https://github.com/CentOS/container-index/blob/master/index.d/eclipse.yml#L70).

benoitf:
@TylerJewell base Dockerfile is using alpine so .ubuntu extension is not a candidate. so you mean Dockerfile.alpine ? BTW I think it's easier to have minimal image (based on alpine) using name Dockerfile (this name is also handled in editors, etc for syntax highlighting)

Dockerfile.centos
```diff
@@ -0,0 +1,58 @@
...
+# are made available under the terms of the Eclipse Public License v1.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-v10.html
+#
+# Contributors:
+#   Dharmit Shah  - Initial implementation
+#   Mario Loriedo - Improvements
+#
+# To build it, run in the repository root:
+#  `docker build -t registry.centos.org/eclipse/che-server -f Dockerfile.centos .`
```
benoitf, line 12: do we need to have registry.centos.org/ when trying to build the image ?

dharmit: @benoitf You don't need to have it. It's already up and running on the CentOS infra.

Dockerfile.centos
```diff
@@ -0,0 +1,58 @@
...
+# To run it:
+#  docker run --net=host \
+#             --name che \
+#             -v /var/run/docker.sock:/var/run/docker.sock \
+#             -v /home/user/che/lib:/home/user/che/lib-copy \
+#             -v /home/user/che/workspaces:/home/user/che/workspaces \
+#             -v /home/user/che/storage:/home/user/che/storage \
+#             registry.centos.org/eclipse/che-server
+#           
+FROM registry.centos.org/centos/centos:latest
```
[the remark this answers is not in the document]
dharmit: They're the same.

Dockerfile.centos
benoitf: I mean, in the comment, why do we need to specify the registry prefix, as user wanted to try doesn't need it

Dockerfile.centos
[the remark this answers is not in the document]
benoitf: why we don't specify FROM centos:latest then ?

Dockerfile.centos
dharmit: Right, if it's being built locally, a user probably doesn't need it. But for che-launcher image to be able to find it properly on local setup, it would be a good idea to name/tag it that way. Let me know if that understanding is incorrect. 

Also, I've taken a cue from the build statement [here](https://github.com/eclipse/che/blob/master/Dockerfile#L12).

Dockerfile.centos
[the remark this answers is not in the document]
dharmit: Because the registry mentioned here is local to the pipeline. That helps reduce the build time and saves network resources as well.

TylerJewell:
Can you explain what it means to be local to the pipeline, please?  I am not understanding that syntax and the differences between having a fully explicit registry vs the default centos in the FROM.

Florent - you are right about alpine.  So dockerfiles with alpine get no extension and dockerfiles from other base images do get an extension.

I noticed that haproxy and nginx images once tagged append -alpine to the version tag for alpine images and nothing for other images types.

dharmit:
@TylerJewell 

> Can you explain what it means to be local to the pipeline, please? I am not understanding that syntax and the differences between having a fully explicit registry vs the default centos in the FROM.

It's in the local network. Registry and the systems on which images get built are in the same network. 

As far as the differences are concerned, CentOS registry is working on providing features of scanning and testing the images before they're pushed to registry. Also, registry.centos.org is canonical upstream for container workloads for CentOS project.

Dockerfile.centos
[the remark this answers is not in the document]
benoitf: @dharmit but for users wanted to use this Dockerfile, they are not local to the pipeline.

Dockerfile.centos
benoitf: on a local setup, there is a way to specify the name of the local image for example

benoitf: Else it would act for me, like storing in eclipse/che repository a file that is internal to registry.centos.org and use privately

Dockerfile.centos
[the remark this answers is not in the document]
benoitf: Else it would act for me, like storing in eclipse/che repository a file that is internal to registry.centos.org and used privately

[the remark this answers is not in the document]
dharmit: @benoitf 

> but for users wanted to use this Dockerfile, they are not local to the pipeline

Agreed. Instead of using the Docker hub, it's pulling from registry.centos.org.

> Else it would act for me, like storing in eclipse/che repository a file that is internal to registry.centos.org and used privately

Nothing's getting used privately. Anyone can pull from registry.centos.org. Also, anyone can build the image using the exact same Dockerfile on their local system.

If that doesn't answer your question, I'm not sure I understand your concern here.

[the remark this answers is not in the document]
l0rd: @benoitf @TylerJewell  registry.centos.org is a public registry of Docker images (as the Docker hub). Since it's maintained from CentOS team it's a better source for CentOS base images than docker hub is (e.g. images there are available earlier). Plus we are going to publish che-server:xx-centos on registry.centos.org so it makes more sense to use it as source in this case.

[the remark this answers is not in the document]
TylerJewell: My question about what you wanted to do with the pipeline makes sense.  I think the only fixes remaining are:
1. Adjust the Dockerfile so that it pulls from a local assembly that is built instead of downloading from nightly the assembly.
2. Comine the RUN commands into a single RUN command to remove layers.

Then we should be ok to merge.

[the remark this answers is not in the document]
dharmit: @l0rd @TylerJewell does that look OK now?

Dockerfile.centos
```diff
@@ -0,0 +1,53 @@
...
+
+ENV LANG=C.UTF-8 \
+    JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.102-1.b14.el7_2.x86_64/jre \
```
benoitf: as you're installing "java" with yum
how can you be sure that JAVA_HOME will be installed in this folder ?
if there are updates on JAVA in centos latest image, it will break, no ?

Dockerfile.centos
```diff
@@ -0,0 +1,53 @@
...
+    PATH=${PATH}:${JAVA_HOME}/bin \
+    CHE_HOME=/home/user/che \
+    DOCKER_VERSION=1.6.0 \
+    DOCKER_BUCKET=get.docker.com
+
+RUN yum -y update update && \
+    yum -y install ca-certificates curl openssl java sudo && \
```
benoitf: if you install ca-certificates and curl and openssl to only download docker, it should be removed after docker has been downloaded to reduce image size.

Dockerfile.centos
```diff
@@ -0,0 +1,53 @@
...
+    sed -i 's/Defaults    requiretty/#Defaults    requiretty/g' /etc/sudoers && \
+    rm -rf /tmp/* /var/cache/yum
+
+EXPOSE 8000 8080
+
+USER user
+
+ADD assembly/assembly-main/target/eclipse-che-*/eclipse-che-* /home/user/che/
+
+ENV CHE_HOME /home/user/che
```
benoitf: this ENV instruction can be added to the existing ENV command at the top of the file

cccp.yml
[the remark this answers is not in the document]
benoitf: I don't see why a file needs to be added in a repository for building it on a CI server.
If each CI needs to add config files in the repository, it will be a mess.

Dockerfile.centos
benoitf: I still don't get why default image would be registry.centos.org/eclipse/che-server in this instruction as we can build on our own and then it will be local and we could use it on any registry. So registry part is not required

l0rd: @benoitf I think that's to be consistent with the comments on other Dockerfiles https://github.com/eclipse/che/blob/master/Dockerfile#L21. The registry and repository correspond to the ones used by the CI. And that's still a command that a user can execute safely on its machine. If a user want to push the resulting image (to the docker hub or to another registry) he will need to re-tag it anyway with his <registry>/<repo>/che-server.

benoitf: to be more explicit. Dockerfile.centos is producing a drop-in replacement for Dockerfile. so the image name produced should be the same.
It's the CI that may deploy it on its own registry by adding the registry prefix.
But all Dockerfiles should produce the same image name. As they're alternatives.

Dockerfile.centos
l0rd: @benoitf I agree with you on that but I'm not sure there is a better option. @dharmit can we specify the `JAVA_HOME` path in centos without being so explicit about the JDK version? Like [it's done for the alpine image](https://github.com/eclipse/che/blob/master/Dockerfile#L26).

Dockerfile.centos
```diff
@@ -0,0 +1,53 @@
...
+    JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.102-1.b14.el7_2.x86_64/jre \
+    PATH=${PATH}:${JAVA_HOME}/bin \
+    CHE_HOME=/home/user/che \
+    DOCKER_VERSION=1.6.0 \
+    DOCKER_BUCKET=get.docker.com
+
+RUN yum -y update update && \
+    yum -y install ca-certificates curl openssl java sudo && \
+    curl -sSL "https://${DOCKER_BUCKET}/builds/Linux/x86_64/docker-${DOCKER_VERSION}" -o /usr/bin/docker && \
+    chmod +x /usr/bin/docker
```
kbsingh: is there a reason why we are not using the distro provided docker here ? that would better line up with the entire stack, both for buildtime and runtime.

kbsingh (changes requested):

Dockerfile.centos
TylerJewell: I think it has to do with limitations in how they distribute Docker 1.6, which uses a different location and distro than others.

kbsingh: could you point me at something that quantifies the need to remain with docker 1.6 ?

TylerJewell: We use docker 1.6 in our images which are responsible for creating other images to avoid having incompatibilities with different daemons.  The docker daemon is installed by users outside of our control and can vary in version and configuration.  We do the same in the Che launcher image too.

kbsingh: ack, thanks. I'll find a more appropriate venue for the 1.6 convesation; but to summarise present state : docker1.6 is used, in order to enable the widest possible user deployment stack selection.

TylerJewell: Correct, yes.  If you are interested I can tomorrow dig up the old issues where some of the inconsistencies were uncovered that led us to this approach.

l0rd (changes requested):
@dharmit I've built and tested your Dockerfile and it works like a charm :-) However I think we still need to solve the following points to be able to merge:
- Move the declaration of environment variable `CHE_HOME` with the previous `ENV` declarations
- Remove some package (like curl) as soon as they are not needed anymore
- Look if it's possible to specify `JAVA_HOME` without hardcoding the java version
- Look at a solution for file `cccp.yml`. That's a file that is needed for CentOS pipeline. From the point of view of a developer this file is probably useless and confusing. And it looks to me that it provides the `job id` only and that this is something the CentOS pipeline already knows. Are we sure we can't avoid it? Would it be possible to put it in a distinct repository?

Dockerfile.centos
benoitf: Maybe I wasn't clear enough but I didn't talk about pushing it to a repository. To use it locally, you don't need to push on a registry, just build it. And then you can still have a name without registry part.

dharmit:
@l0rd 

> Move the declaration of environment variable CHE_HOME with the previous ENV declarations

done.

> Remove some package (like curl) as soon as they are not needed anymore

can't remove `curl` and `ca-certificates` as they're dependencies for packages like `yum`, etc.

> Look if it's possible to specify JAVA_HOME without hardcoding the java version

done.

> Look at a solution for file cccp.yml. That's a file that is needed for CentOS pipeline. From the point of view of a developer this file is probably useless and confusing. And it looks to me that it provides the job id only and that this is something the CentOS pipeline already knows. Are we sure we can't avoid it? Would it be possible to put it in a distinct repository?

I've added more info to the file (as comments) to better explain its purpose.

benoitf:
@dharmit 

> can't remove curl and ca-certificates as they're dependencies for packages like yum, etc.

I don't see why you couldn't remove them as you install curl and ca-certificates few lines before with yum so yum can work without these packages and they're optional

dharmit:
@benoitf removed them from the `yum install` command since those packages are pre-installed.

dharmit:
@l0rd I've added additional info to `cccp.yml` explaining its purpose. Can you please review?

TylerJewell:
@dharmit @l0rd - what are modern day best practices for the files that should go into the root of a repository?  I'm asking because I'm looking at the root of Che repository.  I see the following:
1. Hidden . files, which are used by command-line systems.
2. Markdown files for root-level documentation
3. License, which is mandated by Eclipse Foundation
4. Maven pom for building
5. Che CLI for start / stop / update
6. And then we are starting to have some other items:

`Chefile &`Vagrantfile`, which is really about simplifying getting started with the product, but ultimately not essential for the core product.  And if we introduce the`cccp.yml file` then it's another flavor of this.  I'm wondering - is there a way where Chefile, Vagrantfile, and cccp.yml could be saved as dotfiles for cleanliness purposes?

l0rd (commented):
LGTM

l0rd (approved):

l0rd:
@TylerJewell I think we can rename `cccp.yml` to `.cccp.yml` but I wouldn't dot-prefix Vagrantfile.

TylerJewell:
+1 to the renaming idea.

dharmit:
@l0rd renamed, and did `s/build/include`.

l0rd:
@dharmit we are ready to merge here but you need to sign [the CLA](https://www.eclipse.org/legal/CLA.php) before. And once #2786 get merged we will need to adapt CentOS Dockerfile too.

maxandersen:
> @dharmit we are ready to merge here but you need to sign the CLA before. And once #2786 get merged we will need to adapt CentOS Dockerfile too.

The ip validation (https://dev.eclipse.org/eclipse-webhook/services/status_details.php?id=580461f6bb47e) complains about emails not listed in the commit. Are you sure the ip validation have been rerun against the latest push ?

dharmit:
@maxandersen 

> Are you sure the ip validation have been rerun against the latest push ?

Nope, I'm not sure if they've rerun against the latest push. I'm seeing that since first time I pushed.

TylerJewell:
It has been buggy for awhile.  As long as a pr submitter tells us in the thread that they have executed the cla, we have permission to merge.

maxandersen:
You can use https://projects.eclipse.org/user/cla/validate to check the email too and it does validate dharmit@redhat.com as having signed CLA.

TylerJewell:
@l0rd has commit rights - so will let him as maintainer drive merge, branch deletion, and selection of milestone.

benoitf:
@maxandersen @TylerJewell the CLA check is not buggy.
The commit provided is having "Author: Dharmit Shah shahdharmit@gmail.com"
while CLA is being for dharmit@redhat.com 
so this PR shouldn't have been merged until commit author have been set to dharmit@redhat.com

maxandersen:
@benoitf okey so I now learned to not trust github UI since it does not show the actual author. Note taken! I 

@dharmit you should remember to set the right author not just the signed-off part.

Issue #1801, opened 2016-07-18
Language Server Protocol: Signature Help feature


vparfonov:
The signature help request is sent from the client to the server to request signature information at a given cursor position.
Changed: in 2.0 the requests uses TextDocumentPositionParams with a proper textDocument and position property. In 1.0 the uri of the referenced text document was inlined into the params object.
https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#signature-help

PR #2645, opened 2016-09-29
#1801 implement 'Signature Help' LS feature

evidolob:
### What does this PR do?

implement 'Signature Help' LS feature
https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#signature-help-request
#1801


ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/editorconfig/TextEditorConfiguration.java
```diff
@@ -97,4 +98,11 @@
      */
     @Nullable
     ChangeInterceptorProvider getChangeInterceptorProvider();
+
+    /**
+     * Return the {@link SignatureHelpProvider}
+     * @return the signature help provider
+     */
+    @Nullable
+    SignatureHelpProvider getSignatureHelpProvider();
```
benoitf, line 107: other new interface like SignatureHelp is using `Optional<>`
shouldn't we use `Optional<SignatureHelpProvider>` here as well ?

evidolob: Maybe, but we need to replace all other methods in this class to return `Optional<>` which will be very big changes in  Editor API. In future we definitely should avoid returning `null` and use `Optional<>`

vparfonov (approved):

Issue #2538, opened 2016-09-22
Rename of workspace prevent its start


garagatyi:
After renaming of workspace it is not possible to start that workspace.

**Reproduction Steps:**
1. Create new Workspace from stack
2. Check that it starts well
3. Open swagger
4. Get workspace object
5. Change name in workspace object
6. Update workspace with swagger
7. Open UD
8. Start renamed workspace

**Expected behavior:**

Renamed workspace starts.

**Observed behavior:**

UD hangs. Shows loader forever. Api calls to `http://localhost:8080/wsmaster/api/workspace/che:tetetetererrere` returns 404 with message `Workspace with name 'tetetetererrere' in namespace 'che' doesn't exist`

**Che version:**    5.0.0-M3-Snapshot
**OS and version:** Ubuntu 14.04
**Docker version:** 1.12
**Che install:**    native

**Additional information:**
- Problem can be reliably reproduced, doesn't happen randomly: [Yes]

benoitf:
and if you rename workspace through UD ?

garagatyi:
the same

riuvshin:
seems it is a blocker

PR #2652, opened 2016-09-29
Remove workspace name from model

voievodin:
### What does this PR do?

Uses `workspace.config.name` throughout the components instead of `workspace.name`.
#2538
### PR type
### Minor change checklist
- [ ] New API required?
- [ ] API updated
- [x] Tests passed

@skabashnyuk, @akorneta review please

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/model/impl/WorkspaceImpl.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/model/impl/WorkspaceImpl.java
@@ -12,6 +12,7 @@ package org.eclipse.che.api.workspace.server.model.impl;
 
 import org.eclipse.che.account.shared.model.Account;
 import org.eclipse.che.account.spi.AccountImpl;
+import org.eclipse.che.api.core.jdbc.jpa.eclipselink.DescriptorEventAdapter;
 import org.eclipse.che.api.core.model.workspace.Workspace;
 import org.eclipse.che.api.core.model.workspace.WorkspaceConfig;
 import org.eclipse.che.api.core.model.workspace.WorkspaceRuntime;
@@ -19,6 +20,7 @@ import org.eclipse.che.api.core.model.workspace.WorkspaceStatus;
 import org.eclipse.che.api.machine.server.model.impl.SnapshotImpl;
 import org.eclipse.che.api.workspace.server.jpa.WorkspaceEntityListener;
 import org.eclipse.che.commons.lang.NameGenerator;
+import org.eclipse.persistence.descriptors.DescriptorEvent;
 
 import javax.persistence.Basic;
 import javax.persistence.CascadeType;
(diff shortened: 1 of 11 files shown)


core/che-core-api-jdbc/src/main/java/org/eclipse/che/api/core/jdbc/jpa/eclipselink/AbstractDescriptorEventListener.java
```diff
@@ -0,0 +1,71 @@
...
+import org.eclipse.persistence.descriptors.DescriptorEventListener;
+
+import java.util.Vector;
+
+/**
+ * A skeletal implementation of {@link DescriptorEventListener}.
+ *
+ * @author Yevhenii Voevodin
+ */
+public abstract class AbstractDescriptorEventListener implements DescriptorEventListener {
```
skabashnyuk: Can you use DescriptorEventAdapter?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/model/impl/WorkspaceImpl.java
```diff
@@ -62,7 +63,7 @@
                             query = "SELECT w FROM Workspace w")
         }
 )
-@EntityListeners(WorkspaceEntityListener.class)
+@EntityListeners({WorkspaceEntityListener.class, WorkspaceImpl.SyncNameOnUpdateAndPersistEventListener.class})
```
skabashnyuk, line 66: Can you add a comment about what this listeners is doing?

skabashnyuk (approved):

core/che-core-api-jdbc/src/main/java/org/eclipse/che/api/core/jdbc/jpa/eclipselink/AbstractDescriptorEventListener.java
evoevodin: I'd prefer to let the current name, afaik awt provides some skeletal implementations called adapters, but from the _Effective Java 2nd ed_:

> While interfaces are not permitted to contain method implementations, using
interfaces to define types does not prevent you from providing implementation
assistance to programmers. You can combine the virtues of interfaces and
abstract classes by providing an abstract skeletal implementation class to go
with each nontrivial interface that you export. The interface still defines the
type, but the skeletal implementation takes all of the work out of implementing it.
By convention, skeletal implementations are called AbstractInterface, where
Interface is the name of the interface they implement. For example, the Collec-
tions Framework provides a skeletal implementation to go along with each main
collection interface: AbstractCollection, AbstractSet, AbstractList, and
AbstractMap. Arguably it would have made sense to call them SkeletalCollec-
tion, SkeletalSet, SkeletalList, and SkeletalMap, but the Abstract conven-
tion is now firmly established.

akorneta (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/model/impl/WorkspaceImpl.java
voievodin: Okay, i will add more docs for a new listener

core/che-core-api-jdbc/src/main/java/org/eclipse/che/api/core/jdbc/jpa/eclipselink/AbstractDescriptorEventListener.java
voievodin: ok

Issue #2518, opened 2016-09-21
Add ability to list/edit stack's tags 


ashumilova:
User should be able to define stack's tags, that will be used to filter the project templates.

olexii4:
2d

PR #2668, opened 2016-09-30
CHE-2518: add ability to add/remove stack's tags

olexii4:
### What does this PR do?

Add ability to add/remove stack's tags 
#2518

tags

### DIFF
--- a/dashboard/src/app/stacks/stack-details/stack.controller.ts
+++ b/dashboard/src/app/stacks/stack-details/stack.controller.ts
@@ -16,6 +16,27 @@
  * @author Ann Shumilova
  */
 export class StackController {
+  $log: ng.ILogService;
+  $filter: ng.IFilterService;
+  $timeout: ng.ITimeoutService;
+  $location: ng.ILocationService;
+  $mdDialog: angular.material.IDialogService;
+
+  loading: boolean;
+  isLoading: boolean;
+  isCreation: boolean;
+
+  stackId: string;
+  stackName: string;
+  stackContent: string;
+  invalidStack: string;
+
+  stack: any;
+  editorOptions: any;
+  changesPromise: any;
+
+  cheStack;
+  cheNotification;
 
   /**
    * Default constructor that is using resource injection
@@ -57,14 +78,14 @@ export class StackController {
    *
    * @returns {{stack}} new stack template
    */
-  getNewStackTemplate() {
+  getNewStackTemplate(): any {
     this.stackName = 'New Stack';
-    let stack = {};
+    let stack: any = {};
     stack.name = this.stackName;
     stack.description = '';
     stack.source = {};
     stack.source.origin = '';
-    stack.source.type = ''
+    stack.source.type = '';
     stack.components = [];
     stack.tags = [];
     stack.workspaceConfig = {};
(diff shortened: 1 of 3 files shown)


dashboard/src/app/stacks/stack-details/stack.controller.ts
```diff
@@ -16,12 +16,31 @@
...
+  stackContent: string;
+  invalidStack: string;
+
+  $log;
+  stack;
+  $filter;
+  cheStack;
+  $timeout;
+  $location;
+  $mdDialog;
```
benoitf: for information, you may use https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/angular-material/angular-material.d.ts#L114 type
as we have angular-material definitions there https://github.com/eclipse/che/blob/master/dashboard/tsd.json#L26

the same for Angular object, you can use TypeScript definition (which are not any)

dashboard/src/app/stacks/stack-details/stack.controller.ts
```diff
@@ -103,9 +122,35 @@ export class StackController {
   }
 
   /**
+   * Handle stack's tag adding.
+   *
+   * @param tag stack's tag
+   * @returns {*}
+   */
+  handleTagAdding(tag) {
```
benoitf: you may add type of argument
like tag : any (or string I've not checked) and as there is a return value, you may add it at the end of the method

`handleTagAdding(tag : string) : string`

benoitf (approved):

dashboard/src/app/stacks/stack-details/stack.html
```diff
@@ -47,6 +47,28 @@
     </div>
   </che-label-container>
 
+  <!-- stack's tags -->
+  <che-label-container class="stack-tags-label"
+                       che-label-name="Tags"
+                       che-label-description="Tags define templates available for workspace">
```
ashumilova: "Tags define the templates ..." on the mockup

dashboard/src/app/stacks/stack-details/stack.html
```diff
@@ -47,6 +47,28 @@
...
+                md-on-remove="stackController.updateEditorContent()"
+                secondary-placeholder="Enter stack's tags">
+        <md-chip-template>
+          <div class="stack-library-text">{{$chip | uppercase}}</div>
+          <div class="stack-library-btn">
+            <i class="fa fa-close"></i>
+          </div>
+        </md-chip-template>
+      </md-chips>
+      <div class="stack-tags-reset" data-ng-click="stackController.stack.tags.length = 0">Reset</div>
```
ashumilova: Make Reset button not 100% width of the row - now click below the input causes clearing the tags.

dashboard/src/app/stacks/stack-details/stack.html
olexii4: OK

dashboard/src/app/stacks/stack-details/stack.controller.ts
olexii4: done

dashboard/src/app/stacks/stack-details/stack.controller.ts
olexii4: done

ashumilova (approved):

akurinnoy (approved):

Issue #2670, opened 2016-09-30
Workspace configurations are not saved on Che server restart.


JamesDrummond:
On persistent storage with H2/JPA in che and docker containers

datasource is used as 
[https://github.com/eclipse/che/blob/dda5acf61ebe9e89a9dda6b6afe738b6fd09bfa6/assembly/assembly-main/src/assembly/tomcat/conf/server.xml#L53](https://github.com/eclipse/che/blob/dda5acf61ebe9e89a9dda6b6afe738b6fd09bfa6/assembly/assembly-main/src/assembly/tomcat/conf/server.xml#L53)

and folder for db is using
[https://github.com/eclipse/che/blob/dda5acf61ebe9e89a9dda6b6afe738b6fd09bfa6/assembly/assembly-main/src/assembly/bin/che.sh#L611](https://github.com/eclipse/che/blob/dda5acf61ebe9e89a9dda6b6afe738b6fd09bfa6/assembly/assembly-main/src/assembly/bin/che.sh#L611)

so database is saved into ${CHE_HOME}/db/

but when reading [https://eclipse-che.readme.io/docs/usage-docker#workspace-storage](https://eclipse-che.readme.io/docs/usage-docker#workspace-storage) we can use `-e CHE_DATA_FOLDER`to customize the host directory where to save the data

[https://github.com/eclipse/che/blob/master/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties#L28-L29](https://github.com/eclipse/che/blob/master/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties#L28-L29)

**Reproduction Steps:**
1. Start Che server with CLI or che-launcher.  
2. Create and open a new workspace.
3. Make changes to workspace configuration.  
4. Stop che server.
5. Repeat steps 1 and 2. Configurations are not there.

**Expected behavior:**

All workspace configurations are changed after restart.

**Observed behavior:**

Workspace configurations are not saved after restart.

**Che version:**    5.0.0 master(M3)   

**Additional information:**
- Problem started happening recently, didn't happen in an older version of Che: Yes
- Problem can be reliably reproduced, doesn't happen randomly: Yes

benoitf:
adding labels asked by @JamesDrummond

riuvshin:
М4 already released. moved to M5

riuvshin:
@bmicklea please also move it to M5 milestone plan

[...]

PR #2685, opened 2016-10-03
Change the configuration of the root folder for the che db

akorneta:
### What does this PR do?

Avoids hardcoded value for che db root
#2670
### PR type

@skabashnyuk, @benoitf, @evoevodin

### DIFF
--- a/assembly/assembly-main/src/assembly/bin/che.sh
+++ b/assembly/assembly-main/src/assembly/bin/che.sh
@@ -608,7 +608,7 @@ call_catalina () {
 
   ### Cannot add this in setenv.sh.
   ### We do the port mapping here, and this gets inserted into server.xml when tomcat boots
-  export JAVA_OPTS="${JAVA_OPTS} -Dport.http=${CHE_PORT} -Dche.home=${CHE_HOME} -Dh2.baseDir=${CHE_HOME}/db/"
+  export JAVA_OPTS="${JAVA_OPTS} -Dport.http=${CHE_PORT} -Dche.home=${CHE_HOME}"
   export SERVER_PORT=${CHE_PORT}
 
   # Launch the Che application server, passing in command line parameters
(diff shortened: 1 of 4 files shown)


skabashnyuk (approved):

benoitf (approved):

core/che-core-api-jdbc/src/main/java/org/eclipse/che/api/core/jdbc/jpa/guice/JpaInitializer.java
```diff
@@ -11,18 +11,30 @@
...
 
+    @Inject(optional = true)
+    @Named("che.conf.storage")
+    private String storageRoot;
+
     @Inject
-    public JpaInitializer(PersistService persistService) {
+    private void initialize(PersistService persistService) {
+        if (storageRoot != null) {
+            System.getProperties().put("h2.baseDir", Paths.get(storageRoot).resolve("db").toString());
```
benoitf: note that you've also System.setProperty(key, value)

core/che-core-api-jdbc/src/main/java/org/eclipse/che/api/core/jdbc/jpa/guice/JpaInitializer.java
```diff
@@ -11,18 +11,30 @@
...
  * See <a href="https://github.com/google/guice/wiki/JPA">doc</a>
  *
  * @author Yevhenii Voevodin
  */
+@Singleton
 public class JpaInitializer {
 
+    @Inject(optional = true)
+    @Named("che.conf.storage")
+    private String storageRoot;
```
benoitf: as it's an optional injection, what will result for h2.baseDir if storageRoot is not defined ? What will be default H2 storage directory ?

benoitf (approved):
Sound like issue about optional inject is solved !

Issue #1756, opened 2016-07-16
Add ability to download workspace starting outputs from IDE loader


TylerJewell:
We should be able to download the logs from the console. We need to make another way, simpler and more obvious so that the user can download the outputs in case something goes wrong.

slemeur:
Attached corresponding mockup.
New loader in the IDE + download logs button.

![loading-workspace-in-ide - v2](https://cloud.githubusercontent.com/assets/1636769/17665000/5e13c792-62f8-11e6-956a-9f77eddd0e55.png)

slemeur:
Updated design.

![loading-workspace-in-ide - v2](https://cloud.githubusercontent.com/assets/1636769/18133994/11ffe08c-6f9d-11e6-81c1-b9668567099a.png)

We need to make the part for the outputs being maximized

We will also take care about making the width of the tree in the Console panel as minimal as possible.

vitaliy-guliy:
Spent 3 days

PR #2686, opened 2016-10-03
CHE-1756 Add ability to download workspace starting outputs from IDE loader

vitaliy-guliy:
Adds the "Download outputs" hyperlink to the loader.
Adds "Download" button to output consoles.

https://github.com/eclipse/che/issues/1756

### DIFF
--- a/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/dialogs/DialogFactory.java
+++ b/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/dialogs/DialogFactory.java
@@ -207,6 +207,40 @@ public interface DialogFactory {
                                   @Nullable CancelCallback cancelCallback);
 
     /**
+     * Create an input dialog with the specified initial value.
+     * <p/>
+     * The {@code initialValue} may be pre-selected. Selection begins
+     * at the specified {@code selectionStartIndex} and extends to the
+     * character at index {@code selectionLength}.
+     *
+     * @param title
+     *         the window title
+     * @param label
+     *         the label of the input field
+     * @param initialValue
+     *         the value used to initialize the input
+     * @param selectionStartIndex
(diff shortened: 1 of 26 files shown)


vitaliy-guliy:
download1
download2

slemeur:
Discussed.
Instead of having a popup we should directly name the file and start the download. This would require less interactions and be less error prone.

We should name the file with the following pattern:
{workspace-name}-{machine-name}-{date}-{time}.log

slemeur (changes requested):
Need to remove the popup

vitaliy-guliy:
Popup is removed, file name is changed according described above.

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/dialogs/DialogFactory.java
```diff
@@ -207,6 +207,40 @@ InputDialog createInputDialog(@NotNull @Assisted("title") String title,
...
+     *         the number of characters of the {@code initialValue} to be selected
+     * @param okButtonLabel
+     *         label for OK button
+     * @param inputCallback
+     *         the callback used on OK
+     * @param cancelCallback
+     *         the callback used on cancel
+     * @return an {@link InputDialog} instance
+     */
+    InputDialog createInputDialog(@NotNull @Assisted("title") String title,
```
benoitf, line 234: is this method required as we don't have the popup anymore ?

vitaliy-guliy: We can leave it for the future. Please give a look at the create folder popup. There are two buttons Cancel and OK.
It would be better to have Cancel and Create instead.

ashumilova (approved):

vparfonov (approved):

PR #2691, opened 2016-10-04
Allow to use bare "alpine" image

benoitf:
### What does this PR do?

alpine-sh

Add support for alpine image

It uses sh instead of bash (but allowing to configure shell through "shell" attribute of commands)
It also need https://github.com/eclipse/che-lib/pull/14 to have static go binaries 
Workspace agent on Alpine uses OpenJDK8 as glibc is not provided by default (and Oracle JDK is requiring it)
Terminal agent is also updated to provide /bin/bash only if it's available else it provides /bin/sh access
### Previous behavior

### New behavior
- use of /bin/sh for docker command invocation instead of /bin/bash
- allow to still use bash by providing a custom "shell" / "/bin/bash" attribute
### PR type
### Minor change checklist
- [ ] New API required?
- [ ] API updated
- [x] Tests passed


TylerJewell:
Super slick.  +1

TylerJewell:
I noticed that there are updates done for the ws-agent and terminal agents.  Will the other agents that have been packaged work ok?

benoitf:
@garagatyi @tolusha @mmorhun @vkuznyetsov  ?

benoitf:
@TylerJewell yes I think

tolusha (approved):

mmorhun:
LGTM

PR #2694, opened 2016-10-04
change deprecated package manager to a new one

olexii4:
### What does this PR do?

Change deprecated package manager(https://github.com/DefinitelyTyped/tsd) to a new one(https://github.com/typings/typings).

It fixed TypeScript errors


benoitf (approved):
I think the move to do, as TypeScript 2.0 is now officially released is to use @types/angular + TypeScript 2.0

example : https://www.npmjs.com/package/@types/angularjs

as tsd or typings are both deprecated now

but I'm +1 to remove "old deprecated tsd"

I've found this article about @types : https://blogs.msdn.microsoft.com/typescript/2016/06/15/the-future-of-declaration-files/

akurinnoy (approved):

Issue #2015, opened 2016-08-01
Simplify names of che.properties to be user friendly


garagatyi:
Tyler Jewell:

> I have gone through and done a revision of the default che.properties file. Every property in a comment `#` is the name of the existing property.  The uncommented names are the proposed new property names. I grouped properties by the end user names that we give to concepts as opposed to the internal names, which are not easily understood.

```
#
# Copyright (c) 2012-2016 Codenvy, S.A.
# All rights reserved. This program and the accompanying materials
# are made available under the terms of the Eclipse Public License v1.0
# which accompanies this distribution, and is available at
# http://www.eclipse.org/legal/epl-v10.html
#
# Contributors:
#   Codenvy, S.A. - initial API and implementation
#

### CHE SERVER
# Folder where Che will store internal data objects
che.database=${che.home}/db
#che.conf.storage=${che.home}/storage

# API service. Browsers initiate REST communications to Che server with this URL
che.api=http://localhost:${SERVER_PORT}/wsmaster/api
#api.endpoint=http://localhost:${SERVER_PORT}/wsmaster/api

### WORKSPACES
# The location where your workspaces (and their projects) are stored by Che server.
# This property works for Docker for Windows, Mac, and Linux.
# On vbox for windows, this property is ignored and workspaces stored:
# c:\%userprofile%\AppData\Local\Eclipse Che\
che.workspace.storage=${che.home}/workspaces
che.workspace.storage.create_folders=true

# Your projects are synchronized from the Che server into the machine running each
# workspace. This is the directory in the ws runtime where your projects are mounted.
che.workspace.projects.storage=/projects
# che.machine.projects.internal.storage=/projects

# Configures proxies used by runtimes powering workspaces
che.workspace.http_proxy=
che.workspace.https_proxy=
che.workspace.no_proxy=
#http.proxy=
#https.proxy=
#no_proxy=

# During the stop of the workspace automatically creates a snapshot if the value is {true},
# Otherwise just stops the workspace.
che.workspace.auto_snapshot=true
# workspace.runtime.auto_snapshot=true

# During the start of the workspace automatically restore it from a snapshot if the value is {true},
# Otherwise create a new workspace.
che.workspace.auto_restore=true
# workspace.runtime.auto_restore=true

# Java command line options used to start Che agent in workspace runtime
che.workspace.java_opts=-Xms256m -Xmx2048m -Djava.security.egd=file:/dev/./urandom
[...]

PR #2707, opened 2016-10-05
CHE-2015 Rename configuration properties

mkuznyetsov:
### What does this PR do?

Rename configuration properties to user-friendly ones.
Add aliases file to support old user configuration

https://github.com/eclipse/che/issues/2015

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che_aliases.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che_aliases.properties
@@ -1,0 +1,60 @@
+#
+# Copyright (c) 2012-2016 Codenvy, S.A.
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License v1.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-v10.html
+#
+# Contributors:
+#   Codenvy, S.A. - initial API and implementation
+#
+
+che.database=che.conf.storage
+che.api=api.endpoint
+che.workspace.projects.storage=che.machine.projects.internal.storage
+che.workspace.http_proxy=http.proxy
+che.workspace.https_proxy=https.proxy
+che.workspace.no_proxy=no_proxy
+che.workspace.auto_snapshot=workspace.runtime.auto_snapshot
+che.workspace.auto_restore=workspace.runtime.auto_restore
+che.workspace.java_opts=che.machine.java_opts
+che.workspace.volume=machine.server.extra.volume
+che.workspace.ssh_connection_timeout_ms=machine.ssh.connection_timeout_ms
+che.workspace.terminal_linux_amd64=machine.server.terminal.path_to_archive.linux_amd64
+che.workspace.terminal_linux_arm7=machine.server.terminal.path_to_archive.linux_arm7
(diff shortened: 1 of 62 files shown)


skabashnyuk (approved):

vzhukovs (changes requested):
OpenShift plugin need also to be updated with this changes

mkuznyetsov:
OpenShift plugin has its own PR https://github.com/codenvy/plugin-openshift/pull/39

vzhukovs (approved):

TylerJewell:
@mkuznyetsov @skabashnyuk - I have a couple requests before we merge this PR.  First, it's good work.  I know that there are a lot of details to work through and it was not simple.  I would like this PR to be merged at the same time that we merge:

https://github.com/eclipse/che/pull/2786, which will need to merge and push to Eclipse:

https://github.com/eclipse/che-dockerfiles/pull/29, which will then also require testing of the Che CLI and verification that the docs are current.

If we are going to make a bunch of properties changes like this, we need to end to end test che.properites with changes to che-server, che-launcher and CLI, so that if we break something, which is likely, that we can quickly fix it.  These changes span multiple repositories so while I have tried to test everything locally, it is best for us to merge it all at once, and then to have a very aggressive testing cycle around it.

mkuznyetsov:
@TylerJewell 
Now I'm doing final checks for this PR. Also, it turned out that purpose of aliases is not to support old configuration, but old code, that relies on old configuration.
So, here's how it is working:
- Only new property names will be allowed to be used in configuration (conf. files, system properties, env. variables).
- All our code will be also using new properties.
- Aliases are used in case, if someones old code is using old property names.

About those other PRs, I'd like to know if they're are final and ready to merge, and who will be merging them. I see that it should not conflict with this PR. (I don't see properties, that I've renamed).

TylerJewell:
@mkuznyetsov - I see your questions on who will merge the other PRs.  Please connect with your TL and identify someone (maybe you?) who shoudl merge those PRs.  I wrote them both, and I have done personal testing, but now we need someone in the core team to review, comment, merge, test, and verify.  I supsect that we will break some of our CI systems when we do this, so we will need very good testing of che-launcher, che cli, and che-server.

We should do this at the beginning of the next sprint, as this PR will be merged at the beginning of M7.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che_aliases.properties
```diff
@@ -0,0 +1,60 @@
...
+che.docker.ip=machine.docker.local_node_host
+che.docker.ip.external=machine.docker.local_node_host.external
+che.docker.network=docker.che_host_network
+che.docker.registry_for_snapshots=machine.docker.snapshot_use_registry
+che.docker.namespace=machine.docker.snapshot.registry_namespace
+che.docker.unused_containers_cleanup_min=machine.docker.unused_containers_cleanup_period_min
+che.docker.api=docker.api.version
+che.docker.tcp_connection_timeout_ms=docker.connection.tcp.connection_timeout_ms
+che.docker.tcp_connection_read_timeout_ms=docker.connection.tcp.read_timeout_ms
+che.docker.swap=machine.docker.memory_swap_multiplier
```
skabashnyuk: new line

skabashnyuk (approved):

PR #2717, opened 2016-10-05
CLI: Add new command - ws - for listing & logs of workspaces

TylerJewell:
### What does this PR do?

1: Adds in functionality to let users list all workspaces and then display the logs of a running workspace.

2: Provides some limited discovery - if no running workspaces, returns an error. If 1 running workspace, will automatically display the logs for that.  If 2+ workspaces, lists the workspaces and their ws-id asking user to pick which one they want.

```
che ws list                            List all workspaces
che ws logs [<ws-id>]                  Get the agent logs for a workspace
```

3: Fixes a bug in the --networking command where the use of the new curl command requires --net=host to operate properly. This bug was previously unreported, but was discovered in testing the info changes.

4:  Adds in logic so that INFO: command either prints out no pre-info (such as what we need for certain long-output command) or prints out `(che <command-name>):` as a prefix.

5: Simplifies all error: messages to remove any product name from it.
### PR type

### DIFF
--- a/cli.sh
+++ b/cli.sh
@@ -73,8 +73,6 @@ Usage: ${CHE_MINI_PRODUCT_NAME} [COMMAND]
     dir down                           Stop workspace running in current directory
     dir status                         Display status of ${CHE_MINI_PRODUCT_NAME} in current directory
     dir ssh                            Make SSH connection into workspace mapped to current directory
-    ws list                            List all workspaces
-    ws logs [<ws-id>]                  Get the agent logs for a workspace
     action <action-name> [--help]      Start action on ${CHE_MINI_PRODUCT_NAME} instance
     compile <mvn-command>              SDK - Builds Che source code or modules
     test <test-name> [--help]          Start test on ${CHE_MINI_PRODUCT_NAME} instance
@@ -128,7 +126,7 @@ parse_command_line () {
     CHE_CLI_ACTION="help"
   else
     case $1 in
-      start|stop|restart|update|info|profile|action|dir|ws|mount|compile|test|help|-h|--help)
+      start|stop|restart|update|info|profile|action|dir|mount|compile|test|help|-h|--help)
         CHE_CLI_ACTION=$1
       ;;
       *)
@@ -142,7 +140,6 @@ parse_command_line () {
 execute_cli() {
   case ${CHE_CLI_ACTION} in
     start|stop|restart)
-      # remove "che" arg by shifting it
       shift 
       load_profile
       execute_che_launcher "$@"
@@ -156,12 +153,6 @@ execute_cli() {
       load_profile
       execute_che_dir "$@"
     ;;
-    ws)
-      # remove "ws" arg by shifting it
-      shift
-      load_profile
-      execute_che_ws "$@"
-    ;;    
     action)
       # remove "action" arg by shifting it
       shift
@@ -606,7 +597,6 @@ load_profile() {
     fi
 
     source ~/."${CHE_MINI_PRODUCT_NAME}"/profiles/"${CHE_PROFILE}"
-    info "${CHE_PRODUCT_NAME}: Loaded profile ${CHE_PROFILE}"
   fi
 }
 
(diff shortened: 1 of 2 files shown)


TylerJewell:
@xcoulon @benoitf - FYI.  Xavier, I modified your logic a bit because the CLI on the user's machine can have different YEAR / MONTH / DAY.  Instead, I grab all log files and print them out.

cli.sh
```diff
@@ -750,6 +759,70 @@ execute_che_dir() {
...
+              CURRENT_WS_PORT=$(docker inspect --format='{{ (index (index .NetworkSettings.Ports "22/tcp") 0).HostPort }}' ${CHE_WS_CONTAINER_ID})
+            fi
+            echo "$CHE_WS_CONTAINER_ID       $(has_ssh ${CHE_WS_CONTAINER_ID} && echo "y" || echo "n")           $CURRENT_WS_PORT"
+          done
+          return
+        fi
+      fi
+
+      info "Displaying logs for container $WS_ID"
+      docker_exec exec ${WS_ID} find /home/ -iname "catalina-0.log" -exec tail -n200 {} \;
```
xcoulon: @TylerJewell  My initial code was using the current date to compute the path to the catalina logs. Also, I used the `-f` flag on the `tail` command to keep displaying upcoming logs.
Also, why do you call the `docker_exec()` function here ? Why not just use the `docker` command ?

xcoulon: Ha sorry, I missed the comment above where you explained that you changed the logic to skip the YEAR / MONTH / DAY part (although, I wonder why those were different). Or did you need to check the logs for another date ?

TylerJewell: I removed the `-f` because it was consistent with all of our other commands that we have in the CLI, which always return from execution of the command itself. So, as a general rule, I'd like to try and preserve that functionality.  The only deviation from that rule right now is with `che mount which has to run indefinitely and cannot be sent to execute in the background.

`docker_exec()` is an internal function that we use in the CLI to determine which type of docker command to run. If you are on windows, there is an environment variable that must be passed in order for volume mounts to work properly.  The function is probably not nicely named because docker_exec is really to just execute the `docker` command, not `docker exec`. So there is some word play.

The issue I had with keeping the YEAR / MONTH / DAY was that there was no way to have those variables executed within the container. It was always executed on the host system first, so there was always going to be the possibility of a date mismatch.

TylerJewell: I think a proper long term solution would be to find a way to enable `docker logs -f <container-id>` on these workspace containers.  I'll talk to Alex about what we need to do to activate that. I think it has to do with the way that the services are started.

benoitf: I think it would be nice to allow to get live logs  as well 
`che ws logs`(print current)
`che ws -f logs`(print live)

And I also agree that it could be nice to get logs through docker logs command but I think these logs are coming from the streams of the CMD that we launch. As it's launched by an agent it may be difficult.

BTW now that we've tons of agents, I would also be able to get logs for each agent as well.

Is this command is also working in compose mode as well ? or it will try to get logs on each machine of the workspace id ?

cli.sh
```diff
@@ -779,7 +852,7 @@ execute_che_mount() {
 
   # If extra parameter provided, then this is the port to connect to
   if [ $# -eq 1 ]; then
-    info "${CHE_MINI_PRODUCT_NAME} mount: Connecting to remote workspace on port ${1}"
+    info "${CHE_MINI_PRODUCT_NAME} mount: Connecting to workspace on port ${1}"
```
benoitf: We should be using PRODUCT_NAME and not mini name, no ?

cli.sh
TylerJewell: Right now - it only grabs a single machine set of logs. If there are multiple machines that are running, then it asks the user to choose which machine that they should connect to.  One of the improvements that I could make is to scan all of the machines that are running, and only list the workspace machines that are running a dev agent on it, and then allow for connection into that.

cli.sh
TylerJewell: I simplified things.  I have two modes for INFO now.  for certain outputs like `che info` where there is a lot of streming output, I do not print out any additional information other than `INFO:`.  For others, I print out `($CHE_MINI_PRODUCT_NAME <command-name>): <statement>`

Issue #2594, opened 2016-09-26
Consider Adding Default Profiles for Maven Build


l0rd:
OpenJDK [is not currently supported](https://github.com/eclipse/che/wiki/Development-Workflow) to build Che.

Is there a technical reason for that? Would it be hard to support it? 

Without OpenJDK support it's not possible to build Che on CentOS or RHEL platform with Red Hat supported packages only.

snjeza:
I have built che on Fedora 24 using OpenJDK 1.0.0_102.

ibuziuk:
@l0rd @snjeza I have tried to build the whole che against openjdk version "1.8.0_91" on Ubuntu 16.04. Basically, after using --fail-at-end flag I have failures in three plugins:
- Che Plugin :: SVN :: Server ........................ FAILURE
- Che Plugin :: Maven :: Extension Maven Server ...... FAILURE
- Che Plugin :: Maven :: Maven Server Impl ........... FAILURE

All seems to be server related

ibuziuk:
@l0rd @snjeza hmmm.... I have the very same failure against oracle "1.8.0_91"

snjeza:
I have used

mvn -DskipTests=true -Dfindbugs.skip=true -Dskip-validate-sources clean install

benoitf:
could you list failure errors ?

snjeza:
There aren't any failure errors if we skip the tests. When I include them, I get the same errors as you get. In my opinion, they aren't related to OpenJDK.

ibuziuk:
I'm doing -X right now, will let you know once finished

ibuziuk:
@benoitf agree with @snjeza it looks like there is just a bunch of failing tests:
-  **che-plugin-svn-ext-server**

`Tests run: 7, Failures: 0, Errors: 7, Skipped: 0, Time elapsed: 0.728 sec <<< FAILURE! - in org.eclipse.che.plugin.svn.server.SubversionApiITest`

`Tests run: 6, Failures: 0, Errors: 6, Skipped: 0, Time elapsed: 0.518 sec <<< FAILURE! - in org.eclipse.che.plugin.svn.server.SubversionProjectImporterTest`

all due to ERROR: **/bin/bash: svnadmin: command not found**

Probably, someone removed the command without amending the tests ?
-  **maven-server-impl**

7 test failures in **org.eclipse.che.maven.server.ProjectResolverTest** due to NPE in **setUp()**

```
[...]

PR #2733, opened 2016-10-07
Adding 'fast' profile for building Che faster by skipping unit tests,…

ibuziuk:
### What does this PR do?

'fast' profile for building Che faster by skipping unit tests, license checks and other enforcement features

https://github.com/eclipse/che/issues/2594
### New behavior

"mvn clean verify -Pfast" will allow to build che with the following  system properties enabled:

```
    -DskipTests=true
    -Dfindbugs.skip=true
    -Dskip-validate-sources
    -Dmdep.analyze.skip=true
    -Dlicense.skip=true
    -Dgwt.compiler.localWorkers=2 -T 1C
```
### PR type
### Minor change checklist
- [ ] New API required?
- [ ] API updated
- [x] Tests passed
### Major change checklist
- [ ] New API required?
- [ ] API updated
- [ ] Tests passed

Signed-off-by: Ilya Buziuk ibuziuk@redhat.com

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -849,4 +849,18 @@
             <url>https://maven.codenvycorp.com/content/repositories/codenvy-public-snapshots/</url>
         </pluginRepository>
     </pluginRepositories>
+    <profiles>
+        <!-- Profile for building Che faster by skipping unit tests, license checks and other enforcement features -->
+        <profile>
+            <id>fast</id>
+            <properties>
+                <findbugs.skip>true</findbugs.skip>
+                <gwt.compiler.localWorkers>2 -T 1C</gwt.compiler.localWorkers>
+                <license.skip>true</license.skip>
+                <mdep.analyze.skip>true</mdep.analyze.skip>
+                <skip-validate-sources>true</skip-validate-sources>
+                <skipTests>true</skipTests>
+            </properties>
+        </profile>
+    </profiles>
 </project>


benoitf (approved):

benoitf:
ci-build

vzhukovs (approved):

benoitf:
@ibuziuk After some tests of this profile it looks like -Pfast is not disabling the validate source profile

using -Dskip-validate-sources is working but using -Pfast not.

```
$ mvn -Pfast help:active-profiles |grep validate-sources
 - validate-sources (source: org.eclipse.che.parent:maven-parent-pom:5.0.0-M6-SNAPSHOT)
 - validate-sources (source: org.eclipse.che.parent:maven-parent-pom:5.0.0-M6-SNAPSHOT)
 - validate-sources (source: org.eclipse.che.parent:maven-parent-pom:5.0.0-M6-SNAPSHOT)
...
```

```
$ mvn -Pfast -Dskip-validate-sources help:active-profiles |grep validate-sources

```

ibuziuk:
@benoitf oh, this is trickier than I expected it to be. Basically,  one can not activate a profile when another one is activated (chaining activation is prohibited in maven). `skip-validate-sources` actually enables another profile - https://github.com/eclipse/che-parent/blob/master/pom.xml#L1651 . So, the only solution for this situation would be activating both profiles via the same property e.g.

```
<activation>
  <property>
    <name>fast</name>
  </property>
</activation>
```

I guess, if we change the fast profile's activation^ and add **fast** property to  **validate-source** profile activation both should be enabled via `mvn clean verify -Dfast`

benoitf:
@ibuziuk note that we also have skip-enforce property.
also I'm not sure it's about chained profile but just that property is read and enable the provile before another profile set value to false
I'm then not sure to have fast to disable their activation (because it's disabled profile if property is set)

ibuziuk:
@benoitf I do believe if we may have smth. like that:

```
      <profile>
            <id>validate-sources</id>
            <activation>
                <property>
                    <name>!skip-validate-sources</name>
                    <name>fast</name>
                </property>
            </activation>
```

AFAIK, This will not disable `validate-sources` profile but rather add one more condition when it should be enabled. Maven`<activation>` block is a list of 'or' statements and profile will be activated as soon as the first criteria is met

ibuziuk:
@benoitf sorry I was wrong about 'or' statements, it used to be like that before v3.2.2 but now it is fixed and it is actually the list of 'and' https://issues.apache.org/jira/browse/MNG-4565

benoitf:
OK

TylerJewell:
Cool.

ibuziuk:
Trying to figure out how it is possible to configure 'or' logic. It seems (and I would really like to be wrong here) that that after the "fix" it is not possible to setup "OR" logic for profile activation :/

ibuziuk:
@benoitf it is not possible to configure **IF** condition for maven profile activation unless using a copy-paste approach -  having two identical profiles with different activation logic. At the same time, chaining profile activation is also prohibited. So, there are two possible things we can do:
1. Remove skip-validate-source from fast profile (using the profile would include resource validation step )
2. duplicate skip-validate-source profile with different activation logic e.g.  

```
<activation>
  <property>
    <name>fast</name>
  </property>
</activation>

```

and change fast profile activation logic the same^ way

WDYT ?

PR #2735, opened 2016-10-07
Fixes #2728: Respect user settings when sending formatting options to LS

### LINKED ISSUE #2728: Language Server formatting options are hardcoded
Formatting options passed to language servers are hardcoded to:
- `insertSpaces = true`
- `tabWidth = 3` (why 3 ?!)

The `insertSpaces = true` value comes from here: https://github.com/eclipse/che/blob/master/plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/LanguageServerFormatter.java#L179

The `tabWidth = 3` value seems to come from here: https://github.com/eclipse/che/blob/master/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/editorconfig/DefaultTextEditorConfiguration.java#L35

**Reproduction Steps:**
1. Launch a Node stack with the JSON LS Agent activated
2. Open a JSON file in the editor
3. Format the document

**Expected behavior:**

The document is formatted according to the current user preferences.

**Observed behavior:**

The document is formatted with a hardcoded settings. The user preferences are not applied.

**Che version:**    latest master
**OS and version:** Fedora 24
**Docker version:** 1.10.3
**Che install:**    local build

**Additional information:**
- Problem started happening recently, didn't happen in an older version of Che: [No]
- Problem can be reliably reproduced, doesn't happen randomly: [Yes]

kaloyan-raev:
### What does this PR do?

Replaces the hard coded formatting options sent to language servers with those configured in user preferences.

Language Server formatting options are hardcoded #2728
### Previous behavior

Documents are formatted with hardcoded "tab size" and "insert spaces" (if editor uses a language server, e.g. JSON editor)
### New behavior

Documents are formatted with the "tab size" and "insert spaces" from the user preferences.
### PR type


benoitf:
ci-build

PR #2757, opened 2016-10-11
Annotation used is not the expected one

benoitf:
### What does this PR do?

Fix invalid imports
### Previous behavior

Usage of invalid classes
### New behavior

Do not use internal classes

Change-Id: I9a55703457218022487025e2570864f661d005bf


skabashnyuk (approved):

benoitf:
ci-build

PR #2767, opened 2016-10-12
Provide full project configuration to CreateProjectHandler

tareksha:
### What does this PR do?

Allow project creation handlers in extensions to obtains more information about the project that is bing created.

Some CreateProjectHandlers can benefit a lot more if they obtain full project configuration, most importantly:
1.  Access more information, e.g. mix-in types and source details.
2. Integration more easily with other code that works with project configurations.
### Previous behavior

Only `Map<String,AttributeValue>` is passed to `CreateProjectHandler`.
### New behavior

A safe, full copy of `ProjectConfig` is passed to `CreateProjectHandler`.

Signed-off-by: Tareq Sharafy tareq.sha@gmail.com

### DIFF
--- a/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectManager.java
+++ b/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectManager.java
@@ -27,7 +27,6 @@ import org.eclipse.che.api.project.server.handlers.ProjectHandlerRegistry;
 import org.eclipse.che.api.project.server.importer.ProjectImportOutputWSLineConsumer;
 import org.eclipse.che.api.project.server.importer.ProjectImporter;
 import org.eclipse.che.api.project.server.importer.ProjectImporterRegistry;
-import org.eclipse.che.api.project.server.type.AttributeValue;
 import org.eclipse.che.api.project.server.type.BaseProjectType;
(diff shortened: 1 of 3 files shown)


tareqhs:
@skabashnyuk

gazarenkov:
I proposed just Attributes instead of full Config on purpose, to not to give unnecessary (IMO) information.
Please provide a real case when you believe other than Attributes needed. And moreover I think only Provided Attributes are necessary for generator if you have some in mind.

tareksha:
Hi @gazarenkov ,
`ProjectConfig` is needed because we had some cases where the list of mix-in types is needed by the generator. I suggested passing `ProjectConfig` because we had it in the legacy `CreateModuleHandler` and it turned useful, although not a must.
Please allow me to ask: what is the purpose of avoiding "too much information"? per my understanding there is hard wiring between each project type and a single `CreateProjectHandler` which is a part of the Che extension.

gazarenkov:
Hi @tareqhs 
The reason is that information is not necessarily known before project created.
Also, this handler is per ProjectType so other than "key" types (in particular other mixin PTs) normally should not be taken into account in the handler.  
So, I would consider a list of attributes which belong to key PT instead of the whole config is more clean solution. 
So far I can not imagine the case when the whole Config would be useful, if you can provide one I would probably change my opinion.

vparfonov:
not actual any more

Issue #2590, opened 2016-09-26
PHP support


kaloyan-raev:
This issue collects all items to be done in order to provide additional PHP support within Che.
- [x] Basic PHP plugin #2160
- [x] Enable syntax highlighting for *.phtml files #2263
- [x] Register [HvyIndustries/crane](https://github.com/HvyIndustries/crane) as the PHP language server #2282
- [x] Replace [HvyIndustries/crane](https://github.com/HvyIndustries/crane) with [felixfbecker/php-language-server](https://github.com/felixfbecker/php-language-server) as the PHP language server #2830
  - [x] InitializeParams should include an empty ClientCapabilities #2314
- [x] Enable PHP and JSON language server agents for all PHP stacks #2639
- [x] Add Zend stack (PHP 7, Zend Server and Z-Ray) #2700
- [ ] Add PHP + database compose stack with example
  Blocked by:
  - [x] Fix for using the official mysql image in composite stacks #3049
  - [ ] DB data is lost after workspace restart #3054
- [x] Composer project importer #2785
- [ ] Add more project templates:
  - [x] [Zend Framework](https://github.com/zendframework/ZendSkeletonApplication) #2785
  - [ ] [Apigility](https://github.com/zfcampus/zf-apigility-skeleton)
  - [x] [Zend Expressive](https://github.com/zendframework/zend-expressive-skeleton) #2785
  - [x] [Astro Splash demo app](https://github.com/AndrewCarterUK/AstroSplash) #2700
- [x] Out-of-the-box code completion and validation in the `composer.json` editor #2474
- [x] Integrate Zend Debugger https://github.com/eclipse/che/pull/3202
  - [x] Install Zend Debugger in the default PHP stack eclipse/che-dockerfiles#55
  - [x] Add "Run/Debug PHP script" commands to the PHP and Zend stacks #3277
  - [x] Documentation for PHP Debugging codenvy/che-docs#37
- [ ] Integrate Xdebug
- [x] Unit testing with PHPUnit #3987
  - [x] Decouple testing framework API from Java test specific frameworks #3452
  - [x] Install PHPUnit runtime in the PHP stack https://github.com/eclipse/che-dockerfiles/pull/62
- [x] Add Zend to the contributors list on the marketing home page.
- [x] Update PHP getting started guide in docs to cover all PHP scenarios

General bugs affecting the PHP support too:
- [x] Editor does not filter code completion proposals from language server #2592
- [x] Formatting the whole document makes a big mess #2437
[...]

PR #2785, opened 2016-10-14
Composer project importer

kaloyan-raev:
### What does this PR do?

[Composer](https://getcomposer.org/) is a popular PHP dependency manager (similar to Bower for
Javascript). A natural way to create a new project in PHP is by materializing it from a Composer package like this:

``` sh
composer create-project laravel/laravel /projects/laravel
```

The above command fetches the `laravel/laravel` Composer package from a central repository (https://packagist.org/) and installs it at the `/project/laravel` folder on the local file system. The advantage against cloning a git repository is that `composer create-project` will install all required
dependencies.

This PR contributes a new Composer plugin that:
- Defines a Composer project importer that can import the project source from a Composer package.
- Defines an Import Project Registrar that registers the Composer importer into the import project wizard.

composer-importer

The Composer project importer uses the 'location' property of the
SourceStorage model for the package name. As result the "source" of
project templates can be defined like this:

``` json
"source": {
  "type": "composer",
  "location": "zendframework/zend-expressive-skeleton",
  "parameters": {}
}
```

The new project importer requires the `composer` executable to be installed in the stack and available in the PATH. This is the usual case for all PHP stacks.

Additional changes:
- The 'Archive' importer category is renamed to 'Package'. See [this comment](https://github.com/eclipse/che/pull/2785#issuecomment-253809695).
- New project samples using the 'composer' source type:
  - Zend Framework Skeleton Application
  - Zend Expressive Skeleton Application
- PHP support #2590
### New behavior

- New "composer" source type is available for Project Templates. The "location" property provides the package name of the composer package. The [Templates](https://eclipse-che.readme.io/docs/template) must be updated
- New "composer" project importer is available for the Import Project wizard. It allows users to create new PHP projects from the specified Composer package.
- The existing "astro-splash" project sample is updated to use the new "composer" source type.
- New project samples:
  - Zend Framework Skeleton Application
  - Zend Expressive Skeleton Application


TylerJewell:
At this rate of improements, Kaloyan is going to be a Che committer before CheConf.  @bmicklea @vparfonov - please schedule code review / QA into the _current_ sprint.

kaloyan-raev:
The second commit renames the 'Archive' importer category to 'Package'. We used to have only the Zip importer in the Archive category, but this PR also adds the Composer one. Renaming the category to 'Package' fits both Zip and Composer importers, whereas 'Archive' is not good for the Composer importer.

I kept the ARCHIVE enum class as deprecated for backward compatibility.

This commit can be easily reverted if this change is not accepted.

bmicklea:
Very cool @kaloyan-raev!

bmicklea:
@kaloyan-raev - is this PR ready for review?  Or are you still working on this?

kaloyan-raev:
@bmicklea The PR is ready for review.

bmicklea:
@vparfonov - please review for M7

gazarenkov:
Hi @kaloyan-raev 

Good things and good case, congrats 
But it seems importer is not really an abstraction to use for this purpose since Composer bound to PHP (unlike neutral source code storages) and it seems it has more explicit behavioural nature than just source import.

It seems PHP - Composer relationship looks exactly like Java - Maven one.
The way I propose will also allow you to use all the other functionality offered by Composer, not only project bootstrapping with "create_project" option. 
So, I propose to organize it this way:
- Make Composer PT which extends PHP PT,  
  [see Maven PT](https://github.com/eclipse/che/blob/master/plugins/plugin-maven/che-plugin-maven-server/src/main/java/org/eclipse/che/plugin/maven/server/projecttype/MavenProjectType.java) .
  Include it into "PHP" group so in the Project Wizard it will be listed along with "bare PHP" PT [see Maven wizard](https://github.com/eclipse/che/tree/master/plugins/plugin-maven/che-plugin-maven-ide/src/main/java/org/eclipse/che/plugin/maven/client/wizard).
- To make it possible to bootstrap project make default Composer generator use your code with "create_project" option. Or you can make it even more advanced (maybe later) making 2 different generators (from scratch and bootstrapping, see how it is done on Maven PT for Simple/Archetype generators, [see Maven generator](https://github.com/eclipse/che/blob/master/plugins/plugin-maven/che-plugin-maven-server/src/main/java/org/eclipse/che/plugin/maven/server/projecttype/handler/MavenProjectGenerator.java) ).

kaloyan-raev:
@gazarenkov Thanks for the review and the advice.

Indeed, the PHP - Composer relationship is similar to the Java - Maven. I will look at the Maven plugin as you have suggested and try to rework this PR.

gazarenkov:
Cool, feel free to ask me or @vparfonov if you need some help with that, good luck.

kaloyan-raev:
@gazarenkov I looked at the Maven plugin. I agree that from a design point of view it would be better to move the creation of a Composer project from the Import Project wizard to the Create Project wizard.

However, the main motivation for this PR was to make it possible to have project templates that are created directly from a composer package. I would still need the project importer to achieve this, right?

kaloyan-raev:
@gazarenkov @vparfonov Could you look at my last question?

vparfonov:
@kaloyan-raev Hi, I don't know how composer work but look like we don't need it as **importer**. Importer means that you  have some remote, for example VCS. If composer just generate some files like maven archetype plugin it's org.eclipse.che.api.project.server.handlers.CreateProjectHandler in our Project API.

kaloyan-raev:
@vparfonov Quote from the `composer create-project` command [documentation](https://getcomposer.org/doc/03-cli.md#create-project):

> You can use Composer to create new projects from an existing package. This is the equivalent of doing a git clone/svn checkout followed by a "composer install" of the vendors.

So it does fetch source code. And it also installs the required dependencies. It is the preferred way in the PHP world to start coding on a project template.

My main goal is to have Che project templates created from Composer packages. I still need the project importer to achieve this, don't I? Let me know if I can achieve it in another way.

I can still refactor the GUI part and plug into the Create Project wizard, instead of to the Import Project wizard.

kaloyan-raev:
I removed the IDE plugin that extends the Import Project wizard from this PR, so we can focus entirely on the ability to create project templates from a Composer package, i.e. with `source.type = composer`.

I will prepare a new PR with IDE plugin extending that Create Project wizard after we merge this PR.

vparfonov:
@kaloyan-raev we have fixed Maven Archetype Project generator here https://github.com/eclipse/che/commit/9951639518d5f74cdce24b529ed2914ea9d8220e

i think you can do the same with composer

kaloyan-raev:
@vparfonov  Can you give me an example how I can use the Maven Archetype Project generator in a project template?

vparfonov:
@kaloyan-raev maybe i don't understand your question, but i say about Maven Archetype only like sample.
So i think you need to add CreateProjectHandler for project type "composer" and move code:

`
new ProcessBuilder("composer", "create-project", packageName, projectPath, "--no-progress")
                 .redirectErrorStream(true).start();
`

to them.

kaloyan-raev:
@vparfonov My goal is to be able to define project templates with `source.type = composer` here: https://github.com/eclipse/che/blob/master/ide/che-core-ide-templates/src/main/resources/samples.json

I need this new `source.type`, because I want to fetch the project's source code from a composer package instead of a git repository.

Will this be possible if I implement the CreateProjectHandler that you suggest? I have the impression that defining new source types is possible only via project importers.

gazarenkov:
Hi, @kaloyan-raev 

Well, I understand your point but this kind of identifier (like:  "type": "composer",
 "location": "kaloyan-raev/astrosplash") can not be in the "source" conceptually, since the place you identify does not contain a project source, so you are not able to sync it with your "local" sources.

On the other hand I like the idea to be able to declaratively define project generator, but currently there are no API functionality to do that, so we decided to add one more method to Project API which would allow this sort of thing adding generator properties to project creating JSON.

There is the issue for it which I hope will be implemented soon #2937  
@vparfonov aware of it

So, I think you will have to rework Composer stuffs as discussed above (Project Type like Java - Maven relationship)

kaloyan-raev:
> Well, I understand your point but this kind of identifier (like: "type": "composer",
> "location": "kaloyan-raev/astrosplash") can not be in the "source" conceptually, since the place you identify does not contain a project source, so you are not able to sync it with your "local" sources.

@gazarenkov You are wrong. The "kaloyan-raev/astrosplash" package contains the actual project source code. One of the things that the `composer create-project` command is to fetch this source code from the Composer repository to a folder on the local file system.

I don't see any significant difference to how the ZIP importer is implemented, for example. The "location" for the ZIP importer is a URL of a ZIP file. The ZIP importer downloads the file and then extracts it using the `unzip` command. How is this conceptually different to fetching the source code with the `composer create-project` command?

gazarenkov:
I told the criteria upon which it can not be considered as an imported and serialized as a project source:
- there should be a way to implement the sync back (it can be easily implemented with zip, it just not considered as a storage for prod)
- it should be project type agnostic 

So far a do not think php composer meets them.

kaloyan-raev:
> there should be a way to implement the sync back (it can be easily implemented with zip, it just not considered as a storage for prod)

Can you explain what do you mean by "sync back"? What does it mean to sync back to a zip or to a git/svn repository?

> it should be project type agnostic

Can you explain why this limitation is important?

I don't think that the proposed approach with Maven Archetypes is a good fit here. Composer packages should not be compared Maven Archetypes. Maven Archetypes require some parameters to be passed, so they are used to replace placeholders in the archetype's source code. There is no such thing in Composer packages. It is better to think about Composer packages like actual Maven packages, e.g. a PetClinic app deployed in a Maven repo.

I can implement the Composer project type and the respective CreateProjectHandler, but this will only give me a solution for creating non-sample projects in the Che IDE's Create Project wizard. It does not give me the solution for project samples defined in the `samples.json`, which are only possible to be created via a project importer.

gazarenkov:
Sync back means that user also can commit the changes back to the source according its to location and properties.

"Can you explain why this limitation is important?"
It is not a limitation, it is a purpose.
A generator conceptually is bound to concrete project nature. 
Suits the Java/Maven or PHP/Composer case. 
You are not able to use the composer for other than PHP projects are you?

"I can implement the Composer project type and the respective CreateProjectHandler, but this will only give me a solution for creating non-sample projects in the Che IDE's Create Project wizard. It does not give me the solution for project samples defined in the samples.json, which are only possible to be created via a project importer."
As I told - see the #2937 it will solve this limitation.

bmicklea:
@kaloyan-raev we are pushing this to M8 because M7 has to be closed tomorrow. If this is merged tomorrow let me know and I'll update milestone to M7.

bmicklea:
@kaloyan-raev please resolve the conflicts

kaloyan-raev:
Conflicts resolved.

vparfonov:
Hi @kaloyan-raev. We added new method in Project API [ProjectService.java#L211](https://github.com/eclipse/che/blob/master/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java#L211) it accepts  [NewProjectConfigDto.java](https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/dto/NewProjectConfigDto.java ), so you can set in `Map<String, String> options` all parameters that you need for project generation. Also you need to register 
CreateProjectHandler for your project type and move logic from your ComposerProjectImporter.java to it.

kaloyan-raev:
@vparfonov I am trying to rework the PR as you suggest. Now I have a Composer project type extending the PHP project type. I also have a CreateProjectHandler.

But I am failing to pass any `options` or `attributes` from the project template definition in the `samples.json` to the CreateProjectHandler. The `onCreateProject()` method is called, but both `attributes` and `options` are empty.

Here is my definition in sampled.json:

```json
{
    "name": "zend-expressive",
    "displayName": "zend-expressive",
    "path": "/zend-expressive",
    "description": "A skeleton application using the Zend Exressive framework. Expressive is a minimalist PSR-7 middleware framework for PHP with routing, DI container, optional templating, and optional error handling capabilities.",
    "projectType": "composer",
    "mixins": [],
    "attributes": {
      "language": [
        "php"
      ],
      "package": [
        "zendframework/zend-expressive-skeleton"
      ]
    },
    "options": {
      "package": "zendframework/zend-expressive-skeleton"
    },
    "modules": [],
    "problems": [],
    "source": {},
    "commands": [
      {
        "name": "configure",
        "type": "custom",
        "commandLine": "chmod 777 ${current.project.path}/data/cache; echo \"<VirtualHost *:80>\n    DocumentRoot ${current.project.path}/public\n    SetEnv APPLICATION_ENV 'development'\n    <Directory ${current.project.path}/public>\n        DirectoryIndex index.php\n        AllowOverride All\n        Require all granted\n    </Directory>\n</VirtualHost>\" | sudo tee /etc/apache2/sites-available/000-default.conf",
        "attributes": {
          "previewUrl": ""
        }
      }
    ],
    "links": [],
    "category": "Samples",
    "tags": [
      "Zend"
    ]
  }
```

Note that I am specifying the "package" value in both `options` and `attributes` just for the tests. I am fine using any of them. And I am not really sure what is the difference between them.

vparfonov:
@kaloyan-raev  Can you push your changes to your branch i will check it, thx?

kaloyan-raev:
@vparfonov Meanwhile I was able to debug the problem. I created PR #3375 with a potential fix.

kaloyan-raev:
@vparfonov I reworked the PR. I take advantage of both the `CreateProjectHandler` and `ProjectInitHandler`. In the `CreateProjectHandler.onCreateProject()` method I create the project from the Composer package, but without installing the dependencies. The dependencies are installed in the `ProjectInitHandler.onProjectInitialized()`. This way I also cover the use case where the project is imported from a source location (Git, SVN, ZIP) and then configured as a Composer project.

The output of the executed Composer commands are displayed in a Composer console output in the IDE.

The plugin also contributes a page to the Create Project wizard where the user can specify a Composer package, which the create the project from.

All of the above works pretty nice, but I face some issues that should be addressed separately:
1. The dashboard does not report any progress as console output during the "Creating project" phase. This is due to a limitation of the API. The `ProjectImporter.importSources()` has a `LineConsumerFactory` parameter, which is not available in the `CreateProjectHandler.onCreateProject()` and `ProjectInitHandler.onProjectInitialized()` methods.
1. The dashboard waits for both `onCreateProject()` and `onProjectInitialized()` to finish before showing the Web IDE. Installing the Composer dependencies may take long minutes as they are downloaded from the Internet. It's not a nice user experience to just sit and wait in the "Creating project" status of the dashboard. It would be much better if the dashboard switches to the Web IDE after `onCreateProject()` finishes and executes `onProjectInitialized()` in the background. The user will see the progress of installing the dependencies in the console and can do other tasks in the mean time.
1. Same for the Web IDE's Create Project wizard. The wizard stays modal until both `onCreateProject()` and `onProjectInitialized()` finish. It would be better if the wizard executes `onProjectInitialized()` in the background.
1. The `onProjectInitialized()` is executed twice when a new project is created from a built-in sample with the Web IDE's Create Project wizard. The second call comes from updating the project after it is created. I am not sure if this update call is really necessary. This double call to `onProjectInitialized()` causes issues to some Composer projects and returns with error.
1. There is some strange flickering of the Composer console output. When a new line is added the view is scrolled to the top and then back to the bottom. It's very annoying.

bmicklea:
@kaloyan-raev I am going to move this to 5.1 - we are trying to finalize M9 which will be the 5.0 GA release and it seems there are more changes needed before this is ready to merge.

vparfonov:
will be merged in https://github.com/eclipse/che/pull/3726

PR #2790, opened 2016-10-16
CLI - Add support for unison profiles

TylerJewell:
### What does this PR do?

Adds support for unison profiles which allow users to use path specifications and regular expressions to decide what is allowed (or not allowed) to be synchronized as part of a che-mount.  This can be pushed live ahead of the updates to the che-mount utility.


TylerJewell:
@benoitf

benoitf (approved):

Issue #2365, opened 2016-09-07
command instruction not supported in compose file


ghost:
**Reproduction Steps:**

Use:

`command: tail -f /dev/null`

in the compose file.

**Expected behavior:**

Command is executed

**Observed behavior:**

Error parsing compose file

Currently command is accepted as:

`command: [tail, -f, /dev/null]`

garagatyi:
Please provide an example of compose file you used to reproduce behavior

ghost:
```
services:
 db:
  image: codenvy/mysql
  environment:
   MYSQL_USER: petclinic
   MYSQL_PASSWORD: password
  mem_limit: 2147483648
  command: service mysql start
 dev-machine:
  image: codenvy/ubuntu_jdk8
  mem_limit: 2147483648
  depends_on:
    - db
```

garagatyi:
Since command nature is an array of strings and this instruction is parsed in YAML as a single string compose has to have custom parsing rules to use such syntax. 
We haven't implemented that custom logic yet.

bmicklea:
@vkuznyetsov can you check to see if it can be completed for 18-Oct?

riuvshin:
Moved to M7 because it is still in progress.

PR #2807, opened 2016-10-18
CHE-2365: Add deserializer for 'command' field ComposeServiceImpl.

AndrienkoAleksandr:
### What does this PR do?

This pull request improve compose file parser to parse "command" field for the ComposeServiceImpl like simple string line or string array.

https://github.com/eclipse/che/issues/2365
### Previous behavior

Compose parser can parse ComposeServiceImpl#command only like string array.
### New behavior

Compose parser can parse ComposeServiceImpl#command like string array and simple text line.

@garagatyi @evoevodin @tolusha @mmorhun @sleshchenko  review please.


tolusha:
I don't see any tests.

AndrienkoAleksandr:
@tolusha If this solution is normal I'll write test.

tolusha:
As far it is ok.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/rule/CommandDeserializer.java
```diff
@@ -0,0 +1,110 @@
...
+     *         json parser
+     * @param ctxt
+     *         deserialization context
+     * @throws IOException
+     *         in case I/O error. For example element to parsing contains
+     *         invalid yaml field type defined for this field by yaml document model.
+     * @throws JsonProcessingException
+     */
+    @Override
+    public List<String> deserialize(JsonParser jsonParser, DeserializationContext ctxt) throws IOException, JsonProcessingException {
```
voievodin: Provided by the documentation example:

``` yaml
  image: codenvy/mysql
  environment:
    MYSQL_USER: petclinic
    MYSQL_PASSWORD: password
  mem_limit: 2147483648
  command: service mysql start
```

it is clear yaml format, why do you use _JSON_ parser for parsing command line from yaml format?

As far as i understand you should add support for `command: service mysql start` when this command is parsed from yaml to object by `ComposeFileParser`, do i miss something?

voievodin (changes requested):
Please answer my questions/comments

voievodin:
> @tolusha If this solution is normal I'll write test.

Why that? It's hard to say whether use-case is covered if you don't provide any tests, you can provide implementation agnostic tests and you won't need to retwrite them, moreover when you write test first, it is much simpler to design/develop solution for the problem.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/rule/CommandDeserializer.java
voievodin: okay, it's clear now, i just don't think that it's good to get something like `JsonProcessingException` when you are dealing with yaml file

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/ComposeServiceImpl.java
```diff
@@ -26,6 +29,7 @@
 public class ComposeServiceImpl {
     @JsonProperty("container_name")
     private String              containerName;
+    @JsonDeserialize(using = CommandDeserializer.class)
```
voievodin, line 33: Is there any way to avoid direct usage of jackson annotations right here?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/rule/CommandDeserializer.java
```diff
@@ -0,0 +1,110 @@
...
+        TreeNode tree = jsonParser.readValueAsTree();
+        List<String> commands;
```
voievodin: imho the code looks better when you avoid this list e.g:

``` java
final TreeNode tree = jsonParser.readValueAsTree()
if (tree.isArray()) {
    return toCommand(...);
}
if (tree instanceof TextNode) {
   return ...
}
throw new IOException(...)
```

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/rule/CommandDeserializer.java
```diff
@@ -0,0 +1,110 @@
...
+
+        if (tree.isArray()) {
+            commands = toCommand((ArrayNode)tree);
+        } else if (tree instanceof TextNode) {
+            TextNode node = (TextNode)tree;
+            commands = asList(node.asText().trim().split(SPLIT_COMMAND_REGEX));
+        } else {
+            throw new IOException(format("Field '%s' should be simple text or string array.", jsonParser.getCurrentName()));
```
voievodin: > should be simple text or string array

Is it really `IOException`? As far as i understand it is related to the format we expect, in this case i suppose client should get `BadRequestException`(after mapping), but what kind of exception client gets if you throw `IOException` from here?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/rule/CommandDeserializer.java
AndrienkoAleksandr: JsonProcessingException I'll delete it from signature, because it's unused.

dmytro-ndp:
@AndrienkoAleksandr: let's use package with more obvious name "**deserializer**" instead of "**rule**" for storing _CommandDeserializer_ class:
`org.eclipse.che.api.environment.server.compose.deserializer`

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/CommandDeserializer.java
```diff
@@ -0,0 +1,107 @@
...
+        TreeNode tree = jsonParser.readValueAsTree();
+
+        if (tree.isArray()) {
+            return toCommand((ArrayNode)tree);
+        }
+        if (tree instanceof TextNode) {
+            TextNode textNode = (TextNode)tree;
+            return asList(textNode.asText().trim().split(SPLIT_COMMAND_REGEX));
+        }
+        throw new RuntimeException(format("Field '%s' should be simple text or string array.", jsonParser.getCurrentName()));
```
garagatyi: Deserializer already declares exceptions that can be thrown. Runtime exception can be replaced with something else I suppose. Does Jackson have an example of deserializators creation to find out what is the proper way to throw an error.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/CommandDeserializer.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    private List<String> toCommand(ArrayNode arrayCommandNode) {
+        List<String> commands = new ArrayList<>();
+
+        for (TreeNode treeNode : arrayCommandNode) {
+            if (treeNode instanceof TextNode) {
```
garagatyi: What will happen if node is not TextNode?

garagatyi (changes requested):
Please consider moving compose parsing code out of API packages and modules.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/CommandDeserializer.java
dmytro-ndp: Class JsonMappingException is more suitable here.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/CommandDeserializer.java
AndrienkoAleksandr: We should throw exception. I will add test for that and  fix code.

AndrienkoAleksandr:
Pull request updated. Moved compose parser to separated module.

plugins/plugin-docker/che-plugin-docker-compose/src/main/java/org/eclipse/che/compose/parser/BuildContextImpl.java
```diff
@@ -8,7 +8,7 @@
  * Contributors:
  *   Codenvy, S.A. - initial API and implementation
  *******************************************************************************/
-package org.eclipse.che.api.environment.server.compose;
+package org.eclipse.che.compose.parser;
 
 import java.util.HashMap;
```
garagatyi: Can you rename this class since it is not implementation of anything

plugins/plugin-docker/che-plugin-docker-compose/src/main/java/org/eclipse/che/compose/parser/ComposeEnvironmentImpl.java
```diff
@@ -8,7 +8,7 @@
  * Contributors:
  *   Codenvy, S.A. - initial API and implementation
  *******************************************************************************/
-package org.eclipse.che.api.environment.server.compose;
```
garagatyi, line 11: Can you rename this class since it is not implementation of anything

plugins/plugin-docker/che-plugin-docker-compose/src/main/java/org/eclipse/che/compose/parser/BuildContextImpl.java
AndrienkoAleksandr: ok

plugins/plugin-docker/che-plugin-docker-compose/src/main/java/org/eclipse/che/compose/parser/ComposeEnvironmentImpl.java
AndrienkoAleksandr: OK

AndrienkoAleksandr:
ci-build

plugins/plugin-docker/che-plugin-docker-compose/src/main/java/org/eclipse/che/compose/parser/ComposeModule.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * @author Alexander Andrienko
+ */
+public class ComposeModule extends AbstractModule {
+    @Override
+    protected void configure() {
+        MapBinder<String, EnvironmentFileParser> mapBinder = MapBinder.newMapBinder(binder(),
+                                                                                    String.class,
+                                                                                    EnvironmentFileParser.class,
+                                                                                    Names.named("machine.docker.compose.parsers"));
```
garagatyi: I think that it is unique enough to avoid named injection

plugins/plugin-docker/che-plugin-docker-compose/src/main/java/org/eclipse/che/compose/parser/ComposeModule.java
```diff
@@ -0,0 +1,31 @@
...
+        mapBinder.addBinding("compose").toInstance(new ComposeFileParser());
```
garagatyi: It is better to bind to class instead of instance

plugins/plugin-docker/che-plugin-docker-compose/src/main/java/org/eclipse/che/compose/parser/EnvironmentFileParser.java
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.compose.parser;
+
+import org.eclipse.che.api.core.ServerException;
+
+/**
+ * @author Alexander Andrienko
```
garagatyi: Docs?

plugins/plugin-docker/che-plugin-docker-compose/src/main/java/org/eclipse/che/compose/parser/EnvironmentFileParser.java
```diff
@@ -0,0 +1,21 @@
...
+ */
+public interface EnvironmentFileParser {
+    ComposeEnvironment parse(String recipeContent, String contentType) throws IllegalArgumentException,
```
garagatyi: Docs?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
```diff
@@ -48,13 +49,13 @@
     protected static final String SERVER_CONF_LABEL_PROTOCOL_SUFFIX = ":protocol";
     protected static final String SERVER_CONF_LABEL_PATH_SUFFIX     = ":path";
 
-    private final ComposeFileParser composeFileParser;
-    private final RecipeDownloader  recipeDownloader;
+    private final Map<String, EnvironmentFileParser> composeFileParsers;
```
garagatyi: Please rename this field

garagatyi: This class should not know anything about compose

plugins/plugin-docker/che-plugin-docker-machine/pom.xml
```diff
@@ -90,6 +90,10 @@
             <artifactId>che-plugin-docker-client</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.eclipse.che.plugin</groupId>
+            <artifactId>che-plugin-docker-compose</artifactId>
```
garagatyi: Can you point me where compose is used in docker machine plugin?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalDockerModule.java
```diff
@@ -64,6 +64,7 @@ protected void configure() {
                          .toProvider(org.eclipse.che.plugin.docker.machine.local.provider.DockerApiHostEnvVariableProvider.class);
 
         install(new org.eclipse.che.plugin.docker.machine.DockerMachineModule());
+        install(new org.eclipse.che.compose.parser.ComposeModule());
```
garagatyi: You can make it Dynamodule and not install it manually

garagatyi (changes requested):

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalDockerModule.java
AndrienkoAleksandr: Ok

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
AndrienkoAleksandr: Ok

plugins/plugin-docker/che-plugin-docker-compose/src/main/java/org/eclipse/che/compose/parser/ComposeModule.java
AndrienkoAleksandr: Yes, you are right.

riuvshin:
ci-build

plugins/plugin-docker/che-plugin-docker-compose/src/main/java/org/eclipse/che/plugin/docker/compose/yaml/ComposeEnvironmentParser.java
```diff
@@ -0,0 +1,168 @@
...
+    public String toYaml(ComposeEnvironment composeEnvironment) throws IllegalArgumentException {
+        checkNotNull(composeEnvironment, "Compose environment should not be null");
+        try {
+            return YAML_PARSER.writeValueAsString(composeEnvironment);
+        } catch (JsonProcessingException e) {
+            throw new IllegalArgumentException(e.getLocalizedMessage(), e);
+        }
+    }
+
+    private String getContentOfRecipe(EnvironmentRecipe environmentRecipe) throws ServerException {
```
garagatyi, line 128: Please move that to general environment parser

plugins/plugin-docker/che-plugin-docker-compose/src/test/java/org/eclipse/che/plugin/docker/compose/yaml/EnvironmentContextTest.java
```diff
@@ -25,17 +30,18 @@
 /**
  * @author Dmytro Nochevnov
  */
+@Listeners(MockitoTestNGListener.class)
 public class EnvironmentContextTest {
```
garagatyi: I don't get why it is named EnvironmentContext, please rename it if you can

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
```diff
@@ -40,29 +35,30 @@
  * @author Alexander Garagatyi
  */
 public class EnvironmentParser {
-    private static final List<String> types = Arrays.asList("compose", "dockerimage", "dockerfile");
+
+    private static final String       COMPOSE      = "compose";
```
garagatyi: These values should not be hardcoded

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
```diff
@@ -80,26 +76,28 @@ public CheServicesEnvironmentImpl parse(Environment environment) throws IllegalA
...
         switch (envType) {
-            case "compose":
-                cheServicesEnvironment = parseCompose(environment.getRecipe());
+            case COMPOSE:
+                EnvironmentRecipeParser composeParser = environmentParsers.get(COMPOSE);
+                cheServicesEnvironment = composeParser.parse(recipe);
                 break;
-            case "dockerimage":
-            case "dockerfile":
+            case DOCKER_IMAGE:
```
garagatyi: These values should not be hardcoded

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
```diff
@@ -228,15 +184,7 @@ private CheServicesEnvironmentImpl parseDocker(Environment environment) {
             }
         }
 
-        return composeEnvironment;
-    }
-
-    private String getContentOfRecipe(EnvironmentRecipe environmentRecipe) throws ServerException {
```
garagatyi, line 234: Looks like this method doesn't depend on environment type, so you can leave it here

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
```diff
@@ -154,65 +152,23 @@ private void normalizeMachine(String name, CheServiceImpl service, ExtendedMachi
...
-                                                            .withPorts(service.getPorts())
-                                                            .withVolumes(service.getVolumes())
-                                                            .withVolumesFrom(service.getVolumesFrom());
-
-            services.put(composeServiceEntry.getKey(), cheService);
-        }
-        return new CheServicesEnvironmentImpl().withServices(services);
-    }
-
     private CheServicesEnvironmentImpl parseDocker(Environment environment) {
```
garagatyi: This method should be moved into separate parser also

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentRecipeParser.java
```diff
@@ -0,0 +1,37 @@
...
+    /**
+     * Returns {@link CheServicesEnvironmentImpl} with parameters parsed
+     * from {@link EnvironmentRecipe} content.
+     *
+     * @param environmentRecipe
+     *         recipe with environment definition
+     * @throws IllegalArgumentException
+     *         in case invalid argument in the recipe content of the {@link EnvironmentRecipe}
+     * @throws ServerException
+     *         when any error occurs
```
garagatyi: I think it is thrown in certain cases like network outage or inability to download something. Is it better then to explain it as "When parsing fails due to some internal server error or inability to parse environment due to other reasons"

garagatyi (changes requested):

plugins/plugin-docker/che-plugin-docker-compose/src/main/java/org/eclipse/che/plugin/docker/compose/yaml/ComposeEnvironmentParser.java
AndrienkoAleksandr: General environment parser does not use it.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
AndrienkoAleksandr: General environment parser does not use it.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentRecipeParser.java
AndrienkoAleksandr: ok

plugins/plugin-docker/che-plugin-docker-compose/src/test/java/org/eclipse/che/plugin/docker/compose/yaml/BuildContextTest.java
```diff
@@ -8,36 +8,44 @@
...
+ * Test deserialization field {@link ComposeServiceImpl#build}
+ * in the {@link ComposeEnvironmentParser}.
+ *
  * @author Mario Loriedo
  */
+@Listeners(MockitoTestNGListener.class)
 public class BuildContextTest {
 
-    ComposeFileParser parser = new ComposeFileParser();
+    @InjectMocks
```
garagatyi, line 41: Which mocks?

plugins/plugin-docker/che-plugin-docker-compose/src/test/java/org/eclipse/che/plugin/docker/compose/yaml/EnvironmentDeserializerTest.java
```diff
@@ -23,19 +29,24 @@
...
+ *
  * @author Dmytro Nochevnov
  */
-public class EnvironmentContextTest {
+@Listeners(MockitoTestNGListener.class)
+public class EnvironmentDeserializerTest {
 
-    ComposeFileParser parser = new ComposeFileParser();
+    @InjectMocks
+    private ComposeEnvironmentParser parser;
```
garagatyi, line 47: Which mocks?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/parser/DockerEnvironmentParser.java
```diff
@@ -0,0 +1,54 @@
...
+import static com.google.common.base.Preconditions.checkArgument;
+
+/**
+ * Parser for creating {@link CheServicesEnvironmentImpl} from {@link Environment}
+ * which contains a single {@link ExtendedMachine}.
+ *
+ * @author Alexander Garagatyi
+ * @author Alexander Andrienko
+ */
+public interface DockerEnvironmentParser extends TypeSpecificEnvironmentParser {
```
garagatyi: Maybe replace with abstract class?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
```diff
@@ -80,28 +70,25 @@ public CheServicesEnvironmentImpl parse(Environment environment) throws IllegalA
...
+        Set<String> envTypes = environmentParsers.keySet();
+
+        if (!envTypes.contains(envType)) {
+            throw new IllegalArgumentException(format("Environment type '%s' is not supported. " +
+                                                      "Supported environment types: %s",
+                                                      envType,
+                                                      Joiner.on(", ").join(envTypes)));
         }
 
+        TypeSpecificEnvironmentParser parser = environmentParsers.get(envType);
```
garagatyi, line 89: What if we don't have parser of some type?

garagatyi (changes requested):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
garagatyi: And what if this class is called directly? Will you get NPE?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/parser/DockerEnvironmentParser.java
```diff
@@ -40,7 +40,7 @@
      * @throws IllegalArgumentException
      *         in case if {@code environment} is multiple machine environment
      */
-    default String getMachineName(Environment environment) throws IllegalArgumentException {
+    String getMachineName(Environment environment) throws IllegalArgumentException {
```
garagatyi: protected?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/EnvironmentParser.java
garagatyi: Sorry, I missed check.

garagatyi (approved):
This PR really cleanups the code. Good job!

garagatyi:
@evoevodin can you take a look again

plugins/plugin-docker/che-plugin-docker-compose/src/main/java/org/eclipse/che/plugin/docker/compose/yaml/ComposeEnvironmentParser.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+    /**
+     * Converts Docker Compose environment model into YAML file.
+     *
+     * @param composeEnvironment Docker Compose environment model file
+     * @throws IllegalArgumentException
+     *         when argument is null or conversion to YAML fails
+     */
+    public String toYaml(ComposeEnvironment composeEnvironment) throws IllegalArgumentException {
+        checkNotNull(composeEnvironment, "Compose environment should not be null");
```
voievodin: 1. NPE should be thrown when argument is null
2. Why don't you just use `Objects` or `Preconditions` methods for such checks?
3. Don't list runtime exceptions like `IllegalArgumetnException` in the `throws` list, but still document them
4. Why the other methods of this class don't check their arguments?

plugins/plugin-docker/che-plugin-docker-compose/src/main/java/org/eclipse/che/plugin/docker/compose/yaml/ComposeEnvironmentParser.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    private CheServicesEnvironmentImpl asCheEnvironment(ComposeEnvironment composeEnvironment) {
+        Map<String, CheServiceImpl> services = Maps.newHashMapWithExpectedSize(composeEnvironment.getServices().size());
+        for (Map.Entry<String, ComposeServiceImpl> composeServiceEntry : composeEnvironment.getServices()
+                                                                                           .entrySet()) {
+            ComposeServiceImpl service = composeServiceEntry.getValue();
+
+            CheServiceBuildContextImpl buildContext = null;
+            if (service.getBuild() != null) {
```
voievodin: What if you move this section below the declaration of `cheService` then you will be able to avoid buildConext variable e.g.

``` java
if (service.getBuild() != null) {
    cheService.setBuild(new CheServiceBuildContextImpl().withContext(service.getBuild().getContext())
                                                        .withDockerfilePath(service.getBuild().getDockerfile())
                                                        .withArgs(service.getBuild().getArgs()))
}
```

Moreover i think that creation of `CheServiceBuildContextImpl` should be replaced with constructor call

plugins/plugin-docker/che-plugin-docker-compose/src/main/java/org/eclipse/che/plugin/docker/compose/yaml/ComposeEnvironmentParser.java
```diff
@@ -0,0 +1,173 @@
...
+                                                            .withLinks(service.getLinks())
+                                                            .withMemLimit(service.getMemLimit())
+                                                            .withNetworks(service.getNetworks())
+                                                            .withPorts(service.getPorts())
+                                                            .withVolumes(service.getVolumes())
+                                                            .withVolumesFrom(service.getVolumesFrom());
+
+            services.put(composeServiceEntry.getKey(), cheService);
+        }
+        return new CheServicesEnvironmentImpl().withServices(services);
```
voievodin: What if we add a suitable constructor for this purpose?

``` java
return new CheServicesEnvironmentImpl(services);
```

plugins/plugin-docker/che-plugin-docker-compose/src/main/java/org/eclipse/che/plugin/docker/compose/yaml/ComposeEnvironmentParser.java
```diff
@@ -0,0 +1,173 @@
...
+
+
+    /**
+     * Checks that object reference is not null, throws {@link IllegalArgumentException} otherwise.
+     *
+     * <p>Exception uses error message built from error message template and error message parameters.
+     */
+    private static void checkNotNull(Object object, String errorMessageTemplate, Object... errorMessageParams) {
+        if (object == null) {
+            throw new IllegalArgumentException(format(errorMessageTemplate, errorMessageParams));
```
voievodin: Wrong exception, see `Preconditions.checkNotNull` | `Objects.requireNonNull`.
You don't need this method, reuse existing one.

plugins/plugin-docker/che-plugin-docker-compose/src/main/java/org/eclipse/che/plugin/docker/compose/yaml/ComposeEnvironmentParser.java
garagatyi: +1 first suggestion. I don't see how usage of additional constructor here will simplify code.

plugins/plugin-docker/che-plugin-docker-compose/src/main/java/org/eclipse/che/plugin/docker/compose/yaml/deserializer/CommandDeserializer.java
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * See more for parsing compose file
+ * {@link ComposeEnvironmentParser}.
+ * <p>Note: this deserializer works for json too.</p>
+ *
+ * @author Alexander Andrienko
+ */
+public class CommandDeserializer extends JsonDeserializer<List<String>> {
+
+    private static final String SPLIT_COMMAND_REGEX = "[ \n\r]+";
```
voievodin, line 69: From the [documentation](http://docs.oracle.com/javase/6/docs/api/java/util/regex/Pattern.html)     `\s` stays for `[ \t\n\x0B\f\r]` does it fit your needs?

plugins/plugin-docker/che-plugin-docker-compose/src/main/java/org/eclipse/che/plugin/docker/compose/yaml/ComposeEnvironmentParser.java
voievodin: It will reduce count of called methods

voievodin (approved):
The changes are okay, please fix the small remarks and go for it! good job

plugins/plugin-docker/che-plugin-docker-compose/src/main/java/org/eclipse/che/plugin/docker/compose/yaml/deserializer/CommandDeserializer.java
garagatyi: I'm not sure that we should allow here anything except space and \t. @AndrienkoAleksandr why did you choose this regexp?

garagatyi (changes requested):

plugins/plugin-docker/che-plugin-docker-compose/src/main/java/org/eclipse/che/plugin/docker/compose/yaml/deserializer/CommandDeserializer.java
AndrienkoAleksandr: Because user can has a long command and can break line. On Linux break line is "\n", earlier Mac "\r", Windows "\r\n"

garagatyi (approved):

PR #2812, opened 2016-10-18
Fix policies hashCode

akorneta:
### What does this PR do?

Fix hash code of factory policies

@evoevodin please take a look.


voievodin (approved):
Next time use [templates](https://github.com/evoevodin/intellij-idea-tools/tree/master/templates/equals-hashcode/collections-never-null) for generation

Issue #2683, opened 2016-10-03
Add the way to figure out that workspace is currently snapshotting


ashumilova:
User will be able to do two things with running workspace. 
-  Stop it with snapshot
- Stop it without snapshot.

It means after RUNNING state workspace can have two states STOPPING or SNAPSHOTTING.
- RUNNING->STOPPING|ERROR
- RUNNING->SNAPSHOTTING|ERROR->STOPPING|ERROR

PR #2829, opened 2016-10-19
Add workspace 'SNAPSHOTTING' status

voievodin:
### What does this PR do?

Adds a new workspace status called _SNAPSHOTTING_ workspace is in snapshotting status if and only if it creates snapshots of it's machines right at the time status retrieved.

Moreover from now, there is no way to create snapshot of the same workspace simultaneously by 
sending 2 parallel requests to stop the workspace, only one of those requests will pass the status check.

Along with that new query parameter for `DELETE /api/workspace/<id>/runtime`(stop workspace) method was added, the query parameter is _create-snapshot_.
if _create-snapshot=true_ - snapshot will be created before workspace is stopped, 
if _create-snapshot=false_ - snapshot won't be created before workspace is stopped,
if _create-snapshot_ isn't send-  default behaviour will be used(server will look at workspace attributes and if nothing found it will use default configured value from `workspace.runtime.auto_snapshot` property value)
#2683

@skabashnyuk, @sleshchenko, @garagatyi, @akorneta please review

### DIFF
--- a/wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/SnapshotImpl.java
+++ b/wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/SnapshotImpl.java
@@ -44,7 +44,12 @@ import java.util.Objects;
                 @NamedQuery(name = "Snapshot.findSnapshots",
                             query = "SELECT snapshot " +
                                     "FROM Snapshot snapshot " +
-                                    "WHERE snapshot.workspaceId = :workspaceId")
+                                    "WHERE snapshot.workspaceId = :workspaceId"),
+                @NamedQuery(name = "Snapshot.findByWorkspaceAndEnvironment",
+                            query = "SELECT snapshot " +
+                                    "FROM Snapshot snapshot " +
+                                    "WHERE snapshot.workspaceId = :workspaceId " +
+                                    "  AND snapshot.envName = :envName")
         }
 )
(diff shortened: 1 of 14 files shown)


akorneta (approved):

sleshchenko (approved):

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/SnapshotImpl.java
```diff
@@ -44,7 +44,12 @@
                 @NamedQuery(name = "Snapshot.findSnapshots",
                             query = "SELECT snapshot " +
                                     "FROM Snapshot snapshot " +
-                                    "WHERE snapshot.workspaceId = :workspaceId")
+                                    "WHERE snapshot.workspaceId = :workspaceId"),
+                @NamedQuery(name = "Snapshot.findByWorkspaceAndEnvironment",
```
skabashnyuk, line 48: Check please that you have indexes on both fields

skabashnyuk (approved):

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/model/impl/SnapshotImpl.java
voievodin: checked, all the indexes exist!

PR #2836, opened 2016-10-19
Fix git config test failure

amisevsk:
### What does this PR do?

Changes property used for testing git config implementation to an unused one instead of `difftool.prompt`.

Fixes issue with running tests in `che-core-git-impl-jgit` when `difftool.prompt` option is set globally.
### Previous behavior

JGit pulls the global config file when creating a repository. While running tests, if `difftool.prompt` is set in the global `.gitconfig` file, this will causes tests to fail while building `Che Core :: Git Impl JGit`.
### New behavior

Changes `ConfigTest.java` to use nonsense property `test.prop`, so that there is no conflict.

Signed-off-by: Angel Misevski amisevsk@redhat.com


TylerJewell:
lgtm - @vkuznyetsov?

benoitf:
ci-build

vinokurig (approved):

Issue #2454, opened 2016-09-15
Configuring of named variable with underscore in name doesn't work if it was set using environment variables


garagatyi:
It is possible to configure Che using environment variables instead of property file.
Supposed way of doing that is:
env variable **CHE_SOME_PROPERTY** is converted into **some.property**.

But it is not possible to configure property if it contains underscore, e.g. **some.variable_with_underscore**.

**Che version:**    5.0.0-M2
**OS and version:** Linux (behavior will be the same for any OS)
**Che install:**    Docker container

**Additional information:**
- Problem started happening recently, didn't happen in an older version of Che: [Yes - there was no env variable injection before]
- Problem can be reliably reproduced, doesn't happen randomly: [Yes]

TylerJewell:
Cannot reproduce.

Use double underscore to create underscore.  See the docs.

garagatyi:
I reproduced it. Here is no replacing of double underscores with single https://github.com/eclipse/che/blob/master/core/commons/che-core-commons-inject/src/main/java/org/eclipse/che/inject/CheBootstrap.java#L272

TylerJewell:
I still cannot reproduce this.  If you use the CLI, then the underscores are converted.

garagatyi:
Can you provide steps to reproduce that it is working?

TylerJewell:
1. Get CLI
2. `export CHE_PROPERTY_machine_ws__agent_max__start__time__ms=1000`
3. `che start`

PR #2837, opened 2016-10-19
Che server should connect to ws-agent on internal URL (#2030)

amisevsk:
### What does this PR do?

Enables che-server to use the internal address of wsagent containers when it is running
in docker through setting an environment variable. 

When `CHE_DOCKER_USE_INTERNAL_CONTAINER_ADDRESS` is set to "true", the
`ServerProperties` objects returned as part of `DockerInstanceRuntimeInfo#getServers()`
will use the internal address of the container as obtained from docker inspect. 

If the environment variable is not set, then behaviour is unchanged.

This is a WIP so comments / suggestions are greatly appreciated.

Che server should connect to ws-agent on internal URL (#2030)
### Previous behavior

Value used for internal address could cause issues contacting wsagent if firewall interfered.
### New behavior

Che-server can now optionally communicate with wsagent directly.
### PR type
### Minor change checklist
- [ ] New API required?
- [ ] API updated
- [x] Tests updated
- [x] Tests passed

Signed-off-by: Angel Misevski amisevsk@redhat.com

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
@@ -143,7 +143,7 @@ che.docker.ip=NULL
 che.docker.ip.external=NULL
 
 # If true, then uses the internal address and port of workspace Docker containers (i.e. within the Docker
-# Docker network) instead of the address and port provided by the Docker daemon. May be necessary if the
+# Docker network) instead of the docker host address and the port provided by the Docker daemon. May be necessary if the
 # Che server is unable to contact workspace agents, but will prevent communication if the server and workspace
 # are not on the same network.
 che.docker.ip.use_internal_address=false
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
@@ -142,6 +142,12 @@ che.docker.ip=NULL
 # This is unusual, but happens for example in Docker for Mac when containers are in a VM.
 che.docker.ip.external=NULL
 
+# If true, then uses the internal address and port of workspace Docker containers (i.e. within the Docker
+# Docker network) instead of the address and port provided by the Docker daemon. May be necessary if the
+# Che server is unable to contact workspace agents, but will prevent communication if the server and workspace
+# are not on the same network.
+che.docker.ip.use_internal_address=false
+
 # Provides a Docker network where Che server is running.
 # Workspace containers created by Che will be added to this Docker network.
 # Communications between the Che server and container occur over this network.
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
@@ -310,7 +310,7 @@ public class DockerInstanceRuntimeInfo implements MachineRuntimeInfo {
         return servers;
     }
 
-    protected Map<String, ServerImpl> getServersWithFilledPorts(final String externalHostame, final String internalHostname, final Map<String, List<PortBinding>> exposedPorts) {
+    protected Map<String, ServerImpl> getServersWithFilledPorts(final String externalHostname, final String internalHostname, final Map<String, List<PortBinding>> exposedPorts) {
         final HashMap<String, ServerImpl> servers = new LinkedHashMap<>();
 
         for (Map.Entry<String, List<PortBinding>> portEntry : exposedPorts.entrySet()) {
@@ -330,7 +330,7 @@ public class DockerInstanceRuntimeInfo implements MachineRuntimeInfo {
 
             servers.put(portProtocol, new ServerImpl(null,
                                                      null,
-                                                     externalHostame + ":" + externalPort,
+                                                     externalHostname + ":" + externalPort,
                                                      null,
                                                      new ServerPropertiesImpl(null, internalHostnameAndPort, null)));
         }
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
@@ -88,6 +88,7 @@ public class DockerInstanceRuntimeInfo implements MachineRuntimeInfo {
     private final String                      containerExternalHostname;
     private final String                      containerInternalHostname;
     private final Map<String, ServerConfImpl> serversConf;
+    private final boolean                     useInternalAddress;
 
     @Inject
     public DockerInstanceRuntimeInfo(@Assisted ContainerInfo containerInfo,
@@ -95,7 +96,8 @@ public class DockerInstanceRuntimeInfo implements MachineRuntimeInfo {
                                      @Assisted("internalhost") String containerInternalHostname,
                                      @Assisted MachineConfig machineConfig,
                                      @Named("machine.docker.dev_machine.machine_servers") Set<ServerConf> devMachineSystemServers,
-                                     @Named("machine.docker.machine_servers") Set<ServerConf> allMachinesSystemServers) {
+                                     @Named("machine.docker.machine_servers") Set<ServerConf> allMachinesSystemServers,
+                                     @Named("che.docker.ip.use_internal_address") boolean useInternalAddress) {
         this.info = containerInfo;
         this.containerExternalHostname = containerExternalHostname == null ? containerInternalHostname : containerExternalHostname;
         this.containerInternalHostname = containerInternalHostname;
@@ -108,6 +110,7 @@ public class DockerInstanceRuntimeInfo implements MachineRuntimeInfo {
                                                                srvConf.getPort() :
                                                                srvConf.getPort() + "/tcp",
                                                     ServerConfImpl::new));
+        this.useInternalAddress = useInternalAddress;
     }
 
     @Override
@@ -315,11 +318,21 @@ public class DockerInstanceRuntimeInfo implements MachineRuntimeInfo {
             String portProtocol = portEntry.getKey();
             // we are assigning ports automatically, so have 1 to 1 binding (at least per protocol)
             String externalPort = portEntry.getValue().get(0).getHostPort();
+
+            // If we are sending messages to the container directly, we don't want to use the mapped ports.
+            String internalHostnameAndPort;
+            if (useInternalAddress) {
+                String internalPort = portProtocol.split("/")[0];
+                internalHostnameAndPort = internalHostname + ":" + internalPort;
+            } else {
+                internalHostnameAndPort = internalHostname + ":" + externalPort;
+            }
+
             servers.put(portProtocol, new ServerImpl(null,
                                                      null,
                                                      externalHostame + ":" + externalPort,
                                                      null,
-                                                     new ServerPropertiesImpl(null, internalHostname + ":" + externalPort, null)));
+                                                     new ServerPropertiesImpl(null, internalHostnameAndPort, null)));
         }
 
         return servers;
@@ -355,4 +368,5 @@ public class DockerInstanceRuntimeInfo implements MachineRuntimeInfo {
 
         return serversConf;
     }
+
 }
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalDockerInstanceRuntimeInfo.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalDockerInstanceRuntimeInfo.java
@@ -29,13 +29,10 @@ import java.util.Set;
  * directly (e.g. when running on Docker for Mac or behind a NAT). If the external address is not provided it defaults
  * to the internal one.
  *
- * <p>If environment variable {@code CHE_DOCKER_IP_USE__INTERNAL__ADDRESS} or
- * property {@code che.docker.ip.use_internal_address} is true, is true, the IP Address from
+ * <p>If property {@code che.docker.ip.use_internal_address} is true, is true, the IP Address from
  * {@link ContainerInfo}, if available, is used for internal address instead.<br>
- * <p>Value of host can be retrieved from property {@code che.docker.ip} or
- * from environment variable {@code CHE_DOCKER_IP}.<br>
- * <p>Value of external hostname can be retrieved from property {@code che.docker.ip.external} or
- * from environment variable {@code CHE_DOCKER_IP_EXTERNAL}.<br>
+ * <p>Value of host can be retrieved from property {@code che.docker.ip}. <br>
+ * <p>Value of external hostname can be retrieved from property {@code che.docker.ip.external}. <br>
  * Environment variables override properties.
  *
  * @author Alexander Garagatyi
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalDockerInstanceRuntimeInfo.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalDockerInstanceRuntimeInfo.java
@@ -16,6 +16,7 @@ import org.eclipse.che.api.core.model.machine.MachineConfig;
 import org.eclipse.che.api.core.model.machine.ServerConf;
 import org.eclipse.che.commons.annotation.Nullable;
 import org.eclipse.che.plugin.docker.client.json.ContainerInfo;
+import org.eclipse.che.plugin.docker.client.json.NetworkSettings;
 import org.eclipse.che.plugin.docker.machine.DockerInstanceRuntimeInfo;
 
 import javax.inject.Inject;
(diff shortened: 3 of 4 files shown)


TylerJewell:
This improvement addresses issues on certain linux systems where Che does not work with iptables or firewalld. By connecting over the internal address, communication flow will not go through the firewall which blocks access.

TylerJewell:
-[ ] Requires updates to the Networking docs for Eclipse Che page to explain the scenario.

benoitf:
ci-build

amisevsk:
I've rebased this PR, and dropped a commit that fixed an issue that has been since fixed in master (see d74b24fb073c6e0134ed9ab9e8f8c25551100c00).

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
```diff
@@ -310,16 +316,33 @@ public String projectsRoot() {
     protected Map<String, ServerImpl> getServersWithFilledPorts(final String externalHostame, final String internalHostname, final Map<String, List<PortBinding>> exposedPorts) {
         final HashMap<String, ServerImpl> servers = new LinkedHashMap<>();
 
+        boolean useMappedPorts = true;
+        if (useInternalContainerAddresses()) {
```
garagatyi: It is better to do this inside of constructor to keep main code clean

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
```diff
@@ -310,16 +316,33 @@ public String projectsRoot() {
...
+            if (info.getNetworkSettings() != null && internalHostname.equals(info.getNetworkSettings().getIpAddress())) {
+                useMappedPorts = false;
```
garagatyi: Can you elaborate why we should care about **internalHostname** field if usage of IP address provided by Docker is set by configuration?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
```diff
@@ -310,16 +316,33 @@ public String projectsRoot() {
...
             String portProtocol = portEntry.getKey();
             // we are assigning ports automatically, so have 1 to 1 binding (at least per protocol)
             String externalPort = portEntry.getValue().get(0).getHostPort();
+
+            // If we are sending messages to the container directly, we don't want to use the mapped ports.
+            String internalHostnameAndPort;
+            if (useMappedPorts) {
+                internalHostnameAndPort = internalHostname + ":" + externalPort;
+            } else {
+                String internalPort = portEntry.getKey().split("/")[0];
```
garagatyi: It is better to use portProtocol variable here  instead of **portEntry.getKey()**

garagatyi (changes requested):
Please add fixes in accordance to comments

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
amisevsk: This was meant mostly as a sanity check, although even as that I'm not sure it's handled as well as it could be. `internalHostname` is still the address used to communicate with wsagent, and I was concerned that it may be possible to create an instance of `DockerInstanceRuntimeInfo` with `internalHostname` not matching `info.getNetworkSettings().getIpAddress()`, in which case contacting wsagent would fail even if `internalHostname` is valid. 

If you feel that it's not a concern, I will remove it.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
```diff
@@ -355,4 +378,18 @@ public String projectsRoot() {
...
         return serversConf;
     }
+
+    /**
+     * Checks environment variable {@code CHE_DOCKER_USE_INTERNAL_CONTAINER_ADDRESS} to determine if
+     * direct addresses of containers should be used.
+     * @return true if {@code CHE_DOCKER_USE_INTERNAL_CONTAINER_ADDRESS} is "true", false otherwise.
+     */
+    protected boolean useInternalContainerAddresses() {
+        String useInternalContainerAddresses = System.getenv(CHE_DOCKER_USE_INTERNAL_CONTAINER_ADDRESS);
```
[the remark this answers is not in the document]
amisevsk: Sorry, could you clarify this point? Do you mean that it is preferable to control this behaviour through a property defined in `che.properties`?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
garagatyi: My point that if someone configured property/env variable that enables **useInternalContainerAddresses mode** then he definitely want to use it. And If it does match internalHostname then this mode doesn't make sense.
So I treat this mode as an alternative to default mode with set internalHostname.
Please correct me if I misunderstood the goal of this contribution.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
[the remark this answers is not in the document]
garagatyi: Exactly! And it can also be configured using env variable that match rules of CHE of conversion env variable into property.
The idea is that we use only named injection in Java code and CheBootstrap handles injection of all properties and environment variables and has everything needed to improve code readability.
If you want to know more about that, please, ask.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
amisevsk: You're right, that was the goal. I'll remove this check and rename the variable for clarity.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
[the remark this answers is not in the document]
amisevsk: Thank you, that makes a lot of sense! I've updated the PR to use a property (currently called `che.docker.ip.useinternaladdress` -- open to suggestions).

A few questions regarding the conversion of env vars to properties though: 

1) in `CheBootstrap`,  I see that env vars are converted by lowercasing them and replacing underscores with periods. However, some of the properties have underscores in their name, which makes it impossible to set them with environment variables, and prevents the new property here from having underscores (`che.docker.ip.use_internal_address` would make more sense).

2) In `LocalDockerInstanceRuntimeInfo`, it looks like this convention is ignored: the methods `externalHostnameWithPrecedence` and `internalHostnameWithPrecedence` use environment variables to override their behavior -- is this intentional or should it be fixed?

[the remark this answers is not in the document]
garagatyi: 1) I submitted a PR to fix that. I'm going to merge it soon. #2454 
2) Probably it was implemented before we merged feature with automatic conversion. Or we missed it in PR review. 
I believe that it should be fixed. @amisevsk If you want you can fix it also.
@TylerJewell should env variables be renamed to match properties names or it is better to try to use new property aliases feature?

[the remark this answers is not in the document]
amisevsk: @garagatyi Thanks for pointing me to that PR -- I've renamed the property to `che.docker.ip.use_internal_address` for readability. This has the side effect of not being able to override the property with an environment variable until your change is merged.

I've modified `LocalDockerInstanceRuntimeInfo` to not use environment variables anymore and updated the docs. Previously, the properties `che.docker.ip` and `che.docker.ip.external` were overridden through both the environment variable `CHE_DOCKER_IP`/`CHE_DOCKER_IP_EXTERNAL` but also through `CHE_DOCKER_MACHINE_HOST` and `CHE_DOCKER_MACHINE_HOST_EXTERNAL`. With the change only the former has an effect.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
```diff
@@ -315,11 +318,21 @@ public String projectsRoot() {
             String portProtocol = portEntry.getKey();
             // we are assigning ports automatically, so have 1 to 1 binding (at least per protocol)
             String externalPort = portEntry.getValue().get(0).getHostPort();
+
+            // If we are sending messages to the container directly, we don't want to use the mapped ports.
+            String internalHostnameAndPort;
+            if (useInternalAddress) {
+                String internalPort = portProtocol.split("/")[0];
+                internalHostnameAndPort = internalHostname + ":" + internalPort;
```
garagatyi, line 326: In that case it is supposed that browser will connect to ephemeral port or exposed port? I'm asking because as far as I remember it should be exposed port, but I'm not sure.

garagatyi (changes requested):

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
garagatyi: Do you really mean exposed port or ephemeral port which is mapped to exposed port?

@l0rd Maybe you can help to understand how this workflow is supposed to work? I know you explain such things very clear.

l0rd: @amisevsk there is a typo in variable `externalHostame`. You haven't introduced that (that was probably me 😊) but it would be cool if you could fix it => `externalHostname`

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalDockerInstanceRuntimeInfo.java
```diff
@@ -58,45 +51,57 @@ public LocalDockerInstanceRuntimeInfo(@Assisted ContainerInfo containerInfo,
...
     private static String internalHostnameWithPrecedence(String internalHostnameProperty,
-                                                              String internalHostnameAssisted) {
+                                                         String internalHostnameAssisted,
+                                                         NetworkSettings networkSettings,
+                                                         boolean useInternalAddress) {
+
+        if (useInternalAddress) {
+            String containerHostName = null;
+            if (networkSettings != null) {
+                containerHostName = networkSettings.getIpAddress();
```
l0rd, line 94: Is this IP address the IP of the Che server container? What we need is instead the IP addresses of the workspace containers.

l0rd (changes requested):

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -142,6 +142,12 @@ che.docker.ip=NULL
 # This is unusual, but happens for example in Docker for Mac when containers are in a VM.
 che.docker.ip.external=NULL
 
+# If true, then uses the internal address and port of workspace Docker containers (i.e. within the Docker
+# Docker network) instead of the address and port provided by the Docker daemon. May be necessary if the
```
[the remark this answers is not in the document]
amisevsk: I've updated the sentence.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalDockerInstanceRuntimeInfo.java
l0rd: @amisevsk you are right. Sorry about that :-)

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
l0rd: @amisevsk I understand now that `internaHostname` can be either `containerIP` or `internalHostname` here. The logic to set the right is in class `LocalDockerInstanceRuntimeInfo`.

l0rd (approved):

l0rd:
LGTM good job @amisevsk!

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
garagatyi: Ok, now it is clearer for me. Thanks @l0rd and @amisevsk. Mario explains such things in a really clear way, as always. :smile:

garagatyi:
@l0rd @amisevsk Honestly existing code is already over-complicated. And this PR makes what is happening here even less clear. I'll suggest a way how to refactor this code below. Please share your thoughts about this idea. Do you think that this approach is clearer and more extensible than current one?

We can refactor this code in such a way:
DockerInstanceRuntimeInfo accepts:
- servers internal host (nullable, can be IP)
- container JSON data object
- provider of new class which represents host&port evaluation strategy 

... other things not related to this topic
it doesn't accept:
- servers external host, since it is always set to null for now - looks like it is useless now

HostPortEvaluationStrategyProvider (name just for example, feel free to use another name) uses property that defines chosen strategy impl and set of strategies impls and return needed HostPortEvaluationStrategy implementation. 
HostPortEvaluationStrategyProvider accepts internal host and container JSON and pass it to HostPortEvaluationStrategy impl.
DockerInstanceRuntimeInfo calls HostPortEvaluationStrategy#getServer(String exposedPortProtocol).
HostPortEvaluationStrategy impl forms external/internal addresses using its own logic.
In that case each vendor may bind his own implementation of HostPortEvaluationStrategy and set it as current one for Che assembly.

My description looks a bit complicated when I read it, but it is not really complex. So please ask me for an explanation if this algorithm is not clear enough to you.

l0rd:
@garagatyi I like your approach and I agree that current code is overcomplicated. Just a couple of comments:
- Why do you think that servers external host is useless? It should be used when using docker4mac or with K8s/OpenShift etc..
- Would it be ok to do the refactoring as a separate PR? That would make the refactoring PR easier to review.

garagatyi:
> Why do you think that servers external host is useless? It should be used when using docker4mac or with K8s/OpenShift etc..

I mean it is useless in DockerInstanceRuntimeInfo constructor. Later I would want to refactor DockerInstance code again and remove DockerNode because it looks odd to me. I think we can move internal/external address setting into suggested HostPortEvaluationStrategy.
WDYT?

> Would it be ok to do the refactoring as a separate PR? That would make the refactoring PR easier to review

Do you suggest to merge this PR and make another one with refactoring or just open another one with the refactoring? If first then we will have to pass 2 QA cycles - for 1st and 2nd PRs.

l0rd:
> I mean it is useless in DockerInstanceRuntimeInfo constructor. Later I would want to refactor DockerInstance code again and remove DockerNode because it looks odd to me. I think we can move internal/external address setting into suggested HostPortEvaluationStrategy.
> WDYT?

Ok I see what you mean. It makes sense. And we should move internalhost parameter to HostPortEvaluationStrategy too.

> Do you suggest to merge this PR and make another one with refactoring or just open another one with the refactoring? If first then we will have to pass 2 QA cycles - for 1st and 2nd PRs.

I guess that's not ideal because QA cycles are manual and take time. That's ok I will talk with @amisevsk to see how to handle this.

amisevsk:
@garagatyi I agree that the current setup is probably more complicated than it should be, and would be happy to do the refactor you suggest.

For my understanding: Currently, `DockerInstanceRuntimeInfo` takes internal/external hostname, the `ContainerInfo` json, and Sets of `ServerConf`. When `DockerInstanceRuntimeInfo#getServers()` is called, `DockerInstanceRuntimeInfo`

1) gets port mappings and labels from `ContainerInfo` json
2) calls `DockerInstanceRuntimeInfo#getServersWithFilledPorts()`, which
    1) parses the ports json object from (1)
    2) for each port mapping (e.g. `22/tcp=[PortBinding{hostIp='0.0.0.0', hostPort='32898'}]`) gets the protocol (22/tcp) and the external port (32898)
    3) adds an entry in a `Map<String, ServerImpl>` which pairs the desired port (32898) with internal and external hostname (from the constructor)
3) The Map from (2) is passed to `DockerInstanceRuntimeInfo#addRefAndUrlToServers()` which updates each `ServerImpl`, adding ref (e.g 22/tcp -> ssh) and relevant parts of `ServerPropertiesImpl` (e.g. path, internal URL)
4) The updated Map from (3) is passed to `DockerInstanceRuntimeInfo#AddDefaultReferenceForServersWithoutReference()`, which puts a default value for ref if it is not set.

The result of (4) is what is returned by `getServers()`. 

If I understand you correctly, you're suggesting that steps 2-4 are moved into the `HostPortEvaluationStrategy` impl, and `DockerInstanceRuntimeInfo#getServers()` instead iterates over the port mappings obtained in (1), calling `HostPortEvaluationStrategy#getServer(String portProtocol)` and adding the returned `ServerImpl` to the map it returns.

That makes sense to me. A few things I'm still not clear on though. It's maybe important to note that with the set up I'm running, the constructor for `DockerInstanceRuntimeInfo` is invoked from `LocalDockerInstanceRuntimeInfo`.

1) How do we define HostPortEvaluationStrategies? Currently, the modes of function defined by properties seem to be 
    - Default: use InternalHostname for internal and external address, and external ports
    - If containerExternalHostname is provided, use that for external address
    - (My changes) If property `che.docker.ip.use_internal_address` is set, ephemeral ports should be used on internal addresses instead. Additionally, we assume that `containerInternalHostname` is set correctly so that a route can be established.

    This gets muddy quickly; these don't really represent different strategies for evaluating address and ports, so much as they are two options applied on top of default. This means potentially 4 different strategies, since the two options don't affect one another. Or is there one default strategy that takes these options into account, with the potential for more strategies in the future?

2) I'm not sure I totally understand the role of `LocalDockerInstanceRuntimeInfo`, as it seems to just call the `DockerInstanceRuntimeInfo` constructor with the parameters we want.
It is where the options listed above actually take effect, since 
	- Its constructor is injected with `che.docker.ip.external`, which it uses as `containerExternalHostname` in `DockerInstanceRuntimeInfo`
	- With my changes above, it's also resposible for setting `containerInternalHostname` correctly

   This kind of obfuscates the effects of these properties and I think adds to the confusion. Would the refactor get rid of this class, and move this functionality into `HostPortEvaluationStrategy`? I like this idea since with my changes, the property has to be injected twice -- in `LocalDockerInstanceRuntimeInfo` to set `containerInternalHostname` correctly, and in `DockerInstanceRuntimeInfo` to cause it to use ephemeral ports.

Sorry for the novel, I want to make sure I understand.

garagatyi:
> If I understand you correctly, you're suggesting...

Yes, It is what I'm thinking about.

> How do we define HostPortEvaluationStrategies? ... 

Yes, I suppose it should be a set of different strategies. They can be simple enough and don't need to override some default values. Apparently we will have some default strategy. I assume that we can set property with some strategy as a default and allow to override strategy with properties instead of changing assembly by binding needed strategy in Guice module. If vendor wants to implement new strategy it can contribute it into Che or bind it in customized assembly.

> I'm not sure I totally understand the role of LocalDockerInstanceRuntimeInfo ... 

Yes, I believe we should get rid of LocalDockerInstanceRuntimeInfo

amisevsk:
I've opened a new pull request which includes the refactor suggested above: #3282

amisevsk:
Closed; issue will be solved in #3282

Issue #2811, opened 2016-10-18
Some commands that have worked through bash shell, can work wrong (regression)


musienko-maxim:
**Reproduction Steps:**

After setup of **sh** as default shell in the Codenvy - some commands that was saved early will be wrong. For example some workspace used command like: `echo -e 'some text for check' > ~/check.txt.`

**Expected behavior:**
After execution the command - should be created the check.txt file with next content: **some text for check**

**Observed behavior:**
The content in the file will be like: **-e some text for check**  (-e flag will be added into file)

This problem may do many unexpected results for commands, that were saved in the workspaces created earlier. The same problem can be in the factories, which use bash commands. Many  users may be puzzled.

**Che version:**  5.0.0-M6-SNAPSHOT  
**OS and version:** Ubuntu 14.04 LTS
**Docker version:**  1.12.2
**Che install:**   Docker container

JamesDrummond:
Add current-sprint label.

benoitf:
@musienko-maxim please check with next nightly

musienko-maxim:
The test (LoginBySshTest) that covers this feature - passed. I think issue can be closed.

PR #2838, opened 2016-10-19
CHE-2811 Launch all user commands by using bash if shell is not defined.

benoitf:
### What does this PR do?

For example now that sh is default one, previous user commands may fail. If the commands are required to be launched with sh, then /bin/sh will need to be defined as shell invoker but default is kept to bash

https://github.com/eclipse/che/issues/2811
### Previous behavior

it is now executing all user + internal commands with /bin/sh which may fail for user commands if user has a previous command that was using bash
so "all internal commands" use sh by default, user commands use bash
### New behavior

Keep bash for user defined method unless the attribute "shell" is overrided to be an other interpreter

Change-Id: I19e2065bfd9e6ee93a5d899cb67cd5b9ece76f62


benoitf:
@vparfonov ?

vparfonov (approved):

Issue #2687, opened 2016-10-03
Compose Syntax Links Not Working.


JamesDrummond:
Compose syntax `links` is currently not working. The workaround is to use `depends_on` in lieu of `links`.

**Expected behavior:**

Use `Links` in compose recipe for use in creating stacks/workspaces.

**Observed behavior:**

Use `Links` in compose recipe for use in creating stacks/workspaces will fail. Have to use `depends_on` instead.

**Che version:**    5.0.0-M1   

**Additional information:**
- Problem started happening recently, didn't happen in an older version of Che: No
- Problem can be reliably reproduced, doesn't happen randomly: Yes

bmicklea:
This is critical to M6.

riuvshin:
Moved to M7

mmorhun:
Done

PR #2850, opened 2016-10-20
Fix broken links section in docker-compose recipe

mmorhun:
### What does this PR do?

Fix broken links section in docker-compose recipe.
#2687
### Previous behavior

If links section is present in recipe than workspace will fail to start.
### New behavior

Links and aliases works now.


mmorhun:
@garagatyi @benoitf @l0rd @AndrienkoAleksandr review please

mmorhun:
I'll add tests a bit later, maybe in M7

AndrienkoAleksandr:
+1

gazarenkov:
What is the reason of names like "...CorrectSource"?
It is supposed to have some "...IncorrectSources" and things as well?
Things like this makes the code less readable

mmorhun:
Agree, I'll change the name.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/NetworkSettings.java
```diff
@@ -17,7 +17,7 @@
 
 /** @author andrew00x */
 public class NetworkSettings {
-    private String   ipAddress;
+    private String   iPAddress;
```
benoitf: I would keep ipAddress as it was as method name is getIpAddress() not getIPAddress()

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
```diff
@@ -930,6 +933,45 @@ private Machine getMachineWithCorrectSource(MachineImpl machine, MachineSource m
...
+     * If alias is omitted then service name will be used.
+     *
+     * @param serviceToStart
+     *         service which is starting now
+     * @param launchedServices
+     *         already running services
+     */
+    @VisibleForTesting
+    void replaceServiceToContainerNameInLinks(CheServiceImpl serviceToStart, List<CheServiceImpl> launchedServices) {
+        final List<String> containerLinks = serviceToStart.getLinks();
```
benoitf: we couldn't use Java8 stream/filter/collect for the whole method ?

benoitf (commented):
For me, the tests should be added before merging as I don't see how we can be sure without testing.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/NetworkSettings.java
mmorhun: Yes, `ipAddress` is conventional name , but `iPAddress` is needed for parsing value from json. Exactly like  `iPPrefixLen`, `iPv6Gateway` in the same class.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
mmorhun: Yes, we can, but I think that it is redundant here.
If we used stream here it would made code more complicated and we would had additional expenses for stream creation/collection here (note, that in many cases list with links contains 0 or 1 element, i.e. how often and how many links you use in docker compose?)

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/NetworkSettings.java
l0rd: I agree with @benoitf that iPAddress is weird. Wouldn't it possible to use `@SerializedName` annotation instead?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
```diff
@@ -887,47 +889,86 @@ Instance startMachine(LineConsumer machineLogger,
...
+                                                     .build();
+
+        }
+        return machineWithNormalizedSource;
+    }
+
+    /**
+     * Replaces linked to this service's name with container name which represents the service in links section.
+     * The problem is that a user writes names of other services in links section in compose file.
+     * But actually links are constraints and they values should be names of containers (not services) to be linked.
```
l0rd: and they values => and their values

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
```diff
@@ -887,47 +889,86 @@ Instance startMachine(LineConsumer machineLogger,
...
+     * @param serviceToStart
+     *         service which is starting now
+     * @param services
+     *         all services in environment
+     */
+    @VisibleForTesting
+    void replaceServiceToContainerNameInLinks(CheServiceImpl serviceToStart, Map<String, CheServiceImpl> services) {
+        final List<String> containerLinks = serviceToStart.getLinks();
+        for (int i = 0;  i < containerLinks.size(); i++) {
+            String link = containerLinks.get(i);
```
l0rd: Extracting the body of the for loop in a new method would make it more readable

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
```diff
@@ -887,47 +889,86 @@ Instance startMachine(LineConsumer machineLogger,
...
+     *
+     * @param serviceToStart
+     *         service which is starting now
+     * @param services
+     *         all services in environment
+     */
+    @VisibleForTesting
+    void replaceServiceToContainerNameInLinks(CheServiceImpl serviceToStart, Map<String, CheServiceImpl> services) {
+        final List<String> containerLinks = serviceToStart.getLinks();
+        for (int i = 0;  i < containerLinks.size(); i++) {
```
l0rd: You could probably use for each here (`for (String link: containersLinks) {`)

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
```diff
@@ -887,47 +889,86 @@ Instance startMachine(LineConsumer machineLogger,
...
+            String link = containerLinks.get(i);
+            String serviceName;
+            String serviceAlias;
+            int colonIndex = link.indexOf(':');
+            if (colonIndex != -1) {
+                serviceName = link.substring(0, colonIndex);
+                serviceAlias = link.substring(colonIndex + 1);
```
l0rd: What happen if user provide a (bad) link with an empty alias like: `serviceDb:`

l0rd: `split()` method is probably more appropriate here instead of manually parsing the links string.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
```diff
@@ -887,47 +889,86 @@ Instance startMachine(LineConsumer machineLogger,
...
+            } else {
+                serviceName = link;
+                serviceAlias = null;
+            }
 
+            CheServiceImpl serviceLinkedTo = services.get(serviceName);
+            if (serviceLinkedTo != null) {
```
l0rd: An error should be returned if the user is trying to link a service that does not exist (i.e. serviceLinkedTo == null)

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
```diff
@@ -887,47 +889,86 @@ Instance startMachine(LineConsumer machineLogger,
...
+                String containerNameLinkedTo = serviceLinkedTo.getContainerName();
+                containerLinks.set(i, (serviceAlias == null) ?
```
l0rd: Instead of modifying current containersLinks list you could make this method return a new list

wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/environment/server/CheEnvironmentEngineTest.java
```diff
@@ -895,6 +895,43 @@ public void shouldBeAbleToRemoveSnapshot() throws Exception {
         verify(instanceProvider).removeInstanceSnapshot(machineSource);
     }
 
+    @Test
+    public void shouldReplaceServiceNameWithContainerNameAndAliasInLinks() {
```
l0rd: You are testing 2 things here (links with and without alias). You could probably split this test in 2 separate tests.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
mmorhun: thx

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
mmorhun: No, I cannot. I need to change the list. I can use `Iterator` here, but I think that current version is more readable.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
mmorhun: Taking into account, that we have only one loop in the method and this loop actually is method's body and body of this loop cannot be reused somewhere and if so, we should pass iterator and links list into new method as parameters, IMO, there is no point to extract loop body into another method. @garagatyi WDYT?

garagatyi: Since you are going to refactor this code lets figure out what can be cleaned up after that refactoring

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
mmorhun: I think, that error should be reported. Please note, that things like that are validated by `CheEnvironmentValidator`.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
mmorhun: If user provide link with empty alias, like `serviceDb:`, than `CheEnvironmentValidator` will throw an exception and a user will be notified about syntax error.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
mmorhun: But in case of no links we need additional logic which made code a bit complicated or return new empty list when it can be untouched, which is a bit weird to me.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
```diff
@@ -584,16 +584,60 @@ private void normalize(String namespace,
...
+     * @param serviceToNormalizeLinks
+     *         service which links will be normalized
+     * @param services
+     *         all services in environment
+     */
+    @VisibleForTesting
+    void normalizeLinks(CheServiceImpl serviceToNormalizeLinks, Map<String, CheServiceImpl> services) {
+        final List<String> containerLinks = serviceToNormalizeLinks.getLinks();
+        for (int i = 0;  i < containerLinks.size(); i++) {
+            String serviceNameAndAliasToLink[] = containerLinks.get(i).split(":"); // a link has format: 'name:alias' or 'name'
```
garagatyi: Pass second argument to split

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
```diff
@@ -584,16 +584,60 @@ private void normalize(String namespace,
...
+            String serviceAlias = (serviceNameAndAliasToLink.length > 1) ? serviceNameAndAliasToLink[1] : null;
+            CheServiceImpl serviceLinkTo = services.get(serviceName);
+            if (serviceLinkTo != null) {
+                String containerNameLinkTo = serviceLinkTo.getContainerName();
+                containerLinks.set(i, (serviceAlias == null) ?
+                                      containerNameLinkTo + ':' + serviceName :
+                                      containerNameLinkTo + ':' + serviceAlias);
+            } else {
+                // should never happens. Errors like this should be filtered by CheEnvironmentValidator
+                LOG.error("Attempt to link non existing service {} to {} service.", serviceName, serviceToNormalizeLinks);
```
garagatyi: It is better to throw IllegalArgumentExcption

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
```diff
@@ -584,16 +584,60 @@ private void normalize(String namespace,
...
+        final List<String> containerLinks = serviceToNormalizeLinks.getLinks();
+        for (int i = 0;  i < containerLinks.size(); i++) {
+            String serviceNameAndAliasToLink[] = containerLinks.get(i).split(":"); // a link has format: 'name:alias' or 'name'
+            String serviceName = serviceNameAndAliasToLink[0];
+            String serviceAlias = (serviceNameAndAliasToLink.length > 1) ? serviceNameAndAliasToLink[1] : null;
+            CheServiceImpl serviceLinkTo = services.get(serviceName);
+            if (serviceLinkTo != null) {
+                String containerNameLinkTo = serviceLinkTo.getContainerName();
+                containerLinks.set(i, (serviceAlias == null) ?
+                                      containerNameLinkTo + ':' + serviceName :
```
garagatyi: Is it needed? I suppose it already works in such a way?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
```diff
@@ -584,16 +584,60 @@ private void normalize(String namespace,
...
+     *
+     * @param serviceToNormalizeLinks
+     *         service which links will be normalized
+     * @param services
+     *         all services in environment
+     */
+    @VisibleForTesting
+    void normalizeLinks(CheServiceImpl serviceToNormalizeLinks, Map<String, CheServiceImpl> services) {
+        final List<String> containerLinks = serviceToNormalizeLinks.getLinks();
+        for (int i = 0;  i < containerLinks.size(); i++) {
```
garagatyi: I would suggest to simplify code with stream API - it looks overcomplicated

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
```diff
@@ -667,6 +711,7 @@ private void startEnvironmentQueue(String namespace,
         try {
             machineProvider.createNetwork(networkId);
 
+            EnvironmentHolder environmentHolder;
```
garagatyi: Is it still used?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
```diff
@@ -736,7 +781,7 @@ private void startEnvironmentQueue(String namespace,
                 boolean queuePolled = false;
                 try (StripedLocks.WriteLock lock = stripedLocks.acquireWriteLock(workspaceId)) {
                     ensurePreDestroyIsNotExecuted();
-                    EnvironmentHolder environmentHolder = environments.get(workspaceId);
+                    environmentHolder = environments.get(workspaceId);
```
garagatyi: Is it still used?

garagatyi (changes requested):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
mmorhun: ok

wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/environment/server/CheEnvironmentEngineTest.java
```diff
@@ -895,6 +895,65 @@ public void shouldBeAbleToRemoveSnapshot() throws Exception {
         verify(instanceProvider).removeInstanceSnapshot(machineSource);
     }
 
+    @Test
+    public void shouldReplaceServiceNameWithContainerNameInLinks() {
```
garagatyi, line 919: Please ensure that code of MachineProviderImpl is also covered

garagatyi (changes requested):

mmorhun:
@garagatyi @benoitf @l0rd take a look again please

garagatyi (approved):
Good job!

l0rd:
Nice @mmorhun! LGTM

benoitf (approved):

PR #2856, opened 2016-10-21
Render hover messages according to LSP specification

kaloyan-raev:
### What does this PR do?

If the hover message consists of multiple `MarkedString`s then they should be joined via double newline, so they are rendered as separate Markdown paragraphs.

If a `MarkedString` has the `language` property provided then it should be rendered as a Markdown code block.
- PHP support #2590
### Previous behavior

hover-before
### New behavior

hover-after


benoitf:
ci-build

Issue #2816, opened 2016-10-18
Find in path does not work


ghost:
Open any project. Hit Ctrl+Shift+F to search for some text. 0 occurrences found even though files contain the text being searched.

ghost:
Reopened since Ctrl+Shift+F does not work for php, xml, h, c and cpp files.

PR #2857, opened 2016-10-21
CHE-2816. Correctly applying exclude filters at indexing files

RomanNikitenko:
### What does this PR do?
- avoid to duplicate '*' for searching request
- correctly applying exclude filters at indexing files

Current behavior: we need to have exclude filters here https://github.com/eclipse/che/blob/master/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/search/impl/LuceneSearcher.java#L95 
and we have some filters for files that should NOT be indexed. But  org.eclipse.che.api.vfs.search.MediaTypeFilter  includes in result files that SHOULD  be indexed. 
New behavior: the media type filter will include in result files with media type that should NOT be indexed.
#2816


skabashnyuk (changes requested):
`correctly applying exclude filters at indexing files`
This is an unclear statement. Please specify why and how are you going to fix it.

RomanNikitenko:
@skabashnyuk I've added some explanation to clarify how I'm going to fix it

skabashnyuk (approved):

vparfonov (approved):

Issue #2624, opened 2016-09-28
Array format unsupported for environment in compose file


ghost:
Docker compose environment is accepted as array and dictionary - https://docs.docker.com/compose/compose-file/#/environment. However, array format is unsupported in Che.

**Reproduction Steps:**
1. Create a workspace with the following compose file

```
services: 
 dev-machine: 
  image: codenvy/ubuntu_jdk8
  environment: 
   - MYSQL_ROOT_PASSWORD=root 
   - MYSQL_DATABASE=utqg 
```

**Expected behavior:**
Support dictionary and array.

```
environment:
  RACK_ENV: development
  SHOW: 'true'
  SESSION_SECRET:

environment:
  - RACK_ENV=development
  - SHOW=true
  - SESSION_SECRET
```

**Observed behavior:**

```

Parsing of recipe of environment 'wksp-rkq9' failed. Error: Parsing of environment configuration failed. Can not deserialize instance of java.util.LinkedHashMap out of START_ARRAY token
at [Source: java.io.StringReader@29a1353; line: 5, column: 4] (through reference chain: org.eclipse.che.api.environment.server.compose.model.ComposeEnvironmentImpl["services"]->java.util.LinkedHashMap["dev-machine"]->org.eclipse.che.api.environment.server.compose.model.ComposeServiceImpl["environment"])
```

JamesDrummond:
@TylerJewell Does this need to be put into 5.0.0-M5?

TylerJewell:
Let R&D decide when they take it.

dmytro-ndp:
Error handling:
![wrong_env_value_exception-24_10_2016](https://cloud.githubusercontent.com/assets/1197777/19638587/519098d0-99dd-11e6-8a88-48a3a49b5923.png)

riuvshin:
moved to M7

PR #2865, opened 2016-10-23
Support array format for environment in compose file

dmytro-ndp:
### What does this PR do?

For issue #2624 it adds support of array format of environment variables of _docker compose file_ into the **ComposeServiceImpl** class.
### Note:

next variant of class [ComposeServiceImpl](https://github.com/eclipse/che/blob/af4de272e06ec8213eab5a7c4f69ad72088b3d2e/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/ComposeServiceImpl.java#L29):

```
    /** Parse dictionary in view of:
       "environment:
         key1: value1
         key2: value2"
    */
    public void setEnvironment(Map<String, String> environment) {
        this.environment = environment;
    }

    /** Parse array in view of:
       "listEnvironment:
         - key1=value1
         - key2=value2"
    */
    public void setEnvironment(List<String> listEnvironment) {
        // convert array to Map<String, String>{ key1: value1, key2: value2 }
        try {
            this.environment = listEnvironment.stream()
                                              .collect(Collectors.toMap(item -> item.split("=")[0],
                                                                        item -> item.split("=")[1]));
        } catch (Exception e) {
            throw new RuntimeException(format("Unsupported value '%s'.", listEnvironment.toString()));
        }
    }
```

leads to parsing error:

> java.lang.IllegalArgumentException: Parsing of environment configuration failed. Conflicting setter definitions for property "environment": org.eclipse.che.api.environment.server.compose.ComposeServiceImpl#setEnvironment(1 params) vs org.eclipse.che.api.environment.server.compose.ComposeServiceImpl#setEnvironment(1 params)
>     at org.eclipse.che.api.environment.server.compose.ComposeFileParser.parse(ComposeFileParser.java:51)

@garagatyi: please, review this request.

Signed-off-by: Dmytro Nochevnov dnochevnov@codenvy.com


TylerJewell:
Nice!  Can you add an example of both array and dictionary format of the same compose syntax?

dmytro-ndp:
@TylerJewell:

> Can you add an example of both array and dictionary format of the same compose syntax?

#### Array format

```
services: 
 dev-machine: 
  image: codenvy/ubuntu_jdk8
  environment: 
   - MYSQL_ROOT_PASSWORD=root 
   - MYSQL_DATABASE=utqg
```

#### Dictionary format

```
services: 
 dev-machine: 
  image: codenvy/ubuntu_jdk8
  environment: 
   MYSQL_ROOT_PASSWORD: root 
   MYSQL_DATABASE: utqg 
```

It is impossible to use both formats, array and dictionary, in the one environment section:
array_and_dictionary_in_one_environment_error_24-10-2016

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
```diff
@@ -0,0 +1,74 @@
...
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import static java.lang.String.format;
+
+/**
+ * @author Dmytro Nochevnov
```
garagatyi, line 28: Please add description of class to javadocs

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
```diff
@@ -0,0 +1,74 @@
...
+                 - key2=value2"
+            */
+            if (environment instanceof List) {
+                // convert array to Map<String, String>{ key1: value1, key2: value2 }
+                return ((List<String>) environment).stream()
+                                                   .collect(Collectors.toMap(item -> item.split("=")[0],
+                                                                             item -> item.split("=")[1]));
+            }
+        } catch (Exception e) {
+            throw new RuntimeException(format("Unsupported value '%s'.", environment.toString()));
```
garagatyi: It is better to avoid throwing of another type of exception than it is declared in overridden method

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
```diff
@@ -0,0 +1,74 @@
...
+            /* Parse array in view of:
+               "environment:
+                 - key1=value1
+                 - key2=value2"
+            */
+            if (environment instanceof List) {
+                // convert array to Map<String, String>{ key1: value1, key2: value2 }
+                return ((List<String>) environment).stream()
+                                                   .collect(Collectors.toMap(item -> item.split("=")[0],
+                                                                             item -> item.split("=")[1]));
```
garagatyi: Usage of 2 splits makes parser heavier then it could be because it is regexp parsing. Please use it only once. Also use splitting with limited number of result entries.

garagatyi: It is also possible that there is no second item in an array and index out of bound exception will be thrown in that case.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
```diff
@@ -0,0 +1,74 @@
...
+            }
+        } catch (Exception e) {
+            throw new RuntimeException(format("Unsupported value '%s'.", environment.toString()));
+        }
+
+        /* Work around empty environment */
```
garagatyi: Please use double slashes for single line comments

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
```diff
@@ -0,0 +1,74 @@
...
+        if (environment instanceof String) {
```
garagatyi, line 65: Please describe (not in code - just in comments to PR) how it works for better understanding of other maintainers

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
```diff
@@ -0,0 +1,74 @@
...
+            if (((String) environment).isEmpty()) {
+                return new HashMap<>();
+            }
+
+            throw new RuntimeException(format("Unsupported value '%s'.", environment.toString()));
```
garagatyi: Method already declares 2 exceptions, why not use one of them in that case?

wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/environment/server/compose/EnvironmentContextTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * @author Dmytro Nochevnov
+ */
+public class EnvironmentContextTest {
+
+    ComposeFileParser parser = new ComposeFileParser();
+
+    @Test
+    public void shouldParseDictionaryTypeEnvironment() throws ServerException {
```
garagatyi: Please add tests for next cases:
- map format, value of variable is empty
- map format, value of variable contains colon sign
- map format, no colon in entry
- array format, no equal sign in entry
- array format, empty value of variable
- array format, value of variable contains equal sign

garagatyi (changes requested):
General approach is OK, but cleanup is needed

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
dmytro-ndp: You are right, This case is being handled by `catch (Exception e)` statement below.

dmytro-ndp: Please, explain what limit did you mean:

> Also use splitting with limited number of result entries.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
dmytro-ndp: In case of empty "environment:" section Jackson returns empty String value from jsonParser.readValueAs(Object.class), but I can reproduce it only in test.
Dashboard seems remove empty "environment" section at all:
empty_environment_section_2016-10-26

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
dmytro-ndp: IOException is not suitable here, IMHO.
Jackson handles RuntimeException correctly in this place
wrong_env_value_exception-24_10_2016:

garagatyi: Can't say that this error is friendly. Can we do something to improve it readability and avoid java code in error message?

dmytro-ndp: Yes. We can throw **JsonMappingException**.
It's more reasonable and clear.

dmytro-ndp:
@garagatyi: please, review fixup above.

dmytro-ndp:
@garagatyi: please, review fixup above.

garagatyi:
Please fix build

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
```diff
@@ -0,0 +1,88 @@
...
+            /* Parse array in view of:
+               "environment:
+                 - key1=value1
+                 - key2=value2"
+            */
+            if (environment instanceof List) {
+                // convert array to Map<String, String>{ key1: value1, key2: value2 }
+                Map<String, String> map = new HashMap<>();
+                for (String item : (List<String>) environment) {
+                    String[] splitResult = item.split("=");
```
garagatyi: Limit split result array size

wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/environment/server/compose/EnvironmentContextTest.java
```diff
@@ -0,0 +1,171 @@
...
+            },
+
+            // array type environment
+            {"services: \n"
+             + " dev-machine: \n"
+             + "  image: codenvy/ubuntu_jdk8\n"
+             + "  environment:\n"
+             + "   - MYSQL_ROOT_PASSWORD=root\n"
+             + "   - MYSQL_DATABASE=db",
+             ImmutableMap.of("MYSQL_ROOT_PASSWORD", "root",
```
garagatyi, line 80: Please add test for value with equal sign in array form of environment

wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/environment/server/compose/EnvironmentContextTest.java
```diff
@@ -0,0 +1,171 @@
...
+            // array format, empty value of variable
+            {"services: \n"
+             + " dev-machine: \n"
+             + "  image: codenvy/ubuntu_jdk8\n"
+             + "  environment:\n"
+             + "   =value ",
+             "Parsing of environment configuration failed. Unsupported value '=value'.*(?s).*"
+            },
+
+            // array format, value of variable contains equal sign
```
garagatyi: I suppose the reason why this sample fails is that dash is missing

garagatyi (changes requested):
Please fix build and commented lines

garagatyi:
@sleshchenko Please take a look

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
```diff
@@ -0,0 +1,88 @@
...
+                        throw ctxt.mappingException(format("Unsupported value '%s'.", item));
+                    }
+
+                    map.put(splitResult[0].trim(), splitResult[1].trim());
+                }
+
+                return map;
+            }
+        } catch (JsonMappingException je) {
+            throw je;
```
sleshchenko: Why you can't just not catch this exception? Then it will be rethrown automatically, no?

garagatyi:
Looks like build was broken in master, not in this branch

wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/environment/server/compose/EnvironmentContextTest.java
```diff
@@ -0,0 +1,171 @@
...
+    @DataProvider
+    public Object[][] incorrectContentTestData() {
+        return new Object[][] {
+            // unsupported type of environment
+            {"services: \n"
+             + " dev-machine: \n"
+             + "  image: codenvy/ubuntu_jdk8\n"
+             + "  environment:\n"
+             + "   true",
+             "Parsing of environment configuration failed. Unsupported type 'class java.lang.Boolean'.*(?s).*"
```
sleshchenko: Why do you need `.*` before and after `(?s)`. after is enough I think. Mb you mean `\\.(?s).*`

dmytro-ndp:
@garagatyi, @sleshchenko: please, review new fixup.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
```diff
@@ -0,0 +1,88 @@
...
+        } catch (Exception e) {
+            throw jsonMappingException;
+        }
+
+        // work around empty environment
+        if (environment instanceof String) {
```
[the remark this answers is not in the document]
dmytro-ndp: Because of this block doesn't throw an error.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
dmytro-ndp: Because I want to provide more specific error message with wrong item content.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
dmytro-ndp: I would rewrite the code to don't use "split()" at all as an error-prone in this case

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
```diff
@@ -85,4 +83,21 @@
         */
         throw ctxt.mappingException(format("Unsupported type '%s'.", environment.getClass()));
     }
+
+    /**
+     * Parse "key=value" item.
```
garagatyi: It is common practise to separate main description of class from additional description and from params list with `</p>` line. Please consider such formatting.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
```diff
@@ -85,4 +83,21 @@
...
+     * @param item - environment array item to parse
+     * @return map of key/value which are divided by "=" sign in item.
+     * @throws IllegalArgumentException if there is no "=" sign in item, or if there is empty key in item
+     */
+    private Map<String, String> parseEnvironmentItem(String item) throws IllegalArgumentException {
+        if (!item.contains(ARRAY_ITEM_DIVIDER) || item.indexOf(ARRAY_ITEM_DIVIDER) == 0) {
+            throw new IllegalArgumentException(format(UNSUPPORTED_VALUE_MESSAGE, item));
+        }
+
+        String key = item.substring(0, item.indexOf(ARRAY_ITEM_DIVIDER));
```
garagatyi: I think it is better to use single `indexOf` call instead of `2 indexOf + contains`

garagatyi (approved):
LGTM
Please consider suggested improvements

dmytro-ndp:
@garagatyi:

> Looks like build was broken in master, not in this branch

There is an issue to fix build: https://github.com/eclipse/che/issues/2888

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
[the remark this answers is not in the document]
sleshchenko: So, when can this block https://github.com/eclipse/che/pull/2865/files#diff-f420bd07395539f8527c26d508ecdd7fR47 throw exception?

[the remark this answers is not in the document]
dmytro-ndp: Which block?

dmytro-ndp:
@sleshchenko: please, take a look on the fresh commit.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /**
+     * Parse "key=value" item.
+     * @param item - environment array item to parse
+     * @return map of key/value which are divided by "=" sign in item.
+     * @throws IllegalArgumentException if there is no "=" sign in item, or if there is empty key in item
+     */
+    private Map<String, String> parseEnvironmentItem(String item) throws IllegalArgumentException {
+        if (item.indexOf(ARRAY_ITEM_DIVIDER) < 1) {
```
sleshchenko: It would be better to save result of item.indexOf(ARRAY_ITEM_DIVIDER) in variable to avoid calcucation it three times

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
```diff
@@ -0,0 +1,103 @@
...
+        /* Work around unsupported type of environment content, for example, 'Boolean' in case of content:
+           "environment:
+             true"
+        */
+        throw ctxt.mappingException(format("Unsupported type '%s'.", environment.getClass()));
+    }
+
+    /**
+     * Parse "key=value" item.
+     * @param item - environment array item to parse
```
sleshchenko: Java doc is still unformatted correctly

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
```diff
@@ -0,0 +1,103 @@
...
+           "environment:
+             - key1=value1
+             - key2=value2"
+        */
+        try {
+            if (environment instanceof List) {
+                // convert array to Map<String, String>{ key1: value1, key2: value2 }
+                Map<String, String> map = new HashMap<>();
+                for (String item : (List<String>) environment) {
+                    map.putAll(parseEnvironmentItem(item));
```
sleshchenko: I'd make private method `Map<String, String> parseEnvironmentItem(List<String> items)` because it is quite strange to  invoke method putAll when you return one and only one element. Or you can return Pair. I prefer first way

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
```diff
@@ -0,0 +1,103 @@
...
+                }
+
+                return map;
+            }
+        } catch (IllegalArgumentException ie) {
+            throw ctxt.mappingException(ie.getMessage());
+
```
sleshchenko, line 59: remove extra line

sleshchenko (changes requested):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/compose/deserializer/EnvironmentDeserializer.java
dmytro-ndp: It looks like a correctly formatted.

sleshchenko: Take a look good documented class here https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/WorkspaceDao.java#L27
Also you can check your formatting by quick java doc feature. If you use it you'll see that there is double `-` after parameter

sleshchenko: Style is described here https://github.com/codenvy/codenvy-dev-resources/wiki/Formatting#generally

dmytro-ndp: thank you!

dmytro-ndp:
@sleshchenko: please, review next fresh fixup.

sleshchenko (approved):

Issue #2629, opened 2016-09-28
IDE is trying to get projects while the workspace is not started


slemeur:
The workspace is not yet started, but the IDE is displaying a loader indicating that it is trying to get the projects.

![screen shot 2016-09-28 at 14 21 53](https://cloud.githubusercontent.com/assets/1636769/18913297/6602ebac-8587-11e6-8aee-aee153fff5cf.png)

Either it is a side effect of the changes that we introduced with the new loader - or it is a wrong sequencing of the event. 

Getting the project should not be triggered before the workspace is fully started.

**Che version:**    5.0.0-M4

riuvshin:
moved to M8

PR #2867, opened 2016-10-24
Get projects after workspace has been already initialized

vzhukovs:
This commit also includes API reorganization. Move client service implementations from the core-api to the core-app. Code cleanup. Removed dependency to the `javax.inject` from the core-api to avoid duplicate using `@Inject` annotation from the `com.google.inject` and `javax.inject`. (Client side uses only `com.google.inject.Inject` annotation)

Created two components:
**WsAgentInitializer** - which is responsible to initialize workspace component on the client side. Component consumes developer machine object at startup stage.
**WsAgentMessageBusProvider** - which is responsible to initialize message bus after workspace has been initialized.

API movements:
`OAuthServiceClient` was moved from the `org.eclipse.che.ide.api.auth` to the `org.eclipse.ide.api.oauth`

Related issue: https://github.com/eclipse/che/issues/2629

@vparfonov @ashumilova @azatsarynnyy review it, please.


ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/WsAgentInitializer.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Initialize client's workspace agent component with specific developer machine. Usually, this interface
+ * is not intended to be implemented by the third-party components.
+ *
+ * @author Vlad Zhukovskyi
+ * @see DevMachine
+ * @since 5.0.0
+ */
+public interface WsAgentInitializer {
```
azatsarynnyy: As far as I understood this interface is used in IDE app only and not intended to be used by extensions.
So, if it isn't a part of IDE public API why do we need it in IDE API? May be it would be better to move it in IDE app module?

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/machine/WsAgentStateController.java
```diff
@@ -139,7 +150,7 @@ public WsAgentState getState() {
         return state;
     }
 
-    /** Returns instance of {@link MachineMessageBus}. */
+    /** {@inheritDoc} */
     public Promise<MessageBus> getMessageBus() {
```
azatsarynnyy, line 155: _@Override_ annotation is missed

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/machine/WsAgentStateController.java
```diff
@@ -81,23 +93,22 @@ public WsAgentStateController(WorkspaceServiceClient workspaceServiceClient,
...
     }
 
-    /**
-     * Initialize state of the agent.
-     *
-     * @param devMachine
-     *         development machine instance
-     */
+    /** {@inheritDoc} */
     public void initialize(DevMachine devMachine) {
```
azatsarynnyy: _@Override_ annotation is missed

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/ssh/SshServiceClientImpl.java
```diff
@@ -29,6 +31,7 @@
  *
  * @author Sergii Leschenko
  */
+@Singleton
```
azatsarynnyy: Why did you add this annotation?
It's already bound as singleton in CoreGinModule.

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/machine/WsAgentStateController.java
```diff
@@ -81,23 +93,22 @@ public WsAgentStateController(WorkspaceServiceClient workspaceServiceClient,
...
         this.availableServices = new ArrayList<>();
     }
 
-    /**
-     * Initialize state of the agent.
-     *
-     * @param devMachine
-     *         development machine instance
-     */
+    /** {@inheritDoc} */
```
azatsarynnyy: What is the purpose of an empty javadoc which contains _@inheritDoc_ tag only?
Javadoc tool and all popular IDEs override it automatically when generating documentation.
IMO we should add it only in case of 'real' overriding javadoc like [here](https://github.com/eclipse/che/blob/master/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/wizard/AbstractWizardPage.java#L34).
WDYT?

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/oauth/OAuthServiceClientImpl.java
```diff
@@ -8,18 +8,21 @@
...
 import org.eclipse.che.ide.rest.AsyncRequestCallback;
 import org.eclipse.che.ide.rest.AsyncRequestFactory;
 import org.eclipse.che.ide.rest.RestContext;
 
-import javax.inject.Inject;
-
 /**
  * @author Sergii Leschenko
  */
+@Singleton
```
azatsarynnyy: Why did you add this annotation?
It's already bound as singleton in CoreGinModule.

azatsarynnyy (commented):
:+1:

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/machine/WsAgentInitializer.java
vzhukovs: Yes, you're right, I'll move it to the IDE app

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/machine/WsAgentStateController.java
vzhukovs: It's just personal habbit. I'll remove it.

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/machine/WsAgentStateController.java
vzhukovs: Fixed

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/machine/WsAgentStateController.java
vzhukovs: Fixed

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/ssh/SshServiceClientImpl.java
vzhukovs: Yes, sure, fixed

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/oauth/OAuthServiceClientImpl.java
vzhukovs: Also fixed

ashumilova (approved):

azatsarynnyy (approved):

Issue #2832, opened 2016-10-19
Add servers to Agent description and propagate them to Workspace Runtime


gazarenkov:
**Motivation**:  Some agents (such as ws, terminal etc) suppose launching named servers which have to be visible runtime same way as configured Servers.

Add Servers to Agent declaration (List<Server> getServers() method) and make sure they propagated to Workspace Runtime

PR #2868, opened 2016-10-24
CHE-2832: Add servers to Agent description and propagate them to Workspace Runtime

tolusha:
### What does this PR do?

Adds servers to Agent description and propagates them to Workspace Runtime

https://github.com/eclipse/che/issues/2832


tolusha:
@vparfonov @garagatyi 
pls review

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/MachineInstanceProvider.java
```diff
@@ -51,8 +56,8 @@ Instance startService(String namespace,
                           boolean isDev,
                           String networkName,
                           CheServiceImpl service,
-                          LineConsumer machineLogger)
-            throws ServerException;
+                          LineConsumer machineLogger,
+                          List<? extends ServerConf> providedServers) throws ServerException;
```
garagatyi: This class works with kinda docker terminology. It knows nothing about serversConf and other Che model staf. So it is better to move this code into CheEnvironmentEngine that should apply serverConf to CheServiceImpl to describe docker container. Then server conf can be recovered from docker container labels/exposed ports. You can find that logick in DockerInstanceRuntimeInfo class

garagatyi (changes requested):

tolusha:
@garagatyi 
updated

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
```diff
@@ -376,6 +384,7 @@ public Instance startMachine(String workspaceId,
             try {
                 InstanceProvider provider = machineInstanceProviders.getProvider(machineConfig.getType());
                 machine.setId(generateMachineId());
+                addAgentsProvidedServers(machine, agents);
```
garagatyi, line 387: Looks like it is not called on start of the whole environment

garagatyi (changes requested):

tolusha:
@garagatyi 
updated

slemeur:
nice!

garagatyi (approved):
Good job!

Issue #1787, opened 2016-07-18
Remove workspace id from websocket connection path to master


vparfonov:
Workspace statuses are now tracked by various websocket connections per workspace id. It creates a lot of problems in order to handle the high number of websockets. (especially in dashboard)
Providing one connection with various channels will decrease the number of persistent websocket connections. Thus the clients will be able to use one connection to listen statuses of all workspaces, instead on multiple ones.

TylerJewell:
@vparfonov - is there a PR for this?  Why add to 4.6 milestone if PR is not readied.

PR #2907, opened 2016-10-27
CHE-1787: Remove workspace id from websocket connection path to master

vinokurig:
### What does this PR do?

Removes workspace id from websocket connection path to master.
#1787
### Previous behavior

Separate web-socket connection was created for each workspase.
### New behavior

All workspaces are using one web-socket connection.

@vparfonov @akurinnoy @ashumilova Please review


dashboard/src/components/api/che-websocket.factory.ts
```diff
@@ -54,17 +54,12 @@ export class CheWebsocket {
   }
 
 
-  getBus(workspaceId) {
-    var currentBus = this.sockets.get(workspaceId);
-    if (!currentBus) {
+  getBus() {
```
benoitf, line 58: this kind of change needs to be applied on getRemoteBus as well

benoitf (changes requested):

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/web.xml
```diff
@@ -23,7 +23,7 @@
     </context-param>
     <context-param>
         <param-name>org.eclipse.che.websocket.endpoint</param-name>
-        <param-value>/ws/{ws-id}</param-value>
+        <param-value>/ws</param-value>
```
benoitf, line 26: could we have http redirect from /ws/<id> to /ws ?
for backward compliance for existing clients ?

vinokurig: This is a network issue. This needs investigation.

benoitf: ok

benoitf (approved):
Except the fact that I would want to have a redirect from /ws/<id> to /ws for any existing clients, ok for me

akurinnoy (approved):

vparfonov (approved):

ashumilova (approved):

benoitf:
@vinokurig  is that https://github.com/eclipse/che/blob/47193dc85bcfd00f2dc03dc7f86c22f4cf6dd4e4/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceServiceLinksInjector.java#L133-L135 should have been updated ?

vinokurig:
@benoitf Thank you, I will fix that.

benoitf:
@vinokurig I've found another location, I will provide PR to fix it

vinokurig:
@benoitf ok, thanks

PR #2914, opened 2016-10-27
CODENVY-932 Stopped IDE should provide clear state and options to the user

vitaliy-guliy:
Handle start and stop the workspace outside the IDE

https://github.com/codenvy/codenvy/issues/932


vitaliy-guliy:
https://www.youtube.com/watch?v=6os2U28MpjQ

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/workspace/WorkspaceEventsHandler.java
```diff
@@ -399,8 +419,13 @@ protected void onMessageReceived(WorkspaceStatusEvent statusEvent) {
 
                 case STOPPED:
                     loader.setSuccess(LoaderPresenter.Phase.STOPPING_WORKSPACE);
-                    unSubscribeHandlers();
-                    eventBus.fireEvent(new WorkspaceStoppedEvent(workspace));
+
+                    try {
```
vparfonov: why we need try ..catch block here?

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/workspace/perspectives/general/AbstractPerspective.java
```diff
@@ -46,6 +46,14 @@
  */
 public abstract class AbstractPerspective implements Presenter, Perspective, ActivePartChangedHandler {
 
+    private enum State {
```
vparfonov, line 57: How it related to STOP Workspace Event?

ashumilova (approved):

vparfonov:
@evidolob please check this PR around restoring Parts

plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/machine/MachineStatusHandler.java
```diff
@@ -119,6 +120,12 @@ private void handleMachineRunning(final String machineId, final WorkspaceRuntime
         if (machine == null) {
             return;
         }
-        eventBus.fireEvent(new MachineStateEvent(machine, RUNNING));
+
+        try {
+            eventBus.fireEvent(new MachineStateEvent(machine, RUNNING));
```
azatsarynnyy: Firing an event wrapped in try block...
Never saw it before in our codebase.
That's interesting. Can you explain, please, what's the motivation of this

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/macro/MacroRegistryImpl.java
```diff
@@ -39,9 +38,7 @@ public MacroRegistryImpl() {
     public void register(Set<Macro> macros) {
         for (Macro macro : macros) {
             final String name = macro.getName();
-            if (this.macros.containsKey(name)) {
-                Log.warn(MacroRegistryImpl.class, "Command macro '" + name + "' is already registered.");
```
azatsarynnyy, line 43: Please, don't remove it.
This may be **really** useful to find out why some **Macro** isn't registered in IDE without needing to launch SuperDevMode.

vitaliy-guliy: There is too much outputs in browser console when restarting the workspace.

azatsarynnyy: > too much outputs in browser console

Is it a real reason why you've removed that code? :smirk_cat: 
If it is then the main reason of that warning must be fixed but removing the logging is the bad way to fix it, of course.

Ok, let's think about that all together :family_man_boy_boy: because we're using such approach in several places and I think it's really important to discuss it.

Imagine the following situation: some developer works on his own extension and wants to register his implementation of Macro. But another Macro with the same id/name is already registered in IDE. How we should inform the plugin's developer (not IDE user!) that he should choose another unique id/name for registering a Macro? Or do you think we shouldn't at all?

@vitaliy-guliy @vparfonov @evidolob @dkuleshov I'm interested in your opinion?

IMO informing **plugin's developer** about such kind of situations in browser's DevConsole is the best way in our case. And I think it's one the main purpose of the browser's DevConsole - outputting really useful messages for the developer in order to help him to solve his problems in the shortest period of time.

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/workspace/WorkspaceEventsHandler.java
azatsarynnyy: So if it's really the general problem should we wrap every firing an exception to such try block in order to take care about all problems of all handlers?
I doubt it.
I think that all exceptions in handlers are the problems of those handlers. And code that fires some event shouldn't worry about those exceptions in handlers.

IMO think such wrapping would bring a bad practice to our code.
Anyway, I'm really interested in your opinion.

plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/machine/MachineStatusHandler.java
vparfonov: Handler should resolve problem with exceptions, but  not then you fire event

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/workspace/WorkspaceEventsHandler.java
vitaliy-guliy: Yes, I have thought about possible solutions and think having a special (extended) event bus would be great.

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/macro/MacroRegistryImpl.java
azatsarynnyy: @vitaliy-guliy it's really nice that you're worried about cleanliness of the browser's DevConsole and you can propose your suggestions for discussion with team in another PR

plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/machine/MachineStatusHandler.java
vitaliy-guliy: I don't know how much handlers handle the event and few developers handle exceptions in their code ( null pointer exception for example ). So here I have to guarantee the event will be fired and following code will be executed.

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/macro/MacroRegistryImpl.java
vitaliy-guliy: It's not a problem to revert the code.
And try imaging how much you will be happy when you develop your code and can not find your output because console has the output which informs you about nothing but disturbs.
Please think about other developers.

plugins/plugin-machine/che-plugin-machine-ext-client/src/main/java/org/eclipse/che/ide/extension/machine/client/machine/MachineStatusHandler.java
azatsarynnyy: Yeah. You're right. Of course you don't know how much handlers are listening for the event.
Because when you're firing an event you shouldn't worry about amount of handlers. This is responsibility of EventBus

vitaliy-guliy: Yes. To have a special event bus which guarantee each registered handler will be invoked would be a supergreat!

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/macro/MacroRegistryImpl.java
azatsarynnyy: Exactly! Strongly agree with your opinion.
That's why I urge to log only really helpful messages to the DevConsole.
Messages that appear only when some problem is occurred but not every time.
Messages that may really help developer to solve his problem.
And logging in this registry exactly that case.

vitaliy-guliy: There are huge amount of messages in the console when restarting the workspace.
I had full console of the following
  Command macro ... is already registered.
  Command macro ... is already registered.
  Command macro ... is already registered.
  Command macro ... is already registered.

azatsarynnyy: Ok. That's means that some problem has appeared recently.
So, thanks to this message, now we know that some code tries to register several macros with the same ID or the same macro registered several times. And we must react somehow to this problem (report an issue, fix in another PR).
But removing the log message doesn't solve the actual problem. It only hides it.

vitaliy-guliy: OK. Added a commit reverting the code.

vparfonov (approved):

Issue #2308, opened 2016-09-02
Git history window is too small


bmicklea:
When using the Git History in the UI the actual history window at the bottom is much too small, it makes it hard to use.
1. Open a project with a long git history
2. Select a step in the history
3. Try to read the git history in a very small window :(

**Expected behavior:**
It would be nice to be able to resize the git history viewing window so that on a large screen I can see much more of the history.

**Che version:**    4.7
**OS and version:**  Mac OSX 10.11.6 and beta.codenvy.com 4.7
**Docker version:** 1.12.1
**Che install:**    Docker container

RomanNikitenko:
The history window with default size looks like:
![default_size](https://cloud.githubusercontent.com/assets/5676062/19598744/1dc7b5cc-97a5-11e6-9569-d3443465c638.png)
We can resize this one like this, for example:
![resize](https://cloud.githubusercontent.com/assets/5676062/19598751/2f69df1c-97a5-11e6-90a0-4b152300e07a.png)
@bmicklea 
Could you clarify what new behavior do you prefer?
Do we need change default size or we need find a more appropriate place for this window?

bmicklea:
Yes, please make the default size 3x wider.

bmicklea:
@RomanNikitenko is this still in code review?  If so can we get a review and merge please?

bmicklea:
@vparfonov this is part of 5.0.0 isn't it?

RomanNikitenko:
We have some merged changes related to setting size of part stacks. So I need to check and update my pull request.

RomanNikitenko:
AFAIK we have the new visual Git history window, so the issue can be closed as fixed by https://github.com/eclipse/che/pull/3799
@bmicklea WDYT?

bmicklea:
Fixed by #3799

PR #2931, opened 2016-10-28
CHE-2308. Use own size of part when part stack size is too small for the content of the part

RomanNikitenko:
### What does this PR do?

We use default size for all part stacks now.
After applying my changes we will have next behavior:
 - for example you open Git History first, it uses 550 width. You open Pull Request panel while Git History was still displayed, it uses 550 width too.
 - you open Git History first, it uses 550 width. You close Git History and then open Pull Request panel while nothing was displayed on the right part, it uses default size (260) width.
- you open Pull Request first, it uses 260 width. You open Git History while Pull Request was still displaying on the right, it uses 550 width.
- you open Pull Request first, it uses 260 width. You close Pull Request and then open Git History panel while nothing was displayed on the right part, it uses 550 width.

This behavior will be apply for all part stacks. 
We use default size (260) for all parts except Git history part (it will be 550). 
#2308

history_pullreq_menu

### DIFF
--- a/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/PartStackPresenter.java
+++ b/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/PartStackPresenter.java
@@ -41,6 +41,9 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 
+import static org.eclipse.che.ide.api.parts.PartPresenter.DEFAULT_PART_SIZE;
+import static org.eclipse.che.ide.api.parts.PartPresenter.MIN_PART_SIZE;
+
 /**
  * Implements "Tab-like" UI Component, that accepts PartPresenters as child elements.
  * <p/>
@@ -53,13 +56,6 @@ import java.util.Map;
  * @author Valeriy Svydenko
  */
 public class PartStackPresenter implements Presenter, PartStackView.ActionDelegate, PartButton.ActionDelegate, PartStack {
-
-    /** The default size for the part. */
-    private static final double DEFAULT_PART_SIZE = 260;
-
-    /** The minimum allowable size for the part. */
-    private static final int MIN_PART_SIZE = 100;
-
     private final WorkBenchPartController         workBenchPartController;
     private final PartsComparator                 partsComparator;
     private final Map<PartPresenter, Constraints> constraints;
@@ -113,8 +109,8 @@ public class PartStackPresenter implements Presenter, PartStackView.ActionDelega
         };
 
         if (workBenchPartController != null) {
-            this.workBenchPartController.setSize(DEFAULT_PART_SIZE);
             this.workBenchPartController.setMinSize(MIN_PART_SIZE);
+            this.workBenchPartController.setSize(DEFAULT_PART_SIZE);
         }
 
         currentSize = DEFAULT_PART_SIZE;
@@ -210,15 +206,8 @@ public class PartStackPresenter implements Presenter, PartStackView.ActionDelega
 
         if (state == State.MINIMIZED) {
             state = State.NORMAL;
-
-            if (currentSize < MIN_PART_SIZE) {
-                currentSize = DEFAULT_PART_SIZE;
-            }
-
-            workBenchPartController.setSize(currentSize);
-            workBenchPartController.setMinSize(MIN_PART_SIZE);
+            updateWorkBenchPartSize();
             workBenchPartController.setHidden(false);
-
         } else if (state == State.COLLAPSED) {
             // Collapsed state means the other part stack is maximized.
             // Ask the delegate to restore part stacks.
@@ -227,11 +216,7 @@ public class PartStackPresenter implements Presenter, PartStackView.ActionDelega
             }
 
         } else if (state == State.NORMAL) {
-            if (workBenchPartController.getSize() < MIN_PART_SIZE) {
-                workBenchPartController.setMinSize(MIN_PART_SIZE);
-                workBenchPartController.setSize(DEFAULT_PART_SIZE);
-            }
-
+            updateWorkBenchPartSize();
             workBenchPartController.setHidden(false);
         }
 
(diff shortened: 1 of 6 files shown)


slemeur:
Looks good for Git History.
Does that mean that it changed the default sizes for all panels displayed on the right? 

I'm asking this because, obviously the default width for the _Pull Request_ Panel and the _Git History_ should not be the same - it should depend on the feature displayed.

RomanNikitenko:
@slemeur 
You are right, these changes apply for tooling part stack, so for Pull Request Panel  as well as for the Git History.
As far as I remember we had issue some times ago to avoid jumps when we are switching between parts of the same part stack: between Events and Processes area for example. So I have applied this behavior for Pull request panel and  Git History panel.
But I'm interested in your opinion - WDYT?

slemeur:
I think that changing the sizes of the panel, once it has been opened would be ugly.

Imagine the following use case:
1- Open _Pull Request_ panel --> small width for right part
2- Open _Git History_ panel --> large width for right part
3- Open _Pull Request_ panel --> small width for right part
It will show a lot of redrawing which would not really be friendly to the user.

There are probably a tradeoff possible, by increasing the width of the right part - only if the user try to display a panel which require it.

We would have the following use cases:
1- Open _Pull Request_ panel --> small width for right part
2- Open _Git History_ panel --> large width for right part
3- Open _Pull Request_ panel --> width for right part stay large
And until the user has not opened _git history_ we keep displaying it small.

This way we avoid constant redrawing and situation where the right part is too small for the content of the panel.

Does it make sense to you? 

Maybe - for the time being it is overkill, but we can fill an improvement for that behavior.

RomanNikitenko:
@slemeur 
So do you prefer different default size for Git History panel and  for Pull request panel?

Before my changes we had default size for both panel: 260
After my changes default size for both panel is: 750

Do you prefer follow default size:
- for Git History panel - 750
- for Pull request panel - 260

slemeur:
Yeah, but I'd like the width being updated only if more space is needed.

So for example:
- you open _Git History_ first, it uses 750 width. You open _Pull Request_ panel while _Git History_ was still displayed, it uses 750 width too.
- you open _Git History_ first, it uses 750 width. You close _Git History_ and then open _Pull Request_ panel while nothing was displayed on the right part, it uses 260 width.
- you open _Pull Request_ first, it uses 260 width. You open _Git History_ while _Pull Request_ was still displaying on the right, it uses 750 width.
- you open _Pull Request_ first, it uses 260 width. You close _Pull Request_ and then open _Git History_ panel while nothing was displayed on the right part, it uses 750 width.

750 by default for **all** panel is too much.

RomanNikitenko:
thank you, Stevan!
Now it more clear for me.
But I still have questions.
Current behavior also includes the following: we remember size when user resizes a part stack. For example:  
- user opens Processes panel with default size = 260
- user resizes it to 550
- user closes Processes panel

At this moment we remember 550 as size for bottom part stack.
So we will use 550 size when user opens Processes panel or Events panel as well.
I think we need to override this behavior for right part stack, otherwise corresponding the new behavior imagine the following use case:
- user opens Git History first, it uses default size. 
- user resizes it to a half of screen size.
- user opens Pull Request panel while Git History was still displayed - we remember this size as size for right part stack
- Pull Request panel is displayed with a half of screen size.

What size we need to remember for right part stack?
Do we need to remember own size for each part (not for part stack)?
So do we need to have different way of handling left, right and bottom part stacks?

slemeur:
Thanks Roman. 

We need to find a solution which will be the same for each part stacks, not only for right part stack - otherwise it will be confusing for the user. 

I think that it does not change the main goal that we want to achieve, we want to avoid situations where the user will see the size changing as he is switching between panels. 

Let's keep the rule defined with the previous comments and override it once the user manually resized a part (all: left, right, bottom). If a user manually resized a part, we use it as default for all panels of the part.

TylerJewell:
@vparfonov @RomanNikitenko - lets plan to update these files and get them merged in the next sprint. If you have 30 minutes this week, it would be nice to go ahead and do that now.  Cleaning up old PRs.

RomanNikitenko:
@slemeur @vparfonov I have updated pull request - please review.

slemeur:
@RomanNikitenko: Have you applied the latest comment I gave on this PR?

btw - you can use reviewers for the PR, it helps anybody to see who must validate the PR before merging it.
screen shot 2017-01-19 at 15 06 07

RomanNikitenko:
@slemeur 
- I have updated video which displays new behavior
- thank you for your advice about reviewers - I will use it!

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/PartStackPresenter.java
```diff
@@ -496,6 +476,15 @@ public void onTabClicked(@NotNull TabItem selectedTab) {
...
         selectActiveTab(activeTab);
+
+        final double workbenchPartSize = workBenchPartController.getSize();
+        if (state != State.MAXIMIZED && isSizeOverridden(workbenchPartSize)) {
+            currentSize = workbenchPartSize;
+        }
+
+        if (activePart != null) {
+            updateWorkBenchPartSize();
+        }
```
vitaliy-guliy, line 487: Could you please add comments on new blocks you are added?
It will help developers to understand the logic of the code.

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/PartStackPresenter.java
```diff
@@ -510,10 +499,56 @@ public void showPartMenu(int mouseX, int mouseY) {
...
+
+        if (workBenchPartSize < partSize) {
+            return partSize; //use own size of part when part need more size than current size of workbench part
+        } else {
+            return workBenchPartSize; //use current size of workbench part when this one has enough place for the part
+        }
+    }
+
+    private boolean isSizeOverridden(double size) {
```
vitaliy-guliy, line 541: Could you please add a little comments on methods?

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/PartStackPresenter.java
RomanNikitenko: Done!

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/PartStackPresenter.java
RomanNikitenko: Done!

vparfonov (approved):

RomanNikitenko:
Fixed by #3799

PR #2940, opened 2016-10-31
Add handling of starting workspace errors

vinokurig:
### What does this PR do?
Handle exceptions when making `workspaceServiceClient.startById()` request.

### Previous behavior
If ,for example, user's RAM limit is exceeded or started workspaces number limit is exceeded, no error will be showed in IDE when starting workspace.

@vparfonov please review


vitaliy-guliy:
LGTM

vparfonov (approved):

PR #2946, opened 2016-10-31
CHE-2883 Git importer - convertToTopLevelProject feature

### LINKED ISSUE #2883: Samples configuration improvements
Usually if some vendor wants to provide set of samples they put them in single repository
like 
- https://github.com/artik-iot/artik-sdk-examples
- https://github.com/platformio/platformio-examples 

Our samples configuration provide ability to configure to clone individual modules with `keepDir` parameter. But usability of this is pretty low.

 First of all we see all subset of directories of upper level. Second  - only root become a project. What is quite unusable. 
I'm proposing to add one more parameter `convertToTopLevelProject` that is `false` by default and activated only if  `keepDir` is not empty and `keepVcs` is false. If it's configured to `true` project will be moved to upper level and converted to project. We need to have `keepVcs=false` because we need to do some adjustments in source tree and after that it's not possible to do any operation with git and I think it's ok, it will be used for demo purposes in most cases.

skabashnyuk:
### What does this PR do?
Fixes #2883
Allow to configure git importer to convert subdirectory
cloned during sparce checkout to top level project.

Parameter name - convertToTopLevelProject.
This parameter working if keepDir is not empty and keepVcs=false

### DIFF
--- a/wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitProjectImporter.java
+++ b/wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitProjectImporter.java
@@ -10,6 +10,7 @@
  *******************************************************************************/
 package org.eclipse.che.api.git;
 
+import com.google.common.base.Strings;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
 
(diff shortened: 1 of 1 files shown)


vparfonov (approved):

benoitf:
if this feature is only implemented by git provider, should property name be prefixed by git ?
Also it looks strange that we need to forgot git metadata. I think that if the goal is to see a folder as a top level project in the IDE, then it should be fixed on UI / model, not as part of the importer.

As when working on a local computer, I first clone project locally, then I can choose which folder to import as root project (and not forgetting git metadata) so it should be the same there. For me it looks like a hack/workaround.

skabashnyuk:
> if this feature is only implemented by git provider, should property name be prefixed by git ?

I think no. Potentially it can be extend to other VCS providers.

> Also it looks strange that at we need to forgot git metadata. I think that if the goal is to see a >folder as a top level project in the IDE, then it should be fixed on UI / model, not as part of the >importer.

Right now our UI does not provide functionality that you describe.

Issue #2927, opened 2016-10-28
Detect and stop/remove Machines not counted by API (regression)


skryzhny:
We have a job run by timer that clears containers which are not counted by API - these still consume resources. After introducing ability to launch multimachine workspaces in 5.0.0-M5 we changed the naming scheme of containers and these job stops functioning.

Need to fix it and make some tests. Probably we need new property with regexp for detecting containers launched by che/codenvy, so that regexp can be changed without need to modify code. 

**Reproduction Steps:**
1. Start che/codenvy and let it run for a while(several hours/days), launching different kind of workspaces
2. Stop che/codenvy
3. Run a `docker ps -a`, you will probably see some containers that were launched by che/codenvy and not stopped/removed properly.

There are more probability of seeing such problem when runing multiple MachineNodes.

**Expected behavior:**
All containers belonged to che/codenvy that are not counted by API stopped/removed.

**Observed behavior:**
Containers belonged to che/codenvy that are not counted by API are still running or not removed

**Che version:**    5.0.0-M5  
**OS and version:** Centos 7.2
**Docker version:** 1.11.2
**Che install:**   Codenvy

**Additional information:**
- Problem started happening recently, didn't happen in an older version of Che: Yes
- Problem can be reliably reproduced, doesn't happen randomly: Yes

AndrienkoAleksandr:
I checked jobs DockerContainerCleaner on Che and Codenvy. It works normal. I can't reproduce this bug. I will add logs for containers which were checked by DockerContainerCleaner, but job didn't clean up them, because they are active for our API.

PR #2950, opened 2016-10-31
CHE-2927: Add logs for active containers checked by DockerContainerCleaner.

AndrienkoAleksandr:
### What does this PR do?
Add logs for active containers checked by DockerContainerCleaner.

https://github.com/eclipse/che/issues/2927

### Previous behavior
We logs clean up inactive containers. 

@garagatyi @mmorhun review please.

Signed-off-by: Aleksandr Andrienko <aandrienko@codenvy.com>

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleaner.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleaner.java
@@ -23,6 +23,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 
@@ -31,7 +33,7 @@ import static org.eclipse.che.plugin.docker.client.params.RemoveContainerParams.
 import static org.eclipse.che.plugin.docker.machine.DockerContainerNameGenerator.ContainerNameInfo;
 
 /**
- * Job for periodically clean up inactive docker containers
+ * Job for periodically clean up inactive docker containers and log list active containers.
  *
  * @author Alexander Andrienko
  */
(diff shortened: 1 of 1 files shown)


AndrienkoAleksandr:
ci-build

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleaner.java
```diff
@@ -80,6 +85,19 @@ public void run() {
...
         }
+        logActiveContainers(activeContainers);
+    }
+
+    private void logActiveContainers(List<String> activeContainers) {
+        if (activeContainers.isEmpty()) {
+            LOG.info("No active containers were detected.");
+        } else {
+            StringBuilder activeContainersReport = new StringBuilder();
+            activeContainersReport.append(activeContainers.size() > 1 ? "Containers are registered" : "Container is registered");
```
garagatyi: please simplify it. It is just for maintenance purposes so we don't have to sync message with amount of containers

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleaner.java
```diff
@@ -80,6 +85,19 @@ public void run() {
...
+            activeContainersReport.append(" in the api:");
+            activeContainers.forEach(containerName -> activeContainersReport.append("\n\t").append(containerName));
```
garagatyi: Why not use list to string ?

garagatyi (changes requested):

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerContainerCleaner.java
AndrienkoAleksandr: Ok

garagatyi (approved):
Look good!

mmorhun (approved):

PR #2953, opened 2016-11-01
Rework Tck tests to use TckListener instead of TckModuleFactory

sleshchenko:
### What does this PR do?
Rework Tck tests to use TckListener instead of TckModuleFactory

-

### Previous behavior
Earlier there was not ability to close entity manager factory on finish of tck test. So it could broke tests when there were one or more tests (non tck) that used EntityManagerFactory and some tck test that did not close entity manager factory.
Also there were issues with child injectors which guice uses.

### New behavior
There is ability to close entity manager after tck testing. And TckListener creates new injector for each tck test that solves problem with child injector.


sleshchenko:
@skabashnyuk @evoevodin @akorneta Please review my changes

akorneta (approved):

voievodin (approved):

Issue #2820, opened 2016-10-19
"No workspace selected. Unable to open IDE" displayed while opening the IDE


slemeur:
There is a odd message displayed while loading the IDE as an iframe from the dashboard.

![loading-bug-no-workspace-loaded](https://cloud.githubusercontent.com/assets/1636769/19511662/6f286f92-95e8-11e6-8190-dcaa257688ef.png)

See attached animated gif:
![loading-bug-che](https://cloud.githubusercontent.com/assets/1636769/19511746/db5bbd90-95e8-11e6-9883-57f60fe6e3bd.gif)

See issue: https://github.com/codenvy/codenvy/issues/985

We need to display this message, only when it is accurate and no workspace has been selected. We should not display it while opening the IDE.
We need to remove the placeholder message from:
https://github.com/eclipse/che/blob/master/dashboard/src/app/ide/ide.html

**Che version:**    5.0.0-M5

**Additional information:**
- Problem started happening recently, didn't happen in an older version of Che: No
- Problem can be reliably reproduced, doesn't happen randomly: No

akurinnoy:
time spent: 2d

PR #2957, opened 2016-11-01
CHE-2820: Don't show "No workpsace selected" message when IDE is loading

akurinnoy:
### What does this PR do?
It prevents to show message "No workspace selected. Unable to open IDE" while IDE is loading.

#2820 

### DIFF
--- a/dashboard/src/app/ide/ide.controller.ts
+++ b/dashboard/src/app/ide/ide.controller.ts
@@ -9,41 +9,57 @@
  *   Codenvy, S.A. - initial API and implementation
  */
 'use strict';
+import IdeSvc from './ide.service';
+import IdeIFrameSvc from './ide-iframe/ide-iframe.service';
+import {RouteHistory} from '../../components/routing/route-history.service';
+import {CheWorkspace} from '../../components/api/che-workspace.factory';
 
 /**
  * This class is handling the controller for the IDE
  * @author Florent Benoit
  */
 class IdeCtrl {
+  $rootScope: ng.IRootScopeService;
+  $timeout: ng.ITimeoutService;
+  ideSvc: IdeSvc;
+  ideIFrameSvc: IdeIFrameSvc;
+  cheWorkspace: CheWorkspace;
+
+  hasData: boolean;
+  workspaces: any[];
+  selectedWorkspace: any = null;
+  selectedWorkspaceExists: boolean = true;
+  selectedWorkspaceName: string = null;
 
   /**
    * Default constructor that is using resource
    * @ngInject for Dependency injection
    */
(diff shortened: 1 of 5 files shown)


olexii4 (approved):

ashumilova (approved):

dashboard/src/app/ide/ide.controller.ts
```diff
@@ -9,39 +9,57 @@
...
-    $rootScope.wantTokeepLoader = true;
+    (this.$rootScope as any).showIDE = false;
+    (this.$rootScope as any).wantTokeepLoader = true;
+
+    this.selectedWorkspaceExists = true;
 
     // search the selected workspace
-    let namespace = $routeParams.namespace;
-    let workspace = $routeParams.workspaceName;
+    let namespace = ($routeParams as any).namespace;
```
benoitf: I think that $routeParams['namespace'] wouldn't require a cast to any

dashboard/src/app/ide/ide.controller.ts
```diff
@@ -9,39 +9,57 @@
...
     this.ideSvc = ideSvc;
     this.ideIFrameSvc = ideIFrameSvc;
     this.$rootScope = $rootScope;
     this.cheWorkspace = cheWorkspace;
     this.$timeout = $timeout;
-    this.selectedWorkspace = null;
-    this.$rootScope.showIDE = false;
 
-    $rootScope.wantTokeepLoader = true;
+    (this.$rootScope as any).showIDE = false;
```
benoitf: $rootScope['showIDE'] = false ?

dashboard/src/app/ide/ide.controller.ts
akurinnoy: we have 'no-string-literal' rule enabled which disallows object access via string literals.
we can disable this rule or use another cast syntax:
(<any>$routeParams).namespace

benoitf: ok it's because I have the feeling that ($routeParams as any) adds complexity when reading code. and because typescript definition says that we have an array not an object.

akurinnoy: ok, I got it
could I disable those rule in tslint config?

benoitf: yes this or your proposal of defining specific interface, it's great
we then avoid runtime errors, by having compilation errors

ashumilova (approved):

benoitf (approved):

dashboard/src/components/typings/che.d.ts
```diff
@@ -0,0 +1,23 @@
+import che = _che;
```
benoitf, line 11: I think copyright header is missing there

akurinnoy: fixed

Issue #1661, opened 2016-07-05
Unable to remove SSH key when including invalid characters


Blacktiger:
In che, a user can create a VCS SSH key that cannot be deleted.

**Reproduction Steps:**
1. Open Preferences > SSH > VCS
2. Click Generate Key or Upload Key
3. Enter a hostname including the protocol (https://github.com or similar)
4. Try to delete the key

**Expected behavior:**

There are two possibilities here:
- The host name of the key is not accepted if it contains bad characters (colon, slash, etc)
- The key is properly deleted

**Observed behavior:**

The key is not deleted.

**Che version:**    All versions up to 4.4

**Additional information:**
- Problem started happening recently, didn't happen in an older version of Che: No
- Problem can be reliably reproduced, doesn't happen randomly: Yes

This looks to me like a simple mismatch between user expectations and system requirements. The field for the hostname is called "Host (w/o port)". For me at least, because it mentions that the port should be excluded - that causes me to think of urls with the protocol on the front even though that is not what the field is for. Perhaps the field should be called Hostname instead of host to be less confusing. Regardless, the hostname should probably be rejected if it was entered with invalid characters since the rest api doesn't allow using those invalid characters to delete the key later.

ddementieva:
It's a known bug. Thanks for reporting it.

ddementieva:
https://jira.codenvycorp.com/browse/CHE-816

raafaar:
I'm having the same issue. While it is solved, **Is there a way to manually delete the key?** For example, using the CLI. I'm asking because I need to sync my project and it seems to fail because of the existing wrong hostname.

ddementieva:
@raafaar This is API problem, not purely the client one.

raafaar:
@ddementieva the problem for me is that although I deleted and recreated the workspace, (using the same recipe) the key is still there. I need to find a way to delete it. And that is because I can't connect to the git service and may be is for the wrong key.

ghost:
@Blacktiger if you go to `/storage` you will find a handful of jsons, `preferences.json` being one of them. You may try deleting this key directly there.

bmicklea:
@skabashnyuk can you take this in the next sprint please?

michealharrington:
I'm using the hosted eclipse che by red hat and I have run into this issue. I delete all my workspaces and can longer create or run any workspace. It keeps complaining about my bad ssh key. I'm not sure how to get rid of the key.

bmicklea:
[...]

PR #2960, opened 2016-11-02
CHE-1661: Fix deletion of ssh key by name with slash

vinokurig:
### What does this PR do?
Changes `name` parameter in delete ssh key pair rest method from `PathParam` to `QueryParam` to be able to perform delete ssh pair request with name that contains slashes. 

https://github.com/eclipse/che/issues/1661

### Previous behavior
If user generates an ssh key pair with name that contains slash (https://github.com) and then he removes it, the delete request will fail

@sleshchenko @skabashnyuk please review


sleshchenko:
Why did you move remove pair methods? Also you should rework getPair https://github.com/eclipse/che/pull/2960/files#diff-a9bc2a25767f466231ee9e153fc07e5fR174 methods.

vinokurig:
@sleshchenko I've created a new PR(https://github.com/eclipse/che/pull/3000) where I used regexp instead query parameters because adding query parameter to `getPair` amd `removePair` methods leads to conflict.

PR #2964, opened 2016-11-02
Fix links for websockets

benoitf:
### What does this PR do?

I've discovered using 'che dir' that links were not updated when workspace id has been removed for websocket connections

With this fix workspace id is no longer in links
https://github.com/eclipse/che/issues/1787

### Previous behavior
links are not correct

### New behavior
Fix links that are no longer valid since https://github.com/eclipse/che/pull/2907

Change-Id: I4103f7f2a316361ddc9dfd26c7b47e2096f99855


benoitf:
@vinokurig

TylerJewell:
Lets merge this - we need to get past these stability issues asap.

vinokurig (approved):

PR #2974, opened 2016-11-03
Always respond with an error if trying to start a running workspace

voievodin:
The workspace status must be changed to `STARTING` synchronously during start call, while machines start itself should be asynchronous. This is important for resources/limits checks, as they rely on the workspace statuses. Right now it is possible to send 2 parallel requests to start the workspace and get unlucky timing, which won't start 2 workspaces but both request will return _200ok_, the PR chages improve such behaviour and so services will respond 200ok only, and only if an attempt to start the workspace will be really performed.

@garagatyi please review


garagatyi (approved):
LGTM

garagatyi:
@sleshchenko Please take a look

sleshchenko (approved):

PR #3002, opened 2016-11-07
Make it possible to use relative links to recipe in workspace environment

### LINKED ISSUE #2769: Use relative link to recipe in workspace environment if possible
If recipe is stored in our storage we can convert link to relative. This will allow to not domain in our storage and change domain in future.

mshaposhnik:
### What does this PR do?
To ease data migration between domains, we should store local recipe links in a relative form. And when necessary, downloader should place current domain by itself.

Fixes #2769


skabashnyuk (approved):

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/util/RecipeDownloader.java
```diff
@@ -65,9 +65,12 @@ public RecipeImpl getRecipe(MachineConfig machineConfig) throws MachineException
...
             // add user token to be able to download user's private recipe
-            final String apiEndPointHost = apiEndpoint.getHost();
-            final String host = targetUriBuilder.build().getHost();
-            if (apiEndPointHost.equals(host)) {
+            final URI recipeUri = targetUriBuilder.build();
+            if (!recipeUri.isAbsolute() && recipeUri.getHost() == null) {
+                targetUriBuilder.scheme(apiEndpoint.getScheme())
+                                .host(apiEndpoint.getHost())
+                                .port(apiEndpoint.getPort())
+                                .replacePath(apiEndpoint.getPath() + location);
```
garagatyi, line 73: Is it possible that api endpoint doesn't have trailing slash? In that case I suppose it would be better to use .replacePath().path()

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/util/RecipeDownloader.java
```diff
@@ -105,9 +108,12 @@ public String getRecipe(String location) throws ServerException {
...
```
garagatyi, line 116: See above

voievodin (approved):

wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/environment/server/CheEnvironmentEngineTest.java
```diff
@@ -298,7 +296,7 @@ public void shouldSetDockerfileContentInsteadOfUrlIfUrlPointsToCheApiOnEnvironme
         // given
         EnvironmentImpl env = createEnv();
         String machineName = "machineWithDockerfileFromApi";
-        String additionalServiceComposeFilePart = "\n  " + machineName + ":\n    build:\n      context: " + API_ENDPOINT + "/recipe/12345";
```
garagatyi, line 301: I assume that your changes should not influence these tests

garagatyi (changes requested):

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/util/RecipeDownloader.java
mshaposhnik: AFAIK api endpoint never contain trailing slash

wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/environment/server/CheEnvironmentEngineTest.java
mshaposhnik: I's due to regexp change in CheEnvironmentEngine.

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/util/RecipeDownloader.java
garagatyi: ok

garagatyi (approved):

PR #3006, opened 2016-11-07
Improve default services start strategy.

AndrienkoAleksandr:
### What does this PR do?
Improve services start strategy. Add detail message when user create compose environment with circle dependency to itself.

@garagatyi please review.
Signed-off-by: Aleksandr Andrienko <aandrienko@codenvy.com>

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/DefaultServicesStartStrategy.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/DefaultServicesStartStrategy.java
@@ -17,7 +17,7 @@ import org.eclipse.che.api.environment.server.model.CheServiceImpl;
 import org.eclipse.che.api.environment.server.model.CheServicesEnvironmentImpl;
 
 import java.util.HashMap;
-import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -29,7 +29,8 @@ import static java.lang.String.format;
 /**
  * Finds order of Che services to start that respects dependencies between services.
  *
- * author Alexander Garagatyi
+ * @author Alexander Garagatyi
+ * @author Alexander Andrienko
  */
 public class DefaultServicesStartStrategy {
     /**
@@ -55,7 +56,6 @@ public class DefaultServicesStartStrategy {
             throws IllegalArgumentException {
 
         HashMap<String, Integer> weights = new HashMap<>();
-        Set<String> machinesLeft = new HashSet<>(services.keySet());
 
         // create machines dependency graph
         Map<String, Set<String>> dependencies = new HashMap<>(services.size());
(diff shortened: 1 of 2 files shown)


AndrienkoAleksandr:
ci-build

AndrienkoAleksandr:
ci-build

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/DefaultServicesStartStrategy.java
```diff
@@ -67,32 +67,20 @@
                                                                               service.getVolumesFrom().size());
 
             for (String dependsOn : service.getDependsOn()) {
-                if (!services.containsKey(dependsOn)) {
-                    throw new IllegalArgumentException(
-                            format("Dependency '%s' in machine '%s' points to not known machine.",
-                                   dependsOn, serviceEntry.getKey()));
-                }
+                checkDependency(dependsOn, serviceEntry.getKey(), services, "A service can not depend on itself");
```
garagatyi: We don't have 'service' entity in our public API. User would not understand what that mean if environment is not docker compose.
And message doesn't point to a machine with the error.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/DefaultServicesStartStrategy.java
```diff
@@ -67,32 +67,20 @@
...
 
             // links also counts as dependencies
             for (String link : service.getLinks()) {
                 String dependency = getServiceFromLink(link);
-                if (!services.containsKey(dependency)) {
-                    throw new IllegalArgumentException(
-                            format("Dependency '%s' in machine '%s' points to not known machine.",
-                                   dependency, serviceEntry.getKey()));
-                }
+                checkDependency(dependency, serviceEntry.getKey(), services, "A service can not link to itself");
```
garagatyi: We don't have 'service' entity in our public API. User would not understand what that mean if environment is not docker compose.
And message doesn't point to a machine with the error.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/DefaultServicesStartStrategy.java
```diff
@@ -67,32 +67,20 @@
...
             }
             // volumesFrom also counts as dependencies
             for (String volumesFrom : service.getVolumesFrom()) {
                 String dependency = getServiceFromVolumesFrom(volumesFrom);
-                if (!services.containsKey(dependency)) {
-                    throw new IllegalArgumentException(
-                            format("Dependency '%s' in machine '%s' points to not known machine.",
-                                   dependency, serviceEntry.getKey()));
-                }
+                checkDependency(dependency, serviceEntry.getKey(), services, "A service can not contain 'volumes_from' to itself");
```
garagatyi: We don't have 'service' entity in our public API. User would not understand what that mean if environment is not docker compose.
And message doesn't point to a machine with the error.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/DefaultServicesStartStrategy.java
```diff
@@ -101,48 +89,34 @@
...
-                            Optional<String> maxWeight = dependencies.get(service)
-                                                                     .stream()
-                                                                     .max((o1, o2) -> weights.get(o1).compareTo(weights.get(o2)));
-                            // optional can't be empty because size of the list is checked above
-                            //noinspection OptionalGetWithoutIsPresent
-                            weights.put(service, weights.get(maxWeight.get()) + 1);
-                            machinesLeft.remove(service);
-                            weightEvaluatedInCycleRun = true;
-                        }
+                Map.Entry<String, Set<String>> serviceEntry = it.next();
```
garagatyi, line 96: I think that extracting key and value here would make code more readable

garagatyi (changes requested):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/DefaultServicesStartStrategy.java
AndrienkoAleksandr: Done.

garagatyi (approved):

skabashnyuk (approved):

AndrienkoAleksandr:
ci-build

PR #3013, opened 2016-11-08
[WIP] Hackathon merge

garagatyi:
### What does this PR do?
 - Refactors workspace bootstrapping code:
    - Addition of projects volumes to ws-machines and applying of agents moved to infrastructure provisioner. It can be overriden in any implementation of Che and will be responsible for providing all infrastructure needed for Che implementatipn functioning. 
    - Allow to configure environment network driver.
    - Allow to add Agents from the code instead of special files. These agents can override agents configured in files if identifiers are equal.
    - Adds possibility to host agents binaries from CHE master
    - Reworks agents in a way when they can be downloaded from CHE master if they are not found on FS.
    - Adds workspace ID to internal representation of environment to allow to use it for internal purposes of machine bootstrapping.
- Renames some properties.
- Refactors AgentConfigApplier to incapsulate additional code that was used with this class before.


plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalCheInfrastructureProvisioner.java
```diff
@@ -0,0 +1,80 @@
...
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import java.io.IOException;
+import java.util.Map;
+
+import static java.lang.String.format;
+
+/**
+ * @author Alexander Garagatyi
```
benoitf, line 32: There is nothing about what this class does

wsmaster/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/model/impl/AgentImpl.java
```diff
@@ -44,10 +44,10 @@ public AgentImpl(String id,
         this.name = name;
         this.version = version;
         this.description = description;
-        this.dependencies = dependencies;
-        this.properties = properties;
+        this.dependencies = dependencies != null ? dependencies : new ArrayList<>();
```
benoitf: getters are doing this check so it looks odd to have it at both places
if it's immutable it should be Collection.emptyList()

garagatyi:
@skabashnyuk @evoevodin @mshaposhnik @sleshchenko Please review

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/os/WindowsPathEscaper.java
```diff
@@ -0,0 +1,61 @@
...
+     * @param path path on Windows. Can be unix-style path also.
+     * @return unix-style path
+     */
+    public String escapePath(String path) {
+        String esc;
+        if (path.indexOf(":") == 1) {
+            // check and replace only occurrence of ":" after disk label on Windows host (e.g. C:/)
+            // but keep other occurrences it can be marker for docker mount volumes
+            // (e.g. /path/dir/from/host:/name/of/dir/in/container                                               )
+            esc = path.replaceFirst(":", "").replace('\\', '/');
```
mshaposhnik, line 51: Doesn't like File.separator :) ?

assembly/assembly-main/src/assembly/tomcat/conf/server.xml
```diff
@@ -139,6 +139,10 @@
       <Host name="localhost"  appBase="webapps"
             unpackWARs="true" autoDeploy="false">
 
+
+        <!-- Provide Che Agent binaries (like workspace agent, terminal) -->
+        <Context docBase="${che.home}/lib" path="/agent-binaries" />
```
skabashnyuk, line 144: I propose to use separate folder for that. Let's call it `/agent-binaries`

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/os/WindowsPathEscaper.java
```diff
@@ -0,0 +1,61 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.commons.lang.os;
+
+/**
+ * Escapes Windows path to unix-style path.
```
skabashnyuk, line 14: what for?

wsmaster/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/model/impl/AgentImpl.java
skabashnyuk: firstNotNull?

assembly/assembly-main/src/assembly/tomcat/conf/server.xml
benoitf: separate folder by duplicating the existing binaries ? because agents are inside CHE_HOME/lib for now

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/os/WindowsPathEscaper.java
garagatyi: It's not my code. And I don't want to introduce here additional changes in this PR because it will delay merge significantly.

assembly/assembly-main/src/assembly/tomcat/conf/server.xml
garagatyi: I suppose that we can move agents binaries to any folder, no?

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/os/WindowsPathEscaper.java
garagatyi: It is used in several places in around docker usage in both Che and Codenvy. I think it is time to move that code in a single place since it is bad practice to use the same private methods in several classes.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalCheInfrastructureProvisioner.java
garagatyi: Yes, I'll fix that. It is because this PR is still in WIP

wsmaster/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/model/impl/AgentImpl.java
garagatyi: Getters have a bug, I'll rewrite them. The bug is that they return mutable collection that doesn't influence their state.

assembly/assembly-main/src/assembly/tomcat/conf/server.xml
benoitf: I'm ok but it might take some time to validate this change no ?

garagatyi: @skabashnyuk are you ok if we postpone that and do after merging that PR?

skabashnyuk: yes

skabashnyuk (approved):

Issue #2876, opened 2016-10-25
File watchers don't always work as expected


skabashnyuk:
[Short description of problem here]

**Reproduction Steps:**
1. Start workspace
2. Create directory `aaa`  from terminal in `/projects` folder

**Expected behavior:**
To see `aaa` project with `blank` project type.

**Observed behavior:**
Nothing happened.

**Che version:**    master  
**OS and version:** OSX   

![dir_bug](https://cloud.githubusercontent.com/assets/1614429/19679934/71a4bba4-9aac-11e6-91f1-342468d833c5.gif)

musienko-maxim:
On nightly5.codenvy-stg.com works as expected

PR #3035, opened 2016-11-09
Correct processing reload depth when there is not any resource exists

vzhukovs:
Correct processing reload depth when container is trying to synchronize, when there is no any resource in the workspace root. Depth in this case should be 1 by default.

Related issue: https://github.com/eclipse/che/issues/2876

@vparfonov @skabashnyuk review it, please.


vparfonov (approved):

riuvshin:
please merge this now, we can't wait due to release

PR #3036, opened 2016-11-09
CODENVY-804 : Use of userManager to get the userId (when injecting ssh keys)

benoitf:
https://github.com/codenvy/codenvy/issues/804

### Previous behavior
No ssh keys are injected (machine or workspace keys)

### New behavior
SSH keys are injected as expected

In Eclipse Che, Owner of a machine or namespace of a workspace is the same than the user-id

ssh service is using user-id while username is given so there is never a match

It fixes it by asking userManager the userId.

Change-Id: I30a20a52ffc3149a8f732edeb8ab86db827c9d91

### DIFF
--- a/plugins/plugin-machine/che-plugin-machine-ext-server/pom.xml
+++ b/plugins/plugin-machine/che-plugin-machine-ext-server/pom.xml
@@ -54,6 +54,10 @@
         </dependency>
         <dependency>
             <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-user</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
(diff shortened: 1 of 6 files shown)


benoitf:
@sleshchenko kind of old bug, if you can review my PR

TylerJewell:
Yes - approved.

sleshchenko (approved):
OK as temporary solution. Because we should fix the main problem of this bug here https://github.com/eclipse/che/blob/master/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java#L313-313
We should set user id for machine creator.

benoitf:
linked to https://github.com/eclipse/che/issues/3042

PR #3051, opened 2016-11-10
Update Bitnami stacks to latest versions

amrecio:
### What does this PR do?

Updates the Bitnami Stacks to the latest version. The containers now use Bitnami's own "minideb" image as a base, which reduces the size of the containers in comparison with the previous version.

It doesn't reference an issue.

### New behavior
Bitnami stacks updated to latest version.

### DIFF
--- a/ide/che-core-ide-stacks/src/main/resources/stacks.json
+++ b/ide/che-core-ide-stacks/src/main/resources/stacks.json
@@ -1446,39 +1446,36 @@
     "id": "bitnami-express",
     "creator": "Bitnami",
     "name": "Bitnami Express",
-    "description": "Default Express stack with Node.js 6.4.0, Express 4.14.0, MongoDB 3.2.7.",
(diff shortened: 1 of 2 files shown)


benoitf:
ci-build

benoitf (approved):

slemeur:
Thanks @amrecio .

We make the default stacks provided by Che allocating 2GB of RAM, this ensure that the workspace's agents that Che is injecting will get enough ressources and the user will get a workspace with decent ressources to work with. 
User still have the ability to change the RAM allocated but by default workspaces are created with 2GB of RAM.

Could you please update your PR?

amrecio:
Of course, no problem. I've update the PR as requested.

Thanks for the clarification.

amrecio:
Would it be possible to update the PR to bump the revision of the Laravel container image? The current one works but it contains an outdated version of MariaDB.

slemeur:
Yes sure @amrecio

amrecio:
Done.

slemeur (approved):

benoitf:
ci-build

PR #3053, opened 2016-11-11
Enable to find the container deployed by Kubernetes.

monaka:
### What does this PR do?

On vanilla Docker, the container-ID and the hostname in the container will be same.
Kubernetes (a.k.a k8s) names container-ID in Docker as specified naming rules.
So the hostname and the container-ID are not same.
This patch enables get container-ID deployed by K8s.

None.

### Previous behavior

Just returns the result of `hostname`

### New behavior

In case KUBERNETES_SERVICE_HOST has the value, find container-id that have hostname and `POD`.
KUBERNETES_SERVICE_HOST is always defined by the K8s runtime.

### DIFF
--- a/assembly/assembly-main/src/assembly/bin/docker.sh
+++ b/assembly/assembly-main/src/assembly/bin/docker.sh
@@ -157,7 +157,12 @@ get_che_server_container_id() {
   # Returning `hostname` doesn't work when running Che on OpenShift/Kubernetes.
   # In these cases `hostname` correspond to the pod ID that is different from
   # the container ID
-  hostname
+
+  if [[ -z "${KUBERNETES_SERVICE_HOST+x}" ]]; then
+    hostname
+  else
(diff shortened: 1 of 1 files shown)


tolusha:
@garagatyi 
Could you have a look please.

garagatyi:
I think it is better to be reviewed by @TylerJewell

TylerJewell:
@benoitf - we can take a look at this change after we do the consolidation of the Che cli and codenvy cli.

One thing I noticed is that the check for null variable should use the built in bash syntax of +x.

monaka:
It's better to use `if [[ -z "${KUBERNETES_SERVICE_HOST+x}"  ]]` ?

monaka:
We don't need this patch since the master branch was improved.
Now I'm checking eclipse/che-server:5.0.0-M8 (with no hack) with my K8s cluster.
Thanks for your improvement. I close it.

Issue #2768, opened 2016-10-12
Implement schema initialization/migration components


voievodin:
The issue #2110 defines the strategy for schema initialization/migration, th strategy 
must be implemented.
Consider using [Flyway](https://flywaydb.org/) as a tool for resolving/performing sql migrations, it provides a simple way to customize the process of searching migration scripts and version them in a single way.

Along the implementation the following issues should be solved:
- migration schema location/packaging
- migration tests
- migration scripts in TCKs
- TCK for different datasources(postresql/h2)
- Existing overheads for generating foreign kes must be removed from the code
- Entity columns must be explicitly named

Issue #2110, opened 2016-08-12
Study Initial schema generation and how to make changes in model after migration to single database


skabashnyuk:
Time to time we are doing some small changes in our model. Sometimes we just adding new fields
But sometimes https://github.com/eclipse/che/issues/1820 we change completely structure of some object.  
We need to figure out how :
1. Change structure of database from previus version or much earlier version
2. What to do with existing data during migration.

voievodin:
After a discussion and some study and investigation process it was decided to define an explicit database schema based on a current version generated by EclipseLink. This approach allows us to create native migration scripts, avoid useless mappings for those entries which need foreign keys, use all the power of abilities of SQL standard.
Consider the following scripts hierarchy:

<pre>
   resources/
     /sql
        /5.0.0
          1.init.sql
        /5.0.0-M1
          1.rename_fields.sql
          2.add_workspace_constraint.sql
          /postgresql
            2.add_workspace_constraint.sql
        /5.0.1
          1.stacks_migration.sql
</pre>

From the example
- There is a directory called `sql` which contains initialization/migration scripts
- There are version directories under the `sql` like `5.0.0` or `5.0.0-M1` these directories contain 
  scripts for specific versions
- There are sql scripts under project version duiretories line `1.init.sql` or `1.rename_fields.sql` the naming of scripts is pretty simple: the first is a script version in a project versions it's `1` in `1.init.sql`
  then description of changes(if necessary) `init` in `1.init.sql` and then the file extension `.sql` in `1.init.sql`
- There is a directory under `5.0.0-M1` called `postresql` if current database provider is posgresql then 
  the script from `5.0.0-M1/posgresql/2.add_workspace_constraint.sql` will be used instead of `5.0.0-M1/2.add_workspace_constraint.sql`, so basically if the same script name is provided in provider-specific directory then this script will be used instead 

in this way we can easily create initialization scripts, extend an existing schema and manage overrides for different database. It also allows us to develop one version migrations in separate task simultaneously by versioning scripts under current project version.

PR #3063, opened 2016-11-14
Rework database schema loading

voievodin:
Adds new approach for initializing database schemas and performing migrations.
[Flyway](https://flywaydb.org/) with some custom modifications is used as database migration engine.
The issues are #2768 and #2110. 

### Configuration properties

```properties
# if database contains some data(schema is already initialized), 
# flyway will initialize its state without any migrations up to baseline version
db.schema.flyway.baseline.enabled=true
# up to this version <= migration scripts won't be applied on the existing database
db.schema.flyway.baseline.version=5.0.0.8.1
# migration scripts prefix(empty by default) e.g. it's "v" in "v1.init.sql" script name
db.schema.flyway.scripts.prefix=
# migration scripts suffixes(.sql)
db.schema.flyway.scripts.suffix=.sql
# the real script version is between prefix and first separator appearance
# e.g script version is "1" for the script with name "1.init.sql" according to given config
db.schema.flyway.scripts.version_separator=__
# where to search for scripts, it may be either classpath or filesystem
# if location starts with "classpath:" or with nothing then migrations resolver will try 
# to find scripts in classpath, otherwise if scripts starts with "filesystem:" then 
# it will lookup filesystem for scripts, the scripts must be structure in a way described below.
# It may contain several locations separated with ","
db.schema.flyway.scripts.locations=classpath:che-schema
# resource name to lookup data source
db.jndi.datasource.name=java:/comp/env/jdbc/che
```
### Scripts directory structure :file_folder:
Let's say we have 2 locations in classpath 

```
che-schema/
    5.0.0-M8/
        1__init.sql
    5.0.0-M9
        1__rename_tables.sql
        2__insert_predefined_data.sql
        postgresql/
            2__insert_predefined_data.sql
codenvy-schema/
    5.0.0-M8/
        1.1__init_codenvy.sql
    5.0.0-M9
        2.1__create_new_tables.sql
```
With the simplest configuration:
- prefix - empty
- suffix - `.sql`
- version separator - `__`
- locations - `classpath:che-schema,classpath:codenvy-schema`

If the current database driver is `posgresql` then the following migrations will be resolved and applied:

| db version        | script name                              | location                     |  picked vendor specific |
| -------------------- |:----------------------------------------:| --------------------------:|--------------------------------:|
|5.0.0.8.1            |  1__init.sql                                |  che-schema           | no                                   |
| 5.0.0.8.1.1        | 1.1__init_codenvy.sql               |  codenvy-schema   | no                                   |
| 5.0.0.9.1           | 1__rename_tables.sql              |  che-schema           | no                                   |
| 5.0.0.9.2           | 2__insert_predefined_data.sql |  che-schema           | yes                                  |
| 5.0.0.9.2.1        | 2.1__create_new_tables.sql     |  codenvy-schema   | no                                    |

@sleshchenko, @akorneta, @skabashnyuk please review the pull request


akorneta (approved):

core/che-core-db/src/main/java/org/eclipse/che/core/db/schema/impl/flyway/CustomSqlMigrationResolver.java
```diff
@@ -0,0 +1,201 @@
...
+        final ListMultimap<String, SqlScript> scripts = finder.findScripts(flywayConfiguration);
+        LOG.debug("Found scripts: {}", scripts);
+
+        // filter sql scripts according to current db provider
+        final ListMultimap<String, SqlScript> versionToScripts = ArrayListMultimap.create();
+        for (String name : scripts.keySet()) {
+            final List<SqlScript> candidates = scripts.get(name);
+            final Map<String, SqlScript> vendorToScript = new HashMap<>();
+            for (SqlScript candidate : candidates) {
+                final String vendorName = candidate.vendor == null ? DEFAULT_VENDOR_NAME : candidate.vendor;
```
skabashnyuk: firstNotNull?

core/che-core-db/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,36 @@
...
+    </appender>
+
+    <appender name="file" class="ch.qos.logback.core.FileAppender">
+        <File>target/log/codenvy.log</File>
+        <encoder>
+            <pattern>%-41(%date[%.15thread]) %-45([%-5level] [%.30logger{30} %L]) - %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <!--logger name="com.codenvy.factory" level="DEBUG"/-->
```
sleshchenko: let's remove this commented line and change log level from `INFO` to `ERROR`

pom.xml
```diff
@@ -848,6 +860,16 @@
                 <version>${che.version}</version>
             </dependency>
             <dependency>
+                <groupId>org.flywaydb</groupId>
```
sleshchenko: I think you should move these dependencies to [che-dependencies](https://github.com/eclipse/che-dependencies/blob/master/pom.xml) project

voievodin: I will, thank you for the reminder

wsmaster/che-core-api-ssh/src/main/java/org/eclipse/che/api/ssh/server/model/impl/SshPairImpl.java
```diff
@@ -43,20 +44,24 @@
         }
 )
 @IdClass(SshPairPrimaryKey.class)
+@Table(name = "sshkeypair")
 public class SshPairImpl implements SshPair {
     @Id
+    @Column(name = "owner")
```
sleshchenko, line 50: Looks like we should remove `user` field

sleshchenko (approved):

skabashnyuk (approved):

PR #3075, opened 2016-11-15
INFRA-14: Limits number of pids

tolusha:
### What does this PR do?
Limits the number of processes that can be forked inside a cgroup. Set -1 for unlimited.
Works on nodes where kernel version is equal or greater than 4.3

https://github.com/codenvy/infrastructure/issues/14

### Previous behavior
Docker sets limits by itself. By default it is 512.

### New behavior
Admin might set its own limits.

Some docs 
https://blog.docker.com/2016/02/docker-engine-1-10-security/
https://docs.docker.com/engine/reference/commandline/run/


tolusha:
Pls review
@garagatyi 
@mmorhun

garagatyi (approved):

PR #3084, opened 2016-11-16
CODENVY-657: Do not log SourceNotFoundException

vinokurig:
### What does this PR do?
Do not log `SourceNotFoundException` when starting workspace.

https://github.com/codenvy/codenvy/issues/657

Related to https://github.com/codenvy/codenvy/pull/1181

@garagatyi please review


tolusha (approved):

skabashnyuk:
Why?

vinokurig:
@skabashnyuk If user will create a worcspace from config with wrong recipe location and then will start it, ERROR will appear in logs. This situation shouldn't be logged because it is normal situation.

skabashnyuk (approved):

garagatyi (approved):

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -805,13 +805,17 @@ private String buildImage(final DockerConnection dockerConnection,
                         progressMonitor.updateProgress(progressStatus);
                     }
 
-                    throw new DockerException("Docker image build failed. Image id not found in build output.", 500);
+                    throw new ImageNotFoundException("Docker image build failed. Image id not found in build output.");
```
garagatyi: This exception doesn't fit this case

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -805,13 +805,17 @@ private String buildImage(final DockerConnection dockerConnection,
...
                 });
 
                 return imageIdFuture.get();
             } catch (ExecutionException e) {
                 // unwrap exception thrown by task with .getCause()
-                throw new DockerException(e.getCause().getLocalizedMessage(), 500);
+                if (e.getCause() instanceof ImageNotFoundException) {
+                    throw new ImageNotFoundException(e.getCause().getLocalizedMessage());
```
garagatyi, line 821: why not **throw e.getCause()** in any case?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -647,6 +649,11 @@ void performAsyncStart(WorkspaceImpl workspace,
                                   rmEx.getLocalizedMessage());
                     }
                 }
+                for (Throwable cause : getCausalChain(ex)) {
+                    if (cause instanceof SourceNotFoundException) {
```
garagatyi, line 653: I'm not sure that in that case this exception can be thrown

garagatyi (changes requested):

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
vinokurig: fixed by throwing exception according to error message from docker

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
vinokurig: `e.getCause()` returns `Throwable`, buildImage() method can not throw such exception

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
vinokurig: This exception will be thrown in both cases.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -798,6 +798,12 @@ private String buildImage(final DockerConnection dockerConnection,
                 Future<String> imageIdFuture = executor.submit(() -> {
                     ProgressStatus progressStatus;
                     while ((progressStatus = progressReader.next()) != null) {
+                        if (progressStatus.getError() != null) {
+                            String errorMessage = progressStatus.getError();
+                            if (errorMessage.matches("Error: image .+ not found")) {
```
garagatyi, line 803: And if error doesn't match this pattern what will happen?

garagatyi (changes requested):

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
vinokurig: It will be another error that will be thrown in https://github.com/eclipse/che/pull/3084/files/8dd478fad8c533fd0a6a71e3bd37ebf2434ef3d3#diff-fca1a783572d79a80aa755d17f907b91R814

garagatyi (approved):

PR #3103, opened 2016-11-17
Dto streaming

tareksha:
### What does this PR do?

Eliminates the redundant creation of full in-memory JSON tress and then strings when reading DTOs from requests or writing them to responses. Applies to general consumption of `DtoFactory` as well, not JAX-RS services only.

* Performance.
Current method always constructs a full replica of the DTO objects as an in-memory `JsonElement` tree, and then asks Gson to dump the tree as a full string. There are two issues here:
  * Gson can generate the JSON in a stream-based method based on `JsonWriter` which eliminates the need for any intermediate trees. Actually no JSON elements are created at all here.
  * The whole method of generating/parsing full JSON texts is not needed, the JAX-RS `CheJsonProvider` passes stream `Reader` and response `Writer` that can used directly by Gson, eliminating even the need of an intermediate full JSON texts.
* Maintainability.
A large part of the generated DTO implementation code is there to replicate the DTO data in `JsonElement` trees, bit by bit. The code section in the generator that takes care of generating this java code in the tool is the most complicated area in the already hard to maintain generator. Gson can do this automatically and more efficiently (previous point).

### Previous behavior
* Full JSON trees built when converting DTOs to JSONs.
* Full JSON text is created before parsing DTOs from JAX-RS requests and (vise versa for responses).

### New behavior
* JSON trees eliminated by using Gson to generate the DTOs objects directly.
* Full JSON texts eliminated by using request `Reader`s and response `Writer`s directly in `DtoFactory`.
* Smaller files generated for DTO implementations: workspace API reduced by 30%, machine API by 29%

#### Changelog
Removes redundant in-memory JSON trees for faster DTO parsing performance.

Signed-off-by: Tareq Sharafy <tareq.sha@gmail.com>

### DIFF
--- a/core/che-core-api-dto/src/main/java/org/eclipse/che/dto/generator/DtoImpl.java
+++ b/core/che-core-api-dto/src/main/java/org/eclipse/che/dto/generator/DtoImpl.java
@@ -32,6 +32,7 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 /** Abstract base class for the source generating template for a single DTO. */
 abstract class DtoImpl {
@@ -174,6 +175,16 @@ abstract class DtoImpl {
             addSuperGetters(dto, getters);
         }
         return new ArrayList<>(getters.values());
+    }
+
+    protected Set<String> getSuperGetterNames(Class<?> dto) {
+        final Map<String, Method> getters = new HashMap<>();
+        Class<?> superDto = getSuperDtoInterface(dto);
+        if (superDto != null) {
+            addDtoGetters(superDto, getters);
+            addSuperGetters(superDto, getters);
+        }
+        return getters.keySet();
     }
 
     /**
--- a/core/che-core-api-dto/src/main/java/org/eclipse/che/dto/generator/DtoImpl.java
+++ b/core/che-core-api-dto/src/main/java/org/eclipse/che/dto/generator/DtoImpl.java
@@ -177,6 +177,9 @@ abstract class DtoImpl {
         return new ArrayList<>(getters.values());
     }
 
+    /**
+     * Get the names of all the getters in the super DTO interface and upper ancestors.
+     */
     protected Set<String> getSuperGetterNames(Class<?> dto) {
         final Map<String, Method> getters = new HashMap<>();
         Class<?> superDto = getSuperDtoInterface(dto);
--- a/core/che-core-api-dto/src/test/java/org/eclipse/che/dto/definitions/DTOHierarchy.java
+++ b/core/che-core-api-dto/src/test/java/org/eclipse/che/dto/definitions/DTOHierarchy.java
@@ -46,5 +46,18 @@ public final class DTOHierarchy {
         void setParentField(String parentField);
 
         ChildDto withParentField(String parentField);
+
+        ChildDto getShadowedField();
+
+        void setShadowedField(ChildDto v);
     }
+
+    @DTO
+    public interface GrandchildDto extends ChildDto {
+
+        GrandchildDto getShadowedField();
+
+        void setShadowedField(GrandchildDto v);
+    }
+
 }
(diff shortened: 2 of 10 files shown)


skabashnyuk:
Hello @tareqhs  thanks for this PR. Do you have any statistic that confirms performance optimization?

tareqhs:
Hi @skabashnyuk working on this right now

tareksha:
Hi @skabashnyuk ,
Here are my benchmark results:

**Setup**
Prepare a random `WorkspaceDto`, composed of:
- 10 `ProjectConfigDto`, each composed of
  - 10-character name
  - 10-character type
  - 10-character path
  - 10 `Link` entries, each with a short name and 118-character command line
- 10 `CommandDto`, each composed of
  - 10-character name
  - 10-character type
  - 5 simple attributes
- 3 `EnvironmentDto`, each composed of
  - 3 `ExtendedMachineDto`, each with 5 attributes

All times below are in milliseconds.

**Benchmark 1**
Prepare a reusable `Writer` that has zero writing overhead (similar to /dev/null).
_START TIMER_
Call `JsonSerializable.toJson(Writer)` on the test DTO 20,000 times.
_STOP TIMER_
Because `JsonSerializable.toJson(Writer)` was introduced in the fix, the non-fixed benchmark actually does `writer.write(dtoObj.toJson())`

Without fix :  6601, 5569, 5931, 5479, 5486, 5838, 5523, 5556, 5521, 5495
With fix : 2945, 2714, 2692, 2890, 2704, 2738, 2700, 2705, 2724, 2756

Omitting the bootstrap pass, average benchmark time is **reduced by 52%**

**Benchmark 2**
Convert the full DTO to a JSON. Resulting text is made of 42,320 characters.
Open 20,000 UTF-8 streams that read the same content, all in-memory.
_START TIMER_
Call `DtoFactory.createDtoFromJson(InputStream,Class)` on all the open streams, one after another.
_STOP TIMER_

Without fix :  6758, 6460, 6480, 6492, 6602, 6455, 6431, 6478, 6400, 6617
With fix : 2533, 2361, 2337, 2301, 2347, 2294, 2300, 2367, 2451, 2474

Omitting the bootstrap pass, average benchmark time is **reduced by 63%**

tareksha:
hi @skabashnyuk , any update on this ?

skabashnyuk:
ci-build

tareksha:
@skabashnyuk  seems the ci passes now

core/che-core-api-dto/src/main/java/org/eclipse/che/dto/generator/DtoImpl.java
```diff
@@ -176,6 +177,16 @@ protected String getJsonFieldName(Method getterMethod) {
         return new ArrayList<>(getters.values());
     }
 
+    protected Set<String> getSuperGetterNames(Class<?> dto) {
```
skabashnyuk, line 183: javadoc

core/che-core-api-dto/src/test/java/org/eclipse/che/dto/definitions/DTOHierarchy.java
```diff
@@ -46,5 +46,18 @@
         void setParentField(String parentField);
 
         ChildDto withParentField(String parentField);
+
```
skabashnyuk, line 49: What  is the purpose of changes in DTOHierarchy class?

tareksha: shadowed fields cause caused initial failure, fixed this and added them to make the behavior is consistent
@skabashnyuk

skabashnyuk: Can you elaborate a bit more. explanation is not clear for me.

tareksha: Gson fails if a class has a shadowed field (a field with the same name as another field in a base class). To avoid this shadowed fields are omitted from generated classes and this test verifies the behavior. @skabashnyuk

tareksha:
hi @skabashnyuk , any update on this ?

skabashnyuk:
No updates. I'm looking for some free time-frame to make some integration and performance testing internally. Recently we made some performance testing of ws-agent and one of the slowest part was DTO parsing https://github.com/codenvy/codenvy/issues/1466. Interesting how your PR can affect on this.

skabashnyuk:
Take a look on https://cloud.githubusercontent.com/assets/1614429/21721914/fd47cc40-d432-11e6-827b-4e8c287fc75d.png

bmicklea:
Added Changlog entry.

PR #3125, opened 2016-11-18
Codenvy 1127 - Avoid of concurrent snapshots on same node.

mshaposhnik:
### What does this PR do?
Allows to override DockerInstance class

CODENY-1127


mkuznyetsov (approved):
ok

PR #3032, opened 2016-11-09
Exec agent Che IDE client 

dkulieshov:
### What does this PR do?
- `ExecAgentCommandManager` - new exec agent Che IDE client
- Improvements/fixes/changes to `JsonRPC` protocol
- Improvements/fixes/changes to `websocket` protocol

### What issues does this PR fix or reference?
This is an adaptation of a new exec agent implementation located in [exec-agent](https://github.com/eclipse/che/tree/exec-agent) branch, the work is done under #1947 issue.

#### Websocket protocol
1. Websocket implementation on IDE side now support custom endpoints. This is done by calling improved initializer:
`org.eclipse.che.ide.websocket.ng.impl.WebSocketInitializer#initialize(String endpointId, String url)`
where 
  - `endpointId` is a high level websocket connection identifier (e.g. "exec-agent", "ws-agent", etc.)
  - `url` is a URL of websocket endpoint the connection is established to.
After a connection is initialized and established an endpoint identifier can be used to choose where to send a message via a transmitter, an instance of `org.eclipse.che.ide.websocket.ng.WebSocketMessageTransmitter#transmit(String endpointId, String message)`
2. Multi-protocol support is considered redundant and removed, now websocket supports only json rpc protocol transfers, though core routines for addition of support of new protocols are still present.
3. Websocket messages are now plain text messages thus parsing and analyzing is moved to a higher level consumers (e.g. json rpc service dispatchers).

#### JsonRPC protocol
1. Added automation on different levels, requests and responses are now parsed automatically and the business objects are now passed to actual receivers and transmitters.
2. As client-side websocket implementation supports multiple endpoints json rpc receivers for client side now requires a new way of binding in a dependency injection container. This is now done like this: 
```
GinMapBinder.newMapBinder(binder(), String.class, RequestHandler.class)
                    .addBinding("exec-agent" + '@' + "connected")
                    .to(ConnectedEventHandler.class);
```
where
- `exec-agent` is an endpoint identifier
- `connected` is a method name in terms of JsonRpc 2.0 protocol.
At the same time for server-side registration procedure nothing is changed.
3. Changed API of request and notification receivers. At the moment it is done by extending of `org.eclipse.che.ide.jsonrpc.RequestHandler` class for client-side and 
[...]

Issue #1947, opened 2016-07-26
Rework clients to use new command API of new exec agent


garagatyi:
In #1943 new command API with new machine agent will be introduced. We need adapt clients to use that API.

JamesDrummond:
@skabashnyuk Please add your team label.

dkulieshov:
As I see the new command API with the new machine agent is introduced in a pull request https://github.com/eclipse/che/pull/2163

PR #3143, opened 2016-11-22
Json RPC and websocket architecture reorganization

dkulieshov:
Replicates changes done under #1947 and contained within https://github.com/eclipse/che/pull/3032 including websocket and json rpc services modification but excluding exec agent client adaptation itself. The main idea here is to separate exec agent and jsonrpc/websocket changes into several independent PRs.


dkulieshov:
I have to admit that one of the main changes here is the reorganization of request handling subsystem, now we have `RequestHandler` as an abstract class while perviously it was an interface. And that fact rises some important questions here. 

Abstract class obviously imposes some restrictions however in our case besides that it also gives simplification in terms of method-handler guice/gin binding and allows us to automate quite a big amount of work that previously was done by json rpc user (like json rpc request parsing and params DTO generation, request result generation and json rpc response generation).

Unfortunately I don't really see the way we can keep the benefits and at the same time change an abstract class to an interface since we're unable to use reflection on client-side.

vparfonov (approved):

Issue #3119, opened 2016-11-18
Add Sourcegraph TypeScript Language Server to the Che 


vparfonov:
Need to add  Sourcegraph TypeScript Language Server  to the Che standard package 

https://github.com/sourcegraph/javascript-typescript-langserver

PR #3145, opened 2016-11-22
 add TypeScript language server

evidolob:
### What does this PR do?
add TypeScript language server
#3119


tolusha:
A lot of JavaDoc missed.

tolusha:
What about removing `js` prefix in names?
For instance: 
* `ls-typescript` instead of `ls-js-ts`
* `org.eclipse.che.ls.typescript.binaries` instead of `org.eclipse.che.ls.js-ts.binaries`

plugins/plugin-web/che-plugin-web-ext-shared/src/main/java/org/eclipse/che/plugin/web/shared/Constants.java
```diff
@@ -0,0 +1,31 @@
...
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.plugin.web.shared;
+
+/**
+ *
+ */
```
azatsarynnyy, line 15: empty javadoc

plugins/plugin-web/che-plugin-web-ext-shared/src/main/java/org/eclipse/che/plugin/web/shared/Constants.java
```diff
@@ -0,0 +1,31 @@
...
+    public static final String LANGUAGE  = "language";
+
+    /** TS Project Type ID */
+    public static String TS_PROJECT_TYPE_ID = "typescript";
+
+    /** TS Language */
+    public static String TS_LANG = "typescript";
+
+    /** Default extension for TS files */
+    public static String TS_EXT = "ts";
```
azatsarynnyy, line 27: It seems that this constant is never used so may be remove it?

evidolob: Now it's used

vparfonov (approved):

PR #3150, opened 2016-11-23
update che-select component

akurinnoy:
### What does this PR do?
This PR updates che-select component in order to be able to show validation messages.

It's needed for
https://github.com/codenvy/codenvy/pull/1210

### Previous behavior
che-select_1

### New behavior
che-select_2

Signed-off-by: Oleksii Kurinnyi <okurinnyi@codenvy.com>


ashumilova (approved):

dashboard/src/components/widget/select/che-select.directive.ts
```diff
@@ -16,34 +16,76 @@
...
   /**
    * Default constructor that is using resource
    * @ngInject for Dependency injection
    */
   constructor() {
-    this.restrict = 'E';
-
-    this.replace = true;
-    this.transclude = true;
-    this.templateUrl = 'components/widget/select/che-select.html';
```
benoitf, line 29: why do we need to use HTML template inside typescript instead of defining html code in html file ?

akurinnoy: This approach allows me to explicitly associate element name and ng-messages directive to properly show a validation message.
Additional pros is it will work faster as we avoid extra request.

benoitf (approved):

PR #3166, opened 2016-11-24
Correct handling parent element during creating new resource

vzhukovs:
Correct handling selected element when user is trying to create new file or folder.

@skabashnyuk review it, please.


vzhukovs:
@ashumilova @vparfonov review it also.

ashumilova (approved):

Issue #3183, opened 2016-11-28
Open create workspace failed.


mmjordan:
[Short description of problem here]
I run Che inside a Docker container. But, when I clicked the "**Add Workspace**" button, the create-workspace page was note displayed. Only displayed : 
1: {{workspaceDetailsController.isCreationFlow ? 'New Workspace' : workspaceDetailsController.workspaceName}}

2: {{workspaceDetailsController.invalidWorkspace}}
**Reproduction Steps:**

1. [First Step]
Run che inside a Docker container with the default command:
docker run --rm -t -v /var/run/docker.sock:/var/run/docker.sock eclipse/che:5.0.0-latest start
2. [Second Step]
Connect to che: http://192.168.99.104:8080
3. [Other Steps...]
Click"workspace->Add workspace".

**Expected behavior:**

[Describe expected behavior here]
create-workspace page displayed.
**Observed behavior:**

[Describe observed behavior here]
Only displayed : 
1: {{workspaceDetailsController.isCreationFlow ? 'New Workspace' : workspaceDetailsController.workspaceName}}

2: {{workspaceDetailsController.invalidWorkspace}}

**Che version:**    [eclipse/che:5.0.0-M7]   
**OS and version:** [Windows 7.1]   
**Docker version:** [1.12.3, build 6b644ec]   
**Che install:**    [Docker container]   

**Additional information:**
* Problem can be reliably reproduced, doesn't happen randomly: [Yes]

* Execption show in the browser console:
TypeError: this.usedNamesList.includes is not a function
    at e.a.value (http://192.168.99.104:8080/dashboard/scripts/app-d88f825ffd.js:2043:5367)
    at e.a.value (http://192.168.99.104:8080/dashboard/scripts/app-d88f825ffd.js:2043:2201)
    at new e (http://192.168.99.104:8080/dashboard/scripts/app-d88f825ffd.js:2043:1177)
    at i (http://192.168.99.104:8080/dashboard/scripts/vendor-ae4aa9ca8d.js:74:19553)
    at Object.o [as instantiate] (http://192.168.99.104:8080/dashboard/scripts/vendor-ae4aa9ca8d.js:74:19645)
    at http://192.168.99.104:8080/dashboard/scripts/vendor-ae4aa9ca8d.js:75:9704
    at link (http://192.168.99.104:8080/dashboard/scripts/vendor-ae4aa9ca8d.js:154:3904)
[...]

PR #3188, opened 2016-11-28
CHE-3183: avoid using includes function

ashumilova:
### What does this PR do?

Avoids using includes function due to that it's supported by major browser versions.

https://github.com/eclipse/che/issues/3183

Signed-off-by: Ann Shumilova <ashumilova@codenvy.com>


benoitf:
we might use babel-polyfill to fix that as well (code can still use includes) as we do in unit tests
https://github.com/eclipse/che/blob/5a97ea1e232518453585983760e8959003963440/dashboard/karma.conf.js#L22

benoitf (approved):

akurinnoy (approved):

olexii4 (approved):

Issue #2993, opened 2016-11-07
Cascade deletion of projects with the same storage fails


voievodin:
If remove user who has more than one workspace with projects that have the same source storage, 
removal transaction will fail. Check out the [commit](https://github.com/eclipse/che/commit/80022eebd4b45a3fc48cd9c973058726225c0dc3) in branch [removal_test](https://github.com/eclipse/che/tree/removal_test) which adds breaking use-case.

The problem is that eclipselink collects removal _dependecies_ information in hash map where key 
is removal object and value is a hash set of removal object dependencies. 
SourceStorage `equals` and `hashCode` don't include synthetic `id` field which is present in object itself and makes 2 source storage different, so in this case eclipselink sees 2 different source storages as one  object(because the other object fields are the same) and builds dependency graph like `source -> [project1, project2]`, but sometimes it happens that `project1` is already removed and no longer a part of this unit of work which causes the problem.

Trace from test:
```
org.eclipse.che.api.core.ServerException: Exception [EclipseLink-6004] (Eclipse Persistence Services - 2.6.2.v20151217-774c696): org.eclipse.persistence.exceptions.QueryException
Exception Description: The object [ProjectConfigImpl{name='workspace1-project2', path='/workspace1-project2', description='workspace1-project2-description', type='workspace1-project2type', mixins=[mixin1, mixin2], attributes={attribute1=[value1], attribute3=[value3], attribute2=[value2]}, source=SourceStorageImpl{type='source-type', location='source-location', parameters={param2=value, param1=value}}}], of class [class org.eclipse.che.api.workspace.server.model.impl.ProjectConfigImpl], with identity hashcode (System.identityHashCode()) [945,150,386], 
is not from this UnitOfWork object space, but the parent session's.  The object was never registered in this UnitOfWork, 
but read from the parent session and related to an object registered in the UnitOfWork.  Ensure that you are correctly
registering your objects.  If you are still having problems, you can use the UnitOfWork.validateObjectSpace() method to 
help debug where the error occurred.  For more information, see the manual or FAQ.
```

Putting `@Cachable(false)` on those objects which contain synthetic identifiers, the test will not throw that exception, but it still requires investigation.

PR #3192, opened 2016-11-28
Add synthetic id to hashCode & equals methods

voievodin:
When the entity like `Workspace` contains some internal object which are unique only in the entity context it may happen that cascade removal of such entities doesn't work properly.

#### The problem:
Such entities as `WorkspaceConfig`, `Command`, `Environment`, `ProjectConfig` are unique in the certain workspace context and can be identified with their name + dependency primary key, but name is an updatable attribute which means that it is bad candidate for primary key. That's why for such kind of objects we generate their own synthetic identifier like this:

```java
@Id
@GeneratedValue
private Long id;
```
This approach allows us to save those object along with owning objects.
As the identifier is a certain DB related attribute we don't include it to the equals and hashcode methods.
But when we remove several entities in a single transaction EclipseLink builds dependency graph to perform removal in the right order, so project config entities are removed before the workspace config(dependency owner) and the graph is stored in the current session hashmap, where key is entity and values are the set of depending on the key entities. 
So if object `A1` is removed in the same transaction with object `A2` and those objects contain the same values for all the fields instead of id then, dependency graph will be like this:
```
A1 -> [ depends_on_A1]
A2 -> [ depends_on_A1, depends_on_A2 ]
```
because of the same hashcode & equals behaviour for A1 and A2.

The same happens with project source storages and project configs, as blank project & project samples refer to the source storage with the same fields.
And deletion fails with an error like this:
```
org.eclipse.che.api.core.ServerException: Exception [EclipseLink-6004] (Eclipse Persistence Services - 2.6.2.v20151217-774c696): org.eclipse.persistence.exceptions.QueryException
Exception Description: The object [ProjectConfigImpl{name='workspace1-project2', path='/workspace1-project2', description='workspace1-project2-description', type='workspace1-project2type', mixins=[mixin1, mixin2], attributes={attribute1=[value1], attribute3=[value3], attribute2=[value2]}, source=SourceStorageImpl{type='source-type', location='source-location', parameters={param2=value, param1=value}}}], of class [class org.eclipse.che.api.workspace.server.model.impl.ProjectConfigImpl], with identity hashcode (System.identityHashCode()) [945,150,386], 
is not from this UnitOfWork object space, but the parent session's.  The object was never registered in this UnitOfWork, 
but read from the parent session and related to an object registered in the UnitOfWork.  Ensure that you are correctly
registering your objects.  If you are still having problems, you can use the UnitOfWork.validateObjectSpace() method to 
help debug where the error occurred.  For more information, see the manual or FAQ.
``` 

#### The solution:

The simplest solution will be to add synthetically generated identifier into the hashcode & equals methods which i basically did and return object without database specific information from data access layer, it may be easily achieved by usage of model based copy constructors.
So for the object `Workspace workspace1` which contains all the database specific information the object `Workspace workspace2 = new WorkspaceImpl(workspace1)` won't contain such information.

#2993 

### Previous behavior
Sometimes cascade deletion fails.

### New behavior
Cascade deletion works just fine.

@skabashnyuk, @akorneta, @sleshchenko please review the changes

### DIFF
--- a/wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/jpa/JpaFactoryDao.java
+++ b/wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/jpa/JpaFactoryDao.java
@@ -38,6 +38,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.StringJoiner;
+import java.util.stream.Collectors;
 
 import static java.lang.String.format;
 import static java.util.Collections.singletonList;
@@ -65,14 +66,14 @@ public class JpaFactoryDao implements FactoryDao {
         } catch (RuntimeException ex) {
             throw new ServerException(ex.getLocalizedMessage(), ex);
         }
-        return factory;
+        return new FactoryImpl(factory);
     }
 
     @Override
     public FactoryImpl update(FactoryImpl update) throws NotFoundException, ConflictException, ServerException {
         requireNonNull(update);
         try {
-            return doUpdate(update);
+            return new FactoryImpl(doUpdate(update));
         } catch (DuplicateKeyException ex) {
             throw new ConflictException(ex.getLocalizedMessage());
         } catch (RuntimeException ex) {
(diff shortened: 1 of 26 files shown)


skabashnyuk (approved):

akorneta (approved):

sleshchenko (approved):

voievodin:
ci-build

Issue #3044, opened 2016-11-10
Remove temporary workspaces on server shutdown


voievodin:
Temporary workspaces are removed only when workspace stop method is called directly, 
so if che server is stopped and temporary workspaces are running they won't be removed, unless you start and stop them again.

**Reproduction Steps:**

1. Start a temporary workspace
2. Stop che server in any possible way e.g. `che.sh stop`
3. Start che server

**Expected behavior:**
Temporary workspaces removed

**Observed behavior:**
Temporary workspace are listed with `STOPPED` status

**Additional information:**
* Problem can be reliably reproduced, doesn't happen randomly: YES

The simplest way to fix this issue is to remove temporary workspace on `WorkspaceStatus[STOPPED]` event.

mshaposhnik:
Fixed by https://github.com/eclipse/che/pull/3245

PR #3245, opened 2016-12-01
CHE-3044 Remove temporary workspaces on server shutdown

mshaposhnik:
### What does this PR do?
Adds listener which removes temporary workspaces on server shutdown and startup (to make sure they are deleted in case of emergency shutdown)

#3044 

### Previous behavior
Is stop method wasn't called explicitly on temp workspace, it may stay on the DB after server shutdown.

### New behavior
They will be removed on shutdown or startup


wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/jpa/JpaWorkspaceDao.java
```diff
@@ -159,6 +161,26 @@ public WorkspaceImpl get(String name, String namespace) throws NotFoundException
         }
     }
 
+    @Override
+    @Transactional
+    public List<WorkspaceImpl> getWorkspaces(boolean isTemporary, int skipCount, int maxItems) throws ServerException {
```
sleshchenko, line 166: I think you should use Page object here instead of List

mshaposhnik: IMO its a huge overhead for a method which results doesn't supposed to return via rest.

sleshchenko: OK :smile:

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/model/impl/WorkspaceImpl.java
```diff
@@ -60,7 +60,10 @@
                 @NamedQuery(name = "Workspace.getByName",
                             query = "SELECT w FROM Workspace w WHERE w.account.name = :namespace AND w.name = :name"),
                 @NamedQuery(name = "Workspace.getAll",
-                            query = "SELECT w FROM Workspace w")
+                            query = "SELECT w FROM Workspace w"),
+                @NamedQuery(name = "Workspace.getByTemporary",
+                            query = "SELECT w FROM Workspace w WHERE w.isTemporary = :temporary")
+
```
skabashnyuk, line 66: index on isTemporary field?

wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/workspace/server/TemporaryWorkspaceRemoverTest.java
```diff
@@ -0,0 +1,85 @@
...
+    static final int COUNT_OF_WORKSPACES = 250;
+
+    @Mock
+    private WorkspaceDao workspaceDao;
+
+    @InjectMocks
+    private TemporaryWorkspaceRemover remover;
+
+    @Test
+    public void shouldRemoveTemporaryWorkspaces() throws Exception {
```
garagatyi, line 50: Please add tests for paging, catching exception, preconditions, sql query,

wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/workspace/server/spi/tck/WorkspaceDaoTest.java
```diff
@@ -189,6 +189,27 @@ public void shouldRemoveWorkspace() throws Exception {
         workspaceDao.get(workspace.getId());
     }
 
+
+    @Test
+    public void shouldGetWorkspacesByNonTemporary() throws Exception {
```
garagatyi, line 194: Please add tests for paging, catching exception, preconditions.

garagatyi (dismissed):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/WorkspaceDao.java
```diff
@@ -142,4 +142,21 @@
      *         when any other error occurs during workspaces fetching
      */
     List<WorkspaceImpl> getWorkspaces(String userId) throws ServerException;
+
+
+    /**
+     * Gets workspaces by temporary attribute.
```
akorneta, line 148: please add the doc which describes exceptions in situation when
`maxItems` and `skipCount` are invalid.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/TemporaryWorkspaceRemover.java
```diff
@@ -0,0 +1,80 @@
...
+            removeTemporaryWs();
+        } catch (ServerException | ConflictException e) {
+            LOG.warn("Unable to cleanup temporary workspaces on startup: " + e.getLocalizedMessage(), e);
+        }
+    }
+
+    @PreDestroy
+    void shutdown() {
+        try {
+            removeTemporaryWs();
```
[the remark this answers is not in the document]
mshaposhnik: At least 2 quick ideas why they have to be removed on shutdown:
 - they will not be present in DB dumps etc
 - startup will be faster

voievodin:
On workspace stop, if [runtimes.stop](https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java#L700) fails with an error, temporary workspace won't be removed. Can you also please deal with that?

wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/workspace/server/TemporaryWorkspaceRemoverTest.java
```diff
@@ -0,0 +1,85 @@
...
+                return ((int)argument) > COUNT_OF_WORKSPACES;
+            }
+        }), anyInt());
+
+        remover.removeTemporaryWs();
+
+        verify(workspaceDao, times(COUNT_OF_WORKSPACES)).remove(anyString());
+    }
+
+    private List<WorkspaceImpl> createEntities(int number) throws TckRepositoryException {
```
sleshchenko: When does this method throw TckRepositoryException?

mshaposhnik: Copy-paste driven development

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/TemporaryWorkspaceRemover.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    @VisibleForTesting
+    void removeTemporaryWs() throws ServerException, ConflictException {
+        final int count = 100;
+        int skip = 0;
+        List<WorkspaceImpl> workspaces = workspaceDao.getWorkspaces(true, skip, count);
+        while (!workspaces.isEmpty()) {
+            for (WorkspaceImpl workspace : workspaces) {
+                workspaceDao.remove(workspace.getId());
```
sleshchenko: I'd add try catch here to continue removing other ones when a error occurs during removing one workspace

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/WorkspaceDao.java
mshaposhnik: Added

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/model/impl/WorkspaceImpl.java
mshaposhnik: [Added](https://github.com/eclipse/che/pull/3245/files#diff-1a465722d7fa9129eb2396bb04889c08)

akorneta (approved):

voievodin (approved):

PR #3251, opened 2016-12-01
Update che.env to reflect latest properties

TylerJewell:
### What does this PR do?
1. Adds in missing, but commonly used properties into the new che.env template
2. Fixes some mis-named properties so that they are consiste with the new naming scheme.
3. Adds documentation for end users to understand how to configure certain properties


TylerJewell:
@benoitf @riuvshin 

@skabashnyuk - when we do the testing for running Che or Codenvy behind a proxy, we should do that testing with the Che CLI and the Codenvy CLI, as I think the properties are slightly different for each.  I've updated the properties to the best of my ability, but needs to be revisited as part of a test suite.

benoitf (approved):

riuvshin (approved):

skabashnyuk:
@TylerJewell I think Florent as author can assist with Che CLI. Now we will be more focused on Codenvy in Docker.

TylerJewell:
No.  the cli is identical so when we do testing for codenvy we should donsame tests for Che please.

PR #3252, opened 2016-12-01
CHE-2598 Added support for specifying shared docker volume lib for ws…

sbryzak:
### What does this PR do?

This pull request adds a new feature allowing a named Docker volume to be used to share client libraries between workspaces

CHE-2598

### New behavior
Added support for sharing client libraries (ws-agent, etc) using a named Docker volume.  Usage:

docker run -t -v /var/run/docker.sock:/var/run/docker.sock -e "CHE_LIB_VOLUME=che-lib-wsagent" eclipse/che-launcher start

In this example, "che-lib-wsagent" is the name of the Docker volume.  To create a named volume in Docker:

docker volume create --name che-lib-wsagent

See https://docs.docker.com/engine/reference/commandline/volume_create/ for more information.

…agent

Signed-off-by: Shane Bryzak <sbryzak@redhat.com>


TylerJewell:
Hi @sbryzak - this is an excellent addition!  Though we need to work on some things because we are about to deprecate the che-launcher in favor of the new eclipse/che-cli which simplifies a lot of things.

So I recommend the following changes:
1. Rename in the che.properties machine.docker.volumes.lib to "che.docker.volumes_from".

2. We also need to add in support for this into the new Che CLI.  The new CLI minimizes the use of environment variables unless they are absolutely necessary.  One of hte things that we do is to automatically set Che variables based upon the docker run syntax in the new CLI.  So for example, if you have mounted -v /some/path:/cli, then we know that you want the cli.log saved to the host path.  We do the same for data directories and dev mode if you happen to set :/repo.  So internal values can be defined automatically based upon the flags in the docker run command.  Would it make sense here for a user to just start che with "docker run <normal-stuff> --volumes-from=<value>, and then inside of the CLI we extract the <value> using introspection and if that value is set, then we will automatically set the property for Che?  If this is a yes, then there is a bit of work we need to do on the new CLI, but we can guide you on how ot do that.

sbryzak:
Hi @TylerJewell, I've renamed the property to che.docker.volumes_from as you recommended.  The changes I made to the launcher start up Che with the named library volume mounted as -v ${CHE_LIB_VOLUME}:/data/lib, so most certainly this mapping for /data/lib could be introspected, however when the workspaces are launched the --volumes-from is actually set to "che-server", which is the container which the workspace container copies its volumes from (and not the name of the volume specified by CHE_LIB_VOLUME).  This means that it's not quite a simple matter of introspecting the library volume mapping and using that same value in --volumes-from.

TylerJewell:
@sbryzak - I was hoping to do some hands on testing with this over the weekend, but was not able to get to it. Do you have any urgent need for this to be merged, or can you wait a few more days for me to get to it?

sbryzak:
@TylerJewell No urgent need for merging, please let me know if it needs any further work when you are able to take a look at it.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -165,6 +165,10 @@ che.docker.namespace=NULL
 # Docker unused containers cleanup period
 che.docker.unused_containers_cleanup_min=60
 
+# Docker container to use volumes-from for workspace containers, used to share ws-agent 
+# without using host directory
+che.docker.volumes_from=che-server
```
garagatyi: I think it is better to point in property name that the volume is for agents distribution.
@TylerJewell WDYT?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -504,6 +507,11 @@ private String createContainer(String workspaceId,
                                            .collect(toMap(Function.identity(),
                                                           value -> new PortBinding[0])))
                   .withVolumesFrom(toArrayIfNotNull(service.getVolumesFrom()));
+        
+        if (volumesFromContainerName != null) {
+        	hostConfig.withVolumesFrom(new String[] { volumesFromContainerName });
```
garagatyi: There is **addSystemWideContainerSettings** method for such code

garagatyi (changes requested):

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
TylerJewell: I agree.  We already have 'che.workspace.volume=' in their for the agent mounting of shared folders.  So to be consistnent we should then name this 'che.workspace.voluems_from='.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -504,6 +507,11 @@ private String createContainer(String workspaceId,
                                            .collect(toMap(Function.identity(),
                                                           value -> new PortBinding[0])))
                   .withVolumesFrom(toArrayIfNotNull(service.getVolumesFrom()));
+        
+        if (volumesFromContainerName != null) {
```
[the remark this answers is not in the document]
TylerJewell: +1

dockerfiles/launcher/launcher_funcs.sh
```diff
@@ -166,6 +166,22 @@ docker_run_with_debug() {
   fi
 }
 
+docker_run_with_volumes_lib() {
+  if has_lib_volume; then
+    docker_run_with_debug -v "${CHE_LIB_VOLUME}":/data/lib "$@"
```
TylerJewell, line 193: Ok - you will need to wait for the new Che CLI to be merged and then we need to refactor what you are doing with the che-launcher, as the che-launcher will be no more in favor of a newer solution.  

https://github.com/eclipse/che/pull/3301

Am I understanding you correctly in that the eclipse/che-server image will have two requirements:
1. <local-path>:/data/lib to the eclipse/che-server 'docker run' command?
2. AND, che.workspace.volumes_from= must be set within the che.properties in the che-server?

Are both required?  If so, in reading the description, it is a bit unclear for me what each part is providing. could you just explain that.  And when you do that, I can give some pointers on how to refactor the new CLI to support this.

sbryzak:
@mshaposhnik I've annotated the parameter with **@Nullable**, thanks for pointing out this discrepancy.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
sbryzak: I've renamed this property to che.workspace.volumes_from

dockerfiles/launcher/launcher_funcs.sh
sbryzak: Hi @TylerJewell , yes both requirements are needed.  The first step is required to map the named Docker volume to the /data/lib directory (inside the che-server container).  The second requirement is needed to tell any workspace containers that start up to use the same volumes that the che-server container is using, i.e. since che-server has a mapping for /data/lib to the named Docker volume, the workspace will have that same mapping also and therefore be sharing the ws-agent resources.

TylerJewell: @sbryzak - Hey Shane, any thoughts on the modifications required?  If it would help, i can do a screen share for 45 minutes any time this week to guide you through what we should do.  It would not take long, I think.

I'd like to add the M9 label for this issue, but cannot do it unless we agree on these additional changes, which are needed.

TylerJewell: The new CLI code is linked to in my previous comment - those links are all pointing to the internals of the new CLI.

TylerJewell:
@sbryzak - since you have two issues that need to make improvements into the new CLI, I am happy to do a screen share to help get you educated on how the new CLI works, and we can go through and make any necessary changes in there together.

dockerfiles/launcher/launcher_funcs.sh
sbryzak: Ah, no problem, I should have checked them first :)  I'm looking at the scripts now.

sbryzak: My pull request should now contain the required changes to [che.env](https://github.com/eclipse/che/pull/3252/commits/c3a6d78fc47f2f19d550224195842935fb485fc7#diff-e4ebb68a2356f42877c3d809f0e569ad).  I've left the previous changes to the launcher scripts intact for now, even though from my understanding the new CLI will make them redundant.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -132,6 +132,7 @@
     private final long                                          cpuPeriod;
     private final long                                          cpuQuota;
     private final WindowsPathEscaper                            windowsPathEscaper;
+    private final String                                        volumesFromContainerName;
```
garagatyi: It id better to call it volumesFrom because it is convenient term in docker ecosystem

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -515,6 +518,11 @@ private String createContainer(String workspaceId,
                                            .collect(toMap(Function.identity(),
                                                           value -> new PortBinding[0])))
                   .withVolumesFrom(toArrayIfNotNull(service.getVolumesFrom()));
+        
+        if (volumesFromContainerName != null) {
+        	hostConfig.withVolumesFrom(new String[] { volumesFromContainerName });
```
garagatyi: Here you overwrite value set in line 520. I believe you should append instead.

garagatyi (changes requested):

sbryzak:
@garagatyi I've made those changes as you requested.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -515,6 +518,17 @@ private String createContainer(String workspaceId,
...
                   .withVolumesFrom(toArrayIfNotNull(service.getVolumesFrom()));
+        
+        if (volumesFrom != null) {
+        	List<String> values = new ArrayList<String>();
+        	values.add(volumesFrom);
+        	
+        	if (hostConfig.getVolumesFrom() != null && hostConfig.getVolumesFrom().length > 0) {
+        	    values.addAll(Arrays.asList(hostConfig.getVolumesFrom()));
+        	}
+        	hostConfig.withVolumesFrom(toArrayIfNotNull(values));
```
garagatyi: In that case values can not be NULL

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -515,6 +518,17 @@ private String createContainer(String workspaceId,
                                            .collect(toMap(Function.identity(),
                                                           value -> new PortBinding[0])))
                   .withVolumesFrom(toArrayIfNotNull(service.getVolumesFrom()));
+        
+        if (volumesFrom != null) {
```
garagatyi: Please cover this code with unit tests. See MachineProviderImplTest

garagatyi: Can you move addition of this volumesFrom into **addSystemWideContainerSettings** method. There is a code with pretty the same logic.

garagatyi (changes requested):
@TylerJewell Can you review CLI changes? I'm not sure I can review it myself.

TylerJewell (approved):

TylerJewell:
@garagatyi - I have reviewed for the CLI and made some improvements to the language in the che.env file. It's good to go.  @sbryzak - we have all properties start off as commented out, so that the admin can uncomment them if necessary.  Looks good.

bmicklea:
@sbryzak we are in ramp down for M9 (which will be 5.0) so we're moving this to 5.1. Codenvy team will take it in the next sprint to make sure it's closed out.

skabashnyuk:
@sbryzak  can you comment https://github.com/eclipse/che/pull/3252#discussion_r93197624  and https://github.com/eclipse/che/pull/3252#discussion_r93194813?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
sbryzak: @garagatyi There is no access to the hostConfig object within the addSystemWideContainerSettings method - is there a way to access this configuration object that I am overlooking?

No problem with adding some unit tests.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
sbryzak: Ah yes of course, I've updated this line.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
sbryzak: I've added a new unit test for the volumesFrom parameter, some of the API in the test framework was a little confusing but hopefully it's ok.

garagatyi: It doesn't have to have access to host config. Take a look on how volumes are set [there](https://github.com/garagatyi/che/blob/5266cd930e71c3628ebbd13858676d597b3c8391/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java#L579-L579). If it is not clear for you how to set do that, please, ping me.

garagatyi: Yes, looks like you've added test correctly despite complexity of test framework. Do you have any suggestion how to make it less confusing?

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/MachineProviderImplTest.java
```diff
@@ -697,6 +697,30 @@ public void shouldBindCommonAndDevVolumesToContainerOnDevInstanceCreationFromRec
...
+    	service.setVolumes(cheServerVolumes);
+    	createInstanceFromRecipe(isDev, service);
+    	
+        ArgumentCaptor<CreateContainerParams> argumentCaptor = ArgumentCaptor.forClass(CreateContainerParams.class);
+        verify(dockerConnector).createContainer(argumentCaptor.capture());
+        verify(dockerConnector).startContainer(any(StartContainerParams.class));
+        
+        final String[] actualVolumesFrom = argumentCaptor.getValue().getContainerConfig().getHostConfig().getVolumesFrom();
+        assertTrue(Arrays.asList(actualVolumesFrom).contains(expectedVolumesFrom));
+    }
```
garagatyi, line 723: Please add one more test with volumes from set in che service to ensure that combination of volumesfrom from che service and predefined volumesfrom don't wipe-out each other.

sbryzak: I've added an additional test to cover this.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
sbryzak: These issues have both been fixed.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
sbryzak: The code has been added to addSystemWideContainerSettings.  I'm not sure what to suggest to make the test framework less confusing, perhaps after I spend more time with it I'll get some ideas. :)

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/MachineProviderImplTest.java
```diff
@@ -697,6 +697,56 @@ public void shouldBindCommonAndDevVolumesToContainerOnDevInstanceCreationFromRec
...
+    	service.setVolumes(cheServerVolumes);
+    	service.setVolumesFrom(cheServerVolumesFrom);
+    	createInstanceFromRecipe(isDev, service);
+    	
+        ArgumentCaptor<CreateContainerParams> argumentCaptor = ArgumentCaptor.forClass(CreateContainerParams.class);
+        verify(dockerConnector).createContainer(argumentCaptor.capture());
+        verify(dockerConnector).startContainer(any(StartContainerParams.class));
+        
+        final String[] actualVolumesFrom = argumentCaptor.getValue().getContainerConfig().getHostConfig().getVolumesFrom();
+        assertTrue(Arrays.asList(actualVolumesFrom).contains(expectedVolumesFrom));
```
garagatyi, line 748: If one **volumesFrom** is provided in **MachineProviderImpl** and another in **CheServiceImpl** then looks like **.getHostConfig().getVolumesFrom()** should return array of length 2. But **expectedVolumesFrom** doesn't contain **expectedVolumesFrom**. Am I missing something?

BTW Testng has **assertEqualsNoOrder** for comparing arrays.

sbryzak: Hi @garagatyi , perhaps my understanding is off so please correct me, but doesn't **CheServiceImpl** represent the Che service, while the **MachineProviderBuilder** provides the machine (i.e. workspace) instances? Testing for an assertion that **.getHostConfig().getVolumesFrom()** returns a length of 2 causes the test to fail, so it is not the case that a **volumesFrom** is being added twice.

You asked me earlier if I had any suggestions for making the tests easier to understand - perhaps some kind of documentation that explained line by line what was happening in the test would be helpful.  As it stands, there seems to be a lot going on that I'm a bit unclear about, which means all I can do is make assumptions about what is happening based on the method and variable names.  I think some docs would help to reduce this confusion.

garagatyi: Sorry I confused you because of my mistake. I'll try to explain what I meant and where is my mistake.

MachineProviderImpl accepts in constructor some additional settings, e.g. volumes, /etc/hosts entries, env variables, etc. So these values should be added into every container (or every dev-machine container). They are system-wide.
On the other hand there is CheServiceImpl which represents particular machine configuration.
So if CheServiceImpl and MachineProviderImpl have volumes configured they should be combined in getHostConfig.
But I missed that MachineProviderImpl doesn't accept volumesFrom.

So my comment is not relevant. I'm really sorry that I confused you.

garagatyi:
@sbryzak Have you tried to run this branch? Is it complete or need additional PRs to finish sharing of agents using volumesFrom instead of volumes?

sbryzak:
@garagatyi I've tested the branch and it seems to work correctly, no further PRs are required.

garagatyi:
I have feeling that if we comment regular mounts of [wsagent](https://github.com/eclipse/che/blob/master/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalCheInfrastructureProvisioner.java#L99-L99) and [terminal](https://github.com/eclipse/che/blob/master/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalCheInfrastructureProvisioner.java#L109-L109) of agents in this PR Che won't work.
It looks like volumesfrom from this PR are not used in agents installation sequence. WDYT?

TylerJewell:
In a discussion with Red Hat today, the feeling is that the new CLI provides most of functionality that was going to be provided by this. We will close this PR and then open a new PR, related to providing a default data volume (or an internal docker volume).

l0rd:
I'm closing this PR because in the meantime the benefits (making lib volume configurable) have already been introduced in Che.

PR #3265, opened 2016-12-02
CHE-3264 - Starting the che server as an user

### LINKED ISSUE #3264: Starting the che server as an user (implementing the CHE_USER property)
When CHE_USER environment variable is set Che server should be run as the specified user (otherwise, by default, it's run as root). That's currently broken.
See https://issues.jboss.org/browse/CHE-6

**Che version:**   any
**OS and version:**  any
**Docker version:**  any
**Che install:**    docker 

**Additional information:**

* Problem started happening recently, didn't happen in an older version of Che: No
* Problem can be reliably reproduced, doesn't happen randomly: Yes

snjeza:
### What does this PR do?

Fixes #3264 

#3264 

The following is an example of starting che as the "test" user:

```
docker run -p:8080:8080 \
        --name che \
	--rm \
	-u `id -u test` \
        -e CHE_USER=test \
        -v /tmp/data:/data \
	-v /etc/group:/etc/group:ro \
           eclipse/che-server:ver
```

- as the "user" user (the same as without the patch):

```
docker run -p:8080:8080 \
           --name che \
           --rm \
           -v /tmp/data:/data \
           eclipse/che-server:ver
```

To check which user has started che, enter the following:

```
docker exec che id
```

The "test" user has to be in the docker group.
Workspaces are still started as the "user" user.

BTW, the patch also fixes several issues related to permissions when the che server starts the first time.

To start a workspace (with the codenvy/ubuntu_jdk8 image, for instance) as an user, we could do the following: 

- the user has to have NOPASSWD sudo privilege. It runs the "CMD sudo /usr/sbin/sshd -D && tail -f /dev/null" command
This can be done by creating the sudo group and adding the user to it (in the base OS).

- the user has the /home/user base directory with required artifacts (tomcat, maven, etc.)
We can solve this issue by creating the following script:

```
CHE_USER=`id -un`
if [ "$CHE_USER" != user ]; then
  sudo chown -R ${CHE_USER}:docker /home/user
  sudo ln -s /home/user `eval echo "~${CHE_USER}"`
fi
```

The script should be run immediately after starting the workspace and before starting some other agent.
If this idea is acceptable, I can also create a patch to start the workspace as an user.

### DIFF
--- a/dockerfiles/launcher/launcher_funcs.sh
+++ b/dockerfiles/launcher/launcher_funcs.sh
@@ -110,22 +110,44 @@ docker_run() {
     -v "$CHE_DATA_LOCATION" \
     -p "${CHE_PORT}":"${CHE_PORT}" \
     --restart="${CHE_RESTART_POLICY}" \
-    --user="${CHE_USER}" \
     -e "CHE_LOG_LEVEL=${CHE_LOG_LEVEL}" \
     -e "CHE_IP=$CHE_HOST_IP" \
     --env-file=$ENV_FILE \
     "$@"
- 
    rm -rf $ENV_FILE > /dev/null
+}
+
+get_user_id() {
+   CHE_USER_UID=$(docker run -t \
+     -v /etc/passwd:/etc/passwd:ro,Z \
+     -v /etc/group:/etc/group:ro,Z \
+     alpine id -u ${CHE_USER})
+   CHE_USER_GID=$(docker run -t \
+     -v /etc/passwd:/etc/passwd:ro,Z \
+     -v /etc/group:/etc/group:ro,Z \
+      alpine getent group docker | cut -d: -f3)
+   echo -n "${CHE_USER_UID}" | tr '\r' ':'; echo -n ${CHE_USER_GID}
+}
+
+docker_run_with_che_user() {
+   if [ "${CHE_USER}" != "root" ]; then
+     docker_run -e CHE_USER=${CHE_USER} \
+      -v /etc/group:/etc/group:ro,Z \
+      -v /etc/passwd:/etc/passwd:ro,Z \
+      --user=$(get_user_id) \
+      "$@"
+   else
+     docker_run --user="${CHE_USER}" "$@"
+   fi
 }
 
 docker_run_if_in_vm() {
   # If the container will run inside of a VM, additional parameters must be set.
   # Setting CHE_IN_VM=true will have the che-server container set the values.
   if is_docker_for_mac || is_docker_for_windows || is_boot2docker; then
-    docker_run -e "CHE_IN_VM=true" "$@"
+    docker_run_with_che_user -e "CHE_IN_VM=true" "$@"
   else
-    docker_run "$@"
+    docker_run_with_che_user "$@"
   fi
 }
 
--- a/dockerfiles/che/entrypoint.sh
+++ b/dockerfiles/che/entrypoint.sh
@@ -226,11 +226,27 @@ init() {
   export CHE_DATA="/data"
   CHE_DATA_HOST=$(get_che_data_from_host)
 
+  CHE_USER=${CHE_USER:-root}
+  export CHE_USER=$CHE_USER
+  if [ "$CHE_USER" != "root" ]; then
+    if [ ! $(getent group docker) ]; then
+      echo "!!!"
+      echo "!!! Error: The docker group doesn't exist."
+      echo "!!!"
+      exit 1
+    fi
+    export CHE_USER_ID=`id -u ${CHE_USER}`:`getent group docker | cut -d: -f3`
+    sudo chown -R ${CHE_USER}:docker ${CHE_DATA}
+    sudo chown -R ${CHE_USER}:docker ${CHE_HOME}
+  fi
   ### Are we going to use the embedded che.properties or one provided by user?`
   ### CHE_LOCAL_CONF_DIR is internal Che variable that sets where to load
   if [ -f "/conf/che.properties" ]; then
     echo "Found custom che.properties..."
     export CHE_LOCAL_CONF_DIR="/conf"
+    if [ "$CHE_USER" != "root" ]; then
+      sudo chown -R ${CHE_USER}:docker ${CHE_LOCAL_CONF_DIR}
+    fi
   else
     echo "Using embedded che.properties... Copying template to ${CHE_DATA_HOST}/conf."
     mkdir -p /data/conf


TylerJewell (approved):

TylerJewell:
Nice work @snjeza.  When this is merged, this docs page needs to be updated with instructions about how to run the Che server as a particular user.

https://dash.readme.io/project/eclipse-che/v5.0/docs/usage-docker-server

@jamesdrummond - this will need to be updated in the converted docs.

@snjeza - on the tactics necessary to run the workspace as particular user - both of those solutions seem viable. The question I have is whether we can insert those modifications into images after they have been started as workspaces?  Almost as a type of agent?  I ask this because we have been working hard to allow the idea that any kind of base image can be used as a che workspace, and trying to avoid having permanent base images that are needed in order to run with Che.

snjeza:
> The question I have is whether we can insert those modifications into images after they have been started as workspaces

@TylerJewell 
Yes, we can do it. I have created a separate branch - https://github.com/snjeza/che/tree/starting_workspace_as_user.  It includes this PR.
The base OS has to contain the sudo group. A user must be a member of this group. The group ID isn't important, since the /etc/sudoers file uses the group name, but not the group ID.

 When you start che with:
 
```
docker run -p:8080:8080 \
        --name che \
	--rm \
	-u `id -u test` \
        -e CHE_USER=test \
        -v /tmp/data:/data \
	-v /etc/group:/etc/group:ro \
           eclipse/che-server:ver
```

che will start a workspace as the "test" user.
If necessary, che mounts the /etc/group and /etc/passwd files and run a script enables che agents to be started as the "test" user.
I have tested the codenvy/ubuntu_jdk8 image. The image hasn't been changed.
I suppose other codenvy images have similar features (use the "user" user, the "/home/user" home directory and the "sudo" group).

assembly/assembly-main/src/assembly/bin/docker.sh
```diff
@@ -78,6 +78,13 @@ init() {
   export CHE_DATA="/data"
   CHE_DATA_HOST=$(get_che_data_from_host)
 
+  CHE_USER=${CHE_USER:-user}
+  export CHE_USER=$CHE_USER
+  if [ "$CHE_USER" != "user" ]; then
+    export CHE_USER_ID=`id -u ${CHE_USER}`
```
benoitf: question: where CHE_USER_ID is used ?

assembly/assembly-main/src/assembly/bin/docker.sh
```diff
@@ -78,6 +78,13 @@ init() {
   export CHE_DATA="/data"
   CHE_DATA_HOST=$(get_che_data_from_host)
 
+  CHE_USER=${CHE_USER:-user}
+  export CHE_USER=$CHE_USER
```
benoitf: why do we need to export CHE_USER ?

assembly/assembly-main/src/assembly/bin/docker.sh
```diff
@@ -78,6 +78,13 @@ init() {
...
+  if [ "$CHE_USER" != "user" ]; then
+    export CHE_USER_ID=`id -u ${CHE_USER}`
+  fi
+  sudo chown -R ${CHE_USER}:docker ${CHE_DATA}
```
benoitf: if permissions are ok (like windows/mac) why do we need a chown ?

assembly/assembly-main/src/assembly/bin/docker.sh
```diff
@@ -87,6 +94,7 @@ init() {
   if [ -f "/conf/che.properties" ]; then
     echo "Found custom che.properties..."
     export CHE_LOCAL_CONF_DIR="/conf"
+    sudo chown -R ${CHE_USER}:docker ${CHE_LOCAL_CONF_DIR}
```
benoitf: if permissions are ok (like windows/mac) why do we need a chown ?

assembly/assembly-main/src/assembly/bin/docker.sh
```diff
@@ -121,9 +129,9 @@ init() {
     export CHE_WORKSPACE_STORAGE_CREATE_FOLDERS=false
   fi
 
-  # Ensure that the user "user" has permissions for CHE_HOME and CHE_DATA
-  sudo chown -R user:user ${CHE_HOME}
-  sudo chown -R user:user ${CHE_DATA}
+  # Ensure that the user has permissions for CHE_HOME and CHE_DATA
+  sudo chown -R ${CHE_USER}:docker ${CHE_HOME}
+  sudo chown -R ${CHE_USER}:docker ${CHE_DATA}
```
benoitf: if permissions are ok (like windows/mac) why do we need a chown ?

l0rd: `sudo chown -R ${CHE_USER}:docker ${CHE_DATA}` is already done at line 86. Why do you need to do it again?

snjeza:
> why do we need to export CHE_USER ?
> question: where CHE_USER_ID is used ?

The variables are used when starting a workspace as an user. See https://github.com/snjeza/che/tree/starting_workspace_as_user
Maybe its is enough to export CHE_USER_ID only.

> if permissions are ok (like windows/mac) why do we need a chown ?

The chown command is copied from https://github.com/eclipse/che/blob/master/assembly/assembly-main/src/assembly/bin/docker.sh#L125

I have also added the chown command to several other places, because I had a problem when starting the che server the first time or changing a user.
The patch has been tested on Linux (Fedora F24).

assembly/assembly-main/src/assembly/bin/docker.sh
```diff
@@ -259,7 +267,7 @@ check_docker
 init
 
 # run application
-"${CHE_HOME}"/bin/che.sh run &
+sudo -u $CHE_USER -E sh -c "${CHE_HOME}/bin/che.sh run" &
```
l0rd: Have you tested if Che can be cleanly stopped? For example, when stopping Che, running workspaces (if they exist) are stopped too? I'm wondering that because SIGTERM signals are usually not well propagated when wrapping the command with a shell (i.e. `sh -c`).

assembly/assembly-main/src/assembly/bin/docker.sh
l0rd: I've the same concern as @benoitf. Have you forgot to remove CHE_USER_ID?

l0rd (changes requested):
@snjeza nice job. I've added some comments inline. Shouldn't you patch che_launcher too? Currently option `--user="${CHE_USER}` is used in Che launcher. With your approach you need `-e CHE_USER=${CHE_USER}` instead.

snjeza:
> I've added some comments inline. Shouldn't you patch che_launcher too? Currently option --user="${CHE_USER} is used in Che launcher. With your approach you need -e CHE_USER=${CHE_USER} instead.

I have updated the patch.
You have to use the following command:
```
docker run --rm -t -v /var/run/docker.sock:/var/run/docker.sock -v /etc/passwd:/etc/passwd:ro,Z -v /etc/group:/etc/group:ro,Z -e CHE_USER=test eclipse/che-launcher:ver  start
```

> I've the same concern as @benoitf. Have you forgot to remove CHE_USER_ID?

It is used when starting a workspace as an user
See https://github.com/snjeza/che/tree/starting_workspace_as_user (https://github.com/eclipse/che/commit/edb85aecdc790303fb29a88a1fe10eb1d9395e7d)

> sudo chown -R ${CHE_USER}:docker ${CHE_DATA} is already done at line 86. Why do you need to do it again?

I have updated the patch.

> Have you tested if Che can be cleanly stopped? For example, when stopping Che, running workspaces (if they exist) are stopped too?

It works properly when the /var/run/docker.sock file has a valid permission (666).

l0rd:
> I have updated the patch.

I can't see any update 😢 . Is it me or you have failed to update your commit? 

> (about $CHE_USER_ID) It is used when starting a workspace as an user. See https://github.com/snjeza/che/tree/starting_workspace_as_user (edb85ae)

But that's another branch right? In this case $CHE_USER_ID should be introduced in your second branch, not this one.

> It works properly when the /var/run/docker.sock file has a valid permission (666)

`/var/run/docker.sock` should not have permission 666 (this gives root access to everybody). It should be 644 or more restrictive ([c.f. §3.4 of this great document on Docker security](https://benchmarks.cisecurity.org/tools2/docker/CIS_Docker_1.12.0_Benchmark_v1.0.0.pdf)) and Che should work with 644 permissions.

snjeza:
> I can't see any update ?? . Is it me or you have failed to update your commit?

I have just updated the patch.

snjeza:
> /var/run/docker.sock should not have permission 666 (this gives root access to everybody). It should be 644 or more restrictive (c.f. §3.4 of this great document on Docker security) and Che should work with 644 permissions.

That could be a problem. We can't use the docker remote API if the /var/run/docker.sock file has the 644 permission.

```
$ sudo chmod 644 /var/run/docker.sock
$ docker run  -u `id -u test` --rm -it -v /etc/passwd:/etc/passwd:ro,Z -v /etc/group:/etc/group:ro,Z -v /var/run/docker.sock:/var/run/docker.sock codenvy/ubuntu_jdk8 /bin/bash
$ echo -e "GET /containers/json HTTP/1.0\r\n" | nc -U /var/run/docker.sock
nc: unix connect failed: Permission denied
```

```
$ sudo chmod 666 /var/run/docker.sock
$ docker run  -u `id -u test` --rm -it -v /etc/passwd:/etc/passwd:ro,Z -v /etc/group:/etc/group:ro,Z -v /var/run/docker.sock:/var/run/docker.sock codenvy/ubuntu_jdk8 /bin/bash
$ echo -e "GET /containers/json HTTP/1.0\r\n" | nc -U /var/run/docker.sock
HTTP/1.0 200 OK
Content-Type: application/json
Server: Docker/1.12.3 (linux)
Date: Tue, 06 Dec 2016 11:59:20 GMT
...
```

If a user is a member of the root group, the Remote Docker API can be used when the /var/run/docker.sock has the following permission:
```
srw-rw-r-- 1 root root 0 Dec  6 12:02 /var/run/docker.sock
```

dockerfiles/launcher/launcher_funcs.sh
```diff
@@ -105,17 +105,32 @@ is_docker_for_windows() {
...
-    -v "$CHE_DATA_LOCATION" \
-    -p "${CHE_PORT}":"${CHE_PORT}" \
-    --restart="${CHE_RESTART_POLICY}" \
-    --user="${CHE_USER}" \
-    -e "CHE_LOG_LEVEL=${CHE_LOG_LEVEL}" \
-    -e "CHE_IP=$CHE_HOST_IP" \
-    --env-file=$ENV_FILE \
-    "$@"
- 
+   if [ "${CHE_USER}" != "root" ]; then
```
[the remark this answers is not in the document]
snjeza: I will try to do it.

l0rd:
@snjeza: user `test` should belong to the docker (or root) group. [That's always required when running docker as non-root user](https://docs.docker.com/engine/installation/linux/ubuntulinux/#/manage-docker-as-a-non-root-user).

snjeza:
> @snjeza: user test should belong to the docker group. That's always required when running docker as non-root user.

I know that, but the Remote Docker API doesn't works in that case 

The Remote Docker API works when we call the following commands: 

```
sudo chgrp root /var/run/docker.sock
sudo chmod 664 /var/run/docker.sock
sudo usermod -a -G root test
docker ...
echo ...
```

but doesn't work for the following commands: 

```
sudo chgrp docker /var/run/docker.sock
sudo chmod 664 /var/run/docker.sock
sudo usermod -a -G docker test
docker ... 
echo ...
```

snjeza:
> Wouldn't it be more coherent (and easier to read) if you defined a function docker_run_with_che_user and move the if condition there as it's done for docker_run_with_debug, docker_run_with_external_hostname, docker_run_with_conf etc...?

I have updated the patch.

> Do we really need to add --user="id -u ${CHE_USER}"? I think that if we remove that Che would still be started as user $CHE_USER and the launcher could be run without having to mount/etc/groupand/etc/passwd`. What do you think?

I have removed --user="id -u ${CHE_USER}", but Che still starts as the "user" user

l0rd:
@snjeza for the permissions problems try:
```bash
docker run  -u `id -u test`:`getent group docker | cut -d: -f3`   \
                    --rm -it -v /etc/passwd:/etc/passwd:ro,Z      \
                    -v /etc/group:/etc/group:ro,Z                 \
                    -v /var/run/docker.sock:/var/run/docker.sock  \
                    codenvy/ubuntu_jdk8                           \
                   /bin/bash
```

snjeza:
> for the permissions problems try:

@l0rd 
It works. I have updated the patch.

l0rd:
Please check if Che is stopped cleanly:
- start Che using the launcher
- create a new workspace
- stop Che using the launcher
- use docker ps to verify that container has been stopped

And to avoid mounting /etc/hosts and /etc/passwd/ when running the launcher
you can use a technique to retrieve uid similar to the one used to get host
IP address: add a new function (get_user_uid) in launcher_funcs.sh. This
function would run a docker container with -v /etc/passwd:/etc/passwd that
will output the user uid.

On Tue, 6 Dec 2016 at 22:59, Snjeza <notifications@github.com> wrote:

for the permissions problems try:

@l0rd <https://github.com/l0rd>
It works. I have updated the patch.

—
You are receiving this because you were mentioned.


Reply to this email directly, view it on GitHub
<https://github.com/eclipse/che/pull/3265#issuecomment-265286200>, or mute
the thread

snjeza:
> Please check if Che is stopped cleanly:

It works fine.

> And to avoid mounting /etc/hosts and /etc/passwd/ when running the launcher

I have tried the following function:
```
get_user_id() {
  docker run -t -v /etc/passwd:/etc/passwd:ro,Z -v /etc/group:/etc/group:ro,Z alpine echo -n "`id -u ${CHE_USER}`:`getent group docker | cut -d: -f3`"
}
```

but a docker within the launcher still requires mounting /etc/passwd and /etc/group

l0rd:
@snjeza sorry for answering so late. Function `get_user_id()` is exactly what I was thinking of but I was surprised that the launcher still requires mounting /etc/password and /etc/group. 

So I've looked at it closely and I found out that the problem is that instructions `id -u ${CHE_USER}` and `getent group docker | cut -d: -f3` are interpreted before command `docker run` is executed. These instructions are executed inside the launcher container and that's why you need to have `passwd` and `group` mounted there.

To avoid that you can change your function to something like

```bash
get_user_id() {
 CHE_USER_UID=$(docker run -t \
    -v /etc/passwd:/etc/passwd:ro,Z \
    -v /etc/group:/etc/group:ro,Z \
    alpine id -u ${CHE_USER})
 CHE_USER_GID=$(docker run -t \
    -v /etc/passwd:/etc/passwd:ro,Z \
    -v /etc/group:/etc/group:ro,Z \
    alpine getent group docker | cut -d: -f3)
 echo -n "${CHE_USER_UID}" | tr '\r' ':'; echo -n ${CHE_USER_GID}
}
```

snjeza:
@l0rd 
It works. I have updated the patch.

TylerJewell:
@benoitf - do you have a plan for how we merge these improvements with the new Che CLI that we have merged in master?

assembly/assembly-main/src/assembly/bin/docker.sh
```diff
@@ -121,6 +129,9 @@ init() {
     export CHE_WORKSPACE_STORAGE_CREATE_FOLDERS=false
   fi
 
+  # Ensure that the user has permissions for CHE_HOME and CHE_DATA
+  sudo chown -R ${CHE_USER}:docker ${CHE_HOME}
```
benoitf: chown should be only done if required

snjeza: If we don't call chown, che will start only if we do that manually before starting the che server.

l0rd (approved):

l0rd:
Great job @snjeza ! These changes should make Che nicer to use on Linux desktops. 

I agree with @benoitf that you should verify the real need for `chown` instructions. Default user in eclipse/che-server Dockerfile has recently changed from `user` to `root`. 

Besides that everything is good for me.

l0rd:
@benoitf and @TylerJewell: if your idea is to remove `docker.sh` maybe @snjeza could just commit changes to `launcher_funcs.sh` and discard all changes to `docker.sh` to get that merged?

benoitf (changes requested):
che server is using root user by default
so all chown operations should happen only if we're not with root or if we don't have permissions

benoitf:
@l0rd sure

snjeza:
I have updated the patch.

benoitf (approved):

benoitf:
assembly/assembly-main/src/assembly/bin/docker.sh is now at https://github.com/eclipse/che/blob/master/dockerfiles/che/entrypoint.sh

assembly/assembly-main/src/assembly/bin/docker.sh
```diff
@@ -78,6 +78,13 @@ init() {
   export CHE_DATA="/data"
   CHE_DATA_HOST=$(get_che_data_from_host)
 
+  CHE_USER=${CHE_USER:-root}
+  export CHE_USER=$CHE_USER
+  if [ "$CHE_USER" != "root" ]; then
+    export CHE_USER_ID=`id -u ${CHE_USER}`:`getent group docker | cut -d: -f3`
```
benoitf: what is happening if docker group is not there ?

snjeza: I have updated the patch.

benoitf (approved):
Thanks for having updated the PR

benoitf:
I let @l0rd to do the merge

l0rd:
🎉  Well done @snjeza! Are you going to work on the same issue for workspace users now?

TylerJewell:
Also, now that this is merged, it needs to be tested with the new CLI, too @l0rd .

snjeza:
@TylerJewell 
I have tested the master branch.
The following command will start che as the "test" user

```
docker run --rm -t -v /var/run/docker.sock:/var/run/docker.sock -e CHE_USER=test eclipse/che-launcher:ver start
```

snjeza:
@l0rd 
Yes, I have already started work on it.

TylerJewell:
@snjeza - the eclipse/che-launcher image is about to be deprecated and removed.

Can you provide the syntax for how you would use this syntax with eclipse/che-server, and then I can look at a potential improvement to the new CLI?

snjeza:
@TylerJewell 
```
docker run -p 8080:8080 \
           --name che \
	   --rm \
	   -u `id -u test` \
           -e CHE_USER=test \
           -v /tmp/data:/data \
	   -v /etc/group:/etc/group:ro \
           eclipse/che-server:ver 
```

TylerJewell:
@snjeza - two things:
1) If you do not want to change the user name, can you still mount `/etc/group` and `/etc/passwd`?
2) If you are using docker-compose syntax what is the equivalent service command for the `-u` entry that you provided?

This will be the template that will need to be modified to support this with the new CLI.
https://github.com/eclipse/che/blob/master/dockerfiles/init/modules/compose/templates/docker-compose.yml.erb

snjeza:
@TylerJewell 

> If you do not want to change the user name, can you still mount /etc/group and /etc/passwd?

After removing  the "user" user, we can do it.

> If you are using docker-compose syntax what is the equivalent service command for the -u entry that you provided?

We would need to add "user: $UID" to docker-compose.yml and export the $UID variable

PR #3202, opened 2016-11-29
Zend Debugger for PHP

tolusha:
### What does this PR do?
Adds support for debugging applications that runs on PHP with Zend Debugger on board.

### What issues does this PR fix or reference?
- PHP Support [#2590](https://github.com/eclipse/che/issues/2590)

### New behavior
- Possibility to add and run new _PHP - ZEND DEBUGGER_ debug configuration in Che
- Possibility to debug local PHP applications in CLI or Web mode

### Author
@bartlomiej-laczkowski

PR #3277, opened 2016-12-05
Add "Run/Debug PHP script" commands to the PHP and Zend stacks

kaloyan-raev:
### What does this PR do?

Adds commands to the PHP and Zend stacks for running and debugging PHP scripts, a.k.a. CLI apps:
- "run php script" - executes the PHP script open in the current editor
- "debug php script" - executes the PHP script open in the current editor in debug mode, i.e. it defines the required `QUERY_STRING` env var, so a debug session is opened to the Zend Debugger engine.

Both commands have a check if there is a file currently opened in the editor. If this is not the case, a user friendly message is printed.

This PR should be merge after merging the following two PRs:
- Zend Debugger for PHP #3202
- Install Zend Debugger in the default PHP stack https://github.com/eclipse/che-dockerfiles/pull/55

### Previous behavior
The user had to type these commands manually in the Terminal. While it is relatively easy for the "run php script" command, the `QUERY_STRING` env var is complex and requires copying the value from some doc page.

### New behavior
Easier running and debugging of PHP scripts by executing Che commands.

Signed-off-by: Kaloyan Raev <kaloyan.r@zend.com>

### DIFF
--- a/ide/che-core-ide-stacks/src/main/resources/stacks.json
+++ b/ide/che-core-ide-stacks/src/main/resources/stacks.json
@@ -590,7 +590,7 @@
     "id": "php-default",
     "creator": "ide",
     "name": "PHP",
-    "description": "Default PHP Stack with PHP 5.5, most popular extensions and dockerimager.",
+    "description": "Default PHP Stack with PHP 5.6, most popular extensions and dockerimager.",
     "scope": "general",
     "tags": [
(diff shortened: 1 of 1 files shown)


vparfonov (approved):

vparfonov:
@tolusha

tolusha (approved):

PR #3280, opened 2016-12-05
CHE-51: Fixing multi-machine workspace creation (e.g. MySql-Java stack)

ibuziuk:
### What does this PR do?
Fixes default (MySQL-Java) workspace creation

https://issues.jboss.org/browse/CHE-51
### Previous behavior
Error while creating workspace from multi-machine stack (MySQL - JAVA) 
Openshift service with this name already exist 

### New behavior
WorkspaceID is generated randomly if CHE_WORKSPACE_ID env var does not exist. Running non-dev container as root. For vanilla Che launch (MySQL-JAVA stack) dev-machine UID - 1000, db -0

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/OpenShiftConnector.java
@@ -29,8 +29,11 @@ import com.openshift.restclient.model.IReplicationController;
 import com.openshift.restclient.model.IService;
 import com.openshift.restclient.model.IServicePort;
 import com.openshift.restclient.model.deploy.DeploymentTriggerType;
+
+import org.apache.commons.lang.RandomStringUtils;
 import org.eclipse.che.plugin.docker.client.json.ContainerCreated;
 import org.eclipse.che.plugin.docker.client.json.ContainerInfo;
+import org.eclipse.che.plugin.docker.client.json.ImageConfig;
 import org.eclipse.che.plugin.docker.client.json.PortBinding;
 import org.eclipse.che.plugin.docker.client.params.CreateContainerParams;
 import org.eclipse.che.plugin.docker.client.params.RemoveContainerParams;
@@ -43,7 +46,6 @@ import javax.inject.Named;
 import javax.inject.Singleton;
 import java.io.IOException;
 import java.util.*;
-import java.util.regex.Matcher;
 import java.util.stream.Stream;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
@@ -67,6 +69,9 @@ public class OpenShiftConnector {
     public static final String CHE_DEFAULT_EXTERNAL_ADDRESS = "172.17.0.1";
     public static final String CHE_SERVER_INTERNAL_IP_ADB = "172.17.0.5";
     public static final String CHE_CONTAINER_IDENTIFIER_LABEL_KEY = "cheContainerIdentifier";
+    public static final String UID_ROOT = "0";
+    public static final String UID_USER = "1000";
+    public static final int CHE_WORKSPACE_AGENT_PORT = 4401;
 
     private final IClient            openShiftClient;
     private final IResourceFactory   openShiftFactory;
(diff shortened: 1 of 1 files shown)


ibuziuk:
@l0rd could you please take a look ?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/OpenShiftConnector.java
```diff
@@ -285,7 +292,7 @@ private String createOpenShiftDeploymentConfig(IProject cheProject,
         ((DeploymentConfig) dc).setName(dcName);
         ((DeploymentConfig) dc).setNamespace(cheProject.getName());
 //        ((DeploymentConfig) dc).getNode().get("spec").get("template").get("spec").get("dnsPolicy").set("Default");
-        ((DeploymentConfig)dc).getNode().get("spec").get("template").get("spec").get("securityContext").get("runAsUser").set("1000");
+//        ((DeploymentConfig)dc).getNode().get("spec").get("template").get("spec").get("securityContext").get("runAsUser").set("1000");
```
l0rd: If not needed this line can be removed

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/OpenShiftConnector.java
```diff
@@ -311,15 +318,21 @@ private void addContainer(IDeploymentConfig dc, String workspaceID, String image
         dc.getContainer(containerName).setImagePullPolicy(IMAGE_PULL_POLICY_IFNOTPRESENT);
 
         ModelNode dcFirstContainer = ((DeploymentConfig)dc).getNode().get("spec").get("template").get("spec").get("containers").get(0);
+        
+        boolean isDevMachine = isDevMachine(exposedPorts);
+        // Run non-dev machine as root user
+        String UID = isDevMachine ? UID_USER : UID_ROOT;
```
l0rd: What if it's not a devMachine and the user is not root? I'm thinking about a [multi-machine workspace] where the non-dev machine default user is not root. For these scenarios wouldn't it possible to retrieve the image default user?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/OpenShiftConnector.java
```diff
@@ -311,15 +318,21 @@ private void addContainer(IDeploymentConfig dc, String workspaceID, String image
...
-        dcFirstContainer.get("securityContext").get("runAsUser").set("1000");
+        dcFirstContainer.get("securityContext").get("runAsUser").set(UID);
 
         // LivenessProbe
-        dcFirstContainer.get("livenessProbe").get("tcpSocket").get("port").set(4401);
-        dcFirstContainer.get("livenessProbe").get("initialDelaySeconds").set(120);
-        dcFirstContainer.get("livenessProbe").get("timeoutSeconds").set(1);
+        // TODO: CHE-53 Improve workspaces liveness probe for non-dev machines
+        if (isDevMachine) {
+            dcFirstContainer.get("livenessProbe").get("tcpSocket").get("port").set(4401);
```
l0rd: Now that you have defined constant CHE_WORKSPACE_AGENT_PORT you could use it here too.

l0rd (commented):
Nice work @ibuziuk! I've added a few comments

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/OpenShiftConnector.java
ibuziuk: @l0rd thanks for review. For now the logic is the following - all dev mashines are run as user (UID 1000) and all non-dev machines as root (UID 0). Basically, I just checked what UID is used for MySQL-Java stack in by running Che without OpenShift.  I will try to update PR with the logic of getting user from image config

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/OpenShiftConnector.java
ibuziuk: ok, I will correct this

ibuziuk:
@l0rd PR has been updated

l0rd (approved):

l0rd:
Thank you @ibuziuk!

Issue #3281, opened 2016-12-05
Build is forcing all contributions to be copyrighted to Codenvy


gorkem:
maven license check plugin is failing the build if the copyright owner for a file is different from Codenvy. I do not think we should be adding such restrictions on an open source project.

tolusha:
@gorkem 
Hello. You can configure exclusion in the `pom.xml` of your module.

```
            <plugin>
                <groupId>com.mycila</groupId>
                <artifactId>license-maven-plugin</artifactId>
                <configuration>
                    <excludes>
                        ...
                    </excludes>
                </configuration>
            </plugin>
```

ibuziuk:
@tolusha but it means that all companies will be facing this issue while contributing to the project and will need to configure excludes for every single change. Why there should be a check of copyright owner at all ? AFAIK, according to the Eclipse rules the one who creates a file put the copyright owner in the license header. Other contributors that make changes in the file should add info in the "Contributors" section.

skabashnyuk:
@ibuziuk There is an Eclipse foundation requirement to have license on all files. We did that with mycila license-maven-plugin.  If you know  how better it can be configured please introduce new PR. I have only one rule - it have to be done automatically with maven. Without automation this requirement is hard to achieve.

ibuziuk:
@skabashnyuk fair enough ;-) I will work on this issue

gorkem:
@ibuziuk take a look at http://www.mojohaus.org/license-maven-plugin/ it looks like it has better configuration options.

skabashnyuk:
@ibuziuk FYI https://github.com/mycila/license-maven-plugin/issues/119

ibuziuk:
@skabashnyuk yup, I have seen this upstream issue. IMO it is a bit different from what we are trying to achieve - {INITIAL COPYRIGHT OWNER} should be any company

ibuziuk:
After some investigation I came to the conclusion that what we are trying to achieve is not possible to configure via mycila plugin. It is doable with mojohaus one though. BUT... there are a few things that should be taken in consideration:

1. The mojohaus plugin is more flexible in configuration due to it's [Header model](http://www.mojohaus.org/license-maven-plugin/header.html). Basically, license header consists of several parts which are separated by special characters. As a result, the valid license header generated by default would look the following way:

```
/*-
 * #%L
 * %%
 * Copyright (C) 2012 - 2016 Codenvy, S.A.
 * %%
[...]

PR #3282, opened 2016-12-05
Refactor DockerInstanceRuntimeInfo#getServers() (#2030)

### LINKED ISSUE #2030: Che server should connect to ws-agent on internal URL

amisevsk:
### What does this PR do?

Enables che-server to communicate directly with wsagent containers when running in Docker through use of a property in `che.properties` (`che.docker.ip.use_internal_address`)

Following feedback on #2837, adds interfaces **ServerEvaluationStrategyProvider**, **ServerEvaluationStrategy**, and class **DefaultServerEvaluationStrategy** to replace `DockerInstanceRuntimeInfo#getServers()`. `ServerEvaluationStrategyProvider` provides a `ServerEvaluationStrategy`, which supports the `getServer(String portProtocol)` method. This simplifies the `getServers()` method in `DockerInstanceRuntimeInfo` by abstracting the functionality into a simpler class, while also allowing for extensibility in the future through different implementations of `ServerEvaluationStrategy`.

Fixes #2030, and addresses suggested refactor in #2837.

### Previous behavior
Che-server would always ping wsagent on external port, which could cause issues. E.g., if wsagent is available on ports
   - 172.17.0.5:4401 (*ephemeral* port - within **docker0** network)
   - 172.17.0.1:32701 (exposed port - accessible from host network)
then che-server will try to connect to wsagent at 172.17.0.1:32701 in all cases. This connection can be blocked by the firewall in certain cases, preventing workspaces from starting correctly.

### New behavior
Che-server can now optionally communicate with wsagent directly (i.e. via 172.17.0.5:4401, above), assuming both containers are on the same docker network. This avoids the issue of the host firewall blocking traffic.

### PR Checklist
- [x] Tests Added
- [x] Tests Passed


amisevsk:
@garagatyi @l0rd Thoughts?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
```diff
@@ -234,124 +229,20 @@ public String projectsRoot() {
 
     @Override
     public Map<String, ServerImpl> getServers() {
+        ServerEvaluationStrategy strategy = provider.getStrategy(info, serversConf, internalHost);
```
garagatyi: I think that I imaged this code in a wrong way. Because it looks just like I described and it looks not very good. 
What do you think if:
1. Strategies are injected in MapBinder
2. Some strategy provider is bound as provider of Server**s**EvaluationStrategy
3. In constructor Server**s**EvaluationStrategy is injected
4. In constructor strategy is assigned to a field.
5. In getServers we call strategy.getServers(info, internalHost) and it returns Map<String, ServerImpl> with filled ports and addresses.

@l0rd @amisevsk WDYT?

garagatyi (changes requested):
This PR is much closer to ideal solution then the previous one. Good job!
Now I figured out what was my mistake when I suggested new design for that. Please elaborate on what you think about suggestion I made in comments.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DefaultServerEvaluationStrategy.java
```diff
@@ -0,0 +1,253 @@
...
+        if (ports.isEmpty()) {
+            return;
+        }
+
+        // Make sure we set internal/external address according to precedence (see doc)
+        String containerAddress = containerInfo.getNetworkSettings().getIpAddress();
+        String dockerAddress    = containerInfo.getNetworkSettings().getGateway();
+        this.useEphemeralPorts = true;
+
+        if (internalAddress != null) {
```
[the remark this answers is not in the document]
amisevsk: I'm split on this -- on one hand, splitting the functionality in DefaultServerEvaluationStrategy into two strategies (default and e.g. local) would simplify the logic in determining which port and address to use, while on the other I feel like the two strategies would be very similar. The majority of the getServers() method would be identical -- we would pretty much just remove [the conditional that sets internalAddressAndPort](https://github.com/amisevsk/che/blob/ace0b7cceccbe1b7ab0b096f5f7ea1a218046289/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DefaultServerEvaluationStrategy.java#L183-L188).

I agree with you that using the true/false property is ugly, though. I'd like to hear Mario's thoughts.

For OpenShift, IIRC we actually don't use the new strategy -- we depend on the `che.docker.ip.external` property.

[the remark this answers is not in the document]
garagatyi: If we refactor code in such a way that common code would go to some abstract class or something like that strategies would be super simple. Probably in that case default strategy can have logic of ephemeral/exposed port choice.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -151,6 +151,12 @@ che.docker.ip=NULL
 # This is unusual, but happens for example in Docker for Mac when containers are in a VM.
 che.docker.ip.external=NULL
 
+# If true, then uses the internal address and port of workspace Docker containers (i.e. within the Docker
+# Docker network) instead of the docker host address and the port provided by the Docker daemon. May be necessary if the
+# Che server is unable to contact workspace agents, but will prevent communication if the server and workspace
+# are not on the same network.
+che.docker.ip.use_internal_address=false
```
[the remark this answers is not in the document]
amisevsk: I've changed it so that property refers to strategy name, and strategies are injected in MapBinder.

amisevsk:
I've added a commit that changes some things around. 

1. Instead of getServer(portProtocol), the method is now getServer**s**(containerInfo, internalHost, serverConfMap)
1. ServerEvaluationStrategy is now abstract, and hosts most of the getServers() code.
2. Implementations of ServerEvaluationStrategy only need to implement abstract getInternalAddressesAndPorts() and getExternalAddressesAndPorts() methods.
4. Strategy implementations are injected in MapBinder, which is injected into ServerEvaluationStrategyProvider. ServerEvaluationStrategyProvider now implements Provider\<ServerEvaluationStrategy\>

The tests are currently broken, I can fix them later -- if we're going in this direction then I think DockerInstanceRuntimeInfo#getServers() tests should instead test Strategies directly anyways.

@garagatyi WDYT?

TylerJewell:
Looks great, @amisevsk.  I have one addition for you to make. We are about to make the switch so that Docker is the preferred way to run Che.  This means that a che.env file will the default way that people configure Che.  The template for this has properties that are converted into che.properties overides. An underscore '_' represents a dot '.'.  And a double underscore represents an underscore.  Can you add in similar docs that you have for che.properties into che.env?  I think it belongs either in the networking or in the docker section.

https://github.com/eclipse/che/blob/master/dockerfiles/init/manifests/che.env#L112

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceRuntimeInfo.java
amisevsk: I've tried to implement this in the latest commit -- see message below.

amisevsk:
@TylerJewell Thanks for the update! This looks like a great change.

I'll add an entry to che.env -- is there a preferred way to surface the server evaluation strategy options? Should javadoc in classes refer to the property (che.properties) or the environment variable (che.env)? 

Also, for my understanding: the new che.env file is generally defining the environment variables that are currently set by the che-launcher docker image, correct?

TylerJewell:
@amisevsk - sorry this slipped through my fingers for a couple days.

To answer your questions:
1. The che.env is a file that end users will configure that will set all properties, whether they are things that need to go into che.properties or not, as there are environment variables like CHE_HOST which need to be set, but not necessarily into the che.properties. Doing a customer che.properties is hard for end users as well, becuase then they have to figure out how to get it inside the JVM or the container we designated.  So we have implemented a system that will take these values and use puppet to transform them into properties if necessary.  We also have support for any custom property applied in the che.env file being inserted for the user into a custom che.properties.

2. So you just need to modify the templated che.env with your custom property, and then test it with the new CLI.  To test the new CLI you may need to build it locally.

cd /dockerfiles/base <---- build this
cd /dockerfiles/cli <---- build this

You can then run it:
`docker run -it --rm -v /var/run/docker.sock:/var/run/docker.sock -v <path>:/data eclipse/che-cli:nightly start`

TylerJewell:
@garagatyi - we need you to review the changes and add any further comment. We are ready to merge, otherwise.

TylerJewell (approved):

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerMachineFactory.java
```diff
@@ -81,12 +80,9 @@ DockerNode createNode(@Assisted("workspace") String workspaceId,
      * Creates {@link DockerInstanceRuntimeInfo} instance using assisted injection
      *
      * @param containerInfo description of docker container
-     * @param containerExternalHostname docker host external hostname (used by the browser)
-     * @param containerInternalHostname docker host internal hostname (used by the wsmaster)
```
garagatyi, line 85: please add docs for this parameter

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ServerEvaluationStrategyProvider.java
```diff
@@ -0,0 +1,60 @@
...
+
+    private ServerEvaluationStrategy strategy;
+
+    @Inject
+    public ServerEvaluationStrategyProvider(Map<String, ServerEvaluationStrategy> strategies,
+                                            @Named("che.docker.server_evaluation_strategy") String chosenStrategy) {
+        if (strategies.containsKey(chosenStrategy)) {
+            this.strategy = strategies.get(chosenStrategy);
+        } else {
+            LOG.warn(String.format("Property che.docker.server_evaluation_strategy=%s"
```
garagatyi: It will be hard to debug and realize what happened if developer made a typo in strategy name for example. So it is better to fail component start in that case instead of usage of default strategy.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ServerEvaluationStrategy.java
```diff
@@ -0,0 +1,197 @@
...
+     * @param containerInfo the ContainerInfo object that describes the container.
+     * @param internalAddress address passed into {@code getServers}; used as fallback if address cannot be
+     *        retrieved from containerInfo.
+     * @return a Map of port protocol (e.g. "4401/tcp") to address (e.g. "localhost:32317")
+     */
+    protected abstract Map<String, String> getExternalAddressesAndPorts(ContainerInfo containerInfo,
+                                                                        String internalAddress);
+
+    @Inject
+    protected ServerEvaluationStrategy(@Nullable @Named("che.docker.ip")          String internalAddressProperty,
```
garagatyi: We put constructor before other methods in Che source code. Please follow that to not distract other devs.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ServerEvaluationStrategy.java
```diff
@@ -0,0 +1,197 @@
...
+        if (containerInfo.getConfig() != null && containerInfo.getConfig().getLabels() != null) {
+            labels = containerInfo.getConfig().getLabels();
+        }
+
+        Map<String, String> internalAddressesAndPorts = getInternalAddressesAndPorts(containerInfo, internalHost);
+        Map<String, String> externalAddressesAndPorts = getExternalAddressesAndPorts(containerInfo, internalHost);
+
+        Map<String, ServerImpl> servers = new LinkedHashMap<>();
+
+        for (String portProtocol : portBindings.keySet()) {
```
garagatyi, line 103: Please add unit tests for that class

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ServerEvaluationStrategy.java
```diff
@@ -0,0 +1,197 @@
...
+            String internalAddressAndPort = internalAddressesAndPorts.get(portProtocol);
+            String externalAddressAndPort = externalAddressesAndPorts.get(portProtocol);
+            ServerConfImpl serverConf = getServerConfImpl(portProtocol, labels, serverConfMap);
+
+            // Add protocol and path to internal/external address, if applicable
+            String internalUrl = null, externalUrl = null;
```
garagatyi: Please do not declare several variables in one line. We don't use such syntax in Che.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/LocalDockerServerEvaluationStrategy.java
```diff
@@ -0,0 +1,82 @@
...
+        super(internalAddress, externalAddress);
+    }
+
+    @Override
+    protected Map<String, String> getInternalAddressesAndPorts(ContainerInfo containerInfo, String internalHost) {
+        String internalAddressContainer = containerInfo.getNetworkSettings().getIpAddress();
+
+        String internalAddress;
+        boolean useExposedPorts = true;
+        if (internalAddressContainer != null && !internalAddressContainer.isEmpty()) {
```
garagatyi: Use IsNullOrEmpty  - with it code usually looks clearer

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/LocalDockerServerEvaluationStrategy.java
```diff
@@ -0,0 +1,82 @@
...
+            internalAddress = internalAddressContainer;
+        } else {
+            internalAddress = internalHost;
```
garagatyi, line 67: If internal host provided from another component why it is not prevail?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/LocalDockerServerEvaluationStrategy.java
```diff
@@ -0,0 +1,82 @@
...
+            useExposedPorts = false;
```
garagatyi, line 68: Can you elaborate on why in that case strategy uses ephemeral ports?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/LocalDockerServerEvaluationStrategy.java
```diff
@@ -0,0 +1,82 @@
...
+                addressesAndPorts.put(portEntry.getKey(), internalAddress + ":" + exposedPort);
+            } else {
+                addressesAndPorts.put(portEntry.getKey(), internalAddress + ":" + ephemeralPort);
+            }
+        }
+        return addressesAndPorts;
+    }
+
+    @Override
+    protected Map<String, String> getExternalAddressesAndPorts(ContainerInfo containerInfo, String internalHost) {
```
garagatyi, line 87: Please add unit tests for this class

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DefaultServerEvaluationStrategy.java
```diff
@@ -0,0 +1,83 @@
...
+                                 internalAddressContainer :
+                                 internalHost;
+
+        Map<String, List<PortBinding>> portBindings = containerInfo.getNetworkSettings().getPorts();
+
+        return getAddressesAndPorts(internalAddress, portBindings);
+    }
+
+    @Override
+    protected Map<String, String> getExternalAddressesAndPorts(ContainerInfo containerInfo, String internalHost) {
```
garagatyi, line 74: Please add unit tests for that class

garagatyi (changes requested):
Please address my comments

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/LocalDockerServerEvaluationStrategy.java
amisevsk: Internal Host is a holdover from before -- it ultimately comes from `DockerNode#getHost()`. Prior to these changes, this value was used only when running Che Server as a standalone tomcat server. I've left it here as a fallback, but it could probably be removed from LocalDockerServerEvaluationStrategy altogether. It has lowest precedence because that was the way it was used before: [it's the same as internalHostnameAssisted here](https://github.com/eclipse/che/blob/11b3b9de67cd9f7c40e3641b748e97af8ae3c89a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalDockerInstanceRuntimeInfo.java#L93). Note that when running Che using che-launcher, the env variable `CHE_DOCKER_MACHINE_HOST_INTERNAL` is set to [the value defined here](https://github.com/eclipse/che/blob/11b3b9de67cd9f7c40e3641b748e97af8ae3c89a/dockerfiles/launcher/launcher_funcs.sh#L273)

The code that defines internalHost is in [DockerConnectorConfiguration](https://github.com/eclipse/che/blob/11b3b9de67cd9f7c40e3641b748e97af8ae3c89a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnectorConfiguration.java#L153).

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/LocalDockerServerEvaluationStrategy.java
amisevsk: We need to use ephemeral ports whenever we are not using the direct IP address of the workspace container. Internal host requires ephemeral ports for communication, but if you think it can safely be removed from this strategy, this could be simplified.

TylerJewell (approved):
I thought about it, Angel, and could not identify a better name.  So let's roll with it!

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DefaultServerEvaluationStrategy.java
amisevsk: I've rewritten ServerEvaluationStrategyTest.java to be more clear, it tests getInternalAddress() and getExternalAddress() for the strategies.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/LocalDockerServerEvaluationStrategy.java
amisevsk: I've rewritten ServerEvaluationStrategyTest.java to be more clear, it tests getInternalAddress() and getExternalAddress() for the strategies.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/LocalDockerServerEvaluationStrategy.java
amisevsk: Fixed throughout. Thank you.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ServerEvaluationStrategy.java
amisevsk: Most of the getServers() logic is tested in DockerInstanceRuntimeInfoTest, as part of the tests there. I can move it out if that's preferred, but I think of ServerEvaluationStrategy as tightly coupled with DockerInstanceRuntimeInfo.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ServerEvaluationStrategyProvider.java
amisevsk: That makes a lot of sense.

What is the preferred way of doing this in Che? I've changed it to throw a ServerException with a message now, but the error that appears while launching a workspace includes errors about being unable to inject the constructor, which is not very clean. Since I'm implementing `Provider<ServerEvaluationStrategy>`, we can't throw an exception in the get() method. I suppose implementing Provider is not strictly required, WDYT?

bmicklea:
@amisevsk we're in ramp down for M9 (our GA release), we've moved the milestone to 5.1 for now but if you're able to get the conflicts resolved and docs and tests added in time we'll add it to M9.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ServerEvaluationStrategy.java
garagatyi: If something is implemented in base class it is better to test that in tests for base class. Then we won't have to test that in all implementations

garagatyi (changes requested):
Good job! This is really big and complex contribution! And  I believe it really simplifies code.
The only thing needed is separation of strategies tests into separate test classes. I provided more details in code comment.

Please also squash your git history to avoid pushing commits that are parts of a single task and can not be used separately.

garagatyi:
@musienko-maxim Please run QA cycle for that  PR.

musienko-maxim:
Can you resolve all conflicts? QA team will try to launch qa session after this

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/ServerEvaluationStrategyTest.java
```diff
@@ -0,0 +1,358 @@
...
+        when(containerInfo.getConfig()).thenReturn(containerConfig);
+        when(containerConfig.getLabels()).thenReturn(Collections.emptyMap());
+    }
+
+    /**
+     * Test: che.docker.ip property takes highest precedence for internal address
+     * @throws Exception
+     */
+    @Test
+    public void defaultStrategyShouldUseInternalIpPropertyToOverrideContainerInfo() throws Exception {
```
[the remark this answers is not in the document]
amisevsk: Fixed, thank you.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ServerEvaluationStrategy.java
amisevsk: Done.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ServerEvaluationStrategyProvider.java
amisevsk: The further issue that MachineRuntimeInfo (and thus DockerInstanceRuntimeInfo) do not leave space for throwing a ServerException. Is there an internal way to signal an error otherwise? 

Alternatively, we could add throws declarations to MachineRuntimeInfo, and adjust as necessary, but DockerInstanceRuntimeInfo is used pretty widely in the codebase.

amisevsk:
I've fixed the tests and cleared the merge conflict. 

There is one outstanding issue, of signalling that `che.docker.server_evaluation_strategy` has a typo, that is outstanding. I would like @garagatyi 's input on how to best deal with this case. Once that is dealt with I will squash these commits into a few logical chunks.

Apart from that, the ip-validation check is failing due to commit 9186d0c, but that will be fixed when the history is squashed. 

Finally, current builds fail without the -Dskip-validate-sources flag, due to issue #3281. I'm not sure about how best to proceed there.

@musienko-maxim Is anything else needed before QA can begin?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ServerEvaluationStrategyProvider.java
garagatyi: I think that declaring throwing of an exception in MachineRuntimeInfo is better to omit because, as you said, it is used in a lot of places and bothering all that code with catching exception would make a lot of code much more complicated.

So I think we need to find another way to achieve what you want. 

Can you elaborate on what problem you have faced with lack of exception throwing?

amisevsk: For useability, it would be ideal if typing the property incorrectly would cause workspaces to fail to start, displaying a message saying something along the lines of "Server Evaluation Strategy \<property\> does not correspond to any implemented strategy". The user could read their typo, and make the appropriate fix in che.env without too much trouble.

I'm not familiar enough with the relevant sections of the code to see an easy way to fail workspace starting and also display this message to the user without throwing an exception. This works relatively well, except that the only place I can throw this exception is in the ServerEvaluationStrategyProvider constructor, which litters the error message with "failed to inject constructor" and other information about the internals of Che. 

My initial approach was to add a error message to the log and use the default strategy, but I agree that this can easily lead to confusion. Another option is to log an error and use a no-op implementation of ServerEvaluationStrategy -- this would cause launching a workspace to fail, but the user would still have to view the logs to find their error.

How would you like me to surface this error to the user?

musienko-maxim:
Thanks! It's enough.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ServerEvaluationStrategyProvider.java
garagatyi: If you set this class to be eager singleton Guice will have to create it on Che bootstrap and mistake in strategy name would prevent successful start of Che. Maybe this variant is ok for you? 
In addition to that check and error throwing can be moved to **@PostConstruct** method. Then maybe message of Che start fail would be more satisfactory.

garagatyi:
Unfortunately we can't merge PR with build failing because of source validation. So options here are:
- waiting until #3281 is fixed
- disable license validation for some files in pom.xml, merge PR, wait fix of #3281, remove disabled license validation

WDYT?

garagatyi (approved):

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ServerEvaluationStrategyProvider.java
amisevsk: After playing around with it for awhile, I think I've got something that works. 

The Exception is thrown in a @PostConstruct method, which causes workspace startup to fail and display a user-readable message. Binding ServerEvaluationStrategyProvider as a Singleton caused an issue where Che would still start without displaying errors (except for in the logs), but be in an inconsistent state and have to be stopped from the docker cli directly.

amisevsk:
I've squashed the history into three commits: 
    - The first adds ServerEvaluationStrategy, the Provider, and implmentations,
    - The second adds test cases for ServerEvaluationStrategy
    - The third changes existing files to use ServerEvaluationStrategy.

I can squash these further, but I feel like having three separate steps means that the history is more readable in general.

@garagatyi I'm happy to add excludes entries for the problem files and make a note to remove them after #3281 is resolved. I assume that the preferred location for this in the che-parent pom.xml, correct?

garagatyi:
You can leave these 3 commits.

You need to edit nearest pom.xml to exclude file from license check. Take a look at this [pom](https://github.com/eclipse/che/blob/abc1367c1d3eb4311c2aa760fdfd4da38c0669bd/core/commons/che-core-commons-lang/pom.xml#L73)

amisevsk:
The modified files have been excluded from license checks -- build should complete normally now.

garagatyi:
@amisevsk I found that default strategy behaves in a bit different way than code in master. It fetches gateway from docker container inspection. But current code doesn't use gateway at all. Can you elaborate on why did you introduced that change?

garagatyi:
Also in ServerEvaluationStrategy fields internalAddressProperty externalAddressProperty are not used by require them in constructor. It limits usage of that class since some implementation might want to not use (and not know) about these properties. WDYT if we remove these fields?
Also Inject annotation is useless in that class since it is abstract, so constructor is always called by another constructor.

garagatyi:
Could you also fix links in javadocs:
```
{@link ServerConf}
{@link DefaultServerEvaluationStrategy#serverConf}
{@link DefaultServerEvaluationStrategy#labels}

amisevsk:
@garagatyi The reason for the container inspection change is that I needed to reproduce functionality in LocalDockerInstanceRuntimeInfo, and getting gateway from ContainerInfo seemed like the simplest way to do this. 

In the current master branch, LocalDockerInstanceRuntimeInfo is responsible for calling the DockerInstanceRuntimeInfo constructor. [Here](https://github.com/eclipse/che/blob/6091d007b0f045ddc78807b2036fbdeaa0c979ed/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalDockerInstanceRuntimeInfo.java#L96), it sets the value that is passed as containerInternalHostname to DockerInstanceRuntimeInfo to an environment variable (CHE_DOCKER_MACHINE_HOST). If this env var is not null, it overrides any other value, including the property in che.properties and che.env. The value of CHE_DOCKER_MACHINE_HOST is set:

- `CHE_DOCKER_MACHINE_HOST="{$CHE_IP}"` in [che.sh](https://github.com/eclipse/che/blob/6091d007b0f045ddc78807b2036fbdeaa0c979ed/assembly/assembly-main/src/assembly/bin/che.sh#L189)
- `CHE_IP=$CHE_HOST_IP` in [launcher_funcs.sh](https://github.com/eclipse/che/blob/6091d007b0f045ddc78807b2036fbdeaa0c979ed/dockerfiles/launcher/launcher_funcs.sh#L114)
- `CHE_HOST_IP=${CHE_HOST_IP:-${DEFAULT_DOCKER_HOST_IP}}` in [launcher.sh](https://github.com/eclipse/che/blob/6091d007b0f045ddc78807b2036fbdeaa0c979ed/dockerfiles/launcher/launcher.sh#L94)
- Finally, [DEFAULT_DOCKER_HOST_IP](https://github.com/eclipse/che/blob/6091d007b0f045ddc78807b2036fbdeaa0c979ed/dockerfiles/launcher/launcher.sh#L55) is set via [get_docker_host_ip()](https://github.com/eclipse/che/blob/6091d007b0f045ddc78807b2036fbdeaa0c979ed/dockerfiles/launcher/launcher_funcs.sh#L273) in launcher_funcs.sh

The values of the variables in get_docker_host_ip() are defined [here](https://github.com/eclipse/che/blob/6091d007b0f045ddc78807b2036fbdeaa0c979ed/dockerfiles/launcher/launcher.sh#L26). I figured that using the value from ContainerInfo yields the same functionality in a cleaner and easier to read way.

I admit that I have not been able to test the Boot2docker case fully, but as far as I know, as long as the workspace container is attached to the bridge network, gateway is not null after a docker inspect. The change also means that the property che.docker.ip takes precedence as intended (since before it would get overridden). If we cannot get gateway from ContainerInfo, behaviour is the same as if CHE_DOCKER_MACHINE_HOST is null in the current master.

I'm not sure if this affects the codenvy side of things, but if CHE_DOCKER_MACHINE_HOST is used there, you should be able to have identical functionality by setting CHE_DOCKER_IP instead.

Let me know if the above is clear -- I've been on holiday for a little bit so my memory might be hazy.

Regarding the other changes -- I think they're all good. I'll implement them when I have a moment later today.

garagatyi:
Ok, it's clear, thank you. I think this behavior is OK, we just need to check that it works on OSX and Windows. QA will take care of testing on these OSs.

amisevsk:
@garagatyi Sorry about the delay, I've fixed the javadocs and moved the internalAddressProperty and externalAddressProperty into the concrete classes. The changes have been squashed into commit e5bc686.

garagatyi:
I'm also on holidays )
Thanks that you addressed my comments. These were optional, just advises on how to make code clearer.

amisevsk:
I've rebased to fix merge conflicts and updated header copyright years. It looks like IP validation is failing due to lack of Signed-off-by footers in master. 

@garagatyi Is there anything else I can do to help with merging this change?

l0rd:
@amisevsk I think you merged commit f8ba55f076ec064079450152a087bedb127aaf1b in your branch instead of rebasing on top of it

amisevsk:
@l0rd Woops! Thank you, I don't know how that happened. It should be fixed now.

l0rd (approved):

garagatyi:
@amisevsk Since this PR is tagged with 5.1 milestone we can't merge it now. We will have to wait for 5.0 release. Then I'll ask QA team to re-run QA cycle (previous QA on this PR passed) to ensure that changes in master made since previous QA cycle didn't brake anything in your PR.

TylerJewell:
@amisevsk - 5.0.0 is released.  @garagatyi, @l0rd - lets go ahead and merge this now early in our cycle, and we'll have QA execute tests within the master branch.  We have a window of a few days to merge major PRs for 5.1 and do not want to wait for the QA cycle.

amisevsk:
@garagatyi @TylerJewell Thanks! Let me know if there are any QA issues during the next cycle.

garagatyi:
@amisevsk Thank you that you were patient with all my comments in review. But now code looks muuuch better than before your contribution!

Issue #3185, opened 2016-11-28
Request for Lombok plugin


KakashiMamun:
There is no lombok plugin. So lombok related methods highlighted as error

**Reproduction Steps:**

1. User Lombok annotation

**Expected behavior:**
Shouldn't be highlighted as error.

**Observed behavior:**

Lombok used classes shows error.

**Che version:**    [5.0.0-M7]     
**Che install:**     Docker container   

**Additional information:**

If anyone explain me about jdt implementation for eclipse che i will try to implement.

KakashiMamun:
https://github.com/KakashiMamun/plugin-lombok

TylerJewell:
Thanks for the link - will you be issuing a pull request that can be tracked?  The pull request will have a complete list of recommended changes, CI components, and doc improvements necessary to become part of the product.

KakashiMamun:
Yes @ tyler. Because lombok is not in che-parent and I used lombok in my plugin dependacy. In my Local I build with -Denforcer.skip=true. but i reckon in codenvy-ci it will fail. so someone should add lombok version 1.16.10 in che-parent. Let me know if you want to do make a PR for that.

PR #3288, opened 2016-12-06
Plugin lombok

KakashiMamun:
### What does this PR do?
This added plugin support for lombok in eclipse che workspace

#3185

### Previous behavior
Previously lombok annotations was not suported in eclipse che. Hence Lombok related code shows error 
in IDE.

### New behavior
This PR has changes made to lombok code to work with eclipse che. But the catch is Lombok works as a java agent and in eclipse che case the wsagent tomcat has to be started with -javaagent:lombok.jar 
and the jar has to be in the workspace image. Details of enabling the plugin can be found in README.md file of the plugin. The enabling of the plugin is a little bit messy now. But I will try to improve if i can have guideline from che developers.

### DIFF
--- a/assembly/assembly-wsagent-war/pom.xml
+++ b/assembly/assembly-wsagent-war/pom.xml
@@ -245,6 +245,17 @@
             <artifactId>javax.servlet-api</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.plugin</groupId>
+            <artifactId>che-plugin-lombok-server</artifactId>
+            <scope>runtime</scope>
+            <exclusions>
+                <exclusion>
+                    <artifactId>lombok</artifactId>
+                    <groupId>org.projectlombok</groupId>
+                </exclusion>
+            </exclusions>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
--- a/plugins/plugin-lombok/che-plugin-lombok-server/src/main/java/lombok/launch/Main.java
+++ b/plugins/plugin-lombok/che-plugin-lombok-server/src/main/java/lombok/launch/Main.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright (C) 2014-2015 The Project Lombok Authors.
+ * 
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ * 
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ * 
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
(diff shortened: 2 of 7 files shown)


KakashiMamun:
I would like to add that the lombok dependancy that i used in my plugin pom.xml is not in eclipse che depndancy management pom. so i guess it will fail the build. What i should do ?

TylerJewell:
Hi @KakashiMamun - I think you are wanting to get a dependency added to the che-parent, which is in a different repository:

https://github.com/eclipse/che-parent

Is this what you are seeking?

KakashiMamun:
Yes @ tyler. Because lombok is not in che-parent and I used lombok in my plugin dependacy. In my Local I build with -Denforcer.skip=true. but i reckon in codenvy-ci it will fail. so someone should add lombok version 1.16.10 in che-parent. Let me know if you want to do make a PR for that.

TylerJewell:
Yes, it would need to be a PR for the che-parent repository, thanks.  And then a maintainer will look at this at some point after we make the 5 GA release.

etlweather:
while we wait for this plugin to be merged, is there a simple way to add lombok.jar to Che's runtime classpath (not the project classpath)?

TylerJewell:
Create a custom stack.  The docs have the instructions for what to do.  This will give you what you seek.

etlweather:
@TylerJewell - sorry, I'm a bit new to Eclipse Che and I don't know all the ins and outs of it. I do have a custom stack but I can't find in the instructions as to what to do. 

Do you mean the instructions at https://github.com/KakashiMamun/plugin-lombok ? 

This require building Che from source which seems a bit much to add a dependency...

TylerJewell:
The stacks are the runtimes that users select when creating workspaces.  You can add these rather easily.  See more here - https://www.eclipse.org/che/docs/workspace/stacks/index.html.

Create a custom stack with your JAR pre-packaged into the runtime, and then it's available from commands or in the terminal, or even by your projects.

etlweather:
Ok, I did that. Where I get bugged is this. In regular Eclipse, we add `javaagent:lombok.jar` to the eclipse.ini file as a jvm arg.

What process needs to have that added to it in Che?

I tried adding it to the Tomcat that is started for the `ws-agent`. It is pointing correctly to the lombok jar. When I do that, I get errors showing in the dev-machine output "NoClassDefFoundError: org/eclipse/jdt/internal/compiler/ast/ASTNode`

So it seems lombok is not loading with the right classloader or Che use a different compiler?

TylerJewell:
This product is not like a desktop IDE.  It's a distributed system with many different services running in different locations.  I am sorry, but your question is not clear as to what you are trying to achieve.  Adding a JAR into Che itself to be used by a project is not a concept that makes sense.  

Have you studied the architecture and are you aware of the differences between che, workspaces, workspace agents, and the IDE?  Please, make sure you study these items - as I suspect what you need is a custom stock, but not to modify Che itself.

KakashiMamun:
@etlweather , The class is not found because lombok agent classlaoder is different from tomcat classlaoder. that's why I repacked lombok with thread classloader. but to do that i had to rebuild che. it would be nice to know how to use using stack.

assembly/assembly-wsagent-war/pom.xml
```diff
@@ -245,6 +245,17 @@
             <artifactId>javax.servlet-api</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.plugin</groupId>
+            <artifactId>che-plugin-lombok-server</artifactId>
+            <scope>runtime</scope>
+            <exclusions>
+                <exclusion>
+                    <artifactId>lombok</artifactId>
```
skabashnyuk, line 254: why it's in excludes here?

assembly/assembly-wsagent-war/pom.xml
```diff
@@ -245,6 +245,17 @@
             <artifactId>javax.servlet-api</artifactId>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.plugin</groupId>
+            <artifactId>che-plugin-lombok-server</artifactId>
+            <scope>runtime</scope>
```
skabashnyuk, line 251: why it's runtime?

plugins/plugin-lombok/che-plugin-lombok-server/src/main/java/lombok/launch/Main.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright (C) 2014-2015 The Project Lombok Authors.
+ *
```
skabashnyuk, line 3: Why it's not EPL?

skabashnyuk (changes requested):
@KakashiMamun  Can you comment on this https://github.com/eclipse/che-dependencies/pull/22#issuecomment-265409568

TylerJewell:
We are closing this PR as we do not have a way to maintain ongoing.  Ok to reopen with a strategy in place.

Issue #3287, opened 2016-12-06
Create GET method to return snapshot configuration


skabashnyuk:
On UD we need to know how our snapshots is configured. We need to know that over HTTP.
Idea is to make similar to ```/user/settings``` service to return ```auto_restore``` , ```auto_snapshot``` configuration. 

Need for https://github.com/codenvy/codenvy/issues/937

PR #3289, opened 2016-12-06
Add Workspace settings service

mkuznyetsov:
### What does this PR do?
Adds ability to view server auto-snapshot and auto-restore settings through API

https://github.com/eclipse/che/issues/3287

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
@@ -18,6 +18,7 @@ import io.swagger.annotations.ApiResponses;
 import io.swagger.annotations.Example;
 import io.swagger.annotations.ExampleProperty;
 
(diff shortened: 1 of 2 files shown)


wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -704,6 +714,14 @@ public WsAgentHealthStateDto checkAgentHealth(@ApiParam(value = "Workspace id")
         return check;
     }
 
+    @GET
```
sleshchenko, line 715: Should we add swagger docs here?

sleshchenko (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
mkuznyetsov: Added swagger annotation

skabashnyuk (approved):

PR #3296, opened 2016-12-06
Remove redundant info log call

vzhukovs:
Remove redundant log info call that displays stored json object which describes application state.

@mshaposhnik @ashumilova review it, please.


mshaposhnik (approved):
Thanks

ashumilova (approved):

Issue #3276, opened 2016-12-05
Not able to run che with cli


skabashnyuk:
Che is not starting from cli
**Reproduction Steps:**

1. ```docker run -it --rm  -v /var/run/docker.sock:/var/run/docker.sock -v /home/sj/.che:/che eclipse/che-cli:nightly start```
2. docker logs che

**Expected behavior:**

che is not working

**Observed behavior:**

```
[sj@sj-i7:1]$ docker logs -f che
time="2016-12-05T13:45:49Z" level=fatal msg="Get http:///var/run/docker.sock/v1.18/containers/json: dial unix /var/run/docker.sock: permission denied. Are you trying to connect to a TLS-enabled daemon without TLS?"
Error when running "docker ps":
time="2016-12-05T13:45:49Z" level=fatal msg="Post http:///var/run/docker.sock/v1.18/containers/create: dial unix /var/run/docker.sock: permission denied. Are you trying to connect to a TLS-enabled daemon without TLS?"
time="2016-12-05T13:45:49Z" level=fatal msg="Post http:///var/run/docker.sock/v1.18/containers/create: dial unix /var/run/docker.sock: permission denied. Are you trying to connect to a TLS-enabled daemon without TLS?"
time="2016-12-05T13:45:49Z" level=fatal msg="Post http:///var/run/docker.sock/v1.18/containers/create: dial unix /var/run/docker.sock: permission denied. Are you trying to connect to a TLS-enabled daemon without TLS?"
Using embedded assembly...
Using embedded che.properties... Copying template to /home/user/che/conf.
time="2016-12-05T13:45:49Z" level=fatal msg="Get http:///var/run/docker.sock/v1.18/containers/json: dial unix /var/run/docker.sock: permission denied. Are you trying to connect to a TLS-enabled daemon without TLS?"
```

**Che version:**    5.0.0-M8 snapshot  
**OS and version:** Ubuntu 16.04   
**Docker version:** 1.12.3   
**Che install:**  che cli

TylerJewell:
```
[sj@sj-i7:che][master]$ ls -la /var/run/docker.sock
srw-rw---- 1 root docker 0 лис 24 11:51 /var/run/docker.sock
[sj@sj-i7:che][master]$
```

 ```DOCKER_OPTS="-H tcp://0.0.0.0:2375 --insecure-registry 172.19.20.18:5000```

skabashnyuk:
```
[...]

PR #3297, opened 2016-12-06
Start eclipse/che-server docker image as root by default

benoitf:
### What does this PR do?

Start eclipse/che-server docker image as root by default

https://github.com/eclipse/che/issues/3276

### Previous behavior
User user had to be in docker group but with a specific gid which is not always the case

### New behavior
Start che server as root user by default

Use root user by default to start che-server
another PR is about specifying how to run che-server with a custom uid/gid
https://github.com/eclipse/che/pull/3265

Change-Id: Ib2635dc9b8364b92caa768dc8e4de0603cbcf14f

### DIFF
--- a/assembly/assembly-main/src/assembly/bin/docker.sh
+++ b/assembly/assembly-main/src/assembly/bin/docker.sh
@@ -121,10 +121,6 @@ init() {
     export CHE_WORKSPACE_STORAGE_CREATE_FOLDERS=false
   fi
 
-  # Ensure that the user "user" has permissions for CHE_HOME and CHE_DATA
-  sudo chown -R user:user ${CHE_HOME}
-  sudo chown -R user:user ${CHE_DATA}
-
   # Move files from /lib to /lib-copy.  This puts files onto the host.
   rm -rf ${CHE_DATA}/lib/*
   mkdir -p ${CHE_DATA}/lib  
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,18 +34,9 @@ RUN echo "http://dl-4.alpinelinux.org/alpine/edge/community" >> /etc/apk/reposit
     apk add --update ca-certificates curl openssl openjdk8 sudo bash && \
     curl -sSL "https://${DOCKER_BUCKET}/builds/Linux/x86_64/docker-${DOCKER_VERSION}" -o /usr/bin/docker && \
     chmod +x /usr/bin/docker && \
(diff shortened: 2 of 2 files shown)


skabashnyuk:
What I need to rebuild to test?

benoitf:
@skabashnyuk here are the steps:

```
$ git pull branch CHE-3276
$ cd assembly/assembly-main
$ mvn clean install

# go back to root folder
$ cd ../../
$ docker build -t eclipse/che-server:nightly .

```
and then launch cli

TylerJewell (approved):

l0rd (approved):

assembly/assembly-main/src/assembly/bin/docker.sh
```diff
@@ -121,10 +121,6 @@ init() {
     export CHE_WORKSPACE_STORAGE_CREATE_FOLDERS=false
   fi
 
-  # Ensure that the user "user" has permissions for CHE_HOME and CHE_DATA
-  sudo chown -R user:user ${CHE_HOME}
-  sudo chown -R user:user ${CHE_DATA}
```
l0rd, line 126: @benoitf I think that these may be needed for the workspaces (that are still run as user `user`)

benoitf: @l0rd issue is that user 'user' may have another uid so it doesn't work as well.

benoitf: @l0rd also I think workspaces only need access to /projects which is done there
https://github.com/eclipse/che/blob/5fbaf3d9ac45344b5596742b27d9335652d3d0ca/wsmaster/che-core-api-agent/src/main/resources/agents/scripts/org.eclipse.che.ws-agent.script.sh#L38

TylerJewell:
What about agents and other software that you want to install outside of /projects?

assembly/assembly-main/src/assembly/bin/docker.sh
l0rd: @benoitf that makes sense 👍

benoitf:
agents need to have sudo access so they have root access to do what they want. ?

TylerJewell:
yep - got it.

benoitf:
@skabashnyuk do you have tested ?

riuvshin (approved):

skabashnyuk (approved):

PR #3300, opened 2016-12-06
CHE-53: Adding liveness probe for non-dev machines

ibuziuk:
### What does this PR do?
Adds OpenShift Liveness probe for non-dev machines, if terminal agent is injected 

https://issues.jboss.org/browse/CHE-53
### Previous behavior
Liveness probe was not added for non-dev machines

### New behavior
Liveness probe was not added for non-dev machines, if terminal agent is injected


ibuziuk:
@l0rd could you please review ?

l0rd (approved):

PR #3301, opened 2016-12-06
Che CLI Improvements

TylerJewell:
### What does this PR do?
1. Adds in error message to check for missing "-it" on the command line.

2. Adds in better, friendlier help when missing different volume mounts.

3. Changes the ":/che" mount to be ":/data" to be consistent for different assemblies.  Replaced a hard coded "/che" reference in the docker-compose template, and instead passed in the CHE_CONTAINER_ROOT into puppet to be used.

4. Adds a new cleanup() method to catch any exit scenario. If the product exits for any reason with return code "3", we will print out a message to the user to tell them how they can rerun to get the cli.log.

5. If in development mode, adds a check to pull the CHE_DEBUG_PORT from the user's che.env file and then performs a port check on that during the pre-flight checks to verify that it is available.  We will need to create a matching improvement for Codenvy assembly to solve this issue.
https://github.com/codenvy/codenvy/issues/1282

6. Improved the usage output format. Also moved usage() so that it can appear before any docker checks.

7. Removed the need to have ':/cli' volume mount. We now just write the cli.log every time to the ':/data' folder.

8. Adds a new check that if the system is not in --offline mode, then verifies that the DNS resolver is working to dockerhub.com. If not, prints out an error message with users on how to fix it with their Docker settings.

To test this version of the CLI:
1. git clone http://github.com/eclipse/che
2. git checkout cli-improvements
3. cd dockerfiles/base;  bash build.sh
4. cd dockerfiles/cli; bash build.sh

The syntax has slightly changed, so please check the /docs/README.md in this branch for the updates.


benoitf (approved):

dockerfiles/base/scripts/base/startup.sh
```diff
@@ -251,20 +255,40 @@ init() {
   source "${SCRIPTS_CONTAINER_SOURCE_DIR}"/cli.sh
 }
 
+cleanup() {
+  RETURN_CODE=$?
+
+  # CLI developers should only return '3' in code after the init() method has completed.
+  # This will check to see if the CLI directory is not mounted and only offer the error
+  # message if it isn't currently mounted.
+  if [ $RETURN_CODE -eq "3" ]; then
```
garagatyi, line 269: As far as I understand we should check that no code is using exit or return with exit code that differs from 3 in that case. Am I right? Have you grepped CLI sources such a code?

dockerfiles/cli/scripts/entrypoint.sh
```diff
@@ -8,48 +8,44 @@
...
-    CHE_HOST                             IP address or hostname where ${CHE_FORMAL_PRODUCT_NAME} will serve its users
-    CLI_DEBUG                            Default=false. Prints stack trace during execution
-    CLI_INFO                             Default=true. Prints out INFO messages to standard out
-    CLI_WARN                             Default=true. Prints WARN messages to standard out
-    CLI_LOG                              Default=true. Prints messages to cli.log file
+OPTIONAL DOCKER PARAMETERS:
+  -e CHE_HOST=<YOUR_HOST>              IP address or hostname where ${CHE_FORMAL_PRODUCT_NAME} will serve its users
+  -e CHE_PORT=<YOUR_PORT>              Port where ${CHE_FORMAL_PRODUCT_NAME} will bind itself to
+  -v <LOCAL_PATH>:/data/instance       Where instance, user, log data will be saved
+  -v <LOCAL_PATH>:/data/backup         Where backup files will be saved
```
benoitf: /data/ should be replaced by ${CHE_CONTAINER_ROOT} (to previous lines)

docs/README.md
```diff
@@ -154,7 +158,7 @@ While connected to the Internet, download Che's Docker images:
 ```
 docker run eclipse/che-cli offline
 ``` 
-The CLI will download images and save them to `/che/backup/*.tar` with each image saved as its own file. The `/backup` folder will be created as a subdirectory of the folder you volume mounted to `:/che`. You can optionally save these files to a differnet location by volume mounting that folder to `:/backup`. The version tag of the CLI Docker image will be used to determine which versions of dependent images to download. There is about 1GB of data that will be saved.
+The CLI will download images and save them to `/che/backup/*.tar` with each image saved as its own file. The `/backup` folder will be created as a subdirectory of the folder you volume mounted to `:/data`. You can optionally save these files to a differnet location by volume mounting that folder to `:/backup`. The version tag of the CLI Docker image will be used to determine which versions of dependent images to download. There is about 1GB of data that will be saved.
```
benoitf: /che/backup/*.tar <-- /data ?

docs/README.md
```diff
@@ -178,58 +182,51 @@ When invoked with the `--offline` parameter, the CLI performs a preboot sequence
...
-    CHE_HOST                             IP address or hostname where ${CHE_MINI_PRODUCT_NAME} will serve its users
-    CLI_DEBUG                            Default=false.Prints stack trace during execution
-    CLI_INFO                             Default=true. Prints out INFO messages to standard out
-    CLI_WARN                             Default=true. Prints WARN messages to standard out
-    CLI_LOG                              Default=true. Prints messages to cli.log file
+USAGE: 
+  docker run <DOCKER_PARAMETERS> ${CHE_IMAGE_FULLNAME} [COMMAND]
+
+MANDATORY DOCKER PARAMETERS:
+  -v <LOCAL_PATH>:${CHE_CONTAINER_ROOT}         Where user, instance, and log data saved
```
benoitf, line 189: I'm not sure all the variables ${CHE_CONTAINER_ROOT}, ${CHE_FORMAL_PRODUCT_NAME} will work there ?

docs/README.md
```diff
@@ -255,15 +252,15 @@ INFO: (che-cli start): API: http://10.0.75.2/swagger
 While we provide `nightly`, `latest`, and `5.0.0-latest` [redirection versions](#nightly-and-latest) which are tags that simplify helping you retrieve a certain build, you should always run Cche with a specific version label to avoid [redirection caching issues](#nightly-and-latest). So, running `docker run eclipse/che-cli` is great syntax for testing and getting started quickly, you should always run `docker run eclipse/che-cli:<version>` for production usage.
 
 #### Volume Mounts
-If you volume mount a single local folder to `<your-local-path>:/che`, then Che creates `/che/che.env` (configuration file), `/che/instance` (user data, projects, runtime logs, and database folder), and `/che/backup` (data backup folder).
+If you volume mount a single local folder to `<your-local-path>:/data`, then Che creates `/che/che.env` (configuration file), `/che/instance` (user data, projects, runtime logs, and database folder), and `/che/backup` (data backup folder).
```
benoitf: /data/che.env, /data/instance /data/backup

riuvshin:
Do we need those changes to be included in M8 ?

dockerfiles/base/scripts/base/startup.sh
TylerJewell: In the current CLI implementation, there are no methods that return a '3'. So, this is laying a groundwork so that as we improve the CLI over time, if there are CLI crashes or hard error messages that we print out, we should return this error code to trigger this help message.

TylerJewell: The other thing we could do, @riuvshin - is not even bother with a :/cli volume mount. Instead, we could always have the cli.log written to :/data volume mount which is the root mount that is always required.

garagatyi: +1

dockerfiles/cli/scripts/entrypoint.sh
TylerJewell: Fixed

docs/README.md
TylerJewell: Yes, fixed.

docs/README.md
TylerJewell: They do work there.  Tested.

docs/README.md
TylerJewell: Found about 10 of these to fix, thanks.

dockerfiles/base/scripts/base/startup.sh
TylerJewell: Fixed.

dockerfiles/base/scripts/base/cli/cli-functions.sh
```diff
@@ -188,6 +185,25 @@ grab_offline_images(){
...
     done
+  else
+    # If we are here, then we want to run in networking mode.
+    # If we are in networking mode, we have had some issues where users have failed DNS networking.
+    # See: https://github.com/eclipse/che/issues/3266#issuecomment-265464165
+    local HTTP_STATUS_CODE=$(curl -I -k dockerhub.com -s -o /dev/null --write-out "%{http_code}")
+    if [[ ! $HTTP_STATUS_CODE -eq "301" ]]; then
+      info "Welcome to $CHE_FORMAL_PRODUCT_NAME!"
+      info ""
+      info "We could resolve DockerHub using DNS."
```
benoitf: Should the word 'not' be placed after could ?

dockerfiles/base/scripts/base/cli/cli-functions.sh
```diff
@@ -188,6 +185,27 @@ grab_offline_images(){
       fi
       info "init" "Loading ${file##*/}..."
     done
+  else
+    # If we are here, then we want to run in networking mode.
+    # If we are in networking mode, we have had some issues where users have failed DNS networking.
+    # See: https://github.com/eclipse/che/issues/3266#issuecomment-265464165
+    local HTTP_STATUS_CODE=$(curl -I -k dockerhub.com -s -o /dev/null --write-out "%{http_code}")
```
mshaposhnik, line 192: This check doesn't work behind the proxy. Need to export HTTP_PROXY variables or smthn like that.
And same for [this](https://github.com/eclipse/che/pull/3301/files#diff-3d98ff722a023cb7673112d6d34d43a9R438) check.

docs/README.md
TylerJewell: Yep - will make a lot of docs improvements in another branch as we will be restruturing it to support jekyll.

dockerfiles/base/scripts/base/cli/cli-functions.sh
TylerJewell: Fixed.

PR #3315, opened 2016-12-07
set nightly versions for M9-SN iteration

riuvshin:
this need to make M9-SNAPSHOT available as nightly tag

@benoitf @skabashnyuk 

later this will be automated during release process

### DIFF
--- a/dockerfiles/action/Dockerfile
+++ b/dockerfiles/action/Dockerfile
@@ -10,6 +10,6 @@
 # use:
 #    docker run -v /var/run/docker.sock:/var/run/docker.sock eclipse/che-action [command]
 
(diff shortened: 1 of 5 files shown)


benoitf:
https://github.com/eclipse/che/blob/master/dockerfiles/lib/dto-pom.xml#L20

and

https://github.com/eclipse/che/blob/master/dockerfiles/lib/dto-pom.xml#L26

should be set to M9-SNAPSHOT as well

riuvshin:
@benoitf good catch

benoitf (approved):

skabashnyuk (approved):

Issue #2977, opened 2016-11-03
Eclipse Che CLI Refactoring + Improvements


TylerJewell:
This epic covers a number of improvements to refactor the Che CLI to:
1. Allow for offline installation mode.
2. Eliminate duplicate code between che-server, che-launcher, che.sh, CLI.
3. Simplify the CI management of dependent containers.
4. Allow for "custom cli assemblies" so products like OpenShift and Codenvy can reuse + extend CLI.
5. Move utility containers from eclipse/che-dockerfiles to eclipse/che repository.
6. Port the ARTIK IDE to make use of the improved CLI.
7. Add configuration management to Che installations
8. Support docker-compose as an orchestrator.
9. Provide multipe orchestrator launch mechanisms with a single configuration.
10. Simplify configuration of Che into a set of environment variables.
11. Provide backup / restore functionality for Che.

The work for this epic is in the 'cli-refactoring' branch.  The order of the specification below indicates the order of work to be done.

#### CLI Consolidation
We want to consolidate the CLIs that exist between Che and Codenvy, and move all of the meaningful functionality into Che.  Most of this work has already been captured in the `cli-refactoring` branch.  The key elements of this work are:
- [x] Support 'che init', "che download', 'che config', 'che start', 'che stop', 'che restart', 'che destroy', 'che offline'.
- [x] Configure the che-server docker container and the docker-compose script template that is provided to have all che-server logs and machine logs written to the `/instance/config` folder. This would be identical to the way that Codenvy does it, and it is ensuring that the folder mounts of the various orchestrators are done in a proper way.
- [x] Add support for 'che backup` and `che restore`, to match the cross-platform backup and recovery capabilities that are now reflected in Codenvy.
- [x] Migrate existing CLI function for 'che compile', 'che mount', 'che dir', 'che ssh', 'che test" into the new CLI.  We will not be migrating "che profile" into the new CLI.  That will die.
- [x] Perform another CLI refactoring to create updated versions of eclipse/che-launcher and codenvy/che-launcher, which provide the important functions of `start()`, `restart()`, `stop()`, `config()`, `init()`, `version()`, `backup()`, `restore()`, `upgrade()`, `offline()`, `download()`.  This will move the heft of the CLI into a single container, and then make the CLI an optional wrapper.  This also has an affect of no longer requiring docker compose 1.8 on the host as we can provide that client in the container.  Specification details are at the bottom.

- [x] Need to test: workspace logs written to disk, ssh, sync, dir commands.

[...]

PR #3318, opened 2016-12-07
CHE-2977 "move the core Che Dockerfiles and Dockerfile.centos into /dockerfiles/che

benoitf:
### What does this PR do?
Move Dockerfiles to dockerfiles directory
merge docker.sh and che.sh --> entrypoint.sh

### Previous behavior
Files were at root folder and we had docker.sh and che.sh

### New behavior
Files are in dockerfiles folder
also docker.sh script is moved from the assembly folder to the docker image folder and named entrypoint.sh

Change-Id: I7404f8f2896e8a7eea83ecab57a6c4c623c68abe


benoitf:
@TylerJewell 
@riuvshin it will impact CI jobs of building the image (it also bring a build.sh script -> requires to have called mvn clean install in assembly/assembly-main)
@dharmit / @l0rd it should have impact on CI job of centos as Dockerfile is moved to /dockerfiles subdirectory

TylerJewell (approved):

riuvshin (approved):

dockerfiles/che/Dockerfile.centos
```diff
@@ -46,7 +46,7 @@ EXPOSE 8000 8080
 
 USER user
 
-ADD assembly/assembly-main/target/eclipse-che-*/eclipse-che-* /home/user/che/
+ADD assembly /home/user/che
 
 ENTRYPOINT [ "/home/user/che/bin/che.sh", "-c" ]
```
l0rd: entrypoint should be updated here too

dockerfiles/che/build.sh
```diff
@@ -0,0 +1,31 @@
...
+ASSEMBLY_DIR=$(echo ${DIR}/../../assembly/assembly-main/target/eclipse-che-*/eclipse-che-*)
+
+# Remove current copy of the assembly if present
+if [[ -d "${DIR}/assembly" ]]; then
+  echo "Remove previous assembly folder"
+  rm -rf ${DIR}/assembly
+fi
+
+# Copy assembly
+cp -r ${ASSEMBLY_DIR} ${DIR}/assembly
```
l0rd: Instead of copying the assembly each time (and Docker will copy it one more time) I would rather:
```bash
docker build -t eclipse/che-server -f dockerfiles/che/Dockerfile .
```
from the repository root

dockerfiles/che/Dockerfile
```diff
@@ -9,7 +9,7 @@
 #   Codenvy S.A
 #
 # To build it, run in the repository root:
-#  `docker build -t codenvy/che-server .`
+#  `docker build -t eclipse/che-server .`
```
l0rd, line 12: This doesn't work anymore. I would rather:
```
docker build -t eclipse/che-server -f dockerfiles/che/Dockerfile .
```
from the root of the repo. 
Dockerfile could copy assembly from `assembly/assembly-main/target/eclipse-che-*/eclipse-che-*` as before.

l0rd (changes requested):
Thank you for the heads up on that @benoitf. We will handle the necessary modifications for centOS containers pipeline. You can move the Dockerfiles in the meantime.

dockerfiles/che/Dockerfile
benoitf: the issue #2977 is to not build anymore the docker images from the root but from each folder
the build.sh script is handling that. Maybe I should drop the build instruction from the dockerfile by using docker build but only using build.sh

dockerfiles/che/Dockerfile.centos
benoitf: why should I update the entrypoint ? the current dockerfile is not using docker.sh script (which is now entrypoint.sh) so maybe there were some reasons ?

dockerfiles/che/build.sh
benoitf: the issue #2977 is to not build anymore the docker images from the root but from each folder

benoitf: As a later improvement I would like to copy assembly only if folder has been updated

BTW relaunching twice the build.sh command if assembly has not been changed is reusing cache layer

l0rd:
Building context is always sent (i.e. assembly is always copied)

l0rd:
I think the reason is that Dockerfile.centos hasn't been updated when the
entrypoint of the default Dockerfile has been updated.

dockerfiles/che/build.sh
l0rd: Building context is always sent (i.e. assembly is always copied)

l0rd:
@benoitf @TylerJewell don't get me wrong, I think that moving dockerfiles in a specific folder is a great choice that will make the folder organization a lot cleaner. However I don't understand why you want to

> not build anymore the docker images from the root but from each folder

I see two drawbacks:
1. The Dockerfile build would be a two step operation (copy + docker build) instead of one. It takes longer, there are more chances to make a mistake, it's not the standard way to go and it's more complicated.
2. The copy of the assembly is usually the most expensive operation in the che-server Dockerfile build. We would do the most expensive operation of the build one more time with no particular benefit.

And I don't see any benefit compared to one build command from the root folder:

```bash
docker build -t eclipse/che-server -f dockerfiles/che/Dockerfile .
```

dockerfiles/che/Dockerfile.centos
benoitf: ok, I wanted to keep Dockerfile.centos as it was but if you say it's a mistake, then I will add usage of entrypoint.sh

TylerJewell:
@l0rd - this just comes down to having a more consistent framework for building.  We have done a lot of effort so that every /dockerfiles sub-folder has a consistent build experience on mac, windows, or linux.  

And so it just comes down to making the build experience better, and then a side benefit is that the root of the repository is cleaner.

benoitf:
@l0rd @TylerJewell 
what if I merge assembly/assembly-main into dockerfiles/che
building che will be done in a single self-contained folder and running che-server will be always done with docker so it should match both of the expectations:
- self-contained folder
- no recopy of files/folder as assembly and docker images are produced in the same root folder

and we've the same lifecycle : building the assembly is building the image as well
so we're sure we're always in sync locally and don't forget to build image after building assembly

TylerJewell:
1. As we discussed in slack a little bit, I think we have to have an /assembly/assembly-main/target output. Even if this is unrunnable, it is still a packaged assembly that is docker-ready.
2. We can then optionally push that assembly into /dockerfiles/che, or we can have the build script in /dockerfiles/che check & pull a copy from the assembly location.
3. Finally, separately, we could consider adding a flag for the mvn install phase that will have the assembly/assembly-main also build the Docker image from within the assembly, so that it is built, though I would not know how to do that with the Dockerfile for this image located in another location.

l0rd:
@benoitf @TylerJewell that's a proposal to improve speed of Docker build without sacrificing consistency and neatness 😄 

https://github.com/eclipse/che/pull/3336

benoitf:
I've rewritten all the pull request to merge che.sh and docker.sh into entrypoint.sh
it also solve the issue https://github.com/eclipse/che/issues/3324

benoitf:
ok maybe if os is supporting it, I could use hard link of zip assembly and add zip instead of directory in Dockerfile
then there will be no copy in that case
It will require some changes to rename directory or have an alias but it may work as well

l0rd:
Hard link to the .zip file should be faster than the copy of the folder. And would take less space on disc too. I'm not sure that's easy to do on Windows tough.

skabashnyuk (changes requested):
we not ready to this: Only way to start Che is through che-server docker image

TylerJewell:
@skabashnyuk - you cannot reject a PR without providing a list of actions that need to be taken.  Please itemize whether any additional testing is needed.

skabashnyuk:
Split this PR on two PRs. First fix some bugs,  second remove ability to run native tomcat.

TylerJewell:
I would prefer not to do that. The removing of Tomcat is a clean way to solve the bugs.  Please itemize anything in CI or QA systems we should build to be ready to remove Tomcat as a native approach.

skabashnyuk:
1. Dev team have to say that it is able to do all regular development that it can do with current tomcat
2. We need to have at least a couple of weeks without major outages with the approach how we run che and codenvy with docker.

TylerJewell:
Agree on #1.  We will not wait for #2 - that is not a goal that I have.

benoitf:
@l0rd I've updated to use hard link of files instead of a copy
I've also tested hard link on a Windows 10 VM and with git bash it works on my side.

It also now using the path of the internal archive (then directory /home/user/che is changed to /home/user/eclipse-che-5.0.0-M9-SNAPSHOT) as it is using the .tgz file for the hard link

l0rd (approved):

l0rd:
@benoitf I approved your PR and will close #3336. I still believe that using the repository root as the build context is a better solution than using an hard link. But I understand that you want to use the same convention for every image (build context is dockerfile folder) and I must admit that I'm the only one defending my position so I won't block on that.

vparfonov:
Could not use own Che build 

`docker run -it --rm -v /var/run/docker.sock:/var/run/docker.sock -v /home/vetal/che:/data -v /home/vetal/codenvy/projects/che:/repo eclipse/che-cli:nightly start`

```
WARNING: No swap limit support
WARN: Your local eclipse/che-cli:nightly image is older than the version on DockerHub.
WARN: Run 'docker pull eclipse/che-cli:nightly' to update your CLI.
INFO: (che cli): Checking registry for version 'nightly' images
Notice: Compiled catalog for f30f78d9e0ce.ua.codenvy-dev.com in environment production in 0.11 seconds
Notice: Applied catalog in 0.04 seconds
INFO: (che config): Customizing docker-compose for running in a container
        port 8080 (http):       [AVAILABLE] 
        port 8000 (debug):       [AVAILABLE]

INFO: (che start): Starting containers...
Couldn't find env file: /che/instance/config/che.env
```

benoitf:
@skabashnyuk I've updated PR to not remove native scripts
it will be in a separate pull request

skabashnyuk (approved):

PR #3327, opened 2016-12-08
Improve container label handling (replaceLabels() method)

amisevsk:
### What does this PR do?
Improves OpenShiftConnector#createContainer() and OpenShiftConnector#inspectContainer() to properly store and retrieve labels from CreateContainerParams.

Details:
1. Kubernetes annotations are can consist only of alphanumeric characters and `'-'`, `'_'`, `'.'`. This means that the labels provided have to be converted into valid Kubernetes annotations. This is currently done by converting `':' -> '.'` and `'/' -> '_'`, which means that if a label contains `'.'` or `'_'`, it cannot be converted and is skipped. Additionally, padding has to be added to keys and values since they must start and end with an alphanumeric character (and sometimes we encounter e.g. "/api").
2. Considered attaching the annotations to the Pod instead of the Service, but AFAICT these annotations are not persisted if the pod is recreated by OpenShift. An alternative would be to assign the label to the Deployment Config used to create the pod. Thoughts/suggestions here are welcome.

### Previous behavior
Labels returned for inspectContainer() were hard-coded.

### New behavior
Labels are stored as annotations on the Service assigned to the Pod, and can be retrieved.

Signed-off-by: Angel Misevski <amisevsk@redhat.com>

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/OpenShiftConnector.java
@@ -46,6 +46,9 @@ import javax.inject.Named;
 import javax.inject.Singleton;
 import java.io.IOException;
 import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
@@ -62,6 +65,13 @@ public class OpenShiftConnector {
     private static final String OPENSHIFT_API_VERSION = "v1";
     private static final String CHE_WORKSPACE_ID_ENV_VAR = "CHE_WORKSPACE_ID";
     private static final String CHE_OPENSHIFT_RESOURCES_PREFIX = "che-ws-";
+
+    /** Prefix used for che server labels */
+    protected static final String CHE_SERVER_LABEL_PREFIX  = "che:server";
+    /** Padding to use when converting server label to DNS name */
+    protected static final String CHE_SERVER_LABEL_PADDING = "0%s0";
+    /** Regex to use when matching converted labels -- should match {@link CHE_SERVER_LABEL_PADDING} */
+    protected static final Pattern CHE_SERVER_LABEL_KEY    = Pattern.compile("^0(.*)0$");
 
     private static final String OPENSHIFT_SERVICE_TYPE_NODE_PORT = "NodePort";
     private static final int OPENSHIFT_LIVENESS_PROBE_DELAY = 120;
@@ -130,7 +140,9 @@ public class OpenShiftConnector {
         String[] volumes = createContainerParams.getContainerConfig().getHostConfig().getBinds();
 
         IProject cheProject = getCheProject();
-        createOpenShiftService(cheProject, workspaceID, exposedPorts);
+
+        Map<String, String> additionalLabels = createContainerParams.getContainerConfig().getLabels();
+        createOpenShiftService(cheProject, workspaceID, exposedPorts, additionalLabels);
         String deploymentConfigName = createOpenShiftDeploymentConfig(cheProject,
                                                                       workspaceID,
                                                                       imageName,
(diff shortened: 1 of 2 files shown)


plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/OpenShiftConnector.java
```diff
@@ -585,14 +594,90 @@ private boolean isTerminalAgentInjected(final Set<String> exposedPorts) {
...
+        Map<String, String> names = new HashMap<>();
+        for (Map.Entry<String, String> entry : labels.entrySet()) {
+            String key = entry.getKey();
+            String value = entry.getValue();
+
+            if (value == null) {
+                continue;
+            }
+            // Check for potential problems with conversion
+            if (!key.startsWith(CHE_OPENSHIFT_RESOURCES_PREFIX)) {
```
l0rd: s/CHE_OPENSHIFT_RESOURCES_PREFIX/CHE_SERVER_LABEL_PREFIX ?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/OpenShiftConnector.java
```diff
@@ -585,14 +594,90 @@ private boolean isTerminalAgentInjected(final Set<String> exposedPorts) {
...
+                LOG.warn("Expected CreateContainerParams label key " + entry.getKey() +
+                         " to start with " + CHE_SERVER_LABEL_PREFIX);
+            } else if (key.contains(".") || key.contains("_") || value.contains("_")) {
+                LOG.error(String.format("Cannot convert label %s to DNS Name: contains '-' or '.'", entry.toString()));
```
l0rd: I would be more explicit in the log here: "Cannot convert label .... because these are used as escape characters."

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/OpenShiftConnector.java
```diff
@@ -585,14 +594,90 @@ private boolean isTerminalAgentInjected(final Set<String> exposedPorts) {
...
+    protected Map<String, String> convertLabelsToKubernetesNames(Map<String, String> labels) {
+        Map<String, String> names = new HashMap<>();
+        for (Map.Entry<String, String> entry : labels.entrySet()) {
+            String key = entry.getKey();
+            String value = entry.getValue();
+
+            if (value == null) {
+                continue;
+            }
+            // Check for potential problems with conversion
```
l0rd, line 637: We are just converting characters ':' and '/'. What happens if another char, let's say '?', is present in one of the labels? Shouldn't we check that key and label match regex ([-A-Za-z0-9_.:/]*) too?

l0rd (changes requested):
Great PR @amisevsk! Label conversion was the trickiest part I think. I have added some comment/questions. I will merge the PR as soon as you answer to these.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/OpenShiftConnector.java
amisevsk: Silly oversight on my part; fixed.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/OpenShiftConnector.java
amisevsk: That is a good idea. I've added a check against that regex before putting a label into the map and improved some of the documentation surrounding the process.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/OpenShiftConnector.java
```diff
@@ -585,14 +594,90 @@ private boolean isTerminalAgentInjected(final Set<String> exposedPorts) {
...
+                LOG.warn("Expected CreateContainerParams label key " + entry.getKey() +
+                         " to start with " + CHE_SERVER_LABEL_PREFIX);
+            } else if (key.contains(".") || key.contains("_") || value.contains("_")) {
+                LOG.error(String.format("Cannot convert label %s to DNS Name: contains '-' or '.'", entry.toString()));
+                continue;
+            }
+
+            // Convert keys: e.g. "che:server:4401/tcp:ref" -> "che.server.4401-tcp.ref"
+            key = key.replaceAll(":", ".").replaceAll("/", "_");
+            // Convert values: e.g. "/api" -> ".api" -- note values may include '-'
```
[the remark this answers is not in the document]
amisevsk: Values with `':'` will currently be skipped. This method is written for the current labels produced by Che -- values are protocol (e.g. http), ref (e.g. tomcat8-debug), and path (e.g. /api). I've written it to deal with those cases specifically, since if this changes (e.g. labels start to include other special characters), chances are more in-depth changes will be necessary.

This is kind of tricky -- the conversion methods are written for the specific structure of Che labels. Values don't currently seem to contain `':'`, so while the method could be written to handle that case, there is no reason to assume that that is the special character that *will* be used. Basically, I don't think we should use up one of the allowed special characters for an unnecessary case. 

Example: We replace `':'` and `'/'` now, but the replacement of `':'` is not necessary. Later, Che adds labels that contain `'*'` and `'#'`. Since we are already using `'.'` and `'_'` as replacement chars, it looks like we can't support the new labels. The person fixing the issue would have to look at labels to see that values don't contain `':'`. 

Let me know if this makes sense; I'm not terribly happy with how fragile these methods are, so I'm open to suggestions.

l0rd (approved):

PR #3336, opened 2016-12-09
CHE-2977 Changes to allow build from root of repository

l0rd:
New build of che server take ~10 seconds more on my machine (c.f. video)

That's annoying because there are days when I build che a looooooooot of times 😃 

This is a proposal to improve the speed without sacrificing the coherence of new Dockerfiles organization.

[asciicast](https://asciinema.org/a/avsn02hsb70gsz6ovvudrp315)


benoitf:
the issue is that we still need to execute at the root folder
and you're adding tons of complex ignore patterns in root .dockerignore

benoitf:
> That's annoying because there are days when I build che a looooooooot of times 😃

BTW building che is not forcing to rebuild the che docker image
In dev mode it is using the .tgz directly
so you even don't need to build che docker image at all

TylerJewell:
I think our intention is that dev mode is supposed to solve your issue, and that we shouldn't have to restructure the build structure.  If you volume mount your Che repo to :/repo with the new CLI, what is the performance of that in between builds?  Shouldn't have to rebuild the image at all.

l0rd:
@benoitf 

> we still need to execute at the root folder

False. Please have a look the video

> you're adding tons of complex ignore patterns in root .dockerignore

I added 20 characters. That's 1 line, 1 ignore pattern. Do you really consider that "tons of complex ignore patterns"? 

> building Che is not forcing to rebuild the che docker image

I mostly work with OpenShift and I need to rebuild the image.

l0rd:
@TylerJewell 

> we shouldn't have to restructure the build structure

I've just changed one line of a script to avoid copying 250MB. Do you consider that "restructuring the build structure"? I think there is a misunderstanding here. 

> Shouldn't have to rebuild the image at all

We can't use the new CLI to deploy on OpenShift. We need to build the image. But that's not the point. The point is that copying the assembly is useless. It can be avoided and we can preserve the new build structure too! What's the point to continue copying these 250MB?

benoitf:
> @benoitf
> we still need to execute at the root folder
> False. Please have a look the video
> 

it is, you build dockerfiles from root folder and sending build context which is the root folder
it is only playing on the word "root"


> you're adding tons of complex ignore patterns in root .dockerignore
> I added 20 characters. That's 1 line, 1 ignore pattern. Do you really consider that "tons of complex ignore patterns"?
> 

we do not need anymore this dockerignore file (which include tons of patterns) as we're using a local build context from the directory where Dockerfile is

>building Che is not forcing to rebuild the che docker image
>I mostly work with OpenShift and I need to rebuild the image.

I think the workflow is still taking too long if you need to rebuild each time as well the full binary + image each time

l0rd:
> it is, you build dockerfiles from root folder and sending build context which is the root folder
it is only playing on the word "root"

Ok so we agree script ./build.sh is executed from folder `dockerfiles/che`.
And yes `docker build` is executed from root folder. If needed we can run it from dockerfiles/che too:
```bash
ROOT=$(cd ${DIR}../../; pwd)
docker build -t eclipse/che -f dockerfiles/che/Dockerfile ${ROOT}
```
 But the result is the same. What's the point of changing that?

benoitf:
@l0rd let say we need another dockerfile for building a che image from another folder of che

for example we want to package dashboard into an image
so by keeping "build with root context of /"
we will have a build context with assembly-main + we will have to enable dashboard folder in .dockerignore

so : building dashboard would include a build context with assembly-main + dashboard
and building che image will have the same context (then why do we will need dashboard files in context when building che image ?)

I assume we should not need in the build context the binaries that are not linked to the image we want to build

let's image we build image for the workspace agent, exec agent by using always the root path
it will increase the size of the root build context as well and may not be expected

l0rd:
> I assume we should not need in the build context the binaries that are not linked to the image we want to build

Agree

> so by keeping "build with root context of /"

The PR modifies eclipse/che Dockerfile build only. Other Dockerfiles are still built from their folder, not root folder. In `build.include`:

```bash
BUILD_CONTEXT=${BUILD_CONTEXT:-$DIR}
```

If not specified, $BUILD_CONTEXT defaults to Dockerfile's folder.

benoitf:
yes but idea of issue is to bring consistency 
if one docker image should be build with the root context and the other from their own directories it is not consistent at all.
why .dockerignore file at the root folder should be only for che image and why this one should be different from the others

l0rd:
> it is not consistent at all

Don't you think that the copy of the assembly is a break to the consistency too? Others `build.sh` don't do that copy right? And that inconsistency costs 10s more. 
 
By the way I still consider it consistent:
- all dockerfiles are in `dockerfiles` folder
- all images are built using the `build.sh` script
- `build.include` is used to build all Dockerfiles
- build context is configurable for all Dockerfiles

Another benefit is that the build context is configurable for all the Dockerfiles, not just `eclipse/che`. That's cool because we can avoid duplicating code sharing the build context amongst different images and reusing the same bash scripts.

TylerJewell:
@l0rd @skabashnyuk @benoitf - well, I think we have the basic principle down:
1. Each Docker image, should have a single folder with assets.
2. That folder should have the self-contained context of what is built with those assets.  For example, the /init/ image has the appropriate template files as sub-folders.
3. If we were to follow this strategy, then the che-server image should also have its associated dependencies.  One idea that Florent had was that the /dockerfiles/che/ folder should have all of its dependnecies.  So, we could either copy it from another folder, or we could have this folder be the master assembly.  We wouldn't need to call it "assembly-main". Instead, we would just have a `/dockerfiles/che/assembly/pom.xml which would have the assembly files necessary to build the Che assembly.  When you build the image, it would first check to see if the assembly is built, and if not, build that. You could optionally build the assembly without building the image.

If we did this, then the /assembly folder in the root, really wouldn't be for the assembly. It would be for building libraries and packages which are needed for the assembly to be built.  In essence, this would be renamed as `/lib`, because that is what is in there - libraries and packages built from the code.  

This is probably cleaner, but causes me a huge "ugh" because we have dozens if not 100 pages of documentation in our wiki and in the docs that talk about development in /assembly/assembly-main/, and all of that would have to be updated to basically say:
1. Build code first in either /core, /plugins, /dashboard.
2. Package these code changes by rebuilding /lib - either ws-master or ws-agent.
3. Rebuild an assembly.
4. Run the assembly by either building a new Che docker image, or you can mount the source repo directly.

benoitf:
@l0rd I've made a fixup on pull request https://github.com/eclipse/che/pull/3318 to use hard link (ln) and then switch from directory copy to a file copy.
The build context is smaller and we do not recopy directories twice (as it was in the original pull request 3318) so I suppose we no longer need this one if you agree on using file links

benoitf:
it could be interesting if adding .tar.gz is faster or not than the copy of the directory

PR #3342, opened 2016-12-09
Add manually publishing cascade events in implementation of DAOs instead of using EntityListeners

sleshchenko:
### What does this PR do?
Add:
-  rethrowing original exception while cascade operation
-  manually publishing cascade events in implementation of DAOs instead of using EntityListeners.
-  flushing of data by EntityManager#flush to produce consistency exception in right place.
-  CascadeEventLister that is able to throw any of Exception. If listener throws exception then all data changes should be canceled.
-  base events(PersistEvent, RemoveEvent, UpdateEvent). CascadeEventService throws only right exceptions for each use case. When thrown by listener exception should not be thrown by contract of this operation, then it will be wrapped into ServerException.

https://github.com/eclipse/che/issues/3399

### Previous behavior
Post[Persist|Remove|Update] events are thrown EntityListener, and it is not allowed to change(store new, remove or update existing) any objects on these events.
User is not able to see original exception while cascade removing.

### New behavior
New approach allows us to perform any operations on post events or rollback transaction if is fails. 
So post events must be thrown by implementations of DAOs. It allows to perform any operation on event and rollback transaction if one of subscriber fails.
To fails original operation event subscriber should extend `CascadeEventSubscriber`


akorneta (approved):

voievodin (approved):

voievodin (changes requested):
This approach won't propagate the original exception

core/che-core-db/src/main/java/org/eclipse/che/core/db/cascade/CascadeContext.java
```diff
@@ -8,21 +8,24 @@
...
+ * Context that is used only for sharing the state
+ * of the cascading operation among subscribers.
  *
  * @author Anton Korneta
+ * @author Sergii Leshchenko
  */
-public class RemovalContext {
-    private Exception cause;
+public class CascadeContext {
+    private ApiException cause;
```
mshaposhnik: For me, REST-api oriented exception looks a bit strange here.

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/jpa/JpaFactoryDao.java
```diff
@@ -159,7 +161,9 @@ protected FactoryImpl doUpdate(FactoryImpl update) throws NotFoundException {
         if (update.getWorkspace() != null) {
             update.getWorkspace().getProjects().forEach(ProjectConfigImpl::prePersistAttributes);
         }
-        return manager.merge(update);
+        FactoryImpl merged = manager.merge(update);
+        manager.flush();
```
mshaposhnik, line 164: Is those changes are necessary if we didn't publish any events here?

core/che-core-db/src/main/java/org/eclipse/che/core/db/cascade/CascadeContext.java
sleshchenko: The same as throwing REST-api oriented exception by Dao. But we do it. :) And we save original REST-api oriented exception that is thrown by listener for rethrowing

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/jpa/JpaFactoryDao.java
sleshchenko: Yes, because this operation can be performed by listener. And then integrity exception will be catched in wrong place

mshaposhnik: I can hardly imagine the listener which will create factory in some case :)) But for purposes of uniformity it's ok.

core/che-core-db/src/main/java/org/eclipse/che/core/db/cascade/CascadeEventSubscriber.java
```diff
@@ -8,31 +8,41 @@
...
     }
 
     /**
-     * Receives notification about cascade removal event.
+     * Receives notification about cascade event.
+     *
+     * <p>If the method throws an exception it will be set to context
+     * to break event publishing and rethrow exception.
+     * Original exception can be rethrown to publisher
+     * or it can be wrapped in {@link SecurityException} instance.
```
voievodin: `SecurityException` ?

sleshchenko: Oooops :smile:

voievodin (approved):

akorneta (approved):

skabashnyuk (approved):

PR #3345, opened 2016-12-09
CLI Fixup

TylerJewell:
### What does this PR do?
1. Adds `che-host` to the `NO_PROXY` setting for workspaces if `NO_PROXY` is going to be set.
2. Replaces error for missing -it TTY to just warning, to support CI systems.
3. Fixes issue with version check. If the docker daemon was less than the 1.10 docker client embedded in the CLI, then the command would fail before getting to the version check. This uses a new tactic of testing the output of `docker ps` and if the error message has certain text, then we are sure that we have a version compatibility issue.
4. Added a fix where cli_pre_init was being called before the initial images had been downloaded. And this method depends upon one of the images, so the order was out of sequence. Also hides the docker pull output from the initial images.
5. Adds an improvement to the `cmd_offline()` method to allow for saving workspace images to the offline folder for loading when Che starts. If no parameters are provided, then all known workspace images will be downloaded and saved. You can run `--list` to get a list of valid images and `--image:<name>` to save only a single workspace stack image.
6. Fixes issues where `CHE_HOST` set on the command line was not properly written into `che.env` during init() phase. This was due to a duplicate pre_init method that was removed.

@mshaposhnik @riuvshin


dockerfiles/base/scripts/base/docker.sh
```diff
@@ -251,16 +246,11 @@ check_docker_networking() {
...
-  if [[ ! -t 1 ]]; then
-    info "Welcome to ${CHE_FORMAL_PRODUCT_NAME}!"
-    info ""
-    info "We did not detect a valid TTY."
-    info ""
-    info "TTY Syntax:"
-    info "    Add '-it' to your 'docker run' command."
-    return 2
+  if [ ! -t 1 ]; then
+    warning "Did not detect TTY - interactive mode disabled"
```
benoitf, line 252: it no return/exit code is returned and license is asked it will end without helping user no ?

TylerJewell: We only ask user for input during destroy and init phases.  What are you thinking of as a solution?

riuvshin (approved):

mshaposhnik (approved):
LGTM

dockerfiles/base/scripts/base/docker.sh
benoitf: image

my issue is that user may not understand or see the warning about interactive mode disabled because when he will start the cli he may think cli is behaving badly.

Maybe it's not about checking interactive mode but about detecting result of the read option when asking for the license

TylerJewell: @benoitf - we will address the TTY issue in another PR.

PR #3360, opened 2016-12-13
CLI Fixes - Proxy, Native Mode

TylerJewell:
### What does this PR do?
1. Adds --noproxy for curl() method since curl command does not inherit that from environment variables.
2. Fixes issue where there are two names required for the docker-compose.yml file needed by other CLIs.


mshaposhnik (approved):
Updated a bit, there was a problem with quotes around the arguments

mshaposhnik:
And BTW we need same method override for CODENVY like:
```
if [ ! -z ${CODENVY_HOST+x} ]; then
  	   NO_PROXY_CONFIG_FOR_CURL=("--noproxy" "${CODENVY_HOST} ")
  	 fi
```

PR #3364, opened 2016-12-13
Remove unnecessary -ti

ghost:
`-ti` is not required when running Puppet image. Removing it unblocks launching Che from an Electron client app

@riuvshin @benoitf please approve


benoitf:
@eivantsov what about https://github.com/eclipse/che/blob/acaf9ca8c4c89989c186829b51dffc6ee9b81b4e/dockerfiles/base/scripts/base/commands/cmd_config.sh#L99 ?

ghost:
Missed it. Fixed now

riuvshin (approved):

benoitf:
does it still display diff in the log when -t flag is not there ?

PR #3366, opened 2016-12-13
Machine.getOwner should provide userId and not namespace

### LINKED ISSUE #3042: machine.getOwner() should return userId and not namespace
When calling machine.getOwner() it returns a namespace while it should return userId
https://github.com/eclipse/che/blob/master/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java#L313-313

**Che version:**    5.0.0-M7-SNAPSHOT

**Additional information:**
* Problem can be reliably reproduced, doesn't happen randomly: [Yes]

benoitf:
Fix #3042

### Previous behavior
Machine.getOwner() was returning namespace

### New behavior
machine.getOwner() is returning userId and not a namespace

Change-Id: I64b4cd3931fec502073ead0115a312430f2434b8

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
@@ -305,6 +305,7 @@ public class MachineProviderImpl implements MachineInstanceProvider {
                                                       service.getId(),
                                                       workspaceId);
 
+            final String userId = EnvironmentContext.getCurrent().getSubject().getUserId();
             MachineImpl machine = new MachineImpl(MachineConfigImpl.builder()
                                                                    .setDev(isDev)
                                                                    .setName(machineName)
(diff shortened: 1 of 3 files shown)


sleshchenko (approved):

sleshchenko:
But please fix commit message, because as for me `Fix #3042` is bad commit message.

benoitf:
@sleshchenko 
commit message is

```
Fix #3042
machine.getOwner() should return userId and not a namespace
```


The first line Fix is for https://help.github.com/articles/closing-issues-via-commit-messages/

sleshchenko:
Yes, let me correct me. Please fix commit subject :smile: `Fix #3042` is [commit subject](http://chris.beams.io/posts/git-commit/#separate) and `machine.getOwn...  not a namespace` is message commit body. That why we'll see only `Fix #3042` by default with github, gitg or gitk. So it would be better to see clear message without check commit details

sleshchenko:
Cool.
> The first line Fix is for https://help.github.com/articles/closing-issues-via-commit-messages/

It's new information for me. Thx :+1:

skabashnyuk (approved):

PR #3376, opened 2016-12-13
CHE-3359 - Starting a workspace as an user

### LINKED ISSUE #3359: Starting a workspace as an user
CHE-3264 enables starting the che-server as an user. 
Now, we have to enable starting a workspace as an user.

See #3264 for more detail.

**Che version:**    any
**OS and version:** any
**Docker version:** any
**Che install:**   docker

snjeza:
### What does this PR do?

Fixes #3359 

#3359

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
@@ -186,6 +186,8 @@ che.docker.cpuset_cpus=NULL
 che.docker.cpu_period=0
 che.docker.cpu_quota=0
 
+# Adds options for che user id
+che.docker.user_id=
 # Adds options when mounting the /projects volume.
 che.docker.volumes_projects_options=Z
 
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
@@ -134,7 +134,7 @@ public class MachineProviderImpl implements MachineInstanceProvider {
     private final long                                          cpuQuota;
     private final WindowsPathEscaper                            windowsPathEscaper;
     private final String[]                                      dnsResolvers;
-    private final String                                        cheUserId;
+    private final String                                        cheDockerUserId;
 
     @Inject
     public MachineProviderImpl(DockerConnectorProvider dockerProvider,
@@ -161,7 +161,7 @@ public class MachineProviderImpl implements MachineInstanceProvider {
                                WindowsPathEscaper windowsPathEscaper,
                                @Named("che.docker.extra_hosts") Set<Set<String>> additionalHosts,
                                @Nullable @Named("che.docker.dns_resolvers") String[] dnsResolvers,
-                               @Nullable @Named("che.docker.user_id") String cheUserId)
+                               @Nullable @Named("che.docker.user_id") String cheDockerUserId)
             throws IOException {
         this.docker = dockerProvider.get();
         this.dockerCredentials = dockerCredentials;
@@ -170,7 +170,7 @@ public class MachineProviderImpl implements MachineInstanceProvider {
         this.doForcePullOnBuild = doForcePullOnBuild;
         this.privilegedMode = privilegedMode;
         this.snapshotUseRegistry = snapshotUseRegistry;
-        this.cheUserId = cheUserId;
+        this.cheDockerUserId = cheDockerUserId;
         // use-cases:
         //  -1  enable unlimited swap
         //  0   disable swap
@@ -295,7 +295,7 @@ public class MachineProviderImpl implements MachineInstanceProvider {
 
             docker.startContainer(StartContainerParams.create(container));
 
-            if (!Strings.isNullOrEmpty(cheUserId)) {
+            if (!Strings.isNullOrEmpty(cheDockerUserId)) {
                 changeWorkingDirOwner(container, machineLogger);
             }
 
@@ -552,7 +552,7 @@ public class MachineProviderImpl implements MachineInstanceProvider {
               .withEntrypoint(toArrayIfNotNull(service.getEntrypoint()))
               .withLabels(service.getLabels())
               .withNetworkingConfig(networkingConfig)
-              .withUser(cheUserId)
+              .withUser(cheDockerUserId)
               .withEnv(service.getEnvironment()
                               .entrySet()
                               .stream()
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
@@ -36,6 +36,7 @@ import org.eclipse.che.commons.lang.concurrent.LoggingUncaughtExceptionHandler;
 import org.eclipse.che.commons.lang.os.WindowsPathEscaper;
 import org.eclipse.che.plugin.docker.client.DockerConnector;
 import org.eclipse.che.plugin.docker.client.DockerConnectorProvider;
+import org.eclipse.che.plugin.docker.client.Exec;
 import org.eclipse.che.plugin.docker.client.ProgressLineFormatterImpl;
 import org.eclipse.che.plugin.docker.client.ProgressMonitor;
 import org.eclipse.che.plugin.docker.client.UserSpecificDockerRegistryCredentialsProvider;
@@ -52,12 +53,14 @@ import org.eclipse.che.plugin.docker.client.json.network.EndpointConfig;
 import org.eclipse.che.plugin.docker.client.json.network.NewNetwork;
 import org.eclipse.che.plugin.docker.client.params.BuildImageParams;
 import org.eclipse.che.plugin.docker.client.params.CreateContainerParams;
+import org.eclipse.che.plugin.docker.client.params.CreateExecParams;
 import org.eclipse.che.plugin.docker.client.params.GetContainerLogsParams;
 import org.eclipse.che.plugin.docker.client.params.PullParams;
 import org.eclipse.che.plugin.docker.client.params.RemoveContainerParams;
 import org.eclipse.che.plugin.docker.client.params.RemoveImageParams;
 import org.eclipse.che.plugin.docker.client.params.RemoveNetworkParams;
 import org.eclipse.che.plugin.docker.client.params.StartContainerParams;
+import org.eclipse.che.plugin.docker.client.params.StartExecParams;
 import org.eclipse.che.plugin.docker.client.params.TagParams;
 import org.eclipse.che.plugin.docker.client.params.network.ConnectContainerToNetworkParams;
 import org.eclipse.che.plugin.docker.client.params.network.CreateNetworkParams;
@@ -131,6 +134,7 @@ public class MachineProviderImpl implements MachineInstanceProvider {
     private final long                                          cpuQuota;
     private final WindowsPathEscaper                            windowsPathEscaper;
     private final String[]                                      dnsResolvers;
+    private final String                                        cheUserId;
 
     @Inject
     public MachineProviderImpl(DockerConnectorProvider dockerProvider,
@@ -156,7 +160,8 @@ public class MachineProviderImpl implements MachineInstanceProvider {
                                @Named("che.docker.cpu_quota") long cpuQuota,
                                WindowsPathEscaper windowsPathEscaper,
                                @Named("che.docker.extra_hosts") Set<Set<String>> additionalHosts,
-                               @Nullable @Named("che.docker.dns_resolvers") String[] dnsResolvers)
+                               @Nullable @Named("che.docker.dns_resolvers") String[] dnsResolvers,
+                               @Nullable @Named("che.docker.che.user.id") String cheUserId)
             throws IOException {
         this.docker = dockerProvider.get();
         this.dockerCredentials = dockerCredentials;
@@ -165,6 +170,11 @@ public class MachineProviderImpl implements MachineInstanceProvider {
         this.doForcePullOnBuild = doForcePullOnBuild;
         this.privilegedMode = privilegedMode;
         this.snapshotUseRegistry = snapshotUseRegistry;
+        if (Strings.isNullOrEmpty(cheUserId)) {
+            this.cheUserId = System.getenv("CHE_USER_ID") == null ? "" : System.getenv("CHE_USER_ID");
+        } else {
+            this.cheUserId = cheUserId;
+        }
         // use-cases:
         //  -1  enable unlimited swap
         //  0   disable swap
@@ -289,6 +299,10 @@ public class MachineProviderImpl implements MachineInstanceProvider {
 
             docker.startContainer(StartContainerParams.create(container));
 
+            if (!Strings.isNullOrEmpty(cheUserId)) {
+                changeWorkingDirOwner(container, machineLogger);
+            }
+
             readContainerLogsInSeparateThread(container,
                                               workspaceId,
                                               service.getId(),
@@ -336,7 +350,31 @@ public class MachineProviderImpl implements MachineInstanceProvider {
         }
     }
 
-    @Override
+    // sets the home directory for a user defined with CHE_USER
+    private void changeWorkingDirOwner(String containerId, LineConsumer machineLogger) throws MachineException {
+        String prepareCmd = "CHE_USER=`id -un`;"
+                + "sudo chown -R ${CHE_USER}:docker /home/user;"
+                + "sudo ln -s /home/user `eval echo \"~${CHE_USER}\"`";
+        final String[] command = {"/bin/sh", "-c", prepareCmd};
+        Exec exec;
+        try {
+            exec = docker.createExec(CreateExecParams.create(containerId, command).withDetach(false));
+        } catch (IOException e) {
+            LOG.error(format("Error occurs while initializing command %s in docker container %s: %s",
+                                              Arrays.toString(command), containerId, e.getMessage()), e);
+            return;
+        }
+        try {
+            if (exec != null) {
+                docker.startExec(StartExecParams.create(exec.getId()), new LogMessagePrinter(machineLogger));
+            }
+        } catch (IOException e) {
+            LOG.error(format("Error occurs while executing command %s in docker container %s: %s",
+                                              Arrays.toString(exec.getCommand()), containerId, e.getMessage()), e);
+        }
+    }
+
+	@Override
     public void createNetwork(String networkName) throws ServerException {
         try {
             docker.createNetwork(CreateNetworkParams.create(new NewNetwork().withName(networkName)
@@ -520,6 +558,7 @@ public class MachineProviderImpl implements MachineInstanceProvider {
               .withEntrypoint(toArrayIfNotNull(service.getEntrypoint()))
               .withLabels(service.getLabels())
               .withNetworkingConfig(networkingConfig)
+              .withUser(cheUserId)
               .withEnv(service.getEnvironment()
                               .entrySet()
                               .stream()
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
@@ -361,9 +361,7 @@ public class MachineProviderImpl implements MachineInstanceProvider {
             return;
         }
         try {
-            if (exec != null) {
-                docker.startExec(StartExecParams.create(exec.getId()), new LogMessagePrinter(machineLogger));
-            }
+            docker.startExec(StartExecParams.create(exec.getId()), new LogMessagePrinter(machineLogger));
         } catch (IOException e) {
             LOG.error(format("Error occurs while executing command %s in docker container %s: %s",
                                               Arrays.toString(exec.getCommand()), containerId, e.getMessage()), e);
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
@@ -161,7 +161,7 @@ public class MachineProviderImpl implements MachineInstanceProvider {
                                WindowsPathEscaper windowsPathEscaper,
                                @Named("che.docker.extra_hosts") Set<Set<String>> additionalHosts,
                                @Nullable @Named("che.docker.dns_resolvers") String[] dnsResolvers,
-                               @Nullable @Named("che.docker.che.user.id") String cheUserId)
+                               @Nullable @Named("che.docker.user_id") String cheUserId)
             throws IOException {
         this.docker = dockerProvider.get();
         this.dockerCredentials = dockerCredentials;
@@ -170,11 +170,7 @@ public class MachineProviderImpl implements MachineInstanceProvider {
         this.doForcePullOnBuild = doForcePullOnBuild;
         this.privilegedMode = privilegedMode;
         this.snapshotUseRegistry = snapshotUseRegistry;
-        if (Strings.isNullOrEmpty(cheUserId)) {
-            this.cheUserId = System.getenv("CHE_USER_ID") == null ? "" : System.getenv("CHE_USER_ID");
-        } else {
-            this.cheUserId = cheUserId;
-        }
+        this.cheUserId = cheUserId;
         // use-cases:
         //  -1  enable unlimited swap
         //  0   disable swap
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalCheInfrastructureProvisioner.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalCheInfrastructureProvisioner.java
@@ -72,11 +72,7 @@ public class LocalCheInfrastructureProvisioner extends DefaultInfrastructureProv
         this.wsAgentVolumeProvider = wsAgentVolumeProvider;
         this.dockerExtConfBindingProvider = dockerExtConfBindingProvider;
         this.terminalVolumeProvider = terminalVolumeProvider;
-        if (Strings.isNullOrEmpty(cheUserId)) {
-            this.cheUserId = System.getenv("CHE_USER_ID");
-        } else {
-            this.cheUserId = cheUserId;
-        }
+        this.cheUserId = cheUserId;
         if (!Strings.isNullOrEmpty(projectsVolumeOptions)) {
             this.projectsVolumeOptions = ":" + projectsVolumeOptions;
         } else {
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalCheInfrastructureProvisioner.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalCheInfrastructureProvisioner.java
@@ -51,6 +51,9 @@ public class LocalCheInfrastructureProvisioner extends DefaultInfrastructureProv
     private final DockerExtConfBindingProvider dockerExtConfBindingProvider;
     private final TerminalVolumeProvider       terminalVolumeProvider;
     private final String                       projectsVolumeOptions;
+    public static final String ETC_PASSWD_VOLUME = "/etc/passwd";
+    public static final String ETC_GROUP_VOLUME = "/etc/group";
+    private final String                      cheUserId;
 
     @Inject
     public LocalCheInfrastructureProvisioner(AgentConfigApplier agentConfigApplier,
@@ -60,7 +63,8 @@ public class LocalCheInfrastructureProvisioner extends DefaultInfrastructureProv
                                              @Nullable @Named("che.docker.volumes_projects_options") String projectsVolumeOptions,
                                              WsAgentVolumeProvider wsAgentVolumeProvider,
                                              DockerExtConfBindingProvider dockerExtConfBindingProvider,
-                                             TerminalVolumeProvider terminalVolumeProvider) {
+                                             TerminalVolumeProvider terminalVolumeProvider,
+                                             @Nullable @Named("che.docker.user_id") String cheUserId) {
         super(agentConfigApplier);
         this.workspaceFolderPathProvider = workspaceFolderPathProvider;
         this.pathEscaper = pathEscaper;
@@ -68,6 +72,11 @@ public class LocalCheInfrastructureProvisioner extends DefaultInfrastructureProv
         this.wsAgentVolumeProvider = wsAgentVolumeProvider;
         this.dockerExtConfBindingProvider = dockerExtConfBindingProvider;
         this.terminalVolumeProvider = terminalVolumeProvider;
+        if (Strings.isNullOrEmpty(cheUserId)) {
+            this.cheUserId = System.getenv("CHE_USER_ID");
+        } else {
+            this.cheUserId = cheUserId;
+        }
         if (!Strings.isNullOrEmpty(projectsVolumeOptions)) {
             this.projectsVolumeOptions = ":" + projectsVolumeOptions;
         } else {
@@ -115,6 +124,9 @@ public class LocalCheInfrastructureProvisioner extends DefaultInfrastructureProv
         environmentVars.put(CheBootstrap.CHE_LOCAL_CONF_DIR, DockerExtConfBindingProvider.EXT_CHE_LOCAL_CONF_DIR);
         devMachine.setEnvironment(environmentVars);
 
+        if (!Strings.isNullOrEmpty(cheUserId)) {
+            envConfig.getMachines().entrySet().forEach(entry -> addVolume(internalEnv, entry.getKey()));
+        }
         // apply basic infra (e.g. agents)
         super.provision(envConfig, internalEnv);
     }
@@ -126,4 +138,24 @@ public class LocalCheInfrastructureProvisioner extends DefaultInfrastructureProv
 
         super.provision(machineConfig, internalMachine);
     }
+
+    private void addVolume(CheServicesEnvironmentImpl internalEnv, String machineName) {
+        addVolume(internalEnv, machineName, ETC_PASSWD_VOLUME);
+        addVolume(internalEnv, machineName, ETC_GROUP_VOLUME);
+    }
+
+    private void addVolume(CheServicesEnvironmentImpl internalEnv, String machineName, String volume) {
+        String path;
+        if (projectsVolumeOptions != null && projectsVolumeOptions.contains("Z")) {
+            path = format("%s:%s:ro,Z", volume, volume);
+        } else {
+            path = format("%s:%s:ro", volume, volume);
+        }
+        internalEnv.getServices()
+                .get(machineName)
+                .getVolumes()
+                .add(SystemInfo.isWindows() ? pathEscaper.escapePath(path)
+                                            : path);
+    }
+
 }
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/MachineProviderImplTest.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/MachineProviderImplTest.java
@@ -25,6 +25,7 @@ import org.eclipse.che.commons.subject.SubjectImpl;
 import org.eclipse.che.plugin.docker.client.DockerConnector;
 import org.eclipse.che.plugin.docker.client.DockerConnectorConfiguration;
 import org.eclipse.che.plugin.docker.client.DockerConnectorProvider;
+import org.eclipse.che.plugin.docker.client.Exec;
 import org.eclipse.che.plugin.docker.client.ProgressMonitor;
 import org.eclipse.che.plugin.docker.client.UserSpecificDockerRegistryCredentialsProvider;
 import org.eclipse.che.plugin.docker.client.json.ContainerConfig;
@@ -33,11 +34,13 @@ import org.eclipse.che.plugin.docker.client.json.ContainerInfo;
 import org.eclipse.che.plugin.docker.client.json.ContainerState;
 import org.eclipse.che.plugin.docker.client.json.Volume;
 import org.eclipse.che.plugin.docker.client.params.CreateContainerParams;
+import org.eclipse.che.plugin.docker.client.params.CreateExecParams;
 import org.eclipse.che.plugin.docker.client.params.InspectContainerParams;
 import org.eclipse.che.plugin.docker.client.params.PullParams;
 import org.eclipse.che.plugin.docker.client.params.RemoveContainerParams;
 import org.eclipse.che.plugin.docker.client.params.RemoveImageParams;
 import org.eclipse.che.plugin.docker.client.params.StartContainerParams;
+import org.eclipse.che.plugin.docker.client.params.StartExecParams;
 import org.eclipse.che.plugin.docker.client.params.TagParams;
 import org.eclipse.che.plugin.docker.machine.node.DockerNode;
 import org.mockito.ArgumentCaptor;
@@ -70,6 +73,7 @@ import static java.util.stream.Collectors.toMap;
 import static org.eclipse.che.plugin.docker.machine.DockerInstanceProvider.DOCKER_FILE_TYPE;
 import static org.eclipse.che.plugin.docker.machine.DockerInstanceProvider.MACHINE_SNAPSHOT_PREFIX;
 import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyObject;
 import static org.mockito.Matchers.anyString;
 import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.doThrow;
@@ -93,6 +97,7 @@ public class MachineProviderImplTest {
     private static final int     MEMORY_SWAP_MULTIPLIER = 0;
     private static final String  ENV_NAME               = "env";
     private static final String  NETWORK_NAME           = "networkName";
+    private static final String  EXEC_ID                = "exec";
 
     @Mock
     private DockerConnector dockerConnector;
@@ -123,6 +128,9 @@ public class MachineProviderImplTest {
 
     @Mock
     private WindowsPathEscaper pathEscaper;
+
+    @Mock
+    private Exec exec;
 
     private MachineProviderImpl provider;
 
@@ -158,6 +166,8 @@ public class MachineProviderImplTest {
         when(dockerConnector.inspectContainer(any(InspectContainerParams.class))).thenReturn(containerInfo);
         when(containerInfo.getState()).thenReturn(containerState);
         when(containerState.isRunning()).thenReturn(false);
+        when(dockerConnector.createExec(any(CreateExecParams.class))).thenReturn(exec);
+        when(exec.getId()).thenReturn(EXEC_ID);
     }
 
     @AfterMethod
@@ -335,6 +345,45 @@ public class MachineProviderImplTest {
         ArgumentCaptor<CreateContainerParams> argumentCaptor = ArgumentCaptor.forClass(CreateContainerParams.class);
         verify(dockerConnector).createContainer(argumentCaptor.capture());
         assertTrue(argumentCaptor.getValue().getContainerConfig().getHostConfig().isPrivileged());
+    }
+
+    @Test
+    public void shouldStartExecWithCheUserId() throws Exception {
+        final String user = "test";
+        provider = new MachineProviderBuilder().setUser(user)
+                                               .build();
+        final boolean isDev = true;
+        CheServiceImpl service = createService();
+        service.setVolumes(null);
+        createInstanceFromRecipe(service, isDev);
+        ArgumentCaptor<CreateContainerParams> argumentCaptor = ArgumentCaptor.forClass(CreateContainerParams.class);
+        verify(dockerConnector).createContainer(argumentCaptor.capture());
+        verify(dockerConnector).startContainer(any(StartContainerParams.class));
+        ArgumentCaptor<CreateExecParams> argumentCaptorExec = ArgumentCaptor.forClass(CreateExecParams.class);
+        verify(dockerConnector).createExec(argumentCaptorExec.capture());
+        String prepareCmd = "CHE_USER=`id -un`;"
+                + "sudo chown -R ${CHE_USER}:docker /home/user;"
+                + "sudo ln -s /home/user `eval echo \"~${CHE_USER}\"`";
+        final String[] command = {"/bin/sh", "-c", prepareCmd};
+        assertEquals(argumentCaptorExec.getValue().getCmd(), command);
+        verify(dockerConnector).startExec(eq(StartExecParams.create(EXEC_ID)), anyObject());
+        assertEquals(argumentCaptor.getValue().getContainerConfig().getUser(), user);
+    }
+
+    @Test
+    public void shouldNotStartExecWithoutCheUserId() throws Exception {
+        provider = new MachineProviderBuilder().build();
+        final boolean isDev = true;
+        CheServiceImpl service = createService();
+        service.setVolumes(null);
+        createInstanceFromRecipe(service, isDev);
+        ArgumentCaptor<CreateContainerParams> argumentCaptor = ArgumentCaptor.forClass(CreateContainerParams.class);
+        verify(dockerConnector).createContainer(argumentCaptor.capture());
+        verify(dockerConnector).startContainer(any(StartContainerParams.class));
+        ArgumentCaptor<CreateExecParams> argumentCaptorExec = ArgumentCaptor.forClass(CreateExecParams.class);
+        verify(dockerConnector, never()).createExec(argumentCaptorExec.capture());
+        verify(dockerConnector, never()).startExec(eq(StartExecParams.create(EXEC_ID)), anyObject());
+        assertTrue(argumentCaptor.getValue().getContainerConfig().getUser() == null);
     }
 
     @Test
@@ -708,6 +757,33 @@ public class MachineProviderImplTest {
                            .containsAll(asList("9090", "8080")));
     }
 
+    @Test
+    public void shouldAddUserOnDevInstanceCreationFromRecipe() throws Exception {
+        final String user = "test";
+        provider = new MachineProviderBuilder().setUser(user)
+                                               .build();
+        final boolean isDev = true;
+        CheServiceImpl service = createService();
+        service.setVolumes(null);
+        createInstanceFromRecipe(service, isDev);
+        ArgumentCaptor<CreateContainerParams> argumentCaptor = ArgumentCaptor.forClass(CreateContainerParams.class);
+        verify(dockerConnector).createContainer(argumentCaptor.capture());
+        verify(dockerConnector).startContainer(any(StartContainerParams.class));
+        assertEquals(argumentCaptor.getValue().getContainerConfig().getUser(), user);
+    }
+
+    @Test
+    public void shouldAddUserOnDevInstanceCreationFromSnapshot() throws Exception {
+        final String user = "test";
+        provider = new MachineProviderBuilder().setUser(user)
+                                               .build();
+        final boolean isDev = true;
+        createInstanceFromSnapshot(isDev);
+        ArgumentCaptor<CreateContainerParams> argumentCaptor = ArgumentCaptor.forClass(CreateContainerParams.class);
+        verify(dockerConnector).createContainer(argumentCaptor.capture());
+        verify(dockerConnector).startContainer(any(StartContainerParams.class));
+        assertEquals(argumentCaptor.getValue().getContainerConfig().getUser(), user);
+    }
     @Test
     public void shouldAddBindMountAndRegularVolumesOnInstanceCreationFromRecipe() throws Exception {
         String[] bindMountVolumesFromMachine = new String[] {"/my/bind/mount1:/from/host1",
@@ -1463,6 +1539,7 @@ public class MachineProviderImplTest {
         private String           networkDriver;
         private String           parentCgroup;
         private String           cpuSet;
+        private String           user;
         private long             cpuPeriod;
         private long             cpuQuota;
         private String[]         dnsResolvers;
@@ -1481,6 +1558,7 @@ public class MachineProviderImplTest {
             extraHosts = emptySet();
             memorySwapMultiplier = MEMORY_SWAP_MULTIPLIER;
             pidsLimit = -1;
+            user = null;
         }
 
         public MachineProviderBuilder setDevMachineEnvVars(Set<String> devMachineEnvVars) {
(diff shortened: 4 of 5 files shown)


plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -340,7 +347,27 @@ public Instance startService(String namespace,
         }
     }
 
-    @Override
+    private void prepareContainer(String containerId, LineConsumer machineLogger) throws MachineException {
+        String prepareCmd = "CHE_USER=`id -un`;"
+                + "sudo chown -R ${CHE_USER}:docker /home/user;"
+                + "sudo ln -s /home/user `eval echo \"~${CHE_USER}\"`";
```
l0rd: Why do you need to create this symbolic link?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -340,7 +347,27 @@ public Instance startService(String namespace,
         }
     }
 
-    @Override
+    private void prepareContainer(String containerId, LineConsumer machineLogger) throws MachineException {
```
l0rd: name `prepareContainer` doesn't say much about this method. What about `changeWorkingDirOwner`?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -340,7 +347,27 @@ public Instance startService(String namespace,
...
+            throw new MachineException(format("Error occurs while initializing command %s in docker container %s: %s",
+                                              Arrays.toString(command), containerId, e.getMessage()), e);
+        }
+        try {
+            docker.startExec(StartExecParams.create(exec.getId()), new LogMessagePrinter(machineLogger));
+        } catch (IOException e) {
+            throw new MachineException(format("Error occurs while executing command %s in docker container %s: %s",
+                                              Arrays.toString(exec.getCommand()), containerId, e.getMessage()), e);
+        }
+	}
```
l0rd: I think that last curly bracket is not correctly aligned

l0rd:
@snjeza your PR looks ok to me. How have you tested your PR? Can you please tell me the command line you have used to start Che and the simple steps of your tests?  

@garagatyi and @benoitf I think you should have a look too

snjeza:
> Why do you need to create this symbolic link?

Che supposes a workspace runs as the "user" user that has a specific home directory. The command sets the same environment for any user starting the workspace.

> your PR looks ok to me. How have you tested your PR? Can you please tell me the command line you have used to start Che and the simple steps of your tests?

You start che as described in https://github.com/eclipse/che/pull/3265

```
docker run -p 8080:8080 \
           --name che \
	   --rm \
	   -u `id -u test` \
           -e CHE_USER=test \
           -v /tmp/data:/data \
	   -v /etc/group:/etc/group:ro \
           eclipse/che-server:ver 
```

create a workspace and call 

```
docker exec <workspace_container_id> id
```

If you start a workspace as usually, you will get the user "user(1000)". When using the above command, you will get the "test" user and its ID. The test user must have the sudo NOPASSWD permission (must be a member of the sudo group).

garagatyi:
@snjeza Can you elaborate on what this PR does and what for. It is not clear in most cases what PRs do, so we are welcoming in PRs body detailed explanation of all information  needed for reviewers.

snjeza:
> @snjeza Can you elaborate on what this PR does and what for. It is not clear in most cases what PRs do, so we are welcoming in PRs body detailed explanation of all information needed for reviewers.

CHE-3264 implements the CHE_USER environment variable. However, only the che server is started as a user defined with CHE_USER.
Workspaces still start as the "user" user. This PR enables a workspace to start as a user defined with the CHE_USER environment variable.

By executing the following command, you can see which user started the che server or a workspace:
```
docker exec -it <server_container_id> id
docker exec -it <workspace_container_id> id
```

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -294,6 +297,10 @@ public Instance startService(String namespace,
 
             docker.startContainer(StartContainerParams.create(container));
 
+            if (System.getenv("CHE_USER_ID") != null) {
+                changeWorkingDirOwner(container, machineLogger);
```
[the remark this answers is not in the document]
snjeza: When che starts a workspace, it expects that there is the "user" user that has the /home/user directory as the home directory with a specific content.
You can check that by the following:
```
docker run -it codenvy/ubuntu_jdk8 id
docker run -it codenvy/ubuntu_jdk8 ls -la /home/user
```

The changeWorkingDirOwner method sets the same environment for a user defined with CHE_USER which enables us to use the existing che images.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -294,6 +297,10 @@ public Instance startService(String namespace,
 
             docker.startContainer(StartContainerParams.create(container));
 
+            if (System.getenv("CHE_USER_ID") != null) {
```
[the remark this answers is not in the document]
snjeza: I'm not sure how to do that.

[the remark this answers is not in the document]
benoitf: @snjeza   https://github.com/eclipse/che/blob/e8840780afc9ba8ebeaa4af9730f5e749cda4ebc/plugins/plugin-java/che-plugin-java-ext-lang-server/src/main/java/org/eclipse/che/plugin/java/server/rest/WsAgentURLProvider.java#L45

benoitf (dismissed):
FYI @snjeza master branch has been rewritten on docs import to discard images files so you'll have to rebase your branch to be in sync with latest master. I can help around this task if you need help

garagatyi:
I don't understand why should we change user inside of workspace containers and change ownership of files. It is quiet risky operation to me that can raise a lot of issues with some specific software in containers. 
Please provide more information on why we need that.

TylerJewell:
The basic expectation is that a workspace is the ownership of the user, not of Che. And so if the end user or admin thinks of a workspace runtime as "their" property, then they want to have root-like behaviors. Their project may require a certain user identity or other override behaviors. So this would give an admin the ability to configure a workspace as a particular user.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalCheInfrastructureProvisioner.java
```diff
@@ -89,7 +94,26 @@ public void provision(Environment envConfig, CheServicesEnvironmentImpl internal
                    .add(SystemInfo.isWindows() ? pathEscaper.escapePath(projectFolderVolume)
                                                : projectFolderVolume);
 
+        if (System.getenv(CHE_USER_ID) != null) {
+            addVolume(internalEnv, devMachineName, ETC_PASSWD_VOLUME);
```
garagatyi: Please add unit tests to cover changed code. Unfortunately there are no tests yet for this class. But you can cover only your code if you will.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -528,6 +556,7 @@ private String createContainer(String workspaceId,
               .withEntrypoint(toArrayIfNotNull(service.getEntrypoint()))
               .withLabels(service.getLabels())
               .withNetworkingConfig(networkingConfig)
+              .withUser(getUser())
```
garagatyi: Will container start if there is no such user in container?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -341,7 +348,28 @@ public Instance startService(String namespace,
         }
     }
 
-    @Override
+    // sets the home directory for a user defined with CHE_USER
+    private void changeWorkingDirOwner(String containerId, LineConsumer machineLogger) throws MachineException {
+        String prepareCmd = "CHE_USER=`id -un`;"
+                + "sudo chown -R ${CHE_USER}:docker /home/user;"
+                + "sudo ln -s /home/user `eval echo \"~${CHE_USER}\"`";
```
garagatyi, line 353: Mario asked that and I haven't found answer. Why do you need to create this symbolic link?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalCheInfrastructureProvisioner.java
```diff
@@ -89,7 +94,26 @@ public void provision(Environment envConfig, CheServicesEnvironmentImpl internal
...
+
+	private void addVolume(CheServicesEnvironmentImpl internalEnv, String devMachineName, String volume) {
+        String path;
+        if (projectsVolumeOptions != null && projectsVolumeOptions.contains("Z")) {
+            path = format("%s:%s:ro,Z", volume, volume);
+        } else {
+            path = format("%s:%s:ro", volume, volume);
+        }
+        internalEnv.getServices()
+                .get(devMachineName)
```
garagatyi: Why mapping of these files are needed for dev machines only and not all the machines?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -294,6 +297,10 @@ public Instance startService(String namespace,
 
             docker.startContainer(StartContainerParams.create(container));
 
+            if (!Strings.isNullOrEmpty(getUser())) {
+                changeWorkingDirOwner(container, machineLogger);
```
garagatyi: Is it OK that previous user in container is already running main process of container?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -341,7 +348,28 @@ public Instance startService(String namespace,
         }
     }
 
-    @Override
+    // sets the home directory for a user defined with CHE_USER
+    private void changeWorkingDirOwner(String containerId, LineConsumer machineLogger) throws MachineException {
+        String prepareCmd = "CHE_USER=`id -un`;"
+                + "sudo chown -R ${CHE_USER}:docker /home/user;"
```
garagatyi, line 352: Why user home dir here and in link uses hardcoded path /home/user? User in container can have another name. And if it is root it can be in /root instead.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
garagatyi: Sorry I misunderstood it a bit. In that case main process will run under new user but files on container's FS will be under old user. It can raise problem with that container. WDYT?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
garagatyi: And user may not have home directory at all. And main process of container may have some  permissions problems because of some hardcoded permissions on FS in container but not in home folder.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
garagatyi: Probably with mounting of /etc/passwd and /etc/group user kinda already exists in container. But there is possibility of permissions problems.

l0rd: @garagatyi user will be the same that is running in che-server container. The same env variable CHE_USER_ID is used by both workspace and che-server.

l0rd: By the way I think the purpose of this PR is to solve permissions problems. Let's say that I'm logged in my Linux box using user `mario`. When starting a Che workspace I would expect it to run as user `mario` too. With permission to open files that I can read and only those. This user does not exist in the original recipe docker image but mounting the /etc/passwd will do the trick.

TylerJewell:
Alex - I generally agree with that basic philosophy.  I actually don't know how we haven't been getting user complaints on this limitation in higher volume.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
garagatyi: Yes, I got it. Thanks for additional explanation.

snjeza:
I have updated the patch.

l0rd:
@snjeza I'm sorry but I'm getting lost here 😄  

Could you remember me what you have you changed since the last commit? Looking at current PR I can't remember the difference of this last update with the previous one.

And there is a question about the commands executed in the container (`chown` and `ln -s`) that hasn't been answered.

I think we are really close to merge this PR but we need these details from you.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -341,7 +351,31 @@ public Instance startService(String namespace,
...
+        Exec exec;
+        try {
+            exec = docker.createExec(CreateExecParams.create(containerId, command).withDetach(false));
+        } catch (IOException e) {
+            LOG.error(format("Error occurs while initializing command %s in docker container %s: %s",
+                                              Arrays.toString(command), containerId, e.getMessage()), e);
+            return;
+        }
+        try {
+            if (exec != null) {
```
garagatyi: This method is not covered by tests, can you cover it?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -341,7 +351,31 @@ public Instance startService(String namespace,
         }
     }
 
-    @Override
+    // sets the home directory for a user defined with CHE_USER
+    private void changeWorkingDirOwner(String containerId, LineConsumer machineLogger) throws MachineException {
+        String prepareCmd = "CHE_USER=`id -un`;"
+                + "sudo chown -R ${CHE_USER}:docker /home/user;"
```
garagatyi, line 352: What if user is root in a container? His home folder will be /root instead of /home/user. Or even if user has different username - not the "user".

snjeza:
@l0rd 
> Could you remember me what you have you changed since the last commit? Looking at current PR I can't remember the difference of this last update with the previous one.

Regarding @garagatyi  requests, I have added the following:
- a dependency injection instead of the getenv method
- mapping /etc/passwd, /etc/group for all machines, not for a dev machine only
- two tests

> And there is a question about the commands executed in the container (chown and ln -s) that hasn't been answered.

I have already answered it.

>> Why do you need to create this symbolic link?

> Che supposes a workspace runs as the "user" user that has a specific home directory. The command sets the same environment for any user starting the workspace.

The home directory contains some required files (maven, tomcat, ...). Che won't start a workspace if those files/directories aren't present. 

> What if user is root in a container? His home folder will be /root instead of /home/user. Or even if user has different username - not the "user".

The command won't do anything. We don't know what to do in this case because we don't know what user is a default user. Do you know an image I could test?

> This method is not covered by tests, can you cover it?

I have tried to do it, but the che mockup docker container returns null when calling "docker.createExec". I don't know how to test this method.

@garagatyi 

How to fix the issue with @Named("env.MYVAR") when MYVAR isn't set? Che throws an exception in this case.

garagatyi:
> The command won't do anything. We don't know what to do in this case because we don't know what user is a default user. Do you know an image I could test?

I suppose that if current user is root and user "user" exists in s container this command will do unneeded chown.
I think you can try to use official ubuntu image to check situation when user is root.
And probably in bitnami stacks user is named in a different way, but I'm not sure.

> I have tried to do it, but the che mockup docker container returns null when calling "docker.createExec". I don't know how to test this method.

Here is an example how to use mocks to stub container creation. You can do pretty the same for exec. The difference is that you can do it in you specific test instead of setup method because this stubbing is not needed by other tests. 
https://github.com/eclipse/che/blob/master/plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/MachineProviderImplTest.java#L136-L136
If it is unclear on how to do that, please, notify me and I'll try to help you on that.

> How to fix the issue with @Named("env.MYVAR") when MYVAR isn't set? Che throws an exception in this case.

Yes, we don't have such behavior for now. What can I suggest instead is to create property. Then it will be possible to inject it, to have default value as null and to override it with environment variable.
If you are not sure how to do that I can explain in details how this technic works in Che.

l0rd:
| The home directory contains some required files (maven, tomcat, ...). Che won't start a workspace if those files/directories aren't present.

This means that the workspaces Docker images have been built assuming that they will be run by user `user`. Adding a symbolic link + chown is just a workaround. We should address the root cause in the Docker images instead. Once we modify the Docker images we can remove the workaround but I think we can do that work in a distinct PR. We should open an issue to track that. @snjeza and @garagatyi what do you think?

garagatyi:
@l0rd Sorry I don't understand what do you mean. If you want to run containers by user "myuser" there is no possibility to fix root issue that it was not considered by author of docker image.
Can you elaborate on which another issue we can create and what can we fix to remove that workaround

l0rd:
@garagatyi Let's take [Java default stack](https://github.com/eclipse/che-dockerfiles/blob/master/recipes/ubuntu_jdk8/Dockerfile). Tomcat and maven are installed in `/home/user` folder: 
```dockerfile
RUN mkdir /home/user/tomcat8 /home/user/apache-maven-$MAVEN_VERSION
```
This should be changed. Along with everything is related to a specific `user`. Docker images should be built to be a run as any user. For example tomcat and maven could be installed  in `/che/apps/` that is not related to a particular user.

garagatyi:
I don't see the reason why stack author have to predict that someone would want to change user in container he designed. Also users can create workspaces from custom recipes instead of stacks. In that case we can't rely on what user exists in container and how software is configured in that container.

garagatyi:
Can you remind me why it is needed to use particular user in containers instead of those author of stack has configured?

l0rd:
Yep https://github.com/eclipse/che/pull/3376#discussion_r94806449

garagatyi:
Thank you.
So how do you see solution for stacks that use user named in a different way or even root? I mean solution as part of this PR. Do you think we should solve this problem?

l0rd:
I would solve the image problem in a separate PR. @snjeza has provided a workaround (using `ln -s` and `chmod`). The workaround is executed only if che override the user (i.e. env var CHE_USER_ID is provided) so stacks that are not built to be run with any "myuser" will still work if CHE_USER_ID is not provide.

garagatyi:
But if user wants to run Che as his host user have to understand that custom recipes are not supported, right?

l0rd:
If a custom recipe expect that the workspace will be run with a particular user (for example `root` or `user`) then yes I agree that it will fail. But I think that recipes should not give for granted that they will run as a specific user. That's like an application that need to be run as root. It's a bad practice that should be discouraged.

snjeza:
> Here is an example how to use mocks to stub container creation. You can do pretty the same for exec. 
The difference is that you can do it in you specific test instead of setup method because this stubbing is not needed by other tests.
https://github.com/eclipse/che/blob/master/plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/MachineProviderImplTest.java#L136-L136
If it is unclear on how to do that, please, notify me and I'll try to help you on that.

>> How to fix the issue with @Named("env.MYVAR") when MYVAR isn't set? Che throws an exception in this case.

> Yes, we don't have such behavior for now. What can I suggest instead is to create property. Then it will be possible to inject it, to have default value as null and to override it with environment variable.
If you are not sure how to do that I can explain in details how this technic works in Che.

I have updated the patch.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -157,6 +157,8 @@ che.docker.cpuset_cpus=NULL
 che.docker.cpu_period=0
 che.docker.cpu_quota=0
 
+# Adds options for che user id
+che.docker.che.user.id=
```
garagatyi: Please do not use che twice in property name. We also have convention on usage of underscore to separate works of some term and dots to separate structural elements.
What about `che.docker.user_id` ? @TylerJewell since you are managing user-friendly names, please take a look

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -166,6 +171,11 @@ public MachineProviderImpl(DockerConnector docker,
         this.doForcePullOnBuild = doForcePullOnBuild;
         this.privilegedMode = privilegedMode;
         this.snapshotUseRegistry = snapshotUseRegistry;
+        if (Strings.isNullOrEmpty(cheUserId)) {
+            this.cheUserId = System.getenv("CHE_USER_ID") == null ? "" : System.getenv("CHE_USER_ID");
```
garagatyi: To override value of this property env variable with special name should be used. Letters are capitalized, underscores are transformed into double underscores, dots into single underscores.
In that case injection of property doesn't need to use **System.getenv** at all. Developer injects property and Che bootstrapping sequence take care on reading env variable and injecting it as  property.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalCheInfrastructureProvisioner.java
```diff
@@ -35,21 +35,32 @@
...
                                              @Named("che.workspace.projects.storage") String projectFolderPath,
-                                             @Nullable @Named("che.docker.volumes_projects_options") String projectsVolumeOptions) {
+                                             @Nullable @Named("che.docker.volumes_projects_options") String projectsVolumeOptions,
+                                             @Nullable @Named("che.docker.che.user.id") String cheUserId) {
         super(agentConfigApplier);
         this.workspaceFolderPathProvider = workspaceFolderPathProvider;
         this.pathEscaper = pathEscaper;
         this.projectFolderPath = projectFolderPath;
+        if (Strings.isNullOrEmpty(cheUserId)) {
+            this.cheUserId = System.getenv("CHE_USER_ID");
```
garagatyi: Take a look [at ](https://github.com/eclipse/che/pull/3376/files#r95524647)

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalCheInfrastructureProvisioner.java
```diff
@@ -89,7 +100,30 @@ public void provision(Environment envConfig, CheServicesEnvironmentImpl internal
                    .add(SystemInfo.isWindows() ? pathEscaper.escapePath(projectFolderVolume)
                                                : projectFolderVolume);
 
+        if (cheUserId != null) {
```
garagatyi: Is it on purpose that [there](https://github.com/eclipse/che/pull/3376/files#diff-fb8552dea5c1316ac9fd0b0988377f77R307) you also check for emptiness and here you don't

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/MachineProviderImplTest.java
```diff
@@ -318,6 +321,29 @@ public void shouldBeAbleToCreateContainerWithPrivilegeMode() throws Exception {
...
+        service.setVolumes(null);
+        createInstanceFromRecipe(isDev, service);
+        ArgumentCaptor<CreateContainerParams> argumentCaptor = ArgumentCaptor.forClass(CreateContainerParams.class);
+        verify(dockerConnector).createContainer(argumentCaptor.capture());
+        verify(dockerConnector).startContainer(any(StartContainerParams.class));
+        String prepareCmd = "CHE_USER=`id -un`;"
+                + "sudo chown -R ${CHE_USER}:docker /home/user;"
+                + "sudo ln -s /home/user `eval echo \"~${CHE_USER}\"`";
+        final String[] command = {"/bin/sh", "-c", prepareCmd};
+        Exec exec = dockerConnector.createExec(CreateExecParams.create(CONTAINER_ID, command).withDetach(false));
```
garagatyi: Here you call mocked methods. Instead of that it is needed to verify that they were called. Just like you did with **createContainer** and **startContainer**.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -528,6 +566,7 @@ private String createContainer(String workspaceId,
               .withEntrypoint(toArrayIfNotNull(service.getEntrypoint()))
               .withLabels(service.getLabels())
               .withNetworkingConfig(networkingConfig)
+              .withUser(cheUserId)
```
garagatyi: Have you tried what will happen if you pass here empty string? Is it valid value for docker?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -294,6 +304,10 @@ public Instance startService(String namespace,
 
             docker.startContainer(StartContainerParams.create(container));
 
+            if (!Strings.isNullOrEmpty(cheUserId)) {
```
garagatyi: Can you also cover in test that if property is not set then user is not set on container creation and exec is not called.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
snjeza: Yes, it is. The ContainerConfig class uses the empty string. See https://github.com/eclipse/che/blob/master/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/ContainerConfig.java#L44

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
snjeza: I have tried to rename CHE_USER_ID to CHE_PROPERTY_che_docker_user__id. However, che converts env. variable using cli.sh that is called before configuring CHE_USER_ID (entrypoint.sh) so that CHE_PROPERTY_che_docker_user__id isn't converted properly.

If che server starts with
```
docker run ... eclipse/che-server 
```

cli.sh won't be called.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
garagatyi: ok, thanks

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -153,7 +157,8 @@ public MachineProviderImpl(DockerConnector docker,
                                @Named("che.docker.cpu_period") long cpuPeriod,
                                @Named("che.docker.cpu_quota") long cpuQuota,
                                WindowsPathEscaper windowsPathEscaper,
-                               @Named("che.docker.extra_hosts") Set<Set<String>> additionalHosts)
+                               @Named("che.docker.extra_hosts") Set<Set<String>> additionalHosts,
+                               @Nullable @Named("che.docker.che.user.id") String cheUserId)
```
garagatyi: Looks like property should be **che.docker.user_id**

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
garagatyi: I don't mean binding  CHE_PROPERTY_che_docker_user__id. I mean that now we use properties. Then to overwrite it we add env variable into che.env file.

garagatyi:
@snjeza Please be aware that when you push with force into your PRs you make me review whole PR line by line from the beginning. If it is possible, please, use additional commits (you may use fixup commits) on top of the PR while it is under review.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
snjeza: Could you, please, explain how to use the property when starting the che server using:

```
docker run ... eclipse/che-server
```

garagatyi: I think you can do something like:
```
docker run ... -e "CHE_DOCKER_USER__ID=myuser" eclipse/che-server
```

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/MachineProviderImplTest.java
```diff
@@ -319,6 +322,29 @@ public void shouldBeAbleToCreateContainerWithPrivilegeMode() throws Exception {
...
+        createInstanceFromRecipe(isDev, service);
+        ArgumentCaptor<CreateContainerParams> argumentCaptor = ArgumentCaptor.forClass(CreateContainerParams.class);
+        verify(dockerConnector).createContainer(argumentCaptor.capture());
+        verify(dockerConnector).startContainer(any(StartContainerParams.class));
+        String prepareCmd = "CHE_USER=`id -un`;"
+                + "sudo chown -R ${CHE_USER}:docker /home/user;"
+                + "sudo ln -s /home/user `eval echo \"~${CHE_USER}\"`";
+        final String[] command = {"/bin/sh", "-c", prepareCmd};
+        Exec exec = verify(dockerConnector).createExec(CreateExecParams.create(CONTAINER_ID, command).withDetach(false));
+        if (exec != null) {
```
garagatyi: In this test user is set, so exec has to be executed.

garagatyi (approved):
Looks like now everything is well.
@benoitf @l0rd Can you take a look again?
@musienko-maxim Please schedule QA cycle for this PR.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -153,7 +157,8 @@ public MachineProviderImpl(DockerConnector docker,
                                @Named("che.docker.cpu_period") long cpuPeriod,
                                @Named("che.docker.cpu_quota") long cpuQuota,
                                WindowsPathEscaper windowsPathEscaper,
-                               @Named("che.docker.extra_hosts") Set<Set<String>> additionalHosts)
+                               @Named("che.docker.extra_hosts") Set<Set<String>> additionalHosts,
+                               @Nullable @Named("che.docker.user_id") String cheUserId)
```
benoitf: maybe rename variable from cheUserId to cheDockerUserId

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
benoitf: could we get home directory ?
why are we not using $HOME or getent passwd with user

benoitf (changes requested):

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
l0rd: @benoitf do you agree that $HOME does not correspond to /home/user here right? By the way `/home/user/`

l0rd (approved):

l0rd:
We are good to go for me. Thanks for your patience @snjeza!

benoitf:
@l0rd about the recipe you're referencing
https://github.com/eclipse/che-dockerfiles/blob/master/recipes/ubuntu_jdk8/Dockerfile#L8

I still see that /home/user is defined as being the home folder so getting this entry will work instead of hardcoded /home/user

l0rd:
@benoitf Since the container is started using a specific UID, $HOME should be the home of the corresponding user. Maybe with some code it will be more clear:
```bash
docker run -ti \ 
          --entrypoint="sh" \
          -u 1001 \
          -v /etc/group:/etc/group:ro \
          codenvy/ubuntu_jdk8
$ echo $HOME
/home/docker
```

benoitf:
@l0rd how about
```
$ find / -type d -user dockremap -prune
/home/user

```

l0rd:
@benoitf didn't know the existence of that dockremap user. What is it for? Is it a v1.13 feature? 

By the way that command didn't work on linux with docker 1.12 

```
docker run -ti \ 
          --entrypoint="sh" \
          -u 1001 \
          -v /etc/group:/etc/group:ro \
          codenvy/ubuntu_jdk8
$ find / -type d -user dockremap -prune
find: ‘dockremap’ is not the name of a known user
```

benoitf:
OK :-/ replacing userid by uid 1000 should work but then it's not so good

benoitf:
also is `/etc/subuid` and `/etc/subgid` files may help to get users defined in the image

snjeza:
> could we get home directory ?
why are we not using $HOME or getent passwd with user

If che starts as another user ("test", for instance), it won't work. /home/user is the home folder for the default workspace user.

benoitf:
@snjeza yes I've understood this is why I've asked alternatives on find /subuid, etc

benoitf:
@snjeza also chown on a directory that does not exists will display an error

TylerJewell:
As we are getting close to the discussion being finalized, can we start to think a little bit about how to support this and the documentation that we (or I) should write in support of this.  I think that this is the right direction to go, but the concern here is supportability of the unknown.  None of us have a really good grasp of the potential support issues that could appear as a result of this, so understanding where things can go wrong will be helpful.

If we can start to answer these questions, I'll make a PR in the docs to capture some of these concepts.
1. Can Che itself be started as a different user?  If so, what is the default user that eclipse/che-server starts as, and then how would one change it? 
2. What are the advantages or disadvantages to running the che server with a different user?  I think the answer to this one is that currently we write all of the files to :/data as root, and this would change that to be the user set when Che starts.
3. Then, how does a user of Che specify which user their workspace will start as?  Is this configurable per workspace or is it system-wide?

snjeza:
> also chown on a directory that does not exists will display an error

I have already commented it. See https://github.com/eclipse/che/pull/3376#issuecomment-271357087

benoitf:
@snjeza at the link I only see

> And there is a question about the commands executed in the container (chown and ln -s) that hasn't been answered.
I have already answered it.

but my point is that I don't see that we check if directory exists before performing the chown.

snjeza:
> Can Che itself be started as a different user? If so, what is the default user that eclipse/che-server starts as, and then how would one change it?

Yes, it can. Root is the default user.
In order to start che as another user, you have to do the following:
```
docker run ...\
 -u `id -u <username>`:`getent group docker | cut -d: -f3`\
 -e CHE_USER=<username> \
 -v /etc/group:/etc/group:ro \
 -v /etc/passwd:/etc/passwd:ro  \
...
```
In order to start a workspace, the <username> user has to be a member of the docker group and have sudo nopasswd access. 


>    What are the advantages or disadvantages to running the che server with a different user? I think the answer to this one is that currently we write all of the files to :/data as root, and this would change that to be the user set when Che starts.
>    Then, how does a user of Che specify which user their workspace will start as? Is this configurable per workspace or is it system-wide?

Currently, a user defined with CHE_USER will start a workspace. If CHE_USER isn't set, a user will be defined with an image. We could introduce a separate property/env variable if we want different users to start che and workspaces.

TylerJewell:
Ok.  So the syntax that you are using `CHE_USER=` would work only for the `eclipse/che-server` image if you directly launch it bypassing the CLI.  Have you tested using the CLI launcher with setting a value in `che.env` to pass along?  I would think that we would want two values:
```
# Sets the value of what eclipse/che-server docker container will be launched as a user
# Default is root
#CHE_SERVER_USER=root

# Sets the value of what user workspaces will be started as
# Default = user defined within stack image definition
# Setting this value overrides user defined within a stack - use with caution
#CHE_WORKSPACE_USER=root
```

Per a discussion with Mario, we would like to only have a single value to start - if the Che server is set to a particular user, then all workspaces will have the same user.  This will keep the syntax simpler to start.

snjeza:
@TylerJewell I will try

skabashnyuk:
Note: the person who is going to merge this pr in master have to run all tests including Selenium for che and codenvy.

snjeza:
@benoitf @TylerJewell 
I have updated the patch to test for existing the /home/user folder.
Still have a problem to start che using cli. I have tried the following:
```
docker run --rm \
 -u `id -u test`:`getent group docker | cut -d: -f3` \
 -e CHE_USER=test \
 -v /tmp/data:/data \
 -v /etc/group:/etc/group:ro \
 eclipse/che start
```
che returns the following:
```
WARN: Bound 'eclipse/che' to 'eclipse/che:5.1.2'
WARN: Did not detect TTY - interactive mode disabled
/scripts/base/paths.sh: line 26: /data/test: Permission denied
ERROR: Unable to write files to your host
ERROR: Have you enabled Docker to allow mounting host directories?
ERROR: Did you give our CLI rights to create files on your host?
```
and 
```
docker run --rm \
 -u 1001 \
 -e CHE_USER=test \
 -v /tmp/data:/data \
 -v /etc/group:/etc/group:ro \
 eclipse/che start
```
che returns
```
INFO: Welcome to Eclipse Che!
INFO: 
INFO: You are missing a mandatory parameter:
INFO:    1. Mount 'docker.sock' for accessing Docker with unix sockets.
INFO:    2. Or, set DOCKER_HOST to Docker's daemon location (unix or tcp).
INFO: 
INFO: Mount Syntax:
INFO:    Start with 'docker run -it --rm -v /var/run/docker.sock:/var/run/docker.sock' ...
INFO: 
INFO: DOCKER_HOST Syntax:
INFO:    Start with 'docker run -it --rm -e DOCKER_HOST=<daemon-location> ...'
```

Che cli doesn't parse the command line properly when there isn't the -u option.

TylerJewell:
Hi @snjeza.

First, the user should *not* provide `-e CHE_USER=<value>` on the command line. The only parameters that we will ever allow on the command line is going to be CHE_HOST.  All other parameters must be configured and set within the `che.env` file.  

Second, the `che.env` file is passed into a special image that uses Puppet to take those configuration values and then generate a custom `docker-compose.yml` file, which is used to actually start and stop Che with the `eclipse/che-server` image.  Think of it as a form of image chaining.

So, your command line is trying to get the CLI to inherit certain properties about parameters and user ID, which the CLI itself does not need to inherit.  Instead, we need to modify things so that the configuration in the `che.env` is passed into Puppet and then it generates a valid `docker-compose.yml`.

The items that you need:
1. Update the `che.env` template that is installed into the user's `:/data` mount when you do a first `init`.   This file lives in the `eclipse/che-init` image, and it's part of a [puppet manifest here](  https://github.com/eclipse/che/blob/master/dockerfiles/init/manifests/che.env).

2. Then, these values from `che.env` are passed into a puppet image and translated into [puppet variables in this file](https://github.com/eclipse/che/blob/master/dockerfiles/init/manifests/che.pp)

3. Then, these puppet variables are available to be applied to templates.  Puppet takes the templates and generates actual instance files in the user's `/data/instance` mount.  The template for the Docker compose that we use to actually start Che is here:  https://github.com/eclipse/che/blob/master/dockerfiles/init/modules/compose/templates/docker-compose.yml.erb.  You can see that there is a service that launches the actual Che server.  And here is where you would apply the value that was set in the `che.env` file.

Now, once you start messing with files in the `eclipse/che-init`, you have to rebuild your images. Build `dockerfiles/base`, followed by `dockerfiles/init`, followed by `dockerfiles/cli`. This will generate a CLI image locally that has all of those changes for you to test.

You can skip this and make life easier for you by just running the CLI with `-v <che-git-repo-path>:/repo`.  When you do this, the CLI will use the files for `dockerfiles/init` from your hard drive instead of those in the image, so you can test these modifications without building new images.

TylerJewell:
@snjeza - I also meant to say, that if all of that was a bit complicated to follow, I have another option for you, too.  If you run Che normally, in the `/data/instance` folder will be a `docker-compose.yml` file.  If you can modify this file to reflect how you want Che to start with the applied user, and then verify that it works, it's very easy for me to take those improvements and then apply the templating concepts and modifications to the che.env.  That may be a faster way to closure.

snjeza:
@TylerJewell 
You have to add user to che.env
```
CHE_USER=test
```
and use the following docker-compose file
```
che:
  image: eclipse/che-server:5.1.2
  env_file:
    - '/tmp/data/instance/config/che.env'
  volumes:
    - '/var/run/docker.sock:/var/run/docker.sock'
    - '/tmp/data/instance/data:/data'
    - '/tmp/data/instance/logs:/logs'
    - '/tmp/data/instance/config:/conf'
    - '/etc/group:/etc/group:ro'
  ports:
    - '8080:8080'
  restart: always
  container_name: che
  user: 1001:1001
```

1001:1001 is the user id of the test user and the group id of the docker group. 
You can use the following command to get such a line:
```
echo -n user: `id -u test`:`getent group docker | cut -d: -f3`
```

snjeza:
@TylerJewell 
I am using the following commands on fedora:

```
$ docker --version
Docker version 1.13.0, build 49bf474
$ docker-compose --version
docker-compose version 1.9.0, build 2585387
$ grep CHE_USER che.env 
CHE_USER=test
$ echo  user: `id -u test`:`getent group docker | cut -d: -f3`
user: 1001:1001
$ grep test /etc/passwd
test:x:1001:1002::/home/test:/bin/bash
$ grep docker /etc/group
docker:x:1001:test,snjeza
$ cat docker-compose.yml 
che:
  image: eclipse/che-server:5.1.2
  env_file:
    - 'che.env'
  volumes:
    - '/var/run/docker.sock:/var/run/docker.sock'
    - '/tmp/data/instance/data:/data'
    - '/tmp/data/instance/logs:/logs'
    - '/tmp/data/instance/config:/conf'
    - '/etc/group:/etc/group:ro'
  ports:
    - '8080:8080'
  restart: always
  container_name: che
  user: 1001:1001

$ docker-compose up
Starting che
Attaching to che
che    | Using embedded assembly...
che    | Using embedded che.properties... Copying template to /tmp/data/instance/data/conf.
...
che    | 2017-02-01 17:03:19,856[main]             [INFO ] [o.a.c.http11.Http11NioProtocol 488]  - Starting ProtocolHandler ["http-nio-8080"]
che    | 2017-02-01 17:03:19,870[main]             [INFO ] [o.a.catalina.startup.Catalina 642]   - Server startup in 14142 ms
...

$ docker exec che id
uid=1001(test) gid=1001(docker)

```

TylerJewell:
Thanks - the template and syntax is clear.  Not sure when I will get to the CLI improvements to support this.  So you may beat me to it.  But the work to do is clear now.

TylerJewell:
@snjeza - I have a prototype partially working, but need to wait for this PR to get merged before submitting a new one.  We are refactoring the CLI for maintainability.  https://github.com/eclipse/che/pull/4006

TylerJewell:
@snjeza - well, my prototype is not working. I had it working on the host, but when I started the work to port it to run within the CLI, I am running into some challenges.  When we run an `id` command from within the CLI, that is returning the CLI's container's `id`, but not the one from the host.  Do you have any suggestions on how we can get the host's id that is provided by the user from wtihin a docker container?

Some things:
1. We could require the user to add a `--user` flag to the CLI's docker run syntax. However, this would require the end user admin to discover the uid / gid of the user he wants to use on his host to pass into the container.
2. If we require admins to discover their uid / gid on the host, then the syntax is going to be different for linux, mac, and windows.  While `id` will work on a windows host, `getent` will not.

snjeza:
@TylerJewell 

>  well, my prototype is not working. I had it working on the host, but when I started the work to port it to run within the CLI, I am running into some challenges. When we run an id command from within the CLI, that is returning the CLI's container's id, but not the one from the host. Do you have any suggestions on how we can get the host's id that is provided by the user from wtihin a docker container?

If you mount /etc/passwd and /etc/group, the host and a docker container will have the same user/group id and name.
In the example I have mentioned above, 1001 is the id of the "test" user from the host and a container.
entrypoint.sh sets the CHE_HOST_ID env variable that contains the user's id:gid based on the CHE_HOST env variable

> If we require admins to discover their uid / gid on the host, then the syntax is going to be different for linux, mac, and windows. While id will work on a windows host, getent will not.

Docker doesn't recognize users/groups' names from the host's /etc/passwd and /etc/group. I suppose docker sets a user before mounting volumes. id/getent is a workaround. We should find or create a similar program/script on Windows. 
I have tested the PR only on Fedora 24.

TylerJewell:
@snjeza - I am not sure I am following the sequence of what you are proposing.  Let's make sure that we are discussing the same goals.

1. We have a single `docker run ...` syntax for starting / stopping Che which is identical for all OS.
2. Inside of the che.env file, someone could set a CHE_USER=user.
3. We need a way for the CLI container to obtain the host's uid/gid, preferrably without modifying the syntax of the CLI container `docker run ...` syntax.  

I can see how mounting /etc/passwrd and /etc/group works for native Linux executions of the CLI, but they would not work for Docker for Windows, Mac, or Boot2docker, as you'd just be mounting the VM's identity, and not grabbing the uid / gid from the host itself.  So my thinkin was that we could make use of `docker run -u <uid:gid>` as a way to pass in the host's gid / uid into the CLI container. 

I suppose that in this situation, it would not be necessary for `che.env` to have `CHE_USER` becuase the CLI has it set on the command line.  The CLI would just have to forward the identity properly into the launch of eclipse/che-server.  

But if we provided only s upport for `-u <uid>:<gid>` on the command line of the CLI, then it would turn into a docs issue where we have instructions for how to identify the uid and gid for each OS in situations where you wanted to run as a different user ... or we just say that this is a Linux-only feature.

snjeza:
> So my thinkin was that we could make use of docker run -u <uid:gid> as a way to pass in the host's gid / uid into the CLI container.

Can we add "<uid:gid>" to a container as an env variable (CHE_USER_ID, for instance)?

> ...or we just say that this is a Linux-only feature.

@l0rd 
Could you please answer this question?

TylerJewell:
Even with that approach, it still requires the end user to first determine their host uid gid. I think that is disappointing but may be the only option.

l0rd:
Running as a particular user has sense on Linux hosts. That's where Che running as root can mess with files permissions or can be a concern for security reasons. Docker 
4 mac and win hasn't the same problems. For this reason I think that making it a Linux only feature is ok.

TylerJewell:
Ok, this probably does make things easier if we assume that Docker is running natively on a linux host.  So this changes the configuration to be:
1. Inside the CLI, detect the -u uid:gid parameter that is passed in.
2. During any eclipse/che-server invocation, pass the same parameters.

This would not, then, require a `CHE_USER=` parameter in the che.env. I don't think we would do any sort of tests around the uid / gid - we would assume that the values passed in are the appropriate ones to pass through.

TylerJewell:
@snjeza @l0rd - I have opened a PR which will change the CLI to support this syntax. 
https://github.com/eclipse/che/pull/4050

You can check out this branch and then to test it on your linux systems:
1. cd /dockerfiles/init && bash build.sh
2. cd /dockerfiles/base && bash build.sh
3. cd /dockerfiles/cli && bash build.sh
```
docker run -it --rm 
  -v /var/run/docker.sock:/var/run/docker.sock 
  -v /c/tmp:/data 
  --user <uid:gid> 
    eclipse/che-cli:nightly config --skip:nightly --skip:pull
```
After running this, if the user is not a root user passed in, then the /instance/docker-compose.yml file will have the additional user option provided with the value that you have set.  We do not perform any checks to see if your host is linux or another variant.

So, this CLI PR will need to be merged at the same time as this PR.

1. Please revolve branch conflicts.
2. Please test CLI branch thoroughly on your linux systems
3. Please add Changelog and Release Notes entry to the main posting of this PR.  The release notes would be the same as the marketing content that we would send out to all users when we make the release.

Once these are done, we will be ready to merge, set milestone and include in upcoming release.

sunix:
hi @snjeza, could you rebase/fix conflicts with master ?

snjeza:
@sunix I have rebased the patch.

sunix:
Thanks @snjeza will review and test this week

snjeza:
@sunix 
You can test this PR in the following way:

```
# build che with the PR
$ git fetch origin pull/3376/head:che-3359
$ git checkout che-3359
$ mvn clean install
$ cd dockerfiles/che
$ ./build.sh
# select a user; the user must have the sudo nopasswd privilege and must be a member of the docker group
# I have created the "test" user with userid=1001; docker gid=1001
$ grep test /etc/passwd
test:x:1001:1002::/home/test:/bin/bash
$ grep docker /etc/group
docker:x:1001:test,snjeza
$ sudo chown -R 1001:1001 /tmp/data/
# run che as a user; 
$ docker run -it \
 --user 1001:1001 --rm \
 -v /tmp/data:/data \
 -v /etc/passwd:/etc/passwd\
 -v /etc/group:/etc/group \
 eclipse/che:nightly start --fast
 
# The --fast option skips pulling an image when it exists, i.e., docker will use a local eclipse/che-server:nightly image
# open the che dashborad and create a workspace
# check that the che server and the workspace start as a user with id=1001
$ docker exec -it che id
uid=1001(test) gid=1001(docker)
$ docker exec -it <workspace_container_id> id
uid=1001(test) gid=1001(docker) groups=1001(docker)
```

You have to use Linux to test this feature.

TylerJewell:
@snjeza - you should not have to use `CHE_USER`.  What happens if you drop the `--fast` parameter for bootup?

snjeza:
@TylerJewell che starts as the root user

```
$ docker run -it \
  --user 1001:1001 --rm \
  -v /tmp/data:/data \
  -v /etc/passwd:/etc/passwd \
  -v /etc/group:/etc/group \
  eclipse/che:nightly start

INFO: (che cli): nightly - using docker 17.03.0-ce / native
INFO: (che download): Pulling image eclipse/che:nightly

nightly: Pulling from eclipse/che
709515475419: Already exists 
e50c49019c4a: Pull complete 
56c2701a8f06: Pull complete 
9bcb79f184df: Pull complete 
9241a5c9b8a9: Pull complete 
7c0086b14556: Pull complete 
188b1f780e80: Pull complete 
1765d0e493f3: Pull complete 
Digest: sha256:17d4458e8219a9ca3cd3d141fde2ba1feb91be2f27a8d1232a3946714e347003
Status: Downloaded newer image for eclipse/che:nightly

WARN: Pulled new 'nightly' image - please rerun CLI

$ docker run -it \
  --user 1001:1001 --rm \
  -v /tmp/data:/data \
  -v /etc/passwd:/etc/passwd \
  -v /etc/group:/etc/group \
  eclipse/che:nightly start

INFO: (che cli): nightly - using docker 17.03.0-ce / native
INFO: (che download): Pulling image eclipse/che:nightly

nightly: Pulling from eclipse/che
Digest: sha256:17d4458e8219a9ca3cd3d141fde2ba1feb91be2f27a8d1232a3946714e347003
Status: Image is up to date for eclipse/che:nightly

WARN: (che init): 'nightly' installations cannot be upgraded to non-nightly versions
INFO: (che init): Installing configuration and bootstrap variables:
INFO: (che init):   CHE_HOST=172.20.0.1
INFO: (che init):   CHE_VERSION=nightly
INFO: (che init):   CHE_CONFIG=/tmp/data
INFO: (che init):   CHE_INSTANCE=/tmp/data/instance
INFO: (che config): Customizing docker-compose for running in a container
         mem (1.5 GiB):           [OK]
         disk (100 MB):           [OK]
         port 8080 (http):        [AVAILABLE]
         conn (browser => ws):    [OK]
         conn (server => ws):     [OK]

INFO: (che start): Starting containers...
INFO: (che start): Services booting...
INFO: (che start): Server logs at "docker logs -f che"
INFO: (che start): Booted and reachable
INFO: (che start): Ver: nightly
INFO: (che start): Use: http://172.20.0.1:8080
INFO: (che start): API: http://172.20.0.1:8080/swagger

$ docker exec -it che id
uid=0(root) gid=0(root) groups=0(root),1(bin),2(daemon),3(sys),4(adm),6(disk),10(wheel),11(floppy),20(dialout),26(tape),27(video)
```

TylerJewell:
oh my god - there was a refactoring of the CLI and the `check_user` method was not included.  **head slap**. This method does one thing, and that is to extract `--user` value and assign it to `CHE_USER`.

TylerJewell:
@snjeza - here is the PR that fixes that little mistake.  The function was always in the CLI but in the refactoring it was no longer being invoked.  https://github.com/eclipse/che/pull/4350

garagatyi:
@snjeza @l0rd is this PR still actual?

l0rd:
@garagatyi I think that this PR is useful and we can have someone working on it to rebase it. But it would be better to rebase it on top of the SPI branch, not master. What do you think?

garagatyi:
@l0rd On which infra do you want to support it?

l0rd:
@garagatyi in OpenShift this is automatically handled by the infra (by default user id is choosen arbitrarly) so I guess that Docker infra is missing that at the moment.

garagatyi:
I'm not sure what for it is needed in case of Docker infrastructure. And the solution is tricky which may lead to unexpected behavior of software in containers

skabashnyuk:
closing. There is no interest in this pr.

PR #3393, opened 2016-12-14
Change OpenShiftConnector to subclass of DockerConnector

amisevsk:
### What does this PR do?
Makes OpenShiftConnector a subclass of DockerConnector, and adds DockerConnectorProvider class to allow injection where needed. 

DockerConnector and OpenShiftConnector are bound to a MapBinder in LocalDockerModule, and injected into DockerConnectorProvider, along with the value of new property `che.docker_connector.provider`. This change required replacing DockerConnector with DockerConnectorProvider in constructors of all classes that use DockerConnector directly, but allows OpenShiftConnector to be used selectively without breaking normal functionality. With the changes, all tests cases pass, and when `che.docker_connector.provider` is set to 'default', Che behaves as before.

@l0rd @ibuziuk WDYT?

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
@@ -225,3 +225,5 @@ che.openshift.username=openshift-dev
 che.openshift.password=devel
 che.openshift.project=eclipse-che
 che.openshift.serviceaccountname=cheserviceaccount
+
+che.docker_connector.provider=openshift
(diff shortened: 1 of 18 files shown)


assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -226,4 +226,9 @@ che.openshift.password=devel
 che.openshift.project=eclipse-che
 che.openshift.serviceaccountname=cheserviceaccount
 
+# Which implementation of DockerConnector to use in managing containers. In general,
+# the base implementation of DockerConnector is appropriate, but OpenShiftConnector
+# is necessary for deploying Che on OpenShift. Options:
+#     - 'default'   : Use DockerConnector
+#     - 'openshift' : use OpenShiftConnector
 che.docker_connector.provider=openshift
```
TylerJewell: How about che.docker.connector=<value>

amisevsk: @TylerJewell That is a lot better, thank you. I was having trouble coming up with a good name, for some reason.

On this topic, do you think it's important to include this field in `che.env` for the new CLI? I'm inclined to think that this is too specialized and would just add clutter, but I'm not fully familiar with the new CLI and which use cases it's intended for.

TylerJewell:
I think we can leave it out of the new environment file as it would not be something the everyday admin would need to configure.

PR #3402, opened 2016-12-15
Update Bitnami stacks to latest minideb version

amrecio:
### What does this PR do?

Updates the base image of the Bitnami Stacks (minideb) to its latest versions.

It doesn't reference an issue.

### New behavior
Base image of the Bitnami Stacks updated to latest version.

### DIFF
--- a/ide/che-core-ide-stacks/src/main/resources/stacks.json
+++ b/ide/che-core-ide-stacks/src/main/resources/stacks.json
@@ -1475,7 +1475,7 @@
     ],
     "source": {
       "type": "image",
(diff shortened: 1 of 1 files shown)


slemeur (approved):
OK for me, but we would need QA to test the stacks.

@musienko-maxim: Could you please organize tests on the stacks?

musienko-maxim:
I have checked the stacks on che in the Docker. All workspaces from the Bitnami Stacks have been launched without errors.

ashumilova (approved):

Issue #3369, opened 2016-12-13
Git history panel is not opened (regression)


artaleks9:
I try to open the git history panel, but it is not opened.

**Reproduction Steps:**
* Open a project with git repository in the IDE.
* Go to 'Git' > 'Show History..' 

**Expected behavior:**
* Should be opened the git history panel

**Observed behavior:**
* Git history panel is not opened.

**Che version:**  5.0.0-M9    
**OS and version:** Ubuntu 14.04   
**Docker version:** 1.12   
**Che install:**   Docker container   

**Additional information:**
* Problem can be reliably reproduced, doesn't happen randomly: [Yes]
* See the attachment:
![git_history_panel](https://cloud.githubusercontent.com/assets/5408906/21144144/3ea1cbda-c152-11e6-92cd-176c304109a9.gif)

vitaliy-guliy:
Spent 1d 4h.

PR #3414, opened 2016-12-16
CHE-3369 Git history panel is not opened

vitaliy-guliy:
Fix appearing of the git history panel when the part is hidden.
Fix appearing of the git history panel after restoring the IDE state.
Fix the behavior of any hidden panel after restoring the IDE state.

https://github.com/eclipse/che/issues/3369


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/PartStackPresenter.java
```diff
@@ -145,14 +145,8 @@ public void addPart(@NotNull PartPresenter part) {
...
-            workBenchPartController.setHidden(true);
-
-            TabItem selectedItem = getTabByPart(part);
-
-            if (selectedItem != null) {
-                selectedItem.unSelect();
-            }
-
+            TabItem tab = getTabByPart(part);
+            onTabClicked(tab);
```
benoitf, line 149: tab is always non null ?
because signature of onTabClicked is requiring notNull but getTabByPart can return null
  public void onTabClicked(@NotNull TabItem selectedTab) {

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/workspace/perspectives/general/AbstractPerspective.java
```diff
@@ -373,46 +364,44 @@ public void loadState(@NotNull JsonObject state) {
...
-                activeParts.add(provider.get());
+        if (partStackJSON.hasKey("HIDDEN") && partStackJSON.getBoolean("HIDDEN")) {
+            partStack.minimize();
+            return;
+        }
+
+        if (partStackJSON.hasKey("SIZE")) {
+            double size = partStackJSON.getNumber("SIZE");
+
+            // Size of the part must not be less 100 pixels.
```
benoitf, line 399: comment is for <100 but code is == 0 ?

ashumilova (approved):

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/PartStackPresenter.java
vitaliy-guliy:
```
        if (containsPart(part)) {
            TabItem tab = getTabByPart(part);
            onTabClicked(tab);
            return;
        }
```
Yes, the case is confuses a little bit. But here **tab** must be not **null** because of there is a verification before.

vparfonov (approved):

PR #3428, opened 2016-12-19
Adapt tests for perspectives

svor:
### What does this PR do?
Sets correct value for testing perspectives


vitaliy-guliy (approved):
Thanks

vparfonov (approved):

PR #3436, opened 2016-12-19
CODENVY-1337 Logs of workspace containers and agents are not shown when WS starts in IDE

vitaliy-guliy:
The part of logs are lost and not shown in the dev-machine console.

https://github.com/codenvy/codenvy/issues/1337


vparfonov (approved):

benoitf (changes requested):
the branch needs to be rebased first due to history rewrite
If you need help, I can assist you

PR #3455, opened 2016-12-21
Added macros for product name

bmicklea:
Allows products that rebrand Che to easily reuse Che docs and have their own product name substituted for Che's.


benoitf (approved):
I suppose then we don't go with custom plugin ? 
https://github.com/codenvy/codenvy/pull/1367
because there wasn't any new feedback on it

JamesDrummond (approved):

Issue #2866, opened 2016-10-24
Project import gets org.eclipse.jdt.core.JavaModelException


KakashiMamun:
[Project import gets org.eclipse.jdt.core.JavaModelException]

**Reproduction Steps:**

I have a project with these modles

-     <module>scheduler-build</module>
-     <module>scheduler-typelayout</module>
-     <module>scheduler-dto</module>
-     <module>scheduler-biz</module>
-     <module>scheduler-bizcore</module>
-     <module>scheduler-front/aggregator-pom.xml</module>
-     <module>scheduler-batch</module>
-     <module>scheduler-webapi</module>
-     <module>scheduler-webapi-spec</module>
-     <module>scheduler-webapi-client</module>

As you can see the 'front' project have an aggregator pom rather than a pom xml. the aggregator pom
has these modules

-     <module>./client-side-pom.xml</module>
-     <module>./pom.xml</module>

when i import my project to che it gives me the following error

org.eclipse.jdt.core.JavaModelException: Can't find parent folder: /hue-scheduler/hue-scheduler-front/aggregator-pom.xml/.codenvy
    at org.eclipse.jdt.internal.core.JavaProject.writeFileEntries(JavaProject.java:2023) ~[org.eclipse.jdt.ui-5.0.0-M6-SNAPSHOT.jar:5.0.0-M6-SNAPSHOT]
    at org.eclipse.jdt.internal.core.JavaModelManager$PerProjectInfo.writeAndCacheClasspath(JavaModelManager.java:2789) ~[org.eclipse.jdt.ui-5.0.0-M6-SNAPSHOT.jar:5.0.0-M6-SNAPSHOT]
    at org.eclipse.jdt.internal.core.JavaModelManager$PerProjectInfo.writeAndCacheClasspath(JavaModelManager.java:2802) ~[org.eclipse.jdt.ui-5.0.0-M6-SNAPSHOT.jar:5.0.0-M6-SNAPSHOT]
    at org.eclipse.jdt.internal.core.SetClasspathOperation.executeOperation(SetClasspathOperation.java:77) ~[org-eclipse-jdt-core-repack-5.0.0-M6-SNAPSHOT.jar:5.0.0-M6-SNAPSHOT]
    at org.eclipse.jdt.internal.core.JavaModelOperation.run(JavaModelOperation.java:808) ~[org.eclipse.jdt.ui-5.0.0-M6-SNAPSHOT.jar:5.0.0-M6-SNAPSHOT]
[...]

PR #3457, opened 2016-12-21
#2866 fix importing maven project with relative path in modules

evidolob:
### What does this PR do?
handle relative path and including 'pom.xml in modules declarations
#2866


plugins/plugin-maven/che-plugin-maven-server/src/main/java/org/eclipse/che/plugin/maven/server/core/project/PathUtil.java
```diff
@@ -0,0 +1,157 @@
...
+import org.apache.commons.io.output.NullWriter;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.function.Supplier;
+
+/**
+ * Set of util methods for path strings.
+ */
+public class PathUtil {
```
azatsarynnyy: I think that [che-core-commons-lang](https://github.com/eclipse/che/tree/master/core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/os) is better place for this util class.
Especially since we already have a special package **org.eclipse.che.commons.lang.os** there for the platform-specific utils.

vparfonov (approved):

PR #3473, opened 2016-12-22
Docs maven build.

JamesDrummond:
The PR adds maven build to docs. I moved all files to docs/src/main/ similar to the other folders so that the target build directory would not conflict with maven build. I currently use git clone to get images from my repo https://github.com/JamesDrummond/che-doc-images until an official one is created so that build can be tested properly. We can eventually get the images to build into a jar file that can be downloaded from maven repo but for now this additional work that can be done later.

The pom.xml creates a war file of the static html for the tomcat server and a jar file with doc source that can be uploaded to maven repo for use with codenvy doc build.

Also, added `{% include base.html %}` to all markdown files to allow relative paths. This was not working correctly before.


benoitf (approved):

skabashnyuk (changes requested):
I don't like an idea to use docker as requirement to build che. Right now docker is an option to build che or codenvy. If we accept this PR it would be hard or close to impossible to build che in che or codenvy in codenvy. I think all actions that you are doing in docker can be done without it. Can you change you pr and use jekyll without docker

riuvshin:
totally agree with @skabashnyuk, we should never call docker during regular maven build. (we do this only for special profiles)

riuvshin (changes requested):
totally agree with @skabashnyuk, we should never call docker during regular maven build. (we do this only for special profiles)

bmicklea:
@JamesDrummond I know you're not a maven expert - is this something you can do or do you need help from someone on the team with more maven experience?  I don't want you doing this over the next two days if it's something that someone else could do in a couple of hours.

JamesDrummond:
che-in-che and codenvy-in-codenvy currently require docker in workspace machine to run the binaries on host daemon. Is there any other reasons this will not work?

TylerJewell:
Maven is our build system - we do not have a docker dependency for that.  Maven should always be native.

JamesDrummond:
This now builds without docker. However there are many things I probably should not be doing like git and exec copy command. I need to hand this off to the dev team now to fix what ever needs to be fixed. I have made the war and jar files that will need to be handled as well.

bmicklea:
@skabashnyuk and @riuvshin - who on the dev team can finish this off?  We need this completed as soon as possible.

skabashnyuk:
To finish what? 
Entire picture is not clear.

JamesDrummond:
We are putting docs into new repo. This PR will not be done.

PR #3480, opened 2016-12-22
CHE-3369 Git history panel is not opened

vitaliy-guliy:
Fixes appearing of Git history panel.

https://github.com/eclipse/che/issues/3369

### DIFF
--- a/plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/history/HistoryPresenter.java
+++ b/plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/history/HistoryPresenter.java
@@ -134,7 +134,7 @@ public class HistoryPresenter extends BasePresenter implements HistoryView.Actio
             isViewClosed = false;
         }
 
-        partStack.setActivePart(HistoryPresenter.this);
+        partStack.setActivePart(this);
     }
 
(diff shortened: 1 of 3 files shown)


ashumilova (approved):

vparfonov (approved):

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/history/HistoryPresenter.java
```diff
@@ -132,9 +132,9 @@ public void showDialog(Project project) {
         if (isViewClosed) {
             workspaceAgent.openPart(this, PartStackType.TOOLING);
             isViewClosed = false;
-        } else {
-            partStack.setActivePart(this);
         }
+
+        partStack.setActivePart(HistoryPresenter.this);
```
azatsarynnyy: it can be simplified to `partStack.setActivePart(this)`

vitaliy-guliy: Yes, thanks.
This line was in the inner class and I forgot to remove prefix `HistortyPresenter.`

vitaliy-guliy: Fixed!

Issue #3463, opened 2016-12-21
Duplicate entry in project explorer after import


dkulieshov:
When we import project we have two same entries in a project explorer window however after we refresh page it all looks fine.

Steps to reproduce:
1. Start java workspace
2. Import `che` project
3. Enjoy

Branch: https://github.com/eclipse/che/tree/dynamic-fw

PR #3486, opened 2016-12-22
Fix duplicating node during the project import

vzhukovs:
Don't add a new node with the same path to the tree if such node already exists.

Related issue: #3463


dkulieshov (approved):
Excellent

PR #3490, opened 2016-12-22
Added {% include base.html %} to all markdown. 

JamesDrummond:
Added {% include base.html %} to all markdown. Add license information to some files. Added docs/assets/images to .gitignore . Remove assembly as it should not be used for building. Only the _config.yml file should ever be used for building static html docs.

### What does this PR do?
The {{ base }} variable does not inherit from base html files. There we need to include the base.html in each markdown. It is included directly below the front matter so it should be easy for contributors to add for new pages.


bmicklea (approved):

slemeur (approved):

slemeur:
Seems your commits are not signed off - so the ip-validation is failing

PR #3521, opened 2016-12-26
CHE-2528: Start workspace after renaming of environment

mmorhun:
### What does this PR do?
When snapshot(s) is not found on workspace start, notifies a user with message in machine start log and starts machine from its recipe.

https://github.com/eclipse/che/issues/2528

### Previous behavior
Failed to start workspace after renaming environment with snapshot(s).

### New behavior
Workspace starts without snapshot(s) in case when snapshot(s) is not found.


mmorhun:
WIP

garagatyi (approved):
@evoevodin Please take a look

voievodin (approved):

PR #3530, opened 2016-12-27
CODENVY-1392: fix SSH agent start check

### LINKED ISSUE #1392: Workspace start occasionally fails if docker image doesn't contain ssh and rsync
Workspace start occasionally fails if docker image doesn't contain ssh and rsync. 

Related to:
#1271 
#1371 

**Reproduction Steps:**

Use bitnami stacks to create and start WS several times.

**Expected behavior:**

These WSs start successfully. 

**Observed behavior:**

Sometimes WS start fails and sometimes succeeds. 

**Codenvy version:** m9-snapshot
**OS and version:** Ubuntu 16.04

**Additional information:**

Problem only started happening recently, didn't happen in an older version of Codenvy: [Yes] (because of other reasons such WSs always failed start before merging solution for #1271)
Problem can be reliably reproduced: [No - it's random]

garagatyi:
### What does this PR do?
Chenge check of launch of SSH agent.
Add agent launching checker that checks if command exists
in machine.
Add agent launching checker that checks if port is listened
in machine.
Add agent launching checker that forward check to
several specified checkers.

Fixes codenvy/codenvy#1392

### Previous behavior
Sometimes SSH agent was treated as launched but immediate SSH into container failed.

### New behavior
When SSH agent is treated as launched immediate SSH into container works.

### DIFF
--- a/wsmaster/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/AbstractAgentLauncher.java
+++ b/wsmaster/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/AbstractAgentLauncher.java
@@ -97,7 +97,6 @@ public abstract class AbstractAgentLauncher implements AgentLauncher {
         throw new ServerException(errMsg);
     }
 
-
     protected InstanceProcess start(Instance machine, Agent agent) throws ServerException {
         Command command = new CommandImpl(agent.getId(), agent.getScript(), "agent");
         InstanceProcess process = machine.createProcess(command, null);
--- a/wsmaster/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/CommandExistsAgentChecker.java
+++ b/wsmaster/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/CommandExistsAgentChecker.java
@@ -1,0 +1,54 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2016 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.api.agent.server.launcher;
+
+import org.eclipse.che.api.agent.shared.model.Agent;
+import org.eclipse.che.api.core.ConflictException;
+import org.eclipse.che.api.core.model.machine.Command;
+import org.eclipse.che.api.core.util.ListLineConsumer;
+import org.eclipse.che.api.machine.server.exception.MachineException;
+import org.eclipse.che.api.machine.server.model.impl.CommandImpl;
+import org.eclipse.che.api.machine.server.spi.Instance;
+import org.eclipse.che.api.machine.server.spi.InstanceProcess;
+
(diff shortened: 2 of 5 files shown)


wsmaster/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/AbstractAgentLauncher.java
```diff
@@ -97,7 +97,6 @@ public void launch(Instance machine, Agent agent) throws ServerException {
         throw new ServerException(errMsg);
     }
 
-
```
skabashnyuk, line 100: Do you really need this change?

wsmaster/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/CommandExistsAgentChecker.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    public boolean isLaunched(Agent agent, InstanceProcess process, Instance machine) throws MachineException {
+        Command command = new CommandImpl(format("Wait for %s, try %d", agent.getId(), ++counter),
+                                          checkingCommand,
+                                          "test");
+
+        try (ListLineConsumer lineConsumer = new ListLineConsumer()) {
+            InstanceProcess waitProcess = machine.createProcess(command, null);
+            waitProcess.start(lineConsumer);
+            return lineConsumer.getText().endsWith("[STDOUT] 0");
```
skabashnyuk, line 49: Do we need to check locale of this command?

skabashnyuk (approved):

wsmaster/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/MappedPortIsListeningAgentChecker.java
```diff
@@ -0,0 +1,54 @@
...
+        // normalize port/transport value
+        this.exposedPort = exposedPort.contains("/") ? exposedPort : exposedPort + "/tcp";
+    }
+
+    @Override
+    public boolean isLaunched(Agent agent, InstanceProcess process, Instance machine) throws MachineException {
+        Server server = machine.getRuntime().getServers().get(exposedPort);
+        if (server != null) {
+            try {
+                LOG.error("Ping ssh agent");
```
tolusha: I think this is debug info.

tolusha (approved):

wsmaster/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/AbstractAgentLauncher.java
garagatyi: Why not?

wsmaster/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/MappedPortIsListeningAgentChecker.java
garagatyi: Yes, nice catch! Thank you.

wsmaster/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/CommandExistsAgentChecker.java
garagatyi: It works in the same way as ProcessIsLaunchedChecker. Why do you think locale matters here?

PR #3551, opened 2016-12-28
cli dev mode consistency

riuvshin:
after adding mounts for agents in dev mode we had inconsistent checks in CLI if those packages were not built. So I;ve simply moved check is api server is built from `docker.sh` to `cmd_config` and now we do that check exaclty before copy api server from target to `instance/dev` dir.

the sequence was like
1) check if API server is built
2) do pulling images + config
3) check if ws agent and terminal agent are built.
4) start

now sequence is more clear
1) do pull + config
2) check API, ws agent, terminal agent are built 
3) start

### DIFF
--- a/dockerfiles/base/scripts/base/commands/cmd_config.sh
+++ b/dockerfiles/base/scripts/base/commands/cmd_config.sh
@@ -36,19 +36,6 @@ cmd_config() {
     docker_run -v "${CHE_HOST_CONFIG}":/copy \
                -v "${CHE_HOST_DEVELOPMENT_REPO}"/dockerfiles/init:/files \
                   $IMAGE_INIT
-
-    # in development mode to avoid permissions issues we copy tomcat assembly to ${CHE_INSTANCE}
-    # if ${CHE_FORMAL_PRODUCT_NAME} development tomcat exist we remove it
-    if [[ -d "${CHE_CONTAINER_INSTANCE}/dev" ]]; then
-        log "docker_run -v \"${CHE_HOST_INSTANCE}/dev\":/root/dev alpine:3.4 sh -c \"rm -rf /root/dev/*\""
-        docker_run -v "${CHE_HOST_INSTANCE}/dev":/root/dev alpine:3.4 sh -c "rm -rf /root/dev/*"
-        log "rm -rf \"${CHE_HOST_INSTANCE}/dev\" >> \"${LOGS}\""
-        rm -rf "${CHE_CONTAINER_INSTANCE}/dev"
-    fi
-    # copy ${CHE_FORMAL_PRODUCT_NAME} development tomcat to ${CHE_INSTANCE} folder
-    mkdir -p "${CHE_CONTAINER_INSTANCE}/dev/${CHE_MINI_PRODUCT_NAME}-tomcat"
-    cp -r "$(echo $CHE_CONTAINER_DEVELOPMENT_REPO/$CHE_ASSEMBLY_IN_REPO)/." \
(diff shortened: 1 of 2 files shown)


garagatyi (approved):

TylerJewell:
I need to get into my laptop for a proper review, but I believe we are inverting the checks.  The original intent was to check for the existence of the assemblies and dependency files in bootstrap mode before any commands are called.  So I believe we have moved checks to the wrong location.

Also we should not be rewriting the version file on every config.  It is written only during init and becomes a permanent fixture of the installation and only changes during an upgrade.

dockerfiles/base/scripts/base/commands/cmd_config.sh
```diff
@@ -45,22 +57,18 @@ cmd_config() {
         log "rm -rf \"${CHE_HOST_INSTANCE}/dev\" >> \"${LOGS}\""
         rm -rf "${CHE_CONTAINER_INSTANCE}/dev"
     fi
+
+    if [[ ! -d $(echo ${CHE_CONTAINER_DEVELOPMENT_REPO}/${CHE_ASSEMBLY_IN_REPO}) ]]; then
```
TylerJewell, line 61: I would like us to move this into the docker.sh, so that all of the checks of the pre-reqs (in this case modules properly built) is handled before normal execution of a command.

In this method, we should then take actions that assume the presence of the pre-reqs.

dockerfiles/base/scripts/base/commands/cmd_config.sh
```diff
@@ -36,7 +36,19 @@ cmd_config() {
...
+
+  info "config" "Generating $CHE_MINI_PRODUCT_NAME configuration..."
+  # Run the docker configurator
+  generate_configuration_with_puppet
+
+  # Replace certain environment file lines with their container counterparts
+  info "config" "Customizing docker-compose for running in a container"
+
+  # Write the installed version to the *.ver file into the instance folder
+  echo "$CHE_VERSION" > "${CHE_CONTAINER_INSTANCE}/${CHE_VERSION_FILE}"
```
TylerJewell, line 49: This line should not be here. The only time we should ever write the version file is init(), or as part of a re-init, which is part of an upgrade.

The new CLI is structured so that once you have init() a folder, you cannot change the version using che.env. We removed that ability as it introduced user errors.

TylerJewell (changes requested):

TylerJewell:
Two changes requested.

1) It should be the case that we check that the modules for the server & agents exist in the docker.sh first.  And then we can go into normal cycle.

2) We should only write the version file in init()

TylerJewell:
This PR probably needs to be refactored after we merge the PR that breaks apart :/repo from --debug concept.

dockerfiles/base/scripts/base/commands/cmd_config.sh
riuvshin: I've moved that line within `cmd_config`. I don;t understand your concern because that line is  already in `cmd_config` I've just moved it before dev mode things which is more accurate and more consistent.

dockerfiles/base/scripts/base/commands/cmd_config.sh
riuvshin: this could not be in docker.sh because it is used by both che and codenvy where assemblies are different we need custom logic for both and if we will move it `docker.sh` it will be same for both which does not work.

riuvshin:
That PR does not change logic it change only sequence of things to be more consistent, this does not bock or blocked by anything.

TylerJewell:
I have incorporated all of these changes into this PR - https://github.com/eclipse/che/pull/3581.  You can close this now.

PR #3555, opened 2016-12-28
Minor dockerfiles fixes

l0rd:
Signed-off-by: Mario Loriedo <mloriedo@redhat.com>

### What does this PR do?
- Update Dockerfile.centos
- Remove parameter on all calls to `build` function (because it's not used)
- Avoid building twice `lib` folder in script `build-all.sh`
- Fix formatting in `cmd_debug.sh`


TylerJewell (approved):

benoitf:
I would keep arguments to build command as we may need it in the future if we add some enhancements

l0rd:
@benoitf I would add arguments to build command if/when build command will use them. Not before.

benoitf (approved):

Issue #3082, opened 2016-11-16
Reduce amount of calls on get workspaces API call 


garagatyi:
Now get workspaces API `/api/workspace` returns list of all users workspaces. Each workspace in this list contains runtime object. Each runtime object contains description of all docker machines in a workspace. This means that on each request to get list of workspaces we make big amount of requests to different backends (RDB, Docker API). 
We should remove runtime description from workspaces from this request. If client needs runtime description it should call get workspace API `/api/workspace/<ws-id>`.

PR #3557, opened 2016-12-28
Do not fetch runtimes when getting list of all workspaces

mkuznyetsov:
### What does this PR do?
Removes getting workspace runtimes info on 'get all workspaces' service

https://github.com/eclipse/che/issues/3082


garagatyi (approved):

voievodin (changes requested):
Returned workspaces must still contain correct status

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -536,6 +536,23 @@ public Instance getMachine(String workspaceId, String machineId) throws NotFound
...
     }
 
+    /**
+     * Return status of the workspace
+     *
+     * @param workspaceId
+     *         ID of requested workspace
+     * @return workspace status
+     */
+    public WorkspaceStatus getStatus(String workspaceId) {
```
garagatyi, line 546: Please add tests

garagatyi (changes requested):

mkuznyetsov:
Now returns workspace status

dashboard/src/app/projects/create-project/create-project.controller.ts
```diff
@@ -1046,7 +1046,7 @@ export class CreateProjectController {
    * @param workspace existing workspace
    */
   checkExistingWorkspaceState(workspace: any): void {
-    if (workspace.runtime) {
+    if (workspace.status === 'RUNNING') {
```
voievodin, line 1049:
```javascript
if (workspace.status === 'RUNNING')
```
is not the same to 
```javascript
if (workspace.runtime)
```
it looks like 
```javascript
if (workspace.status !== 'STOPPED')
```
is the correct option here

voievodin:
@ashumilova please review client-side changes

dashboard/src/app/projects/create-project/create-project.controller.ts
ashumilova: 'RUNNING' is fine here, cause wait for this status to start importing projects

garagatyi (approved):

ashumilova (approved):

voievodin (approved):

mkuznyetsov:
same logic done for WorkspaceService#getByNamespace()

Issue #3511, opened 2016-12-23
Delete docs folder


JamesDrummond:
Delete folder https://github.com/eclipse/che/tree/master/docs. Moved docs to dedicated repo.

PR #3580, opened 2016-12-30
Remove docs which are now located at che-docs.

JamesDrummond:
Signed-off-by: James Drummond jdrummond@codenvy.com


TylerJewell:
We should merge this after:
1. We are confident that we have staged the docs on eclipse.org/che/docs
2. We have also updated the che wiki, CONTRIBUTING.md, and CUSTOMIZING.md documents so that they are pointing to the appropriate repository for where documentation source is, and where the static HTML is at.

JamesDrummond:
Related issue https://github.com/eclipse/che/issues/3511

Issue #3546, opened 2016-12-28
Make CLI show che-server start logs in dev mode 


garagatyi:
Sometimes in development che-server doesn't start. It can happen if developer did mistake in configuration, etc.
In that case CLI waits che-server too long. It would improve Che developers experience if these logs were shown by default, so developer identify problem immediately after looking at errors in logs.

**Che version:**    5.0.0-Snapshot
**OS and version:** Ubuntu 16.04
**Docker version:** 12.04
**Che install:**    Docker container

TylerJewell:
This was requested and discussed in last week's team leaders meeting.  The general feeling was that it would. Real the container philosophy.  Roman was a -1 on it.  I was as well because it introduces blocking behavior when we don't want to.

I cannot remember who was the engineer that raised it for consideration.

garagatyi:
What do you mean blocking behavior? AFAIK CLI already waits server start, so what would be change in behavior?

TylerJewell:
Log output of Che server would have a tailing affect making a start operation that is non blocking blocking.  The feeling of the team leads was that we had the docker logs option there to show the outputs for you.  You can write a local script that starts Che and shows the log output at the same time in the background but mixing the two would deviate from the container philosophy.

garagatyi:
We already wait for deploy of API war. It is the longest operation in che-server boot sequence. And exactly that step sometimes fails because of developers mistakes. And this issue is intended to address that. So as far as I can see it can be implemented with no impacts on time of boostrapping and any change in blocking/non/blocking behavior.

garagatyi:
It is essential because in case developer mises some property Che won't start and CLI will not show anything. So developer has to wait some time after boot. Then consider if it was to long, if yes check logs. It is not convenient.
If you don't want to expose them by default let it be some flag (variable) that enables that behavior.

With native Che we didn't have such problem because it showed logs, so I consider lack of that feature as regression

TylerJewell:
So the requested logic would be:
```
If development_mode then
  if 'eclipse/che start -f' then
    during start, run 'docker logs -f che-server &'
  fi
  after successful boot, then kill the docker logs command
fi
```

@riuvshin needs to comment on whether this is viable, or breaking the design of the CLI. This is not a regression as this CLI was never intended to have this type of behavior. And I am not sure that it makes sense yet.

garagatyi:
As far as I understand current logic is:
 - run compose start 
 - wait until container is running
[...]

Issue #3542, opened 2016-12-28
Debug Che binaries within che-server without mounting /repo


garagatyi:
Sometimes I need to debug some version of Che or some branch but I don't have compiled sources of that state of Che. I would want to debug Che without compilation of sources. But now it is not implemented.

**Che version:**    5.0.0-Snapshot

* Problem started happening recently, didn't happen in an older version of Che: [No]
* Problem can be reliably reproduced, doesn't happen randomly: [Yes]

TylerJewell:
The cli requires that you run it in dev mode and that an assembly would exist on your local repo.

What are the circumstances where this doesn't work?

garagatyi:
I was in the middle of development of some feature and needed to make fast debugging session of master. I had to commit my changes, checkout to master, recompile master to debug. And then checkout my branch, recompile it to continue my work. With codenvy it is even worse because I have to do all of that for 2 repos.

TylerJewell:
That workflow is the expected workflow.  I am not sure I see what should be improved?

garagatyi:
Improvement is ability to debug che server without usage of local binaries

TylerJewell:
Where would the binaries come from?  There is not an image that has them - so you have to build those binaries anyways.

garagatyi:
We always have binaries in che-server image

TylerJewell:
Your specific scenario would not work then.  Because of you have to debug master then you always have to build binaries.  The images that are on dockerhub are only tagged images or the nightly, neither of which is correlated to master.

So you just want a way to debug an existing image if that tagged version has the binaries you want in it.

garagatyi:
yes, and nightly is what I need to debug from time to time. Another use-case is to debug some tag to confirm submitted bug.

TylerJewell:
So the objective is to be able to run development_mode on an existing `eclipse/che-server` Docker image without using the module or assembly that is on disk. Intead, using the binaries that are already packaged within the particular `eclipse/che-serve`r image that is referenced during the bootstrap phase.

I see two ways that this could be implemented:
[...]

PR #3581, opened 2016-12-30
CLI - Separate debugging apart from local repo

TylerJewell:
### What does this PR do?
This PR creates a separation of concern between the concept of using a local repository `:/repo` apart from activating development mode + debugging `--debug`. In the existing CLI, these two concepts are mixed together, so that anytime you mount a local repository, debugging of the che server is activated. By separating these two concerns, a user can now activate debugging for images within the shipped Docker image. Or, a user can also use local binaries in production mode.

1. Activates debugging for any command with the option `--debug`
2. Removes automatic setting of dev mode if `:/repo` is mounted.
3. If `:/repo` is not mounted AND `--debug` is activated, then debugs binaries shipping within eclipse/che-server 
4. Fixes bug where `CHE_DEBUG_PORT` was not being used in the `docker-compose.yml` template.
5. If `--debug` displays the Che server streaming output to the foreground, and then kills it after server is booted.
6. Enables usage of local binaries in production mode.
7. Fixes long term bug on windows where some directories are not deleted unless command rerun.

#3542
#3546

### Previous behavior
If `:/repo` was mounted, then dev mode was automatically activated. But there was no way to debug internal binaries. Now the only way to activate debug mode is with `--debug` added to any command. The `:/repo` is now optional, only if you want to debug your source repo.

### Testing
To test this, there are modifications to three images - you need to rebuild `dockerfiles/base`, `dockerfiles/cli`, and `dockerfiles/init`. Or you can run:
```
docker run --rm -it -v /var/run/docker.sock:/var/run/docker.sock \
        -v <PATH>:/data \
        -v <PATH-TO-CHE>/dockerfiles/base/scripts/base:/scripts/base \
        eclipse/che-cli:nightly start --debug
```

### Codenvy Updates
The Codenvy and ARTIK CLIs now inherit from the Che CLI. This could be breaking behavior for those inherited CLIs. Things to update:
1. `CHE_DEVELOPMENT_MODE` is no longer `on` or `off`. There is now a `CHE_DEBUG` which is `true` or `false`. Rather it's `development` or `production`. There is now a `debug_server()` method which will tell you if debugging is on.

2. You can now call `local_repo()` to find out if a local repository has been activated or not. In the puppet configuration of cmd_config(), special attention must be taken as we now have different syntax for four configurations - debug or no-debug and local repo or no-local-repo.

3.  In the puppet module for Docker compose, the debug port should be set if `CHE_DEBUG=true`. However, the local assembly should only be mounted to the container if `local_repo() = true`.  Before, the variable was meaning double - that it was for both dev mode and for using a repo. Now we have a hard separation.

When this is merged, also merge:
https://github.com/codenvy/che-docs/pull/19
https://github.com/codenvy/codenvy/issues/1458


TylerJewell:
Do not merge this until:
1. Codenvy has been tested with the modifications.
2. Docs have a matching pull request

benoitf (approved):

dockerfiles/base/scripts/base/commands/cmd_config.sh
```diff
@@ -30,8 +30,8 @@ cmd_config() {
   fi
 
   # Development mode
```
garagatyi: Does this comment still relevant?

dockerfiles/base/scripts/base/commands/cmd_config.sh
```diff
@@ -41,11 +41,17 @@ cmd_config() {
     # if ${CHE_FORMAL_PRODUCT_NAME} development tomcat exist we remove it
     if [[ -d "${CHE_CONTAINER_INSTANCE}/dev" ]]; then
         log "docker_run -v \"${CHE_HOST_INSTANCE}/dev\":/root/dev alpine:3.4 sh -c \"rm -rf /root/dev/*\""
-        docker_run -v "${CHE_HOST_INSTANCE}/dev":/root/dev alpine:3.4 sh -c "rm -rf /root/dev/*"
+
+        # Super weird bug - sometimes, the RM command doesn't wipe everything, so we have to repeat it a couple times
+        until directory_is_clean; do
+          docker_run -v "${CHE_HOST_INSTANCE}/dev":/root/dev alpine:3.4 sh -c "rm -rf /root/dev/${CHE_MINI_PRODUCT_NAME}-tomcat" > /dev/null 2>&1  || true
```
garagatyi, line 59: Why do you change path from /root/dev/* to /root/dev/${CHE_MINI_PRODUCT_NAME}-tomcat" ?

dockerfiles/base/scripts/base/commands/cmd_config.sh
```diff
@@ -76,7 +82,21 @@ generate_configuration_with_puppet() {
     CHE_ENV_FILE="${CHE_HOST_INSTANCE}/config/$CHE_MINI_PRODUCT_NAME.env"
   fi
 
-  if [ "${CHE_DEVELOPMENT_MODE}" = "on" ]; then
+  if debug_server; then
+    DEV_MODE="development"
```
garagatyi: CHE_ENVIRONMENT looks better than DEV_MODE

dockerfiles/base/scripts/base/commands/cmd_config.sh
```diff
@@ -86,33 +106,42 @@ generate_configuration_with_puppet() {
...
-                  -e \"CHE_DEVELOPMENT_REPO=${CHE_HOST_DEVELOPMENT_REPO}\" \
+                  -e \"CHE_REPO=${CHE_REPO}\" \
                   -e \"CHE_ASSEMBLY=${CHE_ASSEMBLY}\" \
                   --entrypoint=/usr/bin/puppet \
                       $IMAGE_INIT \
                           apply --modulepath \
                                 /etc/puppet/modules/ \
-                                /etc/puppet/manifests/${CHE_MINI_PRODUCT_NAME}.pp --show_diff"
+                                /etc/puppet/manifests/${CHE_MINI_PRODUCT_NAME}.pp --show_diff ${WRITE_LOGS}"
   else
```
garagatyi: It looks like both branches of if clause are equal now, WDYT?

dockerfiles/base/scripts/base/commands/cmd_config.sh
```diff
@@ -86,33 +106,42 @@ generate_configuration_with_puppet() {
...
   eval ${GENERATE_CONFIG_COMMAND}
 }
+
+directory_is_clean() {
+  if [[ -d "${CHE_CONTAINER_INSTANCE}/dev/${CHE_MINI_PRODUCT_NAME}-tomcat" ]]; then
+    return 1
+  else
+    return 0
+  fi
+}
```
garagatyi: Please add trailing new line at the end of file

dockerfiles/base/scripts/base/commands/cmd_destroy.sh
```diff
@@ -46,26 +46,14 @@ cmd_destroy() {
                                    && rm -rf /root${CHE_CONTAINER_ROOT}/instance \
                                    && rm -rf /root${CHE_CONTAINER_ROOT}/${CHE_MINI_PRODUCT_NAME}.env\""
 
-  docker_run -v "${CHE_HOST_CONFIG}":/root${CHE_CONTAINER_ROOT} \
-                alpine:3.4 sh -c "rm -rf /root${CHE_CONTAINER_ROOT}/docs \
-                               && rm -rf /root${CHE_CONTAINER_ROOT}/instance \
-                               && rm -rf /root${CHE_CONTAINER_ROOT}/${CHE_MINI_PRODUCT_NAME}.env" > /dev/null 2>&1  || true
-
   # Super weird bug.  For some reason on windows, this command has to be run 3x for everything
```
garagatyi, line 48: Does docker have an issue for that?

dockerfiles/base/scripts/base/commands/cmd_destroy.sh
```diff
@@ -46,26 +46,14 @@ cmd_destroy() {
...
-                    alpine:3.4 sh -c "rm -rf /root${CHE_CONTAINER_ROOT}/docs \
-                                   && rm -rf /root${CHE_CONTAINER_ROOT}/instance \
-                                   && rm -rf /root${CHE_CONTAINER_ROOT}/${CHE_MINI_PRODUCT_NAME}.env" > /dev/null 2>&1  || true
-    fi
-  fi
-
+  until directory_is_empty; do
+    docker_run -v "${CHE_HOST_CONFIG}":/root${CHE_CONTAINER_ROOT} \
+                  alpine:3.4 sh -c "rm -rf /root${CHE_CONTAINER_ROOT}/docs \
+                                 && rm -rf /root${CHE_CONTAINER_ROOT}/instance \
```
garagatyi: Have you checked whether multiple runs of this command is still needed when && replaced with ; ?

dockerfiles/base/scripts/base/commands/cmd_destroy.sh
```diff
@@ -78,3 +66,12 @@ cmd_destroy() {
   fi
 }
 
+directory_is_empty() {
```
garagatyi, line 68: It is a bit confusing to have both directory_is_empty and directory_is_clean functions. Is it on purpose?

dockerfiles/base/scripts/base/commands/cmd_destroy.sh
```diff
@@ -78,3 +66,12 @@ cmd_destroy() {
...
+   if [[ -d "${CHE_HOST_CONFIG}/docs" ]] ||
+      [[ -d "${CHE_HOST_CONFIG}/instance" ]] ||
+      [[ -f "${CHE_HOST_CONFIG}/${CHE_MINI_PRODUCT_NAME}.env" ]]; then
+    return 1
+  else
+    return 0
+  fi
+}
```
garagatyi: Please add trailing new line at the end of file

riuvshin:
afaik we wanted only possibility to enable debug for prod mode, if repo is mounted debug must be activated automatically.

garagatyi:
@riuvshin Yes, It is what I requested originally. But @l0rd asked about complete separation of debug and repo mounting. So I believe Tyler is going to implement that.

dockerfiles/base/scripts/base/commands/cmd_config.sh
```diff
@@ -76,7 +82,21 @@ generate_configuration_with_puppet() {
     CHE_ENV_FILE="${CHE_HOST_INSTANCE}/config/$CHE_MINI_PRODUCT_NAME.env"
   fi
 
-  if [ "${CHE_DEVELOPMENT_MODE}" = "on" ]; then
+  if debug_server; then
+    DEV_MODE="development"
+    WRITE_LOGS=""
```
riuvshin, line 95: does that mean that puppet logs will be shown only in debug mode? debug mode should only activate debug and do not change any other behaviour

dockerfiles/base/scripts/base/commands/cmd_start.sh
```diff
@@ -112,7 +112,50 @@ cmd_restart() {
...
+    info "start" "Booted and reachable"
+    info "start" "Ver: $(get_installed_version)"
+    info "start" "Use: ${DISPLAY_URL}"
+    info "start" "API: ${DISPLAY_URL}/swagger"
+    if debug_server; then
+      DISPLAY_DEBUG_URL=$(get_debug_display_url)
+      info "start" "Debug: ${DISPLAY_DEBUG_URL}"
+    fi
+  else
+    error "(${CHE_MINI_PRODUCT_NAME} start): Timeout waiting for server. Run \"docker logs ${CHE_SERVER_CONTAINER_NAME}\" to inspect the issue."
```
garagatyi, line 157: looks like kill $LOG_PID > /dev/null 2>&1 is needed here also

dockerfiles/base/scripts/base/commands/cmd_config.sh
```diff
@@ -86,33 +106,42 @@ generate_configuration_with_puppet() {
                   -v \"${CHE_HOST_DEVELOPMENT_REPO}/dockerfiles/init/modules\":/etc/puppet/modules:ro \
                   -e \"CHE_ENV_FILE=${CHE_ENV_FILE}\" \
                   -e \"CHE_CONTAINER_ROOT=${CHE_CONTAINER_ROOT}\" \
-                  -e \"CHE_ENVIRONMENT=development\" \
+                  -e \"CHE_ENVIRONMENT=${DEV_MODE}\" \
                   -e \"CHE_CONFIG=${CHE_HOST_INSTANCE}\" \
                   -e \"CHE_INSTANCE=${CHE_HOST_INSTANCE}\" \
-                  -e \"CHE_DEVELOPMENT_REPO=${CHE_HOST_DEVELOPMENT_REPO}\" \
+                  -e \"CHE_REPO=${CHE_REPO}\" \
```
garagatyi, line 121: Can you elaborate on why did you change CHE_HOST_DEVELOPMENT_REPO. As I can see it is still used in other places.

garagatyi:
I tried it and here is what I found:
- logs of che-server are not streamed when debug mode is enabled
- --debug with no repo mounted doesn't allow to debug che-server

TylerJewell:
I think you do not have the right images.  You need to wipe everything and then rebuild /base, /init, /cli in that order.

garagatyi:
I rebuilt everything with build-all script

TylerJewell:
I have never used that script - and I think we should assume that it's not perfect.  Because you need to make sure that /cli image inherits from the right /base image.

garagatyi:
Sorry, I forgot to switch branch :flushed:
Now everything works like a charm in Che. I didn't try Codenvy yet.
BTW looks like final output was aligned on purpose, but now it is not:
```
INFO: (che start): Ver: nightly
INFO: (che start): Use: http://172.17.0.1:8080
INFO: (che start): API: http://172.17.0.1:8080/swagger
INFO: (che start): Debug: http://172.17.0.1:8000
```

TylerJewell:
Yeah, I did the formatting on purpose.  I figure in debugging mode, developers do not care about formatting.  I'll respond to other comments later today.

dockerfiles/base/scripts/base/commands/cmd_config.sh
TylerJewell: Good catch. Updated.

dockerfiles/base/scripts/base/commands/cmd_config.sh
TylerJewell: I didn't make this path change.  I believe that @riuvshin had made this path change in another issue that was already reviewed and merged. I believe it has to do with being clearer in the in the instance folder on what parts are Tomcat vs. which parts are not.

dockerfiles/base/scripts/base/commands/cmd_config.sh
TylerJewell: Updated.

dockerfiles/base/scripts/base/commands/cmd_config.sh
garagatyi: Then can you rebase your branch or merge master into it

dockerfiles/base/scripts/base/commands/cmd_config.sh
TylerJewell: They are different.  In rows 85-97, we have a different logic for what this variable will contain dependening upon whether you have activated debugging or not.  So with this logic, you can now have four combinations: using a repo (yes / no) or debugging (write to logs yes / no).

TylerJewell: I have decided that the if() logic in this branch is no longer needed.  So I have simplified this down to remove the if logic and generate a single command with variables.

dockerfiles/base/scripts/base/commands/cmd_config.sh
TylerJewell: Thanks

dockerfiles/base/scripts/base/commands/cmd_destroy.sh
TylerJewell: No, I haven't created an issue.  I wasn't sure it was a docker issue for it.  We discovered this a couple months ago when we first implemented the `cmd_destroy` function.  We were seeing that we had to 3x remove stuff in that function and now it's coming up again.  I can create that issue for them.

dockerfiles/base/scripts/base/commands/cmd_destroy.sh
TylerJewell: When I put a semicolon there, then on windows it only requires a single loop to work.  Nice catch!

dockerfiles/base/scripts/base/commands/cmd_destroy.sh
TylerJewell: They had different syntax of what to check.  So I keep both as `directory_is_empty` and then add a qualifier for the one in config now.

dockerfiles/base/scripts/base/commands/cmd_config.sh
TylerJewell: Yes, it means that puppet logs will be streamed if you have `--debug` only.  If you mount a repository, but do not provide `--debug` then they would not be streamed.

dockerfiles/base/scripts/base/commands/cmd_start.sh
TylerJewell: Moved the kill operation to be outside of hte initial loop so that it's killed first, displayed second.

dockerfiles/base/scripts/base/commands/cmd_config.sh
TylerJewell: It was just come cleanup to keep things simpler.  The puppet templates had logic where both the repo and debugging logic was using the same variable. And this was needed to separate the logic across two variables.

TylerJewell:
I believe this issue is ready for merging.  Will update the matching Codenvy issue.

l0rd (approved):

riuvshin (approved):

garagatyi (approved):

Issue #3564, opened 2016-12-28
eclipse/che-ip returns 2 IP addresses as CHE_HOST


ghost:
I started an Ubuntu instance on digital ocean and ran cli:

`INFO: (che init):   CHE_HOST=138.68.77.169 10.19.0.8`

```
root@ubuntu-2gb-fra1-01:~# ifconfig
docker0   Link encap:Ethernet  HWaddr 02:42:58:be:fd:01  
          inet addr:172.17.0.1  Bcast:0.0.0.0  Mask:255.255.0.0
          inet6 addr: fe80::42:58ff:febe:fd01/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:401 errors:0 dropped:0 overruns:0 frame:0
          TX packets:303 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:0 
          RX bytes:35748 (35.7 KB)  TX bytes:54238 (54.2 KB)

eth0      Link encap:Ethernet  HWaddr ca:18:b8:42:99:90  
          inet addr:138.68.77.169  Bcast:138.68.79.255  Mask:255.255.240.0
          inet6 addr: fe80::c818:b8ff:fe42:9990/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:329732 errors:0 dropped:0 overruns:0 frame:0
          TX packets:24948 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000 
          RX bytes:483680037 (483.6 MB)  TX bytes:2256880 (2.2 MB)

eth0:1    Link encap:Ethernet  HWaddr ca:18:b8:42:99:90  
          inet addr:10.19.0.8  Bcast:10.19.255.255  Mask:255.255.0.0
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1

lo        Link encap:Local Loopback  
          inet addr:127.0.0.1  Mask:255.0.0.0
          inet6 addr: ::1/128 Scope:Host
          UP LOOPBACK RUNNING  MTU:65536  Metric:1
          RX packets:160 errors:0 dropped:0 overruns:0 frame:0
          TX packets:160 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1 
          RX bytes:11840 (11.8 KB)  TX bytes:11840 (11.8 KB)

[...]

PR #3582, opened 2017-01-02
Exclude IP addresses from aliased interfaces

TylerJewell:
### What does this PR do?
Adds a different filter to eliminate IP address aliases. In the case where a single network interface returns multiple matching IP addresses, filters out any from a networking interface ":#" at the end, so that the function only returns a single IP address.

#3564 

### Previous behavior
Returns multiple IP addresses - one for each matching interface.

### Tests written?
No

### Docs updated?
N/A


ghost (approved):

benoitf (approved):
OK for me as long as it still work on mac and windows

TylerJewell:
I tested on both - got proper IP address each time.

PR #3600, opened 2017-01-04
CHE-58: Refactoring introducing 'KubernetesResourceHelper' class

ibuziuk:
### What does this PR do?
OpenShiftConnector refactoring

https://issues.jboss.org/browse/CHE-58

### DIFF
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/test/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnectorTest.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/test/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnectorTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2012-2016 Red Hat, Inc.
+ * Copyright (c) 2012-2017 Red Hat, Inc.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v1.0
  * which accompanies this distribution, and is available at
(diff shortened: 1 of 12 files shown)


ibuziuk:
@l0rd could you please take a look ?

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/KubernetesResourceHelper.java
```diff
@@ -0,0 +1,93 @@
...
+import io.fabric8.kubernetes.api.model.EnvVarBuilder;
+import io.fabric8.kubernetes.api.model.IntOrString;
+import io.fabric8.kubernetes.api.model.KubernetesResource;
+import io.fabric8.kubernetes.api.model.ServicePort;
+
+/**
+ * Provides functionality for {@link KubernetesResource} transformation e.g.
+ * ContainerPort, ServicePort, EnvVar etc.
+ */
+public class KubernetesResourceHelper {
```
l0rd: Instead of an Helper class I would rather create a distinct class for each resource (e.g. KubernetesContainer, KubernetesService, KubernetesEnvVar). WDYT?

plugins/plugin-docker/che-plugin-openshift-client/src/test/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnectorTest.java
```diff
@@ -67,10 +68,13 @@
     private OpenShiftConnector openShiftConnector;
     
     private KubernetesLabelConverter kubernetesLabelConverter;
+    
+    private KubernetesResourceHelper kubernetesResourceHelper;
```
l0rd: We can probably split this big test class in smaller classes now

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/KubernetesResourceHelper.java
ibuziuk: good point

ibuziuk:
@l0rd PR has been updated

l0rd (approved):

Issue #3464, opened 2016-12-21
Can't stop exec agent process after refresh page


dkulieshov:
Steps to reproduce:
1. Create maven project
2. Run standard maven build command
3. Refresh page
4. Enjoy

Expected behavior: process can be stopped after page refresh
Actual behavior: process cannot be stopped after page refresh

PR #3613, opened 2017-01-05
CHE-3464: fix a problem with stopping the process

svor:
### What does this PR do?
Adds `process_status`  to the subscription to machine's processes.

https://github.com/eclipse/che/issues/3464

### Previous behavior
Client doesn't get response event about stopping a process after refresh page

### New behavior
Client gets response event about stopping a process after refresh page


dkulieshov (approved):

vparfonov (approved):

vparfonov:
ci-build

Issue #3645, opened 2017-01-09
Alter Chedir Syntax


TylerJewell:
The current Che dir syntax in the new CLI introduces an inconsistency that we should change. The current syntax is:
```
# Clone a source code repo that you want to convert into a workspace
git clone https://github.com/che-samples/web-java-spring-petclinic
cd web-java-spring-petclinic

# Convert the repository into a workspace running in a private Che server
docker run  -it --rm -v /var/run/docker.sock:/var/run/docker.sock 
                    -v <path>:/data 
                      eclipse/che dir <path-to-web-java-spring-petclinic-folder> up
```

It used to be "che dir up" to mean current directory. This is more complicated and also introduces two different approaches to using a path.  We should only use a single approach for managing a path, which is to volume mount that path.

We should preserve the syntax that appears after the image name - so "eclipse/che dir [up|down]".  Everything else should be a volume mount as a Docker option.

The syntax:
1. If <path>:/chedir volume mounted, then we assume that the <path> folder volume mounted == to the git repository that should be used with che dir.

2. If :/dir not volume mounted, then we will assume that the folder mounted to <path>:/data is the Chedir folder, and in this case, the :/data volume mount would be both the Che install and the location of the Chefile.

- [x] Update the syntax summary in the entrypoint.sh to include optional volume mount to :/chedir
- [x] Volume mount this path, and then check that it is valid and writable in the CLI.
- [x] If :/chedir is not volume mounted, use the path set to :/data as the volume mount and print warning message to CLI output indicating that both the Che install and chedir path are identical.
- [x] There is an issue with `cmd_offline` where the Docker images for action, ssh, chedir and other non-runtime utilities are not downloaded as part of the offline operation. These utilities should be hard coded into the offline mechanism some way. We can start by hard coding them into the cmd_offline command.
- [x] We should find all instances within the CLI where we use hard coded image names (not inherited from the registry file) and replace the tags with digest values instead.  For example, we frequently use `alpine:3.4` as an image for a utility, and we should replace that with a digest value. The same would go for the utilities, such as `eclipse/che-dir:nightly`, `eclipse/che-ip:nightly`, `eclipse/che-action:nightly`, and `eclipse/che-test:nightly`.  We liberally use these image names throughout the CLI in different locations. We should add these "utility" image names and digests as constants in the startup.sh, and then reuse those constant values within the other CLI commands.

Commands altered:
1. cmd_backup
2. cmd_start
3. cmd_config
4. cmd_offline
[...]

PR #3646, opened 2017-01-10
Che CLI Refactoring

TylerJewell:
### What does this PR do?
1. Changes Chedir behavior to use the path set by mounting `:/chedir` instead of passing it as a parameter. If `:/chedir` is not volume mounted then uses the value mounted to `:/data` and prints warning.
2. Adds in all UTILITY images into cmd_offline, so that they are available in any offline mode.
3. Places all UTILITY image names as hard coded constants in `startup.sh`, and then uses these constants in all CLI files.

#3645 

### Previous behavior
Previously had syntax of `docker run DOCKER_OPTIONS eclipse/che dir <local-path> up`.  This moves the <local-path> to be derived from a volume mount with DOCKER_OPTIONS instead.

### Docs updated?

### DIFF
--- a/dockerfiles/base/scripts/base/startup.sh
+++ b/dockerfiles/base/scripts/base/startup.sh
@@ -82,6 +82,14 @@ init_constants() {
 
   DEFAULT_CHE_LICENSE=false
   CHE_LICENSE=${CHE_LICENSE:-${DEFAULT_CHE_LICENSE}}
+
+  # Replace all of these with digests
+  UTILITY_IMAGE_ALPINE="alpine:3.4"
+  UTILITY_IMAGE_CHEIP="eclipse/che-ip:nightly"
+  UTILITY_IMAGE_CHEACTION="eclipse/che-action:nightly"
+  UTILITY_IMAGE_CHEDIR="eclipse/che-dir:nightly"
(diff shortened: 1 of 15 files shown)


benoitf (commented):
I've added a fix into the branch

dockerfiles/base/scripts/base/startup.sh
```diff
@@ -82,6 +82,14 @@ init_constants() {
 
   DEFAULT_CHE_LICENSE=false
   CHE_LICENSE=${CHE_LICENSE:-${DEFAULT_CHE_LICENSE}}
+
+  # Replace all of these with digests
+  UTILITY_IMAGE_ALPINE="alpine:3.4"
+  UTILITY_IMAGE_CHEIP="eclipse/che-ip:nightly"
```
benoitf, line 88: should we use ${CHE_VERSION} instead of nightly ? until it's using digests

benoitf (approved):

slemeur:
Updated docs for the PR:
https://github.com/codenvy/che-docs/pull/51

slemeur (approved):

slemeur (approved):

Issue #3578, opened 2016-12-30
Unexpected losing of  ocus in the terminal after minimize/maximize operations


musienko-maxim:
**Reproduction Steps:**

1. Go to the terminal, type for example `ps ax` command several times
2. Maximize terminal panel and type the `ps ax` commands one more time
3. Minimize the terminal try to type some text.

**Expected behavior:**
The terminal should scroll to last line, the text should be typed.
**Observed behavior:**
Terminal does not scroll position to last line. The focus is lost.  

**Che version:**   5.0.0
**OS and version:** Ubuntu 14 TLS
**Docker version:** 1.12.5
**Che install:**    Docker container

svor:
Scrolling to the bottom will be fixed after applying xterm.js https://github.com/eclipse/che/issues/3529

PR #3654, opened 2017-01-10
CHE-3578: set focus on terminal after resizing

svor:
### What does this PR do?
Sets focus on the terminal after resizing.

https://github.com/eclipse/che/issues/3578

### Previous behavior
Focus is lost when the terminal was minimized or maximized  

### New behavior
Focus sets when size of the terminal is changed


vzhukovs:
@vparfonov I suppose that this PR will break changes made by this previous commit https://github.com/eclipse/che/commit/3da5a96b3a1ed82723fc348a066c1cdf6d148d7c#diff-e15936ac6286c2e6559b1194493b5b30L245

Because this method is called when terminal first time appears on the UI (when IDE starts).

vparfonov (approved):

Issue #3664, opened 2017-01-11
Improve error message when starting nightly from a latest configuration


slemeur:
When user is using the CLI with the image `eclipse/che` and moving to use `eclipse/che:nightly` we are printing the following error message.

```
WARNING: No kernel memory limit support
INFO: Proxy: HTTP_PROXY=, HTTPS_PROXY=, NO_PROXY=*.local, 169.254/16
INFO: (che cli): Checking network... (hint: '--fast' skips version and network checks)
INFO: (che cli): Loading cli...
ERROR:
ERROR: Your CLI version 'eclipse/che:nightly' does not match your installed version '5.0.0'.
ERROR:
ERROR: The 'nightly' CLI is only compatible with 'nightly' installed versions.
ERROR: You may not 'che upgrade' from 'nightly' to a numbered (tagged) version.
ERROR:
ERROR: Run the CLI as 'eclipse/che:<version>' to install a tagged version.
```

This message is not helpful and doing an `upgrade` will not work. 

In general, I would think that we should catch when the user is starting `nightly` because it behaves differently than a tag version, and should be used in parallel of a default installation.

Would be nice to inform the user that, he can try nightly by installing it on another folder to be mount. 

**Che version:**    5.0.0
**OS and version:**    MAC OS 10.12.2
**Docker version:** 1.12.3

TylerJewell:
I do not understand the issue with the message.  We do not allow nightly and non nightly installs to mix.  So the message is clear on what the error is.

benoitf:
AFAIK It may not be clear that there is already an installation in the /data folder. Or alternative ways for the end-user

For example
```Your CLI version 'eclipse/che:nightly' does not match your installed version '5.0.0'.```
we could add
```Your CLI version 'eclipse/che:nightly' does not match your installed version '5.0.0' in /home/users/host-folder.```

and a message like :
```You may use a separate ${CHE_FORMAL_PRODUCT_NAME} by providing another /data volume mount folder.```

TylerJewell:
I see what you all are coming from now.  This is an easy improvement.

TylerJewell:
Added to the PR.

Issue #3640, opened 2017-01-06
Prevent Che start if 'nightly' or 'latest' images are not current.


TylerJewell:
The `nightly` and `latest` images are convenience images desigend to make lives of Che developers easier. However, there are circumstances where changes to the core `nightly` image for the CLI requires that all of the underlying dependency images are updated with it.  Getting the most recent CLI nightly, but leaving the `eclipse/che-init:nightly` or `eclipse/che-server:nightly` as older versions can cause problems.

A savvy developer can use `--pull` or `--force` to forcibly clean this up, but when the errors occur it is not obvious that this is the approach to clean things up.

We have tried looking at the `created-date` within the local image cached on a computer with the matching parameter on DockerHub to see if there is a difference. However, the property from dockerhub only reflects when the image was uploaded and does not match the property on the local image, so these values cannot be used for a meaningful comparison.

In doing some tests, it is possible from within the CLI of a Docker container, to check the current date and compare it to the date that certain images were written to disk. We could therefore implement some logic such as:
1. Get the current date of the Docker engine host (Docker VMs can have dates not matching host). On Windows, while today is January 6th, my HyperV VM has the date as December 26th, so we will need to baseline dates.

2. Get the date that the eclipse/che-cli:nightly image was written to disk.

3. If the date the image is written to disk is >24 hours, then we should display the warning message.

4. We could then get the list of dependency images that Che requires (after its been loaded) and then perform a similar check for each one of those. If the dependency images were written to disk >24 hours, we could then detect that other images are older, would they like us to assert a `--force` where we `rmi` and `pull` the image to guarantee it is current.

Another variation of this is that any time you run a `nightly` or `latest` image we could display a table to stdout that shows all of the images loaded by Che, their name, and how many days it has been since they were written to disk with a hint that `--force` will repull all images. So this way any time a developer sees that some of their images are >1 day old, they know that any issues could just require a repull.

```
# Get the docker image id
docker image --no-trunc eclipse/che-cli:nightly

# List the particular file and gets it date written to disk
docker run -it -v /var/lib/docker:/var/lib/docker alpine ls -al /var/lib/docker/image/aufs/imagedb/content/sha256 | grep <image-id> | <cut-to-get-date>
```

TylerJewell:
[...]

PR #3671, opened 2017-01-11
CLI improvements for handling of nightly images

TylerJewell:
### What does this PR do?
1. Improves the error message that is printed when running nighlty image against install that is not nightly
2. Changes behavior to always use --pull whenever init() or config() a nightly image.
3. If --fast, skips the default pull behavior.
4. Changes the nightly CLI check. It now checks the date at which the image this container was created from was written to disk. If it was written to disk > 24 hours ago, it repulls the latest image and then stops with a message to rerun the CLI.

#3664 
#3640 

### Previous behavior

### New behavior

### Tests written?
Yes/No

### Docs updated?

### DIFF
--- a/dockerfiles/base/scripts/base/cli/cli-functions.sh
+++ b/dockerfiles/base/scripts/base/cli/cli-functions.sh
@@ -382,8 +382,7 @@ timestamp_date_iso8601() {
 }
 
 # Compare two dates with ISO 8601 format and return
-# true if the first date is less than the second date (with a 1hour period)
-# else false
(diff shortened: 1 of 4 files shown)


dockerfiles/base/scripts/base/cli/cli-functions.sh
```diff
@@ -374,10 +385,7 @@ timestamp_date_iso8601() {
 # true if the first date is less than the second date (with a 1hour period)
```
benoitf: need to adjust the comment (now it's 24hour)

PR #3683, opened 2017-01-11
fix cli tty issue

riuvshin:
we should not use `-t` in CLI because this break CLI usage from non tty environments like CI.


TylerJewell:
Let's get verification from @benoitf that removing tty from _test, _dir, _action will not affect the user output at all.  If he +1 that, then you have my approval.

TylerJewell (approved):

benoitf:
I'm unsure for actions like ssh as we need a tty afaik

TylerJewell:
Ok - so @riuvshin - let's make sure that we are not removing tty for end user initiated commands which are _ssh, _dir, and maybe _action.

riuvshin:
@benoitf @TylerJewell reverted

PR #3685, opened 2017-01-12
New changeslog

mkuznyetsov:
### What does this PR do?
New changeslog:
- excluded issues for unreleased versions
- excluded issues with `kind/question` label

https://github.com/codenvy/infrastructure/issues/34


mkuznyetsov:
command used to generate changeslog:
```
github_changelog_generator --bugs-label "**Issues fixed with 'bugs' label:**" --pr-label "**Pull requests merged:**" --enhancement-label "**Issues with 'enhancement' label:**" --bug-labels="kind/bug" --enhancement-labels="kind/enhancement" --exclude-labels="kind/question" --no-unreleased
```

skabashnyuk:
github_changelog_generator has no functionality to add custom header. But you can simply workaround it with bash functions.

TylerJewell:
@skabashnyuk - while the new changelog is better than what we have right now, I find the segments still quite confusing.

1. What is a "closed issues" category?  All issues are closed in the end, so that would imply that both issues and PRs are double counted.

2. I don't think we can include issues and PRs in the same list, as they double count each other.  Every issue requires at least 1 PR to be associated with it - so if we have PRs and other stuff, then we end up double counting issues.

Is there a way to **only** show PRs merged, but to have the segments broken down by labels?  I know you explained that the generator has some limitations here.  But what we have in this sample seems more confusing, not less.

skabashnyuk:
1. All closed issues except labeled as --bug-labels="kind/bug" --enhancement-labels="kind/enhancement" --exclude-labels="kind/question" 
2. We can include or exclude pr or issues or combine.

github_changelog_generator  has issely understandable set of options plz take a look
https://github.com/skywinder/github-changelog-generator/wiki/Advanced-change-log-generation-examples

TylerJewell:
Yeah, I have read that selection, but that is a technical answer.  I am just studying the final outcome, and encourage you to look at it as well.  I am just studying the final sections that are produced and they do not add up, so hope you can suggest a diffrent set of parameters that make sense.

These are the headers that appear in the changelog:
Closed Issues
Closed PRs
Issues with enhancmeent label
Issues with bugs label

These are not mutually exclusive - so can you show me a configuration of the generator and what the headers would be so we can analyze those?

skabashnyuk:
https://github.com/eclipse/che/pull/3685#issuecomment-272107064
this is how generator configuration is lookslike

TylerJewell:
I don't think that I can comment on the changelog generator configuration.  I am just commenting on the output, and poiting out to you that the four titles generated are creating duplicate issues and overlap in meaning.  So asking engineering to revisit the changelog configuration to avoid that confusion, if possible.

skabashnyuk:
Can you give an example of duplicated issues and overlap in meaning that you are trying to avoid?

TylerJewell:
1. Closed Issues - this, in generally would mean all issues + PRs closed.  So the average user would see all of the issues in here as duplicates of those in PRs closed.   
2. We do not want to duplicate "issues" with "prs".  PRs is the only list that matters, because if there is not a PR for it, then it wasn't really a closed issue.  So we should not be including anything from the regular issues list.
3. For all of the PRs, we would then want to segment them by kind/bugs or kind/enhancement or kind/task.  So instead of showing "Closed PRs" it would just be "kind/bugs" and we would list all closed PRs that were kind/bug label, for example.  If we add up all of the sections together, then they would total the complete list of closed PRs fo the milestone.

skabashnyuk:
I become more confident that your requirements is not possible  to implement  with  ```github_changelog_generator```. Shall I close this PR?

TylerJewell:
I think we have to close this PR, continue the conversation on the root infrastructure issue, and now call it a blocker, as we have fundamentally broken Changelog right now.  I saw a maven generator which does not apply any labels - it just shows all of the commits.  Maybe we just use that instead for now.

skabashnyuk:
Not able to implement with generator. Change log will be made manually.

PR #3690, opened 2017-01-12
fix che-ip

### LINKED ISSUE #3684: cli nightly does not work on different OS or storage drivers
after we merged this https://github.com/eclipse/che/commit/cd5db87c9a86999e63ac773aec296427c8606506

we have problems on mac and centOs 

error is 

```
INFO: (codenvy cli): Checking network... (hint: '--fast' skips version, network, and nightly checks)
INFO: (codenvy cli): Loading cli...
date: can't stat '/var/lib/docker/image/aufs/imagedb/content/sha256/d4bbec79a4ac34b1ead51573f9c05b2e48c1eedc218b5783ed5339f3a9518fb4': No such file or directory
```

this happen because this path `/var/lib/docker/image/aufs/` maybe different for different OS and storage drivers. in production systems we use special data pool devices for docker storage and afaik you can just go there and navigate through files "layers" they all compressed and not accessible.

@TylerJewell said that `docker inspect -f '{{ .Created }}' codenvy/cli-saas:nightly` this does not work for us because this date is not the date when image was written to disk, but why we need to know when it was written? As I understand it should be enough to know when image was built like if image was built today - ok this is accurate one if yesterday - no. We can assume here that nightly image we update each day. So I don't understand why that CREATED date does not work for us, I don;t know why we need to know when it was written to disk because it maybe written today but image can be outdated one, i.e we should know when locally pulled CLI image was built and if it was built more than 24h ago - we do pull it and ask to relaunch CLI

@benoitf @TylerJewell wdyt guys?

riuvshin:
need to fix https://github.com/eclipse/che/issues/3684

on some OS che-ip does not work because output is different.
here I've added reg-ex `.*` which assume that scope can be differ from `global` and some extra words could appear like `dynamic`.

### DIFF
--- a/dockerfiles/ip/entrypoint.sh
+++ b/dockerfiles/ip/entrypoint.sh
@@ -40,7 +40,7 @@ if test -z ${NETWORK_IF}; then
 fi
 
 ip a show "${NETWORK_IF}" | \
-            grep "scope global ${NETWORK_IF}" | \
+            grep -e "scope.*${NETWORK_IF}" | \
             grep -v ':' | \
             cut -d/ -f1 | \
             awk '{print $2}'


benoitf (approved):

benoitf:
Also could you check that https://github.com/eclipse/che/issues/3564 is still ok ?

riuvshin:
@benoitf just tested with @eivantsov on digital ocean ubuntu. works.

PR #3694, opened 2017-01-12
Improve project type resolver

olexii4:
Signed-off-by: Oleksii Orel <oorel@codenvy.com>

### What does this PR do?
Move auto check project type logic into separate service.

### Release Note
- [x] Added to release note

### DIFF
--- a/dashboard/src/app/projects/create-project/create-project.controller.ts
+++ b/dashboard/src/app/projects/create-project/create-project.controller.ts
@@ -688,119 +688,24 @@ export class CreateProjectController {
     return projects;
   }
 
-  resolveProjectType(workspaceId, projectName, projectData, deferredResolve) {
-    let projectDetails = projectData.project;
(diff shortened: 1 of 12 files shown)


benoitf:
@olexii4 could you explain a little bit more the "Improve project type resolver"

why it was not good before, why now it's better
is it for example : move logic from controller to the API ?

olexii4:
@benoitf 
I moved some of auto check type logic from controller to api as additional of project's update. Because we need it in several places. But the best way - move auto check type logic into separate service. WDYT?

benoitf (approved):
@olexii4 yes it makes sense

olexii4:
done.

akurinnoy (approved):

ashumilova (approved):

slemeur:
@riuvshin this fix must be included for 5.1.1

PR #3695, opened 2017-01-12
revert back pulling CLI if local image older than 24h

riuvshin:
after I've reverted https://github.com/eclipse/che/commit/f7ca10c86a12bc5a6da766dfcf281f2b656c23c3 this I've also removed logic which do pull CLI and exit if CLI is older than 24h
so I'm adding it back with this PR.

### DIFF
--- a/dockerfiles/base/scripts/base/cli/cli-functions.sh
+++ b/dockerfiles/base/scripts/base/cli/cli-functions.sh
@@ -400,7 +400,10 @@ function verify_nightly_accuracy() {
     elif $(newer_date_period \
          $(timestamp_date_iso8601 "${LOCAL_CREATION_DATE}") \
          $(timestamp_date_iso8601 "${REMOTE_CREATION_DATE}")); then
-     warning "There is a newer ${CHE_IMAGE_FULLNAME} image on DockerHub."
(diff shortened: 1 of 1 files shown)


benoitf (approved):

TylerJewell:
We need to do testing on this again - changed the logic.

TylerJewell:
Fixes #3684

dockerfiles/base/scripts/base/cli/cli-functions.sh
```diff
@@ -381,12 +381,14 @@ function verify_nightly_accuracy() {
   # than the one stored on DockerHub.
   if is_nightly; then
 
-    REMOTE_NIGHTLY_JSON=$(curl -s https://hub.docker.com/v2/repositories/${CHE_IMAGE_NAME}/tags/nightly/)
+#    REMOTE_NIGHTLY_JSON=$(curl -s https://hub.docker.com/v2/repositories/${CHE_IMAGE_NAME}/tags/nightly/)
```
benoitf: we can drop the commented line

riuvshin: done

benoitf (approved):

PR #3700, opened 2017-01-12
New tests for CLI, refactoring of CLI scripts, fix destroy cmd

mlabuda:
Signed-off-by: mlabuda <mlabuda@redhat.com>

### What does this PR do?

- Refactoring of existing scripts to be usable for testing
- First draft of the new CLI tests (tests for prompts, init command, destroy command)
- Fixed destroy  command - argument "--cli" did not removed cli.log file from local data folder

### Previous behavior
- CLI destroy command did not delete cli.log file if "--cli" argument was passed
- there were no tests for the new CLI

### Tests written?
Yes

### Changelog
- added functional [bats](https://github.com/sstephenson/bats) tests for the Che CLI

### Release notes
Che CLI is backed by functional [bats](https://github.com/sstephenson/bats) tests. Those tests run inside of a container `eclipse/che-bats`. Test execution is performed on the existing local code by mounting base directory to the test container, so you can be sure nothing got broken by modifying the Che CLI scripts.

### Docs
https://github.com/codenvy/che-docs/pull/78


dockerfiles/cli/cli_tests.bats
```diff
@@ -0,0 +1,170 @@
+#!/usr/bin/env bats
+# Copyright (c) 2012-2017 Red Hat, Inc
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License v1.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-v10.html
+#
+# HOW TO RUN TEST?
+# To run tests, ensure you have:
+#     - installed bats
```
benoitf: Could we use a docker container to run the tests.
I would avoid to have bats as a requirement installed on the computer. (if it's there ppl can use it directly) but as all of dockerfiles are running with docker, I would expect to run tests within a docker container as well.

Also, as we have a build.sh script in /base folder, the script could have option to launch automatically the tests (with an option to skip the tests as well) so dev only launch a single command.

dockerfiles/cli/cli_tests.bats
```diff
@@ -0,0 +1,170 @@
...
+setup() {
+  clean_docker
+} 
+
+clean_docker() {
+  if [ "$(docker ps -q | wc -l)" -gt 0 ]; then
+    docker kill $(docker ps -q)
+  fi
+  if [ "$(docker ps -qa | wc -l)" -gt 0 ]; then
+    docker rm $(docker ps -qa)
```
benoitf: Why do we need to kill all docker containers running ?

dockerfiles/cli/cli_tests.bats
```diff
@@ -0,0 +1,170 @@
...
+  output=$(usage || true)
+
+  #WHEN
+  result=$(docker run -v /var/run/docker.sock:/var/run/docker.sock eclipse/che-cli:nightly || true)
+
+  #THEN
+  [[ $result == *$usage* ]]
+}
+
+@test "test 'init' and 'destroy --quiet' with existing dir" {
```
benoitf: could we separate tests for commands in a separate bats file ?
So it's easy to find where tests are defined for a command.
IMHO adding all the tests in a single unit file will result in something really huge and difficult to maintain

dockerfiles/cli/cli_tests.bats
mlabuda: It's safer to have clean docker daemon for test execution. Once there will be more tests and some of them would fail, there could be e.g. used ports (e.g. there is already running che-server or any other docker image using required ports...) which would cause failure of next test. Or there could be blocked other resources etc. 

Also in test execution, if there would fail a test which has been working with a specific port, resource or there would be named container and other test would use same named container, it would result into complicated teardown function. With clean docker daemon - no running containers, I provide clean environment for test execution and run of tests would be "isolated".

dockerfiles/cli/cli_tests.bats
mlabuda: Sure thing. It popped in my head that with having all tests in one file it would be huge, several hundreds LoC. I will also move tests from dockerfiles/cli folder to dockerfiles/base/tests folder, it will improve user experience with tests execution. Cli folder was my first choice as a dir for CLI but it makes more sense to have it in scripts folder.

dockerfiles/cli/cli_tests.bats
benoitf: but it doesn't work in a CI system where you may have multiple containers already running for other projects
could the clean could be limited only to the images that the CLI is managing ?
like these images : https://github.com/eclipse/che/blob/master/dockerfiles/cli/version/nightly/images

mlabuda: Yeah, I am aware of that and I had an idea of running those tests in a separate job. I could use named containers and kill only those, but it still won't ensure I will have an environment where tests can run successfully. There still could be occupied port (e.g. 8080 by a docker container) and all CLI tests running on top of started Che would fail in setup.

Would it be a problem to have a separate job in CI system just for test execution?

benoitf: we could have requirements performed in a test as well.
Like if port 8080 is not available then it fails the requirements
or find a random port then start che on this random avilable port instead of using 8080

dockerfiles/cli/cli_tests.bats
```diff
@@ -0,0 +1,170 @@
...
+  #GIVEN
+  TMP_PATH1=${TEST_DIR_PATH}1
+  mkdir -p $TMP_PATH1
+
+  #WHEN
+  docker run -it --rm -v /var/run/docker.sock:/var/run/docker.sock -v $TMP_PATH1:/data eclipse/che-cli:nightly init
```
l0rd: Why don't you call directly `cli_init` here instead of starting a new docker container? This looks more like an integration test rather than a unit test: it may fails for different reasons other than the `init` function and it will take more time to execute it (that's not significant now but think when you will have 100+ tests)

mlabuda:
Ok, I changed it a bit:
- split into several bats files
- extracted test base (helper) script
- created a docker image for test execution (for now it is at dockerhub as mlabuda/bats-in-dind:0.1)
- attached Dockerfile to create a docker image for tests execution.
- created shell scripts for simple test execution - just execute all-tests.sh script

benoitf:
@mlabuda why do we need to do docker in docker and use privilege mode ? In CLI we use for example the mount of the docker.sock allowing to use sibling containers and not dind.
also, where is Dockerfile content of the "mlabuda/bats-in-dind" image ?

Also what about the fact to use more "unit testing" than "integration testing" by not launching a docker container for each test but only launching script ?

TylerJewell:
We have some inconsistent file names now:
1. Most files have underscores '_' to separate names, but some files have dashes '-'.  See 'all-tests.sh' vs. 'cmd_network.sh' vs. 'cli-functions.sh'.  I propose that we standardize on the underscore, just because there are fewer files to change.

2. We have sometimes 'funcs' and 'functions'.  See 'startup-funcs.sh' vs. 'cli-functions.sh'.  I am ok with using the shorthand form.

@benoitf - why do the cli-functions need to be in a separate folder `/cli`?  I think it would be ok for them to be in the `/scripts/base` folder along side the other utility scripts.  And then all of the commands just go into `/commands`.

benoitf:
@TylerJewell there is no need to be in a separate folder. It only requires to be in a `base` folder (or subfolder)

TylerJewell:
Then I propose that we move it as part of @mlabuda pull request.  @mlabuda - will you make these file name and location changes?  I would then change /cli/cli-functions.sh to just be /utilities.sh or /library.sh to have a simpler name.

l0rd:
@benoitf @mlabuda dind can be a good idea to avoid dependency on local docker daemon and to have consistent tests even if other containers are running locally. However I would use it differently and make it optional:

**STEP 1 (Optional)**: Run dind official image
`docker run --privileged --name docker-for-cli-tests -d docker:1.12-dind`

**STEP 2**: Run `bats` tests using a custom image based on docker:1.12 (not dind version):
```bash
docker run --rm                         \
     -v $BASE_DIR:/tests                \
     --link docker-for-cli-tests:docker \ # If not executing STEP 1 remove this line and mount the docker socket instead
     bats-docker:1.12                   \
     bats sometests.bats
```
where `bats-docker` is an image built on top of `docker:1.12` where `bats` is installed.

mlabuda:
@TylerJewell sure, I can include file name changes. What do you mean by /cli/cli-functions.sh? I dont see such file there or any similar file name.

@benoitf I wanted isolated environment for test execution, that's why I run it in dind. Dockerfile for the dind in bats image is the Dockerfile in this commit - AFAIK to upload Dockerfile to dockerhub it probably has to be automated build and to have automated build, the dockerfile would have to be in some repo and if I would use mine repo, it would have to be then change to be the eclipse che repo.... So basically this is draft to get dockerfile to this repo and then setting up automated builds on dockerhub under proper organization (not mine repo). Workaround for egg-chicken problem.

Also I want to do automated unit tests as well. For now there are those high level system tests. Automated tests would need more refactoring (to be code testable - mostly because of initialization of environment to be able to run unit tests - env. vars...).

TylerJewell:
This file: https://github.com/eclipse/che/tree/master/dockerfiles/base/scripts/base/cli

dockerfiles/base/tests/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM docker:1.12.6-dind
+
+RUN apk add --no-cache bash
+
+RUN curl -o /tmp/v0.4.0.tar.gz -L https://github.com/sstephenson/bats/archive/v0.4.0.tar.gz \
+	&& tar -x -z -f /tmp/v0.4.0.tar.gz -C /tmp \
+	&& bash /tmp/bats-0.4.0/install.sh /usr/local \
+	&& rm -rf /tmp/*
+
+WORKDIR /home/testing
```
benoitf: there is a missing Header in the file
Also we should include a build.sh script to build it

benoitf: BTW with remark of @l0rd about dind we may probably drop this custom image

dockerfiles/base/all-tests.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
```
benoitf: we are missing a way to call tests from build.sh script. Also do we need the file to be named "all-tests.sh"(or to be in a separate file)

mlabuda:
@TylerJewell ok, my bad, I was looking into dockerfiles/cli files.

dockerfiles/base/tests/Dockerfile
mlabuda: Sure, I will drop this file.

dockerfiles/base/all-tests.sh
mlabuda: Now, when we decided to do it other way, this all tests "suite" can be removed and test execution commands can be added to build.sh. I intended to include all-tests.sh execution in build.sh. To have all-tests.sh a place to track .bats files to execute and to have just plain shell script calling in build.sh.

benoitf: ok

mlabuda:
Ok, now it should be ok. Few comments to PR:
- tests run inside a docker container from image based on che-base
- tests are functional - here few more words. I spent one day trying to do tests like you suggested, but it is not possible with scripts as they are now. The set up just to run a cmd_command function was huge, plenty of lines, env. vars settings.... Problem is that the code is too tightly coupled. To have code in a shape to be able to write unit tests, it requires huge refactoring - to functions behave like functions = take arguments and get some result or change some state etc. Bcs now. even though there are plenty of bash functions, they are not testable as units. So for now I suggest to create few high level tests to be sure that what should work, works. Then, once I have time (bcs. now I have few more important tasks to do), I would do refactoring of CLI scripts to be unit-testable.
- tests are run by calling in build.sh script but I put tests to separate file - to have it cleaner, because more tests will be added to all_tests.sh and I would like to keep build.sh simple
- tests along the way contains also fix for 'destroy --cli' command

benoitf:
ok, also as you've lot of different changes, maybe it will be easier/faster for you to provide small pull requests instead of maintaining a lot of changes for a long time.
Like one that fix destroy command, one that refactor code, one that add the tests
so it will be easier to review, get approval and merge them.

because I think code refactor and fix for destroy are straightforward.

dockerfiles/base/tests/Dockerfile
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
```
benoitf: shouldn't this file be at dockerfiles/tests/ directory ?
as base is the base image, and other images extending base are in dockerfiles/directory

dockerfiles/base/build.sh
```diff
@@ -6,7 +6,10 @@
...
 
 IMAGE_NAME="eclipse/che-base"
-. $(cd "$(dirname "$0")"; pwd)/../build.include
+base_dir=$(cd "$(dirname "$0")"; pwd)
+. $base_dir/../build.include
 
 init "$@"
 build
+
+source $base_dir/all_tests.sh
```
benoitf: we can't use source there because it doesn't work with "sh" (like dash)
we've to use "." to include content

mlabuda:
Sure, next time I will do smaller pull requests. This one is bigger bcs. it is initial testing PR to set up tests structure...

dockerfiles/base/tests/Dockerfile
mlabuda: there is already dockerfiles/test folder and it has different Dockerfile. Having dockerfiles/tests and dockerfiles/test would be ambiguous.

dockerfiles/base/all_tests.sh
```diff
@@ -0,0 +1,24 @@
...
+# Copyright (c) 2012-2017 Red Hat, Inc
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License v1.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-v10.html
+#
+# Contributors:
+#   Marian Labuda - Initial Implementation
+
+# Build test image
```
benoitf: maybe we should separate "building of test image" (like being at /dockerfiles/tests) from the test execution that will stay there.

dockerfiles/base/build.sh
mlabuda: Ok, I will change that. Didn't know that the build does not use bash.

benoitf: FYI bash can be used in all cli code (because we know for sure that bash is installed inside the docker image) while here we don't know if end-user has bash or not (and shebang is only /sh)

mlabuda: Ok, understand. I did not realize there could be users who would develop/run from source eclipse-che and would not have bash, only Bourne shell.

dockerfiles/base/tests/Dockerfile
benoitf: it may be "script-tests" or "unit-test" or only "bats" as we're using bats 
but BTW, do we require to build our own custom bats docker image ?  none of bats images of dockerhub is working ?
for launcher we were using
https://github.com/eclipse/che/blob/5.1.0/dockerfiles/launcher/launcher_funcs_tests.bats#L14

mlabuda: yes, we do, the bats used by launcher does not have docker inside, thus tests for launcher were failing where docker was involved. Problem of bats images is that they don't have docker. Another problem is that those images are not lightweight. On the contrary this image I used is base on eclipse/che-base which is based on alpine so it is really small image and has almost everything we need for tests.

benoitf: ok 
@TylerJewell do you have an idea of name for this new "bats" image ? if it needs to stay in `/dockerfiles/base/tests` (which is not like the other images) or move out to `/dockerfiles/<--some-name-->` but we have already a "tests" image.
I would think "bats" name is ok for me

dockerfiles/base/all_tests.sh
benoitf: the build of the image will be done as well by CI job.
so user won't need to build image locally as well

mlabuda:
Ok, I applied comments:
- creates dockerfiles/bats dir with Dockerfile and build.sh to build eclipse/che-bats image
- updated build.include to be more robust - process '--skip-tests' flag and it parse arguments to get tag
- tests run by default in dockerfiles/base/build.sh; skipped if '--skip-tests' flag is provided
- split test execution from test image build

benoitf (approved):
looks fine for me

TylerJewell (approved):

l0rd (approved):

dockerfiles/base/scripts/base/startup.sh
```diff
@@ -5,417 +5,7 @@
...
-  # a custom CLI assembly in their container and can set global variables which are 
-  # specific to that implementation of the CLI. This method must be called after
-  # networking has been established and initial images downloaded.
-  cli_pre_init
-
-  cli_init "$@"
-  cli_parse "$@"
-  cli_execute "$@"
-}
+source /scripts/base/startup_funcs.sh
```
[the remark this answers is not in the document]
mlabuda: @TylerJewell those file names are changed as you requested in comment above, I moved cli/cli-functions.sh to library.sh and we are using short notation "funcs" instead of "functions" and underscore instead of dash.

Regarding mounting scripts, it is still preserved. The tests uses different image and mount a local location of a scripts to the same location inside the test container and the test container then mount this location to /base/scripts, so the tests run with local scripts, which are mounted to the container. You can see it e.g. at https://github.com/eclipse/che/pull/3700/files#diff-1786fd5485e1f71d9b8ab5197898dcdbR42 where SCRIPTS_DIR is mounted to /scripts/base. This approach is used to test code in its current state.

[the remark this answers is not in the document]
benoitf: yes AFAIK @TylerJewell it's you that ask these names
and launch will be preserved

TylerJewell:
Thank you @mlabuda - there are now a couple new things that we must finish up (new rules of merge that we instituted this week for maintainers).  

1. Please add a matching PR to github.com/codenvy/che-docs that discusses how to run these unit tests.   We would be adding in content that describes the work at the bottom of this section.  https://www.eclipse.org/che/docs/setup/cli/index.html#cli-development.

2. Please add a statement (one line) which should appear in the Changelog.md

3. Please add a set of sentences or paragraphs that you would like for us to use in the release notes (marketing) that we send out to all users when we do a release.

Then we can go ahead and merge.

FYI - @mlabuda @benoitf - I am not sure it is related to this, but yesterday I was doing testing on a VirtualBox system with boot2docker on Windows and I was not able to volume mount :/scripts/base properly.  No matter how I tried to do it, I would just get an error.  I think the issue is that my Che repository was not in %userprofile%, and in perhaps the bootstrap, if on boot2docker, we should check all volume mounts to ensure that they are a sub-directory of %userprofile%

benoitf:
@TylerJewell on your boot2docker issue, is there a way to grab %userprofile% from a docker container ? if the user has not provided it

TylerJewell:
I will need to test this - I don't think it's possible today unless the user passes it in.

mlabuda:
Linkes with docs PR, added changelog line and release notes in the initial comment of this PR.

slemeur:
Perfect for me! Thanks @mlabuda ;)

benoitf:
@mlabuda calling ./build-all.sh from top level dockerfiles directory failed on my side
```
...
Script run successfully: eclipse/che-base:nightly
/Users/benoitf/Documents/git/codenvy/che/dockerfiles/base/test.sh: line 11: cd: base: No such file or directory
/Users/benoitf/Documents/git/codenvy/che/dockerfiles/base/test.sh: line 12: /../build.include: No such file or directory
```

calling individual build.sh is working fine

but I've on this run:
```
$ ./base/test.sh 
No tag provided, using nightly as default
Running tests in container from image eclipse/che-bats:nightly
Running functional bats tests for CLI prompts and usage
1..3
ok 1 test CLI prompt to provide volume for docker sock
ok 2 test CLI prompt to provide directory for user data
ok 3 test CLI 'usage' when running container without command
Running functional bats tests for init and destroy
1..4
not ok 1 test 'init' and 'destroy --quiet' with existing dir
# (in test file /Users/benoitf/Documents/git/codenvy/che/dockerfiles/base/tests/cmd_init_destroy_tests.bats, line 19)
#   `docker run -v $SCRIPTS_DIR:/scripts/base -v /var/run/docker.sock:/var/run/docker.sock -v $tmp_path:/data eclipse/che-cli:nightly init' failed with status 2
# INFO: Proxy: HTTP_PROXY=, HTTPS_PROXY=, NO_PROXY=*.local, 169.254/16
# WARN: Did not detect TTY - interactive mode disabled
# INFO: (che cli): Checking network... (hint: '--fast' skips version, network, and nightly checks)
# INFO: (che cli): Loading cli...
# INFO: (che download): Pulling image eclipse/che-cli:nightly
# 
# nightly: Pulling from eclipse/che-cli
# Digest: sha256:4b099974a0359e27a85d540dc1d4afccbae7541e4f96cf4daa55c27b7ef5d906
# Status: Downloaded newer image for eclipse/che-cli:nightly
# 
# WARN: Pulled new 'nightly' image - please rerun CLI
```

benoitf (changes requested):
I had two issues by checking the PR in its current state

mlabuda:
@benoitf thanks for catching those. First error was caused by nested sourcing, fixed it. Second is the behavior of the che-cli nightly - if there is a newer version of che-cli:nightly on the dockerhub, pull it and exit current command (there is a prompt to rerun it then). Fixed it by initial pull of the che-cli image in the test.sh.

benoitf:
@mlabuda but if I build a local image of the cli I would expect that it test mine, not a new one pulled from the remote.
Else how do I test my local code ?

mlabuda:
Tests mount local scripts to the che-cli image and overlayer existing scripts, so the local scripts are used instead. Because the che-cli dockerfile does not change, almost never, it is quite ok to use this approach. If it is not OK and you would like to also use the same che-cli image tag as the e.g. build.sh script produces, I can modify it.

benoitf:
@mlabuda yes I think it would be better if we use what we build locally (if we have build it). Because if we do heavy changes on che-cli Dockerfile (an example) it may be better to check what we've build. Else we may think it's ok, push it and then it will break later when image will be built on CI and pull locally

benoitf:
@mlabuda also as a side note, could you use --fixup when you perform new commits to your branch so we can see what has been changed so it's easier to track new changes)

mlabuda:
Okey. Tests now use same TAG for che-cli image as the one used for execution of  test.sh script or running of build.sh script. If no tag provided to scripts, default 'nightly' is used.

benoitf (approved):

TylerJewell:
Congrats - @mlabuda on your first major PR to be accepted.  We will now hold you accountable for making sure the rest of us add tests where appropriate :)

mlabuda:
@TylerJewell sure thing :) I will try to make next PRs less painful ;)

benoitf:
@riuvshin a new CI job will be required for https://github.com/eclipse/che/blob/master/dockerfiles/bats/

PR #3717, opened 2017-01-13
CHE-58: Fixing compilation error after rebase

ibuziuk:


l0rd:
LGTM Thank you @ibuziuk!

amisevsk:
Compilation without -Dskip-validate-sources still fails; the Copyright header in `DockerConnectorProvider.java` needs to have its year updated. We can just include this in a future PR though.

ibuziuk:
@amisevsk good catch. Maybe it would be better to fix it now, so that we will be ready to provide neat ready-to-review initial OpenshiftConnector PR once CQs are resolved?

l0rd:
+1 to fix it as soon as possible

ibuziuk:
@l0rd PR with license fix - https://github.com/eclipse/che/pull/3730

PR #3731, opened 2017-01-13
[cli] Fix #3655 : Check docker-compose file validity, result of docker-compose up and containers

### LINKED ISSUE #3655: CLI must check if all containers started properly
today CLI checks only is api server is available, but sometimes some containers may not start this usually happen due to docker bugs like after docker engine restart some container unable to start and the only way to make it work remove stopped container and restart. There is a lot of other possible scenarios where some container may not start so we must check every single container is it at least running container.

here is command how to do that check with inspect this command return `true` or `false`
```docker inspect -f {{.State.Running}} $CONTAINER_ID```

so all we neeed to do is to iterate across all codenvy containers and check if all is in running state.

here is a command how to get all existed (even stopped) codenvy containers:
```docker ps -qa --filter="name=codenvy_*"```

but that approach could be better if we will find a way how to do that check by container name because I think we could have situation where container was not created so we will not have `container id` to check. (Im not sure if that situation is possible)

### PR DESCRIPTION
**Pull request by benoitf**
Fix for https://github.com/eclipse/che/issues/3655

- check validity of docker-compose file by calling config method
- check result of "docker-compose up" command
- check each container of docker-compose services if they’re running or not

Change-Id: Ie7ef9cc301c2605ec40db70f00d06f1fca694e4c

### DIFF
--- a/dockerfiles/base/scripts/base/commands/cmd_start.sh
+++ b/dockerfiles/base/scripts/base/commands/cmd_start.sh
@@ -46,14 +46,16 @@ cmd_start() {
   COMPOSE_UP_COMMAND="docker_compose --file=\"${REFERENCE_CONTAINER_COMPOSE_FILE}\" -p=\"${CHE_MINI_PRODUCT_NAME}\" up -d"
 
   ## validate the compose file (quiet mode)
-  docker_compose --file=${REFERENCE_CONTAINER_COMPOSE_FILE} config -q || (error "Invalid docker compose file content at ${REFERENCE_CONTAINER_COMPOSE_FILE}" && return 2)
+  if local_repo; then
+    docker_compose --file=${REFERENCE_CONTAINER_COMPOSE_FILE} config -q || (error "Invalid docker compose file content at ${REFERENCE_CONTAINER_COMPOSE_FILE}" && return 2)
+  fi
 
   if ! debug_server; then
     COMPOSE_UP_COMMAND+=" >> \"${LOGS}\" 2>&1"
   fi
 
   log ${COMPOSE_UP_COMMAND}
-  eval ${COMPOSE_UP_COMMAND} || (error "Error while running compose up command" && tail -30 ${LOGS} && error "The command compose UP has failed. Latest 30lines of log files ${CHE_HOST_CONFIG}/cli.log have been printed." && return 2)
+  eval ${COMPOSE_UP_COMMAND} || (error "Error during 'compose up' - printing 30 line tail of ${CHE_HOST_CONFIG}/cli.log:" && tail -30 ${LOGS} && return 2)
 
   check_if_booted
 }
--- a/dockerfiles/base/scripts/base/commands/cmd_start.sh
+++ b/dockerfiles/base/scripts/base/commands/cmd_start.sh
@@ -45,12 +45,16 @@ cmd_start() {
   info "start" "Starting containers..."
   COMPOSE_UP_COMMAND="docker_compose --file=\"${REFERENCE_CONTAINER_COMPOSE_FILE}\" -p=\"${CHE_MINI_PRODUCT_NAME}\" up -d"
 
+  ## validate the compose file (quiet mode)
+  docker_compose --file=${REFERENCE_CONTAINER_COMPOSE_FILE} config -q || (error "Invalid docker compose file content at ${REFERENCE_CONTAINER_COMPOSE_FILE}" && return 2)
+
   if ! debug_server; then
     COMPOSE_UP_COMMAND+=" >> \"${LOGS}\" 2>&1"
   fi
 
   log ${COMPOSE_UP_COMMAND}
-  eval ${COMPOSE_UP_COMMAND}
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: dockerfiles/base/scripts/base/commands/cmd_start.sh [hunk 1]
```diff
@@ -45,12 +45,16 @@ cmd_start() {
   info "start" "Starting containers..."
   COMPOSE_UP_COMMAND="docker_compose --file=\"${REFERENCE_CONTAINER_COMPOSE_FILE}\" -p=\"${CHE_MINI_PRODUCT_NAME}\" up -d"
 
+  ## validate the compose file (quiet mode)
+  docker_compose --file=${REFERENCE_CONTAINER_COMPOSE_FILE} config -q || (error "Invalid docker compose file content at ${REFERENCE_CONTAINER_COMPOSE_FILE}" && return 2)
```
TylerJewell: What are the circumstances where the end user has provided a valid che.env, but this configuration check would fail?

dockerfiles/base/scripts/base/commands/cmd_start.sh
```diff
@@ -45,12 +45,16 @@ cmd_start() {
...
+
   if ! debug_server; then
     COMPOSE_UP_COMMAND+=" >> \"${LOGS}\" 2>&1"
   fi
 
   log ${COMPOSE_UP_COMMAND}
-  eval ${COMPOSE_UP_COMMAND}
+  eval ${COMPOSE_UP_COMMAND} || (error "Error while running compose up command" && tail -30 ${LOGS} && error "The command compose UP has failed. Latest 30lines of log files ${CHE_HOST_CONFIG}/cli.log have been printed." && return 2)
```
TylerJewell: I think this would be a bit more concise:

```
(error "Error during 'compose up' - printing 30 line tail of ${CHE_HOST_CONFIG}/cli.log:" && tail -30 ${LOGS} && return 2)
```

dockerfiles/base/scripts/base/commands/cmd_start.sh
```diff
@@ -159,3 +165,26 @@ check_if_booted() {
...
+    local IS_RUNNING_CONTAINER=$(docker inspect -f {{.State.Running}} ${CONTAINER_ID_MATCHING_SERVICE_NAME})
+    debug "Running state of container ${CONTAINER_ID_MATCHING_SERVICE_NAME} is ${IS_RUNNING_CONTAINER}"
+
+    if [[ ${IS_RUNNING_CONTAINER} != "true" ]]; then
+      error "The container with ID ${CONTAINER_ID_MATCHING_SERVICE_NAME} of docker-compose service ${DOCKER_COMPOSE_SERVICE_NAME} is not running, aborting."
+      docker inspect ${CONTAINER_ID_MATCHING_SERVICE_NAME}
+      return 2
+    fi
+  done <<< "${LIST_OF_COMPOSE_CONTAINERS}"
+}
```
TylerJewell: newline

dockerfiles/base/scripts/base/commands/cmd_start.sh
```diff
@@ -136,8 +140,10 @@ check_if_booted() {
     info "start" "Server logs at \"docker logs -f ${CHE_SERVER_CONTAINER_NAME}\""
   fi
 
+  check_containers_are_running
```
TylerJewell, line 145: Why do you need to check this before the server is booted command?

TylerJewell (changes requested):

dockerfiles/base/scripts/base/commands/cmd_start.sh
benoitf: maybe with only che.env it's not possible but a dev could break the docker-compose file

dockerfiles/base/scripts/base/commands/cmd_start.sh
benoitf: ok

dockerfiles/base/scripts/base/commands/cmd_start.sh
benoitf: It's to check that before and after the containers are still running

dockerfiles/base/scripts/base/commands/cmd_start.sh
benoitf: Ok

TylerJewell (approved):

riuvshin (approved):
+1 but please add requested changes from @TylerJewell

dockerfiles/base/scripts/base/commands/cmd_start.sh
benoitf: fixed

dockerfiles/base/scripts/base/commands/cmd_start.sh
benoitf: I've added that check only if "local_repo" is used.

dockerfiles/base/scripts/base/commands/cmd_start.sh
benoitf: fixed

benoitf:
@TylerJewell are the changes ok for you now ?

TylerJewell:
Yes - ok by me.

PR #3732, opened 2017-01-13
Update cli-functions.sh

riuvshin:
needed for https://github.com/eclipse/che/issues/3684
remove return statement due to issue where there is no changes to CLI nightly  image CLI think that is obsolete but it is actual as image was not changed.

### DIFF
--- a/dockerfiles/base/scripts/base/cli/cli-functions.sh
+++ b/dockerfiles/base/scripts/base/cli/cli-functions.sh
@@ -381,43 +381,22 @@ function verify_nightly_accuracy() {
   # than the one stored on DockerHub.
   if is_nightly; then
 
-    # Retrieve info on current nightly
-    LOCAL_CREATION_DATE=$(docker inspect --format="{{.Created }}" ${CHE_IMAGE_FULLNAME})
-    CURRENT_DATE=$(date)
+    CURRENT_DIGEST=$(docker images -q --no-trunc --digests ${CHE_IMAGE_FULLNAME})
 
-    # Unfortunatley, the "last_updated" date on DockerHub is the date it was uploaded, not created.
-    # So after you download the image locally, then the local image "created" value reflects when it
-    # was originally built, creating a situation where the local cached version is always older than
-    # what is on DockerHub, even if you just pulled it.
-    # Solution is to compare the dates, and only print warning message if the locally created ate
-    # is less than the updated date on dockerhub.
-
-    if $(newer_date_period \
-         $(timestamp_date_iso8601 "${LOCAL_CREATION_DATE}") \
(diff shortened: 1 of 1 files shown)


TylerJewell:
I think we have some logic that can detect if the pull was successful.  I will find that and we will return if you really pull and not return otherwise.

benoitf (approved):

riuvshin:
@TylerJewell @benoitf I think we should not check the date, because date will be always obosolete if we will not have changes to CLI. do you know is it possible to get image ID? from docker hub? what it we will compare local image ID and remote it they are different we do pull. (not sure if that works)

benoitf:
@riuvshin unfornatelly the digests are not available on dockerhub. This is why at a time we wanted that ci push that information somewhere each time a nightly was pushed

riuvshin:
@benoitf does digest -eq image id? maybe image id is available..

benoitf:
no we don't have it until we pull image locally :-/
with previous v1 registry we had but not with v2

TylerJewell:
Why don't we do a `docker inspect` of the existing image to get the digest ID, then do a `pull`, and then we do a new `docker inspect` after the pull to see if the digest has changed.  If the digest has not changed, we continue operation.  If the digest has changed, then we stop?

riuvshin:
@TylerJewell that could work, or we can
1 get current image id
2 do pull
3 compare image id with id after pull, 
3a if they are the same we proceed
3b if they are different we exit CLI and ask to rerun CLI

lol it seems it is exactly the same :D

TylerJewell:
I have a branch that is working with the digest comparison.  Let me push it up and see what people think.  I think that the digest check is a more reliable version for us to use.

TylerJewell:
Ok - so I have updated the branch with a different form of check.
1. Capture the existing images's digest with `docker images -q --no-trunc --digests`
2. Pull the image if `--fast` is not provided.
3. Capture the new images digest
4. And then compare - if the digests are different, stop the CLI and ask for a rerun

benoitf:
always pulling by default for nightly seems to be the best option. 
Maybe we could store in /data folder the date of the check and then not pull every time we launch the command.

TylerJewell:
Happy to add that improvement if you don't think the --fast optimization is enough.

I can only place that check into directories that have been initialized so it is not universal.  I like just having the fast option.

benoitf:
using fast option is OK for now we can improve later based on users feedback

TylerJewell:
Let's get @riuvshin input and if he is ok will merge.

riuvshin:
+1 (i cant approve this )

TylerJewell (approved):

Issue #3179, opened 2016-11-26
Failed to change port (regression)


ne0h:
Hi,
I do not manage to get che running with a non-default port (9080). I tried on Arch Linux and Ubuntu 14.04 LTS. 8080 works but changing the port via `export CHE_PORT=9080` does not work.
Error message in the terminal where I start docker:

```
INFO: ECLIPSE CHE: Found image eclipse/che-server:5.0.0-latest
INFO: ECLIPSE CHE: Starting container...
INFO: ECLIPSE CHE: Server logs at "docker logs -f che-server"
INFO: ECLIPSE CHE: Server booting...
---------------------------------------
ERROR: !!!
ERROR:      ECLIPSE CHE: Timeout waiting for server. Run "docker logs che-server" to inspect the issue.
ERROR: !!!
---------------------------------------
```

Docker log is attached here: [https://gist.github.com/ne0h/7e07dd868bdb2a8fc42130471be261b0](url)

**Reproduction Steps:**

1. Follow the guide: [https://www.eclipse.org/che/getting-started/download/](url) to download the startup file
2. `export CHE_PORT=9080`
3. `che start`

**Expected behavior:**

Everything is fine.

**Observed behavior:**

Che "does not come up".

**Che version:**    5.0.0-latest   
**OS and version:**Ubuntu 14.04 LTS, Arch Linux   
**Docker version:** 1.12.3, build 6b644ec 
**Che install:**    Docker container 

**Additional information:**

* Problem started happening recently, didn't happen in an older version of Che: No
* Problem can be reliably reproduced, doesn't happen randomly: Yes

Thanks in advance!
Max

TylerJewell:
Maximillian - I am investigating.  In the mean time, you can bypass the CLI and [launch the server] (https://eclipse-che.readme.io/docs/usage-docker-server) with this Docker syntax. And you can override the ports more directly with that.

TylerJewell:
Confirmed the issue with che-launcher. There is an adaptaptation regression that is causing some sort of variable passing error.

TylerJewell:
@ne0h - found the issue and  got a fix in.  This will show up in our nightly's shortly.

https://github.com/eclipse/che-dockerfiles/pull/50#pullrequestreview-10216972

TylerJewell:
@ne0h - CI systems updated and this issue should no longer be active.

ne0h:
@TylerJewell - Thanks a lot for your work!

[...]

PR #3734, opened 2017-01-14
Fixes mapping of custom port with CHE_PORT

TylerJewell:
### What does this PR do?
Fixes an issue where changing ports would prevent Che from starting due to an unexpected use of `CHE_PORT` being passed into `eclipse/che-server`.  Previously, we always assumed that the internal `eclipse/che-server` container would boot on port `8080` and then we just provide a port mapping from the custom port to the internal port. The new configuration approach actually passes `CHE_PORT` into `eclipse/che-server`, which causes the server to start internally on a different port. 

This quick fix maps the external port to the internal port.

#3179


riuvshin (approved):
good catch!

benoitf (approved):

PR #3749, opened 2017-01-16
Fix bug in CLI Output if user has overridden CHE_HOST

TylerJewell:
### What does this PR do?
Fixes issue uncovere by @garagatyi where the CLI output in two locations shows the wrong CHE_HOST value if the user has overridden CHE_HOST in the che.env file.

This implements a printout mechanism that uses Docker to print out the value and passes in `che.env` as an input into that.

### DIFF
--- a/dockerfiles/base/scripts/base/commands/cmd_debug.sh
+++ b/dockerfiles/base/scripts/base/commands/cmd_debug.sh
@@ -21,8 +21,7 @@ cmd_debug() {
   info "${CHE_PRODUCT_NAME}_CONFIG            = ${CHE_HOST_CONFIG}"
   local CHE_HOST_LOCAL=${CHE_HOST}
   if is_initialized; then
-    CHE_HOST_LOCAL=$(docker_run --env-file="${REFERENCE_CONTAINER_ENVIRONMENT_FILE}" \
-                                ${UTILITY_IMAGE_ALPINE} sh -c 'echo $CHE_HOST')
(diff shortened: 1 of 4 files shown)


dockerfiles/base/scripts/base/commands/cmd_debug.sh
```diff
@@ -19,7 +19,12 @@ cmd_debug() {
   info "${CHE_PRODUCT_NAME}_VERSION           = ${CHE_VERSION}"
   info "${CHE_PRODUCT_NAME}_INSTANCE          = ${CHE_HOST_INSTANCE}"
   info "${CHE_PRODUCT_NAME}_CONFIG            = ${CHE_HOST_CONFIG}"
-  info "${CHE_PRODUCT_NAME}_HOST              = ${CHE_HOST}"
+  local CHE_HOST_LOCAL=${CHE_HOST}
+  if is_initialized; then
+    CHE_HOST_LOCAL=$(docker_run --env-file="${REFERENCE_CONTAINER_ENVIRONMENT_FILE}" \
```
garagatyi: Maybe create function that retrieves variable by name? It is possible that other variables in different places will be also needed at some point

TylerJewell: Yeah, that is a good idea.  Let me think about how to do such a thing.

riuvshin (approved):
sweeet!

benoitf (approved):

Issue #3667, opened 2017-01-11
Use consistent versions of Che images used between nightly/latest


slemeur:
Utilities images are always pointing to "nightly" images.
https://github.com/eclipse/che/pull/3646#pullrequestreview-15885715

So that when you start a command using the eclipse/che:latest, it retrieves images that are `nightly`.

As a user, it feels strange and confusing. Would be expected that if starting with `latest` it uses always `latest` images, and if starting with `nightly` it uses always `nightly` images.

This will make debugging scenarios for users and customers hard to reproduce. 

**Che version:**   5.0.0, 5.0.1

riuvshin:
we already discussed this with @TylerJewell this is a bad thing. we agreed that we will have all those images in `dockerfiles/cli/version/<version>/images_utilities` file same as `images` file.
but not sure that @TylerJewell will be able to do that for 5.0.1.  that issue can be really critical only if  nightly will become incompatible, so we probably can delay this to 5.1.0

TylerJewell:
We will use this issue created by @slemeur as the tracking issue.

benoitf:
Fixed by https://github.com/eclipse/che/pull/3751

Issue #3666, opened 2017-01-11
Not consistent version of Che used with chedir utility


slemeur:
Starting Che using chedir utility with the command: `che dir up` is not starting the right Che version. 

**Reproduction Steps:**
1. Start Che using che dir command. (latest or nightly versions of eclipse/che CLI)
2. Check version: 5.0.0-M7

**Che version:**    5.0.0 , 5.0.1
**OS and version:**    MAC OS 10.12.2
**Docker version:** 1.12.3

benoitf:
Fixed by https://github.com/eclipse/che/pull/3751

PR #3751, opened 2017-01-16
[cli] #3666 do not hardcode version of utilities images to nightly

benoitf:
1. : we have `images-bootstrap` file inside the base image (with variable named BOOTSTRAP_IMAGE_*)  . For example BOOTSTRAP_IMAGE_ALPINE and BOOTSTRAP_IMAGE_CHEIP
2. : `images` file (with variable named IMAGE_*) (we have this file currently)
3. : `images-utilities` file (with variable named UTILITY_IMAGE_*) inside the base image For example UTILITY_IMAGE_CHEDIR

then we can apply the required policies
1. all images of images-bootstrap are loaded when we bootstrap cli
2. all images of images are loaded after bootstrap
3. all utilities images are pulled when one of them is asked

- removes all previous calls to the get_image_manifest as it’s performed by cli once. (In the past it was in another docker image but now it’s a local file parsing)
- update offline and rmi commands

Linked to #3666 and #3667

Change-Id: I7f768bd70d59de561fd469cf5228666b24ed2ded


riuvshin:
afaik @TylerJewell wanted to have those images in separate file to avoid pulling them by default.

benoitf:
UTILITY_IMAGE_ALPINE and UTILITY_IMAGE_CHEIP  are mandatory image so even putting in another file we will still need to pull them

riuvshin:
@benoitf Im +1 but please create same PRs for codenvy versions and customer-saas

TylerJewell:
I think that we can either leave images in the same file or in different files.  But we should honor the following policies:
1. The mandatory images are pulled as part of the bootstrap process no matter what. There is only two now.
2. The runtime images are pulled as part of the normal sequence related to every execution of the CLI after the bootstrap phase.  This should only be the runtime images and not the utility images.
3. We should load the utility images (all of them) if we execute a command that loads a utility. We would do this because these images do take some bulk, and we shouldn't necessarily load them unless necessary.

The cmd_offline then needs to be updated, too, as we now hard code all of the utility images to be downloaded, so this function should now use the images listed in registry instead of hard coding.

benoitf:
ok so what if we have one file for each of the policies
1. : we have `images-bootstrap` file (with variable named BOOTSTRAP_IMAGE_*)  . For example BOOTSTRAP_IMAGE_ALPINE and BOOTSTRAP_IMAGE_CHEIP
2. : `images` file (with variable named IMAGE_*) (we have this file currently)
3. : `images-utilities` file (with variable named UTILITY_IMAGE_*) For example UTILITY_IMAGE_CHEDIR


then we can apply the required policies
1. all images of images-bootstrap are loaded when we bootstrap cli
2. all images of images are loaded after bootstrap
3. all utilities images are pulled when one of them is asked

TylerJewell:
Yes, ok with this.  We also have `images-stacks`, too.

benoitf:
Ok I've updated this PR.
@slemeur do we have some doc around this internal stuff ? I haven't found

@riuvshin I will issue PR for codenvyartik and customer-saas before we merge it

riuvshin (approved):

TylerJewell (approved):

TylerJewell:
@benoitf - based upon another conversation that I am having with @JamesDrummond - it seems that we need to take a look at cmd_rmi.sh in the context of these changes. It seems that this function only removes `/images` file, but is not doing it for the other files.  So we should probably update the rmi function to be consistent - similar to what is done with the offline method.

TylerJewell (changes requested):
We need to now revisit the rmi function as it's not removing all of the images that are in the registry.

RMI is supposed to remove all images in the registry, but not the CLI itself.

benoitf:
I was also thinking. Do we need that bootstrap_images be in the cli and not in the base itself.
For example if we later add a new bootstrap image, we will need to update all CLIs (like che, codenvy, artik, etc) but it may look strange to have to update these cli while the images are more for the core itself which is the cli-base

so moving `cli/version/nightly/images-bootstrap` to be at `base/scripts/base/images-bootstrap`

TylerJewell:
Yes, I believe this makese sense.  @riuvshin - FYI, this moves some important stuff to a different location.

benoitf:
@TylerJewell @riuvshin I think now it's matching your expectations
I removed as well image puppet and registry as they're not used

TylerJewell (approved):

riuvshin (approved):

benoitf:
I'm waiting CI to be ok before merging this PR
https://ci.codenvycorp.com/view/nightly-builds/job/che-base-docker-image-nightly/ is failing due to https://github.com/eclipse/che/pull/3700#issuecomment-274094774

riuvshin:
@benoitf i've added ci job for bats, now it is there but base image still failing https://ci.codenvycorp.com/view/nightly-builds/job/che-base-docker-image-nightly/84/console

I see there 4 `not OK` tests

riuvshin:
ok now build fixed. it was lack space on disk.

PR #3753, opened 2017-01-16
Fix #3686 : Property around external docker has been changed in a refactoring

### LINKED ISSUE #3686: Che nightly does not start on Windows 10
Syntax used:

`docker run -ti -v /c/Users/Codenvy/che-conf:/data -v /var/run/docker.sock:/var/run/docker.sock eclipse/che-cli:nightly start --pull`

When starting a workspace, server and agent can connect to each other, but the browser cannot reach workspace agent:

image

For whatever reason, it tries to connect to VM's docker0 network interface IP instead of the VM IP which is `10.0.75.2`

Playing with CHE_HOST did not help.

It does not happen with eclipse/che-cli:latest

benoitf:
but it has not been updated everywhere

I did a look for 10 minutes as blocker https://github.com/eclipse/che/issues/3686 is not solved for a long time?
it seems to be broken by https://github.com/eclipse/che/pull/3282 merge

### What does this PR do?
Fix boot of Che workspaces on mac (and maybe windows) with CLI

https://github.com/eclipse/che/issues/3686

Change-Id: I6da2c6f33cc063dda08b953eb2fab16abe4a0c86

### DIFF
--- a/dockerfiles/che/entrypoint.sh
+++ b/dockerfiles/che/entrypoint.sh
@@ -264,12 +264,11 @@ init() {
   sed -i "/che.workspace.terminal_linux_amd64=/c\che.workspace.terminal_linux_amd64=${CHE_DATA_HOST}/lib/linux_amd64/terminal" $CHE_LOCAL_CONF_DIR/che.properties
   sed -i "/che.workspace.terminal_linux_arm7=/c\che.workspace.terminal_linux_arm7=${CHE_DATA_HOST}/lib/linux_arm7/terminal" $CHE_LOCAL_CONF_DIR/che.properties
 
-  # CHE_DOCKER_MACHINE_HOST_EXTERNAL must be set if you are in a VM.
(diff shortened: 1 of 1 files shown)


TylerJewell (approved):

TylerJewell:
Thanks for finding this.  I had assumed that this was internal to the Che server code because alex had indicated internal restructuring.  what must have happened is that they have stopped using the old property name and only now use the new property name.  Or, the mapping of the old property to new property somehow is failing.

Thanks for fixing this, Florent.

riuvshin (approved):

amisevsk:
Sorry I missed this in #3282. I think the issue was that previously, [`LocalDockerInstanceRuntimeInfo`](https://github.com/eclipse/che/blob/229a40933a9283553eb0e30bb1fd39451e75057a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalDockerInstanceRuntimeInfo.java#L81) was still relying on old env vars from before the switch to using che.properties. That class was removed in #3282, resulting in this problem.

Issue #3492, opened 2016-12-23
Avoid caching problem in http resources


skabashnyuk:
1. All resources except that have word "cache" in names should not be cached and must be revalidated.
2. Resources that have in name word "cache" have to be cached for long period. Ex: 1 month.

PR #3771, opened 2017-01-17
CHE-3492; Add filters for force or disable caching on the given resources paths.

mshaposhnik:
### What does this PR do?
Add filters for force or disable caching on the given resources paths.

#3492 
### Previous behavior
None of the resources have proper cache control headers.

### New behavior
Resources will have proper cache control headers.

### Tests written?
Yes

### Docs updated?

#### Changelog
Add filters for force or disable caching on the given resources paths.


benoitf:
> What issues does this PR fix or reference?
CHE-3492

maybe we should provide hyperlink to the issue ? because CHE-3492 is not a valid github name while #3492 is

skabashnyuk:
done

benoitf:
does it break http etag ?
https://en.wikipedia.org/wiki/HTTP_ETag

https://github.com/eclipse/che/blob/8e5967613daddac84d9e3ea0ef47d53d08dbd11e/core/che-core-api-core/src/main/java/org/eclipse/che/everrest/ETagResponseFilter.java

mshaposhnik:
@benoitf  No, it shouldn't. We use etag filter only for JSON-s, while the caching is oriented for IDE resources, which is mostly .js .css etc. But even if it is, etag  will just  help, so if the cached resource changed on the server side, another e-tag hash will be published and resource will be re-get from browser.

benoitf:
@mshaposhnik ok thanks I was not sure if this filter was applying or not to JSON resources

mshaposhnik:
@benoitf Should we apply those filters to dashbard war (in Codenvy we do block all caching by pattern `(?!_sso/).*$`, e.g. block caching for anything except `_sso` (which is excluded for login purposes i think))

benoitf:
@mshaposhnik AFAIK you could do it on js and css of dashboard. Resources are minified and then each release has its own filename
like
scripts/app-cc9dbfedf4.js
scripts/vendor-b2dfa25cb3.js
styles/app-5e7a694a55.css

mshaposhnik:
@benoitf So, i don't see the reason why we should not cache them. If it changes, it will have another stamp and re-downloaded. Why do we blocked caching in codenvy, do you remember?

mshaposhnik:
https://github.com/codenvy/codenvy/blob/master/assembly/onpremises-ide-packaging-war-dashboard/src/main/java/com/codenvy/onpremises/deploy/DashboardServletModule.java#L50 that is what i mean

benoitf:
@mshaposhnik yes I've understood.
probably a copy/paste error. BTW dashboard may still load some fragments (not .js or .css that are not minified and then have unique identifiers) so cache could not apply on the whole files. Those files are loaded from "assets/" folder of dashboard.war

mshaposhnik:
@skabashnyuk @benoitf  pls review

benoitf (approved):

skabashnyuk (approved):

PR #3800, opened 2017-01-18
Fixes display URL error when not initialized

TylerJewell:
### What does this PR do?
We recently made an improvement to have more flexibility in the display URL that is printed out.  We derive the value of `CHE_HOST` from a `che.env` file. However, this function can be called when the `:/data` folder has not been initialized generated an error that it could not find the `che.env` file.  

This adds a check to only perform this call if the folder has been initialized otherwise default to the in-memory value.

Note - all new users and installations will see this error.  It will not stop the starting of the system, but it is an ugly looking error.

#### Changelog
Fixes display URL error when not initialized

### DIFF
--- a/dockerfiles/base/scripts/base/cli/cli-functions.sh
+++ b/dockerfiles/base/scripts/base/cli/cli-functions.sh
@@ -52,7 +52,11 @@ get_display_url() {
 
   # If the user has modified che.env with a custom CHE_HOST, we need to detect that here
   # and not use the in-memory one which is always set with eclipse/che-ip.
-  local CHE_HOST_LOCAL=$(get_value_of_var_from_env_file CHE_HOST)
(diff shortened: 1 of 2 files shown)


TylerJewell:
@riuvshin - we should issue a 5.1.1 fix with this.

benoitf (approved):

riuvshin (approved):

riuvshin:
ok, I will think tomorrow how we can do bugfix without release. (as there is only CLI changes I will need think how to do a tag of this.)

TylerJewell:
@benoitf - based upon another conversation that I am having with @JamesDrummond - it seems that we need to take a look at cmd_rmi.sh in the context of these changes. It seems that this function only removes `/images` file, but is not doing it for the other files.  So we should probably update the rmi function to be consistent - similar to what is done with the offline method.

PR #3813, opened 2017-01-19
Rework EnvironmentContext#getSubject method to not return null subject

sleshchenko:
### What does this PR do?
Rework EnvironmentContext#getSubject method to not return null subject. It allows to avoid null pointer exceptions in case missing component which should set subject for request in custom packaging.

### Previous behavior

### New behavior

### Tests written?
No

### Docs updated?

#### Changelog
Rework EnvironmentContext#getSubject method to not return null subject

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
@@ -798,8 +798,8 @@ public class WorkspaceManager {
     }
 
     private String sessionUserNameOr(String nameIfNoUser) {
-        final Subject subject;
-        if (EnvironmentContext.getCurrent() != null && (subject = EnvironmentContext.getCurrent().getSubject()) != null) {
+        final Subject subject = EnvironmentContext.getCurrent().getSubject();
+        if (!subject.isAnonymous()) {
             return subject.getUserName();
         }
         return nameIfNoUser;
--- a/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequest.java
+++ b/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequest.java
@@ -206,7 +206,7 @@ public class DefaultHttpJsonRequest implements HttpJsonRequest {
                                                                               UnauthorizedException,
                                                                               ConflictException,
                                                                               BadRequestException {
-        final String authToken = getAuthenticationToken();
+        final String authToken = EnvironmentContext.getCurrent().getSubject().getToken();
         final boolean hasQueryParams = parameters != null && !parameters.isEmpty();
         if (hasQueryParams || authToken != null) {
             final UriBuilder ub = UriBuilder.fromUri(url);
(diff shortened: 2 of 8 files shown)


gazarenkov (approved):

skabashnyuk (approved):

voievodin (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -799,7 +799,7 @@ private Subject sessionUser() {
 
     private String sessionUserNameOr(String nameIfNoUser) {
         final Subject subject;
-        if (EnvironmentContext.getCurrent() != null && (subject = EnvironmentContext.getCurrent().getSubject()) != null) {
+        if ((subject = EnvironmentContext.getCurrent().getSubject()) != Subject.ANONYMOUS) {
```
benoitf: shouldn't we use !EnvironmentContext.getCurrent().getSubject().equals(Subject.ANONYMOUS) instead of != ?
(I agree that it's testing for now the memory pointer but it may have side effect in future as I don't see tests)

sleshchenko: @benoitf 
When can comparing by `==` work incorrect? Looks like only in case changing Subject from interface to class and making this field non final and setting some values.... I can't give the full variant :)
Equals can works incorrect if we change anonymous implementation to SubjectImpl("0000-000-00", "Anonynous", null, false) then equals won't show that it is really anonymous that are returned by EnvironmentContext, or it is some Subject with id "0000-000-00" and name "Anonynous". (I agree that it is crazy example)
Do you think that using equals there is correct variant?

benoitf: @sleshchenko maybe a test is missing in https://github.com/sleshchenko/che/blob/cd0fb094cfac053c8bb618fac13200ac1880f17b/wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/workspace/server/WorkspaceManagerTest.java
to ensure it will always work as designed

sleshchenko: @benoitf So this method is used only for logging. I don't know any way to test it and I think we can miss it(Mb you know?). Also we could make this method package private and test only its functionality, but I don't like this way in this case. Because it will be more testing of `EnvironmentContext#getSubject` method than `sessionUserNameOr` functionality :) 
Mb the best way is adding isAnonymous() method to Subject and don't use subject comparing in such use cases.

benoitf: yes I was thinking the same on having isAnonymous() so it delegates the test to Subject and not to all callers

sleshchenko: @benoitf Please, review my new changes

benoitf: @sleshchenko looks perfect for me

PR #3823, opened 2017-01-20
Rework personal account to be optional for packaging.

sleshchenko:
### What does this PR do?
Makes personal account as optional by reworking UserImpl object not to include personal account.
Adds methods in AccountManager to add abilities to provide personal account manually.
It allows not to have personal accounts in each packaging.
Add missed indexes for type field of account and foreign key to account id in workspace table. Indexes for foreign keys is required to speed up getting records by this field. Also it can speed up another operation like removing because this field can be used by data base for consistency checking.

### Previous behavior
Personal accounts are integrated to UserImpl. In this way each user has his own personal account.

### New behavior
Reworked personal account not to be required for each packaging.

#### Changelog
- Internal refactoring: move account out of user entity
- Allow personal account to be optional
- Add missed indexes for type field of account and foreign key to account id in workspace table. 

#### Release Notes
- N/A

#### Docs PR
N/A

### Docs updated?
N/A

### Tests written?
Yes

### DIFF
--- a/wsmaster/che-core-api-account/src/main/java/org/eclipse/che/account/api/AccountManager.java
+++ b/wsmaster/che-core-api-account/src/main/java/org/eclipse/che/account/api/AccountManager.java
@@ -12,6 +12,8 @@ package org.eclipse.che.account.api;
 
 import org.eclipse.che.account.shared.model.Account;
 import org.eclipse.che.account.spi.AccountDao;
+import org.eclipse.che.account.spi.AccountImpl;
+import org.eclipse.che.api.core.ConflictException;
 import org.eclipse.che.api.core.NotFoundException;
 import org.eclipse.che.api.core.ServerException;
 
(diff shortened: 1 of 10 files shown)


skabashnyuk (approved):

akorneta (approved):

wsmaster/che-core-sql-schema/src/main/resources/che-schema/5.2.0/2__move_account_out_of_user.sql
```diff
@@ -0,0 +1,23 @@
...
+-- which accompanies this distribution, and is available at
+-- http://www.eclipse.org/legal/epl-v10.html
+--
+-- Contributors:
+--   Codenvy, S.A. - initial API and implementation
+--
+
+ALTER TABLE Usr ADD name VARCHAR(255);
+
+UPDATE Usr
```
voievodin: can you please use lowercase for all the names except of SQL keywords(In the same script table name is in both lower case and uppercase).

sleshchenko: Yes, I can. Thank for remark. I'll fix it

wsmaster/che-core-sql-schema/src/main/resources/che-schema/5.2.0/2__move_account_out_of_user.sql
```diff
@@ -0,0 +1,23 @@
...
+SET name = ( SELECT name
```
evoevodin: Please try 

```sql
UPDATE usr
INNER JOIN account ON usr.account_id = account.id
SET user.name = account.name
```
and also indexing usr.account_id before

voievodin (approved):

Issue #3760, opened 2017-01-17
Amend previous commit works incorrectly


artaleks9:
When I try to change commit message  by the 'Amend previous commit' in the commit form, I have unexpected error.

**Reproduction Steps:**
* Open in the IDE a project (e.g. web-java-spring from template).
* Make a change in a file and perform commit.
* Try to change the massage of commit by 'Amend previous commit'. 

**Expected behavior:**
commit message should be changed. 

**Observed behavior:**
nothing is changed and there is unexpected error.

Che version: 5.1.0-SNAPSHOT
OS and version: Ubuntu 14.04
Docker version: 1.12

**Che cli.log output:** [cli.log.zip](https://github.com/eclipse/che/files/710629/cli.log.zip)

**Additional information:**
* Problem can be reliably reproduced, doesn't happen randomly: [Yes]
* See the attachment:
![amend_prev_commit](https://cloud.githubusercontent.com/assets/5408906/22016148/d653e7c6-dcad-11e6-9e41-11ac859b776e.gif)

vinokurig:
Waiting for QA Tests adoption

PR #3833, opened 2017-01-20
CHE-3760: Fix git commit not staged specified path with amend

vinokurig:
### What does this PR do?
Fix throwing exception if working directory is clean or nothing added to commit.
Add a workaround to allow empty commit with amend and specified paths.

https://github.com/eclipse/che/issues/3760

#### Changelog
BugFix: Fix throwing exception on Git commit, if working directory is clean or nothing added to commit.

#### Release Notes
BugFix N/A

#### Docs PR
BugFix N/A

### DIFF
--- a/wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
+++ b/wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
@@ -547,49 +547,81 @@ class JGitConnection implements GitConnection {
     @Override
     public Revision commit(CommitParams params) throws GitException {
         try {
-            String message = params.getMessage();
-            GitUser committer = getUser();
-            if (message == null) {
-                throw new GitException("Message wasn't set");
+            // Check repository state
+            RepositoryState repositoryState = repository.getRepositoryState();
+            if (!repositoryState.canCommit()) {
+                throw new GitException(format(MESSAGE_COMMIT_NOT_POSSIBLE, repositoryState.getDescription()));
             }
+            if (params.isAmend() && !repositoryState.canAmend()) {
+                throw new GitException(format(MESSAGE_COMMIT_AMEND_NOT_POSSIBLE, repositoryState.getDescription()));
+            }
+
+            // Check committer
+            GitUser committer = getUser();
             if (committer == null) {
                 throw new GitException("Committer can't be null");
             }
-
-            //Check that there are staged changes present for commit, or any changes if is 'isAll' enabled, otherwise throw exception
-            Status status = status(StatusFormat.SHORT);
-            if (!params.isAmend() && !params.isAll()
-                && status.getAdded().isEmpty() && status.getChanged().isEmpty() && status.getRemoved().isEmpty()) {
-                throw new GitException("No changes added to commit");
-            } else if (!params.isAmend() && params.isAll() && status.isClean()) {
-                throw new GitException("Nothing to commit, working directory clean");
-            }
-
             String committerName = committer.getName();
             String committerEmail = committer.getEmail();
             if (committerName == null || committerEmail == null) {
                 throw new GitException("Git user name and (or) email wasn't set", ErrorCodes.NO_COMMITTER_NAME_OR_EMAIL_DEFINED);
             }
-            if (!repository.getRepositoryState().canCommit()) {
-                Revision rev = newDto(Revision.class);
-                rev.setMessage(format(MESSAGE_COMMIT_NOT_POSSIBLE, repository.getRepositoryState().getDescription()));
-                return rev;
+
+            // Check commit message
+            String message = params.getMessage();
+            if (message == null) {
+                throw new GitException("Message wasn't set");
             }
 
-            if (params.isAmend() && !repository.getRepositoryState().canAmend()) {
-                Revision rev = newDto(Revision.class);
-                rev.setMessage(format(MESSAGE_COMMIT_AMEND_NOT_POSSIBLE, repository.getRepositoryState().getDescription()));
-                return rev;
+            Status status = status(StatusFormat.SHORT);
+            List<String> specified = params.getFiles();
+
+            List<String> staged = new ArrayList<>();
+            staged.addAll(status.getAdded());
+            staged.addAll(status.getChanged());
+            staged.addAll(status.getRemoved());
+
+            List<String> changed = new ArrayList<>(staged);
+            changed.addAll(status.getModified());
+            changed.addAll(status.getMissing());
+
+            List<String> specifiedChanged = specified.stream()
+                                                     .filter(path -> changed.stream().anyMatch(c -> c.startsWith(path)))
+                                                     .collect(Collectors.toList());
+
+            // Check that there are changes present for commit, if 'isAmend' is disabled
+            if (!params.isAmend()) {
+                // Check that there are staged changes present for commit, or any changes if 'isAll' is enabled
+                if (status.isClean()) {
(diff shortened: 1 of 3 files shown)


skabashnyuk (approved):

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -564,14 +589,27 @@ public Revision commit(CommitParams params) throws GitException {
             }
 
             CommitCommand commitCommand = getGit().commit()
-                                                  .setCommitter(committerName, committerEmail).setAuthor(committerName, committerEmail)
+                                                  .setAllowEmpty(params.isAmend())
```
sleshchenko: I'd move this setting to `.setAmend(...)` and add comment why is it needed

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -564,14 +589,27 @@ public Revision commit(CommitParams params) throws GitException {
...
-            }
+            /*
+            When committing  with amend flag specified paths that don't have any changes, JGitInternalException will be thrown.
+            According to setAllowEmpty method documentation, setting this flag to true must allow such commit,
+            but it won't because Jgit has a bug: https://bugs.eclipse.org/bugs/show_bug.cgi?id=510685.
+            As a workaround, add only changed and specified paths to specified paths of the commit command, and throw exception,
+            if other changes are present, but the list of changed and specified paths is empty, to prevent committing other paths.
+            TODO Remove throwing exception when the bug will be fixed.
+            */
+            if (!specified.isEmpty() && !(params.isAll() ? changed.isEmpty() : staged.isEmpty()) && specifiedChanged.isEmpty()) {
```
sleshchenko: I'd move this check above to other checks before creating commit comand

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -537,13 +539,36 @@ public Revision commit(CommitParams params) throws GitException {
...
+                                           .collect(joining());
+            if (!errorMessage.isEmpty()) {
+                throw new GitException(errorMessage);
+            }
+
+            // Check that there are changes present for commit, if 'isAmend' is disabled
+            if (!params.isAmend()) {
+                // Check that there are staged changes present for commit, or any changes if 'isAll' is enabled
+                if (status.isClean()) {
+                    throw new GitException("Nothing to commit, working directory clean");
```
sleshchenko, line 600: Should we throw exception that are related to changes and then about missing commiter's name and email?
Please check behavior of native git and do in the same way.
Any way I think we should check repository state and https://github.com/eclipse/che/pull/3833/files#diff-7f323c0167c366f41270f82a35ed368dR579 and only then throw exception related to changes.

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
sleshchenko: Please double check that it is right check. Looks like we should throw exception [here ](https://github.com/eclipse/che/pull/3833/files#diff-7f323c0167c366f41270f82a35ed368dR569) in case when specified is not empty and specifiedChanged is empty.

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
vinokurig: done

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
vinokurig: Moved the check. Added tests for this exception.

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
vinokurig: Moved the check

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -528,50 +530,86 @@ protected void onEndTask(String taskName, int workCurr, int workTotal, int perce
     @Override
     public Revision commit(CommitParams params) throws GitException {
         try {
-            String message = params.getMessage();
-            GitUser committer = getUser();
-            if (message == null) {
-                throw new GitException("Message wasn't set");
+            // Check repository state
+            if (!repository.getRepositoryState().canCommit()) {
```
sleshchenko: Is `getRepositoryState` easy operation? I'd save result to variable and then reuse it

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -528,50 +530,86 @@ protected void onEndTask(String taskName, int workCurr, int workTotal, int perce
...
 
             CommitCommand commitCommand = getGit().commit()
-                                                  .setCommitter(committerName, committerEmail).setAuthor(committerName, committerEmail)
+                                                  .setCommitter(committerName, committerEmail)
+                                                  .setAuthor(committerName, committerEmail)
                                                   .setMessage(message)
                                                   .setAll(params.isAll())
-                                                  .setAmend(params.isAmend());
+                                                  .setAmend(params.isAmend())
+                                                  // Allow empty commits with specified paths with amend
```
sleshchenko: comment says about commits with specified paths but code in next line doesn't do it. Fix code or comment

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -528,50 +530,86 @@ protected void onEndTask(String taskName, int workCurr, int workTotal, int perce
...
+                    throw new GitException("No changes added to commit");
+                }
+            /*
+            By default Jgit doesn't allow to commit not changed specified paths. According to setAllowEmpty method documentation,
+            setting this flag to true must allow such commit, but it won't because Jgit has a bug:
+            https://bugs.eclipse.org/bugs/show_bug.cgi?id=510685. As a workaround, specified paths of the commit command will contain
+            only changed and specified paths. If other changes are present, but the list of changed and specified paths is empty,
+            throw exception to prevent committing other paths. TODO Remove this check when the bug will be fixed.
+            */
+            } else if (!specified.isEmpty() && !(params.isAll() ? changed.isEmpty() : staged.isEmpty()) && specifiedChanged.isEmpty()) {
```
sleshchenko: To make comment clear I'd rewrite it as 
```
if (!params.isAmend()) {
  ...
} else {
    // comment with TODO to remove this block
    if (...) {
       ...
    }
}
```

sleshchenko (approved):
Other looks good to me!

Issue #3834, opened 2017-01-20
CLI - Custom Assembly, Version Bug, Latest Checks


TylerJewell:
This issue is a collection of various improvements.
1. The version command was broken with changes to the DockerHub API. The API now uses pagination to display lists of tags.  So we will only display the first 10 tags and then print out the next pagination link to display the older versions.  Only perform this check if we are not in offline mode.

2. Move offline mode to be a check that happens during bootstrap, not later when we are testing networking mode.

3. Updates some of the code to use `is_offline()` and `is_fast()` utility functions.

4. Add in concept of `:/assembly` mount which will directly mount an assembly and bypass the `:/repo` mode, which uses the repo for the assembly AND other stuff like puppet manifests.

5. If you are not in offline mode, now performs a check of the current image's version against the first 10 versions returned from DockerHub.  If a newer version is available on DockerHub, the version is printed out as a warning.

PR #3835, opened 2017-01-20
[cli] Version command fix, add :/assembly mount option

TylerJewell:
### What does this PR do?
1. Fixes broken version command where DockerHub now uses pagination for listing tags. Also only performs this check if we are not in offline mode.

2. Adds in new `:/assembly` mount option. If set, will use this assembly location for binaries.  If both `:/repo` and `:/assembly` are both set, then `:/assembly` will be used as precedence.
 
3. Fixes a bug where `CHE_IMAGE_NAME` was not properly derived if the image name did not have a colon as a tag at the end of it.

4. If you are not in offline mode, now performs a check of the current image's version against the first 10 versions returned from DockerHub.  If a newer version is available on DockerHub, the version is printed out as a warning.

5. Adds in new utility functions to improve performance of preflight checks.  Now, the port checking accepts an array as input and then generates a single `docker run` command with all ports in the array to perform a single check.  if this check passes, then we print out all of the ports at the same time.  Only if that command fails do we iterate through each port individually.  

#3834 

### Changelog
Add `:/assembly` volume mount to CLI to direct mount binaries without a repo
Fix CLI `version` command to handle DockerHub pagination
CLI adds check for newer versions on DockerHub
CLI preflight check perf improvements

### Release Notes
With the CLI, you can now volume mount `<path>:/assembly` to mount a folder with Che binaries in it.  You can get a similar behavior already by mounting `:/repo`, but this mount also makes use of local CLI and puppet manifest files. Mounting `:/assembly` will only take a new assembly as a set of binaries. This is ideal for custom assembly and plugin developers that are creating assembly binaries with maven.

### Docs PR
https://github.com/eclipse/che-docs/pull/90


benoitf (dismissed):
```
$ docker run -it -v /var/run/docker.sock:/var/run/docker.sock -v /tmp/che-cli:/data eclipse/che-cli:nightly version --fast
WARN: Skipping dockerhub network check...
WARN: Skipping nightly image check...
WARN: Skipping version compatibility check...

Your CLI version is 'nightly'.
Your installed version is 'nightly'.
Available on DockerHub:
  latest
  5.1.1
  nightly
  5.1.0
  5.0.1
  5.0.0
  5.0.0-latest
  5.0.0-M8
```

riuvshin (approved):

TylerJewell:
Sure, absolutely.

Request has come from @skabashnyuk.  The scenario is that they are building a custom assembly utility to generate minimum custom assembly with custom plugins.  And then they want to let user to build that.

The custom assembly that they builld has an /assembly folder and a /plugins folder, but no /dockerfiles folder. So they want to use the CLI to run the assembly after it has been generated.  This allows for just using the binary without also using the puppet manifests.

riuvshin:
I see, thx, so if you want to do same for codenvy please make sure you mounting not only tomcat all in one but agents too. because in CHE agents in the che assembly but in codenvy they are separate and should be mounted to another container.

TylerJewell:
We should continue that dialog in the Codenvy repo.  I ran into that issue and didn't have a good solution because it would require three volume mounts.  so to start, I only mounted the Tomcat assembly and ignored the wsagent and terminal assemblies (however those ARE mounted if you mount :/repo)

skabashnyuk (approved):

dockerfiles/base/scripts/base/library.sh
```diff
@@ -345,9 +344,17 @@ verify_version_compatibility() {
   ##      - If they don't match and one is nightly, fail
   ##      - If they don't match, then if CLI is older fail with message to get proper CLI
   ##      - If they don't match, then if CLLI is newer fail with message to run upgrade first
-
   CHE_IMAGE_VERSION=$(get_image_version)
 
+  # Only check for newer versions if not in offline mode.
+  if ! is_offline; then
+    NEWER=$(compare_versions $CHE_IMAGE_VERSION)
```
benoitf, line 351: If I use a tagged version ( not latest alias) I'm not sure we shouldn't do the check. Only in case of latest

dockerfiles/base/scripts/base/commands/cmd_version.sh
```diff
@@ -20,16 +20,66 @@ cmd_version() {
...
+    if [ $NUMBER_OF_VERSIONS -gt $DISPLAY_LIMIT ]; then
+      OLDER_VERSION=$(echo $VERSION_LIST_JSON | jq '.next')
+      text "  See older versions at: $OLDER_VERSION\n"
+    fi
+  fi
+}
+
+# Compares $1 version to the first 10 versions listed as tags on Docker Hub
+# Returns "" if $1 is newest, otherwise returns the newest version available
+compare_versions() {
```
benoitf: As core of cli is calling fonction it should not be long to cmd_version

dockerfiles/base/scripts/base/commands/cmd_version.sh
```diff
@@ -20,16 +20,66 @@ cmd_version() {
...
+  local VERSION_LIST_JSON=$(curl -s https://hub.docker.com/v2/repositories/${CHE_IMAGE_NAME}/tags/)
+  local NUMBER_OF_VERSIONS=$(echo $VERSION_LIST_JSON | jq '.count')
+
+  DISPLAY_LIMIT=10
+  if [ $DISPLAY_LIMIT -gt $NUMBER_OF_VERSIONS ]; then 
+    DISPLAY_LIMIT=$NUMBER_OF_VERSIONS
+  fi
+
+  # Strips off -M#, -latest version information
+  BASE_VERSION=$(echo $1 | cut -f1 -d"-")
```
benoitf: How does it behave with milestone
It's not clear to me
It's a good candidat to a unit test

dockerfiles/base/scripts/base/library.sh
TylerJewell: When I started developing this feature I was only going to do the check if it was latest.  But while implementing it, I saw more value in it for the tagged versions!  

The tagged versions was how I was able to implement some tests for it.  But I thought about leaving it in and then discussing the trade offs in this issue with others.

dockerfiles/base/scripts/base/commands/cmd_version.sh
TylerJewell: The milestone and rc markers were causing failures on version comparison tests.   And so for this exercise I figured that the version comparison test did not need to account for the milestone markers until we consider them again in the future.

So this strips off the milestone and just uses the numerical comparator.

dockerfiles/base/scripts/base/commands/cmd_version.sh
TylerJewell: Yes, agree.

dockerfiles/base/scripts/base/library.sh
```diff
@@ -517,3 +524,92 @@ wait_until_server_is_booted() {
...
+
+  # Need to evaluate a string otherwise the ports will be seen as strings and not parameters
+  EXECUTION_STRING="docker run -d ${DOCKER_PORT_STRING} \
+                      --name fake ${UTILITY_IMAGE_ALPINE} \
+                          httpd -f ${HTTPD_PORT_STRING} -h /etc/ > /dev/null 2>&1"
+
+  eval ${EXECUTION_STRING}
+  NETSTAT_EXIT=$?
+
+  docker rm -f fake > /dev/null 2>&1
```
benoitf: I've tried on ubuntu and mac and I don't need to use httpd to fail if port mapping is already in use

```
benoitf@ubuntu:~/che/dockerfiles$ docker run -p 22:22  -ti alpine:3.4 sh -c "echo hello"
docker: Error response from daemon: driver failed programming external connectivity on endpoint angry_hawking (a1ffa937fac9bf49ddfa8bfd1cc42199c887b7e4647cc48e628015d3216c1533): Error starting userland proxy: listen tcp 0.0.0.0:22: bind: address already in use.
```
but well it's not a big issue

another item : why do we need to launch container as a daemon and then remove it later.
it is not working on a single command (no daemon/background) ?

TylerJewell: Ok - so I replaced the `httpd` command with your syntax and it worked on Windows, too.  And it was faster execution.  It also does not require a daemon execution, but hte other httpd approach did.  So good improvement.

benoitf (approved):
nice improvements

slemeur:
This issue has been merged without:
- milestone information
- updated docs
- updated release note

TylerJewell:
This PR had the old template so forgot about those items.  Added.

PR #3837, opened 2017-01-20
add new version 5.1.1

riuvshin:
add bugfix version


dockerfiles/cli/version/latest.ver
```diff
@@ -1 +1 @@
-5.1.0
\ No newline at end of file
+5.1.1
```
benoitf, line 1: will it fail if we have a newline there ?

benoitf (approved):

dockerfiles/cli/version/latest.ver
riuvshin: idk we never had new line here... I think it will work but not sure

benoitf: ok up to you if you're a tester or not ^^

riuvshin: I don;t want break things now, I will test this later if it works I will add new line

PR #3848, opened 2017-01-23
CLI - Fix error with version check on nightly

TylerJewell:
Signed-off-by: Tyler Jewell tjewell@codenvy.com <tjewell@codenvy.com>

With the new commit made where we check for newer versions on DockerHub, this check was being done for all versions including nightly.  The `compare_versions` method would break if you had a nightly image. This fix adds a test to only do this check if you are not nightly.

Also, this cleans up some of the code around boot2docker checks. Unfortunately, since we are running our client inside of a Docker container, all we can do is print a warning message that you are on boot2docker and that you need to verify your host mounts.

### Changelog and Release Note Information
**Changelog**: Skips DockerHub newer version check if using `nightly` image

### Release Note
- [x] Added to release note

### DIFF
--- a/dockerfiles/base/scripts/base/library.sh
+++ b/dockerfiles/base/scripts/base/library.sh
@@ -368,7 +368,7 @@ verify_version_compatibility() {
   ##      - If they don't match, then if CLLI is newer fail with message to run upgrade first
   CHE_IMAGE_VERSION=$(get_image_version)
 
-  # Only check for newer versions if not in offline mode.
+  # Only check for newer versions if not in offline mode and not nightly.
   if ! is_offline && ! is_nightly; then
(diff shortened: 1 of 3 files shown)


dockerfiles/base/scripts/base/library.sh
```diff
@@ -369,7 +369,7 @@ verify_version_compatibility() {
   CHE_IMAGE_VERSION=$(get_image_version)
 
   # Only check for newer versions if not in offline mode.
```
benoitf: we should add in the comment : "not in offline mode and not nightly"

benoitf (approved):

slemeur:
Added to RN - can be merged

PR #3866, opened 2017-01-24
CODENVY-1579: show server's address along with URL

akurinnoy:
Signed-off-by: Oleksii Kurinnyi <okurinnyi@codenvy.com>

### What does this PR do?
This PR improves Service section of machine config to show server's address along with URL.

che-workspace-running-runtime-servers

https://github.com/codenvy/codenvy/issues/1579

#### Changelog
Show server's address along with URL.

#### Release notes
- [x] Fixed displaying both server's URL and address for runtime and undefined protocol

### Docs Pull Request

https://github.com/eclipse/che-docs/pull/93

### DIFF
--- a/dashboard/src/app/workspaces/workspace-details/environments/list-servers/edit-server-dialog/edit-server-dialog.html
+++ b/dashboard/src/app/workspaces/workspace-details/environments/list-servers/edit-server-dialog/edit-server-dialog.html
@@ -56,8 +56,10 @@
                    ng-model="editServerDialogController.protocol"
                    type="type"
                    aria-label="Server protocol"
-                   ng-pattern="/^[A-Za-z0-9_\-\.]+$/">
-          <div ng-message="pattern">The reference should not contain special characters like space, dollar, etc.</div>
+                   ng-pattern="/^[A-Za-z0-9_\-\.]+$/"
+                   required>
+          <div ng-message="required">A protocol is required.</div>
+          <div ng-message="pattern">The protocol should not contain special characters like space, dollar, etc.</div>
         </che-input>
       </div>
     </che-label-container>
--- a/dashboard/src/app/workspaces/workspace-details/environments/list-servers/list-servers.controller.ts
+++ b/dashboard/src/app/workspaces/workspace-details/environments/list-servers/list-servers.controller.ts
@@ -60,7 +60,7 @@ export class ListServersController {
   buildServersList(): void {
     this.serversList = this.lodash.map(this.servers, (server: IEnvironmentManagerMachineServer, reference: string) => {
       let serverItem: IServerListItem = angular.extend({}, {reference: reference}, server);
-      serverItem.protocol = serverItem.protocol ? serverItem.protocol : 'http';
+      serverItem.protocol = serverItem.protocol ? serverItem.protocol : '-';
       return serverItem;
     });
   }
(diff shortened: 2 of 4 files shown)


ashumilova (approved):

benoitf (approved):
nice !

gazarenkov:
Do we really need two lines row (address and URL)? Note, there will be Address for all the Servers.
So, why not to make it as usual with 2 columns with clear titles?
(Column name "Runtime" sounds about nothing)

TylerJewell:
I agree.  @gazarenkov - is the URL part even needed?  I think most people can put together the protocol and address to figure it out.  It's just redundant data at that point.

gazarenkov:
Yes.
It is ready to use URL, including path (can be added to the Server definition too btw). So
- it is just convenient to not to concatenate it every time on client
- protocol, host and port may not be enough (especially if URL is modified due to proxy)
- moreover in a case of infra under reverse proxy (such as codenvy.io) it may be even impossible to to form right external URL since host and path may be not the same

olexii4 (approved):

akurinnoy:
@gazarenkov We did it because otherwise we would run into this problem:

screenshot

slemeur:
Yes so basically, we are going to have space problems. A lot of user would have to scroll horizontaly which is in general not convenient.

slemeur:
What is your proposal @gazarenkov ?

For the table's UI, we agree with your comment, with @akurinnoy we tried to display this table with one more column, but text is truncated and table looks worst than with 2 lines. 
We thought about hiding the URL and just use a button, but we thought that it would not allow simple copy/paste action.

We do not really understand your latest comment, if you can explain clearly what you'd like to be displayed, it would be helpful and maybe solve properly the problem.

benoitf:
Can we merge this PR ? it is there since a long long time

ashumilova:
+1 on merging

slemeur:
Yes fine for me.

slemeur (approved):

cyberglad:
2017_05_12_09_54_22_codenvy_hello_world
I may sound stupid, but I don't see any icon at my upper-right corner, as I have seen on your screenshots. How do I switch to Service section?

ashumilova:
@cyberglad by icon in the upper-right corner, i think, you mean machines/operation perspective switcher. Yes, we removed it. Now to see and modify servers - need to go to concrete workspace details in dashboard. 
image
Select runtime tab and expand the details of necessary machine. There you will find Servers section.
Note, that workspace should be running.

LiteSoul:
@ashumilova So you removed the Machine Perspective Switcher, so when you search on google trying to find Where to see the server ip one is running, the answers are wrong because there is no more icon.

Second, now we have to go to the Dashboard, then the workspace, then Runtime, then Dev-machine, then Server... whoa that's incredible hard. Knowing the ip is Fundamental and i'ts one of the most important tools we need.

I hope you can make it into a simple 1 click inside the ide... thanks

ashumilova:
@LiteSoul Thank you for feedbacks, we will take it into account to improve UX.

cyberglad:
Anna

I still cannot connect to  http://node22.codenvy.io:33605/ , though it should be possible with the following code:

@Override
public void configure() throws Exception {
from("jetty:http://0.0.0.0:8080")
.process(exchange -> {exchange.getIn().setBody("Hello World");});
}
is there any firewall to open?
(I don't have any firewalls in my house, in fact I have tried at varius places).
Rgds
Yuri

PR #3867, opened 2017-01-24
change AGENT_EXTERNAL_PORT to real ephemeral port

riuvshin:
@TylerJewell I've made this PR because I think this should fix issues on our nightly env where we are not able to run codenvy due to failed network checs. the problem is that in test we trying to use random port `12345` as ephemeral but ephemeral port rage is `32768:65535` and `12345` port is blocked by firewall while ephemeral range is available.

#### Changelog
Change AGENT_EXTERNAL_PORT to real ephemeral port


benoitf (approved):

TylerJewell (approved):

riuvshin:
to proof that this will work I;ve opened 12345 port on firewall and issue is gone, and as we already have ephemeral ports open on our infra it will work.

benoitf:
@riuvshin yes BTW test is closer than how it performs so it's good

PR #3870, opened 2017-01-24
Added functional tests for CLI start command

mlabuda:
Signed-off-by: Marian Labuda <mlabuda@redhat.com>

### What does this PR do?
Added basic functional tests for CLI start command. Plus few minors:
- fixed usage of CLI image in cli prompts and usage tests suite
- extended test base with function to get free port and find out whether a passed port is free 

#### Changelog
Added basic CLI 'start' functional tests

**Relase Notes and Docs PR** N/A
- [x] Added to release note in community section

### DIFF
--- a/dockerfiles/base/test.sh
+++ b/dockerfiles/base/test.sh
@@ -14,10 +14,18 @@ BATS_BASE_DIR=$(cd "$(dirname "$0")"; pwd)
 init "$@"
 IMAGE_NAME="eclipse/che-bats:$TAG"
 
+# Runs functional CLI tests in a docker container.
+# Pass a file name of functional bats tests as an argument.
+# The file has to be placed in tests folder in directory containing this script
+run_test_in_docker_container() {
+  docker run -v $BATS_BASE_DIR:$BATS_BASE_DIR -e CLI_IMAGE_TAG=$TAG -e BATS_BASE_DIR=$BATS_BASE_DIR -v /var/run/docker.sock:/var/run/docker.sock $IMAGE_NAME bats $BATS_BASE_DIR/tests/$1
+}
+
 echo "Running tests in container from image $IMAGE_NAME"
 echo "Running functional bats tests for CLI prompts and usage"
-docker run -v $BATS_BASE_DIR:$BATS_BASE_DIR -e CLI_IMAGE_TAG=$TAG -e BATS_BASE_DIR=$BATS_BASE_DIR -v /var/run/docker.sock:/var/run/docker.sock $IMAGE_NAME bats $BATS_BASE_DIR/tests/cli_prompts_usage_tests.bats
+run_test_in_docker_container cli_prompts_usage_tests.bats
 echo "Running functional bats tests for init and destroy commands"
-docker run -v $BATS_BASE_DIR:$BATS_BASE_DIR -e CLI_IMAGE_TAG=$TAG -e BATS_BASE_DIR=$BATS_BASE_DIR -v /var/run/docker.sock:/var/run/docker.sock $IMAGE_NAME bats $BATS_BASE_DIR/tests/cmd_init_destroy_tests.bats
+run_test_in_docker_container cmd_init_destroy_tests.bats
 echo "Running functionals bats tests for start command"
-docker run -v $BATS_BASE_DIR:$BATS_BASE_DIR -e CLI_IMAGE_TAG=$TAG -e BATS_BASE_DIR=$BATS_BASE_DIR -v /var/run/docker.sock:/var/run/docker.sock $IMAGE_NAME bats $BATS_BASE_DIR/tests/cmd_start_tests.bats
+run_test_in_docker_container cmd_start_tests.bats
+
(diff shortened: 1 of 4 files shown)


TylerJewell (approved):

TylerJewell:
Nice improvement.  I learned something new by reading this about how to find open ports.

benoitf (commented):
I forgot to publish my comments

dockerfiles/base/tests/cmd_start_tests.bats
```diff
@@ -0,0 +1,49 @@
...
+  #THEN
+  [[ $(docker inspect --format="{{.State.Running}}" che) -eq "true" ]]
+  ip_address=$(docker inspect -f {{.NetworkSettings.Networks.bridge.IPAddress}} che)
+  curl -fsS http://${ip_address}:8080  > /dev/null
+}
+
+@test "test cli 'start' with custom port" {
+  #GIVEN
+  tmp_path=${TESTRUN_DIR}/start2
+  free_port=$(get_free_port)
```
benoitf, line 39: do we need get_free_port method ? AFAIK in docker we're in isolated env

dockerfiles/base/test.sh
```diff
@@ -17,6 +17,7 @@ IMAGE_NAME="eclipse/che-bats:$TAG"
 echo "Running tests in container from image $IMAGE_NAME"
 echo "Running functional bats tests for CLI prompts and usage"
 docker run -v $BATS_BASE_DIR:$BATS_BASE_DIR -e CLI_IMAGE_TAG=$TAG -e BATS_BASE_DIR=$BATS_BASE_DIR -v /var/run/docker.sock:/var/run/docker.sock $IMAGE_NAME bats $BATS_BASE_DIR/tests/cli_prompts_usage_tests.bats
-echo "Running functional bats tests for init and destroy"
+echo "Running functional bats tests for init and destroy commands"
 docker run -v $BATS_BASE_DIR:$BATS_BASE_DIR -e CLI_IMAGE_TAG=$TAG -e BATS_BASE_DIR=$BATS_BASE_DIR -v /var/run/docker.sock:/var/run/docker.sock $IMAGE_NAME bats $BATS_BASE_DIR/tests/cmd_init_destroy_tests.bats
-
+echo "Running functionals bats tests for start command"
+docker run -v $BATS_BASE_DIR:$BATS_BASE_DIR -e CLI_IMAGE_TAG=$TAG -e BATS_BASE_DIR=$BATS_BASE_DIR -v /var/run/docker.sock:/var/run/docker.sock $IMAGE_NAME bats $BATS_BASE_DIR/tests/cmd_start_tests.bats
```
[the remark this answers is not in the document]
mlabuda: Sure thing, I will extract subcommand as a function.

dockerfiles/base/tests/test_base.sh
```diff
@@ -19,15 +19,39 @@ fi
...
 
 remove_named_container() {
-  if [ $(docker ps -a --format '{{.Names}}' | grep $1 | wc -l) -eq 1 ]; then
+  if [[ $(docker ps -a --format '{{.Names}}' | grep $1 | wc -l) -eq 1 ]]; then
     echo "Removing named container $1"
     docker rm $1 1>/dev/null
   fi
 }
+
+# Pass a port as an argument to check whether is free or not
```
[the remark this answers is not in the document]
mlabuda: I don't think so. The test_base.sh is parent of all tests, if I would move all utility functions to a library_tests.sh, than the test_base.sh would have only few constants in header, sourced library and nothing else (no functions). It would be a transitive sourcing and not needed, I think.

dockerfiles/base/tests/cmd_start_tests.bats
mlabuda: Yes, we need it. I was looking into Che and it is using default network interface docker0 (On my side 172.17.0.1). Che server runs in a container with an IP (here 172.17.0.4). Che server is accessible via port 8080. I can access to the server via 172.17.0.4:8080 but also via 172.17.0.1:8080. But the server also expose ports and forward ports to localhost. So the server is accessible also on localhost:8080. The port 8080 can be found in state LISTEN in netstat output. Same applies also to workspaces, but those take ports from 32xxx (at least now when I was running them, haven't look closer to scripts what is the algorithm to choose ports). And those are forwarded as well.

Ok now to the reason why I need free ports. First problem could be caused if there would be other docker images running and those would use port required by a che server. It would fail. Second problem is having a service running on localhost (e.g. I have running wildfly locally, which uses port 8080). And this would fail too.

benoitf: I still think it is strange. Because the check is performed inside the container while che for example use a docker port mapping.
So I don't see if when we check inside the container we can check the host ports if we don't have a port mapping

mlabuda: Workspace map ports to free ports, but the che-server which is using this free port function is using same port and if on host is the port used, I cannot start the che server there. Therefore I cannot run che server with CHE_PORT=alreadyusedport and tests would fail.

mlabuda: Well, I did more digging. The way to get free ports is working on a host, but inside a container it is not working, basically because container itself is isolated from other containers... I will have to find a way to get free ports from inside, at the moment it is not working as expected, unfortunately (it takes first port from the range in free ports function, because output of netstat is empty).

benoitf: yes this was my first impression. But it's not a big issue

dockerfiles/base/tests/test_base.sh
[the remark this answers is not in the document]
benoitf: ok perfect

benoitf (approved):

slemeur:
Added to Release Note - ok for me

benoitf:
While it works on my computer when I tested, it is not on the CI server: https://github.com/eclipse/che/pull/3902

PR #3875, opened 2017-01-24
Chefile handling of inner Dockerfile content

### LINKED ISSUE #3704: Chefile Dockerfile Error: Fail launching agent Terminal
I am trying to embed minimal Dockerfile content in a Chefile and I keep getting, "Error: Fail launching agent Terminal".  I've tried to strip out everything unnecessary from the eclipse/alpine_jdk8 image, which does work if I simply put "FROM eclipse/alpine_jdk8".

**Reproduction Steps:**
Create a Chefile that contains:
```
workspace.runtime.docker.content=`
FROM docker:1.12.0
RUN apk add --no-cache bash openssh sudo && \
    adduser -S user -h /home/user -s /bin/bash -G root -u 1000 && \
    echo "%root ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers && \
    echo "SECRET" >> /tmp/passwd && \
    echo "SECRET" >> /tmp/passwd && \
    cat /tmp/passwd | passwd user && \
    chown -R user /home/user/
EXPOSE 22
USER user
WORKDIR /projects
CMD sudo /usr/bin/ssh-keygen -A && \
    sudo /usr/sbin/sshd -D && \
    sudo su - && \
    tail -f /dev/null
`
```
Then, go into the directory containing the Chefile and run something like:
docker run -t --rm -v /var/run/docker.sock:/var/run/docker.sock -v $PWD/.che:/data eclipse/che dir $PWD up

If using Windows PowerShell, replace the last $PWD in the command above with something like:
/c/users/REPLACE_WITH_THE_REST_OF_YOUR_PATH

This is the error you'll see in the terminal:
```
INFO: (che dir): Workspace created
INFO: (che dir): Workspace booting...
ERROR: (che dir): Error when starting the workspace{"error":"Start of environment default failed. Error: Fail launching
agent Terminal. Workspace ID:workspace6a1uxqckh8c0fvlo","eventType":"ERROR","prevStatus":"STARTING","workspaceId":"works
pace6a1uxqckh8c0fvlo"}
```
**Che version:**
5.0.0

**OS and version:**
Windows 10 Pro via PowerShell

**Docker version:**
1.13.0-rc6

**Che cli.log output:**
Fri Jan 13 05:12:53 UTC 2017
INFO: (che cli): Checking network... (hint: '--fast' skips version and network checks)
/usr/bin/curl  "-I -k dockerhub.com -s -o /dev/null --write-out %{http_code}"
DEBUG: docker_run
DEBUG: is_boot2docker
DEBUG: is_initialized
DEBUG: cli_parse
DEBUG: update_image_if_not_found
[0;32mINFO:[0m (che download): Checking for image 'eclipse/che-dir:nightly'...
found
DEBUG: docker_run

benoitf:
### What does this PR do?

Fix #3704 by handling recipe content as a raw content and handle # for comments correctly

1. Multiline content is escaped by default while it shouldn’t for Dockerfile content.
Users may add “String.raw” but it’s not obvious so add it automatically

2. Handle # comments correctly

Fix #3704 

#### Changelog
Handle dockerfile content in Chefile as raw string

**Release Notes**:
bug only

### Docs Pull Request
no doc, bug fixing.

### DIFF
--- a/dockerfiles/lib/src/utils/string-utils.ts
+++ b/dockerfiles/lib/src/utils/string-utils.ts
@@ -9,8 +9,39 @@
  *   Codenvy, S.A. - initial API and implementation
  */
 
+/**
+ * Handle some operations on strings.
+ * @author Florent Benoit
+ */
 export class StringUtils {
+
+    /**
+     * Check if the given value string i starting with searchString
+     * @param value the original string
+     * @param searchString the value to search
+     * @returns {boolean} true if it starts with
+     */
     static startsWith(value:string, searchString:string):boolean {
         return value.substr(0, searchString.length) === searchString;
     }
+
+    /**
+     * Remove from the given string all the given comments starting by #
+     * @param value
+     * @returns {string}
+     */
+    static removeSharpComments(value : string) : string {
+        return value.replace(/^#.*/gm, '');
+    }
+
+    /**
+     * If string literals es6 def for workspace, use raw strings to keep dockerfile content
+     * @param value
+     * @returns {string}
+     */
+    static keepWorkspaceRawStrings(value : string) : string {
+        return value.replace(/^(workspace.)(.*)=\`(.*)$/gm, '$1$2=String.raw`$3');
+    }
+
 }
+
--- a/dockerfiles/base/scripts/base/commands/cmd_dir.sh
+++ b/dockerfiles/base/scripts/base/commands/cmd_dir.sh
@@ -20,7 +20,5 @@ cmd_dir() {
 
   # Not loaded as part of the init process to save on download time
   load_utilities_images_if_not_done
-  docker_run -it ${UTILITY_IMAGE_CHEDIR} "$@"
-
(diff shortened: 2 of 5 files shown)


dockerfiles/lib/src/utils/string-utils.ts
```diff
@@ -9,8 +9,38 @@
...
+    /**
+     * If string literals es6 def for workspace, use raw strings to keep dockerfile content
+     * @param value
+     * @returns {string}
+     */
+    static keepWorkspaceRawStrings(value : string) : string {
+        return value.replace(/^(workspace.)(.*)=\`(.*)$/gm, '$1$2=String.raw`$3');
+    }
+
 }
```
TylerJewell: Newline

TylerJewell (approved):
Approved - but there is a newline to fix.

dockerfiles/lib/src/utils/string-utils.ts
benoitf: fixed

benoitf:
@TylerJewell fixed

PR #3877, opened 2017-01-24
CODENVY-1443: fix incorrect usage of non bind-mount volumes

garagatyi:
### What does this PR do?
Fix incorrect usage of non bind-mount volumes of docker.
Add tests.
Add other minor fixes.

Discovered in development of codenvy/codenvy#1443


skabashnyuk (approved):

TylerJewell:
Nice improvement, @garagatyi - good work.

benoitf:
AFAIK "Add other minor fixes." in description of the PR is not telling what are the fixes

It could be applied to any bug fix
then we don't know what is really fixing the issue from what is "improvements"

garagatyi:
Here minor improvements are related to internals, so no significant influence on Che users is supposed.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -524,6 +523,16 @@ private String createContainer(String workspaceId,
                               .map(entry -> entry.getKey() + "=" + entry.getValue())
                               .toArray(String[]::new));
 
+        // If volume contains colon then it is bind volume, otherwise - pure volume.
+        hostConfig.setBinds(service.getVolumes()
+                                   .stream()
+                                   .filter(v -> v.contains(":"))
+                                   .toArray(String[]::new));
+        config.setVolumes(service.getVolumes()
```
voievodin: Don't you think that it makes more sense to go for each volume and make `contains(":")` check once for each item?

voievodin (approved):
The code is fine. I am not into the field so up to guys to decide whether the solution is ok.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
garagatyi: ok, I'll do that.

PR #3897, opened 2017-01-25
Remove not needed classes. Web.xml cleanup

vparfonov:
1. Removed classes:

- org.eclipse.che.filters.EnvironmentInitializationFilter
- org.eclipse.che.api.local.CheGuiceEverrestServlet
- org.eclipse.che.ProjectConfigAdapter

2. web.xml fixing 

#### Changelog
Remove not needed classes. Web.xml cleanup


skabashnyuk (approved):

assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterServletModule.java
```diff
@@ -48,7 +49,7 @@ protected void configureServlets() {
         filter("/*").through(CorsFilter.class, corsFilterParams);
 
         filter("/api/*").through(org.eclipse.che.api.local.filters.EnvironmentInitializationFilter.class);
-        serveRegex("^/api((?!(/(ws|eventbus)($|/.*)))/.*)").with(org.eclipse.che.api.local.CheGuiceEverrestServlet.class);
+        serveRegex("/api((?!(/(ws|eventbus)($|/.*)))/.*)").with(GuiceEverrestServlet.class);
```
mshaposhnik: `^` is still needed here IMHO

mshaposhnik:
So after the merge, wsmaster-local module will contain only one class - DummyTokenValidator. Maybe we can move it somewhere and remove module at all ? @skabashnyuk wdyt

assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterServletModule.java
vparfonov: yes, you are right. thx

Issue #3797, opened 2017-01-18
Creating or import project in the running workspace on the dashboard does not work (regression)


artaleks9:
When I try to create or import a project in the running workspace on the dashboard, then the project is not created.

**Reproduction Steps:**
* Create workspace and run it.
* Go to 'Dashboard' -> 'Create Workspace' and create new project from template in the running existing workspace.

**Expected behavior:**
* the project should be created and opened in the IDE.

**Observed behavior:**
* the project is not created.

**Che version:** 5.1.0-SNAPSHOT
**OS and version:** Ubuntu 14.04
**Docker version:** 1.12

**Che cli.log output:** [cli.log.zip](https://github.com/eclipse/che/files/714221/cli.log.zip)

**Additional information:**
* Problem can be reliably reproduced, doesn't happen randomly: [Yes]
* See the attachments:
![create_project_running_ws](https://cloud.githubusercontent.com/assets/5408906/22069590/9fcf3788-dda1-11e6-9052-5d07a13406ff.gif)
![import_project_running_ws](https://cloud.githubusercontent.com/assets/5408906/22069601/a92d9888-dda1-11e6-92c4-93cd99f0aaff.gif)

riuvshin:
@ashumilova  please take a look on this seems this pretty critical

PR #3898, opened 2017-01-25
CHE-3797 fix project import in case when workspace is running

olexii4:
Signed-off-by: Oleksii Orel <oorel@codenvy.com>
### What does this PR do?
Fix project import in case when workspace is running.

https://github.com/eclipse/che/issues/3797

#### Changelog
Fix bug with infinite expectation during project import in case with running workspace.

**Release notes**: 
- [x] Fix bug with infinite expectation during project import in case with running workspace. 
- [x] Add possibility to catch websocket url error for target workspace before project import.

### DIFF
--- a/dashboard/src/components/api/che-workspace.factory.ts
+++ b/dashboard/src/components/api/che-workspace.factory.ts
@@ -45,7 +45,7 @@ export class CheWorkspace {
   workspaces: Array<che.IWorkspace>;
   subscribedWorkspacesIds: Array<string>;
   workspaceAgents: Map<string, CheWorkspaceAgent>;
-  workspacesByNamespace: Map<string, any>;
+  workspacesByNamespace: Map<string, Array<che.IWorkspace>>;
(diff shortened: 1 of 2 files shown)


benoitf (approved):

akurinnoy (approved):

dashboard/src/components/api/che-workspace.factory.ts
```diff
@@ -177,7 +177,7 @@ export class CheWorkspace {
     });
   }
 
-  getWorkspacesByNamespace(namespace) {
+  getWorkspacesByNamespace(namespace: string): che.IWorkspace {
```
ashumilova: Here the list is returned.

ashumilova (approved):

dashboard/src/components/api/che-workspace.factory.ts
olexii4: Done

Issue #3896, opened 2017-01-25
Package is not displayed in the 'Select Path' form


artaleks9:
I have a plain java project and I create command for java project in the 'Commands' form. But when I try to select path, then package is not displayed in the 'Select Path' widget. Also the name of project is duplicated.

**Reproduction Steps:**
* Import and configure plain java project (e.g. https://github.com/idexmai1/plainJavaProject.git)
* Open the 'Commands' widget, create new command for java plain project and press 'Browse' button.
* Expand the project into 'Select Path' form.

**Expected behavior:**
* package should be displayed in the path.

**Observed behavior:**
* folders are displayed in the path instead package 
* also the project name is duplicated.

**Che version:** 5.2.0-SNAPSHOT
**OS and version:** Ubuntu 14.04
**Docker version:** 1.12

**Che cli.log output:**  [cli.log.zip](https://github.com/eclipse/che/files/729635/cli.log.zip)

**Additional information:**
* Problem can be reliably reproduced, doesn't happen randomly: [Yes]
* See the attachments:
![plain-java](https://cloud.githubusercontent.com/assets/5408906/22290658/d5a6b444-e30a-11e6-823a-1d88e8009531.png)
![plain_project](https://cloud.githubusercontent.com/assets/5408906/22290662/dad17daa-e30a-11e6-814a-ad66040bfbb1.png)

artaleks9:
The bug is fixed and not reproduced.

PR #3909, opened 2017-01-26
Avoid duplicate projects in application context

vzhukovs:
This PR prevent duplicating project in application context. Duplicating may be occurred when project imports via UI, in that case IDE knows that a new project is appeared and additionally receives a notification from the server that project has been created. Besides there was a little code cleanup performed in _Select main class_ dialog. Related issue: #3896


vzhukovs:
Closed due to no reproducing issue anymore

PR #3917, opened 2017-01-26
Add system service that allows to prepare system to shutdown

### LINKED ISSUE #3892: Create SystemService
We want to provide some REST API for CLI to make some activity before system shutdown.

 - ```POST /system/stop```
 Will iterate over all running workspaces and suspend them, also it will prevent new workspace starting.
 - ```GET  /system/status```
 Will return json with system ```state:running|preparetoshutdown|readytoshutdown ```

Moved from https://github.com/codenvy/codenvy/issues/1568#issuecomment-275040020

voievodin:
Resolves #3892.

Two API methods are available:

* _POST /api/system/stop_ - stops corresponding system services, for now it's workspace service.
Basically it stops all the RUNNING workspaces snapshotting them before if configured to do so,
also it interrupts start of currently STARTING workspaces.

* _GET /api/system/state_ - returns current system state(only status for now).
By default system status is RUNNING, but after stop is called or che server is
stopped directly system status is changed like the following:
```
RUNNING -> PREPARING_TO_SHUTDOWN -> READY_TO_SHUTDOWN
```
Status changes are followed by status changed events, to recieve these
events client has to subscribe on *system:status* channel.

So client can use different ways to track system status changes.
Doesn't matter whether API method is used or che server is stopped directly,
events will be sent anyway.

#### Changelog
Add system service that allows to prepare system to shutdown


benoitf:
For user feedback as part of CLI I would be interested in having a countdown.

IMHO user may think the cli is hanging if we don't display some interesting outputs.
it could be the current action of the graceful-stop item or size of remaining actions to perform before the graceful-stop is done, etc

voievodin:
Well i think it's not a problem for system service to provide you additional information about stopping process.
What if i reach system events set with such events as:

```json
{
    "type" : "STOPPING_SERVICE", 
    "service" : "workspace"
}
```
```json
{
    "type" : "SERVICE_ITEM_STOPPED", 
    "id" : "workspace0x123456",
    "service" : "workspace"
}  
```
```json
{
    "type" : "SERVICE_STOPPED", 
    "service" : "workspace"
}  
```

So you can display something like the following: 
```
Stopping service: workspace
   'workspace123' is stopped
   'workspace234' is stopped
   'workspace345' is stopped
   service stopped

Stopping service: ldap
...
```
@skabashnyuk, @benoitf what do you think?

benoitf:
@evoevodin if we can have the current/total number as well 
```
'workspace123' is stopped (12/36)
```

```
{
    "type" : "SERVICE_ITEM_STOPPED", 
    "id" : "workspace0x123456",
    "service" : "workspace",
    "current" : "12",
    "total": "36"
}  
```

voievodin:
@benoitf as far as i understand from the [comment](https://github.com/eclipse/che/issues/3892#issuecomment-275434966) we should go forward with current implementation as it solves the problem.

Can you please create an issue and describe the kind of events CLI needs - to show the progress, so we can move topic discussion there?

benoitf (approved):

benoitf:
@evoevodin https://github.com/eclipse/che/issues/3921

wsmaster/che-core-api-system/src/main/java/org/eclipse/che/api/system/server/SystemService.java
```diff
@@ -0,0 +1,58 @@
...
+    private final SystemManager manager;
+
+    @Inject
+    public SystemService(SystemManager manager) {
+        this.manager = manager;
+    }
+
+    @POST
+    @Path("/stop")
+    @ApiResponses({@ApiResponse(code = 204, message = "The system is preparing to stop"),
```
sleshchenko, line 54: Please add @ApiOperation

wsmaster/che-core-api-system/pom.xml
```diff
@@ -0,0 +1,192 @@
...
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-commons-lang</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.everrest</groupId>
+            <artifactId>everrest-websockets</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
```
sleshchenko, line 79: Please add [logback-classic](https://github.com/eclipse/che/blob/25f18e1ce2b207f3aef9e6cabc605cd26bc42cfa/wsmaster/che-core-api-workspace/pom.xml#L147-L151) test dependency and [logback-test](https://github.com/eclipse/che/blob/09353bc7a41377a06c9dc6a4c4f042d3c1883ffc/agents/che-core-api-agent/src/test/resources/logback-test.xml) to test resources

wsmaster/che-core-api-system/src/test/java/org/eclipse/che/api/system/server/SystemManagerTest.java
```diff
@@ -0,0 +1,85 @@
...
+    private EventService eventService;
+
+    @Captor
+    private ArgumentCaptor<SystemStatusEvent> eventsCaptor;
+
+    private SystemManager systemManager;
+
+    @BeforeMethod
+    public void init() {
+        MockitoAnnotations.initMocks(this);
```
sleshchenko, line 57: Why do you need this? MockitoTestNGListener should init mocks, no?

voievodin: If i don't do it explicitly i start getting errors like this:
```
org.mockito.exceptions.verification.TooManyActualInvocations: 
runtimes.stop(<any>);
Wanted 1 time:
-> at org.eclipse.che.api.workspace.server.WorkspaceManagerTest.shouldStopWorkspaceEventIfSnapshotCreationFailed(WorkspaceManagerTest.java:509)
But was 11 times. Undesired invocation:
-> at org.eclipse.che.api.workspace.server.WorkspaceManager.lambda$stopAsync$3(WorkspaceManager.java:813)
```

Which probably means that mockitong doesn't do it's job well, do i miss something ?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -191,22 +190,31 @@ public WorkspaceRuntimes(EventService eventsService,
...
         EnvironmentImpl environment = copyEnv(workspace, envName);
         String workspaceId = workspace.getId();
         CompletableFuture<WorkspaceRuntimeImpl> cmpFuture;
         StartTask startTask;
         try (@SuppressWarnings("unused") Unlocker u = locks.writeLock(workspaceId)) {
-            ensurePreDestroyIsNotExecuted();
+            checkIsNotTerminated("start the workspace");
+            if (isStartRefused.get()) {
+                throw new ConflictException(format("Start of the workspace '%s' is rejected by the system, " +
+                                                   "no more workspaces are allow to start",
```
garagatyi: Looks like typo `allow -> allowed`

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -575,67 +583,85 @@ public Instance getMachine(String workspaceId, String machineId) throws NotFound
...
-        sharedPool.terminateAndWait();
+    /**
+     * Once called no more workspaces are allowed to start, {@link #startAsync}
+     * will always throw an appropriate exception. All the running workspaces
+     * will continue running, unless stopped directly.
+     *
+     * @return true if this is the caller is the one who refused start,
+     * otherwise if start is being already refused returns false
+     */
+    public boolean refuseWorkspacesStart() {
```
garagatyi, line 601: Can you elaborate on why we need both refuse and shutdown?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -812,9 +838,8 @@ private boolean compareAndSetStatus(String id, WorkspaceStatus from, WorkspaceSt
     }
 
     /** Removes state from in-memory storage in write lock. */
-    private void removeState(String workspaceId) {
+    private void removeState(String workspaceId) throws ServerException {
```
garagatyi: Can you point me why it throws ServerException, looks like I'm missed something.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -656,6 +665,70 @@ public Instance getMachineInstance(String workspaceId,
...
+     * Shuts down workspace service and waits for it to finish, so currently
+     * starting and running workspaces are stopped and it becomes unavailable to start new workspaces.
+     *
+     * @throws InterruptedException
+     *         if it's interrupted while waiting for running workspaces to stop
+     * @throws IllegalStateException
+     *         if component shutdown is already called
+     */
+    public void shutdown() throws InterruptedException {
+        if (!runtimes.refuseWorkspacesStart()) {
```
garagatyi, line 678: Please cover by tests
WorkspaceManager#shutdown
WorkspaceRuntimes#isAnyRunning
WorkspaceRuntimes#213
SystemManager#shutdown

garagatyi (commented):
Looks good.
Can you take a look on my comments?

benoitf:
question: DTOs shouldn't be in a maven separated module "che-core-api-system-shared" ?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
voievodin: Workspace manager knows about snapshot creation strategy, so it MUST stop running workspaces to snapshot them before(if it's configured to do so). As workspace runtimes state is managed by this component we need middle state that allows us to handle running workspace and rejects starts of new ones, but stop/snapshot through the API is still allowed. On the other hand component shutdown cleans in memory storage and doesn't allow to make any modification operations in locks, so in this way we guarantee that threads that still working with some in memory data won't continue working and will stop immediately as they meet `isShutdown = true` in the lock. 
So starting thread that wants to change its status from STARTING -> RUNNING is either interrupted by workspace manager or failed because of service shutdown, or everything is okay and workspace status changed, but there is no way to get null workspace or workspace in inconsistent state while it has to be STARTING.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
voievodin: termination check throws this exception

voievodin:
They should, i will move them to separate module thx

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
garagatyi: ok, thanks

benoitf:
@evoevodin 
could we as part of the `/state`json output
get links[]

with something like for workspaces:
```json
{
"rel": "get system events channel",
"href": "ws://localhost:8080/wsmaster/api/ws",
"method": "GET",
"parameters": [
{
"required": true,
"valid": [],
"name": "channel",
"defaultValue": "system:state"
}
]
```

because from client perspective, we don't know the name of channel
we have to go in source code and find
https://github.com/eclipse/che/blob/fb98dc453ef1191229226bb5c54db7a102967773/wsmaster/che-core-api-system/src/main/java/org/eclipse/che/api/system/server/SystemEventsWebsocketBroadcaster.java#L44

wsmaster/che-core-api-system/src/main/java/org/eclipse/che/api/system/server/SystemManager.java
```diff
@@ -0,0 +1,115 @@
...
+     * @see SystemStatus
+     */
+    public SystemStatus getSystemStatus() {
+        return statusRef.get();
+    }
+
+    /** Synchronously stops corresponding services. */
+    private void doStopServices() {
+        LOG.info("Preparing system to shutdown");
+        eventService.publish(new SystemStatusChangedEvent(RUNNING, PREPARING_TO_SHUTDOWN));
```
benoitf, line 94: I think there is an error on status/prevStatus
```java
 SystemStatusChangedEvent(SystemStatus status, SystemStatus prevStatus) {
```
so it means prev status was PREPARING_TO_SHUTDOWN and new status is RUNNING while IMHO it's the opposite

wsmaster/che-core-api-system/src/main/java/org/eclipse/che/api/system/server/SystemManager.java
```diff
@@ -0,0 +1,115 @@
...
+        try {
+            workspaceManager.shutdown();
+            statusRef.set(READY_TO_SHUTDOWN);
+            eventService.publish(new SystemStatusChangedEvent(PREPARING_TO_SHUTDOWN, READY_TO_SHUTDOWN));
```
benoitf, line 98: reverse order of status ?

wsmaster/che-core-api-system/src/main/java/org/eclipse/che/api/system/server/SystemManager.java
voievodin: thx

voievodin:
@benoitf 
added links to the response:
```json
{
   "status":"RUNNING",
   "links":[
      {
         "href":"ws://localhost:8080/wsmaster/api/ws",
         "parameters":[
            {
               "required":true,
               "name":"channel",
               "valid":[],
               "defaultValue":"system:state"
            }
         ],
         "rel":"system.state.channel",
         "method":"GET"
      }
   ]
}
```

benoitf:
@evoevodin thanks

skabashnyuk (approved):

garagatyi (approved):

PR #3924, opened 2017-01-27
Package assembly-ide-war as classes artifact

skabashnyuk:
It's very hard for plugin developers to extend assembly-ide-war.
To do that they need to copy all dependencies declared assembly-ide-war
because assembly-ide-war as war dependency can be reused in dependency section of pom.
If we package it as classes it can be reused to inject transitive dependencies
Examples how to reuse https://github.com/skabashnyuk/sample/blob/master/assembly/assembly-ide-war/pom.xml#L31-L35

#### Changelog
Package assembly-ide-war as classes artifact

### DIFF
--- a/assembly/assembly-ide-war/pom.xml
+++ b/assembly/assembly-ide-war/pom.xml
@@ -458,6 +458,7 @@
                 <artifactId>maven-war-plugin</artifactId>
                 <configuration>
                     <packagingExcludes>%regex[WEB-INF\/lib\/(?!.*j2ee).*.jar]</packagingExcludes>
+                    <attachClasses>true</attachClasses>
                 </configuration>
             </plugin>
             <plugin>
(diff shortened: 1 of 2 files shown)


benoitf (approved):
@skabashnyuk is that it's working as well with war overlays ?
https://maven.apache.org/plugins/maven-war-plugin/overlays.html

BTW this improvement is fine for me as well

skabashnyuk:
>is that it's working as well with war overlays ?
>https://maven.apache.org/plugins/maven-war-plugin/overlays.html

Don't know. I don't see how they are related.

riuvshin (approved):

benoitf:
@skabashnyuk well overlays allows to extend an existing war file

skabashnyuk:
>well overlays allows to extends an existing war file

well - yes. But this pr oriented on ability to inherit jar  dependencies  of war so we can reuse it in GWT compilation.

PR #3926, opened 2017-01-27
[cli] fix issue around logs of che-server not written in host folder and update path of machines logs

### LINKED ISSUE #3913: No che-server logs in logs folder
New Che CLI creates logs folder in instance folder but doesn't write there logs of che-server.
Also there is folder <instance-folder>/data/logs where logs of machines are stored.
For example Codenvy CLI stores logs of both in <instance-folder>/logs/codenvy.

**Che version:**    5.2.0-SNAPSHOT

**OS and version:**    Ubuntu 16.04

**Docker version:** 1.12.5

benoitf:
### What does this PR do?
Fix #3913 : che-server container should reuse property of CHE_LOGS_DIR if externally defined (else we couldn't configure the path externally)
Also machine logs should go in /logs/ folder as well (not in /data)

#3913 

#### Changelog
Che logs should go in /logs folder. server log was not present.

**Release Notes**: N/A (bug)

### Docs Pull Request
N/A as bugfix
https://www.eclipse.org/che/docs/setup/getting-started/index.html is not talking about logs


dockerfiles/init/modules/che/templates/che.env.erb
```diff
@@ -23,7 +22,8 @@ JPDA_ADDRESS=<%= scope.lookupvar('che::che_debug_port') %>
 
 CHE_DATABASE=/data
 CHE_WORKSPACE_STORAGE=/data/workspaces
-CHE_WORKSPACE_LOGS=/data/logs/machine
+CHE_LOGS_DIR=/logs
```
garagatyi, line 25: If it is hardcoded here then it is not possible to reassign it in user's che.env. Am I correct?

garagatyi (approved):

riuvshin (approved):

dockerfiles/init/modules/che/templates/che.env.erb
benoitf: @garagatyi I've moved the definition few lines after its initial location so the LOGS_DIR are at the same place, I've not introduced a new hardcoded property.

AFAIK the definition of these values are "hardcoded" by CLI as it's inside the docker container. (for Che and Codenvy)

benoitf: @garagatyi you may want to be able to override these values ?

garagatyi: No, just thinking about changed behavior [here](https://github.com/eclipse/che/pull/3926/files#diff-1da4013f60e1b92b5613cf9680423a88R63). But now I see that this line sets [that one](https://github.com/eclipse/che/pull/3926/files#diff-1da4013f60e1b92b5613cf9680423a88R63).

slemeur:
Added to RN.
Thanks

slemeur (approved):

dockerfiles/init/modules/che/templates/che.env.erb
benoitf: @garagatyi yes correct, this is the fix. Before it was hardcoded in che-server docker image. Now it is set only if not provided. And cli is providing the property so it handle it correctly.

garagatyi: ok

TylerJewell (approved):

PR #3935, opened 2017-01-27
improve addon to have own modules

riuvshin:
needed for https://github.com/codenvy/customer-saas/issues/126

those changes allows addon.pp have module which allow not only set variables but create new files if needed

### DIFF
--- a/dockerfiles/base/scripts/base/commands/cmd_config.sh
+++ b/dockerfiles/base/scripts/base/commands/cmd_config.sh
@@ -116,6 +116,7 @@ generate_configuration_with_puppet() {
     # Handle override/addon
     if [ -d "/repo/dockerfiles/init/addon" ]; then
       WRITE_PARAMETERS+=" -v \"${CHE_HOST_DEVELOPMENT_REPO}/dockerfiles/init/addon/addon.pp\":/etc/puppet/manifests/addon.pp:ro"
+      WRITE_PARAMETERS+=" -v \"${CHE_HOST_DEVELOPMENT_REPO}/dockerfiles/init/addon/modules/\":/etc/puppet/modules/:ro"
     fi
   fi
 
@@ -133,7 +134,7 @@ generate_configuration_with_puppet() {
                   --entrypoint=/usr/bin/puppet \
                       $IMAGE_INIT \
                           apply --modulepath \
-                                /etc/puppet/modules/ \
+                                /etc/puppet/modules/:/etc/puppet/addon/ \
                                 /etc/puppet/manifests/ --show_diff ${WRITE_LOGS}"
 
   log ${GENERATE_CONFIG_COMMAND}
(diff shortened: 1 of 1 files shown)


garagatyi:
Please, provide explanation what this PR does. Since I'm not expert in CLI and puppet it is not clear for me.

dockerfiles/base/scripts/base/commands/cmd_config.sh
```diff
@@ -116,6 +116,7 @@ generate_configuration_with_puppet() {
     # Handle override/addon
     if [ -d "/repo/dockerfiles/init/addon" ]; then
       WRITE_PARAMETERS+=" -v \"${CHE_HOST_DEVELOPMENT_REPO}/dockerfiles/init/addon/addon.pp\":/etc/puppet/manifests/addon.pp:ro"
+      WRITE_PARAMETERS+=" -v \"${CHE_HOST_DEVELOPMENT_REPO}/dockerfiles/init/addon/modules/\":/etc/puppet/addon/:ro"
```
benoitf: we should add a check if /repo/dockerfiles/init/addon/modules exists before adding the mount (so it's optional in repository that don't need modules)


also is it working if we add module directly in /etc/puppet/modules ? like /etc/puppet/modules/addon

so we don't need the change /etc/puppet/modules/:/etc/puppet/addon/

garagatyi: Does it work on systems with selinux enabled? Doesn't ro,Z needed instead?

riuvshin: 1) will add check.
2) no because `/etc/puppet/modules` is mounted in dev mode already so another mount in that location will remove previous mount content. I actually did as you saying first time but it failed then I realized that i need to add it separately and include in `modulepath`

benoitf: @riuvshin : about 2. : it would be in a subfolder : /etc/puppet/modules/addon (not as /etc/puppet/modules)

benoitf:
```
WRITE_PARAMETERS+=" -v \"${CHE_HOST_DEVELOPMENT_REPO}/dockerfiles/init/addon/modules/\":/etc/puppet/modules/addon/:ro"
```

riuvshin: I don't know I've did same way as it done for another mounts.

riuvshin:
@benoitf I've added checks

dockerfiles/base/scripts/base/commands/cmd_config.sh
riuvshin: no it can't be sub folder, puppet will not be able to find it. in module path puppet expects scructure where `/etc/puppet/modules/<module_name>/{manifests},{modules}` but we will have `/etc/puppet/modules/addon/<module_name>/{manifests},{modules}` I've tried this.

benoitf (approved):

garagatyi (approved):

PR #3940, opened 2017-01-28
CLI: Add multi container launch, memory preflight check, disk preflight check

TylerJewell:

### What does this PR do?
Fixes some errors in the Che launcher and templates that prevent the CLI from launching multiple instances of Che at the same time. The containers that were launched were all given the same container name `che` and they now can be overridden on the command line with `-e CHE_CONTAINER=<name>`. This value is then passed into the Docker compose generated configuration.

#### Changelog
Add `CHE_CONTAINER=<name>` option to set explicit container name with CLI
Add `CHE_CONTAINER_PREFIX=<name>` to let Che auto-set a container name with a prefix
Refactors CLI `usage()` method to inherit from Che base
Add memory and disk preflight check
Add CLI `--skip:preflight`, `--skip:postflight`, `--skip:nightly`, `--skip:network`, `--skip:pull` global options
Refactor CLI  `cmd_start` to remove duplication and add postflight check

**Release Notes**:
You can now launch multiple Che containers on the same host using the CLI. Previously, while we let you have port assignment, all Che containers were assigned the same container name. Any time you change the port with `CHE_PORT` we will set the Che container name to be `<prefix>-<port>`. You can change the prefix by adding `-e CHE_CONTAINER_PREFIX=<name>` with the CLI. If you are using the default port `8080` then we do not append the port number to the container name.  You can optionally override the entire name with a UUID by setting `CHE_CONTAINER=<name>` on the command line.

### Docs Pull Request
https://github.com/eclipse/che-docs/pull/112


dockerfiles/base/scripts/base/startup_funcs.sh
```diff
@@ -71,8 +71,8 @@ init_constants() {
   CHE_COMPOSE_FILE="docker-compose-container.yml"
   CHE_HOST_COMPOSE_FILE="docker-compose.yml"
 
-  DEFAULT_CHE_SERVER_CONTAINER_NAME="${CHE_MINI_PRODUCT_NAME}"
-  CHE_SERVER_CONTAINER_NAME="${CHE_SERVER_CONTAINER_NAME:-${DEFAULT_CHE_SERVER_CONTAINER_NAME}}"
```
benoitf: AFAIK this is a breaking change for existing clients
CHE_SERVER_CONTAINER_NAME --> CHE_CONTAINER_NAME

TylerJewell:
Ok - I can fix this.

riuvshin:
@TylerJewell why we need this? I mean what is the value of running several CHE instances in the same time? and what will happen in codenvy usecase?
I see you've added container name but what about other resources like exposed ports ? 
can you please give any examples when this changes in not avoidable?

TylerJewell:
Well, in the past week two separate requests for this have arrived.  So it seems that it is essential if you want to use this for launching a Che farm on a single vm.

This would not have any affect for codenvy as we do not allow port alterations in codenvy and that is required to launch multiple instances at the same time.

In this case an admin will have to set both this value and the value of port to be different values each time to get unique operation.

One idea I have is to not let users set this value but instead set it to the value of the port.

@riuvshin - another use case is chedir - you can use this utility on different directories to launch different che instances, so it is required for that.

TylerJewell:
I've added in a fix for the backwards compatibility. I also now make using `CHE_CONTAINER` an optional property by using it only as a prefix for the container name.  If the port is `8080` then we use the prefix as the container name. However, if you set any different port, then we set the container name to `<prefixname>-<CHE_PORT>`.

TylerJewell:
@benoitf - refactored the code so that we can now support both a UUID and a prefix approach.

dockerfiles/base/scripts/base/startup_funcs.sh
```diff
@@ -403,12 +527,17 @@ start() {
...
+  set -- "${@/\-\-skip\:network/}"
+  set -- "${@/\-\-skip\:pull/}"
   
-    # The pre_init method is unique to each assembly. This method must be provided by 
+  # The cli_post_init method is unique to each assembly. This method must be provided by 
   # a custom CLI assembly in their container and can set global variables which are 
   # specific to that implementation of the CLI. This method must be called after
   # networking has been established and initial images downloaded.
-  cli_pre_init
+  cli_post_init
```
benoitf: I think here it's odd to have renamed cli_pre_init to cli_post_init

previous flow was
 ```
 cli_pre_init
 cli_init "$@"
```

so cli_pre_init is called before cli_init

with this change the flow is

```
cli_pre_init
init "$@"
cli_post_init
cli_init
```

so we see cli_post_init being called before cli_init which is not IMHO expected

- pre_init and post_init should be before/after init method
- cli_pre_init and cli_post_init should be before/after cli_init

Mattlk13 (approved):

Mattlk13 (approved):

Mattlk13 (approved):

Mattlk13 (approved):

dockerfiles/base/scripts/base/startup_funcs.sh
TylerJewell: Yeah, that is the expected flow for me.  We have two types of init - `init` and `cli_init`.  And in this situation the `pre` and `post` reflect the actions before and after the bootstrap phase, which is the `init` but not the `cli_init`.

I think what we call `cli_init` is just another phase of the init process.  So instead of it being just a single lump sum activity, we break the initialization into five phases.  Maybe we should rename all of the phases to be explicit names instead of pre / post.

benoitf: the name of the function is not correct

per existing patterns, we shouldn't call a function named `cli_post_init` before actual function `cli_init` is not called

only a `pre_init` and `post_init` functions could be around `init` call

TylerJewell: I think we would be better suited to renaming the hooks to be `post_init` and `pre_init`.

benoitf: yes my issue is only with the name, not that we shouldn't add new hooks

riuvshin (approved):

benoitf (approved):

benoitf:
I've updated method name for pre/post

PR #3956, opened 2017-01-31
Implement OpenShift pull, tag, and inspectImage

amisevsk:
Implement OpenShiftConnector `pull`, `tag`, and `inspectImage` through
ImageStreams. Makes changes to how containers are created.

Currently snapshots do not work, as `commit` and `push` methods are not
implemented (these are being worked on by @ibuziuk currently). Additionally, 
pulling a stack from a private repository is not yet implemented.

The changes necessitate reworking `OpenShiftConnector#createContainer()`
to accomodate using ImageStreams instead of the underlying Docker implementation.

### What does this PR do?
Adds implementations of `pull`, `tag` and `inspectImage` to OpenShiftConnector.

Additionally, this PR refactors the helper methods used in OpenShiftConnector to be
static, as they store no state and serve as utility methods. This makes OpenShiftConnector
somewhat cleaner.

#### Changelog
Implement OpenShiftConnector `pull`, `tag`, and `inspectImage` through
ImageStreams. Makes changes to how containers are created.

### Docs Pull Request
https://github.com/eclipse/che-docs/issues/76

### DIFF
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
@@ -8,6 +8,7 @@
  * Contributors:
  *   Red Hat, Inc. - initial API and implementation
  *******************************************************************************/
+
 package org.eclipse.che.plugin.openshift.client;
 
 import java.io.IOException;
@@ -28,16 +29,18 @@ import javax.inject.Named;
 import javax.inject.Singleton;
 
 import org.apache.commons.lang.RandomStringUtils;
-import org.apache.commons.lang.StringUtils;
 import org.eclipse.che.plugin.docker.client.DockerApiVersionPathPrefixProvider;
 import org.eclipse.che.plugin.docker.client.DockerConnector;
 import org.eclipse.che.plugin.docker.client.DockerConnectorConfiguration;
 import org.eclipse.che.plugin.docker.client.DockerRegistryAuthResolver;
+import org.eclipse.che.plugin.docker.client.ProgressMonitor;
 import org.eclipse.che.plugin.docker.client.connection.DockerConnectionFactory;
+import org.eclipse.che.plugin.docker.client.exception.ImageNotFoundException;
 import org.eclipse.che.plugin.docker.client.json.ContainerCreated;
 import org.eclipse.che.plugin.docker.client.json.ContainerInfo;
 import org.eclipse.che.plugin.docker.client.json.ContainerListEntry;
 import org.eclipse.che.plugin.docker.client.json.ImageConfig;
+import org.eclipse.che.plugin.docker.client.json.ImageInfo;
 import org.eclipse.che.plugin.docker.client.json.NetworkCreated;
 import org.eclipse.che.plugin.docker.client.json.PortBinding;
 import org.eclipse.che.plugin.docker.client.json.network.ContainerInNetwork;
@@ -52,6 +55,9 @@ import org.eclipse.che.plugin.docker.client.params.RemoveContainerParams;
 import org.eclipse.che.plugin.docker.client.params.RemoveNetworkParams;
 import org.eclipse.che.plugin.docker.client.params.StartContainerParams;
 import org.eclipse.che.plugin.docker.client.params.StopContainerParams;
+import org.eclipse.che.plugin.docker.client.params.InspectImageParams;
+import org.eclipse.che.plugin.docker.client.params.PullParams;
+import org.eclipse.che.plugin.docker.client.params.TagParams;
 import org.eclipse.che.plugin.docker.client.params.network.ConnectContainerToNetworkParams;
 import org.eclipse.che.plugin.docker.client.params.network.CreateNetworkParams;
 import org.eclipse.che.plugin.docker.client.params.network.DisconnectContainerFromNetworkParams;
@@ -61,6 +67,7 @@ import org.eclipse.che.plugin.openshift.client.kubernetes.KubernetesContainer;
 import org.eclipse.che.plugin.openshift.client.kubernetes.KubernetesEnvVar;
 import org.eclipse.che.plugin.openshift.client.kubernetes.KubernetesLabelConverter;
 import org.eclipse.che.plugin.openshift.client.kubernetes.KubernetesService;
+import org.eclipse.che.plugin.openshift.client.kubernetes.KubernetesStringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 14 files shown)


amisevsk:
@ibuziuk @l0rd Please take a look when you have time.

benoitf:
ci-build

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -593,34 +799,43 @@ private String getVolumeName(String volume) {
...
-                            .endMetadata()
-                            .done();
-                    return normalizedID;
-                }
+                                .withLabel(OPENSHIFT_DEPLOYMENT_LABEL, deploymentName)
+                                .list()
+                                .getItems();
+
+            if (pods.size() < 1) {
+                throw new DockerException(String.format("Pod with deployment name %s not found",
```
ibuziuk: @amisevsk might be more relevant to throw OpenShiftException now ?

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/kubernetes/KubernetesEnvVar.java
```diff
@@ -23,17 +23,17 @@
 /**
  * Provides API for managing Kubernetes {@link EnvVar}
  */
-public class KubernetesEnvVar {
+public final class KubernetesEnvVar {
```
ibuziuk, line 26: @amisevsk for all final util classes I would also add private constructor

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
amisevsk: One reason I didn't do this is that OpenShiftException is unchecked, and so it wouldn't be obvious that it could be thrown when using the methods. DockerException is an IOException.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/kubernetes/KubernetesEnvVar.java
amisevsk: Fixed

benoitf:
@amisevsk you need to rebase your branch (there is a conflict)

ibuziuk:
@amisevsk works just fine in most cases. However, I can reproduce pretty often the following error:
`
Unable to start workspace agent. Error when trying to start the workspace agent: Start of environment 'default' failed. Error: Image eclipse/ubuntu_jdk8:latest not found.
`
It is not reproducible while debugging and I guess it might be coupled that 5 seconds delays might be not enough for Image metadata to be obtained / pulled. WDYT ?

amisevsk:
@ibuziuk sounds like a likely issue; I'll look into it.

amisevsk:
@ibuziuk I've been able to reproduce the issue occasionally, but yeah, it doesn't happen when debugging or if the image has already been pulled. 

It's happening because OpenShift is taking longer than 5 seconds to get metadata about the docker repository. I've increased the max wait time to 10 seconds, which will hopefully fix the issue. However, since it doesn't always occur, I can't be sure. If it does occur again, could you manually check that the data is pulled (Builds -> Images -> \<image-name\> in OpenShift console)? That would confirm that it's just a timeout issue.

OpenShift doesn't seem to offer a way to monitor the progress here, so we are kind of stuck waiting. The (now 20 second) maximum wait time between pull and tag shouldn't have much of an impact in practical use, since the image stream tags are created quickly once the image has been pulled.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -165,25 +163,53 @@ public OpenShiftConnector(DockerConnectorConfiguration connectorConfiguration,
...
+        String imageStreamTagName = KubernetesStringUtils.getTagNameFromRepoString(imageForDocker);
+
+        // imageStreamTagName is not enough to fill out a pull spec; it is only the tag, so we
+        // have to get the ImageStreamTag from the tag, and then get the full ImageStreamTag name
+        // from that tag. This works because the tags used in Che are unique.
+        ImageStreamTag imageStreamTag = getImageStreamTagFromRepo(imageStreamTagName);
+        String imageStreamTagPullSpec = imageStreamTag.getMetadata().getName();
+
+        // Next we need to get the address of the registry where the ImageStreamTag is stored
+        String imageStreamName = imageStreamTagPullSpec.split(":")[0];
```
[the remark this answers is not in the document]
amisevsk: Agreed. I didn't do it before because OpenShift *should* complain that an ImageStreamTag name does not contain ":", but I'm not sure. Fixed.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -165,25 +163,53 @@ public OpenShiftConnector(DockerConnectorConfiguration connectorConfiguration,
...
+        ImageStream imageStream = openShiftClient.imageStreams()
+                                                 .inNamespace(openShiftCheProjectName)
+                                                 .withName(imageStreamName)
+                                                 .get();
+        String registryAddress = imageStream.getStatus()
```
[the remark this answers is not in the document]
amisevsk: Fixed.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -408,6 +434,194 @@ public Network inspectNetwork(InspectNetworkParams params) throws IOException {
                             .withEnableIPv6(false);
     }
 
+    @Override
+    public void pull(final PullParams params, final ProgressMonitor progressMonitor) throws IOException {
+        // This method does not cause the relevant image to actually be pulled to the local
```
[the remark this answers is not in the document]
amisevsk: Good idea -- done.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
amisevsk: With the latest commit, I've changed OpenShiftException to a subclass of IOException (like DockerException) and used it instead of DockerException.

l0rd (approved):

l0rd:
ci-build

amisevsk:
I've squashed and rebased onto master. Additionally, I've added a minor fix for tests broken by commit 41541e0.

PR #3960, opened 2017-01-31
Remove native launch scripts of Eclipse Che (replaced now by Docker CLI)

benoitf:
### What does this PR do?
Now, docker CLI is the way to start Eclipse Che so we don't need anymore "native scripts" and Vagrantfile

command to start che is : `docker run eclipse/che`
to use local binaries : add `-v <path-to-che-cloned-directory>:/repo
`

It was originally made as part of Pull Request https://github.com/eclipse/che/pull/3318
but it has been delayed to get stability on docker CLI and developers using it
https://github.com/eclipse/che/pull/3318#issuecomment-266115368

### Previous behavior
Che was launched with shell scripts or with Vagrant

### New behavior
Che is launched with CLI

### Tests written?
No

#### Changelog
Native scripts like cli.sh, che.sh, etc have been removed

**Release Notes**:
In 5.x a new cli based on docker has been introduced. It is allowing to start/stop/upgrade che with a standard way across all platforms.
So the native scripts that were relying on some specific operating system requirements are no longer needed and then are removed.

 `docker run eclipse/che` is the way to go

### Docs updated?
AFAIK in the docs we don't write how to use these scripts.

Change-Id: I8649830049974043b69f6725480c1799cfe325cd


TylerJewell:
We are going to need a Changelog entry and a Release Notes for this.  Removing this is an advisory that we need to state to developers who read the release notes and what the expected action is for them to take now.

benoitf:
@TylerJewell ok I've added changelog and release notes. The issue is that it's a fix of fix of fix as we already had pull requests removing those files. It was the last one missing. So maybe there will duplicates in changelog/release notes.

TylerJewell (approved):

PR #3961, opened 2017-01-31
CODENVY-1407: Log agent output on fail

mmorhun:
### What does this PR do?
Logs agent output if it fails to start.

https://github.com/codenvy/codenvy/issues/1407

#### Changelog
Add output of failed agent into tomcat logs.

#### Release Notes
N/A

#### Docs PR
N/A


agents/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/AbstractAgentLauncher.java
```diff
@@ -66,13 +68,33 @@ public AbstractAgentLauncher(long agentMaxStartTimeMs,
...
+     * @throws ServerException
+     *         if script execution failed
+     */
     @Override
     public void launch(Instance machine, Agent agent) throws ServerException {
         if (isNullOrEmpty(agent.getScript())) {
             return;
         }
+        ListLineConsumer agentLogger = new ListLineConsumer();
+        LineConsumer lineConsumer = new AbstractLineConsumer() {
```
garagatyi, line 77: CompositeLineConsumer?

garagatyi:
Can you show an example of output if ssh connection fails?

agents/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/AbstractAgentLauncher.java
```diff
@@ -66,13 +68,33 @@ public AbstractAgentLauncher(long agentMaxStartTimeMs,
...
+     * The machine should be started.
+     * Retrieves agent's logs if its start fails.
+     *
+     * @param machine
+     *         the machine instance in which agent will be started. The machine should be started.
+     * @param agent
+     *         the agent to start
+     * @throws ServerException
+     *         if script execution failed
+     */
```
tolusha: Doc should be moved to interface.

agents/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/AbstractAgentLauncher.java
mmorhun: Possible, but in that case we shouldn't close `CompositeLineConsumer`, only `agentLogger` (which is part of composite consumer).

agents/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/AbstractAgentLauncher.java
```diff
@@ -66,13 +68,33 @@ public AbstractAgentLauncher(long agentMaxStartTimeMs,
         this.agentLaunchingChecker = agentLaunchingChecker;
     }
 
+    /**
+     * Launches agent script in machine and waits until agent starts.
+     * The machine should be started.
+     * Retrieves agent's logs if its start fails.
```
garagatyi: As far as I can see it logs agent logs if it fails, but stores it in any case. Is it needed to add logging related logic into javadocs?

garagatyi (approved):
I'm not big fan of putting agents logs into memory and into logs, but ok with that if we need it.
@tolusha Please take a look

tolusha (approved):

Issue #3886, opened 2017-01-24
CLI - Create graceful and forced stop options


TylerJewell:
Eclipse Che is implementing a REST service that will offer a graceful shutdown of all services. Today, when we stop the server, Tomcat will be stopped if it has not naturally stopped within 4 minutes. If you have a large number of running workspaces, you may want them to all be stopped (with appropriate snapshots) before initiating Tomcat shutdown.  This REST service executes this process.

We need to extend the CLI to support two forms of stop:
1. forced stop (essentially what we do now where we just stop all containers without concern for internals)
2. graceful stop (default - which will execute the REST service).

The REST [service is documented here](https://github.com/codenvy/codenvy/issues/1568).  For Codenvy variations of this implemementation, it will require an admin token to complete.

The syntax:
```
eclipse/che stop [--force] [--user username] [--password password]
```
The `[--force]` is the current approach to stopping. The new default would invoke the REST service first.  We would then need to implement a periodic polling mechanism that prints out appropriate status while the stopping process is completing.  In the Codenvy variant, we will need to also ask for username / password on the command line to proceed.  These parameters can optionally be provided as parameters.

benoitf:
issue for che for REST API is https://github.com/eclipse/che/issues/3892

TylerJewell:
We discussed some implementation details:
1. We may opt to use custom action to interact with REST service.
2. We will not need any prompts for Che shutdown - we may need them for Codenvy shutdown.

benoitf:
Adding pull request https://github.com/eclipse/che/pull/3968

PR #3968, opened 2017-01-31
[cli] graceful stop command

benoitf:
### What does this PR do?
Introduce graceful stop command for cli
by default, `stop`command is performing a graceful stop : stop all services then stop and remove containers
using `--force`the graceful stop is not done : we stop containers and remove containers

auth can be used with `--user` and `--password`

#3886

#### Changelog
Add graceful stop (disable with --force) in CLI

**Release Notes**:
We have introduced a way to perform a graceful stop of Che. Previously, Che server stops would be timeout based, meaning that if any workspaces were running when the Che server was stopped, they would continue to run. We have introduced a graceful stop that will initiate shutdown of all workspaces before initiating a shutdown of the server. If workspaces are configured to auto-snapshot during stop, then graceful stop will wait for all workspaces to be snapped before stopping the server. 

The default CLI `stop` command now performs a graceful stop. It stops all services then stops and remove containers. You can optionally add a `--force` option to stop and remove containers without waiting for workspaces to stop. For authenticated servers like Codenvy, authentication can be provided by `--user` and `--password`.

default graceful stop: `docker run  ... eclipse/che stop`
skip graceful stop: `docker run  ... eclipse/che stop --force`
use auth : `docker run  ... eclipse/che stop --user <myemail> --password <my-secret>`

### Docs Pull Request
https://github.com/eclipse/che-docs/pull/117
https://github.com/codenvy/codenvy-docs/pull/61

### DIFF
--- a/dockerfiles/base/scripts/base/commands/cmd_destroy.sh
+++ b/dockerfiles/base/scripts/base/commands/cmd_destroy.sh
@@ -37,7 +37,7 @@ cmd_destroy() {
     return;
   fi
 
-  cmd_stop
+  cmd_stop --force
 
   info "destroy" "Deleting instance and config..."
 
--- a/dockerfiles/base/scripts/base/commands/cmd_start.sh
+++ b/dockerfiles/base/scripts/base/commands/cmd_start.sh
@@ -125,14 +125,8 @@ cmd_start_check_agent_network() {
   fi
 }
 
-cmd_stop() {
-  debug $FUNCNAME
-
-  if [ $# -gt 0 ]; then
-    error "${CHE_MINI_PRODUCT_NAME} stop: You passed unknown options. Aborting."
-    return
-  fi
-
+# stop containers booted by docker compose and remove them
+stop_containers() {
   info "stop" "Stopping containers..."
   if is_initialized; then
     log "docker_compose --file=\"${REFERENCE_CONTAINER_COMPOSE_FILE}\" -p=$CHE_MINI_PRODUCT_NAME stop -t ${CHE_COMPOSE_STOP_TIMEOUT} >> \"${LOGS}\" 2>&1 || true"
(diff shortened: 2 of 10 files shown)


TylerJewell:
Since this is a new feature then the release notes to have a deeper explanation and example.

benoitf:
@TylerJewell ok I've updated release notes section

riuvshin:
what will happen if auth is not provided ?
like in first example `docker run ... eclipse/che stop`

benoitf:
@riuvshin if auth is required it will tell you that you can't perform operation as you need authentication

riuvshin:
so maybe we have to add `--force` here  https://github.com/eclipse/che/blob/master/dockerfiles/base/scripts/base/commands/cmd_destroy.sh#L40 because in case of destroy who will care about auth

benoitf:
@riuvshin yes good catch

benoitf:
@riuvshin what about cmd_stop in cmd_upgrade ?

riuvshin:
in `cmd_upgrade`  we must do graceful stop, so I think we have to change `cmd_upgrade` to ask user do manually stop with auth before `cmd_upgrade` same as we do for `cmd_backup` https://github.com/eclipse/che/blob/master/dockerfiles/base/scripts/base/commands/cmd_backup.sh#L29 

@TylerJewell @benoitf wdyt guys?

benoitf:
@riuvshin yes looks like the proper solution for me. So the user can decide if he performs a stop right now, delay it and do it with force or not

TylerJewell:
Yes, I think that makes sense to me, too.

benoitf:
ok, I've added the changes

TylerJewell:
I updated the release notes comment.

TylerJewell (approved):

riuvshin (approved):

Issue #3811, opened 2017-01-19
Move Che Orion Editor resources to the che-lib 


vparfonov:
In case we don't change those js and css files a lot we don't need to contain them in che/plugins repository.

PR #3979, opened 2017-02-01
CHE-3811: move js file to the che-lib/che-orion-editor

vparfonov:
### What does this PR do?

Move all JS files of Orion from plugin-orion to the che-lib repository. This allow as avoid problem with compilation plugin-orion on Windows OS because during compilation we modify  some JS files with using patch utils which not available on Windows OS machine. 
So for now we move this step to the che-lib, by default our contributors don't need modify this code. 

Related to: https://github.com/eclipse/che-lib/pull/18

#3811 

#### Changelog
Move all JS files of Orion from plugin-orion to the che-lib repository


azatsarynnyy (approved):

vparfonov:
ci-build

JamesDrummond:
Will this effect documentation? 

- Is the code editor still a plugin?
https://github.com/eclipse/che-docs/blame/955f43f787f53462d55ef09598a71eb52cbe31bf/src/main/_docs/write-che-ide-plugins/plugins-code-editors.md#L102

- I think embed still makes sense.
https://github.com/eclipse/che-docs/blame/master/src/main/_docs/write-che-ide-plugins/plugins-embed-htmljs.md#L11

vparfonov:
@JamesDrummond any changes in docs not required, this pretty internal stuff. Script still be loaded in same way by IDE application, we just change location for avoiding compiling problem on Windows machine.

JamesDrummond (approved):

evidolob (approved):

Issue #3064, opened 2016-11-14
Add error logging on predestroy


akorneta:
We have components that clean up after themselves on stopping of the application server. Such components can produce exceptions on predestroy phase and some of this exceptions where handled without logging by `DUMMY` implementation of `DestroyErrorHandler`, so it is therefore proposed to add logging for such cases.

skabashnyuk:
@akorneta  can you provide examples of list of such classes?

PR #3990, opened 2017-02-01
CHE-3064: Add log errors on pre-destroy phase

vinokurig:
### What does this PR do?
Log the errors on pre-destroy phase

https://github.com/eclipse/che/issues/3064

#### Changelog
Errors that might happen on application server stopping, will be logged.

#### Release Notes
Technical debt N/A


core/commons/che-core-commons-inject/src/main/java/org/eclipse/che/inject/lifecycle/DestroyErrorHandler.java
```diff
@@ -21,12 +23,8 @@
     void onError(Object instance, Method method, Throwable error);
 
     /**
-     * Implementation of DestroyErrorHandler that ignore errors, e.g. such behaviour is required for annotation {@link
+     * Implementation of DestroyErrorHandler that log errors, e.g. such behaviour is required for annotation {@link
```
akorneta: actually `@javax.annotation.PreDestroy` does require the logging of errors and provided documentation can confused a bit.

akorneta (approved):

skabashnyuk (approved):

PR #4023, opened 2017-02-06
Adds system graceful stop tracking events

### LINKED ISSUE #3921: Add tracking events to the System graceful stop
#3892 is introducing a system "graceful shutdown" provided by initial PR #3917

Clients like CLI might be interested to display nice output to users during the stop

it could be something like @evoevodin suggested (I've added extra current/total)

```json
{
    "type" : "STOPPING_SERVICE", 
    "service" : "workspace"
}
```
```json
{
    "type" : "SERVICE_ITEM_STOPPED", 
    "id" : "workspace0x123456",
    "service" : "workspace",
    "current" : "12",
    "total": "36"
}  
```
```json
{
    "type" : "SERVICE_STOPPED", 
    "service" : "workspace"
}  
```

Then CLI could display  
```
Stopping service: workspace
   'workspace123' is stopped [12/36]
   'workspace234' is stopped [13/36]
   'workspace345' is stopped [14/36]
   service stopped

Stopping service: ldap
...
```

voievodin:
### What does this PR do?

Adds 3 new tracking events published during system shutdown, they are:

* STOPPING_SERVICE - published before service is stopped. This event is always published
```json
{
        "type" : "STOPPING_SERVICE",
        "service" : "workspace"
}
```
* SERVICE_ITEM_STOPPED - published each time service item(like workspace) is stopped.
This event may be published 0 or N times between (STOPPING_SERVICE and SERVICE_STOPPED) events.
```json
{
        "type" : "SERVICE_ITEM_STOPPED",
        "service" : "workspace",
        "item" : "workspace0x12345",
        "current" : 13,
        "total" : 32
}
```
The fields `current` & `total` are optional and either present or missing both.
So if service manages quantity of its items then "current" & "total" should 
be present in the event, otherwise they must not be included to the event body.

* SERVICE_STOPPED - published after service is stopped. This event is always published
```json
{
        "type" : "SERVICE_STOPPED",
        "service" : "workspace"
}        
```

Resolves #3921 

#### Changelog
Publish tracking events during system graceful stop

#### Release Notes
System events set is reached with 3 new events: _STOPPING_SERVICE_, 
_SERVICE_ITEM_STOPPED_, _SERVICE_STOPPED_. These events provide more useful 
information about services being stopped during graceful system shutdown.

#### Docs PR
N/A

### DIFF
--- a/wsmaster/che-core-api-system-shared/src/main/java/org/eclipse/che/api/system/shared/dto/SystemServiceEventDto.java
+++ b/wsmaster/che-core-api-system-shared/src/main/java/org/eclipse/che/api/system/shared/dto/SystemServiceEventDto.java
@@ -1,0 +1,31 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2017 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.api.system.shared.dto;
+
+import org.eclipse.che.dto.shared.DTO;
+
+/**
+ * DTO for system service events.
+ *
+ * @author Yevhenii Voevodin
+ */
+@DTO
+public interface SystemServiceEventDto extends SystemEventDto {
+
+    /**
+     * Returns the name of the service described by this event.
+     */
+    String getService();
+
+    void setService(String service);
+
+    SystemServiceEventDto withService(String service);
+}
--- a/wsmaster/che-core-api-system-shared/src/main/java/org/eclipse/che/api/system/shared/event/service/SystemServiceEvent.java
+++ b/wsmaster/che-core-api-system-shared/src/main/java/org/eclipse/che/api/system/shared/event/service/SystemServiceEvent.java
@@ -1,0 +1,59 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2017 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.api.system.shared.event.service;
+
+import org.eclipse.che.api.system.shared.event.SystemEvent;
+
+import java.util.Objects;
+
+/**
+ * The base class for system service events.
+ *
+ * @author Yevhenii Voevodin
+ */
+public abstract class SystemServiceEvent implements SystemEvent {
+
+    protected final String serviceName;
+
+    protected SystemServiceEvent(String serviceName) {
+        this.serviceName = Objects.requireNonNull(serviceName, "Service name required");
+    }
+
+    public String getServiceName() {
+        return serviceName;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (!(obj instanceof SystemServiceEvent)) {
+            return false;
+        }
+        final SystemServiceEvent that = (SystemServiceEvent)obj;
+        return Objects.equals(getType(), that.getType())
+               && Objects.equals(serviceName, that.serviceName);
+    }
+
+    @Override
+    public int hashCode() {
+        int hash = 17;
+        hash = 31 * hash + Objects.hashCode(getType());
+        hash = 31 * hash + Objects.hashCode(serviceName);
+        return hash;
+    }
+
+    @Override
+    public String toString() {
+        return getClass().getSimpleName() + "{eventType='" + getType() + "', serviceName=" + serviceName + "'}";
+    }
+}
--- a/wsmaster/che-core-api-system-shared/src/main/java/org/eclipse/che/api/system/shared/event/service/SystemServiceItemStoppedEvent.java
+++ b/wsmaster/che-core-api-system-shared/src/main/java/org/eclipse/che/api/system/shared/event/service/SystemServiceItemStoppedEvent.java
@@ -1,0 +1,98 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2017 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.api.system.shared.event.service;
+
+import org.eclipse.che.api.system.shared.event.EventType;
+import org.eclipse.che.commons.annotation.Nullable;
+
+import java.util.Objects;
+
+/**
+ * See {@link EventType#SERVICE_ITEM_STOPPED} description.
+ *
+ * @author Yevhenii Voevodin
+ */
+public class SystemServiceItemStoppedEvent extends SystemServiceEvent {
+
+    private final String item;
+
+    private Integer total;
+    private Integer current;
+
+    public SystemServiceItemStoppedEvent(String serviceName, String item) {
+        super(serviceName);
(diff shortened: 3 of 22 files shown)


wsmaster/che-core-api-system-shared/src/main/java/org/eclipse/che/api/system/shared/dto/SystemServiceEventDto.java
```diff
@@ -0,0 +1,41 @@
...
+ * DTO for system service events.
+ *
+ * @author Yevhenii Voevodin
+ */
+@DTO
+public interface SystemServiceEventDto extends SystemEventDto {
+
+    /** Creates a new dto from given event. */
+    static SystemServiceEventDto fromEvent(SystemServiceEvent event) {
+        SystemServiceEventDto dto = DtoFactory.newDto(SystemServiceEventDto.class);
```
sleshchenko: Using of `....dto.server.DtoFactory` may break gwt client when it will use it.

wsmaster/che-core-api-system-shared/src/main/java/org/eclipse/che/api/system/shared/event/service/SystemServiceItemStoppedEvent.java
```diff
@@ -0,0 +1,96 @@
...
+        hash = 31 * hash + super.hashCode();
+        hash = 31 * hash + item.hashCode();
+        hash = 31 * hash + Objects.hashCode(total);
+        hash = 31 * hash + Objects.hashCode(current);
+        return hash;
+    }
+
+    @Override
+    public String toString() {
+        return "SystemServiceItemStoppedEvent{" +
```
sleshchenko, line 90: Looks like it will have next format 
```
SystemServiceItemStoppedEvent{item='1', ....} class org.eclipse.che.api.system.shared.event.service.SystemServiceItemStoppedEvent {evenType='someType',...}
```
I'd rewrite it with using getters from super class

wsmaster/che-core-api-system-shared/src/main/java/org/eclipse/che/api/system/shared/event/service/SystemServiceEvent.java
```diff
@@ -0,0 +1,59 @@
...
+    public int hashCode() {
+        int hash = 17;
+        hash = 31 * hash + Objects.hashCode(getType());
+        hash = 31 * hash + Objects.hashCode(serviceName);
+        return hash;
+    }
+
+    @Override
+    public String toString() {
+        return getClass() + "{evenType='" + getType() + "', serviceName='" + serviceName + "'}";
```
sleshchenko: I'd change `getClass()` to `getClass().getSimpleName`

wsmaster/che-core-api-system/src/test/java/org/eclipse/che/api/system/server/SystemEventsWebsocketBroadcasterTest.java
```diff
@@ -0,0 +1,87 @@
+package org.eclipse.che.api.system.server;
```
sleshchenko, line 11: Execute `mvn license:format` to fix licenses.

sleshchenko (approved):
Other looks good to me! :+1:

wsmaster/che-core-api-system-shared/src/main/java/org/eclipse/che/api/system/shared/dto/SystemServiceEventDto.java
voievodin: Thanks! I reworked it to use delegation rules instead

voievodin:
ci-build

wsmaster/che-core-api-system/src/main/java/org/eclipse/che/api/system/server/SystemEventsWebsocketBroadcaster.java
```diff
@@ -54,13 +67,41 @@ public void onEvent(SystemEvent event) {
...
+         *         channel to send message to
+         * @param message
+         *         message to send
+         * @throws Exception
+         *         when any error occurs
+         */
+        void sendMessage(String channel, String message) throws Exception;
+    }
+
+    private static class WebsocketMessageSender implements MessageSender {
```
garagatyi: We already have pretty the same class org.eclipse.che.api.core.util.WebsocketMessageConsumer. Please consider usage of it.

wsmaster/che-core-api-system/src/main/java/org/eclipse/che/api/system/server/WorkspaceServiceTermination.java
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * Propagates workspace stopped events as {@link SystemServiceStoppedEvent} events.
+     */
+    private class WorkspaceStoppedEventsPropagator implements EventSubscriber<WorkspaceStatusEvent> {
+
+        private final int           totalRunning;
+        private final AtomicInteger currentlyStopped;
+
+        private WorkspaceStoppedEventsPropagator() {
+            this.totalRunning = workspaceManager.getRunningWorkspacesIds().size();
```
garagatyi, line 62: I missed what will happen with WSs that are still starting at that time? Are their IDs covered by this method?

garagatyi (commented):
Looks good! Code is clear and well tested.
Also I think that there is a place to cleanup existing code, please consider improvements that I explained in inline comments.

wsmaster/che-core-api-system/src/main/java/org/eclipse/che/api/system/server/WorkspaceServiceTermination.java
voievodin: Yes

voievodin:
ci-build please

wsmaster/che-core-api-system/src/main/java/org/eclipse/che/api/system/server/WorkspaceServiceTermination.java
garagatyi: Then method name `getRunningWorkspacesIds` looks a bit confusing IMO

voievodin: I consider running as not stopped, because if it's running you can't start another one, which is true for starting/snapshotting/stopping workspaces. WorkspaceRuntimes provides methods like `isAnyRunning`, `hasRuntime`, `getRuntimesIds` that work exactly in the same way(with not-stopped workspaces). Do you have any proposal?

garagatyi: We have running state of WS. That's why this method name looks confusing to me. On the other hand hasRuntimes doesn't confuse me because term differs. 
I would say getActiveWSIds but we don't have term active so it's not good to use it here. 
I don't have good suggestion, so it is up to you to decide how it should be named.

wsmaster/che-core-api-system/src/main/java/org/eclipse/che/api/system/server/DtoConverter.java
```diff
@@ -0,0 +1,83 @@
...
+import org.eclipse.che.api.system.shared.event.service.SystemServiceEvent;
+import org.eclipse.che.api.system.shared.event.service.SystemServiceItemStoppedEvent;
+import org.eclipse.che.dto.server.DtoFactory;
+
+/**
+ * Converts events to corresponding DTOs.
+ *
+ * @author Yevhenii Voevodin
+ */
+final class DtoConverter {
```
garagatyi: Can you make it public?

garagatyi (approved):
Looks good!

voievodin:
@JamesDrummond @eivantsov can you please verify that changelog & release notes are okay

ghost (approved):

wsmaster/che-core-api-system-shared/src/main/java/org/eclipse/che/api/system/shared/event/service/SystemServiceEvent.java
```diff
@@ -0,0 +1,59 @@
...
+    public int hashCode() {
+        int hash = 17;
+        hash = 31 * hash + Objects.hashCode(getType());
+        hash = 31 * hash + Objects.hashCode(serviceName);
+        return hash;
+    }
+
+    @Override
+    public String toString() {
+        return getClass().getSimpleName() + "{evenType='" + getType() + "', serviceName=" + serviceName + "'}";
```
sleshchenko: typo **even**Type

JamesDrummond (approved):
Changelog and release notes look good to me.

skabashnyuk (approved):

Issue #3993, opened 2017-02-01
error in cli output


riuvshin:
when I was running nightly server updates I've got following error in CLI output

```
[0;32mINFO:[0m (codenvy config): Generating codenvy configuration...
[mNotice: Compiled catalog for 1511be7f32f2.ec2.internal in environment production in 0.55 seconds[0m
[mNotice: Applied catalog in 0.09 seconds[0m
[0;32mINFO:[0m (codenvy config): Customizing docker-compose for running in a container
[0;32mINFO:[0m (codenvy start): Preflight checks
WARNING: bridge-nf-call-iptables is disabled
WARNING: bridge-nf-call-ip6tables is disabled
/scripts/base/library.sh: line 309: [: .: integer expression expected
         mem (1.5 GiB):           [0;32m[OK][0m
         disk (100 MB):           [0;32m[OK][0m
         port 80 (http):          [0;32m[AVAILABLE][0m
         port 443 (https):        [0;32m[AVAILABLE][0m
         port 2181 (zookeeper):   [0;32m[AVAILABLE][0m
         port 5000 (registry):    [0;32m[AVAILABLE][0m
         port 23750 (socat):      [0;32m[AVAILABLE][0m
         port 23751 (swarm):      [0;32m[AVAILABLE][0m
         port 32000 (jmx):        [0;32m[AVAILABLE][0m
         port 32001 (jmx):        [0;32m[AVAILABLE][0m
         port 8000 (debug):        [0;32m[AVAILABLE][0m
         port 9000 (lighttpd):    [0;32m[AVAILABLE][0m
         conn (browser => ws):    [0;32m[OK][0m
         conn (server => ws):     [0;32m[OK][0m

[0;32mINFO:[0m (codenvy start): Starting containers...
Creating codenvy_rsyslog_1
```

this is somehow related to version checks https://github.com/eclipse/che/blob/master/dockerfiles/base/scripts/base/library.sh#L309

@TylerJewell can you please take a look on this

TylerJewell:
I've tried running a whole series of tests and unable to reproduce so far.  I think this maybe was an image update issue somehow.

riuvshin:
@TylerJewell oki, let's close this and I will reopen if I will see it again, because Im not able to reproduce this  also

benoitf:
if you can use --trace option if you see it again

riuvshin:
Reproduced issue on stg
trace output
```
[...]

PR #4028, opened 2017-02-06
CLI - Fix for CHE-3993 - Proper integer comparison

TylerJewell:

### What does this PR do?
This PR adds a function to the base CLI to use awk to do numerical comparisons. Our default function was having problems if the decimal place was in a different location, and this improvement makes it simpler to do these checks.

This check was being used when comparing RAM and disk sizes to minimums.

#3993 

#### Changelog
Add awk function to do numerical comparisons in CLI but reverted by PR #4035.

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/dockerfiles/base/scripts/base/commands/cmd_start.sh
+++ b/dockerfiles/base/scripts/base/commands/cmd_start.sh
@@ -87,20 +87,23 @@ cmd_start_check_host_resources() {
   HOST_RAM=$(echo ${HOST_RAM#*:} | xargs)
   HOST_RAM=${HOST_RAM% *}
 
+  COMPARE=$(echo "$HOST_RAM<$CHE_MIN_RAM" | bc -l)
+
   PREFLIGHT=""
-  if $(less_than "31.37" "1.5"); then
+  if [[ "$COMPARE" = "0" ]]; then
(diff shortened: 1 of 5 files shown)


dockerfiles/base/scripts/base/commands/cmd_start.sh
```diff
@@ -87,28 +87,30 @@ cmd_start_check_host_resources() {
   HOST_RAM=$(echo ${HOST_RAM#*:} | xargs)
   HOST_RAM=${HOST_RAM% *}
 
+  COMPARE=$(echo "$HOST_RAM<$CHE_MIN_RAM" | bc -l)
```
benoitf: we could use awk so we don't need to add another tool inside the image

/ # awk 'BEGIN { print (1.543 >= 5.234) ? 0  : 1}'
1
/ # awk 'BEGIN { print (15.43 >= 5.234) ? 0  : 1}'
0

riuvshin (approved):

TylerJewell:
Per @benoitf request, I removed `bc` and added in a comparison using awk instead.

benoitf (approved):

Issue #4024, opened 2017-02-06
Use of smallest possible workspace (/container)


benoitf:
As part of https://github.com/eclipse/che/issues/3748 and Diagnostic agent, idea is to use the smallest possible workspace

Using the smallest possible workspace result in having also the smallest possible image
alpines images are among the smallest possible images.

so basically I would be able to:
- [x] use an alpine image (without glibc installed) (PR submitted : https://github.com/eclipse/che/pull/4027 )
- [x] use software provided by default in alpine without adding extra packages at runtime requiring internet connection.  ( PR submitted : https://github.com/eclipse/che/pull/4029 )
- [x] allow to use openJDK 1.8 alpine image directly (for now if images has no cmd command it is rejected) ( PR #4702 )

It would be also very nice to
- [ ] avoid the installation of extra agents (like terminal or exec agent if we don't require them. For now, these agents are installed as a dependency of the workspace agent)

**Reproduction Steps:**
Create a workspace with openjdk:8-jre-alpine and try to start a workspace.

PR #4029, opened 2017-02-06
Allow to use wget in addition to curl when installing workspace agent + terminal agent

benoitf:
### What does this PR do?

For example alpine images have natively wget tool installed (in a busybox version) but not curl by default
So, we could check that either curl or wget is installed instead of only checking if curl is installed as we only need to download binaries which wget can perform easily as well
It is speeding up the first boot of the workspaces based on this configuration. (small alpine images)

#4024 

#### Changelog
Allow to use wget instead of only curl for installing ws agent or terminal agent

#### Release Notes
Allow to use wget instead of only curl for installing ws agent or terminal agent

#### Docs PR
N/A

Change-Id: I7d02fb70f5f2fb03432072f4f33ff27532c8601e

### DIFF
--- a/wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
+++ b/wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
@@ -12,7 +12,16 @@
 unset PACKAGES
 unset SUDO
 command -v tar >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" tar"; }
-command -v curl >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" curl"; }
+CURL_INSTALLED=false
+WGET_INSTALLED=false
+command -v curl >/dev/null 2>&1 && CURL_INSTALLED=true
+command -v wget >/dev/null 2>&1 && WGET_INSTALLED=true
+
+# no curl, no wget, install curl
+if [ ${CURL_INSTALLED} = false ] && [ ${WGET_INSTALLED} = false ]; then
+  { PACKAGES=${PACKAGES}" curl"; }
+fi
+
 test "$(id -u)" = 0 || SUDO="sudo -E"
 
 LOCAL_AGENT_BINARIES_URI="/mnt/che/ws-agent.tar.gz"
@@ -233,13 +242,20 @@ eval "DOWNLOAD_AGENT_BINARIES_URI=${DOWNLOAD_AGENT_BINARIES_URI}"
 
 if [ -f "${LOCAL_AGENT_BINARIES_URI}" ] && [ -s "${LOCAL_AGENT_BINARIES_URI}" ]
 then
-    AGENT_BINARIES_URI="file://${LOCAL_AGENT_BINARIES_URI}"
+    tar zxf "${LOCAL_AGENT_BINARIES_URI}" -C ${CHE_DIR}/ws-agent
 else
     echo "Workspace Agent will be downloaded from Workspace Master"
     AGENT_BINARIES_URI=${DOWNLOAD_AGENT_BINARIES_URI}
+    if [ ${CURL_INSTALLED} = true ]; then
+      curl -s  ${AGENT_BINARIES_URI} | tar  xzf - -C ${CHE_DIR}/ws-agent
+    else
+      # use wget
+      wget -qO- ${AGENT_BINARIES_URI} | tar xzf - -C ${CHE_DIR}/ws-agent
+    fi
+
 fi
 
-curl -s  ${AGENT_BINARIES_URI} | tar  xzf - -C ${CHE_DIR}/ws-agent
+
 
 ###############################################
 ### ws-agent run command will be added here ###
--- a/wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
+++ b/wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
@@ -19,7 +19,7 @@ command -v wget >/dev/null 2>&1 && WGET_INSTALLED=true
 
 # no curl, no wget, install curl
 if [ ${CURL_INSTALLED} = false ] && [ ${WGET_INSTALLED} = false ]; then
-  { PACKAGES=${PACKAGES}" curl"; }
+  PACKAGES=${PACKAGES}" curl";
 fi
 
 test "$(id -u)" = 0 || SUDO="sudo -E"
@@ -249,6 +249,9 @@ else
     if [ ${CURL_INSTALLED} = true ]; then
       curl -s  ${AGENT_BINARIES_URI} | tar  xzf - -C ${CHE_DIR}/ws-agent
     else
+      # replace https by http as wget may not be able to handle ssl
+      AGENT_BINARIES_URI=$(echo ${AGENT_BINARIES_URI} | sed 's/https/http/g')
+
       # use wget
       wget -qO- ${AGENT_BINARIES_URI} | tar xzf - -C ${CHE_DIR}/ws-agent
     fi
--- a/agents/exec/src/main/resources/org.eclipse.che.exec.script.sh
+++ b/agents/exec/src/main/resources/org.eclipse.che.exec.script.sh
@@ -12,7 +12,16 @@
 unset PACKAGES
 unset SUDO
 command -v tar >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" tar"; }
-command -v curl >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" curl"; }
+CURL_INSTALLED=false
+WGET_INSTALLED=false
+command -v curl >/dev/null 2>&1 && CURL_INSTALLED=true
+command -v wget >/dev/null 2>&1 && WGET_INSTALLED=true
+
+# no curl, no wget, install curl
+if [ ${CURL_INSTALLED} = false ] && [ ${WGET_INSTALLED} = false ]; then
+  PACKAGES=${PACKAGES}" curl";
+fi
+
 test "$(id -u)" = 0 || SUDO="sudo -E"
 
 CHE_DIR=$HOME/che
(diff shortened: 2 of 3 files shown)


JamesDrummond:
@benoitf Should we add wget or curl to custom recipe requirements?

benoitf:
@JamesDrummond it's not a requirement : if it's not there it's installed

JamesDrummond (approved):

garagatyi:
I don't see pros in installing curl if scripts are adapted to work with wget. Wget is more lightweight and widespread. It is also usually pre-installed in most linux distros

garagatyi:
@tolusha WDYT?

tolusha:
@benoitf 
Would new script be able to download agent binaries from mounted directory?

garagatyi:
So I'm +1 on this PR. I also +1 on installing wget instead of curl if both are missing. 
@tolusha It doesn't have to download it. It can check existence of file on local FS. I took 1 look at PR content and looks like it does exactly that.

tolusha (approved):

garagatyi:
I also see that all of the language servers agents (c#, json, php, python, TS) can be improved in the same way. It would be great if someone managed that.

garagatyi:
@benoitf WDYT about installing wget instead of curl in case both are missing? BTW I don't see use-case when usage of curl is needed at all since I can't imagine situation when curl is present and wget is not.

voievodin (approved):

benoitf:
so after some investigation, my changes are working fine on http
but it fails on https :-/ so for codenvy.io it doesn't work :-(( we need for https to install wget and/or ca-certificates packages

@tolusha 
> @benoitf
> Would new script be able to download agent binaries from mounted directory?
> 
yes it works when agent binaries are on a mounted directory. For these one we don't use curl or wget, we use file directly

@garagatyi 
> @benoitf WDYT about installing wget instead of curl in case both are missing? BTW I don't see use-case when usage of curl is needed at all since I can't imagine situation when curl is present and wget is not.
yes I'm ok as well. It's just that at first I didn't want to replace curl by wget if both are missing to keep more compliant stuff with the existing. But I'm fine with that as well

garagatyi:
> but it fails on https :-/ so for codenvy.io it doesn't work :-(( we need for https to install wget and/or ca-certificates packages

Can you elaborate on that. I'm not really following you.

garagatyi:
Thank for the detailed explanation. Situation is clear for me now.

garagatyi:
I think @riuvshin is against serving it by non secured protocol.
So IMO it is up to you to choose the way (curl or wget + certificates).
Also it can be useful to check whether protocol is http before installing certificates. We can also check if installation is needed to not perform apt-get update if it is not needed. But these are just additional features, so they are not mandatory.

benoitf:
@garagatyi yes I was thinking as you, to check http protocol (http/https) before requiring extra packages and to install packages only if they're really required

@riuvshin as I understand, the access to the resources are accessed by the swarm nodes that are close to the workspace master so I'm not sure to see what are the benefits on forcing https over http for /agent-binaries/. It's a read-only operation and we don't provide any security/token details.

riuvshin:
@benoitf does this pr cover using `http` for `/agent-binaries/` or it is still something to do?

benoitf:
@riuvshin It does not cover this way but it's something that could be great if you agree.

small recap:
alpine images come with wget busybox installed by default
in Che (if not mounted) and in Codenvy, agents are downloaded on haproxy `/agent-binaries`

for now, it is using curl. So on alpine it requires to install curl but we have wget
so the current PR fixes the fact that if wget is installed we avoid to install curl (we boot faster and limit downloads)
But for https (which is the case for codenvy.io and other ppl using codenvy with https), we have the issue that wget busybox doesn't support ssl
so : 
option 1. for `/agent-binaries` we allow both http and https (agents will use http)
option 2. it means that on alpine images, we will have to install extra packages (like openssl, GNU wget, etc) only to download the binary while with http support no extra package is required

riuvshin:
ok I've got the idea need to check if we can do this with hapoxy, i mean allow some path tobe http when https is on

benoitf:
@riuvshin http://stackoverflow.com/a/36794795/4201908
http://stackoverflow.com/a/25007004/4201908

riuvshin:
@benoitf here is my branch I believe this will work https://github.com/codenvy/codenvy/commit/835a17baebe43b16ed096e6b19435a12fca4e97c

benoitf:
@riuvshin Ok I'm not sure how to test ssl install quickly

riuvshin:
@benoitf that is not easy to test locally I think I can play with it on stg tomorrow

agents/exec/src/main/resources/org.eclipse.che.terminal.script.sh
```diff
@@ -12,7 +12,16 @@
...
 command -v tar >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" tar"; }
-command -v curl >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" curl"; }
+CURL_INSTALLED=false
+WGET_INSTALLED=false
+command -v curl >/dev/null 2>&1 && CURL_INSTALLED=true
+command -v wget >/dev/null 2>&1 && WGET_INSTALLED=true
+
+# no curl, no wget, install curl
+if [ ${CURL_INSTALLED} = false ] && [ ${WGET_INSTALLED} = false ]; then
+  { PACKAGES=${PACKAGES}" curl"; }
```
garagatyi: I think this statement can be simplified by removing curly brackets

agents/exec/src/main/resources/org.eclipse.che.terminal.script.sh
```diff
@@ -149,23 +158,49 @@ eval "LOCAL_AGENT_BINARIES_URI=${LOCAL_AGENT_BINARIES_URI}"
...
+    LOCAL_AGENT_PATH=${LOCAL_AGENT_BINARIES_URI}
 elif [ -f $(echo "${LOCAL_AGENT_BINARIES_URI}" | sed "s/-${PREFIX}//g") ]; then
     AGENT_BINARIES_URI="file://"$(echo "${LOCAL_AGENT_BINARIES_URI}" | sed "s/-${PREFIX}//g")
+    LOCAL_AGENT_PATH=$(echo "${LOCAL_AGENT_BINARIES_URI}" | sed "s/-${PREFIX}//g")
 else
-    echo "Terminal Agent will be downloaded from Workspace Master"
     AGENT_BINARIES_URI=${DOWNLOAD_AGENT_BINARIES_URI}
 fi
 
+AGENT_BINARIES_URI=${DOWNLOAD_AGENT_BINARIES_URI}
```
garagatyi: Looks like either this or line 169 is not needed

wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
```diff
@@ -12,7 +12,16 @@
...
 command -v tar >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" tar"; }
-command -v curl >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" curl"; }
+CURL_INSTALLED=false
+WGET_INSTALLED=false
+command -v curl >/dev/null 2>&1 && CURL_INSTALLED=true
+command -v wget >/dev/null 2>&1 && WGET_INSTALLED=true
+
+# no curl, no wget, install curl
+if [ ${CURL_INSTALLED} = false ] && [ ${WGET_INSTALLED} = false ]; then
+  { PACKAGES=${PACKAGES}" curl"; }
```
garagatyi: See above

wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
```diff
@@ -233,13 +242,20 @@ eval "DOWNLOAD_AGENT_BINARIES_URI=${DOWNLOAD_AGENT_BINARIES_URI}"
...
-    AGENT_BINARIES_URI="file://${LOCAL_AGENT_BINARIES_URI}"
+    tar zxf "${LOCAL_AGENT_BINARIES_URI}" -C ${CHE_DIR}/ws-agent
 else
     echo "Workspace Agent will be downloaded from Workspace Master"
     AGENT_BINARIES_URI=${DOWNLOAD_AGENT_BINARIES_URI}
+    if [ ${CURL_INSTALLED} = true ]; then
+      curl -s  ${AGENT_BINARIES_URI} | tar  xzf - -C ${CHE_DIR}/ws-agent
+    else
+      # use wget
+      wget -qO- ${AGENT_BINARIES_URI} | tar xzf - -C ${CHE_DIR}/ws-agent
```
garagatyi: Wget downloading section for terminal agent is more complex. Can you elaborate on why they differ and why we can't make them pretty the same?

riuvshin (approved):

wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
benoitf: I don't know, I follow the same approach than curl/terminal agent section which is already complex ^^ https://github.com/eclipse/che/blob/master/agents/exec/src/main/resources/org.eclipse.che.terminal.script.sh#L162-L166

It's due to different sed expressions
sed 's/\${PREFIX}/'${PREFIX}'/g'); vs sed 's/-\${PREFIX}//g'); 

like if we need to support different naming scheme for the terminal agent file but I don't know the reason

benoitf:
I need to make a change to replace https by http if https is used

wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
garagatyi: I don't mean PREFIX stuff. I mean that terminal script uses spider mode in wget and check if current wget doesn't have certificates stuff

benoitf:
ok, PR updated

agents/exec/src/main/resources/org.eclipse.che.terminal.script.sh
benoitf: updated

agents/exec/src/main/resources/org.eclipse.che.terminal.script.sh
benoitf: updated

wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
benoitf: updated

garagatyi (approved):

skabashnyuk (approved):

benoitf:
@JamesDrummond it's ok for release notes ?

garagatyi:
This should improve UX in certain situations a lot!

PR #4049, opened 2017-02-07
OpenShiftConnector OAuth token support

l0rd:
### What does this PR do?
Add OpenShiftConnector OAuth token support. OpenShift connector authentication can be done via an OAuth token. 

N/A

#### Changelog
Added OpenShiftConnector OAuth token support.

#### Release Notes
N/A (internal change)

#### Docs PR
N/A (will be documented as part of the larger OpenShift documentation)

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
@@ -272,6 +272,7 @@ db.jndi.datasource.name=java:/comp/env/jdbc/che
 
 # OpenShift related properties
 che.openshift.endpoint=https://192.168.64.2:8443/
+che.openshift.token=
 che.openshift.username=openshift-dev
 che.openshift.password=devel
 che.openshift.project=eclipse-che
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
@@ -131,11 +131,13 @@ public class OpenShiftConnector extends DockerConnector {
     private final int             openShiftLivenessProbeTimeout;
 
     @Inject
-    public OpenShiftConnector(DockerConnectorConfiguration connectorConfiguration,
+    public OpenShiftConnector(ConfigBuilder configBuilder,
+                              DockerConnectorConfiguration connectorConfiguration,
                               DockerConnectionFactory connectionFactory,
                               DockerRegistryAuthResolver authResolver,
                               DockerApiVersionPathPrefixProvider dockerApiVersionPathPrefixProvider,
                               @Named("che.openshift.endpoint") String openShiftApiEndpoint,
+                              @Named("che.openshift.token") String openShiftToken,
                               @Named("che.openshift.username") String openShiftUserName,
                               @Named("che.openshift.password") String openShiftUserPassword,
                               @Named("che.openshift.project") String openShiftCheProjectName,
@@ -149,9 +151,16 @@ public class OpenShiftConnector extends DockerConnector {
         this.openShiftLivenessProbeDelay = openShiftLivenessProbeDelay;
         this.openShiftLivenessProbeTimeout = openShiftLivenessProbeTimeout;
 
-        Config config = new ConfigBuilder().withMasterUrl(openShiftApiEndpoint)
-                .withUsername(openShiftUserName)
-                .withPassword(openShiftUserPassword).build();
+        Config config;
+        if (StringUtils.isNotBlank(openShiftToken)) {
+            config = configBuilder.withMasterUrl(openShiftApiEndpoint)
+                    .withOauthToken(openShiftToken)
+                    .build();
+        } else {
+            config = configBuilder.withMasterUrl(openShiftApiEndpoint)
+                    .withUsername(openShiftUserName)
+                    .withPassword(openShiftUserPassword).build();
+        }
         this.openShiftClient = new DefaultOpenShiftClient(config);
     }
 
@@ -265,7 +274,6 @@ public class OpenShiftConnector extends DockerConnector {
     }
 
     /**
-     * @param docker
      * @param container
      * @return
      * @throws IOException
@@ -982,7 +990,6 @@ public class OpenShiftConnector extends DockerConnector {
      * When container is expected to be run as root, user field from {@link ImageConfig} is empty.
      * For non-root user it contains "user" value
      *
-     * @param dockerConnector
      * @param imageName
      * @return true if user property from Image config is empty string, false otherwise
      * @throws IOException
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
@@ -29,6 +29,7 @@ import javax.inject.Named;
 import javax.inject.Singleton;
 
 import org.apache.commons.lang.RandomStringUtils;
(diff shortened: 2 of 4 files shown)


amisevsk:
LGTM. This will be very helpful for deployments.

ibuziuk:
+1  Very important piece for OpenShift integration. I would just probably extract config generation logic to separate method(s)

skabashnyuk:
```che.openshift.token ``` what is it?

File: plugins/plugin-docker/che-plugin-openshift-client/src/test/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnectorTest.java [hunk 1]
```diff
@@ -43,6 +46,7 @@
     private static final String   OPENSHIFT_DEFAULT_USER_PASSWORD = "devel";
     private static final int      OPENSHIFT_LIVENESS_PROBE_DELAY = 300;
     private static final int      OPENSHIFT_LIVENESS_PROBE_TIMEOUT = 1;
+    private static final String   OPENSHIFT_DEFAULT_TOKEN = "91XMfu-FuNDkGjcIh6b0y1EtCvztGeSsSqRrWhBfyL8";
```
skabashnyuk, line 44: are you sure it's secure to do put it here?

l0rd: @skabashnyuk that's a test class and that's a sample token. What is exactly your security concern?

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -272,6 +272,7 @@ db.jndi.datasource.name=java:/comp/env/jdbc/che
 
 # OpenShift related properties
 che.openshift.endpoint=https://192.168.64.2:8443/
+che.openshift.token=
```
benoitf, line 275: should we use `che.openshift.token=NULL`

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -135,6 +136,7 @@ public OpenShiftConnector(DockerConnectorConfiguration connectorConfiguration,
                               KubernetesService kubernetesService,
                               DockerApiVersionPathPrefixProvider dockerApiVersionPathPrefixProvider,
                               @Named("che.openshift.endpoint") String openShiftApiEndpoint,
+                              @Named("che.openshift.token") String openShiftToken,
```
benoitf: @Nullable ?

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -152,9 +154,16 @@ public OpenShiftConnector(DockerConnectorConfiguration connectorConfiguration,
         this.openShiftLivenessProbeDelay = openShiftLivenessProbeDelay;
         this.openShiftLivenessProbeTimeout = openShiftLivenessProbeTimeout;
 
-        Config config = new ConfigBuilder().withMasterUrl(openShiftApiEndpoint)
-                .withUsername(openShiftUserName)
-                .withPassword(openShiftUserPassword).build();
+        Config config;
+        if (StringUtils.isNotBlank(openShiftToken)) {
```
benoitf: usually che code is using https://google.github.io/guava/releases/19.0/api/docs/com/google/common/base/Strings.html#isNullOrEmpty(java.lang.String)

benoitf (approved):

plugins/plugin-docker/che-plugin-openshift-client/src/test/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnectorTest.java
skabashnyuk: ok if it's only for test

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
l0rd: how is that better?

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
l0rd: yep will fix it thanks

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
l0rd: You are right, I didn't want to import another package in this class so I used the available one. But maybe we should use guava when possible instead.

l0rd:
@ibuziuk thanks that's a good idea to extract the config in another method. I will do that.

l0rd:
@ibuziuk, @benoitf I've updated the PR. Can you please have a look?

ibuziuk:
+1 to apply

benoitf (approved):

bmicklea:
For release notes and docs - is this adding an OpenShift oAuth login option for user accounts like we have today for Google, GitHub and MSFT?  Or is this adding oAuth to OpenShift for use in the workspaces?

l0rd:
@bmicklea this is for the built-in OAuth server of OpenShift (it's not about google, github etc...).

bmicklea (commented):
Thanks for the clarification Mario - I've updated the Changelog, release notes and docs section of your description with my suggestions. If you agree with those let me know and I can approve. If not we can work on tweaking them.

l0rd:
@bmicklea Release notes and docs are ok for me. Thanks for updating it.

bmicklea (approved):

benoitf:
This PR requires a rebase as there is a conflict

l0rd:
@benoitf thanks. I'm working on that. I will rebase it (rebuild and test it) and merge it.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -272,6 +272,7 @@ db.jndi.datasource.name=java:/comp/env/jdbc/che
...
 che.openshift.username=openshift-dev
 che.openshift.password=devel
 che.openshift.project=eclipse-che
```
aslakknutsen, line 278: Are these supported as ENV variables?

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -272,6 +272,7 @@ db.jndi.datasource.name=java:/comp/env/jdbc/che
 
 # OpenShift related properties
 che.openshift.endpoint=https://192.168.64.2:8443/
+che.openshift.token=
 che.openshift.username=openshift-dev
 che.openshift.password=devel
```
aslakknutsen: This would be per user call, not per che starter instance?

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
aslakknutsen: ENV variables or directly as OpenShift Secrets.

benoitf: properties that star by che.* can be overrided by ENV variables

che.openshift.token --> CHE_OPENSHIFT_TOKEN=""

PR #4050, opened 2017-02-07
CLI - set uid:gid of Che and workspaces

TylerJewell:

### What does this PR do?
This PR adapts the Che CLI to support letting users provide a different uid:gid combination to be applied to the Che server and its resulting workspaces. Users can pass `--user uid:gid` on the `docker run ...` command line and this will be passed into the eclipse/che-server container.  If no user is provided, the default is set to root.

This feature should only be used with Linux hosts and will have odd behaviors on Windows.

Only merge this PR when the user improvements of @snjeza are merged.
https://github.com/eclipse/che/pull/3376

#### Changelog
Add `--user uid:gid` and `-e CHE_USER=uid:gid` option to CLI to override user identity of Che container

#### Release Notes
You can now run Eclipse Che's container with a different user identity on Linux or Mac. The default is to run the Che container as a root user. You can now pass `--user uid:gid` or `-e CHE_USER=uid:gid` on the command line as Docker options to the `eclipse/che` Docker image. This image will launch the `eclipse/che-server` image with the same `uid:gid` combination along with mounting `/etc/group` and `etc/passwd`. When Che is run as a user, all files written from within the Che server to the host (such as `che.env` or `cli.log` will be written to disk with the custom user as the owner of the files. Changing the user is not  available on Windows.

#### Docs PR
N/A - covered by the main PR


snjeza:
@TylerJewell 

You should mount /etc/passwd and /etc/group.
A user must have the write permission to the data directory.
We don't have to apply https://github.com/eclipse/che/pull/3376 to test this PR. It works with eclipse/che-server:nightly.

TylerJewell:
@snjeza - fixed by adding those additional mounts to the compose file if you are changing the user.

But help me a little bit - why isn't #3376 needed here?  I thought #3376 added internal modifications to the che server necessary to support running the server under a different user?

snjeza:
@TylerJewell 
>But help me a little bit - why isn't #3376 needed here? I thought #3376 added internal modifications to the che server necessary to support running the server under a different user?

#3376 is related to running a workspace as an user.  #3265 is related to running the che server as an user.

TylerJewell:
Yes but the expectation we discussed with Mario was that if you set the Che server to run under a single user we would enforce that with workspaces to have the same as well.  So we are missing that coupling.

TylerJewell:
@l0rd @snjeza - once we address the last item on running a workspace as a user, then I think this is ready for merge.

snjeza:
> Yes but the expectation we discussed with Mario was that if you set the Che server to run under a single user we would enforce that with workspaces to have the same as well. So we are missing that coupling.

You are right, but this PR can be tested/applied without #3376. If we use --user <XXX>, the che server should start as <XXX> user. When we apply #3376, a workspace should be started as <XXX> user.

I have tried to start che using your PR.
It works when I add the following patch:
```
diff --git a/dockerfiles/base/scripts/base/startup_02_pre_docker.sh b/dockerfiles/base/scripts/base/startup_02_pre_docker.sh
index 290dfb9..ee05557 100644
--- a/dockerfiles/base/scripts/base/startup_02_pre_docker.sh
+++ b/dockerfiles/base/scripts/base/startup_02_pre_docker.sh
@@ -404,7 +404,7 @@ check_interactive() {
 check_user() {
   CHE_USER=""
   CHE_USER=$(docker inspect --format='{{.Config.User}}' $(get_this_container_id))
-  if custom_user; then
+  if ! custom_user; then
     # Add checks / warnings if a custom user is set
     CHE_USER="root"
   fi
diff --git a/dockerfiles/che/entrypoint.sh b/dockerfiles/che/entrypoint.sh
index b2ca8ed..43441a6 100755
--- a/dockerfiles/che/entrypoint.sh
+++ b/dockerfiles/che/entrypoint.sh
@@ -237,9 +237,10 @@ init() {
       echo "!!!"
       exit 1
     fi
-    export CHE_USER_ID=`id -u ${CHE_USER}`:`getent group docker | cut -d: -f3`
-    sudo chown -R ${CHE_USER}:docker ${CHE_DATA}
-    sudo chown -R ${CHE_USER}:docker ${CHE_HOME}
+    export CHE_USER_ID=${CHE_USER}
+    sudo chown -R ${CHE_USER} ${CHE_DATA}
+    sudo chown -R ${CHE_USER} ${CHE_HOME}
+    sudo chown -R ${CHE_USER} ${CHE_LOGS_DIR}
   fi
   ### Are we going to use the embedded che.properties or one provided by user?`
   ### CHE_LOCAL_CONF_DIR is internal Che variable that sets where to load
@@ -247,7 +248,7 @@ init() {
     echo "Found custom che.properties..."
     export CHE_LOCAL_CONF_DIR="/conf"
     if [ "$CHE_USER" != "root" ]; then
-      sudo chown -R ${CHE_USER}:docker ${CHE_LOCAL_CONF_DIR}
+      sudo chown -R ${CHE_USER} ${CHE_LOCAL_CONF_DIR}
     fi
   else
     echo "Using embedded che.properties... Copying template to ${CHE_DATA_HOST}/conf."
```
Call the following commands to run the che server as an user:

```
$ git fetch origin pull/4050/head:che-4050
$ git checkout che-4050
# apply the patch
$ cd dockerfiles/init
$ ./build.sh
$ cd ../base
$ ./build.sh
$ cd ../cli
$ ./build.sh 
$ cd ../che
$ ./build.sh
$ grep test /etc/passwd
test:x:1001:1002::/home/test:/bin/bash
$ grep docker /etc/group
docker:x:1001:test,snjeza
$ sudo chown -R test:docker /tmp/data
$ docker run -it --rm \
  -v /tmp/data:/data \
  -v /etc/group:/etc/group:ro \
  --user 1001:1001 \
    eclipse/che-cli:nightly config --skip:nightly --skip:pull
$ mkdir ~/test
$ cd ~/test
$ cp /tmp/data/instance/docker-compose.yml .
$ docker-compose up
...
$ $ docker exec che id
uid=1001(test) gid=1001(docker)
```

Che-cli creates the following docker-compose.yml:
```
# ###################################
# This file is generated by puppet
# PLEASE DON'T MODIFY BY HAND
# ###################################

che:
  image: eclipse/che-server:nightly
  env_file:
    - '/tmp/data/instance/config/che.env'
  volumes:
    - '/var/run/docker.sock:/var/run/docker.sock'
    - '/tmp/data/instance/data:/data'
    - '/tmp/data/instance/logs:/logs'
    - '/tmp/data/instance/config:/conf'
    - '/etc/group:/etc/group:ro'
  ports:
    - '8080:8080'
  restart: always
  container_name: che
  user: 1001:1001
```
and adds the following env variable to che.env
```
CHE_USER=1001:1001
```

We have to mount /etc/passwd and /etc/group when running che-cli. Otherwise, che-cli will return an error.
A user that runs the che server, has to have the write permission to the /data directory (/tmp/data in my example).

TylerJewell:
Thanks for the pointers @snjeza.  I did some refactoring to make the code more manageable long term.  It was ok on my windows machine.  I also provided the ability to set the uid:gid through either use of `--user` or with `-e CHE_USER=` syntax.  Why don't you give this a shot and if it's good we'll merge it at the beginning of the next release cycle after 5.3 is released tomorrow or Thursday.

snjeza:
@TylerJewell 

It works fine.

TylerJewell:
@l0rd - what is the merge plan here?  At a minimum we should wait for 5.3 to release tomorrow.  Will we also co-release this with the changes that allow for setting user identity within a workspace, too or should we merge this separately?

PR #4056, opened 2017-02-08
Fix locale in CentOS based image

l0rd:
When using CentOS based image a lot of `UnmappableCharacterException` are logged in che-server output when creating a new workspace. This is related to this issue in CentOS base image: https://github.com/CentOS/sig-cloud-instance-images/issues/71  

#### Changelog
Fix bad locale in CentOS based image

#### Release Notes
Improved CentOS based Che docker image

#### Docs PR
That's a fix. No docs needed

### DIFF
--- a/dockerfiles/che/Dockerfile.centos
+++ b/dockerfiles/che/Dockerfile.centos
@@ -8,17 +8,16 @@
 #   Dharmit Shah  - Initial implementation
 #   Mario Loriedo - Improvements
 #
-# To build it, run in the current folder after executing build.sh:
-#  `docker build -t registry.centos.org/eclipse/che-server -f Dockerfile.centos .`
+# To build it, in current folder:
+#  cp Dockerfile.centos Dockerfile
+#  ./build.sh
 #
 # To run it:
-#  docker run --net=host \
-#             --name che \
-#             -v /var/run/docker.sock:/var/run/docker.sock \
-#             -v /home/user/che/lib:/home/user/che/lib-copy \
-#             -v /home/user/che/workspaces:/home/user/che/workspaces \
-#             -v /home/user/che/storage:/home/user/che/storage \
-#             registry.centos.org/eclipse/che-server
+#  docker run -d -p 8080:8080 \
+#            --name che \
+#            -v /var/run/docker.sock:/var/run/docker.sock \
+#            -v ~/.che/workspaces:/data \
+#            eclipse/che-server:nightly
 #           
 FROM registry.centos.org/centos/centos:latest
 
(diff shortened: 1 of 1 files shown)


benoitf (approved):

dockerfiles/che/Dockerfile.centos
```diff
@@ -8,17 +8,16 @@
...
-#             -v /var/run/docker.sock:/var/run/docker.sock \
-#             -v /home/user/che/lib:/home/user/che/lib-copy \
-#             -v /home/user/che/workspaces:/home/user/che/workspaces \
-#             -v /home/user/che/storage:/home/user/che/storage \
-#             registry.centos.org/eclipse/che-server
+#  docker run -d -p 8080:8080 \
+#            --name che \
+#            -v /var/run/docker.sock:/var/run/docker.sock \
+#            -v ~/.che/workspaces:/data \
+#            eclipse/che-server:nightly
```
JamesDrummond, line 20: Not sure we want to require nightly. `eclipse/che-server:nightly` => `eclipse/che-server`. Should change `-v ~/.che/workspaces:/data` => `-v <LOCAL_PATH>:/data`.

JamesDrummond (changes requested):

dockerfiles/che/Dockerfile.centos
benoitf: about the first point if you use build script by default it will produce nightly so if you need to run this newly image you'll have to use that version

JamesDrummond: Still not sure if nightly should be used. Who will use the dockerfile? Seems better to allow user to choose version. Seems like it possible but I have not investigated https://github.com/eclipse/che/blob/master/dockerfiles/build.include#L33 .

l0rd: @JamesDrummond I like `~/.che/` more because I can just copy paste and it works. I don't need an extra step of editing the command. But I understand you want to make it clear that the path is configurable. I just wanted to be sure that you understand why I prefer `~/.che/` (and potentially other users would prefer it too). If you still think that setting `<LOCAL_PATH>` is a better choice I will change that. 

What is the issue with nightly? I understand you want to change that but I don't understand why. I'm asking because in the past I've fixed an issue with **non** nightly tags and I ran into another issue recently. So when I build it I usually use the default tag (nightly) to improve my chance to build/run successfully.

JamesDrummond: This is why I ask who are the users going to be for the dockerfile. ~/.che will not work on a windows machine. Using nightly label is fine but I don't think it should be required. What if you want to produce a stable version of the centos to host on a repo such as dockerhub. Again I am not sure of the use case of this dockerfile. If you would like to keep ~/.che and nightly in place because that is what the target use will be I can approve. Just let me know.

l0rd: James this Dockerfile is an alternative to the che-server alpine based Dockerfile. The image should be used by individuals or organizations that prefer running Che on CentOS then alpine.

Instructions in the Dockerfiles should be for developers that know what they are doing and want hack the Dockerfile (to fix a bug in the Dockerfile for example). A user that just want to run the image should pull the pre-built images available from the Dockerhub and in registry.centos.org (and run it using the CLI).

JamesDrummond: Great thanks.

JamesDrummond (approved):

Issue #3988, opened 2017-02-01
graceful-stop shouldn't fail if system is not in RUNNING state


voievodin:
Current approach is to fail if system is not RUNNING.
But as we expose public API for preparing system to shutdown it makes sense to analyze 
current system state and wait for READY_TO_SHUTDOWN instead of failing immediately.
The example of output i got when after i played with API and called `cli stop`.
```
Status: Image is up to date for eclipse/che:nightly

INFO: Waiting for graceful stop of services...
ERROR: (che action/graceful-stop): Status is not in RUNNING state so the graceful can't be called.
```

Currently the system can't go back to running state, so if system is in the state different from running all that cli needs to do is to wait for it to become READY_TO_SHUTDOWN. 
Proposed action map for cli:
```
RUNNING -> call stop and wait for system to be in READY_TO_SHUTDOWN state
READY_TO_SHUTDOWN -> do nothing(maybe print something useful)
ANY_OTHER_STATE -> wait for system to be in READY_TO_SHUTDOWN state
```

PR #4063, opened 2017-02-08
[cli] graceful stop should work with any system state

benoitf:
### What does this PR do?
cli graceful stop should work with any system state
until now it was expecting remote state was RUNNING.

Now if state is:
  - RUNNING -> call stop and wait for system to be in READY_TO_SHUTDOWN state
  - READY_TO_SHUTDOWN -> do nothing(maybe print something useful)
  - ANY_OTHER_STATE -> wait for system to be in READY_TO_SHUTDOWN state

#3988 

#### Changelog
[cli] graceful stop work with any remote system state

#### Release Notes
N/A

#### Docs PR
N/A

Change-Id: I55a8b5ecb2b016d31dbbd95b8793af4952d8116b

### DIFF
--- a/dockerfiles/lib/src/api/wsmaster/system/system.ts
+++ b/dockerfiles/lib/src/api/wsmaster/system/system.ts
@@ -64,7 +64,6 @@ export class System {
                 websocketLink = stateLink.getHref();
             }
         });
-
         return this.websocket.getMessageBus(websocketLink + '?token=' + this.authData.getToken());
     }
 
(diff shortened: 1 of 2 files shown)


JamesDrummond (approved):
Approval of changelog and release notes.

TylerJewell:
@JamesDrummond - on anything related to CLI, you can skip over.  I'll be responsible for changelog and release notes.  

@benoitf - do you have some various sample outputs that you can share?  I think we need to update the release notes to be more explanatory.  They need full paragraphs and maybe an example or two.

benoitf:
@TylerJewell well the output is the same in all cases. It's just that now we handle cases where remote system is not in running state. But for 99% of ppl there is also no change as it's the default case. Other cases only happen if rest API is called by something else than cli

TylerJewell (approved):

voievodin (approved):

PR #3954, opened 2017-01-30
CHE-3953 - Starting workspace on openshift 1.4.1

snjeza:
Signed-off-by: Snjezana Peco <snjezana.peco@redhat.com>

### What does this PR do?

Fixes #3953 

### What issues does this PR fix or reference?

#3953 

#### Changelog
Bug fix for Starting workspace on openshift 1.4.1 `servicePort.setProtocol(protocol.toUpperCase());`

PR #4066, opened 2017-02-08
Fix test : make protocol uppercase

vparfonov:
Fix test after https://github.com/eclipse/che/pull/3954/files 
Signed-off-by: Vitalii Parfonov <vparfonov@codenvy.com>


vparfonov:
fixed here : https://github.com/eclipse/che/pull/4065

PR #4083, opened 2017-02-09
[cli] Fix CLI loading behaviorCli loading 

TylerJewell:
### What does this PR do?
Fixes the root cause of this issue. https://github.com/codenvy/codenvy/issues/1736.  In situations where scripts were recursively calling `cmd_lifecycle`, old variables were overriding other variables.  The simple fix of setting the variables in `cmd_lifecycle` to `local` avoided this overwriting problem.

https://github.com/codenvy/codenvy/issues/1736

#### Changelog
[cli] Prevent overwriting variables with recursive `cmd_lifecycle` statement
[cli] Added `--skip:scripts` hidden flag which will skip loading scripts from repo, if :/repo mounted

#### Release Notes
N/A

#### Docs PR
N/A


benoitf (approved):

dockerfiles/base/scripts/base/commands/cmd_config.sh
```diff
@@ -29,11 +29,14 @@ images will use those binaries instead of their embedded ones.\n"
 }
 
 pre_cmd_config() {
-  true
+  :
 }
 
-cmd_config() {
+post_cmd_config() {
+  :
```
benoitf, line 36: could we avoid to define empty methods and check if they're defined ?

benoitf (approved):

dockerfiles/base/scripts/base/commands/cmd_config.sh
TylerJewell: It works both ways - you don't have to have empty methods, but I found it cleaner to have empty methods for all of the base just to be pedantic.

You don't need these in any CLI assemblies built from base.

dockerfiles/base/scripts/base/startup_05_pre_exec.sh
```diff
@@ -32,13 +32,14 @@ cli_parse () {
 
 cli_execute() {
   cmd_lifecycle "$@"
+    echo "!!!!!!!!!!!!!!!!!!!!!!!"
```
riuvshin: do we need this echo? looks like debug thing

dockerfiles/base/scripts/base/commands/cmd_config.sh
TylerJewell: Ooops... fixed

riuvshin (approved):

PR #4085, opened 2017-02-09
Refactor inspectContainer() and add commit(), removeImage()

amisevsk:
Adds `commit()` and `removeImage()` implementations to `OpenShiftConnector`. This requires some refactoring of existing methods (mostly refactoring repeated processes into their own methods -- e.g. creating an `ImageStreamTag` and getting image info from a tag).

Additionally, refactors `inspectContainer()` method to remove a call to `DockerConnector`, instead obtaining the same information from what's available through the OpenShift API. This fixes an issue where the IP address of a workspace was unavailable from the `ContainerInfo` returned by `DockerConnector`.

The benefit of these changes is that -- except for exec-related methods -- most uses of `DockerConnector` are removed when running on OpenShift. Note however that custom stacks are still not supported.

### What does this PR do?
Refactors `inspectContainer()` to use ImageStreamTags instead of Docker, and adds `commit()` and `removeImage()` methods. 

#### Changelog
Improved OpenShift connector independence from DockerConnector

#### Release Notes
(Internal changes -- not ready for public release)

#### Docs PR
Issue tracking work on OpenShift integration docs: https://github.com/eclipse/che-docs/issues/76.

### DIFF
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
@@ -28,7 +28,6 @@ import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
 
-import org.apache.commons.lang.RandomStringUtils;
 import org.eclipse.che.commons.annotation.Nullable;
 import org.eclipse.che.plugin.docker.client.DockerApiVersionPathPrefixProvider;
 import org.eclipse.che.plugin.docker.client.DockerConnector;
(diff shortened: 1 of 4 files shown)


plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -505,86 +499,112 @@ public void pull(final PullParams params, final ProgressMonitor progressMonitor)
...
      */
     @Override
     public void tag(final TagParams params) throws ImageNotFoundException, IOException {
-        // E.g. `docker tag oldImageRepo newImageRepo`
+        // E.g. `docker tag sourceImage targetImage`
         String sourceImage = params.getImage();        // e.g. eclipse/ubuntu_jdk8
         String targetImage = params.getRepository();   // e.g. eclipse-che/<identifier>
         String tag = params.getTag();
-        if (tag == null || tag.isEmpty()) {
+        if (sourceImage.contains(":")) {
```
l0rd: You split images string into tag and image name multiple times in the code. To avoid duplication you can probably extract that logic in a new method.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -505,86 +499,112 @@ public void pull(final PullParams params, final ProgressMonitor progressMonitor)
...
-            Thread.currentThread().interrupt();
-            throw new IOException(e.getLocalizedMessage(), e);
-        }
-    }
 
     @Override
     public ImageInfo inspectImage(InspectImageParams params) throws IOException {
 
         String image = params.getImage();
+        if (image.contains(":")) { // Image to inspect passed in with tag, which we don't support
```
l0rd: This code is duplicated in removeImage too. You may want to extract it to a new method

l0rd:
@amisevsk you should rebase your PR

l0rd:
@amisevsk that's not related to this PR but you could have a look at that: in method  `getImageStreamTagFromRepo()` the size to truncate imageStreamTagName is 30 but shouldn't it be 63 (aka MAX_CHARS)?

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -266,24 +274,25 @@ public void putResource(PutResourceParams params) throws IOException {
...
 
-        Pod pod = getChePodByContainerId(info.getId());
+        Pod pod = getChePodByContainerId(containerId);
         if (pod == null ) {
-            LOG.warn("No Pod found by container ID {}", info.getId());
+            LOG.warn("No Pod found by container ID {}", containerId);
             return null;
         }
 
+        String podPullSpec = pod.getSpec().getContainers().get(0).getImage();
```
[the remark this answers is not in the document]
amisevsk: Fixed

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
amisevsk: Fixed, there's a few methods in `KubernetesStringUtils` that can do most of the work now.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
amisevsk: Fixed, this was an oversight by me -- the method already existed.

amisevsk:
I've rebased the PR and added some test cases.

@l0rd For the `getImageStreamTagFromRepo()` issue, the tag name has to be truncated since, as far as I can tell, OpenShift limits the length of the ImageStream *name + tag* to 63, not the tag by itself. This means that if the original repo name is long, the entire length of the ImageStream tag won't fit in that field.

Another option I considered (instead of truncating) was to do a `startsWith()` comparison, but I feel like this might be less readable, since we'd have to manipulate the `ImageStreamTag` name to get just the tag.

l0rd (approved):

PR #4087, opened 2017-02-10
use official openjdk image

riuvshin:
### What does this PR do?
use official openjdk image which will allow us to fix jdk version and avoid unexpected updates / upgrades

#### Changelog
Use the official OpenJDK image.

### DIFF
--- a/dockerfiles/che/Dockerfile
+++ b/dockerfiles/che/Dockerfile
@@ -20,18 +20,15 @@
 #             -v /home/user/che/storage:/home/user/che/storage \
 #             codenvy/che-server
 #           
-FROM alpine:3.4
+FROM openjdk:8u111-jdk-alpine
 
 ENV LANG=C.UTF-8 \
-    JAVA_HOME=/usr/lib/jvm/default-jvm/jre \
-    PATH=${PATH}:${JAVA_HOME}/bin \
     DOCKER_VERSION=1.6.0 \
     DOCKER_BUCKET=get.docker.com \
     CHE_IN_CONTAINER=true
 
 RUN echo "http://dl-4.alpinelinux.org/alpine/edge/community" >> /etc/apk/repositories && \
-    apk upgrade --update && \
-    apk add --update ca-certificates curl openssl openjdk8 sudo bash && \
+    apk add --update curl openssl sudo bash && \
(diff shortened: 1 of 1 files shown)


benoitf (dismissed):

benoitf:
@riuvshin do we need jdk and not the jre ?

riuvshin:
@benoitf we discussed that and we agreed to use jdk this may help us in some cases

benoitf:
@riuvshin what are these cases ?

riuvshin:
@benoitf main point was that we didn't tested running on jre that much as we always used jdk,
do you think jre is enough ? and we don't need jdk to run che or codenvy? maybe some java opts that we pass or debug scenarios may require jdk?

benoitf:
@riuvshin I still don't see what should these java opts and debug scenario.
The user doesn't connect to che master from IDE terminal. And the tools can be as well installed with apk at runtime if required for a one time debugging.
but using JRE saves extra  10MB

riuvshin:
yes we know that switch to jre will save few mb for us but for now the problem that there is no one who can say that we don;t need jdk here we just don;t know is this required or not, so for now we want change only official image

riuvshin:
@benoitf @skabashnyuk updated to use `8u111-jre-alpine` tag

dockerfiles/che/Dockerfile
```diff
@@ -20,18 +20,15 @@
...
-    JAVA_HOME=/usr/lib/jvm/default-jvm/jre \
-    PATH=${PATH}:${JAVA_HOME}/bin \
     DOCKER_VERSION=1.6.0 \
     DOCKER_BUCKET=get.docker.com \
     CHE_IN_CONTAINER=true
 
 RUN echo "http://dl-4.alpinelinux.org/alpine/edge/community" >> /etc/apk/repositories && \
-    apk upgrade --update && \
-    apk add --update ca-certificates curl openssl openjdk8 sudo bash && \
+    apk add --update curl openssl sudo bash && \
```
benoitf, line 31: maybe we could drop as well curl and openssl after docker is downloaded (but it's not related to this PR)

benoitf (approved):

dockerfiles/che/Dockerfile
riuvshin: im not sure why we added this here so it require another discussion

skabashnyuk (approved):

bmicklea (approved):

PR #4119, opened 2017-02-13
Inject MAVEN_OPTS environment variable along with JAVA_OPTS to dev-machine

### LINKED ISSUE #4108: Maven server should be able to use JAVA_OPTS as MAVEN_OPTS
Maven server should be able to do that otherwise resolving deps will fails behind the proxy.

voievodin:
### What does this PR do?
Adds `MAVEN_OPTS` to the list of environment variables injected into container.
Environment variable value is equal to the value provided by `che.workspace.maven.options` property, if the value of property is not set then `che.workspace.java_opts` value is used for maven opts. 

Test case:
- start che 
- create and start a command `echo $MAVEN_OPTS`
- output is `-Xms256m -Xmx2048m -Djava.security.egd=file:/dev/./urandom`

Fixes #4108 

#### Changelog
Inject `MAVEN_OPTS` environment variable into dev-machine

#### Release Notes
`MAVEN_OPTS` environment variable is injected into dev machine along with `JAVA_OPTS`.
Property value can be changed within `che.workspace.maven.options` configuration property. 
Default value is set to the value provided by `che.workspace.java_opts` configuration property.

#### Docs PR
N/A


benoitf:
IMHO it's unclear to me why a plugin property needs to be within the `che.workspace.*` namespace with `che.workspace.maven_opts` name. 
I can use che without Maven. It only target maven projects (with maven plugin) so I think it should match the name of the plugin. (I have not big idea but maybe `che.plugin.java.maven_opts` or something like that
BTW properties name are also reviewed by @TylerJewell 


I also don't see why this registration is performed in src/main/java/org/eclipse/che/plugin/docker/machine/ext/DockerExtServerModule.java
Why is it not done inside Maven plugin code ?

I think it could be a nightmare if each project type plugin register his property on `che.workspace.*` ?

voievodin:
agree, going to rework it

voievodin:
@benoitf updated

benoitf (approved):

garagatyi (approved):

vparfonov (approved):

ghost (approved):

skabashnyuk (approved):

voievodin:
@TylerJewell are you okay with `che.plugin.maven.maven_opts` property name?

TylerJewell:
Not really.  The name is too long and unclear.  I need time to think of better solution.  It is also not clear when a user needs to set this. But if you are a user going in and changing a plugin property is dangerous.  So I am -1 on the naming.

voievodin:
I can remove property and use `MAVEN_OPTS=JAVA_OPTS` which solves the issue.
Is it okay for you?

TylerJewell:
No - that feels like window dressing.

I am unclear of the objective and when would a user need to configure thisnfor Che or codenvy?   The property name should be defined by what we need the end user to think about.

ghost:
A normal usecase implies that a users does not touch it, and Maven inherits options from Java. What exactly isn't clear in the name?

TylerJewell:
What are the situations where a user would need to change this value for Che or codenvy?  

It is unclear whether this is for admins to configure a user developer or if it is intended as an internal plugin property only for customizers of plugins.

ghost:
@TylerJewell it's for admins - who else has access to a Codenvy installation.

In case of Che, users may edit it if they need any additional Maven options.

voievodin:
@TylerJewell i changed the property name, also defined it in _che.env_ along with _JAVA_OPTS_ please check

dockerfiles/init/manifests/che.env
```diff
@@ -101,7 +101,16 @@
 #     agents within the workspace when it boots, then the workspace will not be started.
 #CHE_MACHINE_WS_AGENT_MAX_START_TIME_MS=300000
 
-
+# Workspace Agent Java Options
+#   Java command line options will be exported as a value of the JAVA_OPTS
+#   environment variable in the 'development' machine.
+#CHE_WORKSPACE_JAVA__OPTS=-Xms256m -Xmx2048m -Djava.security.egd=file:/dev/./urandom
```
TylerJewell: We have a problem now in that the Java and Maven options have different formats.
CHE_WORKSPACE_MAVEN_OPTIONS
CHE_WORKSPACE_JAVA__OPTS

Admins will expect common syntax, so propose:
CHE_WORKSPACE_MAVEN_OPTIONS
CHE_WORKSPACE_JAVA_OPTIONS

dockerfiles/init/manifests/che.env
```diff
@@ -101,7 +101,16 @@
 #     agents within the workspace when it boots, then the workspace will not be started.
 #CHE_MACHINE_WS_AGENT_MAX_START_TIME_MS=300000
 
-
+# Workspace Agent Java Options
+#   Java command line options will be exported as a value of the JAVA_OPTS
```
TylerJewell: Please use this comment.  I think we should avoid the idea of "export" because to an end user, an "export" is something that comes *out* of the workspace.  When I think we are really talking about taking parameters from one location and injecting them into another location.

```
Java command line options to be added to JVM's that run agents within workspaces.
```

dockerfiles/init/manifests/che.env
```diff
@@ -101,7 +101,16 @@
...
+#   environment variable in the 'development' machine.
+#CHE_WORKSPACE_JAVA__OPTS=-Xms256m -Xmx2048m -Djava.security.egd=file:/dev/./urandom
+
+# Workspace Agent Maven Options
+#   Maven command line options will be exported as a value of the MAVEN_OPTS
```
TylerJewell:
```
Maven command line options added to JVM's that run agents within workspaces. Defaults to the same value of CHE_WORKSPACE_JAVA_OPTIONS.
```

TylerJewell (changes requested):

TylerJewell:
Thank you @evoevodin and @eivantsov for explaining the use cases.  I think it is clear.  I have some recommended changes that we should make for Che first, and then propagate those changes for Codenvy second.  This is really close.

dockerfiles/init/manifests/che.env
benoitf: I still don't get why MAVEN is a parameter of the workspace
it's only used if there is a maven project while AFAIK JAVA_OPTS is used by the ws agent

TylerJewell:
According to @evoevodin - the maven part is part of every workspace.

But even so, I am generally -1 on having parameters in a che.env that customize how a single plugin would work.  It will be very confusing to explain to end users and admins when they should configure certain plugins in a Che-server configuration file.  So I think of the naming of these parameters as workspace-specific parameters, not plugin specific.

benoitf:
I'm not sure it's "Every workspace" but more "workspace that has maven projects (and that has maven tool installed on its docker container")
@TylerJewell I'm not sure that someone that have nodejs will see what maven_opts is for.

TylerJewell:
We have maven tools installed everywhere our ws agent runs according to dev.  This makes it more of a ws concept for me right now

benoitf:
Started a `ubuntu/go` workspace
Open a terminal:
```
user@8be3a9bfa3c9:/projects$ mvn                                                                                                                                                                                                  
bash: mvn: command not found               
```

```
2017-02-13 16:25:58,980[ost-startStop-1]  [WARN ] [e.c.p.m.s.c.c.ClasspathManager 87]   - Maven server not started looks like you don't have Maven in your path
Mon Feb 13 16:25:59 UTC 2017 - [localhost-startStop-1] Product-specified preferences called before plugin is started
```

voievodin:
@benoitf the point here is that maven-plugin is a mandatory part of ws-agent doesn't matter whether maven installed or not, maven-plugin is a part of ws-agent e.g. as git-plugin is.
I think that the property `che.plugin.maven.maven_opts` is a perfect name for plugin configuration property, but on the other hand i understand that it may be a bit confusing for admins to configure plugins in che.env.
Nevertheless it doesn't matter what the property is IT's injected into environment in this and previous version as well.

benoitf:
@evoevodin yes I see and I understand the goal. My issue is that It's a dependency of ws-agent, not the workspace itself. We could have in theory workspaces without workspace agent and then without maven plugin.

TylerJewell:
In the future yes and we would have to change many parameters then and how the system admin configured them.  I think that is an issue for the future.

TylerJewell (approved):

PR #4073, opened 2017-02-09
Implement new format workspace key {namespace}/{workspaceName} or {workspaceId}

sleshchenko:
### What does this PR do?
Implement new format workspace key:
1. {workspaceId}
2. {namespace}/{workspaceName}
Note that namespace can contain '/' symbol so workspace key can look like {namespace_part1}/{namespace_part2}/.../{namespace_partN}/{workspaceName}.

Also this PR contains required changes for IDE and Dashboard.

### What issues does this PR fix or reference?
https://github.com/eclipse/che/issues/4061

#### Changelog
Reworked workspace key to new format where namespace can contain `/` symbols.

#### Release Notes
Reworked workspace key to new format where namespace can contain `/` symbols.

#### Docs PR
Will be explained as part of the documentation for teams.

Issue #4061, opened 2017-02-08
Rework workspace key to new format where namespace can contain `/` symbols


sleshchenko:
Rework workspace key to new format where namespace can contain `/` symbols. 
Now workspace key have following format:
* {workspaceId}
* {namespace}:{workspaceName}
* :{workspaceName}

It is needed to rework workspace key to have following format:
* {workspaceId}
* {namespace}/{workspaceName}
Note that namespace can contains `/` so workspace key can looks like {namespace_part1}/{namespace_part2}/.../{namespace_partN}/{workspaceName}.

PR #4123, opened 2017-02-13
Make backward compliance for the workspace keys with ":" and optional namespace value

benoitf:
### What does this PR do?
PR https://github.com/eclipse/che/pull/4073 has broken any client using existing workspace keys
Restore the handling of previous keys format

#4061 

#### Changelog
Restore backward compliance broken on workspace keys by  https://github.com/eclipse/che/pull/4073
so clients don't need to be updated

#### Release Notes
Restore backward compliance broken on workspace keys https://github.com/eclipse/che/pull/4073

#### Docs PR
N/A

Change-Id: I884e8adef2f01774f00a9f6c08ca834fc22b5917

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
@@ -895,13 +895,27 @@ public class WorkspaceManager {
     }
 
     private WorkspaceImpl getByKey(String key) throws NotFoundException, ServerException {
+
+        int lastColonIndex = key.indexOf(":");
         int lastSlashIndex = key.lastIndexOf("/");
-        if (lastSlashIndex == -1) {
+        if (lastSlashIndex == -1 && lastColonIndex == -1) {
             // key is id
             return workspaceDao.get(key);
         }
-        final String namespace = key.substring(0, lastSlashIndex);
-        final String wsName = key.substring(lastSlashIndex + 1);
+
+        final String namespace;
+        final String wsName;
+        if (lastColonIndex == 0) {
+            // no namespace, use current user namespace
+            namespace = EnvironmentContext.getCurrent().getSubject().getUserName();
+            wsName = key.substring(1);
+        } else if (lastColonIndex > 0) {
+            wsName = key.substring(lastColonIndex + 1);
+            namespace = key.substring(0, lastColonIndex);
+        } else {
+            namespace = key.substring(0, lastSlashIndex);
+            wsName = key.substring(lastSlashIndex + 1);
+        }
         return workspaceDao.get(wsName, namespace);
     }
 
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
@@ -180,7 +180,7 @@ public class WorkspaceManager {
      * <p> Key rules:
      * <ul>
      * <li>@Deprecated : If it contains <b>:</b> character then that key is combination of namespace and workspace name
-     * <li>@Deprecated : <b></>:workspace_name</b> is valid abstract key and user will be detected from Environment.
+     * <li>@Deprecated : <b></>:workspace_name</b> is valid abstract key and current user name will be used as namespace
      * <li>If it doesn't contain <b>/</b> character then that key is id(e.g. workspace123456)
      * <li>If it contains <b>/</b> character then that key is combination of namespace and workspace name
      * </ul>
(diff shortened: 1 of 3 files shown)


gazarenkov:
I would not leave old format until we have really serious reason for that.
Do we know some third-party clients sensitive to this change? 
Otherwise, the sooner we change it - the better.

Instead I'd add add exception to the service like this:
if(key.indexOf(":"))
    throw new ServerException("Since version ... the format ... is not supported. Use ... instead");

benoitf:
@gazarenkov : Why is it a problem to get backward compliance ?
All links are using new formats, it's just for ppl retrieving in the previous format, it's still working.

Workspace KEY is something of the core part and not in a sub-sub part of plug-In so it's something that you should not break. And even more when backward compliance is possible

gazarenkov:
Sure, it requires changes and yes, may be not pleasant, it is obvious.
But need to estimate what is actually cheaper for us for now and for future. since it requires maintaining API options which we do not plan to support in future.
So, I need estimation of real risk, as I asked, before we make a decision.

benoitf:
@gazarenkov risk is high (and of course you can't estimate because by definition, with a public API you don't maintain all clients) as it will be introduced for a version without deprecation time or introducing the new format.
Also you can have clients (like CLI) that will break usage of ppl given their workspace name in their usual format. (users won't be happy) (side note : also I don't see any reason to force ppl to give a namespace when it belongs directly to their account. It requires extra typing effort and I see it as a regression).
Also by definition a client is not colocated so you can have the same client connecting to different systems. And there is no way to know the level of the API on each system (because it returns today the same version since like a couple of years)

For real estimate and better result, you need to first deprecate when introducing a new format, keep things for a while, then remove deprecate. Not : replace directly.

gazarenkov:
What you said is pretty obvious :). What I said is I would not blindly follow procedure you mentioned , I need to estimate real risk to not to be the hostage of API which is opposite to the concept.  

So, I have not thrown away any option yet, I asked if you know real clients other than ones we know about which may be impacted.

Why not to consider making CLI backward compatible to the old syntax if it is some important to you BTW?

benoitf:
here we derivate from the original issue that is currently, the CLI (and maybe other clients) are broken due to breaking changes, while at any time, code should be shippable and this PR fix the current state.

about the future of cli:
> Why not to consider making CLI backward compatible to the old syntax if it is some important to you BTW?

I would say that it's better when all clients are still compliant by doing nothing on their side than having to update each clients ? Because the effort is pretty easy on the server side.
BTW, how the client could know what the server side support as format as the API level is the same before and after the change while API is not working in the same way ?

I'm not arguing BTW that API should never be changed. Sometimes it requires changes but it's just that when changes are immediate, it breaks and it can be avoided by having timeframe and deprecation time for clients to have time to update.
So by having some period of time, CLI could be updated but it's not planned for the current sprint while API is broken right now.
And yes, I'm still discussing about API versioning, API level so that clients can work smoothly.


so yes, why not updating clients but:
- Server side in this case can handle backward compliance easily
- A correct API level on server side will help clients to address all versions of APIs (here it fails)
- A versioning of the API on server side will allow timeframe for clients to update while server side code is cleaner for the current latest version. (And then it allows the server side to drop old support after a period of some releases)

ashumilova (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -895,13 +895,27 @@ private WorkspaceImpl doCreateWorkspace(WorkspaceConfig config,
     }
 
     private WorkspaceImpl getByKey(String key) throws NotFoundException, ServerException {
+
+        int lastColonIndex = key.indexOf(":");
```
sleshchenko, line 901: Should we fix doc for WorkspaceManager#getWorkspace https://github.com/eclipse/che/pull/4073/files#diff-349ffaa889025cd6acd2d5b512c2e687L183 ?

benoitf: @sleshchenko that's a good question.
I will add info with deprecated marker

sleshchenko (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
benoitf: @sleshchenko https://github.com/eclipse/che/pull/4123/commits/07cd05da8823c6fe5f5d8f56f89bda98b939a107
is it ok ?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -179,6 +179,8 @@ public WorkspaceImpl createWorkspace(WorkspaceConfig config,
      *
      * <p> Key rules:
      * <ul>
+     * <li>@Deprecated : If it contains <b>:</b> character then that key is combination of namespace and workspace name
+     * <li>@Deprecated : <b></>:workspace_name</b> is valid abstract key and user will be detected from Environment.
```
sleshchenko: I'd rewrite to something like `and current user name will be used as namespace`

benoitf: ok I've used previous javadoc
now I've updated with your remark

slemeur (approved):

PR #4129, opened 2017-02-13
CODENVY-1749: add possibility to set dns resolvers

garagatyi:
### What does this PR do?
Add possibility to set DNS resolving servers.
If Che is set to use custom DNS that is resolvable using custom DNS resolving server admin should set env var: **CHE_DNS_RESOLVERS**
It may contain several values in comma-separated format, e.g. 
```
CHE_DNS_RESOLVERS=10.10.10.10,8.8.8.8
```

Related to codenvy/codenvy#1749

#### Changelog
Add possibility to set custom DNS resolvers for workspace **(master too)** container runtimes.

#### Release Notes
We now let you override the default DNS resolvers used by Che and you workspace runtimes. The default behavior is for Che and workspaces to inherit DNS resolution from the host. You can add `CHE_DNS_RESOLVERS=<ip-list>` to `che.env` to customize how DNS resolution will occur within the Che server and workspace containers. Services such as agents and servers running within Che workspaces will usually require custom DNS resolution when there are internal providers behind a proxy or firewall.

#### Docs PR
https://github.com/eclipse/che-docs/pull/139


garagatyi:
@bmicklea @eivantsov @JamesDrummond Please review.

TylerJewell:
Let's change it to be CHE_DNS_RESOLVERS inside of che.env.

If you can provide an explanation of what this does (or when an admin would need to configure this) then I can write release notes and identify docs changes that are needed.

Is there a matching codenvy PR?

TylerJewell:
@garagatyi - I will review this - this will not need @bmicklea or @JamesDrummond review.

riuvshin (approved):
LGTM from configuration point of view

JamesDrummond:
I think this needs docs. I created PR for docs that is in review.

TylerJewell:
@JamesDrummond - the dock PR links to a che PR - it's not a docs PR.

I am not sure this requires docs - we don't document every configuration property in or docs.

garagatyi:
@TylerJewell I updated env variable name. Also I added my comments for docs in bold, please take a look.

skabashnyuk (approved):

garagatyi:
Also not all the parts of the system inherit DNS resolution rules from host. Nginx container doesn't respect that and use 8.8.8.8 for dns resolution. So user might be surprised if **by default** we inherit from host but WSs don't start because of Nginx.

garagatyi:
@JamesDrummond If you still need docs please use my comments from this PR to change explanation in your PR.

riuvshin (approved):

garagatyi:
@riuvshin I had to rework PR, so now empty string is used instead of NULL. It is needed because Java code can't interpret NULL as empty array

TylerJewell (approved):

riuvshin (approved):

garagatyi:
@TylerJewell please take a look at release notes. I added my comments into it.

TylerJewell:
Ok for you to merge.  I can fix release notes after merge when have time to review.

JamesDrummond:
@TylerJewell I updated the docs PR link. I somehow copied the wrong url. I think documentation on this would be helpful to those evaluating the product features.

@garagatyi What is `master` that you refer to above?

garagatyi:
@JamesDrummond I meant ws-master aka che-server

JamesDrummond:
@garagatyi I rewrote the docs to include che server. https://github.com/eclipse/che-docs/pull/139/files#diff-ba9f2804ca5dc20e010ba41f64ef1cac . Please review.

TylerJewell:
I updated the release notes based upon @garagatyi comments

Issue #4141, opened 2017-02-14
Use Persistent Volumes Claims when creating workspaces


snjeza:
Usage of hostPath volumes is not supported on multi-node OpenShift clusters. We should use PersistentVolumeClaims instead.

This issue is to change OpenShiftConnector::createContainer to use PersistentVolumeClaims.

See https://issues.jboss.org/browse/CHE-60

**OS and version:**    any

skabashnyuk:
won't fix. out of date.

PR #4142, opened 2017-02-14
CHE-4141 - Use Persistent Volumes Claims when creating workspaces

snjeza:

### What does this PR do?

Uses Persistent Volumes Claims when creating workspaces in OpenShift

#4141 

The PR requires PV/PVC as described in https://issues.jboss.org/browse/CHE-59

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
@@ -283,7 +283,8 @@ che.openshift.project=eclipse-che
 che.openshift.serviceaccountname=cheserviceaccount
 che.openshift.liveness.probe.delay=300
 che.openshift.liveness.probe.timeout=1
-che.openshift.persistent.volume.claim=claim-che-data
+che.openshift.workspaces.pvc.name=claim-che-data
+che.openshift.workspaces.pvc.quantity=10Gi
 
 # Which implementation of DockerConnector to use in managing containers. In general,
 # the base implementation of DockerConnector is appropriate, but OpenShiftConnector
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
@@ -283,6 +283,7 @@ che.openshift.project=eclipse-che
 che.openshift.serviceaccountname=cheserviceaccount
 che.openshift.liveness.probe.delay=300
 che.openshift.liveness.probe.timeout=1
+che.openshift.persistent.volume.claim=claim-che-data
 
 # Which implementation of DockerConnector to use in managing containers. In general,
 # the base implementation of DockerConnector is appropriate, but OpenShiftConnector
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
@@ -283,7 +283,7 @@ che.openshift.project=eclipse-che
 che.openshift.serviceaccountname=cheserviceaccount
 che.openshift.liveness.probe.delay=300
 che.openshift.liveness.probe.timeout=1
-che.openshift.workspaces.pvc.name=claim-che-data
+che.openshift.workspaces.pvc.name=claim-che-workspace
 che.openshift.workspaces.pvc.quantity=10Gi
 
 # Which implementation of DockerConnector to use in managing containers. In general,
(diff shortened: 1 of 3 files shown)


assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -284,6 +284,7 @@ che.openshift.project=eclipse-che
 che.openshift.serviceaccountname=cheserviceaccount
 che.openshift.liveness.probe.delay=300
 che.openshift.liveness.probe.timeout=1
+che.openshift.persistent.volume.claim=claim-che-data
```
l0rd: Since this property is about PVC for workspaces I would rather call it: `che.openshift.workspaces.pvc.name`. What do yo think?

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -136,6 +142,7 @@
     private final String          openShiftCheServiceAccount;
     private final int             openShiftLivenessProbeDelay;
     private final int             openShiftLivenessProbeTimeout;
+    private final String          persistentVolumeClaim;
```
l0rd: I would rather call it `workspacesPersistentVolumeClaim` or `workspacesPVC` to make it clear that it will be shared by all workpaces.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -972,34 +981,77 @@ private ContainerInfo createContainerInfo(Service svc,
...
     private List<VolumeMount> getVolumeMountsFrom(String[] volumes, String workspaceID) {
         List<VolumeMount> vms = new ArrayList<>();
-        for (String volume : volumes) {
-            String mountPath = volume.split(":",3)[1];
-            String volumeName = getVolumeName(volume);
-
+        PersistentVolumeClaim pvc = getClaimCheWorkspace();
+        if (pvc != null) {
             VolumeMount vm = new VolumeMountBuilder()
+                .withMountPath("/data")
```
l0rd: Can you please extract the string "/data" to a constant?

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -972,34 +981,77 @@ private ContainerInfo createContainerInfo(Service svc,
...
 
+    private PersistentVolumeClaim getClaimCheWorkspace() {
+        PersistentVolumeClaimList pvcList = openShiftClient.persistentVolumeClaims().list();
+        for(PersistentVolumeClaim pvc:pvcList.getItems()) {
+            if (persistentVolumeClaim.equals(pvc.getMetadata().getName())) {
+                return pvc;
+            }
+        }
+        Map<String, Quantity> requests = new HashMap<>();
+        requests.put("storage", new Quantity("10Gi"));
```
l0rd: Can you extract "10Gi" to a constant. Or better would be if you use a new che property for that to make it configurable.

l0rd (commented):
Thanks @snjeza for this PR. 

Please review my comments and could you explain what happens if: 
- che-server and che-workspace share the same PVC
- che-server and che-workspace use 2 distinct PVC

snjeza:
> Please review my comments and could you explain what happens if:

>    che-server and che-workspace share the same PVC
>    che-server and che-workspace use 2 distinct PVC

You can use the same or a different PVC for the server and a workspace. It is just important to ensure that the workspace's PVC includes the required workspace's directories.

> Can you extract "10Gi" to a constant. Or better would be if you use a new che property for that to make it configurable.

I have added a new property. We can also create PVC manually (similarly as in CHE-59) so that PVCs are configurable in any case.

I have updated the PR.

l0rd:
If che-server and workspaces use distinct PVC how do you "ensure ensure that the workspace's PVC includes the required workspace's directories". Is that a manual step that need to be done?

snjeza:
> If che-server and workspaces use distinct PVC how do you "ensure ensure that the workspace's PVC includes the required workspace's directories". Is that a manual step that need to be done?

Yes, it is. We can use PV labels.

l0rd:
Ok. Let me think about that. But we will probably merge this PR and open a
new ticket to address the distinct PVC issue.

snjeza:
We could also use the volumeName and claimRef properties. See https://docs.openshift.org/latest/dev_guide/persistent_volumes.html#persistent-volumes-volumes-and-claim-prebinding

snjeza:
@l0rd 
Could you, please, take a look at https://issues.jboss.org/browse/CHE-100?focusedCommentId=13366512&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13366512?

snjeza:
Moved to https://github.com/eclipse/che/pull/4308

PR #4154, opened 2017-02-16
CODENVY-1403: fix bug in machine removal

garagatyi:
### What does this PR do?
Fix bug in machine removal.
Bug occurs when container is unavailable without stop of machine.
In that case we don't stop machine instance and not cleanup
resources.

Related to codenvy/codenvy#1403

#### Changelog
Fix machine stop in case docker container was stopped or killed not by Codenvy.

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
@@ -11,7 +11,6 @@
 package org.eclipse.che.api.environment.server;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.base.Strings;
 
 import org.eclipse.che.api.agent.server.AgentRegistry;
 import org.eclipse.che.api.agent.server.exception.AgentException;
@@ -31,6 +30,7 @@ import org.eclipse.che.api.core.model.workspace.Environment;
 import org.eclipse.che.api.core.model.workspace.ExtendedMachine;
 import org.eclipse.che.api.core.model.workspace.ServerConf2;
 import org.eclipse.che.api.core.notification.EventService;
+import org.eclipse.che.api.core.notification.EventSubscriber;
 import org.eclipse.che.api.core.util.AbstractLineConsumer;
 import org.eclipse.che.api.core.util.LineConsumer;
 import org.eclipse.che.api.core.util.MessageConsumer;
@@ -123,6 +123,7 @@ public class CheEnvironmentEngine {
     private final Pattern                        recipeApiPattern;
     private final ContainerNameGenerator         containerNameGenerator;
     private final AgentRegistry                  agentRegistry;
+    private final WorkspaceSharedPool            sharedPool;
 
     private volatile boolean isPreDestroyInvoked;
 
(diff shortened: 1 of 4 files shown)


wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
```diff
@@ -1279,6 +1286,65 @@ private CheServiceImpl machineConfigToService(MachineConfig machineConfig) throw
...
+        MachineStatusEvent destroyedEvent = newDto(MachineStatusEvent.class)
+                .withEventType(MachineStatusEvent.EventType.DESTROYED)
+                .withDev(instance.getConfig().isDev())
+                .withMachineId(machineId)
+                .withWorkspaceId(workspaceId)
+                .withMachineName(instance.getConfig().getName())
+                .withError(message);
+
+        try {
+            if (!Strings.isNullOrEmpty(message)) {
```
mshaposhnik: How it can became empty ??

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
```diff
@@ -158,7 +160,12 @@ public CheEnvironmentEngine(SnapshotDao snapshotDao,
                                                 "/recipe/.*$)|(^/recipe/.*$)");
         this.containerNameGenerator = containerNameGenerator;
 
-        eventService.subscribe(new MachineCleaner());
+        eventService.subscribe(event -> {
```
voievodin: Can you please subscribe outside the constructor e.g.
```java
@PostConstruct
private void subscribe() {
    eventSubscriber.subscribe(...)
}
```
As you implicitly expose instance which hasn't been fully constructed yet

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
```diff
@@ -1279,6 +1286,65 @@ private CheServiceImpl machineConfigToService(MachineConfig machineConfig) throw
...
+
+        String message = "Machine is destroyed";
+        if (eventType == OOM) {
+            message = message +
+                      ". The processes in this machine need more RAM. This machine started with " +
+                      instance.getConfig().getLimits().getRam() +
+                      "MB. Create a new machine configuration that allocates additional RAM or increase " +
+                      "the workspace RAM limit in the user dashboard.";
+        }
+        MachineStatusEvent destroyedEvent = newDto(MachineStatusEvent.class)
```
voievodin: What if we move this event to the place where it's published
```java
eventService.publish(newDto(...));
```

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
garagatyi: :smile: copied from older code. Good catch!

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/CheEnvironmentEngine.java
garagatyi: ok

mshaposhnik (approved):

skabashnyuk (approved):

garagatyi:
@mshaposhnik @evoevodin I updated PR

voievodin (approved):

mmorhun (approved):

PR #4155, opened 2017-02-16
update docker hub url in cli tests

riuvshin:
### What does this PR do?
in cli tests we checked old url for docker hub `dockerhub.com` due to some reasons that url is down really often so we switch to new url

#### Changelog
update docker hub url in cli tests

### DIFF
--- a/dockerfiles/base/scripts/base/library.sh
+++ b/dockerfiles/base/scripts/base/library.sh
@@ -102,7 +102,7 @@ initiate_offline_or_network_mode(){
       # Removing this info line as it was appearing before initial CLI output
 #      info "cli" "Checking network... (hint: '--fast' skips nightly, version, network, and preflight checks)"
       local HTTP_STATUS_CODE=$(curl -I -k hub.docker.com -s -o /dev/null --write-out '%{http_code}')
-      if [[ ! $HTTP_STATUS_CODE -eq "301" ]]; then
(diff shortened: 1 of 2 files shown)


TylerJewell (changes requested):
Hi.  It looks like you found duplicate method. The one in library should be removed.

benoitf (approved):
if we can avoid duplicated code, that's better

riuvshin:
@TylerJewell @benoitf yeah I was wondering why when guys trying skip tests it still fail I believe that this one in library was not skipped

dockerfiles/base/scripts/base/library.sh
```diff
@@ -74,56 +74,6 @@ server_is_booted() {
   fi
 }
```
benoitf: AFAIk this method is called from https://github.com/eclipse/che/blob/b446d114250d6e4e7b23e57a045929d9b3f20a6d/dockerfiles/base/scripts/base/startup_01_init.sh#L283

riuvshin: should I remove this line ?

riuvshin: @TylerJewell should I remove that line? I don;t know how to detect where that method was used

riuvshin:
we agreed to clean up duplicate things after release

benoitf (approved):

TylerJewell (approved):

PR #4175, opened 2017-02-17
[cli] Add More Checks

TylerJewell:

### What does this PR do?
Reorganizes the initialization to be easier to follow.  Standardizes all of the `init_` methods with a common name.  Places all script loading (mostly) in the same location for readability. The only exception is the lazy loading of command scripts and the library, which happens in the final library.

Also adds in a check that if you are mounting `:/repo` and you have not `--skip:scripts` then will compare using diff all of the files inside your CLI image to those mounted from the repo.  If they differ, then the CLI does a hard fail with an error message.

Fixed a bug in the CLI where if curling DockerHub for the version list fails, then an unhelpful message appears. Added '--skip:network' option to skip over this along with detecting a failed connection to DockerHub. Also added in all of the curl commands that we execute with their output to be included into the 'info --bundle' diagnostic.

Matching PRs:
1. Codenvy - https://github.com/codenvy/codenvy/pull/1800
2. SaaS - https://github.com/codenvy/saas/pull/147
3. ARTIK - https://github.com/codenvy/artik/pull/348

#### Changelog
[cli] Simplify execution of initialization methods
[cli] Add diff for repo mode
[cli] Add checks for all curl commands and ensure '--skip:network' applied

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/dockerfiles/base/scripts/base/startup_04_pre_cli_init.sh
+++ b/dockerfiles/base/scripts/base/startup_04_pre_cli_init.sh
@@ -75,7 +75,7 @@ cli_init() {
   fi
 }
 
-verify_nightly_accuracy() {
+cli_verify_nightly() {
   # Per request of the engineers, check to see if the locally cached nightly version is older
   # than the one stored on DockerHub.
   if is_nightly; then
@@ -95,7 +95,7 @@ verify_nightly_accuracy() {
   fi
 }
 
-verify_version_compatibility() {
+cli_verify_version() {
   # Do not perform a version compatibility check if running upgrade command.
   # The upgrade command has its own internal checks for version compatibility.
   if [[ "$@" == *"upgrade"* ]]; then
--- a/dockerfiles/base/scripts/base/startup_04_pre_cli_init.sh
+++ b/dockerfiles/base/scripts/base/startup_04_pre_cli_init.sh
@@ -73,17 +73,6 @@ cli_init() {
       fi
     fi
   fi
-
-  # Special function to perform special behaviors if you are running nightly version
-  verify_nightly_accuracy
-
-  # Do not perform a version compatibility check if running upgrade command.
-  # The upgrade command has its own internal checks for version compatibility.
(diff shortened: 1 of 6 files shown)


dockerfiles/cli/scripts/pre_init.sh
```diff
@@ -17,7 +17,7 @@ pre_init() {
   ADDITIONAL_GLOBAL_OPTIONS=""
  
   # This must be incremented when BASE is incremented by an API developer
-  CHE_CLI_API_VERSION=1
+  CHE_CLI_API_VERSION=2
```
riuvshin, line 20: as I understand this will require PRs in codenvy, saas, and artik

TylerJewell: Yes, have opened PRs for all of them.

dockerfiles/base/scripts/base/startup_04_pre_cli_init.sh
```diff
@@ -125,7 +124,7 @@ verify_version_compatibility() {
   CHE_IMAGE_VERSION=$(get_image_version)
 
   # Only check for newer versions if not in offline mode and not nightly.
```
benoitf: comment needs to be adjusted

dockerfiles/base/scripts/base/startup_04_pre_cli_init.sh
```diff
@@ -125,7 +124,7 @@ verify_version_compatibility() {
...
-  if ! is_offline && ! is_nightly; then
+  if ! is_offline && ! is_nightly && ! is_fast && ! skip_network; then
```
benoitf: comment before this line needs to be adjusted

benoitf (approved):

dockerfiles/base/scripts/base/startup_04_pre_cli_init.sh
TylerJewell: Thanks - fixed

PR #4185, opened 2017-02-20
Add Eclipse KeyScheme as Extension

akervern:
Signed-off-by: Arnaud Kervern <akervern@nuxeo.com>

### What does this PR do?
It adds an Eclipse based key schema as a new IDE extension.
And expose actions ids as contants to let them accessible from another module.

But not sure if you prefer to have this Eclipse schema par default; or the previous one. The originating code set Eclipse as the default one, but it was empty; so all key bindings were the global ones.

#### Changelog
- New Eclipse based Key Schema

#### Release Notes
- New Eclipse based Key Schema


TylerJewell:
Thanks for making this contribution.  @JamesDrummond @akervern - the release notes will need to be written as if they were end user documentation.  So probably some questions about how to activate / deactivate this schema.  And we should probably also create a list of all of the bindings and their key strokes generated as well.  

I think we would largely have the release notes + documentation PR be identical in this regard.  I don't think we have ever generated key-binding or schema related documentation, so this might be the first instance of that.  It shouldn't be a lot of docs, but enough for end users to figure out their way.

JamesDrummond (changes requested):
Would it make sense to map the keybinding as well? Maybe `keyBinding.getGlobal().addKey(new KeyBuilder().action().alt().charCode('n').build(), NAVIGATE_TO_FILE);` => `keyBinding.getGlobal().addKey(new KeyBuilder().action().alt().charCode(dictionaryObj.get(NAVIGATE_TO_FILE)).build(), NAVIGATE_TO_FILE);`

akervern:
Good idea, but i'm not sure to get exactly what you want. For me, an interesting improvement could be to have a map of keyBinding per action and registrating them at once using a lambda, it could make the code more readable and could be the first step for allowing import the key binding from a definition file outside of the compiled code to ease modification / contribution.

Can you explain why you would like to extract chars key without taking into account the modifiers; and I think it can be confusing as `CharCodeWithModifiers` handles uppercase chars the same way as a lowercase char + `shift` modifier.

JamesDrummond:
@akervern Yes I didn't take into account modifiers. I like your proposed idea. Are you able to implement this? Would be a great step forward.

akervern:
Hi, is this commit enough for you? Or do you need something more?

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/core/StandardComponentInitializer.java
```diff
@@ -121,6 +121,25 @@
...
+    public static final String OPEN_RECENT_FILES   = "openRecentFiles";
+    public static final String DELETE_ITEM         = "deleteItem";
+    public static final String NEW_FILE            = "newFile";
+    public static final String CREATE_PROJECT      = "createProject";
+    public static final String IMPORT_PROJECT      = "importProject";
+    public static final String CLOSE_ACTIVE_EDITOR = "closeActiveEditor";
+    public static final String SIGNATURE_HELP      = "signatureHelp";
+    public static final String RENAME              = "renameResource";
+    public static final String SHOW_REFERENCE      = "showReference";
+
```
vparfonov: Can you move this declarations to the che/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/action/IdeActions.java ?

plugins/plugin-keybinding-eclipse/plugin-keybinding-eclipse-ide/pom.xml
```diff
@@ -0,0 +1,112 @@
...
+        </resources>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>analyze</id>
+                        <configuration>
+                            <skip>true</skip>
```
skabashnyuk, line 72: What for?

plugins/plugin-keybinding-eclipse/plugin-keybinding-eclipse-ide/pom.xml
```diff
@@ -0,0 +1,112 @@
...
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
```
skabashnyuk: What for?

plugins/plugin-keybinding-eclipse/plugin-keybinding-eclipse-ide/src/main/resources/org/eclipse/che/plugin/keybinding/eclipse/EclipseKeyBinding.gwt.xml
```diff
@@ -0,0 +1,27 @@
...
+    <inherits name="com.google.gwt.i18n.I18N"/>
+    <inherits name="com.google.gwt.json.JSON"/>
+    <inherits name='org.eclipse.che.ide.Api'/>
+    <inherits name='org.eclipse.che.ide.ui.CodenvyUI'/>
+    <inherits name="com.google.gwt.inject.Inject"/>
+    <inherits name="org.eclipse.che.api.project.Project"/>
+    <inherits name="org.eclipse.che.ide.Core"/>
+
+    <source path=""/>
+</module>
```
skabashnyuk: new line

akervern:
@skabashnyuk You're right; copied/pasted from another plugin and useless.

plugins/plugin-keybinding-eclipse/plugin-keybinding-eclipse-ide/pom.xml
akervern: In fact, after tries; not sure why and I can't investigate. 

But `dependency` plugin looks to not be able to resolve those dependencies as `used` and throw an error.
Of course, artifact won't compile if I remove them. So, there is certainly something wrong somewhere as several extensions are using this flag.

vparfonov (approved):

JamesDrummond (approved):

vparfonov:
@akervern Fixed conflict and merged here https://github.com/eclipse/che/pull/4394

akervern:
Cool Thanks!

Issue #4091, opened 2017-02-10
Move factory functionality from codenvy to che except dashboard part


skabashnyuk:
Move factory functionality from codenvy to che except dashboard part

mshaposhnik:
Done in branch.

PR #4192, opened 2017-02-21
Init factory schema in che database

voievodin:
### What does this PR do?
Inits schema for factory in che database.
All the factory schema table names start with _che_factory_.

Also added ability to generate persistence.xml dynamically which is simpler approach for tests
comparing to the current one, it :
- ensures all dependencies are defined
- it is based on a real server url, user, password, so changes made to those values wouldn't require persistence.xml changes 
- allows to adjust jpa related configuration in a single place(module)

`H2TestHelper.inMemoryDefault()` & `H2TestHelper.shutdownDefault` methods are deprecated, suggested to use `H2DBTestServer` instead, all the classes using these methods were rewritten to use db server.

#4091 

#### Changelog
Init factory schema in che database

#### Release Notes
Factory data is now stored in che database

#### Docs PR
N/A


mshaposhnik:
Maybe add some indexes? E.g. `username + factory_name`, `factory_id` in images e.t.c

wsmaster/che-core-sql-schema/src/main/resources/che-schema/5.4.0/1__add_factory.sql
```diff
@@ -0,0 +1,164 @@
...
+ALTER TABLE che_factory ADD CONSTRAINT fk_che_f_ide_id FOREIGN KEY (ide_id) REFERENCES che_factory_ide (id);
+--------------------------------------------------------------------------------
+
+
+-- Factory Images --------------------------------------------------------------
+CREATE TABLE che_factory_image (
+    image_data   BYTEA,
+    media_type   VARCHAR(255),
+    name         VARCHAR(255),
+    factory_id   VARCHAR(255)
```
skabashnyuk: factory_id  - NOT NULL?

wsmaster/che-core-sql-schema/src/main/resources/che-schema/5.4.0/1__add_factory.sql
```diff
@@ -0,0 +1,164 @@
...
+);
+-- constraints
+ALTER TABLE che_factory_image ADD CONSTRAINT fk_che_factory_image_factory_id FOREIGN KEY (factory_id) REFERENCES che_factory (id);
```
skabashnyuk: index on  factory_id

wsmaster/che-core-sql-schema/src/main/resources/che-schema/5.4.0/1__add_factory.sql
```diff
@@ -0,0 +1,164 @@
...
+    id              VARCHAR(255),
+
+    PRIMARY KEY (entity_id)
+);
+--------------------------------------------------------------------------------
+
+
+-- Factory action properties ---------------------------------------------------
+CREATE TABLE che_factory_action_properties (
+    action_entity_id        BIGINT,
```
skabashnyuk: action_entity_id not null and index?

wsmaster/che-core-sql-schema/src/main/resources/che-schema/5.4.0/1__add_factory.sql
```diff
@@ -0,0 +1,164 @@
...
+
+    PRIMARY KEY (id)
+);
+--------------------------------------------------------------------------------
+
+
+-- Factory action --------------------------------------------------------------
+CREATE TABLE che_factory_action (
+    entity_id       BIGINT          NOT NULL,
+    id              VARCHAR(255),
```
skabashnyuk, line 29: not null?

skabashnyuk: index

wsmaster/che-core-sql-schema/src/main/resources/che-schema/5.4.0/1__add_factory.sql
```diff
@@ -0,0 +1,164 @@
...
+ALTER TABLE che_factory_ide ADD CONSTRAINT fk_che_f_ide_on_app_loaded_id FOREIGN KEY (on_app_loaded_id) REFERENCES che_factory_on_app_loaded_action (id);
+--------------------------------------------------------------------------------
+
+
+-- Factory ---------------------------------------------------------------------
+CREATE TABLE che_factory (
+    id                  VARCHAR(255)         NOT NULL,
+    name                VARCHAR(255),
+    version             VARCHAR(255)         NOT NULL,
+    created             BIGINT,
```
skabashnyuk, line 135: not null?

voievodin:
going to add indexes + add unique constraint for _name + factory_id_

wsmaster/che-core-sql-schema/src/main/resources/che-schema/5.4.0/1__add_factory.sql
voievodin: not sure it has to be not null, 
https://github.com/eclipse/che/blob/0e8b21ad3ff5da80a7a03ddf3f0b281cc3594dcb/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/factory/Action.java#L23-L23

sleshchenko: IDE won't perform action when it will have nullable id

voievodin: you mean that null is possible ?

sleshchenko: I think it must be not null

wsmaster/che-core-api-user/src/test/java/org/eclipse/che/api/user/server/jpa/JpaTckModule.java
```diff
@@ -12,40 +12,51 @@
...
-
 /**
  * @author Yevhenii Voevodin
  */
 public class JpaTckModule extends TckModule {
 
     @Override
     protected void configure() {
-        install(new JpaPersistModule("main"));
+        H2DBTestServer server = new H2DBTestServer();
```
sleshchenko, line 47: Why do you need instance creation here instead of H2DBTestServer.startDefault() ?

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/model/impl/ActionImpl.java
```diff
@@ -31,21 +31,21 @@
...
     private Long entityId;
 
     @Column(name = "id")
     private String id;
 
     @ElementCollection
-    @CollectionTable(name = "action_properties", joinColumns = @JoinColumn(name = "action_entityid"))
-    @MapKeyColumn(name = "properties_key")
-    @Column(name = "properties")
+    @CollectionTable(name = "che_factory_action_properties", joinColumns = @JoinColumn(name = "action_entity_id"))
```
sleshchenko, line 46: Why do you use plural for che_factory_action_propertie**s** table name here but single for [che_factory_image](https://github.com/eclipse/che/pull/4192/files#diff-e520532e8b7fceec9cedf14236d578ffR89) table name?

PR #4209, opened 2017-02-22
CHE-117 kubernetes-client will autodetect auth details both inside and outsid…

rawlingsj:
…e openshift so no need to use config

### What does this PR do?
For more information including the order configuration is decided see https://github.com/fabric8io/kubernetes-client#configuring-the-client

TL;DR when running locally simply connect to your OpenShift cluster using `oc login` and kubernetes-client will use the token from the filesystem.  Similarly when running in a pod on OpenShift kubernetes-client will use the token automatically mounted inside the pod by OpenShift.

https://issues.jboss.org/browse/CHE-117

#### Changelog
 - OpenShift connector uses filesystem tokens to authenticate with OpenShift API server

#### Release Notes
 - OpenShift connector uses filesystem tokens to authenticate with OpenShift API server

### DIFF
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
@@ -143,10 +143,6 @@ public class OpenShiftConnector extends DockerConnector {
                               DockerConnectionFactory connectionFactory,
                               DockerRegistryAuthResolver authResolver,
                               DockerApiVersionPathPrefixProvider dockerApiVersionPathPrefixProvider,
-                              @Named("che.openshift.endpoint") String openShiftApiEndpoint,
-                              @Nullable @Named("che.openshift.token") String openShiftToken,
-                              @Nullable @Named("che.openshift.username") String openShiftUserName,
-                              @Nullable @Named("che.openshift.password") String openShiftUserPassword,
                               @Named("che.openshift.project") String openShiftCheProjectName,
                               @Named("che.openshift.serviceaccountname") String openShiftCheServiceAccount,
                               @Named("che.openshift.liveness.probe.delay") int openShiftLivenessProbeDelay,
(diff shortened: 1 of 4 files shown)


ibuziuk:
ci-build

ibuziuk:
@l0rd It looks like I can not trigger the build ? can you 
@rawlingsj +1 to apply the PR

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -158,12 +154,7 @@ public OpenShiftConnector(ConfigBuilder configBuilder,
...
-        Config config = getOpenShiftConfig(configBuilder,
-                                           openShiftApiEndpoint,
-                                           openShiftToken,
-                                           openShiftUserName,
-                                           openShiftUserPassword);
-        this.openShiftClient = new DefaultOpenShiftClient(config);
+        this.openShiftClient = new DefaultOpenShiftClient();
     }
 
     private Config getOpenShiftConfig(ConfigBuilder configBuilder,
```
l0rd: I think this method is not used anymore and can be safely removed

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -143,10 +143,6 @@ public OpenShiftConnector(ConfigBuilder configBuilder,
                               DockerConnectionFactory connectionFactory,
                               DockerRegistryAuthResolver authResolver,
                               DockerApiVersionPathPrefixProvider dockerApiVersionPathPrefixProvider,
-                              @Named("che.openshift.endpoint") String openShiftApiEndpoint,
```
l0rd, line 146: We should remove the corresponding properties in https://github.com/eclipse/che/blob/master/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties#L282 too

l0rd (commented):
Thanks @rawlingsj! That really helps. If you think they makes sense can you please address my 2 comments?

l0rd:
@ibuziuk I cannot trigger the build neither

rawlingsj:
sure thing - I force pushed the original commit, seemed easier than creating another commit and the n squash after for the small change, but not sure how you folks prefer to do things.  Anyways PR updated :)

benoitf:
ci-build

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
rawlingsj: done

l0rd (approved):

l0rd:
Thank you @benoitf ;-)

PR #4230, opened 2017-02-24
Ensure OpenShift pod is terminated before returning from removeContainer

amisevsk:
### What does this PR do?
Prevents issue where it is possible to start workspace while pod is terminating. When this occurs, Che creates a new pod but workspace start fails. This pod must be removed manually before that workspace can be launched again.

This is done by a) waiting for pod to terminate before returning from removeContainer, and b) removing created deployment and service when an exception occurs during createContainer.

Additionally, increases maximum wait time for retreiving OpenShift image stream metadata.

#### Changelog
- Fix issue where it was lose track of running workspace in OpenShift.

#### Docs PR
- No change to documentation, is a bugfix.

### DIFF
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
@@ -97,8 +97,6 @@ import io.fabric8.kubernetes.api.model.VolumeMount;
 import io.fabric8.kubernetes.api.model.VolumeMountBuilder;
 import io.fabric8.kubernetes.api.model.extensions.Deployment;
 import io.fabric8.kubernetes.api.model.extensions.DeploymentBuilder;
-import io.fabric8.kubernetes.api.model.extensions.ReplicaSet;
-import io.fabric8.kubernetes.api.model.extensions.ReplicaSetList;
 import io.fabric8.openshift.api.model.ImageStream;
 import io.fabric8.openshift.api.model.ImageStreamTag;
 import io.fabric8.openshift.client.DefaultOpenShiftClient;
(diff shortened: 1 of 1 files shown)


amisevsk:
@ibuziuk If you have a moment, could you check if this PR helps with the `ImageStream not found` issue? I've stopped encountering it on this branch.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -967,6 +932,47 @@ private ContainerInfo createContainerInfo(Service svc,
...
+            LOG.info("Removing OpenShift Deployment {}", deployment.getMetadata().getName());
+            openShiftClient.resource(deployment).delete();
+        }
+
+        // Wait for all pods to terminate before returning.
+        List<Pod> pods = openShiftClient.pods()
+                                        .inNamespace(openShiftCheProjectName)
+                                        .withLabel(OPENSHIFT_DEPLOYMENT_LABEL, deploymentName)
+                                        .list()
+                                        .getItems();
```
l0rd: Why don't you just get the size of the list? And then iterate until this size get to zero? A for loop in a while loop complicated. You could avoid the for loop if you just get the size of the pods list at every iteration.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -967,6 +932,47 @@ private ContainerInfo createContainerInfo(Service svc,
...
+        try {
+            boolean podsTerminated = false;
+            while (!podsTerminated) {
```
l0rd: What if a pod never terminates? Shouldn't we wait OPENSHIFT_WAIT_POD_TIMEOUT and then exit?

l0rd (commented):
Thank you @amisevsk for this PR. I've added 2 comments.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
amisevsk: That's a lot cleaner, thanks. Fixed.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
amisevsk: Fixed. Although, if a pod never terminates, the corresponding workspace will be blocked from launching, breaking a fair bit of Che.

l0rd (approved):

PR #4241, opened 2017-02-28
Implement getEvents method in OpenShiftConnector

l0rd:
### What does this PR do?
Provide a dummy implementation of method getEvents in OpenShiftConnector.

#### Changelog
- Enable to bootstrap Che on OpenShift without mounting the Docker Socket

#### Docs PR
- N/A


benoitf:
ci-build

amisevsk (approved):
LGTM. This is necessary to get Che running on OpenShift without the Docker Socket. We can implement getEvents() later if necessary.

PR #4269, opened 2017-02-28
Expose Che server and workspaces through one single TCP port

l0rd:
### What does this PR do?
Add a new [ServerEvaluationStrategy](https://github.com/eclipse/che/blob/master/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ServerEvaluationStrategy.java): SinglePortLocalDockerServerEvaluationStrategy. 

This server strategy can be enabled adding `che.docker.server_evaluation_strategy=single-port`
in `che.properties`.

When the single-port strategy is enabled a workspace server address has the format `server-ref.workspace-id.che-address` like for example:
```
terminal.79rfwhqaztq2ru2k.che.local
````

To use the single-port strategy a reverse proxy should be used. This PR comes with the changes needed to use OpenShift embedded reverse-proxy too.

#2847 #1560

#### Changelog
- Added single-port server evaluation strategy to expose one single TCP port for all Che services

#### Release Notes
- Expose Che server and workspace servers through one TCP port adding `che.docker.server_evaluation_strategy=single-port`
in `che.properties`.

#### Docs PR
How to use single-port strategy with OpenShift embedded reverse proxy will described in the upcoming documentation to run Che with OpenShift.

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ServerEvaluationStrategy.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ServerEvaluationStrategy.java
@@ -61,10 +61,13 @@ public abstract class ServerEvaluationStrategy {
      * @param internalAddress
      *         address passed into {@code getServers}; used as fallback if address cannot be
      *         retrieved from containerInfo.
+     * @param serverConfMap
+     *         additional Map of {@link ServerConfImpl}.
      * @return a Map of port protocol (e.g. "4401/tcp") to address (e.g. "localhost:32317")
      */
     protected abstract Map<String, String> getExternalAddressesAndPorts(ContainerInfo containerInfo,
-                                                                        String internalAddress);
+                                                                        String internalAddress,
+                                                                        Map<String, ServerConfImpl> serverConfMap);
 
     /**
      * Constructs a map of {@link ServerImpl} from provided parameters, using selected strategy
@@ -100,7 +103,7 @@ public abstract class ServerEvaluationStrategy {
         }
 
         Map<String, String> internalAddressesAndPorts = getInternalAddressesAndPorts(containerInfo, internalHost);
-        Map<String, String> externalAddressesAndPorts = getExternalAddressesAndPorts(containerInfo, internalHost);
+        Map<String, String> externalAddressesAndPorts = getExternalAddressesAndPorts(containerInfo, internalHost, serverConfMap);
 
         Map<String, ServerImpl> servers = new LinkedHashMap<>();
 
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
@@ -206,9 +206,14 @@ che.docker.ip.external=NULL
 #   - 'default':      internal address is address of docker host and ephemeral port are used
 #   - 'docker-local': internal address is address of container within docker network, and exposed ports
 #                     are used.
+#   - 'single-port':  internal address is set as in docker-local strategy, external address is composed
+#                     using the workspace server name, workspace ID and Che external address
+#                     (e.g terminal.79rfwhqaztq2ru2k.che.local)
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

**Comment by garagatyi** [comment]
Can you point in issue description relation of this PR to mentioned issues. Some people might think that it fixes both issues whereas it is just related to it.

Please re-phrase changelog to reflect then fact that it works with external reverse proxy (Openshift only or can be another one?).

Can you elaborate on how to access terminals from 2 machines in single workspace with this notation?

**Comment by TylerJewell** [comment]
@l0rd - thank you for the contribution!

Let's identify a plan for what we would like to have accomplished before we merge this PR.  First, I have added a couple additional maintainers as reviewers for this issue. We'd like to make sure that any changes here are compatible with other custom assemblies, particularly Codenvy as well.

Some items that I'd like to organize for this PR:
1. A demo review of this implementation with a couple maintainers for a Q&A.
2. Instructions on how to deploy this configuration with traefik or nginx.
3. Discuss how to expose configuring this in the `che.env` file.
4. Discuss a configuration of the `che.env` that also launches the reverse proxy.
5. Updating the docs to reflect these various configurations - particularly port requirements.
6. Are there any implications to the new diagnostic page in the UD?

**Comment by benoitf** [comment]
@TylerJewell 
if URL provided by REST API are in the format "terminal.79rfwhqaztq2ru2k.che.local" that is not directly reachable yes that would be an issue for diagnostic tool that connect directly to the URLs provided by server side.

**Comment by l0rd** [comment]
@TylerJewell please note that this single port strategy is enabled only if property `che.docker.server_evaluation_strategy` is set to `singel-port`. And by default it is not. 

In other words this PR, as the rest of the OpenShift connector related PR, should not break anything and can be merged safely. And even if I agree with all your points this PR should not be blocked by these. We can merge it now (that will help us and won't break anything) and still continue improving it together via other PRs in the next weeks.

**Comment by benoitf** [comment]
for me the wording "single-port" is not obvious as it requires extra front-end tools to use that. It could be seen by end users as a property that can be used directly while it won't work directly.

**Comment by TylerJewell** [comment]
@l0rd - while the initiation of this was for openshift, it's a core evaluation strategy, so will only want to merge this into the system until we have provided for enhancements to the various utilities that will be affected by it - including diagnostic utility, CLI, configuration and docs.  Otherwise, if we have a policy that we'll commit the core and then delay the other PRs, there is a chance that the other PRs are never completed.

File: plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ServerEvaluationStrategy.java [hunk 1]
```diff
@@ -61,10 +61,13 @@
      * @param internalAddress
      *         address passed into {@code getServers}; used as fallback if address cannot be
      *         retrieved from containerInfo.
+     * @param serverConfMap
+     *         additional Map of {@link ServerConfImpl}.
```
garagatyi, line 65: It is unclear from javadocs why this additional map is needed. Please explain it in javadocs.

benoitf (changes requested):
While I agree more than 100% on the final goal, I find the implementation of the strategy too specific for a given product.
Let's say we use a different reverse proxy, we would need another pattern, etc

Could it be possible to name it instead of "single-port" something more generic ?

and instead of always using pattern
`ref.workspace-id.che-address`

to allow to pickup the pattern through a list of predefined properties
like in your case it could be
"<server-name>.<workspace-id>.<something>"

but we could have
<che.protocol>://<che-host>:<che-port>/<workspace-id>-<server-port>

or any pattern built upon a set of properties

l0rd:
@garagatyi @benoitf @TylerJewell thanks for your review. I've opened a second PR (#4351) to merge that work in the openshift-connector branch as it is to get it tested/run by our CI. In the meantime I will push more commits on this PR to address your comments and to get that merged to master too.

TylerJewell:
I am going to invite Florent to help us explore how we can get it ready for a master merge.  It is a challenging exercise but one we have in our objectives for the year.

benoitf:
I think we can close this PR
(overlap with https://github.com/eclipse/che/pull/5115 and https://github.com/eclipse/che/pull/5366 + https://github.com/eclipse/che/pull/5052 )

l0rd:
We have merged that on openshift connector branch a long time ago.

PR #4283, opened 2017-03-01
Implement getContainerLogs method in OpenShiftConnector

snjeza:
### What does this PR do?

Implements the getContainerLogs  method in OpenShiftConnector.

### DIFF
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
@@ -11,8 +11,10 @@
 
 package org.eclipse.che.plugin.openshift.client;
 
+import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -32,6 +34,7 @@ import org.eclipse.che.plugin.docker.client.DockerApiVersionPathPrefixProvider;
 import org.eclipse.che.plugin.docker.client.DockerConnector;
 import org.eclipse.che.plugin.docker.client.DockerConnectorConfiguration;
 import org.eclipse.che.plugin.docker.client.DockerRegistryAuthResolver;
+import org.eclipse.che.plugin.docker.client.LogMessage;
 import org.eclipse.che.plugin.docker.client.MessageProcessor;
 import org.eclipse.che.plugin.docker.client.ProgressMonitor;
 import org.eclipse.che.plugin.docker.client.connection.DockerConnectionFactory;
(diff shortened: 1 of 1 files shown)


benoitf:
ci-build

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -606,6 +613,49 @@ public String commit(final CommitParams params) throws IOException {
...
+    @Override
+    public void getContainerLogs(final GetContainerLogsParams params, MessageProcessor<LogMessage> containerLogsProcessor)
+            throws IOException {
+        String container = params.getContainer(); // container ID
+        Pod pod = getChePodByContainerId(container);
+        if (pod != null) {
+            String podName = pod.getMetadata().getName();
+            boolean[] ret = new boolean[1];
+            ret[0] = false;
+            try (LogWatch watchLog = openShiftClient.pods().inNamespace(openShiftCheProjectName).withName(podName)
```
amisevsk, line 625: I think it might be cleaner to move the logs processing into a separate class. `OpenShiftConnector` is getting very long and complicated.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -606,6 +613,49 @@ public String commit(final CommitParams params) throws IOException {
...
+
+                    @Override
+                    public void onClose(KubernetesClientException cause) {
+                        ret[0] = true;
+                    }
+
+                };
+                openShiftClient.pods().inNamespace(openShiftCheProjectName).withName(podName).watch(watcher);
+                Thread.sleep(5000);
+                InputStream is = watchLog.getOutput();
```
amisevsk, line 644: I've had some issues with output from kubernetes-client and using a reader directly. It might be worth waiting for PR #4305 and then reusing the `KubernetesOutputAdapter` class added there. You can then submit a 

```java
InputStreamPumper(watchLog.getOutput(), 
                  new KubenetesOutputAdapter(LogMessage.type.DOCKER, containerLogsProcessor));
```
and it should handle the output correctly.

amisevsk (changes requested):
I think it might be worth modifying this PR to use some of the changes in #4305.

amisevsk:
Also we should merge these changes into the openshift-connector branch instead.

snjeza:
Moved to https://github.com/eclipse/che/pull/4309

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
snjeza: When you merging #4305, I will do it.

PR #4287, opened 2017-03-02
Boolean server property to enable or disable auto start of a workspac…

sunix:
### What does this PR do?
Adding a boolean server property to enable or disable auto start of a workspace when accessing to it through its url
The property could be set with
- `che.workspace.auto_start=true` or `che.workspace.auto_start=false` in che.properties
- `CHE_WORKSPACE_AUTO__START=true` or `CHE_WORKSPACE_AUTO__START=false` in che.env

The default value true doesn't change the current behavior: accessing to the workspace through its URL will start it ( if stopped)
 
https://issues.jboss.org/browse/CHE-116

#### Changelog

Add `che.workspace.auto_start=<bool>` property to set whether URL acceptance auto-starts a workspace 

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
@@ -40,6 +40,9 @@ che.workspace.auto_snapshot=true
 # Otherwise create a new workspace.
 che.workspace.auto_restore=true
 
+# By default, when users access to a workspace with its URL the workspace
+# automatically starts if it is stopped. You can set this to false to disable this.
+che.workspace.auto_start=true
 
 # Workspace threads pool configuration, this pool is used for workspace related
 # operations that require asynchronous execution e.g. starting/stopping/snapshotting
--- a/dockerfiles/init/manifests/che.env
+++ b/dockerfiles/init/manifests/che.env
@@ -262,6 +262,11 @@
 #CHE_WORKSPACE_AUTO__SNAPSHOT=false
 #CHE_WORKSPACE_AUTO__RESTORE=false
 
+#    By default, if a user accesses a workspace from its unique URL (instead of the
+#    dashboard interface), then the workspace is automatically started by Che.
+#    You can set this property to false to disable this behavior.
+#CHE_WORKSPACE_AUTO__START=true
+
 # Private Docker Images
 #     If pushing snap images to a registry requires authenticated access to the
 #     registry. Or, if your stacks reference private images which require authenticated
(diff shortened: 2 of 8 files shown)


TylerJewell (approved):

TylerJewell:
I updated the language in the `che.env`.  This does not need a docs or a release notes comment.

l0rd (approved):

garagatyi (approved):

sunix:
I have failing unit test, will fix it

sunix:
ci-build

benoitf:
ci-build

voievodin:
The properties
```
che.workspace.auto_snapshot
che.workspace.auto_restore
```
are enforced by server, i mean that workspace service decides whether workspace is going to be recovered from snapshot or not. Maybe it makes sense to name the property more certain, so it would be clear from the name that IDE won't start the workspace by default, and the property is not related to API at all.
e.g. `che.workspace.client.auto_start` or `che.workspace.client.ide.auto_start` WDYT?

sunix:
@evoevodin  Not sure, we don't tell if it's going to be implemented client or server side.

The properties

    che.workspace.auto_snapshot
    che.workspace.auto_restore

don't specify if it implemented server or client side.
These are user settings. Users expect a behavior but don't care if it is done server or client side or wherever.
Do you mind if we keep like it is?

TylerJewell:
@evoevodin - I am -1 on the name change.  I thought about something like that and it just introduces unnecessary complexity.  Big goal to keep the names as short as possible.

voievodin (approved):

riuvshin (commented):
those changes may cause failures in codenvy saas and artik, someone from codenvy side must add appropriate PRs to fix those changes

sunix:
I can set default values to missing properties to avoid breaking codenvy, saas, artik assemblies but would need to use google annotations instead of javax @Named and @Inject not sure this is wanted:

    @com.google.inject.Inject(optional = true)
    @com.google.inject.name.Named(CHE_WORKSPACE_AUTO_SNAPSHOT)
    boolean                                     cheWorkspaceAutoSnapshot = true;
    @com.google.inject.Inject(optional = true)
    @com.google.inject.name.Named(CHE_WORKSPACE_AUTO_RESTORE)
    boolean                                     cheWorkspaceAutoRestore  = true;
    @com.google.inject.Inject(optional = true)
    @com.google.inject.name.Named(CHE_WORKSPACE_AUTO_START)
    boolean                                     cheWorkspaceAutoStart    = true;

or do we keep it like this and add the missing attribute in che.properties of codenvy, saas, artik assemblies ?

riuvshin:
@sunix I've added PRs for codenvy and artik can you also please take a look

sunix:
@riuvshin yes it looks good to me let me know when these are merged

riuvshin:
@sunix merged, now you can merge this pr

riuvshin (approved):

PR #4293, opened 2017-03-02
fix env property name

riuvshin:
env var should reflect actual propery name which is: https://github.com/eclipse/che/blob/fix_property/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties#L105

### What does this PR do?
fixes ENV property name

#### Changelog
fix env property name


TylerJewell (approved):

skabashnyuk (commented):
up2 @garagatyi

garagatyi (approved):

PR #4308, opened 2017-03-03
CHE-4141 - Use Persistent Volumes Claims when creating workspaces

snjeza:
Moved from https://github.com/eclipse/che/pull/4142

#### Changelog
Use Persistent Volumes Claims when creating workspaces


l0rd (approved):
Good job @snjeza !

PR #4326, opened 2017-03-06
[cli] - Add check for exec on file system

TylerJewell:
### What does this PR do?
This adds a check for the ability to execute a file on the file system mounted to `:/data`.  Also adds in the ability to skip this check if `--fast` is provided on the command line.  We previously checked to see if the file system is writable, but never checked to see if files could execute on the system. 

Also adds in `--follow` option to the start command which will provide blocking following of Che logs to the output screen. This will not let you know when the server has finished booting. The user must provide CTRL-C on the command line to kill this process.

https://github.com/codenvy/codenvy/issues/1877#issuecomment-284386189

#### Changelog
[cli] Add check to ensure `:/data` can execute files written to it from a container
[cli] Add `--follow` parameter to start command to follow logs while booting

#### Release Notes
In the CLI, as part of the `start` command, you can now optionally provide the `--follow` flag which will provide a blocking follow of the logs of the Che server.

#### Docs PR
https://github.com/eclipse/che-docs/pull/164
https://github.com/codenvy/docs/pull/94

### DIFF
--- a/dockerfiles/base/scripts/base/commands/cmd_start.sh
+++ b/dockerfiles/base/scripts/base/commands/cmd_start.sh
@@ -16,6 +16,7 @@ help_cmd_start() {
   text "Starts ${CHE_MINI_PRODUCT_NAME} and verifies its operation\n"
   text "\n"
   text "PARAMETERS:\n"
+  text "  --follow                          Displays server logs to console and blocks until user interrupts\n"
   text "  --force                           Uses 'docker rmi' and 'docker pull' to forcibly retrieve latest images\n"
   text "  --no-force                        Updates images if matching tag not found in local cache\n"
   text "  --pull                            Uses 'docker pull' to check for new remote versions of images\n"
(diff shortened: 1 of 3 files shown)


dockerfiles/base/scripts/base/startup_02_pre_docker.sh
```diff
@@ -428,7 +428,10 @@ init_check_mounts() {
   fi
 
   # Verify that we can write to the host file system from the container
-  check_host_volume_mount
+  if ! is_fast; then 
+    check_host_volume_mount
+  fi
+return 2
```
garagatyi: Looks like indentation is not alligned

garagatyi (approved):

dockerfiles/base/scripts/base/startup_02_pre_docker.sh
TylerJewell: Oops - indentation is correct.  That return statement should not be there.  Removed.

dockerfiles/base/scripts/base/commands/cmd_start.sh
```diff
@@ -16,6 +16,7 @@ help_cmd_start() {
   text "Starts ${CHE_MINI_PRODUCT_NAME} and verifies its operation\n"
   text "\n"
   text "PARAMETERS:\n"
+  text "  --follow                          Displays server logs to console and blocks until user interrupts\n"
```
garagatyi, line 19: +1 really useful feature

garagatyi (approved):

PR #3405, opened 2016-12-15
Make snapshots binaries removal asynchronous

voievodin:
### What does this PR do?

- Makes snapshots binaries removal asynchronous(while metadata removal is still synchronous)
- Publishes `WorkspaceStatusEvent.RUNNING` after workspace changes status from `SNAPSHOTTING` back to `RUNNING`
- Reaches `WorkspaceStatusEvent` with _previous workspace status_ field
- Introduces small fixes & improvements to workspace codebase

### What issues does this PR fix or reference?
Resolves #3314 & resolves #3258.

PR #4328, opened 2017-03-06
Get rid of recover from snapshot dialog

### LINKED ISSUE #3611: Remove explicit workspace snapshot creation from IDE
- Remove UI controls from the machine perspective _Machine -> Create Snapshot_
- Remove dialog that asks whether to restore from snapshot or not
- Deprecate workspace api methods for creating/getting snapshots information 

So IDE simply uses workspace start method without any additional logic, while server is responsible for restoring workspace when configuration says so and snapshot exists.

voievodin:
### What does this PR do?
Removes recover from snapshot dialog from workspace start sequence. 
The behaviour of IDE is now the same to dashboard, so recovery is enforced 
by either of:
- start workspace request query parameter `restore` 
- workspace attribute `auto_restore` 
- system configuration property `che.workspace.auto_restore` 

The dialog view was:
image

It's still possible to choose whether to recover from snapshot or not after workspace stop:
image

Along with that i turned on snapshot status notifications:
image

image.

Resolves #3611 

#### Changelog
Explicit snapshot creation becomes an internal approach.
Recover from snapshot dialog is removed from IDE.
Workspace REST API methods related to snapshots are deprecated.

#### Release Notes
API methods for creating and getting workspace snapshots are deprecated.
The reason is that snapshotting itself becomes an internal approach which is going to be 
used on workspace stop only.
```
GET  /api/workspace/<id>/snapshot
POST /api/workspace/<id>/snapshot
```
Pull request #3405 introduced a deprecation of `WorkspaceStatusEvent.getEventType()` the alternative would be to use `WorkspaceStatusEvent.getStatus()`, but it turned out that usage of combination of _status_ + _previous status_ + _optional error_ is not always convenient as the usage of the single event type, so it's reverted and not deprecated any more.

#### Docs PR
N/A

As a dialog alternative it could make sense to provide a kind of control like 'checkbox' on dashboard which would allow to decide per workspace whether to snapshot/recover it(discussion is welcome).

### DIFF
--- a/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/workspace/WorkspaceServiceClient.java
+++ b/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/workspace/WorkspaceServiceClient.java
@@ -124,9 +124,8 @@ public interface WorkspaceServiceClient {
      *         if <code>false</code> workspace will not be restored from snapshot
      *         even if auto-restore is enabled and snapshot exists
      * @return a promise that resolves to the {@link WorkspaceDto}, or rejects with an error
-     * @see WorkspaceService#startById(String, String, Boolean, String)
      */
-    Promise<WorkspaceDto> startById(String id, String envName, boolean restore);
+    Promise<WorkspaceDto> startById(String id, String envName, Boolean restore);
 
     /**
      * Stops running workspace.
@@ -273,6 +272,7 @@ public interface WorkspaceServiceClient {
      * @return a promise that will provide a list of {@link SnapshotDto}s, or rejects with an error
      * @see WorkspaceService#getSnapshot(String)
      */
+    @Deprecated
     Promise<List<SnapshotDto>> getSnapshot(String workspaceId);
 
     /**
@@ -283,6 +283,7 @@ public interface WorkspaceServiceClient {
      * @return a promise that will resolve when the snapshot has been created, or rejects with an error
      * @see WorkspaceService#createSnapshot(String)
      */
+    @Deprecated
     Promise<Void> createSnapshot(String workspaceId);
 
     /**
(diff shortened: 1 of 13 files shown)


skabashnyuk (approved):

vparfonov (approved):

mmorhun:
@evoevodin could you please explain how I can start a workspace ignoring its snapshot if auto restoring is turned on and I've closed tab after workspace stop?

voievodin:
> As a dialog alternative it could make sense to provide a kind of control like 'checkbox' on dashboard which would allow to decide per workspace whether to snapshot/recover it(discussion is welcome).

From the API perspective you have 3 configuration options. From the UI no way, unless you go and edit workspace config directly on dashboard.

bmicklea:
@slemeur any opinions?

slemeur:
I see two different use cases:
1 - User should be able to configure the default behavior per workspace
2- User should be able to by-pass the default behavior

**Configuring default behavior per workspace**
In the workspace details: we can add a section to let the user being able to select how it should behave.

**Allowing the user to by-pass the default behavior**
- In the left-sidebar: when doing a right click on the workspace, we could add an option to start the workspace with the different behaviors. 
- In the list of workspace: we could add the same options on the "cog" button
- In the workspace details: we could add an arrow on the right of the "Open". When clicked it would show the different options.

slemeur (approved):

JamesDrummond:
Deprecation requires notifying users prior to implementation. Do we plan to notify user in our release notes before merging?

I think we need to provide what @slemeur describes above before removing this option.

bmicklea:
I don't think this is a deprecation - it's a change in behaviour but not a
deprecation.

We'll need to document it though.
-- 

Brad Micklea | Codenvy | bmicklea@​codenvy.​com | 4​16​.7​07​.07​92

benoitf (approved):

vparfonov (approved):

voievodin:
I've added a workspace attribute `snapshotted_at=1234567890` which is present only if workspace has a snapshot, it points to the time when the snapshot was created. This allows to implement the logic described by the #4337 without getting workspace snapshots from the client.

JamesDrummond (approved):

voievodin:
QA report on the changes

```
[TEST] Development state:
[TEST] Thu Mar 9 14:23:07 EET 2017. Failed tests (0):
[TEST]
[TEST] Comparing with https://ci.codenvycorp.com/view/qa/job/che-integration-tests/250/
[TEST] If a test failed then it is NOT marked as regression.
[TEST]
[TEST] NO REGRESSION, great job!
```

PR #4334, opened 2017-03-06
Improve eclipse ssh agent on unknown Linux stacks

lehmanju:
### What does this PR do?
It improves the situation on custom docker images using a Linux OS which is not listed explicitly in the ssh server script file. (SSH server agent)

Originally I wanted to run a custom docker image with Arch Linux installed (in Codenvy). Currently this does not work because Arch Linux is not supported as docker container. However, the script basically checks whether sshd works or not. On supported systems the package is installed automatically, if it does not work. On unsupported systems the script currently exits regardless of a working sshd server.

This change at least **tries** whether it works or not. If not, you can still exit with an error code.

#### Changelog
Improve behavior of SSH installation script on unknown Linux stacks

### DIFF
--- a/agents/ssh/src/main/resources/org.eclipse.che.ssh.script.sh
+++ b/agents/ssh/src/main/resources/org.eclipse.che.ssh.script.sh
@@ -26,7 +26,7 @@ if [ -f /etc/centos-release ]; then
 fi
 
 ###############################
-### Install Needed packaged ###
+### Install Needed packages ###
 ###############################
 
 # Red Hat Enterprise Linux 7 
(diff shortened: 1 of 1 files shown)


TylerJewell:
@lehmanju - thanks for the contribution.  A maintainer will take a look and consider the details.  Just to confirm, have you filled out the Eclipse CLA and signed your commit?  It looks like the IP validation test is failing so far.

lehmanju:
@TylerJewell - I did sign the ECA on eclipse.org but the error still remains ... What do I have to do to fix that? Thx.

TylerJewell:
Can you ammend the commit so that it contains the "Signed-off-by: Author Name <email@address.com>"?  If you do this, the automatic validation should pick it up.

lehmanju:
**done**

TylerJewell (approved):

TylerJewell:
Fixed - thanks for your contribution!

Issue #3559, opened 2016-12-28
Problems when building new dockerfiles


l0rd:
I'm using: 
- Che version 5.0.0
- Mac OSX 10.11.6
- Docker 1.13.0-rc4

Trying to build centOS based Che I came up against some problems that I'm listing below.

**1/ Bulding dockerfiles/lib fails**
The error is 
```
./lib-typescript/src/api/dto/che-dto.ts(10126,9): error TS2393: Duplicate function implementation.
The command '/bin/sh -c cd /lib-typescript/ && npm install && cd /runtime && npm install && /lib-typescript/node_modules/.bin/tsc --project /lib-typescript/ && mv /lib-typescript/lib /che-lib && cd /lib-typescript/src && find . -name "*.properties" -exec install -D {} /che-lib/{} \;    && rm -rf /lib-typescript && mv /runtime/node_modules /che-lib && rm -rf /runtime' returned a non-zero code: 2
```
to reproduce it run
```bash
cd dockerfiles/lib
./build.sh
```

**2/ Version `nightly` of `che-base` is used when building che-cli `latest`**
Running `dockerfiles/cli/build.sh <tag>` builds image `eclipse/che-cli:<tag>`. But the base image used to build it is always `eclipse/che-base:nightly`. There is no way to use `eclipse/che-base:<tag>` instead.

**3/ I wasn't able to run Che with a version that is not in dockerfiles/cli/version/**
Running `dockerfiles/build-all.sh latest-centos` builds all images with tag `latest-centos`. When I try to start Che using that version I get an error:
```bash
docker run -ti --rm -v /Users/mloriedo/.che:/data -v /var/run/docker.sock:/var/run/docker.sock eclipse/che-cli:latest-centos start
INFO: (che cli): Checking registry for version 'latest-centos' images

We could not find version 'latest-centos'. Available versions:
 4.7.2
 5.0.0-M5
 5.0.0-M6
 5.0.0-M7
 5.0.0-M8
 latest
 latest.ver
 nightly

Set CHE_VERSION=<version> and rerun.
```
And setting CHE_VERSION doesn't help:
```bash 
docker run -e CHE_VERSION=latest-centos -ti --rm -v /Users/mloriedo/.che:/data -v /var/run/docker.sock:/var/run/docker.sock eclipse/che-cli:latest-centos start
```

[...]

Issue #4313, opened 2017-03-05
Generate Che Docker Images with Custom Names & Registries


TylerJewell:
Running Eclipse Che requires a cascading set of Docker images. There are a number of scenarios where the default values of these image names and tags should be changed:
1. Che engineers working in a branch would like dedicated set of images for their branch.
2. Custom assembly generator would like to apply branding to image names
3. CI systems would like to run tests against a branch in isolation and need separate images

We can envision a pattern of Docker image names that match: 
```
<organization>/<prefix>-<name>:<version>
```
where `<organization>` defaults to `eclipse`; `<prefix>` defaults to `che`; `<name>` reflects the image name that we use like `server`, `init`, or `base`; and `<version>` reflects either the provided version or branch name the images are built from. We can have defaults for these values, or let developers override these with command line parameters in the build scripts used for generating Docker images.

To have a tailored set of Docker images to run Che, the following updates must be made:
1. The name of `eclipse/che-base:<version>` must be modified. I think we can assume that the bootstrap and utilities image names that are contained within the base are not allowed to be changed.  TODO: Is this really needed - maybe generation of the base image doesn't need custom naming.

2. The `/dockerfiles/init` docker image would need to be built with updated name.

3. The `/dockerfiles/che` docker image would need to be built with updated name.

4. The `/dockerfiles/cli/version` directory would need to be updated with a new version-specific folder that contains a new `images` file that uses the names generated above.

5. The `/dockerfiles/cli/Dockerfile` would need to be updated to use the new name for the base image in the FROM syntax.  This would only be necessary if re-generation of the base image is done.

6. These Docker images would have to be included in the custom assembly from the archetype generator.

7. We would need to update the build scripts in the `/dockefiles` directory to have a `build-che.sh` script that just builds the images necessary to run che - base, init, che, cli.  This would accept the appropriate flags for setting the organization, prefix, and version names. There would be an option to skip tests.

riuvshin:
do we really want generate this? why don't we just document this out how to prepare docker files for   your custom assembly this is easy. 
Why we should rename things again? this will have a huge affect on everything on on CLI, CI, release and dev processes, we possibly will lose pulling history and etc... for what? 
Maybe I don't understand something but we don't need change any names we have good examples like artik and saas projects which have own CLI @benoitf created that and afaik it was really easy.

TylerJewell:
[...]

PR #4336, opened 2017-03-06
Customize image names during image build

TylerJewell:
### What does this PR do?
This PR tackles the challenge of generating custom image names to run Che in a custom assembly. The issue is discussed here - https://github.com/eclipse/che/issues/4313#issuecomment-284398059.

The build scripts have been modified to generate images of the format `$ORGANIZATION/$PREFIX-$NAME:$TAG`. The default organization is `eclipse`, the default tag is `nightly`, and the default prefix is `che`.  You can proivde `--organization:<org>`, `--prefix:<prefix>`, or `--tag:<tag>` on the command line to customize the images that are generated.

`build-all.sh` script has been renamed to `build.sh`and allow to build the minimum set of images to generate a new Che server (with build.sh <dirs>), the CLI, and unit tests for these custom image names.  If you build the CLI image, we overwrite `/dockerfiles/cli/version/$TAG/images` file to include the customized names that have been provided before this image is built, so the registry within the Docker image is accurate.

https://github.com/eclipse/che/issues/4313

#### Changelog
[cli] Apply custom image names for Che and associated CLI images

#### Release Notes
You can now build Che with custom image names and also generate custom CLI images that reference those custom names. Changing the image name of Che has a number of beneficial use cases including allowing us to create tailored continuous integration flows with custom images for each branch, letting ecosystem developers generate custom image names for their custom assemblies, and engineers can create images tailored to the work they are doing without it conflicting with released images.

The build scripts have been modified to generate images of the format `$ORGANIZATION/$PREFIX-$NAME:$TAG`. The default organization is `eclipse`, the default tag is `nightly`, and the default prefix is `che`.  You can proivde `--organization:<org>`, `--prefix:<prefix>`, or `--tag:<tag>` on the command line to customize the images that are generated.

There is an updated `build.sh <dirs>` script which can now build any number of Docker images. You can provide the list of directories that it should build and it will iterate through each one. Also, builds the minimum set of images to generate a new Che server, the CLI, and unit tests for these custom image names.  If you build the CLI image, we overwrite `/dockerfiles/cli/version/$TAG/images` file to include the customized names that have been provided before this image is built, so the registry within the Docker image is accurate.


riuvshin:
why we need custom images for base images? as I understand we need only custom images for for custom packages like saas which reuse eclipse/* base images.

TylerJewell:
Custom assembly authors want the ability to rename the base images for branding purposes.

riuvshin:
hm... this sounds strange to me, no one wants to put branding on dependencies, authors just using them. I believe that custom assembly authors want put branding on their own images for custom package but definitely not for dependencies. We don;t rename or repackage GWT framework we just use it to build our own product. 
And just imagine how hard for a user will be build own whole stack of images instead of using already built ones as dependencies

TylerJewell:
Then they can skip building other base images.  But this improvement modifies the build scripts to enable changing one or all image names.

benoitf:
Let say we want to publish che images for each pull request/some branches so everyone can locally test the changes
We need for that to build locally the base image so we're sure we are using that image and not the official base image (for example if a pull request is modifying the base image, we need to ensure it is working as well)
But these images, while they will be built, they may not be published

riuvshin:
@benoitf totally agree this is useful! my concern was only about forcing users rebuild whole stack images for "branding" purposes.

benoitf:
@riuvshin AFAIK I think it's separated

TylerJewell:
It is just simpler this way to guarantee mutability.  They are already cloning the source tree and could modify any number of files.  So this guarantees isolation and a clean build.

benoitf:
@TylerJewell maybe instead of having another script `build-che.sh`

we could rename build-all.sh to build.sh and have options to build all images (default) or some subset (like --something-to-define)

TylerJewell:
Yeah that is a good idea florent.  How goes the testing?

riuvshin:
@TylerJewell yeah but only in case where they really need customize base images. I think in most cases they will just use tags that we provide

benoitf:
@TylerJewell In progress. BTW can we remove from this PR the changes on --follow and exec or merge https://github.com/eclipse/che/pull/4326 so that we only have build changes

TylerJewell:
Then they can skip the build all steps and just build a single image.  But I will find debugging and supporting those accounts very difficult.  

It is easier to ask a support case to build all and have them get a custom base, init, che, cli image that passes unit tests.  And then the chain of accountability for debugging is cleaner.

TylerJewell:
I am going to merge 4236 this am.

benoitf:
I’ve done some changes
- Added optional .require files in directories. Allow us to detect dependencies of the image (for example cli is requiring base and bats)
- Add macros on Dockerfile. For example cli image is based on base image, so add macro to be able to change the parent image to be like FROM ${BUILD_ORGANIZATION}/${BUILD_PREFIX}-base:${BUILD_TAG}
- fix the path to version files (when building from top level folder it was not writing in the correct file)
- fix tests that tries to pull images (add skip pull)
- replace build-all.sh to be build.sh with more advanced options
```
     build.sh <list of directories to build>  : build all the specified images and their dependencies
     build.sh : build all the images
```

for example if I need to rebuild cli (with base as dependency) I only need to do on dockerfiles folder `: ./build.sh cli`
If I need to rebuild as well init
`./build.sh init cli`

benoitf:
The new enhancements should fix https://github.com/eclipse/che/issues/3559

benoitf:
- added template for images file
- fix test for checking online dockerhub images

benoitf:
add fixes when checking remote dockerhub on not yet existing images

TylerJewell:
@benoitf - I was able to build + run a custom Che in a Docker image.  +1 on this and ok for it to merge.

benoitf:
should we remove "wip" on PR title ?

TylerJewell:
Yes - it's definitely not wip now

benoitf:
@riuvshin you want to give a try before merging ?

benoitf (approved):

PR #4349, opened 2017-03-07
[WIP] CHE-120 Workspace creation as non-privileged container

ibuziuk:
https://issues.jboss.org/browse/CHE-120

#### Changelog
CHE-120 Workspace creation as non-privileged container

### DIFF
--- a/ide/che-core-ide-stacks/src/main/resources/stacks.json.centos
+++ b/ide/che-core-ide-stacks/src/main/resources/stacks.json.centos
@@ -1,0 +1,2139 @@
+[
+  {
+    "id": "java-mysql",
+    "creator": "ide",
+    "name": "Java-MySQL",
+    "description": "Multi-machine environment with Default Java Stack and MySQL database",
+    "scope": "general",
+    "tags": [
+      "Java 1.8, Tomcat 8, MySQL 5.7"
+    ],
+    "components": [
+      {
+        "name": "JDK",
+        "version": "1.8.0_45"
+      },
+      {
+        "name": "Maven",
+        "version": "3.2.2"
+      },
+      {
+        "name": "Tomcat",
+        "version": "8.0.24"
+      }
+    ],
+    "source": {
+      "type": "image",
+      "origin": "eclipse/ubuntu_jdk8"
+    },
+    "workspaceConfig": {
+      "environments": {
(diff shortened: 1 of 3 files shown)


ibuziuk:
@l0rd am I correct that the plan is renaming `stacks.json.centos` to `stacks.json` before build e.g. we are planning to leave `stacks.json` without modifications

ibuziuk:
FYI, I have also created an issue for publishing images to `registry.centos.org` from **openshift-connector** branch https://issues.jboss.org/browse/CHE-152

l0rd (commented):
Thank you @ibuziuk ! Some comments:
- `stacks.json` should not be modified, we are adding "modified" images just in `stacks.json.centos`
- When modifying file `org.eclipse.che.ws-agent.script.sh` you should modify `org.eclipse.che.ws-agent.json`
- Instead of removing the sudo commands you could add and if condition that checks if the folder already exists (if the folder already exists the sudo command is not executed). With this approach other stacks will continue working.

ibuziuk:
@l0rd Thanks for review. PR has been updated

> stacks.json should not be modified, we are adding "modified" images just in stacks.json.centos

So, we are going to have smth. like that in build instructions : 

```
cd ide/che-core-ide-stacks/src/main/resources
mv stacks.json stacks.json.orig
cp stacks.json.centos stacks.json
cd ../../../../
mvn clean install

cd ide/che-core-ide-stacks/src/main/resources
mv stacks.json.orig  stacks.json
cd ../../../../
```
?

> When modifying file org.eclipse.che.ws-agent.script.sh you should modify org.eclipse.che.ws-agent.json

Hmmm.. do not really understand why this should be done. `org.eclipse.che.ws-agent.json` is just an agen's metadata - I can not figure out what should be changed there

> Instead of removing the sudo commands you could add and if condition that checks if the folder already exists (if the folder already exists the sudo command is not executed). With this approach other stacks will continue working.

Good point - fixed

l0rd:
> Hmmm.. do not really understand why this should be done. org.eclipse.che.ws-agent.json is just an agen's metadata - I can not figure out what should be changed there

You are right. A few months ago the shell script was duplicated in a json file but that should have been fixed now.

l0rd (approved):

PR #4350, opened 2017-03-07
[cli] Add back in check for `--user` on command line

TylerJewell:
In the refactoring to simplify the CLI for 5.4, the `check_user` method was not included and it seems that the original `init()` method was left in tact.  This adds that check back into the bootstrap sequence and removes the redundant method.
 
https://github.com/eclipse/che/pull/3376#issuecomment-284819093

#### Changelog
Add back in check for `--user` on command line

#### Release Notes
N/A


snjeza:
It works fine.

PR #4363, opened 2017-03-08
Implement getVersion method in OpenShiftConnector

snjeza:
What does this PR do?

Implements the getVersion method in OpenShiftConnector.
OpenShift REST API has the version method, but the kubernetes client doesn't implement it. I have created a simple extension that returns version information.

### DIFF
--- a/plugins/plugin-docker/che-plugin-openshift-client/pom.xml
+++ b/plugins/plugin-docker/che-plugin-openshift-client/pom.xml
@@ -36,6 +36,11 @@
             <artifactId>guava</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
+            <artifactId>okhttp</artifactId>
+            <version>3.4.1</version>
+        </dependency>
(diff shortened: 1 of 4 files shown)


plugins/plugin-docker/che-plugin-openshift-client/pom.xml
```diff
@@ -36,6 +36,11 @@
             <artifactId>guava</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.squareup.okhttp3</groupId>
+            <artifactId>okhttp</artifactId>
+            <version>3.4.1</version>
```
skabashnyuk: Guys. I know it's your branch. But could you follow the general rule to not to declare versions in maven submodules. It can save us some time in a future.

snjeza:
The PR requires https://github.com/eclipse/che-dependencies/pull/34

TylerJewell:

benoitf:
work is done as part of https://github.com/eclipse/che/pull/6222

Issue #4353, opened 2017-03-08
"/fakeagent" is already in use by container 


monaka:
Run a lot of eclipse/che:nightly with difference ports at the same time.

Some coitainers can't spawn che-server. Log is like below.

```
Mar 08 08:35:23 che-node-00 dockerd[1056]: time="2017-03-08T08:35:23.950391448Z" level=error msg="Handler for POST /v1.23/containers/create returned error: Conflict. The container name \"/fakeagent\" is already in use by container fffd092867c7e46e801be9b50999f52543b564c55ba6cf4098f6249831b01e6f. You have to remove (or rename) that container to be able to reuse that name."
```

**OS and version:**    

CoreOS 1339.0.0. But maybe OS independent.

TylerJewell:
Hi @monaka - ok, so the startup sequence during the preflight checks uses Docker to start and stop test servers to create simulated connections between a browser and a container.  This container is named `fakeagent`.  

It seems that you must be having some sort of condition that is causing the preflight checks to be started, but our normal cleanup of those preflight checks are not being completed properly.  The cleanup of the tests has us remove the fakeagent.

Or, perhaps if you are starting many containers at the same time, a second container is attempting to run preflight checks with `fakeagent` before the first server that is booting has finished to remove the `fakeagent`.  

So, some options:
1. You can add `--skip:preflight` to skip these checks and this container will never be launched.
2. You can sequence your starting of containers so that you do not start one Che before the others have finished.
3. You can hack the CLI to randomize the container name of `fakeagent` (it's in the `library.sh` file) so that different starts will have different container names to avoid any sort of race condition.

TylerJewell:
I have started a branch with some proposed changes.  These are not well tested as I am on a plane and bandwidth is awful, so use at your own risk right now until they are more broadly covered.

https://github.com/eclipse/che/tree/che-4353

monaka:
IMO, the best solution is "3." variant.
Users will run `eclipse/che` with different `CHE_PORT` in the multi che-server environment.
So I think this issue can be fixed by just adding suffix `-$CHE_PORT` to `fakeagent`.
(This strategy is same as che-server's container name.)

TylerJewell:
Yeah, I agree with the usage of `CHE_PORT`. - committed and tested a solution for this.  
https://github.com/eclipse/che/pull/4378

ghost:
PR merged - closing the issue.

monaka:
Oh..., the issue is still available...

[...]

PR #4378, opened 2017-03-09
[cli] Make preflight test containers with unique name

TylerJewell:
### What does this PR do?
Solves race condition when starting multiple Che servers quickly where the preflight test containers were sharing the same name and running into conflicts. This sets the name of preflight containers to be unique by appending the port of the Che server.

https://github.com/eclipse/che/issues/4353

#### Changelog
[cli] Give containers used in preflight tests unique name
 
#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/dockerfiles/base/scripts/base/library.sh
+++ b/dockerfiles/base/scripts/base/library.sh
@@ -265,13 +265,13 @@ generate_random_string() {
 start_test_server() {
   export AGENT_INTERNAL_PORT=80
   export AGENT_EXTERNAL_PORT=32768
-  export AGENT_CONTAINER_NAME="fakeagent-"$(generate_random_string 10)
+  export AGENT_CONTAINER_NAME="fakeagent-$CHE_PORT"
 
   # Start mini httpd server to run simulated tests
   docker run -d -p $AGENT_EXTERNAL_PORT:$AGENT_INTERNAL_PORT --name $AGENT_CONTAINER_NAME \
              ${BOOTSTRAP_IMAGE_ALPINE} httpd -f -p $AGENT_INTERNAL_PORT -h /etc/ >> "${LOGS}"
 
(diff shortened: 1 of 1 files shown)


dockerfiles/base/scripts/base/library.sh
```diff
@@ -256,21 +256,28 @@ docker_compose() {
   fi
 }
 
+# Return srandom character alphanumeric string (upper and lower case) of length $1
+generate_random_string() {
```
benoitf: I don't see where this function is used ?

TylerJewell: It's not anymore.  My first commit was going to use it with a 10 character code.  And then I decided to use `CHE_PORT` instead as the unique ID.  But I wanted to leave this utility method in the library for future use.

benoitf: could we only add it if we require it ?

TylerJewell: I am sure that is typical, but I can see a number of situations in the near future where we may want to use it.  It wasn't the easiest thing to dig up - so is it inappropriate to leave in unused methods?

benoitf: yes I would say that we only let code really used in master. Dead code is usually deleted.
you may add the code to the pull request so if you need to find it again, you could get it.
Or paste it to a gist

TylerJewell: ok

```
# Return srandom character alphanumeric string (upper and lower case) of length $1
generate_random_string() {
  NEW_UUID=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w $1 | head -n 1)
  echo $NEW_UUID
}
```

benoitf (approved):

Issue #4296, opened 2017-03-02
Webkit font-smoothing bug when the IDE is opened in dashboard iframe


vitaliy-guliy:
It looks Webkit browser has a bug with font rendering when a page is opened in iframe and contains flash object.

The problem with font rendering is on the first screenshot. On the second the flash object is hidden and text is rendered properly.

![font1](https://cloud.githubusercontent.com/assets/1655894/23552203/59eaf776-0023-11e7-982f-559e4099fe8f.png)

![font2](https://cloud.githubusercontent.com/assets/1655894/23515297/6cc09d30-ff72-11e6-8694-b50da00e4d9a.png)

StackOverflow has a discussion about that.
http://stackoverflow.com/questions/11748082/webkit-font-smoothing-bug-when-iframe-with-flash-is-on-page-is-there-a-work-aro

vitaliy-guliy:
Hiding the flash by setting `visibility: hidden` does not allow perform copying.
The better way is to use HTML5 clipboard api.

PR #4388, opened 2017-03-10
CHE-4296 Webkit font-smoothing bug when the IDE is opened in dashboad iframe

vitaliy-guliy:
Fixes rendering of the fonts when the IDE is opened in iframe.

https://github.com/eclipse/che/issues/4296


vitaliy-guliy:
Copy does not work with hidden flash. Need to find another solution.

PR #4390, opened 2017-03-10
Revert "CHE-120 Workspace creation as non-privileged container. Movin…

l0rd:
The PR was still work in progress when I've merged it. Let's revert it since it broke creating a simple java workspace.

#### Changelog
Revert CHE-120 Workspace creation as non-privileged container.


ibuziuk:
@l0rd It was decided not to have stacks.json.centos and just update original one, right ?

l0rd:
@ibuziuk correct

ibuziuk (approved):

Issue #4076, opened 2017-02-09
Disable forcing file saving on editor close operation


vzhukovs:
There is a proposal to change behavior of closing editor. The last one doesn't check whether opened file is dirty or not and force file saving. This type of operation is redundant and may be removed.

Partially fix this issue: https://github.com/eclipse/che/issues/3891

PR #4393, opened 2017-03-10
Trigger file saving operation on editor closing only when file has changes

vzhukovs:
### What does this PR do?

This changes proposal disable automatically file saving on editor closing.

The previous version doesn't care about whether file changed or not and automatically fired file saving. New behavior takes care about current opened file, and if file is dirty then fire file saving.

Signed-off-by: Vladyslav Zhukovskii <vzhukovskii@codenvy.com>

https://github.com/eclipse/che/issues/4076

#### Changelog
Trigger file saving operation on editor closing only when file has changes

#### Release Notes
Trigger file saving operation on editor closing only when file has changes

#### Docs PR
N/A


vzhukovs:
Selenium testing wasn't found any regression

vparfonov (approved):

Issue #4381, opened 2017-03-09
Workspace agent script should not contain sudo commands


ibuziuk:
`org.eclipse.che.ws-agent.script.sh` contains several sudo commands for  java / tar / curl  package installations and creation of the `/projects` dir. Those sudo commands can not be executed on OpenShift online, so we had to patch stack's Dockerfile (for now is done only for vertx [1] ) in order to include packages required by ws-agent and create `/projects` directory. By doing this, no sudo commands would be run as part of the ws-agent script. This ws-agent "sudo cleanup" could be done relatively easy step-by-step without breaking stacks that would still require sudo execution from ws-agent :

- The following check should be added to the ws-agent script [2]

```
if [ ! -d "/projects" ]; then
    ${SUDO} mkdir -p /projects
    ${SUDO} sh -c "chown -R $(id -u -n) /projects"
fi
```
- Gradually stacks' Dockerfile would be patched with required instructions and both versions of stacks (sudo / non-sudo) would  be supported by the ws-agent script.  

[1] https://github.com/eclipse/che-dockerfiles/pull/70 
[2] https://github.com/eclipse/che/pull/4349/files#diff-1bf4c5348e6dab351e2402e9888727aeR39

ghost:
@tolusha can you comment on consequences of removing sudo from the script?

l0rd:
@eivantsov @tolusha I don't think that we want to remove sudo from the script. The goal is to avoid executing sudo commands if these are not needed (e.g. if the image has already the packages and if the folders are already created).

l0rd:
And then adapt "official stacks" to avoid sudo commands.

The benefit would be that we can run a workspaces without the need to have root privileges in container (that's a good practice).

ghost:
@l0rd of course all stacks that Che provides has required deps for a ws agent and terminal. AFAIK, the script checks if the required packages are installed and if they are, sudo gets never executed.

l0rd:
@eivantsov ok so we all agree that we should avoid (when it's possible) to execute sudo commands.

That's a [sudo command](https://github.com/eclipse/che/blob/master/wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh#L37) that is always execuded isn't it? And the vert.x stack didn't had curl and tar so there were 2 more sudo commands needed. We are talking about improving that.

vpavlin:
[...]

PR #4395, opened 2017-03-10
CHE-4381 Do not create '/projects' dir if already exists

ibuziuk:
### What does this PR do?
Do not create '/projects' dir if already exists via workspace agent script

https://github.com/eclipse/che/issues/4381

#### Changelog
 Do not create '/projects' dir if already exists via workspace agent script

#### Release Notes
n/a

#### Docs PR
n/a

### DIFF
--- a/wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
+++ b/wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
@@ -34,8 +34,11 @@ fi
 MACHINE_TYPE=$(uname -m)
 
 mkdir -p ${CHE_DIR}
-${SUDO} mkdir -p /projects
-${SUDO} sh -c "chown -R $(id -u -n) /projects"
+
+if [ ! -d "/projects" ]; then
+    ${SUDO} mkdir -p /projects
+    ${SUDO} sh -c "chown -R $(id -u -n) /projects"
+fi
 
 
 INSTALL_JDK=false


ibuziuk:
After some tests, it looks like the fix is not working as expected - "projects" dir is already created before the ws-agent script is involved (probably mounted as volume) for stacks like `ubuntu_jdk8`

tolusha:
@ibuziuk 
I can start workspace but  I can't create any projects

ibuziuk:
@tolusha thanks for the tests. Could you please clarify against what stacks have you created a workspace (is it vertx one ) ? I think that it is not possible to create a project due to permissions issue - 
Basically, /projects is mounted as volume [1] and `
${SUDO} sh -c "chown -R $(id -u -n) /projects"` is never called because `/projects` dir already exists by the time ws-agent script is called

[1] https://github.com/eclipse/che/blob/master/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalCheInfrastructureProvisioner.java#L97-L97

tolusha:
@ibuziuk 
It was a `default` stack based on `eclipse/ubuntu_jdk8`

TylerJewell:
@ibuziuk - any update on this - would be nice to include this improvement in the release that requires merge at the end of this week.

l0rd:
@ibuziuk I think that this PR was needed to run Che on OpenShift and is not needed anymore. Am I right?

ibuziuk:
This one should be closed. Basically, it was pushed only in `openshift-connector` branch and reverted short after it. Proper fix with addition of arbitrary user check[1] (user with random UID without sudo right) is part of "Openshift connector improvements" PR https://github.com/eclipse/che/pull/5052

[1] https://github.com/eclipse/che/pull/5052/files#diff-1bf4c5348e6dab351e2402e9888727aeR49

Issue #4249, opened 2017-02-28
Remove redundant VFS layer


vparfonov:
The benefits that the VFS provided no more actual for current logic and architecture. There is no need for multi tenancy support as long as we're running file system on a single workspace agent instance. 
We are not using/not implemented most of file system items attributes facilities.   As long as any user can intrude a file system via ssh we cannot effectively use file system locks provided by VFS.
We cannot rely on index update routines as we're not tracking external file system events but only track VFS events.

The removal of virtual file system layer will give us the following benefits:
- Removing extra layer - simplified logic, less source cod.
- Reuse of JDK file related libraries - more reliable code
- Make file/text searcher based on file watchers - make index more stable
- Remove synchronization required for multi tenancy - performance improvements
- The general performance and architecture impact is yet to be investigated and analyzed

PR #4400, opened 2017-03-13
VFS API deprecation

dkulieshov:

### What does this PR do?
Sets deprecation on major VFS components, VFS components are now considered deprecated and will be replaced by standard JDK routines.

The benefits that the VFS provided no more actual for current logic and architecture. There is no need for multi tenancy support as long as we're running file system on a single workspace agent instance. 
We are not using/not implemented most of file system items attributes facilities.   As long as any user can intrude a file system via ssh we cannot effectively use file system locks provided by VFS.
We cannot rely on index update routines as we're not tracking external file system events but only track VFS events.

The removal of virtual file system layer will give us the following benefits:
- Removing extra layer - simplified logic, less source cod.
- Reuse of JDK file related libraries - more reliable code
- Make file/text searcher based on file watchers - make index more stable
- Remove synchronization required for multi tenancy - performance improvements
- The general performance and architecture impact is yet to be investigated and analyzed

https://github.com/eclipse/che/issues/4249

#### Changelog
VFS components are now considered deprecated and will be replaced by standard JDK routines.

### DIFF
--- a/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/VirtualFile.java
+++ b/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/VirtualFile.java
@@ -26,6 +26,7 @@ import java.util.Map;
  *
  * @author andrew00x
  */
+@Deprecated
 public interface VirtualFile extends Comparable<VirtualFile> {
     /**
      * Gets name.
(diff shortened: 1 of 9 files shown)


slemeur:
The PR description must explain the purpose of deprecating the APIs (by reminding explanations from #4249) and by what they are going to be replaced.

svor (approved):

vparfonov (approved):

Issue #2847, opened 2016-10-20
Packaging Che to run on OpenShift


TylerJewell:
This is a specification for all of the work to be done to productize Eclipse Che to run and operate on  Kubernetes-based OpenShift.  This is a particularly challenging task as we have to decouple the internal Docker management operations of Che into an abstraction layer that can delegate those functions into an OpenShift provider, which itself must map those commands to OpenShift APIs.  Additionally, OpenShift was originally designed for applications that are fairly well self-contained, but Che is an application which itself is also responsible for creating and managing containers, and needs to have certain control hooks into the underlying orchestration system.  

@l0rd - has started on a prototype that has programmatic controls over how to create and destroy workspaces using only OpenShift APIs. This is the work product of what is needed to productize all of this work.

- [x] Che core needs to implement Service Provider Interface which abstracts core Docker functions to delegate to a provider. We will have providers for Docker & OpenShift initially. A swarm provider is likley as Codenvy runs itself on a Swarm implementation.  #2355
>
- [ ] Enabling Che workspaces to launch, run, and stop within a OpenShift pod.
  - [x] Use OpenShift API instead of Docker API to start and stop Che workspaces #3798
  - [x] Switch between OpenShift API and Docker API using a property in che.properties #3798
  - [x] Add the needed OpenShift connector dependencies https://github.com/eclipse/che-dependencies/pull/23
  - [ ] Extend Che CLI to generate an OpenShift Template and deploy it
>
- [x] (*) Remove --privileged container constraint for running Che workspaces in OpenShift.  Privileged flag is required for any situation where we are not using the OpenShift APIs, we are exposing services on ephemeral ports, we are using host path volumes.
  - [x] (*) Replace all the remaining call to the Docker API with OS API #3031
  - [x] (*) Use Persistent Volumes Claims
  - [x] (*) Workspaces services should be exposed on port 80 or 443 only
>
- [x] Documentation and marketing updates: 
  - [x] Deployment of Che on OpenShift https://github.com/codenvy/che-docs/issues/76
  - [x] Custom assembly and template for running Che with OpenShift plugin (also deploying into OS)
  - [x] Marketing sub-site at www.eclipse.org/che to treat the OS custom assembly as a specialty product

(*) work has been merged in [openshift-connector](https://github.com/eclipse/che/tree/openshift-connector) branch only.

vazexqi:
[...]

PR #4411, opened 2017-03-13
[OpenShift Integration] Add implementation of getEvents() to avoid busy wait

amisevsk:
### What does this PR do?
Add implementation of getEvents() to OpenShiftConnector. The added method simply discards events as they arrive, but can (and should) be extended to convert OpenShift events into Docker events to be tracked by Che in the future. 

Adding the no-op implementation avoids the infinite loop of the previous empty method, lowering CPU usage.

Component of #2847.

#### Changelog
- Reduce CPU usage when running on OpenShift

#### Release Notes
- N/A


l0rd (approved):
Thanks @amisevsk !

PR #4418, opened 2017-03-14
ARTIK-318: Add close to JGit clone operation

vinokurig:
If Jgit clone command is executed without close, Jgit process will capture `pack` file in `.git` folder.
As a result workspace folder is not deleted after deleting workspace in IDE on Windows OS.

### What does this PR do?
Add `close` operation to result of Jgit's clone command execution to free resources.

https://github.com/codenvy/artik/issues/318

#### Changelog
Fix bug when workspace folder is not deleted after deleting workspace in IDE on Windows OS.

#### Release Notes
Bugfix N/A

#### Docs PR
Bugfix N/A

### DIFF
--- a/wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
+++ b/wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
@@ -514,7 +514,7 @@ class JGitConnection implements GitConnection {
                 }
             });
 
-            executeRemoteCommand(remoteUri, cloneCommand, params.getUsername(), params.getPassword());
+            ((Git)executeRemoteCommand(remoteUri, cloneCommand, params.getUsername(), params.getPassword())).close();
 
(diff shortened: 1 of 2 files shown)


ghost (approved):

wsagent/che-core-git-impl-jgit/src/main/java/org/eclipse/che/git/impl/jgit/JGitConnection.java
```diff
@@ -514,7 +514,7 @@ protected void onEndTask(String taskName, int workCurr, int workTotal, int perce
                 }
             });
 
-            executeRemoteCommand(remoteUri, cloneCommand, params.getUsername(), params.getPassword());
+            ((Git)executeRemoteCommand(remoteUri, cloneCommand, params.getUsername(), params.getPassword())).close();
```
skabashnyuk: It looks like this is a wrong place to do that. 
Are you sure?

skabashnyuk (approved):

sleshchenko (approved):

PR #4424, opened 2017-03-14
Implement getSystemInfo method in OpenShiftConnector

snjeza:
What does this PR do?

Implements the getSystemInfo method in OpenShiftConnector.

The PR implements getSystemInfo partially.
The method requires the cluster-admin role which has already been added for the che service account.

This method isn't called anywhere in che. In order to test it, you need to debug che and call the method manually.

### DIFF
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
@@ -61,6 +61,7 @@ import org.eclipse.che.plugin.docker.client.json.ImageInfo;
 import org.eclipse.che.plugin.docker.client.json.NetworkCreated;
 import org.eclipse.che.plugin.docker.client.json.NetworkSettings;
 import org.eclipse.che.plugin.docker.client.json.PortBinding;
+import org.eclipse.che.plugin.docker.client.json.SystemInfo;
 import org.eclipse.che.plugin.docker.client.json.network.ContainerInNetwork;
 import org.eclipse.che.plugin.docker.client.json.network.Ipam;
 import org.eclipse.che.plugin.docker.client.json.network.IpamConfig;
@@ -100,6 +101,8 @@ import org.slf4j.LoggerFactory;
 
 import io.fabric8.kubernetes.api.model.Container;
 import io.fabric8.kubernetes.api.model.ContainerBuilder;
+import io.fabric8.kubernetes.api.model.Node;
+import io.fabric8.kubernetes.api.model.NodeSystemInfo;
 import io.fabric8.kubernetes.api.model.PersistentVolumeClaim;
 import io.fabric8.kubernetes.api.model.PersistentVolumeClaimBuilder;
 import io.fabric8.kubernetes.api.model.PersistentVolumeClaimList;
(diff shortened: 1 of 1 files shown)


plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -1428,4 +1432,38 @@ private boolean isTerminalAgentInjected(final Set<String> exposedPorts) {
     private boolean isDevMachine(final Set<String> exposedPorts) {
         return exposedPorts.contains(CHE_WORKSPACE_AGENT_PORT + "/tcp");
     }
+
+    @Override
+    public SystemInfo getSystemInfo() throws IOException {
+        PodList chePods = openShiftClient.pods().inNamespace(this.openShiftCheProjectName)
+                .withLabel(OPENSHIFT_DEPLOYMENT_CONFIG_LABEL, OPENSHIFT_DEPLOYMENT_CONFIG_VALUE).list();
```
amisevsk: Is this step necessary? I feel like explicitly checking for `deploymentconfig=che` is prone to error if the deploy script is changed. Is it sufficient to just do `openShiftClient.pods().inNamespace(...).list()`?

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -1428,4 +1432,38 @@ private boolean isTerminalAgentInjected(final Set<String> exposedPorts) {
...
+            Node node = openShiftClient.nodes().withName(pod.getSpec().getNodeName()).get();
+            NodeSystemInfo nodeInfo = node.getStatus().getNodeInfo();
+            SystemInfo systemInfo = new SystemInfo();
+            systemInfo.setKernelVersion(nodeInfo.getKernelVersion());
+            systemInfo.setOperatingSystem(nodeInfo.getOperatingSystem());
+            systemInfo.setID(node.getMetadata().getUid());
+            int containers = 0;
+            List<Pod> pods = openShiftClient.pods().inNamespace(this.openShiftCheProjectName).list().getItems();
+            for (Pod p:pods) {
+                containers += p.getSpec().getContainers().size();
```
amisevsk: In our current implementation, we are using 1 container/pod, so `containers` should just be equal to the number of pods in the eclipse-che namespace.

amisevsk (changes requested):
Thanks @snjeza! Just a few comments

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -1428,4 +1430,34 @@ private boolean isTerminalAgentInjected(final Set<String> exposedPorts) {
     private boolean isDevMachine(final Set<String> exposedPorts) {
         return exposedPorts.contains(CHE_WORKSPACE_AGENT_PORT + "/tcp");
     }
+
+    @Override
+    public SystemInfo getSystemInfo() throws IOException {
```
amisevsk: I think we should move the method up, to come right after `startExec()`. That way, the code is easier to read and all public methods are grouped together.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -1428,4 +1430,34 @@ private boolean isTerminalAgentInjected(final Set<String> exposedPorts) {
...
+            Pod pod = chePods.getItems().get(0);
+            Node node = openShiftClient.nodes().withName(pod.getSpec().getNodeName()).get();
+            NodeSystemInfo nodeInfo = node.getStatus().getNodeInfo();
+            SystemInfo systemInfo = new SystemInfo();
+            systemInfo.setKernelVersion(nodeInfo.getKernelVersion());
+            systemInfo.setOperatingSystem(nodeInfo.getOperatingSystem());
+            systemInfo.setID(node.getMetadata().getUid());
+            int containers = 0;
+            List<Pod> pods = openShiftClient.pods().inNamespace(this.openShiftCheProjectName).list().getItems();
+            containers += pods.size();
```
amisevsk: These three lines can be merged into just one:

``` java
int containers = openShiftClient.pods().inNamespace(this.openShiftCheProjectName).list().getItems().size();
```

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -1428,4 +1430,34 @@ private boolean isTerminalAgentInjected(final Set<String> exposedPorts) {
...
+            int images = node.getStatus().getImages().size();
+            systemInfo.setContainers(containers);
+            systemInfo.setImages(images);
+            systemInfo.setName(node.getMetadata().getName());
+            String[] labels = new String[node.getMetadata().getLabels().size()];
+            final int[] i = {0};
+            node.getMetadata().getLabels().forEach((k,v)->
+                labels[i[0]++] =  k + "=" + v
```
amisevsk: WDYT about using streams here? 

```java
String[] labels = node.getMetadata()
                      .getLabels()
                      .entrySet().stream().map(e -> String.format("%s=%s", e.getKey(), e.getValue()))
                      .toArray(String[]::new);
```

amisevsk (changes requested):
Sorry to bring it up again, a few more things.

snjeza:
@amisevsk 
I have updated the PR.

amisevsk (approved):

TylerJewell:

benoitf:
work is done by https://github.com/eclipse/che/pull/6218

Issue #4361, opened 2017-03-08
Use a single port to route all traffic


slemeur:
# Goals
Along with the work done for the [integration with Openshift](https://github.com/eclipse/che/pull/4269), we need to unify this mecanism for Che.

The idea would be to embed a reverse proxy in order to allow the single-port strategy.
This will likely simplify the administration and configuration of Che/Codenvy by reducing the number of port necessary using a single one.

We would also let the user being able to configure the port to get exposed .

### Linked Issues:
https://github.com/eclipse/che/issues/1560
https://github.com/eclipse/che/pull/4269
https://github.com/eclipse/che/issues/3037

slemeur:
Linking with Traefik issue: https://github.com/containous/traefik/pull/1257

benoitf:
for Che, it's done

slemeur:
Excellent !

PR #4440, opened 2017-03-15
[Wip] Introduce reverse proxy for all http protocol + custom template server strategy

benoitf:
## What does this PR do?
Introduce reverse proxy for all http protocol and custom server strategy
- traefik plugin register docker label into the docker container
- custom strategy redirect to traefik hosts

apply machine name and workspace id labels to all containers
reverse-proxy

#4361 

#### Changelog
[cli] Add support for launching traefik reverse proxy for traffic routing through single port
Add new server evaluation strategy to channel all non-SSH workspace activity through a single port
Add Traefix plugin to notify Traefik servers with configuration updates for workpace events

#### Release Notes
Due to popular demand, you can now setup Che and all of its workspaces to operate all traffic through a single port. Previously, a browser would communicate to Che, each workspace agent, and the servers users run in a workspace on different ports. This required environments behind a firewall to open a number of ports, which poses security issues for certain accounts.

You can now set `CHE_DOCKER_SERVER__EVALUATION__STRATEGY` in the `che.env file to `single-port` or `no-proxy`. By setting it to `single-port`, all browser traffic to Che or any workspace will be routed through the value that you have set to `CHE_PORT`, or 8080 if not set. Setting this property will transform the launch sequence of Che to launch a Traefik reverse proxy. The reverse proxy will act as the traffic end point for all browser communications. When a new workspace is started or stopped, Che will update Traefik's configuration with rules for how browser traffic should be routed to Che or a workspace.

The `no-proxy` evaluation strategy makes all traffic into a form that has the encoding of where each workspace, server, and container is located behind the proxy.
```
<serverName>.<machineName>.<workspaceId>.<wildcardNipDomain>:<chePort>
```

If you set the evaluation strategy to `no-proxy`, this sets Che to the older behavior where there is no Traefik proxy and all traffic is not routed on different ports.

#### Docs PR
WiP

Change-Id: Id2fa9fb4764f2c049a0b81601892e855ce019828


dockerfiles/init/modules/compose/templates/docker-compose.yml.erb
```diff
@@ -23,16 +23,36 @@ che:
...
 <% if scope.lookupvar('che::che_user') != 'root' -%>
   user: <%= scope.lookupvar('che::che_user') -%>
 <% end -%>
 <% if ! @dns_resolvers.empty? -%>
 <%= "  dns:" + "\n" + @dns_resolvers.split(",").map { |val| "    - #{val}" }.join("\n") %>
 <% end -%>
+
+
+traefik:
+  image: florentbenoit/traefik
```
skabashnyuk: any plans to reuse this image https://hub.docker.com/_/traefik/ ?

dockerfiles/init/modules/traefik/templates/traefik.toml.erb
```diff
@@ -0,0 +1,89 @@
...
+
+#[web]
+#address = ":7070"
+
+[entryPoints]
+  [entryPoints.http]
+  address = ":8080"
+
+[docker]
+watch = true
```
skabashnyuk: new line?

dockerfiles/init/modules/compose/templates/docker-compose.yml.erb
benoitf: yes sure, it's still a WiP
it was to allow me to use Traefik with my PR merged
when Traefik will release 1.3.x version I will switch in my PR

TylerJewell (changes requested):
Can you elaborate a little bit on why we needed a custom plugin for traefik directly?  I understand the role of the custom server evaluation strategy (nicely implemented, BTW).  But why also the additional traefix plugin?  Is it something about updating the traefix server with a new configuration from within Che?

If that is "yes", could there be any sort of race conditions where the browser attempts to do something before Che has had a chance to update traefik with a new configuration?

dockerfiles/init/modules/compose/templates/docker-compose.yml.erb
```diff
@@ -23,16 +23,36 @@ che:
...
+  command: --logLevel=DEBUG
+  links:
+    - 'che:che'
+  labels:
+    traefik.enable: "false"
+  ports:
+    - '<%= scope.lookupvar('che::che_port') -%>:<%= scope.lookupvar('che::che_port') -%>'
+    - '7070:7070'
```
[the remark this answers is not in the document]
benoitf: It's to reach admin console ( if enabled in config file)

benoitf:
Traefik plugin is adding docker labels into docker container before container is started
These labels are used by Traefik to handle routes dynamically ( container is stopped, routes are deleted)
So there is no race condition. Routes are added before browser try to use them

TylerJewell:
I believe that the release notes that I drafted would also largely act as the documentation update that we need to provide into che-docs once we are ready to merge this.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -209,10 +209,14 @@ che.docker.ip.external=NULL
 #   - 'default':      internal address is address of docker host and ephemeral port are used
 #   - 'docker-local': internal address is address of container within docker network, and exposed ports
 #                     are used.
+#   - 'custom': The evaluation strategy may be customized through a template property.
 # The 'docker-local' strategy may be useful if a firewall prevents communication between che-server and
 # workspace containers, but will prevent communication when che-server and workspace containers are not
 # on the same Docker network.
-che.docker.server_evaluation_strategy=default
+che.docker.server_evaluation_strategy=custom
```
garagatyi, line 218: Looks like you change default behavior of Che to work behind concrete proxy. 
It is really backward incompatible change. I think we should consider it as an additional feature instead of mandatory part of Che.
Some companies might want to embed Che into their system and don't have to use that proxy.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -209,10 +209,14 @@ che.docker.ip.external=NULL
...
+
+che.docker.custom.external.template=<serverName>.<machineName>.<workspaceId>.<wildcardNipDomain>:<chePort>
```
garagatyi, line 220: This template doesn't provide list of macros that can be used to format the link. Can you add it here?

Also I think there are some other values that can be used here, like exposed port, published port, etc.

In addition to comment about changing default behavior looks like used pattern will require wildcard DNS. Not every user has/wants dns resolution control.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -209,10 +209,14 @@ che.docker.ip.external=NULL
...
+
```
garagatyi: I would recommend to replace `workspaceId` with `workspaceName` since it is more user friendly

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
```diff
@@ -0,0 +1,335 @@
...
+            return String.format("%s.%s", getExternalIp(externalAddress), "nip.io");
+        }
+
+        /**
+         * Gets a Wildcard domain based on the ip using an external provider xip.io
+         *
+         * @return wildcard domain
+         */
+        protected String getWildcardXipDomain(String externalAddress) {
+            return String.format("%s.%s", getExternalIp(externalAddress), "xip.io");
```
garagatyi, line 392: Does it use some cloud service or I missed something?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
```diff
@@ -0,0 +1,335 @@
...
+                         .findFirst().get();
+        }
+
+        /**
+         * Gets the workspace Machine Name from the config of the given container
+         *
+         * @return machine name of the workspace
+         */
+        protected String getMachineName() {
+            return Arrays.stream(env).filter(env -> env.startsWith(CHE_MACHINE_NAME_PROPERTY))
```
garagatyi, line 359: What do you think about usage of labels instead of env variables. I think it may be better, but I'm not sure.

garagatyi (changes requested):
This PR is complex and changes a lot.
Probably it would require several review session from me and other engineers to review it carefully.
Also I don't feel like changing default port exposure method. If this is not mockup but real contribution into master I would want to discuss that with you, @TylerJewell and @gazarenkov

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
benoitf: issue is that we haven't the workspaceName value when creating containers (only workspaceId)

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
benoitf: Thanks for the review, it's still in progress.
I may consider to be able to replace all existing strategies by this one (I mean, that user can have fine grained strategy)
so it may include exposed port, published port, etc

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
benoitf: For some properties I agree labels is better
But I may think it's nice to have machine Name as env variable
For example, I may use the machine name in my prompt terminal instead of the default `$`character

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
garagatyi: Oh, then it is OK to keep it default for the sake of simplicity of development

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
benoitf: if xip or nip is used, it uses a remote service.
but user can configure its own dns (adding wildcard dns entries) and then it can goes away from xip, nip, etc.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
garagatyi: At least do not remove possibility to bind another strategy. Some vendors might want to use something very specific for their needs.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
garagatyi: Maybe we should include it. But it can be done as an improvement and is not related to this contribution

benoitf:
> Also I don't feel like changing default port exposure method. If this is not mockup but real
> contribution into master I would want to discuss that with you, @TylerJewell and @gazarenkov

could you explain what you're meaning by "I don't feel like changing default port exposure" ?

If your concern is about having by default reverse proxy, for my current wip branch this is the case so there is no setup. But we may change it to be enabled upon option.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
garagatyi: I don't mean to not add env variables. Just to consider usage labels instead of env vars for proxy and other infrastructural needs

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
benoitf: I think workspaceName won't be enough, we should include namespace as well (or for you the workspace name includes the namespace)
Because two ppl can have the same workspace name (on different workspace) and in that case it won't work as it will provide host collision

garagatyi:
I meant to not require proxy in default Che packaging (especially native Che). You said that it is for the sake of simplicity in development so I'm ok with that.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
benoitf: ah for sure, for now yes I don't want to remove existing strategies or change the default
but removal of prepackaged strategies could be done in a next major release if we can tune strategies only by customizing properties. (BTW this is a different story)

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
garagatyi: Yes I know, but it might be better for UX because WS ID is not really readable.

benoitf:
@garagatyi ok yes it is for simplicity on my branch. Compile and run

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
garagatyi: ok, thanks for the explanation

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
benoitf: I'm already using labels for Traefik. Only machine name is added as env variable.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
benoitf: I fully agree with you. I had the same feeling when at first I wanted to use namespace/workspaceName but I only found workspaceId in containers.

`ws-agent.devMachine.alexander.java8.my-domain.com` is far better than `ws-agent.devMachine.workspace129898793.my-domain.com`

I just didn't wanted to make big changes to existing code to add namespace/ workspace name as container labels.
It doesn't impact the reverse proxy mechanism, only introducing new configuration properties. So we can iterate on that.

benoitf:
>This PR is complex and changes a lot.
>Probably it would require several review session from me and other engineers to review it carefully.
>Also I don't feel like changing default port exposure method. If this is not mockup but real >contribution into master I would want to discuss that with you, @TylerJewell and @gazarenkov

@garagatyi this is why I've started to move some code of this PR to new small PRs so it's easy to review (like the new one about WORKSPACE_ID and MACHINE_NAME env variables)

garagatyi (commented):
The interceptor is applied even when it is not needed. This means that interceptor does a lot of things and applied by default even when server evaluation strategy differs.
Because of usage of interceptor it is not possible to configure server evaluation strategies from single place (probably properties). It also uses configuring by interceptor binding.
This can lead to different errors from Che maintainers and those who would want to extend Che.

Because all of this I'm -1 on usage of interceptor for this case. 
Please consider another way of alternating of evaluation strategy.

garagatyi:
@benoitf I have one idea related to this PR. Looks like you need interceptor only because you don't really know all the exposed ports and image labels. It prevents you from figuring out the whole list of references of servers. 
So idea is to NOT use ports and labels that are not part of workspace definition. It is unlikely that someone will use them in Che because they are usually inherited from parent images.
So if we agreed to not publish exposed ports that are not defined as server in workspace config then we don't need to inspect docker image. [One](https://github.com/moby/moby/issues/13322#issuecomment-226958010) might treat running unknown images with unknown ports as not very clever thing and I kinda agree.

So all you need is to add additional labels into container. List of labels is based on workspace definition. Looks like [infrastructure provisioner](https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/environment/server/InfrastructureProvisioner.java) can help. I'm not sure that current implementations of provisioner is applicable in that case, but I think it can be rewritten. 
And, of cause, there is another, pretty clean, solution. Usage of API of proxy from component that rewrites URLs. It knows runtime state, it can access proxy API to setup needed rules.

benoitf:
@garagatyi I need interceptor to be able to define labels before container is started (so Traefik is able to get that information from the container)

>  It is unlikely that someone will use them in Che because they are usually inherited from parent images. So if we agreed to not publish exposed ports that are not defined as server in workspace config then we don't need to inspect docker image.

I think your assumption is wrong. If I inherit from a Tomcat image I would use the 8080 ports exposed in the parent image. (without having to add custom che metadata)
So I cannot agree as your proposal is not a global solution as it removes support of valid use-cases.


> So all you need is to add additional labels into container. List of labels is based on workspace >definition. Looks like infrastructure provisioner can help. I'm not sure that current implementations >of provisioner is applicable in that case, but I think it can be rewritten.

I use the current SPI. Currently it's not possible. If one day it is possible, I will upgrade or migrate.
Interceptors are, in the current state (note that the PR is almost a couple of month delayed...) the only working solution.


@garagatyi your suggested "pretty clean" solution on"API proxy" doesn't work for Che. (usage of API proxy) as labels need to be added to the container before the container is started. (Traefik/Docker setup)
I'm using API only in codenvy as we already have a zookeeper container to manage data. 

Also, interceptor  binding or infrastructure bindings happen through their module configuration. So it will not be faster. 

It seems you dislike implementation but not for a technical reason. If current SPI is not so "SPI friendly" it shouldn't block new features for 2 months until it is enhanced or rewritten.

garagatyi:
> I need interceptor to be able to define labels before container is started 

I meant that with suggested approach it is not needed at all. 

> I think your assumption is wrong. If I inherit from a Tomcat image I would use the 8080 ports exposed in the parent image.

If someone wants to run his project in product X it is common to use concepts of product X to configure his project. What if I use tomcat base image but don't want to expose port 8080 because it is not secured. I just can't stop Che publishing it. And If I could decide which ports should be opened I would 
 be able to control what I want to expose to the world. 
Docker compose doesn't have publish all ports in config. User of compose HAS to set which ports should be published. So I wouldn't say that assumption is wrong - it just differ from your POV.

> your suggested "pretty clean" solution on"API proxy" doesn't work for Che. (usage of API proxy) as labels need to be added to the container before the container is started.

It supposed to NOT use labels, but proxy API instead. I suggested this option as usage of interceptor even when strategy is not used looks a bit odd to me.

> It seems you dislike implementation but not for a technical reason.

I submitted my concerns about interceptors, so I don't see the reason to think so.

>  If current SPI is not so "SPI friendly" it shouldn't block new features for 2 months until it is enhanced or rewritten.

I wonder why you see any connection between new SPI and this feature. I didn't state that new SPI needed for that feature.  Also I added my recent comments to try to help with further development of this feature.

benoitf:
> If someone wants to run his project in product X it is common to use concepts of product X to configure his project. What if I use tomcat base image but don't want to expose port 8080 because it is not secured. I just can't stop Che publishing it. And If I could decide which ports should be opened I would
> be able to control what I want to expose to the world.
> Docker compose doesn't have publish all ports in config. User of compose HAS to set which ports should be published. So I wouldn't say that assumption is wrong - it just differ from your POV.

yes, I see that default is : let me access to all ports exposed by my docker image
Also this issue is not linked to this PR. Because for now, all exposed ports are available through a dynamic port while with the PR, it is available through a single port. (looks like a side issue, not related to this PR)

> It supposed to NOT use labels, but proxy API instead. I suggested this option as usage of interceptor even when strategy is not used looks a bit odd to me.

The interceptor when strategy is not used is a no-op interceptor. You may also eventually remove the plugin module in a custom assembly and then interceptor is gone.
Again, this is not a problem of the interceptor but this is due to Che/Guice integration limitation where we can't disable server-side plug-ins at runtime, all the server modules have to be set when we define the assembly at compile time. While, user/developer might expect that if a server side plug-in was not enabled at runtime, we should not load code at all from this plug-in.

The code that you suggest is part of the Traefik plug-in so I would suppose the implementation choice is depending on the implementation of this plug-in. Traefik has a direct connection to Docker so I see no-value for che to add some extra layers to do third party communication.
This third-party layer will only be setup for codenvy as we have zookeeper container.
Also note that with current implementation when a container is removed, the routes are removed as well.
Infrastructure provisioner has no "removal state" as well this is why it can't be used as well.

you're making some assumptions on this PR as if there were not pro/cons work done on my side previously before providing this PR.
For me, current solution is the best non-intrusive and elegant solution for Che in order to support Traefik.
You remove the plug-in, it removes the support. And best of all : nothing to change in the core or SPI of Che.

skabashnyuk:
@benoitf is this PR are feature complete? Are you going to add something? Is there missing feature? What is your merging plan?

benoitf:
@skabashnyuk  I think this one will never be merged in that state as it enable by default Single port while default option should be kept.
For the merging plan, I'm creating separate small PRs in replacement to this one.

Plan is the following :
once https://github.com/eclipse/che/pull/4928 is merged (need your approval as well)
I will submit PR to introduce a traefik plug-in based on 1.3.0-RC1 (https://github.com/containous/traefik/releases/tag/v1.3.0-rc1 release few days ago so I don't need anymore a custom docker image made by myself)
Then another PR will add a toggle to enable this plug-in in che
and another or a couple of PR for enabling a different plug-in (based this time on Zookeeper/Traefik and not only Docker/Traefik) for codenvy.

skabashnyuk:
Ok. I'm proposing to start merging things like #4928 and keep current configuration as default and " single port" feature as optional

benoitf:
yes that's the plan.

benoitf:
This PR will be closed.
Work has ben done on 
https://github.com/eclipse/che/pull/4928
and newer https://github.com/eclipse/che/pull/5115

This is as discussed, an optional feature and the interceptor is not bound if single port property is missing/not enabled.

PR #4444, opened 2017-03-16
Set Github token through workspace master Rest API. Added a force act…

sunix:
Set Github token through workspace master Rest API. Added a force activation property variable to register Github Oauth provider even without client id/secret

Signed-off-by: Sun Seng David Tan <sutan@redhat.com>

### What does this PR do?
- add a new REST operation to set a Oauth token to an existing provider
- add a property `che.oauth.github.forceactivation` to force registration of Github Oauth provider, even without client id/secret (actually set `NULL` string for these value)

https://issues.jboss.org/browse/CHE-151

#### Changelog

Set Github token through workspace master Rest API. Added a force activation property variable to register Github Oauth provider even without client id/secret

#### Release Notes

Set Github token through workspace master Rest API. Added a force activation property variable to register Github Oauth provider even without client id/secret

### DIFF
--- a/assembly/assembly-wsagent-war/src/main/webapp/WEB-INF/classes/codenvy/che-machine-configuration.properties
+++ b/assembly/assembly-wsagent-war/src/main/webapp/WEB-INF/classes/codenvy/che-machine-configuration.properties
@@ -50,8 +50,6 @@ oauth.github.authuri= https://github.com/login/oauth/authorize
 oauth.github.tokenuri= https://github.com/login/oauth/access_token
 #redirected uris
 oauth.github.redirecturis= http://localhost:${SERVER_PORT}/che/api/oauth/callback
-# register github even without client id and secret
-oauth.github.forceactivation=false
 
 git.server.uri.prefix=git
 
--- a/assembly/assembly-wsagent-war/src/main/webapp/WEB-INF/classes/codenvy/che-machine-configuration.properties
+++ b/assembly/assembly-wsagent-war/src/main/webapp/WEB-INF/classes/codenvy/che-machine-configuration.properties
@@ -50,6 +50,8 @@ oauth.github.authuri= https://github.com/login/oauth/authorize
 oauth.github.tokenuri= https://github.com/login/oauth/access_token
 #redirected uris
 oauth.github.redirecturis= http://localhost:${SERVER_PORT}/che/api/oauth/callback
+# register github even without client id and secret
+oauth.github.forceactivation=false
 
 git.server.uri.prefix=git
 
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
@@ -142,7 +142,8 @@ che.oauth.github.clientsecret=NULL
 che.oauth.github.authuri= https://github.com/login/oauth/authorize
 che.oauth.github.tokenuri= https://github.com/login/oauth/access_token
 che.oauth.github.redirecturis= http://localhost:${SERVER_PORT}/wsmaster/api/oauth/callback
-
+# register github even without client id and secret
+che.oauth.github.forceactivation=false
 
 ### DOCKER PARAMETERS
 # Docker is the default machine implementation within Che. Workspaces are powered by machines
--- a/plugins/plugin-github/che-plugin-github-oauth2/src/main/java/org/eclipse/che/security/oauth/GitHubOAuthAuthenticator.java
+++ b/plugins/plugin-github/che-plugin-github-oauth2/src/main/java/org/eclipse/che/security/oauth/GitHubOAuthAuthenticator.java
@@ -39,7 +39,7 @@ public class GitHubOAuthAuthenticator extends OAuthAuthenticator {
                                     @Nullable @Named("che.oauth.github.redirecturis") String[] redirectUris, //
                                     @Nullable @Named("che.oauth.github.authuri") String authUri, //
                                     @Nullable @Named("che.oauth.github.tokenuri") String tokenUri, //
-                                    @Nullable @Named("che.oauth.github.forceactivation") Boolean forceActivation) throws IOException {
+                                    @Named("che.oauth.github.forceactivation") boolean forceActivation) throws IOException {
         if (!isNullOrEmpty(clientId) //
             && !isNullOrEmpty(clientSecret) //
             && !isNullOrEmpty(authUri) //
@@ -50,9 +50,12 @@ public class GitHubOAuthAuthenticator extends OAuthAuthenticator {
             return;
         }
 
-        if (forceActivation != null && forceActivation) {
+        if (forceActivation && //
+            !isNullOrEmpty(authUri) && //
+            !isNullOrEmpty(tokenUri) && //
+            redirectUris != null && redirectUris.length != 0) {
+
             configure("NULL", "NULL", redirectUris, authUri, tokenUri, new MemoryDataStoreFactory());
-            return;
         }
     }
 
--- a/plugins/plugin-github/che-plugin-github-oauth2/src/main/java/org/eclipse/che/security/oauth/GitHubOAuthAuthenticator.java
+++ b/plugins/plugin-github/che-plugin-github-oauth2/src/main/java/org/eclipse/che/security/oauth/GitHubOAuthAuthenticator.java
@@ -30,23 +30,29 @@ import java.net.HttpURLConnection;
 import java.net.URL;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
-
 /** OAuth authentication for github account. */
 @Singleton
 public class GitHubOAuthAuthenticator extends OAuthAuthenticator {
     @Inject
-    public GitHubOAuthAuthenticator(@Nullable @Named("che.oauth.github.clientid") String clientId,
-                                    @Nullable @Named("che.oauth.github.clientsecret") String clientSecret,
-                                    @Nullable @Named("che.oauth.github.redirecturis") String[] redirectUris,
-                                    @Nullable @Named("che.oauth.github.authuri") String authUri,
-                                    @Nullable @Named("che.oauth.github.tokenuri") String tokenUri) throws IOException {
-        if (!isNullOrEmpty(clientId)
-            && !isNullOrEmpty(clientSecret)
-            && !isNullOrEmpty(authUri)
-            && !isNullOrEmpty(tokenUri)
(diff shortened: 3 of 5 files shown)


skabashnyuk (changes requested):
Conceptually -1 
Explained opinion here https://github.com/eclipse/che/pull/4438

assembly/assembly-wsagent-war/src/main/webapp/WEB-INF/classes/codenvy/che-machine-configuration.properties
```diff
@@ -50,6 +50,8 @@ oauth.github.authuri= https://github.com/login/oauth/authorize
 oauth.github.tokenuri= https://github.com/login/oauth/access_token
 #redirected uris
 oauth.github.redirecturis= http://localhost:${SERVER_PORT}/che/api/oauth/callback
+# register github even without client id and secret
+oauth.github.forceactivation=false
```
garagatyi: I don't see the place in code that uses this property. Can you point to it?

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -142,7 +142,8 @@ che.oauth.github.clientsecret=NULL
 che.oauth.github.authuri= https://github.com/login/oauth/authorize
 che.oauth.github.tokenuri= https://github.com/login/oauth/access_token
 che.oauth.github.redirecturis= http://localhost:${SERVER_PORT}/wsmaster/api/oauth/callback
-
+# register github even without client id and secret
+che.oauth.github.forceactivation=false
```
garagatyi, line 146: Can you elaborate on what will happen if token is not injected but this property is set to true and client secret/id are unset. 
What will happen on clicking OAuth button in IDE?
What will happen on an operation with github?

plugins/plugin-github/che-plugin-github-oauth2/src/main/java/org/eclipse/che/security/oauth/GitHubOAuthAuthenticator.java
```diff
@@ -30,23 +30,29 @@
...
-    public GitHubOAuthAuthenticator(@Nullable @Named("che.oauth.github.clientid") String clientId,
-                                    @Nullable @Named("che.oauth.github.clientsecret") String clientSecret,
-                                    @Nullable @Named("che.oauth.github.redirecturis") String[] redirectUris,
-                                    @Nullable @Named("che.oauth.github.authuri") String authUri,
-                                    @Nullable @Named("che.oauth.github.tokenuri") String tokenUri) throws IOException {
-        if (!isNullOrEmpty(clientId)
-            && !isNullOrEmpty(clientSecret)
-            && !isNullOrEmpty(authUri)
-            && !isNullOrEmpty(tokenUri)
+    public GitHubOAuthAuthenticator(@Nullable @Named("che.oauth.github.clientid") String clientId, //
```
garagatyi, line 37: It is uncommon for existing codebase to use trailing slashes, can you elaborate on why they are needed?

plugins/plugin-github/che-plugin-github-oauth2/src/main/java/org/eclipse/che/security/oauth/GitHubOAuthAuthenticator.java
```diff
@@ -30,23 +30,29 @@
...
+            && !isNullOrEmpty(clientSecret) //
+            && !isNullOrEmpty(authUri) //
+            && !isNullOrEmpty(tokenUri) //
             && redirectUris != null && redirectUris.length != 0) {
 
             configure(clientId, clientSecret, redirectUris, authUri, tokenUri, new MemoryDataStoreFactory());
+            return;
+        }
+
+        if (forceActivation != null && forceActivation) {
```
garagatyi: If you make forceActivation primitive then nullness check can be avoided.

plugins/plugin-github/che-plugin-github-oauth2/src/main/java/org/eclipse/che/security/oauth/GitHubOAuthAuthenticator.java
```diff
@@ -30,23 +30,29 @@
...
+            configure("NULL", "NULL", redirectUris, authUri, tokenUri, new MemoryDataStoreFactory());
+            return;
```
garagatyi: Return is not needed here

plugins/plugin-github/che-plugin-github-oauth2/src/main/java/org/eclipse/che/security/oauth/GitHubOAuthAuthenticator.java
```diff
@@ -30,23 +30,29 @@
...
+            && !isNullOrEmpty(authUri) //
+            && !isNullOrEmpty(tokenUri) //
             && redirectUris != null && redirectUris.length != 0) {
 
             configure(clientId, clientSecret, redirectUris, authUri, tokenUri, new MemoryDataStoreFactory());
+            return;
+        }
+
+        if (forceActivation != null && forceActivation) {
+            configure("NULL", "NULL", redirectUris, authUri, tokenUri, new MemoryDataStoreFactory());
```
garagatyi: If client id/secret are not configured how other params (redirectUris, etc) can be used? Should they be unset too?

wsmaster/che-core-api-auth/src/main/java/org/eclipse/che/security/oauth/OAuthAuthenticationService.java
```diff
@@ -203,4 +205,27 @@ protected OAuthAuthenticator getAuthenticator(String oauthProviderName) throws B
         }
         return oauth;
     }
+
+    @POST
+    @Path("token")
+    @Consumes(MediaType.APPLICATION_JSON)
+    public void setToken(@Required @QueryParam("oauth_provider") String oauthProvider, //
```
garagatyi, line 212: Looks like one of  map's entry can growth infinitely because of this API, and it is not restricted so any user of the product can cause OOM

JamesDrummond:
@skabashnyuk Can you explain a little more why you are not ok with this PR conceptually?

wsmaster/che-core-api-auth/src/main/java/org/eclipse/che/security/oauth/OAuthAuthenticationService.java
sunix: It means that `EnvironmentContext.getCurrent().getSubject().getUserId();` can return any userId ? or be dynamically changed ? are you sure of that ?
If you are sure of that, so getToken shouldn't be available because anyone could still others' token ...

plugins/plugin-github/che-plugin-github-oauth2/src/main/java/org/eclipse/che/security/oauth/GitHubOAuthAuthenticator.java
sunix: It's for readability. But I can remove it ...

plugins/plugin-github/che-plugin-github-oauth2/src/main/java/org/eclipse/che/security/oauth/GitHubOAuthAuthenticator.java
sunix: Can it be a default null value (so it's false by default) ?

plugins/plugin-github/che-plugin-github-oauth2/src/main/java/org/eclipse/che/security/oauth/GitHubOAuthAuthenticator.java
sunix: I use the default formatter in Che. So these were to keep the current formatting

skabashnyuk:
@sunix are you 100% sure that there is no way(even theoretical) to hack the  system like Che or Codenvy with this method? How confident you are?
```
@POST @Path("token"
```
@l0rd @benoitf wdyt? Is this only my paranoia :)?

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
sunix: These video shows an attempt to perform a git operation or github operation (would be similar to clicking in OAuth button):
1. attempt with default che.oauth.github.forceactivation false
2. attempt with che.oauth.github.forceactivation true without setToken
3. attempt with che.oauth.github.forceactivation true and setting the token

- Github operation (upload ssh key - old demo, but will show what happen) https://youtu.be/y9cux67RQHY It shows a 404 in a pop in one case + notification error of Oauth failure
- Git operation (push) https://youtu.be/5D6XZlDyPKE . It shows a notification error

sunix:
@skabashnyuk I think that User's data and access to other system is more important than anything.
But if I were a hacker, I would try to get the token and have access to the users Github ( or whatever) and steal informations or change rights if I can. Setting the token ... ? I really don't see how this can be used.

In my opinion, setting the token and use another system to do the oauth stuff is actually more secured than letting Codenvy/Che doing the whole thing: the system responsible for oauth would focus on security ... where as we (Che) want to focus our effort on providing the best development tool.

skabashnyuk:
Can you comment
https://oauth.net/articles/authentication/ 
Injection of access tokens
An additional (and very dangerous) threat occurs when clients accept access tokens from sources other than the return call from the token endpoint. This can occur for a client that uses the implicit flow (where the token is passed directly as a parameter in the URL hash) and don't properly use the OAuth state parameter. This issue can also occur if different parts of an application pass the access token between components in order to "share" access among them. This is problematic because it opens up a place for access tokens to potentially be injected into an application by an outside party (and potentially leak outside of the application). If the client application does not validate the access token through some mechanism, it has no way of differentiating between a valid token and an attack token.

This can be mitigated by using the authorization code flow and only accepting tokens directly from the authorization server's token enpdoint, and by using a state value that is unguessable by an attacker.


Are you sure it's not related to what you are proposing?

sunix:
Yes OK. Thanks for this. The plan is: we will add validation of token.
@skabashnyuk how does it sounds ? or you still think we shouldn't add a setToken at all ?

TylerJewell:
@sunix - I think your plan sounds appropriate.  What is the plan to revise and then get this merged?

vparfonov:
@sunix @skabashnyuk any update here, look like not actual for now?

sunix:
not needed anymore thanks

PR #4466, opened 2017-03-17
Implement top method in OpenShiftConnector

snjeza:
What does this PR do?

Implements the top method in OpenShiftConnector.

This method isn't called anywhere in che. In order to test it, you need to debug che and call the method manually.


TylerJewell:

benoitf:
work is done as part of https://github.com/eclipse/che/pull/4466

PR #4472, opened 2017-03-17
[OpenShift] Modifications to the shell scripts that bootstrap the servers

l0rd:
### What does this PR do?
- Remove `RunAs` instruction when starting a workspace. Container will run as an arbitrary user (usually with a UID that doesn't match any host UID) and with security constraint `restricted` (that's the most secure scc in OpenShift).
- Modify the scripts that bootstrap the servers (wsagent, terminal etc...) to avoid usage of `sudo` when the container is run as an arbitrary user. That works only if some modified [Docker images are used](https://github.com/redhat-developer/che-dockerfiles).
- Change `stacks.json` to point to these modified Docker images
- Disable SSH 

#### Changelog
Modifications to the shell scripts that bootstrap the servers

### DIFF
--- a/agents/go-agents/src/main/resources/org.eclipse.che.terminal.script.sh
+++ b/agents/go-agents/src/main/resources/org.eclipse.che.terminal.script.sh
@@ -13,7 +13,7 @@ unset PACKAGES
 unset SUDO
 command -v tar >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" tar"; }
 command -v curl >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" curl"; }
-test "$(id -u)" = 0 || SUDO="sudo -E"
+test "$(id -u)" = 0 || test -f ${HOME}/is_arbitrary_user || SUDO="sudo -E"
 
 CHE_DIR=$HOME/che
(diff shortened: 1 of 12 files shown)


ibuziuk:
@l0rd for now this change is supposed to work only with **vertx** & **ubuntu_jdk8** patched versions of stacks, right ?

ibuziuk:
@l0rd For me PR worked just fine against rhche/ubuntu_jdk8 stack but for rhche/vertx I got permission denied: 

```
[dev-machine] [STDOUT] Terminal Agent will be downloaded from Workspace Master
[dev-machine] [STDERR] mkdir: cannot create directory '//che': Permission denied
[dev-machine] [STDERR] 
[dev-machine] [STDERR] gzip: stdin: unexpected end of file
[dev-machine] [STDERR] tar: Child returned status 1
[dev-machine] [STDERR] tar: Error is not recoverable: exiting now
[dev-machine] [STDERR] /bin/sh: line 177: //che/terminal/che-websocket-terminal: No such file or directory

```
I guess, this is weird because ssh agent has been removed from vertx stack

l0rd:
I haven't patched vert.x stack yet

Il 20 mar 2017 12:23, "Ilya Buziuk" <notifications@github.com> ha scritto:

ibuziuk:
@l0rd than everything is expected I suppose

ibuziuk (approved):

PR #4477, opened 2017-03-20
use docker-make for images building and pushing

jizhilong:
Hi guys,
 I'm a newbie of eclipse/che and the author of [docker-make](https://github.com/CtripCloud/docker-make)-a cli tool for building,tagging,and pushing a bunch of related docker images.

I noticed that eclipse/che builds several Docker images for different parts of the whole system, and these images are managed by some bash scripts since docker has no official `docker-compose` like tool for image building.

We also encountered such scenarios in some of our private projects- building several related images from one code base.Though bash scripts and makefiles can help to automate images management, it also has issues for unclear configurations, bad readability, and unnecessary  copy of scripts for new projects.

So we developed `docker-make` to make such works yaml-configurable, just as we do on containers with the power of `docker-compose`, and I'm very glad to introduce this tool to you.

### What does this PR do?
Introduce in  [docker-make](https://github.com/CtripCloud/docker-make) for building and pushing che's various docker images.

No related issues.

### DIFF
--- a/.docker-make.yml
+++ b/.docker-make.yml
@@ -1,0 +1,105 @@
+builds:
+  base:
+    context: /dockerfiles/base
+    dockerfile: Dockerfile
+    pushes:
+      - "on_branch:master=eclipse/che-base:nightly"
+
+  lib-dto:
+    context: /dockerfiles/lib
+    dockerfile: Dockerfile.dto
+    extract:
+      - /tmp/dto/:./dto.tar
+
+  lib:
+    context: /dockerfiles/lib
+    dockerfile: Dockerfile
+    depends_on:
+      - lib-dto
+    pushes:
+      - "on_branch:master=eclipse/che-lib:nightly"
+
+  action:
+    context: /dockerfiles/action
+    dockerfile: Dockerfile
+    depends_on:
+      - lib
+    rewrite_from: lib
+    pushes:
+      - "on_branch:master=eclipse/che-action:nightly"
+
+  bats:
+    context: /dockerfiles/bats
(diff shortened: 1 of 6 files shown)


TylerJewell:
@jizhilong - we certainly want to applaud your new project and making attempts to bring some sanity to the process relating to building Docker images that have complex sets of dependencies, and offering up a pull request on how this might work within Eclipse Che.

However, it may be a bit premature for us to think about merging it into our systems as we have a number of requirements that have to be addressed:
1. We have complex logic related to naming and tagging the organization, repository, prefix for images.
2. Our CI systems cannot easily install the utility natively, so they would all have to be rewritten for docker
3. We have a dozen or so custom assemblies that inherit from che's base, and they have similar build / CI approaches, so any upstream fundamental change has downstream ripples that have to be tested
4. We have optional built in integrated unit testing for our images as well that need to be reflected
5. We have cross-GitHub repository dependencies such as codenvy/cli depending upon che's base image.
6. We have some complex versioning logic that then implies certain tagging.
7. It's unclear why, but you were modifying some core files in our maven assembly, and this is something that you should try to avoid.

I think we'd like to see this utility get broader usage and for our engineers time to experiment more.  We all agree that having a nice YML format for describing image build dependencies is definitely an improvement for the industry.

jizhilong:
@TylerJewell  thanks for such a detailed explanation, I'll close this PR for now.

TylerJewell:
I think there is a ton of potential with what you are doing.  So it just has to be inroduced into projects carefully as what you provide would be fundamental to how different projects set up their CI systems, for example.

PR #4478, opened 2017-03-20
Rework firing events

tolusha:
### What does this PR do?
Move events firing from UserDao to UserManager

https://github.com/codenvy/codenvy/issues/1923

#### Changelog
not required

#### Release Notes
not required

#### Docs PR
not required


wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserManager.java
```diff
@@ -252,6 +253,34 @@ public long getTotalCount() throws ServerException {
...
+        }
+
+        eventService.publish(new BeforeUserRemovedEvent(user)).propagateException();
         userDao.remove(id);
     }
+
+    private class RemovePreferencesBeforeUserRemovedEventSubscriber extends CascadeEventSubscriber<BeforeUserRemovedEvent> {
+        @Override
+        public void onCascadeEvent(BeforeUserRemovedEvent event) throws Exception {
+            preferencesDao.remove(event.getUser().getId());
```
sleshchenko: Why you can't do it in `UserManager#doRemove` method?

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/jpa/JpaUserDao.java
```diff
@@ -237,16 +228,13 @@ protected void doUpdate(UserImpl update) throws NotFoundException {
     }
 
     @Transactional(rollbackOn = {RuntimeException.class, ServerException.class})
-    protected Optional<UserImpl> doRemove(String id) throws ServerException {
+    protected void doRemove(String id) throws ServerException {
```
sleshchenko: When can this method throw ServerException?

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserManager.java
voievodin: nice catch, it makes sense to move this code from listeners to manager `doRemove`

voievodin (approved):

sleshchenko (approved):

wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserManager.java
```diff
@@ -91,32 +96,20 @@ public User create(User newUser, boolean isTemporary) throws ConflictException,
...
-        }
+        doCreate(user, isTemporary);
+        eventService.publish(new UserCreatedEvent(user));
         return user;
     }
 
+    @Transactional(rollbackOn = {RuntimeException.class, ApiException.class})
+    protected void doCreate(UserImpl user, boolean isTemporary) throws ConflictException, ServerException {
+        userDao.create(user);
+        eventService.publish(new PostUserPersistedEvent(new UserImpl(user))).propagateException();
```
sleshchenko, line 107: Mb It'd better to publish event after profile creation and preferences setting. It'd allow listener to put something into preferences or profile attributes, otherwise we'll need different events for profile and preferences when there will be such listener.
For me it's OK to listen user persisted event and put something into profile because UserManager#create -> create profile automatically and UserManager publish this event. So it's the same as expect that profile will exist after user creation by UserManager#create
WDYT?

tolusha: `PostUserPersistedEvent` should be thrown before profile creation because `user` and `profile` are different entities. 
If we need to put something into the profile we have to listen to `UserCreatedEvent`. It is designed for that exact purpose.
@evoevodin wdyt ?

voievodin: The current approach is the same to the previous one, the only difference is that real transaction is initiated instead of the fake one, which probably means that use-cases we have are handled correctly. I would stay for current implementation as long as we don't know for sure what the other use-cases are.

tolusha:
[Selenium test report](https://github.com/eclipse/che/files/858047/report.txt)

skabashnyuk (approved):

Issue #4372, opened 2017-03-09
Make factory name mandatory


voievodin:
The problem
---
For each factory e.g.
```json
{
  "v": "4.0",
  "name": "my-factory",
  "id": "i0rx6mlihjaayugt",
  "creator": {
    "name": "username",
    "created": 1465332525959,
    "userId": "user1234567"
  },
  "workspace": { }
}
```
the combination of `factory.name` + `factory.creator.userId` must be *unique*, so the factory
may be accepted as `/<namespace>/<factory-name>` where the namespace is the name of the user who is the creator of the factory. 
Those factories which don't have name, cannot be accepted in the format described above.

Right now multiple factories with the same name may exist, as the unique constraint is missing.
In this case the first factory returned from factory service will be used.

The actual problem here is that the unique constraint like this requires partial unique index, which is not a common feature of database systems. It could look like this:

```sql
CREATE UNIQUE INDEX idx ON factory (user_id, name) WHERE (user_id IS NOT NULL)
```
While it works for PostgreSQL it doesn't for H2.

Proposed solution
---
Make factory name mandatory on database level + add a unique constraint on _(user_id, name)_.

Each time factory service receives a factory without name it must set its name to some value e.g. _factory id_ or generate a new factory name, so in this way clients are still not required to send factory name so everything works like before

For those factories in database that don't have name create a migration script that sets names equal to _factory ids_(or the values generated by the same principle as the values set by service) e.g:

| factory_id        | factory_name           | user_id  |
| ------------- |:-------------:| -----:|
| factory123      | null | user123 |
| factory234      | null | user123 |

:arrow_down: becomes :arrow_down: 

| factory_id        | factory_name           | user_id  |
| ------------- |:-------------:| -----:|
| factory123      | factory123 | user123 |
| factory234      | factory234 | user123 |

For those factories in database that have duplicated _name + user_id_ create a migration script that 
leaves the very first factory with such a name as it is and modifies all the other factories in a way defined by the script from the previous statement. So the factories with such a name are accepted as they were(the first one) e.g:

| factory_id        | factory_name           | user_id  |
| ------------- |:-------------:| -----:|
| factory123      | example | user123 |
| factory234      | example | user123 |

:arrow_down: becomes :arrow_down: 

| factory_id        | factory_name           | user_id  |
| ------------- |:-------------:| -----:|
| factory123      | example | user123 |
| factory234      | factory234 | user123 |

voievodin:
[...]

PR #4521, opened 2017-03-22
CHE-4372; add unique constraing for factory name + user

mshaposhnik:
### What does this PR do?
Makes factory name mandatory on database level (auto generated if not set by user) 

#4372 

#### Changelog
Factory name made mandatory on database level

#### Release Notes
n/a

#### Docs PR
N/A


wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryManager.java
```diff
@@ -133,8 +141,12 @@ public Factory updateFactory(Factory update, Set<FactoryImage> images) throws Co
                                                                                   ServerException {
         requireNonNull(update);
         final AuthorImpl creator = factoryDao.getById(update.getId()).getCreator();
+        FactoryImpl updateImpl = new FactoryImpl(update, images);
+        if (isNullOrEmpty(updateImpl.getName())) {
```
sleshchenko: It would be better not to generate new name when update doesn't contain it. Let's set old name when it is null or empty in update.

wsmaster/che-core-sql-schema/src/main/resources/che-schema/5.6.0/3__create_factory_name_plus_userid_index.sql
```diff
@@ -0,0 +1,16 @@
...
+-- if any.  The intellectual and technical concepts contained
+-- herein are proprietary to Codenvy S.A.
+-- and its suppliers and may be covered by U.S. and Foreign Patents,
+-- patents in process, and are protected by trade secret or copyright law.
+-- Dissemination of this information or reproduction of this material
+-- is strictly forbidden unless prior written permission is obtained
+-- from Codenvy S.A..
+--
+
+CREATE UNIQUE INDEX index_name_plus_userid ON che_factory (user_id, name);
```
sleshchenko: `index_name_plus_userid` -> `index_che_factory_name_user_id`

sleshchenko: Looks like next script also does it. Do we need do it twice?

wsmaster/che-core-sql-schema/src/main/resources/che-schema/5.6.0/postgresql/3__create_factory_name_plus_userid_index.sql
```diff
@@ -0,0 +1,42 @@
...
+-- herein are proprietary to Codenvy S.A.
+-- and its suppliers and may be covered by U.S. and Foreign Patents,
+-- patents in process, and are protected by trade secret or copyright law.
+-- Dissemination of this information or reproduction of this material
+-- is strictly forbidden unless prior written permission is obtained
+-- from Codenvy S.A..
+--
+
+UPDATE che_factory
+SET name = concat('f', right(id, 9)) WHERE name IS NULL OR name = '';
```
sleshchenko: Script name does not match the content

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryManager.java
mshaposhnik: In this case you cant re-assign the name to another factory. I don't know if it's good.

sleshchenko: Why you can't re-assign the name? You can send name in update then name will be updated. Or you can miss it, then old name will be used (but not generated new)

mshaposhnik: For me it looks stupid, if you deleting the factory name on Dashb, and it returns back after update :). But seems others are with you, so ok.

mshaposhnik: So seems it must not be generated on update at all.

skabashnyuk (approved):

wsmaster/che-core-sql-schema/src/main/resources/che-schema/5.6.0/postgresql/3__create_factory_name_plus_userid_index.sql
```diff
@@ -0,0 +1,46 @@
...
+--
+
+
+-- Generate names where they are not set.
+UPDATE che_factory
+SET name = concat('f', right(id, 9)) WHERE name IS NULL OR name = '';
+
+-- Make names unique for the same user, e.g if there is more than one factory with same name and user,
+-- leave first one and rename others.
+WITH dupes AS
```
voievodin: Can we do this migration on codenvy tables before we move factory to che?
Because it's weird to see such a postgresql specific migration in che, while che doesn't have factories yet.
Moreover it allows to define unique constraint next to _che_factory_ declaration.

mshaposhnik: Yes I can move naming uniques into 1.1 in codenvy, then 1.2 will do the migration, and costraint will be just in `che_factory` table definition.

akorneta (approved):

voievodin (approved):

PR #4525, opened 2017-03-22
Update Docker Compose tests to fix test failure

amisevsk:
### What does this PR do?
Updating to Jackson 2.7.7 causes tests in the docker compose plugin to fail. This is due to the fact that the tests expect empty values in dictionaries to be parsed as the empty string, whereas jackson 2.7.7 parses them as null (as specified by the yaml spec).

Modifies the affected tests to explicitly use an empty string (i.e. "") instead of an empty value.

Fixes failed tests in `CommandDeserializerTest.java` and `EnvironmentDeserializerTests.java`

#### Changelog
Update Docker Compose tests to fix test failure


amisevsk:
@l0rd WDYT?

l0rd (approved):

l0rd:
Thanks @amisevsk you can merge that.

Issue #339, opened 2016-02-15
Custom Java formatting rules


sunix:
User should be able to choose which formatting rules to use
- at least basic rules, like use tab or 2 spaces for indent
- the best: import Eclipse formatting rule

Also need to make possible set formatting per project 
- [x] Add ability set formatting per project and for workspace #5724
- [x] Develop UI panel for changing Java code formatter preferences #5720
- [x] Develop service for importing Java formatter preferences form the file #5721

sunix:
Sorry @TylerJewell , this is still a major issue, I am reopening it.

sunix:
To me this is fixed with https://github.com/eclipse/che/pull/6077
I managed to import the google java formatter for eclipse from https://github.com/google/styleguide
and it seems to work well

PR #4556, opened 2017-03-24
Various fixes to the PR panel

sunix:
1. The user already using the oauth Github token to create the fork, upload the ssh key and create the PR so it makes sense to git push with GH token. Changing AddSshForkRemoteStep with AddHttpForkRemoteStep in GitHubContributionWorkflow.java
2. It doesn't work if we change the local branch to use. Base upstream branch to merge into will change. This is because the panel is supposed to work from a Factory where attribute "base upstream branch to merge to" is set. We should fix it somehow: user should be able to choose the upstream branch to merge into. Also, We shouldn't update this branch when selecting another local branch. https://www.youtube.com/watch?v=H4_cJJWQ5DY

https://issues.jboss.org/browse/CHE-106

#### Changelog
Various fixes to the PR panel

### DIFF
--- a/plugins/plugin-pullrequest-parent/che-plugin-pullrequest-ide/src/main/java/org/eclipse/che/plugin/pullrequest/client/steps/InitializeWorkflowContextStep.java
+++ b/plugins/plugin-pullrequest-parent/che-plugin-pullrequest-ide/src/main/java/org/eclipse/che/plugin/pullrequest/client/steps/InitializeWorkflowContextStep.java
@@ -89,7 +89,7 @@ public class InitializeWorkflowContextStep implements Step {
                     context.setOriginRepositoryOwner(vcsHostingService.getRepositoryOwnerFromUrl(originUrl));
                     context.setOriginRepositoryName(vcsHostingService.getRepositoryNameFromUrl(originUrl));
 
-                    context.setContributeToBranchName(getContributeToBranchName(context.getProject()));
+                    setContributeToBranchName(context);
 
                     executor.done(InitializeWorkflowContextStep.this, context);
                 } else {
@@ -100,7 +100,24 @@ public class InitializeWorkflowContextStep implements Step {
         };
     }
 
-    private String getContributeToBranchName(final ProjectConfig project) {
+    protected void setContributeToBranchName(Context context) {
+        String contributeToBranchName = getBranchFromProjectMetadata(context.getProject());
+
+        if (contributeToBranchName != null) {
+            context.setContributeToBranchName(contributeToBranchName);
+            return;
+        }
+
+        vcsServiceProvider.getVcsService(context.getProject()) //
+                          .getBranchName(context.getProject()) //
+                          .then( //
+                          (String branchName) -> {
+                              context.setContributeToBranchName(branchName);
+                              context.getProject().getSource().getParameters().put("branch", branchName);
+                          });
+    }
+
+    private String getBranchFromProjectMetadata(final ProjectConfig project) {
         final Map<String, List<String>> attrs = project.getAttributes();
         if (attrs.containsKey(CONTRIBUTE_TO_BRANCH_VARIABLE_NAME) && !attrs.get(CONTRIBUTE_TO_BRANCH_VARIABLE_NAME).isEmpty()) {
             return attrs.get(CONTRIBUTE_TO_BRANCH_VARIABLE_NAME).get(0);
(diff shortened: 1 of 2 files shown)


plugins/plugin-pullrequest-parent/che-plugin-pullrequest-ide/src/main/java/org/eclipse/che/plugin/pullrequest/client/steps/InitializeWorkflowContextStep.java
```diff
@@ -100,7 +100,24 @@ public void apply(final List<Remote> remotes) throws OperationException {
...
-    private String getContributeToBranchName(final ProjectConfig project) {
+    protected void setContributeToBranchName(Context context) {
+        String contributeToBranchName = getBranchFromProjectMetadata(context.getProject());
+
+        if (contributeToBranchName != null) {
+            context.setContributeToBranchName(contributeToBranchName);
+            return;
+        }
+
+        vcsServiceProvider.getVcsService(context.getProject()) //
```
skabashnyuk: could you please remove ```//``` at the end of lines. It's unusual for che such kind of formatting technique.

voievodin:
Could you please describe changes a bit.
Does it mean that you will always use https remote even if you have ssh key?

plugins/plugin-pullrequest-parent/che-plugin-pullrequest-ide/src/main/java/org/eclipse/che/plugin/pullrequest/client/steps/InitializeWorkflowContextStep.java
sunix: Can you elaborate what's wrong with that ?
If the default formatting rules in Che doesn't let the Che users format their code the way they want then we are going nowhere: https://github.com/eclipse/che/issues/339

sunix:
@evoevodin I've updated the description
It means that in all the cases we already have the GH token that we already use for create the fork, create the PR (uploading the ssh key) etc ... so it makes sense for the Github workflow to reuse it to Git push.
It means that we don't always upload the ssh key even if we already can push with https and the GH OAuth Token

voievodin (approved):

mshaposhnik (approved):

plugins/plugin-pullrequest-parent/che-plugin-pullrequest-ide/src/main/java/org/eclipse/che/plugin/pullrequest/client/steps/InitializeWorkflowContextStep.java
skabashnyuk: >Can you elaborate what's wrong with that ?

Nothing. It's just looks weird to add your personal preference formatting and to be
surprised that other committees may not understand it.

About formatting.
This is a formatting style for intellij. Developed by @florent and @evoevodin
https://github.com/codenvy-legacy/dev/blob/cded88bbe9b6cf006ac5ba5b99ca9069a87109cc/che-codestyle/src/main/resources/che-codestyle-intellij-new.xml

Docker command to run without Intellij
```
docker run -ti \
            -u root \
            --net="host" \
            --privileged=true \
            -e DISPLAY=${DISPLAY} \
            -v /tmp/.X11-unix:/tmp/.X11-unix \
            -v ${HOME}/.Xauthority:/home/developer/.Xauthority \
            -v  {path to sources}:/cheformat/sources ksmster/che-format
```
If you don't like this format we can discuss it in che-dev mailist.

skabashnyuk (approved):

plugins/plugin-pullrequest-parent/che-plugin-pullrequest-ide/src/main/java/org/eclipse/che/plugin/pullrequest/client/steps/InitializeWorkflowContextStep.java
sunix: OK I will remove the ```//``` and open a discussion in che-dev.
About the docker command, it's not that I don't trust you but ... I won't run that with root :)

skabashnyuk: heh

PR #4562, opened 2017-03-25
Remove dashabord from che-parent submodules to avoid building it

l0rd:
This PR is to avoid compilation of the dashboard component (and to stay with version 5.0.0)

#### Changelog
Remove dashabord from che-parent submodules to avoid building it

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -31,7 +31,7 @@
         <module>agents</module>
         <module>plugins</module>
         <module>samples</module>
-        <module>dashboard</module>
+        <!--<module>dashboard</module>-->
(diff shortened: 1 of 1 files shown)


ibuziuk:
Dashboard still is not loading for me - https://snag.gy/9eQEAU.jpg

`Failed to load resource: the server responded with a status of 405 (Method Not Allowed)` might not be coupled with dashboardactually...

ibuziuk:
@l0rd oh, false alarm - this problem might be coupled with my fixes - building from this branch works fine.
+1 to apply

PR #4563, opened 2017-03-25
Setting rwx permissions for all on /data/ in case it's not mounted

l0rd:
If /data/ is not bind mounted from the host and if the container is started as an arbitrary user, a permission error will occur.

Creating the folder and chmod it solves the problem.

#### Changelog
Setting rwx permissions for all on /data/ in case it's not mounted

### DIFF
--- a/dockerfiles/che/Dockerfile.centos
+++ b/dockerfiles/che/Dockerfile.centos
@@ -50,3 +50,4 @@ ENTRYPOINT ["/entrypoint.sh"]
 ADD eclipse-che.tar.gz /home/user/
 RUN mkdir /logs && chmod 0777 /logs
 RUN chmod -R 0777 /home/user/
+RUN mkdir /data && chmod 0777 /data


TylerJewell:
Is there a matching issue for the non Centos dockerfile?
Also - please make sure you get at least one +1 on any changes before
merging, too.


l0rd:
@TylerJewell that has been merged in `openshift-connector` branch. Not `master`.

TylerJewell:
Then perhaps we should not have any PRs for thins going into another
branch.

But the question is still relevant - openshift branch or not - as
everything eventually needs to go into master.


l0rd:
@TylerJewell we are currently 4 engineers working on that branch so doing PRs is a good way to ask for a code review before merging. That is what we have been doing during the last 4 months or so. What we may do is to add a prefix to every PR title so that you can identify it easily (e.g. [OpenShift]).

Even if this is a minor fix that solves a corner case of our OpenShift scenario (`/data/` volume not mounted and che-server run as an arbitrary, not privileged user) I agree with you that we should add that to the Alpine based Dockerfile too.

In general we have been working on improving the support for running Che server and Che workspaces as an unprivileged and arbitrary users (when in the container `sudo` is not allowed and the UID doesn't match any of the host's UID). This commit is one small step of that "epic" and we definitely should port that to master as soon as we are comfortable with the quality of the code with respect to non OpenShift scenarios. That work is decoupled from the OpenShift API related work and therefore should be easily merged.

TylerJewell:
I will give some thought while on holiday about whether we need to have a
policy doc or some light rules around long-lived branches, PRs against
non-master branches, and the so forth.  Watching this branch mature and
develop has revealed elements of policy, best practice or procedure that we
have not considered before, and should be codified into our project
practices and guidelines.


l0rd:
@TylerJewell +1

PR #4564, opened 2017-03-27
Add support for resource requests/limits when running on Openshift

amisevsk:
### What does this PR do?
Add resource requests and limits to workspace Pods when running on OpenShift. Workspace memory limit is obtained from the che REST API, while memory request, cpu limit, and cpu request are determined by new che properties

- "che.openshift.workspace.memory.request"
- "che.openshift.workspace.cpu.limit"
- "che.openshift.workspace.cpu.request"

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
@@ -294,6 +294,9 @@ che.openshift.liveness.probe.delay=300
 che.openshift.liveness.probe.timeout=1
 che.openshift.workspaces.pvc.name=claim-che-workspace
 che.openshift.workspaces.pvc.quantity=10Gi
+che.openshift.workspace.cpu.limit=1
+# Override memory limit used for openshift workspaces. String, e.g. 1300Mi
+che.openshift.workspace.memory.override=NULL
 
 # Which implementation of DockerConnector to use in managing containers. In general,
 # the base implementation of DockerConnector is appropriate, but OpenShiftConnector
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
@@ -170,6 +170,8 @@ public class OpenShiftConnector extends DockerConnector {
     private final String          cheWorkspaceStorage;
     private final String          cheWorkspaceProjectsStorage;
     private final String          cheServerExternalAddress;
+    private final String          cheWorkspaceCpuLimit;
+    private final String          cheWorkspaceMemory;
 
     @Inject
     public OpenShiftConnector(DockerConnectorConfiguration connectorConfiguration,
@@ -183,7 +185,10 @@ public class OpenShiftConnector extends DockerConnector {
                               @Named("che.openshift.workspaces.pvc.name") String workspacesPersistentVolumeClaim,
                               @Named("che.openshift.workspaces.pvc.quantity") String workspacesPvcQuantity,
                               @Named("che.workspace.storage") String cheWorkspaceStorage,
-                              @Named("che.workspace.projects.storage") String cheWorkspaceProjectsStorage) {
+                              @Named("che.workspace.projects.storage") String cheWorkspaceProjectsStorage,
+                              @Named("che.openshift.workspace.cpu.limit") String cheWorkspaceCpuLimit,
+                              @Nullable @Named("che.openshift.workspace.memory.override") String cheWorkspaceMemory) {
+
 
         super(connectorConfiguration, connectionFactory, authResolver, dockerApiVersionPathPrefixProvider);
         this.cheServerExternalAddress = cheServerExternalAddress;
@@ -194,6 +199,8 @@ public class OpenShiftConnector extends DockerConnector {
         this.workspacesPvcQuantity = workspacesPvcQuantity;
         this.cheWorkspaceStorage = cheWorkspaceStorage;
         this.cheWorkspaceProjectsStorage = cheWorkspaceProjectsStorage;
+        this.cheWorkspaceCpuLimit = cheWorkspaceCpuLimit;
+        this.cheWorkspaceMemory = cheWorkspaceMemory;
 
         this.openShiftClient = new DefaultOpenShiftClient();
     }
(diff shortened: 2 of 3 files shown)


amisevsk:
@l0rd I've squashed some new changes in to the commit above and rebased my branch. 

Changes:
- Remove resource requests entirely
- Convert che memory limit to MiB
    - note: che property `che.workspace.default_memory_mb` (and the corresponding env var) seem to be overriden somewhere in Che. To test memory limits you need to use the dashboard currently.
- Remove memory limit from `inspectContainer()` output. The memory limit was used only in `CgroupOOMDetector.java`, which seems to no longer be in use.

l0rd:
Thank you Angel. Since current property to set workspace max memory doesn't
work and we don't want to use the dashboard in our use case, wouldn't it
better if we add a new distinct property to set the workspace pod memory
limit?

Il 28 mar 2017 19:54, "Angel Misevski" <notifications@github.com> ha
scritto:

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -294,6 +294,9 @@ che.openshift.liveness.probe.delay=300
 che.openshift.liveness.probe.timeout=1
 che.openshift.workspaces.pvc.name=claim-che-workspace
 che.openshift.workspaces.pvc.quantity=10Gi
+che.openshift.workspace.cpu.limit=1
+# Override memory limit used for openshift workspaces. String, e.g. 1300Mi
+che.openshift.workspace.memory.limit=
```
skabashnyuk: it would be nice if name of the property contains dementions like ```che.workspace.default_memory_mb```
Ex ```che.openshift.workspace.memory.limit_mb=```
Wdyt?

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -252,6 +256,13 @@ public ContainerCreated createContainer(CreateContainerParams createContainerPar
         String[] volumes = createContainerParams.getContainerConfig().getHostConfig().getBinds();
 
         Map<String, String> additionalLabels = createContainerParams.getContainerConfig().getLabels();
+
+        long memoryLimitBytes = createContainerParams.getContainerConfig().getHostConfig().getMemory();
+        String memoryLimit = Long.toString(memoryLimitBytes / 1048576) + "Mi";
```
skabashnyuk: FYI good tool to parse such values ```org.eclipse.che.commons.lang.Size```

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -1093,6 +1109,9 @@ private Deployment createOpenShiftDeploymentInternal(String workspaceID,
                                     .withLivenessProbe(getLivenessProbeFrom(exposedPorts))
                                     .withCommand(command)
                                     .withVolumeMounts(getVolumeMountsFrom(volumes, workspaceID, withSubpath))
+                                    .withNewResources()
+                                        .withLimits(resourceLimits)
```
skabashnyuk, line 1125: lines looks shifted

plugins/plugin-docker/che-plugin-openshift-client/src/test/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnectorTest.java
```diff
@@ -38,7 +38,8 @@
     private static final String   OPENSHIFT_DEFAULT_WORKSPACE_STORAGE = "/data/workspaces";
     private static final String   OPENSHIFT_DEFAULT_WORKSPACE_PROJECTS_STORAGE = "/projects";
 	private static final String   CHE_DEFAULT_SERVER_EXTERNAL_ADDRESS = "che.openshift.mini";
-	
+	private static final String   CHE_WORKSPACE_CPU_LIMIT = "1";
+
```
skabashnyuk: 40-41-43 linse looks shifted

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
amisevsk: We could do something like this, I didn't do it initially since OpenShift interprets the string provided. So, for example, `1Gi == 1024Mi == 1073741824` are all valid values for memory. CPU is measured as e.g. 1 = 1 CPU, or 500m = 500 millicores. I wasn't sure the best way to include dimensions, since the strings are passed directly to OpenShift, so anything that is valid as a resource limit in OpenShift is valid here.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
amisevsk: Ah thank you, I knew I had seen this class before, but couldn't find it very quickly when I was writing this. I'm not sure the output provided by `Size` will work perfectly, since OpenShift interprets `M` as "mega", and `Mi` as mebi. I'll have to test it more.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
amisevsk: It is shifted on purpose; the dsl for openshift-client has nested elements, so it's more readable to indent those elements to expose the structure. The only methods available between `withNewResources()` and `endResources()` are `withLimits()` and `withRequests()`. A more obvious example of this is a little further down, where the `Deployment` is created.

plugins/plugin-docker/che-plugin-openshift-client/src/test/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnectorTest.java
amisevsk: Ah thanks, that was due to a rogue tab character on the above line :+1:

amisevsk:
@l0rd I looked into adding such a property, but couldn't decide on a good way to do it easily. If we use the property to override what is provided to che-server, then it could get confusing down the road. 

Either way, the limit can be set via the dashboard when creating a workspace, or through the workspace API.

l0rd (approved):

Issue #4452, opened 2017-03-16
Multiple Che Servers Cleanup Wrong Workspaces 


JamesDrummond:
When running two Che servers on same docker daemon, a che server will destroy/remove workspace containers belonging to the another che server. There is currently now way based on the normalized container name to determine a workspace containers associated che server parent.

Currently working on fix in [fix-cleanup-2-che-servers-560 branch](https://github.com/eclipse/che/compare/fix-cleanup-2-che-servers-560)

ghost:
Maybe you just lack RAM?

JamesDrummond:
I don't think it's the problem with ram. I think it is has something to do with running codenvy and che on the same docker daemon. When only che or codenvy at any given time I no longer have this issue. Can someone use both che and codenvy on same docker daemon with ws running in each to see if they can reproduce. Is there a clean up routing the che/codenvy runs on containers with certain names? When there is a ws that it does not know about it auto destroys it. There could also be an issue with multiple che or codenvy servers running on same daemon as well.

Example docker container name ` workspace20dk10jiriqw9qu5_machinesgckw3zwlvoo0rh7_admin_dev-machine` and docker image name `eclipse-che/workspace20dk10jiriqw9qu5_machinesgckw3zwlvoo0rh7_admin_dev-machine`

ghost:
Codenvy cleaner kills Che ws containers as it thinks they are orphaned.

JamesDrummond:
We should provide a different naming scheme where the server name is included like `workspace20dk10jiriqw9qu5_machinesgckw3zwlvoo0rh7_admin_dev-machine_codenvy_codenvy_1`. I am pretty sure che also kills codenvy ws too. Also, two codenvy servers or two codenvy servers on the same docker daemon would kill each other ws containers.

ghost:
You cannot have 2 Codenvy installations on one server. But otherwise, you are right. I wouldn't say it's a bug though since 2 Codenvy installations on one Docker daemon has never been a use case that we keep in mind. 2 Che installations - yes, possible - Che farm with servers on different ports. But I think there's already a similar issue. @TylerJewell and @garagatyi and myself were actively discussing it but I cannot find it now.

garagatyi:
Multiple CHEs would do the same.

garagatyi:
@eivantsov https://github.com/eclipse/che/issues/4178

garagatyi:
Fixes #4178

slemeur:
[...]

PR #4565, opened 2017-03-27
[WIP] Multiple Che Servers Cleanup Workspaces Incorrectly 

JamesDrummond:
When running two Che servers on same docker daemon, a che server will destroy/remove workspace containers belonging to the another che server. There is currently no way based on the normalized container name to determine a workspace containers associated che server parent.

This PR provides a different naming scheme for workspace machine containers where the server's environment variable `HOSTNAME` included

Current name schema for workspace machine container:
`workspace20dk10jiriqw9qu5_machinesgckw3zwlvoo0rh7_che_dev-machine`

New name schema for workspace machine container with environment variable `HOSTNAME=44538d8f467a`:
`workspace20dk10jiriqw9qu5_machinesgckw3zwlvoo0rh7_serverid-44538d8f467a_che_dev-machine`. 

https://github.com/eclipse/che/issues/4452

#### Changelog
Fix workspace machine cleaner to only clean containers belonging the che server that spawned it. This allows for multiple che servers to co-exist on the same docker daemon.  

#### Release Notes
Fix workspace machine cleaner to only clean containers belonging the che server that spawned it. This allows for multiple che servers to co-exist on the same docker daemon.  

Signed-off-by: James Drummond jdrummond@codenvy.com


JamesDrummond:
Need help with failing tests:
https://github.com/eclipse/che/pull/4565/files#diff-92e7162e3c949f42760601c74a8b864eR233
https://github.com/eclipse/che/pull/4565/files#diff-92e7162e3c949f42760601c74a8b864eR235
https://github.com/eclipse/che/pull/4565/files#diff-92e7162e3c949f42760601c74a8b864eR236
https://github.com/eclipse/che/pull/4565/files#diff-92e7162e3c949f42760601c74a8b864eR248

garagatyi:
@JamesDrummond IP validation failed. Have you sign eclipse agreement?

JamesDrummond:
Yes I have signed. I just checked too. Think it maybe I don't do commit -s which is not an option with eclipse che menus and direct edits in github.

image

garagatyi:
@JamesDrummond I worked on this PR as you asked and reworked it completely for code simplification.
Here is my current diff https://github.com/eclipse/che/compare/master...garagatyi:jamesFix?expand=1
BTW I found that It is not possible to use HOSTNAME variable because it can be not set in some environments.
I'll try to figure out another strategy. Probably with help of @riuvshin

JamesDrummond:
@garagatyi Feel free to merge into my branch when you are complete. What cases would HOSTNAME not be available? It defaults to container ID when not set. If it's not available could we just default to value `NA` or something similar?

garagatyi:
in some shells this variable is just not set. Also Che doesn't have to be run in container, so it is possible it will get situation when this variable is not available and we will get bug report. So I believe that we should not create this situation at all.

JamesDrummond:
Again if variable is not there default to a value. Not sure what shells you are referring to. If the server is run inside docker container the shell will always be the same. If che is run outside of docker container let's default to a value. In such a situation where running outside docker container I don't think multiple servers will be running anyways and we don't support this method to my knowledge. This is not the normal use of che so I am ok with it defaulting to a value.

JamesDrummond:
One thing though if the user sets this environment variable manually the regex search might fail. So we should check if HOSTNAME is valid. Maybe replace `_` in HOSTNAME value with `-` is a good idea to solve the possible problem.

garagatyi:
HOSTNAME is set in bash. In sh it is not set. And this shell is very widely used. 
I don't think that running multiple Che instances natively is invalid usecase. AFAIK native Che is valid usecase. 
BTW if we do this then cleaning up of orphaned machines will be broken since every restart container will get new hostname and Che will think that existing containers belong to another instance of Che.

JamesDrummond:
I am ok with the cleanup being "broken" in this way. Stopping workspace should be done gracefully. If not then it is up to user to cleanup all workspace container manually or may provide such functionality to be done manually through cli. 

Running multiple che instances natively is a invalid use case to me. If you feel differently we can start a chat in slack channel to discuss with others. The work around for developers not wanting to use docker container for server(not sure why you wouldn't) would be to set the environment variable manually each time `HOSTNAME=[unique name] [commands]`

FYI sh shell does provide HOSTNAME env var this too. 

```
>docker run -ti --rm alpine sh
/ # env | grep HOSTNAME
HOSTNAME=14c922b30b48
```

JamesDrummond:
Again if for some strange reason HOSTNAME is not there we set a default value. Also if there is a setup that causes this issue developer can set `HOSTNAME=[unique name] [commands]` natively as mentioned above or set `-e HOSTNAME=[unique name]` in docker command. Though I am not sure if cli will carry this over to to server container.

garagatyi:
I merged master into your branch and reworked it in accordance with my vision.
In current state it introduces posibility to set server ID but it is not set by default. 
If it is not set Che behaves in the same way as before.

garagatyi:
> I am ok with the cleanup being "broken" in this way

In that case we will have completely remove existing feature! Are you sure that we should do this?

garagatyi:
> Stopping workspace should be done gracefully. If not then it is up to user to cleanup all workspace container manually or may provide such functionality to be done manually through cli.

If you believe that we should do this then I suggest to discuss it with Tyler. If he agreed then fixing #4452 should be done in absolutely another way.

JamesDrummond:
We could do this another way which might take a little more effort but better in the end. We can check if any currently running workspace machine container has an associated running che servers before removing the workspace. This way if there are no che server containers matching workspace machine container serverId then it's default behavior is to remove unknown workspaces like it does currently. Truncated container ID of all containers can also be found by `docker ps -aq`. We can compare workspace machine container serverId with all running container ID's.

Native server I don't know how to do this but I don't care for this use case. It's currently this way now so I would like to complete the bug fix for container approach, then we can provide an enhancement/bug fix for native later on if to difficult to do now.

garagatyi:
Tyler said that we should support native Che in the save way we support Che in docker. 
@TylerJewell can you comment when you will be available whether we should implement something that brakes native Che and then try to fix it in following releases?

JamesDrummond:
@garagatyi OK looked at your code and I think I see where you are going. I made a few modifications that I need to test. I think it will give us both what we want.

TylerJewell:
Alex - am not able to comment and review this issue for a number of weeks.
Please consider discussing this with another maintainer, particularly
@benoitf @skabashnyuk.


skabashnyuk:
@TylerJewell I think this issue have to be revisited and(or) probably rewritten by @garagatyi  or other developer. We think this code will not work. Please let me know about priority of this  task.

TylerJewell:
Can you summarize, please, the range of problems that we are having and
would need to fix for multi-server che?  I thought we resolved all
remaining open issues.


skabashnyuk:
AFAIK proposed pr will not work.  That's why I suggest to set priority to take it into development.
> I thought we resolved all remaining open issues.

Unfortunately we don't.

TylerJewell:
Please document scenarios where it fails.

Tyler

skabashnyuk:
this code never works. It's just a proof of concept that is not working. There is no scenario when it's working.

skabashnyuk:
I'm proposing to close this pr and let @garagatyi  and co to implement it if we need this functionality.

TylerJewell:
I do not accept the blanket statement.  

The code works and does what was advertised. What we need are clear scenarios where multi Che is not starting or stopping, the root cause of the conflict and then identifying modifications to make to support it.

Let's not use blanket statements without supporting evidence.

Tyler

skabashnyuk:
ci-build

TylerJewell:
-1 on closing it

I want clear analysis and identification of scenarios that need to be fixed and a possible set of issues that can be scheduled in future before we close this pr.

Tyler

skabashnyuk:
@JamesDrummond  can you rebase this code and ensure that code is working as expected. It's quite old we need to make sure everything is working

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerContainerNameGenerator.java
```diff
@@ -100,12 +149,20 @@ public String getMachineId() {
         public String getWorkspaceId() {
             return workspaceId;
         }
+        
+        /**
+         * Return workspaceId of the docker container.
```
garagatyi, line 154: Returns `serverId`

TylerJewell:
-1 - please let's not have PMs responsible for rebasing code.  They cannot be guaranteed to make sure it is proper.  I am just concerned about consistency here.

Tyler

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerAbandonedResourcesCleaner.java
```diff
@@ -96,23 +96,27 @@ public void run() {
      * Cleans up CHE docker containers which don't tracked by API any more.
      */
     @VisibleForTesting
-    void cleanContainers() {
+    void cleanContainers(@Nullable @Named("che.server_id") String cheServerId) {
```
garagatyi, line 99: Looks like this code doesn't compile since this method is called without arguments

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerAbandonedResourcesCleaner.java
```diff
@@ -96,23 +96,27 @@ public void run() {
...
-                    } catch (NotFoundException e) {
-                        cleanUpContainer(container);
-                    } catch (Exception e) {
-                        LOG.error(format("Failed to check activity for container with name '%s'. Cause: %s",
-                                         containerName, e.getLocalizedMessage()), e);
+                if (optional.isPresent() ) {
+                    if(optional.get().getServerId().equals(cheServerId)){
+                        try {
+                            // container is orphaned if not found exception is thrown
+                            if(optional.get().getServerId().equals("") || ){
```
garagatyi, line 109: Assumption that it is empty string violates encapsulation principle. I suggest to not expose serverId identifier in ContainerNameInfo at all - this is information not needed for other classes. So in that case if server ID doesn't match `DockerContainerNameGenerator.parse(containerName)` can return empty optional and container is treated as non-matching.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerContainerNameGenerator.java
```diff
@@ -10,32 +10,64 @@
...
+                                                          MACHINE_ID_GROUP + "_" +
+                                                          "%s" + // che server ID group
+                                                          ".+$";
 
-    private static final String  CONTAINER_NAME_REGEX   = "^" + NODE_HOST_GROUP + WORKSPACE_ID_GROUP + "_" + MACHINE_ID_GROUP + "_.+$";
-    private static final Pattern CONTAINER_NAME_PATTERN = Pattern.compile(CONTAINER_NAME_REGEX);
+    private final Pattern containerNamePattern;
+    private final String  cheServerIdGroup;
+
+    @Inject(optional = true)
```
garagatyi, line 53: Injection of this property is not declared in the PR. So this PR in current state doesn't fix mentioned issue.

skabashnyuk:
Code is not compiling
```
[INFO] --- maven-compiler-plugin:3.3:compile (default-compile) @ che-plugin-docker-machine ---
[INFO] Changes detected - recompiling the module!
[INFO] Compiling 50 source files to /Users/sj/dev/src/codenvy/che/plugins/plugin-docker/che-plugin-docker-machine/target/classes
/Users/sj/dev/src/codenvy/che/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/DockerAbandonedResourcesCleaner.java:109: error: illegal start of expression
                            if(optional.get().getServerId().equals("") || ){
                                                                          ^
1 error
[INFO] BUILD FAILURE
```
@TylerJewell  can you clarify. Will @JamesDrummond  finish his work or plf team should finish this work? Who will test this code? Who will be responsible for it?

garagatyi:
@TylerJewell it doesn't compile not because of changes in master

davidwindell:
This bug will affect anyone using a basic compose setup with multiple che-servers on one machine like so https://github.com/eclipse/che/issues/4878#issuecomment-297589303

ghost:
@JamesDrummond @TylerJewell what are we going to do with this PR? The original problem - 2+ Che servers on one host  - is still there

davidwindell:
Any movement on this? I'm currently running a hacked assembly to work around the issue!

TylerJewell:
We will not work on this - it will be resolved by the move of che on kubernetes instead - and that will be the proposed scale out solution going forward.

davidwindell:
@TylerJewell Is there any more info about this move to Kubernetes and what that actually means? Our infrastructure has to run on AWS so this could be a deal-breaker for us. As an open source project I'd have thought we would be able to run it wherever we like :/

UPDATE: Just seen the blog article https://che.eclipse.org/multi-user-eclipse-che-kubernetes-6aef521a36f5

Here's hoping the Kubernetes "infrastructure choice" will work on any orchestrator/platform i.e. Docker Cloud and AWS ;)

JamesDrummond:
@davidwindell Personally, I think Kubernetes is a much better infrastructure than Docker swarm which is currently used. I am excited about Kubernetes implementation. Don't know much benefit's for single users cases of Che with Kubernetes but for Codenvy/Multi-User it will be great. Really with unless there are big changes to Codenvy that break usability for average/single user(don't see how) I would be looking at Codenvy now that it will be Eclipse Public Licensed!! This is great for many different reasons. Codenvy for the average user provides authentication and let's the user invite others to their person/private server that may have custom stacks or factories. As a workaround could you start multiple Docker daemons as discussed at [https://stackoverflow.com/questions/32334167/is-it-possible-to-start-multiple-docker-daemons-on-the-same-machine](https://stackoverflow.com/questions/32334167/is-it-possible-to-start-multiple-docker-daemons-on-the-same-machine)

@TylerJewell You can close this if you feel it will not be a problem in Kubernetes or is tracked elsewhere.

TylerJewell:
@davidwindell - the infrastructure choice of Docker or k8s should be cloud agnostic.  You should be able to pick any VM with any OS that supports Docker, install k8s, and it will work.  So your lock-in fears should go away.

There are many services that install k8s for you as well, and in the end you get a lot more flexibility in how the workloads will scale at the end of the day.

davidwindell:
It sounds like we'll be forced to move away from Docker Swarm if we want to continue using Che, I'll have to think on this some more but it's a shame you are dropping support.

TylerJewell:
We have never supported Docker Swarm with Che before.  We had Swarm as an embedded engine within Codenvy and you had to install Codenvy to get that embedded engine. Swarm did an ok job as it was simple to install, but it really fell down in some complex enterprise scenarios for us.  Migrating onto k8s will mean there is a large swath of enterprise issues we no longer have to worry about building ourselves.

davidwindell:
We took [this](https://eclipse.org/che/docs/setup/docker/), perhaps wrongly, as tantamount to support.

> The eclipse/che-server Docker image is appropriate for running Che within clusters, orchestrators, or by third party tools with automation.

TylerJewell:
We will still have Che on Docker support, but it will not include much of the Codenvy team management. You can run that single container as a single-user instance, and then deploy it onto many containers at the same time using Swarm. We just will not test the Swarm configuration for it.  We think of the single-container instance of Che as ideal for dev teams building Che or those that want their own private Che server.  That single container does not have any multi-user identity - so multi-user identity, server scale out, permissions, organizations, policies - these will all be in the multi-server packaging of Che.  This sort of capability requires proxies and database to run properly.  We will let you run this on Docker as well, but we will deploy multiple containers to achieve this without scalability.

davidwindell:
That's clearer, thanks. We have a working multi-user setup now simply using IP restriction.

The only thing that stops this working is the way Che does a cleanup. This would be *so* easy to fix by simply adding an additional string to the workspace container names so the Che server only cleans up those it identifies as belonging to itself.

For now, we just turn the cleanup off in a source hack.

TylerJewell:
Yeah, but the downside of that scenario is that you are over paying for resources as you have to spin up a che server for each user. The multi-user che with all of the codenvy capabilities built into it gives a complete management system on a single che server, and then delegate the workspaces to different pods that can each have their own scaling policy.

In terms of managing teams >5 people with any sort of precision, the che on k8s will offer so much more control and stability.

JamesDrummond:
@davidwindell Sorry if I confused you with Docker swarm that I was stating for Codenvy. However, K8s YAML config file in MY opinion is much more robust than Docker compose which Che currently uses to implement for multi-machine workspaces. Docker compose syntax can be used for single machine too but most people just use dockerfile syntax as compose if generally used/useful for multi-containers [http://www.eclipse.org/che/docs/devops/runtime-recipes/index.html](http://www.eclipse.org/che/docs/devops/runtime-recipes/index.html) . For better or worse I see Docker compose as almost being Docker swarm which you can see why from this article [https://codefresh.io/blog/deploy-docker-compose-v3-swarm-mode-cluster/](https://codefresh.io/blog/deploy-docker-compose-v3-swarm-mode-cluster/). I am hoping Codenvy devs will allow for multi-machine recipes/workspaces for Che to be defined by K8s YAML config syntax. K8s YAML config to me is a bit more complex from my short experience with it but I could do more things with it and clustering is easier to me with it. Sorry for lack of examples but I would just try K8s to see for yourself if you already have experience with Docker compose if you hven't tried it already[https://www.mirantis.com/blog/introduction-to-yaml-creating-a-kubernetes-deployment/](https://www.mirantis.com/blog/introduction-to-yaml-creating-a-kubernetes-deployment/).

davidwindell:
@TylerJewell It would still be great if - by fixing this issue - we/others could run our own proxy in front of multiple single-user servers after these changes are brought in as that would mean we wouldn't have to lock in to Kubernetes. We could then have (as we do now) one very simple Docker Compose file that runs on a single high-performance cloud box for up to 10 users.

You said...

> We will let you run this on Docker as well, but we will deploy multiple containers to achieve this without scalability.

...does that mean we WILL be able to continue using Docker Cloud (orchestrator for Docker Compose) AND/OR AWS Elastic Container Service to run our Che Cluster, just that we won't have the added scaling benefits etc of Kubernetes?

@JamesDrummond Do you know of any good Orchestrators for Kubernetes on AWS?

TylerJewell:
It's possible, but it's not something we would test or support because we will have it built into Che. The default multi-user deployment of Che on docker will include its own database and proxy. This is how Codenvy runs, and this functionality is moving into Che.  It will only require Docker and run only on a single server.

davidwindell:
Fingers crossed the default Codenvy cleanup won't cleanup others on the same Docker server then ;)

davidwindell:
I've come across [Rancher](http://rancher.com/) which looks super __awesome__ and supports Kubernetes AND Docker Swarm.

It would be great to see Che supporting Swarm and K8s

davidwindell:
A very good insight behind Redhat's.Kubernetes motivations https://www.theregister.co.uk/2017/07/18/kubernetes_anyone_but_aws/

JamesDrummond:
@davidwindell Good article :) . Also keep in mind that many enterprise customers need secure "cloud" solutions behind their firewall and cannot use services like AWS but needs scalability/portability with their own servers. Openshift from what I know(could be wrong) can be installed behind a firewall for enterprises even though Redhat to my knowledge is pushing their Openshift online services more. Check out [https://www.openshift.org/](https://www.openshift.org/) and [https://github.com/openshift/origin](https://github.com/openshift/origin). I used it briefly and it seems like its a cli/wrapper/frontend for Kubernetes to make your life easier.

Also take a look at [https://docs.openshift.org/latest/getting_started/administrators.html#running-in-a-docker-container](https://docs.openshift.org/latest/getting_started/administrators.html#running-in-a-docker-container) to see how to run Openshift Origins in a Docker container. This is how I actually started to use Openshift Origins but it's important to note that Docker running on Fedora, CentOS, and Red Hat Enterprise Linux (RHEL) hosts is required. I typically use Scaleways but had to use Linode to get this to work as Scaleways only supports Docker on Ubuntu.

davidwindell:
@JamesDrummond private cloud is a good point (thankfully not something we contend with). Rancher is a really good candidate here, It can run in a private cloud and can provision Kubernetes, Swarm, Mesos, etc clusters with a few clicks (including on RHEL). We're sold!

PR #4572, opened 2017-03-27
CODENVY-1936: Add alias index.docker.io for docker hub

mmorhun:
### What does this PR do?
Adds alias `index.docker.io` for docker hub container registry. This alias can be  used for configuration of `CHE_DOCKER_REGISTRY` and `CHE_DOCKER_REGISTRY_AUTH_<docker-hub-registry-name>_URL` properties.
Adds tests.

https://github.com/codenvy/codenvy/issues/1936

#### Changelog
Added alias `index.docker.io` for docker hub.

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerRegistryAuthResolver.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerRegistryAuthResolver.java
@@ -115,6 +115,10 @@ public class DockerRegistryAuthResolver {
             Map<String,AuthConfig> normalizedAuthConfigMap = new HashMap<>(authConfigs);
             normalizedAuthConfigMap.put(DEFAULT_REGISTRY, normalizedAuthConfigMap.remove("docker.io"));
             return normalizedAuthConfigMap;
+        } else if (authConfigs.containsKey("index.docker.io")) {
+            Map<String,AuthConfig> normalizedAuthConfigMap = new HashMap<>(authConfigs);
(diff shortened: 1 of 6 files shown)


ghost (approved):

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerRegistryAuthResolver.java
```diff
@@ -115,6 +115,10 @@ public String getXRegistryConfigHeaderValue(@Nullable AuthConfigs paramAuthConfi
             Map<String,AuthConfig> normalizedAuthConfigMap = new HashMap<>(authConfigs);
             normalizedAuthConfigMap.put(DEFAULT_REGISTRY, normalizedAuthConfigMap.remove("docker.io"));
             return normalizedAuthConfigMap;
+        } else if (authConfigs.containsKey("index.docker.io")) {
```
garagatyi: What If both "docker.io" and "index.docker.io" are present?

mmorhun: Then only `docker.io` will be respected.

garagatyi: Maybe it would be better to handle this situation?

mmorhun: ok, I wiil handle it

mmorhun:
ci-build

garagatyi (approved):

skabashnyuk (approved):

PR #4575, opened 2017-03-27
PR panel: Code refactoring

sunix:
### What does this PR do?
This is just an extract method for code readability:
There is a getBranchFromProjectMetadata() method that get the branch name from project metadata.
We set the branch name to the very same project metadata that will be retrieved later during the flow by getBranchFromProjectMetadata(). It make sense to name it setBranchToProjectMetadata so reader understand that we are playing with the same project metadata (whatever it is).

#### Changelog

PR panel: Clean code/refactoring. Extract method getBranchFromProjectMetadata()

### DIFF
--- a/plugins/plugin-pullrequest-parent/che-plugin-pullrequest-ide/src/main/java/org/eclipse/che/plugin/pullrequest/client/steps/InitializeWorkflowContextStep.java
+++ b/plugins/plugin-pullrequest-parent/che-plugin-pullrequest-ide/src/main/java/org/eclipse/che/plugin/pullrequest/client/steps/InitializeWorkflowContextStep.java
@@ -113,7 +113,7 @@ public class InitializeWorkflowContextStep implements Step {
                           .then(
                           (String branchName) -> {
                               context.setContributeToBranchName(branchName);
-                              context.getProject().getSource().getParameters().put("branch", branchName);
+                              setBranchToProjectMetadata(context.getProject(), branchName);
                           });
     }
 
@@ -129,6 +129,10 @@ public class InitializeWorkflowContextStep implements Step {
             }
         }
         return null;
+    }
+
+    private void setBranchToProjectMetadata(final ProjectConfig project, String branchName) {
+        project.getSource().getParameters().put("branch", branchName);
     }
 
     private Operation<PromiseError> errorSettingUpOriginRepoOp(final WorkflowExecutor executor, final Context context) {


skabashnyuk:
Describe goal of improvement?  It's not clear what and why you are doing. Why it's better then we have now?

sunix:
This is just an extract method for code readability:
There is a  getBranchFromProjectMetadata() method that get the branch name from project metadata.
We set the branch name to the very same project metadata that will be retrieved later during the flow by getBranchFromProjectMetadata(). It make sense to name it setBranchToProjectMetadata so reader understand that we are playing with the same project metadata (whatever it is).

skabashnyuk:
>This is just an extract method for code readability:

What if we just add a comment? 

Do you have any other rationality (like extracted method will be reused in other place) of this change because of my subjective taste this subjective improvement has subjective value. :) So I rather prefer to not to change unless other contributors (code maintainers prefered) think the same as you.

sunix:
I don't like comments in general ...
Here I tried to apply 3 advices you can find in Clean code (Robert C Martin):
- Small method size
- Avoiding comments that usually become obsolete
- Meaningful method name

voievodin:
Hi @sunix. Does it make sense to make this method a part of the `Context` like 
```
void setProjectParam(String name, String value) { ... } 

void setProjectBranchName(String branchName) { setProjectParam("branch", branchName); }
```
so it becomes reusable by other steps

sunix:
@evoevodin yes, it can be part of another pull request

mshaposhnik:
Pls reopen to master

voievodin:
> @evoevodin yes, it can be part of another pull request

Well, why not to make it a part of the context right now?
- it's still matches pr title
- it becomes reusable

skabashnyuk:
agree with @evoevodin

TylerJewell:
@sunix - status, please?  Would be nice to include this in the upcoming release and have it merged this week.

benoitf:
@sunix : what is the state ?

sunix:
@benoitf @TylerJewell to me it's ok to merge, it's a minor change and I won't have time to make more changes on this as requested by @mshaposhnik and @skabashnyuk.

benoitf:
@sunix it still requires a description/changelog/etc

TylerJewell:
Yes, let's get the meta information together.  The other improvements could be done as another PR per the reviewer's comments.

vparfonov:
@skabashnyuk @evoevodin What the problem here? Is it ok now, can you approve it?

voievodin:
The diff is the same, there is nothing wrong with it. My point is that it worth nothing to make the code which sets the branch name reusable by making it a part of the context.

voievodin (approved):
I'm fine with the changes

PR #4581, opened 2017-03-27
PR workflow: removing review factory step

sunix:
This is an open discussion. I think we should remove it from the default Github workflow as the Che server may not be accessible from outside and a factory URL to a non public Che server may confuse reviewers. Codenvy Github workflow could keep this factory if this is wanted.


benoitf:
I would prefer to have a flag to enable/disable

sunix:
@benoitf a server side boolean property flag ?

benoitf:
@sunix yes.

voievodin:
I like the changes, but you have to make _Codenvy_ to work as it does(keep generating factories).
So please either add configuration which by default says to generate factory or bind a separate workflow for Codenvy.

conceptually about localhost factories is up to @skabashnyuk to decide.

voievodin (approved):

skabashnyuk:
Agree with @evoevodin  Codenvy should work as it's now. I would like to keep flow same for che and codenvy

skabashnyuk:
>server may not be accessible from outside

It' depends. it may or may not.

sunix:
I am closing this one and am working on https://github.com/eclipse/che/pull/4645 to add the server side flag

Issue #4554, opened 2017-03-24
"Download sources" command fails quietly when artifact doesn't have sources


dmytro-ndp:
**Reproduction Steps:**
1. Create and open workspace from java stack.
1. Import project https://github.com/dmytro-ndp/java-with-external-libs.git and configure it as maven project.
1. Open **SimpleLogger.java** file in editor.
1. Set cursor on `log4jLogger.info("Info from log4jLogger");` command and open it's declaration (top menu _Assistant > Open Declaration_).
1. Wait until **Category** class is opened and then click on "**Download sources**" link at the top of the editor.

**Expected behavior**
IDE displays error message about impossibility to load sources, for example:
> Cannot download sources
> Sources not found for: log4j:log4j:1.2.12

**Actual misleading behavior**
IDE did nothing.
There was error message received from server on request to **http://172.19.20.32:33035/api/maven/server/download/sources?projectpath=/console&fqn=org.apache.log4j.Category**:
```{"message":"Internal Server Error occurred, error time: 2017-03-24 16:08:20"}```

![screenshot from 2017-03-24 18 08 46](https://cloud.githubusercontent.com/assets/1197777/24304063/d102bac6-10c0-11e7-9693-6269af4f4ff3.png)

**Screencast**:
![out](https://cloud.githubusercontent.com/assets/1197777/24304095/f16b13f8-10c0-11e7-9361-fa512b4ed7bd.gif)

There was next error message in the ws agent log
```
2017-03-24 16:32:14,961[nio-4401-exec-7]  [ERROR] [c.a.c.r.RuntimeExceptionMapper 47]   - Internal Server Error occurred, error time: 2017-03-24 16:32:14
java.lang.RuntimeException: java.rmi.ServerException: RemoteException occurred in server thread; nested exception is: 
	java.rmi.ServerException: RemoteException occurred in server thread; nested exception is: 
	java.rmi.UnmarshalException: error unmarshalling arguments; nested exception is: 
	java.lang.ClassNotFoundException: org.eclipse.aether.resolution.ArtifactResolutionException (no security manager: RMI class loader disabled)
	at org.eclipse.che.plugin.maven.server.MavenServerWrapper.perform(MavenServerWrapper.java:127) ~[che-plugin-maven-server-5.6.0-SNAPSHOT.jar:5.6.0-SNAPSHOT]
[...]

PR #4613, opened 2017-03-29
#4554 fix Class not found exception in maven server

evidolob:
### What does this PR do?
Fixes `ClassNotFoundException` when download maven artifact sources.

#4554

### DIFF
--- a/plugins/plugin-maven/maven-server/maven-server-impl/src/main/java/org/eclipse/che/maven/server/MavenServerImpl.java
+++ b/plugins/plugin-maven/maven-server/maven-server-impl/src/main/java/org/eclipse/che/maven/server/MavenServerImpl.java
@@ -425,8 +425,9 @@ public class MavenServerImpl extends MavenRmiObject implements MavenServer {
                                                                              new ArtifactRequest(RepositoryUtils.toArtifact(artifact),
                                                                                                  remoteRepositories, null));
         return MavenModelUtil.convertArtifact(RepositoryUtils.toArtifact(artifactResult.getArtifact()), localRepository);
-        } catch (ArtifactResolutionException e) {
-            MavenServerContext.getLogger().info(e);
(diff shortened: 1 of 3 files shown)


plugins/plugin-maven/maven-server/maven-server-impl/src/main/java/org/eclipse/che/maven/server/MavenServerImpl.java
```diff
@@ -425,8 +425,9 @@ public MavenArtifact resolveArtifact(MavenArtifactKey artifactKey, List<MavenRem
                                                                              new ArtifactRequest(RepositoryUtils.toArtifact(artifact),
                                                                                                  remoteRepositories, null));
         return MavenModelUtil.convertArtifact(RepositoryUtils.toArtifact(artifactResult.getArtifact()), localRepository);
-        } catch (ArtifactResolutionException e) {
-            MavenServerContext.getLogger().info(e);
+        } catch (ArtifactResolutionException ignored) {
+            //we need ignore exception, it's some times has class that client doesn't has
```
skabashnyuk, line 429: may be system.error.println()?

evidolob: All sdtout/stderr of the maven sever is logged by client, but 'ArtifactResolutionException' is normal when maven can't find or download artifact, so I not sure that this exception need to be logged.

skabashnyuk: up2you

vparfonov (approved):

dmytro-ndp (approved):
It definitely fixes issue #4554
screenshot from 2017-03-30 16 19 58

PR #4632, opened 2017-03-29
Do not prevent developers to use instance and config folder in package names

skabashnyuk:
### What does this PR do?
Removes ```instance``` and  ```config``` folders from gitignore section.
This ignore section was added in the time of che in docker development and it's not longer relevant.

 blocking developers to use words ```instance``` and  ```config``` as package names

#### Changelog
Allow developers to use words ```instance``` and  ```config``` as package names

#### Release Notes
not needed

#### Docs PR
not needed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,8 +9,6 @@ target/
 bin/
 test-output/
 maven-eclipse.xml
-instance/
-config/
 
 # Idea #
 ##################


garagatyi (approved):

riuvshin (approved):

vparfonov (approved):

gazarenkov (approved):

tolusha:
+1

benoitf:
I would have replace it by using prefix / so it only applies at root level

TylerJewell:
yes, I agree - that is a better solution.  Can you please adjust it, Sergii


garagatyi:
@TylerJewell does CLI still use 2 separate folders - instance and config?

TylerJewell:
The default is now just /instance.  But a user can separate /config from
/instance in rare circumstances if they configure it to do so, which would
be very rare.

PR #4637, opened 2017-03-30
PR workflow: flag to enable/disable the generation of the review factory url.

sunix:

 - `CHE_PULLREQUEST_GENERATE__REVIEW__FACTORY=true` in che.env
 -  `che.pullrequest.generate_review_factory=true` in che.properties


TylerJewell:
Sun Tan - the configuration property is too long.  I am not sure which PM
reviewed the property name, but let's come up with some different names.
Please reopen the PR and suggest some alternatives.

I'm on holiday for the next couple weeks and only have little time.  But
this is something that needs to be reviewed, appropriate comments added
into the properties template, and discussed further.


sunix:
@TylerJewell This is the good one: https://github.com/eclipse/che/pull/4645

PR #4641, opened 2017-03-30
PR workflow: flag to enable/disable the generation of the review factory url.

sunix:

 - `CHE_PULLREQUEST_GENERATE__REVIEW__FACTORY=true` in che.env
 -  `che.pullrequest.generate_review_factory=true` in che.properties


TylerJewell:
Sun Tan:

Please reopen this PR.  The property name is too long and needs to be
simplified and rationalized with other properties used. And then we need
the matching section put into the properties template. This review is
something that I would typically do, but I am out on holiday for another
couple weeks.

You can make some suggestions and show what the template looks like.  I do
not have access to github on my vacation - just some gmail email.


sunix:
@TylerJewell This is the good one: https://github.com/eclipse/che/pull/4645

PR #4645, opened 2017-03-30
PR workflow: flag to enable/disable the generation of the review factory url

sunix:

 - `CHE_PR=true` in che.env
 -  `che.pr=true` in che.properties

(PS: changed property name after discussion with Tyler)


plugins/plugin-pullrequest-parent/che-plugin-pullrequest-ide/src/main/java/org/eclipse/che/plugin/pullrequest/client/rest/PullRequestWorkflowServiceClientImpl.java
```diff
@@ -0,0 +1,43 @@
...
+import org.eclipse.che.api.promises.client.Promise;
+import org.eclipse.che.ide.api.app.AppContext;
+import org.eclipse.che.ide.rest.AsyncRequestFactory;
+import org.eclipse.che.ide.rest.StringMapUnmarshaller;
+
+import static org.eclipse.che.ide.MimeType.APPLICATION_JSON;
+import static org.eclipse.che.ide.rest.HTTPHeader.ACCEPT;
+import static org.eclipse.che.ide.rest.HTTPHeader.CONTENT_TYPE;
+
+public class PullRequestWorkflowServiceClientImpl implements PullRequestWorkflowServiceClient {
```
benoitf: class is missing description

plugins/plugin-pullrequest-parent/che-plugin-pullrequest-ide/src/main/java/org/eclipse/che/plugin/pullrequest/client/rest/PullRequestWorkflowServiceClientImpl.java
```diff
@@ -0,0 +1,43 @@
...
+    protected AsyncRequestFactory asyncRequestFactory;
+    protected String              baseHttpUrl;
```
benoitf: do we need those fields with protected ?

plugins/plugin-pullrequest-parent/che-plugin-pullrequest-ide/src/main/java/org/eclipse/che/plugin/pullrequest/client/steps/GenerateReviewFactoryStep.java
```diff
@@ -35,30 +39,45 @@
...
         this.notificationManager = notificationManager;
+        this.pullRequestWorkflowService = pullRequestWorkflowService;
         this.factoryService = factoryService;
     }
 
     @Override
     public void execute(final WorkflowExecutor executor, final Context context) {
+        pullRequestWorkflowService.getSettings() //
+                                  .then((Map<String, String> settings) -> {
+                                      if (!Boolean.parseBoolean(settings.getOrDefault(CHE_PULLREQUEST_GENERATE__REVIEW__FACTORY, "true"))) {
```
benoitf: I think we don't need a map there, we can inject a property with name "che.pullrequest.generate_review_factory"

plugins/plugin-pullrequest-parent/che-plugin-pullrequest-server/src/main/java/org/eclipse/che/plugin/pullrequest/server/PullRequestWorkflowService.java
```diff
@@ -0,0 +1,51 @@
...
+
+import com.google.common.collect.ImmutableMap;
+
+import static org.eclipse.che.plugin.pullrequest.shared.Constants.CHE_PULLREQUEST_GENERATE__REVIEW__FACTORY;
+import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
+
+@Path("/pullrequestwf")
+@Api(value = "/pullrequestwf", //
+     description = "Pull Request Workflow REST API")
+public class PullRequestWorkflowService {
```
benoitf: missing javadoc

sunix:
About the name of the property, i suggest:
- `che.pullrequest.review_url`
- `che.pullrequest.review_factory`
- `che.pr.generate_review_factory`

TylerJewell:
Too complicated, too long, and inconsistent,  I would like time to think
about it more - can this wait until I am off holiday?

Do we have explanations for all three parameters and when / how / why they
are needed?  What would the docs say on them - that will help me think of
the names.


plugins/plugin-pullrequest-parent/che-plugin-pullrequest-ide/src/main/java/org/eclipse/che/plugin/pullrequest/client/steps/GenerateReviewFactoryStep.java
sunix: We can have the rest end point `pullrequestwf/reviewurl` that returns true or false.
pullRequestWorkflowService.shouldGenerateReviewUrl() that returns a boolean

sunix:
@TylerJewell https://www.youtube.com/watch?v=5BrajzHIaxs
these are suggestions to generate or not the review factory link when creating a pullrequest from the panel

plugins/plugin-pullrequest-parent/che-plugin-pullrequest-ide/src/main/java/org/eclipse/che/plugin/pullrequest/client/steps/GenerateReviewFactoryStep.java
benoitf: IMHO looks better for me

sunix: ok will update the PR soon with this change

sunix:
@TylerJewell is it ok if we keep the current name `che.pullrequest.generate_review_factory` ? and change it after if needed ?

TylerJewell:
It is not ok - as it creates backward compatibility issues.  This needs to hold until there is appropriate reviews.

Tyler

TylerJewell:
Unfortunately github is blocked for me down here so am having to do some reviews without the typical tools at my fingertips.

Can you send over the che.env comment desrcription and name for all three of the proposed properties?  I will try to do a write up this weekend.

Tyler

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -137,6 +137,14 @@ che.auth.access_denied_error_page=/error-oauth
 # Reserved user names
 che.auth.reserved_user_names=
 
+### Pull Request Generation
+# If true (default), then generates a Factory URL embedded within the PR
+# generated by Che for provider that manages the pull request. The PR
+# factory is a URL that generates a workspace that contains the commit of
+# the PR itself. Setting this to false will generate a PR without an
+# embedded Factory URL. Generate PRs using the pull request panel in the IDE.
+che.pr=true
```
voievodin, line 146: > CHE_PR=true in che.env
> che.pr=true in che.properties
> (PS: changed property name after discussion with Tyler)

I have no idea why the property called `che.pr` is clearer than `che.pr.generate_review_factory`.
First thing i though was that this property enables or disables PR panel appearance.
if we need a short name for it, let's make it at least related to factory e.g. `che.pr.factory=true` 

wdyt @TylerJewell

voievodin:
@sunix Have you considered named strategies for generating factories.
e.g.
```
factory_gen_strategy=none // <- not generated
factory_gen_strategy=workspace // <- default
factory_gen_strategy=repo // <- generated from repo
```
what do you think about it?

mshaposhnik:
Pls reopen on master now

benoitf:
I would prefer to keep the PR and only change the target branch

vparfonov:
@sunix is it still actual ?

sunix:
not needed anymore thanks

PR #4652, opened 2017-03-30
Use pods instead of deployments for workspaces on OpenShift

amisevsk:
### What does this PR do?
- Replace deployments with pods in OpenShiftConnector.
- Set pod restartPolicy to "Never". Before, pods would get restarted but be stuck, since che-server wouldn't know to restart the agents. Deployments do not support "never" as a restart policy, since their purpose is to make sure their pods are always running.
- Clean up some methods since we can now get the workspace pod directly.


l0rd (approved):

TylerJewell:

amisevsk:
Closing this PR as using terminating Pods has issues with some OpenShift setups. Changes superceded by https://github.com/eclipse/che/pull/4838

Issue #4516, opened 2017-03-22
Add new goal with existing name 


Ohrimenko1988:
This is issue related with epic https://github.com/eclipse/che/issues/4511

In the command explorer we have folders who named on upper register , but in the edit command form in the "Goal" drop-down list we have names in the down register , and if try to type existing folder name with upper register in the "Goal" field and click "Save" button , we will see that new command was created not in this folder , but in its duplicate .
**Reproduction Steps:**
- Go to the web - java - spring project 
- Click on "Commands" tab in the left panel
- Click "+" button on any goal 
- Choose any command type
- In the opened commands editor go to the "Goal" text field and input exist goal name in upper register
- Click "Save" button .
- Pay attention to the command explorer .

**Expected behavior :**
New command will be added to the inputted goal .

**Observed behavior :**
New command was added in the new created goal

**See attachment :**

![anim](https://cloud.githubusercontent.com/assets/22724619/24201390/91b406c6-0f18-11e7-9f3e-7a4ea04c032f.gif)
al with inputted name .

slemeur:
I agree with that bug.

At this moment it's not critical or blocker as we are reworking the ability to create a new goal as well. But this need to be handled as well.

I'm linking the issue with: https://github.com/eclipse/che/issues/4417

azatsarynnyy:
Fixed by https://github.com/eclipse/che/pull/4656

PR #4656, opened 2017-03-31
CHE-4516: Check command goal name uniqueness ignoring string case considerations on new goal creation

azatsarynnyy:
### What does this PR do?
Since Commands Explorer displays all goals names in uppercase we need to check newly created goal name for uniqueness ignoring string case considerations in order to avoid situations like on this screenshot:
image

So this PR adds checking goal name uniqueness ignoring string case considerations on new goal creation.

Now user will be proposed to enter another goal's name:
goal

#4516 

#### Changelog
Now all command goals names must be unique ignoring string case considerations.

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/command/editor/page/goal/GoalPage.java
+++ b/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/command/editor/page/goal/GoalPage.java
@@ -21,6 +21,7 @@ import org.eclipse.che.ide.command.editor.EditorMessages;
 import org.eclipse.che.ide.command.editor.page.AbstractCommandEditorPage;
 import org.eclipse.che.ide.command.editor.page.CommandEditorPage;
 
+import java.util.Optional;
 import java.util.Set;
 
 /**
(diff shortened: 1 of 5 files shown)


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/command/editor/page/goal/GoalPage.java
```diff
@@ -60,13 +61,10 @@ public IsWidget getView() {
 
     @Override
     protected void initialize() {
-        final String goalId = editedCommand.getGoal();
-        final CommandGoal goal = goalRegistry.getGoalForId(goalId);
-
-        goalInitial = goal.getId();
+        goalInitial = editedCommand.getGoal();
```
dkulieshov: Shouldn't it be `initialGoal`? I'm a bit off the context, but this name seems quite odd to me.

azatsarynnyy: Sure, thx. Fixed

vparfonov (approved):

PR #4662, opened 2017-03-31
Fix: Add Nullable parameter to che.docker.ip.external in OpenShiftConnector

amisevsk:
### What does this PR do?
The property che.docker.ip.external can be null, but OpenShiftConnector does not include the annotation. This prevents Che from initialising if e.g. running on docker without the property set.

#### Changelog
Add Nullable parameter to che.docker.ip.external in OpenShiftConnector

### DIFF
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
@@ -178,7 +178,7 @@ public class OpenShiftConnector extends DockerConnector {
                               DockerConnectionFactory connectionFactory,
                               DockerRegistryAuthResolver authResolver,
                               DockerApiVersionPathPrefixProvider dockerApiVersionPathPrefixProvider,
-                              @Named("che.docker.ip.external") String cheServerExternalAddress,
+                              @Nullable @Named("che.docker.ip.external") String cheServerExternalAddress,
                               @Named("che.openshift.project") String openShiftCheProjectName,
                               @Named("che.openshift.liveness.probe.delay") int openShiftLivenessProbeDelay,
                               @Named("che.openshift.liveness.probe.timeout") int openShiftLivenessProbeTimeout,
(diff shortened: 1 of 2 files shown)


plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -982,6 +982,9 @@ private void createOpenShiftRoute(String serviceName,
                                       String workspaceName) {
 
         String routeName = CHE_OPENSHIFT_RESOURCES_PREFIX + workspaceName + "." + serverRef;
+        if (cheServerExternalAddress == null) {
+            throw new NullPointerException("Property che.docker.ip.external must be set when using openshift.");
```
l0rd: I would rather use `IllegalArgumentException` here

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -178,7 +178,7 @@ public OpenShiftConnector(DockerConnectorConfiguration connectorConfiguration,
                               DockerConnectionFactory connectionFactory,
                               DockerRegistryAuthResolver authResolver,
                               DockerApiVersionPathPrefixProvider dockerApiVersionPathPrefixProvider,
-                              @Named("che.docker.ip.external") String cheServerExternalAddress,
+                              @Nullable @Named("che.docker.ip.external") String cheServerExternalAddress,
```
l0rd, line 181: Shouldn't you use @NotNull instead?

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
amisevsk: Done.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
amisevsk: It has to be `Nullable` to match the `ServerEvaluationStrategies`. Otherwise we break functionality on the Docker side, since this class is injected at bootstrap.

l0rd: Ok

l0rd (approved):

PR #4676, opened 2017-04-03
Fix php56 name

ghost:
#### Changelog
Fix php56 stack name so that it does not conflict with the default php stack name.


skabashnyuk (approved):

slemeur:
link to https://github.com/eclipse/che/pull/4633

Issue #4507, opened 2017-03-22
NPE in FileWatcherService


mmorhun:
In some cases after deleting of project subdirectories NPE in `FileWatcherService` occurs.

**Reproduction Steps:**
1. Create a workspace from Java stack
2. Clone https://github.com/jboss-developer/jboss-eap-quickstarts.git and configure as a blank project
3. Unfold the project and delete all files and folders using one operation (use shift and left mouse button to select  all files faster)
4. Click on `dev-machine` tab in `Processes` panel and wait a bit.

**OS and version:**  Ubuntu 16.04
**Additional info:**
Problem can be reliable reproduced.
CHE version: 5.6.0-SNAPSHOT
Logs with NPE: 
[FileWatcherService_NPEs.txt](https://github.com/eclipse/che/files/861189/FileWatcherService_NPEs.txt)

PR #4691, opened 2017-04-04
NPE in FileWatcherService

dkulieshov:
### What does this PR do?
Fixed NPE in file watcher service and tree file walker. Exceptions happen because of processing of stale file descriptors, in other words in this case we try to process already removed files. This mostly happens on massive file move/remove operations and normally should be ignored.

https://github.com/eclipse/che/issues/4507

#### Changelog
Fixed NPE in file watcher service that appeared to happen in some rare cases


svor (approved):

RomanNikitenko (approved):

ghost (approved):
Tested with angular2 project which is auto generated. No npe or no such file errors. Good job @dkuleshov

vparfonov (approved):

Issue #4501, opened 2017-03-21
[Factory] Replace factory buttons


slemeur:
The factory buttons have not been used a lot.
It’s probably good time to remove the service that we have, which was allowing to use logo in the button - and just provide an SVG badge as on dashboard. 
We would remove their mention from docs too.

We also need to replace the factory button that we are providing from the dashboard.

skabashnyuk:
@slemeur is that ok for this issue if we marked
```
    @GET
    @Path("/{id}/image")
```
and 
```
    @POST
    @Consumes(MULTIPART_FORM_DATA)
```
as deprecated 
Is that the goal of this issue?

slemeur:
Yes for deprecating.

I will also update the buttons styles (I planned to do that today)

slemeur:
Close: will be fixed with #2049

PR #4694, opened 2017-04-04
Deprecate some FactoryService methods related to factory button 

skabashnyuk:
### What does this PR do?
Deprecate FactoryService methods related to factory button 

https://github.com/eclipse/che/issues/4501

#### Changelog
Methods related to factory button functionality in FactoryService become deprecated


slemeur:
@skabashnyuk you forgot to sign-off ;)

slemeur (approved):

skabashnyuk:
done

JamesDrummond:
@skabashnyuk Will this deprecation effect users? If so we need to follow https://github.com/eclipse/che/wiki/Development-Process#deprecation-policy I would think. Need release notes also if it effects users.

skabashnyuk:
@JamesDrummond no effect. For che for sure. For now it's only declaration. Methods will be removed in next versions

JamesDrummond (approved):

PR #4695, opened 2017-04-04
Fix fetch user return promise

ashumilova:
Signed-off-by: Ann Shumilova <ashumilova@codenvy.com>

### What does this PR do?

Reworks fetch current user to return promise and handle 304.
#### Changelog
[UD] reworked fetch current user to return promise on error and handle 304 status properly.

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/dashboard/src/components/api/che-user.factory.ts
+++ b/dashboard/src/components/api/che-user.factory.ts
@@ -373,11 +373,15 @@ export class CheUser {
   fetchUser(): ng.IPromise<any> {
     let promise = this.remoteUserAPI.get().$promise;
     // check if if was OK or not
-    promise.then((user: che.IUser) => {
+    return promise.then((user: che.IUser) => {
       this.user = user;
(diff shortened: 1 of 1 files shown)


akurinnoy (approved):

dashboard/src/components/api/che-user.factory.ts
```diff
@@ -373,11 +373,15 @@ export class CheUser {
   fetchUser(): ng.IPromise<any> {
     let promise = this.remoteUserAPI.get().$promise;
     // check if if was OK or not
-    promise.then((user: che.IUser) => {
+    return promise.then((user: any) => {
```
benoitf: user is no longer of che.IUser type ?

ashumilova: no, my mistake - fixed

benoitf (approved):

PR #4702, opened 2017-04-04
Nocmdimage

### LINKED ISSUE #4310: Improve UX of usage of docker images that are source of containers that exits right after start
Now if user uses some image that doesn't have `cmd`, `entrypoint` or have shell interpreter there Che shows weird error on workspace start. 

For such cases we can inspect image and if it is such image set entrypoint/cmd into container.
Since we can't detect all possible variants where container exist immediately after start we should also inspect container after start and if it has state stopped then return pretty error with the explanation on why container exited.

This enhancement was asked by @benoitf and @skryzhny 

**Reproduction Steps:**
1. Create WS from image ubuntu
1. Start this WS

garagatyi:
### What does this PR do?
Improves UX when user used docker image that needs additional command or use interactive command/entrypoint by default. E.g. ubuntu, openjdk, etc.
In some cases we can recognize this situation and change command to fix it.
When we can't recognize this situation but container stops right after start we send custom error message that explains why that error happens. 
Here are a few screenshots on how it looks:
<details>

selection_051

selection_050

selection_047

</details>

Fixes #4310 

#### Changelog

Adds guidance for user when containers exit unexpectedly in certain situations.

#### Release Notes

**Improved UX For Unexpected Container Exits**
We have created additional guidance for users when machines exit immediately after startup. This is often caused because the machine recipe requires interaction or additional arguments to start. Although we can't prevent the machine from stopping we now display a popup notification to the user asking them to check the machine recipe for commands that would require interaction or additional arguments.

#### Docs PR

N/A

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
@@ -12,6 +12,7 @@ package org.eclipse.che.plugin.docker.machine;
 
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Strings;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 
 import org.eclipse.che.api.core.NotFoundException;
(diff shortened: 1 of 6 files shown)


benoitf (approved):
nice UX improvement

skabashnyuk (approved):

mmorhun (approved):

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -591,6 +617,39 @@ private void addSystemWideContainerSettings(String workspaceId,
...
+            containerConfig.setEntrypoint((String[])null);
+        }
+    }
+
+    // Inspect container right after start to check if it is running,
+    // otherwise throw error that command should not exit right after container start
+    protected void checkContainerIsRunning(String container) throws IOException, ServerException {
+        ContainerInfo containerInfo = docker.inspectContainer(container);
+        if ("exited".equals(containerInfo.getState().getStatus())) {
+            throw new ServerException("We detected that machine container exited right after start. " +
```
bmicklea: Please change message to: 
>We detected that a machine exited unexpectedly. This may be caused by a container in interactive mode or a container that requires additional arguments to start. Please check the container recipe.

bmicklea (commented):
Also updated changelog and release notes.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
garagatyi: done

garagatyi:
@bmicklea please add in changelog and release notes another improvement: we now allow to run official images such as ubuntu because we detect that it will exit right after start and change the command in container. It is described in PR description.

bmicklea (approved):

Issue #3914, opened 2017-01-26
Error Launching SSH Agent


hypertoast:
My server is accessible via a PublicIP. I can connect to the dashboard and go to create a new workspace. It fails at the "Launching SSH Agent". I have tried running it on multiple OS (Ubuntu, Debian, CentOS). All have the same problem... Any help is appreciated.

**Reproduction Steps:**
Create a workspace (any stack)
Workspace loads until the following line appears: [STDOUT] ssh-keygen: generating new host keys: RSA1
The error popup will appear after a few minutes
"Unable to start workspace agent. Error when trying to start the workspace agent: Start of environment 'default' failed. Error: Fail launching agent SSH. Workspace ID:workspacesrla9fgjiux3nutt"

**Che version:**    
Your CLI version is '5.1.2'.
Your installed version is '5.1.2'.

**OS and version:**    
OS: CentOS Linux release 7.3.1611 (Core)

**Docker version:**    
Client:
 Version:      1.13.0
 API version:  1.25
 Go version:   go1.7.3
 Git commit:   49bf474
 Built:        Tue Jan 17 09:55:28 2017
 OS/Arch:      linux/amd64

Server:
 Version:      1.13.0
 API version:  1.25 (minimum version 1.12)
 Go version:   go1.7.3
 Git commit:   49bf474
 Built:        Tue Jan 17 09:55:28 2017
 OS/Arch:      linux/amd64
 Experimental: false

**Eclipse Browser Window Output:**
`[STDOUT] 2017-01-26 14:39:56,860[main] [INFO] [o.a.c.http11.Http11NioProtocol 488] - Starting ProtocolHandler ["http-nio-4401"]
[STDOUT] 2017-01-26 14:39:56,865[main] [INFO] [o.a.catalina.startup.Catalina 642] - Server startup in 7747 ms
[STDOUT] 2017-01-26 14:39:57,005[nio-4401-exec-2] [WARN ] [o.e.c.impl.ContainerResponse 93] - Client has aborted connection. Response writing omitted.
[STDOUT] 2017-01-26 14:39:57,005[nio-4401-exec-1] [WARN ] [o.e.c.impl.ContainerResponse 93] - Client has aborted connection. Response writing omitted.
[dev-machine] [STDOUT] ssh-keygen: generating new host keys: RSA1`

**Che cli.log output:**  
[...]

PR #4708, opened 2017-04-05
CHE-3914: Fix launching SSH agent in some cases

mmorhun:
### What does this PR do?
Fixes launching SSH agent error, when according to security setting of host OS, a user in a docker container cannot see processes, which belongs to container's root.

Before this PR, to check whether SSH agent is started successfully we checked two things:
- SSH process is running inside container
- exposed port for SSH connection is opened and has been listening

But the first check is redundant, because if ssh fail to run it won't stand on port and second check will fail. More, that check could prevent workspace start when it shouldn't.
So, this PR just deletes redundant checking of launching ssh process in workspace container. Check that port has been listening is enough here.

#3914 

#### Changelog
Fixed SSH agent launch error when host security prevents a user from seeing container root's processes.

#### Release Notes
N/A

#### Docs PR
N/A


skabashnyuk (commented):
up2 @garagatyi

TylerJewell:
changelog needs to be more explicity on what is being fixed (which case)
vs. the generic "some cases"


garagatyi (approved):

bmicklea (approved):
updated changelog comment

voievodin (approved):

PR #4723, opened 2017-04-06
CHE-4721: fix agents install scripts

### LINKED ISSUE #4721: Agent installation scripts fail to download binaries if both curl/wget are not installed in container
We merged an [improvement](https://github.com/eclipse/che/commit/950dd692c0148da45e99b48b4a1f32cf7a0ad7e7) that allowed to use wget instead of curl in agents installation scripts.
Looks like this feature wasn't tested on machines when no wget/curl are present.
Now start of exec agent fails because of that.
Agents scripts that should be fixed:
- exec
- terminal
- wsagent

**Reproduction Steps:**
1. Create workspace with a machine from image that don't have neither curl nor wget.
1. Start workspace

**OS and version:**    
Ubuntu 16.04

**Diagnostics:**

garagatyi:
### What does this PR do?
Fixes agents installation scripts in case neither of curl/wget is installed in machine.

Fixes #4721 

#### Changelog
Fix agents installation scripts in case neither of curl/wget is installed in machine.


riuvshin (approved):

skabashnyuk (approved):

riuvshin:
Thanks @garagatyi !

Issue #4497, opened 2017-03-21
[Bug] Snapshot removing failed. Snapshot attributes are not valid


rhughes89:
Cannot delete snapshots from the remote registry at docker hub. Pushes and pulls to registry just fine, however it fails to delete old snapshots from registry resulting in ALOT of orphan snapshots in registry. I have a multi-machine workspace consisting of 5 machines and it left around 40 orphan snapshots over time.

**Reproduction Steps:**

- configure che to work with a docker hub organization repository for saving snapshots 
- start a multi-machine workspace
- stop the workspace, verify that it saves to registry
- start the workspace and load from snapshot
- stop the workspace without saving snapshot

**OS and version:**    
- Che running on AWS EC2 t2.medium. 
- Ubuntu 16.04
- Eclipse Che 5.4

**Diagnostics:** 
Log:
```
2017-03-21 18:16:26,551[aceSharedPool-3]  [ERROR] [o.e.c.a.w.s.WorkspaceRuntimes 529]   - Couldn't remove snapshot 'snapshot7r82ivljtss3ls1n', workspace id 'workspacem52qibob1zahcdiy'
org.eclipse.che.api.machine.server.exception.SnapshotException: Snapshot removing failed. Snapshot attributes are not valid
        at org.eclipse.che.plugin.docker.machine.DockerInstanceProvider.removeInstanceSnapshot(DockerInstanceProvider.java:149) ~[che-plugin-docker-machine-5.4.0-SNAPSHOT.jar:5.4.0-SNAPSHOT]
        at org.eclipse.che.api.environment.server.CheEnvironmentEngine.removeSnapshot(CheEnvironmentEngine.java:557) ~[che-core-api-workspace-5.4.0-SNAPSHOT.jar:5.4.0-SNAPSHOT]
        at org.eclipse.che.api.workspace.server.WorkspaceRuntimes.removeBinaries(WorkspaceRuntimes.java:508) [che-core-api-workspace-5.4.0-SNAPSHOT.jar:5.4.0-SNAPSHOT]
        at org.eclipse.che.api.workspace.server.WorkspaceRuntimes.removeBinaries(WorkspaceRuntimes.java:525) [che-core-api-workspace-5.4.0-SNAPSHOT.jar:5.4.0-SNAPSHOT]
[...]

PR #4744, opened 2017-04-10
CHE-4497: Delete snapshots from docker hub

mmorhun:
### What does this PR do?
Distinguish Docker Hub as special registry on snapshot deletion operation and adds corresponding authorization header to the request. This adds ability to automatically deletion of workspace snapshots from Docker Hub.
This implementation uses v1 registry API because Docker Hub doesn't expose corresponding API method of v2 registry API.

https://github.com/eclipse/che/issues/4497

#### Changelog
Added ability to delete workspace snapshots from Docker Hub.

#### Release Notes
N/A

#### Docs PR
N/A


garagatyi (approved):

bmicklea (commented):
I don't think this changes the workflow for users does it?  They still use the product as they have, but we clean up things better if they're using DockerHub as their registry - correct?

mmorhun:
@bmicklea yes, you are right

bmicklea (approved):
Removed release note entry - it's not needed because this doesn't affect user workflow.

skabashnyuk (approved):

PR #1335, opened 2016-05-23
Fix project.version issue in dependencyManagement: Use maven property…

sunix:
… for che version instead. Do not force maven child projects to use the Che version.

PR #4752, opened 2017-04-10
Fix Maven dependency manangement: Use ${che.version} instead of ${project.version} to allow projects with different version to use Che dependencies.

sunix:
### What does this PR do?
Fix Maven dependency manangement: Use ${che.version} instead of ${project.version} to allow projects with different version to use Che dependencies.

Updates #1335

#### Changelog

Fix Maven dependency manangement: Use ${che.version} instead of ${project.version} to allow projects with different version to use Che dependencies.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -856,7 +856,7 @@
             <dependency>
                 <groupId>org.eclipse.che.plugin</groupId>
                 <artifactId>che-plugin-requirejs</artifactId>
-                <version>${project.version}</version>
+                <version>${che.version}</version>
             </dependency>
             <dependency>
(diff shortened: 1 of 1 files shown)


skabashnyuk (approved):

riuvshin:
what about this one https://github.com/sunix/che/blob/a5539bbab232464a14282eb6cc15d3df75d19eb5/pom.xml#L859 ?

sunix:
@riuvshin thanks ... :) will fixup

riuvshin (approved):

sunix:
@JamesDrummond why `target/branch` label ?

JamesDrummond:
My mistake.

PR #4760, opened 2017-04-11
Create workspace dir in server instead of workspace pod on Openshift

amisevsk:
### What does this PR do?
Create workspace directory on PVC before launching workspace instead of starting a Pod to create the directory. This step is necessary because if a workspace directory does not exist, it will be created with root permissions when mounted on a subpath.

Assumes that the workspace PVC is mounted in the server pod. In addition to mounting the PVC on che-server, the property `che.openshift.workspace.pvc.mount.path` must match the mount path of the workspace PVC.

See: https://github.com/fabric8io/fabric8-online/pull/140


l0rd (approved):
Thank you Angel! That improves a lot the time needed to start a workspace.
My only concern is that with this solution workspaces share their PV with che server. And since PV is RWO that implies that the server and the workspaces need to run on the same node. We should probably release the PV before starting the workspace...and claim it again once a workspace is stopped.
But I approve it as it is since that's an improvement with respect to the pre-existing situation. And we will discuss how to implement further improvements.

ibuziuk (approved):

amisevsk:
Closing this issue as it won't work with RWO Persistent Volumes.

PR #4762, opened 2017-04-11
Avoid System.exit calls that kills the whole Maven jvm or IDE running it

### LINKED ISSUE #4458: Use of exec-maven-plugin:exec:java with System.exit() kills build silently
In many project, for example in /che/assembly/assembly-ide-war/pom.xml, we use the exec:java goal of the maven plugin "exec-maven-plugin".
In this particular instance, the maven plugin call org.eclipse.che.util.GwtXmlGenerator. This main class catches exceptions and call System.exit(1). Since the exec:java goal does not start a separate VM to execute the main class, this immediately terminates the maven process. 
The problem is much worse when using the Eclipse IDE, since in that case, the System.exit() happens inside the process that runs the IDE. Basically, running an incremental build reilably terminates the IDE.
I see three options:
1. Don't use the exec-maven plugin, but instead write mojos like DtoGeneratorMojo, for example
2. Don't call System.exit(), but instead throw an exception. I would expect the exec:java goal to catch excptions and handle them properly.
3. Use exec:exec instead of exec:java

sunix:
### What does this PR do?
Removing System.exit calls that kill the whole Maven jvm or IDE running them.

Fix #4458

#### Changelog
Fix maven build execution that kills Eclipse IDE when building Che

### DIFF
--- a/ide/che-core-ide-generators/src/main/java/org/eclipse/che/util/DtoFactoryVisitorRegistryGenerator.java
+++ b/ide/che-core-ide-generators/src/main/java/org/eclipse/che/util/DtoFactoryVisitorRegistryGenerator.java
@@ -44,22 +44,17 @@ public class DtoFactoryVisitorRegistryGenerator {
      * Entry point. --rootDir is the optional parameter.
      *
      * @param args
+     * @throws IOException 
      */
-    public static void main(String[] args) {
-        try {
-            File rootFolder = GeneratorUtils.getRootFolder(args);
-            System.out.println(" ------------------------------------------------------------------------ ");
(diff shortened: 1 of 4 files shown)


skabashnyuk:
How did you test this code?

skabashnyuk (approved):

vparfonov (approved):

sunix:
@davidfestal @tsmaeder can you try test this ?

vrubezhny (approved):
Works fine for me with Eclipse Oxygen

skabashnyuk:
@sunix  I meant, have you tested master Che to work with your changes?

PR #4799, opened 2017-04-12
CHE-180: Creating and closing OpenShiftClient in every method of Open…

ibuziuk:
### What does this PR do?
Potential fix for https://issues.jboss.org/browse/CHE-180

https://issues.jboss.org/browse/CHE-180

### Docker image
ibuziuk/che-server:closeable - https://hub.docker.com/r/ibuziuk/che-server/tags/

### DIFF
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
@@ -162,7 +162,6 @@ public class OpenShiftConnector extends DockerConnector {
 
     private Map<String, KubernetesExecHolder> execMap = new HashMap<>();
 
-    private final OpenShiftClient openShiftClient;
     private final String          openShiftCheProjectName;
     private final int             openShiftLivenessProbeDelay;
     private final int             openShiftLivenessProbeTimeout;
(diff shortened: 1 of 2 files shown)


ibuziuk:
@amisevsk @l0rd please review the code meticulously (formatting was really monotonous task, so I might made some typos)

ibuziuk:
Seems to work fine so far on minishift locally, trying on free-int now

amisevsk (approved):
LGTM. I've tested the changes and it looks like everything works.

I'm still worried about the overhead of starting so many instances of OpenShiftClient, but this is a lot better than the IOExceptions we were facing before.

ibuziuk:
@amisevsk I believe there is no need to worry about creating many instances of client. OpenShift client should be treated as DB connection IMO - it implements Autocloseabe interface and must be closed once the operation on cluster is finished (the good think that in comparison with DB connection, creating OpenShift client is a "cheap" operation) One thing I notice, if OpenShift connector is created in method and not closed there will be warning - `Resource leak 'openShiftClient' is never closed`

l0rd (approved):
I confirm that I could not get any "Error while pumping stream". 

And it has no impact on performances: I've started a workspace locally with and without this patch and it takes 1'42'' in both cases :-)

ibuziuk:
@l0rd awesome! shall we merge ?

l0rd:
Unfortunately it does not work on OSO but we figured out that the we couldn't deploy older versions neither. We can exclude that's a regression then and let's merge it!

Issue #4379, opened 2017-03-09
Cannot reduce process panel size.


JamesDrummond:
When the workspace is loaded with  process panel maximized, I cannot reduce the size. This is reproducible only when panel is not resized after initialization so I am not sure how to reproduce. I think it may have to do with workspaces being restarted after timeout.

![issue](https://cloud.githubusercontent.com/assets/20580888/23754472/b3bebaf2-04a2-11e7-9a26-c026409acc03.gif)

PR #4829, opened 2017-04-18
CHE-4379 Cannot reduce process panel size

vitaliy-guliy:
### What does this PR do?
Fix a bug with restoring dimensions when refreshing the browser.
The dimensions are set to default when the perspective was maximized before refreshing the browser.

https://github.com/eclipse/che/issues/4379

#### Changelog
Fix a bug with restoring dimensions when refreshing the browser.


ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/component/StateComponent.java
```diff
@@ -41,6 +41,6 @@
      *
      * @param state the component state object
      */
-    void loadState(@NotNull JsonObject state);
+    void setState(@NotNull JsonObject state);
```
vparfonov: What reason for this renaming?

vitaliy-guliy: Just to make names more clear for developer.
The `StateComponent` interface has
`JsonObject getState()` method and now it has `setState` instead of `loadState`

vparfonov: according to JavaDoc `Called when component should restore his state`. loadSatte more reasonable for me.
@evidolob your opinion?

evidolob: I agree with @vparfonov, in common case `set` prefix used for setters, mutator method for object but that method doesn't has side effects. In this case we have huge side effects, so `set` prefix is confusing, `load` is better.

evidolob (changes requested):
@vitaliy-guliy have you write tests to your fix?

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/component/StateComponent.java
vitaliy-guliy: Actually it changes the state, it sets a new state.
But it's not a problem to me to revert the method name.

evidolob (approved):

vparfonov (approved):

PR #4831, opened 2017-04-18
Update route naming to make it work on OSO

l0rd:
### What does this PR do?
Name routes using the default scheme: ${ROUTE_NAME}-${NAMESPACE_NAME}
That is mandatory to deploy Che on OpenShiftOnline


ibuziuk (approved):
+ 1 to apply

PR #4838, opened 2017-04-19
Openshift pvc refactor

amisevsk:
### What does this PR do?
1. Deletes Replica Sets while deleting other workspace resources, since OpenShift does not clean them automatically.
2. Deletes workspace directories on the workspace PVC when workspaces are deleted. To avoid unnecessary mounting of the PVC, waits until Che server is idled, and all directories simultaneously
    - To override `workspaceFilesCleaner`, the binding in `WsMasterModule` has been changed to suit OpenShift. This is not an ideal solution (not compatible with docker distribution).
    - Since we're not removing the directory immediately, re-creating a workspace with a previously-deleted name before the server idles will result in the previous workspace's files being already present and cause an issue if the same project is imported.
3. Creates directories for workspaces in PVC in a short-lived job pod instead of a full deployment. This makes workspace launch much faster. Additionally, some effort is made to avoid attempting to create directories when they are already there. 

- Removing workspace causes NPE
- Cleanup workspaces resources in PVC on workspace deletion
- Improve PVC set up when creating workspace
- Clean up replica sets when stopping workspace

### Testing
To test creation, creating a workspace, stopping it, and then restarting it is sufficient. The first launch should log `Executing command mkdir in PVC claim-che-workspace for 1 dirs`, whereas the second should not.

To test pvc cleanup, set `che.openshift.server.inactive.stop.timeout.ms` to a small number (e.g. 60 seconds) and delete a created workspace. On idle the workspace folder should be deleted on the PVC.

To test replica set cleanup, `oc get rs` should return no resources when no workspace is running.

### DIFF
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
@@ -128,6 +128,7 @@ import io.fabric8.kubernetes.api.model.VolumeMount;
 import io.fabric8.kubernetes.api.model.VolumeMountBuilder;
 import io.fabric8.kubernetes.api.model.extensions.Deployment;
 import io.fabric8.kubernetes.api.model.extensions.DeploymentBuilder;
+import io.fabric8.kubernetes.api.model.extensions.ReplicaSet;
 import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.Watcher;
 import io.fabric8.kubernetes.client.dsl.ExecWatch;
@@ -315,6 +316,7 @@ public class OpenShiftConnector extends DockerConnector {
                                                  .withName(imageStreamName)
                                                  .get();
         if (imageStream == null) {
+            openShiftClient.close();
             throw new OpenShiftException("ImageStream not found");
         }
         String registryAddress = imageStream.getStatus()
@@ -512,20 +514,20 @@ public class OpenShiftConnector extends DockerConnector {
     public Network inspectNetwork(InspectNetworkParams params) throws IOException {
         String netId = params.getNetworkId();
         ServiceList services;
-        
+
         try (OpenShiftClient openShiftClient = new DefaultOpenShiftClient()) {
             services = openShiftClient.services()
                                       .inNamespace(this.openShiftCheProjectName)
                                       .list();
         }
-        
+
         Map<String, ContainerInNetwork> containers = new HashMap<>();
         for (Service svc : services.getItems()) {
             String selector = svc.getSpec().getSelector().get(OPENSHIFT_DEPLOYMENT_LABEL);
             if (selector == null || !selector.startsWith(CHE_OPENSHIFT_RESOURCES_PREFIX)) {
                 continue;
             }
-            
+
             PodList pods;
             try (OpenShiftClient openShiftClient = new DefaultOpenShiftClient()) {
                 pods = openShiftClient.pods()
@@ -923,7 +925,7 @@ public class OpenShiftConnector extends DockerConnector {
                                           .inNamespace(openShiftCheProjectName)
                                           .list()
                                           .getItems();
-        } 
+        }
 
         // We only get ImageStreamTag names here, since these ImageStreamTags do not include
         // Docker metadata, for some reason.
@@ -948,7 +950,7 @@ public class OpenShiftConnector extends DockerConnector {
     }
 
     private ImageStreamTag getImageStreamTag(final String imageStreamName) {
-        try (OpenShiftClient openShiftClient = new DefaultOpenShiftClient()) { 
+        try (OpenShiftClient openShiftClient = new DefaultOpenShiftClient()) {
             return openShiftClient.imageStreamTags()
                                   .inNamespace(openShiftCheProjectName)
                                   .withName(imageStreamName)
@@ -961,11 +963,11 @@ public class OpenShiftConnector extends DockerConnector {
             ServiceList svcs = openShiftClient.services()
                                     .inNamespace(this.openShiftCheProjectName)
                                     .list();
-    
+
             Service svc = svcs.getItems().stream()
                     .filter(s->s.getSpec().getSelector().containsKey(selectorKey))
                     .filter(s->s.getSpec().getSelector().get(selectorKey).equals(selectorValue)).findAny().orElse(null);
-    
+
             if (svc == null) {
                 LOG.warn("No Service with selector {}={} could be found", selectorKey, selectorValue);
             }
@@ -1006,6 +1008,18 @@ public class OpenShiftConnector extends DockerConnector {
         }
     }
 
+    private List<ReplicaSet> getReplicaSetByLabel(String key, String value) {
+        List<ReplicaSet> replicaSets;
+        try (OpenShiftClient openShiftClient = new DefaultOpenShiftClient()) {
+            replicaSets = openShiftClient.extensions()
+                                         .replicaSets()
+                                         .inNamespace(openShiftCheProjectName)
+                                         .withLabel(key, value)
+                                         .list().getItems();
+            return replicaSets;
+        }
+    }
+
     private Pod getChePodByContainerId(String containerId) throws IOException {
         try (OpenShiftClient openShiftClient = new DefaultOpenShiftClient()) {
             PodList pods = openShiftClient.pods()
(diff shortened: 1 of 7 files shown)


plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -315,6 +316,7 @@ public ContainerCreated createContainer(CreateContainerParams createContainerPar
                                                  .withName(imageStreamName)
                                                  .get();
         if (imageStream == null) {
+            openShiftClient.close();
```
ibuziuk: No need to close client here - it will be closed in finally block

ibuziuk (changes requested):

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
ibuziuk: @amisevsk oh, sorry overlook try block. Still I think that client should be opened only once and closed only once (in finally) in each method. So, maybe it makes sense to wrap the whole method body in try ?

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftWorkspaceFilesCleaner.java
```diff
@@ -0,0 +1,70 @@
+package org.eclipse.che.plugin.openshift.client;
```
ibuziuk, line 12: No license header

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftPvcHelper.java
```diff
@@ -0,0 +1,175 @@
+package org.eclipse.che.plugin.openshift.client;
```
ibuziuk, line 12: No license header

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -1431,57 +1362,49 @@ private void cleanUpWorkspaceResources(String deploymentName) throws IOException
...
+                                                                               workspacesPersistentVolumeClaim);
+                throw new OpenShiftException("Failed to create workspace directory in PVC");
+            }
+        }
+    }
+
+    private String getWorkspaceSubpath(String[] volumes) {
+        String workspaceSubpath = null;
+        for (String volume : volumes) {
+            String mountPath = volume.split(":", 3)[1];
```
ibuziuk: 3 is a magic number for me

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -1006,6 +1008,18 @@ private Deployment getDeploymentByName(String deploymentName) throws IOException
         }
     }
 
+    private List<ReplicaSet> getReplicaSetByLabel(String key, String value) {
+        List<ReplicaSet> replicaSets;
```
ibuziuk: this could be a part of try with resource block I belive

ibuziuk:
@amisevsk I have tested PR on minishift and functionality seems to work just fine - great job!

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftWorkspaceFilesCleaner.java
amisevsk: Fixed.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftPvcHelper.java
amisevsk: Fixed

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
amisevsk: I agree that it's not a great solution. Wrapping the whole method body in a try causes some variables used in the catch to go out of scope. Instead, I've separated out the two distinct uses of OpenShiftClient into two separate try blocks. WDYT?

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
amisevsk: It's a bit of a magic number for me too --  this method is mostly renamed and rearranged from the previous subpaths implementation. I've tried adding some documentation to explain this method but it still could probably be clearer.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
amisevsk: Fixed.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
ibuziuk: sounds good

ibuziuk (approved):

amisevsk:
I've squashed and rebased.

l0rd:
Testing on OSO free-int creation of a workspace failed with the following message:
> Failure executing: POST at: https://kubernetes.default.svc/api/v1/namespaces/mario-loriedo-che/pods. Message: Forbidden!Configured service account doesn't have access. Service account may have been revoked..

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftPvcHelper.java
```diff
@@ -0,0 +1,219 @@
...
+ *
+ * @author amisevsk
+ */
+final class OpenShiftPvcHelper {
+
+    private static final Logger LOG = LoggerFactory.getLogger(OpenShiftPvcHelper.class);
+
+    private static final String POD_PHASE_SUCCEEDED        = "Succeeded";
+    private static final String POD_PHASE_FAILED           = "Failed";
+    private static final String JOB_IMAGE                  = "centos:centos7";
```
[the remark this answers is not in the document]
amisevsk: Done. The image used is now specified by the property `che.openshift.jobs.image`, which is injected into the `OpenShiftWorkspaceFilesCleaner` and `OpenShiftConnector`. Using the workspace's image won't necessarily work, since delete jobs are executed when the server is idled, so there's no workspace image string readily available.

amisevsk:
@l0rd I've got the same issue, I'm looking into it.

amisevsk:
@l0rd The `Failure executing: POST` issue should be fixed now, can you confirm?

I've added another property: `che.openshift.jobs.memorylimit`, and made `OpenShiftPvcHelper` non static so that the properties can be injected there instead of all over the place.d

The issue was the the Pod being created had a memory limit that was less than the limitrange on the cluster. Kubernetes-client sure could've given a more descriptive error message...

l0rd:
@amisevsk I confirm that `Failure executing: POST`! Everything works fine for me. Great work!

l0rd:
I can still see a few `Failed to attach volume` on OpenShift Online. But at third tentative (that's after 10s) the workspace is created successfully.

l0rd (approved):

Issue #4865, opened 2017-04-20
Workspace does not start with Java 1.7 installed


ghost:
Steps to reproduce:

1. Create a workspace from the following recipe:

```
FROM codenvy/jdk7
CMD tail -f /dev/null

```

Workspace agent does not start since it uses JDK 1.7 that is pre installed in the image. However, this piуce of script https://github.com/eclipse/che/blob/master/wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh#L52-L57 should have detected that Java version installed in a workspace is not 1.8 and install openjdk package from a debian repo.

bmicklea:
@tolusha and @vkuznyetsov - this is preventing TIAA from continuing their rollout.

ghost:
Looks like the way script extracts and compares Java version is wrong. Should be an easy fix.

PR #4870, opened 2017-04-21
CHE-4865: Correctly detect installed java version

tolusha:
### What does this PR do?
Correctly detect installed java version

https://github.com/eclipse/che/issues/4865

#### Changelog
Correctly detect installed java version

#### Release Notes
not required

#### Docs PR
not required

### DIFF
--- a/wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
+++ b/wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
@@ -52,8 +52,8 @@ INSTALL_JDK=false
 command -v ${JAVA_HOME}/bin/java >/dev/null 2>&1 || {
     INSTALL_JDK=true;
 } && {
-    java_version=$(${JAVA_HOME}/bin/java -version 2>&1 | sed 's/.* version "\\(.*\\)\\.\\(.*\\)\\..*"/\\1\\2/; 1q')
-    if [ ! -z "${java_version##*[!0-9]*}" ] && [ "${java_version}" -lt "18" ]; then
+    java_version=$(${JAVA_HOME}/bin/java -version 2>&1 | grep version  | awk '{print $NF}' | sed 's/"//g' | cut -d '.' -f2)
+    if [ "${java_version}" -eq "8" ]; then
         INSTALL_JDK=true;
     fi
 }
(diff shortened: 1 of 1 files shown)


ghost (approved):

ghost:
Tested with `codenvy/jdk7`, `codenvy/codenvy` and `eclipse/ubuntu_jdk8` images

wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
```diff
@@ -52,8 +52,8 @@ INSTALL_JDK=false
 command -v ${JAVA_HOME}/bin/java >/dev/null 2>&1 || {
     INSTALL_JDK=true;
 } && {
-    java_version=$(${JAVA_HOME}/bin/java -version 2>&1 | sed 's/.* version "\\(.*\\)\\.\\(.*\\)\\..*"/\\1\\2/; 1q')
-    if [ ! -z "${java_version##*[!0-9]*}" ] && [ "${java_version}" -lt "18" ]; then
+    java_version=$(${JAVA_HOME}/bin/java -version 2>&1 | grep version  | awk '{print $NF}' | sed 's/"//g' | cut -d '.' -f2)
```
garagatyi, line 55: There is no checks that awk is present in container. awk is used in centos check, so I supose in works on centos but how to be sure it works in other images?.

ghost: Tried ubuntu, debian and alpine.

garagatyi: What about user's custom images? Have we declared that awk is mandatory in dev machine?

ghost: Library images for alpine, busybox, ubuntu, debian, centos, fedora and rhel have awk.

Also, in all other agents we use awk - so far so good. So, if we are to check if awk is available, then we should do it across all agent scripts. However, having a custom image without awk is quite uncommon I think (maybe some super stripped images)

garagatyi: ok then

garagatyi (approved):

benoitf (approved):

Issue #4591, opened 2017-03-28
Switch project importing output to the websocket JSONRPC API endpoint


vparfonov:
Need to rewrite client side of IDE to the new implementation of WebSocket communication.
Add new message  transceiver on server side  for IDE, keep current implementation based on channel subscription for Dashboard

PR #4888, opened 2017-04-24
Broadcast project import output through json rpc protocol

vzhukovs:
### What does this PR do?
This changes proposal adds ability to broadcast project importing output through JSON RPC protocol. Although there is a back compatibility for broadcasting output events through Web Socket protocol.

Signed-off-by: Vladyslav Zhukovskii <vzhukovskii@codenvy.com>

#4591 

#### Changelog
Broadcast project import output through json rpc protocol

#### Release Notes
N/A

#### Docs PR
N/A


vzhukovs:
ci-build

gazarenkov:
@vzhukovskii what's the status of this?

vzhukovs:
Ready to merge, just need approve from reviewers

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/projectimport/wizard/ProjectImportOutputJsonRpcSubscriber.java
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @author Vlad Zhukovskyi
+ * @since 5.9.0
+ */
+@Singleton
+public class ProjectImportOutputJsonRpcSubscriber {
+
+    public static final String WS_AGENT_ENDPOINT = "ws-agent";
+
+    private static final String EVENT_IMPORT_OUTPUT_SUBSCRIBE    = "event:import-project:subscribe";
```
dkulieshov: This probably should be changed to something like `importProject/subscribe` according to our json rpc naming conventions

vzhukovs: Fixed

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/jsonrpc/commons/RequestHandlerManager.java
```diff
@@ -134,6 +134,21 @@ public boolean isRegistered(String method) {
         return methodToCategory.containsKey(method);
     }
 
+    public synchronized boolean deregister(String method) {
```
dkulieshov, line 137: As I see we can handle this more accurate here: we can first get category by method's name from `methodToCategory` and than knowing category we can remove method handler from exactly one map of handlers, what do you think?

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
```diff
@@ -351,8 +361,16 @@ public void importProject(@ApiParam(value = "Path in the project", required = tr
                                                                      ServerException,
                                                                      NotFoundException,
                                                                      BadRequestException {
-        projectManager.importProject(path, sourceStorage, force,
-                                     () -> new ProjectImportOutputWSLineConsumer(path, workspace, 300));
+
```
dkulieshov, line 364: I believe it is not enough to rewrite the method here. The main idea that stands by this task is to replace all REST and old websocket calls by new JSON-RPC implementation. I mean we also need to replace the REST call itself (`@Path("/import/{path:.*}")`) by JSON-RPC call (`project/import` or something like that). So eventually we should not have any REST calls in this service, perhaps meanwhile we can have a separate JsonRpcProjectService to decouple REST and JSON-RPC logics. Do you see what I mean?

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
```diff
@@ -351,8 +361,16 @@ public void importProject(@ApiParam(value = "Path in the project", required = tr
...
+        final int delayBetweenMessages = 300;
+
+        final ProjectImportOutputWSLineConsumer wsLineConsumer =
```
dkulieshov, line 367: Do we still need this? Is it used somewhere?

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/importer/BaseProjectImportOutputLineConsumer.java
```diff
@@ -0,0 +1,83 @@
...
+    public void close() throws IOException {
+        executor.shutdown();
+    }
+
+    @Override
+    public void writeLine(String line) throws IOException {
+        try {
+            lineToSendQueue.put(line);
+        } catch (InterruptedException ignored) {
+            // ignore if interrupted
```
dkulieshov: I would add at least a log message here, that should slightly improve the logging :-)

dkulieshov:
We should think on all JSON-RPC based project operations method naming (not only import). In my opinion it should be named something like `project/{operation}/{additional_context}` or something similar, for example for import process output subscription we can call it `project/import/subscribe` while for project import we can call `project/import`. What is your thinking on that?

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/importer/ProjectImportOutputJsonRpcRegistrar.java
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Endpoint registry for broadcasting project import events. Holds registered client's endpoint ids.
+ *
+ * @author Vlad Zhukovskyi
+ * @since 5.9.0
+ */
+@Singleton
+public class ProjectImportOutputJsonRpcRegistrar {
+
+    private static final String EVENT_IMPORT_OUTPUT_SUBSCRIBE    = "importProject/subscribe";
```
azatsarynnyy: Maybe it would be better to place these constants in the shared [Constants.java](https://github.com/eclipse/che/blob/master/wsagent/che-core-api-project-shared/src/main/java/org/eclipse/che/api/project/shared/Constants.java#L16)?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/jsonrpc/commons/RequestHandlerManager.java
vzhukovs: Refactored the code and add switch constrution

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
vzhukovs: Yes, websocket line consumer still needs for dashboard uses. Left for backward compatibility.

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/importer/BaseProjectImportOutputLineConsumer.java
vzhukovs: Added log call

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/importer/ProjectImportOutputJsonRpcRegistrar.java
vzhukovs: Moved to the constants

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
dkulieshov: I see that [issue](https://github.com/eclipse/che/issues/4591) description does not state this and that's a bit odd as that's the main idea here. We can create a separate issue for this if you think that this would require additional work or we can apply changed directly to this PR if that would take not much time, it is up to you.

vzhukovs: The main task was just streaming importer logs over JSON-RPC protocol. @vparfonov and I had discussion about this moment before and we decided to make only logs streaming at first step. Running import operation over JSON-RPC will be next step of this global task.

dkulieshov (approved):

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
vzhukovs: Created separated issue for switching begin project import via JSON-RPC https://github.com/eclipse/che/issues/5410

azatsarynnyy (approved):

Issue #4679, opened 2017-04-03
Test runner launch tests on the same JVM with dev-machine


musienko-maxim:
**Reproduction Steps:**
1) Create java console application from wizard
2) Open  the AppTest class and add   `System.exit(1);`  invocation in the test.
**Expected result:** the test should pass
**Actual result:** the test has invoked System.exit. The workspace 
has crashed. The test can invoke any java code on the JVM of dev-machine! It is very critical for security.
![anim](https://cloud.githubusercontent.com/assets/1640058/24615491/13923ffa-1897-11e7-80a5-dbf98548a364.gif)
**OS and version:** Ubuntu 14.0 TLS, chrome browser

vparfonov:
Done for TestNG test framework

PR #4922, opened 2017-04-26
Temporary fix for https://issues.jboss.org/browse/CHE-214

davidfestal:
Wildfly Swarm / Arquillian tests cannot be run successfully.

Signed-off-by: David Festal <dfestal@redhat.com>

### What does this PR do?

It correctly changes the OS current working directory to the tested project root directory during the execution of the tests. This is necessary for various test framework that assume the tests to be started in the project root directory.

However the true long-term solution for these problems is to refactor the Testing feature to run tests in their own process (see https://issues.jboss.org/browse/CHE-192 and https://github.com/eclipse/che/issues/4679)

### DIFF
--- a/plugins/plugin-testing-java/plugin-testing-junit/che-plugin-testing-junit-server/pom.xml
+++ b/plugins/plugin-testing-java/plugin-testing-junit/che-plugin-testing-junit-server/pom.xml
@@ -22,6 +22,10 @@
     <name>Che Plugin :: Java Testing :: JUnit Server</name>
     <dependencies>
         <dependency>
+            <groupId>com.github.jnr</groupId>
+            <artifactId>jnr-posix</artifactId>
+        </dependency>
+        <dependency>
             <groupId>com.google.inject</groupId>
             <artifactId>guice</artifactId>
         </dependency>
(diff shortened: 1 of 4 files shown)


plugins/plugin-testing-java/pom.xml
```diff
@@ -26,6 +26,15 @@
         <module>plugin-testing-junit</module>
         <module>plugin-testing-testng</module>
     </modules>
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>com.github.jnr</groupId>
+                <artifactId>jnr-posix</artifactId>
+                <version>3.0.38</version>
```
benoitf: introduction of specific version should be done at https://github.com/eclipse/che-dependencies/blob/master/pom.xml  (there is also an openshift branch)

benoitf: also the IP should be checked on https://dev.eclipse.org/ipzilla/query.cgi

davidfestal:
@l0rd could you review this PR please ?

benoitf:
note: it requires https://github.com/eclipse/che-dependencies/pull/42 to be merged first

l0rd (approved):
LGTM

Issue #4670, opened 2017-04-03
The error notification is not informative if copy a file in a folder where a file with this name exists


SkorikSergey:
**Reproduction Steps** for a file:

- Create any project.
- In the Project create a file.
- Copy this file and try to paste it there

**Expected behavior:**
The "Error" notification appears with the message "Resource already exists"

**Observed behavior:**
The "Error" notification appears without the message about the file exists

**Reproduction Steps** for a folder:

- Create any project.
- In the Project create a folder.
- Copy this folder and try to paste it there

**Expected behavior:**
The "Error" notification appears with the message "Resource already exists"

**Observed behavior:**
The "Error" notification appears without the message about the folder exists

**Che version:** 5.6.0-SNAPSHOT
**OS and version:** Ubuntu 16.04 LTS
**Docker version:** 1.12.3
**Che install:** Docker container

**Additional information:**
Problem can be reliably reproduced, doesn't happen randomly: [Yes]
![anim](https://cloud.githubusercontent.com/assets/7760565/24600696/707f7810-185e-11e7-9064-b477e7289635.gif)

PR #4926, opened 2017-04-26
CHE-4670: make copy file operation more informative

svor:
### What does this PR do?
Makes copy file operation more informative.
If a file with its name already exists the window will be shown:
selection_003

https://github.com/eclipse/che/issues/4670

#### Changelog

Make copy file operation more informative

### DIFF
--- a/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/resources/impl/ResourceManager.java
+++ b/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/resources/impl/ResourceManager.java
@@ -544,7 +544,9 @@ public final class ResourceManager {
         checkArgument(!source.getLocation().isRoot(), "Workspace root is not allowed to be copied");
 
         return findResource(destination, true).thenPromise(resource -> {
-            checkState(!resource.isPresent() || force, "Cannot create '" + destination.toString() + "'. Resource already exists.");
+            if (resource.isPresent() && !force){
+                return promises.reject(new IllegalStateException("Cannot create '" + destination.toString() + "'. Resource already exists."));
+            }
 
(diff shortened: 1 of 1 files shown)


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/resources/impl/ResourceManager.java
```diff
@@ -544,7 +544,9 @@ private NewProjectConfigDto asDto(MutableProjectConfig config) {
         checkArgument(!source.getLocation().isRoot(), "Workspace root is not allowed to be copied");
```
vzhukovs, line 544: translate this statement with using `return promises.reject(...)`

vzhukovs (approved):

vparfonov (approved):

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/resources/impl/ResourceManager.java
svor: In this class we have many validations like this:
`checkArgument(expression, errorMessage)`
So, I think we should change many of them but not in this issue.

vzhukovs: then ok

slemeur:
@vparfonov For this kind of PR, you need to notify PM. and request approval
The title of the popup is not appropriate: it's not an error. 
It's not "cannot", as we can overwrite. 

Could we change the messages to:
- Title: "File already exists. Do you want to replace it?"
- Message: "A file with the same name already exists in the destination folder. Do you want to replace it?"
- Actions: "Rename" "Cancel" "Replace"

PR #4930, opened 2017-04-26
Add secret to workspaces for bayesian plugin

amisevsk:
### What does this PR do?
Adds env var to workspace containers to allow the bayesian server to communicate with backend.

This is a bit of a workaround, as currently there doesn't seem to be a good way of figuring out which agents are going to be injected at workspace creation.


l0rd:
If we have time it would be better to move this code in rh-che repository using a patch

Issue #3324, opened 2016-12-08
Che container creates folders with projects in /data on host machine


garagatyi:
Che CLI in docker container creates /data folder on host and writes projects there.

**Che version:**    5.0.0-M9-SNAPSHOT
**OS and version:** Ubuntu 16.04
**Docker version:** 1.12
**Che install:**    Docker container

benoitf:
@garagatyi with image from master branch ?

garagatyi:
Yes

TylerJewell:
Post command line.  We made changes in the cli-improvements branch to change :/che volume mount to now be :/data, but it's only in the branch.

garagatyi:
I tried commands I listed below and I can reproduce it:
```
✔  15:52 ~/workspace/.che $ docker pull eclipse/che-cli:nightly 
nightly: Pulling from eclipse/che-cli
Digest: sha256:d674cebdca3095240f74606d89f75b7e254956a74dc41edf3a08042076891a8e
Status: Image is up to date for eclipse/che-cli:nightly
✔  15:52 ~/workspace/.che $ docker pull eclipse/che-cli:nightly ^C
✔  15:52 ~/workspace/.che $ ./che.sh --nd start --pull
INFO: (che cli): Loading cli...
INFO: (che cli): Checking registry for version 'nightly' images
INFO: (che download): Pulling image puppet/puppet-agent-alpine:4.6.2

4.6.2: Pulling from puppet/puppet-agent-alpine
Digest: sha256:bf93e9f1a92b692efeb3138d25444c478093b6e18818f4242ffc723b04fc6a57
Status: Image is up to date for puppet/puppet-agent-alpine:4.6.2

INFO: (che download): Pulling image registry:2.5.0

2.5.0: Pulling from library/registry
Digest: sha256:1b68f0d54837c356e353efb04472bc0c9a60ae1c8178c9ce076b01d2930bcc5d
Status: Image is up to date for registry:2.5.0

INFO: (che download): Pulling image eclipse/che-init:nightly

nightly: Pulling from eclipse/che-init
Digest: sha256:4f1a8696a0ca8984663d65c13d5bcfbf8b4d47a2fcbfd2d8ee5b09fcc0c8895b
Status: Image is up to date for eclipse/che-init:nightly

INFO: (che download): Pulling image eclipse/che-server:nightly

nightly: Pulling from eclipse/che-server
[...]

PR #4931, opened 2017-04-27
Fix workspace storage path on Docker Cloud

davidwindell:
### What does this PR do?
This PR fixes the problem with workspace data directories already fixed for *some* in https://github.com/eclipse/che/issues/3324 but identified as not complete for all users in https://github.com/eclipse/che/blob/master/dockerfiles/che/entrypoint.sh#L310.

By using the cgroup file we can find the container's real ID without relying on the hostname (which can be changed with `--hostname` or by many clustering tools like Docker Cloud). 

#### Changelog
Fix workspace storage path where hostname does not match container name

### DIFF
--- a/dockerfiles/che/entrypoint.sh
+++ b/dockerfiles/che/entrypoint.sh
@@ -307,10 +307,10 @@ get_che_data_from_host() {
 }
 
 get_che_server_container_id() {
-  # Returning `hostname` doesn't work when running Che on OpenShift/Kubernetes.
+  # Returning `hostname` doesn't work when running Che on OpenShift/Kubernetes/Docker Cloud.
   # In these cases `hostname` correspond to the pod ID that is different from
   # the container ID
-  hostname
+  echo $(basename "$(head /proc/1/cgroup || hostname)");
 }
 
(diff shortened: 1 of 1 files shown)


davidwindell:
Please can this be in 5.10 O:-)

TylerJewell:
Just to set expectations - this is probably something that requires deep investigation.  It would be something probably tied to a 6.0 major release train.

We are analyzing a number of internal structural improvements for six to handle some really interesting scenarios like this.

Tyler

davidwindell:
Ok (although this is a bug fix for 5.x and doesn't change existing scenarios).

For anyone who comes across this, you can use sed within the running che-server and then restart it to get workspace host paths working properly.

`sed -i '313s/hostname/echo \$(basename "\$(head \/proc\/1\/cgroup || hostname)");/' /entrypoint.sh`

TylerJewell:
Sorry - I was leaving a comment for a different issue.  This is a PR - up to Florent and other maintainers on when to merge it. This is super simple, so don't know why this wouldn't be included right away.

benoitf (approved):

PR #4947, opened 2017-04-27
Add Alpine support to ls-json

davidwindell:
### What does this PR do?
Add's Alpine support to ls-json (the same as other agents)

#### Changelog

Fixed Alpine support for ls-json (the same as other agents)

#### Release Notes

Fixed Alpine support for ls-json (the same as other agents)

### DIFF
--- a/agents/ls-json/src/main/resources/org.eclipse.che.ls.json.script.sh
+++ b/agents/ls-json/src/main/resources/org.eclipse.che.ls.json.script.sh
@@ -139,6 +139,14 @@ elif echo ${LINUX_TYPE} | grep -qi "opensuse"; then
         ${SUDO} zypper in nodejs
     }
 
+# Alpine 3.3
+############
+elif echo ${LINUX_TYPE} | grep -qi "alpine"; then
+    test "${PACKAGES}" = "" || {
+        ${SUDO} apk update
+        ${SUDO} apk add ${PACKAGES};
+    }
+
 else
     >&2 echo "Unrecognized Linux Type"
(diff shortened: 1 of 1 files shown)


benoitf (approved):

ghost (approved):

agents/ls-json/src/main/resources/org.eclipse.che.ls.json.script.sh
```diff
@@ -139,6 +139,14 @@ elif echo ${LINUX_TYPE} | grep -qi "opensuse"; then
         ${SUDO} zypper in nodejs
     }
 
+# Alpine 3.3
+############
+elif echo ${LINUX_TYPE} | grep -qi "alpine"; then
```
tolusha, line 144: What about installing nodejs dependencies?
Are they required for alpine?

tolusha (changes requested):

benoitf (changes requested):
CLA is required or Signed-Off header needs to be added https://dev.eclipse.org/eclipse-webhook/services/status_details.php?id=5902f17391d65 for this PR

agents/ls-json/src/main/resources/org.eclipse.che.ls.json.script.sh
davidwindell: :+1: done

benoitf (approved):

tolusha (approved):

slemeur (approved):

PR #4980, opened 2017-05-03
Rework top test menu items

davidfestal:
### What does this PR do?

This PR reworks the behaviour of the top menu `Run -> Test` menu items.
Until now, these menus items were based on the Java Editor only. So when no Java Editor had the focus, the menu items were disabled, even with a Java source file selected in File Explorer. This was leading to an inconsistent behaviour between the Project Explorer `Test` contextual menu and the `Run -> Test` top menu item.

With this PR, the proposed test menu items are based on the current resource, either from the project explorer or the currently opened editor, according to which one has the focus. This brings back consistency between both menus.

https://issues.jboss.org/browse/CHE-227

#### Changelog

Provide a consistent behaviour between the `Run -> Test` top menu item and the `Run Test` project explorer contextual menu.

#### Release Notes

Provide a consistent behaviour between the `Run -> Test` top menu item and the `Run Test` project explorer contextual menu.


davidfestal:
@l0rd Could you review this one ?

benoitf:
ci-build

davidfestal:
@l0rd @benoitf do you think this is ready for merge ?

benoitf:
@vparfonov ?

benoitf (approved):

l0rd (approved):

davidfestal:
@vparfonov @slemeur is there anything more I should do in order to have this PR merged ?

slemeur (approved):

vparfonov:
Need QA confirmation

plugins/plugin-testing-java/plugin-testing-junit/che-plugin-testing-junit-ide/src/main/java/org/eclipse/che/plugin/testing/junit/ide/action/RunAllTestAction.java
```diff
@@ -31,35 +44,63 @@
...
+        
+        Project project = resource.getProject();
+        if (project == null) {
+            e.getPresentation().setEnabledAndVisible(false);
+        }
+        
+        e.getPresentation().setVisible(true);
+
+        String projectType = project.getType();
+        boolean enable = "maven".equals(projectType);
```
vparfonov, line 102: Why we need here checking for "maven" what about "javac" project?

vparfonov:
Basically ok for me except adding checking on "maven" project type. Why we need it?

plugins/plugin-testing-java/plugin-testing-junit/che-plugin-testing-junit-ide/src/main/java/org/eclipse/che/plugin/testing/junit/ide/action/RunAllTestAction.java
davidfestal: @vparfonov : because for the moment the Java (JUnit / TestNG) testing feature is only supported on maven Java projects. that was from the beginning of this feature afaik. But sure as soon as this limitation is dropped, we'll be able to remove this test.

musienko-maxim:
@davidfestal 
So, after selenium cycle tests we have not found regression. But we have 1 question.
We have a specific test with TestNg xml suite. In previous version we could launch of the 'TestNG XML suite' menu from any selected item of the project. But on this moment we cannot launch from this menu (becouse this one is disabled) if we select a class from  /src/main/java or /src/test/javaaa folder. Is it bug or feature? More details in the screencast:
anim

davidfestal:
> But on this moment we cannot launch from this menu (becouse this one is disabled) if we select a class from /src/main/java or /src/test/javaaa folder. Is it bug or feature

Well, when reworking the testing menu, I tried to fix inconsistencies in the overall behavior, and thus had to understand / rethink the menu activation rules. And it seemed to me that, when you're explicitly requesting tests on a Java file, the aim is not to start a test whose input is a specific unrelated XML file. So when I disabled the `TestNG XML Suite` option in this case I just saw this as the expected behavior of the Testing feature. But I may have been wrong.

davidfestal:
@musienko-maxim are you OK with my comment and is it possible to merge now ?

musienko-maxim:
For me ok.

musienko-maxim (approved):

benoitf:
@vparfonov ? can we merge

vparfonov (approved):

benoitf:
ci-build

PR #5008, opened 2017-05-04
Add property to control manual workspace dir creation in OpenShift

amisevsk:
### What does this PR do?
Adds property `che.openshift.precreate.workspace.dirs`. If true, Che will run a separate pod to create a subdirectory in the workspaces PV before launching the pod. If false, this step is skipped.

Creating the directory is necessary in some versions as Kubernetes creates the volumemount subpath with root permissions, preventing modification by workspaces. This has been fixed fairly [recently](https://github.com/kubernetes/kubernetes/pull/43775) and should be disabled if OpenShift supports it, as mounting/unmounting PVs can cause issues. 

I've left the property as true by default, as this is the safer option, and minishift does not seem to support a version of OpenShift with this fix at the moment. Once the fix is standard, this property could probably be removed.

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
@@ -306,6 +306,14 @@ che.openshift.secure.routes=false
 che.openshift.jobs.image=centos:centos7
 che.openshift.jobs.memorylimit=250Mi
 
+# Run job to create workspace subpath directories in persistent volume before launching workspace.
+# Necessary in some versions of OpenShift/Kubernetes as workspace subpath volumemounts are created
+# with root permissions, and thus cannot be modified by workspaces running as user (presents as error
+# importing projects into workspace in Che). Default is "true", but should be set to false if version
+# of Openshift/Kubernetes creates subdirectories with user permissions.
+# Relevant issue: https://github.com/kubernetes/kubernetes/issues/41638
+che.openshift.precreate.workspace.dirs=true
+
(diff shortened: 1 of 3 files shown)


assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -306,6 +306,13 @@ che.openshift.secure.routes=false
 che.openshift.jobs.image=centos:centos7
 che.openshift.jobs.memorylimit=250Mi
 
+# Run job to create workspace subpath directories in persistent volume before launching workspace.
+# Necessary in some versions of OpenShift/Kubernetes as workspace subpath volumemounts are created
+# with root permissions, and thus cannot be modified by workspaces running as user (presents as error
```
l0rd, line 311: You may want add a link to the issue https://github.com/kubernetes/kubernetes/issues/41638

l0rd (approved):
I've added a comment if you have time to address that but otherwise LGTM. 

You should also open a couple of PRs to add a configmap in https://github.com/redhat-developer/rh-che `create_all.sh` and in https://github.com/fabric8io/fabric8-online `cm.yml`

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
amisevsk: Done. Thanks.

amisevsk:
Opened PRs https://github.com/redhat-developer/rh-che/pull/62 and https://github.com/fabric8io/fabric8-online/pull/219 to include this parameter in templates.

PR #5052, opened 2017-05-11
Openshift connector improvements

sunix:
### What does this PR do?
Contains improvements of the openshift-connector that were in the openshift-connector branch and that we concider as "mergeable" 

- [x] https://github.com/eclipse/che-dependencies/pull/43 is required 

#### Changelog

Openshift connector improvements

#### Release Notes

Openshift connector improvements


pom.xml
```diff
@@ -32,7 +32,7 @@
         <module>plugins</module>
         <module>ide/che-ide-core</module>
         <module>samples</module>
-        <module>dashboard</module>
+        <!--<module>dashboard</module>-->
```
benoitf: error ?

pom.xml
```diff
@@ -520,7 +520,7 @@
             <dependency>
                 <groupId>org.eclipse.che.dashboard</groupId>
                 <artifactId>che-dashboard-war</artifactId>
-                <version>${che.version}</version>
+                <version>5.0.0</version>
```
benoitf: error ?

pom.xml
sunix: yes error :) thanks

pom.xml
sunix: yes thanks

agents/ls-csharp/src/main/resources/org.eclipse.che.ls.csharp.script.sh
```diff
@@ -13,7 +13,7 @@ unset PACKAGES
 unset SUDO
 command -v tar >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" tar"; }
 command -v curl >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" curl"; }
-test "$(id -u)" = 0 || SUDO="sudo -E"
+test "$(id -u)" = 0 || test -f ${HOME}/is_arbitrary_user || SUDO="sudo -E"
```
benoitf: can we have more details about this test ?
commit message being : "Modifications to the shell scripts that bootstrap the servers"

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -212,9 +212,14 @@ che.docker.ip.external=NULL
 #   - 'docker-local': internal address is address of container within docker network, and exposed ports
 #                     are used.
 #   - 'custom': The evaluation strategy may be customized through a template property.
+#   - 'single-port':  internal address is set as in docker-local strategy, external address is composed
```
benoitf: I think the single port could reuse the custom strategy as the expression "serverName-workspaceID-cheExternalAddress" can be addressed by using the  template `<serverName>-<workspaceId>-< externalAddresss>`

benoitf: so, AFAIK there is no need to introduce a new "hardcoded" strategy, the strategy based upon templating should provide the expected result

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/idle/ServerIdleDetector.java
```diff
@@ -0,0 +1,109 @@
...
+import org.eclipse.che.api.core.notification.EventSubscriber;
+import org.eclipse.che.api.workspace.server.WorkspaceManager;
+import org.eclipse.che.api.workspace.shared.dto.event.WorkspaceStatusEvent;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.inject.Inject;
+
+@Singleton
+public class ServerIdleDetector implements EventSubscriber<WorkspaceStatusEvent> {
```
benoitf: there is missing a lot of javadoc/comments

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/idle/ServerIdleDetector.java
```diff
@@ -0,0 +1,109 @@
...
+    public ServerIdleDetector(WorkspaceManager workspaceManager,
+                                 EventService eventService,
+                                 @Named("che.openshift.server.inactive.stop.timeout.ms") long timeout) {
+        this.timeout = timeout;
+        this.eventService = eventService;
+        this.workspaceManager = workspaceManager;
+        if (timeout > 0) {
+            this.executor = Executors.newSingleThreadScheduledExecutor();
+            this.future = executor.schedule(this::run, timeout, TimeUnit.MILLISECONDS);
+            LOG.info("Idling che server scheduled [timeout=" + timeout/1000 + " seconds]");
```
benoitf: I'm not sure it requires INFO level and we should probably avoid string concatenation there

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalDockerModule.java
```diff
@@ -56,6 +57,8 @@ protected void configure() {
                   .to(org.eclipse.che.plugin.docker.machine.DefaultServerEvaluationStrategy.class);
         strategies.addBinding("docker-local")
                   .to(org.eclipse.che.plugin.docker.machine.LocalDockerServerEvaluationStrategy.class);
+        strategies.addBinding("single-port")
+                  .to(LocalDockerSinglePortServerEvaluationStrategy.class);
```
skabashnyuk: This looks really similar to @benoitf 's CustomServerEvaluationStrategy what was the issue what it can't handle so you've created LocalDockerSinglePortServerEvaluationStrategy?

wsagent/wsagent-local/src/main/java/org/eclipse/che/idle/WorkspaceIdleFilter.java
```diff
@@ -0,0 +1,100 @@
...
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+
+import org.eclipse.che.WorkspaceIdProvider;
+import org.eclipse.che.api.core.rest.HttpJsonRequestFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+@Singleton
+public class WorkspaceIdleFilter implements Filter {
```
benoitf: I don't see unit tests for this class.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftWorkspaceFilesCleaner.java
```diff
@@ -0,0 +1,97 @@
...
+ * <p>
+ * Since deleting a workspace does not immediately remove its files, re-creating a workspace
+ * with a previously used name can result in files from the previous workspace still being
+ * present.
+ *
+ * @see WorkspaceFilesCleaner
+ * @author amisevsk
+ */
+@Singleton
+public class OpenShiftWorkspaceFilesCleaner implements WorkspaceFilesCleaner {
```
benoitf, line 51: Unit tests are missing

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftPvcHelper.java
```diff
@@ -0,0 +1,232 @@
...
+
+        Pod podSpec = new PodBuilder().withNewMetadata()
+                                       .withName(podName)
+                                   .endMetadata()
+                                   .withNewSpec()
+                                       .withContainers(container)
+                                       .withVolumes(volume)
+                                       .withRestartPolicy("Never")
+                                   .endSpec()
+                                   .build();
```
benoitf, line 159: the indent is looking strange

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/kubernetes/KubernetesContainer.java
```diff
@@ -47,7 +47,7 @@ private KubernetesContainer() {
 
             int portNumber = Integer.parseInt(port);
             String portName = CheServicePorts.get().get(portNumber);
-            portName = isNullOrEmpty(portName) ? exposedPort.replace("/", "-") : portName;
+            portName = isNullOrEmpty(portName) ? "server-" + exposedPort.replace("/", "-") : portName;
```
benoitf, line 50: "server-" should be a constant

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/kubernetes/KubernetesOutputAdapter.java
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Adapter class for passing data from a {@code kubernetes-client} output stream (e.g.
+ * for an exec call) to {@link MessageProcessor}. This class should be passed to a
+ * {@link InputStreamPumper} along with the output of the exec call.
+ *
+ * <p> Output passed in via the {@link #call(byte[])} method is parsed into lines,
+ * (respecting {@code '\n'} and {@code CRLF} as line separators), and
+ * passed to the {@link MessageProcessor} as {@link LogMessage}s.
+ */
+public class KubernetesOutputAdapter implements Callback<byte[]> {
```
benoitf, line 30: I don't see unit tests for this class

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/idle/ServerIdleDetector.java
```diff
@@ -0,0 +1,109 @@
...
+import org.eclipse.che.api.core.event.ServerIdleEvent;
+import org.eclipse.che.api.core.notification.EventService;
+import org.eclipse.che.api.core.notification.EventSubscriber;
+import org.eclipse.che.api.workspace.server.WorkspaceManager;
+import org.eclipse.che.api.workspace.shared.dto.event.WorkspaceStatusEvent;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.inject.Inject;
+
```
skabashnyuk: The similar functionality we had in Codenvy/partial Che. And I think we can potentially port/enable in Eclipse Che. See more 
https://github.com/codenvy/codenvy/tree/master/wsagent/codenvy-hosted-workspace-activity-agent
https://github.com/codenvy/codenvy/tree/master/wsmaster/codenvy-hosted-workspace-activity
https://github.com/eclipse/che/blob/master/agents/go-agents/core/activity/activity.go#L46
https://github.com/eclipse/che/blob/master/agents/go-agents/terminal-agent/main.go#L136
https://github.com/codenvy/codenvy/blob/master/plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/agent/CodenvyAgentModule.java#L43
I prefer that we think to port it in che instead of creating parallel functionality. The profits that I see.
 - General API for many agents like go terminal, ws-agent
 - Not connected to some backend-specific

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftRouteCreator.java
```diff
@@ -0,0 +1,82 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.fabric8.openshift.api.model.DoneableRoute;
+import io.fabric8.openshift.api.model.Route;
+import io.fabric8.openshift.api.model.RouteFluent.SpecNested;
+import io.fabric8.openshift.client.DefaultOpenShiftClient;
+import io.fabric8.openshift.client.OpenShiftClient;
+
```
skabashnyuk, line 21: javadoc. what is it, how it's used?

skabashnyuk (commented):
@benoitf can you review this pr too. Especially code related to LocalDockerSinglePortServerEvaluationStrategy

agents/ls-csharp/src/main/resources/org.eclipse.che.ls.csharp.script.sh
l0rd: For security reasons, by default, a container in OpenShift is run by an arbitrary user. An arbitrary user is a user with a random UID (and not a sudoer in the container).

We use this trick to be able to run a `*scripts.sh` even if the user is an arbitrary user.

File `${HOME}/is_arbitrary_user` is added using the following mechanism: https://github.com/redhat-developer/che-dockerfiles/blob/master/recipes/centos_jdk8/entrypoint.sh#L30

benoitf: could we check instead if sudo access is granted or not ?
https://superuser.com/questions/195781/sudo-is-there-a-command-to-check-if-i-have-sudo-and-or-how-much-time-is-left (so we don't need special trick in images. We can use any existing images)

l0rd: Yes of course that's what we've tried first. It did not work so we added this file to be able to skip `sudo`.

And to use existing images we would still need more work too. That's because in case of arbitrary user we need to modify `/etc/passwd/` and `/etc/group` as soon as the container is started https://github.com/redhat-developer/che-dockerfiles/blob/master/recipes/centos_jdk8/entrypoint.sh#L16.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
l0rd: Agreed. When we introduced the single port strategy the custom strategy wasn't implemented yet. And since then we haven't had the chance to rework it yet. But that's the direction we need to take.

l0rd:
@benoitf @skabashnyuk @garagatyi thanks for your reviews. We have planned to fix some of the issues you have raised during the next days:
- Use the new server evaluation strategy instead of the single port
- Find a better mechanism to run workspaces as arbitrary user
- Improve the implementation of the server and wsagent idling

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/idle/ServerIdleDetector.java
snjeza: It would be nice if we could port the code from the codenvy project.

skabashnyuk:
@l0rd  can you elaborate on 
>Improve the implementation of the server and wsagent idling

are you ok with codenvy way and we should backport it to che?

l0rd:
@skabashnyuk definitely! The only concern is that's not Open Source yet. So what I think we should do now is to remove idling from this PR and open a new PR with idling the codenvy way as soon as it's Open Sourced. @snjeza is going to take care of this.

gorkem:
@skabashnyuk @snjeza You can go ahead and do the port to Che... Welcome to Red Hat.

snjeza:
@skabashnyuk @benoitf @garagatyi I would try to port the activity stuff from the codenvy project to the che project. Is there any documentation or rules on this?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/Exec.java
```diff
@@ -20,7 +20,7 @@
     private final String[] command;
     private final String   id;
 
-    Exec(String[] command, String id) {
+    public Exec(String[] command, String id) {
```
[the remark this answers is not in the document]
amisevsk: Since OpenShiftConnector is in a different package, we cannot implement `startExec()` or `createExec()` without this class being public -- there seems to be no factory method.

sunix:
idling of workspaces will be part of this: https://github.com/eclipse/che/pull/5425

agents/ls-csharp/src/main/resources/org.eclipse.che.ls.csharp.script.sh
sunix: if I exclude these ones from the PR, would this PR be mergeable ? @skabashnyuk @benoitf @garagatyi

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
sunix: @benoitf is it ok with @davidfestal 's https://github.com/eclipse/che/pull/5110

agents/ls-csharp/src/main/resources/org.eclipse.che.ls.csharp.script.sh
benoitf: yes, removing that part would be better for this PR

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
benoitf: @sunix it's missing https://github.com/eclipse/che/pull/5456

davidfestal: This is fixed in PR https://github.com/eclipse/che/pull/5456. But it has not been merged to `openshift-connector-rebased` for now.

davidfestal: https://github.com/eclipse/che/pull/5110 was a first step to the final fix.

sunix:
@snjeza @benoitf @l0rd @davidfestal @skabashnyuk @garagatyi @amisevsk 
We covered pretty much all the comments for this PR, some of them have been excluded, some of them are part of other PRs, some of them are fixed.
I did a last rebase and tested successfully.
Plz give your approval to merge thanks

benoitf:
ci-build

benoitf (approved):

l0rd (approved):
LGTM but please double check that these changes are needed [EnvironmentDeserializerTest.java](https://github.com/eclipse/che/pull/5052/files#diff-2a14633cc3f99981cc2cfb2929caae74) and [CommandDeserializerTest.java](https://github.com/eclipse/che/pull/5052/files#diff-f48357d2975af2ca15e0c53ea4b3be49). Cf #5291 vs fba68ea04b0181c0d2f0d2cd9ed280853f4fb987

sunix:
@l0rd yes, the change were done in both `master` and `openshift-connector-rebased` branches but with different fixes. I've chosen to keep  fba68ea04b0181c0d2f0d2cd9ed280853f4fb987 as I found it cleaner not to pass `null` as values

l0rd:
@sunix ok thanks for the explanation

garagatyi (approved):

benoitf:
Adding @slemeur if some stuff is required on documentation as the code review seems to be ok

slemeur (approved):

Issue #4803, opened 2017-04-13
Bulk operation works wrong after filtering items


SkorikSergey:
**Precondition:**
Should be created some workspaces in account

**Reproduction Steps:**
Filter some workspaces by defined symbols
Click the Bulk checkbox (select all items from this list)
Click the Delete button

**Expected behavior:**
Only filtered workspaces deleted

**Observed behavior:**
All workspaces from list deleted

Che version: 5.8.0-SNAPSHOT
OS and version: Ubuntu 16.04 LTS
![anim](https://cloud.githubusercontent.com/assets/7760565/24999701/401cac7e-2047-11e7-9179-3b10a21b0c28.gif)

PR #5083, opened 2017-05-16
CHE-4803: fix items bulk selection and filtration.

akurinnoy:
### What does this PR do?
This PR adds the helper class which provides selection and filtration of items in a list. 
There were fixed all lists where items can be simultaneously selected and filtered.

#4803 

#### Changelog

[UD] Fixed bug when bulk operations work wrong after items are filtered.

#### Release Notes

N/A - bugfix

#### Docs PR

N/A - bugfix

### DIFF
--- a/dashboard/src/components/widget/list/che-list-helper.factory.ts
+++ b/dashboard/src/components/widget/list/che-list-helper.factory.ts
@@ -1,0 +1,64 @@
+/*
+ * Copyright (c) 2015-2017 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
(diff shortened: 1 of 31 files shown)


ashumilova (approved):

olexii4 (approved):

dashboard/src/components/widget/list/che-list-helper.factory.ts
```diff
@@ -0,0 +1,66 @@
...
+    return helper;
+  }
+
+  /**
+   * Removes specified helper's instance.
+   *
+   * @param id {string} a string to identify helper.
+   */
+  removeHelper(id: string): void {
+    if (this.helpers.has(id)) {
```
benoitf: do we need to check if map has the element, AFAIK delete will return either true or false but it can accept any element

benoitf (approved):

dashboard/src/components/widget/list/che-list-helper.factory.ts
akurinnoy: you're right, I'll remove this check.

Issue #4132, opened 2017-02-14
Technical issue: ProcessPanel deletes widgets not clear. 


AndrienkoAleksandr:
When we delete widgets (terminals or command output widgets) from ProcessPanel, this widgets are still exist in the memory, because in the code we didn't clean up them correctly. This is ProcessPanel performance issue, because references to deleted widgets still exists and we have a problems, for example, with resize process panel, because we resizes existing widgets with already deleted widgets, so ui begin resize with lags.
Main goal this is issue: correctly clean up maps in the ProcessPanelPresenter and ProcessPanelViewImpl.

**OS and version:**
Fedora 24 (os version doesn't matter in this case). 
CHE version: 5.3.0-SNAPSHOT

PR #5087, opened 2017-05-16
CHE-4132: Fix clean up map after widget deletion from ProcessPanel.

AndrienkoAleksandr:
### What does this PR do?
Clean up map on deleting widget from Process Panel. 
Don't show "Stop command" window if workspace has already stopped.

https://github.com/eclipse/che/issues/4132

#### Changelog
Clean up map on deleting widget from Process Panel. 
Don't show "Stop command" window if workspace has already stopped.

#### Release Notes
Clean up map on deleting widget from Process Panel. 
Don't show "Stop command" window if workspace has already stopped.

### DIFF
--- a/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/processes/panel/ProcessesPanelPresenter.java
+++ b/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/processes/panel/ProcessesPanelPresenter.java
@@ -337,7 +337,7 @@ public class ProcessesPanelPresenter extends BasePresenter implements ProcessesP
     @Override
     public void onCloseTerminal(ProcessTreeNode node) {
         closeTerminal(node);
-        view.hideProcessOutput(node.getId());
+        view.removeWidget(node.getId());
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


azatsarynnyy (approved):

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/processes/panel/ProcessesPanelPresenter.java
```diff
@@ -674,12 +664,7 @@ public void onStopCommandProcess(ProcessTreeNode node) {
 
     @Override
     public void onCloseCommandOutputClick(final ProcessTreeNode node) {
-        closeCommandOutput(node, new SubPanel.RemoveCallback() {
-            @Override
-            public void remove() {
-                view.hideProcessOutput(node.getId());
-            }
-        });
+        closeCommandOutput(node, () -> {});
```
vzhukovs, line 667: Maybe it would be better to pass `null` as third argument instead of empty callback?

AndrienkoAleksandr: I think no, because it's can do logic more difficult because we should add null checking in the few methods.

vzhukovs (approved):

vparfonov (approved):

PR #5109, opened 2017-05-16
Gwt logger rework

vparfonov:
### What does this PR do?
keep in our source only 2 changed classes other will get from original lib sources


skabashnyuk (approved):

vparfonov:
ci-build

PR #5110, opened 2017-05-16
[MERGE-AFTER-PR-5366] Multi-container workspace Support

davidfestal:
### What does this PR do?

Fixes the support of multi-container (multi-machine) Che workspaces inside Openshift.

In the current design of the OpenShift connector, each Che machine create is transformed into an OpenShift Pod with a single container. (and not a Pod with many containers).

However, the multi-machine use-case wasn't fully tested and NPEs used to prevent the setup of non-development additional machines. This is now fixed.

Additionally, since machines live in distinct pods, the docker compose internal network feature cannot be used to allow machines communicating with their compose service names.
To fix this, the OpenShift service name, for all non-development Che machines, will be created with the name of the related compose service. This way we can leverage the OpenShift internal Dns server that allows all pods in an OS namespace to communicate through their OS service name.

https://issues.jboss.org/browse/CHE-175
https://issues.jboss.org/browse/CHE-258
https://issues.jboss.org/browse/CHE-259

### DIFF
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/kubernetes/KubernetesLabelConverter.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/kubernetes/KubernetesLabelConverter.java
@@ -64,6 +64,9 @@ public final class KubernetesLabelConverter {
      */
     public static Map<String, String> labelsToNames(Map<String, String> labels) {
         Map<String, String> names = new HashMap<>();
+        if (labels == null) {
+            return names;
+        }
         for (Map.Entry<String, String> label : labels.entrySet()) {
 
             if (!hasConversionProblems(label)) {
@@ -103,6 +106,9 @@ public final class KubernetesLabelConverter {
      */
     public static Map<String, String> namesToLabels(Map<String, String> names) {
         Map<String, String> labels = new HashMap<>();
+        if (names == null) {
+            return labels;
+        }
         for (Map.Entry<String, String> entry: names.entrySet()){
             String key = entry.getKey();
             String value = entry.getValue();
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
@@ -68,6 +68,7 @@ import org.eclipse.che.plugin.docker.client.json.NetworkCreated;
 import org.eclipse.che.plugin.docker.client.json.NetworkSettings;
 import org.eclipse.che.plugin.docker.client.json.PortBinding;
 import org.eclipse.che.plugin.docker.client.json.network.ContainerInNetwork;
+import org.eclipse.che.plugin.docker.client.json.network.EndpointConfig;
 import org.eclipse.che.plugin.docker.client.json.network.Ipam;
 import org.eclipse.che.plugin.docker.client.json.network.IpamConfig;
 import org.eclipse.che.plugin.docker.client.json.network.Network;
@@ -167,6 +168,7 @@ public class OpenShiftConnector extends DockerConnector {
     private static final String CHE_CONTAINER_IDENTIFIER_LABEL_KEY       = "cheContainerIdentifier";
     private static final String CHE_DEFAULT_EXTERNAL_ADDRESS             = "172.17.0.1";
     private static final String CHE_WORKSPACE_ID_ENV_VAR                 = "CHE_WORKSPACE_ID";
+    private static final String CHE_IS_DEV_MACHINE_ENV_VAR               = "CHE_IS_DEV_MACHINE";
     private static final int CHE_WORKSPACE_AGENT_PORT                    = 4401;
     private static final int CHE_TERMINAL_AGENT_PORT                     = 4411;
     private static final String DOCKER_PROTOCOL_PORT_DELIMITER           = "/";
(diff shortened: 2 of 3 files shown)


davidfestal:
@l0rd, @gorkem, could you review this PR please ?

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/kubernetes/KubernetesLabelConverter.java
```diff
@@ -64,31 +64,33 @@ public static String getCheServerLabelPrefix() {
...
-                key = addPadding(key);
-                value = addPadding(value);
-
-                if (matchesKubernetesLabelRegex(key) && matchesKubernetesLabelRegex(value)) {
-                    names.put(key, value);
-                } else {
-                    LOG.error(
-                            "Could not convert label {} into Kubernetes annotation: labels must be alphanumeric with ':' and '/'",
-                            label.toString());
+        if (labels != null) {
```
l0rd: Avoiding to nest the for loop inside the `if` would make it more readable:
```  java
if (labels == null) {
  returns names;
}

for (Map.Entry<String, String> label : labels.entrySet()) {
 ...
}
```

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/kubernetes/KubernetesLabelConverter.java
```diff
@@ -103,25 +105,27 @@ public static String getCheServerLabelPrefix() {
...
-
-            // Remove padding
-            Matcher keyMatcher   = CHE_SERVER_LABEL_KEY.matcher(key);
-            Matcher valueMatcher = CHE_SERVER_LABEL_KEY.matcher(value);
-            if (!keyMatcher.matches() || !valueMatcher.matches()) {
-                continue;
-            }
-            key = keyMatcher.group(1);
-            value = valueMatcher.group(1);
+        if (names != null) {
```
l0rd: Same comment here than above.

l0rd (approved):
If you have time you can extract the for loops outside the if to make code more easy to read. But otherwise LGTM

l0rd:
cc @sunix

ide/che-core-ide-stacks/src/main/resources/stacks.json
```diff
@@ -1,5 +1,74 @@
 [
   {
+    "id": "java-centos-mysql",
```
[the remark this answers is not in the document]
davidfestal: I don't know. @l0rd can tell.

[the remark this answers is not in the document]
l0rd: @davidfestal @sunix it makes sense to me to merge this stack to eclipse/che master

davidfestal:
@l0rd I committed e66d31df1b29033b4ccff9ca9d7ac9c123a28a20 as you suggested. Do you think you could you merge ?

ide/che-core-ide-stacks/src/main/resources/stacks.json
```diff
@@ -1,5 +1,74 @@
...
+      "type": "image",
+      "origin": "rhche/centos_jdk8"
+    },
+    "workspaceConfig": {
+      "environments": {
+        "default": {
+          "machines": {
+            "dev-machine": {
+              "agents": [
+                "org.eclipse.che.terminal", "org.eclipse.che.ws-agent", "com.redhat.bayesian.lsp"
```
sunix: if this is supposed to go to master, it shouldn't include com.redhat.bayesian.lsp agent

l0rd: That's a good point @sunix. I think @davidfestal did it for consistency since other stacks use the same agent here. But I understand that's an issue for you because you are in the middle of a rebasing. I think we have 2 choices:
- @davidfestal remove `com.redhat.bayesian.lsp` and we merge it right now
- @davidfestal change the PR to be merge on `master` branch instead of `openshift-connector`

@sunix are these 2 options ok for you?
@davidfestal which of these 2 options is better for you?

l0rd: When I say change the PR to be merged on `master` I actually mean to split this PR in two:
- one PR for `master` with the modifications of `stacks.json`
- one PR for `openshift-connector` with modifications on the `OpenShiftConnector`

davidfestal:
@l0rd 

> remove `com.redhat.bayesian.lsp` and we merge it right now

Let me choose this option. Then we'll be able to merge on `openshift-connector` branch right now, right ?

davidfestal:
@l0rd @sunix I changed the stacks file to remove the default activation of Bayesian. Is it OK to merge it ?

sunix:
The second problem is that in the branch `openshift-connector-rebased` and `openshift-connector-wip` we did reverted all changes on stack.json, so if you could cherry-pick and fix conflicts + send PR (to openshift-connector-rebased) it would work for me. + we need to plan a task so the changes are applied in rh-che stacks.json as well

davidfestal:
@sunix. Not sure I fully understand what you're proposing. But let me remove the `stacks.json` from this PR and simply add the `stacks.json` change in a PR against master, since @l0rd said the newly-created `java-mysql-centos` stack was expected to finally be merged to `eclipse/che` `master`. This shouldn't bring any conflict like that, right ? Of course we'll have to reintegrate this new stack into the `rh-che` `stacks.json` file.

@l0rd wdyt ?

davidfestal:
@sunix I removed the `stacks.json` from this PR and  created a new PR for it (#5227).
@l0rd I added the changes that also fix https://issues.jboss.org/browse/CHE-258 and https://issues.jboss.org/browse/CHE-259

Is it possible to merge this ?

l0rd:
@davidfestal can you please rebase this on top of branche openshift-connector-rebased

davidfestal:
@l0rd @sunix I rebased to the `openshift-connector-rebased` branch

However this should be rebuilt and tested again when we'll be ready to merge to `openshift-connector-branch` or to `master`.

davidfestal:
@sunix could you review this PR ? it should be able to be merged without any problem / conflicts now.

benoitf:
maybe in the description, it could say that each Eclipse Che machine is transformed into an OpenShift Pod with a single container. (and not a Pod with many containers)

davidfestal:
> maybe in the description, it could say that each Eclipse Che machine is transformed into an OpenShift Pod with a single container. (and not a Pod with many containers)

Sure let me add a more detailed explanation

Issue #4197, opened 2017-02-22
Add CHE_DOCKER_ALWAYS__PULL__IMAGE for run portion, not just build


ghost:
I'd like Che not to check if there's a newer stack image on Docker Hub. However, setting CHE_DOCKER_ALWAYS__PULL__IMAGE to false in che.env does not help:

```
[DOCKER] latest: Pulling from eclipse/ubuntu_jdk8 
[DOCKER] Digest: sha256:69f514b226808be7d251442096bcc61d85dcb77f91f453e1cfe090a8f350c854 
[DOCKER] Status: Image is up to date for eclipse/ubuntu_jdk8:latest 
```

This makes it impossible to use Che in a purely offline mode.

ghost:
Blocker for Artik. Offline mode is impossible.

TylerJewell:
Is the `che.property` value (or alias) `che.docker.always_pull_image`?  If yes, have you tried `machine.docker.pull_image` as well?  There is some code in the core Server that assigns both values to be the same, so I wonder if this alternative version works.

https://github.com/eclipse/che/blob/9370327d4606f98a904c09a45fe9923aa6df5684/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che_aliases.properties#L49

ghost:
That did not work either

ghost:
It turns out this is not a bug but an enhancement request.

CHE_DOCKER_ALWAYS__PULL__IMAGE works only for build, not run.

bmicklea:
Increasing severity as this prevents true offline mode

mmorhun:
Done

zanoubia:
Still having the same problem, I think, with codenvy offline mode.
Although I already downloaded the stacks I need, I can't run workspaces

# docker images
REPOSITORY            TAG                 IMAGE ID            CREATED             SIZE
.
.
eclipse/centos_jdk8   latest              5bd02d36ed35        5 weeks ago         877MB

And I have the following logs in instance/logs/codenvy/che-machines-logs/machineuswhc5e31bn9wnzm/machineId.logs

[DOCKER] Step 1/1 : FROM eclipse/centos_jdk8
[DOCKER]
[DOCKER] [ERROR] Failed to build image: Get https://registry-1.docker.io/v2/: dial tcp: lookup registry-1.docker.io on  <>:53: no such host
[ERROR] Docker image build failed. Image id not found in build output.

mmorhun:
[...]

PR #5112, opened 2017-05-17
CHE-4197: Apply CHE_DOCKER_ALWAYS__PULL__IMAGE to container creation

mmorhun:
### What does this PR do?
Takes into account `CHE_DOCKER_ALWAYS__PULL__IMAGE` value when creating docker container for a workspace machine. This means, that if the property is set to `false` then Che won't try to pull image for a container if it exists locally.

https://github.com/eclipse/che/issues/4197

#### Changelog
Respect the value of CHE_DOCKER_ALWAYS__PULL__IMAGE when creating containers.

#### Release Notes
We've improved the behavior of Che when it's run in a secure environment disconnected from the internet. In the `che.env` there is a property `CHE_DOCKER_ALWAYS_PULL_IMAGE`that when sent to `false` will ensure that Che only checks the local (private) registry for the Docker image. The default value is `true` and in this case Che will always check DockerHub for a newer image version before creating the container.


plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
```diff
@@ -518,7 +520,9 @@ protected void pullImage(CheServiceImpl service,
 
         try {
             boolean isSnapshot = SNAPSHOT_LOCATION_PATTERN.matcher(dockerMachineSource.getLocation()).matches();
-            if (!isSnapshot || snapshotUseRegistry) {
+            boolean isImageExistLocally = isDockerImageExistLocally(dockerMachineSource.getRepository());
+            if (!isSnapshot && (doForcePullImage || !isImageExistLocally) ||
```
garagatyi: In my opinion, this code would be more clear if operands wouldn't rely on remembering operations priority. E.g. next equivalent code looks clearer for me:
```
(!isSnapshot && (doForcePullImage || !isImageExistLocally) || isSnapshot && snapshotUseRegistry)
```

garagatyi (approved):

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java
mmorhun: I divided this with new line, but I do not mind about brackets. Let be so.

bmicklea (approved):
updated release notes and changelog.

mmorhun:
ci-build

skabashnyuk (approved):

PR #5115, opened 2017-05-17
Enable single port exposure on Che

benoitf:
### What does this PR do?
Enable single port exposure on Che

Toggle is made by enabling `CHE_SINGLE_PORT` in the `che.env` file

By enabling single-port, all browser traffic to Che or any workspace will be routed through the value that you have set to CHE_PORT`, or 8080 if not set. Setting this property will transform the launch sequence of Che to launch a Traefik reverse proxy. The reverse proxy will act as the traffic endpoint for all browser communications. When a new workspace is started or stopped, Che will update Traefik's configuration with rules for how browser traffic should be routed to Che or a workspace.

It’s now using an **official Traefik** image (before I was using a custom made image (_florentbenoit/traefik_))

- There is an interceptor with a _kill switch_. It means interceptor is applied only if plug-in is enabled (not only if plug-in is added at compilation)
It is automatically enabled when `CHE_SINGLE_PORT` is turned on
- docker-compose file is handling if the single_port is turned on or off and then add the traefik container and redirect port only if the property is enabled. (not enabled by default)
- using `--debug` flag when launching Eclipse Che is also turning on the traffic web console to view traefik routes
- It is not enabled by default, so it means that without user change, there is no overhead, no useless container started, etc.

#4361

#### Changelog
Enable single port exposure on Che

#### Release Notes
By enabling single-port, all browser traffic to Che or any workspace will be routed through the value that you have set to CHE_PORT`, or 8080 if not set. Setting this property will transform the launch sequence of Che to launch a Traefik reverse proxy. The reverse proxy will act as the traffic endpoint for all browser communications. When a new workspace is started or stopped, Che will update Traefik's configuration with rules for how browser traffic should be routed to Che or a workspace.

#### Docs PR
https://github.com/eclipse/che-docs/pull/229

Change-Id: I12644d9202dadc0b10104f78bb055425ca6611ac


slemeur:
ok for me, just needs the docs ;)

dockerfiles/init/manifests/che.env
```diff
@@ -235,6 +235,16 @@
...
 #CHE_DOCKER_IP_EXTERNAL=NULL
 
+# Usage of single-port routing.
+# By enabling single-port, all browser traffic to Che or any workspace will be routed
+# through the value that you have set to CHE_PORT`, or 8080 if not set. Setting this
+# property will transform the launch sequence of Che to launch a Traefik reverse proxy.
+# The reverse proxy will act as the traffic endpoint for all browser communications.
+# When a new workspace is started or stopped, Che will update Traefik's configuration
+# with rules for how browser traffic should be routed to Che or a workspace.
+CHE_SINGLE_PORT=false
```
garagatyi: This explanation doesn't describe how it interferes with DNS resolution. I wonder whether it does?

plugins/plugin-traefik/plugin-traefik-docker/src/main/java/org/eclipse/che/plugin/traefik/TraefikDockerModule.java
```diff
@@ -0,0 +1,43 @@
...
+public class TraefikDockerModule extends AbstractModule {
+
+    /**
+     * Configure the traefik components
+     */
+    @Override
+    protected void configure() {
+
+        // add logic only if plug-in is enabled.
+        if (Boolean.parseBoolean(firstNonNull(System.getenv("CHE_PLUGIN_TRAEFIK_ENABLED"), "false"))) {
```
garagatyi: I suppose `Boolean.parseBoolean(System.getenv("CHE_PLUGIN_TRAEFIK_ENABLED"))` would do the same, how do you think?

plugins/plugin-traefik/plugin-traefik-docker/src/main/java/org/eclipse/che/plugin/traefik/TraefikCreateContainerInterceptor.java
```diff
@@ -0,0 +1,148 @@
...
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Stream;
+
+import static java.lang.String.format;
+
+/**
+ * Intercept the calls on createContainer on docker Connector.
```
garagatyi: Can you add the purpose of such an addition to the docs? Without context, it is not clear why it does so.

plugins/plugin-traefik/plugin-traefik-docker/src/main/java/org/eclipse/che/plugin/traefik/TraefikCreateContainerInterceptor.java
```diff
@@ -0,0 +1,148 @@
...
+     *         intercepting data of createContainer method on {@link DockerConnector}
+     * @return the result of the intercepted method
+     * @throws Throwable
+     *         if there is an exception
+     */
+    @Override
+    public Object invoke(MethodInvocation methodInvocation) throws Throwable {
+        ServerEvaluationStrategy serverEvaluationStrategy = serverEvaluationStrategyProvider.get();
+        // Abort if custom server evaluation strategy is not enabled.
+        if (!(CustomServerEvaluationStrategy.class.isInstance(serverEvaluationStrategy))) {
```
garagatyi, line 72: Correct me if I'm mistaken but with this code, the interceptor is enabled whenever the custom strategy is applied.
If so what if the custom strategy is used with some external proxy instead of Traefik.
Then the custom strategy looks rather Traefik oriented than custom while from the configuration point of view it looks like it really custom and can be used in a wide range of situations.

plugins/plugin-traefik/plugin-traefik-docker/src/main/java/org/eclipse/che/plugin/traefik/TraefikCreateContainerInterceptor.java
```diff
@@ -0,0 +1,148 @@
...
+            return methodInvocation.proceed();
+        }
+        final CustomServerEvaluationStrategy customServerEvaluationStrategy = (CustomServerEvaluationStrategy)serverEvaluationStrategy;
+
+        // Get the connector
+        DockerConnector dockerConnector = (DockerConnector)methodInvocation.getThis();
+
+        // only one parameter which is CreateContainerParams
+        CreateContainerParams createContainerParams = (CreateContainerParams)methodInvocation.getArguments()[0];
```
garagatyi, line 81: Florent, do you know whether this code works with Openshift based Che? I wonder whether workarounds with replacing DockerConnector with Openshift related entities and this code may work correctly together.

garagatyi (commented):
Overall looks good. It is an interesting feature, good work!
Can you elaborate on commented lines to align reviewers view of these changes?

dockerfiles/init/manifests/che.env
benoitf: You need extra doc ?

plugins/plugin-traefik/plugin-traefik-docker/src/main/java/org/eclipse/che/plugin/traefik/TraefikDockerModule.java
benoitf: yes correct, I will fix

plugins/plugin-traefik/plugin-traefik-docker/src/main/java/org/eclipse/che/plugin/traefik/TraefikCreateContainerInterceptor.java
benoitf: sure

garagatyi:
@sunix @gorkem @benoitf Looks like this PR interferes with PR https://github.com/eclipse/che/pull/5052 in the area of single port strategy.
I have concerns that we need 2 strategies with a single port feature at the same time and even in the review process at the same time.
Maybe we can use server strategy from this PR to use in Openshift packaging?

plugins/plugin-traefik/plugin-traefik-docker/src/main/java/org/eclipse/che/plugin/traefik/TraefikCreateContainerInterceptor.java
benoitf: in Guice module, we do not bind the interceptor if we do not enabled traefik.
https://github.com/eclipse/che/blob/683b6f4b8253cc83c8ce7500020c7bf2ee8a8130/plugins/plugin-traefik/plugin-traefik-docker/src/main/java/org/eclipse/che/plugin/traefik/TraefikDockerModule.java#L35-L40
so it's an extra check in case the interceptor is bound but the strategy is not the expected one.

And you can still enable the custom strategy without enabling the traefik plug-in/interceptor

plugins/plugin-traefik/plugin-traefik-docker/src/main/java/org/eclipse/che/plugin/traefik/TraefikCreateContainerInterceptor.java
benoitf: well due to all changes on https://github.com/eclipse/che/pull/5052 I will check once it is merged but either we can drop support for OpenShift connector or if there are issues, it will be fixed.

dockerfiles/init/manifests/che.env
garagatyi: Right now I want to understand how it would work in the different situation. Then I would understand whether I believe additional docs are needed.

plugins/plugin-traefik/plugin-traefik-docker/src/main/java/org/eclipse/che/plugin/traefik/TraefikCreateContainerInterceptor.java
garagatyi: ok then, thx for the explanation

dockerfiles/init/manifests/che.env
benoitf: I will add docs

plugins/plugin-traefik/plugin-traefik-docker/src/main/java/org/eclipse/che/plugin/traefik/TraefikCreateContainerInterceptor.java
garagatyi: In this particular comment, I meant the current state of Openshift packaging in master, not from that PR. 
I have concerns because DockerConnector is replaced with Openshift patched version and it works in a different way. So I want to ensure that if those components don't work together we know that and document that.

dockerfiles/init/manifests/che.env
benoitf: doc added

plugins/plugin-traefik/plugin-traefik-docker/src/main/java/org/eclipse/che/plugin/traefik/TraefikDockerModule.java
benoitf: fixed

plugins/plugin-traefik/plugin-traefik-docker/src/main/java/org/eclipse/che/plugin/traefik/TraefikCreateContainerInterceptor.java
benoitf: doc added

plugins/plugin-traefik/plugin-traefik-docker/src/main/java/org/eclipse/che/plugin/traefik/TraefikCreateContainerInterceptor.java
benoitf: In the current state there is no issue.

benoitf:
Added link to documentation PR

slemeur (approved):

benoitf:
@garagatyi, about
> I have concerns that we need 2 strategies with a single port feature at the same time and even in the review process at the same time.
> Maybe we can use server strategy from this PR to use in Openshift packaging?

OpenShift doesn't need a front-end (it's what is done in this PR) so this PR doesn't collide with https://github.com/eclipse/che/pull/5052

but#5052 OpenShift PR collide with the existing https://github.com/eclipse/che/blob/52e418bc73c36eaa86c9a427a6014f923de797b7/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java that allow to provide a different naming for workspace urls.

sunix:
@garagatyi  @gorkem @benoitf yes this interferes with #5052 we should discuss with @l0rd about that

benoitf:
@sunix : it doesn't. #5052 is introducing a custom server evaluation strategy provider. This one is not, it reuse a previously one. It only allow to provide a frontend/proxy which is traefik.
But you may enable only the custom server evaluation strategy provider without enabling traefik frontend.

skabashnyuk (approved):

garagatyi:
Well looks like #5052 interferes with an existing solution that is already merged and not related to this PR.
@sunix @gorkem @l0rd Can you take a look at that strategy to figure out whether it can be reused?

benoitf:
Also the discussion started a long time ago on https://github.com/eclipse/che/pull/4269

garagatyi (approved):

sunix:
I works for me as the 2 can live together (we select either one strategy or another in che.env). Though in a second step we should consider using a custom one in openshift connector

gorkem:
@sunix can you create another issue to reuse this and replace the openshift one?

riuvshin:
@benoitf hey, is setting `CHE_SINGLE_PORT=true` enough to test this? CHE does not work for us here.

Issue #4644, opened 2017-03-30
Autocomplete into commands editor does not work after save


artaleks9:
**Reproduction Steps:**
* Open a project in the IDE.
* Create a command into commands editor
* Make some changes and save them.
* Try to use autocomplete in the commands editor

**Expected behavior:**
autocomplete should work

**Observed behavior:**
* autocomplete form is open, but feature does not work

**Che version:** 5.6.0
**OS and version:** Ubuntu 14.04
**Docker version:** 1.13
**Che install:** [Docker container]

**Additional information:**
* Problem can be reliably reproduced, doesn't happen randomly: [Yes]
* See the attachment
![autocomplete_commands_editor](https://cloud.githubusercontent.com/assets/5408906/24501586/d0a89848-1552-11e7-83ad-0585a5bad960.png)

slemeur:
@azatsarynnyy Could you please check this issue?

azatsarynnyy:
@slemeur I confirm, it's reproducible on my side

PR #5133, opened 2017-05-19
CHE-4670: fix a bug with command auto completion

svor:
### What does this PR do?
Makes command auto completion  without errors.

https://github.com/eclipse/che/issues/4644


dkulieshov:
It looks like there is a missing test

dkulieshov:
what is telling me that there is no regression after the change ?

svor:
yes

vparfonov (approved):

PR #5157, opened 2017-05-22
CHE-220: Improving 'removeContainer' API. Refactoring

ibuziuk:
### What does this PR do?
Enhance OpenShift resource deletion for workspaces (scaling deployment down before removing resources). Refactoring

https://issues.jboss.org/browse/CHE-220

### DIFF
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
@@ -130,7 +130,6 @@ import io.fabric8.kubernetes.api.model.VolumeMount;
 import io.fabric8.kubernetes.api.model.VolumeMountBuilder;
 import io.fabric8.kubernetes.api.model.extensions.Deployment;
 import io.fabric8.kubernetes.api.model.extensions.DeploymentBuilder;
-import io.fabric8.kubernetes.api.model.extensions.ReplicaSet;
 import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.Watcher;
 import io.fabric8.kubernetes.client.dsl.ExecWatch;
@@ -142,8 +141,6 @@ import io.fabric8.openshift.api.model.DoneableDeploymentConfig;
 import io.fabric8.openshift.api.model.Image;
 import io.fabric8.openshift.api.model.ImageStream;
 import io.fabric8.openshift.api.model.ImageStreamTag;
-import io.fabric8.openshift.api.model.Route;
-import io.fabric8.openshift.api.model.RouteList;
 import io.fabric8.openshift.client.DefaultOpenShiftClient;
 import io.fabric8.openshift.client.OpenShiftClient;
 import io.fabric8.openshift.client.dsl.DeployableScalableResource;
@@ -158,8 +155,9 @@ import io.fabric8.openshift.client.dsl.DeployableScalableResource;
 @Singleton
 public class OpenShiftConnector extends DockerConnector {
     private static final Logger LOG                                      = LoggerFactory.getLogger(OpenShiftConnector.class);
-    public static final String CHE_OPENSHIFT_RESOURCES_PREFIX           = "che-ws-";
-    public static final String OPENSHIFT_DEPLOYMENT_LABEL               = "deployment";
+    public static final String CHE_OPENSHIFT_RESOURCES_PREFIX            = "che-ws-";
+    public static final String OPENSHIFT_DEPLOYMENT_LABEL                = "deployment";
+    public static final int OPENSHIFT_WAIT_POD_TIMEOUT                   = 240;
 
     private static final String CHE_CONTAINER_IDENTIFIER_LABEL_KEY       = "cheContainerIdentifier";
     private static final String CHE_DEFAULT_EXTERNAL_ADDRESS             = "172.17.0.1";
@@ -168,7 +166,6 @@ public class OpenShiftConnector extends DockerConnector {
     private static final int CHE_TERMINAL_AGENT_PORT                     = 4411;
     private static final String DOCKER_PROTOCOL_PORT_DELIMITER           = "/";
     private static final int OPENSHIFT_WAIT_POD_DELAY                    = 1000;
-    private static final int OPENSHIFT_WAIT_POD_TIMEOUT                  = 240;
     private static final int OPENSHIFT_IMAGESTREAM_WAIT_DELAY            = 2000;
     private static final int OPENSHIFT_IMAGESTREAM_MAX_WAIT_COUNT        = 30;
     private static final long OPENSHIFT_POD_TERMINATION_GRACE_PERIOD     = 0;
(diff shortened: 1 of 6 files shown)


ibuziuk:
cc @mlabuda

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -158,8 +155,9 @@
 @Singleton
 public class OpenShiftConnector extends DockerConnector {
     private static final Logger LOG                                      = LoggerFactory.getLogger(OpenShiftConnector.class);
-    public static final String CHE_OPENSHIFT_RESOURCES_PREFIX           = "che-ws-";
-    public static final String OPENSHIFT_DEPLOYMENT_LABEL               = "deployment";
+    public static final String CHE_OPENSHIFT_RESOURCES_PREFIX            = "che-ws-";
+    public static final String OPENSHIFT_DEPLOYMENT_LABEL                = "deployment";
+    public static final int OPENSHIFT_WAIT_POD_TIMEOUT                   = 240;
```
sunix, line 160: are these ones real changes or just formatting ?

ibuziuk: real change & 2 deliberate shifts of '=' sign

mlabuda (changes requested):
One comment, otherwise LGTM.

sunix (approved):
LGTM

ibuziuk:
@mlabuda there is no comment ;-)

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftContainerRemover.java
```diff
@@ -0,0 +1,139 @@
...
+import io.fabric8.openshift.api.model.Route;
+import io.fabric8.openshift.client.DefaultOpenShiftClient;
+import io.fabric8.openshift.client.OpenShiftClient;
+
+@Singleton
+public class OpenShiftContainerRemover {
+    private static final Logger LOG = LoggerFactory.getLogger(OpenShiftContainerRemover.class);
+
+    public void removeContainer(final String deploymentName, final String namespace) throws IOException {
+        scaleDownDeployment(deploymentName, namespace);
```
mlabuda: did you check that waiting in scaleDownDeployment works? If you remember, there is that bug with k8 client in deployment config waitings (https://github.com/fabric8io/kubernetes-client/issues/699). If it is same for deployments, then it waits for fixed time and then continue...

mlabuda (changes requested):
Ah, my comment somehow disappeared between posting a comment and submitting review. See comment now.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftContainerRemover.java
ibuziuk: @mlabuda it works for me just fine

mlabuda: If it works, +1 ;)

mlabuda (approved):

l0rd (commented):
Great job @ibuziuk but I would rename `OpenShiftContainerRemover` class and `removeContainer()` method because we are not removing a container but a deployment. Maybe `OpenShiftDeploymentCleaner` and `cleanDeploymentResources()` are more indicated

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftContainerRemover.java
```diff
@@ -0,0 +1,139 @@
...
+import io.fabric8.kubernetes.client.Watch;
+import io.fabric8.kubernetes.client.Watcher;
+import io.fabric8.kubernetes.client.dsl.FilterWatchListDeletable;
+import io.fabric8.kubernetes.client.dsl.ScalableResource;
+import io.fabric8.openshift.api.model.Route;
+import io.fabric8.openshift.client.DefaultOpenShiftClient;
+import io.fabric8.openshift.client.OpenShiftClient;
+
+@Singleton
+public class OpenShiftContainerRemover {
```
amisevsk: WDYT about making this class a static utility class? That way we don't have to inject it in OpenShiftConnector, and can instead refer to static methods. The OpenShiftConnector constructor is already getting quite unwieldy.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftContainerRemover.java
```diff
@@ -0,0 +1,139 @@
...
+            }
+
+            if (service != null) {
+                LOG.info("Removing OpenShift Service {}", service.getMetadata().getName());
+                openShiftClient.resource(service).delete();
+            }
+        }
+    }
+    
+    private void waitUntilWorspacePodIsDeleted(final String deploymentName, final String namespace) throws OpenShiftException {
```
amisevsk: Typo: waitUntilWor**k**spacePodIsDeleted

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftRouteCreator.java
```diff
@@ -19,17 +21,18 @@
...
     private static final String TLS_TERMINATION_EDGE = "edge";
     private static final String REDIRECT_INSECURE_EDGE_TERMINATION_POLICY = "Redirect";
 
-    public static void createRoute (final String namespace,
-                                    final String workspaceName,
-                                    final String openShiftNamespaceExternalAddress,
-                                    final String serverRef,
-                                    final String serviceName,
-                                    final boolean enableTls) {
+    public void createRoute (final String namespace,
```
amisevsk, line 30: What do we gain by making this class non-static?

amisevsk (changes requested):
One typo that should be fixed, looks good to me apart from that. I've added a few comments for clarification as well.

amisevsk:
+1 to the rename suggested by @l0rd

amisevsk (approved):

ibuziuk:
@amisevsk @l0rd rename & typo are fixed
in terms of static vs. singleton components (in our context it can be also rephrased to procedure programming vs OOP)

Basically, the rule-of-thumb is that static methods should be used for pure utilities like `KubernetesResourceUtil.java` and code reuse should be achieved via composition in OOP (via DI in our case).

However, I do agree that with the increasing number of components the constructor is becoming polluted with the args.  In Spring, for example, there is no such problem with polluted constructors - 
`@Autowired` and `@Value` annotations are used directly on class fields for injecting components / env vars. Maybe, it is possible to reimplement DI in OpenShiftConnector and inject properties without using constructor ? 

In general, I would  prefer to use `@Singleton` annotation for stateless components as much as possible

amisevsk:
@ibuziuk Alright, makes sense. :+1:

l0rd (approved):

l0rd:
If @sunix is ok we can merge

ibuziuk:
cherry-picking to `openshift-connector-rebased` https://github.com/sunix/che/pull/1

ibuziuk:
cherry-picking to origin's `openshift-connector-rebased` https://github.com/eclipse/che/pull/5191

PR #5168, opened 2017-05-23
CODENVY-2199: Do not snapshot /tmp folder

mmorhun:
### What does this PR do?
Prevents adding `/tmp` folder from a workspace container into a snapshot.
Added ability to specify a user for docker exec.

https://github.com/codenvy/codenvy/issues/2199

#### Changelog
Ignore `/tmp` folder when creating workspace snapshot.

#### Release Notes
Snapshots No Longer Capture `tmp`
We have implemented a small but important change in how we handle snapshots. In order to speed up snapshotting when stopping a workspace, and snapshot restore when workspaces startup we are no longer snapshotting the `tmp` directory. In many cases this has trimmed >50% of the total snapshot size and noticeably increased the speed of snapshotting and restoring.

#### Docs PR
https://github.com/eclipse/che-docs/pull/232

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
@@ -531,6 +531,7 @@ public class DockerConnector {
      */
     public Exec createExec(final CreateExecParams params) throws IOException {
         final ExecConfig execConfig = new ExecConfig().withCmd(params.getCmd())
+                                                      .withUser(params.getUser())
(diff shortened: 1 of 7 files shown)


bmicklea:
Is this true for Che as well?  Or only Codenvy?

mmorhun:
@bmicklea for both Che and Codenvy

bmicklea:
This will require docs changes that will have to be approved before I can approve this PR: https://github.com/eclipse/che-docs/pull/232

bmicklea:
@slemeur do you think we need to mention this in the release notes?

slemeur:
@bmicklea : Yes I think we can have a section for the enhancement of the snapshotting policy. (based on the docs)

bmicklea (approved):

garagatyi (approved):

skabashnyuk (approved):

PR #5176, opened 2017-05-23
Wait and check LS process

vparfonov:
Signed-off-by: Vitalii Parfonov <vparfonov@codenvy.com>

### What does this PR do?
_Temporary solution until 5.12 or 5.13 (version)_

In future need to provide some service that can watch for LS process and notify user in case in some reason it  stopped.
For now we just check it once start it before connect to it.  Ask with delay in 5 seconds this delay chose empirical in normal state should be enough for start or fail process. If after 5 seconds process not alive we notify client about problem.


tolusha:
What does 5.12 mean?

tolusha (approved):

evidolob (approved):

slemeur:
What version are you targeting for this "quick-fix" ? 
Do you want this to be release as part of a bug fix from 5.11.x ?

dkulieshov (approved):

vparfonov:
@slemeur yes, it will be part of bugfix

PR #5210, opened 2017-05-26
Use Authorization header instead of session-access-key cookie

voievodin:
CSRF protection https://github.com/codenvy/codenvy/pull/2213 affected CLI.
CLI or any other client different from web-browser SHOULD NOT use cookies for authorization but SHOULD use either _Authorization_ header or query parameter.


riuvshin (approved):

benoitf (approved):

riuvshin:
fix tested, works like a charm!

PR #5226, opened 2017-05-29
added subscription manager

dkulieshov:

### What does this PR do?
Adds remote subscription manager

#### Changelog
added remote subscription manager

#### Release Notes
added remote subscription manager

### DIFF
--- a/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/notification/RemoteSubscriptionManager.java
+++ b/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/notification/RemoteSubscriptionManager.java
@@ -1,0 +1,93 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2017 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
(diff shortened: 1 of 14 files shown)


azatsarynnyy (approved):

TylerJewell:
This description is not clear.

1. Needs a specification to explain what this capabiltiy is.
2. There is a huge amount of files changed - also needs clear documentation in the release notes on which clints would use this subscription manager and maybe an example.

vparfonov:
@TylerJewell this PR for `spi` branch not for `master` so any release notes don't need. 
Branch `spi` still in development

TylerJewell:
Ok, if we play strictly by the rules - then yes release notes are not needed. However, this is a large chunk of code. Take 15 minutes to add some documentation to the release notes in the description of this issue and then it's captured forever in the records.  Otherwise, when the full SPI branch is merged, I doubt there will be notes on what this particular subsystem does.  If you take 15 minutes to add some documentation each time there is a PR like this, then it does not turn into a 15 days effort when the whole branch merges.

dkulieshov:
Hi,

This pull request is yet in progress and the technical specification is being discussed so development is not over and technical specification is not finalized in any sense. The description and/or documentation will be added if needed as it is required by the workflow when the development is finished.

I feel flattered to know that you consider the number of changes is huge, but formally many of those are license changes, so that is not really worth attention.

gazarenkov (approved):

Issue #3629, opened 2017-01-06
Make loading default stacks more obvious


ashumilova:
Now the stacks.json rewrites the existing stacks each time the user restarts che server. In this case - user makes his changes to existing stacks, restarts che and looses the changes. It should be better and more clear for user to load stacks.json only on the first che start  - to fill database with predefined values.

skabashnyuk:
We are proposing following concept: 
 - remove stacks.json and template.json
 - Move stacks and template initialization to database script

 That makes two different strategies for two different group of users.

1. Codenvy and Che users if they want to customize their installation they need to go to dashboard and remove or add or modify any stack they want.
   If they want to copy some stack from one installation to another they just need to copy\paste it on dashboard.

2. Users who is making new assembly of Codenvy and Che they need to extend database initialization script. Remove everything that is not needed
add new stacks and templates. Can be simplified by creating stack from dashboard and some tooling with database.

TylerJewell:
So there are two types of issues:
1. When we have a custom assembly, and admins initialized their database with a default set of stacks / templates, and then users update it within the dashboard, those user changes are overwritten.  We should avoid this.

2. We would need some details on the custom assembly scenario on what it means to extend the database initialization script.  This sounds complicated, even though it is just a text file.  So, please keep in mind that someone who runs an archetype to get a custom assembly will need a simple way to find where stacks / templates are loaded by database, edit it, and then debug it if the format they type is wrong.

skabashnyuk:
>? those user changes are overwritten. We should avoid this.
 
That is not going to happen when move stacks.json and template.json to database initialization script

>We would need some details on the custom assembly scenario
Yes sure.  We can simplify it with some tooling like
   - cli export stacks > stack.sql
   - export to file in UD

garagatyi:
It was very confusing when I found that if user started using Che from 5.x.0 and updated his Che after each release 5.x+1.0, 5.x+2.0, ..., 5.x+n.0 he won't get updates in old stacks from these releases.
So looking at [page with changes in stacks](https://github.com/eclipse/che/commits/master/ide/che-core-ide-stacks/src/main/resources/stacks.json) note that changes made in this file is never delevered to a user that started using Che before that release. Except addition of new stacks, they are added into Che.

[...]

PR #5234, opened 2017-05-30
Change public stacks, recipes loading policy

akorneta:
### What does this PR do?
Changes logic of stacks, recipes loading from _always overwrite_ to _load once_ that means following:
_public stacks_ would be loaded from `stacks.json` only once after the first initialization of the database schema, for users with an existing database state, stacks will not be loaded. In order to modify, delete, add another `public stacks`, che user or codenvy admin should use dashboard functionality or stack API.
Added initialization properties to `DBInitializer` which needed to keep information about db state before init start. Also added `CHE_PREDEFINED_STACKS_RELOAD__ON__START` configuration property for backward compatibility, those who prefer previous strategy may set value `true` for loading stacks on server start. 

Note that there will be no way to configure custom stacks location(with `stacks.json` and images).

https://github.com/eclipse/che/issues/3629
is needed for [CODENVY-2208](https://github.com/codenvy/codenvy/issues/2208)

#### Changelog
Updated default stack loading policy.

#### Docs PR

https://github.com/eclipse/che-docs/pull/246

#### Release Notes
Default stacks loading policy was changed, now stacks loaded once on database initialization.
Those who use previous(load per server start) strategy may continue using it by setting `CHE_PREDEFINED_STACKS_RELOAD__ON__START` property value to `true`


wsmaster/che-core-api-system/src/main/java/org/eclipse/che/api/system/server/ServiceTermination.java
```diff
@@ -20,7 +20,7 @@
  *
  * @author Yevhenii Voevodin
  */
-interface ServiceTermination {
+public interface ServiceTermination {
```
mshaposhnik: Нащо?

akorneta: Strategy for service termination changed and now certain service could provide the implementation of its termination.

wsmaster/che-core-api-system/src/test/java/org/eclipse/che/spi/system/server/tck/SystemPropertyDaoTest.java
```diff
@@ -0,0 +1,115 @@
...
+import static org.testng.Assert.assertEquals;
+
+/**
+ * Tests {@link SystemPropertyDao} contract.
+ *
+ * @author Anton Korneta
+ */
+@Listeners(TckListener.class)
+@Test(suiteName = SystemPropertyDaoTest.SUITE_NAME)
+public class SystemPropertyDaoTest {
```
sleshchenko: Please add execution of this test in `integration-tests/postgresql-tck` module

AndrienkoAleksandr (approved):

skabashnyuk (approved):

wsmaster/che-core-api-machine/src/main/java/org/eclipse/che/api/machine/server/recipe/RecipeLoader.java
```diff
@@ -106,16 +125,19 @@ public void start() throws ServerException {
...
-    private InputStream getResource(String resource) throws IOException {
-        File resourceFile = new File(resource);
-        if (resourceFile.exists() && !resourceFile.isFile()) {
-            throw new IOException(String.format("%s is not a file. ", resourceFile.getAbsolutePath()));
-        }
-        InputStream is = resourceFile.exists() ? new FileInputStream(resourceFile)
-                                               : Thread.currentThread().getContextClassLoader().getResourceAsStream(resource);
-        if (is == null) {
-            throw new IOException(String.format("Not found resource: %s", resource));
+    private Reader getResourceReader(String resource) throws IOException {
```
skabashnyuk: ```IoUtil.getResource``` ?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/stack/StackLoader.java
```diff
@@ -96,22 +106,36 @@ private void loadStack(StackImpl stack) {
...
+    /**
+     * Searches for resource by given path.
+     *
+     * @param resource
+     *         path to resource
+     * @return resource InputStream
+     * @throws IOException
+     *         when specified resource was not found
+     */
+    private InputStream getResource(String resource) throws IOException {
```
skabashnyuk: ```IoUtil.getResource``` ?

assembly/assembly-wsmaster-war/pom.xml
```diff
@@ -293,25 +297,38 @@
...
                     <execution>
-                        <id>analyze</id>
+                        <id>add-stack</id>
+                        <phase>prepare-package</phase>
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
                         <configuration>
-                            <skip>true</skip>
+                            <target>
```
skabashnyuk: can you leave it in jar file?

core/che-core-db/src/main/java/org/eclipse/che/core/db/DBInitializer.java
```diff
@@ -46,10 +49,14 @@
 @Singleton
 public class DBInitializer {
```
skabashnyuk, line 51: describe purpose of this property

dockerfiles/che/entrypoint.sh
```diff
@@ -284,10 +282,8 @@ init() {
   mkdir -p ${CHE_DATA}/lib  
   cp -rf ${CHE_HOME}/lib/* "${CHE_DATA}"/lib
 
-  if [[ ! -f "${CHE_DATA}"/stacks/stacks.json ]];then
-    rm -rf "${CHE_DATA}"/stacks/*
-    mkdir -p "${CHE_DATA}"/stacks
-    cp -rf "${CHE_HOME}"/stacks/* "${CHE_DATA}"/stacks
+  if [[ -d "${CHE_DATA}"/stacks ]];then
+    rm -rf "${CHE_DATA}"/stacks
```
skabashnyuk: is this just a cleanup of old data?

sleshchenko: yes

assembly/assembly-wsmaster-war/pom.xml
sleshchenko: I have to edit original file by adding che-in-che to json file. Do you know how to implement it with leaving it in jar file?

dockerfiles/che/entrypoint.sh
benoitf: maybe add a comment that it's for upgrade, we clean "no longer in use" data

sleshchenko (approved):

voievodin (approved):
the code is ok

slemeur:
Good work @akorneta.

There is a discussion to open up on the user experience impact which this PR will introduce on the update policy. 
Let's take the following situation:
- User installs Che 5.10 and get a list of default stacks provided by Che
- User removes / updates / adds stacks on his installation
[Che 5.12 released with a new stacks for new supported technologies]
- User updates to Che 5.12
- User will not have the new stacks provided with Che 5.12

The problem is that the only way for the user to get the new stacks, will be to **manually** update the stacks, by either using the rest API or by going into the dashboard and manually upload the new stacks.json file. 

This is not a proper experience for the end-user, which is expecting that when he is updating his Che, he gets everything new from Che - including the new stacks for newly supported technologies.

There are also other situations such as OSIO, where the stacks needs to be automatically updated by the server as well. 

So I'm -1 on the PR due to the update policy which is satisfying from a user experience perspective.

gorkem:
I agree, I think there needs to be a mechanism that can propagate a new stack to all the users of a master.

ashumilova:
what will be the strategy for modified or removed stacks?

slemeur:
**When a system stack has been removed**
When the system administrator remove a system stack, we do not delete it, we would just hide it. This way, when we update the list of stacks, we know that we will not display it again to the user. 
It would probably be better wording to use "disable" instead of "remove". 

**When a system stack has been modified**
Updating an edited system stack could be tricky to manage. I don't think it is necessary to go into complicated update scenario for the time being.
My proposal would be the following:
- Once a system stack has been edited, we do not update it.
- In docs, and dashboard, when user is editing a system stack, we inform that he will not get the updates from upstream and recommend him to duplicate the stack if he wants to keep getting updates from upstream.

*Note:*
We probably need to keep a timestamp/version for stacks in the metadatas.

gazarenkov:
@gorkem 
"I think there needs to be a mechanism that can propagate a new stack to all the users of a master."
Nothing changed in terms of new stacks accessibility.
The problem we solve is data mess which we have providing new version of Che.
Want to separate data from new version functionality.

gazarenkov (approved):

benoitf:
@gazarenkov I would say that PR should solve the whole usecases and not only one of the issue as this new behaviour is not backward compatible on what users had previously.

gazarenkov:
@benoitf what kind of backward compatibility problems you see?

benoitf:
@gazarenkov people were getting at each update the new stacks inserted in new Eclipse Che releases. Now they won't.

gazarenkov:
@benoitf yes, so for ones who care we can release new stacks with json (and|or make them available online) and provide import/export facility inside UD. IMO it is very good replacement yet clear behavior, letting users to choose what they need other than forcing everyone to use ours preferences.

l0rd:
@gazarenkov I like the idea to separate code from data. We may need to release stacks.json as a separate artifact then. And make it possible to hot reload of stacks.json (i.e. without restarting Che). Are you planning to do this as well? Or maybe it's already implemented?

Talking about this PR, our issue is that in https://openshift.io (osio) Che UD is not available. That means that there is no way for a user to update a stack. Some UD features will eventually become part of the osio codebase section but I believe most of a stack details will remain "hidden" for users: that was a UX choice. 

So the solution for osio will be to add an extra step when deploying Che to check for stacks.json update and replacing the existing one. That would be easier if stacks.json is a proper artifact with versioned releases.

TylerJewell:
@gazarenkov - I am wondering if we are not violating some simplicity principles. My biggest concern is that each time I revisit this particular issue, I have to re-learn the concepts each time.  They are not sticking permanently because there are so many use cases to consider as it relates to stack management, and then layering in the concept of what is in stacks.json vs. the embedded database is problematic.

Is there any scenario where we eliminate the stacks.json file concept?  And instead within the internal database we keep a repository of stacks in different lists:
1. For a version of Che, we have a list of the official certified stacks for that versoin.
2. We maintain this list for each previous version / shipped version
3. There is the "system stacks", which is a list of the system + user stacks that should be presented in a dashboard.
4. In the "system stacks", an admin can choose whether stacks are a) not upgradable, b) upgradeable, or c) lists all versions.  If "not-upgradeable" then the UD would only display the version of the stack that was previously pinned.  If "upgradeable", then the UD displays the latest available stack.  If "lists all versions", then we display a stack and then have a drop-down widget to let the end user choose from all known versions of that stack.
5. We then add in some capability into the system that lets admins add or remove new stacks.  They can also "pin" existing stacks to be "non-upgradeable", meaning that even if there is a newer version loaded in an upgrade, the system still presents only the version that the admin pinned.
6. If an admin adds a "new" stack that is similar to our stacks, this stack would get a new UUID and its versioning scheme would be on a different path than the ones we ship (all would have same stack type UUID, just differences from versioning)

So what would this do for us in the end:
1. There would only be a single data location for storage of critical information.
2. We would then version our stacks on each release.
3. We have the concept of a UUID for a stack-type, so that we can track the evolution over time.
4. There is a clear path for enterprises to control which stacks they want and how they are used.

skabashnyuk:
I think ```statck upgrading by default``` is anti-pattern and we should try to avoid it. Here are a couple of examples:
1. A user may already create a custom stack similar name what we want to add in next release. And this will produce unavoidable conflict during upgrade

2. A user may don't want our new stack. They are ok with existing stack because the process of development that they use are connected to particular stack and not necessarily compatible with our update.

So I think we should have pattern ```do not upgrade any stack by default``` and require explicit user interaction for any stack upgrade.  AFAIK similar situation with Wordpress. Upgrade of CMS doesn't mean an upgrade of all plugins, usually upgrade will not add new plugins, to upgrade plugin user have to do some actions.

TylerJewell:
I think the longer strategy I wrote is identical to what you wrote @skabashnyuk. The only difference is that I am suggesting we keep a table with every version of every stack.  The default behavior would be "no-upgrade". But why not also allow the additional features that I mentioned where end users could potentially see older and newer versions?

I think the only analogy issue I have with the Wordpress example is that WP is managing truly pluggable plug-ins that are supposed to execute on any version of the product. Don't we have scenarios where we have runtime minimums for certain versions of stacks?  ie, this new stack version requires docker 1.13 and will fail on Docker 1.12.

skabashnyuk:
>The only difference is that I am suggesting we keep a table with every version of every stack. The default behavior would be "no-upgrade". But why not also allow the additional features that I mentioned where end users could potentially see older and newer versions?

Too complicated. I assume that it could be couples of smart guys who wants to see that. But most scenarios should be way more simple.

1. I don't want your new stacks. Go away.
2. I do want to upgrade this one stack.
3. I don't mind. Upgrade them all.

And the way how I see how to do that. Is some kind notification window on UD that can compare the list of existed stacks with ```god blessed``` and provide an opportunity for a user to select between this 3 options.
wdyt?

TylerJewell:
If Che was continuing to be a product for individual developers, then that approach may be suitable. But I think given the direction of Che and the intention of OSIO, it's increasingly clear that stacks (and stack management) is an admin responsibility and not an end user responsibility.

So the administrator (person performing the upgrade) is responsible for setting the stack upgrade policy that everyone must follow.  Depending upon the policy chosen, end users may (or may not) have a choice.

So it seems that when upgrading, the admin must:
1. I don't want your new stacks, go away.
2. I want your new stacks, upgrade all workspaces wherever possible, upgrade all stacks.
3. I want to let the end user choose (in the UD or other client with UX to help guide them).

I think this would have to be managed for the admin either during the upgrade sequence with our Docker installer.  Or, perhaps,  after the upgrade is done and when admin logs in the first time. I would think the upgrade time would be the time where this must be managed.

gazarenkov:
@l0rd let's consider OSIO needs separately. UD uses REST API so it is something what can be easily implemented on your side but I am interested to understand more of what you mean by "most of a stack details will remain "hidden" for users". It may make us think about something else.

l0rd:
@gazarenkov currently in osio users can only choose a technology (e.g. vert.x, spring boot etc...). There is still no option for the user to update the version of the technology stack. That should change so I'm ok to consider osio separately here.

By the way considering stacks as external plugins is fascinating. With their own lifecycle and, why not, a marketplace where the community could contribute publishing their custom stacks.

gazarenkov:
So I would say we all are kinda on the same page:
Any decision about upgrading data (stacks in this case) should be taken consciously by authorized person.
This PR is pointed to this direction, which mean - "go" for merging it I suppose.

benoitf:
@gazarenkov I think it's missing the server side "update part" that client would have to call to manage Tyler's proposal:
- I don't want your new stacks, go away.
- I want your new stacks, upgrade all workspaces wherever possible, upgrade all stacks.
- I want to let the end user choose (in the UD or other client with UX to help guide them).

slemeur:
As this PR introduces regressions in term of end-user experience - it is "no go" to merge

gazarenkov:
What should we do is to sick the simple design principles, such as:
1/ Keep only one place for storing data which is used by the system. If we decide it is DB (which is a natural choice) - it is ONLY DB, i.e. no other data source should be exposed to end-user for reading/writing.
2/ If we want our DB to be pre-initialized for user convenience - we should do it normally once, during DB initialization (other are really special cases)
3/ If we want to have (batch) data import/export/merge (from JSON or other source) - it has to be clear for user what data is expected, no hidden merging.
4/ If we want data integrity and correct merging during batch merging - we need to add versioning.
5/ If we want to have data assess policy for multi-user system (such as data created by Admin vs data created by ordinary User) - it should be based on permissions and not on different data sources (DB and JSON for example, see number 1)

In the case of Stacks we apparently have system design problem (lack of system design I'd say) and as a consequence unexpected system behavior and data inconsistency. 

I consider this PR as a bug fix, we need to fix it now to not to do it later (since is more painful in time).
And I see no contradictions between this and what Tyler said.
Actually it means that we need to: 
(1) use one data source and make the data consistent 
(2) make sure permissions policy work (if data is added/configured by admin it may be visible for others)
(3) add UI for import/export

So, I see no reason of not to fix it.

slemeur:
I agree with almost everything @gazarenkov. You are just forgetting few elements and those elements are related to the experience provided to the end-user - not about the way the stacks are implemented.

Today, when a new stack is added to Che and when a user is upgrading his Che installation, he benefits from that new stack automatically - he doesn't need to do anything.
Stacks are used by community contributors as it is a simple way to get started on contributing on Che and **enrich** the product for all Che users (at least those who are upgrading). [This file](https://github.com/eclipse/che/blob/master/ide/che-core-ide-stacks/src/main/resources/stacks.json) which define the default stacks is probably one of those which received the highest number of PRs - so I think it's important that community contributors can still easily enrich the product and have their stacks enabled for users in an handy way. 

This PR, will remove update mecanism and nothing is there to handle it in an handy way. So the PR is introducing a regression - and that regression will need to be handled by the various clients. We can handle it with dashboard - but not all Che integration are using the dashboard. So the proposed solution by @TylerJewell is actually the best approach.

> So it seems that when upgrading, the admin must:
> - I don't want your new stacks, go away.
> - I want your new stacks, upgrade all workspaces wherever possible, upgrade all stacks.
> - I want to let the end user choose (in the UD or other client with UX to help guide them).

We need to handle those scenarios in order to provide an experience without regressions. This needs some server part in order to handle the "update" scenario which can be called by the various clients.

BTW - In the model, there is no relation between a workspace and the stack used to create it. Once the workspace is created, it is impossible to know which stack (and which version of the stack) was used to create it. It is also not possible to change the stack. 
So there will be no impact on upgrading workspaces (at least for now with the designed model).

gazarenkov:
@slemeur 
I am aware of changing stacks.json and the reason. We shortly discussed it with @l0rd , the reason why it is needed is that actually there the simplest possible way in order to upgrade single user OSIO installation but in fact it looks like workaround for workaround. I think we will assign someones from our JS guys (plugin team) to help with OSIO client side flow (they do not use UD) but in general it is not a problem for them to apply this PR.

BTW, as @ashumilova informed me, there is a Workspace attribute which pointed to stackId so strictly speaking there is a reference. 
>It is also not possible to change the stack.

I would not listed it as advantage, it is rather a workaround caused by not good design.

slemeur:
OK, right on the stackId ;)

From my understanding - it means that clients (which we know or not) will have to handle and maintain the stack update mecanism. This work will have to be "duplicated" by all clients where this could be "simplified" if the update mecanism and logic was handled as close as possible to the datas.

sunix:
To me what is important is the custom product stacks:
- would have their own stacks
- would customize existing stacks from upstream (for instance, just adding an agent but keep the other parameters)
- would remove/hide stacks from upstream

We need a mechanism that is flexible enough so that when stacks from upstream change, custom product stacks get updated automatically. Admin/user changes (from user dashboard) should use the same mechanism on top of it.

Just loading once the stacks is not OK imo

TylerJewell:
@sunix - it is currently a non-goal to support dynamic updates of stacks for an existing installation. That goal would require a different approach to stacks altogether, such as having a registered third-party marketplace for storing, updating, and advertising stacks.  This is scoped outside of Che right now, but is a bigger issue that will be solved with different projects.

sunix:
@TylerJewell I'm not talking about dynamic updates but extensible stacks mechanism.

gazarenkov:
Guys, as an architect of Che I am interested in fixing system design problems, I explained above. This is my main concern.
It was discussed, the fix and strategy to make it less impactful to existed state were proposed (explained above as well). 
I do respect all of your opinions and being a co-lead of the project appreciate that you think of improving the project but, please, let's not try to solve all of them here.
So, I'd propose - If there are no issues with the PR code itself, let's apply the fix, do what we need to do improving UX around and go forward discussing other proposals.
Thanks a lot.

benoitf:
AFAIK leader role is 
operate using open source rules of engagement: meritocracy, transparency, and open participation

https://www.eclipse.org/projects/dev_process/#4_6_Leaders

TylerJewell:
@gazarenkov - I would like to get this PR merged sooner than later as well. If this is only a bug fix, then we should move the discussion to another forum.

However, I broadened the conversation for admins so that I can understand the end-user impact of the proposed PR change.

I've re-read the thread above, and still not quite sure I grok the admin impact of the change.

The logic for me is the following:
a) If this change does not alter how an existing admin / user installs or upgrades their system (ie, does not lead to a docs change), then this is definitely a bug fix. Let's merge and open a new issue for the broader issue.

b) If this change requires an admin to alter the way that they manage an existing stacks.json or the upgrade behavior may be confusing (ie, only something that can be explained in docs), then this bug fix is going to alter fundamental behavior.  If that is true, then I think we would need to pause this particular PR and instead target the system design that you have advocated (single database with upgrade policies).  

I think the proposed system design is getting a lot of +1s, so it's only a question of whether we jump right to it now, or we merge this PR first, and then implement the new policy second.

I just want to minimize the confusion / impact with end user admin as we transition.

ghost:
@slemeur I understand your concerns but here's another use case.

I am a Che user, and I have modified my `stacks.json` so that only my two awesome stacks are displayed for me and my users. I do not want dozens of unnecessary stacks that litter Dashboard and confuse users. Btw, this is the case with all OEMs - all of them want just THEIR stacks, often one or two, so that users have no chances of failing or doing smth wrong.

With the policy that you propose, after an update (great! i have new binaries with a lot of new features and bug fixes), all of a sudden my dashboard is full of cool new stacks, which I unfortunately do not need. What should I do in this case? OK, I will remove these ready to go stacks from UD using UD widgets. But then my server reboots, and these stacks are loaded again?

By the way, right now in Che, users do not get any stack or template updates, since these files are mounted into Che container.

benoitf:
@eivantsov reading the thread, I think nobody was objecting on the fact that some people didn't want to keep their existing stack (no update)
It's just that changing the default policy being the opposite should imply that the previous policy can still be possible.
On a side note, introducing opposite default changes in software is usually introduced in major versions of a product (like 6.x) not in a feature release (5.x) or it means that the previous default policy is still available, people wanted to have 'never update' having to toggle a flag.

sunix:
From what I understand, this will only load once from stacks.json. In openshift.io we have our custom stacks.json integrated to our assembly. These are supported stacks. If there is a fix to do on stacks, we redeploy a new version of the che-server and this update the stacks.json. Our supported stacks.json is versionned with git and it is easier to maintain than going to the default user dashboard .... that we don't have and don't use anyway. If this is breaking the way we maintain our stacks.json, if new versions of stacks.json are not applied when upgrading, i strongly not for merging this PR.

slemeur:
@eivantsov :
**For the update of the stacks.json**
Nobody is saying that we wipeout all the stacks of the database when we upgrade and replace with the content from the `stacks.json`, instead we would have update policies and obviously at each Che upgrades the previously "removed/hidden" stacks by the admin should not reappear. 

**Introducing stacks that the user does not need**
User would still have the ability to update or not. And if he updates, he will get the new stacks. If he is not interested in all of them, the user will still be able to "remove/hide" them. 
In fact, it's better than many other pieces of the product. Because today, when we introduce a plugin, the plugin is available for every users and every workspace - even where it is not necessary (example: SVN plugin). We have the same problem with the project types.

**Reboot of Che server**
Why would we reload the stacks when we reboot Che server. I don't think we should do that.

**Custom Che for OEMs**
Yes agree - in custom Che people also provide only a few set of stacks. They will also be heavily impacted by the regression introduced by this PR and have the same problem if they need their users to update the stack.

**Another use case:LS agent**
When a new LS agent is introduced, it is available to all users. Users might not need it, but it will be there and not possible to remove it - user is able to not use it.
Certain LS agents are working only on certain stacks. Let's say, we add a new LS agent, which provide intellisense capabilities for a new language - which is major new set of features for the product. A user, might want to test that, but without update policies in place (and with the regression introduce by this PR), by default he'll have the agent, but he'll not have the proper stacks to make it working. I don't think that is logicial from a end-user perspective.

ghost:
> at each Che upgrades the previously "removed/hidden" stacks by the admin should not reappear

I am talking about new stacks.

> Why would we reload the stacks when we reboot Che server. I don't think we should do that.

This is how it works now and is intended to work in this PR.

gazarenkov:
@sunix please talk to @l0rd , we discussed OSIO needs already (I told above)

@TylerJewell what I am saying the way it works currently is wrong since the experience is dictated by bad system design. So to me it is definitely a problem and I see no reasons to not to apply the fix on it.
I do not think that will negatively impact on UX as soon as we describe the changes and provide import/export tool. I do believe the impact will be only positive. 

Please read comment of @eivantsov below, it is a real customer case and in fact this customer request finally made me think that there are no reason to wait.

TylerJewell (approved):

TylerJewell:
I am ok to merge.

benoitf:
AFAIK OEM customers have dedicated assemblies/repositories and stack.json is a custom one in their repository. So I'm not sure it's matching the usecase.

slemeur:
What's the cost of implementing an update mecanism from the server side for the stacks? What are the reasons to not implement them?

BTW - according to our policies, this PR can't be merged. It is:
- missing documentation 
- missing Release Note information

gazarenkov:
Here is the compromise we came to make this movement as smooth as possible:
- we add the optional property which will merging existed (in db) stacks with one from stacks.json. Proposed name could be something like "reloadStacksOnStart" (I let you make a correct name according to our policy).
- if this property present and has "true" value - every system start the content of stacks.json is merged to stacks database. 
- We implement the simplest case of merging which for example just replacing stacks DB with stacks.json content (remove and add) or something else but simple
- by default (no property) or =="false" we do not change stacks DB
- in the documentation/changes log we warn that such an automatic merging can be applied carefully by user's own risk (since it can be inconsistent with workspaces db) and removed in time.

sunix:
> - We implement the simplest case of merging which for example just replacing stacks DB with stacks.json content (remove and add) or something else but simple

This wasn't part of the discussion. I would say we implement the best merge we have.

OK for the others as discussed.

Also as discussed,  the deprecated policy doesn't get removed after 2 sprints as usual deprecation but only when we have a replacement solution discussed and working for our use cases.

l0rd:
@gazarenkov 👍

gazarenkov:
@sunix 
If you want to provide something more sophisticated than "simple" merging - do it, this complexity just should not block this PR merging.
Yes, "the best merge we *have* " is good so.

akorneta:
Updated doc + added property `CHE_PREDEFINED_STACKS_RELOAD__ON__START` for backward compatibility
@slemeur please take a look

slemeur:
In the issue description, I don't see the section for documentation pointing to any docs PR.

I think those documentations are affected:
- https://github.com/eclipse/che-docs/blob/master/src/main/_docs/devops/devops-runtime-stacks.md#adding-stacks-to-the-che-default-assembly
- https://github.com/eclipse/che-docs/blob/master/src/main/_docs/setup/setup-managing.md#upgrading
- https://github.com/eclipse/che-docs/blob/master/src/main/_docs/setup/setup-configuration.md

gazarenkov (approved):

ghost:
@slemeur @akorneta and myself will take care of that

dockerfiles/che/entrypoint.sh
```diff
@@ -262,8 +262,6 @@ init() {
   sed -i "/che.workspace.storage=/c\che.workspace.storage=/data/workspaces" $CHE_LOCAL_CONF_DIR/che.properties
   sed -i "/che.database=/c\che.database=/data/storage" $CHE_LOCAL_CONF_DIR/che.properties
   sed -i "/che.template.storage=/c\che.template.storage=/data/templates" $CHE_LOCAL_CONF_DIR/che.properties
-  sed -i "/che.stacks.storage=/c\che.stacks.storage=/data/stacks/stacks.json" $CHE_LOCAL_CONF_DIR/che.properties
-  sed -i "/che.stacks.images=/c\che.stacks.images=/data/stacks/images" $CHE_LOCAL_CONF_DIR/che.properties
```
riuvshin, line 266: are you sure we need remove images?

akorneta: @riuvshin yes, we moved all stacks images [into](https://github.com/eclipse/che/tree/bed4280261d773887f8101e000ca103d983a5525/ide/che-core-ide-stacks/src/main/resources/stacks-images)

riuvshin: this is not docker images list? like this one https://github.com/eclipse/che/blob/master/dockerfiles/cli/version/5.12.0/images-stacks

akorneta: no, its path to folder with stacks images

benoitf: yes it's not docker images.
But how is backward compliance for users that were using the stacks.json with custom images for their stacks. It seems now this is no longer possible.

Also the description of the pull request is not describing the path or the property name to define the path to the custom stacks.json to load

slemeur:
Merge must happen when docs are ready as well.

slemeur:
@akorneta + @gazarenkov + @TylerJewell : Please follow our merging policies. Here the PR has been merged without docs ready and without release note information. The release note details in the description of the issue are not following the requirements mentioned by @gazarenkov.

In order to get that solved, I'm proposing the following release note details:

This version introduces a change in the way [runtime stacks](https://www.eclipse.org/che/docs/devops/runtime-stacks/index.html) are handled. The runtime stacks are used to defined the runtime configuration of workspaces. Che provides a default set of stacks for various technologies and framework that are ready to be used for your workspaces. You can also edit stacks and add your own by going into the "Stacks" administration page.

In order to simplify the management of the stacks, we changed the loading policy and behaviour which impact the upgrade of the list of stacks. The list is initialized when you first install Che and then your list of stacks is kept identical when you upgrade to a newer version of Che. 

For the users who are upgrading from 5.x to 5.13.x, your list of stacks will be kept consistent and not re-initialized. 

It is possible to re-initialized the list of stacks you have. Che is oftenly updating and adding new pre-defined stacks which are contributed by the Che community. If you want to benefits from those stacks and get them integrated when upgrading to a newer version of Che, you can configure the `CHE_PREDEFINED_STACKS_RELOAD__ON__START`  in `che.ev` file to `true`. 
Special warning, this must be done cautiously because the database of stacks will be wipeout (including the ones you added) and will be reinitialized with the `stacks.json`.

If you want to upgrade to a newer version of Che and keep the stacks you added and benefit from the new stacks added you need to manually edit the stacks.json and follow the instructions on this documentation [DOC LINK NEEDED]. 

Please @gazarenkov review.

gazarenkov:
@slemeur @akorneta @eivantsov 
please make a fixes when necessary

1/ For better decoupling data from released code our default policy is population stacks database with predefined set only once - during brand new installation of product (i.e. on empty database)
2/ If you need stacks database to be changed every restart you can configure the CHE_PREDEFINED_STACKS_RELOAD__ON__START=true in che.ev file.
NOTE: such a merging is a workaround for the behavioral compatibility, it does not guarantee correct references with existed workspaces and will be removed in the future, so use it carefully by you own risk.
3/ For adding new stacks from JSON file please use UD Stack Export functionality

ghost:
@slemeur let's continue this dialogue in https://github.com/eclipse/che-docs/pull/246

There are a number of things that you misunderstand, including those about how database is updated (I'll post a comments in the docs PR)

slemeur:
@gazarenkov : what do you mean by:
> 3/ For adding new stacks from JSON file please use UD Stack Export functionality

To add stack, we have to use export functionality? What is that?

@eivantsov What you are saying is not aligned with what @gazarenkov described here:
> We implement the simplest case of merging which for example just replacing stacks DB with stacks.json content (remove and add) or something else but simple"

And also with:

> 2/ If you need stacks database to be changed every restart you can configure the CHE_PREDEFINED_STACKS_RELOAD__ON__START=true in che.ev file.

So, this is why documentation and release note information must be approved before merging those kind of PR.

gazarenkov:
@slemeur I meant stack IMPORT functionality (it was a typo, sorry) which we discussed several times. 

https://github.com/eclipse/che-docs/pull/246 - this is the issue to discuss the docs, please.

slemeur:
@gazarenkov : Where is that feature? I have not seen any PR for that on dashboard. Apparently, that capability was needed in order to merge that PR.

gazarenkov:
https://github.com/codenvy/codenvy/issues/2252

PR #4732, opened 2017-04-06
CHE-102 - Idle detection of che-server and workspaces

snjeza:
### What does this PR do?

Detect when a workspace and che-server are idle

### What issues does this PR fix or reference?

https://issues.jboss.org/browse/CHE-102

PR #5278, opened 2017-06-04
CHE-102 - Idle detection of che-server and workspaces

snjeza:
https://github.com/eclipse/che/pull/4732 has been removed from the openshift-connector-rebased branch. I suppose we will port the code from the codenvy project. However, part of the PR related to idling the che server is openshift specific and doesn't exist in the codenvy project.
Since the PR is being used in OpenShiftWorkspaceFilesCleaner, the branch contains compile errors. This PR fixes them.

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
@@ -340,6 +340,12 @@ che.openshift.workspace.memory.request=NULL
 #
 # - Maximum amount of memory a workspace container can use e.g. 1.3Gi
 che.openshift.workspace.memory.override=NULL
+
+# The Openshift will idle the server if no workspace is run for
+# this length of time.
+che.openshift.server.inactive.stop.timeout.ms=1800000
+
+#
 #
 # Be aware that setting che.openshift.workspace.memory.override
 # will override Che memory limits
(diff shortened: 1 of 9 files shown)


l0rd:
Thank you @snjeza. @sunix please have a look at that.

@snjeza yes we want to remove idling from branch `openshift-connector-rebased`. And then open a second pull request (you should work on that one) to add idling capabilities the codenvy way. This doesn't mean that we are going to throw away all the work done for idling in `openshift-connector` branch but we are going to adapt it to extend the already existing and well tested idling in codenvy.

sunix:
Yes thanks I'm going to review. Actually #4732  has been moved to https://github.com/eclipse/che/commits/openshift-connector-wip, but I didn't noticed the compilation error :/ thanks for reporting that :+1:

snjeza:
@l0rd Idling in the openshift-connector branch contains two features:

1) stops a workspace if it hasn't been active for some time. This isn't openshift specific. There is a similar feature in the codenvy project.
The feature will stop, but won't idle an inactive workspace.
2) if no workspace is run for some time, the che server (the openshift connector) will be idled. This is an openshift specific feature and, in my opinion, there isn't a similar feature in codenvy or in a pure docker environment. This feature is independent on feature #1. The che server will be also idled if all its workspaces are stopped manually. We can apply this patch independently of #1. This feature can't be used in the standard (docker) che.

@skabashnyuk @benoitf @garagatyi @riuvshin I am not sure if I am missing something. Is the codenvy activity feature related only to workspaces or there is a feature that will stop or idle the che server?

garagatyi:
@snjeza You understand correctly - Codenvy implements workspace idling feature, not workspace master idling.

l0rd:
> The feature will stop, but won't idle an inactive workspace.

That should not be a problem.

> We can apply this patch independently of #1. 

Make sense.

PR #5285, opened 2017-06-07
CHE-5169: Add Git configuration agent

### LINKED ISSUE #4969: Git preferences for command line
Setting git preferences such as ssh keys should also configure the command line git. Users should be able to do a commit and push identically without doing any additional configuration from UI an terminal.

### LINKED ISSUE #5169: Sync user ssh keys with git in a dev machine
Che uses JGit, and there's Git installed in a workspace. These two gits do not share user SSH keys that are stored in user prefs (`/api/preferences`). It becomes quite confusing for users who, having uploaded ssh keys using UI Git, expect these keys to be available for a console git in a terminal.

Proposed solution:

1. When a workspace starts get user keys from `/api/preferences` and inject them in a workspace
2. Let a user choose if or not it should happen on workspace start since this might be a security issue - when a workspace is shared with another user, such a user has access to ssh keys. It may be a check box in user dashboard for example.

vinokurig:
### What does this PR do?
Added Git configuration agent that fetches SSH keys and Git username and email from CHE user preferences, and injects to console Git preferences

fixes https://github.com/eclipse/che/issues/5169
fixes https://github.com/eclipse/che/issues/4969

#### Changelog
Added a Git configuration agent that fetches Git SSH keys, username / email from Che preferences, and injects them into the console Git preferences.

#### Release Notes
N/A

#### Docs PR
N/A


gorkem:
@ibuziuk Take a look at this PR and can you comment if this will work with the GH token work that you have just done for OSIO?

tolusha:
@vinokurig 
Git agent name misused. Git agent means that git will be installed and ready to use.
Your PR brings new capability.

vinokurig:
@tolusha How about 'Git injector agent'?

bmicklea:
Updated title and description to clarify the agent's purpose. @vinokurig - does this inject preferences  into the Git console only?  In other words is it ensuring that the preferences match in console and  menu?

ibuziuk:
@gorkem No, it is not related to setting access token during workspace creation (moving @sunix PRs https://github.com/eclipse/che/pull/4444 https://github.com/eclipse/che/pull/4438 to rh-che) This is more related to what was raised up on the mailing list recently by Pete / Burr - not possible to push via terminal OOTB

slemeur:
@vinokurig : Could you explain why you built this capability by using an agent?

ghost:
@slemeur I can explain - We need to make sure that this feature works in any environment. We had an idea to embed it into our base images and start in an entrypoint. However, it won't work with non Eclipse certified images.

@bmicklea what this agent does is that when git performs ssh operations, it gets keys and git prefs (committer name and email) from user prefs, if any.

tolusha:
>  How about 'Git injector agent'?

I don't like either

vinokurig:
@tolusha may be 'Git synchroniser agent' ?

ghost:
Git sync agent is cleaner I think.

bmicklea:
But it's inaccurate isn't it? If I change a git setting in the terminal it doesn't sync it back to the Che preferences panel does it?

vinokurig:
@bmicklea It doesn't change back CHE preferences, may be Git fetcher agent ?

bmicklea:
I think Git Config agent is clearest since it's configuring Git preferences.

agents/git/src/main/resources/org.eclipse.che.git.script.sh
```diff
@@ -0,0 +1,58 @@
...
+chmod +x ${SCRIPT_FILE}
+
+user_name="$(${request} "$api_url/preferences$(if [ -n "$token" ]; then echo "?token=$token"; fi)" | grep -Po '"git.committer.name":.*?[^\\]",' | sed -e "s/\"git.committer.name\":\"//" | sed -e "s/\",//")"
+user_email="$(${request} "$api_url/preferences$(if [ -n "$token" ]; then echo "?token=$token"; fi)" | grep -Po '"git.committer.email":.*?[^\\]",' | sed -e "s/\"git.committer.email\":\"//" | sed -e "s/\",//")"
+git config --global user.name \""$user_name"\"
+git config --global user.email \""$user_email"\"
+
+if [ -z "$(cat /home/user/.bashrc | grep GIT_SSH)" ]
+then
+    printf '\n export GIT_SSH='"$SCRIPT_FILE" >> /home/user/.bashrc
```
ghost: Shouldn't it be `~` since `/home/user` is home directory only in Eclipse images?

vinokurig:
@bmicklea The problem is that the agent doesn't fetch only configs, it also fetches SSH keys

bmicklea:
if it's really just authentication then what about the git auth agent?

vinokurig:
Not just authentication, SSH keys injection + config injection

agents/git/src/main/resources/org.eclipse.che.git.script.sh
vinokurig: reworked to relative path

vinokurig:
I think `Git sync agent` is the most suitable name for this agent @bmicklea @tolusha WDYT?

mmorhun:
@vinokurig what about `Git credentials agent`? WDYT?

bmicklea:
I'm +1 for `git credentials agent`. I don't like `git sync agent` because sync implies something that is bi-directional and responsive but this just goes in one direction.

vinokurig:
Renamed to `Git Credentials agent` @bmicklea @tolusha @skabashnyuk @eivantsov please review

ghost (approved):

agents/git/pom.xml
```diff
@@ -0,0 +1,33 @@
...
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>che-agents-parent</artifactId>
+        <groupId>org.eclipse.che</groupId>
+        <version>5.12.0-SNAPSHOT</version>
+    </parent>
+    <artifactId>git-credentials-agent</artifactId>
+    <name>Git Agent</name>
```
tolusha: Fix name accordingly

agents/git/src/main/resources/org.eclipse.che.git.json
```diff
@@ -0,0 +1,5 @@
+{
+  "id": "org.eclipse.che.git",
```
tolusha: `"id": "org.eclipse.che.git-credentials",`

agents/pom.xml
```diff
@@ -32,6 +32,7 @@
         <module>che-core-api-agent-shared</module>
         <module>che-core-api-agent</module>
         <module>ls-json</module>
+        <module>git</module>
```
tolusha: git-credentials

tolusha (approved):

agents/git/pom.xml
vinokurig: done

agents/git/src/main/resources/org.eclipse.che.git.json
vinokurig: done

agents/pom.xml
vinokurig: done

bmicklea (approved):

agents/git-credentials/src/main/resources/org.eclipse.che.git.script.sh
```diff
@@ -0,0 +1,58 @@
...
+echo '    rm "$key_file"' >> ${SCRIPT_FILE}
+echo 'else' >> ${SCRIPT_FILE}
+echo '    ssh "$@"' >> ${SCRIPT_FILE}
+echo 'fi' >> ${SCRIPT_FILE}
+
+chmod +x ${SCRIPT_FILE}
+
+user_name="$(${request} "$api_url/preferences$(if [ -n "$token" ]; then echo "?token=$token"; fi)" | grep -Po '"git.committer.name":.*?[^\\]",' | sed -e "s/\"git.committer.name\":\"//" | sed -e "s/\",//")"
+user_email="$(${request} "$api_url/preferences$(if [ -n "$token" ]; then echo "?token=$token"; fi)" | grep -Po '"git.committer.email":.*?[^\\]",' | sed -e "s/\"git.committer.email\":\"//" | sed -e "s/\",//")"
+git config --global user.name \""$user_name"\"
```
skabashnyuk, line 52: that means that any commit from console git will be made from the name of workspace owner. Not sure it's good collaboration experience.

ghost: Right, but it was discussed before, I mean security implications and commit authorship.

sunix:
how would this agent behave with multitenant ? (multiple users accessing to the same workspace running the terminal)
It would be bad if I give access to someone to my workspace and he got access to my git credentials

tolusha:
It is the same if you allowed someone to use your laptop. It is bad (
It is supposed that user injects this agent thoughtfully.

sunix:
Sorry, No it is not the same, che is a cloud workspace. Sharing the workspace is not bad but something users expect to have. When you share your laptop to someone else, you don't give access to your user account, nor your Chome saved passwords. To me, all these git users related injection shouldn't go through a shared agent. This is a major security issue and should be reverted IMO. Moreover generated ssh keys are not protected with a passphrase.

gorkem:
I agree with @sunix . We should at least have password protection enabled or remove the SSH key on shared workspaces.

Issue #5276, opened 2017-06-02
Broken links in CUSTOMIZING.md


vnea:
Hello,

All links in [CUSTOMIZING.md](https://github.com/eclipse/che/blob/master/CUSTOMIZING.md) are broken.

slemeur:
Thanks @vnea. 
I'll take that.

Anne369:
@slemeur I also face the problem, could you fix it asap

slemeur:
Hi @Anne369. That issue is prioritized. The page needs to be refactored and simplified since we introduced archetypes and refactored the structure of the docs - but the docs are existing.

Have you explored them? 
https://www.eclipse.org/che/docs/

More particularly, sections regarding extending Che: https://www.eclipse.org/che/docs/assemblies/intro/index.html

You can also find documentation about development workflow here:
https://github.com/eclipse/che/wiki/Development-Workflow

qi-min:
@slemeur Besides the instruction in https://www.eclipse.org/che/docs/, is there any other documents for Che developers. 
I want to extend editor to support new file type. I see examples about how to support code completion and highlight, but I do not see any sample or API about formatting and syntax validation on Editor part https://www.eclipse.org/che/docs/assemblies/sdk-code-editors/index.html

slemeur:
Hi @m7luffy. 
There are various approaches for that:
- You can create a dedicated Language Server which will provide intellisense capabilities: https://www.eclipse.org/che/docs/assemblies/sdk-language-server-protocol/index.html
- Otherwise, you should explorer the various Editor API and apply the same principles you did for code completion and highlighting: https://github.com/eclipse/che/tree/master/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor . Here is how the java formatter is done for example: https://github.com/eclipse/che/blob/master/plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/editor/JavaFormatter.java

PR #5328, opened 2017-06-10
Update customizing.md

slemeur:
### What does this PR do?
Update customzing.md page with up-to-date content and links.
Keep indexed style format.

https://github.com/eclipse/che/issues/5276


TylerJewell (approved):
I modified some of the language to clean things up.  I think this is ok to merge, but I also think we could simplify the extension section.  Now we have three separate sections - extensions, IDE, server-side.  And I think we could just simplify this down to one section so that we do not duplicate certain entries and more closely follow the docs flow.

PR #5305, opened 2017-06-08
Recent changes required access to `/` which is impossible under Openshift

davidfestal:
### What does this PR do?

This PR fixes a recent change in the Che Docker image `entrypoint.sh` file, that prevented execution inside Openshift.

### What issues does this PR fix or reference?

The root cause of the problem was the write-access requirement to the filesystem root, in order to create / modify the `/conf` directory.  

Signed-off-by: David Festal <dfestal@redhat.com>

PR #5344, opened 2017-06-12
adapt che-server entrypoint.sh to environments without write permissions in '/'

riuvshin:
### What does this PR do?
proper fix for permission issue https://github.com/eclipse/che/pull/5305

after discussion with @davidfestal, we decided to check if we have write permissions to `/` we keep old logic which keep CLI and non-CLI use-cases not touched, but if there is no write permissions to `/` like in OS case we set **CHE_LOCAL_CONF_DIR** to **${CHE_DATA}/conf** `(/data/conf)`

#### Changelog
adapt che-server entrypoint.sh to environments without write permissions in '/'

### DIFF
--- a/dockerfiles/che/entrypoint.sh
+++ b/dockerfiles/che/entrypoint.sh
@@ -244,16 +244,25 @@ init() {
   fi
   ### Are we going to use the embedded che.properties or one provided by user?`
   ### CHE_LOCAL_CONF_DIR is internal Che variable that sets where to load
-  export CHE_LOCAL_CONF_DIR="${CHE_DATA_HOST}/conf"
-  if [ -f "${CHE_LOCAL_CONF_DIR}/che.properties" ]; then
-    echo "Found custom che.properties..."
-    if [ "$CHE_USER" != "root" ]; then
-      sudo chown -R ${CHE_USER} ${CHE_LOCAL_CONF_DIR}
+  # check if we have permissions to create /conf folder.
+  if [ -w / ]; then
+    export CHE_LOCAL_CONF_DIR="/conf"
+    if [ -f "/conf/che.properties" ]; then
+      echo "Found custom che.properties..."
+      if [ "$CHE_USER" != "root" ]; then
+        sudo chown -R ${CHE_USER} ${CHE_LOCAL_CONF_DIR}
+      fi
+    else
+      if [ ! -d /conf ]; then
+          mkdir -p /conf
+      fi
+      echo "Using embedded che.properties... Copying template to ${CHE_LOCAL_CONF_DIR}/che.properties"
+      cp -rf "${CHE_HOME}/conf/che.properties" ${CHE_LOCAL_CONF_DIR}/che.properties
     fi
   else
-    if [ ! -d ${CHE_LOCAL_CONF_DIR} ]; then
-        mkdir -p ${CHE_LOCAL_CONF_DIR}
-    fi
+    echo "WARN: parent dir is not writeable, CHE_LOCAL_CONF_DIR will be set to ${CHE_DATA}/conf"
+    export CHE_LOCAL_CONF_DIR="${CHE_DATA}/conf"
+    mkdir -p ${CHE_DATA}/conf
     echo "Using embedded che.properties... Copying template to ${CHE_LOCAL_CONF_DIR}/che.properties"
     cp -rf "${CHE_HOME}/conf/che.properties" ${CHE_LOCAL_CONF_DIR}/che.properties
   fi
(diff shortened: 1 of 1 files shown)


riuvshin:
@davidfestal can you please test this PR with OS environment to make sure it works as expected ? :)

benoitf (approved):

davidfestal:
@riuvshin sure. I'll test is asap

l0rd (approved):
LGTM

dockerfiles/che/entrypoint.sh
```diff
@@ -244,16 +244,25 @@ init() {
   fi
   ### Are we going to use the embedded che.properties or one provided by user?`
   ### CHE_LOCAL_CONF_DIR is internal Che variable that sets where to load
-  export CHE_LOCAL_CONF_DIR="${CHE_DATA_HOST}/conf"
-  if [ -f "${CHE_LOCAL_CONF_DIR}/che.properties" ]; then
-    echo "Found custom che.properties..."
-    if [ "$CHE_USER" != "root" ]; then
-      sudo chown -R ${CHE_USER} ${CHE_LOCAL_CONF_DIR}
+  # check if we have permissions to create /conf folder.
+  if [ -w / ]; then
```
sunix, line 248: maybe also test write access to /conf ?

dockerfiles/che/entrypoint.sh
```diff
@@ -244,16 +244,25 @@ init() {
...
+          mkdir -p /conf
+      fi
+      echo "Using embedded che.properties... Copying template to ${CHE_LOCAL_CONF_DIR}/che.properties"
+      cp -rf "${CHE_HOME}/conf/che.properties" ${CHE_LOCAL_CONF_DIR}/che.properties
     fi
   else
-    if [ ! -d ${CHE_LOCAL_CONF_DIR} ]; then
-        mkdir -p ${CHE_LOCAL_CONF_DIR}
-    fi
+    echo "WARN: parent dir is not writeable, CHE_LOCAL_CONF_DIR will be set to ${CHE_DATA}/conf"
```
davidfestal: It was `CHE_DATA_HOST` and not `CHE_DATA`

davidfestal (changes requested):

dockerfiles/che/entrypoint.sh
davidfestal: that was what was used before your inital change
in commit 4813a1a712abb1e0bb949b1c2dcdc7a7b206fd63
and this is the property that is used in the related Openshift deployment config.

dockerfiles/che/entrypoint.sh
riuvshin: added checks

dockerfiles/che/entrypoint.sh
riuvshin: that was a mistake, `CHE_DATA_HOST` is a path actual for user PC like `/home/roman/.che/data` but we need path actual for container `/data` 

as you see in https://github.com/eclipse/che/commit/4813a1a712abb1e0bb949b1c2dcdc7a7b206fd63 `CHE_DATA_HOST` is only in echo, actual `cp` command using `/data/conf` path.
In fact `CHE_DATA_HOST` is a path which we mount to `CHE_DATA` and inside of a container we use only `CHE_DATA` but for user actual path is `CHE_DATA_HOST` because it is a path on his host filesystem.

riuvshin:
PR updated, I also figured why @davidfestal had:

```
1) No implementation for java.lang.String annotated with @com.google.inject.name.Named(value=che.port) was bound.
  while locating java.lang.String annotated with @com.google.inject.name.Named(value=che.port)
    for the 5th parameter of org.eclipse.che.plugin.docker.machine.CustomServerEvaluationStrategy.<init>(CustomServerEvaluationStrategy.java:85)
  at org.eclipse.che.plugin.docker.machine.local.LocalDockerModule.configure(LocalDockerModule.java:62) (via modules: org.eclipse.che.api.deploy.WsMasterModule -> org.eclipse.che.plugin.docker.machine.local.LocalDockerModule)

1 error
    at com.google.inject.internal.Errors.throwCreationExceptionIfErrorsExist(Errors.java:470)
    at com.google.inject.internal.InternalInjectorCreator.initializeStatically(InternalInjectorCreator.java:155)
    at com.google.inject.internal.InternalInjectorCreator.build(InternalInjectorCreator.java:107)
    at com.google.inject.Guice.createInjector(Guice.java:99)
    at org.everrest.guice.servlet.EverrestGuiceContextListener.getInjector(EverrestGuiceContextListener.java:140)
    at com.google.inject.servlet.GuiceServletContextListener.contextInitialized(GuiceServletContextListener.java:47)
```
it was missing `export` on `CHE_PORT`, that is why it was working only by adding `-e CHE_PORT=8080`

davidfestal:
I tested this on Minishift and can confirm it works as well in this Openshift use-case. Thanks @riuvshin

dockerfiles/che/entrypoint.sh
davidfestal: OK, right

davidfestal (approved):
OK for me to merge

Issue #4096, opened 2017-02-10
Agents installation  specification


skabashnyuk:
- agent installation workflow
- agent installation API (how client should connect to it, stream and interpret data)

garagatyi:
### Glossary:
WS - workspace

### Master - Bootstrapper workflow

1. Master starts WS.
1. Master starts Machine and passes needed environment configuration and an object that contains full definition of agents (including install scripts and dependent agents) in order they are supposed to be started. It is up to implementation to decide how binaries of _Bootstrapper_ should be delivered into machine.
**Agents config file format:**
``` json
[
   {
      "id":"exec",      
      "description":"Agent for command execution",
      "version":"1.0",
      "script":"$HOME/che/exec-agent -addr :4412",
      "servers":{
         "execAgent":{
            "port":"4412/tcp",
            "protocol":"http",
            "path":"/exec"
         }
      }
   }
]
```
3. Master adds Machine into lists of machines waiting for status confirmation.
1. If Machine doesn’t update status that moves it to the next step of booting until timeout expires Master stops machine forcefully and declares that it failed booting. WS start fails too.
1. Machine starts.
1. Machine gets _Bootstrapper_ binaries and starts it.
1. Machine starts _Bootstrapper_ and passes path to agents definition file. 
1. Machine adds _Bootstrapper_ logs into machine logs. It is supposed that logs of this agent should have appropriate prefix, e.g. [_Bootstrapper_].
1. _Bootstrapper_ opens WebSocket connection to master and notifies master that it was started.
1. _Bootstrapper_ starts agents defined by file one-by-one.
1. _Bootstrapper_ notifies master about states of agents.
1. Master waits for successful state of all agents in all machines and performs check of ws-agent that its API is responding. If it is not true and ws-agent start timeout expires WS is treated as failed booting and should be forcefully stopped.

### _Bootstrapper_ - agents workflow:
1. _Bootstrapper_ notifies Master that it started
1. _Bootstrapper_ notifies Master that particular agent starts
1. _Bootstrapper_ executes agent installation script
1. _Bootstrapper_ streams agent logs and states to interested sides
1. _Bootstrapper_ checks agent for installation state:
   1. If Agent contains server
      1. _Bootstrapper_ checks connection to socket on local port of machine
      1. If it is successful agent is treated as started
      1. Otherwise check is repeated with some delay.
      1. If agent boot time expires and last check failed agent is treated as failed
   1. If Agent contains multiple servers
[...]

PR #5345, opened 2017-06-12
Handling bootstrapper & installer events.

mshaposhnik:
### What does this PR do?
Adds handling of bootstrapper & installer  events as in [spec](https://github.com/eclipse/che/issues/4096#issuecomment-283067971)


garagatyi (approved):

gazarenkov:
Let's move Status enums from *Event classes to standalone.
It is better design since strictly speaking Status is not a part of Event (it is "association" relationship rather than "composition")

gazarenkov (approved):

PR #5353, opened 2017-06-12
CHE-280: Adding container's state info to the 'inspectContainer' API

ibuziuk:
### What does this PR do?
Adding container's state info to the 'inspectContainer' API

https://issues.jboss.org/browse/CHE-280

### DIFF
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
@@ -59,6 +59,7 @@ import org.eclipse.che.plugin.docker.client.json.ContainerConfig;
 import org.eclipse.che.plugin.docker.client.json.ContainerCreated;
 import org.eclipse.che.plugin.docker.client.json.ContainerInfo;
 import org.eclipse.che.plugin.docker.client.json.ContainerListEntry;
+import org.eclipse.che.plugin.docker.client.json.ContainerState;
 import org.eclipse.che.plugin.docker.client.json.Event;
 import org.eclipse.che.plugin.docker.client.json.Filters;
 import org.eclipse.che.plugin.docker.client.json.HostConfig;
@@ -105,6 +106,10 @@ import org.slf4j.LoggerFactory;
 
 import io.fabric8.kubernetes.api.model.Container;
 import io.fabric8.kubernetes.api.model.ContainerBuilder;
+import io.fabric8.kubernetes.api.model.ContainerStateRunning;
+import io.fabric8.kubernetes.api.model.ContainerStateTerminated;
+import io.fabric8.kubernetes.api.model.ContainerStateWaiting;
+import io.fabric8.kubernetes.api.model.ContainerStatus;
 import io.fabric8.kubernetes.api.model.DoneableEndpoints;
 import io.fabric8.kubernetes.api.model.Endpoints;
 import io.fabric8.kubernetes.api.model.PersistentVolumeClaim;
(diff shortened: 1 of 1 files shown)


plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -1318,9 +1319,15 @@ private ContainerInfo createContainerInfo(Service svc,
         info.setNetworkSettings(networkSettings);
         info.setHostConfig(hostConfig);
         info.setImage(imageInfo.getConfig().getImage());
+        info.setState(getContainerState(pod));
         return info;
     }
 
+    private ContainerState getContainerState(final Pod pod) {
+        ContainerState state = new ContainerState();
+        state.setStatus(pod.getStatus().getPhase());
```
sunix: Not sure the kubernetes pod phase match with docker status
https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase

> status=(created|restarting|running|removing|paused|exited|dead)

https://docs.docker.com/engine/api/v1.29/#operation/ContainerList

ibuziuk: @sunix pod's states are "Pending / Running / Succeeded / Failed / Unknown" 
What about matching it the following way:
1. created = Succeeded
2. running = Running 
3. dead / exited = Failed

ibuziuk: there is also a possibility to get ContainerState[1] which can be:
```
     * @param running
     * @param waiting
     * @param terminated
```
this model is not also aligned with docker one

[1] http://static.javadoc.io/io.fabric8/kubernetes-api/2.0.38/io/fabric8/kubernetes/api/model/ContainerState.html

amisevsk: @ibuziuk that's sadly probably what's going to be necessary. Another option would be to look at `pod.getStatus.getContainerStatuses()[0].getState()`. According to the kubernetes docs, this *should* be the [output of `docker inspect`](https://kubernetes.io/docs/resources-reference/v1.6/#podstatus-v1-core) but it's not. It might have more / better information though. 

I think running = Running and exited = Failed is good enough for now; it looks like the only use of containerState is to check if that state is `exited` at the moment.

Edit: looks like you slipped in while I was typing that up :)

ibuziuk: @amisevsk all makes sense - lemme update the PR

l0rd: +1 to use ContainerState. And since in a pod there may be more than one container I would prefer `getContainerState` to require the index of the container in the pod or change the method to `getPodContainersState` where the returned value is a collection of `ContainerState`.

ibuziuk: @l0rd thanks for review - working on PR update.

ibuziuk:
PR has been updated

l0rd (approved):
LGTM. Thanks @ibuziuk

sunix (approved):
LGTM

amisevsk (approved):

PR #5366, opened 2017-06-13
Factorize code of `ServerEvaluationStrategy` classes, to use the Custom strategy as the basis of other strategies

davidfestal:
### What does this PR do?

This PR makes the `CustomServerEvaluationStrategy` the basis of all other strategies. This allows having the `LocalDockerSinglePortStrategy` simply use a custom template. 

For the moment the properties in `che.properties` are not changed, and other strategies extend the `custom` strategy with specific settings. So this change should have *no impact* on existing configurations / installations.

https://issues.jboss.org/browse/CHE-277

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
@@ -67,6 +67,12 @@ public class CustomServerEvaluationStrategy extends ServerEvaluationStrategy {
      */
     public static final String CHE_WORKSPACE_ID_PREFIX = "workspace";
 
+
+    /**
+     * name of the macro that indicates if the machine is the dev machine
+     */
+    public static final String IS_DEV_MACHINE_MACRO = "isDevMachine";
+
     /**
      * Used to store the address set by property {@code che.docker.ip}, if applicable.
      */
@@ -407,7 +413,7 @@ public class CustomServerEvaluationStrategy extends ServerEvaluationStrategy {
             globalPropertiesMap.put("wildcardNipDomain", getWildcardNipDomain(externalAddress));
             globalPropertiesMap.put("wildcardXipDomain", getWildcardXipDomain(externalAddress));
             globalPropertiesMap.put("chePort", chePort);
-            globalPropertiesMap.put("isDevMachine", getIsDevMachine());
+            globalPropertiesMap.put(IS_DEV_MACHINE_MACRO, getIsDevMachine());
         }
 
         /**
@@ -421,7 +427,7 @@ public class CustomServerEvaluationStrategy extends ServerEvaluationStrategy {
             }
             ST stringTemplate = new ST(template);
             globalPropertiesMap.forEach((key, value) -> stringTemplate.add(key, 
-                                                                           "isDevMachine".equals(key) ? 
+                                                                           IS_DEV_MACHINE_MACRO.equals(key) ? 
                                                                                Boolean.parseBoolean(value)
                                                                                : value));
             stringTemplate.add("serverName", portsToRefName.get(port));
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
@@ -165,11 +165,11 @@ public class CustomServerEvaluationStrategy extends ServerEvaluationStrategy {
         
         if (isNullOrEmpty(cheDockerCustomExternalTemplate)) {
             return getExposedPortsToAddressPorts(renderingEvaluation.getExternalAddress(), ports, false);
-        } else {
-            return ports.keySet().stream()
+        }
+        
+        return ports.keySet().stream()
                 .collect(Collectors.toMap(portKey -> portKey,
                                           portKey -> renderingEvaluation.render(cheDockerCustomExternalTemplate, portKey)));
-        }
     }
 
 
@@ -281,13 +281,13 @@ public class CustomServerEvaluationStrategy extends ServerEvaluationStrategy {
                     !isNullOrEmpty(gatewayAddressContainer) ?
                     gatewayAddressContainer :
                     this.internalHost;
-            } else {
-                return externalAddressProperty != null ?
+            }
+            
+            return externalAddressProperty != null ?
                     externalAddressProperty :
                     internalAddressProperty != null ?
                     internalAddressProperty :
                     this.internalHost;
-            }
         }
     }
 
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
@@ -171,10 +171,10 @@ public class CustomServerEvaluationStrategy extends ServerEvaluationStrategy {
         if (isNullOrEmpty(cheDockerCustomExternalTemplate)) {
             return getExposedPortsToAddressPorts(renderingEvaluation.getExternalAddress(), ports, false);
         }
-        
+
         return ports.keySet().stream()
-                .collect(Collectors.toMap(portKey -> portKey,
-                                          portKey -> renderingEvaluation.render(cheDockerCustomExternalTemplate, portKey)));
+                  .collect(Collectors.toMap(portKey -> portKey,
+                                            portKey -> renderingEvaluation.render(cheDockerCustomExternalTemplate, portKey)));
     }
 
 
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
@@ -179,8 +179,8 @@ public class CustomServerEvaluationStrategy extends ServerEvaluationStrategy {
         }
 
         return ports.keySet().stream()
-                  .collect(Collectors.toMap(portKey -> portKey,
-                                            portKey -> renderingEvaluation.render(cheDockerCustomExternalTemplate, portKey)));
+                    .collect(Collectors.toMap(portKey -> portKey,
+                                              portKey -> renderingEvaluation.render(cheDockerCustomExternalTemplate, portKey)));
     }
 
 
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
@@ -392,6 +392,7 @@ public class CustomServerEvaluationStrategy extends ServerEvaluationStrategy {
             globalPropertiesMap.put("externalAddress", externalAddress);
             globalPropertiesMap.put("externalIP", externalIP);
             globalPropertiesMap.put("workspaceId", getWorkspaceId());
+            globalPropertiesMap.put("workspaceIdWithoutPrefix", getWorkspaceId().replaceFirst("workspace",""));
             globalPropertiesMap.put("machineName", getMachineName());
             globalPropertiesMap.put("wildcardNipDomain", getWildcardNipDomain(externalAddress));
             globalPropertiesMap.put("wildcardXipDomain", getWildcardXipDomain(externalAddress));
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
@@ -40,7 +40,7 @@ import static com.google.common.base.Strings.isNullOrEmpty;
  * @author Florent Benoit
  * @see ServerEvaluationStrategy
  */
-public class CustomServerEvaluationStrategy extends DefaultServerEvaluationStrategy {
+public class CustomServerEvaluationStrategy extends ServerEvaluationStrategy {
 
     /**
      * Regexp to extract port (under the form 22/tcp or 4401/tcp, etc.) from label references
@@ -58,6 +58,16 @@ public class CustomServerEvaluationStrategy extends DefaultServerEvaluationStrat
     public static final String CHE_MACHINE_NAME_PROPERTY = "CHE_MACHINE_NAME=";
 
     /**
+     * Used to store the address set by property {@code che.docker.ip}, if applicable.
+     */
+    protected String internalAddressProperty;
+
+    /**
+     * Used to store the address set by property {@code che.docker.ip.external}. if applicable.
+     */
+    protected String externalAddressProperty;
+
+    /**
      * The current port of che.
      */
     private final String chePort;
@@ -72,22 +82,70 @@ public class CustomServerEvaluationStrategy extends DefaultServerEvaluationStrat
      */
     private String cheDockerCustomExternalTemplate;
 
-
+    /**
+     * Option to enable the use of the container address, when searching for addresses.
+     */
+    private boolean useContainerAddress;
+    
+    
+    /**
+     * Option to tell if an exception should be thrown when the host defined in the `externalAddress` isn't known
+     * and cannot be converted into a valid IP.
+     */
+    private boolean throwOnUnknownHost = true;
+    
     /**
      * Default constructor
      */
     @Inject
-    public CustomServerEvaluationStrategy(@Nullable @Named("che.docker.ip") String cheDockerIp,
-                                          @Nullable @Named("che.docker.ip.external") String cheDockerIpExternal,
+    public CustomServerEvaluationStrategy(@Nullable @Named("che.docker.ip") String internalAddress,
+                                          @Nullable @Named("che.docker.ip.external") String externalAddress,
                                           @Nullable @Named("che.docker.server_evaluation_strategy.custom.template") String cheDockerCustomExternalTemplate,
                                           @Nullable @Named("che.docker.server_evaluation_strategy.custom.external.protocol") String cheDockerCustomExternalProtocol,
                                           @Named("che.port") String chePort) {
-        super(cheDockerIp, cheDockerIpExternal);
+        this(internalAddress, externalAddress, cheDockerCustomExternalTemplate, cheDockerCustomExternalProtocol, chePort, false);
+    }
+
+    /**
+     * Constructor to be called by derived strategies
+     */
+    public CustomServerEvaluationStrategy(@Nullable @Named("che.docker.ip") String internalAddress,
+                                          @Nullable @Named("che.docker.ip.external") String externalAddress,
+                                          @Nullable @Named("che.docker.server_evaluation_strategy.custom.template") String cheDockerCustomExternalTemplate,
+                                          @Nullable @Named("che.docker.server_evaluation_strategy.custom.external.protocol") String cheDockerCustomExternalProtocol,
+                                          @Named("che.port") String chePort,
+                                          boolean useContainerAddress) {
+        this.internalAddressProperty = internalAddress;
+        this.externalAddressProperty = externalAddress;
         this.chePort = chePort;
         this.cheDockerCustomExternalTemplate = cheDockerCustomExternalTemplate;
         this.cheDockerCustomExternalProtocol = cheDockerCustomExternalProtocol;
+        this.useContainerAddress = useContainerAddress;
     }
 
+    @Override
+    protected Map<String, String> getInternalAddressesAndPorts(ContainerInfo containerInfo, String internalHost) {
+        String internalAddressContainer = containerInfo.getNetworkSettings().getIpAddress();
+        
+        String internalAddress;
+        
+        if (useContainerAddress) {
+            internalAddress = !isNullOrEmpty(internalAddressContainer) ?
+                internalAddressContainer :
+                internalHost;
+        } else {
+            internalAddress =
+                internalAddressProperty != null ?
+                internalAddressProperty :
+                internalHost;
+        }
+        
+        boolean useExposedPorts = useContainerAddress && internalAddress != internalHost;
+
+        return getExposedPortsToAddressPorts(internalAddress, containerInfo.getNetworkSettings().getPorts(), useExposedPorts);
+    }
+
+    
     /**
      * Override the host for all ports by using the external template.
      */
@@ -96,13 +154,17 @@ public class CustomServerEvaluationStrategy extends DefaultServerEvaluationStrat
 
         // create Rendering evaluation
         RenderingEvaluation renderingEvaluation = getOnlineRenderingEvaluation(containerInfo, internalHost);
-
+        
         // get current ports
         Map<String, List<PortBinding>> ports = containerInfo.getNetworkSettings().getPorts();
-
-        return ports.keySet().stream()
-                    .collect(Collectors.toMap(portKey -> portKey,
-                                              portKey -> renderingEvaluation.render(cheDockerCustomExternalTemplate, portKey)));
+        
+        if (isNullOrEmpty(cheDockerCustomExternalTemplate)) {
+            return getExposedPortsToAddressPorts(renderingEvaluation.getExternalAddress(), ports, false);
+        } else {
+            return ports.keySet().stream()
+                .collect(Collectors.toMap(portKey -> portKey,
+                                          portKey -> renderingEvaluation.render(cheDockerCustomExternalTemplate, portKey)));
+        }
     }
 
 
@@ -180,6 +242,11 @@ public class CustomServerEvaluationStrategy extends DefaultServerEvaluationStrat
          * @return the rendering of the template
          */
         String render(String template, String port);
+        
+        /**
+         * Gets default external address.
+         */
+        String getExternalAddress();
     }
 
     /**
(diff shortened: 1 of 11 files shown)


davidfestal:
@benoitf @l0rd could you review this PR please ?

benoitf:
@davidfestal do you have checked https://github.com/codenvy/codenvy/blob/master/plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/CodenvyDockerServerEvaluationStrategy.java is still compiling ?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
```diff
@@ -72,22 +87,70 @@
...
+     */
+    private boolean throwOnUnknownHost = true;
+    
     /**
      * Default constructor
      */
     @Inject
-    public CustomServerEvaluationStrategy(@Nullable @Named("che.docker.ip") String cheDockerIp,
-                                          @Nullable @Named("che.docker.ip.external") String cheDockerIpExternal,
+    public CustomServerEvaluationStrategy(@Nullable @Named("che.docker.ip") String internalAddress,
```
benoitf: maybe it's better to keep the name cheDockerIp / cheDockerIpExternal ?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
```diff
@@ -96,13 +159,17 @@ public CustomServerEvaluationStrategy(@Nullable @Named("che.docker.ip") String c
...
         // get current ports
         Map<String, List<PortBinding>> ports = containerInfo.getNetworkSettings().getPorts();
-
-        return ports.keySet().stream()
-                    .collect(Collectors.toMap(portKey -> portKey,
-                                              portKey -> renderingEvaluation.render(cheDockerCustomExternalTemplate, portKey)));
+        
+        if (isNullOrEmpty(cheDockerCustomExternalTemplate)) {
+            return getExposedPortsToAddressPorts(renderingEvaluation.getExternalAddress(), ports, false);
+        } else {
```
sunix: Is this else necessary ?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
```diff
@@ -72,22 +87,70 @@
...
+        this.externalAddressProperty = externalAddress;
         this.chePort = chePort;
         this.cheDockerCustomExternalTemplate = cheDockerCustomExternalTemplate;
         this.cheDockerCustomExternalProtocol = cheDockerCustomExternalProtocol;
+        this.useContainerAddress = useContainerAddress;
     }
 
+    @Override
+    protected Map<String, String> getInternalAddressesAndPorts(ContainerInfo containerInfo, String internalHost) {
+        String internalAddressContainer = containerInfo.getNetworkSettings().getIpAddress();
```
benoitf: is it final objects ?

benoitf (dismissed):

benoitf:
@garagatyi ?
@davidfestal as long as it works for codenvy strategy, +1 for me

davidfestal:
> @davidfestal do you have checked https://github.com/codenvy/codenvy/blob/master/plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/CodenvyDockerServerEvaluationStrategy.java is still compiling ?

Good point. I didn't check that. Let me have a look.

sunix (changes requested):

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
```diff
@@ -202,12 +274,20 @@ protected OnlineRenderingEvaluation withInternalHost(String internalHost) {
...
-                   gatewayAddressContainer :
-                   this.internalHost;
+        public String getExternalAddress() {
+            if (useContainerAddress) {
+                return externalAddressProperty != null ?
+                    externalAddressProperty :
+                    !isNullOrEmpty(gatewayAddressContainer) ?
+                    gatewayAddressContainer :
+                    this.internalHost;
+            } else {
```
sunix: is this `else` necessary ?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
davidfestal: > Is this else necessary ?

Sure it isn't. Just for style ;-)
Let me change it.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
```diff
@@ -96,13 +159,17 @@ public CustomServerEvaluationStrategy(@Nullable @Named("che.docker.ip") String c
 
         // create Rendering evaluation
         RenderingEvaluation renderingEvaluation = getOnlineRenderingEvaluation(containerInfo, internalHost);
-
+
```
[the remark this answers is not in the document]
davidfestal: OK. Not hard to remove !

davidfestal:
@sunix is it OK for you now ?

davidfestal:
@benoitf after commit https://github.com/eclipse/che/pull/5366/commits/70cb99b00f1ffb966f45cf2e2961a72403ae9dfa the file [CodenvyDockerServerEvaluationStrategy.java](https://github.com/codenvy/codenvy/blob/master/plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/CodenvyDockerServerEvaluationStrategy.java) should compile fine now (very simple to see it from the code since it directly extends `ServerEvaluationStrategy`). However I can checkout the codenvy repo and test it if you want.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
```diff
@@ -82,12 +108,49 @@ public CustomServerEvaluationStrategy(@Nullable @Named("che.docker.ip") String c
...
                                           @Nullable @Named("che.docker.server_evaluation_strategy.custom.external.protocol") String cheDockerCustomExternalProtocol,
                                           @Named("che.port") String chePort) {
-        super(cheDockerIp, cheDockerIpExternal);
+        this(cheDockerIp, cheDockerIpExternal, cheDockerCustomExternalTemplate, cheDockerCustomExternalProtocol, chePort, false);
+    }
+
+    /**
+     * Constructor to be called by derived strategies
+     */
+    public CustomServerEvaluationStrategy(@Nullable @Named("che.docker.ip") String internalAddress,
```
garagatyi: Can you align names of parameters of constructors? E.g. `cheDockerIp == internalAddress`

garagatyi (approved):

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
benoitf: @davidfestal I think this comment was not addressed.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
```diff
@@ -302,7 +382,7 @@ protected void initPortMapping() {
         /**
          * Gets default external address.
          */
-        protected String getExternalAddress() {
+        public String getExternalAddress() {
```
benoitf: do we need public there ?

davidfestal: yes, because I pulled up the method in the `RenderingEvaluation` interface, since it is now also useful in `CustomServerEvaluationStrategy.getExternalAddressesAndPorts()` (in case `cheDockerCustomExternalTemplate` is `null`)

davidfestal:
@benoitf I checked the compilation of codenvy against this PR. it fails, but not due to this PR. in fact the codenvy build fails with the `openshift-connector-rebased` branch due to this change: https://github.com/eclipse/che/commit/ab740c69b4fecfae9a96c6a64dfb563eee2d7a99#diff-eae0d593d9d11b19a0c27f9e468d957bR73 :
The addition of the `useHttpsForExternalUrls() ` abstract method.

I'll fix this anyway, so we won't have any surprise when merging `openshift-connector-rebased ` to master.

benoitf:
@davidfestal ok, waiting for the fix

Also what about the comments on using variable cheDockerIp instead of internalAddress and cheDockerIpExternal for the externalAddress fields ?
it seems you replaced it in custom server strategy with "old" names

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
```diff
@@ -58,6 +58,21 @@
...
 
     /**
+     * Name of the property to get the property that indicates if the machine is the dev machine
+     */
+    public static final String CHE_IS_DEV_MACHINE_PROPERTY = "CHE_IS_DEV_MACHINE=";
+
+    /**
+     * Used to store the address set by property {@code che.docker.ip}, if applicable.
+     */
+    protected String internalAddressProperty;
```
benoitf: we should use cheDockerIp

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
```diff
@@ -58,6 +58,21 @@
...
+
+    /**
+     * Used to store the address set by property {@code che.docker.ip.external}. if applicable.
+     */
+    protected String externalAddressProperty;
```
benoitf: and here cheDockerIpExternal

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
davidfestal: The names used here are those that were used initially in the `DefaultServerEvaluationStrategy`.

davidfestal: all the other strategies had `internalAddressProperty` and `externalAddressProperty` That's why I had kept them.

benoitf: yes but name of properties have been changed during that time so it's better to keep matching name between the name of the property and the value of the field

davidfestal:
@benoitf I fixed the problem that produce a Codenvy build failure, and changed the names of the custom strategy attributes as requested.

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
```diff
@@ -313,14 +393,16 @@ protected String getExternalAddress() {
         protected void populateGlobalProperties() {
             String externalAddress = getExternalAddress();
             String externalIP = getExternalIp(externalAddress);
-            globalPropertiesMap.put("internalIp", internalAddressProperty);
+            globalPropertiesMap.put("internalIp", cheDockerIp);
             globalPropertiesMap.put("externalAddress", externalAddress);
             globalPropertiesMap.put("externalIP", externalIP);
             globalPropertiesMap.put("workspaceId", getWorkspaceId());
+            globalPropertiesMap.put("workspaceIdWithoutPrefix", getWorkspaceId().replaceFirst("workspace",""));
```
benoitf: could you move "workspace" as a class constant ?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
```diff
@@ -99,10 +162,14 @@ public CustomServerEvaluationStrategy(@Nullable @Named("che.docker.ip") String c
 
         // get current ports
         Map<String, List<PortBinding>> ports = containerInfo.getNetworkSettings().getPorts();
-
+
```
sunix: could this change be removed (trailing space)

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
```diff
@@ -99,10 +162,14 @@ public CustomServerEvaluationStrategy(@Nullable @Named("che.docker.ip") String c
 
         // get current ports
         Map<String, List<PortBinding>> ports = containerInfo.getNetworkSettings().getPorts();
-
+        
+        if (isNullOrEmpty(cheDockerCustomExternalTemplate)) {
+            return getExposedPortsToAddressPorts(renderingEvaluation.getExternalAddress(), ports, false);
+        }
+
```
sunix: could trailing space be removed ?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
```diff
@@ -99,10 +162,14 @@ public CustomServerEvaluationStrategy(@Nullable @Named("che.docker.ip") String c
...
+        
+        if (isNullOrEmpty(cheDockerCustomExternalTemplate)) {
+            return getExposedPortsToAddressPorts(renderingEvaluation.getExternalAddress(), ports, false);
+        }
+        
         return ports.keySet().stream()
-                    .collect(Collectors.toMap(portKey -> portKey,
-                                              portKey -> renderingEvaluation.render(cheDockerCustomExternalTemplate, portKey)));
+                .collect(Collectors.toMap(portKey -> portKey,
+                                          portKey -> renderingEvaluation.render(cheDockerCustomExternalTemplate, portKey)));
```
sunix: if there is no change and just formatting, it would be better if these changes are not included in the PR. (if there is a conflict with another branch it can be a harder to merge). Formatting should be done separately to simplify merge IMO.

sunix (changes requested):

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
davidfestal: done

benoitf: this property should be also tested in unit test

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
```diff
@@ -313,14 +398,16 @@ protected String getExternalAddress() {
...
+            globalPropertiesMap.put("internalIp", cheDockerIp);
             globalPropertiesMap.put("externalAddress", externalAddress);
             globalPropertiesMap.put("externalIP", externalIP);
             globalPropertiesMap.put("workspaceId", getWorkspaceId());
+            globalPropertiesMap.put("workspaceIdWithoutPrefix", getWorkspaceId().replaceFirst(CHE_WORKSPACE_ID_PREFIX,""));
             globalPropertiesMap.put("machineName", getMachineName());
             globalPropertiesMap.put("wildcardNipDomain", getWildcardNipDomain(externalAddress));
             globalPropertiesMap.put("wildcardXipDomain", getWildcardXipDomain(externalAddress));
             globalPropertiesMap.put("chePort", chePort);
+            globalPropertiesMap.put("isDevMachine", getIsDevMachine());
```
benoitf: missing unit test

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
```diff
@@ -333,12 +420,26 @@ public String render(String template, String port) {
                 this.initialized = true;
             }
             ST stringTemplate = new ST(template);
-            globalPropertiesMap.forEach((key, value) -> stringTemplate.add(key, value));
+            globalPropertiesMap.forEach((key, value) -> stringTemplate.add(key, 
+                                                                           "isDevMachine".equals(key) ?
```
benoitf: "isDevMachine" should be a class constant ?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java
davidfestal: Aren't you requesting the exact opposite change as the one you requested [previously](https://github.com/eclipse/che/pull/5366#discussion_r121704027)

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategyTest.java
```diff
@@ -129,6 +130,34 @@ public void testWorkspaceIdRule() throws Throwable {
...
+     */
+    @Test
+    public void testWorkspaceIdWithoutPrefixRule() throws Throwable {
+        this.customServerEvaluationStrategy =
+                new CustomServerEvaluationStrategy("10.0.0.1", "192.168.1.1", "<workspaceIdWithoutPrefix>", "http", "8080");
+
+        Map<String, String> portMapping = this.customServerEvaluationStrategy.getExternalAddressesAndPorts(containerInfo, "localhost");
+
+        Assert.assertTrue(portMapping.containsKey("4401/tcp"));
+        Assert.assertEquals(portMapping.get("4401/tcp"), WORKSPACE_ID_VALUE.replaceFirst(CHE_WORKSPACE_ID_PREFIX, ""));
```
benoitf: where do we check that we have value ABCDEFG ? for the workspaceId without prefix ?

plugins/plugin-docker/che-plugin-docker-machine/src/test/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategyTest.java
```diff
@@ -129,6 +130,34 @@ public void testWorkspaceIdRule() throws Throwable {
...
+    @Test
+    public void testIsDevMachine() throws Throwable {
+        this.customServerEvaluationStrategy =
+                new CustomServerEvaluationStrategy("10.0.0.1", "192.168.1.1",
+                                                   "<if(isDevMachine)><workspaceId><else><machineName><endif>", "http", "8080");
+
+        Map<String, String> portMapping = this.customServerEvaluationStrategy.getExternalAddressesAndPorts(containerInfo, "localhost");
+
+        Assert.assertTrue(portMapping.containsKey("4401/tcp"));
+        Assert.assertEquals(portMapping.get("4401/tcp"), WORKSPACE_ID_VALUE);
```
benoitf, line 164: I don't see where we check the value of the template value ? if we get workspaceId or machineName ?

benoitf (approved):

benoitf:
@davidfestal @slemeur maybe there is a missing Changelog section in the description

davidfestal:
> @davidfestal @slemeur maybe there is a missing Changelog section in the description

Well, I don't know since these are only internal changes. No property was changed, and the behavior is the same.

davidfestal:
@sunix is it OK now after https://github.com/eclipse/che/pull/5366/commits/cd27b8b589d29426e59d1a15fdb5223f3d327fad ?

benoitf:
@davidfestal yes but template asks for a one-line changelog for all pull requests
(Release notes and docs are optionals for internal refactoring)

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/LocalDockerServerEvaluationStrategy.java
```diff
@@ -10,19 +10,14 @@
...
 import org.eclipse.che.plugin.docker.client.json.ContainerInfo;
-import org.eclipse.che.plugin.docker.client.json.PortBinding;
 
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-
-import static com.google.common.base.Strings.isNullOrEmpty;
+import com.google.inject.Inject;
+import com.google.inject.name.Named;
```
sunix: changing order of import is not easy to spot what has been added or removed or not changed

davidfestal:
@benoitf please note that it's a PR against the `openshift-connector-rebased` branch that in turns will be merged to master.
Maybe there will be a changelog for the global PR ?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/LocalDockerSinglePortServerEvaluationStrategy.java
```diff
@@ -11,19 +11,11 @@
...
-import org.eclipse.che.plugin.docker.client.json.ContainerInfo;
-import org.eclipse.che.plugin.docker.client.json.PortBinding;
 
-import java.util.Arrays;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.stream.Stream;
+import com.google.inject.Inject;
+import com.google.inject.name.Named;
```
sunix: change imports order should be done in a separate PR

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/LocalDockerServerEvaluationStrategy.java
davidfestal: Obviously due to a "clean imports" in Eclipse.
But let me change this if it blocks your approval.

benoitf:
@davidfestal ah ok I thought it was for master :)

benoitf:
@davidfestal BTW why it isn't directly merged to master ?

davidfestal:
> @davidfestal BTW why it isn't directly merged to master ?

Because the first idea was to have the new `LocalDockerSinglePortServerEvaluationStrategy` (existing only in `openshift-connector` and `openshift-connector-rebased` branches) reuse the new `custom` strategy. But finally, this required some more refactoring. However this started  as a pure Openshift connector change.

davidfestal:
@sunix I fixed the order of the imports

benoitf:
>> @davidfestal BTW why it isn't directly merged to master ?

> Because the first idea was to have the new `LocalDockerSinglePortServerEvaluationStrategy` (existing only in `openshift-connector` and `openshift-connector-rebased` branches) reuse the new `custom` strategy. But finally, this required some more refactoring. However this started  as a pure Openshift connector change.


@davidfestal ok this is because I think that merging into master will allow to decrease the work on merge operations into the openshift-connector-rebased branch, as BTW is being rebased on master.

davidfestal:
> @davidfestal ok this is because I think that merging into master will allow to decrease the work on merge operations into the openshift-connector-rebased branch, as BTW is being rebased on master.

Well, the problem is that it depends on previous changes made on the `openshift-connector-branch`, and is a prerequisite for PR https://github.com/eclipse/che/pull/5110 , which also depends on previous `openshift-connector-branch` changes

slemeur (approved):
Ok for me as it is going to `openshift-connector-rebased`

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/LocalDockerSinglePortServerEvaluationStrategy.java
```diff
@@ -41,57 +33,13 @@
...
-            addressesAndPorts.put(serverKey, serverURL);
-        }
-
-        return addressesAndPorts;
-    }
-
-    private String getWorkspaceServerName(Map<String, String> labels, String portKey) {
-        ServerConfImpl serverConf = getServerConfImpl(portKey, labels, new HashMap<>());
-        if (serverConf == null) {
-            return "server-" + portKey.split("/",2)[0];
```
sunix, line 84: OK these lines have been removed, how do we deal with exposed ports without label ? we did a quick fix here https://github.com/eclipse/che/pull/4970 while waiting for the rebase to work on a refactored and cleanner fix. Would preview urls works with non label exposed ports ?

davidfestal: Seems to me it is managed here:

https://github.com/davidfestal/che/blob/0f9a64170ee72077099a89f458ad557f1ae38380/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/CustomServerEvaluationStrategy.java#L384

sunix: thanks

sunix (approved):

Issue #5338, opened 2017-06-12
User preferences are not cleaned


svor:
Workspace information which stores in user preferences isn't cleaned after deleting the workspace. So as a result a json (user preferences) gets bigger and bigger when user creates new workspace.

**Reproduction Steps:**
1. Create a workspace with some project and open some file;
2. Refresh IDE and look on http://localhost:8080/api/preferences - here you can find an information about your ws;
3. Delete this ws and create new one;
4. If you go to http://localhost:8080/api/preferences again you can find an info about your previous ws.

PR #5367, opened 2017-06-13
CHE-2230: Save preferences every 10 seconds

svor:
### What does this PR do?
Adds a repeating command that is saved user's preferences every 10 seconds. 
This is temporary solution. Should be properly fixed by https://github.com/eclipse/che/issues/5338

https://github.com/codenvy/codenvy/issues/2230

### DIFF
--- a/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/jsonrpc/impl/GsonJsonRpcComposer.java
+++ b/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/jsonrpc/impl/GsonJsonRpcComposer.java
@@ -64,7 +64,7 @@ public class GsonJsonRpcComposer implements JsonRpcComposer {
             JsonArray jsonArray = new JsonArray();
             for (int i = 0; i < paramsList.size(); i++) {
                 JsonElement jsonElement = (JsonElement)paramsList.get(i);
-                jsonArray.set(i, jsonElement);
+                jsonArray.add(jsonElement);
             }
             return DtoFactory.getInstance().createListDtoFromJson(jsonArray.toString(), type);
         }
--- a/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/context/AppContextImpl.java
+++ b/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/context/AppContextImpl.java
@@ -11,6 +11,7 @@
 package org.eclipse.che.ide.context;
 
 import com.google.gwt.core.client.Callback;
+import com.google.gwt.core.client.Scheduler;
 import com.google.inject.Inject;
 import com.google.inject.Provider;
 import com.google.inject.Singleton;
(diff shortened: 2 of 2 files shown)


dkulieshov (approved):

benoitf:
I'm not sure it looks great even as a temporary solution. There is no hooks to do that on some workspace actions ? also 10seconds looks like very often and it's not configurable at all.

skabashnyuk:
is this implemented over websocket or direct http call?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/jsonrpc/impl/GsonJsonRpcComposer.java
```diff
@@ -64,7 +64,7 @@
             JsonArray jsonArray = new JsonArray();
             for (int i = 0; i < paramsList.size(); i++) {
                 JsonElement jsonElement = (JsonElement)paramsList.get(i);
-                jsonArray.set(i, jsonElement);
+                jsonArray.add(jsonElement);
```
skabashnyuk, line 67: what is it? Fix of  some issue in RPC transport?

svor: Yes, it is small fix for the JSON RPC composer

svor:
@benoitf I have no idea which ws action I can use here (when browser tab with ide closes or reloads). If use windowClosing action for some user who has a big text of preferenses, it can't be saved because the browser canceles this request. 
So, @slemeur I don't think that this issue is regression and we need to fix this https://github.com/eclipse/che/issues/5338.
    
@skabashnyuk It is http call. I've tried to use JSON RPC but have found a bug (https://github.com/eclipse/che/issues/5369).

dkulieshov:
Hi,

Hi, as of what I see this is not a regression as this is reproduced on a specific environment and I'm pretty sure that it was not tested on such environments earlier. This seems to be a downside of how we manage user preferences at the moment:
 - We have no possibility to incrementally save user preference data, so we send it all (dozens of KB)
 - We cannot confidently rely on hooks or any other event based mechanism (like close tab action, refresh, etc.) here

So we simply can't update  large user preferences properly and in time when we refresh/close tabs. Also taking into account time limit for this issue it is obvious that [there's no easy way out](https://www.youtube.com/watch?v=rOXaPE6gklI) and I see that the option that is chosen as the most appropriate here is to send it by schedule.

> I'm not sure it looks great even as a temporary solution. 

I can't agree more here. The solution that is presented is quite doubtful however here we have not many options. We either accept this as best afford or leave it as is and wait until user preferences implementation is updated/redesigned. I guess https://github.com/eclipse/che/issues/5338 is not enough for that, will create appropriate issue.

As this is not a regression and not a blocker I would humbly suggest to postpone it until proper fix is ready.

riuvshin:
@svor so it seems we have to close this PR

slemeur:
It is fine for me.

Please @svor and @dkuleshov, cross link the issue with the refactoring needed for preferences.

Issue #5092, opened 2017-05-16
Rework SPI components to allow clients to get current state of Runtime while it is still starting.


skabashnyuk:
To allow clients make a load of workspace more smooth and provide a feeling that it starts faster we should allow retrieving of the current state of runtime while it is in starting state.
Goal:
-  Rework WorkspaceRuntimes and  WorkspaceManager - Workspace runtime should be available during workspace start 
- Revise methods WorkspaceManager.removeWorkspace an DockerAbandonedResourcesCleaner.cleanNetworks . Now we have assumptions that WorkspaceRuntimes.hasRuntime true if a workspace is running. After 1. it wouldn't be true.

gazarenkov:
The contract of Context and Runtime changed.
Now Context is able to create/return Runtime independently of its state (unconditionally getRuntime() added).
So *start()/stop()* methods as well as *status* should be moved from Context to InternalRuntime

PR #5381, opened 2017-06-14
CHE-5092 Fix returning of runtime while it is starting

sleshchenko:
### What does this PR do?
Moves StartSynchronizer to DockerInternalRuntimes because it is unused in DockerRuntimeContext.
Fixes destroying of a newly created machine when start is canceled.
Moves publishing of STARTING workspace event to a place where runtime is already ready.

https://github.com/eclipse/che/issues/5092


gazarenkov (approved):
OK, but let's review implementation of Docker Infra in general

garagatyi (approved):

voievodin (approved):

Issue #4518, opened 2017-03-22
[dashboard] Update the styles of inputs


ashumilova:
Update the styles of inputs due to the following spec:
https://app.zeplin.io/project.html#pid=5875f7218d4086df39a3a603&sid=58ca6362295490ebed2a12eb

Prepare page in dashboard demo components with different states samples.

![image](https://cloud.githubusercontent.com/assets/1611939/26443334/9488d8cc-4140-11e7-8f0e-8779c13b5644.png)
![image](https://cloud.githubusercontent.com/assets/1611939/26443342/a0378e8e-4140-11e7-896c-41281d2170a8.png)

olexii4:
1. After erasing the team name.
![selection_001](https://user-images.githubusercontent.com/6310786/27134392-145bd120-511e-11e7-9e4b-3ea7e33f0571.png)
2. After erasing the workspace cap.
![selection_002](https://user-images.githubusercontent.com/6310786/27134393-173888de-511e-11e7-9157-8f4276aae45f.png)
3. After team selected.
![selection_003](https://user-images.githubusercontent.com/6310786/27134397-1a071e18-511e-11e7-9488-40a0199d5210.png)
4. After adding special characters to the team name.
![selection_004](https://user-images.githubusercontent.com/6310786/27134404-1e38cf9a-511e-11e7-9e43-0a90afecd934.png)
5. After resizing to width < 960px ('xs' 
 and 'sm')
![selection_005](https://user-images.githubusercontent.com/6310786/27134409-2181076c-511e-11e7-8fb8-08b659a8bd1f.png)

slemeur:
Great work @olexii4 ! 

I just have a question on the second screenshot, comparing to the third screenshot I don't get the state that you want to show. Could you describe it?

*Note:* You should add in your comment the description of each state you are showing, this way it will make it easier to other to easily understand ;)

olexii4:
done

PR #5384, opened 2017-06-15
CHE-4518: change styles for input boxes

olexii4:
Signed-off-by: Oleksii Orel <oorel@redhat.com>

### What does this PR do?
Change styles for input boxes.

https://github.com/eclipse/che/issues/4518

#### Changelog
-  Change styles for input boxes for UD.

#### Release Notes
On our road to Che 6, we have refined the input boxes in the Dashboard so that they look prettier and better to interact with. The different states of the input boxes are well defined and they allow the user to properly identify if the input is having the focus and is in edit mode. Similarly, it is now clearer to understand the mandatory fields and when the values are incorrect. As for the styles of the input boxes, they are simpler and easier to identify in the page.

#### Docs PR
N/A

selection_006
selection_007
selection_008


dashboard/src/app/factories/create-factory/create-factory.html
```diff
@@ -25,28 +25,20 @@
...
+                     che-form="factoryNameForm"
+                     che-name="name"
+                     che-place-holder="Name of the factory"
+                     aria-label="Name of the factory"
+                     ng-model="createFactoryCtrl.name"
+                     ng-trim
+                     ng-minlength="3"
+                     ng-maxlength="20"
+                     ng-pattern="/^[ A-Za-z0-9_\-\.]+$/">
+        <div ng-message="pattern">Factory name may contain digits, latin letters, spaces, _ , . , - and should start only with digits, latin letters or underscores</div>
```
benoitf, line 38: You removed the ng-message =required . Does it need to be removed on all che-input-box elements in UD ?

dashboard/src/app/factories/create-factory/git/create-factory-git.html
```diff
@@ -10,19 +10,20 @@
...
+                   che-label-name="Git URL"
+                   che-form="createFactoryGitForm"
+                   che-name="remoteGitURL"
+                   che-place-holder="Repository URL"
+                   che-width="auto"
+                   ng-model="createFactoryGitCtrl.location"
+                   git-url
+                   required>
+      <div ng-message="gitUrl" ng-if="createFactoryGitCtrl.location && createFactoryGitCtrl.location.length > 0">
+        <span>Invalid Git URL</span>
```
benoitf: the required field is there transformed as a span element, on the previous change it was deleted. Could you explain why in some case we need to remove it, or transform it to a span element. We couldn't keep the ng-message required element to be displayed ?

dashboard/src/app/factories/create-factory/create-factory.html
olexii4: I removed "ng-message =required" because we have another solution(red placeholder with an asterisk) in mockups;
selection_010

olexii4: Both messages and an asterisk will be too much.

dashboard/src/app/factories/create-factory/git/create-factory-git.html
olexii4: I hide the error message in a case with required input (the same reason).
I will remove <span> element for this message.

olexii4:
done

dashboard/src/app/factories/create-factory/create-factory.html
benoitf: so, is that ng-message="required" need to be removed on all occurence of che-input-box we have in all pages of UD ?

olexii4: Yes. But we have only three input boxes in che. And I have changed them. 

We have many places with input box in codenvy.  In general, we have three input widgets: che-input,  che-input-box and che-textarea.

benoitf: ah ok :) I thought we had more ^^

benoitf (approved):

akurinnoy (approved):

slemeur (approved):

ashumilova (approved):

benoitf (approved):

akurinnoy (approved):

ashumilova (approved):

PR #5386, opened 2017-06-15
CHE-274 - Improve idling implementation

snjeza:

### What does this PR do?

Fixes https://issues.jboss.org/browse/CHE-274

I have moved the code from the codenvy project, but I'm not sure if that has been done properly.
Could you please review?

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterModule.java
+++ b/assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterModule.java
@@ -162,7 +162,8 @@ public class WsMasterModule extends AbstractModule {
         bindConstant().annotatedWith(Names.named("machine.terminal_agent.run_command"))
                       .to("$HOME/che/terminal/che-websocket-terminal " +
                           "-addr :4411 " +
-                          "-cmd ${SHELL_INTERPRETER}");
+                          "-cmd ${SHELL_INTERPRETER}" +
+                          "-enable-activity-tracking");
         bindConstant().annotatedWith(Names.named("machine.exec_agent.run_command"))
                       .to("$HOME/che/exec-agent/che-exec-agent " +
                           "-addr :4412 " +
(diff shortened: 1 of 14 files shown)


skabashnyuk:
I'm proposing to make this pr to master and when backport to ```openshift-connector-rebased``` if needed. It can help us to reduce the size of ```openshift-connector-rebased```.  wdyt?

l0rd:
@skabashnyuk 👍

l0rd:
@snjeza can you please modify this PR to point to master branch instead of openshift-connector-rebased?

akorneta:
So first thing that is nessesary to fix it is the license([from](https://github.com/eclipse/che/pull/5386/files#diff-22cb4cb28f95479f7a551d21b7b19b94R5), [to](https://github.com/eclipse/che/blob/master/assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterModule.java#L2)), also I think we need to improve the way how we track activity through the websocket connection because if the wsagent server lay down then workspace would be active for a long time while the server will write into logs, or track only requests wdyt  @snjeza, @skabashnyuk, @l0rd?

skabashnyuk:
@akorneta  it can be part of the second issue. I think we good with current behavior.

assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterModule.java
```diff
@@ -162,7 +162,8 @@ protected void configure() {
         bindConstant().annotatedWith(Names.named("machine.terminal_agent.run_command"))
                       .to("$HOME/che/terminal/che-websocket-terminal " +
                           "-addr :4411 " +
-                          "-cmd ${SHELL_INTERPRETER}");
+                          "-cmd ${SHELL_INTERPRETER}" +
```
akorneta, line 165: please add a space after ${SHELL_INTERPRETER}

akorneta:
@skabashnyuk ok, all other good to me

akorneta (approved):

skabashnyuk:
@snjeza can you change target branch to master?

snjeza:
I have created https://github.com/eclipse/che/pull/5425

PR #5388, opened 2017-06-16
CHE-2242: Limit username length to 39 chars

vinokurig:
### What does this PR do?
Limit username length to 39 chars.
1. User is created with oauth:
-- automatically cut username to 39 chars 
-- if username wich consist of 39 charackters already exists, replace the last 4 chars with a random ones.

2. User is created by API
-- If username contains more than 39 chars return the error message: "Username should contain less then 39 chars.".

https://github.com/codenvy/codenvy/issues/2242

#### Changelog
Limit username length to 39 chars

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/wsmaster/che-core-api-account/src/test/java/org/eclipse/che/account/spi/AccountValidatorTest.java
+++ b/wsmaster/che-core-api-account/src/test/java/org/eclipse/che/account/spi/AccountValidatorTest.java
@@ -20,8 +20,13 @@ import org.testng.annotations.DataProvider;
 import org.testng.annotations.Listeners;
 import org.testng.annotations.Test;
 
+import static org.mockito.AdditionalMatchers.not;
 import static org.mockito.Matchers.anyString;
+import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.doThrow;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNotEquals;
+import static org.testng.Assert.assertTrue;
 
 /**
  * Tests of {@link AccountValidator}.
@@ -32,6 +37,8 @@ import static org.mockito.Mockito.doThrow;
  */
 @Listeners(MockitoTestNGListener.class)
 public class AccountValidatorTest {
+
+    private final static String LONG_ACCOUNT_NAME = "accountNameThatContainsMoreThan39Characters";
 
     @Mock
     private AccountManager accountManager;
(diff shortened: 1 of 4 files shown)


wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/UserValidator.java
```diff
@@ -54,6 +54,9 @@ public void checkUser(User user) throws BadRequestException {
         if (isNullOrEmpty(user.getName())) {
             throw new BadRequestException("User name required");
         }
+        if (user.getName().length() > 39) {
+            throw new BadRequestException("Username should contain less then 39 characters");
```
sleshchenko: typo `then` -> `than`

wsmaster/che-core-api-account/src/test/java/org/eclipse/che/account/spi/AccountValidatorTest.java
```diff
@@ -50,7 +57,22 @@ public void testNormalizeAccountName(String input, String expected) throws Excep
     public void testNormalizeAccountNameWhenInputDoesNotContainAnyValidCharacter() throws Exception {
         doThrow(NotFoundException.class).when(accountManager).getByName(anyString());
 
-        Assert.assertTrue(accountValidator.normalizeAccountName("#", "name").startsWith("name"));
+        assertTrue(accountValidator.normalizeAccountName("#", "name").startsWith("name"));
+    }
+
+    @Test
+    public void shouldChangeLast4CharactersInAccountNameIfAccountNameContains39CharactersAndAccountAlreadyExists() throws Exception {
```
sleshchenko, line 64: test name says `..AndAccountAlreadyExists` but I see another logic with `AndAccountDoesNotExist`. If I'm right please fix it and add new test to cover both of them.

wsmaster/che-core-api-account/src/test/java/org/eclipse/che/account/spi/AccountValidatorTest.java
```diff
@@ -62,39 +84,40 @@ public void testValidUserName(String input, boolean expected) throws Exception {
...
+                              {"TEST", "TEST"},
+                              {"test-", "test"},
+                              {"-test", "test"},
+                              {"te_st", "test"},
+                              {"te#st", "test"},
+                              {"-test", "test"},
+                              {"test-", "test"},
+                              {"--test--", "test"},
+                              {"t-----e--s-t", "t-e-s-t"},
+                              {"testTestTestTestTestTestTestTestTestTestTest", "testTestTestTestTestTestTestTestTestTes"}
```
sleshchenko, line 100: I'd prefer to use test value something like `f23456789012345678901234567890` then it'd be easier to understand this check.

wsmaster/che-core-api-account/src/test/java/org/eclipse/che/account/spi/AccountValidatorTest.java
```diff
@@ -62,39 +84,40 @@ public void testValidUserName(String input, boolean expected) throws Exception {
...
+                              {"test 123", false},
+                              {"test@gmail.com", false},
+                              {"test-", false},
+                              {"-test", false},
+                              {"te_st", false},
+                              {"te#st", false},
+                              {"-test", false},
+                              {"test-", false},
+                              {"--test--", false},
+                              {"te--st", false}
```
sleshchenko, line 120: please add invalid value with length more than 39 chars

wsmaster/che-core-api-account/src/test/java/org/eclipse/che/account/spi/AccountValidatorTest.java
sleshchenko: Ohh, I see why you miss this example. So normalization in AccountValidator cut values with more than 39 chars, but for validation it is a valid value. I don't like such behavior. Maybe we can move cutting to 39 chars to UseValidator if we don't need it for all accounts. WDYT?

sleshchenko (changes requested):

vinokurig:
@sleshchenko According to the issue (https://github.com/codenvy/codenvy/issues/2242#issuecomment-304816521), If user is creaded by API and username is longer than 39 chars, exception must be thrown, but if user is created with oaut and username is longer than 39 chars, we must cut the username to 39 chars.

Issue #5093, opened 2017-05-16
Add server health concept into infrastructure workflow.


skabashnyuk:
Add health-check concept into the model of server conf. Add health status into the model of runtime server. Add implementation of server health-checking in Docker SPI impl.

gorkem:
I think @amisevsk can give some insights to how Openshift/K8s handles these.

gazarenkov:
Yea, I think we need more information about how it already implemented.
Ideally, we may let infrastructure to implement it or reuse it (the case of Openshift/K8s AFAIK) to not to reinvent the wheel. Though, not sure yet.
So, @gorkem good idea, thanks.
@amisevsk any suggestions/insights as for Openshift/K8s impl?

amisevsk:
@gazarenkov sorry about the delay. 

I haven't put as much thought into this as will probably be necessary, but I'm pretty sure on the Openshift/K8s side we can manage to monitor workspace health through the built-in functionality. I haven't tested it though, so a Che-side implementation may still be necessary. In short, health checks on Openshift should probably be implemented to use Openshift liveness probes and an event listener -- this looks quite different from the docker implementation. On Openshift (if we went with this implementation), we would not probably not be requesting health checks directly, and instead listening for 'unhealthy' events. In terms of an abstract SPI model, I would prefer to have an assumption of "machine is healthy unless notified otherwise". 

What I would say regarding the SPI is that ideally it's high level enough that it is not responsible for workspace health/readiness, and instead listens for e.g. "workspace unhealthy" events, etc. I think it should be up to the SPI implementation-side on how this is done -- the current implementation for docker is likely redundant when running on Openshift (although we don't currently have a solution and instead use the docker listener anyways).

For Openshift, we can include liveness/readiness probes in the workspace pod spec, and listen for events to detect issues. I think the [available options](https://docs.openshift.org/latest/dev_guide/application_health.html) should be sufficient for Che's uses, and might even be cleaner on the Openshift side, since currently we are working around a lot of Docker specific implementation stuff. 

Feel free to ping me on Slack/github/email if this isn't clear or if I'm misunderstanding the root problem.

gazarenkov:
@amisevsk sorry for delay (something with GH notification on my side :( )

Thanks for the info!
It looks like we can reuse "Readiness probe" to let Infrastructure (SPI implementation) know when Server (or Machine) is up and "Liveness" when it down...

[...]

PR #5405, opened 2017-06-19
CHE-5093: hardcode server status check in docker infra 

garagatyi:
### What does this PR do?
Add hardcoded server status check in Docker infrastructure to
test usage of server statuses events in client load sequence.
Add server url into ServerStatueEvent to ease workflow of client.
Add more info about agent into internal representation of agent.
Use assisted injection to cleanup docker infra code.
Allow to access docker machine before agents start to prevent
crashes of wsagent.

Related to #5093

### DIFF
--- a/infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerInternalRuntime.java
+++ b/infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerInternalRuntime.java
@@ -10,20 +10,28 @@
  *******************************************************************************/
 package org.eclipse.che.workspace.infrastructure.docker;
 
+import com.google.inject.assistedinject.Assisted;
+
+import org.eclipse.che.api.agent.shared.model.Agent;
+import org.eclipse.che.api.agent.shared.model.impl.AgentImpl;
 import org.eclipse.che.api.core.NotFoundException;
 import org.eclipse.che.api.core.model.machine.MachineSource;
 import org.eclipse.che.api.core.model.workspace.runtime.Machine;
 import org.eclipse.che.api.core.model.workspace.runtime.MachineStatus;
 import org.eclipse.che.api.core.model.workspace.runtime.RuntimeIdentity;
+import org.eclipse.che.api.core.model.workspace.runtime.Server;
+import org.eclipse.che.api.core.model.workspace.runtime.ServerStatus;
 import org.eclipse.che.api.core.notification.EventService;
 import org.eclipse.che.api.workspace.server.DtoConverter;
 import org.eclipse.che.api.workspace.server.URLRewriter;
+import org.eclipse.che.api.workspace.server.model.impl.MachineImpl;
 import org.eclipse.che.api.workspace.server.model.impl.MachineSourceImpl;
 import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
 import org.eclipse.che.api.workspace.server.spi.InternalInfrastructureException;
 import org.eclipse.che.api.workspace.server.spi.InternalMachineConfig;
 import org.eclipse.che.api.workspace.server.spi.InternalRuntime;
 import org.eclipse.che.api.workspace.shared.dto.event.MachineStatusEvent;
+import org.eclipse.che.api.workspace.shared.dto.event.ServerStatusEvent;
 import org.eclipse.che.dto.server.DtoFactory;
 import org.eclipse.che.plugin.docker.client.MessageProcessor;
 import org.eclipse.che.workspace.infrastructure.docker.exception.SourceNotFoundException;
@@ -35,6 +43,12 @@ import org.eclipse.che.workspace.infrastructure.docker.snapshot.SnapshotExceptio
 import org.eclipse.che.workspace.infrastructure.docker.snapshot.SnapshotImpl;
 import org.slf4j.Logger;
 
+import javax.inject.Inject;
+import javax.ws.rs.core.UriBuilder;
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -45,6 +59,7 @@ import java.util.Map;
 import java.util.Queue;
 
 import static java.lang.String.format;
+import static java.util.stream.Collectors.toMap;
 import static org.slf4j.LoggerFactory.getLogger;
 
 /**
@@ -54,6 +69,13 @@ public class DockerInternalRuntime extends InternalRuntime<DockerRuntimeContext>
 
     private static final Logger LOG = getLogger(DockerInternalRuntime.class);
 
+    private static Map<String, String> livenessChecksPaths = new HashMap<>();
+
+    static {
+        livenessChecksPaths.put("wsagent", "/api/");
+        livenessChecksPaths.put("exec-agent", "/process");
+    }
+
     private final StartSynchronizer    startSynchronizer;
     private final Map<String, String>  properties;
     private final Queue<String>        startQueue;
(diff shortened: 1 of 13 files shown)


infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerInfraFactory.java
```diff
@@ -0,0 +1,39 @@
...
+import org.eclipse.che.api.core.model.workspace.runtime.RuntimeIdentity;
+import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
+import org.eclipse.che.workspace.infrastructure.docker.model.DockerEnvironment;
+
+import java.util.List;
+
+/**
+ * @author Alexander Garagatyi
+ */
+public interface DockerInfraFactory {
```
voievodin: I would say that this name is not a good one and may be confusing, while `DockerRuntimeFactory` is better candidate. Can you please add a small description which describes a purpose of this interface(solve dependency routine i think). Does it make sense to make this interface package private, so user will not use it for `Runtime` instances creation? Does it make sense to have separate factories for context & runtime?

sleshchenko (approved):

voievodin (approved):

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerInfraFactory.java
garagatyi: I've just changed that

garagatyi:
@evoevodin @sleshchenko I've just changed PR significantly. Could you do another review?

gazarenkov (commented):
checkServerReadiness should not be bound to Installers (Agents) since servers may be provided directly by workspace configuration.
I think you should do it in a course of starting machine but after Installers start.

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerInternalRuntime.java
```diff
@@ -223,6 +267,79 @@ private void startAgents(String machineName, DockerMachine dockerMachine) throws
...
+            url = UriBuilder.fromUri(serverUrl)
+                            .replacePath(livenessCheckPath)
+                            .build()
+                            .toURL();
+        } catch (MalformedURLException e) {
+            // server URL is invalid, should not happen, skip server
+            LOG.error(e.getLocalizedMessage(), e);
+            return;
+        }
+        // max start time 60 seconds
```
akorneta: fix comment to 3 minute

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerInternalRuntime.java
```diff
@@ -54,30 +68,38 @@
 
     private static final Logger LOG = getLogger(DockerInternalRuntime.class);
 
+    private static Map<String, String> livenessChecksPaths = new HashMap<>();
```
sleshchenko: Mb `ImmutableMap.of(...)` would be better here?

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerInternalRuntime.java
```diff
@@ -223,6 +267,79 @@ private void startAgents(String machineName, DockerMachine dockerMachine) throws
...
+        }
+        String livenessCheckPath = livenessChecksPaths.get(serverRef);
+        URL url;
+        try {
+            url = UriBuilder.fromUri(serverUrl)
+                            .replacePath(livenessCheckPath)
+                            .build()
+                            .toURL();
+        } catch (MalformedURLException e) {
+            // server URL is invalid, should not happen, skip server
```
voievodin: _skip server_ in the context of _checkServerReadiness_ method doesn't look good, if it's invalid why not propagate the exception ? If it should not happen why not propagate is as RuntimeException?

garagatyi: I'm not going to leave it here forever, but I want to make it more tolerant to mistakes while we are hard coding checks and haven't defined correct behavior in different situations.

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerInternalRuntime.java
```diff
@@ -378,8 +474,11 @@ public StartSynchronizer() {
             this.machines = new HashMap<>();
         }
 
-        public synchronized Map<String, ? extends Machine> getMachines() {
-            return Collections.unmodifiableMap(machines);
+        public synchronized Map<String, ? extends DockerMachine> getMachines() {
+            if (machines != null) {
+                return Collections.unmodifiableMap(machines);
```
sleshchenko: This method is used only in DockerInternalRuntime, so I think we can skip wrapping machines to unmodifiableMap. BTW I'd inline this with ternary operator :smile:

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerMachineStarter.java
```diff
@@ -0,0 +1,774 @@
...
+public class DockerMachineStarter {
+    private static final Logger LOG                     = getLogger(DockerMachineStarter.class);
+    /**
+     * Prefix of image repository, used to identify that the image is a machine saved to snapshot.
+     */
+    public static final  String MACHINE_SNAPSHOT_PREFIX = "machine_snapshot_";
+
+    public static final Pattern SNAPSHOT_LOCATION_PATTERN = Pattern.compile("(.+/)?" + MACHINE_SNAPSHOT_PREFIX + ".+");
+
+    static final   String            CONTAINER_EXITED_ERROR = "We detected that a machine exited unexpectedly. " +
```
sleshchenko: Strange formatting here, please check that it is OK. Also maybe it can be private.

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerRuntimeContext.java
```diff
@@ -48,46 +42,26 @@
...
-                                NetworkLifecycle dockerNetworkLifecycle,
-                                MachineStarter serviceStarter,
-                                URLRewriter urlRewriter,
                                 AgentSorter agentSorter,
                                 AgentRegistry agentRegistry,
-                                ContextsStorage contextsStorage,
-                                SnapshotDao snapshotDao,
-                                DockerRegistryClient dockerRegistryClient,
-                                EventService eventService)
+                                DockerRuntimeFactory dockerInfraFactory)
```
sleshchenko: I suppose `dockerInfraFactory` should be named as `dockerRuntimeFactory`

sleshchenko (approved):
OK for me.

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerInternalRuntime.java
garagatyi: Now exception is thrown here instead. Thank you for the suggestion!

akorneta (approved):

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerInternalRuntime.java
```diff
@@ -223,6 +267,78 @@ private void startAgents(String machineName, DockerMachine dockerMachine) throws
...
+                // TODO protect with lock, from null, from exceptions
+                DockerMachine machine = startSynchronizer.getMachines().get(machineName);
+                machine.setServerStatus(serverRef, ServerStatus.RUNNING);
+                eventService.publish(DtoFactory.newDto(ServerStatusEvent.class)
+                                               .withIdentity(DtoConverter.asDto(identity))
+                                               .withMachineName(machineName)
+                                               .withServerName(serverRef)
+                                               .withStatus(ServerStatus.RUNNING)
+                                               .withServerUrl(serverUrl));
+                LOG.info("Agent {} of machine {} started", serverRef, machineName);
```
voievodin: `LOG.info("Server ...")` i guess

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerInternalRuntime.java
garagatyi: +1

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerInternalRuntime.java
garagatyi: +1

gazarenkov (approved):

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerMachineStarter.java
garagatyi: ok

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerRuntimeContext.java
garagatyi: +1

garagatyi:
Added suggested changes

Issue #4101, opened 2017-02-10
 Rework docker runtime infrastructure to use bootstrapper events during workspace start


skabashnyuk:
Events specification is [here](https://github.com/eclipse/che/issues/4096#issuecomment-283067971)(section _Bootstrapper events_).

Rework docker infrastructure to use bootstrapper to start agents.
The golang bootstrapper is merged under https://github.com/eclipse/che/pull/5399.

voievodin:
Depends on #5266

voievodin:
@sleshchenko 
As bootstrapper is already part of the SPI branch the goal is changed a bit, there is no need to make synthetic implementation while the real bootstrapper can be used instead, i updated issue description.
Also i created a separate [issue](https://github.com/eclipse/che/issues/5413) for removing unused launchers and agents-related functionality.

Issue #5413, opened 2017-06-20
Get rid of java agent launchers


voievodin:
With a [new](https://github.com/eclipse/che/issues/3971) agents launching approach there is no need to keep unused java code. 
The goal of the issue is to review and remove unused agents related java code/modules.
Rename Agent -> Installer.

Depends on #4101

PR #5414, opened 2017-06-20
CHE-4101 Rework docker infrastructure to use bootstrapper to start agents

sleshchenko:
### What does this PR do?
Fixes workspace stopping when an exception occurs on runtime starting
Reworked docker infrastructure to use bootstrapper to start agents.

https://github.com/eclipse/che/issues/4101

#### Release Notes
N/A

#### Docs PR
N/A


gazarenkov (commented):
1/ On terminology (for code and javadocs) : it is just *Installer* not *Agent Installer* , please do not operate the "Agent" term at all in this context. It is a bootstrapper for Installers
2/ I do not think it is a good idea to calculate endpoint like this: 
apiEndpoint.replace("http", "ws").replace("wsmaster/api", "wsmaster") + "/installer/websocket/";
IMO it should be definite configured with parameter "ws" or "wss" URL

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Bootstrapper.java
```diff
@@ -0,0 +1,176 @@
...
+                               " -runtime-id " + String.format("%s:%s:%s", runtimeIdentity.getWorkspaceId(),
+                                                               runtimeIdentity.getEnvName(),
+                                                               runtimeIdentity.getOwner()) +
+                               " -push-endpoint " + endpoint +
+                               " -push-logs-endpoint " + endpoint +
+                               " -file " + BOOTSTRAPPER_BASE_DIR + CONFIG_FILE,
+                               null);
+
+            //waiting for DONE or FAILED bootstrapper status event
+            bootstrappingFinished = bootstraplatch.await(3, TimeUnit.MINUTES);
```
voievodin: Consider using `CompletableFuture<BootstrapperStatusEvent>` instead of latch + volatile `resultEvent`.
The approach would be:
```java
Bootstrapper( ... ) {
    /// ...
   this.finishedFuture = new CompletableFuture<>();
}
```

When `DONE` | `FAILED` event received, completion propagated, e.g.
```java
private class ReleaseFutureSubscriber implements EventSubscriber<BootstrapperStatusEvent> {
    @Override
    public void onEvent(BootstrapperStatusEvent event) {
        BootstrapperStatus status = event.getStatus();
        //skip starting status event
        if (status.equals(BootstrapperStatus.DONE) || status.equals(BootstrapperStatus.FAILED)) {
           //check boostrapper belongs to current runtime and machine
            RuntimeIdentityDto runtimeId = event.getRuntimeId();
            if (event.getMachineName().equals(machineName)
                && runtimeIdentity.getEnvName().equals(runtimeId.getEnvName())
                && runtimeIdentity.getOwner().equals(runtimeId.getOwner())
                && runtimeIdentity.getWorkspaceId().equals(runtimeId.getWorkspaceId())) {

                finishedFuture.complete(bootstrapperEvent); // <- future completion
                eventService.unsubscribe(this); // <- self unsubscribe
            }
        }
    }
}
```
and then instead of calling `bootstraplatch.await(3, TimeUnit.MINUTES);` call
```java
BootstrapperStatusEvent event = finishedFuture.get(3, TimeUnit.MINUTES);
```
This approach simplifies safe result event publication from event subscriber to its client.

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Bootstrapper.java
```diff
@@ -0,0 +1,176 @@
...
+     */
+    public void bootstrap(List<AgentImpl> agents) throws InfrastructureException {
+        injectBootstrapper(agents);
+
+        this.eventService.subscribe(bootstrapperStatusListener, BootstrapperStatusEvent.class);
+        boolean bootstrappingFinished;
+        try {
+            String endpoint = wsIntallerEndpoint + endpointId.getAndIncrement();
+            dockerMachine.exec(BOOTSTRAPPER_BASE_DIR + BOOTSTRAPPER_FILE +
+                               " -machine-name " + machineName +
```
voievodin: Have you considered creating a command line builder to build this exec?

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Bootstrapper.java
```diff
@@ -0,0 +1,176 @@
...
+                        @Assisted DockerMachine dockerMachine,
+                        EventService eventService,
+                        @Named("che.workspace.che_server_endpoint") String apiEndpoint) {
+        this.eventService = eventService;
+        this.machineName = machineName;
+        this.runtimeIdentity = runtimeIdentity;
+        this.dockerMachine = dockerMachine;
+        this.bootstraplatch = new CountDownLatch(1);
+        this.endpointId = new AtomicInteger();
+        this.wsIntallerEndpoint = apiEndpoint.replace("http", "ws").replace("wsmaster/api", "wsmaster")
```
voievodin: I think it makes sense to create a separate property for websocket endpoint, i guess i saw similar code before.
Consider 
- separate property 
- dedicated [qualifier](https://github.com/google/guice/wiki/BindingAnnotations), so this property can be used like `@WebSocketEndpoint String wsEndpoint`

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Bootstrapper.java
sleshchenko: Yes, I have. But I prefer to leave it without builder. https://gist.github.com/sleshchenko/b6bcfccb6a1dff207265a67dd8a6d916
Or maybe you meant something different?

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Bootstrapper.java
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    private void injectBootstrapper(List<AgentImpl> agents) throws InfrastructureException {
+        dockerMachine.putResource("/home/user/che", Thread.currentThread()
+                                                          .getContextClassLoader()
+                                                          .getResourceAsStream("bootstrapper.tar.gz"));
+        // inject config file
+        File configFileArchive = null;
+        try {
+            configFileArchive = createArchive(CONFIG_FILE, new Gson().toJson(agents));
```
garagatyi: Consider making Gson instance static.

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Bootstrapper.java
```diff
@@ -0,0 +1,176 @@
...
+        Path bootstrapperConfTmp = null;
+        try {
+            bootstrapperConfTmp = Files.createTempDirectory(filename);
+            Path configFile = bootstrapperConfTmp.resolve(filename);
+            Files.copy(new ByteArrayInputStream(content.getBytes()), configFile);
+            Path bootstrapperConfArchive = Files.createTempFile(filename, ".tar.gz");
+            TarUtils.tarFiles(bootstrapperConfArchive.toFile(), configFile.toFile());
+            return bootstrapperConfArchive.toFile();
+        } catch (IOException e) {
+            throw new InfrastructureException("Error occurred while injecting bootstrapping conf. " + e.getMessage(),
```
garagatyi: Consider INternalInfrastructureException here

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Bootstrapper.java
```diff
@@ -0,0 +1,176 @@
...
+        dockerMachine.putResource("/home/user/che", Thread.currentThread()
+                                                          .getContextClassLoader()
+                                                          .getResourceAsStream("bootstrapper.tar.gz"));
+        // inject config file
+        File configFileArchive = null;
+        try {
+            configFileArchive = createArchive(CONFIG_FILE, new Gson().toJson(agents));
+            dockerMachine.putResource(BOOTSTRAPPER_BASE_DIR, new FileInputStream(configFileArchive));
+        } catch (FileNotFoundException e) {
+            throw new InfrastructureException(e.getMessage(), e);
```
garagatyi: Consider INternalInfrastructureException here

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/BootstrapperLogger.java
```diff
@@ -0,0 +1,73 @@
...
+import org.eclipse.che.api.workspace.shared.dto.event.InstallerLogEvent;
+import org.eclipse.che.api.workspace.shared.dto.event.InstallerStatusEvent;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.inject.Inject;
+
+/**
+ * It is needed only for debugging.
+ * //TODO Remove before merge.
```
garagatyi: Let's remove?!

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerMachine.java
```diff
@@ -148,6 +150,14 @@ void setServerStatus(String serverRef, ServerStatus status) {
         server.setStatus(status);
     }
 
+    public void putResource(String targetPath, InputStream sourceStream) throws InfrastructureException {
```
garagatyi, line 153: Consider making it package private.

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerMachine.java
```diff
@@ -148,6 +150,14 @@ void setServerStatus(String serverRef, ServerStatus status) {
...
+        try {
+            docker.putResource(PutResourceParams.create(container, targetPath, sourceStream));
+        } catch (IOException e) {
+            throw new InfrastructureException(e.getMessage(), e);
```
garagatyi: Maybe INternalInfrastructureException?

sleshchenko: exec method throws InfrastructureException it's why I decided to throw InfrastructureException here.
Is InternalInfrastructureException better here?

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerMachine.java
sleshchenko: Why should it be package private while `destroy` and `exec` are public?

garagatyi: I don't say destroy and exec should be public )

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerMachine.java
garagatyi: It is supposed that exception that is probable and expected and say nothing helpful to admin is InfrastructureException, whereas InternalInfrastructureException should take attention of admin as unusual thing that may identify some problem in product

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/BootstrapperLogger.java
sleshchenko: A bit later :smile:

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerMachine.java
sleshchenko: I understand, thanks for the explanation. I'll fix it.

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerMachine.java
sleshchenko: I'd leave it as it is, because I prefer to have the same approach. And there are a lot of classes methods which could have package private visibility (like DockerMachineStarter, DockerMachineStarter etc) but there don't have.

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Bootstrapper.java
sleshchenko: Added named property configured by che.properties

garagatyi (approved):
Looks good!

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Bootstrapper.java
```diff
@@ -0,0 +1,176 @@
...
+                        @Assisted DockerMachine dockerMachine,
+                        EventService eventService,
+                        @Named("che.workspace.che_server_endpoint") String apiEndpoint) {
+        this.eventService = eventService;
+        this.machineName = machineName;
+        this.runtimeIdentity = runtimeIdentity;
+        this.dockerMachine = dockerMachine;
+        this.endpointId = new AtomicInteger();
+        //TODO Add property
+        this.wsIntallerEndpoint = apiEndpoint.replace("http", "ws").replace("wsmaster/api", "wsmaster")
```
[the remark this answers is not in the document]
sleshchenko: Move value of websocket-endpoint-base to configuration

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Bootstrapper.java
sleshchenko: Fixed

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Bootstrapper.java
sleshchenko: Make sens. Fixed

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Bootstrapper.java
voievodin: No, i meant pretty the same thing, i think it can be less verbose e.g.
```java
dockerMachine.exec(BootstrapperCmd.binary(BOOTSTRAPPER_DIR + BOOTSTRAPPER_FILE)
                                  .file(BOOTSTRAPPER_DIR + CONFIG_FILE)
                                  .machine(machineName)
                                  .runtime(runtimeIdentity)
                                  .pushEndpoint(endpoint)
                                  .pushLogsEndpoint(endpoint)
                                  .get(), null);

```

Where bootstrapper cmd can be something like [this](https://gist.github.com/evoevodin/edec8ba1172c474b753b2f996b284284).

Up to you whether to change it or not.

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Bootstrapper.java
```diff
@@ -0,0 +1,185 @@
...
+            throw new IllegalStateException("Bootstrap method must be called only once.");
+        }
+        this.finishEventFuture = new CompletableFuture<>();
+
+        injectBootstrapper(agents);
+
+        this.eventService.subscribe(bootstrapperStatusListener, BootstrapperStatusEvent.class);
+        try {
+            String endpoint = wsIntallerEndpoint + endpointId.getAndIncrement();
+            dockerMachine.exec(BOOTSTRAPPER_DIR + BOOTSTRAPPER_FILE +
```
voievodin: What about other arguments like `-server-check-period` and `-installer-timeout` do you think it makes sense to make configuration properties for them as well?

sleshchenko: Nice catch. I missed these parameters. I'll add configuration properties for them.

sleshchenko:
@gazarenkov 
> 1/ On terminology (for code and javadocs) : it is just Installer not Agent Installer, please do not operate the "Agent" term at all in this context. It is a bootstrapper for Installers

Java docs are fixed.
Required code changes(like renaming agent class to installer) will be done within https://github.com/eclipse/che/issues/5413.

> 2/ I do not think it is a good idea to calculate endpoint like this:
apiEndpoint.replace("http", "ws").replace("wsmaster/api", "wsmaster") + "/installer/websocket/";
IMO it should be definite configured with parameter "ws" or "wss" URL

I agree with you. Added configuration parameter [che.workspace.che_server_ws_endpoint_base](https://github.com/eclipse/che/pull/5414/files#diff-5bbadd86c05d20f13241983c27be08b2R107) for configuring base of wsmaster websocket endpoints that are accessible from docker machines.

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Bootstrapper.java
```diff
@@ -0,0 +1,193 @@
...
+            if (resultEvent.getStatus().equals(BootstrapperStatus.FAILED)) {
+                throw new InfrastructureException(resultEvent.getError());
+            }
+        } catch (ExecutionException e) {
+            throw new InfrastructureException(e.getCause().getMessage(), e);
+        } catch (TimeoutException e) {
+            throw new InfrastructureException("Bootstrapping of machine " + machineName + " reached timeout");
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new InfrastructureException("Bootstrapping of machine " + machineName + "was interrupted");
```
voievodin: space before "was"

voievodin (approved):

gazarenkov (approved):
The only general comment regarding using "ws" meaning "web socket". Note, "ws" also applicable for "workspace". Should we consider to be more explicit in naming, like "websocket", "workspace" instead of "ws"?

PR #5156, opened 2017-05-22
Ability to register language servers for file names instead of extensions #5107

tsmaeder:
### What does this PR do?

This PR allows language server plugins to register language servers for specific file names (for example pom.xml). Until now, it is only possible to register for filename extension (like *.xml). 
The PR introduces two major changes:
1. Multiple extensions/or and file names can be mapped to a single mime type and language id. This solves the problem that the mime type was only registered for the first extension when multiples were given.
2. Language registration and language server launcher registration are separated. This way, we can still mimic the old behavior and register multiple mime types for the same language id. As an additional advantage, this change will be useful when allowing to register multiple language servers for the same language in the future.

### What issues does this PR fix or reference?
fixes https://github.com/eclipse/che/issues/5107

#### Changelog
Ability to register language servers for file names instead of extensions
#### Release Notes
Language Server Support: Add ability to register language servers for file names instead of extensions

#### Docs PR
https://github.com/eclipse/che-docs/pull/230

Issue #4609, opened 2017-03-29
Support for Multiple LSP Languages Servers per File


tsmaeder:
Currently, LSP support assumes that only a single language server can be active for any given file. However, this is makes it impossible to do add-on language servers. For an example, think of a linting plugin for typescript for example. Even if the main language server plugin has no support for a particular linting tool, we could make the linting tool contribute diagnostics and code actions to show/fix warnings from the linter.
Visual Studio Code, for example, supports this (so it IS possible) and describes a set of rules on how to handle the cases where multiple language servers are claiming responsibility for a file, see https://code.visualstudio.com/docs/extensionAPI/vscode-api#_languages.

tsmaeder:
I have made a timeboxed proof of concept implementation and which is based on the following main concepts:

1. Languages and LanguageServers are separated. Languages are configured per file extension and determine the language id for a 
2. A language server registers a description that can match a combination of file paths (via a regular expression) or language ids (determined by matching a language description). When multiple language servers are selected for a given file, they are asked for results (and the results merged) via a set of rules based on a match score (see https://code.visualstudio.com/docs/extensionAPI/vscode-api#_)
3. The selection of language servers to ask for a particular request and the merging of the result is done in the workspace agent. No IDE plugins are required to add a new language server.
4. Language servers are started lazily and asynchronously. Each service call uses the currently available language servers for processing. Since we may start multiple language servers when opening a file, waiting for the startup becomes less attractive. Server capabilities for a file are determined using the servers currently online. (see also https://github.com/eclipse/che/issues/2525). Servers are always initialized with the projects directory as the root. Currently, servers are started per project, but I don't think that makes sense, for example for Java workspaces. If a language server cannot handle multiple projects, that failure must be handled inside the server-specific workspace agent plugin.

I have dumped the code here: https://github.com/tsmaeder/che/tree/multiple_language_servers. I would call the state of the code "barely working". I think it can show that the concept will work, but due to bugs and incompleteness, you can only just do a demo with it.
I think with another 2 weeks of work, I could make this into a working PR.

tsmaeder:
For reference, the investigation was originally done here: https://issues.jboss.org/browse/CHE-173

tsmaeder:
Pictures, or it didn't happen: https://www.youtube.com/watch?v=tNQ3-M2_KVA&feature=youtu.be

tsmaeder:
Merged.

PR #5442, opened 2017-06-22
Add support for using multiple language servers to be registered for the same files #4609

tsmaeder:
### What does this PR do?
This PR allows starting multiple language servers per file. The results for various calls are merged according to https://code.visualstudio.com/docs/extensionAPI/vscode-api. 
Opening an editor blocks until all necessary editors have started. The merged server capabilities for all the applicable servers are computed whenever an editor is opened. One copy of a language server is started per project.

This PR is based on https://github.com/eclipse/che/pull/5156, which is not merged yet.

https://github.com/eclipse/che/issues/4609

#### Changelog
Added support for using multiple language servers to be registered for the same files.

#### Release Notes

The support of the Language Server Protocol is getting better and better in Eclipse Che. Two main aspects have been improved: the ability to register a [language server for a specific filename](https://github.com/eclipse/che/pull/5156) and the ability to [support multiple language servers for a single file](https://github.com/eclipse/che/pull/5442).

Until now, it was only possible to register a language server for a file extension like `*.xml`, but now it is possible to register a language server for a specific filename like  `pom.xml`. The improvements also allow for multiple extensions/or and filenames to be mapped to a single mime type and language id. The language registration and language server launcher registration have also been separated. You can learn more into the [following documentation](https://www.eclipse.org/che/docs/assemblies/sdk-language-server-protocol/index.html).

This version also allows multiple language servers to be used for a single file. When opening a file in an editor, the language servers registered for that file are initialized and started if needed. The results coming from the various language servers are computed together. 

You can see a sample demo here in a python file, where Python language server and another demo language server are both registered and working together. 

27958182-1fe7b30e-6322-11e7-9ecc-fef98ff2c04d

Those fundamental improvements on how language servers are supported in Che are going to provide more flexibility for contributors and more capabilities for users. As an example, it is now possible to provide support for a `pom.xml` file, where a Maven language server and an XML language server are both registered and operating together. 

#### Docs PR
https://github.com/eclipse/che-docs/pull/256


ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/annotation/AnnotationModelImpl.java
```diff
@@ -70,6 +71,7 @@ protected void addAnnotation(final Annotation annotation, final Position positio
         try {
             addPosition(position);
         } catch (BadLocationException ignore) {
+            Log.error(getClass(), "BadLocation for "+annotation);
```
dkulieshov: If I'm not mistaken this line is not properly formatted

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/quickassist/LanguageServerQuickAssistProcessor.java
```diff
@@ -12,7 +12,7 @@
 
 import com.google.gwt.user.client.rpc.AsyncCallback;
 import com.google.gwt.user.client.ui.Widget;
-
+import org.eclipse.che.api.languageserver.shared.dto.DtoClientImpls.CodeActionParamsDto;
```
vparfonov, line 15: Why you use here implementations instead of interfaces?

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/DiagnosticCollector.java
```diff
@@ -22,18 +22,18 @@
      * @param diagnostic
      *         Diagnostic - The discovered diagnostic.
      */
-    void acceptDiagnostic(Diagnostic diagnostic);
+    void acceptDiagnostic(String diagnosticsCollection, Diagnostic diagnostic);
```
evidolob, line 26: Can you add description what `diagnosticsCollection` parameter mean

slemeur:
Nice @tsmaeder !

This is really good, we need to write a nice release note paragraph for that enhancement! 
I think that we should record a small demo for that. Would you be able to record an animated gif? It would perfectly illustrate the use case and the new capability.

I'll provide the release note summary.

Thanks

slemeur:
Hi @tsmaeder. Could you record a small animated gif that we will include in the release note to demo the new capability? (Don't forget to zoom-in on your browser so that it will be visible to the reader); 

Thanks in advance

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/quickassist/LanguageServerQuickAssistProcessor.java
tsmaeder: lsp4j does not use interfaces

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/DiagnosticCollector.java
tsmaeder: yes, not clear

tsmaeder:
@benoitf this change should not be visible to the user until we start USING multiple language servers/file. Does it make sense to put it in the release notes?

benoitf:
@tsmaeder are you sure the question is for me or it's for @slemeur asking for a gif demo ?

slemeur:
@tsmaeder : This issue + #5156 are adding new capability to the system - and they are pretty excellent ! I'll handle the release note section but its goal will not be on the end-user benefits, it will focus on explaining the new capability added to the system and what it will allow to do in the future - this could be interesting for any Che contributors. 

When we will leverage the capability, like `pom.xml` file for example, we will focus the explanations on the what kind of features the user will get.

The animated gif might not be necessary, but as you did a demo during the sprint demo, I thought it could nicely illustrate the explanations.

vparfonov (approved):

tsmaeder:
demo

tsmaeder:
Longer demo video available here: https://youtu.be/-G3WlGskIvs

slemeur:
Thanks a lot @tsmaeder !

slemeur (approved):
:+1:

evidolob (approved):

dkulieshov (approved):

benoitf:
ci-build

Issue #5266, opened 2017-06-02
Review and define agent bootstrapper events


voievodin:
Agent installation [spec](https://github.com/eclipse/che/issues/4096#issuecomment-283067971) defines agent bootstrapper events(see _Agent lifecycle events_).
According to [new generation SPI](https://github.com/eclipse/che/issues/4736), defined events - do not provide enough information. For example each agent event should provide machine identifier as well as its own identifier, it can be either structured id like
```
"workspaceId" : "workspace123",
"environment" : "main",
"machine" : "dev"
```
or it can be a path like 
```
"machine" : "workpsace123/main/dev"
```
or it can be a machine id
```
"machine" : "machine123"
```
Another point is that the format of events names should fit the concept of workspace master events.

The DOD of the issue is to provide a list of jsonrpc2.0 events needed to be produced by bootstrapper, 
which allows to develop java client & installer separately.

voievodin:
I reviewed bootstrapper events, after they approved here i will change the spec.

### Bootstrapper events

- _Bootstrapper_ is ready to push events. Occurs when bootstrapper is configured to push events on start, it connects to the specified websocket url(workspace master probably) and publishes this event before any other event.
```
{
  "jsonrpc": "2.0",
  "method": "bootstrapper/statusChanged",
  "params": {
    "status": "READY",
    "machineName": "my-machine",
    "runtimeId": {
      "workspace": "workspace123",
      "environment": "default",
      "owner": "user123"
    },
    "time": "2017-02-23T16:22:22.50287146+02:00"
  }
}
```
- Bootstrapping done, everything is started ok.
```
{
  "jsonrpc": "2.0",
  "method": "bootstrapper/statusChanged",
  "params": {
    "status": "DONE",
    "machineName": "my-machine",
    "runtimeId": {
      "workspace": "workspace123",
      "environment": "default",
      "owner": "user123"
    },
    "time": "2017-02-23T16:22:22.50287146+02:00"
  }
}
```
- Bootstrapping failed(when any agent fails or any error occurs), followed by disconnect and `os.Exit(1)`.
```
{
  "jsonrpc": "2.0",
  "method": "bootstrapper/statusChanged",
  "params": {
    "status": "FAILED",
    "error" : "couldn't start agent X",
    "machineName": "my-machine",
    "runtimeId": {
      "workspace": "workspace123",
      "environment": "default",
      "owner": "user123"
    },
    "time": "2017-02-23T16:22:22.50287146+02:00"
[...]

PR #5466, opened 2017-06-26
Add infrastructure output endpoint for messages/logs

### LINKED ISSUE #5102: Implement channel with infrastructure specific messages/logs in Docker SPI implementation.
New SPI adds a concept of Infrastructure channel that provides heavyweight events from infrastructure, such as logs of machines, agents. We need to implement it to be able to use it on the client side.
Logs list:
- agents logs
- machines logs
- docker image pull logs
- docker image build logs
- snapshot (docker push) logs

akorneta:
### What does this PR do?
Added infrastructure JSON-RPC endpoint for messages/logs.
Added MachineLogsLineConsumer publishing logs on the output channel.
Log events:

**MachineLogEvent**
``` json
{
   "jsonrpc":"2.0",
   "method":"machine/log",
   "params":{
      "text":"[DOCKER] latest: Pulling from eclipse/ubuntu_jdk8 ",
      "stream":"STDOUT",
      "time":"2017-06-27T17:11:09.306+03:00",
      "runtimeId":{
         "envName":"default",
         "owner":"che",
         "workspaceId":"workspacehay1ymbik5v47noq"
      },
      "machineName":"dev-machine"
   }
}
```
**InstallerLogEvent** from [spec](https://github.com/eclipse/che/issues/4096)

Websocket endpoint:
_ws://${host}/wsmaster/output/websocket/${endpoint-id}_
resolves #5102 

Note that these changes are going to be merged into SPI branch.


skabashnyuk:
Can you describe how it works?

garagatyi:
Installer log event also contains the time and stream fields. Have you considered the addition of them to machine log event?

akorneta:
@skabashnyuk 
Machine logs goes directly to subscribers via `RemoteEventService` because for imitatiion of transmission logs we need one more endpoint on wsmaster(there is no sence to add one more endpoint only for transmittion imitation).
Installer logs publishes from `bootstrapper` endpoint on websocket endpoint: _ws://${host}/wsmaster/output/websocket/${endpoint-id}_ and then goes to subscribers via `RemoteEventService`.

akorneta:
@garagatyi Yes, I considered this option, and I think that for clients would be easier to use the formatted output as it was previously. But we still can discuss it.

voievodin:
+1 it makes sense to add _time + stream_ to machine log event.
Please note that [RFC3339(Nanos)](https://tools.ietf.org/html/rfc3339#section-5.8) time format is used for installer/log events.

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/output/OutputService.java
```diff
@@ -0,0 +1,69 @@
...
+import org.eclipse.che.api.core.notification.RemoteEventService;
+import org.eclipse.che.api.workspace.shared.dto.event.InstallerLogEvent;
+import org.eclipse.che.api.workspace.shared.dto.event.MachineLogEvent;
+
+import javax.annotation.PostConstruct;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+
+/**
+ * Service for handling output.
```
garagatyi: Since service and handling are pretty common word javadocs are not very clear to me. Can you describe more precisely the purpose of the class in javadocs?

akorneta: sure

dkulieshov:
Do I understand it right that you suggest to send log events bypassing the `EventService` using another `RemoteEventService` instead, so we have to event services running at a time? If yes than it seems a bit odd to me that we use one event service for one event and another for another, are we going to create a specific event service for each new event? Could you please describe your idea more as there is nothing in the description.

If I'm not mistaken we have an agreement that all infrastructure events must go through `EventService` and if there is something that bypasses event service than it should not be called an event? I'm very curious about what you think about this.

akorneta:
@dkuleshov 
The main idea of this PR is to send infra logs to clients. We decided that there is no reason to sent log events on _master,_ because on _master_ side there is no interested subscribers in these types of events. Proceeding from this thoughts we added `publish` method to RemoteEventService for publishing to clients directly, not through `EventService`.

I think it is not a problem to send events bypasses `EventService` if sending via `EventService` makes no sense.

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/output/MachineLoggersFactory.java
```diff
@@ -0,0 +1,143 @@
...
+
+        public MachineLogMessageProcessor(BiConsumer<String, String> biConsumer) {
+            this.biConsumer = biConsumer;
+        }
+
+        @Override
+        public void process(LogMessage message) {
+            final LogMessage.Type type = message.getType();
+            if (type == DOCKER) {
+                biConsumer.accept(null, "[DOCKER] " + message.getContent());
```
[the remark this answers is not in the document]
akorneta: I'll add appropriate docs to machine log event.

dkulieshov:
@akorneta 
For me personally, treating log messages as events is a problem itself however this is out of scope of this topic. 

I'm not sure that I follow your point. Are you sure that you are accurate with terminology? Do you mean "publish log events" instead of "send log events"? You're expecting to call `publish` method of `RemoteEventService` so in your case both infrastructure and master are expected to be the same application, so I don't think it is appropriate here to distinguish master and infrastructure sides. From this point of view it is senseless to say "there is no reason to sent log events on master" because `EventService.publish` and `RemoteEventService.publish` will be executed in the same application. With the same success we can say that RemoteEventService.publish is for master side also. `RemoteSubscriptionManager`/`RemoteEventService` is a component of master, so you're interacting with master anyway.

I think we should have a single instance of event service and if you want to bypass event service you must implement a separate mechanism and must not name those messages as events. For me it is simple and clear: if you want to publish an event use single existing event service, if you want to publish something else, implement something else.

garagatyi:
I agree with Dmytro that it is better to not duplicate EventService even though its purpose is significantly different. AFAIK we knew that usage of EventService for logs is not a good solution but decided to start from this solution as the simplest one. Then we can figure out how to do this properly.

voievodin:
I think the idea is not to FLOOD to `EventService` with machine log events, which as i can see are
- docker pull
- docker push
- docker build
- docker logs -f

I agree with @akorneta it makes no sense to listen to logs events on master side.

`RuntimeContext` interface provides a [single](https://github.com/eclipse/che/blob/2399e94727dce82886fbce08fb29fe05f2471358/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/RuntimeContext.java#L93) channel for ALL THE OUTPUT, which means
that client might need to distinguish between docker push/pull, build logs, installer logs etc so 
logs have to be structured, right now they are structured as they provide _method + stream + text_.
If they are not events then what?

I measured the count of log events published to event service for the workspace with 1 machine(default java stack) - the number of logs is _825_ - the full log is available [here](https://gist.githubusercontent.com/evoevodin/d590aea4dbafc54203b3a1ca4b455756/raw/bd7fe1709d64008ab9419ac9448cc4fb09b89290/log).
I truly believe that event publication is cheap, but still why should we do uncessary computations?

If `RemoteEventService` is confusing and not the case, then how log events can be delivered to the interested sides bypassing `EventService`?

akorneta:
@dkuleshov @gazarenkov @garagatyi PR reworked. `EventService` used for infrastructure events, please take a look.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/event/MachineLogJsonRpcMessenger.java
```diff
@@ -0,0 +1,47 @@
...
+import org.eclipse.che.api.core.notification.RemoteSubscriptionManager;
+import org.eclipse.che.api.workspace.shared.dto.event.MachineLogEvent;
+
+import javax.annotation.PostConstruct;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import java.util.Map;
+
+/**
+ * Register subscriber on {@link MachineLogEvent installer log event} for resending
```
sleshchenko: copy/paste error `installer log event` -> `machine log event`

sleshchenko (approved):

dkulieshov:
Looks nice!

Could you guys please remind why do we treat logs as events here, because actually I can't figure out for myself why we mix simple logging messages and event, do I miss something?

voievodin:
I guess you missed my [comment](https://github.com/eclipse/che/pull/5466#issuecomment-311924223).

>`RuntimeContext` interface provides a [single](https://github.com/eclipse/che/blob/2399e94727dce82886fbce08fb29fe05f2471358/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/RuntimeContext.java#L93) channel for ALL THE OUTPUT, which means
that client might need to distinguish between docker push/pull, build logs, installer logs etc so 
logs have to be structured, right now they are structured as they provide _method + stream + text_.
If they are not events then what?

dkulieshov:
@evoevodin How is your comment related to mixing messages and events explanation? Do you mean that if logs must be structured they should be treated as events? That does not make sense to me at all, sorry.

voievodin:
@dkuleshov 
Maybe we understand event in a different way.
As far as i understand event in this discusstion is a synonym for  _JSONRPC Notification_.

The flow
```
Raw Message -> LogMessage -> MachineLogEvent -> client
```
Here `MachineLogEvent` is a DTO and it is needed here only because we need to deliver `LogMessage` to the interested CLIENTS. If we change `MachineLogEvent` to `MachineLog` or similar it won't change the flow!

That's why the first version of this PR introduced the way to <s>publish</s> send `LogMessage` to the clients. That's why we wanted to bypass `EventService`.

It's called event i guess because of the analogy with [InstallerLogEvent](https://github.com/eclipse/che/blob/spi/wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/dto/event/InstallerLogEvent.java) which AFAIU was absolutely [fine](https://github.com/eclipse/che/issues/5266) for everyone including you.
image

dkulieshov:
@evoevodin 

> May be we understand event in a different way.
> As far as i understand event in this discusstion is a synonym for JSONRPC Notification.

As of my vision your assumption is correct: for me JSON-RPC notification and events are objects of different non related categories. The difference is approximately the same as between the terms "sentence" and "idea".

> If we change MachineLogEvent to MachineLog or similar it doesn't change the flow!

I would say it is quite arguable statement as according to my understandings events must go through `EventService` while non-events may go through whatever you want (e.g. you can implement some specific messenger/handler pair for your own purposes). 

So back to my first comment: in my opinion it is okay to bypass `EventService` if we're dealing with non-event entities otherwise they should go through the service. I absolutely support your concern of flooding however I don't think we should treat those structured logs as events (in global sense )  when we bypass the event service but we can treat them as JSON-RPC notifications using specific implementation of messenger and handler. 

@gazarenkov  @vparfonov @skabashnyuk  @garagatyi 
Does that make sense for you?

P.S.
I have to admit that the way I analyzed the specification that you provided leaves much to be desired. However I would be grateful if you do not speculate on the facts in the future. The screenshot that you provided shows my reaction to specific comment, while the specification that you refer on - I did not approve it directly.

garagatyi:
I agree that logs should not be treated as events but as JSON-RPC notifications.
I suppose it is a good moment to decide altogether on how to solve this specific use case. 
There is no value in arguing who has approved what. Let's just find a solution and merge the PR. 

So, the problem is:
- how to send logs to clients as JSON-RPC notifications
- how to not flood EventService with logs because they are heavy and they are not events in fact. 
- how to not change the purpose of subscription manager to not make it the second implementation of event service. 

What consequences of these requirements I see:
- not to use EventService at all
- Only `RemoteSubscriptionManager` knows about subscription, so it has to be involved

@akorneta and @evoevodin suggested adding publish method to RemoteSubscriptionManager.
@dkuleshov you disagreed with this decision. You are the author of this class, so you know it much better than we do. Can you suggest another sollution? We would be pleased to hear from you a better approach.

dkulieshov:
We have subscription manager to subscribe to events remotely, I think we should have another similar entity to subscribe to logs remotely, something like this (naming is preliminary): 
- `RemoteLogsSubscriptionManager` on server side to allow client to subscribe directly for logs that are considered to be important, in this case we will need to rename `RemoteSubscriptionManager` to `RemoteEventSubscriptionManager` to distinguish entities
-  `JsonRpcMessenger` on server side to be able to publish logs from server to client
-  `LogsSubscriber` on client side to make it simple to subscribe for needed logs

Technically it would have similar workflow however logically we will have different entities which may suit the business logic more accurately.

Actually I'm not totally disagree with the decision, I'm in the middle of a discussion on that :-)

garagatyi:
Sorry, I didn't get what how Docker infra would send logs to clients with this approach. Can you elaborate on this?

garagatyi:
We discussed this PR with @dkuleshov and agreed that conceptually it would be better to implement JsonRpcMessenger and use it for logs publishing instead of EventService. But for now it is not a problem to use EventService and it won't make a significant additional load to Che master.

garagatyi (approved):

voievodin:
I am +1 for `JsonRpcMessenger` and dedicated JSONRPC Notification, it was something guys wanted to  achieve from start. But i understand that it might need more time than it was estimated while the performance benefit is tiny, so up to the author whether to merge it like this.

dkulieshov (approved):

gazarenkov (approved):

PR #5471, opened 2017-06-27
BugZilla 518578 : Let users to use .Chefile in addition to Chefile

benoitf:
### What does this PR do?

It allows people to have Chefile hidden by default on Unix systems (like many metadata information not related directly to the project)

https://bugs.eclipse.org/bugs/show_bug.cgi?id=518578

#### Changelog
Handle .Chefile filename in addition to Chefile

#### Release Notes
Handle .Chefile filename in addition to Chefile. If both files are present, an error is reported.

#### Docs PR
https://github.com/eclipse/che-docs/pull/248

Change-Id: Ie290a4550ccbdcc9aac64cb0c3bc60ac434a2ed8

### DIFF
--- a/dockerfiles/lib/src/internal/dir/che-dir.ts
+++ b/dockerfiles/lib/src/internal/dir/che-dir.ts
@@ -75,6 +75,11 @@ export class CheDir {
    */
   dotCheFile: string;
 
+  /**
+   * Alternate path to the Chefile file with .chefile
+   */
+  dotCheFileLowercase: string;
+
   dotCheFolder : any;
   workspacesFolder : any;
   cliFolder : any;
@@ -107,6 +112,7 @@ export class CheDir {
     this.folderName = this.path.basename(this.currentFolder);
     this.cheFile = this.path.resolve(this.currentFolder, 'Chefile');
     this.dotCheFile = this.path.resolve(this.currentFolder, '.Chefile');
+    this.dotCheFileLowercase = this.path.resolve(this.currentFolder, '.chefile');
     this.dotCheFolder = this.path.resolve(this.currentFolder, '.che');
     this.dotCheIdFile = this.path.resolve(this.dotCheFolder, 'id');
     this.dotCheSshPrivateKeyFile = this.path.resolve(this.dotCheFolder, 'ssh-key.private');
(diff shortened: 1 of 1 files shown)


slemeur (approved):
:+1:

dockerfiles/lib/src/internal/dir/che-dir.ts
```diff
@@ -201,8 +208,16 @@ export class CheDir {
...
-      Log.getLogger().debug('No chefile defined, use default settings');
-      return;
+
+      // if there is a .Chefile, use it, else return
+      try {
+        this.fs.statSync(this.dotCheFile);
+        Log.getLogger().debug('The alternate file .Chefile is present at ', this.dotCheFile);
+        this.cheFile = this.dotCheFile;
+      } catch (e) {
+        Log.getLogger().debug('No chefile defined, use default settings');
```
apupier: maybe this error message can be improved now:
- chefile --> Chefile
- and maybe add the possibility to use Chefile: "No Chefile or .Chefile defined, use default settings"

apupier (commented):
We talked about throwing an error if Chefile and .Chefile are both provided. I think that it is not implemented in this PR.
Is it by choice? Don't want to do it anymore? or want to provide it in a separate PR?

benoitf:
@apupier it's not implemented by this PR. I thought that maybe it would help if there is always a winner (no error reported). Chefile was always winning over .Chefile. You think that raising an error if the two are available will be more user friendly ? (error case vs winner case)

apupier:
> @apupier it's not implemented by this PR. I thought that maybe it would help if there is always a winner (no error reported). Chefile was always winning over .Chefile. You think that raising an error if the two are available will be more user friendly ? (error case vs winner case)

I think that it is better to have the error case. it avoids the user to start modifying the wrong Chefile
Do we have a third point of view to settle?

Anyway, I will be happy with both cases (error or winner). Just I think the error one is slightly better.

benoitf:
@apupier I will modify the pull request to throw the error.

benoitf:
ok updated

TylerJewell:
I believe that the appropriate choice would be a `.chefile` as opposed to
`.Chefile` to be consistent with other systems and handling of hidden file
names. Is the implementation case sensitive?


Tyler Jewell // Dev Products // 978-884-5355

benoitf:
It is case sensitive (on operating system that supports it)

benoitf:
so should I keep .Chefile or use .chefile ( in addition to Chefie)

TylerJewell:
I would prefer `.chefile` as the only option. However, I don't see the harm in supporting both `.chefile` and `.Chefile`.

apupier:
from an IDE point of view, the filters are on .* names so there is no difference.

sunix (approved):
LGTM

riuvshin (approved):

benoitf:
I've added support then to Chefile, .chefile and .Chefile

Issue #5316, opened 2017-06-09
Add ability to start tests for TestNG in debug mode


vparfonov:
We partly fix problem for starting test in #4679, now we need to add possibility for start test in debug mode

svor:
The next step is https://github.com/eclipse/che/issues/5316

PR #5472, opened 2017-06-27
New runner for TestNG tests

svor:
### What does this PR do?
Added common interface which helps to add new actions for other test frameworks. 
Improves a mechanism for running TestNG tests. New feature:

- Added an ability to run only one test method from the test class

- Added an ability to debug TestNG testes

anim

https://github.com/eclipse/che/issues/5316

#### Changelog

- Run TestNg tests in the debug mode
- Run only one test method or all methods from the TestNg test class

#### Release Notes
In this version we have been refactoring the foundations which allow to provide test capabilities over the various popular test frameworks. We added default interfaces and mecanism which will allow to bring better test capabilities inside of Eclipse Che. We have also reworked the TestNG runner in order to use those interfaces. You now get extended capabilities with TestNG and soon with other test frameworks we already integrated (JUnit, PHPUnit):
- Ability to run TestNg tests in the debug mode
- Ability to run only one test method or all methods from the TestNg test class

anim


core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/execution/ParametersList.java
```diff
@@ -0,0 +1,30 @@
...
+ *******************************************************************************/
+package org.eclipse.che.commons.lang.execution;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Represent and configure program parameters
+ */
+public class ParametersList {
```
dkulieshov, line 21: I'm curious why do we need a list instead of map for parameters?

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/execution/ParametersList.java
```diff
@@ -0,0 +1,30 @@
...
+    private final List<String> parameters = new ArrayList<>();
+
+    public void add(String name, String value) {
+        parameters.add(name);
+        parameters.add(value);
+    }
+
+    public List<String> getParameters() {
+        return new ArrayList<>(parameters);
```
dkulieshov: I'm not sure but shouldn't we return [unmodifiable list](https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#unmodifiableList(java.util.List)) here?

slemeur:
This is a new feature. You need PM approval + Release note information for that issue. 

Could you please attach an animated gif (and zooming in it so that we can include it in release note)

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -53,37 +52,13 @@
...
-import org.eclipse.che.plugin.docker.client.params.RemoveContainerParams;
-import org.eclipse.che.plugin.docker.client.params.RemoveImageParams;
-import org.eclipse.che.plugin.docker.client.params.network.RemoveNetworkParams;
-import org.eclipse.che.plugin.docker.client.params.StartContainerParams;
-import org.eclipse.che.plugin.docker.client.params.StartExecParams;
-import org.eclipse.che.plugin.docker.client.params.StopContainerParams;
-import org.eclipse.che.plugin.docker.client.params.TagParams;
-import org.eclipse.che.plugin.docker.client.params.TopParams;
-import org.eclipse.che.plugin.docker.client.params.WaitContainerParams;
+import org.eclipse.che.plugin.docker.client.params.*;
```
benoitf: AFAIK wildcard import is not allowed

plugins/plugin-docker/che-plugin-docker-client/src/test/java/org/eclipse/che/plugin/docker/client/DockerConnectorTest.java
```diff
@@ -56,37 +55,13 @@
...
-import org.eclipse.che.plugin.docker.client.params.RemoveContainerParams;
-import org.eclipse.che.plugin.docker.client.params.RemoveImageParams;
-import org.eclipse.che.plugin.docker.client.params.network.RemoveNetworkParams;
-import org.eclipse.che.plugin.docker.client.params.StartContainerParams;
-import org.eclipse.che.plugin.docker.client.params.StartExecParams;
-import org.eclipse.che.plugin.docker.client.params.StopContainerParams;
-import org.eclipse.che.plugin.docker.client.params.TagParams;
-import org.eclipse.che.plugin.docker.client.params.TopParams;
-import org.eclipse.che.plugin.docker.client.params.WaitContainerParams;
+import org.eclipse.che.plugin.docker.client.params.*;
```
benoitf: another wildcard import

plugins/plugin-maven/che-plugin-maven-server/src/test/java/org/eclipse/che/plugin/maven/server/classpath/OutputPathTest.java
```diff
@@ -0,0 +1,233 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+
+import static org.fest.assertions.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+/**
+ *
```
benoitf: it seems the text is missing there

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/execution/ParametersList.java
evidolob: Because java.lang.ProcessBuilder constructor requires list of the parameters and  it's easier to handle as List

plugins/plugin-testing-java/plugin-testing-classpath/src/main/java/org/eclipse/che/plugin/java/testing/AbstractJavaTestRunner.java
```diff
@@ -0,0 +1,143 @@
...
+            throw new RuntimeException("Can't find Compilation Unit.", e);
+        }
+    }
+
+    @Override
+    public int getDebugPort() {
+        return -1;
+    }
+
+    private RuntimeException getRuntimeException(String filePath) {
```
vparfonov, line 140: What target of this method ?

plugins/plugin-testing-java/plugin-testing-testng/che-plugin-testing-testng-runtime/src/main/java/org/testng/CheTestNG.java
```diff
@@ -0,0 +1,72 @@
...
+ * Copyright (c) 2012-2017 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.testng;
```
vparfonov, line 11: @evidolob  Why we have `org.testng` package here?

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/execution/ParametersList.java
svor: I've changed it. Thank you for the remark.

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
svor: Fixed

plugins/plugin-docker/che-plugin-docker-client/src/test/java/org/eclipse/che/plugin/docker/client/DockerConnectorTest.java
svor: Fixed

plugins/plugin-testing-java/plugin-testing-testng/che-plugin-testing-testng-runtime/src/main/java/org/testng/TestingMessageHelper.java
```diff
@@ -0,0 +1,268 @@
...
+            case '\u2028':
+                return 'l';
+            case '\u2029':
+                return 'p';
+            default:
+                return 0;
+        }
+    }
+
+    private static class Pair {
```
vparfonov, line 259: Maybe here you can use org.eclipse.che.commons.lang.Pair?

plugins/plugin-testing-java/plugin-testing-testng/che-plugin-testing-testng-runtime/src/main/java/org/testng/CheTestNG.java
evidolob: We need some package private method

benoitf: is it legal per Eclipse rules to define packages outside `org.eclipse.che` namespace ?

plugins/plugin-testing-java/plugin-testing-classpath/src/main/java/org/eclipse/che/plugin/java/testing/AbstractJavaTestRunner.java
svor: Method returns RuntimeException when have a problem to find java element by path (package or compilation unit). It is an inner private method.

plugins/plugin-testing-java/plugin-testing-testng/che-plugin-testing-testng-runtime/src/main/java/org/testng/TestingMessageHelper.java
evidolob: I don't sink so, this is runtime part of TestNG JVM, I want to keep less external dependencies inside this module as possible

slemeur (approved):

plugins/plugin-testing-java/plugin-testing-testng/che-plugin-testing-testng-server/src/main/java/org/eclipse/che/plugin/testing/testng/server/TestNGRunner.java
```diff
@@ -40,8 +71,114 @@ public TestResult execute(Map<String, String> testParameters) throws Exception {
...
+    private ProcessHandler startTestProcess(IJavaProject javaProject, TestExecutionContext context) {
+        File suiteFile = createSuite(context, javaProject);
+        if (suiteFile == null) {
+            throw new RuntimeException("Can't create TestNG suite xml file.");
+        }
+
+        JavaParameters parameters = new JavaParameters();
+        parameters.setJavaExecutable("java");
+        parameters.setMainClassName("org.testng.CheTestNGLauncher");
+        parameters.setWorkingDirectory("/tmp");
```
davidfestal: Several testing frameworks or test runners (such as arquillian, etc ...) assume that the current working directory should be the root directory of the project.
See [this issue](https://issues.jboss.org/browse/CHE-214?focusedCommentId=13397893&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13397893) as an example.
So it seems that at this line the working directory should be set to the root directory of the project being tested, instead of `/tmp`.
This would also be consistent with the way tests are run inside maven afaik.

davidfestal (changes requested):

plugins/plugin-testing-java/plugin-testing-testng/che-plugin-testing-testng-server/src/main/java/org/eclipse/che/plugin/testing/testng/server/TestNGRunner.java
svor: @davidfestal Each test framework will have own runner where working directory can by set.
In this case working directory is "/tmp" because TestNG framework generates test-output directory which contains some generated files. So, if this dir will be the root directory of the project, we'll see it in the project explorer.

vparfonov: Let;s keep for TestNG runner as is

vparfonov (approved):

evidolob (approved):

dkulieshov (approved):

plugins/plugin-testing-java/plugin-testing-testng/che-plugin-testing-testng-server/src/main/java/org/eclipse/che/plugin/testing/testng/server/TestNGRunner.java
davidfestal: Maybe we just run the corresponding TestNG tests (on an example project) in maven with `mvn test` or simply the standard build command, and compare where the test outputs are generated  ? I assume they will be generated either at the root of the project or maybe inside the `target` directory. In the last case, this would mean that an option certainly exists in testNG to specify the place of the test output directory without changing process working directory to `/tmp`.

IMO we should do that same as what maven does, because as soon as the developer will run the build command, the tests will be run in maven anyway.

davidfestal: More precisely the `-d` command line option allows specifying where the reports will be generated (defaults to test-output).
See http://testng.org/doc/documentation-main.html#running-testng

So it seems to me there should be a solution to both:
- start the test process in the tested project root
- create the test outputs in a place that isn't visible in the explorer.

svor: I see your point. 
So my solution is: 

- Working directory is the project root directory; 
- The directory where the reports will be generated will created in target folder (project output directory) and named test-output if project output directory (for example target) doesn't exist test-output will be created in the project root.

You can see this changes with my last commits.

@davidfestal @vparfonov Is it OK?

davidfestal (approved):

plugins/plugin-testing-java/plugin-testing-testng/che-plugin-testing-testng-server/src/main/java/org/eclipse/che/plugin/testing/testng/server/TestNGRunner.java
davidfestal: @svor OK for me. I approved the PR

vparfonov (approved):

Issue #5215, opened 2017-05-26
Fix Popup for extenal operation file updated.


JamesDrummond:
When I have a log file that is open that get changed often too many popup come up. I propose we notify user of external change in another manner or limit repeats of message. 

![image](https://cloud.githubusercontent.com/assets/20580888/26498099/7173ff16-41f3-11e7-90de-487ee6356f9d.png)

codenvy 5.10.0

davidwindell:
+1 this is really annoying at the moment as we run lots of external scripts, logs, etc. A `.cheignore` might also be an idea to ignore entire files/folders from external operations, etc.

We also seem to be having excessive numbers of these "External operation" messages since 5.11.1 but I'll raise another issue for that.

marciobrst:
+1 any solution?

Anthony-Janin:
Hi I've got the same issue but for all files, event Main.java for example, once I open the file and that I add some text, i've got this popup infinitly....

asacamano:
Just upgraded to  5.14 yesterday - and I saw similar behavior creating a new file and editing it.  Is this the same issue?

Steps to reproduce:

1. Create a workspace with an empty project (e.g. Ubuntu workspace, blank project)
1. Create a new file - docker-compose.yml, or edit the README file.
1. Wait for the file to open in the editor.

Result

An (apparently) endless stream of "External Operation: File docker-compose.tml is updated." notifications.

davidwindell:
Yes, this seems to be the same problem. We're stuck on 5.10 until this is resolved.

slemeur:
@vparfonov + @dkuleshov : Could you please check that behaviour. We don't ave an appropriate fix for that. At least #5495 is not fixing this problem.

Updating the status to blocker.

dkulieshov:
Hi, 

As of what I see here I can say:
- @slemeur according to the description of this issue I don't think we can say that https://github.com/eclipse/che/pull/5495 does not fix the problem
- Many thanks to @asacamano who reported a bug, however it seems that it is not quite related to this specific issue so I created another: https://github.com/eclipse/che/issues/5515
- I also created an issue to improve general external operation notifications behavior: https://github.com/eclipse/che/issues/5514

PR #5495, opened 2017-06-29
CHE-5215. Add ability to exclude tracking files by file watcher

RomanNikitenko:
### What does this PR do?
- Add ability to manage file watcher tracking creation, modification and deletion events.
- Allows to add and remove entries to excludes by 'Add to File Watcher excludes' and 'Remove from File Watcher excludes' actions or by editing projectPath/.che/fileWatcherIgnore file manually
- Change the way of notification about external changes - current behavior: we display popup, new behavior: we display this kind of notifications ONLY in Events panel. We are going to improve this behavior by #5514

#5215 

#### Changelog
Add ability to manage file watcher tracking creation, modification and deletion events.

#### Release Notes
File watchers allow to keep notified about changes on files when they are done from multiple clients. For example, you might be editing files of projects from Che's editor, Che's terminal or any other clients if you are using ssh. The file watcher provide the ability track any creation, modification and deletion on your projects' files. 
In this version, we added the ability to manage the file watcher behaviours. In certain cases, you might not want to keep track of certain files/folders of your projects. We added a `fileWatcherIgnore`which behave similarly to a `.gitignore`. You manage the file yourself by editing the `projectPath/.che/fileWatcherIgnore` or you can do a right click on a file/folder to get the actions 'Add to File Watcher excludes' or 'Remove from File Watcher excludes'.

filewatcherignore

### DIFF
--- a/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/event/ng/FileWatcherExcludesOperation.java
+++ b/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/event/ng/FileWatcherExcludesOperation.java
@@ -1,0 +1,154 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2017 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.ide.api.event.ng;
+
+import com.google.gwt.user.client.rpc.AsyncCallback;
+import com.google.web.bindery.event.shared.EventBus;
+
+import org.eclipse.che.api.core.jsonrpc.commons.JsonRpcPromise;
+import org.eclipse.che.api.core.jsonrpc.commons.RequestHandlerConfigurator;
+import org.eclipse.che.api.core.jsonrpc.commons.RequestTransmitter;
+import org.eclipse.che.api.promises.client.Promise;
+import org.eclipse.che.api.promises.client.PromiseProvider;
+import org.eclipse.che.ide.api.event.WindowActionEvent;
+import org.eclipse.che.ide.api.event.WindowActionHandler;
+
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Tracks and allows to manage the file watcher exclude patterns for tracking creation,
+ * modification and deletion events for corresponding entries.
+ *
+ * @author Roman Nikitenko
+ */
+@Singleton
+public class FileWatcherExcludesOperation implements WindowActionHandler {
+    private static final String ENDPOINT_ID          = "ws-agent";
+    private static final String EXCLUDES_SUBSCRIBE   = "fileWatcher/excludes/subscribe";
+    private static final String EXCLUDES_UNSUBSCRIBE = "fileWatcher/excludes/unsubscribe";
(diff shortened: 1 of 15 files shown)


vparfonov (approved):

vparfonov:
@slemeur I am ok with code but have doubts about usability, need your opinion

dkulieshov (approved):

slemeur:
Thanks @vparfonov for pinging on this.

I have few questions for @RomanNikitenko + @dkuleshov:
- In which case do you think a user will want to remove the watcher for a file? 
- I don't think it fixes #5495. I just think it gives a workaround - but the root cause of the problem is not fixed. The root cause of the problem is that we allow the IDE to display 1000 the same notification for a similar operation.
- Instead of having the user, creating a "fileWatcherExcludes" file, why not having a right click action to "add/remove" to "fileWatcherExcludes", which will handle the creation of the file if it does not exist.
- Will this file "fileWatcherExcludes" be added to "gitignore"?

RomanNikitenko:
@slemeur thank you for your thoughts about this, I'm looking for better solution for this issue, so:
- We hold "fileWatcherExcludes" file in the ".che" directory - so I think it be added to "gitignore" automatically
- Regarding "why not having a right click action" - yes, I thought so too, but I had doubts... So I can add this action, If you have the same opinion
- About "we allow the IDE to display 1000 the same notification for a similar operation"
In the issue "log file that is open that get changed often too many popup come up" case is described, so what is criterion we can use for it? The same notification for a similar operation per 30 sec, one minute, ...? I think we should allow to add entries to excludes and user himself can solve his case.
 BTW, the idea about ignore file is taken from comments to issue. But I still looking for criterion to limit notifications for these operations and I would be grateful for ideas!

slemeur:
@RomanNikitenko Thanks for your clarification.

Let's discuss a bit on the notification itself:
- Is that as important to display that the file has been modified by an external operation? I really wonder if it is needed.
- Is the notification UX we are using here is appropriate? The notification is quite important (= taking a lot of space and take user attention). Would a notification in the editor be more efficient? We could display "This file has been updated from another editing operation". We would display it at first change from an external operation and until the user acknowledge it with a "OK" button?

Otherwise, I'm +1 with the right click action.

dkulieshov:
Hi, 

According to the description of https://github.com/eclipse/che/issues/5215 the problem that @RomanNikitenko fixes here is the ability to manage ignore lists to be able to skip notifications for specific files when they are constantly edited. In our case it is opened log file. However there is a general problem: how to manage those notification in arbitrary situations for arbitrary files when we may have several consequent notifications that eventually result in some kind of a notification flood. 

I created a dedicated issue to find a way we can generally improve notifications behavior https://github.com/eclipse/che/issues/5514 It would be very nice if we could continue the discussion there.

dkulieshov:
@slemeur @RomanNikitenko I agree that right click action is a good idea, but what do you think if we would detect that notification manager begins to flood the IDE with external operation notifications for some file and proactively ask a user if he would like to add that file to ignore list. Does that make sense?

RomanNikitenko:
So within this issue I'm going to add some right click action to have ability to add path of a selected item to "fileWatcherExcludes" file automatically.

davidwindell:
Can a folder be added to the `fileWatcherExcludes` path? Sometimes we have a large number of time-stamped logs with varying names generated.

Something replicating .gitignore or .dockerignore rules would be ideal.

RomanNikitenko:
@davidwindell  Yes, at this moment a folder can be added to the fileWatcherExcludes path. Thank you for your idea about .gitignore/.dockerignore rules - I will check and improve it.

RomanNikitenko:
@vparfonov @dkuleshov @slemeur  I've updated the PR, please review

vparfonov (approved):

slemeur:
Thanks for the update @RomanNikitenko .

slemeur:
Do you confirm that it is possible to exclude a whole folder as well?

RomanNikitenko:
@slemeur yes, it's possible to exclude a whole folder. You can see it on the video.

slemeur (approved):

Issue #5505, opened 2017-06-30
NPE in o.e.c.a.w.s.WorkspaceManager


skryzhny:
Got several ERRORs in logs:
```
2017-06-29 13:46:06,992[aceSharedPool-8]  [ERROR] [o.e.c.a.w.s.WorkspaceManager 780]    - null
org.eclipse.che.api.core.ServerException: null
	at org.eclipse.che.api.environment.server.CheEnvironmentEngine.startInstance(CheEnvironmentEngine.java:1007)
	at org.eclipse.che.api.environment.server.CheEnvironmentEngine.startEnvironmentQueue(CheEnvironmentEngine.java:813)
	at org.eclipse.che.api.environment.server.CheEnvironmentEngine.start(CheEnvironmentEngine.java:268)
	at org.eclipse.che.api.workspace.server.WorkspaceRuntimes.startEnvironmentAndPublishEvents(WorkspaceRuntimes.java:712)
	at org.eclipse.che.api.workspace.server.WorkspaceRuntimes.access$100(WorkspaceRuntimes.java:106)
	at org.eclipse.che.api.workspace.server.WorkspaceRuntimes$StartTask.call(WorkspaceRuntimes.java:977)
	at org.eclipse.che.api.workspace.server.WorkspaceRuntimes$StartTask.call(WorkspaceRuntimes.java:941)
	at org.eclipse.che.commons.lang.concurrent.CopyThreadLocalCallable.call(CopyThreadLocalCallable.java:30)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
Caused by: org.eclipse.che.api.core.ServerException: null
	at org.eclipse.che.plugin.docker.machine.MachineProviderImpl.startService(MachineProviderImpl.java:395)
	at org.eclipse.che.api.environment.server.CheEnvironmentEngine.lambda$startEnvironmentQueue$6(CheEnvironmentEngine.java:785)
	at org.eclipse.che.api.environment.server.CheEnvironmentEngine.startInstance(CheEnvironmentEngine.java:949)
	... 11 common frames omitted
Caused by: java.lang.NullPointerException: null
[...]

PR #5510, opened 2017-06-30
CHE-5505: Fix NPE in DockerConnector

mmorhun:
### What does this PR do?
Fixes NPE in `DockerConnector` when Docker respond with 200 response code and empty stream.

https://github.com/eclipse/che/issues/5505

#### Changelog
Fixed NPE when docker responds with 200 response code and empty stream.

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
+++ b/plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
@@ -1348,7 +1348,8 @@ public class DockerConnector {
         try (InputStreamReader reader = new InputStreamReader(inputStream)) {
             T objectFromJson = GSON.fromJson(reader, clazz);
             if (objectFromJson == null) {
-                throw new IOException("Docker responded with an empty stream.");
+                LOG.error("Docker response doesn't contain any data, though it was expected to contain some.");
+                throw new IOException("Internal server error. Unexpected response body received from Docker.");
             }
             return objectFromJson;
         } catch (JsonParseException e) {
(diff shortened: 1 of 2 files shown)


ghost (approved):

AndrienkoAleksandr (approved):

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/DockerConnector.java
```diff
@@ -1346,15 +1346,23 @@ private String getBuildImageId(ProgressStatus progressStatus) {
 
     protected <T> T parseResponseStreamAndClose(InputStream inputStream, Class<T> clazz) throws IOException {
         try (InputStreamReader reader = new InputStreamReader(inputStream)) {
-            return GSON.fromJson(reader, clazz);
+            T objectFromJson = GSON.fromJson(reader, clazz);
+            if (objectFromJson == null) {
+                throw new IOException("Docker responded with an empty stream.");
```
garagatyi: Empty stream says nothing to a user I suppose. And in certain cases, e.g. 301 or 201, an empty body is OK, so this message can be frustrating a bit. 
WDYT about next code instead:
```
LOG.error("Docker response doesn't contain any data, though it was expected to contain some. Stack trace: {}", Thread.currentThread().getStackTrace());
throw new IOException("Internal server error. Unexpected response body from Docker daemon received.");
```

mmorhun: Agree, this will be better.

garagatyi (approved):

slemeur:
@mmorhun : Please assign milestone and label when merging.

PR #5551, opened 2017-07-06
Stop Docker runtime in case of running runtime error

### LINKED ISSUE #5419: Stop Docker runtime in case of appearance of internal error like OOM
When some Runtime finds out that an internal error happened in the runtime and it can not continue its work it should stop the runtime and notify SPI to clear up WS status.

garagatyi:
This change is relevant to the spi branch only.
### What does this PR do?
Stop Docker runtime in case of running runtime error.
Move workspace statuses management to WorkspaceRuntimes.

Fixes #5419

### DIFF
--- a/infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/monit/AbnormalMachineStopHandler.java
+++ b/infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/monit/AbnormalMachineStopHandler.java
@@ -1,0 +1,25 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2017 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.workspace.infrastructure.docker.monit;
+
+/**
+ * Handles unexpected stop of a Docker machine.
+ *
+ * @author Alexander Garagatyi
+ */
+public interface AbnormalMachineStopHandler {
+    /**
+     * Accepts machine stop error message and handles unexpected stop of a machine.
+     *
+     * @param error
+     */
+    void handle(String error);
+}
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
@@ -21,6 +21,7 @@ import org.eclipse.che.api.core.model.workspace.WorkspaceStatus;
 import org.eclipse.che.api.core.model.workspace.config.Environment;
 import org.eclipse.che.api.core.model.workspace.runtime.RuntimeIdentity;
 import org.eclipse.che.api.core.notification.EventService;
+import org.eclipse.che.api.core.notification.EventSubscriber;
 import org.eclipse.che.api.workspace.server.model.impl.EnvironmentImpl;
 import org.eclipse.che.api.workspace.server.model.impl.RuntimeImpl;
 import org.eclipse.che.api.workspace.server.model.impl.WorkspaceImpl;
@@ -29,6 +30,7 @@ import org.eclipse.che.api.workspace.server.spi.InternalRuntime;
 import org.eclipse.che.api.workspace.server.spi.RuntimeContext;
 import org.eclipse.che.api.workspace.server.spi.RuntimeIdentityImpl;
 import org.eclipse.che.api.workspace.server.spi.RuntimeInfrastructure;
+import org.eclipse.che.api.workspace.shared.dto.event.RuntimeStatusEvent;
 import org.eclipse.che.api.workspace.shared.dto.event.WorkspaceStatusEvent;
 import org.eclipse.che.commons.env.EnvironmentContext;
 import org.eclipse.che.commons.lang.concurrent.ThreadLocalPropagateContext;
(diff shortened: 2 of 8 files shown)


akorneta (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -249,48 +264,45 @@ public Runtime get(String workspaceId) throws NotFoundException, ServerException
...
                                                                              InfrastructureException,
                                                                              ConflictException {
-
+        RuntimeState state = runtimes.get(workspaceId);
+        if (state == null) {
+            throw new NotFoundException("Workspace with id '" + workspaceId + "' is not running.");
+        }
+        if (!state.status.equals(RUNNING)) {
+            throw new ConflictException(
+                    format("Could not start workspace '%s' because its state is '%s'", workspaceId, state.status));
```
voievodin: typo _start_ -> _stop_

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -249,48 +264,45 @@ public Runtime get(String workspaceId) throws NotFoundException, ServerException
...
+        }
+        runtimes.replace(workspaceId, new RuntimeState(state.runtime, STOPPING));
```
voievodin: I'm not sure i understand the whole flow, but it looks like this is a place which should be passed by the one thread only. I think that this version of replace allows you to achieve such behaviour:
```java
if (!runtimes.replace(workspaceId, state, new RuntimeState(state.runtime, STOPPING))) {
    // some other thread already put the new value, handle it somehow
}
```
WDYT?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -301,11 +313,21 @@ public boolean hasRuntime(String workspaceId) {
...
     private void recover() {
         for (RuntimeInfrastructure infra : infraByRecipe.values()) {
             try {
                 for (RuntimeIdentity id : infra.getIdentities()) {
-                    runtimes.put(id.getWorkspaceId(), validate(infra.getRuntime(id)));
+                    // TODO how to identify correct state of runtime
+                    if (runtimes.putIfAbsent(id.getWorkspaceId(),
+                                             new RuntimeState(validate(infra.getRuntime(id)), RUNNING)) != null) {
+                        // should not happen, violation of SPI contract
+                        LOG.error("More than 1 runtime of workspace with id '' found. Skipping all duplicates.");
```
voievodin: Missing _id_ parameter. I think more information should be logged, for example _RuntineIdentity_ of the duplicate

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -335,6 +361,34 @@ private static EnvironmentImpl copyEnv(Workspace workspace, String envName) {
...
+    private class CleanupRuntimeOnAbnormalRuntimeStop implements EventSubscriber<RuntimeStatusEvent> {
+        @Override
+        public void onEvent(RuntimeStatusEvent event) {
+            if (event.isFailed()) {
+                RuntimeState state = runtimes.remove(event.getIdentity().getWorkspaceId());
+                if (state != null) {
+                    eventService.publish(DtoFactory.newDto(WorkspaceStatusEvent.class)
+                                                   .withWorkspaceId(state.runtime.getContext().getIdentity()
+                                                                         .getWorkspaceId())
+                                                   .withPrevStatus(WorkspaceStatus.STOPPING)
```
voievodin: Are you sure it's _STOPPING_ ?
It looks like if the stop is abnormal then the flow should be _RUNNING -> STOPPED_

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/monit/AbnormalMachineStopHandler.java
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.workspace.infrastructure.docker.monit;
+
+/**
+ * @author Alexander Garagatyi
+ */
+public interface AbnormalMachineStopHandler {
```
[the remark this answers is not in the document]
garagatyi: What if the handler doesn't use exception at all.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
garagatyi: tnx

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
garagatyi: There is no guarantee that objects will be equal. Some of the fields might change in the process of start, so it really depends on equal method implementation

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
garagatyi: ok

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
garagatyi: tnx

garagatyi:
@evoevodin I updated PR in accordance with your comments

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
garagatyi: Yep, you are right. I updated the PR

voievodin (approved):

PR #5563, opened 2017-07-07
[OpenShift] Allow to run some extra commands when running the container command.

benoitf:
### What does this PR do?
Allow to run some extra commands when running the container command.

So we can for example add location of the log file
https://github.com/redhat-developer/rh-che/issues/145

#### Changelog
OpenShift connector : allow to start extra commands

#### Release Notes
N/A

#### Docs PR
N/A

Change-Id: I75862d07759fc9c163ff579ae3b4e10d74dd1c93

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalDockerModule.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/LocalDockerModule.java
@@ -26,7 +26,9 @@ import org.eclipse.che.plugin.docker.machine.DockerProcess;
 import org.eclipse.che.plugin.docker.machine.LocalDockerCustomServerEvaluationStrategy;
 import org.eclipse.che.plugin.docker.machine.ServerEvaluationStrategy;
 import org.eclipse.che.plugin.docker.machine.node.DockerNode;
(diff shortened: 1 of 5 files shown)


benoitf:
@l0rd what is the status on this PR ? SPI branch will replace this logic and with side car containers, this kind of commands should become useless

l0rd:
@benoitf I think that's a clever solution and I thank you for that but I don't like the ` workspaceCommandAppender` to be for the OpenShift connector only. I would prefer that it worked for any connector. Having the command appender just for the OpenShift connector doesn't make really sense for me for a few reasons:

- It doesn't matter if Che is running on Docker or OpenShift: appending some info on the container log (e.g. how to access the agents logs) is useful in both cases
- Right now the OpenShift connector is really just an adapter from the Docker API to the OpenShift API and I'm not comfortable with adding here some some logic that is not related to that. 
- Adding a discrepancy between the behaviour of the two connectors is another difficulty that we will need to take care of for the SPI (something I would prefer to avoid). 

So I would rather prefer a solution (not ideal but less fragile) at the Dockerfile level. Or, maybe we should just wait to have the log files in a persistent volumes and that alone should solve the original need.

benoitf:
@l0rd as I said, this will never go to the SPI branch as it's being refactored and with side cars containers this is useless

benoitf:
so I'm closing it as you don't want it.

Issue #5477, opened 2017-06-27
The Save button in the Workspace config tab appears only after adding second symbol to config


SkorikSergey:
**Reproduction Steps:**
Create a workspace and click on the Workspace config tab
Add one symbol to config

**Expected behavior:**
The Save button appeared

**Observed behavior:**
The Save button will appears only after adding second symbol

**Che version:** 5.14.0-SNAPSHOT
**OS and version:** Fedora 25
**Additional information:**
Problem can be reliably reproduced, doesn't happen randomly: [Yes]
![factory-1](https://user-images.githubusercontent.com/7760565/27585214-f36417c2-5b43-11e7-91c7-060f01cb31cf.gif)

bdshadow:
@SkorikSergey @ashumilova the issue can be closed

SkorikSergey:
Fixed.

ashumilova:
@bdshadow Thank you

PR #5566, opened 2017-07-09
CHE-5477: Save button in Workspace config tab appears only after adding second symbol to config

bdshadow:
### What does this PR do?
It fixes the corresponding issue - the save button in the workspace config tab appears only after adding second symbol to config

#5477 

#### Changelog
Changed workspace is passed into the `updateWorkspaceConfigImport` method and must be used further in `switchEditMode` to detect the presence of changes instead of `this.workspaceDetails.config`.


bdshadow:
Looks like i don't have access to add labels. It's really strange that i can't do it for my own PR. Labels (as well as Reviewers, Assignees and the rest on the right column) are uneditable.
Please, add `kind/bugs` and `status/code-review` labels. 
Thank you!

ashumilova (approved):

Issue #3663, opened 2017-01-11
Unable to stop sync process started from CLI


slemeur:
When starting the sync process from the CLI, it is not possible to kill it:
`CRTL + C`does not do anything.

Keyboard: Mac, french layout

**Reproduction Steps:**

1. Started the command as this : 
`docker run -it -v /var/run/docker.sock:/var/run/docker.sock -v /Users/stevanlemeur/dev:/data -v /Users/stevanlemeur/Desktop/local-sync:/sync eclipse/che sync wksp-a22n`
2. The sync process is starting properly informing that we can stop by using `CRTL`+ `C`
3. Input `CRTL`+ `C`

Expected behavior:
The sync process is stopped

Current behavior:
The sync process is not stopped and the only way is to stop the Docker container (using Docker commands)

**Che version:**    5.0.0
**OS and version:**    MAC OS 10.12.2
**Docker version:** 1.12.3

ghost:
Fixed by https://github.com/eclipse/che/pull/5576

PR #5576, opened 2017-07-10
Run sync container in an interactive mode

ghost:
### What does this PR do?

Adds `-ti` to docker run syntax when syncing workspace projects

https://github.com/eclipse/che/issues/3663

### DIFF
--- a/dockerfiles/base/scripts/base/commands/cmd_sync.sh
+++ b/dockerfiles/base/scripts/base/commands/cmd_sync.sh
@@ -48,7 +48,8 @@ cmd_sync() {
   # Determine the mount path to do the mount
   info "mount" "Starting sync process to ${SYNC_MOUNT}"
 
-  docker_run --cap-add SYS_ADMIN \
+  docker_run -ti \
+	     --cap-add SYS_ADMIN \
              --device /dev/fuse \
              --security-opt apparmor:unconfined \
              -e CHE_VERSION=${CHE_VERSION} \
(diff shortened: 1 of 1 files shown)


benoitf (commented):
IMHO it should use the same approach than here
https://github.com/eclipse/che/blob/aed295c7bab11e92b4ad1ec3f82d2e187507baae/dockerfiles/base/scripts/base/library.sh#L338-L348

So if -ti is not given by user, we don't report error as well.

(We had several issues in the past hardcoding -ti because some people don't want/or unable to give that parameter (like in dummy terminal, etc)

ghost:
But in this case I do run che sync with `-ti`. And it worked for me only when I explicitly added `-ti` in actual docker run command.

benoitf:
yes but if sync command is hardcoding `-ti` and that you run `che sync`command without `-ti` it will fail while you might expect you can't hit control+c but it should still work.

ghost:
Not sure I understood the most recent comment. Under what circumstances will this fix lead to a failure?

benoitf:
@eivantsov :

if you run without -it option the command line
`docker run -it -v /var/run/docker.sock:/var/run/docker.sock -v /Users/stevanlemeur/dev:/data -v /Users/stevanlemeur/Desktop/local-sync:/sync eclipse/che sync wksp-a22n` with your fix, this command will fail while it should work (it's just that in that case ctrl+c break shouldn't be available)

quick resume : your fix is requiring that all people that launch `che sync`command should provide `-it` while it shouldn't

ghost:
I see, but `-ti` is what we always advertise in docs https://codenvy.io/docs/user-guide/sync/index.html#use

I have pushed an update - this time using the function.

benoitf:
@eivantsov yes it's advertised like this but for some usecases (like a jenkins job, etc) the terminal can't be interactive this is why we introduced the function.

benoitf (approved):

ghost:
Yes, I understand the reason behind it. Thanks for the pointers.

PR #5585, opened 2017-07-10
CHE-5555: try to fix occasionally failing tests on CI

### LINKED ISSUE #5555: org.eclipse.che.plugin.docker.machine.MachineProviderImplTest.setUp test is failing randomly on CI
We have PR jenkins jobs that fails very often with the error :

```
org.eclipse.che.plugin.docker.machine.MachineProviderImplTest.setUp
Error Message

'getContainerLogs' is a *void method* and it *cannot* be stubbed with a *return value*!
Voids are usually stubbed with Throwables:
    doThrow(exception).when(mock).someVoidMethod();
```

so we need to launch again the tests...so jenkins jobs take longer to perform as we run them more than once.
Maybe there is wrong init/cleanup sequence in the tests

**Reproduction Steps:**
Jenkins PR jobs like
https://ci.codenvycorp.com/job/che-pullrequests-build/org.eclipse.che.plugin$che-plugin-docker-machine/3009/testReport/

garagatyi:
### What does this PR do?
We have very complicated class and tests for it fail occasionally on CI. It is hard to identify problem now and Che 6 probably won't include these tests.
This PR is trying to workaround this problem. It actually disables some code execution using mocking mechanics. 

May fix #5555 

#### Changelog
Fix occasionally failing tests on CI


garagatyi:
@benoitf @evoevodin Guys, WDYT about this approach?

benoitf:
@garagatyi let's try it

benoitf (approved):

voievodin (approved):

Issue #5621, opened 2017-07-12
Workspace is not started correctly sometimes


RomanNikitenko:
Looks like we have some circular dependencies and it leads to case when Workspace can not start correctly.
The problem is can not  reproduced permanently, but happens very often last time.

**Che version:** 5.15.0
**OS and version:**   Fedora 25
![start_workspace](https://user-images.githubusercontent.com/5676062/28116864-a5c0656c-6713-11e7-8ac8-34b01371f89d.png)

**Diagnostics:** 
```[STDOUT] Wed Jul 12 11:53:02 UTC 2017 - [localhost-startStop-1] Product-specified preferences called before plugin is started
[STDOUT] 2017-07-12 11:53:03,077[ost-startStop-1]  [ERROR] [o.a.c.c.C.[.[localhost].[/] 4733]    - Exception sending context initialized event to listener instance of class org.eclipse.che.inject.CheBootstrap
[STDOUT] com.google.inject.CreationException: Unable to create injector, see the following errors:
[STDOUT] 
[STDOUT] 1) Invocation error of method public void org.eclipse.che.api.project.server.ProjectRegistry.initProjects() throws org.eclipse.che.api.core.ConflictException,org.eclipse.che.api.core.NotFoundException,org.eclipse.che.api.core.ServerException,org.eclipse.che.api.core.ForbiddenException on org.eclipse.che.api.project.server.ProjectRegistry@70e29853
[STDOUT]   at org.eclipse.che.api.project.server.ProjectApiModule.configure(ProjectApiModule.java:75) (via modules: org.eclipse.che.wsagent.server.WsAgentModule -> org.eclipse.che.api.project.server.ProjectApiModule)
[STDOUT]   while locating org.eclipse.che.api.project.server.ProjectRegistry
[STDOUT]     for the 3rd parameter of org.eclipse.che.api.project.server.ProjectManager.<init>(ProjectManager.java:97)
[STDOUT]   at org.eclipse.che.api.project.server.ProjectManager.class(ProjectManager.java:74)
[STDOUT]   while locating org.eclipse.che.api.project.server.ProjectManager
[STDOUT]     for the 1st parameter of org.eclipse.che.plugin.testing.phpunit.server.PHPUnitTestRunner.<init>(PHPUnitTestRunner.java:52)
[STDOUT]   while locating org.eclipse.che.plugin.testing.phpunit.server.PHPUnitTestRunner
[...]

PR #5636, opened 2017-07-13
CHE-5621: Avoid using ProjectManager just for taking project root dir

vparfonov:
### What does this PR do?
Avoid using ProjectManager only for taking project root dir instead of we will use special properties 

#5621


pom.xml
```diff
@@ -32,7 +32,7 @@
         <module>plugins</module>
         <module>ide/che-ide-core</module>
         <module>samples</module>
-        <module>dashboard</module>
+        <!--module>dashboard</module-->
```
azatsarynnyy: should dashboard be disabled?

dkulieshov (approved):

pom.xml
vparfonov: fixed

Issue #5595, opened 2017-07-11
Automate generation of changelog


slemeur:
We would like to completely automate the generation of the changelog when doing a release, by using:  https://github.com/skywinder/github-changelog-generator

The idea is the following:
- CI handle the generation of the changelog.
- The changelog is based on PRs merged.
- The changelog has sections to regroup PRs by `kind/*` labels

We will need to see how we can distinguish the security and deprecation issues. (probably by introducing new labels)

riuvshin:
@slemeur bad news here, that tool can't decorate / filter PRs it work only with issues.
so for PRs we can only adjust the title "Merged pull requests"
So... I think we will generate for future versions in same way as simple list of all merged PRs for particular version.

riuvshin:
ci job - https://ci.codenvycorp.com/view/release/job/generate-changelos/

PR #5681, opened 2017-07-17
regenerate changelog

riuvshin:
### What does this PR do?
regenerate change log for CHE

@slemeur please review the format, and here is some notes about that.
I found why changelog generator wasn't working well for us, this is because we have huge amount of issues and PRs and even with auth token github allows only 5000 API calls for 1 hour. With our amount of data even 1 run of change log generator will fail due to exceeding that API calls limit and github will ban us. So to avoid this we should **not** include any issue to changelog and we should **not** use filtering by labels. Also I have to set limit for 1000 items. I understand that this is probably not something we wanted but I don't see what we can do here... I've tried to use several tokens and generate in few iterations but github treat this as ddos attack and block by IP.
So that changelog include ONLY merged PRs in particular version without any filtering. It is simple and I believe it is pretty enough and clear.

https://github.com/eclipse/che/issues/5595

#### Changelog
regenerate change log for CHE


slemeur:
Thanks @riuvshin .

Is that possible to limit the execution of the script to the latest release version?
So for example, we would generate the changelog only for 5.14 and append that to `changelog.md` ?

riuvshin:
@slemeur I still investigating this, but it seems that generator is not that smart... it generate all the time whole changelog! so as you can see due to limit I've applied there is no changelog for 5.4.1 and older.... so potentially we will keep change log only for some fixed amount of latest versions... that sucks and I hope I will figure out how to append... because without this that tool useless.

riuvshin:
@slemeur https://github.com/skywinder/github-changelog-generator/issues/209 since 2015.... 
so I have idea I will try to generate change log for current version only to temp file and then i will write some script to put content of that to CHANGELOG.md file. that sucks but I don't see any other way...

slemeur:
Yeah - I definitely agree that it would be better to keep the whole history. 
Let me know how it goes. If you don't find a solution, we might look for another format.

riuvshin:
@slemeur OK, I've updated PR with regenerated CHANGELOG.md now it contains all **merged prs** for all existing tags. once it will be merged I will start work on automation for generation changelog for current version and append the changelog with it with that approach we should be able to apply some extra filtering for new versions but of all these existing versions we will keep only merged prs. if that OK for you please approve this and I will merge it. 
btw one we will done with CHE I will do same for Codenvy

vparfonov (approved):

slemeur:
The current changelog is already having those information, but the language of PRs' titles were cleaned by @JamesDrummond. 
Can we:
- keep the current changelog from until 5.10
- append the changelogs for each tags between 5.10 and 5.14

Starting from 5.15, the PRs titles have to be reviewed by the PR's reviewers, so we could expect better formatting and language. 

If we can do that, I'll manually clean the changelog between 5.10 and 5.14

riuvshin:
ok I will keep existing changelog and will generate only merged PRs for 5.10.0 - 5.14.0 versions
for 5.15 + we will improve format and filtering I just need to find a way how to generate for current version and how to merge them once it generated.

riuvshin:
@slemeur I've added 5.10-5.14 changelog

slemeur (approved):
We should be good.

Issue #5686, opened 2017-07-17
Implement initial state for openshift spi


sleshchenko:
It is needed to have an initial state of openshift implementation for further development. So in the scope of this issue, it is needed to start workspace on minishift instance and get working IDE.

Issue #5691, opened 2017-07-17
Add using of persistent volumes for storing sources of projects


sleshchenko:
Not to lose local changes in sources of projects after workspace restarting we should use persistent volumes for storing sources of projects.

bmicklea:
Please add correct labels.

gorkem:
Every PV has different characteristic depending on the backend performance may be different? Is this going to be configurable?

akorneta:
@gorkem I'm not sure that I fully understand question, could you please clarify what exactly should be configurable?

benoitf:
there are several implementations of PV : https://docs.openshift.com/enterprise/3.1/install_config/persistent_storage/index.html

akorneta:
@benoitf So we do not manage PVs in che at all. The current issue is about adding PVCs for storing workspace projects. As far as I understand PVC doesn't provide an ability to choose which back end should be used. It is a part of OpenShift instance configuration managed by admins or I missed something?

benoitf:
@akorneta when you claim a persistent volume, you ask for characteristics

https://docs.openshift.com/enterprise/3.0/architecture/additional_concepts/storage.html#persistent-volume-claims
like the access mode, etc.

so maybe you want to be able to choose options of the PV claim.

gorkem:
It is not even that advanced. Some PVCs are slow for small writes but optimized for larger writes. I think we should have the option disable persistent volume storing in case we see that it affects the performance of the IDE.

akorneta:
Current implementation of `OpenShiftConnector` has hard coded access mode to `ReadWriteOne` and has configurable `quantity`. I'm going to make configuration property for `quantity`. 
@benoitf Do you think that it makes sense to make access mode configurable?
> disable persistent volume

@gorkem But then we may lose changes of projects sources after workspace restart or crash. I think it is not good UX. Do you think it is acceptable?

gorkem:
It is not good UX but I was thinking of it as a fallback if PV starts to render IDE unusable.

akorneta:
@gorkem ok it makes sense I'll add appropriate config property

Issue #5690, opened 2017-07-17
Implement environment provisioning in openshift spi


sleshchenko:
Openshift spi implementation doesn't have full environment provisioning like docker implementation does.
Now it adds only `CHE_WORKSPACE_ID` and `CHE_API` environment variables to each pod
It is needed to add servers(missing services and routes objects) and environment variables according to configured installers.

bmicklea:
Please add correct labels.

PR #5716, opened 2017-07-18
OpenShift spi initial implementation

sleshchenko:
### What does this PR do?
Minor fixes:
- Fix location of agent binaries
- Add missed server configs to InstallerDto
- Fix construction of output channel

Changes related to spi reworking to make it reusable by other implementations:
- Move installer and output endpoints to api-workspace module
- Rework Bootstrapping to be able to reuse it for different spi implementations

Add initial state for openshift implementation of spi

https://github.com/eclipse/che/issues/5686


agents/exec/src/main/resources/org.eclipse.che.exec.script.sh
```diff
@@ -160,50 +150,23 @@ eval "LOCAL_AGENT_BINARIES_URI=${LOCAL_AGENT_BINARIES_URI}"
...
-    elif curl -o /dev/null --silent --head --fail $(echo ${AGENT_BINARIES_URI} | sed 's/-\${PREFIX}//g'); then
-      curl -o $(echo ${TARGET_AGENT_BINARIES_URI} | sed 's/\${PREFIX}/'${PREFIX}'/g' | sed 's/file:\/\///g') -s $(echo ${AGENT_BINARIES_URI} | sed 's/-\${PREFIX}//g')
-    fi
-    curl -s $(echo ${TARGET_AGENT_BINARIES_URI} | sed 's/\${PREFIX}/'${PREFIX}'/g') | tar  xzf - -C ${CHE_DIR}
-  else
-    # replace https by http as wget may not be able to handle ssl
-    AGENT_BINARIES_URI=$(echo ${AGENT_BINARIES_URI} | sed 's/https/http/g')
-
-    # use wget
-    WGET_SPIDER="wget --spider"
```
benoitf: why wget is removed ?

agents/terminal/src/main/resources/org.eclipse.che.terminal.script.sh
```diff
@@ -159,50 +149,23 @@ eval "LOCAL_AGENT_BINARIES_URI=${LOCAL_AGENT_BINARIES_URI}"
...
-    elif curl -o /dev/null --silent --head --fail $(echo ${AGENT_BINARIES_URI} | sed 's/-\${PREFIX}//g'); then
-      curl -o $(echo ${TARGET_AGENT_BINARIES_URI} | sed 's/\${PREFIX}/'${PREFIX}'/g' | sed 's/file:\/\///g') -s $(echo ${AGENT_BINARIES_URI} | sed 's/-\${PREFIX}//g')
-    fi
-    curl -s $(echo ${TARGET_AGENT_BINARIES_URI} | sed 's/\${PREFIX}/'${PREFIX}'/g') | tar  xzf - -C ${CHE_DIR}
-  else
-    # replace https by http as wget may not be able to handle ssl
-    AGENT_BINARIES_URI=$(echo ${AGENT_BINARIES_URI} | sed 's/https/http/g')
-
-    # use wget
-    WGET_SPIDER="wget --spider"
```
benoitf: why wget is removed ?

agents/exec/src/main/resources/org.eclipse.che.exec.script.sh
```diff
@@ -12,18 +12,8 @@
...
-
-# no curl, no wget, install curl
-if [ ${CURL_INSTALLED} = false ] && [ ${WGET_INSTALLED} = false ]; then
-  PACKAGES=${PACKAGES}" curl";
-  CURL_INSTALLED=true
-fi
-
-test "$(id -u)" = 0 || SUDO="sudo -E"
+command -v curl >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" curl"; }
+test "$(id -u)" = 0 || test -f ${HOME}/is_arbitrary_user || SUDO="sudo -E"
```
benoitf: why do we have a test on is_arbitrary_user ?

ide/che-core-ide-stacks/src/main/resources/stacks.json
```diff
@@ -2771,5 +2771,50 @@
...
+              "installers": [
+                "org.eclipse.che.exec", "org.eclipse.che.terminal", "org.eclipse.che.ws-agent"
+              ],
+              "servers": {},
+              "attributes" : {
+                "memoryLimitBytes": "2147483648"
+              }
+            }
+          },
+          "recipe": {
```
benoitf, line 2797: AFAIK right now in OpenShift Connector there is a deployment generator of workspace definitions. Here you expect that users have to provide specific data for OSIO ?

agents/terminal/src/main/resources/org.eclipse.che.terminal.script.sh
benoitf: note: support was added as on alpine curl is not provided by default while wget is there

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenshiftInternalRuntime.java
```diff
@@ -0,0 +1,290 @@
...
+                                                 new ServerImpl("http://" + route.getSpec().getHost(),
+                                                                ServerStatus.UNKNOWN));
+                                }
+                            }
+
+                        }
+                    }
+                }
+            }
+        } catch (RuntimeException ignored) {
```
benoitf: silent catch ?

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenshiftInternalRuntime.java
```diff
@@ -0,0 +1,290 @@
...
+                Thread.currentThread().interrupt();
+            }
+            try (OpenShiftClient client = clientFactory.create()) {
+                Pod actualPod = client.pods()
+                                      .inNamespace(pod.getMetadata().getNamespace())
+                                      .withName(pod.getMetadata().getName())
+                                      .get();
+
+                if (actualPod == null) {
+                    throw new InternalInfrastructureException("Can't find creted pod");
```
benoitf: FYI typo on exception

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenshiftInternalRuntime.java
```diff
@@ -0,0 +1,290 @@
...
+                }
+                String status = actualPod.getStatus().getPhase();
+                LOG.info("POD " + actualPod.getMetadata().getName() + " has status " + status);
```
benoitf: no + usage in logger

benoitf (commented):
I may be wrong but it looks like odd for me that the recipe format is a dedicated format. Like if I write a simple dockerfile from eclipse che it won't work on OSIO ?

I don't see usage of persistent volume, env variables, but as it seems to be "initial" implementation, there is a lot of TODO
It looks like that some code is coming from the current openshift connector implementation in master but copyright header/author is changed.

agents/terminal/src/main/resources/org.eclipse.che.terminal.script.sh
sleshchenko: Thanks for comment. I copied/pasted it from openshift-connector branch. I'll revise all changes in installer scripts and notify you.

benoitf: @sleshchenko note: we're now using `openshift-connector-wip` branch of che

sleshchenko: These changes were redundant. I reverted all changes in installers scripts. Thanks for this catch.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenshiftEnvironmentParser.java
```diff
@@ -0,0 +1,112 @@
...
+
+        return openshiftEnvironment;
+    }
+
+    private String getContentOfRecipe(Recipe environmentRecipe) throws InfrastructureException {
+        if (environmentRecipe.getContent() != null) {
+            return environmentRecipe.getContent();
+        } else {
+            try {
+                return recipeDownloader.getRecipe(environmentRecipe.getLocation());
```
garagatyi, line 100: I know that it is similar to Docker infra impl. But it should use recipe retrieved by RuntimeContext

garagatyi (approved):
Good job! Since it is an initial step I believe it is a good start!

ide/che-core-ide-stacks/src/main/resources/stacks.json
l0rd: I don't think it's a bad idea to introduce a new recipe type `openshift`. 

But yes the approach we had with the current OpenShift connector was to be able to reuse current "docker image" recipes. It makes sense to run a `centos_jdk8` Docker image on OpenShift so we didn't feel the need to support the openshift deployment descriptions (yaml or json). 

The only drawback with your approach is that you will need to duplicate all the stacks in `stacks.json` if you want them to run on openshift.

gazarenkov: To make the things clear - the format is not "dedicated", it is "native" for K8s, OS, which I consider as a huge advantage of Che6 SPI.
Showing ability to support it is one of the main goal of initial implementation.
The problem you are talking about can be solved with more than one way:
- defining another environment, i.e. one for Docker, another for OSIO (we can imagine some more for example for Local computer, no docker, no containers), so one or another will work depending on available infrastructure
- adding one more infrastructure type which works on OS but support some other, simplified format (docker image, dockerfile etc)
- extending this infrastructure implementation making it support more than one recipe type (like in Docker impl) - possible by design but I personally would not prefer this way 

And, of course they can be combined.

garagatyi: @benoitf As it is an initial implementation it doesn't have to cover all the use cases. Implementing k8s here seems more interesting since it was quite impossible in the master.
And points 2 and 3 (GA, yes, it's by design) from Gennady's comment would allow using env types from the master in Openshift infra.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenshiftRuntimeContext.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @Override
+    public URI getOutputChannel() throws InfrastructureException {
+        try {
+            final URI apiURI = URI.create(apiEndpoint);
+            return UriBuilder.fromUri(apiURI)
+                             .scheme("https".equals(apiURI.getScheme()) ? "wss" : "ws")
+                             .replacePath(apiURI.getPath().replace("/api", ""))
+                             .path(OUTPUT_WEBSOCKET_ENDPOINT_BASE)
```
gazarenkov: Let's avoid URI transformation providing ws:// URI directly instead (in general case it can be different than http:// )

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/bootstrapper/OpenshiftBootstrapper.java
```diff
@@ -0,0 +1,115 @@
...
+                                 @Named("che.infra.openshift.bootstrapper.timeout_min") int bootstrappingTimeoutMinutes,
+                                 @Named("che.infra.openshift.bootstrapper.installer_timeout_sec") int installerTimeoutSeconds,
+                                 @Named("che.infra.openshift.bootstrapper.server_check_period_sec") int serverCheckPeriodSeconds,
+                                 EventService eventService) {
+        super(machineName, runtimeIdentity, bootstrappingTimeoutMinutes, websocketBaseEndpoint, eventService);
+        try {
+            URL wsmasterEndpoint = new URL(workspaceMasterEndpoint);
+            this.downloadBootstrapperLink = wsmasterEndpoint.getProtocol() + "://" + wsmasterEndpoint.getHost() + ":" +
+                                            wsmasterEndpoint.getPort() +
+                                            "/agent-binaries/linux_amd64/bootstrapper/bootstrapper";
```
gazarenkov: Let's avoid modifying URL. It is temporary is not it?

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenshiftRuntimeContext.java
sleshchenko: Introduced new property `che.websocket.endpoint.base` and added using it while construction of output channel

akorneta (approved):

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/bootstrapper/OpenshiftBootstrapper.java
sleshchenko: Added dedicated property `che.infra.openshift.bootstrapper_binary_url`.

Maybe later we will inject bootstrapper binary in other way but now I have no idea how we can implement it without execution command for downloading like curl or wget.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenshiftEnvironmentParser.java
sleshchenko: @garagatyi Do you mean that parsing and provisioning of the environment should be done by `RuntimeContext` instead of `RuntimeInfrastructure`?

sleshchenko:
@benoitf 
> Like if I write a simple dockerfile from eclipse che it won't work on OSIO ?

You're right. It won't work. Mb later we will support dockerfile/composefile but not in this initial implementation of openshift spi. See https://github.com/eclipse/che/pull/5716#discussion_r128162844

> I don't see usage of persistent volume, env variables

There are few [hard coded env variables](https://github.com/eclipse/che/pull/5716/files#diff-
dc6c75e34aa896716a466e0b3bf1cf98R40), it will be fixed in https://github.com/eclipse/che/issues/5690
For using persistent volume there is another issue https://github.com/eclipse/che/issues/5691

> It looks like that some code is coming from the current openshift connector implementation in master but copyright header/author is changed.

It this comment actual after refactoring? If yes, then please notify me.

And please take a look my PR again after refactoring.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenshiftEnvironmentParser.java
garagatyi: I mean that downloading of environment recipe is done on RuntimeContext.

ide/che-core-ide-stacks/src/main/resources/stacks.json
benoitf: I'm just saying that che is a 'portable workspace' so I was expecting I could use my workspace on all platforms whatever is their native stuff. It's like when I use docker I don't care if it's windows, linux or MacOS as native platform. 
If I need to rewrite the data of my workspace then it's not anymore 'portable'

l0rd:
I think that supporting recipes of type `openshift` is definitely a good thing. So I really like the work you are doing and that is something that we haven't implemented in the OpenShift connector and it's cool that you have thought about that. But I think that before merging this to master you should add the capability to run different kind of recipes on OpenShift too. Not only `openshift` recipes but at least also already built docker images recipes. I'm sure that is something you have already think about but let me note here a couple of reasons why this is important:

- **Portable stacks**: Until now a user was able to use the same stack definition without worrying about what was the underlying Che infrastructure (docker, swarm, openshift, k8s etc...). And that is good because I want to be able to use the same runtime Dockerfile on my laptop, for the CI and in production. If we having different stacks for different environment we loose all the benefits of using Docker containers.

- **Don't break OSIO**:  OSIO Che will be broken if you don't support running already-built docker images recipes. Even if you can map every Docker stack to an OpenShift stack there are some details that need to be deduced at runtime and cannot be hardcoded in stacks.json. For example workspace routes and paths of workspace persistent volume claims.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenshiftEnvironmentParser.java
sleshchenko: I just don't understand how to implement it. As docker infra does: context should have already parsed environment, so we need to download recipe for parsing before context creation. 
WDYT should we do in this case? Should we move parsing of recipe and provisioning to context? Or maybe now leave it as it is and rework later along with docker infra?

garagatyi: Docker infra is a separate implementation. Your implementation doesn't have to do things in the same way as it does. Downloading in docker infra was implemented long time ago. Then we realized that it should use recipe downloaded in context, but haven't changed.

sleshchenko: Doesn't have but for me, the responsibility of infrastructure and context is not clear enough. So I implemented in the same way as docker infrastructure does. So I'd leave it as it is. And rework it later when runtime infrastructure won't do estimation (that require recipe downloading) and it will define when validation and estimation should be done.

skabashnyuk (approved):
@l0rd  I agree with all your points. As we mention the main purpose of this PR is to prove that Che6 SPI is capable of providing native capabilities for Openshift. I understand your concerns about backward compatibility with the state that we have in the master branch, and as @gazarenkov  says  https://github.com/eclipse/che/pull/5716#discussion_r128162844 we have a couple of different options to solve this.

PR #5764, opened 2017-07-20
Fix DtoServer error messages

dmytro-ndp:
### What does this PR do?
Fix error messages in case of there is no implementation of DTO interface available for DtoServer.

@evoevodin, @sleshchenko, @skabashnyuk: please, take a look at this request.

### DIFF
--- a/core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java
+++ b/core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java
@@ -386,8 +386,13 @@ public final class DtoFactory {
     private <T> DtoProvider<T> getDtoProvider(Class<T> dtoInterface) {
         DtoProvider<?> dtoProvider = dtoInterface2Providers.get(dtoInterface);
         if (dtoProvider == null) {
-            throw new IllegalArgumentException("Unknown DTO type " + dtoInterface);
+            if (dtoInterface.isAnnotationPresent(DTO.class)) {
+                throw new IllegalArgumentException("Implementation for DTO type " + dtoInterface + " isn't found.");
+            } else {
+                throw new IllegalArgumentException("Is not a DTO type " + dtoInterface);
+            }
         }
+
         return (DtoProvider<T>)dtoProvider;
     }
 
--- a/core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java
+++ b/core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java
@@ -42,6 +42,8 @@ import java.util.Map;
 import java.util.ServiceLoader;
 import java.util.concurrent.ConcurrentHashMap;
 
+import static java.lang.String.format;
+
 /**
  * Provides implementations of DTO interfaces.
  *
@@ -386,14 +388,34 @@ public final class DtoFactory {
     private <T> DtoProvider<T> getDtoProvider(Class<T> dtoInterface) {
         DtoProvider<?> dtoProvider = dtoInterface2Providers.get(dtoInterface);
         if (dtoProvider == null) {
-            if (dtoInterface.isAnnotationPresent(DTO.class)) {
-                throw new IllegalArgumentException("Provider of implementation for DTO type " + dtoInterface + " isn't found");
+            if (!dtoInterface.isInterface()) {
+                throw new IllegalArgumentException(format("Only interfaces can be DTO, but %s is not", dtoInterface));
+            }
+
+            if (checkOnDtoAnnotation(dtoInterface)) {
+                throw new IllegalArgumentException(format("Provider of implementation for DTO type %s isn't found", dtoInterface));
             } else {
-                throw new IllegalArgumentException("Is not a DTO type " + dtoInterface);
+                throw new IllegalArgumentException(dtoInterface + " is not a DTO type");
             }
         }
 
         return (DtoProvider<T>)dtoProvider;
+    }
+
+    /**
+     * Checks if dtoInterface or its parent have DTO annotation.
+     * @param dtoInterface
+     *          DTO interface
+     * @return true if onle dtoInterface or one of its parent have DTO annotation.
+     */
+    private boolean checkOnDtoAnnotation(Class dtoInterface) {
+        for (Class parent: dtoInterface.getInterfaces()) {
+            if (checkOnDtoAnnotation(parent)) {
+                return true;
+            }
+        }
+
+        return dtoInterface.isAnnotationPresent(DTO.class);
     }
 
     /**
--- a/core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java
+++ b/core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java
@@ -387,7 +387,7 @@ public final class DtoFactory {
         DtoProvider<?> dtoProvider = dtoInterface2Providers.get(dtoInterface);
         if (dtoProvider == null) {
             if (dtoInterface.isAnnotationPresent(DTO.class)) {
-                throw new IllegalArgumentException("Implementation for DTO type " + dtoInterface + " isn't found.");
+                throw new IllegalArgumentException("Provider of implementation for DTO type " + dtoInterface + " isn't found");
             } else {
                 throw new IllegalArgumentException("Is not a DTO type " + dtoInterface);
             }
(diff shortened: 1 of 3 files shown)


core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java
```diff
@@ -386,8 +386,13 @@ public Gson getGson() {
     private <T> DtoProvider<T> getDtoProvider(Class<T> dtoInterface) {
         DtoProvider<?> dtoProvider = dtoInterface2Providers.get(dtoInterface);
         if (dtoProvider == null) {
-            throw new IllegalArgumentException("Unknown DTO type " + dtoInterface);
+            if (dtoInterface.isAnnotationPresent(DTO.class)) {
+                throw new IllegalArgumentException("Implementation for DTO type " + dtoInterface + " isn't found.");
```
voievodin: Maybe it is better to say _"Provider for DTO type ..."_ ?

voievodin (approved):

core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java
dmytro-ndp: Thank you for advice.
The main reason of this request is to mention **implementation** in some way in error message.
But in the method which throws exception we operate with term **provider**.
Next message is enclosing both of reasons:
> Provider of implementation for DTO type ...

voievodin: A good one, go with it

core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java
```diff
@@ -386,8 +386,13 @@ public Gson getGson() {
     private <T> DtoProvider<T> getDtoProvider(Class<T> dtoInterface) {
         DtoProvider<?> dtoProvider = dtoInterface2Providers.get(dtoInterface);
         if (dtoProvider == null) {
-            throw new IllegalArgumentException("Unknown DTO type " + dtoInterface);
+            if (dtoInterface.isAnnotationPresent(DTO.class)) {
+                throw new IllegalArgumentException("Provider of implementation for DTO type " + dtoInterface + " isn't found");
+            } else {
+                throw new IllegalArgumentException("Is not a DTO type " + dtoInterface);
```
skabashnyuk: what if "blah-blah" is not a DTO type
?

dmytro-ndp: Then there will be an IllegalArgumentException:  "Is not a DTO type class blah-blah", won't it?

sleshchenko: I think Sergii means what if rewrite message in following way: `dtoInterface + " is not a DTO type"`

sleshchenko (approved):
It will be much easier to understand a problem with lack of dto implementations after your changes. :+1:

skabashnyuk (approved):

dmytro-ndp:
@skabashnyuk, @evoevodin, @sleshchenko: please, review my new commit with next fixes:
1. Handle non-interface types.
1. Check on DTO annotation among the parent interfaces as well.
1. Fix of error message due to @skabashnyuk suggestion.

core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java
```diff
@@ -386,12 +388,37 @@ public Gson getGson() {
     private <T> DtoProvider<T> getDtoProvider(Class<T> dtoInterface) {
         DtoProvider<?> dtoProvider = dtoInterface2Providers.get(dtoInterface);
         if (dtoProvider == null) {
-            throw new IllegalArgumentException("Unknown DTO type " + dtoInterface);
+            if (!dtoInterface.isInterface()) {
+                throw new IllegalArgumentException(format("Only interfaces can be DTO, but %s is not", dtoInterface));
+            }
+
+            if (checkOnDtoAnnotation(dtoInterface)) {
```
sleshchenko: as for me `check...` method should throw an exception and `is...` or `has...` should return true or false.

core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java
```diff
@@ -386,12 +388,37 @@ public Gson getGson() {
...
         }
+
         return (DtoProvider<T>)dtoProvider;
     }
 
     /**
+     * Checks if dtoInterface or its parent have DTO annotation.
+     * @param dtoInterface
+     *          DTO interface
+     * @return true if onle dtoInterface or one of its parent have DTO annotation.
```
sleshchenko: typo `onle`

dmytro-ndp: Thank you!

core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java
dmytro-ndp: I see. Let's rename it to **hasDtoAnnotation**.

core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java
```diff
@@ -386,12 +388,37 @@ public Gson getGson() {
...
+     */
+    private boolean checkOnDtoAnnotation(Class dtoInterface) {
+        for (Class parent: dtoInterface.getInterfaces()) {
+            if (checkOnDtoAnnotation(parent)) {
+                return true;
+            }
+        }
+
+        return dtoInterface.isAnnotationPresent(DTO.class);
```
voievodin: I think it's better to check it before analyzing interfaces, for example:
```java
@DTO
interface A extends B, C { }
```

the code above will analyze B and C before analyzing A, which i think unreasonable.

dmytro-ndp: Good advice! Thank you!

skabashnyuk (approved):

core/che-core-api-dto/src/test/java/org/eclipse/che/dto/definitions/DTOHierarchy.java
```diff
@@ -60,4 +62,6 @@
         void setShadowedField(GrandchildDto v);
     }
 
+    public interface GrandchildWithoutDto extends Child, Closeable {
```
sleshchenko: Why should it be Closeable? It can confuse =)

sleshchenko: Also Mb you are able to reuse just `Child` class or create `Grandchild` class if you want three level of extending

core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java
```diff
@@ -386,12 +388,41 @@ public Gson getGson() {
...
+            } else {
+                throw new IllegalArgumentException(dtoInterface + " is not a DTO type");
+            }
         }
+
         return (DtoProvider<T>)dtoProvider;
     }
 
     /**
+     * Checks if dtoInterface or its parent have DTO annotation.
```
sleshchenko, line 406: It would be good to add an empty line after summary section.

sleshchenko (approved):

core/che-core-api-dto/src/test/java/org/eclipse/che/dto/definitions/DTOHierarchy.java
dmytro-ndp: Closeable is just additional interface for testing propose. I am ready to produce second child if you  think it's less confusing :-)

core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java
dmytro-ndp: I see. Will fix it.
Thank you for note!

core/che-core-api-dto/src/test/java/org/eclipse/che/dto/definitions/DTOHierarchy.java
sleshchenko: Up to you

PR #5007, opened 2017-05-04
CHE-249: Setting workspace Pod terminationGracePeriodSeconds to zero

ibuziuk:
### What does this PR do?
Setting workspace Pod terminationGracePeriodSeconds to zero

### What issues does this PR fix or reference?
https://issues.jboss.org/browse/CHE-249

PR #5766, opened 2017-07-20
CHE-220: Improving 'removeContainer' API. Refactoring

ibuziuk:
Signed-off-by: Ilya Buziuk <ibuziuk@redhat.com>

### What does this PR do?
Enhances OpenShift resource deletion for workspaces (scaling down deployment before deleting resources) 

https://issues.jboss.org/browse/CHE-220

#### Changelog
Enhances OpenShift resource deletion for workspaces (scaling down deployment before deleting resources) 

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
@@ -40,7 +40,6 @@ import io.fabric8.kubernetes.api.model.VolumeMount;
 import io.fabric8.kubernetes.api.model.VolumeMountBuilder;
 import io.fabric8.kubernetes.api.model.extensions.Deployment;
 import io.fabric8.kubernetes.api.model.extensions.DeploymentBuilder;
-import io.fabric8.kubernetes.api.model.extensions.ReplicaSet;
 import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.Watcher;
 import io.fabric8.kubernetes.client.dsl.ExecWatch;
@@ -52,8 +51,6 @@ import io.fabric8.openshift.api.model.DoneableDeploymentConfig;
 import io.fabric8.openshift.api.model.Image;
 import io.fabric8.openshift.api.model.ImageStream;
 import io.fabric8.openshift.api.model.ImageStreamTag;
-import io.fabric8.openshift.api.model.Route;
-import io.fabric8.openshift.api.model.RouteList;
 import io.fabric8.openshift.client.DefaultOpenShiftClient;
 import io.fabric8.openshift.client.OpenShiftClient;
 import io.fabric8.openshift.client.dsl.DeployableScalableResource;
@@ -173,8 +170,8 @@ public class OpenShiftConnector extends DockerConnector {
     private static final int CHE_WORKSPACE_AGENT_PORT                    = 4401;
     private static final int CHE_TERMINAL_AGENT_PORT                     = 4411;
     private static final String DOCKER_PROTOCOL_PORT_DELIMITER           = "/";
-    private static final int OPENSHIFT_WAIT_POD_DELAY                    = 1000;
     private static final int OPENSHIFT_WAIT_POD_TIMEOUT                  = 240;
+    private static final int OPENSHIFT_WAIT_POD_DELAY                    = 1000;
     private static final int OPENSHIFT_IMAGESTREAM_WAIT_DELAY            = 2000;
     private static final int OPENSHIFT_IMAGESTREAM_MAX_WAIT_COUNT        = 30;
     private static final String OPENSHIFT_POD_STATUS_RUNNING             = "Running";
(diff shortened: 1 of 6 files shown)


plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftDeploymentCleaner.java
```diff
@@ -0,0 +1,139 @@
...
+
+    private void scaleDownDeployment(String deploymentName, final String namespace) throws OpenShiftException {
+        try (OpenShiftClient openShiftClient = new DefaultOpenShiftClient()) {
+            ScalableResource<Deployment, DoneableDeployment> deployment = openShiftClient.extensions()
+                           .deployments()
+                           .inNamespace(namespace)
+                           .withName(deploymentName);
+
+            if (deployment != null) {
+                deployment.scale(0, true);
```
amisevsk, line 59: Does waiting for the scale down take significantly longer than the previous way, or does the scale down take grace termination period into account? I ask because we had the issue on OpenShift where stopping and restarting a pod quickly could case PV issues.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftDeploymentCleaner.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    private void cleanUpWorkspaceResources(final String deploymentName, final String namespace) throws IOException {
+        Deployment deployment = KubernetesResourceUtil.getDeploymentByName(deploymentName, namespace);
+        Service service = KubernetesResourceUtil.getServiceBySelector(OpenShiftConnector.OPENSHIFT_DEPLOYMENT_LABEL, deploymentName, namespace);
+        List<Route> routes = KubernetesResourceUtil.getRoutesByLabel(OpenShiftConnector.OPENSHIFT_DEPLOYMENT_LABEL, deploymentName, namespace);
+        List<ReplicaSet> replicaSets = KubernetesResourceUtil.getReplicaSetByLabel(OpenShiftConnector.OPENSHIFT_DEPLOYMENT_LABEL, deploymentName, namespace);
+
+        try (OpenShiftClient openShiftClient = new DefaultOpenShiftClient()) {
+            if (replicaSets != null && replicaSets.size() > 0) {
```
amisevsk: I worry about the order of removal here, since replica sets are controlled by deployments. It shouldn't be a problem since the deployment is scaled to zero first, but I think the deployment should be deleted before the replica set just in case. IIRC there are cases where a deployment will ensure a replica set exists, causing it to maybe be recreated before the deployment is deleted.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -165,6 +162,7 @@
     public static final String OPENSHIFT_DEPLOYMENT_LABEL               = "deployment";
     public static final String CHE_MOUNTED_WORKSPACE_FOLDER             = "/workspace-logs";
     public static final String WORKSPACE_LOGS_FOLDER_SUFFIX             = "-logs";
+    public static final int OPENSHIFT_WAIT_POD_TIMEOUT                  = 240;
```
amisevsk: It may be worth it to leave this private and just have a similar field in `OpenShiftDeploymentCleaner`. There's no reason the timeout on scaling down deployments *needs* to be the same as the one for waiting for containers to start (in fact, we may want the timeout for scaling down to be less than 4 minutes, that seems like a long time). Ideally, the scale down happens in under 5 seconds.

amisevsk (changes requested):
Couple of comments. +1 for simplifying `OpenShiftConnector` with `KubernetesResourceUtil`

ibuziuk:
@amisevsk btw, it is technically just cherry-picking of https://github.com/eclipse/che/pull/5157 ;-)

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftDeploymentCleaner.java
ibuziuk: there should not be PV issues since this change was applied in `openshift-connector` rebased, but grace termination period is taken into account

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftDeploymentCleaner.java
ibuziuk: ok, lemme change the order

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
ibuziuk: @amisevsk good point, but I'm not sure about 5 seconds - in che starter[1] stop timeout is 150 ms  


https://github.com/redhat-developer/che-starter/blob/master/src/main/resources/application.properties#L21

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftDeploymentCleaner.java
ibuziuk: @amisevsk btw in master I do not see that we use "terminationGracePeriodSeconds", So while testing it on minishift scale down does not happen fast and require some time - this might be the main reason of quota limit issue that was reported today

ibuziuk: I think this PR is not applied to master https://github.com/eclipse/che/pull/5007
But this is a separate conversation, I will provide another PR for that

ibuziuk:
@amisevsk PR has been updated

l0rd:
@ibuziuk can you please test on OSIO if this PR solves https://github.com/redhat-developer/rh-che/issues/217?

ibuziuk:
@l0rd I'm quite sure it will not because we need to also apply - https://github.com/eclipse/che/pull/5007

ibuziuk:
PR for grace period has been sent - https://github.com/eclipse/che/pull/5773

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftDeploymentCleaner.java
amisevsk: Sounds good - thanks @ibuziuk !

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
amisevsk: well, 150 ms is under 5 seconds ;)

amisevsk (approved):

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
ibuziuk: typo, 150 seconds ;-)

l0rd (approved):
LGTM

Issue #5439, opened 2017-06-22
Implement mechanism for providing unique ids for json rpc clients


sleshchenko:
Now id is a part of endpoint address and generation of it is a responsibility of the client. But there is not a way to guarantee that different clients generate a unique identifier.

There are few ways to be sure that different clients use unique identifier:
1. Implement providing identifier from a server when the connection is set up and implement a mechanism to reconnect as existing client. Something like:
- Client connects to a websocket endpoint;
- Server establishes connection and send generated id to client;
- Client receives id and store it;
- If the connection is interrupted, client reconnects and somehow specify its identifier (like query param or by sending a special message).

2. Implement rest service that will generate a unique identifier for clients. Then:
- Client request identifier for itself and store it;
- Client constructs endpoint path by adding received identifier;

3. Also we can investigate if we could robustly identify client without any interaction with him, for example getting client's unique identifier from websocket connection session or something like that.

Second way looks like easier to implement but there still is no guarantee that clients will use implemented service for generation of identifiers. And the second approach force clients to use an identifier on another hand first approach provides identifiers just like additional functionality for reconnection and clients don't have to use it.

benoitf:
what is the implementation choice that has been picked up ? between choice 1, 2 and 3 ?

dkulieshov:
@benoitf It is not yet decided completely, there appeared some technical impediments so we will probably have something in the middle of the first and the second option. I will provide brief spec draft as soon as the prototype will be tested.

benoitf:
@dkuleshov thx for the answer, as I noticed that it was in the current-sprint/in-progress I thought it was decided

dkulieshov:
Adding brief specification from the PR, there is more info in comment in PR itself

### What does this PR do?
This PR provides a new way of interaction between websocket server-side and client-side. 

In order to identify a client on server-side to associate with it some processes or data we must have a unique client ID. In previous version client itself was responsible for ID generation and providing to server-side, that approach had its obvious shortcomings. 

The new approach rearrange responsibilities in a such way that server-side now have a centralized way of generating a unique identifier, while client can get that identifier and use it for its identification further. The JSON-RPC service at the moment has the following api:
 - method name: `websocketIdService/getId`
   method params: none
   method result: **`String`**
   description: this method must be used by a client in order to get identifier that will also be associated with current websocket session.
   example: 

**request**
```json
{
    "jsonrpc": "2.0",
    "method": "websocketIdService/getId",
    "id": 0
}
```
[...]

PR #5775, opened 2017-07-21
Implement mechanism for providing unique ids for JSON-RPC clients

dkulieshov:
### What does this PR do?
This PR provides a new way of interaction between websocket server-side and client-side. 

In order to identify a client on server-side to associate with it some processes or data we must have a unique client ID. In previous version client itself was responsible for ID generation and providing to server-side, that approach had its obvious shortcomings. 

The new approach rearrange responsibilities in a such way that server-side now have a centralized way of generating a unique identifier, while client can get that identifier and use it for its identification further. The JSON-RPC service at the moment has the following api:
 - method name: `websocketIdService/getId`
   method params: none
   method result: **`String`**
   description: this method must be used by a client in order to get identifier that will also be associated with current websocket session.
   example: 

**request**
```json
{
    "jsonrpc": "2.0",
    "method": "websocketIdService/getId",
    "id": 0
}
```
   **response**
```json
{
    "jsonrpc": "2.0",
    "result": "529345982734",
    "id": 0
}
```
 - method name: `websocketIdService/setId`
   method params: **`String`**
   method result: none
   description: this method must be used by a client in order to make the provided client identifier associated with current websocket session
   example: 

**request**
```json
{
    "jsonrpc": "2.0",
    "method": "websocketIdService/setId",
    "params": "529345982734"
}
```
So briefly the workflow is the following:
- Initially client has no websocket identifier
- When it establishes a first websocket connections and the fist websocket session is initialized it calls a corresponding JSON-RPC service (`websocketIdService/getId`) to get a unique identifier
- For all further websocket connection establishments (e.g. reconnection, new endpoint, etc.) it must call a corresponding JSON-RPC service (`websocketIdService/setId`) in order to identify itself in a newly initialized sessions.

https://github.com/eclipse/che/issues/5439

#### Changelog
Provided a way a JSON-RPC client can get ID for websocket interactions

#### Release Notes
Provided a way a JSON-RPC client can get ID for websocket interactions


musienko-maxim:
The selenium test session did not find regression

musienko-maxim (approved):

garagatyi:
This is very nice feature. But it is not clear whether a very simple client can connect and not send get/set id requests. And what would be the behavior of the server in case a client hasn't sent setId request.

voievodin:
Have you considered using an existing approach for identification which implemented 
within client<->exec-agent communication? 
A short description of it:

#### Connect flow

 Client connects to the server let's say using url:
```
ws://host:port/ws-endpoint
```

The first message client receives is [HELLO](https://github.com/eclipse/che/blob/master/agents/go-agents/docs/events.md#connected) 
message from server which provides unique identifier for the client.

```json
{
  "jsonrpc": "2.0",
  "method": "connected",
  "params": {
    "time": "2016-09-24T16:40:05.098478609+03:00",
    "channel": "channel-1",
    "text": "Hello!"
  }
}
```

#### Reconnect flow

Client connects to the server sending identifier as query parameter
```
ws://host:port/ws-endpoint?channel=channel-1
```

I you haven't please consider it, otherwise can you please describe its drawbacks comparing to the approach you implemented?

dkulieshov:
@garagatyi I must apologize for that I did not mention it in the description. In the most basic case when a client establishes a connection server side generates a unique ID which is associated with a session and no client interaction is required at all. However if you need to identify the same client for several websocket sessions (for example when you have several endpoints in a single application) you obviously will need some cross-session identifier and for those purposes we have websocket id service.

So answering your question, when client does not call `setId` it should work just fine within the single websocket session. When client's logic involves several websocket sessions you have to use cross-session ID.

dkulieshov:
@evoevodin 
> I you haven't please consider it, otherwise can you please describe its drawbacks comparing to the approach you implemented?

For me, common sense and logic dictate that if someone (you) offers an alternative solution, then it is he (you) who should explain the advantages of his (yours) option, rather than require of initiator (me) to find shortcomings of an alternative (yours) approach. 

Concentrating on the construct, I would like to say the following: the option you are mentioning looks rather viable for me but I'm not seeing it's strong benefits. I mean calling a JSON-RPC method is a client's obligation but setting a query parameter is also a client's obligation. "Hello message" is okay when client need's an ID, but when client already has that ID it's an obvious overhead (e.g. when client reconnects).

garagatyi:
@dkuleshov Thank you for the explanation! It is what had I expected from UX perspective.

voievodin:
@dkuleshov thank you for the answer. 
It was thoughtless from my side to ask you for such unreasonable comparison, next time i will try to spend more time and bring some real evidences.

vparfonov (approved):

azatsarynnyy (approved):

RomanNikitenko (approved):

skabashnyuk (approved):

svor (approved):

riuvshin (approved):

gazarenkov (approved):

evidolob (approved):

musienko-maxim:
Rerunning of the functional tests after applying new changes -  did not  find new defects

PR #5794, opened 2017-07-24
Various fixes and cleanups in SPI branch

garagatyi:
### What does this PR do?
Adds several fixes, code cleanups, refactorings split in separate commits.

Here is the list of changes:
- Add Alex Garagatyi as code owner for docker infra
- Fixed usage of Port mappings in docker infra
- Fix missing setting of warning field in EnvironmentImpl object
- Fix containers start ordering
- Fix ConcurrentModificationException in Docker infra
- Make Bootstrapper do nothing if there are no installers in a machine
- Add possibility to use PidMode field in Docker infra to join container
- Rework InfrastructureProvisioner impls. Removed DefaultInfrastructureProvisioner. Moved its logic to LocalCheInfrastructureProvisioner.
 - Make InfrastructureProvisioner accept EnvironmentImpl instead of Environment to ease modification of environment configuration.
- Refactored LocalCheInfrastructureProvisioner to clean code.


skabashnyuk (approved):

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/local/LocalCheInfrastructureProvisioner.java
```diff
@@ -32,20 +36,21 @@
...
 
 /**
  * Infrastructure provisioner that uses volumes for workspaces projects storing because it relies on local Che usage.
  *
  * @author Alexander Garagatyi
  */
-public class LocalCheInfrastructureProvisioner extends DefaultInfrastructureProvisioner {
-    private static final List<String> SNAPSHOT_EXCLUDED_DIRECTORIES = Arrays.asList("/tmp");
+// TODO should add default RAM settings, hosts, and all other stuff from MachineStarter
+public class LocalCheInfrastructureProvisioner implements InfrastructureProvisioner {
```
akorneta: Looks like it should be singleton or I missed something?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/bootstrap/AbstractBootstrapper.java
```diff
@@ -86,6 +86,9 @@ public void bootstrap() throws InfrastructureException {
         if (finishEventFuture != null) {
             throw new IllegalStateException("Bootstrap method must be called only once.");
         }
+        if (!hasInstallersToBootstrap()) {
```
sleshchenko: WDYT about moving such logic to InternalRuntime and do not bootstrap at all if installers are missed

garagatyi: Bootstrapper works with each machine, not the whole runtime, InternalRuntime knows nothing about bootstrapper, the bootstrapper is not part of SPI contract. So I do think it is better to not introduce bootstrapper in InternalRuntime

sleshchenko: I mean changes in DockerInternalRuntime
```
List<Intaller> installers = machineCfg.getInstallers();
if (!installers.isEmpty()) {
    bootstrapperFactory.create(name, getContext().getIdentity(), installers, machine).bootstrap();
}
```

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/model/DockerContainerConfig.java
```diff
@@ -135,6 +139,7 @@ public DockerContainerConfig withBuild(DockerBuildContext build) {
     /**
      * Override the default entrypoint.
      */
+    @Nullable
```
sleshchenko, line 142: Are there any reasons to make only this list nullable?

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/local/LocalCheInfrastructureProvisioner.java
```diff
@@ -82,56 +88,93 @@ public LocalCheInfrastructureProvisioner(InstallerConfigApplier installerConfigA
...
-        try {
-            projectFolderVolume = String.format("%s:%s%s",
-                                                workspaceFolderPathProvider.getPath(identity.getWorkspaceId()),
-                                                projectFolderPath, projectsVolumeOptions);
-        } catch (IOException e) {
-            throw new InfrastructureException("Error occurred on resolving path to files of workspace " +
-                                              identity.getWorkspaceId());
+            for (String installer : machineConfig.getInstallers()) {
+                switch (installer) {
+                    case "org.eclipse.che.terminal":
```
voievodin: I've seen a few places where installer identifiers used as string literals, 
what if we create constants for them?

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/local/LocalCheInfrastructureProvisioner.java
```diff
@@ -82,56 +88,93 @@ public LocalCheInfrastructureProvisioner(InstallerConfigApplier installerConfigA
...
+    public void provision(EnvironmentImpl envConfig,
                           DockerEnvironment internalEnv,
                           RuntimeIdentity identity)
             throws InfrastructureException {
         String devMachineName = getDevMachineName(envConfig);
         if (devMachineName == null) {
             throw new InfrastructureException("ws-machine is not found on installers applying");
         }
+        excludeDirsFromSnapshot(internalEnv);
+        addProjects(internalEnv, identity.getWorkspaceId(), devMachineName);
```
sleshchenko, line 102: I don't think that usage of `addProjects` and `addInstallers` here are clear enough because one add volumes, another one adds volumes and env variables. But you can not rework it because they are private methods.

sleshchenko (approved):
OK for me

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/local/LocalCheInfrastructureProvisioner.java
```diff
@@ -82,56 +88,93 @@ public LocalCheInfrastructureProvisioner(InstallerConfigApplier installerConfigA
...
-        // add bind-mount volume for projects in a workspace
-        String projectFolderVolume;
-        try {
-            projectFolderVolume = String.format("%s:%s%s",
-                                                workspaceFolderPathProvider.getPath(identity.getWorkspaceId()),
-                                                projectFolderPath, projectsVolumeOptions);
-        } catch (IOException e) {
-            throw new InfrastructureException("Error occurred on resolving path to files of workspace " +
-                                              identity.getWorkspaceId());
+            for (String installer : machineConfig.getInstallers()) {
```
voievodin: WDYT about
```java
HashSet<String> installers = new HashSet<>(machineConfig.getInstallers());
if (installers.containsKey("org.eclipse.che.terminal")) {
  containerConfig.getVolumes().add(terminalVolumeProvider.get());
}
```
The benefit i see is that this approach allows to eventually switch installers provisioning from static to dynamic

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/bootstrap/AbstractBootstrapper.java
voievodin: i would do both
1. change the contract of the bootstrapper, but rather to throw exception when there is nothing to install, instead of silent quit. What if we do it directly in AbstractBootstrapper constructor? So the contract is that bootstrapper cannot be created with empty list of installers.
2. change the code in internal runtimes(like Sergii suggests) so bootstrapper is not created at all.

garagatyi: I like Sergii's idea and not sure about throwing an exception in the case of an empty list. I'll go with his suggestion.

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/model/DockerContainerConfig.java
garagatyi: These specific fields should be null by default because non-null values can break start of container

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/local/LocalCheInfrastructureProvisioner.java
garagatyi: I agree that would be good to have constants. Do you have an idea on where to place them?

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/local/LocalCheInfrastructureProvisioner.java
garagatyi: These methods do what their names say. `addProjects` adds everything needed for projects folder. `addInstallers` adds everything needed for installers, which may include volumes, env variables, etc. 
This was the motivation for the creation of these methods.

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/local/LocalCheInfrastructureProvisioner.java
garagatyi: Well, it doesn't do anything heavy, so it doesn't have to be a singleton. I'll add the annotation.

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/local/LocalCheInfrastructureProvisioner.java
garagatyi: Looks good! Do we really need to copy the map?

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/local/LocalCheInfrastructureProvisioner.java
voievodin: Maybe in classes like `TerminalInstaller`.

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/local/LocalCheInfrastructureProvisioner.java
voievodin: which map?

akorneta (approved):

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/local/LocalCheInfrastructureProvisioner.java
garagatyi: The problem here is that there are usages on both server and client. Should we use the constant on the server side only?

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/local/LocalCheInfrastructureProvisioner.java
garagatyi: Sorry for confusion, I meant set:
`HashSet<String> installers = new HashSet<>(machineConfig.getInstallers());`

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/local/LocalCheInfrastructureProvisioner.java
voievodin: Using constants on serverside only is better than no constants at all, i would say that client may define its own constants for such purpose. Obviously another approach would be to create shared module for clients(each agent has it's own shared module) and put the constants there which will make class available for gwt & typescript clients.

garagatyi: I will consider this approach in the further development of this module. But I'm not ready to implement it right now

voievodin (approved):

Issue #5687, opened 2017-07-17
Link recipe and configuration machines


sleshchenko:
Launching and resolving (already started) machines is implemented by OpenShift spi in following way:
1. Parse openshift recipe (aka kubernetes list in yaml format).
2. Create all objects from the recipe.
3. Bootstrap all containers with hardcoded installers.
4. For fetching all started machines - fetch all pods in openshift project and resolve servers by existing services and routes.

It is needed to rework this approach to:
- link recipe machines and configuration that are specified in the environment;
- bootstrap only specified in configuration installers;
- maybe store machines not to analyze openshift objects each time;
- publish events about machine status changes.

bmicklea:
@sleshchenko I think this is an enhancement, but can you apply the correct label?

PR #5796, opened 2017-07-24
CHE-5687 Add usage of configured installers

sleshchenko:
### What does this PR do?
Add usage of configured installers instead of hard coded ones.
Move resolving of servers from InternalRuntime to Machine.
Add retrieving all required information for InternalRuntime from Сontext instead of setting it while InternalRuntime instance creation.

https://github.com/eclipse/che/issues/5687

### DIFF
--- a/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenshiftMachine.java
+++ b/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenshiftMachine.java
@@ -10,17 +10,27 @@
  *******************************************************************************/
 package org.eclipse.che.workspace.infrastructure.openshift;
 
+import io.fabric8.kubernetes.api.model.Container;
+import io.fabric8.kubernetes.api.model.ContainerPort;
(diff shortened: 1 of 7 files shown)


infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenshiftMachine.java
```diff
@@ -67,17 +84,117 @@ public String getName() {
 
     @Override
     public Map<String, ? extends Server> getServers() {
-        //TODO https://github.com/eclipse/che/issues/5687
-        return new HashMap<>();
+        //TODO https://github.com/eclipse/che/issues/5688
+        try (OpenShiftClient client = clientFactory.create()) {
+            List<Service> services = client.services()
```
voievodin: It seems that services are used only to get _matchedServices_, so maybe it's better to get rid of the codeblock below and use stream + filter to get _matchedServices_ here?

akorneta (approved):

voievodin (approved):

garagatyi (approved):

Issue #3221, opened 2016-12-01
Git compare fails on huge projects


mmorhun:
When try to compare edited file in huge project with original repository version then internal server error will occur. 

**Reproduction Steps:**
1. Create workspace and import Eclipse Che project (configure as maven)
2. Open any java file and change something
3. Use Menu -> Git -> Compare -> Compare with latest repository version

In case of comparison of whole project Git compare window is appeared, but the same error happens when try to compare a file from the list.

**Expected behavior:**
Compare widget is appeared

**Observed behavior:**
Internal server error in events tab

**Che version:**    5.0.0-M8-SNAPSHOT
**OS and version:**  Ubuntu 16.04
**Docker version:** Version: 1.12.3 API version: 1.24
**Che install:**   tomcat server (zip)

**Additional information:**
* Problem can be reliably reproduced, doesn't happen randomly: Yes

slemeur:
Cannot reproduce

mmorhun:
Still actual, but now the same behaviour even on comparison of whole project (no compare widget appears, but errors in logs).
![screenshot from 2017-04-19 09-00-15](https://cloud.githubusercontent.com/assets/15607393/25165345/b2b199e8-24de-11e7-886c-a1ad6d1581e8.png)

slemeur:
Thanks @mmorhun

ghost:
Still an issue in 5.15-SNAPSHOT. I have imported Che into a Java workspace, edited one class, hit compare and got an internal server error. I think this should be fixed since the feature is unusable for real life enterprise level projects.

mmorhun:
Fixed

PR #5799, opened 2017-07-25
CHE-3221: Fix git compare on submodules.

mmorhun:
### What does this PR do?
Fixes bug which makes git compare with latest version failed if project submodule is active.

https://github.com/eclipse/che/issues/3221

#### Changelog
Fix git compare when using it with the latest version and a project with submodules.

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/resources/Resource.java
+++ b/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/resources/Resource.java
@@ -483,6 +483,16 @@ public interface Resource extends Comparable<Resource> {
     Project getProject();
 
     /**
+     * Returns the {@link Container} which is root project to this resource or {@code null}.
+     * Returns itself for projects without submodules.
+     *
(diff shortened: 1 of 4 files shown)


vzhukovs:
It would be better to move implementation of this method into this utility class `org.eclipse.che.ide.ext.git.client.GitUtil`. For now there is no need to add additional method into resource interface.

mmorhun:
To me it is not clear better or not to move `getRootProject` method into `GitUtils`. In one hand it is used now only in one place inside git plugin, but in the other hand this method theoretically could be needed by other stuff and to avoid code duplicates would be better to keep it in the `Resource` interface.
I am not against to place it in `GitUtils`, and if you think that moving to `GitUtils` will be better - let be so.

vzhukovs (approved):

vinokurig (approved):

bmicklea (approved):
updated changelog entry.

vparfonov (approved):

Issue #5728, opened 2017-07-18
Run JUnit tests which were included to @Suite.SuiteClasses annotation


svor:
Need recognize tests which describes into Test Suite Class and add ability to run their.
Example of test suite class:

```
import org.junit.runner.RunWith;
import org.junit.runners.Suite;

@RunWith(Suite.class)

@Suite.SuiteClasses({
   TestJunit1.class,
   TestJunit2.class
})

public class JunitTestSuite {   
}
```

Issue #5725, opened 2017-07-18
Make possible to run all tests from project and package for JUnit and TestNG frameworks


svor:
Now we have an ability to run JUnit and TestNG tests from test class by using new Test Runer API.
Need to make possible to run all tests from the project or package.

Issue #5422, opened 2017-06-21
Need to investigate ability re-compiles only those portions of a program that have been modified


svor:
Now we have an ability to run/debug tests, but if make some changes in the project firstly need to compile project for getting actual state and then run/debug tests. 

So main idea of this issue is compile only changed part of code automatically before running/debugging tests.

It will help us save a lot of time for example: for compiling some modules in che project take around 45 sec, but if we change only one class we don't need recompile all module just changed and related classes it can take less then 1 second. 

It is a next step after https://github.com/eclipse/che/issues/5316

vparfonov:
Need to investigate ability re-compiles only those portions of a program that have been modified. (Incremental compiler)

vparfonov:
Depends on https://github.com/eclipse/che/issues/6157

gorkem:
Closing as Che 6 editor is no longer under active development

PR #5804, opened 2017-07-25
CHE-5725: Run all java tests from the project or package

svor:
### What does this PR do?

- Run java tests (JUnit4, TestNg) from the Project Explorer
  - Can run one test class

  - Can run tests from package
package
  - Can run tests from project
project

For this case (when the editor isn't active) we have  actions:
selection_001
selection_003

- Run java test from the editor. When the editor is active , the test runner will automatically detect test framework. For this case we have next set of actions:
selection_002

https://github.com/eclipse/che/issues/5725
https://github.com/eclipse/che/issues/5728

#### Changelog

Run all java tests from the project or package

#### Release Notes

Run all java tests from the project or package


davidfestal (commented):
Nice, but it seems it still doesn't manage the on-the-fly compilation of projects before the first test run ?

svor:
Still don't have project compilation before test running. 
Have an issue for this (https://github.com/eclipse/che/issues/5422)

plugins/plugin-testing-java/plugin-testing-junit/che-plugin-testing-junit-ide/src/main/java/org/eclipse/che/plugin/testing/junit/ide/action/JUnitAbstractAction.java
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Abstract action for JUnit test actions.
+ */
+public abstract class JUnitAbstractAction extends RunDebugTestAbstractAction {
+    public static final String JUNIT_TEST_FRAMEWORK = "junit";
+
+    private AppContext appContext;
+    private boolean    isEnable;
+
+    protected TestExecutionContext.TestType testType;
```
vparfonov: TestType not good name it think, as i understand its context for runinig

plugins/plugin-testing-java/plugin-testing-junit/che-plugin-testing-junit-ide/src/main/java/org/eclipse/che/plugin/testing/junit/ide/action/RunJUnitTestAction.java
```diff
@@ -0,0 +1,110 @@
...
+        notification.setStatus(SUCCESS);
+        if (result.isSuccess()) {
+            notification.setTitle("Test runner executed successfully.");
+            testResultPresenter.handleResponse();
+        } else {
+            notification.setTitle("Test runner failed to execute.");
+        }
+    }
+
+    private void onTestRanUnsuccessfully(JsonRpcError exception, StatusNotification notification) {
```
vparfonov: onTestFail() ?

MatousJobanek:
Hi,
are the private methods in [TestingRPCService](https://github.com/eclipse/che/blob/master/wsagent/che-core-api-testing/src/main/java/org/eclipse/che/api/testing/server/TestingRPCService.java) still somewhere used? They seem to be like a zombie for me. The same for the method [AbstractJavaTestRunner#detectTests](https://github.com/eclipse/che/blob/master/plugins/plugin-testing-java/plugin-testing-classpath/src/main/java/org/eclipse/che/plugin/java/testing/AbstractJavaTestRunner.java#L58) (and the related methods) as you are using a new approach for retrieving classes, right?

svor:
Hi @MatousJobanek 
About `TestingRPCService` - it is a service which uses JSON RPC concepts. It register two methods (for detecting and running tests). It means if the client will send a request via websocket with name `RUN_TESTS_METHOD` - server will call method `org.eclipse.che.api.testing.server.TestingRPCService#runTests` the same about `TESTING_RPC_TEST_DETECTION_NAME` and `org.eclipse.che.api.testing.server.TestingRPCService#handleTestDetection`. 

 About `AbstractJavaTestRunner#detectTests` - this method analyzes opened test file and detects tests methods, and which framework we should use. This detection needs only for analyzing opened editor.

plugins/plugin-testing-java/plugin-testing-junit/che-plugin-testing-junit-ide/src/main/java/org/eclipse/che/plugin/testing/junit/ide/action/JUnitAbstractAction.java
svor: Changed to ContextType

plugins/plugin-testing-java/plugin-testing-junit/che-plugin-testing-junit-ide/src/main/java/org/eclipse/che/plugin/testing/junit/ide/action/RunJUnitTestAction.java
svor: It isn't exactly test fail. Renamed to `onTestRanFailed`

MatousJobanek:
Ah, I missed that. Thx for clarification @svor

MatousJobanek:
Have you think about extracting the code from `*TestFinder` classes (`TestNGTestFinder` & `JUnit4TestFinder`) to some shared module - most of the code seems to be same - you know, to bring some higher abstraction. Similar also for `createSuite` method - move most of the code to `AbstractJavaTestRunner` and then just providing a specific implementation of some method would define how the classes will be detected.

svor:
Make sense, I will do it. Thx!

vparfonov (approved):

svor:
ci-test

plugins/plugin-testing/che-plugin-testing-ide/src/main/java/org/eclipse/che/plugin/testing/ide/action/RunDebugTestAbstractAction.java
```diff
@@ -34,55 +35,59 @@
...
-                               SVGResource icon) {
+    private TextEditor          currentEditor;
+    private List<TestPosition>  testPosition;
+    private TestServiceClient   client;
+    private DtoFactory          dtoFactory;
+    private AppContext          appContext;
+    private NotificationManager notificationManager;
+
+    public boolean isEditorInFocus;
+    public boolean isEnable;
```
MatousJobanek: I just came across this part - is it really necessary to have these variables `public` when they are not constants?

MatousJobanek (commented):
Thank you for refactoring the code @svor This is exactly what I was expecting. Apart from the one comment inside of the code I have another proposal for refactoring/improvement. In last two days, I was playing with the IDE side and I noticed that the classes `JUnitAbstractAction` and `TestNGAbstractAction` are completely same. The same I can say about classes `Run*TestAction` and `Debug*TestAction`. Do you have any plan for merging the classes together?

svor:
@MatousJobanek Thank you for reviewing! I've made some optimizations with actions and I think now it's more clear.

evidolob (approved):

Issue #5814, opened 2017-07-26
Recovered che-server has different ip than the one defined in machine /etc/hosts


voievodin:
If che-server container is stopped or killed it might have different ip address from the one defined in the _/etc/hosts/_ in workspace containers. The problem here is that severs( like wsagent) which use che-host won't be able to access che-server after it is restarted and its ip address changed.

workspace container _/etc/hosts_ content
```
user@9eaf1acd3814:~$ cat /etc/hosts
127.0.0.1       localhost
::1     localhost ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
172.17.0.4      che-host
172.17.0.2      9eaf1acd3814
172.18.0.2      9eaf1acd3814
```
che-server network settings
```
"NetworkSettings": {
  "Bridge": "",
  "SandboxID": "55af3ac5098e76d069a7f0c38192f1605b61adcedfa741785dee8c15c083b542",
  "HairpinMode": false,
  "LinkLocalIPv6Address": "",
  "LinkLocalIPv6PrefixLen": 0,
  "Ports": {
    "8000/tcp": [
      {
        "HostIp": "0.0.0.0",
        "HostPort": "8000"
      }
    ],
    "8080/tcp": [
      {
        "HostIp": "0.0.0.0",
        "HostPort": "8080"
      }
    ]
  },
  "SandboxKey": "/var/run/docker/netns/55af3ac5098e",
  "SecondaryIPAddresses": null,
  "SecondaryIPv6Addresses": null,
  "EndpointID": "84c0c6304523670371df941015165e7b7087ec2721f958018def703966df43f7",
  "Gateway": "172.17.0.1",
  "GlobalIPv6Address": "",
  "GlobalIPv6PrefixLen": 0,
  "IPAddress": "172.17.0.4",
  "IPPrefixLen": 16,
  "IPv6Gateway": "",
  "Networks": {
    "bridge": {
      "IPAMConfig": null,
      "Links": null,
      "Aliases": null,
      "NetworkID": "33522275ceb39508c4f7a0e8083c9a759b71aed837f240bf9d5dc50d1b24dd03",
      "EndpointID": "84c0c6304523670371df941015165e7b7087ec2721f958018def703966df43f7",
      "Gateway": "172.17.0.1",
      "IPAddress": "172.17.0.4",
      "IPPrefixLen": 16,
[...]

PR #5815, opened 2017-07-26
Reconnect to push-logs-endpoint if connection is lost

### LINKED ISSUE #5675: Recover servers logs streaming
Bootstrapper continues streaming logs after workspace is started and there is at least one server.
But right now there is no way to _reconnect_ to the logs stream after master is restarted.
I see  several ways to solve this issue:
1. Bootstrapper attempts to reconnect to master once disconnected.
2. Master asks bootstrapper to continue logs streaming sending _push-logs-endpoint_ to it.
3. Master connects to bootstrapper and asks to stream logs using established connection.

I would start from the option 1, as it is the simplest way to achieve recovery.

The logs format is defined in the bootstrapper [spec](https://github.com/eclipse/che/issues/4096#issuecomment-283067971). 

Needed for #5097
Needs #5814

voievodin:
Added ability to reconnect to logs endpoint if previously established connection is lost.
Default reconnect period is _10sec_ but it can be configured with cli parameter:
```
  -logs-endpoint-reconnect-period int
        Time(in seconds) between attempts to reconnect to push-logs-endpoint.
        Bootstrapper tries to reconnect to push-logs-endpoint when previously 
        established connection lost (default 10)
```
It partly resolves #5675.
Another part should be resolved by #5814.

### DIFF
--- a/agents/go-agents/bootstrapper/booter/booter.go
+++ b/agents/go-agents/bootstrapper/booter/booter.go
@@ -31,21 +31,28 @@ const (
 )
 
 var (
-	installers       []Installer
-	runtimeID        RuntimeID
-	machineName      string
-	installerTimeout time.Duration
-	checkPeriod      time.Duration
+	installers                  []Installer
+	runtimeID                   RuntimeID
+	machineName                 string
+	installerTimeout            time.Duration
+	checkPeriod                 time.Duration
+	logsEndpointReconnectPeriod time.Duration
 
 	bus = event.NewBus()
 )
 
+// Connector encloses implementation specific jsonrpc connection establishment.
+type Connector interface {
+	Connect() (*jsonrpc.Tunnel, error)
+}
+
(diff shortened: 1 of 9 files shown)


agents/go-agents/bootstrapper/booter/booter.go
```diff
@@ -288,6 +306,20 @@ func (tb *tunnelBroadcaster) IsDone() bool {
 
 func (tb *tunnelBroadcaster) Close() { tb.tunnel.Close() }
 
+func (tb *tunnelBroadcaster) goReconnect() {
+	go func() {
+		<-time.After(logsEndpointReconnectPeriod)
```
garagatyi: I think it can be replaced with time.Sleep()

garagatyi (approved):
Nice feature!

akorneta (approved):

PR #5817, opened 2017-07-27
CHE-5335: Add Git branch/revision link to project name

### LINKED ISSUE #5335: Show current git branch in the IDE
Show current git branch name in the editors bottom info panel.
git_branch
Clicking on the branch name must show Git branches dialog.

This issue is a part of epic: https://github.com/eclipse/che/issues/5128 
new git panel

vinokurig:
### What does this PR do?
Add git branch/revision link near project name in the project explorer that opens 'Git branches' window.
git branch

fixes https://github.com/eclipse/che/issues/5335

#### Changelog
Added git branch/revision link after project name in the explorer that opens the 'Git branches' window.

#### Release Notes
(Add to a git improvements RN section)

#### Docs PR

N/A


ide/che-core-ide-ui/src/main/java/org/eclipse/che/ide/ui/smartTree/TreeStyles.java
```diff
@@ -36,6 +36,8 @@
 
         String infoTextContainer();
 
+        String vcsHeadLinkContainer();
```
vzhukovs: It would be better to move this style declaration to general resource bundle `org.eclipse.che.ide.Resources.CoreCss` and make it accessible for every component which wants to colorize standard html link. WDYT?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/GitCheckoutHandler.java
```diff
@@ -76,5 +85,8 @@ public void apply(String endpointId, GitCheckoutEventDto dto) {
                 break;
             }
         }
+
+        //Update project attributes from server.
+        resourceManagerFactory.newResourceManager(appContext.getDevMachine()).getWorkspaceProjects();
```
vzhukovs: Use `appContext.getWorkspaceRoot().synchronize();` instead of this construction.

`org.eclipse.che.ide.resources.impl.ResourceManager.ResourceManagerFactory` it's an internal factory to create a new instance of Resource Manager which not update existing state.

bmicklea (approved):
Tweaked changelog and RN in description

ide/che-core-ide-ui/src/main/java/org/eclipse/che/ide/ui/smartTree/TreeStyles.java
vinokurig: But this CSS is needed only for tree node container. It does not related to general html link.

vzhukovs: Ah, then ok.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/GitCheckoutHandler.java
vinokurig: Done, fixed the `synchronize()` method.

vzhukovs (approved):

slemeur:
It's generally ok. 

I see few problems with the ability to switch git branch provided by the hyperlink on the branch name:
- when I'm exploring the workspace's projects, user might expect to click on the project's "row" to select it (and expand/collapse it). I think it will be a bad UX to force the user to click on a specific area of the row to get the project selected/expanded - the user will easily get trapped and open the git's branch popup.
- This behaviour (expand/collapse) will be inconsistent with exploring other nodes of the project tree
- Visually the blue color is used for primary actions in the UI. User's eyes will be more attracted by the git branch than the project name.

For these reasons, I don't think it's a good idea to allow the user to click on the branch name from the project explorer.

vinokurig:
@slemeur
All the raw area except the link is still does the same as an old item. The link becomes underlined on mouse hover so it is inuitivly clear that it is a link and you should move the cursor to the rest of the row's area.

You are right that the link occupies some usefull space and attracts users atantion, but in the other hand, user receives very usefull ability to view / checkout other branch directly from the IDE, without exploring the menu.

BTW it was discussed in the issue: https://github.com/eclipse/che/issues/5335#issuecomment-311988263

slemeur:
@vinokurig : I can understand why you think that it is useful, but you are putting the "git switch branch" as a primary action of the project explorer. We should be very careful with that, as it is not one of the primary goal of the explorer. 
The primary goal of the project explorer is to allow the user to navigate into the projects source code and _display_ meaningful information (file's type, git status, git branch...). The primary actions are related to the navigation - all the other "potentials" actions must be secondary and should not impact the primary actions. 
This action, as a link in the expand/collapse area, really close to the project's name, will impact the usability of the primary actions and does not need to be prominent in the project explorer: available with a right-click action, from the menu is fine and will benefit the UX.

Moreover, you are describing the behaviour to be like an hyperlink. If this is a hyperlink, in user mind, it is expected to open a new page (in the same browser tab or in a new browser tab) - not to bring a popup.

I would simply stick with the initial specification and requirements.

vkuznyetsov:
+1 on using git branch name in PE for quick access to switch branch. The action comes after clicking the git branch name is clear and expected (operating with a git branch). I guess pros here more than cons.

tolusha:
I vote for  @slemeur, his explanation make sense

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/explorer/project/ProjectExplorerPresenter.java
```diff
@@ -438,6 +443,15 @@ public boolean isShowHiddenFiles() {
         return hiddenFilesAreShown;
     }
 
+    @Override
+    public void onVcsBranchClicked(Project project) {
+        showVcsBranchActionProviders.get()
+                                    .stream()
+                                    .filter(provider -> provider.getVcsName().equals(project.getAttribute("vcs.provider.name")))
```
dkulieshov: I would recommend to extract `project.getAttribute("vcs.provider.name")` to calculate it only once instead of for each action provider in a stream, what do you think?

dkulieshov (approved):

dkulieshov:
+1 for @slemeur

dkulieshov:
I actually think we should consider how it is done in IntelliJ and Eclipse IDE as they provide valuable user experience

bmicklea:
The branch info (not link) is helpful in the project explorer. However to switch branches there should be a switch git branch action in the right-click context menu. All other actions are in the right-click context menu so that's where branch switch belongs.

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/part/explorer/project/ProjectExplorerPresenter.java
vinokurig: done

vinokurig:
git_branch

@slemeur reworked the link to text label, please review

slemeur:
thanks @vinokurig. I think the branch name should be displayed differently. I would use italic and a different color (#7A8088). 
See in the following mockup:
ide 6 - git in editor view

vinokurig:
@slemeur Please provide color for white theme

vinokurig:
@slemeur I think this color is too dark for Dark theme
dark
light

How about `#a7a0a0` for Dark theme
dark_new
WDYT?

slemeur:
Your proposal is perfect for me (we'll have to update that with the Che 6 theme cc @vitaliy-guliy). Thanks

vinokurig:
@slemeur I have changed the color, please review the PR

slemeur (approved):

Issue #5818, opened 2017-07-27
[Regression] The code validation is not available for language servers


musienko-maxim:
**Preconditions:** Create a workspace based on C# stack. Create the Asp. application from the Wizard. 

**Reproduction Steps:** 
1. Open the Program.cs.  After init  LS, do some error in the code.
**OS and version:**    Fedora 25, chrome browser, nightly.codenvy-stg.com/ server

**Expected result:**  The error marker should appear in the line with error
**Actual result:**  There is no any markers
**Additional information:**  The problem actual for PHP and Python language servers

The related test: CheckMainFeatureForLanguageTest

tsmaeder:
I can confirm that for python files. First inspection reveals that the python language server is not started. Investigating.

tsmaeder:
Are the language server agents enabled in the workspace in question?

tsmaeder:
Works for me when enabling the the language servers in the workspace. What behaviour are you expecting?

Ohrimenko1988:
This problem also actual for TypeScript language server

ghost:
For Python, hover and auto completion worked for me But not error marking.

musienko-maxim:
The language server agents was enabled. I think the LS process also was run.

tsmaeder:
The notification channel for diagnostics is no longer working. My suspicion is that the registration mechanism in PublishDiagnosticsParamsJsonRpcTransmitter and PublishDiagnosticsReceiver got broken (maybe by the work on unique endpoint ids?). When the request transmitter tries to transmit the diagnostics even, there is one websocket connection registered, but the id is not the one that got recorded in PublishDiagnosticsParamsJsonRpcTransmitter.

tsmaeder:
One more thing: this worked before I rebased on master today. So the problem must have been introduced recently.

dkulieshov:
Quite possible that the new endpoint ID's mechanism may have impact however at first glance I can't tell how exactly as, for example, file watching is okay (according to QA reports) while it uses similar endpoint ID based subscription. I'm going to debug to get more info.

dkulieshov:
[...]

PR #5827, opened 2017-07-27
Fix for the code validation is not available for language servers 

dkulieshov:

### What does this PR do?
The description of the origin and solution is [here](https://github.com/eclipse/che/issues/5818#issuecomment-318431161). Briefly the idea is to replace `wsIdService/setId` JSON-RPC method calls with query params in order to set websocket client identifier before any request is being processed (previously it only could be set in parallel with processing other JSON-RPC requests). So we may have several scenarios:
- Client has or has no websocket ID but it is not expected to have several websocket sessions - client simply establishes a connection, no extra work for client. ID is generated automatically on server side and is used internally, within a single websocket session client is identified.
- Client has no websocket ID while it is expected to have several websocket sessions - after websocket connection is established client calls `wsIdService/getId` (not necessarily right on session opening, obviously depends on business logics) and gets an ID that can be used further for cross-session identification.
- Client has websocket ID while it is expected to have several websocket sessions - it identifies itself (only) during initialization of a websocket session by setting `clientId` query param: `ws://host:port/websocket?clientId=1283719824`

https://github.com/eclipse/che/issues/5818

#### Changelog
Fixed errors in LS diagnostics propagation over JSON-RPC by changing websocket client identification mechanism.

### DIFF
--- a/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/websocket/impl/BasicWebSocketEndpoint.java
+++ b/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/websocket/impl/BasicWebSocketEndpoint.java
@@ -20,6 +20,13 @@ import javax.websocket.OnError;
 import javax.websocket.OnMessage;
 import javax.websocket.OnOpen;
 import javax.websocket.Session;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+import static java.util.Collections.emptyMap;
+import static java.util.Collections.unmodifiableMap;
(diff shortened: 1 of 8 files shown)


dkulieshov:
@vparfonov If I remember right websocket initialization params in the way they are implemented here won't properly work for situations when we add security/authorization token, so I guess the code should be slightly reworked to support such cases.

I mean [this](https://github.com/eclipse/che/pull/5827/files#diff-92e336cd2d05aa1bb7672b08c83a0e70R96) line will not properly work if there are other query parameters except `clientId` however that should not be a problem I thing.

vparfonov (approved):

PR #5835, opened 2017-07-28
Avoid using `sudo` in agent launchers when the current user is not a sudoer

l0rd:
### What does this PR do?
Add a check in agent launchers to verify if the user is a sudoer. Then, if the user is **not a sudoer**, variable `$SUDO` is kept empty.

To check if a user is a sudoer we use the following bash function:
```bash
is_current_user_sudoer() {
    sudo -n true >& /dev/null && return 0 || return 1
}
```

I have added some [bats](https://github.com/sstephenson/bats) files to tests if `is_current_user_sudoer()` works fine with all eclipse/che workspace docker images.

This modification has been tested on both Docker and OpenShift (creation of workspace).  

Starting an agent with a user that was not a sudoer (e.g. UID 10000) was impossible.

#### Changelog
Allowing agents bootstrap by non privileged users

#### Release Notes
Allowing agents bootstrap by non privileged users


agents/exec/src/main/resources/org.eclipse.che.exec.script.sh
```diff
@@ -9,6 +9,15 @@
...
 #
 
+
+is_current_user_root() {
+    test "$(id -u)" = 0 && return 0 || return 1
+}
+
+is_current_user_sudoer() {
+    sudo -n true >& /dev/null && return 0 || return 1
+}
```
benoitf: >& is bash
you should use
```
 is_current_user_sudoer() {
     sudo -n true > /dev/null 2>&1 && return 0 || return 1
 }
```


and maybe it could be simplified to
```
is_current_user_sudoer() {
  sudo -n true > /dev/null 2>&1
}
```

agents/exec/src/main/resources/org.eclipse.che.exec.script.sh
```diff
@@ -9,6 +9,15 @@
 #   Codenvy, S.A. - initial API and implementation
 #
 
+
+is_current_user_root() {
+    test "$(id -u)" = 0 && return 0 || return 1
+}
+
```
benoitf: maybe we could avoid the return operations ?
```
is_current_user_root() {
  test "$(id -u)" = 0
}
```

agents/exec/src/main/resources/org.eclipse.che.exec.script.sh
```diff
@@ -23,7 +32,7 @@ if [ ${CURL_INSTALLED} = false ] && [ ${WGET_INSTALLED} = false ]; then
   CURL_INSTALLED=true
 fi
 
-test "$(id -u)" = 0 || SUDO="sudo -E"
+if is_current_user_root && is_current_user_sudoer; then SUDO="sudo -E"; fi
```
benoitf:
```
is_current_user_root && is_current_user_sudoer || SUDO="sudo -E"
```

 ?

agents/che-core-api-agent/src/test/resources/run_launcher_bats_tests.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
```
benoitf, line 1: is there a way to launch tests as part of the maven build ?

l0rd: any suggestion?

agents/che-core-api-agent/src/test/resources/run_launcher_bats_tests.sh
```diff
@@ -0,0 +1,43 @@
...
+#
+# Contributors:
+#   Mario
+#
+# How to run this script:
+#   cd <root of che local git repository>
+#   sh agents/che-core-api-agent/src/test/resources/run_launcher_bats_tests.sh
+#
+
+images=(bitnami/che-codeigniter:3.1.3-r6 bitnami/che-express:4.15.3-r2 bitnami/che-java-play:1.3.12-r3 bitnami/che-laravel:5.4.23-r1 bitnami/che-rails:5.1.2-r0 bitnami/che-swift:3.1.1-r0 bitnami/che-symfony:3.3.2-r0 eclipse/centos_jdk8 eclipse/centos_jdk8 eclipse/cpp_gcc eclipse/debian_jdk8 eclipse/debian_jre eclipse/dotnet_core eclipse/hadoop-dev eclipse/kotlin eclipse/node eclipse/php eclipse/php:5.6 eclipse/php:gae eclipse/selenium eclipse/ubuntu_android eclipse/ubuntu_go eclipse/ubuntu_jdk8 eclipse/ubuntu_jre eclipse/ubuntu_python:2.7 eclipse/ubuntu_python:gae_python2.7 eclipse/ubuntu_python:latest eclipse/ubuntu_rails kaloyanraev/che-zendserver registry.centos.org/che-stacks/centos-go registry.centos.org/che-stacks/centos-nodejs registry.centos.org/che-stacks/spring-boot registry.centos.org/che-stacks/vertx registry.centos.org/che-stacks/wildfly-swarm tomitribe/ubuntu_tomee_173_jdk8 registry.centos.org/che-stacks/centos-git)
```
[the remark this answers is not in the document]
l0rd: Probably parsing stacks.json. That may be done in another PR.

agents/exec/src/main/resources/org.eclipse.che.exec.script.sh
l0rd: yes that's better. let me change it

agents/exec/src/main/resources/org.eclipse.che.exec.script.sh
l0rd: yes let me fix that

agents/che-core-api-agent/src/test/resources/run_launcher_bats_tests.sh
benoitf: it could be run as part of the profile integration tests and using docker containers 
https://github.com/eclipse/che/blob/185273f8b92fe5523e676255cd7be172e90ee418/plugins/plugin-java-debugger/che-plugin-java-debugger-server/pom.xml#L234-L235

agents/exec/src/main/resources/org.eclipse.che.exec.script.sh
l0rd: Ok your version is more consistent with the rest of the script. Will change that

agents/che-core-api-agent/src/test/resources/run_launcher_bats_tests.sh
l0rd: That would be cool but I can't see how to do `docker exec` using the docker-maven-plugin. It doesn't look so trivial anyway. Do you mind if we address that in a separate PR?

agents/exec/src/main/resources/org.eclipse.che.exec.script.sh
l0rd: fixed

agents/exec/src/main/resources/org.eclipse.che.exec.script.sh
l0rd: fixed

agents/exec/src/main/resources/org.eclipse.che.exec.script.sh
l0rd: fixed

agents/che-core-api-agent/src/test/resources/run_launcher_bats_tests.sh
tolusha: As a general tendency, if something is postponed then it won't be done at all :(
`docker-maven-plugin` allows starting container in privileged mode and then we can execute tests script inside it.
May be I am mistaken but could you clarify what issue you faced here?

benoitf: separate PR is fine for me as this PR should unblock the usage of a specific branch of Che for OpenShift.

l0rd: > As a general tendency, if something is postponed then it won't be done at all :(

I have to admit that you are right @tolusha :-) I was hoping in some softness since that's an improvement of an improvement I added to the codebase (the bats files). But man you don't seem to be much indulgent today so let's try to fix that ;-) 
Let me try to address that today. If takes too long I will ask you again to be indulgent hoping to have more luck the second time.
 
> May be I am mistaken but could you clarify what issue you faced here?

In the [fabric8 docker-maven-plugin](https://github.com/fabric8io/docker-maven-plugin) I've seen a `docker:start` goal but not a `docker:exec`. And in the tests I've written we first start a container (`docker run`) and then executes the tests inside the container (`docker exec`). Maybe `docker:watch` may help but not sure.

l0rd: Thanks @benoitf. At least someone has been indulgent today :-) However I will give it a try and tell you how it goes

agents/exec/src/main/resources/org.eclipse.che.exec.script.sh
```diff
@@ -23,7 +32,7 @@ if [ ${CURL_INSTALLED} = false ] && [ ${WGET_INSTALLED} = false ]; then
   CURL_INSTALLED=true
 fi
 
-test "$(id -u)" = 0 || SUDO="sudo -E"
+is_current_user_root && is_current_user_sudoer || SUDO="sudo -E"
```
garagatyi: Can you elaborate on what will happen if a user is root but sudo command is not installed in a container?

l0rd: @garagatyi if current user is root this code is not setting SUDO variable. Your scenario should work just fine. I've just tested with:

```bash
unset SUDO
is_current_user_root() {
    return 0
}

is_current_user_sudoer() {
    return 0
}

is_current_user_root && is_current_user_sudoer || SUDO="sudo -E"
echo SUDO=${SUDO}

# output sould be: 
# SUDO=
```

But since it doesn't look easy to understand to me neither I've extracted it to a function and made it more readable and covered by the tests (and found a bug in the process):

```bash
set_sudo_command() {
    if is_current_user_sudoer && ! is_current_user_root; then SUDO="sudo -E"; else unset SUDO; fi
}
```

tolusha (approved):

agents/exec/src/main/resources/org.eclipse.che.exec.script.sh
garagatyi: Yes, it is what I had expected. In your example return value of function `is_current_user_sudoer` should return 1 since sudo is not installed. Then 
```
is_current_user_root && is_current_user_sudoer || SUDO="sudo -E"
echo SUDO=${SUDO}
```
echos SUDO=sudo -E

Now the logic of `set_sudo_command` function looks good.

garagatyi (approved):

PR #5839, opened 2017-07-28
CHE-5684: implement URLRewriter for MacOS

### LINKED ISSUE #5684: Implement URLRewriter for Docker infrastructure
Related to SPI branch only.
We replaced the concept of ServerEvaluationStrategy with URLRewriter in spi branch. Now URLRewriter impl doesn't change URLs at all. We need to create impl that can cover Docker infra needs, such as local Linux, remote Linux,  Windows, Mac environments.

garagatyi:
### What does this PR do?
Implement and bind as default impl URLRewriter which makes Docker
infra work on MacOS or Windows. Property 'che.docker.ip.external'
should be used to set the correct external hostname for containers.
Refactored URLRewriter interface to use strings as URLs instead of
URL class since it can't handle non-basic protocols, e.g. ws.

Fixes #5684

### DIFF
--- a/infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/URLRewriterImpl.java
+++ b/infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/URLRewriterImpl.java
@@ -1,3 +1,13 @@
+/*******************************************************************************
+ * Copyright (c) 2012-2017 Codenvy, S.A.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
 package org.eclipse.che.workspace.infrastructure.docker;
 
 import org.eclipse.che.api.core.model.workspace.runtime.RuntimeIdentity;
--- a/infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/URLRewriterImpl.java
+++ b/infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/URLRewriterImpl.java
@@ -1,0 +1,63 @@
+package org.eclipse.che.workspace.infrastructure.docker;
+
+import org.eclipse.che.api.core.model.workspace.runtime.RuntimeIdentity;
+import org.eclipse.che.api.workspace.server.URLRewriter;
+import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
+import org.eclipse.che.api.workspace.server.spi.InternalInfrastructureException;
+import org.eclipse.che.commons.annotation.Nullable;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.ws.rs.core.UriBuilder;
+import javax.ws.rs.core.UriBuilderException;
+import java.net.URI;
+
+import static java.lang.String.format;
+
+/**
+ * {@link URLRewriter} needed for {@link DockerRuntimeInfrastructure} to handle specific cases of
+ * running containers.
+ * </br> For example, containers in MacOS, Windows, etc.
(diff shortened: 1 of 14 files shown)


riuvshin (approved):

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/URLRewriterImpl.java
```diff
@@ -0,0 +1,63 @@
...
+import static java.lang.String.format;
+
+/**
+ * {@link URLRewriter} needed for {@link DockerRuntimeInfrastructure} to handle specific cases of
+ * running containers.
+ * </br> For example, containers in MacOS, Windows, etc.
+ *
+ * @author Alexander Garagatyi
+ */
+public class URLRewriterImpl implements URLRewriter {
```
voievodin: Maybe change the name to something more certain like `ExternalIpURLRewriter` ?

garagatyi: ok

skabashnyuk (approved):

benoitf:
I don't see how we can plug additional url rewriters like when we're using a reverse proxy with that new way. 

Also this URL rewriter is only handling che.docker.ip.external while we had possibility to select by property the custom strategy pattern in Che 5.

How it is handle in this branch ? as I mainly see it as a regression.

garagatyi:
To bind another Url Rewriter you need to replace url rewriter class in constructor with interface of url rewriter and bind it. 

This PR doesn't handle strategies implementation. If it is needed it should be implemented separately.

benoitf:
@garagatyi yes but this is still not configurable as it's hardcoded in the class.
What I see is that each time a new design is bring, each time you drop all features that I added before. This is why  I want to warn it into this PR as it's again the case.
AFAIK SPI should bring new design but keep existing features.

gazarenkov (commented):
@garagatyi can we use uri intead of url (you replaced with string) ?

@benoitf che6 is not backward compatible

benoitf:
@gazarenkov I know it's not backward compatible but architecture design should still take into account previous features which are useful for end users.

garagatyi:
> yes but this is still not configurable as it's hardcoded in the class

Since there are no other strategies it doesn't make sense to implement strategy selection. 

> can we use uri intead of url (you replaced with string) ?

@gazarenkov  we can, but here is the [only place where it is used](https://github.com/eclipse/che/blob/spi/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/InternalRuntime.java#L201) and it is converted into String right there. So even if we change it to URI this fact won't change anything at all.

sleshchenko (approved):

akorneta (approved):

voievodin (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/InternalRuntime.java
```diff
@@ -197,22 +197,11 @@ public final T getContext() {
...
-                        return 101;
-                    }
-
-                    @Override
-                    public String getMessage() {
-                        return "Malformed URL for " + name + " : " + e.getMessage();
-                    }
-                });
+            } catch (InfrastructureException e) {
+                warnings.add(new WarningImpl(101, "Malformed URL for " + name + " : " + e.getMessage()));
```
voievodin, line 204: What is _101_?
Have you considered implementing warning for this certain case e.g.
```java
class MalformedURLWarning implements Warning { 
  static int CODE = 101;
  //...
  int getCode() { return CODE; }
  //...
}
```

garagatyi: it is initial value  - I didn't change it. I suppose we need to think about spec on it and then do something like class you have suggested

gazarenkov (approved):
OK, 
but let's consider (discuss) replace String with URI in all related classes

Issue #5127, opened 2017-05-18
Create docker build profile


voievodin:
Che project uses many technologies and programming languages and requires developer to have all of them installed if he/she wants to build the project. Moreover the versions of all the installed tools must match the version supported by the project.

These technologies include:
- java
- maven
- nodejs
- gulp
- bower 
- golang

Some of che subprojects such as [exec-agent](https://github.com/eclipse/che/tree/master/agents/go-agents) might depend on linux specific stuff like sys calls - which causes problems like [this one](https://github.com/eclipse/che/issues/4758). Obviously sometimes we can workaround such problems by skipping build of some of the artifacts, but it doesn't work always.

The solution i see is creation of a __docker__ build profile which brings the following benefits: 

1. Reduced amount of tools needed to build the project.
Well, we still need _java_,  _maven_ and _docker_ but that's all. 

2. Tool version switch made easier  e.g. if we want to switch golang version from 1.6 to 1.8 we don't need to force user to do that.

3. Better stability, less bugs.
Lets say we require _build-tool-x v11+_, we're using _build-tool-x v11_ for building the product and on our CI and everything works fine but it turns out that _build-tool-x v12_ introduces a bug and developers who have it installed might have build issues.

The *docker* profile could help us to cover these topics.

@TylerJewell, @benoitf, @skabashnyuk, @riuvshin  what do you think about it?

benoitf:
Building with docker is really good, we've start to do it for dockerfiles directory
Also, even if it may include an overhead in performance, it's good to have a universal build that work on all platforms where docker is installed. 

BTW @evoevodin, do you see docker profile being the only way or it's an alternate way when required tools are not setup (or user doesn't want to) of native builds

voievodin:
i see it as alternate way to build the project
```
mvn clean install          <- uses native tools
mvn clean install -Pdocker <- uses docker even if there are native tools installed
```

benoitf:
@evoevodin ok this is how I was understanding it but I wanted to be sure.

skabashnyuk:
won't fix

PR #5848, opened 2017-07-31
Use docker to build dashboard app. (as alternative)

benoitf:
### What does this PR do?
Use docker to build dashboard app. (as alternative)
By using layer for package.json and bower.json file, it saves extra time to get npm or bower dependencies

Also, if dashboard files are not modified, the build is faster.

it targets as well issue #5127

By default :
 "$ mvn clean install" --> use docker
 "$ mvn clean install -Pnative" --> use native tools

as a side note, if all files are untouched, mvn clean  install of dashboard is taking less than 10 seconds instead of like 5mn+

 #5127

#### Changelog
Allow to build dashboard app with docker.

#### Release Notes
N/A

#### Docs PR
N/A

Change-Id: Ia65161b58e8ae70f7799500750a3d87687e34819

### DIFF
--- a/dashboard/.dockerignore
+++ b/dashboard/.dockerignore
@@ -1,0 +1,27 @@
+Dockerfile
+node_modules
+bower_components
+target
+pom.xml
+./typings
+./tmp
+./codeverage
+
+# Eclipse  #
+###################
+
+*.launch
+.classpath
+.project
+.settings/
+bin/
+test-output/
+maven-eclipse.xml
+
+# Idea #
+##################
+*.iml
+*.ipr
+*.iws
+.idea/
+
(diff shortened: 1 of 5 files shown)


olexii4 (approved):

ashumilova (approved):

ashumilova:
@benoitf can you please mention the profile option here as well : https://github.com/eclipse/che/blob/master/dashboard/README.md

akurinnoy (approved):

benoitf:
@ashumilova done. is it ok now ?

ashumilova (approved):

riuvshin:
As I understand we wanted use native tools by default and profile for docker. But I'm +1 on this

benoitf:
@riuvshin yes but after thinking a log around this I was thinking that we have more ppl having docker than all the other required tools. (this is why I used docker by default)
Also it should be a lot faster using docker than native (as we reuse docker layers)

Issue #5786, opened 2017-07-24
Customizing site pages related to authentication and Keycloak


skabashnyuk:
In Codenvy we had a couple of customized pages 
1. Login
2. Create account
3. Recover password
4. Reset password

We need to find out if we need to have it customized for Che with Keycloak

mmorhun:
Created draft of login pages for Keycloak Che theme:
Login page: 
![login](https://user-images.githubusercontent.com/15607393/28676756-2b1385be-72f4-11e7-9d9c-71e93d2689f7.png)
Login page after wrong password : 
![wrongpassword](https://user-images.githubusercontent.com/15607393/28865126-7e60a5b2-7777-11e7-9076-1af1c100a687.png)
Register page:
![register](https://user-images.githubusercontent.com/15607393/28676835-6e265700-72f4-11e7-99e7-0930f6c3a580.png)
Reset password page:
![reset-password](https://user-images.githubusercontent.com/15607393/28676848-7955db3c-72f4-11e7-905c-13684749bf55.png)
Login page after reset password:
![afterresetpassword](https://user-images.githubusercontent.com/15607393/28865075-4854ad7e-7777-11e7-9e03-3ce43d79ea8c.png)
Keycloak error:
![kc-error](https://user-images.githubusercontent.com/15607393/28676953-cc510a46-72f4-11e7-8b88-b0e5797ec21c.png)

mmorhun:
@slemeur could you please review it?

mmorhun:
Done

PR #5859, opened 2017-08-01
Che theme for Keycloak

mmorhun:
### What does this PR do?
Adds draft of Che theme for Keycloak login pages.

https://github.com/eclipse/che/issues/5786

#### Changelog
N/A Hasn't released yet.

#### Release Notes
N/A


riuvshin (approved):

riuvshin:
@mmorhun please check ci build fail report it failed due to 
"Some files do not have the expected license header"

mmorhun:
This error comes from master and has been fixed already.

skabashnyuk (approved):

dockerfiles/init/modules/keycloak/files/che/login/messages/messages_en.properties
```diff
@@ -0,0 +1,240 @@
+product=Eclipse Che
```
benoitf, line 1: how to we keep in sync this file with new keycloack releases ?
what are the changes with upstream project ?

should it be better to apply patches ? so we can track changes easily

dockerfiles/init/modules/keycloak/files/che/login/register.ftl
```diff
@@ -0,0 +1,76 @@
+<#import "template.ftl" as layout>
```
benoitf, line 1: maybe describe the changes with https://github.com/keycloak/keycloak/blob/master/themes/src/main/resources/theme/base/login/register.ftl to better update with new upstream releases.

mmorhun: I am not sure, whether I understood you correctly, but if you meant to roll our changes over base templates - I do not think that it is a good idea here. For example, they could redesign some pages, do we want to apply that? I think that we don't.

dockerfiles/init/modules/keycloak/files/che/login/messages/messages_en.properties
mmorhun: Do we need to replace our current messages with messages from new upstream releases?
I think we have our own Che theme and we shouldn't depend on changes in upstream themes.

mshaposhnik (approved):

dockerfiles/init/modules/keycloak/files/che/login/messages/messages_en.properties
benoitf: I'm not asking to replace "our" messages with "upstream" messages.
but if you upgrade keycloak later, you may have new constants to update/change.
And how do we know what was the "customized" strings and the "upstream" strings.

Maybe that you've only changed product name and that's all but here we don't know what were the full changes.
By knowing the changes, then, it's easier to upgrade keycloak

dockerfiles/init/modules/keycloak/files/che/login/register.ftl
benoitf: @mmorhun how do you manage the upgrade of keycloak then ?
Let say they fix an issue in the base template. How the fix is then applied on our template ? it is not.

skabashnyuk: @benoitf I think we don't know. If it will be an issue we will figure out.

dockerfiles/init/modules/keycloak/files/che/login/messages/messages_en.properties
skabashnyuk: I think at this stage we don't need to dig that far. If and when it will be needed when we will work on that.

dockerfiles/init/modules/keycloak/files/che/login/register.ftl
benoitf: I prefer to deal with it now or the plan around it.

dockerfiles/init/modules/keycloak/files/che/login/messages/messages_en.properties
benoitf: It will be needed for each update of the keycloak component. so the need will come sooner than never.

benoitf (changes requested):
I don't see how we clearly manage the updates of the theme at each update of keycloak

riuvshin:
@benoitf current custom there is under puppet control, we generate it same way as from templates. If we will need to change theme we have to change source files here `dockerfiles/init/modules/keycloak/files` and puppet will regenerate it all and new updated theme will be mounted to keycloack image here https://github.com/eclipse/che/pull/5859/files#diff-d97c076dd428b95a4021cdc15fdebfcbR139
So if we decide to update just a theme OR keycloack version which may cause theme update, all we need to do is update theme sources under puppet and then it will be regenerated and remounted.

But yes it still require some manual kind of merge with upstream template in case of update, so I agree that patching is probably better but it may fail same way if upstream is changed heavily.
and we can't patch theme inside of a container with keycloack we can only mount something else to replace default one.

benoitf:
@riuvshin I'm not talking about how puppet is working.

But how we will ensure that bugfixes and new stuff (like new constant messages) in upstream(keycloak), is backported when updating keycloak version here
https://github.com/eclipse/che/blob/f052d14bac5a470a2e32996d5c042cbb754875c4/dockerfiles/cli/images.template#L7

AFAIK it's error prone as ppl will only bump the version in the images.template (not checking what are the new contants, if there are some fixes in templates, etc)

riuvshin:
@benoitf ah :) well bumping version of key component is not something that happens like that. you can't just change version and commit it without tests of everything related to this. It is exactly same as we update swarm for example, it is one line change https://github.com/codenvy/codenvy/pull/2335/files with new version but to merge this we did huge amount of testing. 
It works same way if you will upgrade some lib version in code. you do upgrade and check is everything is working same way and if not you go and fix it because it is part of upgrade task to make sure that there is no regressions.

mmorhun:
@benoitf addressing you comments:
- If we'll upgrade Keycloak later it could use new constants, but we will use our own.
- We don't have to know what was the "customized" strings and the "upstream" strings. We just use our own from our own theme.
- We'll just update Keycloak. That's it. If they make significant changes in viewing area then describing changes (as you suggested) won't help.
- > Let say they fix an issue in the base template. How the fix is then applied on our template ?
  
  What kind of issue in template do you mean?

benoitf:
@riuvshin again, I'm talking about reporting changes from upstream and I was refering two usecases not covered by your scenario.
my usecases were:
1 - update of string constants.
   You can check that everything is working as you're telling. But let say there is an unexpected error at a time and that then it doesn't find the message as it is not in the constants file. I really doubt that "checking that all is working" will check all errors.

2 - backport of upstream fixes.
 again, you can check that all is working but if a fix was done upstream it will not be applied on your fork automatically.

riuvshin:
@benoitf I see your points but, I don't see how we can do this in other way if no testing...
how we manage situations like that in other components in code?
I mean don't it should be a part of task for upgrade keycloack version to make sure that we have exactly what we expect? we can't just bump version and expect that everything will be just fine.

mmorhun:
@benoitf again
1. > But let say there is an unexpected error at a time and that then it doesn't find the message as it is not in the constants file.

This situation is impossible because we use our own template with our own constants.

2. We change only theme, not engine! If they fix engine we'll get the fix. But what we have not touched is the theme html templates and constants.

benoitf:
> @benoitf again
> 
> But let say there is an unexpected error at a time and that then it doesn't find the message as it is not in the constants file.
> This situation is impossible because we use our ouwn template with our own constants.
> 

here
https://github.com/eclipse/che/blob/71e597727f056eceed8144e390e39625f42cc11e/dockerfiles/init/modules/keycloak/files/che/login/messages/messages_en.properties

are you really using all constants of this file ? AFAIK you've only added product constant on top of forking /opt/jboss/keycloak/themes/base/login/messages/messages_en.properties of the docker image.

shouldn't the file define only specific che theme "product" constant instead of duplication a whole file ?
as "che" theme is based on "base" theme, all other messages should be inherited ?


> We change only theme, not engine! If they fix engine we'll get the fix. But what we have not touched is the theme html templates and constants.

You're forking a constant properties file so fix could be on constant value or introducing new values.

mmorhun:
@benoitf 
> are you really using all constants of this file ? 
> shouldn't the file define only "product" constant ?

No, I use several constants.

Yes, I forked constant properties, what is the problem?
> fix could be on constant value

Again. Then it will change messages in our theme. Do we need that?
> could introducing new values

This won't affect us at all. We use our own templates and constants.

If you want a solution for all cases here, I am afraid, that it is just impossible.
From Keycloak docs (https://keycloak.gitbooks.io/documentation/server_development/topics/themes.html): 
> When extending a theme you can override individual resources (templates, stylesheets, etc.). If you decide to override HTML templates bear in mind that you may need to update your custom template when upgrading to a new release.

benoitf:
> > @benoitf
> > 
> > are you really using all constants of this file ?
> > shouldn't the file define only "product" constant ?
> > No, I use several constants.
> > 
> 
> Yes, I forked constant properties, what is the problem?
> 

It's the problem that I'm referring since my first comment.
We don't know what you've added, what you've changed in the forked file and how to manage upgrade.

mmorhun:
@benoitf I removed forked properties and left only new/redefined. But this could cause changes of  messages on Keycloak updates!

benoitf (approved):

PR #5866, opened 2017-08-01
Fix launchers to call to set_sudo_command

l0rd:
### What does this PR do?
Fix a bug introduced with https://github.com/eclipse/che/commit/a419690730908aac6ddf573e935722179615ddff: function `set_sudo_command` was never called


ghost (approved):

l0rd:
I've rebuild everything and double check that with this patch a workspace could be created under both Docker and OpenShift

tolusha (approved):

PR #5869, opened 2017-08-02
Avoid NPE at using MutableProjectConfig

RomanNikitenko:
### What does this PR do?
We have constructor for MutableProjectConfig without parameters [here](https://github.com/eclipse/che/blob/a652db2c46aea34dc4b2e0e3c554288da4a2e1c0/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/project/MutableProjectConfig.java#L53-L53)
and the fields are not initialized in this case.
These changes allow to avoid NPE in different places on client side where we use MutableProjectConfig

#### Changelog
Avoid NPE at using MutableProjectConfig

### DIFF
--- a/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/project/MutableProjectConfig.java
+++ b/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/project/MutableProjectConfig.java
@@ -20,6 +20,7 @@ import org.eclipse.che.api.machine.shared.dto.CommandDto;
 import java.util.List;
 import java.util.Map;
 
+import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.collect.Lists.newArrayList;
 import static com.google.common.collect.Maps.newHashMap;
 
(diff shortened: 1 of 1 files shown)


ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/project/MutableProjectConfig.java
```diff
@@ -55,7 +55,7 @@ public MutableProjectConfig() {
 
     @Override
     public String getName() {
-        return name;
+        return name != null ? name : "";
```
azatsarynnyy: @RomanNikitenko just FYI - `com.google.common.base.Strings#nullToEmpty`

RomanNikitenko: @azatsarynnyy thank you, I've fixed it!

RomanNikitenko:
@azatsarynnyy @vparfonov I've changed checks for null: com.google.common.base.MoreObjects#firstNonNull instead of com.google.common.base.Strings#nullToEmpty because of in this case this one allows to avoid NPE for collections as well as for string or any object.

azatsarynnyy (approved):

vparfonov (approved):

Issue #2571, opened 2016-09-23
Modifiable agent registry


tolusha:
Every codenvy installation and locally launched che must have agent registry. Currently there is a very basic implementation of the [AgentRegistry](https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/impl/LocalAgentRegistryImpl.java) with several limitations:
- No storage
- Impossible to add/remove/update agents
- Relying on [launcher](https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/AgentLauncher.java)

We need to have a new implementation without limitation mentioned above. 
Consider the case when new agents requires binaries, where they will be located/stored?
Take into account users` permissions.

TylerJewell:
Requirements are clear.  I am not sure on priority for implementation - but we will probably consider it after first release of team capabilities.  It will be interesting to see how we get C/C++ language server.

PR #5876, opened 2017-08-02
CHE-2571: Modifiable agent registry

tolusha:
### What does this PR do?
Introduces persistent layer to store descriptors and scripts of the installers.

https://github.com/eclipse/che/issues/2571

### DIFF
--- a/wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/InstallerRegistry.java
+++ b/wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/InstallerRegistry.java
@@ -10,14 +10,14 @@
  *******************************************************************************/
 package org.eclipse.che.api.installer.server;
 
+import org.eclipse.che.api.core.Page;
 import org.eclipse.che.api.installer.server.exception.IllegalInstallerKeyException;
-import org.eclipse.che.api.installer.server.exception.InstallerConflictException;
+import org.eclipse.che.api.installer.server.exception.InstallerAlreadyExistException;
 import org.eclipse.che.api.installer.server.exception.InstallerException;
 import org.eclipse.che.api.installer.server.exception.InstallerNotFoundException;
 import org.eclipse.che.api.installer.server.impl.InstallerFqn;
 import org.eclipse.che.api.installer.shared.model.Installer;
 
-import java.util.Collection;
 import java.util.List;
 
 /**
@@ -34,7 +34,7 @@ public interface InstallerRegistry {
      *
      * @param installer
      *      the installer to add
-     * @throws InstallerConflictException
+     * @throws InstallerAlreadyExistException
      *         if installer with corresponding {@link InstallerFqn} already exists
      * @throws InstallerException
      *         if unexpected error occurred
@@ -56,13 +56,14 @@ public interface InstallerRegistry {
     /**
      * Removes installer from the registry.
      *
-     * @param fqn
-     *      the installer fully-qualified key
-     *      {@link InstallerFqn}
+     * @param installerKey
+     *         the installer key
+     * @throws IllegalInstallerKeyException
+     *         if specified installer key has wrong format
      * @throws InstallerException
      *         if unexpected error occurred
      */
-    void remove(InstallerFqn fqn) throws InstallerException;
+    void remove(String installerKey) throws InstallerException;
 
     /**
      * Gets {@link Installer} by key.
@@ -85,8 +86,6 @@ public interface InstallerRegistry {
      * @param id
      *         the id of the installer
      * @return list of versions
-     * @throws InstallerNotFoundException
-     *         if installer not found in the registry
      * @throws InstallerException
      *         if unexpected error occurred
      */
@@ -94,13 +93,19 @@ public interface InstallerRegistry {
 
 
     /**
-     * Returns the collection of available installers.
+     * Returns all installers using pagination.
      *
-     * @return collection of installers
+     * @param maxItems
+     *         the maximum number of installers to return
+     * @param skipCount
+     *         the number of installers to skip
+     * @return list of installers or empty list if no installers were found
+     * @throws IllegalArgumentException
+     *         when {@code maxItems} or {@code skipCount} is negative
      * @throws InstallerException
      *         if unexpected error occurred
      */
-    Collection<Installer> getInstallers() throws InstallerException;
+    Page<? extends Installer> getInstallers(int maxItems, int skipCount) throws InstallerException;
 
     /**
      * Traverses dependencies of all listed installers and
--- a/wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/InstallerRegistry.java
+++ b/wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/InstallerRegistry.java
@@ -12,7 +12,7 @@ package org.eclipse.che.api.installer.server;
 
 import org.eclipse.che.api.core.Page;
 import org.eclipse.che.api.installer.server.exception.IllegalInstallerKeyException;
-import org.eclipse.che.api.installer.server.exception.InstallerAlreadyExistException;
+import org.eclipse.che.api.installer.server.exception.InstallerAlreadyExistsException;
 import org.eclipse.che.api.installer.server.exception.InstallerException;
 import org.eclipse.che.api.installer.server.exception.InstallerNotFoundException;
 import org.eclipse.che.api.installer.server.impl.InstallerFqn;
@@ -34,7 +34,7 @@ public interface InstallerRegistry {
      *
      * @param installer
      *      the installer to add
-     * @throws InstallerAlreadyExistException
+     * @throws InstallerAlreadyExistsException
      *         if installer with corresponding {@link InstallerFqn} already exists
      * @throws InstallerException
      *         if unexpected error occurred
--- a/wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/InstallerRegistry.java
+++ b/wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/InstallerRegistry.java
@@ -10,9 +10,11 @@
  *******************************************************************************/
 package org.eclipse.che.api.installer.server;
 
-import org.eclipse.che.api.installer.server.exception.IllegalInstallerKey;
+import org.eclipse.che.api.installer.server.exception.IllegalInstallerKeyException;
+import org.eclipse.che.api.installer.server.exception.InstallerConflictException;
 import org.eclipse.che.api.installer.server.exception.InstallerException;
 import org.eclipse.che.api.installer.server.exception.InstallerNotFoundException;
+import org.eclipse.che.api.installer.server.impl.InstallerFqn;
 import org.eclipse.che.api.installer.shared.model.Installer;
 
 import java.util.Collection;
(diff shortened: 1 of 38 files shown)


wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/exception/InstallerExceptionMapper.java
```diff
@@ -0,0 +1,59 @@
...
+import org.eclipse.che.dto.server.DtoFactory;
+
+import javax.inject.Singleton;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.ext.Provider;
+
+/**
+ * @author andrew00x
```
garagatyi: Looks like these authors haven't implemented this class ;)

wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/exception/InstallerExceptionMapper.java
```diff
@@ -0,0 +1,59 @@
...
+                           .build();
+
+        } else if (exception instanceof InstallerConflictException) {
+            return Response.status(Response.Status.CONFLICT)
+                           .entity(errorAsJson)
+                           .type(MediaType.APPLICATION_JSON)
+                           .build();
+
+        } else if (exception instanceof IllegalInstallerKeyException) {
+            return Response.status(Response.Status.BAD_REQUEST)
```
garagatyi, line 46: I think 403 forbidden should be used instead as server understood request and entity is valid but value in entity prevents fulfilment of the request

wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/spi/InstallerDao.java
```diff
@@ -0,0 +1,136 @@
...
+     * @param maxItems
+     *         the maximum number of installers to return
+     * @param skipCount
+     *         the number of installers to skip
+     * @return list of installers or empty list if no installers were found
+     *         when {@code maxItems} or {@code skipCount} is negative
+     * @throws InstallerException
+     *         when any other error occurs
+     */
+    Page<InstallerImpl> getVersions(String id, int maxItems, long skipCount) throws InstallerException;
```
voievodin: If there is `getByFqn` maybe `getById` makes more sense?

voievodin (approved):
Good man writes good code :+1:

wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/InstallerRegistry.java
```diff
@@ -28,12 +30,47 @@
...
+    /**
+     * Removes installer from the registry.
+     *
+     * @param fqn
+     *      the installer fully-qualified key
+     *      {@link InstallerFqn}
+     * @throws InstallerException
+     *         if unexpected error occurred
+     */
+    void remove(InstallerFqn fqn) throws InstallerException;
```
sleshchenko: I think you should not use `InstallerFqn` object here, and use string key value like `getOrderedInstallers` or `getInstaller` methods do. Or you should rework all methods to use InstallerFqn. I prefer to use string value.

wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/exception/InstallerExceptionMapper.java
sleshchenko: Usually, we use 400 as validation exception. So I wouldn't introduce such approach in this PR because it can be strange that only this service throw 403 on validation exception while another throw 400.
If throwing of 400 in such case is a problem it should be reworked whole all services.

wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/jpa/JpaInstallerDao.java
```diff
@@ -0,0 +1,176 @@
...
+/**
+ * @author Anatolii Bazko
+ */
+@Singleton
+public class JpaInstallerDao implements InstallerDao {
+
+    private final Provider<EntityManager> managerProvider;
+
+    @Inject
+    public JpaInstallerDao(@SuppressWarnings("unused") DBInitializer dbInitializer,
```
sleshchenko, line 46: Can you explain why do you need `dbInitializer` here?

wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/jpa/JpaInstallerDao.java
```diff
@@ -0,0 +1,176 @@
...
+            }
+            return new InstallerImpl(installer);
+        } catch (RuntimeException e) {
+            throw new InstallerException(e.getMessage(), e);
+        }
+    }
+
+    @Override
+    @Transactional
+    public Page<InstallerImpl> getVersions(String id, int maxItems, long skipCount) throws InstallerException {
```
sleshchenko: Manager and Service have the method with the same but do something different. Manager and Service return all versions of installer with id, while dao return installer with all versions with id. It can be confused.
Should we really implement it in such way or we can implement it in the same way?

wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/model/impl/InstallerServerConfigImpl.java
```diff
@@ -0,0 +1,144 @@
...
+@Table(name = "installer_servers")
+public class InstallerServerConfigImpl implements ServerConfig {
+
+    @Id
+    @GeneratedValue
+    @Column(name = "id")
+    private Long id;
+
+    @Column(name = "installer_id", nullable = false)
+    private String installerId;
```
sleshchenko: Are `installerId` and `installerVersion` usefull fields here or they are used only for mapping?
If only for mapping then maybe we can map `InstallerServerConfigImpl` and `InstallerImpl` in other way like map InstallerServerConfigImpl to InstalleImpl by InstallerServerConfigImpl#id field, then looks like `installerId` and `installerVersion` are redundant and you'll be able to create copy constructor for InstallerServerConfigImpl(ServerConfig). WDYT?

wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/spi/InstallerDao.java
tolusha: `getById` name is supposed to return the only installer as opposite to `getVersions` name.

wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/jpa/JpaInstallerDao.java
tolusha: Otherwise, Dao starts working before DB is initialized.

sleshchenko: Have you investigated why other Daos like JpaStackDao, JpaWorkspace, JpaAccountDao, ... ?

wsmaster/che-core-sql-schema/src/main/resources/che-schema/5.16.0/1__add_installer.sql
```diff
@@ -0,0 +1,59 @@
...
+-- which accompanies this distribution, and is available at
+-- http://www.eclipse.org/legal/epl-v10.html
+--
+-- Contributors:
+--   Codenvy, S.A. - initial API and implementation
+--
+
+-- Installer ---------------------------------------------------------------------
+CREATE TABLE installer (
+    id                  VARCHAR(255)         NOT NULL,
```
skabashnyuk: Can you make id is auto generated id, and installer_id+version as an alternative unique key. It will allow in nested tables to depend on single field id.

wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/jpa/JpaInstallerDao.java
sleshchenko: I see. It needs `DbInitializer` because `LocalInstallerRegistry` use this dao in constructor

wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/InstallerRegistry.java
```diff
@@ -28,12 +30,47 @@
 public interface InstallerRegistry {
 
     /**
+     * Adds installer to the registry.
+     *
+     * @param installer
+     *      the installer to add
+     * @throws InstallerConflictException
```
sleshchenko: WDYT about making this exception more concrete. Now it is an analog of rest ConfictException where conflict is related to `409` response code. Mb it would be better to use here exception like `InstallerAlreadyExistException`?

tolusha: sure

wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/jpa/JpaInstallerDao.java
tolusha: Ok

tolusha:
@sleshchenko @skabashnyuk

wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/exception/InstallerAlreadyExistException.java
```diff
@@ -0,0 +1,26 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Codenvy, S.A. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.che.api.installer.server.exception;
+
+/**
+ * Installer conflict error.
```
sleshchenko: Fix doc

wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/InstallerRegistryService.java
```diff
@@ -66,76 +70,85 @@ public InstallerRegistryService(InstallerRegistry installerRegistry) {
...
-                                    .stream()
-                                    .map(DtoConverter::asDto)
-                                    .collect(Collectors.toList());
-        } catch (InstallerException e) {
-            throw new ServerException(e.getMessage(), e);
+            Page<? extends Installer> installers = installerRegistry.getInstallers(maxItems, skipCount);
+            return Response.ok()
+                           .entity(installers.getItems())
+                           .header("Link", createLinkHeader(installers))
+                           .header("TotalItemsCount", installers.getTotalItemsCount())
```
voievodin: 1. Why do you need to know the exact count of the items here?
2. Custom headers usually follow `X-` prefix naming

tolusha: To reconstruct `Page` object at `RemoteInstallerRegistry`

tolusha: AFAIK `X-` prefix is deprecated, but I agree to rename to `Total-Items-Count`

tolusha:
```
 int totalCount = -1;
 +            List<String> totalItemsCountHeader = response.getHeaders().get("TotalItemsCount");
 +
 +            if (totalItemsCountHeader != null && !totalItemsCountHeader.isEmpty()) {
 +                totalCount = Integer.valueOf(totalItemsCountHeader.get(0));
 +            }
 +            return new Page<>(response.asList(InstallerDto.class), skipCount, maxItems, totalCount);
```

voievodin: Ok, thank you.

sleshchenko (approved):

garagatyi (approved):

skabashnyuk (approved):

PR #5879, opened 2017-08-02
Create routes in openshift with the same pattern than the urls provided by naming strategy

benoitf:
### What does this PR do?

For now, routes are created by "hardcoding" some name of services like "tomcat" for 8080
Then we need to add labels in each docker image to set "tomcat" service name for exposing 8080 port.
So for example we have "vertx" dockerfile exposing a "Tomcat-" url while there is no tomcat.

This PR is generating routes by taking care of labels if they're defined without hardcoding the value.

Note: once this PR is merged, we will be able to modify Dockerfile and change "tomcat" label to something more specific (or generic) for the stack.

https://github.com/redhat-developer/rh-che/issues/92

#### Changelog
Create OpenShift routes according to the url naming strategy used.

#### Release Notes
BugFix

#### Docs PR
BugFix

Change-Id: Ie8e0a09d0be8f6d042f542a76c146e2bbe266fc7

example : vertx in the URL

image

image


amisevsk (approved):
LGTM. Great addition @benoitf!

davidfestal (approved):

ibuziuk (approved):

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -302,6 +312,83 @@ private void idleCheServer(ServerIdleEvent event) {
...
+    }
+
+    /**
+     * Gets the mapping between the port (with format 8080/tcp) and the associated label (if found)
+     *
+     * @param labels
+     *         the mapping for known port labels
+     * @param exposedPorts
+     *         the ports that are exposed
+     * @return
```
l0rd: I think you forgot to specify what's returned in the javadoc

l0rd (approved):
LGTM Thanks @benoitf

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
benoitf: fixed, thanks

Issue #5402, opened 2017-06-19
adapt docker images structure to several packagings of CHE


riuvshin:
**Needed for:** https://github.com/eclipse/che/issues/5362

- [x] add postgres docker image to che (needed for keycloak)
- [x] add keycloak docker image to che
- [x] investigate issue with keycloak startup failure when db container not started in time. (add custom entrypoint) - used docker healthcheck feature
- [x] add keycloak configuration
- [x] fix permission issues with mount for keycloak on linux 
- [x] adapt  CHE CLI for che basic and CHE advanced (with keycloak, db, etc)
- [x] adapt init image to handle configuration of different CHE assemblies
- [x] fix cli tests in che-multiuser branch https://github.com/eclipse/che/issues/6042
- [x] create needed CI automation for new images / assemblies
- [x] adapt release release autimation tools to these changes https://github.com/eclipse/che/issues/6636
- [x] clean up CHE CLI if keycloak will manage oauth and etc.

PR #5880, opened 2017-08-02
get rid of hardcoded paths in che-server container

riuvshin:
### What does this PR do?
get rid of hardcoded paths

needed for https://github.com/eclipse/che/issues/5402

### DIFF
--- a/assembly-multiuser/assembly-main/src/assembly/assembly.xml
+++ b/assembly-multiuser/assembly-main/src/assembly/assembly.xml
@@ -21,6 +21,7 @@
         <format>zip</format>
         <format>tar.gz</format>
     </formats>
+    <includeBaseDirectory>false</includeBaseDirectory>
     <dependencySets>
         <dependencySet>
             <useProjectArtifact>false</useProjectArtifact>
@@ -69,7 +70,7 @@
     </dependencySets>
     <fileSets>
         <fileSet>
-            <directory>${project.build.directory}/dependency/eclipse-che-${che.version}</directory>
+            <directory>${project.build.directory}/dependency/</directory>
             <outputDirectory />
             <excludes>
                 <exclude>**/tomcat/webapps/ROOT.war</exclude>
--- a/assembly-multiuser/assembly-main/src/assembly/assembly.xml
+++ b/assembly-multiuser/assembly-main/src/assembly/assembly.xml
@@ -21,7 +21,6 @@
         <format>zip</format>
         <format>tar.gz</format>
     </formats>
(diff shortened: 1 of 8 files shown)


benoitf:
-1 on removing the directory in the zip file. (when generating the assembly)

benoitf:
BTW I understand the problem that you're trying to fix but IHMO the proposed solution is involving too many breaking changes which I think could be avoided by a different solution

riuvshin:
@benoitf the biggest problem here is that each time in the che-server image we have folder with different names which is really painful to have. In codenvy we have have same as I did here. I thought about how can I workaround this without touching this but I didn't found any solution how can I know folder in container before I will start the container to properly mount files in right places. And I afraid that in time we could have more and more bad workarounds around that while it should be well predicted. I will try do dig more, but keeping this will introduce more problems in future...

Maybe is there a way to have folder in zip but with static name? like "eclipse-che" that could help but I don;t see how to do this, may be you @benoitf  @skabashnyuk know?

benoitf:
well, first I said -1 to strip folder in the assemblies (tgz/zip) that we distribute.

You can still strip down the folder in the docker image or make an alias if it's easier for you
For example you can add in the Dockerfile after the add command
`RUN ln -s /home/user/eclipse-che-* /home/user/eclipse-che`

then inside an eclipse che container, you can use `/home/user/eclipse-che/` and it's still backward compliant.

also AFAIK, we do know the version of che when we're generating the template so version property could be part of the template as well.

riuvshin:
@benoitf ah if you worried only about artifacts it is ok I could workaround in dockerfile, but I still think that this is bad thing to have... But I agree that we had that bad thing for a quite a while and some users may rely on that... 
I will think on this more...

benoitf:
@riuvshin for you the bad thing is to have the directory part of the tgz/zip file ?
because I would say that it's a standard in OSS world.

riuvshin:
to me bad thing is not a folder but it's name... each time you extract you have different folder, but ofc we can workaround it

benoitf:
IMHO it's for people that unpack tools in a directory, they don't override their previous binaries of a previous version. (like overriding your current install of Tomcat 8.0.35 by Tomcat 8.0.36 files)
or if you want to have two versions in parallel, you only need to unpack two archives directly, no need to make mkdir, move files, etc

riuvshin:
@benoitf I've updated PR

vparfonov (approved):

benoitf (approved):

skabashnyuk (approved):

PR #5884, opened 2017-08-03
Make OpenShiftConnector add Cluster IP instead of Service name in workspace pods

amisevsk:
### What does this PR do?
Change OpenShiftConnector to be able to run workspaces in a separate namespace

- Make OpenShiftConnector override the `CHE_API` env variable before starting a workspace so that it uses che-host's cluster IP instead of service name. This is necessary since service names are namespaced, while cluster IP is accessible anywhere on the cluster.
- *Minor fix*: Add timeout to OpenShiftPvcHelper, to avoid stalling for a long time when an error occurs (e.g. when there is a PVC issue)

This PR only makes it possible to run workspaces in a separate namespace, it does not add the ability to do so without some customization. Since Che currently only supports one user, there is no way of knowing, within OpenShiftConnector, where the workspace should be started.

Is a necessary step towards running a multitenant Che server on OpenShift. See https://github.com/redhat-developer/rh-che/issues/131.

#### Changelog
(Bug fix + no functional change)

#### Release Notes
(Bug fix; internal change only)

### DIFF
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
@@ -122,6 +122,8 @@ import org.eclipse.che.plugin.openshift.client.kubernetes.KubernetesStringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.common.annotations.VisibleForTesting;
+
 import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
@@ -441,6 +443,9 @@ public class OpenShiftConnector extends DockerConnector {
         Map<String, String> portsToRefName = getPortsToRefName(labels, exposedPorts);
 
         String[] envVariables = createContainerParams.getContainerConfig().getEnv();
+        // Overwrite CHE_API env var with che-host cluster IP instead of service name
+        setClusterIpEnvVar(envVariables);
+
         String[] volumes = createContainerParams.getContainerConfig().getHostConfig().getBinds();
 
         Map<String, String> additionalLabels = createContainerParams.getContainerConfig().getLabels();
@@ -1681,4 +1686,31 @@ public class OpenShiftConnector extends DockerConnector {
         return deploymentName;
     }
 
+    /**
+     *
+     * @param envVariables
+     * @return
+     */
+    @VisibleForTesting
+    protected void setClusterIpEnvVar(String[] envVariables) {
+        // Replace CHE_API env var with ClusterIP for che-host
+        String cheHostClusterIp = getCheHostClusterIp();
+        for (int i = 0; i < envVariables.length; i++) {
+            if (envVariables[i].contains("CHE_API")) {
+                envVariables[i] = envVariables[i].replaceAll("che-host", cheHostClusterIp);
+                return;
+            }
+        }
+        LOG.warn("Could not set CHE_API workspace environment variable to Cluster IP");
+    }
+
+    /**
+     * Get ClusterIP for the che-host service.
+     */
+    @VisibleForTesting
+    protected String getCheHostClusterIp() {
+        try (OpenShiftClient client = new DefaultOpenShiftClient()) {
+            return client.services().withName(OPENSHIFT_CHE_SERVER_SERVICE_NAME).get().getSpec().getClusterIP();
+        }
+    }
 }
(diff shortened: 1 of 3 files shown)


plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -352,6 +352,14 @@ public ContainerCreated createContainer(CreateContainerParams createContainerPar
         Set<String> exposedPorts = getExposedPorts(containerExposedPorts, imageExposedPorts);
 
         String[] envVariables = createContainerParams.getContainerConfig().getEnv();
+
+        // Replace CHE_API env var with ClusterIP for che-host
+        String cheHostClusterIp = getCheHostClusterIp();
+        for (int i = 0; i < envVariables.length; i++) {
+            if (envVariables[i].contains("CHE_API")) {
```
ibuziuk: personal preference, but I would probably extract this in a separate method (JUnit test for it would be also super useful here IMO)

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftPvcHelper.java
```diff
@@ -176,7 +177,7 @@ protected boolean createJobPod(String workspacesPvcName,
         try (OpenShiftClient openShiftClient = new DefaultOpenShiftClient()){
             openShiftClient.pods().inNamespace(projectNamespace).create(podSpec);
             boolean completed = false;
-            while(!completed) {
+            for(int waitCount = 0; (waitCount < JOB_TIMEOUT_SECONDS) && !completed; waitCount++) {
```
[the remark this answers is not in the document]
amisevsk: Yeah I took a second look at this loop and it really made no sense -- completed was never set to true...

Fixed to wait in the same was as OpenShiftConnector.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
amisevsk: Good idea. Fixed, and I've added tests in `OpenShiftConnectorTest.java`

ibuziuk (approved):
:+1: good job @amisevsk

l0rd:
@amisevsk I didn't had the time to do a proper review of this PR but patching `che-host` in OpenShiftConnector doesn't seem a good solution:
- It's fragile, if the name `che-host` is changed (e.g. `che-server`) this doesn't work anymore
- We should sunset OpenShiftConnector so I would prefer a solution that would last even after OpenShiftConnector has been replaced.

So maybe you could find a way to let `che-host` be configurable with an IP or another hostname?
But if this looks too complicate nevermind. Let's merge that now and focus on the OpenShift implementation of the SPI.

amisevsk:
@l0rd Good point. I realize the `che-host` solution is far from ideal, but I'm not sure I see a good way around it. The alternative options as I see them are 
1. Use some other feature of the `che-host` service to get its instance (e.g. selector, but that is set externally to this repository)
2. Add yet another property that tells OpenShiftConnector what to expect for its own service name, which makes running on OpenShift even clumsier.

Additionally, the problem here is slightly deeper -- OpenShift integration depends on the `che-host` name in a few different places. At the moment, workspace -- server communication works only because the Che server service is named `che-host`:
- the property [`che.workspace.che_server_endpoint`](https://github.com/eclipse/che/blob/d81810e8686f9df07b952011f9ca8ff91684bf76/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties#L104) is defined to, by default, use `che-host` as a hostname.
- This property is used to define the `CHE_API` env var [here](https://github.com/eclipse/che/blob/d81810e8686f9df07b952011f9ca8ff91684bf76/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/ApiEndpointEnvVariableProvider.java)
- The provider [`CheDockerExternalHostProvider`](https://github.com/eclipse/che/blob/d81810e8686f9df07b952011f9ca8ff91684bf76/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/local/provider/CheDockerExtraHostProvider.java) is used to add an entry for `che-host` to [/etc/hosts](https://github.com/eclipse/che/blob/d81810e8686f9df07b952011f9ca8ff91684bf76/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/MachineProviderImpl.java#L635)
- When Che is running on Docker, this means that `che-host` resolves to the correct IP address. On OpenShift, `/etc/hosts` is managed by Kubernetes, so the `che-host` entry is never added. Instead, the internal OpenShift DNS allows the same env var, `CHE_API`, to work in workspaces without additional configuration. This means that if the service name for Che server is changed from `che-host`, communication between workspaces and wsmaster will fail regardless of the changes in this PR.

Either way, I think that this is a significant enough issue that it should be addressed separately, so I've created issue #5960 to track this.

benoitf:
@l0rd @amisevsk is that this PR is still actual ?

l0rd:
Nope. This PR had an interest before #6738 (merged this morning).

Il 19 ott 2017 11:34 PM, "Florent BENOIT" <notifications@github.com> ha
scritto:

benoitf:
@l0rd so we can close it ?

l0rd:
Yes we can close it

Issue #5334, opened 2017-06-12
Regression: Hover breaks PHP intellisense


kaloyan-raev:
A simple hovering over an element in the PHP editor breaks the PHP intellisense. I see the below error in the dev-machine output and then no intellisense works anymore.

```
[STDERR] Jun 12, 2017 7:31:08 AM org.eclipse.lsp4j.jsonrpc.json.ConcurrentMessageProcessor run
[STDERR] SEVERE: java.lang.IllegalStateException: Expected STRING but was BEGIN_OBJECT
[STDERR] com.google.gson.JsonSyntaxException: java.lang.IllegalStateException: Expected STRING but was BEGIN_OBJECT
[STDERR] 	at com.google.gson.internal.bind.ReflectiveTypeAdapterFactory$Adapter.read(ReflectiveTypeAdapterFactory.java:200)
[STDERR] 	at com.google.gson.Gson.fromJson(Gson.java:810)
[STDERR] 	at com.google.gson.Gson.fromJson(Gson.java:875)
[STDERR] 	at org.eclipse.lsp4j.jsonrpc.json.adapters.MessageTypeAdapterFactory$Adapter.createMessage(MessageTypeAdapterFactory.java:162)
[STDERR] 	at org.eclipse.lsp4j.jsonrpc.json.adapters.MessageTypeAdapterFactory$Adapter.read(MessageTypeAdapterFactory.java:130)
[STDERR] 	at org.eclipse.lsp4j.jsonrpc.json.adapters.MessageTypeAdapterFactory$Adapter.read(MessageTypeAdapterFactory.java:54)
[STDERR] 	at com.google.gson.Gson.fromJson(Gson.java:810)
[STDERR] 	at com.google.gson.Gson.fromJson(Gson.java:775)
[STDERR] 	at com.google.gson.Gson.fromJson(Gson.java:724)
[STDERR] 	at com.google.gson.Gson.fromJson(Gson.java:696)
[STDERR] 	at org.eclipse.lsp4j.jsonrpc.json.MessageJsonHandler.parseMessage(MessageJsonHandler.java:70)
[STDERR] 	at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.handleMessage(StreamMessageProducer.java:148)
[STDERR] 	at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.listen(StreamMessageProducer.java:77)
[...]

PR #5899, opened 2017-08-03
CHE-5334. Fix hover state

RomanNikitenko:
### What does this PR do?
Fix hover state

#5334

#### Changelog
Fix hover state

### DIFF
--- a/wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/service/TextDocumentService.java
+++ b/wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/service/TextDocumentService.java
@@ -363,7 +363,8 @@ public class TextDocumentService {
                 @Override
                 public boolean handleResult(InitializedLanguageServer element, Hover hover) {
                     if (hover != null) {
-                        result.getContents().addAll(hover.getContents());
+                        HoverDto hoverDto = new HoverDto(hover);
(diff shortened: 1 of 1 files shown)


SkorikSergey (approved):

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/service/TextDocumentService.java
```diff
@@ -363,7 +365,18 @@ public boolean canDo(InitializedLanguageServer element) {
                 @Override
                 public boolean handleResult(InitializedLanguageServer element, Hover hover) {
                     if (hover != null) {
-                        result.getContents().addAll(hover.getContents());
+                        List<Either<String, MarkedString>> contents =
```
tsmaeder: There is a constructor that makes a HoverDto out of a Hover in the generated HoverDto class. It may be simpler and safer to use that one.

dkulieshov (approved):

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/service/TextDocumentService.java
RomanNikitenko: @tsmaeder fixed, thank you!

PR #5915, opened 2017-08-04
Replace workspace object links with key-value links

### LINKED ISSUE #5581: Provide links of the workspace to the client
We already have at least two endpoints which may be useful for clients:
`output endpoint`
`workspace status endpoint`
So we need to rework our workspace links injector in accordance with client`s needs and changes in spi branch.
Note that the set of endpoints may expand.
We also want to simplify Links model. It is supposed to be done as a part of this issue.
The current model is complicated and, lately, we used to use it in the simplest use cases only. So we decided to simplify it significantly.
The current model of an object that contains links is: 
```json
{
    "field1" : "",
    "links" : [
        {
            "href" : "http://some-server.com:9090/path1/path2",
            "rel" : "myserver.api",
            "method" : "GET",
            "produces" : "application/json",
            "consumes" : "application/json",
            "parameters" : [
                {
                    "name" : "parameterName",
                    "defaultValue" : "defvalue1",
                    "description" : "Some user-friendly description",
                    "type" : "string",
                    "required" : true,
                    "valid" : [
                        "valid1",
                        "valid2",
                    ]
                }
            ],
            "requestBody" : {
                "description" : "body user-friendly description"
            }
        }
    ]
}
```
and the new model would be:
```json
{
    "field1" : "",
    "links" : {
        "myserver.api" : "http://some-server.com:9090/path1/path2",
        "anotherserver.api" : "http://another-server.com",
    }
}
```

voievodin:
Resolves #5581.

JSON responses of the Workspace REST API will include:

the following links if the workspace is stopped:
```json
"links": {
  "self": "http://localhost:8080/wsmaster/api/workspace/workspaceeyxr3nqkkuyaj3b7",
  "ide": "http://localhost:8080/che/test2"
}
```
the following links if the workspace is not stopped:
```json
"links": {
  "self": "http://localhost:8080/wsmaster/api/workspace/workspace28t00rschopreajr",
  "ide": "http://localhost:8080/che/test",
  "environment/statusChannel": "ws://localhost:8080/wsmaster/websocket",
  "environment/outputChannel": "ws://localhost:8080/output/websocket"
}
```


akorneta (approved):

garagatyi (approved):

skabashnyuk (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -218,11 +217,11 @@ public WorkspaceDto getByKey(@ApiParam(value = "Composite key",
...
                                                      String namespace) throws ServerException, BadRequestException {
-        return workspaceManager.getByNamespace(namespace, false)
-                               .stream()
-                               .filter(ws -> status == null || status.equalsIgnoreCase(ws.getStatus().toString()))
-                               .map(workspace -> linksInjector.injectLinks(asDto(workspace), getServiceContext()))
-                               .collect(toList());
+        return withLinks(workspaceManager.getByNamespace(namespace, false)
+                                         .stream()
+                                         .filter(ws -> status == null || status.equalsIgnoreCase(ws.getStatus().toString()))
+                                         .map(DtoConverter::asDto)
```
sleshchenko, line 223: Why we can't use `this::asDtoWithLinks` here?

voievodin: asDtoWithLinks throws checked exception

sleshchenko: I see, thanks

sleshchenko (approved):

Issue #5688, opened 2017-07-17
Rework resolving of machine's servers


sleshchenko:
OpenShift spi resolves machines servers when everything is created and ready to work.
It is needed to rework it to resolving all servers before pod creation to avoid analyzing openshift objects on each request and to have the same servers as configuration has in the case when some route, service is not configured properly.

bmicklea:
Please add correct labels.

PR #5916, opened 2017-08-04
CHE-5690 Add provisioning for OpenShift Environment

sleshchenko:
#### Major changes:
Adds provisioning for OpenShift Environment to wit adds all objects which are required for workspace start:
- After user's recipe parsing to OpenShiftEnvironment will be added container ports, services, routes which are required for configured servers.
- Installer configuration is applied to OpenShiftEnvironment, that is:
 -- Add container ports, services, routes that are required for servers of configured installers.
 -- Add environment variables that are configured in properties of configured installers.
 -- Add environment variables that are required by installers or agents after installations: CHE_API for all installers, CHE_WORKSPACE_ID for workspace agent.

#### Minor changes:
- Adds OpenShift Multi Pod stack.
- Change Openshift to OpenShift in names of classes.

https://github.com/eclipse/che/issues/5690


core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/config/ServerConfig.java
```diff
@@ -18,6 +18,15 @@
 public interface ServerConfig {
     /**
      * Port used by server.
+     *
+     * <p>It may contain protocol(TCP or UDP) after '/' symbol.
+     * If protocol is missed TCP will be used by default.
+     * Example:
+     * <ul>
+     *     <li>8080/TCP</li>
```
voievodin: Matching is key-sensitive at least if you're using `ServersMapper`, so you should either fix javadoc or make the mapping case-insensitive

garagatyi (approved):

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftMachine.java
```diff
@@ -86,65 +88,29 @@ public String getName() {
...
-                            if (portNumber.equals(servicePort.getPort())) {
-                                servers.put(Integer.toString(portNumber),
-                                            new ServerImpl("http://" + route.getSpec().getHost(),
-                                                           ServerStatus.UNKNOWN));
-                            }
-                            continue;
-                        }
+                if (matchedServices.contains(route.getSpec().getTo().getName())) {
+                    Map<String, String> annotations = route.getMetadata().getAnnotations();
+                    String serverName = annotations.get(CHE_SERVER_NAME_ANNOTATION);
```
voievodin, line 104: Maybe it makes sense to create a component similar to `org.eclipse.che.workspace.infrastructure.docker.Labels`, so you can define the constants there + implement annotations serialization and deserialization + it may be useful for feature recovery?

akorneta (approved):

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftMachine.java
```diff
@@ -272,6 +217,62 @@ public void onClose(KubernetesClientException cause) {
...
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private static boolean isExposedByService(Container container, Service service) {
+        for (ServicePort servicePort : service.getSpec().getPorts()) {
+            IntOrString targetPort = servicePort.getTargetPort();
+            if (targetPort.getIntVal() != null) {
```
voievodin, line 259: Is it possible to get targetPort without int val part?
I thought that it's either in format _8000_ or _8000/tcp_

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/ServerExposeUtils.java
```diff
@@ -0,0 +1,150 @@
...
+                                                          .stream()
+                                                          .filter(p -> p.getContainerPort().equals(port) &&
+                                                                       p.getProtocol().equals(protocol))
+                                                          .findAny();
+            ContainerPort containerPort;
+
+            if (exposedOpt.isPresent()) {
+                containerPort = exposedOpt.get();
+            } else {
+                containerPort = new ContainerPort(port,
```
voievodin: Consider using `ContainerPortBuilder`

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/ServerExposeUtils.java
```diff
@@ -0,0 +1,150 @@
...
+                                                  null,//hostIp
+                                                  null,//hostPort
+                                                  null,//name
+                                                  protocol);
+                container.getPorts().add(containerPort);
+            }
+
+            exposedPorts.put(portToExpose, new ServicePort("server-" + containerPort.getContainerPort(),
```
voievodin: Consider using `ServicePortBuilder`

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/provision/InstallerConfigApplier.java
```diff
@@ -0,0 +1,131 @@
...
+                                         environment,
+                                         "agents");
+
+                // CHE_API is used by installers for agent binary downloading
+                container.getEnv().removeIf(env -> "CHE_API".equals(env.getName()));
+                container.getEnv().add(new EnvVar("CHE_API", cheServerEndpoint, null));
+
+                // WORKSPACE_ID is required only by workspace agent
+                if (installers.contains("org.eclipse.che.ws-agent")) {
+                    container.getEnv().removeIf(env -> "CHE_WORKSPACE_ID".equals(env.getName()));
```
voievodin, line 95: It looks like this code should be part of 'General' infra independent provision or maybe installer configurations, so docker and openshift and any other implementation won't need to do it over and over again.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftMachine.java
sleshchenko: It makes sense. I considered this way and it will be implemented in scope of https://github.com/eclipse/che/issues/5688

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftMachine.java
sleshchenko: It is possible. Actually, targetPort has int value when it points out to container port like `8080` or it has string value when it points out to container port name like `ws-agent`.

voievodin: ok, thank you

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/provision/InstallerConfigApplier.java
sleshchenko: I agree with you. It would be nice to have common provision mechanism and then what infrastructure will have to implement is just applying prepared changes to its objects.
But now it will not be done in the scope of this issue.
It requires changes in the model because now it doesn't have environment variables anymore. @garagatyi WDYT will we implement this?

skabashnyuk (approved):

voievodin (approved):

Issue #5689, opened 2017-07-17
Add health checks for servers in openshift spi


sleshchenko:
Add health checks for servers in openshift spi for updating server statuses.
Also it is needed to publish statuses of servers to clients.

bmicklea:
Please add correct labels.

PR #5923, opened 2017-08-07
Reuse server health checks in OpenShift infrastructure

akorneta:
Moved server health checks from `org.eclipse.che.workspace.infrastructure.docker.server` into `org.eclipse.che.api.workspace.server.hc`.
Reuse server health checks mechanism for `internalStart` of OpenShift environment.

Resolves: https://github.com/eclipse/che/issues/5689


sleshchenko (approved):

skabashnyuk (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/ServerChecker.java
```diff
@@ -8,7 +8,7 @@
  * Contributors:
  *   Codenvy, S.A. - initial API and implementation
  *******************************************************************************/
-package org.eclipse.che.workspace.infrastructure.docker.server;
+package org.eclipse.che.api.workspace.server.hc;
```
voievodin, line 11: I find it a bit weird that these components are part of the workspace api module.
I clearly understand that spi model is defined under package `org.eclipse.che.api.workspace.server.spi`, so you need a place probably somewhere near the spi definition.

Consider the following module structure
```
--wsmaster
  |--che-workspace-spi
     |--model
     |--hc
     |--bootstrapper
  |--che-workspace-api
```
It allows you to place all the spi related stuff in the same module, which i think is good.
Clearly you might not want to do it under this PR, but what do you think about the structure?
@garagatyi, @sleshchenko, @skabashnyuk what do you guys think?

sleshchenko: I'm good with such approach. It would be better to separate workspace api, workspace spi and default spi stuffs implementations.

garagatyi: I think it can simplify things. @gazarenkov WDYT?

garagatyi (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/ServerChecker.java
garagatyi: but definitely not in this PR in my opinion

voievodin (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/ServerChecker.java
gazarenkov: @sleshchenko what is "default spi stuffs implementations" ?

sleshchenko: Now it is bootstrapping and server health checks. Maybe later we will have others like provisioning installers depending on configured installers.

PR #5931, opened 2017-08-08
CHE-5871: refactor DockerMachineStarter

### LINKED ISSUE #5871: Move docker container settings applying from DockerMachineStarter
Related to SPI branch only.

DockerMachineStarter class is old and big and it doesn't follow single responsibility principle. Because of this, it is hard to write tests to it as part of [this issue](https://github.com/eclipse/che/issues/5575).
We need to refactor it and cover its functionality with tests.
I suggest moving all docker system settings from this class to InfrastructureProvider class impl.
I also suggest that it can be split into several components for work with networks, images, containers.

garagatyi:
### What does this PR do?
Refactor DockerMachineStarter to clean it up and reduce its responsibilities. 
Move container configuration applying to InfrastructureProvisioners.
Move classes to different packages to create some structure and reduce classes mess in a single package.
Uncomment some commented tests.

Fixes #5871 

#### Changelog
Refactor DockerMachineStarter to clean it up


infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerMachine.java
```diff
@@ -153,16 +153,17 @@ public void exec(String script, MessageProcessor<LogMessage> messageProcessor) t
...
         } catch (IOException e) {
             throw new InternalInfrastructureException(e.getMessage(), e);
         }
         try {
-            docker.removeImage(RemoveImageParams.create(this.image).withForce(false));
+            docker.removeImage(RemoveImageParams.create(image).withForce(false));
         } catch (IOException e) {
-            LOG.error("IOException during destroy(). Ignoring.");
+            // TODO make log level warning if we ignoring it or remove ignoring phrase
+            LOG.error("IOException during destroy(). Ignoring.", e);
```
evoevodin: Can you please add exception message to the log

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/local/dod/DockerApiHostEnvVariableProvisioner.java
```diff
@@ -0,0 +1,60 @@
...
+import static java.lang.String.format;
+
+/**
+ * Provides DOCKER_HOST env variable for the sake of access to docker API within docker container
+ *
+ * @author Alexander Garagatyi
+ */
+@Singleton
+public class DockerApiHostEnvVariableProvisioner implements ConfigurationProvisioner {
+    private final Pair<String, String> result;
```
sleshchenko: Looks like you don't need `Pair` here. You can save value of the following expression
`format("tcp://%s:%s", dockerConnectorConfiguration.getDockerHost(), dockerDaemonUri.getPort())`
and then just use `DockerConnectorConfiguration.DOCKER_HOST_PROPERTY` constant as name of env variable.

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/provisioner/hosts/ExtraHostsProvisioner.java
```diff
@@ -0,0 +1,60 @@
...
+    private static final String PROPERTY = "che.workspace.hosts";
+
+    private final List<String> extraHosts;
+
+    @Inject
+    public ExtraHostsProvisioner(@Nullable @Named(PROPERTY) String extraHosts) {
+        if (isNullOrEmpty(extraHosts)) {
+            this.extraHosts = Collections.emptyList();
+        } else {
+            this.extraHosts = Arrays.stream(extraHosts.split(","))
```
voievodin: i'd better inject an array of strings, so there is no need to do the split, AFAIK `StringArrayConverter` will do it for you.

voievodin (approved):

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/local/dod/DockerApiHostEnvVariableProvisioner.java
garagatyi: Nice catch, ty.

infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/provisioner/hosts/ExtraHostsProvisioner.java
garagatyi: Looks like `StringArrayConverter` requires [space between values](https://github.com/garagatyi/che/blob/0e8b21ad3ff5da80a7a03ddf3f0b281cc3594dcb/core/commons/che-core-commons-inject/src/main/java/org/eclipse/che/inject/StringArrayConverter.java#L24-L24). WDYT?

voievodin: It requires zero or many spaces, thus it should be fine

garagatyi: My bad, you are right! I'll replace splitting.

akorneta (approved):

sleshchenko (approved):

skabashnyuk (approved):

Issue #5431, opened 2017-06-22
Add versions to Language Server download links in LS agent install scripts


evidolob:
We need to bind a version of the LS with the version of the Che. Now we don't use versions, so if we change some LS it's will affect all versions of Che.
Here sample how we use LS download link: https://github.com/eclipse/che/blob/f1f076437052942b4810bf2ad0fb2a360ff7967d/agents/ls-csharp/src/main/resources/org.eclipse.che.ls.csharp.script.sh#L18-L18
Need to add some version, it can be LS version, to the link.

https://github.com/eclipse/che/issues/2571

PR #5940, opened 2017-08-08
CHE-5431: Add versioning to installers

tolusha:
### What does this PR do?
Brings versioning to the installers:
* all installers have a version in the descriptors, for instance:
```{
  "id": "org.eclipse.che.ls.json",
  "version": "1.0.1",
  "name": "JSON language server",
  "description": "JSON intellisense",
  "dependencies": [],
  "properties": {}
}
```
* It is allowed now to add version as well as id in `stack.json`:
```
     "installers": [
                "org.eclipse.che.exec", "org.eclipse.che.ls.json:1.0.1"
              ],
```
If version is not set then the latest available will be used.

https://github.com/eclipse/che/issues/5431


core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/Version.java
```diff
@@ -0,0 +1,224 @@
...
+        result = 31 * result + milestone;
+        result = 31 * result + beta;
+        result = 31 * result + (snapshot ? 1 : 0);
+        result = 31 * result + (rc ? 1 : 0);
+        return result;
+    }
+
+    @Override
+    public int compareTo(Version o) {
+        if (major > o.major
```
sleshchenko: Really hard to understand the logic of this if block. Can we rework it to make it more clear? Or add a comment how versions should be compared?

sleshchenko:
@tolusha When actual issues should be solved?
When you wrote

> if user modified a script ... version in json file.

What did you mean about `user`? Because I understand `user` as a person who uses running che instance, or maybe who installs che server from binaries.
And in this case, the user doesn't have the ability to update json file.
If you mean that user is able to edit source code locally, rebuild it and reuse for himself. Then It is not good UX. And we should think about moving it to the external configuration or maybe there should be an ability to modify installers by REST without modifiyng source code.

tolusha:
I meant developer.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/Version.java
```diff
@@ -0,0 +1,224 @@
...
+import java.util.Comparator;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static java.lang.Integer.parseInt;
+import static java.lang.String.format;
+import static java.util.regex.Pattern.compile;
+
+/**
+ * @author Anatoliy Bazko
```
sleshchenko: Please add a comment about format of version

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/Version.java
```diff
@@ -0,0 +1,224 @@
...
+            return false;
+        }
+        if (rc != version.rc) {
+            return false;
+        }
+        if (snapshot != version.snapshot) {
+            return false;
+        }
+
+        return true;
```
sleshchenko: I'd prefer to rewrite it in the following way:
```
return beta == version.beta
          && ...
           && snapshot == version.snapshot;
```

agents/ls-json/src/main/resources/org.eclipse.che.ls.json.json
```diff
@@ -1,5 +1,6 @@
 {
   "id": "org.eclipse.che.ls.json",
+  "version": "1.0.1",
```
sleshchenko, line 3: Can you explain why it should have `1.0.1` but not `1.0.0` while `org.eclipse.che.ls.csharp.json` has `1.0.0`?

tolusha: For Json installer we already have 1.0.0 and 1.0.1 binaries.

agents/ls-typescript/src/main/resources/org.eclipse.che.ls.typescript.script.sh
```diff
@@ -26,7 +26,7 @@ unset PACKAGES
 command -v tar >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" tar"; }
 command -v curl >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" curl"; }
 
-AGENT_BINARIES_URI=https://codenvy.com/update/repository/public/download/org.eclipse.che.ls.typescript.binaries
+AGENT_BINARIES_URI=https://codenvy.com/update/repository/public/download/org.eclipse.che.ls.typescript.binaries/1.0.1
```
sleshchenko, line 29: Can you explain why `ls` agents binaries should be downloaded from `codenvy.com` while `ws-agent`, `terminal` etc will be mounted or downloaded from [ws-master](https://github.com/eclipse/che/blob/CHE-5431/wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh#L39)?
As far as I understand only the latest versions of `ws-master`, `terminal`, etc will be available. Can you confirm that it is OK?

tolusha: We don't have rights to distribute ls binaries.

>As far as I understand only the latest versions of ws-master, terminal, etc will be available. 

confirm

sleshchenko: Thanks!

agents/ls-json/src/main/resources/org.eclipse.che.ls.json.json
sleshchenko: I see. Thanks!

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/Version.java
```diff
@@ -0,0 +1,246 @@
...
+public class Version implements Comparable<Version> {
+
+    private static final String MILESTONE_VERSION_SUFFIX = "-M";
+    private static final String BETA_VERSION_SUFFIX      = "-beta-";
+
+    private static final Pattern VERSION =
+            compile("^(0|[1-9]+[0-9]*)\\.(0|[1-9]+[0-9]*)\\.(0|[1-9]+[0-9]*)(\\.0|\\.[1-9]+[0-9]*|)" +
+                    "(" + MILESTONE_VERSION_SUFFIX + "[1-9]+[0-9]*|)" +
+                    "(" + BETA_VERSION_SUFFIX + "[1-9]+[0-9]*|)" +
+                    "(-RC|)" +
```
sleshchenko: There should not be any difference but for me `(-RC)?` looks better than `(-RC|)` here.
P.S. The same about snapshot, milestone and other sections

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/Version.java
```diff
@@ -0,0 +1,246 @@
...
+        if (!milestoneGroup.isEmpty()) {
+            milestone = parseInt(milestoneGroup.substring(MILESTONE_VERSION_SUFFIX.length()));
+        }
+
+        String betaGroup = matcher.group(6);
+        if (!betaGroup.isEmpty()) {
+            beta = parseInt(betaGroup.substring(BETA_VERSION_SUFFIX.length()));
+        }
+
+        return new Version(parseInt(matcher.group(1)),
```
sleshchenko: Usage of named groups may help avoid error while changing the pattern.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/Version.java
```diff
@@ -0,0 +1,246 @@
...
+                           beta,
+                           !matcher.group(7).isEmpty(),
+                           !matcher.group(8).isEmpty());
+    }
+
+    /**
+     * Checks if version suites for pattern.
+     * For example 3.1.0 version is suited for 3.1.* or 3.*.0
+     */
+    public boolean isSuitedFor(String versionRegex) {
```
sleshchenko: I don't see any usage of this method except test. When may it be useful?

tolusha: It is a legacy.

wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/impl/LocalInstallerRegistry.java
```diff
@@ -88,12 +99,13 @@ public Installer getInstaller(String installerKey) throws InstallerException {
...
-    public List<Installer> getOrderedInstallers(List<String> installers) throws InstallerException {
+    public List<Installer> getOrderedInstallers(List<String> installerKeys) throws InstallerException {
         Map<InstallerFqn, Installer> sorted = new LinkedHashMap<>();
         Set<InstallerFqn> pending = new HashSet<>();
 
-        for (String installer : installers) {
-            doSort(InstallerFqn.parse(installer), sorted, pending);
+        for (String installer : installerKeys) {
+            InstallerFqn installerFqn = InstallerFqn.parse(installer);
+            doSort(stripOffLatestTag(installerFqn), sorted, pending);
```
sleshchenko: It is not defined by the interface that sorting will be performed by using latest versions instead of specified.
Please specify it or rename argument to `installer id` and do not parse it as `InstallerFqn`.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/Version.java
tolusha: Actually it makes difference.
`(-RC)` expects `-RC` but `(-RC|)` experts `-RC` or nothing

sleshchenko: Sorry but I wrote `(-RC)?` but not `(-RC)`.
And from https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html `
> X? -> X, once or not at all

sleshchenko (approved):
OK for me. Are you going to create new issues for described actual issues?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/Version.java
tolusha: Ahh I see.

garagatyi:
We have places where we hardcoded check of installer FQN. Such as these : 
- [InstallerBinariesInfrastructureProvisioner](https://github.com/garagatyi/che/blob/8be2f0e4dce1ba4ebb21c1fb2af08eb5cb676c27/infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/local/installer/InstallerBinariesInfrastructureProvisioner.java#L100-L100)
- [workspace shared util](https://github.com/garagatyi/che/blob/8be2f0e4dce1ba4ebb21c1fb2af08eb5cb676c27/wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Utils.java#L40-L40)
- [AddExecInstallerInEnvironmentUtil](https://github.com/garagatyi/che/blob/0504b313d8851bccc29e3cadb1adbd0b4f151094/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/agent/server/filters/AddExecInstallerInEnvironmentUtil.java#L34-L34)
- [EnvironmentValidator](https://github.com/garagatyi/che/blob/b397c9b75e183e50d8bf4d3a2fb22805ad18925c/infrastructures/docker/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/EnvironmentValidator.java#L93-L93)
- [InstallerConfigApplier](https://github.com/garagatyi/che/blob/949c1a2c26f702db556504951516c13c139a9510/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/provision/InstallerConfigApplier.java#L94-L94)

Maybe I have missed some other places.

Can you clarify how these places are supposed to work with all these changes in the installers versioning?

tolusha:
@garagatyi 
Thx. I will take a look.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/agent/server/filters/AddExecInstallerInEnvironmentUtil.java
```diff
@@ -44,4 +47,9 @@ public static void addExecInstaller(WorkspaceConfigDto workspaceConfig) {
             }
         }
     }
+
+    public static boolean containsTerminalInstallerWithoutExec(List<String> installers) {
```
sleshchenko: Is it used anywhere?

riuvshin (approved):

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/Version.java
```diff
@@ -0,0 +1,246 @@
...
+            throw new IllegalArgumentException(format("Illegal version '%s' format.", version));
+        }
+
+        int hotFix = 0;
+        int milestone = 0;
+        int beta = 0;
+
+        String hotFixGroup = matcher.group("hotfix");
+        if (hotFixGroup != null) {
+            hotFix = parseInt(hotFixGroup.substring(1));
```
sleshchenko: You can rewrite Pattern to avoid evaluating substring while parsing hotFix, milestone and beta versions.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/Version.java
sleshchenko: Then should not we remove it?

wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/model/impl/InstallerImpl.java
```diff
@@ -104,9 +103,11 @@ public InstallerImpl(String id,
                          Map<String, String> properties,
                          String script,
                          Map<String, ? extends ServerConfig> servers) {
+        Version.validate(version);
```
sleshchenko: I'm not sure that it is OK to do validation here. Maybe you should do it by InstallerRegistry, or InstallerService like workspace api does with workspace validation? See WorkspaceValidator.

sleshchenko (approved):
LGTM, except few small comment
:+1:

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/Version.java
```diff
@@ -0,0 +1,238 @@
...
+ *
+ * @author Anatoliy Bazko
+ */
+public class Version implements Comparable<Version> {
+
+    private static final String MILESTONE_VERSION_SUFFIX = "-M";
+    private static final String BETA_VERSION_SUFFIX      = "-beta-";
+
+    private static final Pattern VERSION =
+            compile("^(?<major>0|[1-9]+[0-9]*)\\.(?<minor>0|[1-9]+[0-9]*)\\.(?<bugfix>0|[1-9]+[0-9]*)" +
```
sleshchenko:
```java
Pattern.compile("^(?<major>0|[1-9]+[0-9]*)\\.(?<minor>0|[1-9]+[0-9]*)\\.(?<bugfix>0|[1-9]+[0-9]*)" +
                    "(\\.(?<hotfix>0|[1-9]+[0-9]*))?" +
                    "(" + MILESTONE_VERSION_SUFFIX + "(?<milestone>[1-9]+[0-9]*))?" +
                    "(" + BETA_VERSION_SUFFIX + "(?<beta>[1-9]+[0-9]*))?" +
                    "(?<rc>-RC)?" +
                    "(?<snapshot>-SNAPSHOT)?$");
```
```java
String hotFixGroup = matcher.group("hotfix");
if (hotFixGroup != null) {
    hotFix = parseInt(hotFixGroup);
}
```
WDYT?

tolusha: Agree, it is even simple!

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/Version.java
```diff
@@ -0,0 +1,238 @@
...
+                           hotFix,
+                           milestone,
+                           beta,
+                           matcher.group("rc") != null,
+                           matcher.group("snapshot") != null);
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
```
garagatyi: Isn't it more convenient to use `compareTo` here?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/Version.java
```diff
@@ -0,0 +1,238 @@
...
+    private static final String MILESTONE_VERSION_SUFFIX = "-M";
+    private static final String BETA_VERSION_SUFFIX      = "-beta-";
+
+    private static final Pattern VERSION =
+            compile("^(?<major>0|[1-9]+[0-9]*)\\.(?<minor>0|[1-9]+[0-9]*)\\.(?<bugfix>0|[1-9]+[0-9]*)" +
+                    "([.](?<hotfix>0|[1-9]+[0-9]*))?" +
+                    "(" + MILESTONE_VERSION_SUFFIX + "(?<milestone>[1-9]+[0-9]*))?" +
+                    "(" + BETA_VERSION_SUFFIX + "(?<beta>[1-9]+[0-9]*))?" +
+                    "(?<rc>-RC)?" +
+                    "(?<snapshot>-SNAPSHOT)?$");
```
garagatyi: I can imagine going even further and use something like next lines to eliminate usage of magic strings:
```
private static String MAJOR_GRP_NAME = "major";
...

private static String NON_ZERO_NUM_PART = "[1-9]+[0-9]*";
private static String ALLOW_ZERO_NUM_PART = "0|[1-9]+[0-9]*";

private static String MAJOR_GRP = "(?<major>" + ALLOW_ZERO_NUM_PART + ")";
private static String MINOR_GRP = "(?<minor>" + ALLOW_ZERO_NUM_PART + ")";
private static String BUGFIX_GRP = "(?<bugfix>" + ALLOW_ZERO_NUM_PART + ")";
private static String HOTFIX_GRP = "(\\.(?<hotfix>" + ALLOW_ZERO_NUM_PART + "))?";
private static String MILESTONE_GRP = "(" + MILESTONE_VERSION_SUFFIX + "(?<milestone>" + NON_ZERO_NUM_PART + "))?";
private static String BETA_GRP = "(" + BETA_VERSION_SUFFIX + "(?<beta>" + NON_ZERO_NUM_PART + "))?";
...
VERSION = Pattern.compile("^" + MAJOR_GRP + DOT + MINOR_GRP + DOT + BUGFIX_GRP + HOTFIX_GRP + MILESTONE_GRP + BETA_GRP + RC_GRP + SNAPSHOT_GRP);
```
And get groups by constants instead of magic strings too

core/che-core-api-core/src/test/java/org/eclipse/che/api/core/util/VersionTest.java
```diff
@@ -0,0 +1,166 @@
...
+                               {"1.0.1-SNAPSHOT"},
+                               {"1.0.1-RC-SNAPSHOT"},
+                               {"1.0.1.0-SNAPSHOT"},
+                               {"1.0.1-M1"},
+                               {"1.0.1.1-M1"},
+                               {"1.0.1-M1-SNAPSHOT"},
+                               {"1.0.1.2-M1-SNAPSHOT"},
+                               {"1.0.1.2-beta-1-SNAPSHOT"},
+                               {"1.0.1.2-M1-beta-1-SNAPSHOT"},
+                               {"1.0.1.2-M1-RC-SNAPSHOT"}};
```
garagatyi: Please ensure you haven't missed combinations such as 1.0.1.2-M1-beta-1-RC-SNAPSHOT

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/IoUtil.java
```diff
@@ -136,6 +143,36 @@ public static InputStream getResource(String resource) throws IOException {
         return is;
     }
 
+    /**
+     * Lists all children resources.
```
garagatyi: Javadocs don't say anything about consumer

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/IoUtil.java
```diff
@@ -136,6 +143,36 @@ public static InputStream getResource(String resource) throws IOException {
...
+     *      the root path represented in {@link URI} format
+     * @throws java.io.IOException
+     *      if any i/o error occur
+     * @throws ProviderNotFoundException
+     *      if a provider supporting the URI scheme is not installed
+     */
+    public static void listResources(URI parent, Consumer<Path> consumer) throws IOException {
+        FileSystem fileSystem = null;
+        try {
+            if (!parent.getScheme().equals("file")) {
```
garagatyi: If it is possible that provided URI is relative and this line throws NPE, argument can be used on the left side instead

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/IoUtil.java
```diff
@@ -136,6 +143,36 @@ public static InputStream getResource(String resource) throws IOException {
...
+                try {
+                    fileSystem = FileSystems.newFileSystem(parent, Collections.emptyMap());
```
garagatyi: Can you elaborate on why this if-clause is needed and lines 166-167 are not enough?

wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/InstallerRegistry.java
```diff
@@ -109,10 +109,13 @@
 
     /**
      * Traverses dependencies of all listed installers and
-     * returns properly ordered list of non-duplicated installer descriptions
+     * returns properly ordered list of non-duplicated installer descriptions.
      *
-     * @param installers
-     *         installers to fetch dependencies and order
+     * If any of {@code installerKeys} contains only id then the latest version of this installer
```
garagatyi: FYI empty lines in javadocs are ignored. If you want to put a brake usually `<p/>` is used.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/Utils.java
```diff
@@ -8,10 +8,11 @@
  * Contributors:
  *   Codenvy, S.A. - initial API and implementation
  *******************************************************************************/
-package org.eclipse.che.api.workspace.shared;
```
garagatyi: It is supposed to be used on GWT side also. Can you elaborate on why do you think it should be moved on servers side?

tolusha: Not to add dependency to `che-core-api-installer`

garagatyi: Is there a class that helps GWT code to find dev machine?

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/IoUtil.java
tolusha: To get file system based on `file` scheme the URI has to be like this `file:/`. In other words `newFileSystem` methods accept the root of the file system for 'file' scheme.


So anyway I needed to add `if` clause, so I decided to omit to request file system since it exists anyway and I am able to `Paths.get(parent);` not to afraid arising  `FileSystemNotFoundException`

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/Utils.java
tolusha: I will move `Utils` back to shared module.

wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/impl/LocalInstallerRegistry.java
```diff
@@ -37,39 +44,71 @@
...
-                installerDao.create(installer);
-            } catch (InstallerAlreadyExistsException e) {
-                // ignore
-            }
+    private void doInit(InstallerDao installerDao, Installer i) throws InstallerException {
+        String installerKey = InstallerFqn.of(i).toKey();
+
+        try {
+            Version.validate(i.getVersion());
+        } catch (IllegalArgumentException e) {
```
sleshchenko: Mb then you can throw `InstallerException` by `Version.validate`? Also, I'd create `VersionValidator` class to avoid static calls.

tolusha: `IllegalVersionException` ?

tolusha: Actually I don't see much benefits.

sleshchenko: You wouldn't need to remap this exception three times

tolusha: Then it is better to have 
```
public class InstallerValidator {
 public void validate(Installer installer) throws InstallerException {
 try {
            Version.validate(installer.getVersion());
        } catch (IllegalArgumentException e) {
            throw new InstallerException(format("Illegal version format '%s' for installer '%s'",
                                                installer.getVersion(),
                                                InstallerFqn.of(installer).toKey()));
        }
}
}

Comment (tolusha) [reply wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/impl/LocalInstallerRegistry.java]: WDYT?

Comment (sleshchenko) [reply wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/impl/LocalInstallerRegistry.java]: Sounds good for me.

File: core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/Version.java (hunk 14 above)
Comment (tolusha) [reply core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/Version.java]: What do you mean?

Comment (garagatyi) [reply core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/Version.java]: `o1.compareTo(o2) == 0`

**Comment by tolusha** [comment]
Conflicts resolved, `Utils` moved to another module.
@garagatyi

**Review by garagatyi — APPROVED** [review]

File: wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/Utils.java [hunk 23]
```diff
@@ -37,12 +38,8 @@ public static String getDevMachineName(Environment envConfig) {
         for (Map.Entry<String, ? extends MachineConfig> entry : envConfig.getMachines().entrySet()) {
             // TODO should we use server ref instead of installers?
             List<String> installers = entry.getValue().getInstallers();
-            if (installers != null) {
-                for (String installerKey : installers) {
-                    if (installerKey.equals(WSAGENT_INSTALLER) || installerKey.startsWith(WSAGENT_INSTALLER + ":")) {
```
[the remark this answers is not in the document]
garagatyi: It is already implemented by installers matching, so I see no reason for Anatoliy to change that. I will remove this class and rework logic of matching to server based in my following PR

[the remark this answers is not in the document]
gazarenkov: I see that it is implemented per-installer, that causes my comment in fact.
The question is why ("you see no reason to change that")?
@tolusha @garagatyi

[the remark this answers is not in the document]
garagatyi: because I already deleted it in my branch and there is no sense to double this piece of work in Anatoliy's PR. I will create a PR shortly

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/Version.java
```diff
@@ -0,0 +1,238 @@
...
+ * Version format:
+ * {MajorVersion}.{MinorVersion}.{BugFixVersion}.{HotFixVersion}-M{MilestoneVersion}-beta-{BetaVersion}-RC-SNAPSHOT.
+ * </p>
+ * Mandatory parts are only major, minor and bug fix versions.
+ *
+ * @see #toString()
+ * @see #VERSION
+ *
+ * @author Anatoliy Bazko
+ */
```
skabashnyuk: Can you replace this class with https://maven.apache.org/ref/3.0.3/maven-artifact/apidocs/org/apache/maven/artifact/versioning/ComparableVersion.html?

skabashnyuk (approved):

PR #5945, opened 2017-08-08
Update entrypoint.sh

JamesDrummond:
JAVA_OPTS is set by line 321 `export JAVA_OPTS="${JAVA_OPTS} -Dche.docker.network=bridge"` to `export JAVA_OPTS=" -Dche.docker.network=bridge"` if JAVA_OPTS is empty. This causes 152 and 154 to never set JAVA_OPTS as it will always have a value. I propose to move line 321 to execute after lines 152/154.

### What does this PR do?
Allows JAVA_OPTS to be set in lines 152 and 154.

#### Changelog
Allow default values of JAVA_OPTS to be set in che-server container entrypoint script if no value is given to environment variable JAVA_OPTS by user or cli.

#### Release Notes
N/A

#### Docs PR
N/A

Signed-off-by: James Drummond james@devcomb.com

### DIFF
--- a/dockerfiles/che/entrypoint.sh
+++ b/dockerfiles/che/entrypoint.sh
@@ -156,7 +156,7 @@ call_catalina () {
 
   ### Cannot add this in setenv.sh.
   ### We do the port mapping here, and this gets inserted into server.xml when tomcat boots
-  export JAVA_OPTS="${JAVA_OPTS} -Dport.http=${CHE_PORT} -Dche.home=${CHE_HOME}"
+  export JAVA_OPTS="${JAVA_OPTS} -Dche.docker.network=bridge -Dport.http=${CHE_PORT} -Dche.home=${CHE_HOME}"
   export SERVER_PORT=${CHE_PORT}
 
   # Launch the Che application server, passing in command line parameters
@@ -317,8 +317,6 @@ init() {
     cp -rf "${CHE_HOME}"/templates/* "${CHE_DATA}"/templates
   fi
 
-  # A che property, which names the Docker network used for che + ws to communicate
-  export JAVA_OPTS="${JAVA_OPTS} -Dche.docker.network=bridge"
 }
 
 get_che_data_from_host() {
--- a/dockerfiles/che/entrypoint.sh
+++ b/dockerfiles/che/entrypoint.sh
@@ -149,9 +149,9 @@ call_catalina () {
 
   ### Initialize default JVM arguments to run che
   if [[ "${CHE_BLOCKING_ENTROPY}" == true ]]; then
-    [ -z "${JAVA_OPTS}" ] && JAVA_OPTS="-Xms256m -Xmx1024m -Djava.security.egd=file:/dev/./urandom"
+    [ -z ${JAVA_OPTS} ] && JAVA_OPTS="-Xms256m -Xmx1024m -Djava.security.egd=file:/dev/./urandom"
   else
-    [ -z "${JAVA_OPTS}" ] && JAVA_OPTS="-Xms256m -Xmx1024m"
+    [ -z ${JAVA_OPTS} ] && JAVA_OPTS="-Xms256m -Xmx1024m"
   fi
 
   ### Cannot add this in setenv.sh.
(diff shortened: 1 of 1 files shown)


benoitf:
I'm not sure to follow what is doing the PR

AFAIK sequence in sh is :

`init`
then
`start-tomcat`

so JAVA_OPTS is initialized in init() by
```
export JAVA_OPTS="${JAVA_OPTS} -Dche.docker.network=bridge"
```
(it concatenates che.docker.network to existing options or set new value if not set)

then it start tomcat method we have
```
export JAVA_OPTS="${JAVA_OPTS} -Dport.http=${CHE_PORT} -Dche.home=${CHE_HOME}"
```
which concatenates options as well

so AFAIK the options are always set, no matter if we have a previous JAVA_OPTS or not

riuvshin:
@benoitf is right entrypoint works fine, and btw in che multi user that part actually changed, So there is no value in that pr it will bring only merge conflicts later.

JamesDrummond:
@benoitf The code at [https://github.com/eclipse/che/blob/master/dockerfiles/che/entrypoint.sh#L154](https://github.com/eclipse/che/blob/master/dockerfiles/che/entrypoint.sh#L154) is

`[ -z "${JAVA_OPTS}" ] && JAVA_OPTS="-Xms256m -Xmx1024m -Djava.security.egd=file:/dev/./urandom"`

This means `JAVA_OPTS="-Xms256m -Xmx1024m -Djava.security.egd=file:/dev/./urandom"` will be executed if `[ -z "${JAVA_OPTS}" ]` returns 1. This is a little counter intuitive as "-z String1 - the length of the String1 variable is 0 (zero)" will only return 1 in our case if `${JAVA_OPTS}` is empty. This will never be the case as `${JAVA_OPTS}` is always set before. 

I had particular trouble on CentOS where `-Djava.security.egd=file:/dev/./urandom` is needed and only by me adding `-e JAVA_OPTS="-Xms256m -Xmx1024m -Djava.security.egd=file:/dev/./urandom"` to my docker run command was I able to fix the problem even when I tried setting `-e CHE_BLOCKING_ENTROPY=false`. 

Setting `-e CHE_BLOCKING_ENTROPY=false` is also a possible problem with the code as I think it should be reversed `CHE_BLOCKING_ENTROPY                Starts Tomcat with blocking entropy: -Djava.security.egd=file:/dev/./urandom`. 

```
### Initialize default JVM arguments to run che
  if [[ "${CHE_BLOCKING_ENTROPY}" == true ]]; then
    [ -z "${JAVA_OPTS}" ] && JAVA_OPTS="-Xms256m -Xmx1024m"
  else
    [ -z "${JAVA_OPTS}" ] && JAVA_OPTS="-Xms256m -Xmx1024m -Djava.security.egd=file:/dev/./urandom"
  fi
```

Shouldn't this be the reverse?

```
  if [[ "${CHE_BLOCKING_ENTROPY}" == true ]]; then
      [ -z "${JAVA_OPTS}" ] && JAVA_OPTS="-Xms256m -Xmx1024m -Djava.security.egd=file:/dev/./urandom"
   else
     [ -z "${JAVA_OPTS}" ] && JAVA_OPTS="-Xms256m -Xmx1024m"
  fi
```
 Bellow is some testing I did in the terminal which I am glad I did as quotes cannot be used with test.
```
[root@scw-94b5af dev]# [ -f "" ] && echo hello
[root@scw-94b5af dev]# [ -f " -Dche.docker.network=bridge" ] && echo hello
[root@scw-94b5af dev]# [ -f  ] && echo hello
hello
[root@scw-94b5af dev]# JAVA_OPTS=" -Dche.docker.network=bridge"
[root@scw-94b5af dev]# [ -f $JAVA_OPTS ] && echo hello
[root@scw-94b5af dev]# JAVA_OPTS=
[root@scw-94b5af dev]# [ -f $JAVA_OPTS ] && echo hello
hello
[root@scw-94b5af dev]#
```

skabashnyuk:
@JamesDrummond are you going to continue maintaining this pr or we can close it?

skabashnyuk:
Closing due @riuvshin comment https://github.com/eclipse/che/pull/5945#issuecomment-321176461

Issue #5892, opened 2017-08-03
Add navigation in the diff window to the next/prev file


mmorhun:
Implement `Next` and `Prev` buttons in Git Diff window. This also require adding of `Save changes` button.

mmorhun:
Done.

Issue #5925, opened 2017-08-07
Add hotkeys for the most usefull git actions


mmorhun:
It will be useful to add hotkeys for the most usefull git operations.
Also it will be great to have separate group of shortcuts for git operations in the keybinding list.

List of shortcuts:
- git commit `Alt + c`
- git push `Alt + Shift + c`
- git pull `Alt + p`
- git branches `Ctrl + b`

gazarenkov:
Could you please specify what exactly operations you have in mind, it sounds quite strange otherwise

mmorhun:
It is under discussion now. Actually this issue is the place where others can describe their vision for this.
But we think that we have to add shortcuts for:
- git commit
- git push
- git pull
- git branches

mmorhun:
@slemeur I updated description. Please take a look and approve or give some proposals.

slemeur:
seems okay for me.
Don't forget that those keyboard shortcuts needs to be in the docs + in the `Assistant` -> `Key bindings`

tolusha:
@mmorhun 
Why shortcuts aren't the same as intellij idea has?

mmorhun:
@tolusha because we aren't intellij idea.

tolusha:
Your explanation hits the nail on the head :)

mmorhun:
Done

Issue #5128, opened 2017-05-18
Git dedicated panel and integration


slemeur:
# Goal
The current Eclipse Che version is providing a Git plugin with many of features but the overall experience is tedious for users. Simple actions are needing a lot of interactions and the user flow have been made complex. The current git integration is also lacking of feedbacks to show to the user the changes that have been made. 

The goal of this epic is to rethink the Git experience in order to improve the developer efficiency and provide integrated features (in the editor and in the explorer). 

We'll provide a new Git panel, which will be displayed in the left area. This panel will allow the user to do the most common Git actions more easily and will progressively replace certain of the features currently implemented in the git plugin. 

# Linked issues and discussions
- [x] https://github.com/eclipse/che/issues/1605
- [x] https://github.com/eclipse/che/issues/3614
- [x] https://github.com/eclipse/che/issues/4969
- [x] https://github.com/eclipse/che/issues/3415
- [x] https://github.com/eclipse/che/issues/3416
- [x] https://github.com/eclipse/che/issues/5335
- [x] https://github.com/eclipse/che/issues/5339
- [x] https://github.com/eclipse/che/issues/5340

# Subtasks
- [x] Add git panel to display local changes https://github.com/eclipse/che/issues/6361
- [x] Synchronize list of repositories and the list of local changes https://github.com/eclipse/che/issues/6608
- [x] Apply new styles to the widget displaying local changes https://github.com/eclipse/che/issues/6645
- [x] Port git dedicated panel to che6 https://github.com/eclipse/che/issues/7022
- [x] Git commit window tree files view selection model confused 
https://github.com/eclipse/che/issues/6255
- [ ] Ability to commit files in git dedicated panel https://github.com/eclipse/che/issues/7085
- [ ] Hide panel if there is no git projects
## Details

**Git Panel:**
We'll introduce a new git panel which is composed of 3 main areas:
- Repositories: to select the project to interact with
- History: to select local/remote branches, browse previous changes and interact with the remote repo to fetch, pull or push
- Local: to see the changes and execute a commit.

![ide 6 - git panel v2](https://cloud.githubusercontent.com/assets/1636769/26212523/1cac37bc-3bf6-11e7-8f2b-13b3eb9abc67.png)

Commented:
![ide 6 - git panel v2 - explained](https://cloud.githubusercontent.com/assets/1636769/26212528/272aa278-3bf6-11e7-9676-693bd6a43bec.png)

**Commit Options:**
[...]

PR #5965, opened 2017-08-10
Add ability to switch between files in Git Diff widget

mmorhun:
### What does this PR do?
- Adds ability to switch to the next/previous file in git compare widget. The order is defined by git. The editable content is saved on `Next`/`Previous` button click.
- Adds Save Changes button for git compare widget.
 - Performs refactoring of git-compare-related functionality.
- Fixes compare with deleted file bug.
- Moves Git Diff widget from iframe to IDE.
- Adds hotkeys for `next` and `previous` diff: `alt + ,` and `alt + .` correspondingly. (actually `alt` + `<` or `>` but without `shift`).

https://github.com/eclipse/che/issues/5892

#### Changelog
Added ability to switch between files in Git Diff widget.

#### Release Notes
In this version, we continued to work on improving Git support in Eclipse Che. We have improved the Git diff window so that it's simpler and faster to review changes between two states of code.

git-diff-2

To view the diff, use `Git`->`Compare`->`<Select-to-what>` from main menu. If more than one file has changed the list of changed files will be opened first. To select a file to compare double-click on it or select a file then click the `Compare` button.

Your changes are displayed in the left editor and the file being compared to is on the right. The left editor can be used for editing and fixing your changes.

When you have multiple files to review you have the ability to navigate under all the files that are changed. The number of files that are reviewable is displayed in the title of the wizard. Use `Next` and `Previous` buttons to navigate to the next or previous file correspondingly. We also introduce new keyboard shortcut for those actions `Alt + .` for `Next` and `Alt + ,` for `Previous`.

#### Docs PR
https://github.com/eclipse/che-docs/pull/283


plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/action/CompareWithLatestAction.java
```diff
@@ -91,21 +87,11 @@ public void actionPerformed(ActionEvent e) {
...
+                       ChangedItems changedItems = new ChangedItems(project, diff);
+                       if (changedItems.getFilesQuantity() == 1) {
+                           comparePresenter.showCompareWithLatest(changedItems, null, REVISION);
                        } else {
-                           Map<String, Status> items = new HashMap<>();
-                           for (String item : changedFiles) {
-                               items.put(item.substring(2, item.length()), defineStatus(item.substring(0, 1)));
-                           }
-                           changesListPresenter.show(items, REVISION, null, project);
+                           changesListPresenter.show(changedItems, REVISION, null);
```
tolusha: I personally don't like methods which can accept null values.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ChangedItems.java
```diff
@@ -0,0 +1,107 @@
...
+     *         the project under diff operation
+     * @param diff
+     *         plain result of git diff operation
+     */
+    public ChangedItems(Project project, String diff) {
+        this.project = project;
+
+        changedFilesStatuses = new LinkedHashMap<>();
+        for (String item : diff.split("\n")) {
+            changedFilesStatuses.put(item.substring(2, item.length()), defineStatus(item.substring(0, 1)));
```
tolusha: if `diff.isEmpty()` then you will have an issue

tolusha: Make sure `item.length() > 2`
If `defineStatus(item.substring(0, 1))` throws `IllegalArgumentException` then catch it

tolusha: To have clear code extract variables for `item.substring(2, item.length())` and `defineStatus(item.substring(0, 1))`

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ChangedItems.java
```diff
@@ -0,0 +1,107 @@
...
+
+        length = changedFilesList.size();
+    }
+
+    public Project getProject() {
+        return project;
+    }
+
+    /**
+     * @return number of files in the diff
```
vinokurig: I prefer not to use single `@return` annotation in javadoc, but to have something like: "Returns number of files in the diff"

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ChangedItems.java
```diff
@@ -0,0 +1,107 @@
...
+        this.project = project;
+
+        changedFilesStatuses = new LinkedHashMap<>();
+        for (String item : diff.split("\n")) {
+            changedFilesStatuses.put(item.substring(2, item.length()), defineStatus(item.substring(0, 1)));
+        }
+
+        changedFilesList = new ArrayList<>(changedFilesStatuses.keySet());
+
+        length = changedFilesList.size();
```
tolusha: Redundant field

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ChangedItems.java
```diff
@@ -0,0 +1,107 @@
...
+ */
+public class ChangedItems {
+
+    private final Project                       project;
+    private final LinkedHashMap<String, Status> changedFilesStatuses;
+    private final List<String>                  changedFilesList;
+    private final int                           length;
+
+    /**
+     * Creates user-friendly representation of git diff.
```
vinokurig: What do you mean user-frendly? How does it related to user?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ChangedItems.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.Objects;
+
+import static org.eclipse.che.ide.ext.git.client.compare.FileStatus.defineStatus;
+
+/**
+ * Describes changed files in git comparison process.
+ *
+ * @author Mykola Morhun
+ */
+public class ChangedItems {
```
tolusha: ModifiedFiles ?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -75,45 +78,106 @@ public ComparePresenter(AppContext appContext,
...
+    public void showCompareWithLatest(final ChangedItems changedItems,
+                                      @Nullable final String currentFile,
+                                      final String revision) {
+        this.changedItems = changedItems;
         this.revision = revision;
+
         this.compareWithLatest = true;
+        view.setEnableSaveChangesButton(true);
+
+        setupCurrentFile(currentFile);
```
tolusha: You don't setup anything. Rename it respectively.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -75,45 +78,106 @@ public ComparePresenter(AppContext appContext,
...
+     * Type of comparison to show depends on {@code compareWithLatest} field.
+     */
+    private void showCompareForCurrentFile() {
+        view.setEnableNextDiffButton(currentItemIndex != (changedItems.getFilesQuantity() - 1));
+        view.setEnablePreviousDiffButton(currentItemIndex != 0);
+
+        changedItems.getProject()
+                    .getFile(changedItems.getItemByIndex(currentItemIndex))
+                    .then(file -> {
+                        if (file.isPresent()) {
```
tolusha: if `!file.isPresent()` it means it has been deleted. So, we have to show diff also.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -214,4 +292,55 @@ private void showCompare(final String remoteContent) {
...
+    /** @return relative path of given file from specified project */
+    private Path getRelPath(final Container project, final File file) {
+        return file.getLocation().removeFirstSegments(project.getLocation().segmentCount());
+    }
+
+    /**
+     * Searches for project with git repository to which given file belongs.
+     */
+    @Nullable
+    private Container getGitDir(final File file) {
```
tolusha: `getGitRootFolder`

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -75,45 +78,106 @@ public ComparePresenter(AppContext appContext,
...
         if (status.equals(ADDED)) {
             showCompare("");
             return;
         }
 
-        final Container rootProject = GitUtil.getRootProject(file);
-
-        if (rootProject == null) {
+        final Container gitDir = getGitDir(file);
+        if (gitDir == null) {
```
tolusha: Should we notify user with some message?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -181,30 +232,57 @@ public void showCompareBetweenRevisions(final Path file,
...
 
         CancelCallback cancelCallback = view::hide;
 
         dialogFactory.createConfirmDialog(locale.compareSaveTitle(), locale.compareSaveQuestion(), locale.buttonYes(), locale.buttonNo(),
                                           confirmCallback, cancelCallback).show();
     }
 
+    @Override
+    public void onSaveChangesClicked() {
+        if (compareWithLatest) {
```
tolusha: Why do we have to check if `compareWithLatest` ?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -214,4 +292,55 @@ private void showCompare(final String remoteContent) {
...
+                        eventBus.fireEvent(new FileContentUpdateEvent(comparedFile.getLocation()
+                                                                                  .toString()));
+                    })
+                    .catchError(error -> {
+                        notificationManager.notify(error.getMessage(), FAIL, NOT_EMERGE_MODE);
+                    });
+    }
+
+    /** @return relative path of given file from specified project */
+    private Path getRelPath(final Container project, final File file) {
```
tolusha: It seems like a method for `File` class.

plugins/plugin-git/che-plugin-git-ext-git/src/main/resources/org/eclipse/che/ide/ext/git/client/GitLocalizationConstant.properties
```diff
@@ -29,6 +29,9 @@ button.no=No
 button.fetch=Fetch
 button.push=Push
 button.pull=Pull
+button.save_changes=Save Changes
+button.next_diff=Next
+button.previous_diff=Prev
```
tolusha: `Previous` ?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ChangedItems.java
mmorhun: well, programmer friendly)

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ChangedItems.java
mmorhun: What if it is added or deleted?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/action/CompareWithLatestAction.java
mmorhun: Well, I don't like it too, but here I see no problem. It is well documented and mean undefined.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ChangedItems.java
mmorhun: No, it will end up with empty diff

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ChangedItems.java
mmorhun: ok

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ChangedItems.java
vinokurig: I dont think that Modified or Changed are good words for this class because such words are present in Git Status. This object can contain not only modified/changed but new or deleted
How about `DiffFiles`?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ChangedItems.java
mmorhun: Just cached. But ok, I'll get rid of it.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ChangedItems.java
mmorhun: I think `TouchedItems` is worse. But still open for discussion.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ChangedItems.java
vinokurig: realy? :smile:

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
mmorhun: I disagree to remove it. But name could be changed, ok.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
mmorhun: See no logical difference here.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -75,45 +78,106 @@ public ComparePresenter(AppContext appContext,
     }
 
     /**
-     * Show compare window.
+     * Show compare window for given set of files between given revision and latest code version.
```
vinokurig: `latest code version` is not clear enough, I think it should be `HEAD`

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
mmorhun: Just for case.
The Save Changes button should be disabled when compare between revisions is shown.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -75,45 +78,106 @@ public ComparePresenter(AppContext appContext,
...
      *
-     * @param file
-     *         file name with its full path
-     * @param status
-     *         status of the file
+     * @param changedItems
+     *         ordered touched files
```
vinokurig: What do you mean `touched files` ?

vinokurig: `ordered` -> `sorted`?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
mmorhun: Maybe, but I think it doesn't 'fit' for current `File` interface

plugins/plugin-git/che-plugin-git-ext-git/src/main/resources/org/eclipse/che/ide/ext/git/client/GitLocalizationConstant.properties
mmorhun: Let be so

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
mmorhun: Yes, agree

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
mmorhun: Any files which were changed in any way, e.g. modified, added, deleted etc

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/action/CompareWithLatestAction.java
dkulieshov: I'm a bit out of the scope however in general I agree with @tolusha. In this very example I would recommend to avoid public method calls with `null` values (even well documented). I think it would be nice if you could add a corresponding method so we could call `changesListPresenter.show(changedItems, REVISION)` and it would internally recall `changesListPresenter.show(changedItems, REVISION, null)` or something similar to this.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
mmorhun: ok

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ChangedItems.java
dkulieshov: I'm not sure I understand why do we use term Item here if we're talking exactly about files only according to javadoc? What do you think about something like `AlteredFiles`? In general I think we can't just name this class `ChangedItems` as it does not reflect it's nature, personally I would split it into several classes something like `ChangedItemParser`, `ChangedItemHolder` and `ChangedItem` itself, what do you think?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/commit/CommitPresenter.java
```diff
@@ -76,7 +71,7 @@
     private final ProcessesPanelPresenter consolesPanelPresenter;
 
     private Project      project;
-    private Set<String>  allFiles;
+    private List<String> allFiles;
```
dkulieshov: Could you please explain why we're changing a set for a list? We need to store same files several times?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
mmorhun: `ordered` != `sorted`

vinokurig: `touched` is not clear for me, How about just `Changed Files` ?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
mmorhun: Good catch!

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/commit/CommitPresenter.java
mmorhun: No, we needn't, but git returns modified files in some order, so I just do not want to loose it. Also it is needed to navigate through this list.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
mmorhun: By the way, seems that this is broken in master

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/action/CompareWithLatestAction.java
mmorhun: Good piece of advice, but not for this case, unfortunately...
Otherwise we have to make additional methods with different names (to cover two nullable parameters), which will cause additional logic on the invocations.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ChangedItems.java
mmorhun: `AlteredFiles` is good name, think no one will be against it.
As about splitting, I see no point to do this. Constructor just do its own work. Its logic isn't complicated to move into separate class. And according to current representation of diff, storing data into internal object will make it more complicated and add small overhead I think.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/commit/CommitPresenter.java
tolusha: May be 
`private AlteredFiles alteredFiles;` ?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.Objects;
+
+import static org.eclipse.che.ide.ext.git.client.compare.FileStatus.defineStatus;
+
+/**
+ * Describes changed in any way files in git comparison process.
+ *
+ * @author Mykola Morhun
+ */
+public class AlteredFiles {
```
tolusha, line 29: I would like to see some tests.

mmorhun:
Diff widget with new `Next`/`Previous` and `Save Changes` buttons
diff

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
mmorhun: Done

vinokurig: `AlteredFilesHolder` ?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
```diff
@@ -0,0 +1,106 @@
...
+
+    private final Project                       project;
+    private final LinkedHashMap<String, Status> changedFilesStatuses;
+    private final List<String>                  changedFilesList;
+
+    /**
+     * Git diff representation.
```
vinokurig: This javadoc is not clear for me.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
```diff
@@ -0,0 +1,106 @@
...
+     * @param project
+     *         the project under diff operation
+     * @param diff
+     *         plain result of git diff operation
+     */
+    public AlteredFiles(Project project, String diff) {
+        this.project = project;
+
+        changedFilesStatuses = new LinkedHashMap<>();
+        if (!"".equals(diff)) {
```
vinokurig: `isNullOrEmpty` ?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
```diff
@@ -0,0 +1,106 @@
...
+            for (String item : diff.split("\n")) {
+                changedFilesStatuses.put(item.substring(2, item.length()), defineStatus(item.substring(0, 1)));
+            }
+        }
+
+        changedFilesList = new ArrayList<>(changedFilesStatuses.keySet());
```
vinokurig: The class is called `AlteredFiles, may be it would be better to rename this variable to `alteredFilesList` ?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    public Project getProject() {
```
vinokurig: Javadoc?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    /**
+     * Returns number of files in the diff.
+     */
+    public int getFilesQuantity() {
+        return changedFilesList.size();
+    }
+
+    public boolean isEmpty() {
```
vinokurig: Javadoc?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
```diff
@@ -0,0 +1,106 @@
...
+        return 0 == changedFilesList.size();
+    }
+
+    public Map<String, Status> getChangedFilesMap() {
```
vinokurig: Javadoc?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
```diff
@@ -0,0 +1,106 @@
...
+        return changedFilesStatuses;
+    }
+
+    public List<String> getChangedFilesList() {
+        return changedFilesList;
```
vinokurig: Javadoc?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    public Status getStatusByFilePath(String pathToChangedItem) {
+        return changedFilesStatuses.get(pathToChangedItem);
```
vinokurig: Javadoc?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    public Status getStatusByIndex(int index) {
```
vinokurig: Javadoc?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
```diff
@@ -0,0 +1,106 @@
...
+        return changedFilesStatuses.get(changedFilesList.get(index));
+    }
+
+    public String getFileByIndex(int index) {
+        return changedFilesList.get(index);
```
vinokurig: Javadoc?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -75,45 +80,129 @@ public ComparePresenter(AppContext appContext,
...
+     */
+    private void showCompareForCurrentFile() {
+        view.setEnableNextDiffButton(currentFileIndex != (alteredFiles.getFilesQuantity() - 1));
+        view.setEnablePreviousDiffButton(currentFileIndex != 0);
+
+        alteredFiles.getProject()
+                    .getFile(alteredFiles.getFileByIndex(currentFileIndex))
+                    .then(file -> {
+                        if (file.isPresent()) {
+
```
vinokurig: Please remove empty line

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -75,45 +80,129 @@ public ComparePresenter(AppContext appContext,
...
+                            view.setEnableSaveChangesButton(true);
+
+                            final Container gitDir = getGitDir(file.get());
+                            if (gitDir == null) {
+                                notificationManager.notify(locale.messageFileIsNotUnderGit(file.toString()), WARNING, EMERGE_MODE);
+                                return;
+                            }
+                            final Path relPath = getRelPath(gitDir, file.get());
```
vinokurig: Please move this variable initialization closer to it's usage

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -75,45 +80,129 @@ public ComparePresenter(AppContext appContext,
...
+
```
vinokurig: Please remove empty line

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
mmorhun: Yes, it will be better here, thx

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
mmorhun: see no big difference here, but ok, let be so

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -75,45 +80,129 @@ public ComparePresenter(AppContext appContext,
...
+                                showCompareWithLatestForFile(gitDir.getLocation(),
+                                                             relPath,
+                                                             alteredFiles.getStatusByIndex(currentFileIndex));
+                            } else {
+                                showCompareBetweenRevisionsForFile(gitDir.getLocation(),
+                                                                   relPath,
+                                                                   alteredFiles.getStatusByIndex(currentFileIndex));
+                            }
+
+                        } else { // file is deleted
```
vinokurig: Please place this comment directly above the command

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -75,45 +80,129 @@ public ComparePresenter(AppContext appContext,
...
-        final Container rootProject = GitUtil.getRootProject(file);
+                            if (compareWithLatest) {
+                                this.comparedFile = null;
 
-        if (rootProject == null) {
+                                showCompareWithLatestForFile(gitDir, relPath, alteredFiles.getStatusByIndex(currentFileIndex));
+                            } else {
+                                showCompareBetweenRevisionsForFile(gitDir, relPath, alteredFiles.getStatusByIndex(currentFileIndex));
+                            }
+
```
vinokurig: Please remove empty line

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
mmorhun: Do we need javadoc for obviously methods?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
mmorhun: I think that it is already clear form the method name.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
vinokurig: Why not? if you added a javadoc for one method why don't you add it for others

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
mmorhun: To me it is better to keep this comment here, because it is related to whole `else ` branch

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
```diff
@@ -0,0 +1,116 @@
...
+     * @param diff
+     *         plain result of git diff operation
+     */
+    public AlteredFiles(Project project, String diff) {
+        this.project = project;
+
+        alteredFilesStatuses = new LinkedHashMap<>();
+        if (!isNullOrEmpty(diff)) {
+            for (String item : diff.split("\n")) {
+                alteredFilesStatuses.put(item.substring(2, item.length()), defineStatus(item.substring(0, 1)));
```
tolusha: Validate the length of the item.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -214,4 +308,60 @@ private void showCompare(final String remoteContent) {
...
         });
     }
+
+    /**
+     * Searches for given file in the changes files list and save it sequential number to class field.
+     *
+     * @param currentFile
+     *         name of file to set up as current; if null or invalid, the first one will be chosen.
+     */
+    private void findCurrentFile(@Nullable String currentFile) {
```
tolusha: `private int revealFileIndex(..)`

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -75,45 +80,125 @@ public ComparePresenter(AppContext appContext,
...
+                                            @Nullable final String revisionB) {
+        this.alteredFiles = alteredFiles;
+        this.revisionA = revisionA;
+        this.revisionB = revisionB;
+
+        this.compareWithLatest = false;
 
-        final Container rootProject = GitUtil.getRootProject(file);
+        findCurrentFile(currentFile);
+        showCompareForCurrentFile();
```
tolusha: If `(currentFile == null)` will it breaks `showCompareForCurrentFile` method?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
tolusha: `Root` means not any

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -75,45 +80,125 @@ public ComparePresenter(AppContext appContext,
...
+        this.alteredFiles = alteredFiles;
         this.revision = revision;
+
         this.compareWithLatest = true;
 
-        if (status.equals(ADDED)) {
-            showCompare("");
-            return;
-        }
+        findCurrentFile(currentFile);
```
tolusha: What do you mean by `currentFile` ? Is this a selected file?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -75,45 +80,125 @@ public ComparePresenter(AppContext appContext,
...
-        final Container rootProject = GitUtil.getRootProject(file);
+        findCurrentFile(currentFile);
+        showCompareForCurrentFile();
+    }
+
+    /**
+     * Shows comparison for selected file.
+     * Type of comparison to show depends on {@code compareWithLatest} field.
+     */
+    private void showCompareForCurrentFile() {
```
tolusha: Could we simplify the method?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
mmorhun: Ok, but usually we obtain raw diff from git, so it should be valid.

tolusha: Trust no one!

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
mmorhun: @tolusha maybe `findFileIndexOrFirst(..)` will be better?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
mmorhun: I thought about it. On the one hand it would be good to generalize `gitDir` and `relPath` and invoke `showCompare*` form a single place, but on the other hand I tired to leave some 'frame' for subrepository support which will be dropped by generalization.

bmicklea:
@slemeur has spec'ed out the Git interactions so I'm adding him as a reviewer and removing myself (since he will check Changelog, release notes and other PM activities).

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
```diff
@@ -47,7 +48,7 @@ public AlteredFiles(Project project, String diff) {
         alteredFilesStatuses = new LinkedHashMap<>();
         if (!isNullOrEmpty(diff)) {
             for (String item : diff.split("\n")) {
-                if (item.length() < 3 || item.charAt(1) != ' ') {
+                if (item.length() < 3 || item.charAt(1) != '\t') {
                     throw new IllegalArgumentException("Invalid git diff format.");
```
tolusha: If exception is thrown then it has to be caught properly. 
I am thinking about showing error with an appropriate message and skipping invalid `item`

tolusha:
What hotkeys can developer use?
Next/Prev diif?
Next/Prev file?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
```diff
@@ -0,0 +1,120 @@
...
+     * @param project
+     *         the project under diff operation
+     * @param diff
+     *         plain result of git diff operation
+     */
+    public AlteredFiles(Project project, String diff) {
+        this.project = project;
+
+        alteredFilesStatuses = new LinkedHashMap<>();
+        if (!isNullOrEmpty(diff)) {
```
tolusha: Redundant check

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
```diff
@@ -0,0 +1,120 @@
...
+            for (String item : diff.split("\n")) {
+                if (item.length() < 3 || item.charAt(1) != '\t') {
+                    throw new IllegalArgumentException("Invalid git diff format.");
```
tolusha: LOG.warn with invalid item and skip it.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -181,30 +248,57 @@ public void showCompareBetweenRevisions(final Path file,
...
+                    saveContent(content);
+                }
+            });
+        }
+    }
+
+    @Override
+    public void onNextDiffClicked() {
+        view.getEditableContent(content -> {
+            if (isEdited(content)) {
```
tolusha: Is `if (compareWithLatest) {` check required?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -181,30 +248,57 @@ public void showCompareBetweenRevisions(final Path file,
...
+
+            currentFileIndex++;
+            showCompareForCurrentFile();
+        });
+    }
+
+    @Override
+    public void onPreviousDiffClicked() {
+        view.getEditableContent(content -> {
+            if (isEdited(content)) {
```
tolusha: Is if (compareWithLatest) { check required?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -214,4 +308,60 @@ private void showCompare(final String remoteContent) {
...
+
+        currentFileIndex = alteredFiles.getAlteredFilesList().indexOf(currentFile);
+        if (currentFileIndex == -1) {
+            currentFileIndex = 0;
+        }
+    }
+
+    /** Returns true if user edited content in the compare widget i.e. initial and current isn't equal. */
+    private boolean isEdited(final String newContent) {
+        return compareWithLatest && comparedFile != null && this.localContent != null && !newContent.equals(localContent);
```
tolusha: Don't check `compareWithLatest` here, has nothing to do with edited file.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -214,4 +308,60 @@ private void showCompare(final String remoteContent) {
...
+        }
+
+        currentFileIndex = alteredFiles.getAlteredFilesList().indexOf(currentFile);
+        if (currentFileIndex == -1) {
+            currentFileIndex = 0;
+        }
+    }
+
+    /** Returns true if user edited content in the compare widget i.e. initial and current isn't equal. */
+    private boolean isEdited(final String newContent) {
```
vinokurig: `isDirty` ?

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
tolusha: FYI: `git apply` command fails if file not found and applies patch if not a critical error found.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
mmorhun: I think we should throw an exception here. It will be caught and a user will see it in the events section.
To me it will be better to say that something went wrong than show partial diff.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
mmorhun: Yeap, thx

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
mmorhun: See my previous comment

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
tolusha: I don't mind.

plugins/plugin-git/che-plugin-git-ext-git/src/test/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFilesTest.java
```diff
@@ -0,0 +1,121 @@
...
+    @DataProvider(name = "invalidDiffFileData")
+    private Object[][] getInvalidDiffFileData() {
+        return new Object[][] {
+                {"M "},
+                {"M_Test.java"},
+                {"Test.java"},
+                {" "}
+        };
+    }
+}
```
vinokurig: new line

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
mmorhun: Yes this is required. In case when the widget shows diff between two revisions a user cannot edit any content, so in case of comparison with not latest code no changes is supposed.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
mmorhun: Yes, because of autosave

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
mmorhun: I think this is the very place. I'll rename the method

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
mmorhun: Agree, thx

mmorhun:
@tolusha about hotkeys, please add comment [here](https://github.com/eclipse/che/issues/5925).

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -75,45 +77,106 @@ public ComparePresenter(AppContext appContext,
...
+        this.alteredFiles = alteredFiles;
         this.revision = revision;
+
         this.compareWithLatest = true;
 
-        if (status.equals(ADDED)) {
-            showCompare("");
-            return;
-        }
+        findFileIndexOrFirst(currentFile);
```
tolusha: `revealFileIndex`

tolusha: If method finds index why it doesn't return it?

mmorhun: I think the method name is clear. It says what it does.

mmorhun:
@slemeur can you take a look?

tolusha (approved):

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/ComparePresenter.java
```diff
@@ -75,45 +77,106 @@ public ComparePresenter(AppContext appContext,
...
      *         hash of revision or branch
      */
-    public void showCompareWithLatest(final File file, final Status status, final String revision) {
-        this.comparedFile = file;
+    public void showCompareWithLatest(final AlteredFiles alteredFiles,
+                                      @Nullable final String currentFile,
+                                      final String revision) {
+        this.alteredFiles = alteredFiles;
         this.revision = revision;
+
```
vinokurig: Please remove empty line

mmorhun: This line is a separator

vinokurig: I prefer to see initialization of all variables in constructor without separation, even if the variable is initialized not from constructor parameters.

mmorhun: I am not sure I understood you... This is not constructor this method is just kind of entrypoint which stores some data in presenter fields.

vinokurig: You arre right, this is not a constructor :slightly_smiling_face: this is a method, but anyway, why do you separate block of variable initialization?

mmorhun: Because the first block is set of fields which holds data which is provided by user and the second one is the field which is related to internal logic of the presenter.

plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/compare/AlteredFiles.java
mmorhun: Well, I think the current name is better because this class also parses raw diff.

vinokurig (approved):

vparfonov (approved):

slemeur:
@mmorhun : Could you please record an animated gif to demonstrate the features? I'd like to include it in the future release note. (Don't forget to zoom in with the browser, so that the video is readable)

mmorhun:
@slemeur I updated PR description

slemeur:
Few feedbacks: 
1 - I agree with @tolusha that it's important to map the hotkeys in that wizard, at the same time we will introducing it. 

2 - In term of UX, I'm a little bit confused by the 5 buttons at the the bottom of the wizard:
- It seems not obvious what `Previous` and `Next` actions are allowing. I think the only way for the user to understand that, would be to click on those. As a user, I would try to understand to what `Previous/Next` actions are related to. I'll look for the title of the wizard, but here we use the filename.

3 - The `refresh` button is unclear, but why would I need to refresh? (refresh local changes or from remote?)

4 - Editing while reviewing: I don't think it's clear to the user that he could edit the changes while reviewing them. It's a modal popup opened while user is reviewing his changes changes. Could the `Save Changes` be displayed only if there are changes. 

5 - What happens if the user is editing a file and move to `previous`/ `next` without clicking on `Save Changes` ?

6 - When the user has edited the file - can he discards the changes?

As we have the git panel upcoming, I don't we should not invest too much in that wizard which is creating different editing behaviors and different file navigation experience. My proposals would be the following:
1. The title should be more explicit: "Review and edit changes in: {filename} (1/17)" we put the (1/17) to indicate that there are multiple files which can be reviewed and the currently opened one is the 1st of 17. 
2. We move `Next` / `Previous` to the left. When the file has edits, we disable `Next` / `Previous` and reveal "Save Changes" + "Discards" on the left as well.


I'll just add that the review of the changes will be much more logical and convenient from the upcoming git panel (#5128) as it will not need to create any new UX and reuse only what already exist from a file tree explorer (to navigate from file to file) and a consistent editor behavior (save mecanism, logical ability to edit the file). User will not have to understand or learn anything new.

mmorhun:
1. Yes it will be useful, but we have separate issue for hotkeys (https://github.com/eclipse/che/issues/5925)
2. > As a user, I would try to understand to what Previous/Next actions are related to.

     Well, a user opened Git Diff window, I think it is clear what `Next` and `Previous` mean.

3. This button was already present before my changes, I just did not touch it. (This was added by @vinokurig because orion editor doesn't support on fly update in diff view).
4. For now `Save changes` is enabled when user can edit and save left part. For example, for deleted file this button is disabled (see screencast).
5. The content will be saved.
6. Yes, he can. Just click `Close` button and then `No` in popup.
6.1 > The title should be more explicit: "Review and edit changes in: {filename} (1/17)"
  What if we cannot edit changes for this diff? It is possible to view diff between two revision, so a user cannot modify anything. But I like the idea about (1/17). Maybe "Diff for: {filename} (1/17)"  
6.2 Under reveal you meant enable or show?

@slemeur my proposal here is to merge this PR as is and create separate issue for improvement(and take it into next sprint) because here we have too much things together: refactor of git changes files list / diff window, new feature(the issue itself) and bug fix (compare with deleted files didn't work).

tolusha:
`1.` I stick for adding hotkeys here to have fully functional feature.
`4.` It is obvious that user can edit something in the editor, so nothing wrong here.
+1 to disable/enable `Save` button depending on changes.
`6.` `Ctrl+Z` also works.

+1 for @slemeur proposals to update title and rearrange buttons

slemeur:
@mmorhun:

I don't think we should merge the pull request until we have the following:

**1. Allow the keyboard navigation in the wizard**
- Support tab to switch focus on the form's elements
- I believe `left`, `right`, `up`, `down` arrows should be reserved to the editors
- Proposal `ctrl+left: Previous` and `ctrl+right: Next` 

 **2. Rearrange the buttons and title**
- Display: "Review diff in: {filename} (1/17)" 
- Move Next / Previous to the left. 

For the behaviour with "Save Changes" + "Discards". You explained that catching the changes in the editor might be tricky and take time. So I don't think we should invest more time for that behavior - that's something we already have in place with the default editors, so we should reuse that instead of re-creating a new UI/UX which will not be consistent for the end-user. This problem will be solved when the git panel epic #5128 will be implemented.

mmorhun:
@slemeur makes sense, agree.
But about hotkeys for `Next`/`Previous` diff. `ctrl+left` and `ctrl+right` is already used in editor for jumping between words. Do you mind to use `alt+left` and `alt+right` for jumping between diffs?

azatsarynnyy (approved):

mmorhun:
@vparfonov please review new changes

mmorhun:
@slemeur done.
I used `alt` + `,` / `.` (i.e. `alt` + `<`/`>` without `shift`) because `alt` + `left`/`right` are already used in the editor too.

slemeur:
@mmorhun : Great ! Where are the docs ;) ?

mmorhun:
@slemeur well, I wanted to add them ;) , but I failed to find appropriate place. We have some docs in `Git and SVN` section which mainly about how to use different types of credentials for VCSs. We do not have a page or a section about using git functionality like change branch, commit, push, pull etc. Most of them are intuitively, but some details for git diff widget, like autosave on `next`/`prev` or hotkeys combinations, probably will be useful.
So, should I create a section for git diff widget or just leave it as is?

mmorhun:
Maybe we should create separate page in the docs for using git functionality in CHE to not to mess credentials settings and just IDE details/tips.

slemeur:
You should move forward and add the information about Git Diff in the user documentation - the page you found is fine.

I agree with you that the content of the page `Git and SVN` needs to be revisited and re-organized:
- The Git Using OAuth section is more pointing the configuration of the system. So I don't think we should have that in the user documentation, but in the [Configuration](https://www.eclipse.org/che/docs/setup/configuration/index.html) (which BTW is already listed https://www.eclipse.org/che/docs/setup/configuration/index.html#oauth).
- We should probably re-organize the different sections, so that the explanation on the features are first and then we have explanations how to use it with SSH. 
- The SVN documentation will need to be moved (see #6144)

Let's start on adding the docs for this PR. If you are able to start to re-organize the content, that would be perfect, but I'll help you on that.

Thanks Mykola.

ide/che-core-ide-ui/src/main/java/org/eclipse/che/ide/ui/window/Window.java
```diff
@@ -243,6 +256,17 @@ public void execute() {
...
   }
 
+  private void addButtonAlignment(Button button, ButtonAlignment alignment) {
+    switch (alignment) {
+      case LEFT:
+        button.addStyleName(resources.windowCss().buttonAlignLeft());
+        break;
+      case RIGHT:
+        button.addStyleName(resources.windowCss().buttonAlignRight());
+        break;
```
vparfonov: add default statement or replace with if..else

slemeur (approved):

PR #5971, opened 2017-08-11
Adding sources maven artifact for dashboard-war

sunix:
### What does this PR do?
Adding sources maven artifact for dashboard-war

#### Changelog

Adding sources maven artifact for dashboard-war

#### Release Notes

Adding sources maven artifact for dashboard-war

### DIFF
--- a/dashboard/pom.xml
+++ b/dashboard/pom.xml
@@ -72,31 +72,33 @@
                 <artifactId>build-helper-maven-plugin</artifactId>
                 <executions>
                     <execution>
-                      <id>add-resource</id>
-                      <phase>generate-resources</phase>
-                      <goals>
-                        <goal>add-resource</goal>
-                      </goals>
-                      <configuration>
-                        <resources>
-                          <resource>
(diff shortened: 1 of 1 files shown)


dashboard/pom.xml
```diff
@@ -67,6 +67,37 @@
...
             </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>build-helper-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                      <id>add-resource</id>
+                      <phase>generate-resources</phase>
+                      <goals>
+                        <goal>add-resource</goal>
```
benoitf: AFAIK here you're adding sources as resources 
while we should have a separated artifact including the sources.
Else we will deploy a war file that will include the source code while we're only requiring final js code

so IMHO it's more about add-source goal and using maven-source-plugin

sunix: ok I see will fix it

benoitf (approved):

PR #5990, opened 2017-08-14
CHE-5688 Rework resolving of machine servers

sleshchenko:
### What does this PR do?
Reworks resolving of machine servers in the following way:
- matching existing servers to machines is done while workspace starting;
- OpenShiftMachine machine has all servers while instance creation;
- information about servers that are exposed by particular `Route` is stored in its annotations;

This PR also has code refactoring that should make it easier to write unit tests. Earlier it was so difficult to test code that manages OpenShift objects(like pods, services, routes, projects) by using of OpenShiftClient. It's difficult to mock an operation with OpenShift client. Now OpenShiftPods, OpenShiftServices, OpenShiftRoutes, OpenShiftProjects classes are created and they will encapsulate calls to OpenShiftClient.

https://github.com/eclipse/che/issues/5688


infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftProject.java
```diff
@@ -0,0 +1,80 @@
...
+                  .withNewMetadata()
+                  .withName(projectName)
+                  .endMetadata()
+                  .done();
+        } catch (KubernetesClientException e) {
+            throw new InfrastructureException(e.getMessage(), e);
+        }
+    }
+
+    private Optional<Project> get(String projectName, OpenShiftClient client) throws InfrastructureException {
```
garagatyi: Please, consider usage of null instead of  optional since it is private method and you control single place where it is used

garagatyi (commented):
Refactoring looks really good!

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftProject.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    public OpenShiftServices services() {
+        return services;
+    }
+
+    public OpenShiftRoutes routes() {
+        return routes;
+    }
+
```
gazarenkov: Why not to use getter name pattern (getPods(), getServices(), getRoutes()) ? I suppose it is as good practices for public methods unless you have some special reason?

gazarenkov (commented):
I miss JavaDocs on classes with a real business logic. Please FIX it ASAP with meaningful docs (feel free if help needed), it is difficult/time consuming to follow the program logic otherwise

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftProject.java
sleshchenko: `getPods()` looks like method returns pods in a collection. But in fact, it returns the object for managing this kind of objects. Something like that is implemented in fabric8 OpenShiftClient.
So `project.pods().get()` looks much better for me than `project.pods().get()`. I'd leave it without get prefix.

sleshchenko:
I've added TODOs to add java docs and tests. This PR has WIP status, so I'm working on adding tests and docs and they will be added soon.

gazarenkov:
@sleshchenko sure, I saw the TODOs but they do not help much for the program logic understanding )) ... see my request is not a formal one, I have no doubts you will have comments at the end. I need them to better understand and review the code.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftProject.java
gazarenkov: "So project.pods().get() looks much better for me than project.pods().get()" - not sure I understand what is the difference (typo?).

I am not familiar with fabric8 patterns that good. So, please either explain it here or refer to the third party docs (I mean JavaDocs again, in any case)

sleshchenko: Yes. It was typo. `project.pods().get()` against `project.getPods().get()`. 

Kubernetes client provided by Fabric8 doesn't have any docs, but you can take a look source code https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/KubernetesClient.java#L126

sleshchenko:
@gazarenkov I've fixed TODOs about adding java docs. If I missed java docs somewhere please notify me.

garagatyi (approved):

akorneta (approved):

gazarenkov (approved):

skabashnyuk (approved):

Issue #5542, opened 2017-07-05
Some JSON RPC requests are not processed by RequestHandlerManager


RomanNikitenko:
I found that we don't handle JSON-RPC requests that are configured to receive empty result (e.g. `resultAsEmpty()`) in some strange way. When I try to transmit such request it throws an exception so I guess there is an error.

It is another subject of investigation if we really can receive empty responses according to JSON-RPS specifications however I must admit that such ability is quite important in situations where we just want to check if the request is processed properly and no error occurred.

PR #5991, opened 2017-08-14
Some JSON RPC requests are not processed by RequestHandlerManage

dkulieshov:

### What does this PR do?
Fixes exceptions for some requests that expect result as empty elements.

https://github.com/eclipse/che/issues/5542

#### Changelog
Fixed exceptions for requests that expect result with empty elements.

#### Release Notes
Fixed exceptions for requests that expect result with empty elements.

### DIFF
--- a/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/jsonrpc/commons/transmission/SendConfiguratorFromNone.java
+++ b/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/jsonrpc/commons/transmission/SendConfiguratorFromNone.java
@@ -235,11 +235,11 @@ public class SendConfiguratorFromNone {
         endpointId, requestId, Boolean.class, timeInMillis);
   }
 
-  public JsonRpcPromise<Void> sendAndReceiveResultAsEmpty() {
-    return sendAndReceiveResultAsEmpty(0);
+  public JsonRpcPromise<List<Double>> sendAndReceiveResultAsListOfDouble() {
+    return sendAndReceiveResultAsListOfDouble(0);
   }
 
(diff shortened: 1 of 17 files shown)


vparfonov (approved):

musienko-maxim:
Functional tests have passed without regression

RomanNikitenko (approved):

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/jsonrpc/commons/transmission/SendConfiguratorFromNone.java
```diff
@@ -254,10 +254,10 @@ public void sendAndSkipResult() {
             + method
             + ", "
             + "result list items class: "
```
benoitf, line 256: FYI, I see string concat on Logger debug while we should give object references

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/jsonrpc/commons/transmission/SendConfiguratorFromOne.java
```diff
@@ -304,10 +304,10 @@ public void sendAndSkipResult() {
             + pValue
             + ", "
             + "result list items class: "
-            + Void.class);
+            + Double.class);
```
benoitf, line 307: FYI, I see string concat on Logger debug while we should give object references

dkulieshov: This problem came after we had merged client and server code bases related to JSON-RPC because on client we had different logging format rules. There are many places where it is necessary to fix logging format, so I created a separate issue not to mix different kinds of changes - https://github.com/eclipse/che/issues/6119

PR #6003, opened 2017-08-15
Added ability to configure ds with environment variables

skabashnyuk:
### What does this PR do?
Allows to dynamically configure the JDBC data source for simple and multiuser environments. (H2 or Postgres, respectively).
Datasources is now configured via factories that using environment variables for init parameters.

This is standart Tomcat jdbc configuration parameters.
```
 CHE_JDBC_USERNAME
 CHE_JDBC_PASSWORD
 CHE_JDBC_URL
 CHE_JDBC_DRIVER__CLASS__NAME
 CHE_JDBC_MAX__TOTAL
 CHE_JDBC_MAX__IDLE
 CHE_JDBC_MAX__WAIT__MILLIS
```


assembly-multiuser/assembly-ide-war/pom.xml
```diff
@@ -230,7 +230,7 @@
                                 <include>IDE.jsp</include>
                             </includes>
                         </overlay>
-                        <overlay></overlay>
+                        <overlay />
```
sleshchenko, line 233: Why do we need empty overlay here?

skabashnyuk: empty groupId/artifactId represents the current build
see more https://maven.apache.org/plugins/maven-war-plugin/overlays.html

sleshchenko: I see, thanks!

riuvshin (approved):

core/che-core-db/src/main/java/org/eclipse/che/core/db/JNDIDataSourceFactory.java
```diff
@@ -0,0 +1,55 @@
...
+        poolConfigurationProperties.setProperty("maxTotal", maxTotal);
+        poolConfigurationProperties.setProperty("maxIdle", maxIdle);
+        poolConfigurationProperties.setProperty("maxWaitMillis", maxWaitMillis);
+        dataSource = BasicDataSourceFactory.createDataSource(poolConfigurationProperties);
+    }
+
+
+    @Override
+    public Object getObjectInstance(Object obj, Name name, Context nameCtx, Hashtable<?, ?> environment) throws Exception {
+        LOG.info("This={} obj={} name={} Context={} environment={}", this, obj, name, nameCtx, environment);
```
benoitf, line 59: IMHO it's not info level

core/che-core-db/src/main/java/org/eclipse/che/core/db/JNDIDataSourceFactory.java
```diff
@@ -0,0 +1,55 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.naming.Context;
+import javax.naming.Name;
+import javax.naming.spi.ObjectFactory;
+import java.util.Hashtable;
+import java.util.Properties;
+
+public class JNDIDataSourceFactory implements ObjectFactory {
```
benoitf: why not using https://tomcat.apache.org/tomcat-8.0-doc/api/org/apache/tomcat/jdbc/pool/DataSourceFactory.html ?

mshaposhnik (approved):

core/che-core-db/src/main/java/org/eclipse/che/core/db/JNDIDataSourceFactory.java
skabashnyuk: Agree. Will remove that.

core/che-core-db/src/main/java/org/eclipse/che/core/db/JNDIDataSourceFactory.java
skabashnyuk: AFAIK this is an old version of pool that comes from 
```
        <dependency>
            <groupId>org.apache.tomcat</groupId>
            <artifactId>tomcat-jdbc</artifactId>
            <version>${org.apache.tomcat.version}</version>
        </dependency>
```
But I would like to use the second version of pool  https://tomcat.apache.org/tomcat-8.0-doc/jndi-datasource-examples-howto.html#Database_Connection_Pool_(DBCP_2)_Configurations
```
        <dependency>
            <groupId>org.apache.tomcat</groupId>
            <artifactId>tomcat-dbcp</artifactId>
            <scope>provided</scope>
        </dependency>
```
The main reason why do I need custom factory is the way how I would like to configure it. Tomcat's factories are made to be configured with standard J2EE mechanism aka context.xml, but I would like to use environment variables. With @riuvshin  we considered one more option. Like: using system variables to configure DS in context.xml or server.xml. But found that we print all system variables in tomcat startup, and printing all DS configuration including password doesn't look good. So we decided that custom  ObjectFactory configured over environment variables is a suitable wa

benoitf: I understood why you're using JNDI resources.
but for example https://tomcat.apache.org/tomcat-8.0-doc/api/org/apache/tomcat/dbcp/dbcp2/BasicDataSourceFactory.html is already a JNDI factory. So you could directly use that JNDI factory as well.

Also do you have considered https://github.com/brettwooldridge/HikariCP#initialization
https://brettwooldridge.github.io/HikariCP/ because dbcp2 has not the best performances vs c3p0 or hikari ?

skabashnyuk: >So you could directly use that JNDI factory as well.

Directly I can't. But I use it indirectly over ```BasicDataSourceFactory.createDataSource``` method. 

>Also do you have considered ```HikariCP```

Yes. 

Procs
- very tasty on high load. 

Cons
- Our load is not that high
- We have no experience with it.
- We need a CQ for it.
- I wrote a note about it to @gorkem. But looks like it's not that important at this stage

benoitf: ok thanks

benoitf (approved):

core/che-core-db/src/main/java/org/eclipse/che/core/db/JNDIDataSourceFactory.java
gorkem: I agree with @skabashnyuk this is not a fix worth for doing at this time but we may want to open an issue to revisit in the future.

core/che-core-db/src/main/java/org/eclipse/che/core/db/JNDIDataSourceFactory.java
benoitf: https://github.com/eclipse/che/blob/che-multiuser/core/che-core-db/src/main/java/org/eclipse/che/core/db/JNDIDataSourceFactory.java#L59

It has been merged but it's still INFO level

PR #6008, opened 2017-08-16
Fix OpenShiftConnector.inspectImage() to resolve issue in OS 3.6+

amisevsk:
### What does this PR do?

OpenShiftConnector.inspectImage() uses pods to figure out a workspaces docker image. Previously, this was not a problem as it would match the imagestream tag we create earlier. However, in OpenShift 3.6 and higher, it seems like pods created by deployments use the sha256 of the image as an identifier, and so this fails. This means that workspaces fail to start due to imagestream not found.

To resolve this issue, we instead get the image name through the deployment, which does have the correct value.

#### Changelog
(bugfix, update to support OpenShift 3.6+)

#### Release Notes
(bugfix)


amisevsk:
Note that if you are going to be testing it on minishift 1.4.1 + OpenShift 3.6, the modifications in https://github.com/amisevsk/rh-che/tree/minishift-network might be necessary (for the Red Hat distribution).

benoitf:
ci-build

benoitf:
we merged PR done against master branch

PR #6017, opened 2017-08-16
Do not scale down deployment before resource deletion. Removing watch pod logic (seems to be also broken after 3.6 osio update)

ibuziuk:
### What does this PR do?
Do not scale down deployment before resource deletion
`deployment.scale(0, true);` hangs on osio after update to 3.6 version. Watching pods also seem to hang till timeout occurs.

#### Changelog
(bugfix, update to support OpenShift 3.6+)

#### Release Notes
N / A

#### Docs PR
N / A

### DIFF
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftDeploymentCleaner.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftDeploymentCleaner.java
@@ -12,8 +12,6 @@ package org.eclipse.che.plugin.openshift.client;
 
 import java.io.IOException;
 import java.util.List;
-import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.TimeUnit;
 
 import javax.inject.Singleton;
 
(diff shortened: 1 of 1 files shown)


plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftDeploymentCleaner.java
```diff
@@ -41,26 +32,14 @@
 public class OpenShiftDeploymentCleaner {
     private static final Logger LOG = LoggerFactory.getLogger(OpenShiftDeploymentCleaner.class);
     private static final int OPENSHIFT_POD_DELETION_TIMEOUT = 120;
+    private static final int OPENSHIFT_WAIT_POD_DELAY       = 1000;
```
skabashnyuk, line 35: shall we make this parameter configurable?

ibuziuk: this one is hot-fix. We will not this ugly delays in future

skabashnyuk: ок

benoitf (approved):

Issue #6097, opened 2017-08-26
Che deployment on OpenShift should be as simple as on Docker 


l0rd:
The epic is about the few small steps that would make the deployment of Che on OpenShift consistent with the deployment on Docker:

- [x] Add the scripts and doc to run Che on OpenShift https://github.com/redhat-developer/rh-che/issues/284
- [x] Adapting default Che runtime stacks to run on OpenShift as unprivileged containers https://github.com/redhat-developer/rh-che/issues/155
- [x] Don't abort a workspace bootstrap if SSH agent fails to be started (in an unprivileged container) (https://github.com/redhat-developer/rh-che/issues/346)
- [ ] Extend Che CLI to start Che on OpenShift (https://github.com/redhat-developer/rh-che/issues/418)
- [x] Inject CHE_* environment variables when bootstrapping Che on OpenShift
- [x] Adapt Docker image `eclipse/che-server` (alpine based) to run on OpenShift
- [x] Package Che as a [minishift addon](https://github.com/minishift/minishift-addons) (c.f. https://github.com/minishift/minishift/issues/1264#issuecomment-329120040) (https://github.com/redhat-developer/rh-che/issues/353)

kameshsampath:
@l0rd - how can I issue a PR to this item 5 of this epic ? I have built a script for adding the `GITHUB oAuth2`

kameshsampath:
@l0rd , @benoitf  - we shall discuss on how to get item 5 in the epic, to be done via cli - as we kind of agreed that using script based approach is no good Ux

l0rd:
@riuvshin are you working (or worked) on the following subtask?

`- [ ] Inject CHE_* environment variables when bootstrapping Che on OpenShift`

riuvshin:
@l0rd no, but I was working on allowing to pass any configuration set to CHE pod without adding new `sed`s to deploy script: https://github.com/eclipse/che/pull/7692  but note that this is done ONLY in che6 branch.

ibuziuk:
Tests results of running workspaces with stacks available on che6 against minishift ` v1.9.0+a511b25`:

| Stack  | Workspace starts without errors | Commands for quckstart projects work correctly|
| ------------- | ------------- | ------------- |
| .NET   | :white_check_mark: | Commands work but for some reason preview URL is not resolved `${server.dot.net.server} ` |
[...]

PR #6098, opened 2017-08-26
Scripts to deploy Che on OpenShift

l0rd:
### What does this PR do?
Add 3 scripts that help deploying Che on OpenShift:
- `deploy_che.sh`: Creates OpenShift objects and starts wsmaster on OpenShift
- `wait_until_che_is_available.sh`: Monitors Che deployment. Exit when wsmaster is available.
- `replace_stacks.sh`: Replaces default Che runtime stacks with ["OpenShift supported" stacks](https://github.com/redhat-developer/rh-che/blob/master/builds/fabric8-che/assembly/fabric8-stacks/src/main/resources/stacks.json)

Default runtime stacks still need to be replaced because they don't meet [the criteria to run as unprivileged container](https://docs.openshift.org/latest/creating_images/guidelines.html#openshift-origin-specific-guidelines). This PR is a first step of [this epic](https://github.com/eclipse/che/issues/6097). 

To deploying Che is a matter of running the following command:
```bash
./deploy_che.sh && ./wait_until_che_is_available.sh && ./replace_stacks.sh
```

https://github.com/redhat-developer/rh-che/issues/284

#### Changelog
Added instructions to deploy Che on OpenShift

#### Release Notes
Added instructions to deploy Che on OpenShift

#### Docs PR
https://github.com/eclipse/che-docs/pull/280


dockerfiles/cli/scripts/openshift/replace_stacks.sh
```diff
@@ -0,0 +1,58 @@
...
+     stack_id=$(echo "${stack}" | jq '.id')
+     stack_id_no_quotes="${stack_id%\"}"
+     stack_id_no_quotes="${stack_id_no_quotes#\"}"
+     echo -n "[CHE] Adding stack ${stack_id_no_quotes}..."
+     http_code=$(curl -w '%{http_code}' -s --output /dev/null -X POST --header 'Content-Type: application/json' --header 'Accept: application/json' -d "${stack}" "${CHE_API_ENDPOINT}/stack")
+     echo "done (HTTP code ${http_code})"
+done
+
+echo
+echo
```
riuvshin: new line missing

dockerfiles/cli/scripts/openshift/replace_stacks.sh
```diff
@@ -0,0 +1,58 @@
...
+
+if [ -z "${CHE_API_ENDPOINT+x}" ]; then
+    echo -n "[CHE] Variable \$CHE_API_ENDPOINT is not set. Trying to infer it..."
+    che_host=$(oc get route che -o jsonpath='{.spec.host}')
+    if [ -z "${che_host}" ]; then echo >&2 "[CHE] [ERROR] Failed to infer environment variable \$CHE_API_ENDPOINT. Aborting. Please set it and run ${0} script again."; exit 1; fi
+    CHE_API_ENDPOINT="http://${che_host}/api"
+    echo "done (${CHE_API_ENDPOINT})"
+fi
+
+DEFAULT_NEW_STACKS_URL="https://raw.githubusercontent.com/redhat-developer/rh-che/master/builds/fabric8-che/assembly/fabric8-stacks/src/main/resources/stacks.json"
```
riuvshin: maybe we should think how to avoid direct links to github, maybe it will be better to have this on central same as  OS yaml

l0rd: Publishing stacks.json in maven central is a good idea. That could be a mechanisms to publish and share custom stacks for users. 

By the way this script (`replace_stack.sh`) is just a temporary workaround to use the docker images that are "OpenShift compatible". The long term solution should be to make all default docker images "OpenShift compatible" (c.f. point n.2 of   https://github.com/eclipse/che/issues/6097). When we have that we won't need to replace stacks.json anymore.

dockerfiles/cli/scripts/openshift/replace_stacks.sh
l0rd: Thanks. Fixed.

riuvshin:
ci-build

l0rd:
@benoitf @riuvshin is it ok to merge?

dockerfiles/cli/scripts/openshift/deploy_che.sh
```diff
@@ -0,0 +1,319 @@
...
+  DEFAULT_CHE_KEYCLOAK_DISABLED="true"
+  CHE_KEYCLOAK_DISABLED=${CHE_KEYCLOAK_DISABLED:-${DEFAULT_CHE_KEYCLOAK_DISABLED}}
+  DEFAULT_CHE_DEBUGGING_ENABLED="true"
+  CHE_DEBUGGING_ENABLED=${CHE_DEBUGGING_ENABLED:-${DEFAULT_CHE_DEBUGGING_ENABLED}}
+
+elif [ "${OPENSHIFT_FLAVOR}" == "osio" ]; then
+  # ----------------------
+  # Set osio configuration
+  # ----------------------
+  if [ -z "${OPENSHIFT_TOKEN+x}" ]; then echo "[CHE] **ERROR** Env var OPENSHIFT_TOKEN is unset. You need to set it with you OSO token to continue. To retrieve your token: https://console.starter-us-east-2.openshift.com/console/command-line. Aborting"; exit 1; fi
```
AndrienkoAleksandr: Maybe it should be "with your" instead of "with you"

l0rd: Fixed. Thanks @AndrienkoAleksandr!

riuvshin (approved):

benoitf (commented):
@l0rd well for me I dont see now why the scripts are part of the cli docker image because there is no command from the cli to call these commands
It looks like a placeholder but if it's inside cli it needs to work with cli

l0rd:
@benoitf truth is that I didn't know where to put them. But since next step is to make these scripts used by the CLI ([4th point of the epic](https://github.com/eclipse/che/issues/6097#issue-253080141)) it should make sense to have them already there. That at least would avoid to moving them around in a few days.

AndrienkoAleksandr (approved):

benoitf:
why is that bas been merged ?

benoitf:
Because AFAIK I didn't approved it

l0rd:
That's a misunderstanding. After talking yesterday I thought you were ok to
merge this PR now and work on the integration of these scripts within the
CLI in a separate PR.

benoitf:
Yes probably misunderstanding. I was expecting scripts in their final state when merging them but now it's already merged so... ;-)
It's because we often say we will work on improvements after the merge but the improvements are never coming

PR #6107, opened 2017-08-30
allow additional hostname match

jgroom33:
the regex blocks hostnames that are added to /etc/hosts:
172.16.1.25    foo

### What does this PR do?

This allows hostnames of foo in addition to #.#.#.# for adding a VCS git.


skabashnyuk:
ci-build

tolusha:
@jgroom33 
It would be nice to see some new tests concerning new functionality here
https://github.com/eclipse/che/blob/5b373e6cf5bfeb16f278c4d750fe64361445eeb4/plugins/plugin-git/che-plugin-git-ext-git/src/test/java/org/eclipse/che/ide/ext/git/client/importer/page/GitImporterPagePresenterTest.java#L41

PR #6144, opened 2017-09-05
Remove svn plugin from default assembly

### LINKED ISSUE #5484: Deprecate and Remove Subversion support
Let's start the retirement of the subversion support.

skabashnyuk:
### What does this PR do?
Due to the low popularity svn plugin will be removed from default assembly. 

Fixes https://github.com/eclipse/che/issues/5484

#### Changelog
Svn plugin removed from default assembly

#### Release Notes
Svn plugin removed from default assembly

### DIFF
--- a/assembly/assembly-ide-war/pom.xml
+++ b/assembly/assembly-ide-war/pom.xml
@@ -193,10 +193,6 @@
         </dependency>
         <dependency>
             <groupId>org.eclipse.che.plugin</groupId>
-            <artifactId>che-plugin-svn-ext-ide</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.eclipse.che.plugin</groupId>
             <artifactId>che-plugin-testing-ide</artifactId>
         </dependency>
         <dependency>
--- a/assembly/assembly-wsagent-war/pom.xml
+++ b/assembly/assembly-wsagent-war/pom.xml
@@ -105,10 +105,6 @@
         </dependency>
         <dependency>
             <groupId>org.eclipse.che.plugin</groupId>
-            <artifactId>che-plugin-svn-ext-server</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.eclipse.che.plugin</groupId>
             <artifactId>che-plugin-test-ls-server</artifactId>
         </dependency>
         <dependency>


riuvshin (approved):

bmicklea:
@skabashnyuk do you have usage info for SVN plugin at codenvy.io?

skabashnyuk:
@bmicklea no

skabashnyuk:
https://trends.google.com/trends/explore?date=all&q=git,svn does this counts usage info

bmicklea:
We have existing customers of Codenvy who use the SVN plugin and will be migrating to Che. Why remove SVN plugin if we know it will impact those users - what does it gain for us?

riuvshin:
@bmicklea a while ago we all agreed on roadmap that we have to drop svn.
more info in related issue https://github.com/eclipse/che/issues/5484

bmicklea:
Thanks - I missed that issue somehow. So how will a company that uses Che and SVN move forward? Is there anything easier than them creating a custom Che assembly?

slemeur:
-1 to remove it like this.

We put effort to build the SVN plugin, which can still be interesting for user and for the community. We should not put it under a lost land on the repository where the source code is not built anymore. 

This PR clearly points the plugin modularity issues that we have right now. It's a great opportunity to discuss about how this should be. 

**Important note:** This PR introduces functional regressions. It **has** first to be deprecated - this can't be added into the upcoming release without any sort of communication to the community (changelog, release note, documentation).

gorkem:
We spend engineering time for testing, building and the occasional maintenance for SVN. In addition SVN UI elements takes up valuable screen real estate and needs to be handled as we enhance our UX with things like command palette. So having SVN(or any other feature) is not free because initial implementation was done in the past. @bmicklea @slemeur do you have any metrics on the amount of users that will be effected?

bmicklea:
We have a paying enterprise customer using it. There are likely others on codenvy.io but without per feature metrics I don't know how many.

A couple of things:
- removing it from the default assembly doesn't mean it isn't supported so we'd still need to test and maintain it
- this is a breaking change as Stevan points out so we do need a deprecation notice on the RN and clarification on how to add it to a default assembly for on-prem installs.

gorkem:
I agree, we can not just remove the feature without any notice, this PR is early. I actually created #5484 so that we could do the planning around this over there. 
I suggest we abort this one for now and let @bmicklea and @slemeur work on a retirement plan. :) Will assign the original issue over to PMs.

skabashnyuk:
Going to create a new one for Che6 branch.

PR #6147, opened 2017-09-05
Use concrete node version for UD Dockerfile

riuvshin:
### What does this PR do?
Use concrete node version for UD Dockerfile

#### Changelog
Use concrete node version for UD Dockerfile


benoitf (approved):
nice to set the version so we don't have unexpected tool change

PR #6165, opened 2017-09-07
Check UID before executing chown.

### LINKED ISSUE #5560: Can't boot up ws-agent with huge workspace.
**Reproduction Steps:**

0. Place project on NFS volume.
1. Create huge project (e.g. `npm install` after cloning ng2-admin)
2. Restart workspace.

**OS and version:**    

5.15.0-SNAPSHOT c7c1ee807cc2d85f

**Diagnostics:** 

Small size workspaces work fine on the same instance.
Referring `docker exec -it <container-id> ps -ef`, my ws-agent was blocked in `chown -R usr /projects`.
Is it possible to run `chown` in background ?

monaka:
This will fix #5560.

Signed-off-by: Masaki Muranaka <monaka@monami-ya.com>

### What does this PR do?

Improve speed for changing file owner on projects/
In case home directory is mounted via NFS and a lot of files in there,
 connections to workspace is timed out.

#5560

### DIFF
--- a/wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
+++ b/wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
@@ -55,7 +55,7 @@ MACHINE_TYPE=$(uname -m)
 
 mkdir -p ${CHE_DIR}
 ${SUDO} mkdir -p /projects
-${SUDO} sh -c "chown -R $(id -u -n) /projects"
+${SUDO} sh -c "find /projects -not -uid $(id -u) -print0 | xargs -0 -P3 --no-run-if-empty chown $(id -u)"
 ${SUDO} chmod 755 /projects
 
 


wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
```diff
@@ -55,7 +55,7 @@ MACHINE_TYPE=$(uname -m)
 
 mkdir -p ${CHE_DIR}
 ${SUDO} mkdir -p /projects
-${SUDO} sh -c "chown -R $(id -u -n) /projects"
+${SUDO} sh -c "find /projects -not -uid $(id -u -n) -print0 | xargs -0 -P3 --no-run-if-empty chown $(id -u -n )"
```
tolusha:
```
user@ce7b8a5773c0:/projects$ find /projects -not -uid $(id -u -n) -print0
find: invalid argument `user' to `-uid'
user@ce7b8a5773c0:/projects$ echo $?
1
```

tolusha (changes requested):

wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
tolusha: @monaka 
I guess something is missed

ghost: it should be either `-uid $(id -u)` or `-user $(id -u -n)` First option is better I think.

@monaka can you push changes please and will merge this one?

monaka: I see. I'll push again.

monaka:
@tolusha I checked my code and pushed again.

wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
```diff
@@ -55,7 +55,7 @@ MACHINE_TYPE=$(uname -m)
 
 mkdir -p ${CHE_DIR}
 ${SUDO} mkdir -p /projects
-${SUDO} sh -c "chown -R $(id -u -n) /projects"
+${SUDO} sh -c "find /projects -not -uid $(id -u) -print0 | xargs -0 -P3 --no-run-if-empty chown $(id -u)"
```
garagatyi, line 58: Looks like `find -uid` and `xargs -P3` are not POSIX compliant and not available in certain Linux distributions, such as alpine. 
@skabashnyuk @tolusha Do you think guys it's ok or should we explore this?

tolusha: on alphine 
```
/ # sh -c "find /projects -not -uid $(id -u) -print0 | xargs -0 -P3 --no-run-if-empty chown $(id -u)"
xargs: unrecognized option: P
find: unrecognized: -uid
BusyBox v1.26.2 (2017-06-11 06:38:32 GMT) multi-call bina
```

monaka: Hmm... I reject it by myself and find the another way...

Issue #5810, opened 2017-07-26
Creating a file/folder with '%' in the name returns 400 Bad Request


ghost:
_From @balzer814 on July 25, 2017 15:3_

When adding a file/folder with '%' in the name codenvy returns 400 Bad Request.
It is possible to create a folder with such a name using the terminal, but this folder can later not be opened in the ide.
When creating such a file over the terminal and than open it in the IDE the opened Editor only says 'An error occoured while loading the file.'

**Reproduction Steps:**

1. open new...-> folder dialog
2. type in '%name' as directory name
3. click ok
4. codenvy shows an error message:
![image](https://user-images.githubusercontent.com/23092994/28578382-e436d2d6-7159-11e7-860b-fa1e27455394.png)

**Codenvy version:**    
5.15.0

_Copied from original issue: codenvy/codenvy#2337_

koppor:
This is necessary for [Eclipse Winery](http://www.eclipse.org/winery/) to be able to code files in the browser.

Example folder: https://github.com/winery/test-repository/tree/black/nodetypes/http%253A%252F%252Fwinery.opentosca.org%252Ftest%252Fponyuniverse/unicorn

We use double encoding, because of security counter-measurements at Web servers. See http://eclipse.github.io/winery/dev/Encoding for more links.

balzer814:
This issue also exists with already existing folders that are cloned into che from a git repository

Example:

1. create Git repository "issue5810Repo" on GitHub
2. add folder "problematic%folder"
3. add File "unreachable.txt"
4. import the repo in che using the factory link http://codenvy.io/f?url=https://github.com/balzer814/issue5810Repo
5. after importing the folder can't be opened, deleted or renamed and "unreachable.txt" can't be accessed

slemeur:
thanks @ashumilova !

AndrienkoAleksandr:
QA cycle https://github.com/eclipse/che/tree/CHE-5810-3.

mmorhun:
Merged.

PR #6166, opened 2017-09-07
CHE-5810: Improve client encoding for file or folder path to support special symbols

AndrienkoAleksandr:
### What does this PR do?
Improve client encoding for file or folder path to support special symbols:  "%", "[", "]", "#", ";".

https://github.com/eclipse/che/issues/5810

#### Changelog
Improve client encoding for file or folder path to support special symbols:  "%", "[", "]", "#", ";".

#### Release Notes
Improve client encoding for file or folder path to support special symbols:  "%", "[", "]", "#", ";".

#### Docs PR
Don't need bug Fix.


azatsarynnyy (approved):

ide/commons-gwt/src/main/java/org/eclipse/che/ide/resource/Path.java
```diff
@@ -1143,4 +1176,24 @@ public static Path commonPath(Path... paths) {
 
     return commonPath;
   }
+
+  /** Returns path encoded by segments without device. */
+  public String getEncodedPath() {
```
vzhukovs: Is it tested?

Create separate utility class called `PathEncoder` and move this method as static there. See how `java.net.URLEncoder#encode(java.lang.String, java.lang.String)` method done.

This is not a right place for this method, it should be a static utility method.

vzhukovs (changes requested):

ide/commons-gwt/src/main/java/org/eclipse/che/ide/resource/Path.java
AndrienkoAleksandr: Yes it was tested.

tolusha (approved):

vparfonov (approved):

ide/commons-gwt/src/main/java/org/eclipse/che/ide/util/PathEncoder.java
```diff
@@ -0,0 +1,47 @@
...
+import static com.google.gwt.http.client.URL.encodePathSegment;
+import static org.eclipse.che.ide.resource.Path.SEPARATOR;
+
+import org.eclipse.che.ide.resource.Path;
+
+/**
+ * @author Alexander Andrienko
+ * @author Mykola Morhun
+ */
+public class PathEncoder {
```
tolusha, line 22: Add several tests also.

vzhukovs (approved):

ide/commons-gwt/src/main/java/org/eclipse/che/ide/util/PathEncoder.java
mmorhun: Well, we have only one method to test and it uses native javascript method. I think it won't worst to try to deal with it in tests.

mmorhun: But I tried...

tolusha: It is pure java code. Why do you need js?

mmorhun: Look deeper. `encodeComponent` method uses native js method under the hood.

PR #6171, opened 2017-09-07
Fix gulp test task execution

ashumilova:

### What does this PR do?
Fixes the gulp test task execution.

#### Docs PR
N/A

### DIFF
--- a/dashboard/Dockerfile
+++ b/dashboard/Dockerfile
@@ -8,15 +8,17 @@
 # Build step: $ docker build -t eclipse-che-dashboard
 # It builds an archive file that can be used by doing later
 #  $ docker run --rm eclipse-che-dashboard | tar -C target/ -zxf -
-FROM mhart/alpine-node:6.11.2
+FROM node:6.11.2
 
-RUN apk update && \
-    apk add --no-cache git
+RUN apt-get update && \
+    apt-get install -y git \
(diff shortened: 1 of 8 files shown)


benoitf:
does it need too be activated there as well ?
https://github.com/eclipse/che/blob/master/dashboard/Dockerfile#L20

benoitf (approved):

dashboard/Dockerfile
```diff
@@ -8,15 +8,15 @@
 # Build step: $ docker build -t eclipse-che-dashboard
 # It builds an archive file that can be used by doing later
 #  $ docker run --rm eclipse-che-dashboard | tar -C target/ -zxf -
-FROM mhart/alpine-node:6.11.2
+FROM node:6.11.2
 
-RUN apk update && \
-    apk add --no-cache git
+RUN apt-get update && \
+    apt-get install -y git
```
benoitf: we should add
```
    && apt-get -y clean \
    && rm -rf /var/lib/apt/lists/* && \
```

olexii4 (approved):

dashboard/Dockerfile
benoitf: as now it's switched to debian

akurinnoy (approved):

Issue #6169, opened 2017-09-07
Workspace Agent is slow to boot on macOs


benoitf:
Running the workspace agent is quite slow, especially when I/O is not so fast.
Which is the case on my environment : macOs / docker

**Reproduction Steps:**
Start Eclipse Che workspace, and look at the time used by the workspace agent to boot.

Running profiler shows that it can run faster.

- Some libraries of Tomcat could be removed
- All jars (from WEB-INF/lib) included in the war file are scanned
- All Guice Singletons are initialized at startup but some of them have very costly constructor time. We could use lazy initialization.

PRs:
- [x] Get rid of JSP in assemblies so we can remove JSP stuff from Tomcat https://github.com/eclipse/che/pull/6174
- [x] Lazy init of maven local repository plugin maven server https://github.com/eclipse/che/pull/6178
- [x] Lazy init of tika instance in project service https://github.com/eclipse/che/pull/6179
- [x] workspace agent : avoid scan of WEB-INF/lib jars for classes #6186

garagatyi:
Can you elaborate on ways to make it faster because it is not clear from the description to me

benoitf:
@garagatyi I'm updating the issue. Some PR will be linked

skabashnyuk:
Can we see numbers of perfomance report before we start applaing some changes?

benoitf:
@skabashnyuk On my computer, on docker or minishift the start time of the agent is reduced by 5 seconds (15 seconds before to less than 10 seconds after)

skabashnyuk:
Perfect. 
What is the timings for container start, individual components?
What are the results on different OS OSX vs Linux?

benoitf:
@skabashnyuk I suppose your comment is a little bit odd. In current design, code is eating memory, cpu and I/O and what my PR are doing is that it **reduces memory, cpu and I/O** without changing the logic.
So there is no macOS vs Linux vs Windows. 
Also reducing I/O is good when you've several of workspaces starting on the same physical machine. (and memory is good as well)

skabashnyuk:
@benoitf  I just want to understand what is the price and what is the profit.

benoitf:
@skabashnyuk I see only profit ?

skabashnyuk:
[...]

PR #6174, opened 2017-09-07
There is no Java code inside the IDE.JSP so there is no need to have all JSP stuff

benoitf:
### What does this PR do?
Rename IDE.jsp to IDE.html as we only need to serve the html page.
so we can get rid of all the JSP libraries.

https://github.com/eclipse/che/issues/6169

#### Changelog
Avoid to use JSP for IDE html page that has no java code.

#### Release Notes
N/A

#### Docs PR
N/A

Change-Id: I0c881fc37aedcc22f09fbbf16e49ce43f2dea334

### DIFF
--- a/assembly/assembly-ide-war/src/main/java/org/eclipse/che/IDEController.java
+++ b/assembly/assembly-ide-war/src/main/java/org/eclipse/che/IDEController.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright (c) 2012-2017 Red Hat, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che;
+
+import java.io.IOException;
+import javax.servlet.RequestDispatcher;
+import javax.servlet.ServletException;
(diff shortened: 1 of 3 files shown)


skabashnyuk:
Why do we need IDEController?

benoitf:
@skabashnyuk it's to fulfil this pattern https://github.com/eclipse/che/blob/c858b65b70ffa2686c5084cab8cf2ec25bd3ac08/assembly/assembly-ide-war/src/main/webapp/WEB-INF/web.xml#L63-L66

skabashnyuk:
What blocks us from remove servlet at all and use static html from war file?

skabashnyuk (approved):

benoitf:
@skabashnyuk I didn't wanted to be intrusive on this pull request. (I only replace JSP by a servlet)
Maybe we can remove the servlet and use a filter to do the redirect but I thought it was more a deep change on the logic.
But I agree that removing the servlet at all would be a nice improvement.

benoitf:
@vparfonov @riuvshin is it ok for you ? you want to wait like on monday ?

vparfonov (approved):

riuvshin (approved):

gazarenkov:
@benoitf 
since you edit web.xml
maybe you can also remove 

 ```
<security-role>
        <description>the user role</description>
        <role-name>developer</role-name>
  </security-role>
```

pretty sure it is not used anymore no?

benoitf:
@gazarenkov yes when I've edited file I was thinking the same. I will issue another pull request as I don't like mixing stuff in PR

PR #6179, opened 2017-09-07
Project service: Lazy init of Tika

benoitf:
### What does this PR do?
Lazy init of Tika.
Instead of initializing when constructor is called, initialize it at first call.

#6169 

#### Changelog
lazily init the tika instance

#### Release Notes
N/A

#### Docs PR
N/A

Change-Id: I06872806f8a4c334614e1e15a12b5c533f5d8c11

### DIFF
--- a/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
+++ b/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/ProjectService.java
@@ -113,7 +113,7 @@ import org.slf4j.LoggerFactory;
 @Singleton
 public class ProjectService extends Service {
   private static final Logger LOG = LoggerFactory.getLogger(ProjectService.class);
-  private static final Tika TIKA = new Tika();
+  private static Tika TIKA;
 
   private final ProjectManager projectManager;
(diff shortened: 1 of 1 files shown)


skabashnyuk:
Can we do some sort of provider or so? Or hide initialization logic inside some class instead of doing it in REST service?

benoitf:
@skabashnyuk I fixed the issue, what you're asking is a refactoring on how ProjectService was designed ? AFAIK these changes are handled by maintainers not by contributors.

garagatyi:
@benoitf Can you elaborate on why the kind label is set to bug?

benoitf:
@garagatyi you think it's an enhancement ?
For me it's a bug as creating an instance of an object should be always fast.

garagatyi:
I do believe that marking everything as a bug is not convenient since it tempts ppl to mark everything as a bug to put up the priority and attention of other ppl in the project

benoitf:
I've updated the label to enhancement if it's more convenient for people.

garagatyi:
thanks

vparfonov (approved):

skabashnyuk (approved):

PR #6185, opened 2017-09-07
CHE-6085: Add posibility to choose WS infra impl

### LINKED ISSUE #6085: OpenShift infrastructure doesn't work on Minishift because of dependency on Docker runtime
Related to SPI branch only.
In SPI branch both OpenShift and Docker infrastructures are working. 
On MiniShift Docker infra tries to reach Docker API and fails. 
This prevents us from testing OpenShift infrastructure on MiniShift.
So, we need some way to Disable Docker infrastructure in this case. Probably, it should be done by some property.
Note that it is supposed to be a temporary solution until we realize the proper way to fix it.

garagatyi:
### What does this PR do?
Environment variable CHE_INFRASTRUCTURE_ACTIVE sets which
implementation of workspace runtime infrastructure is used in Che.
Notice that corresponding property does not work - only env var.

Fixes #6085 

#### Changelog
Allow choosing workspace runtime infrastructure implementation with an environment variable.

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterModule.java
+++ b/assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterModule.java
@@ -41,14 +41,14 @@ import org.eclipse.che.api.workspace.server.RemoveWorkspaceFilesAfterRemoveWorks
 import org.eclipse.che.api.workspace.server.adapter.StackMessageBodyAdapter;
 import org.eclipse.che.api.workspace.server.adapter.WorkspaceConfigMessageBodyAdapter;
 import org.eclipse.che.api.workspace.server.adapter.WorkspaceMessageBodyAdapter;
+import org.eclipse.che.api.workspace.server.hc.ServerCheckerFactory;
+import org.eclipse.che.api.workspace.server.hc.ServerCheckerFactoryImpl;
 import org.eclipse.che.api.workspace.server.stack.StackLoader;
 import org.eclipse.che.core.db.schema.SchemaInitializer;
 import org.eclipse.che.inject.DynaModule;
 import org.eclipse.che.plugin.github.factory.resolver.GithubFactoryParametersResolver;
 import org.eclipse.che.workspace.infrastructure.docker.DockerInfraModule;
 import org.eclipse.che.workspace.infrastructure.docker.local.LocalDockerModule;
-import org.eclipse.che.workspace.infrastructure.docker.snapshot.JpaSnapshotDao;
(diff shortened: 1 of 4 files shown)


skabashnyuk:
```CHE_INFRASTRUCTURE_ACTIVE -> CHE_ACTIVE_INFRASTRUCTURE```

**Review by skabashnyuk — APPROVED** [review]

**Review by sleshchenko — APPROVED** [review]
+1 to rename as Sergii suggested.

**Review by akorneta — APPROVED** [review]

**Comment by garagatyi** [comment]
underscore reflects dot:
CHE_INFRASTRUCTURE_ACTIVE -> che.infrastructure.active
What you suggest, I believe, is not consistent with our current strategy of property/env vars naming.

**Comment by sleshchenko** [comment]
I see. Thanks. Did you investigate using configuration property, It would make this option more clear, I mean user would be able to find information about it in che.properties or che.env,

File: assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterModule.java [hunk 1]
```diff
@@ -184,10 +179,14 @@ protected void configure() {
     //        bind(org.eclipse.che.api.agent.server.filters.AddExecInstallerInWorkspaceFilter.class);
     //        bind(org.eclipse.che.api.agent.server.filters.AddExecInstallerInStackFilter.class);
 
-    // FIXME: spi
-    install(new DockerInfraModule());
-    install(new LocalDockerModule());
-    install(new OpenShiftInfraModule());
+    String infrastructure = System.getenv("CHE_INFRASTRUCTURE_ACTIVE");
+    if ("openshift".equals(infrastructure)) {
```
riuvshin, line 183: maybe we should think how to made that part more friendly for new types of infra... just in case if we will decide to port swarm.

garagatyi:
@sleshchenko Since it is conditional binding AFAIK it is not possible to inject property there.

assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterModule.java
garagatyi: My intention was to make it possible to deploy SPI branch of Che on Minishift ASAP. 
AFAIK we don't see a use case with several infras impls bound and are going to have separate assemblies per infra, but don't want to introduce them now. So it is a temporary solution to start using Che on Minishift with minimal efforts.

riuvshin: oki, thanks for info

PR #6186, opened 2017-09-07
workspace master and agent : avoid scan of classes at runtime in order to find DynaModules

benoitf:
### What does this PR do?
Today, ws-che-master and ws-che-agent are deployed using war files. In order to have extensibility, we can apply a `@DynaModule` annotation on guice class modules. Then, at runtime, Servlet container(Tomcat) is scanning the jar files to find the `@DynaModule`classes and then call the `org.eclipse.che.inject.ModuleScanner` class by calling `onStartup`method

When Che is initialized, it calls `CheBootStrap`class that ask the modules found by the scanner by calling `modules.addAll(ModuleScanner.findModules());`

The issue is that we've a lot of scanning operations (performed by SCI (ServletContainerInitializer) .
Following https://wiki.apache.org/tomcat/HowTo/FasterStartUp if we disable all JAR scanning operations we lost the ability to use `DynaModule`which is not good.

The idea is then to still disable scanning at runtime but keep using DynaModule.
By writing a plugin that could scan all `@DynaModule` when generating the war file, we could avoid the scan at runtime and then keep the ability to use `@DynaModule``

This PR introduces this plugin and the use of this plugin inside ws-master and ws-agent.
On the agent, we should also add `<absolute-ordering />`element but it will break archetypes/custom assemblies.

#### How is it working ?

The DynaModule plugin scan all the classes (there is a way to skip some jars that we're sure doesn't contain any `@DynaModule` annotations) and generate a class that provides a `getModules()` method returning all DynaModule modules.
This class is then loaded at runtime and CheBootstrap can load these modules, avoiding the use of scanning classes at runtime.

#### Details

- Plugin uses ASM visitor to scan classes avoiding the need to instantiate classes (if some dependencies are available only at runtime)
- Plugin generates a META-INF/services/org.eclipse.che.inject.ModuleFinder file contaning the generated class.
- Plugin generates the class implementing the ModuleFinder interface (getModules() method) at `process-classes` lifecycle (based on maven lifecycle : process-classes | post-process the generated files from compilation, for example to do bytecode enhancement on Java classes)
- ModuleScanner class also ask ServiceLoader SPI to get modules through ModuleFinder implementations.

then CheBootStrap is able to find all DynaModule modules.

the flow being:
CheBootStrap --> ModuleScanner.getModules() --> ServiceLoader<ModuleFinder> --> GeneratedClass.getModules()

#6169 

#### Changelog
workspace master and agent : avoid scan of classes at runtime in order to find DynaModules 

#### Release Notes
N/A

#### Docs PR
N/A

Change-Id: I8e33a2007a41482df7152d2bf875505053840e8a


skabashnyuk:
I think this is a step back in plugin development experience. Instead of "put your artifact in claspass aka declare in pom.xml" developer has to do some coding. All our archetypes, custom assemblies are based on this. All of that will be broken after this PR.

One of the alternatives is to prescan @Dynamodules on build time and put results in some index file and reuse it in CheButstrup in runtime. 

But let's step back and talk about the real numbers of seconds what we want to improve. At what cost? Does it worth it?

benoitf:
@skabashnyuk 

>  Instead of "put your artifact in claspass aka declare in pom.xml" developer has to do some coding. All our archetypes, custom assemblies are based on this. All of that will be broken after this PR.

AFAIK none of them will be broken. I studied it before and custom assemblies or archetype are using `che-wsagent-core` module. I only update the 'che wsagent' assembly so it impacts only 'che' but benefits to all Eclipse Che end-users.

> One of the alternatives is to prescan @dynamodules on build time and put results in some index file and reuse it in CheButstrup in runtime.

Yes I wanted to create this issue as well that so it can benefits to 'ALL' custom assemblies, archetypes, etc. But it's another item.

> But let's step back and talk about the real numbers of seconds what we want to improve. At what cost? Does it worth it?

For sure it worth it. It's how we optimize server side components. I don't see why Che would not be optimized as well.

garagatyi:
I think that kind of issue should be the enhancement, not the bug.
Listing of all the plugins modules contradicts pluggability which you were discussing on f2f. Can you elaborate on why you decided to even decrease pluggability in this PR?

benoitf:
@garagatyi 

> Listing of all the plugins modules contradicts pluggability which you were discussing on f2f. Can you elaborate on why you decided to even decrease pluggability in this PR?

It's not contradicting pluggability. As you need to rebuild the assembly currently it does not reduce it.
At F2F I was talking about "runtime pluggability" and it is not linked to that part as well as in my demo there was no scan of the classes as well to bring a plugin at runtime.

benoitf:
Also I wouldn't mix "developer experience" with "production experience" as usually there is a "developer mode" and "production mode"

skabashnyuk (changes requested):
-1 I don't believe for now that win of a couple of seconds on ws start are worth of a headache with a support of Eclipse Che assembly and all existing and future assemblies based on Eclipse Che Full assembly. changing off the idea of @Dynamodule approach. 

Can this Module be generated like we doing with gwt.xml for instance? Have you consider other options that don't require Java code changes when you're adding a new plugin.

benoitf:
@skabashnyuk did you read what I've written.
it has **NO** impacts on custom assemblies and archetype.

benoitf:
> Can this Module be generated like we doing with gwt.xml for instance? Have you consider other options that don't require Java code changes when you're adding a new plugin.

yes it can but this is step 2 as described in my previous comments.
But step 2 is not mandatory for having immediate performance improvements.

skabashnyuk:
As an assembly maintainer, I have **opposite** opinion. And think that we should also consider some alternatives. I agree that jar scanning is the place where we can optimize agent performance but I don't agree with the way how it was made.

benoitf:
if you agree that we can optimize, then why it was not optimized ? or why it was not raised when you've done profiling on https://github.com/codenvy/codenvy/issues/1466 ?

garagatyi:
@benoitf @skabashnyuk all of us are maintainers in different parts of the system. Can we decrease the temperature in the room a bit?
@skabashnyuk can you elaborate on how this change will influence custom assemblies and archetypes?

skabashnyuk:
>if you agree that we can optimize, then why it was not optimized ? or why it was not raised when you've done profiling on codenvy/codenvy#1466 ?

Too complicated to solve  ~2 seconds on OSX, on Linux desktop, it's even less when what.  We have other different places to improve where developers suppose to spent much more when 2 seconds. For instance editor, language servers. 

>@skabashnyuk can you elaborate on how this change will influence custom assemblies and archetypes?

It requires additional steps to add the plugin.

benoitf:
>> @skabashnyuk

> can you elaborate on how this change will influence custom assemblies and archetypes?
>It requires additional steps to add the plugin.

I disagree, custom assemblies and archetypes have nothing to do after this PR

garagatyi:
> It requires additional steps to add the plugin.

So, basically, it will break current archetypes, right?

benoitf:
@garagatyi not at all as archetypes are based on ws-agent-core (and this one will scan the modules)

skabashnyuk:
>>@garagatyi not at all as archetypes are based on ws-agent-core (and this one will scan the modules)

And that exactly the problem. Your solution is fixing only this Eclipse Che Full assembly. All other
like archetypes don't get this boost or like Codenvy, or Codenvy OnPrem will get a headache to don't forget to put plugins in two places. If we start solving this performance problem let's solve it once and for all.

benoitf:
it's step 2 of the improvement to make it to the others.
Also you're shifting your initial remark.
At first you said it will be break all the assemblies and I was saying it will not, now you're saying that the boost should be applied for all assemblies

skabashnyuk:
>it's step 2 of the improvement to make it to the others.

I disagree. It means that it will be put on a long shelf and some one will have a headache

>Also you're shifting your initial remark.
>At first you said it will break all the assemblies and I was saying it will not, now you're saying >that the boost should be applied to all assemblies

No. It's just because I was wrong and you are right that Eclipse Che archetype will be no be broken.  They will work as it's now.

slemeur:
This change is going to highly impact the end-user experience, in a positive way. That's only what we should keep in mind. 

Maybe it will create

sunix:
Hi there, we are trying to speedup workspace startup which is very slow in openshift.io.
If there are any startup speedup improvements to make there and there, we should consider them.
I saw @benoitf 's demo and it seems that it's not about 2 secs ... but 40% of an overall time to start a workspace agent which is really awesome IMO.
I don't think it will change the developer experience that much and it won't be that hard to maintain, compared to what we will gain in performance improvement.

skabashnyuk:
> demo and it seems that it's not about 2 secs ... but 40% 

Ok 40%.


@sunix  I totally support the idea to improve performance. But let's take some time to make it general and without changing experience. 
What if WsAgentModuleList.java will be generated like many things in ide war https://github.com/eclipse/che/blob/master/assembly/assembly-ide-war/pom.xml#L268-L325

benoitf:
I would say that we should do more incremental steps instead of trying to apply it everywhere at once. (BTW this approach often fails)

- short term : use it in Che and see how it behaves
- mid term: introduce generation of the DynaModule list and use it in che then in codenvy.
- mid term++: apply it everywhere (custom assemblies, archetype)

so it's better to see how it behaves and to see if there are impacts.

skabashnyuk:
-1 to experiment here.
if it's urgently needed on production lets experiment on production. https://github.com/redhat-developer/rh-che/tree/master/builds/fabric8-che/assembly/assembly-wsagent-war

Don't see a problem to start from ```mid term``` or ```mid term++``` in Eclipse Che from the beginning. This is not kind of problem that some one will die without it.

benoitf:
@skabashnyuk because from what I've seen, it has always failed to apply it everywhere at once.

benoitf:
@skabashnyuk also please clarify your -1.

It's not experimenting. It's optimizing Che first, and then propagate the change slightly to the other assemblies and the archetypes.

Because currently you're -1 is being against gaining seconds when booting up the Che workspace agent. (which can be 40% of the overall time)

skabashnyuk:
-1 
1. duplicate plugin declaration in pom.xml + java file.
2. To make it for Eclipse Che Full assembly only.

Proposal
1. Generate java file with all modules like we doing in https://github.com/eclipse/che/blob/master/assembly/assembly-ide-war/pom.xml#L268-L325
2. Reuse this approach in all assemblies based on Eclipse Che.
3. Turn off scanning annotation for all Eclipse Che web applications. 

> (which can be 40% of the overall time)

Just assumption. My best guess it's 2-5 sec on Mac, Linux can be completely different.

benoitf:
I never said that it will be only for Che forever, it's the principle of using iterative development.
Things are not black and white, there is a pattern that can bring the feature for everyone with small incremental steps.
It's strange to refuse step 1 and only wanting the whole steps at once.

Also when you're saying duplicate plugin declaration in pom.xml + java file.
we're already duplicating many stuff for all Guice modules that are not plugins so this argument is not really true.

skabashnyuk:
Ok. Don't agree that this is technically possible to generate WsAgentModuleList.java during build phase?

benoitf:
I never disagree and I said since my first reply that it's part of the step 2

skabashnyuk:
ok. I'm glad that we have the same view 😄  We both agree that this is just technical exercise. 
And since we both wrote some generators and knows that it's not what hard work. 
What if we collaborate and goes directly to the level 2 with a single step, as a team. You will write generator and adapt maven build, I'll handle archetypes, Codenvy, and all derivative assemblies.

benoitf:
ok let's do that.

gorkem:
Wow, What a read to start my day with. I feel like you guys should be hugging 🤗  though

skabashnyuk:
Comment from https://github.com/eclipse/che-dependencies/pull/55#issuecomment-328962472.
Let's sync our vision for WsAgentModuleList.java generation.
I thought plan would be:
In ws-agent(ws-master) war we have a couple of phases.
1. Before compile phase take a look on all dependencies with reflection tool, filter all classes annotated with ```@Dynamodule``` annotation.
2. Generate from template WsAgentModuleList.java and put it to   ```target/generated```
3. Add ```target/generated``` as maven source folder and compile it.
4. Add   WsAgentModuleList.class to the  tomcat annotation scanning list.

WDYT?

benoitf:
@skabashnyuk as I said in the che-dependencies repository

reflection tool is not a good candidate : it loads the classes . At compile time we may not have all required dependencies (and it is slow) but it's more about the process that there is a flaw.

To be the faster, we need to scan "ALL" the classes even those that are part of the war itself. So it must be at `process-classes`lifecycle to be sure that we scan all the classes.

so the plugin is doing at `process-classes` lifecycle
- scan all the classes
- generate the class providing the `getModules()` list.

and in web.xml absolute-ordering element is used and in context.xml we skip all the jars.

skabashnyuk:
Can we add ALL classes of war to scan annotation list? Or may define a list of it, for example, put it in some folder, etc? So we can move to ```generate-sources``` phase. That will allow us not to directly work with byte code generation that is quite complicated itself.

benoitf:
@skabashnyuk no, fully disabling scanning disable also the SCI components so we don't scan anything anymore at all.
BTW the plugin is ready and fully working.

skabashnyuk:
ok. just worried that it would be not trivial.

benoitf:
FYI here is the updated PR.
I've rewritten it including the description and updating code.

gazarenkov:
+1 for this.
I like decoupling "scanning" DynaModules from SCI in general (even out of startup time scope) and the ability to plug other mechanism via Service loading.

What I miss in discussion around performance though is numbers analyzing the overall start time, what and how much impacting on it in general (Docker, image, container, agent etc) and what we gain in percentage in each (this in particular) case. It would give us engineering view of problem.
@benoitf can you make this profiling/analysis as a separate task?

benoitf:
@gazarenkov 
Some details are there
https://github.com/eclipse/che/issues/6169#issuecomment-328064479

skabashnyuk (approved):

benoitf:
@skabashnyuk do you will handle codenvy and archetype inclusion  ?
Because right now if I merge this PR I think it will break codenvy as it relies on the ws-agent war of che
https://github.com/codenvy/codenvy/blob/master/assembly/assembly-wsagent-war/pom.xml#L56

core/che-core-dynamodule-maven-plugin/src/test/projects/project/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+                <artifactId>che-core-dynamodule-maven-plugin</artifactId>
+                <configuration>
+                    <project implementation="org.eclipse.che.plugin.dynamodule.stub.DynaModuleListGeneratorMojoProjectStub"/>
+                </configuration>
+                <extensions>true</extensions>
+            </plugin>
+        </plugins>
+    </build>
+
+</project>
```
mkuznyetsov: new line?

benoitf: fixed

benoitf:
I've rebased to current master (as it was providing 5.18.0-SNAPSHOT artifact)

mkuznyetsov:
If assembly uses another WAR as overlay, plugin should scan for dynamodules in it. Otherwise we would need to use it's `classes` artifacts

mkuznyetsov:
I've used Set to exclude generation of duplicating dynamodules

mkuznyetsov (approved):

ashumilova (approved):

PR #6202, opened 2017-09-08
Backport rh-che changes to OpenShift deployment script

l0rd:
### What does this PR do?
Backport a fix merged in rh-che repository a few days ago (https://github.com/redhat-developer/rh-che/pull/320) to update the maven central URL where deployment configurations for Che are published.

https://github.com/redhat-developer/rh-che/issues/315

### DIFF
--- a/dockerfiles/cli/scripts/openshift/deploy_che.sh
+++ b/dockerfiles/cli/scripts/openshift/deploy_che.sh
@@ -184,7 +184,7 @@ if ! oc get project "${CHE_OPENSHIFT_PROJECT}" &> /dev/null; then
 
   echo -n "no creating it..."
   oc new-project "${CHE_OPENSHIFT_PROJECT}" &> /dev/null
-  ## TODO we should consider oc apply the latest http://central.maven.org/maven2/io/fabric8/online/packages/fabric8-online-che-quotas-oso/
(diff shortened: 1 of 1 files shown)


benoitf (approved):

garagatyi:
AFAIK @riuvshin is on vacation. He might be offline.
@l0rd Can you elaborate on why you treat this issue as a bug?

benoitf:
it's a bug as it's not using the correct deployment configuration (repository has been changés)

garagatyi:
From my point of view if something changes in a backward incompatible way projects that depend on it don't get bugs, but rather suffer from breaking change. But it is my lamentation, so you can totally ignore it.

garagatyi:
I'm saying that bug makes ppl feel that it is some defect inside of a project. But in some cases, it is just decision made in another place that affected the stability of the project.
Just like in case of the leftpad - someone removed his artifact and thousands of projects got broken. Does it mean that all of those projects got the new bug? I doubt.
But on the other hand, I understand that if something is broken it is treated as a bug. Even when the source of a break is somewhere else.

riuvshin (approved):

l0rd:
@garagatyi Maybe `kind/task` is the right label here ;-)

Issue #6203, opened 2017-09-10
Browser was freeze when running `npm install` in the console.


monaka:
**Reproduction Steps:**

1. Setup node environment
2. Get an application contains packages.json via GitHub or somewhere.
3. Run `npm install`
4. Browser is hang up. Browser's console shows 504 error in `ws://.../connect`
   
This is not 100% reproduction.

**OS and version:**    

* Version :5.18.0-SNAPSHOT
* Revision :5231059555dc1838
* Build Time :2017-09-08 13:45:52

**Diagnostics:** 

I suspect the root cause is this NPE.
![screenshot 2017-09-10 at 15 48 44 - display 2](https://user-images.githubusercontent.com/101795/30246869-ab7db44e-9640-11e7-9e4b-f46492568969.png)

sr229:
@monaka I can't reliably reproduce this, my npm exits fine and there was no 504 reported in my end. Probably connection issues?

monaka:
I don't think the root cause is connection issue.
Traefik report like this.

```
time="2017-09-10T06:40:22Z" level=warning msg="Error dialing `172.18.0.2:4412`: dial tcp 172.18.0.2:4412: getsockopt: connection timed out" 
```

Looks the terminal service on dev-machine is down.

monaka:
I created PR #6205 to show my expectations. Frankly I've not tested that yet. Anyway this will be issue by execution timing.

bmicklea:
@eivantsov can you try and reproduce?  I can't either...

bmicklea:
@monaka thanks for the issue and associated PR - it looks like you've found a bug here and we really appreciate the PR. One of the maintainers has already commented on the PR asking you for some feedback - can't wait to see this merged!

sr229:
@monaka I was able to reproduce the NPE after some trials, so I can confirm this.

davidwindell:
We've also seen this issue quite a lot, but in our case when we copy a massive amount of files/directories from within a docker image into the workspace

SkorikSergey:
Is this still an issue?

davidwindell:
Not for us, resolved with the tree fixes last year

bmicklea:
Closing as fixed - @monaka if this is still an issue please reopen with details.

PR #6205, opened 2017-09-10
FileWatcher service: Add exclusive process in register/refer dirs.

monaka:
(This patch is not checked well. May need more fixes)

### What does this PR do?

Wait loop until the target dir is registered.

#6203

### DIFF
--- a/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/watcher/FileWatcherService.java
+++ b/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/watcher/FileWatcherService.java
@@ -191,9 +191,11 @@ public class FileWatcherService {
     } else {
       try {
         LOG.debug("Starting watching directory '{}'", dir);
-        WatchKey watchKey = dir.register(service, eventKinds, eventModifiers);
-        keys.put(watchKey, dir);
-        registrations.put(dir, 1);
+        synchronized (keys) {
+          WatchKey watchKey = dir.register(service, eventKinds, eventModifiers);
+          keys.put(watchKey, dir);
+          registrations.put(dir, 1);
+        }
       } catch (IOException e) {
         LOG.error("Can't register dir {} in file watch service", dir, e);
       }
@@ -275,7 +277,17 @@ public class FileWatcherService {
     while (running.get()) {
       try {
         WatchKey watchKey = service.take();
-        Path dir = keys.get(watchKey);
+        Path dir;
+        synchronized (keys) {
+          dir = keys.get(watchKey);
+        }
+
+        if (dir == null) {
+          resetAndRemove(watchKey, dir);
+
+          LOG.debug("Reported directory is not registered - skipping.");
+          continue;
+        }
 
         List<WatchEvent<?>> watchEvents = watchKey.pollEvents();
 


dkulieshov:
Hi @monaka,

It is very appreciated that you've discovered such a nasty bug. You're saying that the NPE is thrown because value of variable `dir` [here](https://github.com/eclipse/che/blob/5b373e6cf5bfeb16f278c4d750fe64361445eeb4/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/watcher/FileWatcherService.java#L299) is **`null`** because the directory is not added to the registry yet. It is an excellent observation and to the best of my ability I see only one case where it is possible - if a file system event happens between [this](https://github.com/eclipse/che/blob/5b373e6cf5bfeb16f278c4d750fe64361445eeb4/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/watcher/FileWatcherService.java#L194) and [that](https://github.com/eclipse/che/blob/5b373e6cf5bfeb16f278c4d750fe64361445eeb4/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/watcher/FileWatcherService.java#L195) lines of code are executed.

However I'm not sure that the changes you provided fixes the issue efficiently. Correct me if I'm wrong but I do believe that these lines 
```java
if (dir == null) {
  LOG.debug("Reported directory is not registered - skipping.");
  continue;
}
``` 
are not necessary because according to the use case logic eventually folders are always registered (unless there may be other use cases that I'm missing here). If we agree that no extra **`null`** check is needed I think we can simply keep resetting watch key until the folder is registered (which should take as long as the second of the two mentioned earlier lines is executed - not long):
```diff
diff --git a/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/watcher/FileWatcherService.java b/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/watcher/FileWatcherService.java
index 6c74413af3..4460083dd4 100644
--- a/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/watcher/FileWatcherService.java
+++ b/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/watcher/FileWatcherService.java
@@ -277,6 +277,13 @@ public class FileWatcherService {
         WatchKey watchKey = service.take();
         Path dir = keys.get(watchKey);
 
+        if (dir == null){
+          resetAndRemove(watchKey, dir);
+
+          LOG.debug("Folder is not registered but watch key already has pending events, resetting");
+          continue;
+        }
+
         List<WatchEvent<?>> watchEvents = watchKey.pollEvents();
 
         if (suspended.get()) {
```

What do you think?

monaka:
I think this issue is raised like this even though I've not checked well.

1. Executed line 194. https://github.com/PizzaFactory/che/blob/51378ed07bfc30dec3f86bec1221828764ec8788/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/watcher/FileWatcherService.java#L194
2. Event was raised before executed line 195. https://github.com/PizzaFactory/che/blob/51378ed07bfc30dec3f86bec1221828764ec8788/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/watcher/FileWatcherService.java#L195
3. `keys.get(watchKey);` returns null as line 195 have not executed yet.

So possibly calling `resetAndRemove()` is not the best solution.

monaka:
So it may be better to use `syncronized(lockObj)` instead of the wait loop.

monaka:
I pushed the another plan.
IMO, there is no case that @dkuleshov suggested above. But I added in my new patch for the stability.

dkulieshov:
@monaka 

> I think this issue is raised like this even though I've not checked well.
>
> 1, Executed line 194. https://github.com/PizzaFactory/che/blob/51378ed07bfc30dec3f86bec1221828764ec8788/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/watcher/FileWatcherService.java#L194
> 2. Event was raised before executed line 195. https://github.com/PizzaFactory/che/blob/51378ed07bfc30dec3f86bec1221828764ec8788/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/watcher/FileWatcherService.java#L195
> 3. keys.get(watchKey); returns null as line 195 have not executed yet.
> So possibly calling resetAndRemove() is not the best solution.

Actually that is exactly what I was saying here in my previous [comment](https://github.com/eclipse/che/pull/6205#issuecomment-328515843) :smile: :

>  I see only one case where it is possible - if a file system event happens between [this](https://github.com/eclipse/che/blob/5b373e6cf5bfeb16f278c4d750fe64361445eeb4/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/watcher/FileWatcherService.java#L194) and [that](https://github.com/eclipse/che/blob/5b373e6cf5bfeb16f278c4d750fe64361445eeb4/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/watcher/FileWatcherService.java#L195) lines of code are executed.

Anyway I'm not sure that I follow you when you're saying
> So possibly calling `resetAndRemove()` is not the best solution. 

because in this specific case watch key is valid and not cancelled so this method will only call `java.nio.file.WatchKey#reset` which (according to the [java docs](https://docs.oracle.com/javase/7/docs/api/java/nio/file/WatchKey.html#reset())) 
> Otherwise if there are pending events for the object then this watch key is immediately re-queued to the watch service. If there are no pending events then the watch key is put into the ready state and will remain in that state until an event is detected or the watch key is cancelled.

So it will re-enqueue watch key with all pending events and iterate over the first part of the loop until [this](https://github.com/eclipse/che/blob/6e9825c699d34eef6b8bf36fc0194ff871cbee84/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/vfs/watcher/FileWatcherService.java#L195) line is executed and then continue it's normal work and that is probably precisely what we all want here. Unless I'm missing something important here?

I'm also eager to understand your motivation for introducing synchronized blocks here. I mean I see that it will probably solve the issue but I'm not really sure if we will benefit in some way here by setting additional synchronization, because with or without synchronized blocks we will not go further in the loop until `dir` variable is put to `keys` map: 
- with synchronization block the loop is blocked entirely until `keys` monitor is released 
- in case we will use these lines
 ```diff
+        if (dir == null){
+          resetAndRemove(watchKey, dir);
+
+          LOG.debug("Folder is not registered but watch key already has pending events, resetting");
+          continue;
+        }
```
watch key will be repeatedly re-queued and a new iteration will be repeatedly run until 'dir' is put to 'keys'. Moreover, and that is important part I think, _re-queued_ means that it is put to the end of a queue so if there are more watch keys to take in the watch service they will be taken and processed normally and prior to our unfortunate watch key. In this case the loop won't be blocked at any time and the event processing process will be continuous. And that may be beneficial taking into account that we're talking about intensive file system event stream processing.

Not taking into account extra synchronization (IMHO) the solution looks good to me.

monaka:
@dkuleshov Let me check it. Do you agree with my assumption? (that events are raised between L194 and L195)

dkulieshov:
@monaka, your assumption is absolutely correct in my opinion :+1:

vparfonov (approved):

Issue #6206, opened 2017-09-11
Traefik doesn't restart after rebooting the host OS.


monaka:
The title is all.

**Reproduction Steps:**

1. Start Che with `-e CHE_SINGLE_PORT=true`
2. Reboot your host OS.
3. Run `docker ps`

**OS and version:**    

Recent nightly images..

**Diagnostics:** 

I've not tested but simply it should be added `restart: always` into [there](https://github.com/eclipse/che/blob/5231059555dc1838e0b6388dc98711878d20c05c/dockerfiles/init/modules/compose/templates/docker-compose.yml.erb#L58).

PR #6207, opened 2017-09-11
Setup Traefik container to restart policy 'always'.

monaka:
As descibed in #6206, Traefik will not restart after rebooting the host OS.
So you will fail to reconnect to che-host. 

### What does this PR do?

Adding `restart: always` into docker-compose.yml

#6206 

#### Changelog

Traefik restarts always.

#### Release Notes

Treafik restarts after the host OS was rebooted.


benoitf (approved):
nice improvement @monaka

riuvshin (approved):

PR #6245, opened 2017-09-12
openshift.io-798 Checking only first 20 chars of tag in 'getImageStreamTagFromRepo' for long tags

ibuziuk:
### What does this PR do?
After switching rhche stacks to`registry.devshift.net` ImageStream naming prefix has been significantly increased and creation of workspaces against some stacks (spring-boot) is started to fail. In order to fix it checking only first 20 chars of tag in 'getImageStreamTagFromRepo' for long tags is required

https://github.com/openshiftio/openshift.io/issues/798

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
@@ -1105,7 +1105,7 @@ public class OpenShiftConnector extends DockerConnector {
     // Since repository + tag are limited to 63 chars, it's possible that the entire
     // tag name did not fit, so we have to match a substring.
     String imageTagTrimmed =
-        imageStreamTagName.length() > 30 ? imageStreamTagName.substring(0, 30) : imageStreamTagName;
+        imageStreamTagName.length() > 20 ? imageStreamTagName.substring(0, 20) : imageStreamTagName;
 
     // Note: ideally, ImageStreamTags could be identified with a label, but it seems like
     // ImageStreamTags do not support labels.


plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -1105,7 +1105,7 @@ private ImageStreamTag getImageStreamTagFromRepo(String imageStreamTagName) thro
     // Since repository + tag are limited to 63 chars, it's possible that the entire
     // tag name did not fit, so we have to match a substring.
     String imageTagTrimmed =
-        imageStreamTagName.length() > 30 ? imageStreamTagName.substring(0, 30) : imageStreamTagName;
+        imageStreamTagName.length() > 20 ? imageStreamTagName.substring(0, 20) : imageStreamTagName;
```
l0rd, line 1108: Can't we do, as OpenShift does: trim just the chars to make the image name + tag < 63?

Because I am afraid that if tomorrow we have a image name with 10 chars more (e.g. `registry.../che-stacks/centos-wildfly-swarm`) we will have the problem again.

And the error message we got when facing that problem wasn't clear at all. We lost a few hours to understand what was happening exactly. Can't we anticipate that. For example if we see that the imageTagTrimmed would be < 3 chars, fail with an error message that explains that the image name is too long?

ibuziuk: This requires more investigation. 
> Can't we do, as OpenShift does: trim just the chars to make the image name + tag < 63?

The problem here is that we do not really know what is the length of image name we are trying to find. e.g. while IS creation tag size is trimmed (<63), but when we trying to find image by tag name non trimmed version is used

> Because I am afraid that if tomorrow we have a image name with 10 chars more (e.g. registry.../che-stacks/centos-wildfly-swarm) we will have the problem again.

yes, so now there is an assumption that image name must be less than 43 chars (previously it was lees than 33)

All in all, I agree that this solution is not robust and hard to maintain - so we could treat it as a hot fix

l0rd (approved):

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
l0rd: Ok that makes sense. Can you open an issue to make this more robust (and error message more clear) so that we can plan to fix it in the next sprints?

ibuziuk: @l0rd sure - https://github.com/redhat-developer/rh-che/issues/334

Issue #6216, opened 2017-09-11
Ignore Routes that are specified in OpenShift recipe type


sleshchenko:
Related to SPI branch only.
Current OpenShift spi implementation creates Routes that are specified in OpenShift recipe type, but does not resolve them as servers.
It doesn't make sense to create them while they are not resolved as servers and their publicly accessible URLs are not returned by Workspace API.
As it was decided not to expose automatically something (ports in dockerimage, dockerfile, etc) as servers It is needed to modify current approach with following changes:
- Do not create Routes that are specified in recipe by user;
- If Routes are present in OpenShift Recipe, add a warning(to InternalRuntime warnings list) that they will be ignored. It also could contain information that user can define a server in workspace config to expose a container port.

PR #6259, opened 2017-09-13
Ignore Routes in OpenShift recipe

sleshchenko:
### What does this PR do?
Ignore Routes in OpenShift recipe.
Add warnings when OpenShift recipe contains Routes and they are ignored.

Note that warnings are not delivered to the clients yet. Because `InternalEnvironment` is not exposed by Rest API. It will be implemented later.

https://github.com/eclipse/che/issues/6216


garagatyi (approved):
Looks good!

akorneta (approved):

PR #6262, opened 2017-09-14
CHE-6033: Fix bug when git colors are not updated after git operations

### LINKED ISSUE #6033: Wrong git status displayed in editor's tab after pushing changes
After pushing changes onto the remote repository, the git status color used for the committed is incorrect.

**Reproduction Steps:**
1- Do some changes in existing files
2- Create a new file and include content in it
3- Commit and push to the remote repository

*Expected Output:*
The filename color used in the editor's bar is "normal" = gray after the changes have been pushed - exactly as in the project explorer.

*Current Output:*
The filename color is green, while in the git status that file is clean and the color is "normal" in the project explorer.

See attached snapshot and file "deployment.yml" :
screen shot 2017-08-17 at 17 05 48

**Related to:**
https://github.com/eclipse/che/pull/5722

**OS and version:**    
OSIO Che 1.0.289 - 08/17/2017

vinokurig:
### What does this PR do?
Added events on Git operations that can change Git index during their execution.
We have `GitIndexChangedDetector` that detects changes in `index` file inside `.git` folder,
but that changes are made before a JGit comand has finished, so JGit status, that is called when `index` file is changed, can be not actual. So need to have such events, with actual status inside, that are called when JGit command is finished.

P.S. When working with Git from terminal, everything works fine.

fixes https://github.com/eclipse/che/issues/6033

#### Changelog
Fix bug when git colors are note updated after git operations

#### Release Notes

N/A

#### Docs PR

N/A


wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitJsonRpcMessenger.java
```diff
@@ -50,21 +53,34 @@ private void unsubscribe() {
...
-        .forEach(
-            it ->
-                transmitter
-                    .newRequest()
-                    .endpointId(it)
-                    .methodName("git/checkoutOutput")
-                    .paramsAsDto(event)
-                    .sendAndSkipResult());
+  public void onEvent(GitEvent event) {
+    if (event instanceof GitCheckoutEvent) {
```
tolusha: Make `onEvent` method simpler.  Encapsulate handling different git events in methods. 
```
if () {
  handleEvent1()
} else () {
  handlerEvent2()
}
```

wsagent/che-core-api-git-shared/src/main/java/org/eclipse/che/api/git/shared/IndexChangedEvent.java
```diff
@@ -0,0 +1,32 @@
...
+
+import org.eclipse.che.dto.shared.DTO;
+
+/**
+ * Event for indicating that Git operation that can cause Git Status changes is called.
+ *
+ * @author Igor Vinokur.
+ */
+@DTO
+public interface IndexChangedEvent extends GitEvent {
```
tolusha: If several git commands reveals the same issue then it is recommended to use different events:
* GitCommitEvent
* GitRepositoryCreateEvent
etc.

tolusha (changes requested):

wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitJsonRpcMessenger.java
vinokurig: done

wsagent/che-core-api-git-shared/src/main/java/org/eclipse/che/api/git/shared/IndexChangedEvent.java
vinokurig: done

tolusha (approved):

wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitJsonRpcMessenger.java
```diff
@@ -67,6 +86,26 @@ public void onEvent(GitCheckoutEvent event) {
...
+      status = ((GitResetEvent) event).getStatus();
+    } else if (event instanceof GitRepositoryInitializedEvent) {
+      status = ((GitRepositoryInitializedEvent) event).getStatus();
+    }
+
+    for (String endpointId : endpointIds) {
+      transmitter
+          .newRequest()
+          .endpointId(endpointId)
+          .methodName("event/git-index")
```
dkulieshov: May be we could use here something like `event/git/indexStatusChanged` or `event/git/indexChanged` similar to previously used `event/git/subscribe` just to keep homogeneous method names here?

dkulieshov: I'm actually wondering if we should add one more level to the method naming scope, I mean we could use `event/git/index/subscribe` and then `event/git/index/statusChanged` or `event/git/index/changed`, what do you think?

dkulieshov (approved):

wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitJsonRpcMessenger.java
vinokurig: This chanel is already used in `GitIndexChangedDetector` so I am reusing it here

dkulieshov: Well, I'm just assuming that we probably could change it if it would fit the logic better :-)

vparfonov (approved):

Issue #6162, opened 2017-09-06
After creating new folder via terminal, folder displayed as well configured project in Project Explorer but should be mark as broken project


vparfonov:
**Reproduction Steps:**
1. Create new folder via terminal: `mkdir test`
2. Check Project Explore, just created folder will be displayed as well configured blank project - this is wrong
3. Refresh IDE page 
4. Check Project Explore, just created folder will be displayed as well broken project - this is ok
![2017-09-06-22-44-24](https://user-images.githubusercontent.com/1636592/30131705-6a4f6fc2-9356-11e7-8b10-07a3cd916e75.gif)

PR #6264, opened 2017-09-14
CHE-6162: Fix problem project detection procedure 

vparfonov:
### What does this PR do?
Fix problem project detection procedure. 
Avoid call ResourceNode.updatePresntation if not need

#6162


garagatyi (approved):

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/resources/impl/ProjectImpl.java
```diff
@@ -96,6 +97,11 @@ public SourceStorage getSource() {
     return reference.getSource();
   }
 
+  @Override
+  public List<ProjectProblem> getProblems() {
+    return null;
```
vzhukovs: Maybe it would be better to return actual list of problems? `reference.getProblems()` WDYT?

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/resources/tree/ResourceNode.java
```diff
@@ -189,6 +193,7 @@ public void updatePresentation(@NotNull NodePresentation presentation) {
                 ? nodesResources.hiddenSimpleFolder()
                 : nodesResources.simpleFolder());
       } else if (getData().getResourceType() == PROJECT) {
+        Log.info(getClass(), ((Project) getData()).isProblem());
```
vzhukovs: Non-informative info log

vzhukovs (approved):

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/resources/impl/ProjectImpl.java
vparfonov: yes, thx

PR #6266, opened 2017-09-14
Fix locators in NavigateToFile page object

SkorikSergey:
### What does this PR do?
Change locators in NavigateToFile page object which checks popup of the **Navigate To File** feature.


tolusha (changes requested):
Add debug id to the all items and use them instead of xpaths

SkorikSergey:
All changed locators are used for searching dynamically created web elements .

tolusha:
This is a suggestion panel. You have to investigate corresponding source code to make assumption.

https://github.com/eclipse/che/blob/3cdf3b6eb713ba8a02fec8b3d148c4ef4363c421/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/navigation/NavigateToFileViewImpl.ui.xml#L74

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/NavigateToFile.java
```diff
@@ -45,21 +45,21 @@ public NavigateToFile(
 
   private interface Locators {
     String NAVI_TO_FILE_FORM = "gwt-debug-file-navigateToFile-mainPanel";
-    String FILE_NAME_FIELD = "//div[text()='Enter file name:']/following::input[1]";
-    String FILE_NAME_POPUP = "//div[text()='Enter file name:']/following::table[last()]";
+    String FILE_NAME_FIELD = "gwt-debug-navigateToFile-fileName";
```
tolusha: The name of the constant should reflect its value.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/NavigateToFile.java
```diff
@@ -45,21 +45,21 @@ public NavigateToFile(
...
+    String FILE_NAME_POPUP = "gwt-debug-navigateToFile-suggestionPanel";
```
tolusha: The name of the constant should reflect its value.

tolusha (approved):

PR #6200, opened 2017-09-08
Modify che-server Dockerfile to run on OpenShift

l0rd:
### What does this PR do?
Modify Che alpine-based Dockerfile to make it OpenShift compatible.

### What issues does this PR fix or reference?
The 6th task of #6097

#### Changelog
Image eclipse/che-server:nightly can now be run on OpenShift

#### Release Notes
Image eclipse/che-server:nightly can now be run on OpenShift

PR #6272, opened 2017-09-14
Che dockerfiles: apply chmod only to folders to fool COW

l0rd:
### What does this PR do?
Modify Che alpine and CentOS based Dockerfiles to apply chmod only to folders. 

https://github.com/eclipse/che/pull/6200#issuecomment-329418498: this single instruction at the end of the Dockerfile added a layer of ~185M
```Dockerfile
RUN chmod -R 0777 /home/user
```
That's because of copy on write: every file which permissions have been modified is duplicated in the new layer. To avoid that (and fool cow) we apply chmod only to folders:
```Dockerfile
RUN find /home/user -type d -exec chmod 777 {} \;
```

### DIFF
--- a/dockerfiles/che/Dockerfile
+++ b/dockerfiles/che/Dockerfile
@@ -40,4 +40,4 @@ ENTRYPOINT ["/entrypoint.sh"]
 RUN mkdir /logs /data && \
     chmod 0777 /logs /data
 ADD eclipse-che.tar.gz /home/user/
-RUN chmod -R 0777 /home/user
+RUN find /home/user -type d -exec chmod 777 {} \;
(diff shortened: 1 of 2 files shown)


benoitf (approved):

dockerfiles/che/Dockerfile
```diff
@@ -40,4 +40,4 @@ ENTRYPOINT ["/entrypoint.sh"]
 RUN mkdir /logs /data && \
```
benoitf, line 40: I don't remember why this instruction couldn't be merged with previous RUN instruction ?

l0rd: @benoitf I don't think we have ever discussed that

benoitf: AFAIK It was discussed there https://github.com/eclipse/che/pull/6200#pullrequestreview-61541153

l0rd: Anyway I can't see any problem to do it. But it should be carefully tested.

PR #6277, opened 2017-09-14
CHE-6083: Add ability to run selenium tests against OpenShift infrastructure

### LINKED ISSUE #6083: Modify selenium tests to test OpenShift infrastructure
Related to SPI branch only.
We are eager to check how functional is OpenShift infrastructure comparing to Docker infrastructure in SPI branch. But our selenium tests use dockerimage environment to test different features of the product. 
Right now OpenShift infrastructure doesn't support running dockerimage and dockerfile environments, but support running openshift environment. We need some method to alternate default environment(s) in selenium tests to run tests on OpenShift infrastructure.
It is supposed to be a temporary solution that won't be needed when dockerfile and dockerimage environment are supported by OpenShift infrastructure.

garagatyi:
### What does this PR do?
Allows running QA tests against OpenShift infrastructure from SPI branch deployed to Minishift.
Note it is better to review commit by commit!

Fixes #6083


selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/SeleniumWebDriver.java
```diff
@@ -69,7 +69,7 @@ public SeleniumWebDriver(
       @Named("sys.browser") TestBrowser browser,
       @Named("sys.driver.port") String webDriverPort,
       @Named("sys.grid.mode") boolean gridMode,
-      @Named("sys.driver.port") String webDriverVersion) {
+      @Named("sys.driver.version") String webDriverVersion) {
```
dmytro-ndp, line 72: OMG! 
Good catch, Alex!

selenium/che-selenium-core/bin/webdriver.sh
```diff
@@ -875,69 +878,79 @@ testProduct() {
...
+    fi
+}
 
-    echo "[TEST]"
-    echo "[TEST]"
-    echo -e "[TEST] ${YELLOW}---------------------------------------------------${NO_COLOUR}"
-    echo -e "[TEST] ${YELLOW} RUN UNSTABLE TESTS${NO_COLOUR}"
-    echo -e "[TEST] ${YELLOW}---------------------------------------------------${NO_COLOUR}"
-    echo "[TEST]"
+set -e
```
dmytro-ndp: Could you, please, explain the reasons to use "set -e"?
IMHO handling errors explicitly makes code much more reliable and clear in case like the follow:
```
checkBuild() {
    mvn package
    [[ $? != 0 ]] && { exit 1; }		
}
```

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/user/DefaultTestUser.java
```diff
@@ -30,13 +27,8 @@
...
 
   @Inject
-  public DefaultTestUser(
-      TestUserServiceClient testUserServiceClient,
-      TestWorkspaceServiceClient workspaceServiceClient,
-      TestAuthServiceClient authServiceClient)
-      throws Exception {
-    this.testUser =
-        new TestUserImpl(testUserServiceClient, workspaceServiceClient, authServiceClient);
+  public DefaultTestUser(TestUserImpl testUser) throws Exception {
```
dmytro-ndp, line 30: Good catch, Alex!

selenium/che-selenium-core/bin/webdriver.sh
```diff
@@ -120,9 +117,10 @@ checkParameters() {
         if [[ "$var" =~ --web-driver-version=.* ]]; then :
         elif [[ "$var" =~ --web-driver-port=[0-9]+$ ]]; then :
         elif [[ "$var" == "--http" ]]; then :
-        elif [[ "$var" == "--https" ]]; then  :
+        elif [[ "$var" == "--https" ]]; then :
         elif [[ "$var" == "--che" ]]; then :
         elif [[ "$var" =~ --host=.* ]]; then :
+        elif [[ "$var" =~ --port=.* ]]; then :
```
dmytro-ndp, line 123: New parameter **--port** might be described in the next places as well:
- in the [README.md](https://github.com/eclipse/che/blob/master/selenium/che-selenium-test/README.md)
- in the [setup-selenium.md](https://github.com/eclipse/che-docs/blob/master/src/main/_docs/setup/setup-selenium.md)

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/SeleniumWebDriver.java
garagatyi: tnx

selenium/che-selenium-core/bin/webdriver.sh
garagatyi: `set -e` makes the script fail on any non-zero return code. Pros are:
- you don't have to explicitly check that commands and functions finish gracefully
- if something works not as expected you immediately get exit and error and can create an issue or fix a bug as fast as possible

tolusha: `mvn` returns non-zero exit code if tests failed. So, using `set -e` forces stopping script execution immediately and prevents report generation.

selenium/che-selenium-core/bin/webdriver.sh
tolusha: It is the `spi` branch, updating `setup-selenium.md` is not required.

garagatyi: Thanks!
I've updated readme file. I won't update setup-selenium.md since it is in another repository where there is no spi branch.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/core/UrlUtil.java
```diff
@@ -0,0 +1,39 @@
...
+public class UrlUtil {
+
+  /**
+   * Creates {@link URL} using provided parameters.
+   *
+   * @throws IllegalArgumentException in case URL creation fails because of provided parameters
+   */
+  public static URL url(String protocol, String host, int port, String path)
+      throws IllegalArgumentException {
+    if ((80 == port && "http".equals(protocol)) || (443 == port && "https".equals(protocol))) {
```
tolusha: What are the benefits?

garagatyi: Of what?

skabashnyuk (approved):
Could you please add an instruction how to use it on openshift

selenium/che-selenium-core/bin/webdriver.sh
garagatyi: I haven't noticed that. I'll remove this directive and add explicit check back. Thank you!

garagatyi: done

garagatyi:
@skabashnyuk done

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/core/UrlUtil.java
tolusha: Of having url `http://localhost/api` in comparison with `http://localhost:80/api`?

Anyway, my concern is about using magic constants :)
If would refine `if` clause:

```
int httpsDefaultPort = new URL("https://"+host).getDefaultPort();
int httpDefaultPort = new URL("http://"+host").getDefaultPort();

if ((httpDefaultPort == port && "http".equals(protocol)) || (httpsDefaultPort == port && "https".equals(protocol))) {
...
}
```

garagatyi: I'm afraid it may lead to some weird errors when URL after passing from one component to another may lose default port and some error may happen in tests. It is also weird to see a URL with default port in it. By the way, Che itself would not use the port in URL in this case so it is more consistent.

I don't like magic constants too, but this tool can be used extensively and building URL is quite an expensive operation that involves parsing and validation. I do not think it is a good thing to do on each call.
But I can create constant fields such as:
```
private static final String DEFAULT_HTTP_PORT = 80;
private static final String DEFAULT_HTTPS_PORT = 443;
```
And use it in code. Would it be OK for you?

tolusha: Ok

tolusha (approved):

PR #6363, opened 2017-09-20
interrupt build scripts for Dockerfiles on failure

riuvshin:
### What does this PR do?
interrupt build scripts for Dockerfiles on failure


benoitf (approved):
pls add labels before merging

also the changelog section seems to be optional based on the latest community meeting

PR #6372, opened 2017-09-20
fix docker images build

riuvshin:
### What does this PR do?
fix build of docker images

after adding `set -u` docker images build was failed:
```
../build.include: line 83: IMAGE_ALIASES: unbound variable
```


benoitf (approved):

benoitf:
I would prefer to fix the build.include instead of removing the `set -u` flag

initializing values in build.include with the pattern
`FOO=${FOO:-default}``
should solve the issue of IMAGE_ALIASES (without having to add empty value in all the .sh scripts)

riuvshin:
@benoitf updated :)

PR #6379, opened 2017-09-21
Allow overriding the images used by the CLI

davidfestal:
### What does this PR do?

This PR allows overriding  the images used by the CLI to perform commands.
for example, the following command:

``` 
docker run ... -e IMAGE_INIT=dfestal/che-init:nightly eclipse/che-cli:nightly config
```

will run the `cli config` command, but when running it will use the `dfestal/che-init:nightly` init image, instead of the default `eclipse/che-init:nightly` image.

This makes running specific version of some components with the standard CLI much easier, which can be a great help for development. 

#### Changelog

Allow overriding the images used by the CLI

### DIFF
--- a/dockerfiles/cli/test.sh
+++ b/dockerfiles/cli/test.sh
@@ -31,7 +31,7 @@ fi
 #   The file has to be placed in tests folder in directory containing this script
 # (Optional) second argument is options for a docker run command.
 run_test_in_docker_container() {
-  docker_exec run --rm ${DOCKER_RUN_OPTIONS} \
+  docker_exec run --rm ${DOCKER_RUN_OPTIONS} $2 \
        -v "${BATS_BASE_DIR}":/dockerfiles \
        -e CLI_IMAGE="$ORGANIZATION/$PREFIX-cli:$TAG" \
        -e BATS_BASE_DIR="${BATS_BASE_DIR}" \
--- a/dockerfiles/cli/test.sh
+++ b/dockerfiles/cli/test.sh
@@ -31,7 +31,7 @@ fi
 #   The file has to be placed in tests folder in directory containing this script
 # (Optional) second argument is options for a docker run command.
 run_test_in_docker_container() {
-  docker_exec run --rm ${DOCKER_RUN_OPTIONS} $2 \
+  docker_exec run --rm ${DOCKER_RUN_OPTIONS} \
        -v "${BATS_BASE_DIR}":/dockerfiles \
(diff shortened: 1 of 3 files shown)


benoitf:
https://www.shellcheck.net/
```
  INSTRUCTION=$(echo "${SINGLE_IMAGE}" | sed -e 's/\(.*\)=\(.*\)/\1=${\1:-\2}/g')
                ^-- SC2001: See if you can use ${variable//search/replace} instead.
 
    eval $INSTRUCTION
         ^-- SC2086: Double quote to prevent globbing and word splitting.

```

benoitf:
also, could you add a very simple test case
like https://github.com/eclipse/che/blob/master/dockerfiles/ip/tests/library_tests.bats

to ensure the regexp is valid

riuvshin:
cool!!! is that possible to override multiple images at once?

benoitf:
@davidfestal also it should provide documentation
probably here https://github.com/eclipse/che-docs/blob/71df8bf0d60b5aabfbe2bac2fa2c6d6f11caac36/src/main/_docs/setup/setup-cli.md
but @slemeur may provide the right path

davidfestal:
> cool!!! is that possible to override multiple images at once? 

@riuvshin yes, add the environment variables for all the images you want to override.

davidfestal:
> also, could you add a very simple test case
> like https://github.com/eclipse/che/blob/master/dockerfiles/ip/tests/library_tests.bats

@benoitf  done [here](https://github.com/eclipse/che/pull/6379/commits/6f942cdcc879ae198449b7a59563275dfaaa4c0b)

dockerfiles/cli/test.sh
```diff
@@ -31,7 +31,7 @@ fi
 #   The file has to be placed in tests folder in directory containing this script
 # (Optional) second argument is options for a docker run command.
 run_test_in_docker_container() {
-  docker_exec run --rm ${DOCKER_RUN_OPTIONS} $2 \
```
riuvshin: why `$2` is removed this is used to pass some docker params like  in that test
`run_test_in_docker_container cmd_start_stop_restart_tests.bats --net=host`

davidfestal:
> @davidfestal also it should provide documentation

Done in PR https://github.com/eclipse/che-docs/pull/294

benoitf (approved):

riuvshin (approved):

garagatyi:
Nice improvement!

Issue #6284, opened 2017-09-15
[SPI] Add ability to create all user's workspaces in one OpenShift project


sleshchenko:
Now we use one OpenShift project for OpenShift objects of one workspace. But it doesn't suit for some cases. 
It is needed to add an ability to create all user's workspaces in one OpenShift project. 
So there should be implemented different project choosing strategies:
1. One OpenShift project for one workspace.
2. One OpenShift project for all Che workspaces.
There should be an ability to configure this strategy by external configuration.

To implement this it is needed to label all workspace objects which are related to one workspace with its id. To ensure that services use only pods from the same workspace should be solved by adding the same entry to Service#spec#selector field.

PR #6400, opened 2017-09-22
che-6284: Adding property for specifying project where workspaces will be created for OpenShift infrastructure

ibuziuk:
### What does this PR do?
Adding property for specifying project where workspaces will be created for OpenShift infrastructure.
Once https://github.com/eclipse/che/issues/6284 is implemented it will be possible to pick strategy of workspace creation (new namespace / predefined namespace). However, now it would be really handy to have a posibility to deploy che from `spi` on clusters where project creation is not allowed (osio).

temp solution of https://github.com/eclipse/che/issues/6284

#### Changelog
N / A

#### Release Notes
N / A 

#### Docs PR
N / A

### DIFF
--- a/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
+++ b/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
@@ -90,7 +90,6 @@ public class OpenShiftInternalRuntime extends InternalRuntime<OpenShiftRuntimeCo
   @Override
   protected void internalStart(Map<String, String> startOptions) throws InfrastructureException {
     try {
-      project.cleanUp();
 
       prepareOpenShiftPVCs(getContext().getOpenShiftEnvironment().getPersistentVolumeClaims());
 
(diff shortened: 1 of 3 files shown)


infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
```diff
@@ -90,7 +90,6 @@ public OpenShiftInternalRuntime(
   @Override
   protected void internalStart(Map<String, String> startOptions) throws InfrastructureException {
     try {
-      project.cleanUp();
```
sleshchenko, line 93: Note that all objects in the project will be stopped even when the dedicated project is configured. See https://github.com/eclipse/che/pull/6400/files#diff-fd5e0f37b880417aed392d5dc7020dd6R149

sleshchenko (approved):
OK for me as temporary solution

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
ibuziuk: Yeah, I know. That's why I defined it as a "temp" solution. Basically, having this property for now will allow at least somehow work with spi on prod cluster. And in any case I'm not sure why `project.cleannup()` is called in internal startup (I guess it is 'just in case')

garagatyi (approved):

PR #6415, opened 2017-09-25
Add possibility to disable DockerInstanceStopDetector

garagatyi:
### What does this PR do?
It is needed for OpenShift infra where getting events
is not supported due to security limits.
To disable containers stop detector variable
CHE_DOCKER_ENABLE__CONTAINER__STOP__DETECTOR=false should be
defined.

#### Changelog
Add possibility to disable DockerInstanceStopDetector

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceStopDetector.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceStopDetector.java
@@ -45,7 +45,7 @@ import org.slf4j.LoggerFactory;
 public class DockerInstanceStopDetector {
   private static final Logger LOG = LoggerFactory.getLogger(DockerInstanceStopDetector.class);
 
-  private final boolean enableDetector;
+  private final boolean isEnabled;
   private EventService eventService;
   private DockerConnector dockerConnector;
   private ExecutorService executorService;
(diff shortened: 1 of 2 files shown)


skabashnyuk (approved):

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceStopDetector.java
```diff
@@ -89,7 +97,9 @@ public DockerInstanceStopDetector(
    * @param workspaceId id of a workspace that owns machine
    */
   public void startDetection(String containerId, String machineId, String workspaceId) {
-    instances.put(containerId, Pair.of(machineId, workspaceId));
+    if (enableDetector) {
```
sleshchenko: I'd name this variable as `isEnabled`

sleshchenko (approved):
OK for me

benoitf:
there is a missing test

garagatyi:
yes

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/DockerInstanceStopDetector.java
garagatyi: ok

riuvshin (approved):

benoitf:
@garagatyi I don't see the unit test ?

benoitf:
also I don't see any labels set and milestone

garagatyi:
I didn't add tests.
I added a label. Do I have to add other labels?
As far as I can remember I haven't heard anything about the addition of milestones to PRs.

slemeur:
labels and milestone must be defined when merging a PR

garagatyi:
@slemeur which labels? 
@slemeur @benoitf I would rather not expose this variable. Basically, it is a workaround to be able to run OpenShiftConnector on OSIO. It should be removed after moving to spi

PR #6416, opened 2017-09-25
Do not perform chown and chmod on /projects unless a user is in sudoers

ghost:
### What does this PR do?

Uses and extends an existing function to avoid failed chown and chmod (when a user is not in sudoers) 

https://github.com/redhat-developer/rh-che/issues/298

### DIFF
--- a/wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
+++ b/wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
@@ -18,7 +18,7 @@ is_current_user_sudoer() {
 }
 
 set_sudo_command() {
-    if is_current_user_sudoer || is_current_user_root; then
+    if is_current_user_sudoer; then
         SUDO="sudo -E"
     else
         unset SUDO;
--- a/wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
+++ b/wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
@@ -18,7 +18,7 @@ is_current_user_sudoer() {
 }
 
 set_sudo_command() {
-    if is_current_user_sudoer; then
+    if is_current_user_sudoer && ! is_current_user_root; then
         SUDO="sudo -E"
     else
         unset SUDO;
@@ -58,8 +58,9 @@ fi
 MACHINE_TYPE=$(uname -m)
 
 mkdir -p ${CHE_DIR}
-${SUDO} mkdir -p /projects
-if [ "${SUDO}" = "sudo -E" ]; then
+
+if is_current_user_sudoer; then
+    ${SUDO} mkdir -p /projects
     ${SUDO} sh -c "chown -R $(id -u -n) /projects"
     ${SUDO} chmod 755 /projects
 fi
--- a/wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
+++ b/wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
@@ -62,9 +62,8 @@ mkdir -p ${CHE_DIR}
 if is_current_user_sudoer; then
     ${SUDO} mkdir -p /projects
     ${SUDO} sh -c "chown -R $(id -u -n) /projects"
-    ${SUDO} chmod 755 /projects
+    find /projects -type d -exec ${SUDO} chmod 755 {} \;
 fi
-
 
 INSTALL_JDK=false
 command -v ${JAVA_HOME}/bin/java >/dev/null 2>&1 || {
--- a/wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
+++ b/wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
@@ -18,7 +18,7 @@ is_current_user_sudoer() {
 }
 
 set_sudo_command() {
-    if is_current_user_sudoer && ! is_current_user_root; then SUDO="sudo -E"; else unset SUDO; fi
+    if is_current_user_sudoer && ! is_current_user_root; then SUDO="sudo -E"; else unset SUDO && CHANGE_PROJECTS_PERMISSIONS=false; fi
 }
 
 set_sudo_command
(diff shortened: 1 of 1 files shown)


l0rd (approved):
LGTM and can be merged as it is for me because it has the merit to address some annoying logs that we have since a long time. But clearly this code could be improved: `if []; then : else ...chmod and chown...` works but is not beautiful.

I would rather:

```
if is_current_user_sudoer || is_current_user_root; then
 ... chmod and chown...
fi
```

wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
```diff
@@ -55,8 +55,12 @@ MACHINE_TYPE=$(uname -m)
 
 mkdir -p ${CHE_DIR}
 ${SUDO} mkdir -p /projects
-${SUDO} sh -c "chown -R $(id -u -n) /projects"
-${SUDO} chmod 755 /projects
+if [ "${CHANGE_PROJECTS_PERMISSIONS}" = false ]; then
+    :
```
benoitf: it's looking strange to have a no-op operation in the `if` block and the code in the else block

shouldn't it better to change the condition so the logic is in the "if" block ?

tolusha (approved):

ghost:
@benoitf @l0rd is it ok for you guys? Basically, if a users isn't in sudoers chown and chmod never happen (which is always OpenShift case)

wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
```diff
@@ -18,7 +18,11 @@ is_current_user_sudoer() {
 }
 
 set_sudo_command() {
-    if is_current_user_sudoer && ! is_current_user_root; then SUDO="sudo -E"; else unset SUDO; fi
+    if is_current_user_sudoer; then
```
l0rd: Changing this is an error. Function `is_current_user_sudoer` returns `true` if the user is `root`. As a consequence `SUDO` is set to `sudo -E"` when the script is run by `root` user. The correct behavior should be to `unset SUDO` if user is `root`. 

By the way the goal of your PR is to avoid executing `chown` and `chmod` if the user is arbitrary. For this you don't need to change this function (`set_sudo_command`).

benoitf:
@eivantsov yes

but when I read the previous command line it's looking strange as well
```
${SUDO} mkdir /projects
```
at first it let me think it will try to use SUDO as well (but the variable is unset)
(it's more on the variable name as it's a more a prefix than SUDO but that's a different story)

about your PR, does it work if we only check if SUDO variable is set to something nonnull/not empty ?
`if [ -n "$SUDO" ]`
instead of checking if it's "sudo -E"

ghost:
I think it will work - it's just a different way to check it.

And yes, moving mkdir to if clause makes sense.

wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
```diff
@@ -55,8 +59,10 @@ MACHINE_TYPE=$(uname -m)
 
 mkdir -p ${CHE_DIR}
 ${SUDO} mkdir -p /projects
-${SUDO} sh -c "chown -R $(id -u -n) /projects"
-${SUDO} chmod 755 /projects
+if [ "${SUDO}" = "sudo -E" ]; then
```
l0rd: Here you need to check if the user has `sudoer` privileges.  Why don't you just use `if is_current_user_sudoer; then`. That would be more easy to read.

l0rd (changes requested):

wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
```diff
@@ -54,9 +58,12 @@ fi
 MACHINE_TYPE=$(uname -m)
 
 mkdir -p ${CHE_DIR}
-${SUDO} mkdir -p /projects
-${SUDO} sh -c "chown -R $(id -u -n) /projects"
-${SUDO} chmod 755 /projects
+
+if is_current_user_sudoer; then
+    ${SUDO} mkdir -p /projects
```
benoitf: is it ok if /projects is only created for sudoers ?

wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
```diff
@@ -54,9 +58,12 @@ fi
...
+    ${SUDO} sh -c "chown -R $(id -u -n) /projects"
+    ${SUDO} chmod 755 /projects
+fi
```
benoitf: in che-server image we updated only directory permissions

would it work as well if we only change permissions of folder but not the owner ?
(it can make the operation faster)

ghost: Do you mean this? https://github.com/eclipse/che/blob/master/dockerfiles/che/Dockerfile#L43

benoitf: @eivantsov yes

wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
ghost: I think by the time this script runs `/projects` is already there since wither Docker or OpenShift creates it when the container starts/is created.

wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
benoitf: I was thinking replacing 2 instructions chown + chmod with the only find + exec

wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
l0rd: @benoitf in any case if the user is not a sudoer he cannot create the folder `/projects/`.

l0rd (approved):

wsagent/agent/src/main/resources/org.eclipse.che.ws-agent.script.sh
benoitf: @l0rd user can be `root` user

Issue #6258, opened 2017-09-13
OpenShift infrastructure: add TLS support of routes


ibuziuk:
Currently OpenShift infrastructure does not support TLS which is one of the things that prevents 'spi' to be working on osio. In order to enable TLS the following steps need to be done:
- Having property `che.infra.openshift.secure.routes` for enabling TLS during route creation
- Implement `URLRewriter` which would pick suitable protocol depending on `che.infra.openshift.secure.routes` property
- On OSIO custom router is used which completely ignores the 'host' field and assumes that routes host is supposed to have the following format `$NAME.$NAMESPACE` where:
$NAME = route name
$NAMESPACE = namespace where route has been created.
In theory this supposed to work OOTB, since while route creation the 'host' is not used while route creation, but it still fails for me on osio (looks like there is extra 'che-' path in the route which results in 503 error during workspace creation)

ibuziuk:
Currently can not start workspace on osio. Fails with the following message:
`WebSocket connection to 'wss://servers-app-main-server-4401-ibuziuk-che.8a09.starter-us-east-2.openshiftapps.com/wsagent?clientId=1312596289' failed: Error during WebSocket handshake: Unexpected response code: 503`

garagatyi:
@ibuziuk do you think we should use TLS for user's application too?

ibuziuk:
![image](https://user-images.githubusercontent.com/1461122/30420071-c9509d92-9938-11e7-8580-6aa6bc4edd43.png)
yeah, all the routes including those that would be used by user's app would be secure

garagatyi:
What if user's application is not configured to use https behind the proxy that sends http traffic instead of https?

ibuziuk:
@garagatyi could you provide some example where enabling TLS for route that is used for apps preview will break it? Basically, for routes `insecureEdgeTerminationPolicy` *Redirect* is used which would redirect http requests to https (in our rhche stacks run / debug commands are using http scheme but once user is trying to open it in the browser there would be force redirect to https)

ibuziuk:
![image](https://user-images.githubusercontent.com/1461122/30441882-00e38ac0-997b-11e7-8620-c19980e729f4.png)
Was able to start workspace on prod cluster with tls enabled, but for some reason ws-agent dies after a few seconds after workspace startup with the following error:

```
[...]

PR #6424, opened 2017-09-25
che #6258: Adding TLS support for OpenShift routes

ibuziuk:
### What does this PR do?
Adding TLS support for OpenShift routes

https://github.com/eclipse/che/issues/6258

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
@@ -398,6 +398,9 @@ che.infra.openshift.username=
 che.infra.openshift.password=
 che.infra.openshift.trust_certs=
 
+# Create routes with Transport Layer Security (TLS) enabled
+che.infra.openshift.tls_enabled=false
+
 # Defines OpenShift namespace in which all workspaces will be created.
 # If not set, every workspace will be created in a new project, where project name = workspace id
 che.infra.openshift.project=
--- a/infrastructures/openshift/src/test/java/org/eclipse/che/workspace/infrastructure/openshift/provision/route/TlsRouteProvisionerTest.java
+++ b/infrastructures/openshift/src/test/java/org/eclipse/che/workspace/infrastructure/openshift/provision/route/TlsRouteProvisionerTest.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2012-2017 Red Hat, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.workspace.infrastructure.openshift.provision.route;
+
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import io.fabric8.openshift.api.model.Route;
+import java.util.HashMap;
+import java.util.Map;
+import org.eclipse.che.api.core.model.workspace.runtime.RuntimeIdentity;
+import org.eclipse.che.api.workspace.server.spi.InternalEnvironment;
+import org.eclipse.che.workspace.infrastructure.openshift.environment.OpenShiftEnvironment;
+import org.mockito.Mock;
+import org.mockito.testng.MockitoTestNGListener;
+import org.testng.annotations.Listeners;
+import org.testng.annotations.Test;
+
+/**
+ * Tests {@link TlsRouteProvisioner}.
+ *
+ * @author Ilya Buziuk
+ */
+@Listeners(MockitoTestNGListener.class)
+public class TlsRouteProvisionerTest {
+
+  @Mock private InternalEnvironment environment;
+  @Mock private OpenShiftEnvironment osEnv;
+  @Mock private RuntimeIdentity runtimeIdentity;
+
+  @Test
+  public void doNothingWhenTlsDisabled() throws Exception {
+    TlsRouteProvisioner tlsProvisioner = new TlsRouteProvisioner(false);
+    tlsProvisioner.provision(environment, osEnv, runtimeIdentity);
+    verify(osEnv, never()).getRoutes();
+  }
+
+  @Test
+  public void provisionTlsForRoutes() throws Exception {
+    TlsRouteProvisioner tlsProvisioner = new TlsRouteProvisioner(true);
+    final Map<String, Route> routes = new HashMap<>();
+    when(osEnv.getRoutes()).thenReturn(routes);
+    tlsProvisioner.provision(environment, osEnv, runtimeIdentity);
+    verify(osEnv, times(1)).getRoutes();
+  }
+}
--- a/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInfrastructureProvisioner.java
+++ b/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInfrastructureProvisioner.java
@@ -18,6 +18,7 @@ import org.eclipse.che.api.workspace.server.spi.InternalEnvironment;
 import org.eclipse.che.workspace.infrastructure.openshift.environment.OpenShiftEnvironment;
 import org.eclipse.che.workspace.infrastructure.openshift.provision.UniqueNamesProvisioner;
 import org.eclipse.che.workspace.infrastructure.openshift.provision.installer.InstallerConfigProvisioner;
+import org.eclipse.che.workspace.infrastructure.openshift.provision.route.TlsRouteProvisioner;
 import org.eclipse.che.workspace.infrastructure.openshift.provision.volume.PersistentVolumeClaimProvisioner;
 
 /**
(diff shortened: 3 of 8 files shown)


infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/ServerExposer.java
```diff
@@ -178,6 +181,11 @@ public void expose(String namePrefix, Map<String, ? extends ServerConfig> server
     return exposedPorts;
   }
 
+  private boolean isTlsEnabled() {
+    String isTlsEnabled = System.getenv("CHE_INFRA_OPENSHIFT_ENABLE__TLS");
```
garagatyi: -1 to use property in such a way. In this case, property added though property file won't work. Can you inject it instead?

sleshchenko: You should inject value as named property instead of getting environment variable. I understand why you implement it in this way =) So I'd move this logic to provision step, see [OpenShiftInfrastructureProvisioner](https://github.com/eclipse/che/blob/spi/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInfrastructureProvisioner.java#L30). Then you wouldn't have any problem with property injection.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/TlsURLRewriter.java
```diff
@@ -0,0 +1,57 @@
...
+package org.eclipse.che.workspace.infrastructure.openshift;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import org.eclipse.che.api.core.model.workspace.runtime.RuntimeIdentity;
+import org.eclipse.che.api.workspace.server.URLRewriter;
+import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
+
+/** @author Ilya Buziuk */
+public class TlsURLRewriter implements URLRewriter {
```
sleshchenko: please add tests for this class

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -395,6 +395,9 @@ che.infra.openshift.username=
 che.infra.openshift.password=
 che.infra.openshift.trust_certs=
 
+# Create routes with TLS enabled
+che.infra.openshift.enable_tls=
```
sleshchenko: It would be better to set default value here

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/TlsURLRewriter.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+package org.eclipse.che.workspace.infrastructure.openshift;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import org.eclipse.che.api.core.model.workspace.runtime.RuntimeIdentity;
+import org.eclipse.che.api.workspace.server.URLRewriter;
+import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
+
+/** @author Ilya Buziuk */
```
garagatyi: Can you briefly describe the purpose of this class in the javadocs

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/ServerExposer.java
ibuziuk: thanks for the pointer. lemme try

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/TlsURLRewriter.java
ibuziuk: sure

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
ibuziuk: sounds good

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/TlsURLRewriter.java
ibuziuk: yeah, I will ;-)

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/ServerExposer.java
ibuziuk: ok. lemme try to handle DI better

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/TlsURLRewriter.java
```diff
@@ -0,0 +1,57 @@
...
+  @Inject
+  TlsURLRewriter(@Named("che.infra.openshift.enable_tls") boolean isTlsEnabled) {
+    this.isTlsEnabled = isTlsEnabled;
+  }
+
+  @Override
+  public String rewriteURL(final RuntimeIdentity identity, final String name, final String url)
+      throws InfrastructureException {
+    String rewrittenURL = url;
+    if (isTlsEnabled) {
```
garagatyi: Consider next refactoring:
```
if (isTlsEnabled) {
    rewrittenURL = rewriteURL(url);
}
```
```
private String rewriteURL(final String url) {
    if (isWebSocketURL(url)) {
        url.replaceFirst("^ws:", "wss:");
    } else {
        url.replaceFirst("^http:", "https:");
    }
}
```

garagatyi: This might make code easier for understanding

ibuziuk: true

akorneta (approved):

ibuziuk:
PR has been updated. @garagatyi @sleshchenko could you please take another look ?

sleshchenko:
Maybe, It would be better to remove URL rewriter here and keep all logic related to TLS in one place. Please take a look my https://github.com/sleshchenko/che/commit/bbade68937b6d404733148946bb6fbd29cc40423 draft
@ibuziuk @garagatyi WDYT?

ibuziuk:
@sleshchenko I do not have strong opinion on this. On  one hand, the property would be used only in one class, which is good from the single responsibility principle perspective, but on the other hand the whole point of `UrlRewriter` becomes misleading IMO

garagatyi:
@sleshchenko +1. In case we need URL rewriter for some hacky infrastructure like Docker on OSX we would need to combine workaround for such system and TLS enabling.
And keeping TLS enabling feature in a single class is also valuable for supporting the code

ibuziuk:
ok, makes sense. Let me update the code

ibuziuk:
@sleshchenko PR has been updated, but I'm not sure I understand how the trick with labels supposed to work. 
image
Routes are created  with tls enabled but communication still happens via insecure protocols. Could you elaborrate on this a bit more ?

sleshchenko:
I've made assembly from your branch and:
- Routes are created with TLS enabled;
- Routes have annotations with correct protocols for agents;
- Workspace API returns workspace with correct links to agents;
- Servers checks use correct links.
BUT
Workspace fails to start because servers checks hang out. Maybe it's because of my minishift instance.

So, make sure that you are using correct binary, and then check items which I've described above.

P.S. Servers are resolved here https://github.com/eclipse/che/blob/7cb8a224fcba78f0ed14f1b7b8b282971d3ab57f/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/ServerResolver.java#L67-L74, so If Routes have correct protocols for agents, Workspace API should return correct links which are used for IDE and Dashboard.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -398,6 +398,9 @@ che.infra.openshift.username=
 che.infra.openshift.password=
 che.infra.openshift.trust_certs=
 
+# Create routes with Transport Layer Security (TLS) enabled
+che.infra.openshift.enable_tls=false
```
sleshchenko: I'd rename it to `che.infra.openshift.tls_enabled`

ibuziuk: ok, will do

ibuziuk:
@sleshchenko thanks for your feedback. So, when you are starting workspace you do not have 503s in the browser console?

sleshchenko:
@ibuziuk No, I have not. Please check annotations of created Routes. Are there right values for protocol?

ibuziuk:
@sleshchenko yeah, annotations use the right protocol and it looks like I just had dirty setup. So, if you are ok we could merge this one and I would provide updated scripts & fixes for deployment on osio in the upcoming PR

infrastructures/openshift/src/test/java/org/eclipse/che/workspace/infrastructure/openshift/provision/route/TlsRouteProvisionerTest.java
```diff
@@ -0,0 +1,53 @@
...
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.workspace.infrastructure.openshift.provision.route;
+
+import static org.mockito.Mockito.*;
```
sleshchenko: We shouldn't use import with *

sleshchenko (approved):
Looks good for me.

infrastructures/openshift/src/test/java/org/eclipse/che/workspace/infrastructure/openshift/provision/route/TlsRouteProvisionerTest.java
ibuziuk: fixed

ibuziuk:
ci-build

riuvshin:
@ibuziuk can you please trigger build again, I've just updated whitelist :)

riuvshin (approved):

ibuziuk:
@riuvshin thanks!

ibuziuk:
ci-build

ibuziuk:
ci-build

ibuziuk:
@riuvshin hmm.. looks like build is not triggered after my commands ;-)
https://ci.codenvycorp.com/job/che-pullrequests-build/

riuvshin:
@ibuziuk yeah... seems whitelist based not on usernames... I will check

riuvshin:
ci-build

ibuziuk:
> 23:04:03 FileNotFoundException means that the credentials Jenkins is using is probably wrong. Or the user account does not have write access to the repo.
> 23:04:03 java.io.FileNotFoundException: {"message":"Not Found","documentation_url":"https://developer.github.com/v3/repos/statuses/#create-a-status"}

@riuvshin build is still failing - looks like some jenkins config issue

garagatyi (approved):

skabashnyuk (approved):

PR #6441, opened 2017-09-26
Multi-user Eclipse Che

skabashnyuk:
### steps required before merge
- [x] CQ approval of https://dev.eclipse.org/ipzilla/show_bug.cgi?id=14132

### What does this PR do?
Introduce multi-user Eclipse Che based on Keyclaok identity server.

#### What is it
- Added authentification and user management provided by Keycloak  http://www.keycloak.org/
- Permissions and Organization ported from Codenvy https://codenvy.com/docs/user-guide/permissions/index.html
 

#### How to run it.
```docker run -it -e CHE_MULTIUSER=true -e CHE_HOST=<your ip> -e CHE_KEYCLOAK_AUTH__SERVER__URL=http://<your ip>:5050/auth -v /var/run/docker.sock:/var/run/docker.sock -v ~/.che-multiuser:/data eclipse/che:nightly start --skip:pull --skip:nightly```
#### How to manage it
 - Keycloak configured with two realms. ```Master``` and ```che```. Also we have one user admin/admin in both realm. Admin user in master realm is  - super admin. 
-  Eclipse Che configured for che realm
- We enabled user registration in ```Che``` realm
#### Known limitation
 - swagger would not work. We need to upgrade a version. to support openid authentification https://github.com/eclipse/che/issues/6015
- It's working on local docker. We are going to provide scalable version based on OpenShift on next versions.
- Invitation of non-existent users to Eclipse Che organization https://github.com/eclipse/che/issues/6335
#### How to run it when it is in a branch

To run an multiuser Che version, the following steps are required after building the branch:
 - Rebuild init, cli and che images (in the given sequence). To do that, proceed to folder _dockerfiles/<image_name>_ and run _build.sh_
 - Run Che in a  usual way using cli, with additional parameters:  `-e CHE_MULTIUSER=true` and `--skip:pull --skip:nightly`  
   Full command example:
   `docker run -it --rm -v /var/run/docker.sock:/var/run/docker.sock -v /home/user/.che:/data -e CHE_MULTIUSER=true eclipse/che-cli:nightly start --skip:pull --skip:nightly`
 - MacOS users may need to edit _che.env_ file in the data folder, changing `CHE_HOST` and `CHE_KEYCLOAK_AUTH__SERVER__URL` values to their specific IP.
 
When start is succeeded, the following docker containers should be created:  
 - che, exposing 8080 port;
 - che_keycloak, exposing 5050 port;  
 - che_postgres, exposing 5432 port; 

#### Changelog
Introduce multi-user Eclipse Che based on Keyclaok identity server.

#### Release Notes
Introduce multi-user Eclipse Che based on Keyclaok identity server.

#### Docs PR
https://github.com/eclipse/che-docs/pull/296

### CODE REVIEW

**Comment by skabashnyuk** [comment]
We are still in progress. So no sense for code review yet. But if you have any comments about the procedure of merge - please comment.

**Comment by benoitf** [comment]
I have a single comment
 It's about the process of submitting another big PR with a very long live branch
Why do we need to add 600files at once ?
why can't we add feature step by step ? with short live branches
Extend the current API, allow pluggability of components, etc.

**Comment by skabashnyuk** [comment]
I've almost agreed with you with one except. We don't know for sure at the beginning when it will be merged

**Comment by benoitf** [comment]
I would say that we don't need to know when a feature will be merged to introduce code changes.
Code can be there and not enabled at all as well

File: assembly-multiuser/assembly-wsmaster-war/src/main/webapp/WEB-INF/web.xml [hunk 1]
```diff
@@ -0,0 +1,63 @@
...
+
+    <filter>
+        <filter-name>guiceFilter</filter-name>
+        <filter-class>com.google.inject.servlet.GuiceFilter</filter-class>
+    </filter>
+    <filter-mapping>
+        <filter-name>guiceFilter</filter-name>
+        <url-pattern>/*</url-pattern>
+    </filter-mapping>
+    <security-role>
```
benoitf: AFAIK this is not needed

core/commons/che-core-commons-auth/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+    <packaging>jar</packaging>
+    <name>Che Core :: Commons :: Auth</name>
+    <dependencies>
+        <dependency>
+            <groupId>javax.inject</groupId>
+            <artifactId>javax.inject</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>javax.servlet</groupId>
+            <artifactId>javax.servlet-api</artifactId>
```
benoitf: it should be scope provided

core/commons/che-core-commons-auth/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+        <groupId>org.eclipse.che.core</groupId>
+        <version>5.19.0-SNAPSHOT</version>
+    </parent>
+    <artifactId>che-core-commons-auth</artifactId>
+    <packaging>jar</packaging>
+    <name>Che Core :: Commons :: Auth</name>
+    <dependencies>
+        <dependency>
+            <groupId>javax.inject</groupId>
+            <artifactId>javax.inject</artifactId>
```
benoitf, line 27: scope provided

core/commons/che-core-commons-mail/pom.xml
```diff
@@ -0,0 +1,102 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>che-core-commons-parent</artifactId>
+        <groupId>org.eclipse.che.core</groupId>
+        <version>5.19.0-SNAPSHOT</version>
+    </parent>
+    <artifactId>che-core-commons-mail</artifactId>
+    <packaging>jar</packaging>
+    <name>Che Core :: Commons :: Mail sender</name>
+    <dependencies>
```
benoitf, line 24: missing provided scope for API

core/commons/che-core-commons-mail/src/main/java/org/eclipse/che/mail/MailSender.java
```diff
@@ -0,0 +1,162 @@
...
+        message.setReplyTo(InternetAddress.parse(emailBean.getReplyTo()));
+      }
+      LOG.info(
+          "Sending from {} to {} with subject {}",
+          emailBean.getFrom(),
+          emailBean.getTo(),
+          emailBean.getSubject());
+
+      Transport.send(message);
+      LOG.debug("Mail send");
```
benoitf: Mail sent ?

dashboard/src/index.html
```diff
@@ -11,7 +11,7 @@
 
 -->
 <!doctype html>
-<html class="no-js" ng-app="userDashboard" ng-strict-di>
+<html class="no-js">
```
benoitf: we don't need anymore ng-app ?

dockerfiles/che/build.sh
```diff
@@ -17,15 +19,41 @@ if [ ! -d "${DIR}/../../assembly/assembly-main/target" ]; then
...
-LOCAL_ASSEMBLY_ZIP="${DIR}"/eclipse-che.tar.gz
+BUILD_ASSEMBLY_DIR=$(echo "${DIR}"/../../assembly/assembly-main/target/eclipse-che-*/eclipse-che-*/)
+LOCAL_ASSEMBLY_DIR="${DIR}"/eclipse-che
+
+if [ -d "${LOCAL_ASSEMBLY_DIR}" ]; then
+  rm -r "${LOCAL_ASSEMBLY_DIR}"
+fi
+
+echo "Copying assembly ${BUILD_ASSEMBLY_DIR} --> ${LOCAL_ASSEMBLY_DIR}"
+cp -r "${BUILD_ASSEMBLY_DIR}" "${LOCAL_ASSEMBLY_DIR}"
```
benoitf: why do we need to copy files instead of using the files directly

multiuser/api/che-multiuser-api-permission/pom.xml
```diff
@@ -0,0 +1,306 @@
...
+            <artifactId>javax.annotation-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>javax.inject</groupId>
+            <artifactId>javax.inject</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>javax.ws.rs</groupId>
+            <artifactId>javax.ws.rs-api</artifactId>
+        </dependency>
```
benoitf, line 59: missing provided scope on API

multiuser/permission/che-multiuser-permission-user/pom.xml
```diff
@@ -0,0 +1,148 @@
...
+    </properties>
+    <dependencies>
+        <dependency>
+            <groupId>javax.inject</groupId>
+            <artifactId>javax.inject</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>javax.ws.rs</groupId>
+            <artifactId>javax.ws.rs-api</artifactId>
+        </dependency>
```
benoitf, line 34: missing provided scope on API

dockerfiles/che/build.sh
benoitf: I'm kind of -1 as it takes a lot of time to copy

benoitf (dismissed):

dashboard/bower.json
```diff
@@ -14,6 +14,7 @@
     "angular-charts": "0.2.6",
     "angular-cookies": "1.4.8",
     "angular-dropdowns": "1.0.0",
+    "angular-gravatar": "0.2.4",
```
benoitf, line 17: what is the associate CQ ?

assembly-multiuser/Readme-multiuser.md
```diff
@@ -0,0 +1,12 @@
+### Getting Started with multiuser Che
```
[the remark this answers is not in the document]
skabashnyuk: I'll remove this file. Content added to PR description.

assembly-multiuser/Readme-multiuser.md
```diff
@@ -0,0 +1,12 @@
...
+To run an multiuser Che version, the following steps are required after building the branch:
```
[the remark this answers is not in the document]
skabashnyuk: Yes, for now, when we in a branch we need to rebuild it .

core/commons/che-core-commons-mail/pom.xml
skabashnyuk: Are you talking about javax.mail?

skabashnyuk:
>Could you tell me more about why the private realm has been removed ?

@davidfestal 
We don't use it. Am I wrong?

davidfestal:
> > Could you tell me more about why the private realm has been removed ?

> We don't use it. Am I wrong?

I had thought they were used with the dedicated Keycloak server to allow *creating new users* (which certainly requires more rights that simply authenticating). But I might be the one that is wrong.

skabashnyuk:
@davidfestal AFAIK we don't create new users from our code

dashboard/bower.json
skabashnyuk: https://dev.eclipse.org/ipzilla/show_bug.cgi?id=14132

dockerfiles/che/build.sh
riuvshin: the reason why I did this is because `ln` does not work with folders so I can't just create a link as it was anymore

riuvshin: btw you approved this before https://github.com/eclipse/che/pull/5880

core/commons/che-core-commons-mail/pom.xml
benoitf: javax.inject, javax.annotation api, 
about mail usually only if an implementation is required we provide `mail` artifact  (that contains the implementation (like for the test scope)
else the `javax.mail-api` artifact is used for API calls

dockerfiles/che/build.sh
benoitf: AFAIK `ln` is working with directories.

```
$ mkdir foo
$ touch foo/dummy
$ ln -s foo bar
$ ls -la bar/
-rw-r--r--   1 benoitf  wheel    0 27 sep 11:05 dummy
```

skabashnyuk:
@benoitf let's discuss a state with ```provided``` scope of some dependencies.
AFAIK A single know to me dependency that is provided by Tomcat javax.servlet-api. So no doubts that it have to be provided.
javax.ws.rs and javax.mail are not provided by default. We may try to base on the fact that mail jar is provided by our tomcat. But I'm not sure this is a good idea.
I would say they are both compile dependencies

benoitf:
@skabashnyuk sure, let skip all my comments around provided scope. We can iterate on another PR

dashboard/bower.json
benoitf: thx, I've updated the description

dockerfiles/che/build.sh
riuvshin: yes, but symbolic link does not work with dockerfile `ADD` or `COPY` instructions. That was the issue. And you cant make not symbolic link to a folder

benoitf:
```
$ more Dockerfile 
FROM centos
ADD bar /tmp/bar


$ docker build -t fb/test1 .
Sending build context to Docker daemon  3.584kB
Step 1/2 : FROM centos
 ---> 328edcd84f1b
Step 2/2 : ADD bar /tmp/bar
 ---> 5e99318d01fa
Successfully built 5e99318d01fa
Successfully tagged fb/test1:latest

$ docker run -ti -exec fb/test1 ls -la /tmp/bar/
total 8
drwxr-xr-x 2 root root 4096 Sep 27 09:42 .
drwxrwxrwt 1 root root 4096 Sep 27 09:42 ..
-rw-r--r-- 1 root root    0 Sep 27 09:10 dummy
```

benoitf: ok, when you said "symbolic link" you said "symbolic link to an external folder"

riuvshin: $ mkdir -p src/2
$ touch src/test && touch src/2/test2
$ ln -s src dst

Dockerfile
```
FROM alpine
ADD dst /tmp/dst
```
build `docker build -t fb/test1 .`

```

Step 1/2 : FROM alpine
latest: Pulling from library/alpine
88286f41530e: Pull complete
Digest: sha256:f006ecbb824d87947d0b51ab8488634bf69fe4094959d935c0c103f4820a417d
Status: Downloaded newer image for alpine:latest
 ---> 76da55c8019d
Step 2/2 : ADD dst /tmp/dst
ADD failed: stat /var/lib/docker/tmp/docker-builder623285034/Users/roman/development/codenvy_projects/test/src: no such file or directory
```

riuvshin: external folder, yes, sorry for confusion. So that is why `cp` was used

assembly-multiuser/assembly-wsagent-war/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-war-plugin</artifactId>
+                <configuration>
+                    <attachClasses>true</attachClasses>
+                    <packagingExcludes>WEB-INF/lib/*gwt*.jar,
+                        WEB-INF/lib/gin-*.jar,
+                        WEB-INF/lib/jsr305*.jar,
+                        WEB-INF/classes/org/eclipse/che/wsagent/server/CheWsAgentServletModule.class</packagingExcludes>
```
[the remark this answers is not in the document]
skabashnyuk: Agree it's not ideal. As a variant, this che single user aka sdk assembly specific files can be moved to separate module.  But is this a blocker?

assembly-multiuser/assembly-wsmaster-war/pom.xml
```diff
@@ -0,0 +1,172 @@
...
+                            <targetPath>WEB-INF</targetPath>
+                            <includes>
+                                <include>**/*</include>
+                            </includes>
+                        </resource>
+                    </webResources>
+                    <packagingExcludes>WEB-INF/lib/wsmaster-local*.jar,
+                        WEB-INF/lib/che-core-db-vendor-h2-*.jar,
+                        WEB-INF/classes/org/eclipse/che/api/deploy/CheWsMasterModule.class,
+                        WEB-INF/classes/org/eclipse/che/api/deploy/CheWsMasterServletModule.class</packagingExcludes>
```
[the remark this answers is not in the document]
skabashnyuk: I think this is similar to https://github.com/eclipse/che/pull/6441#discussion_r141306509. and has the same solution.

dashboard/src/index.html
skabashnyuk: @davidfestal can you comment. Why it was removed? https://github.com/redhat-developer/rh-che/blame/master/assembly/dashboard/src/main/patches/src/index.html.patch#L7 ?

davidfestal: It was like that from the beginning of the Keycloak support in the dashboard (initial draft was made by @gorkem).
I think the reason is that angular is now bootstrapped *after* the keycloak initialization, as shown here: https://github.com/redhat-developer/rh-che/blob/master/assembly/dashboard/src/main/patches/src/app/index.module.ts.patch#L31

assembly-multiuser/assembly-wsagent-war/pom.xml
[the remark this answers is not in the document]
benoitf: no it is not

dashboard/src/index.html
skabashnyuk: @davidfestal @benoitf @ashumilova @akurinnoy do you see any blockers if we revert this change?

dockerfiles/che/build.sh
skabashnyuk: @benoitf @riuvshin Are you agree with this solution? Or you have some alternative?

benoitf: an alternative would be to keep symbolic link on the file but adding a symlink inside the image
I wanted to check if running the build command twice was using the cache or not

dashboard/src/index.html
benoitf: I don't know if keycloak need to be initialized before but I'm fine in both cases. But it would be great to not have this lazy loading in index.module.ts

davidfestal: why do you want to revert it ?
it is consistent with the fact that angular is bootstrapped after keycloak initialization also here: https://github.com/eclipse/che/pull/6441/files#diff-23910a1448372bee6229c74377877dfdR112

I'm not sure it would work if you revert this change.

davidfestal: it was part of the design of the keycloak integration to the dashboard from the very beginning. @gorkem wdyt ?

skabashnyuk: Ok. Based on this https://github.com/eclipse/che/pull/6441#discussion_r141333945 and if there are no major objections let's keep it.

benoitf: IMHO I still think we don't require to do a manual initialization of angular

dashboard/src/app/index.module.ts
```diff
@@ -36,12 +36,81 @@ import IdeIFrameSvc from './ide/ide-iframe/ide-iframe.service';
...
+});
+promise.then((keycloakSettings: any) => {
+  keycloakAuth.config = buildKeycloakConfig(keycloakSettings);
+
+  // load Keycloak
+  return keycloakLoad(keycloakSettings).then(() => {
+    // init Keycloak
+    return keycloakInit(keycloakAuth.config);
+  }).then((keycloak: any) => {
+    keycloakAuth.isPresent = true;
```
benoitf, line 106: why do we need to store values in keycloakAuth object as it seems we don't keep a reference to that object ?

dashboard/src/app/index.module.ts
```diff
@@ -36,12 +36,81 @@ import IdeIFrameSvc from './ide/ide-iframe/ide-iframe.service';
...
+    keycloakAuth.keycloak = keycloak;
+    /* tslint:disable */
+    window['_keycloak'] = keycloak;
+    /* tslint:enable */
+  });
+}).catch((error: any) => {
+  console.error('Keycloak initialization failed with error: ', error);
+}).then(() => {
+  angular.bootstrap(document.body, ['userDashboard'], {strictDi: true}); // manually bootstrap Angular
```
benoitf: can this logic be moved instead to the routeProvider initialization
like
https://github.com/eclipse/che/blob/fb48c3d5d6942bd9fcc02f23f511c5b18743b35a/dashboard/src/app/index.module.ts#L118-L135

where we get user preferences before the route changes

it would be more compliant with the existing solution used

dashboard/src/app/index.module.ts
skabashnyuk: @davidfestal @akurinnoy can you comment?

dashboard/src/app/index.module.ts
skabashnyuk: @davidfestal @akurinnoy can you comment?

dashboard/src/app/index.module.ts
akurinnoy: @benoitf this object is referenced above:
```initModule.constant('keycloakAuth', keycloakAuth);```
and used by Keycloak interceptor, see [here](https://github.com/eclipse/che/blob/fb48c3d5d6942bd9fcc02f23f511c5b18743b35a/dashboard/src/components/interceptor/interceptor-config.ts) and [here](https://github.com/eclipse/che/blob/fb48c3d5d6942bd9fcc02f23f511c5b18743b35a/dashboard/src/components/interceptor/keycloak-token-interceptor.ts)

File: dashboard/src/app/index.module.ts (hunk 16 above)
Comment (akurinnoy) [reply dashboard/src/app/index.module.ts]: @benoitf AFAIK no
because application bootstrapping starts routeProvider initialization

File: dashboard/src/app/index.module.ts (hunk 15 above)
Comment (benoitf) [reply dashboard/src/app/index.module.ts:106]: yes sorry I failed with the text window of the review

File: dashboard/src/app/index.module.ts (hunk 16 above)
Comment (benoitf) [reply dashboard/src/app/index.module.ts]: but routeProvider initialization can be delayed. This is what is done in the snippet

Comment (benoitf) [reply dashboard/src/app/index.module.ts]: > resolve - {Object.<string, Function>=} - An optional map of dependencies which should be injected into the controller. If any of these dependencies are promises, the router will wait for them all to be resolved or one to be rejected before the controller is instantiated. If all the promises are resolved successfully, the values of the resolved promises are injected and $routeChangeSuccess event is fired. If any of the promises are rejected the $routeChangeError event is fired. For easier access to the resolved dependencies from the template, the resolve map will be available on the scope of the route, under $resolve (by default) or a custom name specified by the resolveAs property (see below). This can be particularly useful, when working with components as route templates.

**Comment by skabashnyuk** [comment]
Does anybody object if a process of merging will be made with squash? I don't see a reason to keep history.

File: dockerfiles/build.sh [hunk 17]
```diff
@@ -34,7 +34,7 @@ build_directory() {
   shift
   # Calling build.sh
   if [ -e ${directory}/build.sh ] ; then
-    ${directory}build.sh ${OPTIONS} ${ARGS}
+    bash ${directory}build.sh ${OPTIONS} ${ARGS}
```
benoitf: why do we have bash there ? shebang is sh so we shouldn't require bash

dockerfiles/che/build.sh
```diff
@@ -5,6 +5,8 @@
 # which accompanies this distribution, and is available at
 # http://www.eclipse.org/legal/epl-v10.html
 
+set -e
```
benoitf: set -e could be removed

dockerfiles/build.sh
riuvshin: without this it was failed on build che server, idk why but it helped

benoitf: if build.sh is not compliant we should fix it instead of using bash there.

assembly-multiuser/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/multiuser.properties
```diff
@@ -0,0 +1,84 @@
+#
```
benoitf: the path of this file should be within "WEB-INF/classes/che", not codenvy

skabashnyuk: done

dockerfiles/build.sh
riuvshin: I will recheck this

riuvshin: @benoitf here is PR https://github.com/eclipse/che/pull/6513 with proper fixes

bmicklea (approved):

ashumilova (approved):

akurinnoy (approved):

benoitf:
-0
adding keycloak support and having multi-user support is awesome. I love that.
But, there are too many duplicates in the code and it's kind hard to maintain as it's based on some include/exclude/duplicates between the two assemblies.
Besides that, it will be hard for onboarding users as they'll to figure out which assembly they need to use and it's not clear how you could migrate from one to another

skabashnyuk:
Good point @benoitf. We received a couple more feedback like yours. And we have an agreement to have a single assembly in next releases. It's going to improve onboarding and probably remove duplication problems you've mentioned.

davidfestal (approved):

vparfonov (approved):

garagatyi (approved):

multiuser/api/che-multiuser-api-permission/src/main/java/org/eclipse/che/multiuser/api/permission/server/InstanceParameterValidator.java
```diff
@@ -0,0 +1,55 @@
...
+   * @throws BadRequestException if specified {@code instance} is not valid
+   * @throws NotFoundException if specified {@code domain} is unsupported
+   */
+  public void validate(String domain, @Nullable String instance)
+      throws BadRequestException, NotFoundException {
+    checkArgument(domain != null, "Domain id required");
+    if (permissionsManager.getDomain(domain).isInstanceRequired() && instance == null) {
+      throw new BadRequestException("Specified domain requires non nullable value for instance");
+    }
+  }
```
mshaposhnik, line 48: @sleshchenko simplify proposal:
```
  checkArgument(domain != null, "Domain id required");
  checkArgument((permissionsManager.getDomain(domain).isInstanceRequired() && instance == null),
        "Specified domain requires non nullable value for instance"
    );
```

sleshchenko: It can be rewritten with `checkArgument` method. But then condition should be rewritten as
```
checkArgument(!permissionsManager.getDomain(domain).isInstanceRequired() || instance != null), "Specified domain requires non nullable value for instance");
```
So I'd not change it because then condition becomes more complicated.

mshaposhnik: We can remove negation in the `checkArgument` so than condition will be the same (and `domain == null` in first check )

sleshchenko:
This method is named after Precondition from guava. And I think it is
logical to have negation there while it is named in this way.

6 жовт. 2017 р. 14:01 "Max Shaposhnik" <notifications@github.com> пише:

multiuser/api/che-multiuser-api-permission/src/main/java/org/eclipse/che/multiuser/api/permission/server/InstanceParameterValidator.java
sleshchenko: This method is named after Precondition#checkArgument from guava. And I think it is
logical to have negation there while it is named in this way. Up to you change it or not =)

bmicklea:
@skabashnyuk the move to a single assembly noted here is for multi-user only right? We haven’t agreed that we don’t need a single user assembly yet.

skabashnyuk:
@bmicklea it's topic TBD. I'll create an issue for that.

riuvshin (approved):

sainirijul:
Hi,
I am trying to set up a workspace supporting multiple concurrent users. For this, i am using Multiple user assembly of Eclipse che and followed this thread.

On running the below command in windows powershell:
docker run -it -e CHE_MULTIUSER=true -e CHE_HOST= -e CHE_KEYCLOAK_AUTH_SERVER_URL=http:///auth -v /var/run/docker.sock:/var/run/docker.sock -v /c/tmp:/data eclipse/che start

I am getting the below message:
WARN: Bound 'eclipse/che' to 'eclipse/che:5.22.1'
/scripts/base/startup_02_pre_docker.sh: line 14: /data/cli.log: Resource busy
INFO: (che cli): 5.22.1 - using docker 17.12.0-ce / docker4windows
WARN: Newer version '6.0.0-M4' available
INFO: (che start): Booted and reachable
INFO: (che start): Ver: 5.22.1
INFO: (che start): Use: http://localhost:8080
INFO: (che start): API: http://localhost:8080/swagger

Now can you please guide me further steps and other steps if i am missing to set up the workspace which multiple users can access concurrently from different machines ? I could not find simple straight forward documents/references for this. Your help would be highly appreciated.

Looking forward to your response.

PR #6448, opened 2017-09-26
Avoid NPE by checking if map are null and not only empty in OpenShift connector

benoitf:
### What does this PR do?
Avoid NPE by checking if map are null and not only empty in OpenShift connector

https://github.com/redhat-developer/rh-che/issues/338

#### Release Notes
N/A

#### Docs PR
N/A


benoitf:
ci-build

l0rd:
@benoitf that's perfect for labels thanks! I think we have the same problem (NPE) when the image for the runtime stack doesn't have any `EXPOSE <port>`. Do you mind check it and fix that in this PR?

benoitf:
@l0rd I would prefer to do it in another PR as this PR targets a specific issue.

l0rd:
@benoitf ok

l0rd (approved):

PR #6450, opened 2017-09-26
Convert model into dto before returning in service

tolusha:
Convert model into dto before returning in service


sleshchenko (approved):
I'd fix commit message. Other is GOOD!

skabashnyuk (approved):

Issue #5930, opened 2017-08-08
 Run QA in multiuser branch and figure out what should be done to adapt QA to multiuser branch


skabashnyuk:
Run QA in multi-user branch and figure out what should be done to adapt QA to multiuser branch

dmytro-ndp:
[Test report](https://github.com/codenvy/qa/files/1251104/report1503586820.zip):
```
[TEST] =========== RUN OPTIONS ===========================
[TEST] Mode                : grid
[TEST] Tests inclusion     : stable and unstable tests
[TEST] Rerun failing tests : false
[TEST] ===================================================
[TEST] Product Protocol : http
[TEST] Product Host     : 172.17.0.1
[TEST] Tests            : -DrunSuite=/tmp/CheSuite.xml
[TEST] Threads          : 4
[TEST] Web browser      : GOOGLE_CHROME
[TEST] Web driver ver   : 2.31
[TEST] Web driver port  : 4444
[TEST] Additional opts  : -Dgrid.mode=true  
[TEST] ===================================================

Failed tests: 
  AddSshKeyForGitHubTest.generateKey:72 » Timeout Expected condition failed: wai...
  KeyBindingsTest.dialogAboutKeyBindingTest:114 expected [true] but found [false]
  CreateAndDeleteProjectsTest.createAndDeleteProjectTest:98 » Timeout Expected c...
  CheckLaravelStackTest.checkLaravelStack:73 » Timeout Expected condition failed...
  ChangeVariableWithEvaluatingTest.changeVariableTest:140 » Timeout Expected con...
  DebugExternalClassTest.shouldDebugJreClass:139 » Timeout Expected condition fa...
  InnerClassAndLambdaDebuggingTest.shouldDebugInnerClass:148 » Timeout Expected ...
  InnerClassAndLambdaDebuggingTest.shouldDebugMethodLocalInnerClass:136 » Timeout
  InnerClassAndLambdaDebuggingTest.shouldDebugStaticInnerClass:160 » Timeout Exp...
  MultimoduleProjectDebuggingTest.shouldDebugStaticMethod:182 » Timeout Expected...
  MultimoduleProjectDebuggingTest.shouldGoIntoConstructor:130 » Timeout Expected...
  PhpProjectDebuggingTest.shouldDebugCliPhpScriptFromFirstLine:139 » Timeout Exp...
[...]

PR #6456, opened 2017-09-27
Adapts selenium tests to che multiuser changes

akorneta:
### What does this PR do?
Adapts selenium tests to che multiuser changes.
Adds keycloak auth client to allow get and refresh access tokens for test users.
Adds tests startup option:
`--multiuser`
Note that without this parameter, selenium tests will proceed the same flow as on single user che.

Adds configuration properties:
Defines default test user for che single / multiuser selenium tests
`che.test_user.email`
`che.test_user.password`

Defines admin test user for che multiuser selenium tests
`che.admin_user.email`
`che.admin_user.password`

https://github.com/eclipse/che/issues/5930


tolusha:
Is ready to review?

akorneta:
@tolusha yes, in general, things related to che multiuser almost done except the additional user creation.
Another part that related to launch che single user will be done in this pr asap.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/requestfactory/TestAdminHttpJsonRequestFactory.java
```diff
@@ -10,21 +10,32 @@
...
+  private final String adminEmail;
+  private final String adminPassword;
+  private final TestAuthServiceClient authServiceClient;
 
   @Inject
-  public TestAdminHttpJsonRequestFactory(Provider<AdminTestUser> adminProvider) {
-    this.adminProvider = adminProvider;
+  public TestAdminHttpJsonRequestFactory(
+      TestAuthServiceClient authServiceClient,
+      @Named("che.admin_user.email") String adminEmail,
```
dmytro-ndp: It's Che-specific settings which should be re-written in tests for Codenvy onprem/saas. 
I propose to create **TestAdminEmailProvider** and **TestAdminPasswordProvider** interfaces and next implementations for them:
- **CheTestAdminEmailProvider** and **CheTestAdminPasswordProvider** implementations in selenium tests for Che (where _che.admin_user.email_ and _che.admin_user.password_ config properties are using);
- **CodenvyTestAdminEmailProvider** and **CodenvyTestAdminPasswordProvider** implementations for Codenvy (where _codenvy.admin.name_, _codenvy.admin.initial_password_, and _codenvy.admin.password_ config properties are used similarly to the class https://github.com/codenvy/codenvy/blob/master/selenium/codenvy-selenium-test/src/main/java/com/codenvy/selenium/core/user/OnpremAdminTestUser.java)

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/requestfactory/TestAdminHttpJsonRequestFactory.java
```diff
@@ -10,21 +10,32 @@
...
+    this.adminPassword = adminPassword;
+    this.authServiceClient = authServiceClient;
   }
 
   protected String getAuthToken() {
-    return adminProvider.get().getAuthToken();
+    try {
+      return authServiceClient.login(adminEmail, adminPassword);
+    } catch (Exception ex) {
+      throw new RuntimeException("Failed to get access token for admin user");
```
dmytro-ndp: Let's log admin's email here as well so as it could be different.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/requestfactory/TestDefaultUserHttpJsonRequestFactory.java
```diff
@@ -10,22 +10,37 @@
...
   }
 
+  @Override
   protected String getAuthToken() {
-    return testUserProvider.get().getAuthToken();
+    try {
+      return authServiceClient.login(testUser.getEmail(), testUser.getPassword());
+    } catch (Exception ex) {
+      throw new RuntimeException(
+          format("Failed to access get token for user '%s'", testUser.getName()));
```
dmytro-ndp: It's better to log user's email for the full information so as user's name is usual just the part of email before "@"

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/requestfactory/TestDefaultUserHttpJsonRequestFactory.java
```diff
@@ -10,22 +10,37 @@
...
+      TestAuthServiceClient authServiceClient, TestUser testUser) {
+    this.testUser = testUser;
+    this.authServiceClient = authServiceClient;
   }
 
+  @Override
   protected String getAuthToken() {
-    return testUserProvider.get().getAuthToken();
+    try {
+      return authServiceClient.login(testUser.getEmail(), testUser.getPassword());
```
dmytro-ndp: we can use **testUser.getAuthToken()** here, can't we?

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/requestfactory/TestDefaultUserHttpJsonRequestFactory.java
```diff
@@ -10,22 +10,37 @@
...
+import org.eclipse.che.selenium.core.client.TestAuthServiceClient;
+import org.eclipse.che.selenium.core.user.TestUser;
 
-/** @author Dmytro Nochevnov */
+/**
+ * @author Dmytro Nochevnov
+ * @author Anton Korneta
+ */
 @Singleton
 public class TestDefaultUserHttpJsonRequestFactory extends TestHttpJsonRequestFactory {
```
dmytro-ndp: This class looks like redundant after default test user is replaced by regular test user, doesn't it?

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/requestfactory/TestUserHttpJsonRequestFactory.java
```diff
@@ -10,22 +10,33 @@
...
+      TestAuthServiceClient authServiceClient, @Assisted TestUser testUser) {
+    this.authServiceClient = authServiceClient;
+    this.testUser = testUser;
   }
 
+  @Override
   protected String getAuthToken() {
-    return this.authToken;
+    try {
+      return authServiceClient.login(testUser.getEmail(), testUser.getPassword());
```
dmytro-ndp: we can use **testUser.getAuthToken()** here, can't we?

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/requestfactory/TestUserHttpJsonRequestFactory.java
```diff
@@ -10,22 +10,33 @@
...
+    } catch (Exception ex) {
+      throw new RuntimeException(
+          format("Failed to access get token for user '%s'", testUser.getName()));
```
dmytro-ndp: It's better to log user's email for the full information so as user's name is usual just the part of email before "@"

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/requestfactory/TestUserHttpJsonRequestFactory.java
akorneta: Well, we need to guarantee that returned access token is valid and not expired so it's better to use authServiceClient.login call to get/refresh/login user's token

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/requestfactory/TestDefaultUserHttpJsonRequestFactory.java
akorneta: yes you are right

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/user/TestUserFactory.java
```diff
@@ -17,7 +17,13 @@
  * @author Dmytro Nochevnov
  */
 public interface TestUserFactory {
-  TestUser create(@Assisted("email") String email);
 
-  TestUser create(@Assisted("email") String email, @Assisted("password") String password);
+  /** Creates new test user with generated e-mail and password */
+  TestUserImpl create();
```
dmytro-ndp: What is the propose of this method? 
It looks as redundant so as simple `@Inject TestUser test user;` could be used instead of testUserFactory.create()

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/user/TestUserImpl.java
```diff
@@ -45,17 +44,32 @@
   private final TestWorkspaceServiceClient workspaceServiceClient;
 
   /** To instantiate user with generated email and password. */
-  @Inject
+  @AssistedInject
```
dmytro-ndp: Did you test it? 
It will lead to an error of injection when we try to create random user by the command 
`@Inject TestUser testUser;`.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/user/TestUserFactory.java
akorneta: This method needs in case when in tests there need to be some user with generated email and password. So for codenvy I planed to bind the `TestUser` to implementation wich call this method.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/requestfactory/TestAdminHttpJsonRequestFactory.java
akorneta: added

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/core/TestUserServiceClientProvider.java
```diff
@@ -0,0 +1,37 @@
...
+  public TestUserServiceClientProvider(
+      @Named("che.multiuser") boolean isMultiUser,
+      TestUserServiceClientImpl cheTestUserServiceClient) {
+    this.isMultiUser = isMultiUser;
+    this.cheTestUserServiceClient = cheTestUserServiceClient;
+  }
+
+  @Override
+  public TestUserServiceClient get() {
+    return isMultiUser ? cheTestUserServiceClient : cheTestUserServiceClient;
```
dmytro-ndp: Looks like a useless command. IMHO it's better to explicitly define proper user service clients in **CheSeleniumSingleUserModule** and **CheSeleniumMultiUserModule**.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/client/TestUserServiceClient.java
```diff
@@ -10,65 +10,69 @@
...
-    this.apiEndpoint = apiEndpointProvider.get().toString();
-    this.requestFactory = requestFactory;
-  }
+/**
+ * @author Mihail Kuznyetsov
+ * @author Anton Korneta
+ */
+public interface TestUserServiceClient {
 
-  public User getByEmail(String email) throws Exception {
```
dmytro-ndp, line 39: We need this method for https://github.com/codenvy/codenvy/blob/master/selenium/codenvy-selenium-test/src/test/java/com/codenvy/selenium/factory/AuthenticateAndAcceptFactoryThroughGitHubOAuthTest.java#L89-L89

akorneta: I'll adapt this code in codenvy PR

selenium/che-selenium-test/src/test/resources/conf/selenium.properties
```diff
@@ -24,3 +24,21 @@ github.auxiliary.password=
...
+# Define configuration for che default test user.
+# Note that for single user assembly the configuration should be:
+# email: che@eclipse.org
+# password: secret
+# For che multiuser assembly, user must be already created on identity provider service
+che.test_user.email=
+che.test_user.password=
+
+# Defines additional test user
+che.aditional_test_user.email=
```
dmytro-ndp: I can't find the usage of these properties. Can you, please, write an example of usage?

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/user/TestUserImpl.java
akorneta: yes for usecase described [here](https://github.com/eclipse/che/pull/6456#discussion_r141295205) it works pretty fine

selenium/che-selenium-test/src/test/resources/conf/selenium.properties
akorneta: I'm working on it and will notify when it will be done

selenium/che-selenium-test/src/test/resources/suites/CheSuite.xml
```diff
@@ -886,18 +886,6 @@
                 </methods>
             </class>
             <class name="org.eclipse.che.selenium.refactor.types.TestFailTest"/>
-            <class name="org.eclipse.che.selenium.subversion.CommitTest"/>
```
dmytro-ndp: It have not been finally decided to remove support of SVN plugin yet https://github.com/eclipse/che/issues/5484#issuecomment-328075154

selenium/che-selenium-test/src/test/resources/suites/PassingSuite.xml
```diff
@@ -0,0 +1,775 @@
+<!--
```
dmytro-ndp: What is this suite for? We have already had CheSuite.xml, and it is using by default.

dmytro-ndp (changes requested):
IMHO we need to ensure that these changes don't break execution of selenium tests for Single User Che, Codenvy onprem and Codenvy SaaS.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/user/TestUserFactory.java
dmytro-ndp: I see. But we don't need assisted injection, that is TestUserFactory, for creation an instance from constructor without **@Assisted** parameters at all. 
@Inject annotation is able to provide required instance of TestUser.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/requestfactory/CheAdminHttpJsonRequestFactory.java
```diff
@@ -10,34 +10,18 @@
...
 import javax.inject.Inject;
 import org.eclipse.che.selenium.core.client.TestAuthServiceClient;
 
 /** @author Dmytro Nochevnov */
-public class TestAdminHttpJsonRequestFactory extends TestHttpJsonRequestFactory {
-
-  private final String adminEmail;
-  private final String adminPassword;
-  private final TestAuthServiceClient authServiceClient;
+public class CheAdminHttpJsonRequestFactory extends TestUserHttpJsonRequestFactory {
```
dmytro-ndp: Usually we add **Test** word into the name of selenium test framework classes to avoid confusion with Che product codebase.

selenium/che-selenium-test/src/test/resources/conf/selenium.properties
akorneta: @dmytro-ndp  I deleted these properties because for the first step of integration selenium tests with  che multiuser it's overhead.

selenium/che-selenium-test/src/test/resources/suites/PassingSuite.xml
akorneta: removed

akorneta:
Fixed single user che test launching, for codenvy and saas assemblies there will be separate prs. Removed additional che user properties.
@tolusha @dmytro-ndp please take a look.

selenium/che-selenium-test/selenium-tests.sh
```diff
@@ -14,7 +14,24 @@ export CALLER=$(basename $0)
...
+       cheMultiuserValue=true
+       unset params[$index]                     # remove "--multiuser" from parameters
+       break
+   fi
+   let "index+=1"
+done
+
+export CHE_MULTIUSER=$cheMultiuserValue
+
+set -- "${params[@]}"
```
tolusha, line 31: Make it simpler. Just export what you need.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/core/factory/TestFactory.java
```diff
@@ -26,28 +25,34 @@
...
   public TestFactory(
       String factoryUrl,
-      DefaultTestUser owner,
+      TestUser owner,
       FactoryDto factoryDto,
       TestDashboardUrlProvider dashboardUrl,
       TestFactoryServiceClient factoryServiceClient,
-      TestWorkspaceServiceClient workspaceServiceClient) {
+      TestWorkspaceServiceClient workspaceServiceClient,
+      LoginPage loginPage) {
```
tolusha, line 39: You moved TestFactory into the test module. Unfortunately we can't mix page objects and selenium core part. 
Describe usecase.

selenium/che-selenium-test/src/test/resources/suites/CheSuite.xml
tolusha: @akorneta Pls, restore removed tests.

selenium/che-selenium-test/src/test/resources/conf/selenium.properties
```diff
@@ -24,3 +24,17 @@ github.auxiliary.password=
 # Google credentials
 google.user=
 google.password=
+
+# Define configuration for che default test user.
+# Note that for single user assembly the configuration should be:
```
tolusha: It is n't clear if user has to be predefined in single user assembly. Could you clarify pls

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/requestfactory/TestUserHttpJsonRequestFactory.java
dmytro-ndp: But **testUser.getAuthToken()** can refresh user's token every time we call it, can't it?

selenium/che-selenium-test/src/test/resources/conf/selenium.properties
dmytro-ndp: I see. We can add it in future on demand.

selenium/che-selenium-test/selenium-tests.sh
dmytro-ndp: @tolusha: we should remove **--multiuser** parameter from list of parameters because of **webdriver.sh** doesn't recognize it and throws an error.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/user/TestUserFactory.java
akorneta: removed method + changed annotation

selenium/che-selenium-test/src/test/resources/conf/selenium.properties
akorneta: added default values, so for single user che it's not needed to provide default user data

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/requestfactory/TestUserHttpJsonRequestFactory.java
```diff
@@ -10,22 +10,35 @@
...
+    this.authServiceClient = authServiceClient;
+    this.email = email;
+    this.password = password;
   }
 
+  @Override
   protected String getAuthToken() {
-    return this.authToken;
+    try {
+      return authServiceClient.login(email, password);
```
tolusha, line 39: Can we login once and then return authentication token?

selenium/che-selenium-test/selenium-tests.sh
tolusha: That's what I told. In case of multi use assembly:
```
export CHE_MULTIUSER=true
```

And then check environment variable in the code.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/core/CheSeleniumMultiUserModule.java
```diff
@@ -0,0 +1,50 @@
...
+  protected void configure() {
+    bind(TestAuthServiceClient.class).to(KeycloakTestAuthServiceClient.class);
+
+    bind(TestUser.class).to(CheDefaultTestUser.class);
+    bind(AdminTestUser.class).to(CheAdminTestUser.class);
+
+    bind(TestUserServiceClient.class).to(CheTestUserServiceClient.class);
+    bind(HttpJsonRequestFactory.class).to(TestUserHttpJsonRequestFactory.class);
+    bind(TestUserHttpJsonRequestFactory.class).to(TestCheDefaultUserHttpJsonRequestFactory.class);
+    bind(TestMachineServiceClient.class).to(CheTestMachineServiceClient.class);
```
tolusha: There is mess with class names.
`CheTest`, `Test`, `TestChe`

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/core/CheSeleniumSuiteModule.java
```diff
@@ -78,44 +66,33 @@ public void configure() {
...
     install(new FactoryModuleBuilder().build(TestWorkspaceServiceClientFactory.class));
+    install(new FactoryModuleBuilder().build(TestUserFactory.class));
 
-    install(
-        new FactoryModuleBuilder()
-            .implement(TestUser.class, TestUserImpl.class)
-            .build(TestUserFactory.class));
-
-    bind(AdminTestUser.class).to(CheAdminTestUser.class);
+    if (parseBoolean(System.getenv(CHE_MULTIUSER_VARIABLE))) {
```
tolusha, line 99: I can see the same bindings in both CheSeleniumMultiUserModule and CheSeleniumSingleUserModule.
Let's keep them in CheSeleniumSuiteModule

selenium/che-selenium-test/selenium-tests.sh
dmytro-ndp: We should remove _--multiuser_ parameter from parameters because otherwise it will be passed to **webdriver.sh** as it and will cause an error at start of execution of selenium tests:
> (target/bin/webdriver.sh "$TESTS_SCOPE" **$@**)

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/requestfactory/TestUserHttpJsonRequestFactory.java
dmytro-ndp: We need to update token regularly because of little default token timeout around 5 minutes.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/core/CheSeleniumMultiUserModule.java
dmytro-ndp: Let me explain to Anton the naming scheme:
1. Usually we add **Test** prefix to the name of classes which are specific to selenium tests like the following:
> AuthServiceClient -> **Test**AuthServiceClient
> HttpJsonRequestFactory -> **Test**UserHttpJsonRequestFactory
2. In case of different implementation for Che, Codenvy Onprem, Keycloak, Codenvy SaaS, we add name of product at start of class name:
> TestAuthServiceClient -> **Che**TestAuthServiceClient
or -> **Onpem**TestAuthServiceClient
or -> **Keycloak**TestAuthServiceClient

According to this scheme, correct name of class 
> Test**Che**DefaultUserHttpJsonRequestFactory > **Che**TestDefaultUserHttpJsonRequestFactory.

akorneta: ok I'll fix that

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/core/factory/TestFactoryInitializer.java
```diff
@@ -80,7 +80,7 @@ public TestFactory fromUrl(String url) throws Exception {
     HttpJsonRequest httpJsonRequest =
         requestFactory.fromUrl(apiEndpointProvider.get() + "factory/resolver");
     httpJsonRequest.setBody(singletonMap("url", url));
-    HttpJsonResponse response = httpJsonRequest.request();
+    HttpJsonResponse respoTestCheDefaultUserHttpJsonRequestFactorynse = httpJsonRequest.request();
```
dmytro-ndp: typo

dmytro-ndp:
@akorneta: thank you for quick fixup.
I would appropriate it if commit messages will reflect content of changes.

tolusha:
There are 2 requirements to merge into master branch:
* don't mix page objects with core part of selenium tests
* ensure running tests for codenvy/saas projects

tolusha:
It is ok to be merged into multiuser branch and unblock platform team workking

tolusha (approved):

skabashnyuk:
@tolusha  we do our best to met your requirements before merge che-multiuser branch in master

skabashnyuk (approved):

Issue #6461, opened 2017-09-27
Override application state when workspace has beed stopped


vzhukovs:
Problem represents when workspace is stopped but IDE looses focus and following code snipped trigger save state:

`AppContextImpl.java:`
```java
Elements.getWindow()
        .addEventListener(
            Event.BLUR, evt -> appStateManager.get().persistWorkspaceState(getWorkspaceId()));

   ...

 @Override
  public String getWorkspaceId() {
    if (userWorkspace == null) {
      throw new IllegalArgumentException(getClass() + " Workspace can not be null.");
    }

    return userWorkspace.getId();
  }
```

**Reproduction Steps:**
* Stop workspace
* Blur the browser window

Discovered during check #5067

PR #6464, opened 2017-09-27
Do not store preferences when workspace has already stopped

vzhukovs:
### What does this PR do?
This changes proposal fix the problem when IDE could rewrite IDE state when workspace has already stopped. It means that when workspace stops, IDE stores opened files to the preferences and when it looses the focus - state could be rewritten with empty values.

#6461 

#### Changelog
Do not store preferences when workspace has already stopped

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/context/AppContextImpl.java
+++ b/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/context/AppContextImpl.java
@@ -26,6 +26,7 @@ import com.google.inject.Provider;
 import com.google.inject.Singleton;
 import com.google.web.bindery.event.shared.EventBus;
 import elemental.events.Event;
+import elemental.events.EventRemover;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/context/AppContextImpl.java
```diff
@@ -481,7 +490,10 @@ public ActiveRuntime getActiveRuntime() {
 
     @Override
     public void onWindowClosing(WindowActionEvent event) {
-      appStateManager.get().persistWorkspaceState(getWorkspaceId());
+      Workspace workspace = getWorkspace();
```
vparfonov, line 493: What reason call getWorkspace() here?

vzhukovs: Because, `onWindowClosing` may be called when workspace has stopped. And calling `getWorkspaceId` will throw an exception, because current workspace if not provided. So we'll be on the safe side, we'll check if we have a workspace as initialized object.

vparfonov (approved):

PR #6505, opened 2017-09-29
Disable scheduling of methods annotated with @ScheduleRate/@ScheduleDelay if period/delay is negative

### LINKED ISSUE #6383: Starting a second container of eclipse/che will kill all the workspaces container of the first one
I'm trying to run 2 che-server in the same docker engine using the CLI. If I start a second che-server, the che launcher is killing all the workspaces started by the first one.
This issue is very annoying when trying to launch another che instance from a che workspace (che in che)

**Reproduction Steps:**
1. Start a first che-server

        docker run -t --rm  -v /var/run/docker.sock:/var/run/docker.sock -v /home/sutan/che-data:/data eclipse/che:nightly

2. Create and start a workspace from this new che-server
3. Start another che-server with different name and port

        docker run --rm -t -v /var/run/docker.sock:/var/run/docker.sock -v /home/sutan/data-dev:/data -e CHE_CONTAINER=che-dev -e CHE_PORT=8081 eclipse/che:nightly start

4. Then, the workspace created by the first che-server is killed

**OS and version:**    
RHEL 7

sunix:
### What does this PR do?
Disable scheduling of methods annotated with @ScheduleRate/@ScheduleDelay if period/delay is negative

Thanks to this PR, we can fix #6383 by setting  `che.docker.cleanup_period_min=-1`

### DIFF
--- a/core/commons/che-core-commons-schedule/src/main/java/org/eclipse/che/commons/schedule/ScheduleDelay.java
+++ b/core/commons/che-core-commons-schedule/src/main/java/org/eclipse/che/commons/schedule/ScheduleDelay.java
@@ -42,6 +42,9 @@ public @interface ScheduleDelay {
   /** @return - name of configuration parameter for initialDelay */
   String initialDelayParameterName() default "";
 
-  /** @return - name of configuration parameter for delay */
+  /**
+   * @return - name of configuration parameter for delay. A non positive delay value will disable
+   *     the scheduling of the method.
+   */
   String delayParameterName() default "";
 }
--- a/core/commons/che-core-commons-schedule/README.md
+++ b/core/commons/che-core-commons-schedule/README.md
@@ -133,6 +133,30 @@ public class RouterRulesRegistry {
        when statically configured value. Same for delay and delayParameterName.</p>
 </blockquote>
 
+### Disabling scheduling by rate or period
+If a scheduled method has been configured through `delayParameterName` or `periodParameterName`, it is possible to disable
+the scheduling by setting a non positive value to the parameter.
+
+Example: Disabling scheduling of this `registerRoutingRules` method:
+
+```java
+@Singleton // should be eager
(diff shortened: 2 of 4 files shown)


garagatyi:
+1

benoitf (approved):

core/commons/che-core-commons-schedule/src/main/java/org/eclipse/che/commons/schedule/ScheduleDelay.java
```diff
@@ -42,6 +42,9 @@
   /** @return - name of configuration parameter for initialDelay */
   String initialDelayParameterName() default "";
 
-  /** @return - name of configuration parameter for delay */
+  /**
+   * @return - name of configuration parameter for delay. A negative delay value will disable the
```
garagatyi: it is rather non-positive than negative

sunix: right :+1:

skabashnyuk (approved):
Please sync with @slemeur to check if we need some docs or release notes changes.

skabashnyuk:
Couple of words here https://github.com/eclipse/che/blob/master/core/commons/che-core-commons-schedule/README.md would be nice too.

sunix:
@skabashnyuk @garagatyi I've updated the PR with your feedback

garagatyi:
Don't want to be PITA by saying it so late but have you considered using just zero instead of non-positive value? A single value for disabling feature is easier for understanding than a range.

sunix:
@garagatyi I think a single value would confuse people.

garagatyi (approved):

garagatyi:
OK

skabashnyuk (approved):

slemeur (approved):
ok for me

PR #6508, opened 2017-09-29
CHE-5893: Adapt CLI to use keycloak openid authentication

### LINKED ISSUE #5893: Adapt CLI to use keycloak openid authentication
We need to change this code https://github.com/eclipse/che/blob/master/dockerfiles/lib/src/api/wsmaster/auth/auth-data.ts#L129 to use keycloak openid authentication.

1. I think we need a new client in che realm. Let's say 'curl' with "Bearer only" authentification. 
2. Then, change the way how cli obtain token to something like this. ```curl --data "grant_type=password&client_id=che-public&username=admin&password=123456" http://192.168.65.2:5050/auth/realms/che/protocol/openid-connect/token```

See also http://blog.keycloak.org/2015/10/getting-started-with-keycloak-securing.html

garagatyi:
### What does this PR do?
Adapt CLI to use keycloak openid authentication.
I'm still checking how it works, but want reviewers to take a look.

Fixes #5893

### DIFF
--- a/dockerfiles/lib/src/api/wsmaster/auth/auth-data.ts
+++ b/dockerfiles/lib/src/api/wsmaster/auth/auth-data.ts
@@ -36,7 +36,7 @@ export class AuthData {
     private password: string;
     private authType: AUTH_TYPE;
 
-    constructor(cheMasterLocation?: ServerLocation, username?: string, password?: string) {
+    constructor(cheMasterLocation?: string, username?: string, password?: string) {
 
         this.username = username;
         this.password = password;
@@ -55,11 +55,7 @@ export class AuthData {
             }
         } else {
             this.authType = AUTH_TYPE.CODENVY_SSO;
-            if (cheMasterLocation) {
-                this.authServerLocation = cheMasterLocation;
-            } else {
-                this.authServerLocation = new WsMasterLocation();
-            }
+            this.authServerLocation = new WsMasterLocation(cheMasterLocation);
         }
     }
 
@@ -112,15 +108,6 @@ export class AuthData {
             default:
                 error('Che authentication type ' + this.authType + ' is illegal');
         }
-    }
-
-    static parse(remoteUrl: string, username?: string, password?: string) : AuthData {
-        let serverApi: ServerLocation;
-        if (remoteUrl) {
-            serverApi = ServerLocation.parse(remoteUrl);
-        }
-
-        return new AuthData(serverApi, username, password);
     }
 
     private ssoLogin(http) : Promise<boolean> {
--- a/dockerfiles/lib/src/api/wsmaster/auth/auth-data.ts
+++ b/dockerfiles/lib/src/api/wsmaster/auth/auth-data.ts
@@ -12,9 +12,8 @@
 import {RemoteIp} from "../../../spi/docker/remoteip";
 import {Log} from "../../../spi/log/log";
 import {ServerLocation} from "../../../utils/server-location";
-import {DefaultHttpJsonRequest, HttpJsonRequest, HttpJsonResponse} from "../../../spi/http/default-http-json-request";
-import {PermissionDto} from "../permissions/dto/permissiondto";
 import {error} from "util";
+import {WsMasterLocation} from "../wsmaster-location";
 
 /**
  * Defines a way to store the auth data in order to deal with remote REST or Websocket API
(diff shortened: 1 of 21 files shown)


dockerfiles/lib/src/api/wsmaster/auth/auth-data.ts
```diff
@@ -164,38 +165,58 @@ export class AuthData {
...
+            resolve(true);
+        });
+    }
 }
+
+enum AUTH_TYPE {
+    KEYCLOAK,
+    CODENVY_SSO,
+    NO
+}
```
riuvshin: new line

dockerfiles/lib/src/utils/server-location.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    getPort(): number {
+        return this.port;
+    }
+
+    setPort(port : number) {
+        this.port = port;
+    }
+}
```
riuvshin: new line

riuvshin (approved):

benoitf:
I'm not sure to follow the use of ServerLocation and AuthData with some methods with static access and some methods without the static access
It's like if there are duplicates between AuthData and ServerLocation

dockerfiles/lib/src/api/wsmaster/permissions/permissions.ts
```diff
@@ -36,9 +42,9 @@ export class Permissions {
      */
     listPermissions():Promise<Array<DomainDto>> {
 
-        var jsonRequest:HttpJsonRequest = new DefaultHttpJsonRequest(this.authData, '/api/permissions', 200);
+        let jsonRequest: HttpJsonRequest = new DefaultHttpJsonRequest(this.authData, this.apiLocation, '/api/permissions', 200);
```
benoitf: why do we need to provide apiLocation if authData contains the server location ?

garagatyi: Because keycloak is deployed outside of Che API

benoitf: ok I think the main issue comes that global idea was to only provide one object to the instances including all the configuration. So maybe we could have merged one object into another one (or a third one)

so, we could have all auth the authentication data stored in a single object along with the api location

dockerfiles/lib/src/utils/server-location.ts
```diff
@@ -0,0 +1,94 @@
...
+            } else if ('https:' === url.protocol || 'wss:' === url.protocol) {
+                port = 443;
+            }
+        } else {
+            port = +url.port;
+        }
+        return new ServerLocation(url.hostname, port, isSecured);
+    }
+
+    static detectCheMasterLocation(): ServerLocation {
```
benoitf: what if this becomes the empty constructor ?

benoitf: (or a class CheMasterLocation extending ServerLocation)

garagatyi: makes sense

dockerfiles/lib/src/api/wsmaster/permissions/permissions.ts
garagatyi: In case of exec agent of the wsagent location of the API that is being requested has no connection with authentication at all. Do you think we should keep this in the auth-data?
If so then it doesn't look like auth data only and this behavior is not described in the docs of this class.
I like simplicity too, but in this case, I don't see how server of wsagent fits in auth-data class.
If I got something wrong, please, elaborate.

garagatyi: What if we do next:
1. Allow getting API location from AuthData
2. Make HttpJsonRequest constructor accept AuthData and ServerLocation
3. In HttpJsonRequest if provided ServerLocation is null then use server location from AuthData

Or we can alternate 2 and 3 with the addition of a method to HttpJsonRequest which sets ServerLocation for the request. If method is not called ServerLocation from AuthData is used.

dockerfiles/lib/src/utils/server-location.ts
garagatyi: Moved to separate class

garagatyi:
@benoitf please, take a look at [this diff](https://github.com/eclipse/che/compare/che-multiuser...garagatyi:mu_cli_pr_2?expand=1) - it is an example of keeping server location value inside of AuthData. WDYT?

benoitf:
@garagatyi I think it's easier in the new diff

dockerfiles/lib/src/api/wsmaster/permissions/permissions.ts
benoitf: I like this proposal

dockerfiles/lib/src/api/wsmaster/wsmaster-location.ts
```diff
@@ -0,0 +1,51 @@
...
+            port = process.env.CHE_PORT;
+        } else if (secured) {
+            port = 443;
+        } else {
+            port = WsMasterLocation.DEFAULT_PORT;
+        }
+
+        super(hostname, port, secured);
+    }
+}
```
skabashnyuk, line 52: new line?

skabashnyuk (approved):

benoitf (approved):

Issue #6515, opened 2017-10-02
Che multiuser: Move and adapt html templates from Codenvy project


sleshchenko:
Move and adapt HTML templates from `Codenvy` project.
It is required to move HTML templates from `Codenvy` project to send users email notifications about some changes like organization renaming/removing. Also, these changes are required for moving invite API from `Codenvy`.
And just moving won't work because there is `thymeleaf` implementation that should not be used for Che. Instead of `thymeleaf` there should be StringTemplate(ST) simple implementation. So it is needed to do required changes to implement it in this way.

PR #6520, opened 2017-10-02
CHE-6515 Added email notifications

sleshchenko:
### What does this PR do?
Adds template processor.
Adds warning on tomcat start when mail sender is not configured.
Adds sending of email notification when changes occur in an organization. Note that notifications are not styled. Styling will be made during this [issue](https://github.com/eclipse/che/issues/6546).

##### Member added to an Organization
added_to_org

##### Member Removed from an Organization
removed_from_org

##### Organization Renamed
org_remaned

##### Organization Removed
org_deleted

https://github.com/eclipse/che/issues/6515


skabashnyuk (approved):

garagatyi (approved):
Looks good! Would be awesome to have some tests also.

akorneta (approved):

PR #6523, opened 2017-10-02
Map system state API to login filter

### LINKED ISSUE #5893: Adapt CLI to use keycloak openid authentication

garagatyi:
### What does this PR do?
It is needed since it is secured by permissions system which requires
an authenticated user.
Also, use base API endpoint for checking Che in OpenShift deploy script since system state API which was used before is secured now.

Fixes #5893

### DIFF
--- a/dockerfiles/init/modules/openshift/files/scripts/wait_until_che_is_available.sh
+++ b/dockerfiles/init/modules/openshift/files/scripts/wait_until_che_is_available.sh
@@ -44,7 +44,7 @@ elif [ ${SECONDS} -lt ${end} ]; then
   exit 1
 fi
 
-che_http_status=$(curl -s -o /dev/null -I -w "%{http_code}" "${CHE_API_ENDPOINT}/system/state")
+che_http_status=$(curl -s -o /dev/null -I -w "%{http_code}" "${CHE_API_ENDPOINT}/")
 if [ "${che_http_status}" == "200" ]; then  
   echo "[CHE] Che is up and running"
 else
--- a/multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/deploy/KeycloakServletModule.java
+++ b/multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/deploy/KeycloakServletModule.java
@@ -20,12 +20,12 @@ public class KeycloakServletModule extends ServletModule {
   protected void configureServlets() {
     bind(KeycloakAuthenticationFilter.class).in(Singleton.class);
 
-    // Not contains '/websocket', /docs/ (for swagger) and not ends with '/ws' or '/eventbus' or '/settings/' or '/api/system/state'
+    // Not contains '/websocket', /docs/ (for swagger) and not ends with '/ws' or '/eventbus' or '/settings/'
     filterRegex(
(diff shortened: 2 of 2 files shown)


skabashnyuk (approved):

dockerfiles/init/modules/openshift/files/scripts/wait_until_che_is_available.sh
```diff
@@ -44,7 +44,7 @@ elif [ ${SECONDS} -lt ${end} ]; then
   exit 1
 fi
 
-che_http_status=$(curl -s -o /dev/null -I -w "%{http_code}" "${CHE_API_ENDPOINT}/system/state")
+che_http_status=$(curl -s -o /dev/null -I -w "%{http_code}" "${CHE_API_ENDPOINT}/")
```
benoitf, line 47: @garagatyi it's strange. It means that you return `OK` before system is really available.

benoitf: Also OPENSHIFT_TOKEN is a token to access secure endpoint
https://github.com/garagatyi/che/blob/02252426f8e5ef5bb687ca9592573470e9ea866e/dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh#L182-L188

garagatyi: I think it is OpenShift token. Does it work in Che?

benoitf: AFAIK we're in the che for openshift script there

garagatyi: I think that checking /api is pretty the same check, so it's ok to secure this one

benoitf: it's not. If someone has stopped the system you would think system is ready while it is not.

garagatyi: CLI just checks whether the response is 200 or not. System state JSON data are not checked.  It is equal to the check that API responds.

benoitf: so you've found a bug in this script ;-)

garagatyi: Well as far as I understand system state API returns 403 in che-multiuser branch now. Permissions are checked but the user is anonymous, so it should not work in any case.

l0rd (approved):

garagatyi:
@benoitf Do you think something specific should be changed in this PR?

benoitf:
@garagatyi I think but l0rd is owner of the script so it was up to him

riuvshin (approved):

PR #6534, opened 2017-10-02
allow to override che default keycloak config

riuvshin:
### What does this PR do?
allow to override che default keycloak config

@skabashnyuk can you please add more explanations in multiuser.properties for these props


benoitf (approved):

benoitf:
FYI there is a warning on Eclipse Validation

skabashnyuk (approved):

PR #6541, opened 2017-10-03
change predefined host for start tests on the webdriver.sh to localhost

Ohrimenko1988:
### What does this PR do?
- change predefined host for starting selenium tests to "localhost"

### DIFF
--- a/selenium/che-selenium-core/bin/webdriver.sh
+++ b/selenium/che-selenium-core/bin/webdriver.sh
@@ -41,18 +41,7 @@ getRecommendedThreadCount() {
 }
 
 detectDockerInterfaceIp() {
-    local interfaces=("docker"
-                      "docker0")
-
-    for interface in ${interfaces[@]}; do
-        ifconfig ${interface} >/dev/null 2>&1
-        if [[ $? == 0 ]]; then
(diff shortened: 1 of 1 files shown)


garagatyi (changes requested):
I think we should reimplement function `detectDockerInterfaceIp` instead. And use approach used in Che CLI. 
[Here](https://github.com/garagatyi/che/blob/69d51483859d828aa285bcef324244ca236109ad/dockerfiles/cli/scripts/post_init.sh#L9-L9) is an example: 
`DOCKER_IP=$(docker run --rm --net host eclipse/che-ip:nightly)`

riuvshin:
+1 to what @garagatyi said. `hostname -I | awk '{ print $1 }'` will not work on mac for example

tolusha (approved):

akorneta (approved):

skabashnyuk:
```+++```

riuvshin (changes requested):
As @garagatyi said we have to **reimplement** `detectDockerInterfaceIp` function because in that state it does not make any sense

garagatyi (approved):
Looks good!

riuvshin (approved):

Issue #5606, opened 2017-07-11
[regression] Cannot move intelligent command from one goal to another goal


Ohrimenko1988:
**Reproduction Steps:**
- Create custom java stack
- Go to the stack settings and in "Commands" field create command
- Run this stack
- Go to the "CommandsExplorer" in the ide
- Try to move created command to another goal

**Expected behavior:**
Command will be moved

**Observed behavior:**
After choosing a new goal in the command editor, the "Save" button is still disabled, so command can`t be moved. 

**OS and version:**    
CHE 5.15.0 ; Chrome 59.0 ; Fedora 25

PR #6551, opened 2017-10-03
Set default value of the command's goal if it doesn't exist

svor:
### What does this PR do?
Propose to set default value of the command's goal if it doesn't exist. Fixes a problem with editing commands which was created from dashboard

https://github.com/eclipse/che/issues/5606

#### Release Notes

 N/A

#### Docs PR

 N/A


vparfonov (approved):

svor:
ci-build

PR #6561, opened 2017-10-03
Che-multiuser fix login

Ohrimenko1988:
### What does this PR do?
Rework login in the selenium tests and add ability to start tests in the codenvy/codenvy


garagatyi:
Please squash this PR before merge

selenium/che-selenium-test/src/test/resources/conf/selenium.properties
```diff
@@ -28,11 +28,11 @@ google.password=
 # Define configuration for che default test user.
 # For che single user assembly, the default values will be enough.
 # For che multiuser assembly, user must be already created on identity provider service.
-che.test_user.email=che@eclipse.org
-che.test_user.password=secret
+che.test_user.email=test@gmail.com
```
dmytro-ndp: From security reasons it's better not to place real credentials in the public config file.

selenium/che-selenium-test/src/test/resources/conf/selenium.properties
```diff
@@ -28,11 +28,11 @@ google.password=
...
+che.test_user.password=password
 
 # Define configuration for admin test user.
 # Note that admin user needs only for multiuser assembly and
 # must be already created on identity provider service
-che.admin_user.email=
-che.admin_user.password=
+che.admin_user.email=admin@admin.com
```
dmytro-ndp: From security reasons it's better not to place real credentials in the public config file.

akorneta (approved):
lgtm

dmytro-ndp:
@garagatyi: concerning this notice:
> Please squash this PR before merge

we can use "**Squash and merge**" button instead, can't we?

tolusha:
`TestFactory` has to be moved into core part of selenium tests.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/inject/SeleniumTestHandler.java
```diff
@@ -335,4 +343,9 @@ private void shutdown() {
 
   /** Returns parent injector. */
   public abstract Injector createParentInjector();
+
+  /** Returns list of child modules */
+  public List<Module> getChildModules() {
```
tolusha: Rename and make it abstract
`getTestClassModules`

selenium/che-selenium-test/src/test/resources/conf/selenium.properties
dmytro-ndp: Let's stay empty properties.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/inject/SeleniumTestHandler.java
```diff
@@ -335,4 +343,9 @@ private void shutdown() {
 
   /** Returns parent injector. */
   public abstract Injector createParentInjector();
```
tolusha: Let's return modules instead of injector.
`List<Module> getParentModules()`

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/user/TestUserImpl.java
```diff
@@ -47,12 +48,12 @@
...
   @Inject
   public TestUserImpl(
-      TestUserServiceClient userServiceClient,
+      Provider<TestUserServiceClient> userServiceClient,
       TestAuthServiceClient authServiceClient,
       TestWorkspaceServiceClientFactory wsServiceClientFactory)
       throws Exception {
     this(
-        userServiceClient,
+        userServiceClient.get(),
```
tolusha, line 56: You might pass `null` ^(

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/inject/SeleniumTestHandler.java
dmytro-ndp: This method is aimed to provide just child modules, and the list is only used as an argument of method **injector.createChildInjector()**. So, could you, please, explain, why we should rename it to **getTestClassModules** which can includes any kind of modules including parent ones?

Also, if we make **getTestClassModules** an abstract, we obligate implementations to realize this method to return empty list in case of Codenvy Onprem and Codenvy SaaS tests.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/user/TestUserImpl.java
dmytro-ndp: In which situation **null** could be there? There is **@Inject** annotation above the constructor, and **Guice** will throw an error instantly if it will not be able to create an instance of some parameter, will not it?
Otherwise we SHOULD check all parameters on _null_ in EACH constructors where **@Inject** is used, don't we?

garagatyi:
> we can use "Squash and merge" button instead, can't we?

yes

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/user/TestUserImpl.java
tolusha: Agree, null value won't be returned.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/inject/SeleniumTestHandler.java
tolusha: `getChildModules` ok. But make it abstract

dmytro-ndp: okay

tolusha (approved):

dmytro-ndp (approved):

PR #6562, opened 2017-10-03
Fixed issue where download outputs was returning null on workspace start

### LINKED ISSUE #360: Downloaded logs when "Starting workspace runtime" are empty
This is to fix https://github.com/openshiftio/openshift.io/issues/513

JPinkney:
### What does this PR do?
This PR fixes an issue where download outputs would return null while starting a workspace.

rh-che

I believe it would fix https://github.com/redhat-developer/rh-che/issues/360.
Additionally, it fixes eclipse/che behaviour but it does not have a created issue here.

#### Release Notes

N/A


benoitf:
ci-build

l0rd:
@vparfonov are you ok? Can we merge it?

benoitf:
@vparfonov ?

vparfonov (approved):

benoitf:
FYI fix is already in 6.x branch
https://github.com/eclipse/che/blob/che6/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/processes/panel/ProcessesPanelPresenter.java#L1357

Issue #6576, opened 2017-10-04
[Multi-user] Authentication is not enabled in golang based agents


garagatyi:
Authentication is not enabled in golang based Che agents, so anyone can access terminal and exec server.
To enable it add `-enable-auth` flag to command line of the agents. 
Also, we need to add the token to the connection to the terminal since IDE doesn't do this.
And check how this works in CLI, IDE.

**Reproduction Steps:**

**OS and version:**    

**Diagnostics:**

PR #6592, opened 2017-10-05
Enable authentication in golang-based agents

garagatyi:
### What does this PR do?
Also, adapt workspace links injection to provide the token in all needed places, so IDE client connects to the terminal with a token.

Related to #6576

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterModule.java
+++ b/assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterModule.java
@@ -175,6 +175,7 @@ public class WsMasterModule extends AbstractModule {
             "$HOME/che/terminal/che-websocket-terminal "
                 + "-addr :4411 "
                 + "-cmd ${SHELL_INTERPRETER} "
+                + "-enable-auth "
                 + "-enable-activity-tracking");
     bindConstant()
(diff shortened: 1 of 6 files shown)


assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterModule.java
```diff
@@ -175,13 +175,15 @@ protected void configure() {
             "$HOME/che/terminal/che-websocket-terminal "
                 + "-addr :4411 "
                 + "-cmd ${SHELL_INTERPRETER} "
+                + "-enable-auth "
```
sleshchenko: Should not it be defined only in `assemlby-multiuser/assembly-wsmaster-war`? Have you checked that single user is working with your changes?

garagatyi: yep, I'll rework it

garagatyi: I've moved binding to MultiUserCheWsMasterModule and CheWsMasterModule

sleshchenko (approved):

skabashnyuk (approved):

akorneta (approved):

Issue #6265, opened 2017-09-14
Remove Codenvy references in email templates.


tolusha:
Remove Codenvy references in email templates.

### Reference issue
https://github.com/codenvy/codenvy/issues/2389

slemeur:
The goal here is to replace the email templates with the branding from Che.

We need to:
- Replace Codenvy's references to Eclipse Che in titles and body of the emails.
- Change the header, so that it displays the Eclipse Che logo instead of Codenvy
- Change the footer to display Eclipse Che logo, instead of Codenvy one
- Replace the link in the footer to point to Eclipse Che website (when clicking on logo).
- Replace the link to Eclipse Che twitter + Eclipse Che Blog. 
- Remove LinkedIn and Facebook links since we don't have that
- Add a github icon, to point to our github repository

We will also replace some styles:
- The green Codenvy button, needs to be replace by the yellow `##fdb940ff`
- We will remove the All caps from the buttons
- We replace the background color by `#292c2fff`
- We replace font by Roboto

Here is a quick sample mockup (your implementation will be better ;)):
![screen shot 2017-09-18 at 09 24 27](https://user-images.githubusercontent.com/1636769/30532246-724e311c-9c53-11e7-8bef-551710e82241.png)

mmorhun:
Here we have two different types of email-templates:
- Keycloak related mails (verify email, reset password, ...)
- Codenvy related mails (team invite, organization renamed, ... )

This issue will cover Keycloak related emails, the others will be covered in [this](https://github.com/eclipse/che/issues/6335) issue, because we have to port/adapt related to it functionality first.

mmorhun:
We have some limitations here. Keycloak has its own email sender we cannot attach images as multipart in mails. Most of mail services blocks inline base64 images. So the only image in mails is Che logo. I used direct link to the Che logo from official site. But Gmail doesn't support svg images in mails, so for GMail users we have empty footer.

mmorhun:
Done for Keycloak's emails

PR #6597, opened 2017-10-05
Remove opacity from hex formatted colors

mmorhun:
### What does this PR do?
Removes opacity value from hex formatted colors in email styles. This is caused by some email senders (like GMail) which cannot send it correctly. To support wider range of senders we have to remove them.

https://github.com/eclipse/che/issues/6265


riuvshin (approved):

mmorhun:
@benoitf take a look please

mmorhun:
@benoitf take a look please

benoitf (approved):

Issue #5362, opened 2017-06-13
Multi-Users for Che


skabashnyuk:
Adding authentification - is a very frequent request from Eclipse Che community. We think that Keycloak is a rich framework that can provide many interesting features http://www.keycloak.org/about.html

From a technical perspective, it would be a separate assembly inside of Che. We don't want to remove an ability for plugin developers to have an easy and simple assembly without authentification and any heavy components for easy testing of their work. We might reuse the same approach like we did with "single port" feature. When changing some specific parameter on configuration will change the number of containers that would be running. In this case, Keycloak will be running as a separate container.
- [x] Filter to transform Keycloak user to Eclipse Che subject 
- [x] Use Keycloak token to authenticate request in DefaultHttpJsonRequestFactory from wsmaster to wsmaster. https://github.com/eclipse/che/issues/5427
- [x] Keycloak client configurations https://github.com/eclipse/che/issues/5435
- [x] Keycloak server configuration https://github.com/eclipse/che/issues/5436
- [x]  Ws-agent authentication https://github.com/eclipse/che/issues/5434
- [x] CQ for keycloak jars https://github.com/eclipse/che/issues/5437
- [x] Synchronize user profile with keycloak database  https://github.com/eclipse/che/issues/5438
- [x] Customizing site pages related to authentication and Keycloak https://github.com/eclipse/che/issues/5786
- [x] Revise functionality of Profile & Security page on dashboard https://github.com/eclipse/che/issues/5788
- [x] The use of GWT `Frame` objects is not compatible with the Keycloak integration https://github.com/eclipse/che/issues/5745
- [x] Configure PostgreSQL backend for multiuser Che https://github.com/eclipse/che/issues/5836
- [x] Che-multiuser: Not able to create workspace and import sample in same time https://github.com/eclipse/che/issues/5889
- [x] Adapt CLI to use keycloak openid authentication  https://github.com/eclipse/che/issues/5893
- [x] Keycloak token expiration https://github.com/eclipse/che/issues/5927 
- [x] Run QA in multiuser branch and figure out what should be done to adapt QA to SPI 
 https://github.com/eclipse/che/issues/5930
 - [x] Che multiuser: allow to configure database settings over environment variables of che-server image https://github.com/eclipse/che/issues/5962
 - [x] Port server side of Permissions, organization/team , resource management from Codenvy to Eclipse Che https://github.com/eclipse/che/issues/6041
[...]

PR #6606, opened 2017-10-06
Move dependency declaration to che-deps repo

skabashnyuk:
### What does this PR do?
Move dependency declaration to che-deps repo
related to https://github.com/eclipse/che-dependencies/pull/57

n/a

#### Release Notes
n/a

#### Docs PR
n/a

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,6 @@
         <che.docs.version>5.19.0-SNAPSHOT</che.docs.version>
         <che.lib.version>5.19.0-SNAPSHOT</che.lib.version>
         <che.version>5.19.0-SNAPSHOT</che.version>
-        <com.dumbster.version>1.7.1</com.dumbster.version>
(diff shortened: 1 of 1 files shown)


sleshchenko:
Should not we move [jsr305](https://github.com/eclipse/che/blob/938935acfa14956284de60083a0d54df65c3ac02/pom.xml#L56) dependency also somewhere like `che-dependencies` project, as it is not `eclipse che` artifact?

skabashnyuk:
@sleshchenko we could but we don't want to. It's here because it's needed for gwt compilation. But we don't want to give developers signal to use it in compile scope mode because it's declared as build only.

sleshchenko:
@skabashnyuk Thanks for explanation

vparfonov (approved):

riuvshin (approved):

Issue #6406, opened 2017-09-22
SPI: Port AppStatesPreferenceCleaner to spi branch


garagatyi:
Looks like preference cleaning was implemented in master and not ported correctly in spi branch.
We don't have this feature in spi but have a class that implements it in abandoned module che-plugin-docker-machine. Looks like it is the result of the merge of the master in the spi.

**Reproduction Steps:**

**OS and version:**    

**Diagnostics:**

azatsarynnyy:
fixed by https://github.com/eclipse/che/pull/6611

PR #6611, opened 2017-10-06
Port AppStatesPreferenceCleaner to spi branch

azatsarynnyy:
### What does this PR do?
Ports [`AppStatesPreferenceCleaner`](https://github.com/eclipse/che/blob/master/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/cleaner/AppStatesPreferenceCleaner.java) component from the `master` branch. Class has been missed in result of merges master>spi

#6406


garagatyi (approved):

gazarenkov (commented):
To me it is more logical to keep it near Preferences (i.e User API) 
Dependency conflict with WorkspaceRemovedEvent should be resolved moving it to shared module (it should be the good practice to keep Events there since events are considered as cross-modules entity)

garagatyi (approved):

azatsarynnyy:
PR has been updated:
* `AppStatesPreferenceCleaner` moved to `che-core-api-user`;
* `WorkspaceRemovedEvent` and `WorkspaceCreatedEvent` moved to `che-core-api-workspace-shared`.

gazarenkov (approved):

PR #6614, opened 2017-10-06
Running integrations tests on centos/postgresql-96-centos7 image

skabashnyuk:
### What does this PR do?
Running integrations tests on centos/postgresql-96-centos7 image
Same image as we are using on che-multiuser assembly.
Because user that we are using to run che have privileges to install extension
https://github.com/eclipse/che/blob/che-multiuser/dockerfiles/init/modules/postgres/templates/init-che-user.sh.erb#L16 I've also changed the user in the test to superuser.  This is the same behavior as if we run official image https://hub.docker.com/_/postgres/

#### Release Notes
n/a

#### Docs PR
n/a


sleshchenko:
Please take a look `integration tests` module in multiuser

akorneta (approved):

sleshchenko (approved):

Issue #1806, opened 2016-07-18
Language Server Protocol: Rename feature


vparfonov:
The rename request is sent from the client to the server to do a workspace wide rename of a symbol.
https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#rename

evidolob:
The result of rename is list changes that we should apply to files over workspace, I'm not sure that client(browser) is right place to do that. We need to thick how apply that changes on server(ws-agent).

tsmaeder:
@evidolob From the spec, I can't say if the delta you get back is describing an edit that has happened on the back end or if it's a "todo" for the client. Do you have more info?

tsmaeder:
@evidolob I think you are correct. However, we do have an agent in the workspace that could execute the change. We don't really have to have the exact same protocol between the che ide and the language server api ws-agent and the individual language server implementations. Not sure where filed buffers live in che, though.

tsmaeder:
Note that implementing infrastructure for applying workspace edits would also benefit https://github.com/eclipse/che/issues/3957

evidolob:
This issue depends on https://github.com/eclipse/che/pull/5955 pull request

PR #6615, opened 2017-10-06
Language Server Protocol Rename implementation

evidolob:
### What does this PR do?
Add implementation of LSP Rename

#1806

#### Release Notes

Language Server Protocol Rename implementation


wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/service/TextDocumentService.java
```diff
@@ -776,6 +800,115 @@ public boolean handleResult(
...
+    }
+  }
+
+  private List<ExtendedTextEdit> convertToExtendedEdit(List<TextEdit> edits, String filePath) {
+    try {
+      // for some reason C# LS sends ws related path,
+      if (!isStartWithProject(filePath)) {
+        filePath = prefixProject(filePath);
+      }
+      String fileContent =
```
tsmaeder, line 881: Sorry to chime in late, but since we'll face a similar problem with the "references" for the jdt.ls, etc. I think this approach will not scale for large sets of changes. We should be thinking towards a UI where we fetch the concrete line changes ("snippets") when we need them.

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/LanguageServerExtension.java
```diff
@@ -134,12 +150,15 @@ public void onFileOperation(final FileEvent event) {
                 break;
             }
           }
-          //            onOpen(event.getEditor(), event.getFile(), dtoFactory, serviceClient, fileTypeRegister);
+          //            onOpen(event.getEditor(), event.getFile(), dtoFactory, serviceClient,
```
vparfonov: comment code

vparfonov (approved):

PR #6618, opened 2017-10-06
Adopt replace stacks script to multi user

riuvshin:
### What does this PR do?
Adopt replace stacks script to multi user

### DIFF
--- a/dockerfiles/init/modules/openshift/files/scripts/replace_stacks.sh
+++ b/dockerfiles/init/modules/openshift/files/scripts/replace_stacks.sh
@@ -8,25 +8,41 @@
 
 set -e
 
+
+DEFAULT_CHE_ADMIN_USERNAME="admin"
+DEFAULT_CHE_ADMIN_PASSWORD="admin"
+CHE_ADMIN_USERNAME=${CHE_ADMIN_USERNAME:-${DEFAULT_CHE_ADMIN_USERNAME}}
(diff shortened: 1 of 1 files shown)


davidfestal (commented):
This will work when the Keycloak dedicated server is deployed. But not when the OSIO Keycloak server is used instead (ie when `DEFAULT_CHE_KEYCLOAK_DISABLED` is `true` in `deploy_che.sh`).
Because I doubt we can connect as `admin/admin` to get the Keycloak token.

riuvshin:
@davidfestal good point! will fix that!

riuvshin:
@davidfestal Im a bit confused now, becuase https://github.com/eclipse/che/blob/che-multiuser/dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh#L100-L101

```
if [ "${CHE_MULTI_USER}" == "true" ]; then
  DEFAULT_CHE_KEYCLOAK_DISABLED="false"
```

so if this is multi user it will always have `  DEFAULT_CHE_KEYCLOAK_DISABLED="false"`

davidfestal:
> if this is multi-user it will always have DEFAULT_CHE_KEYCLOAK_DISABLED="false"

Oh, sorry, I was speaking of `CHE_DEDICATED_KEYCLOAK`.
of the case when you set this variable to `false` on the command line.

benoitf:
on https://www.shellcheck.net/ I have multiple warnings on CHE_MULTIUSER (should be quoted)

skabashnyuk:
PR for Che6?

riuvshin:
as in CHE6 any stack should work on openshift we don't need this anymore.

Issue #6160, opened 2017-09-06
Import new project from other client (not IDE) works wrong: new project not represented in Project Explorer 


vparfonov:
**Reproduction Steps:**
1. Create/start new workspace
2. Try to import new project via calling Project REST API with curl (replace IP with your)
```
curl 'http://172.19.20.233:32871/api/project/import/spring' -H 'Origin: http://172.19.20.233:8080' -H 'Accept-Encoding: gzip, deflate' -H 'Accept-Language: en-US,en;q=0.8,ru;q=0.6,uk;q=0.4' -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/57.0.2987.133 Safari/537.36' -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'Referer: http://172.19.20.233:8080/che/wksp-8wst' -H 'Connection: keep-alive' --data-binary '{"location":"https://github.com/che-samples/web-java-spring.git", "type":"git", "parameters":{}}' --compressed
```
New project should represent in Project Explorer but not. After pressing F5 projects are shown correctly.

PR #6628, opened 2017-10-09
Send events from server to client about project's changes

svor:
### What does this PR do?
WsAgent has an ability to communicate with clients if some changes were happened  with projects (create, update, import, delete). 
For example if the project was created by calling API request beyond the IDE, the server will send events to each registered clients that some changes in workspace was happened. 

How it works:
1. A client initialised then he sends an event to the server about that by using `ServerSubscriptionBroadcaster`;
2. The server (`ClientSubscriptionHandler`) saves client's endpoint id into some set;
3. When some changes were happened on the server, he (`WorkspaceSyncCommunication`) sends synchronization event to each registered endpointId;
4. Each client (`WorkspaceProjectsSyncer`) catches this event and updates the workspace. 

https://github.com/eclipse/che/issues/6160

#### Release Notes

N/A

#### Docs PR

N/A


svor:
ci-test

dkulieshov (approved):

vparfonov (approved):

Issue #6346, opened 2017-09-19
Fix unstable and outdated selenium tests


dmytro-ndp:
There are several selenium tests which are unstable, that are fail from time to time, or outdated.
According to the [for recent build 43 result](https://ci.codenvycorp.com/view/qa/job/che-integration-tests/43/#showFailuresLink) we have 71 tests failures.
We need to sort out regular tests execution result to recognize those of them which failed because of inaccurate code to achieve 100% selenium tests success execution:
- [x] [CheckBasicFunctionalityInCommandsExplorerTest.shouldCreateAndRunBuildCommand](https://ci.codenvycorp.com/view/qa/job/che-integration-tests/64/testReport/junit/org.eclipse.che.selenium.intelligencecommand/CheckBasicFunctionalityInCommandsExplorerTest/shouldCreateAndRunBuildCommand/)
- [x] [CheckBasicFunctionalityInCommandsExplorerTest.shouldCreateDifferentTypesCommandsInDifferentGoals](https://ci.codenvycorp.com/view/qa/job/che-integration-tests/64/testReport/junit/org.eclipse.che.selenium.intelligencecommand/CheckBasicFunctionalityInCommandsExplorerTest/shouldCreateDifferentTypesCommandsInDifferentGoals/)
- [x] [CheckFindActionFeatureInCheTest](https://ci.codenvycorp.com/view/qa/job/che-integration-tests/56/testReport/junit/org.eclipse.che.selenium.editor/ContextMenuEditorTest/) (#6524)
- [x] [CommandsPaletteTest.newCommandTest](https://ci.codenvycorp.com/view/qa/job/che-integration-tests/55/testReport/org.eclipse.che.selenium.intelligencecommand/CommandsPaletteTest/newCommandTest/) It looks like this bug has been fixed by https://github.com/eclipse/che/pull/6511.
- [x] [ContextMenuEditorTest](https://ci.codenvycorp.com/view/qa/job/che-integration-tests/56/testReport/junit/org.eclipse.che.selenium.editor/ContextMenuEditorTest/) (#6524)
- [x] [CreateNamedFactoryFromDashBoard.createFactoryFromDashBoard](https://ci.codenvycorp.com/view/qa/job/che-integration-tests/50/testReport/org.eclipse.che.selenium.factory/CreateNamedFactoryFromDashBoard/createFactoryFromDashBoard/) (#6440)
- [x] DebugExternalClassTest
[...]

PR #6632, opened 2017-10-09
Fix unstable CheckBasicFunctionalityInCommandsExplorerTest selenium test

SkorikSergey:
### What does this PR do?
We need to fix unstable CheckBasicFunctionalityInCommandsExplorerTest selenium test from intelligencecommand package:

**CheckBasicFunctionalityInCommandsExplorerTest** - sometimes(when starting on CI) the test fails while try to click on the duplicate command button.

https://github.com/eclipse/che/issues/6346

### DIFF
--- a/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/intelligent/CommandsExplorer.java
+++ b/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/intelligent/CommandsExplorer.java
@@ -184,9 +184,10 @@ public class CommandsExplorer {
   }
 
   public void cloneCommandByName(String commandName) {
+    loader.waitOnClosed();
     new WebDriverWait(seleniumWebDriver, LOAD_PAGE_TIMEOUT_SEC)
         .until(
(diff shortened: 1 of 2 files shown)


tolusha (approved):

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/intelligent/CommandsExplorer.java
```diff
@@ -184,9 +184,10 @@ public void waitCommandInExplorerByName(String commandName) {
   }
 
   public void cloneCommandByName(String commandName) {
+    loader.waitOnClosed();
     new WebDriverWait(seleniumWebDriver, LOAD_PAGE_TIMEOUT_SEC)
         .until(
-            ExpectedConditions.visibilityOfElementLocated(
+            ExpectedConditions.elementToBeClickable(
                 By.xpath(
```
musienko-maxim, line 191: Why do we use elementToBeClickable in this place?

musienko-maxim (approved):

PR #6638, opened 2017-10-09
Don't fail if SSH agent doesn't start

l0rd:
### What does this PR do?
In the case of a workspace run by an unprivileged user the bootstrap won't fail when the SSH agent fail to start. Failure of the SSH agent is not blocking a workspace creation anymore. That applies to SSH agent only, failures to other agents will result in aborting workspace creation.

https://github.com/redhat-developer/rh-che/issues/346

#### Release Notes
Workspaces can start even if SSH service is not started (e.g. if the workspace user is not a privileged user).

### DIFF
--- a/agents/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/AbstractAgentLauncher.java
+++ b/agents/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/AbstractAgentLauncher.java
@@ -102,7 +102,14 @@ public abstract class AbstractAgentLauncher implements AgentLauncher {
               "Fail launching agent '%s' in '%s' workspace due to timeout",
               agent.getName(), machine.getWorkspaceId()));
 
-      process.kill();
+      if (this.shouldBlockMachineStartOnError()) {
+        LOG.info("Stopping workspace {}", machine.getWorkspaceId());
+        process.kill();
+      } else {
+        LOG.info(
+            "Continuing workspace bootstrap even if agent {} failed to start.", agent.getName());
+        return;
+      }
     } catch (MachineException e) {
       logAsErrorAgentStartLogs(machine, agent.getName(), agentLogger.getText());
       throw new ServerException(e.getServiceError());
@@ -171,5 +178,10 @@ public abstract class AbstractAgentLauncher implements AgentLauncher {
           machine.getId(),
           machine.getNode().getHost());
     }
+  }
+
+  @Override
+  public boolean shouldBlockMachineStartOnError() {
+    return true;
   }
 }
--- a/agents/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/AgentLauncher.java
+++ b/agents/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/AgentLauncher.java
@@ -31,6 +31,12 @@ public interface AgentLauncher {
   String getMachineType();
 
   /**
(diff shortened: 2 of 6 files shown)


l0rd:
cc @eivantsov

agents/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/AbstractAgentLauncher.java
```diff
@@ -172,4 +180,9 @@ void logAsErrorAgentStartLogs(Instance machine, String agentName, String logs) {
           machine.getNode().getHost());
     }
   }
+
+  @Override
+  public boolean blockMachineStart() {
```
benoitf: the name of the method looks odd for me (I was not expecting that is was returning something)

benoitf:
ci-build

agents/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/AbstractAgentLauncher.java
l0rd: @benoitf any suggestion?

benoitf:
@riuvshin is that all committers could have automatically have their PR checked without requiring another user action ?

agents/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/AbstractAgentLauncher.java
benoitf: isSomething() ?

l0rd: `shouldBlockMachineStart()`?

benoitf: @l0rd yep I think it's clear as well

benoitf: but maybe add that it's in case of error.
shouldBlockMachineStartIfError() (but I agree it's a bit long :D )

or there is the usual "failOnError()" method names

benoitf (approved):

agents/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/AbstractAgentLauncher.java
```diff
@@ -97,12 +97,20 @@ public void writeLine(String line) throws IOException {
           Thread.sleep(agentPingDelayMs);
         }
       }
+
```
benoitf: maybe we can delete this line of this PR

l0rd: done

agents/che-core-api-agent/src/main/java/org/eclipse/che/api/agent/server/launcher/AbstractAgentLauncher.java
l0rd: I changed to `shouldBlockMachineStartOnError()` :-)

ghost (approved):

tolusha (approved):

PR #6639, opened 2017-10-09
Workspace start code cleanup

garagatyi:
### What does this PR do?
Moved async operations from WorkspaceManager to WorkspaceRuntimes
to have an async facility in one place instead of two.
Moved workspace start/stop logging from WorkspaceManager
to WorkspaceRuntimes since WorkspaceManager can not correctly log
them.
Improved logging of workspace start/stop including the addition of new logs.

n/a

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
@@ -11,20 +11,17 @@
 package org.eclipse.che.api.workspace.server;
 
 import static com.google.common.base.MoreObjects.firstNonNull;
-import static com.google.common.base.Throwables.getCausalChain;
 import static java.lang.String.format;
 import static java.lang.System.currentTimeMillis;
 import static java.util.Objects.requireNonNull;
 import static org.eclipse.che.api.core.model.workspace.WorkspaceStatus.RUNNING;
 import static org.eclipse.che.api.core.model.workspace.WorkspaceStatus.STARTING;
 import static org.eclipse.che.api.core.model.workspace.WorkspaceStatus.STOPPED;
-import static org.eclipse.che.api.workspace.shared.Constants.WORKSPACE_STOPPED_BY;
 
 import com.google.inject.Inject;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.CompletableFuture;
 import javax.inject.Singleton;
 import org.eclipse.che.account.api.AccountManager;
 import org.eclipse.che.account.shared.model.Account;
@@ -38,8 +35,6 @@ import org.eclipse.che.api.core.model.workspace.WorkspaceStatus;
 import org.eclipse.che.api.core.notification.EventService;
 import org.eclipse.che.api.workspace.server.model.impl.WorkspaceConfigImpl;
 import org.eclipse.che.api.workspace.server.model.impl.WorkspaceImpl;
-import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
-import org.eclipse.che.api.workspace.server.spi.InternalInfrastructureException;
 import org.eclipse.che.api.workspace.server.spi.WorkspaceDao;
 import org.eclipse.che.api.workspace.shared.event.WorkspaceCreatedEvent;
 import org.eclipse.che.commons.annotation.Nullable;
@@ -69,7 +64,6 @@ public class WorkspaceManager {
   private final WorkspaceDao workspaceDao;
   private final WorkspaceRuntimes runtimes;
   private final AccountManager accountManager;
-  private final WorkspaceSharedPool sharedPool;
   private final EventService eventService;
   private final WorkspaceValidator validator;
 
@@ -79,13 +73,11 @@ public class WorkspaceManager {
       WorkspaceRuntimes runtimes,
       EventService eventService,
       AccountManager accountManager,
-      WorkspaceValidator validator,
-      WorkspaceSharedPool sharedPool) {
+      WorkspaceValidator validator) {
     this.workspaceDao = workspaceDao;
     this.runtimes = runtimes;
     this.accountManager = accountManager;
     this.eventService = eventService;
-    this.sharedPool = sharedPool;
     this.validator = validator;
   }
 
@@ -273,7 +265,7 @@ public class WorkspaceManager {
     }
 
     workspaceDao.remove(workspaceId);
-    LOG.info("Workspace '{}' removed by user '{}'", workspaceId, sessionUserNameOr("undefined"));
+    LOG.info("Workspace '{}' removed by user '{}'", workspaceId, sessionUserNameOrUndefined());
   }
 
   /**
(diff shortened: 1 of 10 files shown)


skabashnyuk (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -305,26 +324,53 @@ public void stop(String workspaceId, Map<String, String> options)
       throw new ConflictException(
           format("Could not stop workspace '%s' because its state is '%s'", workspaceId, status));
     }
+    String stoppedBy =
+        firstNonNull(
+            sessionUserNameOr(workspace.getAttributes().get(WORKSPACE_STOPPED_BY)), "undefined");
```
sleshchenko: Is it possible to have here the same method as WorkspaceManager has? I mean `sessionUserNameOrUndefined`?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -305,26 +324,53 @@ public void stop(String workspaceId, Map<String, String> options)
...
+              } catch (InfrastructureException e) {
+                // remove before firing an event to have consistency between state and the event
+                runtimes.remove(workspaceId);
+                LOG.info(
+                    "Workspace '{}/{}' with id '{}' stopped by user '{}'. Error: {}",
+                    workspace.getNamespace(),
+                    workspace.getConfig().getName(),
+                    workspace.getId(),
+                    stoppedBy,
+                    e);
```
sleshchenko: Are you sure that `Logger` will process exception object correctly? Mb it'd be better to log `e.getMessage`?

sleshchenko (approved):
LGTM

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
garagatyi: This is a bit another use-case

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -344,13 +334,23 @@ public void stopWorkspace(String workspaceId, Map<String, String> options)
 
     requireNonNull(workspaceId, "Required non-null workspace id");
     final WorkspaceImpl workspace = normalizeState(workspaceDao.get(workspaceId), true);
-    checkWorkspaceIsRunningOrStarting(workspace, "stop");
```
akorneta, line 347: I think we need this check because when workspace state is different to `STARTING` / `RUNNING`  we will update its attributes and try to stop it.

akorneta (approved):
other LGTM

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
garagatyi: I moved it back

garagatyi:
@sleshchenko @skabashnyuk I reworked PR since your review, can you take one more look?

garagatyi:
ci-build

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -272,23 +256,74 @@ public WorkspaceStatus getStatus(String workspaceId) {
...
+        String failureCause = "failed";
+        if (e instanceof RuntimeStartInterruptedException) {
+          failureCause = "interrupted";
+          publishWorkspaceStatusEvent(workspaceId, STOPPED, STARTING, e.getMessage());
+        }
+        LOG.info(
+            "Workspace '{}:{}' with id '{}' start {}",
+            workspace.getNamespace(),
+            workspace.getConfig().getName(),
+            workspace.getId(),
```
sleshchenko, line 297: Should not `workspaceId` variable be here? 
And are you sure that workspaceId should be `field` but not a local variable? It's OK for and it is just a question =)

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -272,23 +256,74 @@ public WorkspaceStatus getStatus(String workspaceId) {
...
+            sessionUserNameOr("undefined"));
+        publishWorkspaceStatusEvent(workspaceId, RUNNING, STARTING, null);
+      } catch (InfrastructureException e) {
+        runtimes.remove(workspaceId);
+        String failureCause = "failed";
+        if (e instanceof RuntimeStartInterruptedException) {
+          failureCause = "interrupted";
+          publishWorkspaceStatusEvent(workspaceId, STOPPED, STARTING, e.getMessage());
```
sleshchenko, line 291: Should not be published an event that workspace fails to start in case failing but not the interruption?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -272,23 +256,74 @@ public WorkspaceStatus getStatus(String workspaceId) {
...
+            workspace.getConfig().getName(),
+            workspace.getId(),
+            failureCause);
+        // InfrastructureException is supposed to be an exception that can't be solved
+        // by Che admin, so should not be logged (but not InternalInfrastructureException).
+        // It will prevent bothering the admin when user made a mistake in WS configuration.
+        if (e instanceof InternalInfrastructureException) {
+          LOG.error(e.getLocalizedMessage(), e);
+        }
+        throw new RuntimeException(e);
```
sleshchenko, line 305: Is there any reason to throw the exception here? Maybe logging is enough?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
garagatyi: yep, good catch!

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
garagatyi: I didn't see much difference but will move to local var

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
garagatyi: It is needed to propagate the result (failure or success) to a possible calling side that may be interested in chaining other stages depending on success/failure of this task.

Issue #6650, opened 2017-10-10
Unable to log in with a user if an organization with the same name already exists


sleshchenko:
### Description
Unable to log in with a user if an organization with the same name already exists.

### Reproduction Steps
1. Log in as admin.
2. Create an organization with some name e.g. `test`.
3. Log out.
4. Try to register a new user with the same name.
Expected: User is registered and logged in successfully or gets an error with the message that he has to use another username.
Actual: User is created on `keycloak` server but the dashboard is not opened because of error while importing `keycloak` user into `che-server` database because of name uniqueness constraint.

riuvshin:
@skabashnyuk should we close issue?

PR #6674, opened 2017-10-11
Fixed possible constraint violation with an existed organization.

skabashnyuk:
### What does this PR do?
Usecase is following:
If we have already an organization with name, let's say "org"  and the new user also has name "org" then we will have constraint violation.
 In this case, we will try to create the user with name "org"+random string.

https://github.com/eclipse/che/issues/6650

n/a

#### Release Notes
n/a

#### Docs PR
n/a

### DIFF
--- a/multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/KeycloakEnvironmentInitalizationFilter.java
+++ b/multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/KeycloakEnvironmentInitalizationFilter.java
@@ -99,23 +99,24 @@ public class KeycloakEnvironmentInitalizationFilter extends AbstractKeycloakFilt
     }
   }
 
-  private synchronized User getOrCreateUser(String id, String email, String username)
-      throws ServletException {
+  private User getOrCreateUser(String id, String email, String username) throws ServletException {
     try {
       return userManager.getById(id);
     } catch (NotFoundException e) {
-      final UserImpl cheUser = new UserImpl(id, email, username, generate("", 12), emptyList());
-      try {
-        return userManager.create(cheUser, false);
-      } catch (ConflictException ex) {
+      synchronized (this) {
+        final UserImpl cheUser = new UserImpl(id, email, username, generate("", 12), emptyList());
         try {
-          cheUser.setName(generate(cheUser.getName(), 4));
           return userManager.create(cheUser, false);
-        } catch (ServerException | ConflictException e1) {
-          throw new ServletException("Unable to create new user", e1);
+        } catch (ConflictException ex) {
+          try {
+            cheUser.setName(generate(cheUser.getName(), 4));
+            return userManager.create(cheUser, false);
+          } catch (ServerException | ConflictException e1) {
+            throw new ServletException("Unable to create new user", e1);
+          }
+        } catch (ServerException ex) {
+          throw new ServletException("Unable to create new user", ex);
         }
-      } catch (ServerException ex) {
-        throw new ServletException("Unable to create new user", ex);
       }
     } catch (ServerException e) {
       throw new ServletException("Unable to get user", e);
--- a/multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/KeycloakEnvironmentInitalizationFilter.java
+++ b/multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/KeycloakEnvironmentInitalizationFilter.java
@@ -80,15 +80,21 @@ public class KeycloakEnvironmentInitalizationFilter extends AbstractKeycloakFilt
         throw new ServletException("Cannot detect or instantiate user.");
       }
       Claims claims = (Claims) jwtToken.getBody();
-      User user =
-          getOrCreateUser(
-              claims.getSubject(),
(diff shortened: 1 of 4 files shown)


wsmaster/che-core-api-account/src/test/java/org/eclipse/che/account/spi/tck/AccountDaoTest.java
```diff
@@ -96,6 +96,14 @@ public void shouldThrowConflictExceptionWhenUpdatingAccountWithExistingName() th
     accountDao.update(account);
   }
 
+  @Test(expectedExceptions = ConflictException.class)
+  public void shouldThrowConflictExceptionWhenCreatingAccountWithExistingName() throws Exception {
+    AccountImpl account = accounts[0];
+    account.setName(accounts[1].getName());
+
+    accountDao.create(account);
```
sleshchenko: You try to create user with the existing name and id. To ensure that conflict is thrown because of names conflct it is needed to rewrite this test as
```
AccountImpl account = new AccountImpl(NameGenerator.generate("account", 5), accounts[0].getName(), "test");

accountDao.create(account);
```

wsmaster/che-core-api-account/src/main/java/org/eclipse/che/account/spi/jpa/JpaAccountDao.java
```diff
@@ -111,7 +111,9 @@ public void remove(String id) throws ServerException {
 
   @Transactional
   protected void doCreate(AccountImpl account) {
-    managerProvider.get().persist(account);
+    final EntityManager manager = managerProvider.get();
+    manager.persist(account);
+    manager.flush();
```
sleshchenko, line 115: Please add flushing to `doRemove` method too. See [UserDao](https://github.com/eclipse/che/blob/6e9825c699d34eef6b8bf36fc0194ff871cbee84/wsmaster/che-core-api-user/src/main/java/org/eclipse/che/api/user/server/jpa/JpaUserDao.java#L316).

multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/KeycloakEnvironmentInitalizationFilter.java
```diff
@@ -103,10 +104,17 @@ private synchronized User getOrCreateUser(String id, String email, String userna
...
       try {
-        final UserImpl cheUser = new UserImpl(id, email, username, "secret", emptyList());
         return userManager.create(cheUser, false);
-      } catch (ServerException | ConflictException ex) {
+      } catch (ConflictException ex) {
+        try {
+          cheUser.setName(generate(cheUser.getName(), 4));
+          return userManager.create(cheUser, false);
+        } catch (ServerException | ConflictException e1) {
+          throw new ServletException("Unable to create new user", e1);
```
sleshchenko: With such try catches this method looks unreadable. Please consider to rewrite it somehow, e.g.
```
  private synchronized User getOrCreateUser(String id, String email, String username)
      throws ServletException {
    try {
      Optional<User> user = doGetUser(id);

      if (user.isPresent()) {
        return user.get();
      }

      final UserImpl cheUser = new UserImpl(id, email, username, generate("", 12), emptyList());
      try {
        return userManager.create(cheUser, false);
      } catch (ConflictException ex) {
        cheUser.setName(generate(cheUser.getName(), 4));
        return userManager.create(cheUser, false);
      }
    } catch (ServerException | ConflictException e) {
      throw new ServletException("Unable to create new user", e);
    }
  }

  private Optional<User> doGetUser(String id) throws ServerException {
    try {
      return Optional.of(userManager.getById(id));
    } catch (NotFoundException e) {
      return Optional.empty();
    }
  }
```

wsmaster/che-core-api-account/src/main/java/org/eclipse/che/account/spi/jpa/JpaAccountDao.java
mshaposhnik, line 115: Pls fix AccountDao.doRemove() in the same way.

mshaposhnik (approved):

multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/KeycloakEnvironmentInitalizationFilter.java
mshaposhnik: 1:1 by readability imo...:)

sleshchenko: BTW `throw new ServletException("Unable to create new user", e);` this line is one time in my variant :smiley:

wsmaster/che-core-api-account/src/main/java/org/eclipse/che/account/spi/jpa/JpaAccountDao.java
skabashnyuk: done

skabashnyuk: done

sleshchenko (approved):

skabashnyuk:
ci-test

vparfonov (approved):

Issue #6481, opened 2017-09-28
Remove rewrite  /wsmaster/api/* to /api/*


skabashnyuk:
A long time ago requests to ws-master were like ```http://localhost:8080/api/user```. 
After we decided to change the name of ```api``` war to ```wsmaster```.
To make this change compatible with the previous version we have a rewrite rule.
```RewriteRule ^/api/(.*)$     /wsmaster/api/$1    [L]```
After that, part of our application are making request to ```/api``` some to ```/wsmaster/api```
Internally this solution works more or less. But eventually, it produces small chaos and unpredictable problems. Like I have redirection loop with Keycloak Opeid Filter because I have a request to ```/api``` but it was internally transformed to ```/wsmaster/api``` and result cookies was set on the path ```/wsmaster/api``` but it wasn't expected. 

I think it's time to drop this workaround with RewriteRule in next version of Che.

**Comment by benoitf** [comment]
I would say it was also to be consistent between che and codenvy across all clients

and you can use other flags than `[L]`

**Comment by skabashnyuk** [comment]
Yes. And after Che6 this consistency is not required, it means that is a good time for drop off.

**Comment by mshaposhnik** [comment]
There is no other ways to keep `/api` path than:
 - Rewrite valve;
-  Define app configuration in _Tomcat/conf/server.xml_ and override `Path`
-  Rename war;

@skabashnyuk @benoitf  what we decide ?

**Comment by benoitf** [comment]
@mshaposhnik I would say that the goal was to get rid of the valve so there is option 2 and option 3
and option 3, @gazarenkov does not want to rename it so we've only option 2

We may also add a file into conf/ folder which avoid to modify server.xml

- move wsmaster.war to `tomcat/` folder instead of `tomcat/webapps` folder
- add `conf/Catalina/localhost/api.xml` with content

```
<Context 
  docBase="../wsmaster.war" 
  reloadable="true" 
/>
```

Another topic could be to get rid of tomcat and use Jetty or Undertow

skabashnyuk:
lets go with ```Define app configuration in Tomcat/conf/server.xml and override Path```

benoitf:
AFAIK it's not the best practice to modify server.xml for extensibility.

gazarenkov:
I am OK with any, just see no point to make it more difficult with files manipulation as we already have server.xml in the assembly and it does the job

benoitf:
up to you guys

skabashnyuk:
[...]

PR #6678, opened 2017-10-11
Remove implicit redirection from  `/api` to `/wsmaster/api` for all requests.

mshaposhnik:
### What does this PR do?
Removes implicit redirection from `/api` to `/wsmaster/api` by Tomcat valve.

https://github.com/eclipse/che/issues/6481

#### Release Notes
Removed internall rewrite  `/api` to `/wsmaster/api` 

#### Docs PR

https://github.com/eclipse/che-docs/pull/301/files

### DIFF
--- a/assembly-multiuser/assembly-wsmaster-war/src/main/webapp/META-INF/context.xml
+++ b/assembly-multiuser/assembly-wsmaster-war/src/main/webapp/META-INF/context.xml
@@ -11,7 +11,7 @@
       Red Hat, Inc. - initial API and implementation
 
 -->
-<Context allowCasualMultipartParsing="true" sessionCookiePath="/api">
+<Context allowCasualMultipartParsing="true">
 
     <Resource name="jdbc/che" auth="Container"
               type="javax.sql.DataSource"
--- a/assembly/assembly-ide-war/src/main/webapp/WEB-INF/rewrite.config
+++ b/assembly/assembly-ide-war/src/main/webapp/WEB-INF/rewrite.config
@@ -1,5 +1,4 @@
 
-RewriteRule ^/api/(.*)$     /wsmaster/api/$1    [L]
 RewriteRule ^/factory/resources/(.*)$     /_app/$1    [L]
 RewriteRule ^/factory$     /dashboard/#/load-factory/    [R,NE,QSA]
 RewriteRule ^/factory/$    /dashboard/#/load-factory/    [R,NE,QSA]
--- a/assembly/assembly-wsmaster-war/src/main/webapp/META-INF/context.xml
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/META-INF/context.xml
@@ -11,7 +11,7 @@
       Red Hat, Inc. - initial API and implementation
 
 -->
-<Context allowCasualMultipartParsing="true" sessionCookiePath="/api">
+<Context allowCasualMultipartParsing="true">
 
     <Resource name="jdbc/che" auth="Container"
               type="javax.sql.DataSource"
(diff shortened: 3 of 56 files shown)


benoitf:
It's more linked to the issue than the PR itself, but why wsmaster is deployed at /wsmaster and not on /api ?

skabashnyuk:
AFAIK @TylerJewell decided to rename it.

benoitf:
ok because it looks like it will be more simpler to deploy wsmaster.war on /api as it's basically only providing API stuff.

skabashnyuk (approved):

slemeur (changes requested):
I don't see the value. These changes are going to make the api less readable.

WSMaster is an internal component - it does not have to be exposed in the URL.

`wsmaster/api/factory` for example is looking very bad.

skabashnyuk:
@slemeur what is the value to make it readable. It's not something that end user will see or will have to input. Because that of my technical motivation was to make it work in one way.   ```/wsmaster/api```  or ```/api``` because it provides **real** technical issues.  So if you are against ```/wsmaster/api```  as it is right now - please propose and motivate another way.
BTW factory url that end-user will see **will not** change.

skabashnyuk:
>> @benoitf  ok because it looks like it will be more simpler to deploy wsmaster.war on /api as it's >> basically only providing API stuff.

I don't know. It's more looks like exercise for exercise. If @TylerJewell  in one time decided that wsmaster is more appropriate name then let's finalize this task and play with single url, not with two.
It's quite strange to have two type of url in Eclipse Che, one in Codenvy. I would like to clean up this mess once and for all.

benoitf:
@skabashnyuk I agree on the single URL but I would vote for /api

skabashnyuk:
No problem lets do it in separate issue. 
This one is more related to finalization of rename initiated by @TylerJewell

benoitf:
AFAIK we don't need a separate issue

slemeur:
@skabashnyuk : 
- When you are working on an API, you actually have "end-users" who are developers. It's all about API Developer Experience and yes, names, paths are importants in order to make an API readable.

So I confirm: `/api` is much more simpler, logical, easy to discover and readable than `/wsmaster/api`.

slemeur:
And because we are speaking about API DX, when doing those kind of things, you actually need to document it and notify your community.
I wonder how you can approve the PR, while there is no mention to be published in the release note and no information/help/guidance for our community on the change.

vparfonov:
wsmaster/api/ is logical for me because it really Workspace Master in our terminology

vparfonov (approved):

skabashnyuk:
One point to be ```wsmaster/api/``` basket is to distinct somehow from ws-agent url.
It would be easier for some kind incoming proxies to filter ws-master vs ws-agent urls if they do not start in the same way.

benoitf:
@skabashnyuk I'm not sure `incoming proxies` reason is helping much there.
AFAIK the URLs are not targeting the same port so it's easy to distinct. 
And I would say that for production you may want to use a single port (like with Træfik) then the hosts are different as well so it's easy as well.

slemeur:
@vparfonov : That's exactly the point. It is our terminology, which is logical for us but might not be for external contributors. 

At the end, it'll be your decision to move forward and to define what's the best. 
I wanted to give different elements, so that you can understand that this change is having a bigger impacts that what you may think at first. And those impacts also have to be managed (documentation, communication to the community, ...)

Making things simpler for us should not impact the ultimate goal which is to make the product simple for others (end-users, api consumer, community, contributors).

I give my -1 to the PR, as it is today because it does not meet the merge criterias - that you as maintainers should care about.

gazarenkov:
/api is more clear

but my concern is more that path (which is visible part and I can imagine the cases when it makes sense to have other than default) should be independent on name of war (which is implementation details)

mshaposhnik:
DOCS [PR](https://github.com/eclipse/che-docs/pull/301/files)

benoitf:
@gazarenkov yes I agree as well that the name of the war can be anything.

Issue #6479, opened 2017-09-28
Create A Factory Based On A Repo


kameshsampath:
When we create a  factory from a github repo, it fails to pick the `.factory.json` when the url tend to contain `.git`. Usually when we copy the URLs from GitHub it ends with `.git` e.g. *https://github.com/kameshsampath/che-boot-demo.git* . Because of this the factory create fails and uses the default `codenvy/ubuntu_jdk8` image ref https://github.com/eclipse/che/blob/5b373e6cf5bfeb16f278c4d750fe64361445eeb4/plugins/plugin-urlfactory/src/main/java/org/eclipse/che/plugin/urlfactory/URLFactoryBuilder.java#L64.

It will be nice to sanitise the repo url if it ends with `.git` and use that as `jsonFileLocation`.

**Reproduction Steps:**

Install Che on minishift baed on the instructions https://www.eclipse.org/che/docs/setup/openshift/index.html#deploy-che-on-minishift

**Execute the command** 
http://che-eclipse-che.192.168.64.8.nip.io/f?url=https://github.com/kameshsampath/che-boot-demo.git (modify the che url as per your install)

This will fail with the error as the Repo url is not found because of `.git` in the end 

```
Error when starting agent
Unable to start workspace agent. Error when trying to start the workspace agent: Start of environment 'che-boot-demo.git' failed. Error: We detected that a machine exited unexpectedly. This may be caused by a container in interactive mode or a container that requires additional arguments to start. Please check the container recipe.
```

Now try the same command like http://che-eclipse-che.192.168.64.8.nip.io/f?url=https://github.com/kameshsampath/che-boot-demo this will work well as the image now will pulled correctly.

**OS and version:**    
All OS 

**Diagnostics:** 

Minishift install as per https://www.eclipse.org/che/docs/setup/openshift/index.html#deploy-che-on-minishift

ghost:
And of course, codenvy/ubuntu_jdk is the wrong default image to use. It should be `eclipse/ubuntu_jdk8`

PR #6679, opened 2017-10-11
#6479 - Create workspace with factory json for urls ending with .git

kameshsampath:
Signed-off-by: Kamesh Sampath <kamesh.sampath@hotmail.com>

### What does this PR do?
* Change the default image from `codeenvy/ubuntu_jdk` to `eclipse/ubu…ntu_jdk8`
* Santize urls ending with .git to enable loading of factory files from git repos

#6479

### DIFF
--- a/plugins/plugin-urlfactory/src/main/java/org/eclipse/che/plugin/urlfactory/URLFetcher.java
+++ b/plugins/plugin-urlfactory/src/main/java/org/eclipse/che/plugin/urlfactory/URLFetcher.java
@@ -19,6 +19,8 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.net.URL;
 import java.net.URLConnection;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import javax.inject.Singleton;
 import javax.validation.constraints.NotNull;
@@ -40,6 +42,9 @@ public class URLFetcher {
   /** Maximum size of allowed data. (30KB) */
   protected static final long MAXIMUM_READ_BYTES = 30 * 1000;
 
+  /** The Compiled REGEX PATTERN that can be used for http|https git urls */
+  final Pattern GIT_HTTP_URL_PATTERN = Pattern.compile("(^http[s]?://.*)(.git$)");
+
   /**
    * Fetch the url provided and return its content To prevent DOS attack, limit the amount of the
    * collected data
@@ -50,7 +55,7 @@ public class URLFetcher {
   public String fetch(@NotNull final String url) {
     requireNonNull(url, "url parameter can't be null");
     try {
-      return fetch(new URL(url).openConnection());
+      return fetch(new URL(sanitized(url)).openConnection());
     } catch (IOException e) {
       // we shouldn't fetch if check is done before
       LOG.debug("Invalid URL", e);
@@ -88,5 +93,23 @@ public class URLFetcher {
    */
   protected long getLimit() {
     return MAXIMUM_READ_BYTES;
+  }
+
+  /**
+   * Simple method to sanitize the Git urls like &quot;https://github.com/demo.git&quot; or
+   * &quot;http://myowngit.example.com/demo.git&quot;
+   *
+   * @param url - the String format of the url
+   * @return if the url ends with .git will return the url without .git otherwise return the url as
+   *     it is
+   */
+  String sanitized(String url) {
+    if (url != null) {
+      final Matcher matcher = GIT_HTTP_URL_PATTERN.matcher(url);
+      if (matcher.find()) {
+        return matcher.group(1);
+      }
+    }
+    return url;
   }
 }
--- a/plugins/plugin-urlfactory/src/main/java/org/eclipse/che/plugin/urlfactory/URLFetcher.java
+++ b/plugins/plugin-urlfactory/src/main/java/org/eclipse/che/plugin/urlfactory/URLFetcher.java
@@ -43,7 +43,7 @@ public class URLFetcher {
   protected static final long MAXIMUM_READ_BYTES = 30 * 1000;
 
   /** The Compiled REGEX PATTERN that can be used for http|https git urls */
-  final Pattern GIT_HTTP_URL_PATTERN = Pattern.compile("(^http[s]?://.*)(.git$)");
+  final Pattern GIT_HTTP_URL_PATTERN = Pattern.compile("(?<sanitized>^http[s]?://.*)\\.git$");
 
   /**
    * Fetch the url provided and return its content To prevent DOS attack, limit the amount of the
@@ -107,7 +107,7 @@ public class URLFetcher {
     if (url != null) {
       final Matcher matcher = GIT_HTTP_URL_PATTERN.matcher(url);
       if (matcher.find()) {
-        return matcher.group(1);
+        return matcher.group("sanitized");
       }
     }
     return url;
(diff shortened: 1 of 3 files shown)


plugins/plugin-urlfactory/src/main/java/org/eclipse/che/plugin/urlfactory/URLFetcher.java
```diff
@@ -40,6 +42,9 @@
   /** Maximum size of allowed data. (30KB) */
   protected static final long MAXIMUM_READ_BYTES = 30 * 1000;
 
+  /** The Compiled REGEX PATTERN that can be used for http|https git urls */
+  final Pattern GIT_HTTP_URL_PATTERN = Pattern.compile("(^http[s]?://.*)(.git$)");
```
skabashnyuk: do we need to check ssh urls here?

plugins/plugin-urlfactory/src/main/java/org/eclipse/che/plugin/urlfactory/URLFetcher.java
```diff
@@ -89,4 +94,22 @@ public String fetch(@NotNull URLConnection urlConnection) {
...
+
+  /**
+   * Simple method to sanitize the Git urls like &quot;https://github.com/demo.git&quot; or
+   * &quot;http://myowngit.example.com/demo.git&quot;
+   *
+   * @param url - the String format of the url
+   * @return if the url ends with .git will return the url without .git otherwise return the url as
+   *     it is
+   */
+  String sanitized(String url) {
```
skabashnyuk, line 106: When can it happen?

plugins/plugin-urlfactory/src/main/java/org/eclipse/che/plugin/urlfactory/URLFetcher.java
kameshsampath: am not sure Eclipse factory accepts ssh urls, thats why I thought to check only http(s)

plugins/plugin-urlfactory/src/main/java/org/eclipse/che/plugin/urlfactory/URLFetcher.java
kameshsampath: when i do something like http://<my-che>/f?url=http://github.com/demo/demo.git to create a workspace from factory. As users has 99% chances of using this url `http://github.com/demo/demo.git` as they usually copy the url from github which includes .git. WDYT?

benoitf (approved):

plugins/plugin-urlfactory/src/main/java/org/eclipse/che/plugin/urlfactory/URLFetcher.java
skabashnyuk: I don't see a reason why not.

plugins/plugin-urlfactory/src/main/java/org/eclipse/che/plugin/urlfactory/URLFetcher.java
skabashnyuk: I don't understand this use-case. When do we need to sanitized url?

plugins/plugin-urlfactory/src/main/java/org/eclipse/che/plugin/urlfactory/URLFetcher.java
kameshsampath: Based on the official doc https://www.eclipse.org/che/docs/factory/creating/index.html I don’t see a case for ssh/git protocols. Hence set the pattern currently to http/https.

plugins/plugin-urlfactory/src/main/java/org/eclipse/che/plugin/urlfactory/URLFetcher.java
kameshsampath: Try doing a workspace creation using https://codenvy.io/f?url=https://github.com/eclipse/che, this will work as expected and load the right docker image suited for che development (Assuming it’s differnt from default Ubuntu JDK image). But try the same with https://codenvy.io/f?url=https://github.com/eclipse/che.git, note .git at the end of url parameter the factory creation fails as it will not be able to load .factory.json from the url https://codenvy.io/f?url=https://github.com/eclipse/che.git

kameshsampath:
@benoitf - i just noticed that had typo in my signing email, do i need to touch some file and repush again ? to make IP validation pass ?

benoitf:
@kameshsampath yes you need to amend the commit and force push into your branch

plugins/plugin-urlfactory/src/main/java/org/eclipse/che/plugin/urlfactory/URLFetcher.java
ghost: @skabashnyuk it happens when creating a Factory with a git url as a query param.

ghost (approved):

plugins/plugin-urlfactory/src/main/java/org/eclipse/che/plugin/urlfactory/URLFetcher.java
kameshsampath: @skabashnyuk for the url tests if you use my repo `https://github.com/skshiftio/mydemo` which has a factory JSON in it, actually this should create a workspace with Spring boot image but if you use url parameter like `https://github.com/skshiftio/mydemo.git` then it will load the ubuntu_jdk8 image as it is not able to read the .factory.json.

plugins/plugin-urlfactory/src/main/java/org/eclipse/che/plugin/urlfactory/URLFetcher.java
garagatyi: Please consider next regexp: `"(?<sanitized>^https?://.*)\\.git$"`
Then fetching sanitized part can be done in such a way: `matcher.group("sanitized")`

garagatyi: FYI if one needs to match dot in regexp it should be escaped with 2 backslashes

kameshsampath: thanks @garagatyi  - thought of it and somehow slipped when i implemented it. thanks for it.. I have updated the code

plugins/plugin-urlfactory/src/main/java/org/eclipse/che/plugin/urlfactory/URLFetcher.java
```diff
@@ -40,6 +42,9 @@
   /** Maximum size of allowed data. (30KB) */
   protected static final long MAXIMUM_READ_BYTES = 30 * 1000;
 
+  /** The Compiled REGEX PATTERN that can be used for http|https git urls */
+  final Pattern GIT_HTTP_URL_PATTERN = Pattern.compile("(?<sanitized>^http[s]?://.*)\\.git$");
```
garagatyi, line 46: I believe that `s`  doesn't have to be in square brackets, so it can be removed to simplify regexp if you will.

garagatyi (approved):

garagatyi:
@skabashnyuk can we merge this PR?

skabashnyuk (approved):

garagatyi:
ci-build

garagatyi:
@kameshsampath congrats on the first contribution to the project! Eager to see more contributions from you!
Sorry if the process was slow - we are a bit overwhelmed with big features of the project.

kameshsampath:
Thanks @garagatyi  it was a pleasure working on it and hope to do more over the time

PR #6690, opened 2017-10-12
use 9.6 tag of postgresql image instead of latest

riuvshin:
we should not use `latest` tag to avoid unexpected updates of critical components


davidfestal (approved):
OK for me if Openshift deployment has been tested again with this tag (cannot do the test myself right now).

riuvshin:
@davidfestal currently `latest` = `9.6` but in future it could be different

davidfestal:
> @davidfestal currently latest = 9.6 but in future it could be different

OK

benoitf (approved):

Issue #6654, opened 2017-10-10
Release CHE 5.19.0


riuvshin:
Release CHE 5.19.0

riuvshin:
depends on: https://github.com/eclipse/che/issues/6687, https://github.com/eclipse/che/issues/6643, https://github.com/eclipse/che/issues/6689, https://github.com/eclipse/che/issues/6693

riuvshin:
Done.

PR #6692, opened 2017-10-12
RELEASE: Set next development version

riuvshin:
### What does this PR do?
set next version in master
https://github.com/eclipse/che/issues/6654


tolusha (approved):

vparfonov (approved):

skabashnyuk (approved):

benoitf:
AFAIK what I've understood in the last community meeting is that we could release 5.20 as well from the 5.x tree, not only 5.19.x versions and that it will be called a "5.x" branch

vparfonov:
As i remember 5.19 will be last feature release.For sure we can release 5.20 and even 5.21 if we will need but this is not our plan

riuvshin:
@benoitf we don't want to have 5.20 because next major version will be 6.
now we release 5.19.0 and temporary we set in master 5.19.1 which means NO 5.20.
after some time we will merge 6 to master. in the same time we will have 5.19.x branch for 5.x.x maintenance

benoitf:
the "we" was not what I've listen during the last meeting.
it was:
- tag 5.19.0 and release it as usual in end of sprint
- rh-che product will switch and try to use this version. We need to wait that rh-che is fully using that 5.x release branch before merging spi into master
- once the full "rh-che/osio che/upstream" is working, a branch is created from master into 5.x and spi is merged into master

so this PR should not be limited to 5.19.1-SNAPSHOT. New development can happen (all the stuff required to have something working on OSIO, it may not be bugfixes) before it could be used on OSIO 
And it's not because there is a 6.0-M1 version that there can't have any 5.x releases.

riuvshin:
@benoitf 
I think you've misunderstanded 

- yes we will release 5.19.0 as usual
- yes rh-che will be able to use that new tag AND we will keep 5.19.x branch for it in case we will need do bugfixes but from branch not from master
why we should block `spi` branch ? if don't expect new features / major changes in master after 5.19 release ?
I could set there version 5.20.0 but we will not release 5.20.0 which will be confusing. after 5.19.0 release all releases for 5.x.x will be from branch.

Issue #6687, opened 2017-10-12
[Regression] Cannot open files in the Editor from the Find panel


SkorikSergey:
**Reproduction Steps:**
1. Start a workspace from any stack with project.
2. Press Ctrl+Shift+F, type text for finding and click the **Search** button
3. Select any occurrence and double click on it.

**Expected behavior:**
File opened in the **Editor**

**Observed behavior:**
Nothing happens.

**Failed test:** FindTextFeatureTest

**Che version:** 5.19.0-SNAPSHOT
**OS and version:** Fedora 25
![teams-11-29](https://user-images.githubusercontent.com/7760565/31483862-cbaf13e8-af36-11e7-9640-882d5afcf2db.gif)

riuvshin:
set as blocker because we can't ship in that state especially after adding pagination feature for search. @vparfonov please assign someone to this asap. Thanks

PR #6696, opened 2017-10-12
Fix initialization change markers render

RomanNikitenko:
### What does this PR do?
Initialization of 'Change markers render' gets project info from app context at the moment.
It works perfect when we have selection in project tree or some opened editor.
But it doesn't work for use case when we open first Editor from the Find panel  - at this moment 
selection in the Find panel and we haven't active editor to provide current project.
So I suggest to get project info from resource which we open for initialization this render.

#6687 

### DIFF
--- a/ide/che-core-orion-editor/src/main/java/org/eclipse/che/ide/editor/orion/client/OrionEditorPresenter.java
+++ b/ide/che-core-orion-editor/src/main/java/org/eclipse/che/ide/editor/orion/client/OrionEditorPresenter.java
@@ -129,6 +129,7 @@ import org.eclipse.che.ide.api.parts.PartPresenter;
 import org.eclipse.che.ide.api.parts.WorkspaceAgent;
 import org.eclipse.che.ide.api.preferences.PreferencesManager;
 import org.eclipse.che.ide.api.resources.File;
+import org.eclipse.che.ide.api.resources.Project;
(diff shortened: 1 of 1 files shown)


RomanNikitenko:
ci-test

ide/che-core-orion-editor/src/main/java/org/eclipse/che/ide/editor/orion/client/OrionEditorPresenter.java
```diff
@@ -1210,7 +1211,15 @@ public void onContextMenu(ContextMenuEvent event) {
...
-                  String vcsName = appContext.getRootProject().getAttribute("vcs.provider.name");
+                  VirtualFile file = input.getFile();
+                  Project project =
+                      file instanceof Resource ? ((Resource) file).getProject() : null;
+                  if (project == null) {
+                    resolve.apply(null);
+                    return;
+                  }
+
+                  String vcsName = project.getAttribute("vcs.provider.name");
```
AndrienkoAleksandr: Can you apply checking on the null? User can delete folder .git but attribute "vcs.provider.name" for project still exist...

vinokurig (approved):

vparfonov (approved):

musienko-maxim (approved):

Issue #6587, opened 2017-10-05
[SPI] Rework injecting of machine token into OpenShift machine


sleshchenko:
Multiuser module has `infrastructure/openshift` submodule where is located only  [MultiuserInstallerConfigProvisioner](https://github.com/eclipse/che/blob/spi/multiuser/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/provision/installer/MultiuserInstallerConfigProvisioner.java) which set `USER_TOKEN` environment variable.
But it may make an impression that `multiuser` has different implementation of OpenShift infrastructure.

It is needed to create `MachineTokenProvider` in `che-core-api-workspace` module and move injection of `USER_TOKEN` env variable into `che/infrastructure/openshift` module. Then single user will have a `dummy` implementation for `MachineTokenProvider` and multi user che will have an implementation that provides machine token from [MachineTokenRegistry#getOrCreateToken](https://github.com/eclipse/che/blob/d2bfa4f7f49dcc232ef84df03de81e87df0b3661/multiuser/machine-auth/che-multiuser-machine-authentication/src/main/java/org/eclipse/che/multiuser/machine/authentication/server/MachineTokenRegistry.java#L65).

PR #6700, opened 2017-10-12
Added MachineTokenProvider interface with different implementations

sleshchenko:
### What does this PR do?
Adds MachineTokenProvider interface. There are different implementations for single and multiuser packagings. With these changes, there is no necessity to override anything for implementations of infrastructures in `che multiuser` assembly.

It also contains minor cleaning up of jpa tests and OpenShift provisioning.

https://github.com/eclipse/che/issues/6587


skabashnyuk (approved):

gazarenkov (approved):

akorneta (approved):

garagatyi (approved):

riuvshin (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/ServersChecker.java
```diff
@@ -56,16 +60,17 @@
...
       RuntimeIdentity runtimeIdentity,
       String machineName,
       Map<String, ? extends Server> servers,
-      ServerCheckerFactory serverCheckerFactory) {
+      @Assisted MachineTokenProvider machineTokenProvider) {
     this.runtimeIdentity = runtimeIdentity;
     this.machineName = machineName;
     this.servers = servers;
-    this.serverCheckerFactory = serverCheckerFactory;
     this.timer = new Timer("ServerReadinessChecker", true);
```
garagatyi: Can you align string constant with the new new name

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/ServersChecker.java
```diff
@@ -56,16 +60,17 @@
...
    * @param servers map of servers in a machine
    */
-  public ServersReadinessChecker(
+  @Inject
+  public ServersChecker(
       RuntimeIdentity runtimeIdentity,
       String machineName,
       Map<String, ? extends Server> servers,
-      ServerCheckerFactory serverCheckerFactory) {
+      @Assisted MachineTokenProvider machineTokenProvider) {
```
garagatyi: Am I wrong that assisted annotation should be on other arguments, but not on this one?

garagatyi (approved):
In overall looks good. I would try to avoid interceptor, but since it is not your code we can discuss it separately.
Please, make sure that it is tested on different infrastructures and deployments.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/ServersChecker.java
sleshchenko: You're correct. I've already fixed it locally.

PR #6705, opened 2017-10-12
Workspace start code cleanup

garagatyi:
### What does this PR do?
Moved async operations from WorkspaceManager to WorkspaceRuntimes
to have an async facility in one place instead of two.
Moved workspace start/stop logging from WorkspaceManager
to WorkspaceRuntimes since WorkspaceManager can not correctly log
them.
Improved logging of workspace start/stop including the addition of new logs.
Fixed logging of exception thrown by RuntimeInfrastructure on runtime
start/stop.
Fix docker image deletion bug on stop of a workspace.

Reopened #6639 since it was closed due to removal of the base branch

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
@@ -10,12 +10,14 @@
  */
 package org.eclipse.che.api.workspace.server;
 
+import static com.google.common.base.MoreObjects.firstNonNull;
 import static java.lang.String.format;
 import static java.util.Objects.requireNonNull;
 import static org.eclipse.che.api.core.model.workspace.WorkspaceStatus.RUNNING;
 import static org.eclipse.che.api.core.model.workspace.WorkspaceStatus.STARTING;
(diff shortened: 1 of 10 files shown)


wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -305,25 +339,75 @@ public void stop(String workspaceId, Map<String, String> options)
...
+      try {
+        state.runtime.stop(options);
+
+        // remove before firing an event to have consistency between state and the event
+        runtimes.remove(workspaceId);
+        LOG.info(
+            "Workspace '{}/{}' with id '{}' stopped by user '{}'",
+            workspace.getNamespace(),
+            workspace.getConfig().getName(),
+            workspace.getId(),
```
sleshchenko: Use `workspaceId` here and below

sleshchenko (approved):

akorneta (approved):

skabashnyuk (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
garagatyi: fixed

PR #6706, opened 2017-10-12
Fix WorkspaceDetailsTest selenium test

SkorikSergey:
### What does this PR do?
We need fix **WorkspaceDetail** selenium test:
- add new agent to test(Yaml language server);
- add timeout for switching agents(sometimes agents don't change status if clicking is fast);
- change locator for checking workspace status.


selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/dashboard/DashboardWorkspace.java
```diff
@@ -116,7 +116,7 @@ public String getName() {
     String RUN_WORKSPACE_BTN = "//button/span[text()='Run']";
     String STOP_WORKSPACE_BTN = "//button/span[contains(text(),'Stop')]";
     String DELETE_WORKSPACE_BTN = "//button/span[text()='Delete']";
-    String STATE_WORKSPACE = "//span[contains(@class, '%s')]";
+    String STATE_WORKSPACE = "//div[contains(@class, 'workspace-status')]/span[text()='%s']";
```
dmytro-ndp: It may be reasonable to use debug-id instead of xpath here.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/dashboard/DashboardWorkspace.java
```diff
@@ -917,6 +917,7 @@ public void checkAgentExists(String agentName) {
   }
 
   public void switchAgentState(String agentName) {
+    loader.waitOnClosed();
```
dmytro-ndp: IMHO it's better not to link method itself with preceding commands, that is to wait until loader from previous operation closes.
To make **switchAgentState()** more universal, it's better to pull `loader.waitOnClosed()` up to the caller method.

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/dashboard/WorkspaceDetailsTest.java
```diff
@@ -138,13 +138,15 @@ public void workingWithAgents() {
         (name, value) -> {
           Assert.assertEquals(dashboardWorkspace.getAgentState(name), value);
           dashboardWorkspace.switchAgentState(name);
+          loader.waitOnClosed();
```
dmytro-ndp: Is it really correct placement for the waiting of loader to be closed, so as we had it at the start of **dashboardWorkspace.switchAgentState(name)** method previously in the class https://github.com/eclipse/che/pull/6706/commits/95e0809b34d8a9009a0c1678e83d4d3db5ff52aa#diff-b918b4454f3846c11126b5e25fdcb6d5 ?

dmytro-ndp (approved):

dmytro-ndp:
@SkorikSergey: I suggest to use **WaitUtils.sleepQuietly()** instead of **loader.waitOnClose()** to provide timeout in the test honestly.

dashboard/src/app/workspaces/workspace-status/workspace-status.html
```diff
@@ -1,4 +1,4 @@
 <div class="workspace-status" flex layout="row" layout-align="start center">
   <workspace-status-indicator che-status="status"></workspace-status-indicator>
-  <span class="workspace-status-text">{{status | lowercase}}</span>
+  <span class="workspace-status-text" id="debug-id-workspace-status">{{status | lowercase}}</span>
```
benoitf: why do we need to call the id with "debug-id" prefix ?

dmytro-ndp: You are right - there ought to be added debug-id attribute - not id one.

SkorikSergey: Yes, you are right. I will remove "debug-id" prefix from id name.

tolusha (approved):

PR #6715, opened 2017-10-13
[CHE-6] - Provide ability to create user's workspaces in one OpenShift project

akorneta:
### What does this PR do?
Provides another strategy for placing user's workspaces: _few workspaces inside one OpenShift project_. This strategy is used only in case when `che.infra.openshift.project` is specified, otherwise for each workspace the os project will be created with name that equal to workspace identifier.
Allows user run more than one workspace on os flavours, and also fixes the case of a stop che when the workspace is placed in the same os project with it.

#6284 

#### Release Notes
Provide ability to create and run user's workspaces in one OpenShift project


garagatyi (dismissed):

sleshchenko (approved):
LGTM

skabashnyuk (approved):

garagatyi (approved):

PR #6717, opened 2017-10-13
rework selenium tests in organization package

Ohrimenko1988:
### What does this PR do?
Fix problem with duplicated organization tests in codenvy/codenvy , and add ability to run organization tests on the codenvy/codenvy

https://github.com/codenvy/codenvy/issues/2483

### DIFF
--- a/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/core/CheSeleniumWebDriverRelatedModule.java
+++ b/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/core/CheSeleniumWebDriverRelatedModule.java
@@ -13,13 +13,9 @@ package org.eclipse.che.selenium.core;
 import com.google.inject.AbstractModule;
 import com.google.inject.Provides;
 import javax.inject.Named;
-import org.eclipse.che.selenium.core.client.CheTestTestOrganizationServiceClientImpl;
-import org.eclipse.che.selenium.core.client.TestOrganizationServiceClient;
 import org.eclipse.che.selenium.core.entrance.CookieEntrance;
 import org.eclipse.che.selenium.core.entrance.Entrance;
 import org.eclipse.che.selenium.core.entrance.LoginPageEntrance;
-import org.eclipse.che.selenium.core.provider.TestApiEndpointUrlProvider;
-import org.eclipse.che.selenium.core.requestfactory.TestCheAdminHttpJsonRequestFactory;
 import org.eclipse.che.selenium.pageobject.site.CheLoginPage;
 import org.eclipse.che.selenium.pageobject.site.LoginPage;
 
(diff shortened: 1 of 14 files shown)


selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/core/CheSeleniumWebDriverRelatedModule.java
```diff
@@ -45,4 +49,12 @@ public Entrance getEntrance(
       return new CookieEntrance(seleniumWebDriver);
     }
   }
+
+  @Provides
```
dmytro-ndp: Could you, please, explain, why did you place this provider here?
This module is dedicated for classes which are injecting **SeleniumWebDriver**, but I don't see such field in the **TestOrganizationServiceClient**, and it shouldn't be there according to the propose of that client.

dmytro-ndp:
@Ohrimenko1988: there are new tests in PR like **UserOrganizationTest**, but I don't see any changes in tests suites. How are the new tests going to be started?

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/core/CheSeleniumWebDriverRelatedModule.java
Ohrimenko1988: Because in che we return CheTestTestOrganizationServiceClientImpl and in the codenvy we return other implementation of TestOrganizationServiceClient interface

Ohrimenko1988 (commented):
@dmytro-ndp tests can be started by using package running (  --test=organization.*  )

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/core/CheSeleniumWebDriverRelatedModule.java
dmytro-ndp: I see. But the propose of **CheSeleniumWebDriverRelatedModule** is the classes which are injecting webdriver itself.

dmytro-ndp:
Do you have corresponding changes in the Codenvy repo as well?

Ohrimenko1988:
@dmytro-ndp  yes I have prepared changes in the codenvy/codenvy according to changes in this PR

dmytro-ndp:
Which branch in Codenvy repo contains the changes?

Ohrimenko1988:
@dmytro-ndp  https://github.com/codenvy/codenvy/tree/fix-organization-selenium-tests

dmytro-ndp:
Thank you, @Ohrimenko1988.
[OrganizationSuite.xml](https://github.com/codenvy/codenvy/blob/fix-organization-selenium-tests/selenium/codenvy-selenium-test/src/test/resources/suites/OrganizationSuite.xml) looks an obsolete. Lets remove it.

Ohrimenko1988:
@dmytro-ndp yes, you are absolutely right, thx

dmytro-ndp (approved):
Looks good.

tolusha (approved):

Issue #5326, opened 2017-06-09
Java 9 Support 


gorkem:
Study and determine the tasks for Che to fully support Java 9

Issues reported by @benoitf at date of 16 october 2017:

compiling che with Java 9 blocker : 
- [x] Gwt and its maven plugin are not working with Java9
 need to update to HEAD-MASTER (2.8.2) and new gwt plugin
 https://github.com/eclipse/che/issues/6602

compiling issues with Java9 : plug-Ins errors with workarounds :

- [x] error-prone is not handling Java9 for now (ok with newer version
   + https://github.com/google/error-prone/issues/448

- [x] lot of maven plug-Ins need to be updated to their latest version to handle Java9 correctly
    + [x] https://github.com/eclipse/che/pull/5490 Avoid deprecated configuration for maven assembly plugin

- [x] tools.jar is referenced in debugger server (but tools.jar is gone with Java9)

- [x] plexus-component-metadata needs to be updated as well https://github.com/eclipse/che/blob/513aa19ebd121093746209fcc4b31697b1424331/plugins/plugin-maven/maven-server/maven-server-impl/pom.xml#L214-L217

- [x] eclipselink failures on Java9. Updating to EclipseLink 2.7.0 solves some issues
- [x] Mockito failures on Java9. We should update to a recent version of Mockito (2.10 release)
    - [x] mockito-all artifact is no longer existing
    - [x] mockitong incompatible with Mockito 2.x (usage of removed annotations)

- [x] gwt mockito is not working with Java9
I've provided https://github.com/google/gwtmockito/pull/73 but required to release a new version

- [x] Lucene failures on Java9. Updating to Lucene 7.0.0 solves some issues (current=5.2.1)

- [x] Usage of internal classes of JMX stuff
    - https://github.com/eclipse/che/blob/5b373e6cf5bfeb16f278c4d750fe64361445eeb4/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/SystemInfo.java#L76-L79

- [x] org.reflections and usage of Classpath (assuming that for example current classloader is a URLClassLoader is not working on Java9 as boot classloader is not anymore a URLClassLoader)

### Runtime

- [ ]: wrong path of logging.properties in Tomcat
          http://tomcat.10.x6.nabble.com/svn-commit-r1809434-in-tomcat-trunk-java-org-apache-juli-ClassLoaderLogManager-java-webapps-docs-chal-td5067587.html

(fixed in new Tomcat release)

[...]

PR #6719, opened 2017-10-13
Update Mockito to v2.10 (che6 branch)

benoitf:
note : before merging, the versions of mockito needs to be removed from root pom.xml and moved to che-dependencies repository.

### What does this PR do?
Update Mockito to v2.10

Some changes are required like
 - `anySomething()` --> `nullable(Something.class)` for many tests as in fact we provide null values while any() expect a non-null value
 - add class for `VerificationMode` as there are more than one method now (can't use lambdas)
 - remove cast on `ArgumentMatcher` (now we can use lambdas as there is generics)
 - remove unecessary stubbing (mockito is now reporting un-needed stubs)
 - replace dependency `mockito-all` by` mockito-core` (it is no longer provided in 2.x)

#5326 

#### Release Notes
N/A

#### Docs PR
N/A

Change-Id: Ifcddd24927a69cbbae1bd1dd97e35e900c94a5b3

### DIFF
--- a/assembly/assembly-wsmaster-war/pom.xml
+++ b/assembly/assembly-wsmaster-war/pom.xml
@@ -304,7 +304,7 @@
         </dependency>
         <dependency>
             <groupId>org.mockito</groupId>
-            <artifactId>mockito-all</artifactId>
+            <artifactId>mockito-core</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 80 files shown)


garagatyi:
Does this PR include changes of tests from the master branch or not?

benoitf:
@garagatyi I'm not sure to fully understand but 
it contains all fixes of PR aginst master https://github.com/eclipse/che/pull/6394 + the new fixes for new classes and also some classes have been removed so there are some fixes not applied on that branch

garagatyi (approved):

skabashnyuk (approved):

riuvshin (approved):

skabashnyuk (approved):

Issue #6557, opened 2017-10-03
[SPI] Runtime#getToken must return machine token


sleshchenko:
Runtime#getToken must return machine token but actually Keycloak token is [there](
https://github.com/eclipse/che/blob/spi/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/DtoConverter.java#L76). It should be fixed to make client able to use it for connection to workspace agents.

PR #6732, opened 2017-10-13
Made RuntimeDto return machine token

sleshchenko:
### What does this PR do?
Made `RuntimeDto` return machine token.

Also, this PR contains uncommenting and fixing tests for WorkspaceService and WorkspaceValidator.

https://github.com/eclipse/che/issues/6557

### DIFF
--- a/wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/workspace/server/WorkspaceServiceTest.java
+++ b/wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/workspace/server/WorkspaceServiceTest.java
@@ -10,937 +10,939 @@
  */
 package org.eclipse.che.api.workspace.server;
 
+import static com.jayway.restassured.RestAssured.given;
+import static java.util.Arrays.asList;
+import static java.util.Collections.emptyList;
+import static java.util.Collections.emptyMap;
+import static java.util.Collections.singleton;
+import static java.util.Collections.singletonList;
+import static java.util.Collections.singletonMap;
+import static java.util.stream.Collectors.toList;
+import static org.eclipse.che.api.core.model.workspace.WorkspaceStatus.STARTING;
+import static org.eclipse.che.dto.server.DtoFactory.newDto;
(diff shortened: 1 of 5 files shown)


garagatyi (dismissed):
Thank you for the tests!

wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/workspace/server/WorkspaceServiceTest.java
```diff
@@ -10,937 +10,941 @@
...
+                    "custom", "custom:value")));
+  }
+
+  @Test
+  public void shouldUseUsernameAsNamespaceWhenCreatingWorkspaceWithoutSpecifiedNamespace()
+      throws Exception {
+    final WorkspaceConfigDto configDto = createConfigDto();
+    final WorkspaceImpl workspace = createWorkspace(configDto);
+    when(wsManager.createWorkspace(
+            any(WorkspaceConfig.class), anyString(), anyMapOf(String.class, String.class)))
```
benoitf: with Java8, anyMapOf is not required (and it is deprecated in new releases) any() is enough for type inference

sleshchenko: thanks. I'll fix

benoitf: https://stackoverflow.com/questions/21783219/mockito-anymapof-nested-generics
Matchers.<Map<String, String >>any()

benoitf: BTW it's because I'm thinking of https://github.com/eclipse/che/pull/6719

skabashnyuk (dismissed):

garagatyi (approved):

akorneta (approved):

skabashnyuk (approved):

Issue #6749, opened 2017-10-16
Can't create workspace in Che 5.19.0 behind proxy


bs-matil:
### Description
Can't get a workspace running with http proxy enabled. To me it looks a lot like the no_proxy setting has been ignored within the che server code.

#### Environment:
Start eclipse che behind a http proxy (Port 1234. IP 1.3.4.5) 

**Proxy in che container**

```

JAVA_HTTP_PROXY_PORT=-Dhttp.proxyPort=1234
JAVA_HTTP_PROXY_HOST=-Dhttp.proxyHost=1.3.4.5
JAVA_HTTPS_PROXY_HOST=-Dhttps.proxyHost=1.3.4.5
JAVA_HTTPS_PROXY_PORT=-Dhttps.proxyPort=1234
JAVA_HTTP_PROXY_SET=-Dhttp.proxySet=true
CHE_WORKSPACE_JAVA_OPTIONS=-Xms256m -Xmx2048m -Djava.security.egd=file:/dev/./urandom -Dhttp.proxySet=true -Dhttp.proxyHost=1.3.4.5  -Dhttp.proxyPort=3128 -Dhttps.proxyHost=1.3.4.5 -Dhttps.proxyPort=1234 -Dhttp.nonProxyHosts='che-host||1.3.4.5|localhost|


```

**Proxy in che container**

```

http_proxy=http://1.3.4.5:1234/
https_proxy=http://1.3.4.5:1234/
no_proxy=che-host,,1.3.4.5,localhost

```

Steps
1. Create a Workspace
2. tail the log of container "che"

Result:
che workspace does not boot up. 

### Reproduction Steps
create a workspace behind a http proxy

**OS and version:**   
5.19.0
ubuntu 
docker 17.03.ce 

**Diagnostics:** 
```
2017-10-16 12:53:13,870[aceSharedPool-0]  [INFO ] [o.e.c.a.w.s.WorkspaceRuntimes 676]   - Launching 'org.eclipse.che.exec' agent at workspace workspace20zt3gxmcn724j5z
2017-10-16 12:53:16,170[aceSharedPool-0]  [INFO ] [o.e.c.a.w.s.WorkspaceRuntimes 676]   - Launching 'org.eclipse.che.terminal' agent at workspace workspace20zt3gxmcn724j5z
[...]

PR #6775, opened 2017-10-17
No proxy handling

ghost:
Signed-off-by: Eugene Ivantsov <eivantsov@codenvy.com>

### What does this PR do?

Fixes the bug with setting an empty no_proxy value. Also, a warning is added if proxy is set in docker daemon but no_proxy is not.

https://github.com/eclipse/che/issues/6749


dockerfiles/base/scripts/base/startup_02_pre_docker.sh
```diff
@@ -356,6 +356,10 @@ init_check_docker_networking() {
      [[ ! ${HTTPS_PROXY} = "" ]] ||
      [[ ! ${NO_PROXY} = "" ]]; then
      info "Proxy: HTTP_PROXY=${HTTP_PROXY}, HTTPS_PROXY=${HTTPS_PROXY}, NO_PROXY=${NO_PROXY}"
+     if [[ ${NO_PROXY} = "" ]]; then
+     warning "Potential networking issue discovered!"
+     warning "We have identified that http and https proxies set are but no_proxy is not. This may cause fatal errors. Set no_proxy for your Docker daemon!"
```
benoitf: is there a misswording between `set are` and `are set`

benoitf (approved):

riuvshin (approved):

PR #6779, opened 2017-10-17
Move Docker client to Docker infrastructure packaging

### LINKED ISSUE #6404: SPI: Move docker infrastructure related code from docker-plugin to docker infrastructure
We have code that is placed in module plugin-docker but it is not a plugin at all comparing to what we have in other plugins. It is rather infrastructure related code. So we should move this code into related infrastructure modules.

**Reproduction Steps:**

**OS and version:**    

**Diagnostics:**

garagatyi:
### What does this PR do?
Move Docker client to Docker infrastructure packaging.
Also,  remove OpenShiftConnector as it is replaced with OpenShift
infrastructure.

Fixes #6404


riuvshin (approved):

infrastructures/docker/pom.xml
```diff
@@ -15,185 +15,31 @@
...
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-surefire-plugin</artifactId>
-                <configuration>
-                    <excludes>
-                        <exclude>**/integration/**</exclude>
-                    </excludes>
-                </configuration>
-            </plugin>
-        </plugins>
+        <pluginManagement>
```
benoitf: do we need this pluginManagement section here ?

skabashnyuk (approved):

infrastructures/docker/pom.xml
garagatyi: removed

infrastructures/pom.xml
```diff
@@ -19,7 +19,9 @@
         <version>5.20.0-SNAPSHOT</version>
         <relativePath>../pom.xml</relativePath>
     </parent>
+    <groupId>org.eclipse.che.infrastructure</groupId>
     <artifactId>che-infrastructures-parent</artifactId>
+    <version>5.20.0-SNAPSHOT</version>
```
sleshchenko: I think you can remove version here

infrastructures/docker/infrastructure/pom.xml
```diff
@@ -0,0 +1,203 @@
...
+        </dependency>
+    </dependencies>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <excludes>
+                        <exclude>**/integration/**</exclude>
```
sleshchenko: I don't see `integration` package. Please check that it is needed here

sleshchenko (approved):

infrastructures/pom.xml
garagatyi: ok

infrastructures/docker/infrastructure/pom.xml
garagatyi: Nice catch!

benoitf (approved):
good bye OpenShift Connector 😭

garagatyi:
@benoitf You can still contribute to the new OpenShift infra instead and I hope it will be easier and more straightforward than what we have in master

garagatyi:
@l0rd can you take a look on this PR?

l0rd (approved):
👍

Issue #6770, opened 2017-10-17
find a way how to add new fully activated user in keycloak


riuvshin:
Today we have an issue with running tests against che-multiuser because there is no ready user. Even Admin user should be activated which is manual step and blocks automated run on CI.

PR #6780, opened 2017-10-17
Allow to disable require update password for default admin user

riuvshin:
backport changes from master. Original PR: https://github.com/eclipse/che/pull/6777

needed for: https://github.com/eclipse/che/issues/6770

### DIFF
--- a/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/FindText.java
+++ b/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/FindText.java
@@ -103,6 +103,7 @@ public class FindText {
 
   /** wait the 'Find Text' main form is closed */
   public void waitFindTextMainFormIsClosed() {
+    loader.waitOnClosed();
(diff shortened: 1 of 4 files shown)


selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/FindText.java
```diff
@@ -103,6 +103,7 @@ public void launchFindFormByKeyboard() {
 
   /** wait the 'Find Text' main form is closed */
   public void waitFindTextMainFormIsClosed() {
+    loader.waitOnClosed();
```
benoitf, line 106: do we need that as well ?

benoitf (approved):

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/FindText.java
riuvshin: yes, this is from PR https://github.com/eclipse/che/pull/6762
@SkorikSergey didn't ported that change to che6 and it was already in my branch so I decided to keep it.

skabashnyuk (approved):

PR #6789, opened 2017-10-18
Fix problem with open factory page in the selenium tests

Ohrimenko1988:
### What does this PR do?
Resolve problem with redirect to dashboard when try to open factory page by URL

Issue https://github.com/codenvy/codenvy/issues/2477

### DIFF
--- a/selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/factory/TestFactory.java
+++ b/selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/factory/TestFactory.java
@@ -14,10 +14,12 @@ import org.eclipse.che.api.factory.shared.dto.FactoryDto;
 import org.eclipse.che.selenium.core.SeleniumWebDriver;
 import org.eclipse.che.selenium.core.client.TestFactoryServiceClient;
 import org.eclipse.che.selenium.core.client.TestWorkspaceServiceClient;
+import org.eclipse.che.selenium.core.constant.TestTimeoutsConstants;
 import org.eclipse.che.selenium.core.entrance.Entrance;
 import org.eclipse.che.selenium.core.provider.TestDashboardUrlProvider;
 import org.eclipse.che.selenium.core.user.TestUser;
 import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.support.ui.WebDriverWait;
 
 /** @author Anatolii Bazko */
 public class TestFactory {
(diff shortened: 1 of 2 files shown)


selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/factory/TestFactory.java
```diff
@@ -48,10 +50,10 @@ public TestFactory(
 
   /** Adds authentication token into the browser and opens factory url. */
   public void authenticateAndOpen(SeleniumWebDriver driver) throws Exception {
-    driver.get(dashboardUrl.get().toString());
-
+    String dashboardUrlPath = dashboardUrl.get().toString();
+    driver.get(dashboardUrlPath);
     entrance.login(owner);
-
+    waitUntilTextPresentInUrl(driver, dashboardUrlPath + "#/");
```
dmytro-ndp: "#/" looks like a kind of rudiment

Ohrimenko1988: This is a part which added when dashboard start loading, and exactly this we need check. After that webdriver can open another url without mistakes.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/factory/TestFactory.java
```diff
@@ -48,10 +50,10 @@ public TestFactory(
 
   /** Adds authentication token into the browser and opens factory url. */
   public void authenticateAndOpen(SeleniumWebDriver driver) throws Exception {
-    driver.get(dashboardUrl.get().toString());
-
+    String dashboardUrlPath = dashboardUrl.get().toString();
```
dmytro-ndp: Variable **dashboardUrlPath** looks useless here.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/factory/TestFactory.java
dmytro-ndp: It's a kind of workaround. Can we wait on presence of some element on page?

dmytro-ndp: We already have login page. IMHO it's more reasonable to check if page has already been loaded inside the **entrance.login()** method.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/factory/TestFactory.java
```diff
@@ -75,4 +77,9 @@ private void deleteFactory() throws Exception {
   private boolean isNamedFactory() {
     return factoryDto.getName() != null;
   }
+
+  private void waitUntilTextPresentInUrl(SeleniumWebDriver driver, String textInUrl) {
+    new WebDriverWait(driver, TestTimeoutsConstants.LOAD_PAGE_TIMEOUT_SEC)
+        .until((WebDriver webdriver) -> webdriver.getCurrentUrl().contains(textInUrl));
```
dmytro-ndp: **startWith()** is more reliable here compare with contains()

tolusha (approved):

dmytro-ndp (approved):
Looks good as for me. 
Thank you for quick fixup.

PR #6790, opened 2017-10-18
fixed minor "webscoket" typos

Patricol:

PR #6793, opened 2017-10-18
fixed minor "webscoket" typos

Patricol:
continued from #6790

### DIFF
--- a/agents/go-agents/docs/rest_api.md
+++ b/agents/go-agents/docs/rest_api.md
@@ -191,7 +191,7 @@ The result of the request _GET /process?all=true_
 _POST /process/{pid}/events/{channel}_
 
 - `pid` - the id of the process to subscribe to
-- `channel` - the id of the webscoket channel which is subscriber
+- `channel` - the id of the websocket channel which is subscriber
 - `types`(optional) - the types of the events separated by comma e.g. `?types=stderr,stdout`
 -  `after`(optional) - process logs which appeared after given time will
 be republished to the channel. This method may be useful in the reconnect process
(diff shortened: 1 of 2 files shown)


Patricol:
I can't change both files in one commit from just the website, and two pull requests would be insane to do.
I'll have to figure this out tomorrow.

garagatyi (approved):
Looks good. Thank you for your contribution!

benoitf:
@Patricol did you find time to get ECA added as Signed-Off-By ?

Patricol:
@benoitf 

It looks to me like both commits are signed off identically, and that only one failed the ip-validation check.
What exactly is the issue?

The incorrect(?) data in the ip-validation check's [details](https://dev.eclipse.org/eclipse-webhook/services/status_details.php?id=59e729a7cc7b5) (the users listed are noreply@github.com and web-flow) made me think the problem was no longer coming from my side, and I mistakenly interpreted my email notification of garagatyi's comment as indication that the PR was successful.

I'd assume that the error was caused by me using the github.com website to make the commits; (as it would make some sense for the committer to be "web-flow <noreply@github.com>", mismatching the "signed off by" line and causing the ip-validation check to fail,) but in that case I'd expect the ip-validation check to fail on both commits.

Thank you for your patience, by the way. 
As you may be able to tell from #6790's initial title; I prematurely submitted that PR by pressing enter while typing in its information; so I missed all of the notes about contribution guidelines.

Patricol:
Nevermind; just noticed that only one of the two initial commits in this PR was tested; both would have failed.

So the issue was that you cannot have a compliant signed-off-by when committing from github.com's web-based editor.

Thank you for your time!

benoitf:
@Patricol yes the github web editor is adding noreply@github.com and make ECA fails.

Thanks for your patience and the fix.

Issue #6702, opened 2017-10-12
[CHE 6] Add environment variables configuration to Machine model


garagatyi:
### Description
We need to add an ability to set environment variables configuration in [machine config](https://github.com/eclipse/che/blob/che6/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/config/MachineConfig.java#L21).
I suggest next structure of the MachineConfig object:
```
{
    "installers" : [ ],
    "attributes" : { },
    "servers" : { },
    "env" : {
         "my_env1" : "value1",
         "my_env2" : "value2"
    }
}
```

gazarenkov:
OK with this model extension, can we consider "env" name instead of "envVariables" though?

garagatyi:
ok, I changed the issue description.

PR #6802, opened 2017-10-18
Added environment variables configuration to Machine model

sleshchenko:
### What does this PR do?
Adds environment variables configuration to MachineConfig model.
Adds injecting of configured environment variables into Docker and OpenShift machines.

https://github.com/eclipse/che/issues/6702

#### Release Notes
Added environment variables configuration to Machine model

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/InternalMachineConfig.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/InternalMachineConfig.java
@@ -35,14 +35,27 @@ public class InternalMachineConfig {
   private final List<InstallerImpl> installers;
   // set of servers including ones configured by installers
   private final Map<String, ServerConfig> servers;
+  private final Map<String, String> env;
   private final Map<String, String> attributes;
 
   InternalMachineConfig(MachineConfig originalConfig, InstallerRegistry installerRegistry)
       throws InfrastructureException {
-    this.installers = new ArrayList<>();
-    this.servers = new HashMap<>(originalConfig.getServers());
-    this.attributes = new HashMap<>(originalConfig.getAttributes());
+    this.servers = new HashMap<>();
+    if (originalConfig.getServers() != null) {
+      this.servers.putAll(originalConfig.getServers());
+    }
 
+    this.env = new HashMap<>();
+    if (originalConfig.getEnv() != null) {
+      this.env.putAll(originalConfig.getEnv());
+    }
+
+    this.attributes = new HashMap<>();
+    if (originalConfig.getAttributes() != null) {
+      this.attributes.putAll(originalConfig.getAttributes());
+    }
+
+    this.installers = new ArrayList<>();
     initInstallers(originalConfig.getInstallers(), installerRegistry);
   }
 
@@ -58,6 +71,11 @@ public class InternalMachineConfig {
   /** Returns unmodifiable list of installers configs of the machine. */
   public List<InstallerImpl> getInstallers() {
     return Collections.unmodifiableList(installers);
+  }
+
+  /** Returns unmodifiable map of machine environment variables. */
+  public Map<String, String> getEnv() {
+    return Collections.unmodifiableMap(env);
   }
 
   /** Returns unmodifiable map of machine attributes. */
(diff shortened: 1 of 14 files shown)


wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/InternalMachineConfig.java
```diff
@@ -35,12 +35,14 @@
...
   // set of servers including ones configured by installers
   private final Map<String, ServerConfig> servers;
+  private final Map<String, String> env;
   private final Map<String, String> attributes;
 
   InternalMachineConfig(MachineConfig originalConfig, InstallerRegistry installerRegistry)
       throws InfrastructureException {
     this.installers = new ArrayList<>();
     this.servers = new HashMap<>(originalConfig.getServers());
+    this.env = new HashMap<>(originalConfig.getEnv());
```
garagatyi: What if getEnv returns null?

wsmaster/che-core-sql-schema/src/main/resources/che-schema/5.20.0/1__add_machine_env.sql
```diff
@@ -0,0 +1,20 @@
...
+-- are made available under the terms of the Eclipse Public License v1.0
+-- which accompanies this distribution, and is available at
+-- http://www.eclipse.org/legal/epl-v10.html
+--
+-- Contributors:
+--   Red Hat, Inc. - initial API and implementation
+--
+
+-- Environment machine env ----------------------------------------------
+CREATE TABLE externalmachine_env (
```
garagatyi: Can you elaborate on why external?

garagatyi: I think that script 1 in 5.20.0 already exists

benoitf (approved):

wsmaster/che-core-sql-schema/src/main/resources/che-schema/5.20.0/1__add_machine_env.sql
sleshchenko: Actually, I don't know. I do in this way to be the same with existing tables https://github.com/eclipse/che/blob/36777244e994f1c822a86f27656fd37a82c52ef7/wsmaster/che-core-sql-schema/src/main/resources/che-schema/5.0.0-M8/1__init.sql#L271-L300

sleshchenko: It didn't when I created this file. But now it exists. Nice catch. I'll rename it.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/InternalMachineConfig.java
sleshchenko: The same as in case when `originalConfig.getServers` or `originalConfig.getAttributes` return `null`.
Now `originalConfig` doesn't return null values because it is `MachineConfigImpl`.
We can use InternalMachineConfig as constructor argument here or rewrite constructor in the same way as [InternalMachineConfig](https://github.com/eclipse/che/blob/85314523ed5c6d88c6dce68d645bfc04701a92ed/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/model/impl/MachineConfigImpl.java#L94-L99) has. Which way do you prefer?

wsmaster/che-core-sql-schema/src/main/resources/che-schema/5.20.0/1__add_machine_env.sql
garagatyi: Can we avoid that historical thing

akorneta (approved):

skabashnyuk (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/InternalMachineConfig.java
sleshchenko: I've reworked in a second way. Ping me if you think that a first way is better

garagatyi (approved):

Issue #6238, opened 2017-09-12
Don't create redundant default workspace at start of test


dmytro-ndp:
Selenium tests creates test workspace through REST API by default by invoking [TestWorkspaceProviderImpl.initializePool()](https://github.com/eclipse/che/blob/master/selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/workspace/TestWorkspaceProviderImpl.java#L141) method at start.

But there are situations exist when default test workspace in the pool is redundant:
1. In tests which use dedicated workspace with non-default parameters, for example org.eclipse.che.selenium.languageserver.**CheckMainFeatureForLanguageTest**;
1. In tests which create test workspace in test methods by using factory, for example org.eclipse.che.selenium.factory.**DirectUrlFactoryWithSpecificBranch**
1. In tests which create test workspace in test methods through Dashboard UI, for example org.eclipse.che.selenium.dashboard.stacks.**CheckLaravelStackTest**

In some cases we have too little resources to run even two workspaces of 3 GB size, for example, on codenvy.io or in OCP (in issue https://github.com/eclipse/che/issues/6475). As a result, tests fail because it's impossible to run just single workspace.

Fix selenium tests to recognize such situation in some way and don't create default workspace at start of those tests at all.

riuvshin:
set as blocker because this blocks running tests on OCP where we can run only 1 ws

PR #6853, opened 2017-10-19
Add ability to setup test workspace pool size in selenium tests

dmytro-ndp:
### What does this PR do?
Fix selenium tests not to create test workspace pool by default, which allows to prevent unnecessary workspaces creation. To do so, it provides and implements new parameter of selenium tests:
```
 --workspace-pool-size=[<SIZE>|auto] Size of test workspace pool.
                                     Default value is 0, that means that test workspaces are created on demand.
```

In case of _--workspace-pool-size=**auto**_, real pool size will be equals `(threads - 1) / 2 + 1`.

#6238
https://github.com/eclipse/che/pull/6852


tolusha (approved):

selenium/che-selenium-core/bin/webdriver.sh
```diff
@@ -142,6 +143,7 @@ checkParameters() {
         elif [[ "$var" =~ -P.* ]]; then :
         elif [[ "$var" == "--help" ]]; then :
         elif [[ "$var" == "--compare-with-ci" ]]; then :
+        elif [[ "$var" =~ --workspace-pool-size=auto|[0-9]+$ ]]; then :
```
garagatyi: Have you tried both following variants?
`--workspace-pool-size=auto`
`--workspace-pool-size=2`
I'm not sure that regexp is correct. Can you check whether it should be something like:
`^--workspace-pool-size=((auto)|([0-9]+))$`

selenium/che-selenium-core/bin/webdriver.sh
```diff
@@ -379,20 +384,25 @@ Modes (defines environment to run tests):
...
-        --web-driver-port=<PORT>        To run WebDriver on the specific port, by default: "${WEBDRIVER_PORT}"
-        --threads=<THREADS>             Number of tests that will be run simultaneously. It also means the very same number of
+    Options that go with 'local' mode:
+    --web-driver-version=<VERSION>      To use the specific version of the WebDriver, be default the latest will be used: "${WEBDRIVER_VERSION}"
+    --web-driver-port=<PORT>            To run WebDriver on the specific port, by default: "${WEBDRIVER_PORT}"
+    --threads=<THREADS>                 Number of tests that will be run simultaneously. It also means the very same number of
                                         Web browsers will be opened on the developer machine.
                                         Default value is in range [2,5] and depends on available RAM.
+    --workspace-pool-size=[<SIZE>|auto] Size of test workspace pool.
+                                        Default value is 0, that means that test workspaces are created on demand.
```
garagatyi, line 394: Does it mean that you are changing current default behavior?

selenium/che-selenium-core/bin/webdriver.sh
```diff
@@ -632,7 +643,8 @@ printProposals() {
                           sed -e "s/--suite=[^ ]*//g " | \
                           sed -e "s/--test*=[^ ]*//g " | \
                           sed -e "s/--compare-with-ci\W*[0-9]*//g" | \
-                          sed -e "s/--threads=[0-9]*//g")
+                          sed -e "s/--threads=[0-9]*//g" | \
+                          sed -e "s/--workspace-pool-size=auto|[0-9]*//g")
```
garagatyi, line 647: Please check regexp, see my comment above

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/workspace/TestWorkspaceProviderImpl.java
```diff
@@ -41,50 +39,48 @@
...
   private final TestWorkspaceServiceClient testWorkspaceServiceClient;
   private final TestWorkspaceServiceClientFactory testWorkspaceServiceClientFactory;
   private final WorkspaceDtoDeserializer workspaceDtoDeserializer;
 
+  private ArrayBlockingQueue<TestWorkspace> testWorkspaceQueue;
+  private ScheduledExecutorService executor;
+
   @Inject
   public TestWorkspaceProviderImpl(
+      @Named("sys.workspace_pool_size") String poolSize,
```
garagatyi: Please, do not use injection with `sys.` anymore. Use injection similar to what we do in Che instead - `che.workspace_pool_size` which can be injected using system property, property in a file,  environment variable.

garagatyi (changes requested):
Nice idea! PR looks good overall. Can you address my comments?

@dmytro-ndp @tolusha What if developer executes a single test with this script? can we implement logic in this PR that automatically sets pool size to 1 when there is 1 test to execute?

selenium/che-selenium-core/bin/webdriver.sh
dmytro-ndp: I have tried, both :-)

selenium/che-selenium-core/bin/webdriver.sh
dmytro-ndp: Yes, exactly.

selenium/che-selenium-core/bin/webdriver.sh
dmytro-ndp: I had already checked, before making PR.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/workspace/TestWorkspaceProviderImpl.java
dmytro-ndp: This variable is kind of "**sys.threads**", so I have to do the same with threads as well, haven't it?

dmytro-ndp: Also in this case it doesn't make sense to have a parameter **--workspace-pool-size** if user be able to define the value through the system property or environment variable instead.

selenium/che-selenium-core/bin/webdriver.sh
garagatyi: The reason is that `--workspace-pool-size=aut1` also matches this regexp.

selenium/che-selenium-core/bin/webdriver.sh
garagatyi: I see, thank you.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/workspace/TestWorkspaceProviderImpl.java
garagatyi: It makes sense in case we will want to use selenium tests in some other way then through this script. And since it won't change much I don't see the reason to use approach that differs from Che code approach

dmytro-ndp:
@garagatyi: please, review fresh fixup.

selenium/che-selenium-core/bin/webdriver.sh
garagatyi: Do you think it is worth to fix?

selenium/che-selenium-core/bin/webdriver.sh
```diff
@@ -680,14 +692,15 @@ runTests() {
...
                 -Dche.port=${PRODUCT_PORT} \
-                -Dprotocol=${PRODUCT_PROTOCOL} \
+                -Dche.protocol=${PRODUCT_PROTOCOL} \
                 -Ddocker.interface.ip=$(detectDockerInterfaceIp) \
                 -Ddriver.port=${WEBDRIVER_PORT} \
                 -Ddriver.version=${WEBDRIVER_VERSION} \
                 -Dbrowser=${BROWSER} \
-                -Dthreads=${THREADS} \
+                -Dche.threads=${THREADS} \
+                -Dche.workspace.pool.size=${WORKSPACE_POOL_SIZE} \
```
garagatyi: The previous name I liked more )
We have a convention to use dots to show components separation and underscores to separate words in one component. 
So, consider: `che.workspace_pool_size`

dmytro-ndp: But [SeleniumTestConfiguration](https://github.com/eclipse/che/blob/5.19.0/selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/configuration/SeleniumTestConfiguration.java#L120) replaces "_" on "." in module, so "che.workspace.pool.size" is just an exact name which we should inject into classes.

selenium/che-selenium-core/bin/webdriver.sh
dmytro-ndp: :-) good catch

dmytro-ndp:
@garagatyi:
> @dmytro-ndp @tolusha What if developer executes a single test with this script? can we implement logic in this PR that automatically sets pool size to 1 when there is 1 test to execute?

workspace pool size = 1 means that there should be 2 workspaces run simultaneously, and this PR discard such behavior by default.
The main point is NOT TO CREATE additional workspaces at all by default, and it requires pool size = 0 by default.

garagatyi (approved):

PR #6854, opened 2017-10-20
Edit image list depending on single port policy and single vs multi user assembly

ghost:
Now that different Che assemblies and modes require a different set of docker images, the CLI pulls all of them. As a result, a user trying to start a single user Che pulls 4 extra images. The proposed check modifies the list depending on:

1. Single port policy
2. Single vs Multi user mode

### DIFF
--- a/dockerfiles/base/scripts/base/startup_03_pre_networking.sh
+++ b/dockerfiles/base/scripts/base/startup_03_pre_networking.sh
@@ -93,9 +93,9 @@ get_image_manifest() {
     IMAGE_LIST=$(cat /version/$1/images)
   fi
   if [ -z "${CHE_MULTIUSER:-}" ]; then
-     IMAGE_LIST=$(echo "${IMAGE_LIST}" | sed '/IMAGE_KEYCLOAK/d; /IMAGE_POSTGRES/d; /IMAGE_CHE_MULTI_*/d')
+     IMAGE_LIST=$(echo "${IMAGE_LIST}" | sed '/IMAGE_KEYCLOACK/d; /IMAGE_POSTGRES/d; /IMAGE_CHE_MULTI_*/d')
   else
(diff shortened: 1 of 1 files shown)


benoitf (approved):
so nice

dockerfiles/base/scripts/base/startup_03_pre_networking.sh
```diff
@@ -87,7 +87,15 @@ get_image_manifest() {
 
   # Load images from file
   BOOTSTRAP_IMAGE_LIST=$(cat ${SCRIPTS_BASE_CONTAINER_SOURCE_DIR}/images/images-bootstrap)
-  IMAGE_LIST=$(cat /version/$1/images)
+  if [ -z "${CHE_SINGLE_PORT:-}" ]; then
+    IMAGE_LIST=$(cat /version/$1/images | sed '/IMAGE_TRAEFIK/d')
+        if [ -z "${CHE_MULTIUSER:-}" ]; then
```
benoitf: is that this IF should be inside of CHE_SINGLE_PORT or it's not related ?

ghost: I first check if it's single port and define image list accordingly. Do you mean it should not be if in if? but two if clauses?

benoitf: yes I was wondering why the multiuser is part of the singleport condition
if !single_port
   remove traefik image

if !multiuser
  remove multiuser images

for me they are separated

ghost:
Two if clauses now

benoitf (approved):

riuvshin (approved):

garagatyi (approved):

skabashnyuk:
PR for Che6?

ghost:
https://github.com/eclipse/che/pull/6871

Issue #6855, opened 2017-10-20
[Regression]GDB-Debugging not working in 5.20.0


hkolvenbach:
### Description
 
GDB debugging does not work in Eclipse Che 5.20-SNAPSHOT even though privileged mode is enabled.
- GDB debugging without specifying the port for gdbserver does not work at all
- when correctly specifying the port, debugging crashes instantly
- using GDB from command line to connect to gdbserver works

### Reproduction Steps

- Following the guide in https://eclipse.org/che/docs/ide/debug/index.html#gdb with `Che_DOCKER_PRIVILEGED=true`
- Use C++ stack and cpp console application sample
- add `int test1 = 200;` to the main function to have a variable and set a breakpoint on it
- compile with the run command (or use `gcc -g Hello.cc` directly)
- Start the gdbserver with `gdbserver :4000 a.out`
- Try to connect to gdb through the che gui by going to `run -> debug -> edit debug configuration`, set the port to 4000 as specified above
- The debug window opens briefly and crashes instantly (breakpoints are visible, but no variables. "gdb 7.1.. flashes up briefly in the top right corner of the debug window)

**OS and version:**    
Debian 9 (64bit), Chrome 62

**Diagnostics:** 

![debug-not-working](https://user-images.githubusercontent.com/26924052/31808277-b28d522a-b572-11e7-8b25-bef0747fdb5e.gif)

TODO:
- [x] Pass absolute path to the file where breakpoint is set
- [x] Move all server side tests into the docker container like Java Debugger does

ghost:
Can be potentially caused by gdb update in the stack.

tolusha:
The root of the issue:
```
Can't parse 'No source file named /console-cpp-simple/hello.cc.
Make breakpoint pending on fu' into GdbBreak
```
It means that https://github.com/eclipse/che/blob/master/wsagent/che-core-api-debug-shared/src/main/java/org/eclipse/che/api/debug/shared/model/Location.java#L16
has to consist the absolute path including `/projects`. The issue might be caused the recent Debugger refactoring.

Issue #5254, opened 2017-05-31
debugging golang with gdb - cannot connect to localhost:8080


JCzz:
**Reproduction Steps:**

1. Running Che with the following docker run:
docker run -it --privileged --rm -v /var/run/docker.sock:/var/run/docker.sock -v $(PWD):/data eclipse/che:latest start

2. Create Workspace; selecting go stack with web-go-simple

3. install gdb:
sudo apt-get update
sudo apt-get install gdb

4. In the Che editor menu: "Run / Edit Debug Configuration"
5. Select the "gdb/ remote gdb" with port 8080.
6. Press "Debug"

Error message: "Can not connect to: localhost:8080."

**OS and version:**    
Mac 10.12.3
Docker version: 17.03.1-ce

ghost:
@JCzz first off, you should start your workspaces in a privileged mode. In your che.env file, look for `PRIVILEGED` and set it to true.

Secondly, you should start gdbserver on port 8080, and then connect debugger to it.

Lastly, I do not think Che supports golang debugger with current gdb implementation.

JCzz:
I am not sure what you mean, but I have --privileged in step 1. ?

JCzz:
Can you describe what you do to start the gdbserver?

ghost:
This was you run CLI container in a privileged mode.

As to debugging golag, it is not supported right now in Eclipse Che.

JCzz:
Ok, thank you for the information. Looking forward to hawing go onboard a some point :-)

Issue #4719, opened 2017-04-06
Eclipse Che - Debugging C++ code


MuraliVidhya:
Hi,

In Che workspace (c++ stack), i created new c++ project (basic application with some cout statements). i could run the application successfully. In order to debug - as in https://www.eclipse.org/che/docs/ide/debug/index.html#gdb, configuration is done. But the debug windows does not show anything other than breakpoints, is this a known issue ? Any support would be helpful.

Environment: Linux 
Che : 5.7.0

ghost:
@mvidhya86 currently, using gdb debugger required privileged mode for a workspace container.

In your che.env, uncomment the following line, set it to `true` and restart Che:

`#CHE_DOCKER_PRIVILEGED=false`

This is a Docker security limitation.

MuraliVidhya:
Hi,
I modified che.env and restarted,  on click of debug, i get "Can not connect to: localhost ".

ghost:
Can you provide a screenshot?

MuraliVidhya:
[cid:image001.png@01D2AEE6.0A7BEAD0]





Can you provide a screenshot?

ghost:
I can't see the image, sorry.

mvidhya86:
https://mail.google.com/mail/u/0/s/?view=att&th=15b42a6067e0150b&attid=0.1&disp=attd&safe=1&zw

mvidhya86:
From: Vidhya Murali (RBEI/ETC3)
Sent: Thursday, April 06, 2017 2:57 PM
To: 'eclipse/che' <reply@reply.github.com>; eclipse/che <che@noreply.github.com>
Cc: Mention <mention@noreply.github.com>
Subject: RE: [eclipse/che] Eclipse Che - Debugging C++ code (#4719)

[cid:image001.png@01D2AEED.F0C5E930]

ghost:
@mvidhya86 attach screenshot to the issue on Github. Do not reply via email

ghost:
Works on my side:

![image](https://cloud.githubusercontent.com/assets/5337267/24751711/4c47234a-1ad3-11e7-8070-173476f484ac.png)

Make sure your workspace container runs in a privileged mode (run docker inspect against it). Also, make sure you compile your binary with `-g`

MuraliVidhya:
@eivantsov - In che.env - CHE_DOCKER_PRIVILEGED is set to true. After restarting the container, 
[...]

Issue #4524, opened 2017-03-22
gdbserver doesn't work in workspace with docker's privileged mode = false


dmytro-ndp:
**Reproduction Steps:**
1. Start Che with environment property `CHE_DOCKER_PRIVILEGED=false`.
1. Create CPP workspace with C++ project.
1. Make C++ project and start gdbserver to debug this project by the command:
```cd ${current.project.path} && make && gdbserver :8001 ./a.out```

After that there is error message was displayed in command panel:
> gdbserver: linux_ptrace_test_ret_to_nx: Cannot PTRACE_TRACME: Operation not permited
> gdbserver: linux_ptrace_test_ret_to_nx: status 256 is not WIFSTOPPED!

![screenshot from 2017-03-22 22-40-52](https://cloud.githubusercontent.com/assets/1197777/24220713/0eed3d20-0f54-11e7-8212-7bec4d96842e.png)

At the same time **gdbserver** started correctly when Che had environment property `CHE_DOCKER_PRIVILEGED=true`.

**OS and version:**    
Ubuntu 16.06, Fedora, Linux Mint 18
Eclipse Che ver. 5.6.0-SNAPSHOT

At the same time in Ubuntu 14.04 gdbserver works correctly.

**Selenium tests**: there is test [CppProjectDebuggingTest.shouldDebugCppProject()](https://github.com/codenvy/qa/blob/master/src/test/java/com/codenvy/ide/debugger/CppProjectDebuggingTest.java#L119) which should be enabled after this issue has been fixed.

TylerJewell:
Blocker bugs should only be reserved for data corruption or system outage issues.  This is definitely a P1.

dmytro-ndp:
I see.

ghost:
How are we going to solve this? I think I had this issue before. With new Docker security policies, it is what it is. I do not think this is a Che problem at all. Please see -https://github.com/docker/docker/issues/7276#issuecomment-194137403

Also, do debug a local binary, one does not need to start gdbserver at all. Just gdb is more than enough.

dmytro-ndp:
@eivantsov: thank you for commenting on this. Could you, please, show an example of debugging C++ project without **gdbserver** withing the workspace from our C++ stack in Che by using GDB debug configuration?

ghost:
[...]

PR #6856, opened 2017-10-20
Add support for Docker SecurityOpt (allowing GDB Debugging on unprivileged docker containers)

hkolvenbach:
Could someone please review the use of config variables? I am not entirely sure if I understand the translation from `CHE_DOCKER_SECURITYOPT` in `che.env` to `@Named('che.docker.securityopt')` correctly.
This also needs https://github.com/eclipse/che/issues/6855 to be fixed to be tested

### What does this PR do?
This PR allows to specify the `securityOpt` parameter in `HostConfig` to pass parameters to the docker daemon during container creation. the `securityOpt` parameter is equal to the command line option `--security-opt` of the command line. The value `seccomp:unconfined` allows gdb to run in unprivileged containers. Kubernetes uses `seccomp:unconfined` by default, but docker does not.

Related issues:
https://github.com/eclipse/che/issues/4719
https://github.com/eclipse/che/issues/4524
https://github.com/eclipse/che/issues/5254
https://github.com/eclipse/che/issues/4284

- Add new config variable `che.docker.securityopt` / `CHE_DOCKER_SECURITYOPT`
- Support for GDB in unprivileged docker containers

#### Release Notes

- Support for GDB in unprivileged docker containers

#### Docs PR

https://github.com/eclipse/che-docs/pull/305


benoitf:
@hkolvenbach could you please run the command 
```
mvn -o com.coveo:fmt-maven-plugin:2.0.0:format
```
jenkins CI job is failing to invalid format in your modified files.
thanks

hkolvenbach:
@benoitf sorry, probably still formatted with the old version. should be fixed now

benoitf:
@hkolvenbach yes thanks.
This is related to the new java formatter change I introduced yesterday

skabashnyuk:
Any volunteers to port to che6? @hkolvenbach @tolusha ?

dockerfiles/init/manifests/che.env
```diff
@@ -111,6 +111,15 @@
 #     containers from within their Docker-powered workspace.
 #CHE_DOCKER_PRIVILEGED=false
 
+# Host SecurityOpt Parameters
+#     This parameter allows to specify text values for the docker "--security-opt" parameter
+#     or the HostConfig{SecurityOpt:[]} value of the docker machine API 
+#     (https://docs.docker.com/engine/api/v1.33/#operation/ContainerCreate)
```
garagatyi: We use 1.21 API by default for compatibility with older docker versions. Can you point to that version instead?

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -161,6 +161,11 @@ che.docker.always_pull_image=true
 # to be able to launch their own runtimes which are embedded Docker containers.
 che.docker.privileged=false
 
+# This parameter allows to specify custom security options for the created docker container
+# seccomp:unconfined is the default for kubernetes, but not for docker. This is needed
+# for debugging with gdbserver
+che.docker.securityopt=seccomp:unconfined
```
garagatyi: Do we want to change default settings or just allow to change it? @hkolvenbach @tolusha

garagatyi: Have you tried it on systems that do not support seccomp? Is it just ignored?

plugins/plugin-docker/che-plugin-docker-client/src/main/java/org/eclipse/che/plugin/docker/client/json/ContainerConfig.java
```diff
@@ -333,14 +333,18 @@ public ContainerConfig withExposedPorts(Map<String, Map<String, String>> exposed
     return this;
   }
 
+  // there is no "SecurityOpts" in the docker specification, only HostConfig{"SecurityOpt" .[]}
```
garagatyi: Please, use javadocs comments style to provide documentation, e.g.
```
/*
 * There is no "SecurityOpts" in the docker specification, only HostConfig{"SecurityOpt" .[]}
 */
```

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -161,6 +161,11 @@ che.docker.always_pull_image=true
 # to be able to launch their own runtimes which are embedded Docker containers.
 che.docker.privileged=false
 
+# This parameter allows to specify custom security options for the created docker container
+# seccomp:unconfined is the default for kubernetes, but not for docker. This is needed
+# for debugging with gdbserver
```
garagatyi: Since it is supposed to be injected as an array can you add comment that values are separated by commas

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
garagatyi: Can you elaborate on how to set this option to default docker behavior?

dockerfiles/init/manifests/che.env
```diff
@@ -111,6 +111,15 @@
...
 #CHE_DOCKER_PRIVILEGED=false
 
+# Host SecurityOpt Parameters
+#     This parameter allows to specify text values for the docker "--security-opt" parameter
+#     or the HostConfig{SecurityOpt:[]} value of the docker machine API 
+#     (https://docs.docker.com/engine/api/v1.33/#operation/ContainerCreate) 
+#     This is useful if you want to run e.g. Cpp debugging within the container through gdbserver
+#     See https://github.com/eclipse/che/issues/4284 for details. This is only necessary for docker,
+#     openshift seems to set "seccomp:unconfined" by default
+#CHE_DOCKER_SECURITYOPT="seccomp:unconfined"
```
garagatyi: I think that quotes should not be used here

garagatyi (commented):
Nice PR! 
You used properties injection correctly.
Can you take a look at my inlined comments?

hkolvenbach:
Thanks for the remarks, I will fix them. Regarding the Che6 port: I can look into it, but not sure if I will manage to do it this week.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
hkolvenbach: @garagatyi didn't try it yet, any specific system I should test with?

hkolvenbach: I would say yes to behave in the same way as the kubernetes / openshift implementation does. Any comments on this @tolusha?

garagatyi: I don't know where to test it and whether it is worth checking, just thought how it can influence users.
I tried to put an invalid value of seccomp profile and another security option. Both failed. So I assume that this would prevent Che starting workspaces if users host doesn't support seccomp.

hkolvenbach: @garagatyi default would be an empty parameter

dockerfiles/init/manifests/che.env
hkolvenbach: fixed

dockerfiles/init/manifests/che.env
hkolvenbach: fixed

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
hkolvenbach: thanks for the feedback, i will test with different values for securityopt and see how it behaves. If unsupported values fail, i will leave it empty as default

garagatyi:
@hkolvenbach Che6 has this area massively refactored. If you need any help you can contact me

garagatyi:
@hkolvenbach I think that in case CHE_DOCKER_SECURITYOPT is set to empty value, e.g.: 
```
CHE_DOCKER_SECURITYOPT=
```
the current default behavior will be used, right?
I'm afraid container creation would fail in that case. Can you check? If I'm correct, please, take a look at [this class](https://github.com/eclipse/che/blob/master/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/dns/DnsResolversProvider.java#L33) for the hint.

hkolvenbach:
@garagatyi thanks for the support, i will come back to it when i am working on the che6 port. 

i did a few more tests and can confirm that docker will definitely fail if any unknown value is provided. so i will change it do not provide anything on default. null or empty string seems to work though (at least on linux).

i will also check if an empty string will result in the default behavior.

hkolvenbach:
@garagatyi i just confirmed that container creation will fail with an empty string, only null is allowed. do i understand correctly, that i need to implement a `SecurityOptProvider` similar to the `dnsResolversProvider` and create a seperate named parameter for it? or is there any easier way to check the `che.docker.securityopt` parameter for empty strings and set it to null instead?

garagatyi:
I think that unconfined might be OK as a default value. I just wanted to mention it to hear a feedback from the community if any. Wrong values in other places of che.env would definitely brake workspaces start also, so it is not the case. 
But the empty value has to be fixed because it is definitely what someone might use. Yes, you need a provider similar to DnsResolversprovider to fix that issue. But you don't have to inject provider as named string - just inject provider directly and get value from it in the constructor of MachineProviderImpl.

garagatyi:
@hkolvenbach can you point me out to the place where unconfined profile is used in k8s?

hkolvenbach:
@garagatyi i actually got that from another issue: https://github.com/eclipse/che/issues/4284#issuecomment-301487562 which links to https://github.com/kubernetes/kubernetes/issues/20870. havent tested it myself though

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -161,6 +161,16 @@ che.docker.always_pull_image=true
...
 che.docker.privileged=false
 
+# This parameter allows to specify custom security options for the created docker container.
+# seccomp:unconfined is the default for kubernetes, but not for docker. This is needed
+# for debugging with gdbserver. See https://github.com/eclipse/che/issues/4284 for details.
+# Parameters are passed as an array, so you can add multiple 
+# values comma seperated. Please also see https://docs.docker.com/engine/api/v1.21/#2-endpoints,
+# in particular the "HostConfig":{"SecurityOpt": []} entry. If this parameter is empty,
+# docker blocks certain Syscalls by default https://docs.docker.com/engine/security/seccomp/
+# WARNING: if you give incorrect values, docker gives an error and doesn't start the containre
```
garagatyi: Here is a little typo: `containre` -> `container`. The same typo is in che.env

plugins/plugin-docker/che-plugin-docker-machine-auth/src/main/java/org/eclipse/che/plugin/docker/machine/AuthMachineProviderImpl.java
```diff
@@ -53,6 +53,9 @@ public AuthMachineProviderImpl(
       @Named("machine.docker.machine_volumes") Set<String> allMachinesSystemVolumes,
       @Named("che.docker.always_pull_image") boolean doForcePullOnBuild,
       @Named("che.docker.privileged") boolean privilegedMode,
+      // @Named("che.docker.securityopt") String[] securityOpt,
```
garagatyi: Can you clean up the commented lines?

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
```diff
@@ -161,6 +161,16 @@ che.docker.always_pull_image=true
...
+che.docker.hostconfig.securityopt=NULL
```
garagatyi: I would suggest using the previous name name `che.docker.securityopt` since it is much user-friendly. 
Then inject provider in `MachineProviderImpl.java` and assign field with `this.securityOpt = securityOptProvider.get();`

hkolvenbach:
@garagatyi I tried to implement it exactly like the DnsResolversProvider, however I get the following error on startup:

```
eclipse.che.inject.CheBootstrap
com.google.inject.CreationException: Unable to create injector, see the following errors:

1) No implementation for java.lang.String[] annotated with @com.google.inject.name.Named(value=che.docker.hostconfig.securityopt) was bound.
  while locating java.lang.String[] annotated with @com.google.inject.name.Named(value=che.docker.hostconfig.securityopt)
    for the 1st parameter of org.eclipse.che.plugin.docker.machine.SecurityOptProvider.<init>(SecurityOptProvider.java:33)
  at org.eclipse.che.plugin.docker.machine.SecurityOptModule.configure(SecurityOptModule.java:21) (via modules: org.eclipse.che.api.deploy.WsMasterModule -> org.eclipse.che.plugin.docker.machine.SecurityOptModule)

1 error
	at com.google.inject.internal.Errors.throwCreationExceptionIfErrorsExist(Errors.java:470)
	at com.google.inject.internal.InternalInjectorCreator.initializeStatically(InternalInjectorCreator.java:155)
	at com.google.inject.internal.InternalInjectorCreator.build(InternalInjectorCreator.java:107)
	at com.google.inject.Guice.createInjector(Guice.java:99)
	at org.everrest.guice.servlet.EverrestGuiceContextListener.getInjector(EverrestGuiceContextListener.java:140)
	at com.google.inject.servlet.GuiceServletContextListener.contextInitialized(GuiceServletContextListener.java:47)
	at org.everrest.guice.servlet.EverrestGuiceContextListener.contextInitialized(EverrestGuiceContextListener.java:85)
	at org.apache.catalina.core.StandardContext.listenerStart(StandardContext.java:4727)
	at org.apache.catalina.core.StandardContext.startInternal(StandardContext.java:5189)
	at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:150)
	at org.apache.catalina.core.ContainerBase.addChildInternal(ContainerBase.java:752)
	at org.apache.catalina.core.ContainerBase.addChild(ContainerBase.java:728)
	at org.apache.catalina.core.StandardHost.addChild(StandardHost.java:734)
	at org.apache.catalina.startup.HostConfig.deployWAR(HostConfig.java:952)
	at org.apache.catalina.startup.HostConfig$DeployWar.run(HostConfig.java:1823)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
``` 

any idea what I might have done wrong? The configuration variable is called `CHE_DOCKER_HOSTCONFIG_SECURITYOPT` and is then translated into `che.docker.securityopt` by the securityOptProvider

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
hkolvenbach: I tried that approach, but it didn't work (I couldn't change `MachineProviderImplTest.java` to work with the provider). If no test is fine, I can just delete the test part and go back to `securityOptProvider.get()`

plugins/plugin-docker/che-plugin-docker-machine-auth/src/main/java/org/eclipse/che/plugin/docker/machine/AuthMachineProviderImpl.java
hkolvenbach: fixed

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/codenvy/che.properties
hkolvenbach: fixed

garagatyi:
@hkolvenbach I believe that the reason is default value `NULL`. The problem is that NULL is not converted into array, so injection fails in this case unfortunately

garagatyi:
@eivantsov Can you check GDB on OpenShift?

ghost:
Yes, I will.

hkolvenbach:
@garagatyi @eivantsov from my side I would consider my implementation for che 5.x ready. I have tested it with empty values, single values and multiple values for `CHE_DOCKER_HOSTCONFIG_SECURITYOPT` and this works for docker. 

i will now start to look into the che 6 port.

garagatyi:
@hkolvenbach take a look on direct usage of provider in tests https://github.com/garagatyi/che/commit/2685f22732b69b00a7ed3439e074ec01a91c0b04

hkolvenbach:
@garagatyi thanks, i integrated your changes and adapted everything to use CHE_DOCKER_SECURITYOPT again. somehow your commit doesn't pass validation, should I squash and sign it?

hkolvenbach:
@garagatyi I am a bit lost in the Che6 source code. Could you point me to the file where the configuration for the docker container is generated? I hope i can take it from there

garagatyi:
@hkolvenbach yes, you can squash my commit. I've signed Eclipse agreement so there are no legal issues with that.
Take a look at how DNS settings are done in Che6 https://github.com/eclipse/che/blob/che6/infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/provisioner/dns/DnsSettingsProvisioner.java#L24-L24
https://github.com/eclipse/che/blob/che6/infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/provisioner/dns/DnsResolversProvider.java#L31-L31
https://github.com/eclipse/che/blob/che6/infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerInfraModule.java#L43-L43

ghost:
Just checked gdb on OpenShift. I was able to start gdb and debug a binary using command line.

garagatyi:
Great! @eivantsov maybe we should change defaults to that behavior?

ghost:
@garagatyi i think it's safe enough to do so.

hkolvenbach:
@garagatyi I finally found time to port this pull request to che6. Tested it with empty configuration, 1 configuration value `seccomp:unconfined` and 2 values `seccomp:unconfined,apparmor:unconfined`

here it is https://github.com/eclipse/che/pull/7203

garagatyi (approved):

garagatyi:
@skabashnyuk there is a port for che6 already

riuvshin (approved):

skabashnyuk (approved):

garagatyi:
@benoitf can you take a look?

benoitf (approved):

benoitf:
ci-build

benoitf:
thanks @hkolvenbach

Issue #6858, opened 2017-10-20
Add ability to get user public email from github  


Ohrimenko1988:
For our tests we need to add ability to get user public email from github. This is related with cleanup tests which logins with github user name and password. Becouse when user login by github, created username will be not equals with github user name. Account will be created with github public email, so we need get this email from github for to do cleanup after test.

Ohrimenko1988:
Done

PR #6865, opened 2017-10-20
Add method 'getUserPublicEmail' to TestGitHubServiceClient

Ohrimenko1988:
### What does this PR do?
Add ability to get github user public email by using github api .
(che6 PR https://github.com/eclipse/che/pull/6866)
Issue https://github.com/eclipse/che/issues/6858


selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/client/TestGitHubServiceClient.java
```diff
@@ -203,4 +204,24 @@ private String createBasicAuthHeader(String username, String password)
...
+
+    @SuppressWarnings("unchecked")
+    List<Map<String, String>> properties =
+        response.as(List.class, new TypeToken<List<Map<String, String>>>() {}.getType());
+
+    return properties
+        .stream()
+        .filter(map -> map.containsKey("email"))
+        .map(map -> map.get("email"))
+        .collect(Collectors.joining());
```
dmytro-ndp: What is the propose of joining() here? Did you try to test this method with github account with several public emails?

Ohrimenko1988:
[INFO] Che Core :: IDE :: App ............................. FAILURE
...
Found 4 non-complying files

I think this is not this PR fault

dmytro-ndp:
It's a PR fault so as there are changes in master which should be merged in branch to prevent an error.

dmytro-ndp (changes requested):
https://api.github.com/user/public_emails can return several emails: https://developer.github.com/v3/users/emails/#list-public-email-addresses-for-a-user
Please, take it into account.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/client/TestGitHubServiceClient.java
dmytro-ndp: I would recommend to get first email from the map instead of joining all of them, as a less error-prone variant.

dmytro-ndp (approved):

PR #6882, opened 2017-10-20
Add ability to run selenium tests for che launched on custom port.

AndrienkoAleksandr:
### What does this PR do?

Add ability to run selenium tests for che launched on custom port.

#### Release Notes
Add ability to run selenium tests for che launched on custom port.

#### Docs PR

Signed-off-by: Oleksandr Andriienko <oandriie@redhat.com>


dmytro-ndp (changes requested):
This variant doesn't work without parameter --port, which is not mandatory and should have default value=8080:
> [TEST] http://172.19.20.32 is down

So, I strongly recommend to test changes with at least to use cases:
1. Run tests with correct new parameter.
2. Run tests without new parameter.
3. Run tests with wrong new parameter.

Also, we have already had **--port** parameter added into the selenium tests of _Che6_. So, it have to be extremely useful to review [this commit](https://github.com/eclipse/che/commit/a614b2c88de32d633f4645c249dbf30bd2fb9585#diff-d54a4ec2c34436c239eade6514bb0822) made by @garagatyi.

PR #6946, opened 2017-10-24
Remove unused SystemInfo attributes (cpu, freeMemory, totalMemory) 

benoitf:
### What does this PR do?
In Java9, relying on "internal fields" of classes is not permitted by default. Only public methods should be used.
The SystemInfo class is relying on "internal specific JDK" to get some cpu and memory attributes.

We could delete the JDK implementation and only uses the "Unix" implementation in the class but as it seems this code is never used, I would propose to remove it entirely.

#5326 

#### Release Notes
N/A

#### Docs PR
N/A

Change-Id: I41bd108ce4b2af15693d57a5fa0c72caab932289


benoitf:
ci-build

benoitf:
as discussed on mattermost, some delay is required to ensure it's not used

benoitf:
ci-build

skabashnyuk (approved):

Issue #6904, opened 2017-10-23
[Che-6] Improve CheckMavenPluginTest for more stability


musienko-maxim:
The tests works with project explorer by native selenium clicks but this functionality is already checked in Prject Explorer  tests. But too often usage of native selenium double clicks may have potential problem like StaleElementReferenceException.
I think we should replace native invokations on Javascript for expand project explorer tree

PR #6955, opened 2017-10-24
Rework maven plugin test for more stability

musienko-maxim:
### What does this PR do?
* Remove checking of maven plugin progressor (because it was deleted in che6). But if  it will be returned - we'll enable it .
* Use native javascript invocation for collapse project explorer items that allow avoid of StaleElement exceptions
* Use ProjectService importer instand of importing project from github
* Add the test project to resources
* Improve locator for defining of type of project folders (simple folder, project folder, javaSourceFolder) in the expanded tree
* Use static imports
* Change test for avoid unexpected failing (test#3 fault when fault test#2)

https://github.com/eclipse/che/issues/6904
@dmytro-ndp, @vparfonov, @Ohrimenko1988


musienko-maxim:
ci-build

Ohrimenko1988 (approved):

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java
```diff
@@ -502,7 +502,7 @@ public void waitFolderDefinedTypeOfFolderByPath(String pathToFolder, String type
     String locator =
         "//div[@path='/"
             + pathToFolder
-            + "']//*[local-name() = 'svg' and @id='"
+            + "']/div/*[local-name() = 'svg' and @id='"
```
dmytro-ndp, line 505: It would be better to use just *id* here as identifier, instead of _xpath_, to prevent frequent changing of the code.

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/mavenplugin/CheckMavenPluginTest.java
```diff
@@ -48,97 +50,74 @@
...
-    commandServiceClient.createCommand(
-        "cd /projects/" + PROJECT_NAME + " && git checkout contrib-12042015",
-        CHECKOUT_COMMAND,
-        TestCommandsConstants.CUSTOM,
-        workspace.getId());
-
+    URL resource = getClass().getResource("/projects/check-maven-plugin-test");
+    testProjectServiceClient.importProject(
+        workspace.getId(),
+        Paths.get(resource.toURI()),
```
dmytro-ndp: It's better to use static import here to improve readability of code.

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/mavenplugin/CheckMavenPluginTest.java
```diff
@@ -48,97 +50,74 @@
...
+        PROJECT_NAME,
+        ProjectTemplates.MAVEN_SPRING);
```
dmytro-ndp: It's better to use static import here to improve readability of code.

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/mavenplugin/CheckMavenPluginTest.java
```diff
@@ -48,97 +50,74 @@
...
     ide.open(workspace);
     projectExplorer.waitProjectExplorer();
   }
 
   @Test
-  public void mavenStatusBarShouldDisplayResolvingProjectMessage() {
```
dmytro-ndp, line 71: Why this and other tests were removed from the test class? Does it mean the corresponded functionality disappeared from the Che? Or those tests were useless and tested nothing?

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java
musienko-maxim: I agree, but in this case it does not work because we  use @path for define a folder

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/mavenplugin/CheckMavenPluginTest.java
musienko-maxim: Status bar was removed from che6. But i hope it will be returned and we enable this one again.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java
dmytro-ndp: I see.

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/mavenplugin/CheckMavenPluginTest.java
dmytro-ndp: I see. What about next test *shouldExecuteCommandAndWaitTextInConsole*? It checked message inside console.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java
musienko-maxim: Also i don't have idea what we will be recognize folders with the same type in project tree. We can have many folders with the same type and the same id's. It can be problem and improve this is not trivial task.

vparfonov (approved):

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/mavenplugin/CheckMavenPluginTest.java
musienko-maxim: In this case it unnecessary step. Because it was done for checkout on tag in the just cloned project. After improving we don't need cloning project and checkout on tag it is extra steps that may be unstable.

musienko-maxim: Also we check correct work of maven plugin. Cloning and checkouting are checked in the Git package.

musienko-maxim:
ci-build

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/mavenplugin/CheckMavenPluginTest.java
dmytro-ndp: I see. Thank you for the answer.

dmytro-ndp (approved):

PR #7012, opened 2017-10-26
Fix deployment to Minishift with default behavior

garagatyi:
### What does this PR do?
Fix deployment to Minishift with the default behavior, in particular:
- increased timeout of waiting for Postgres start
- removed second check of Postgres after the first one (didn't get why there were 2 of them)
- fixed bugs in checking for Postgres status
- fixed TLS configuration for Minishift
- renamed deployment configuration


riuvshin (approved):

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
```diff
@@ -108,14 +108,24 @@ else
 fi
 
 CHE_IMAGE_REPO=${CHE_IMAGE_REPO:-${DEFAULT_CHE_IMAGE_REPO}}
-DEFAULT_CHE_IMAGE_TAG="spi"
+DEFAULT_CHE_IMAGE_TAG="nightly"
```
benoitf: nightly is coming from `che6` branch ?

riuvshin: we don't have spi tags any more, it should be nightly as in master. and nightly is default tag. If you want to use some custom tag it is possible to define it in ENV var

benoitf: default should be for now `che6` ?
https://hub.docker.com/r/eclipse/che-server/tags/

sleshchenko (approved):
Good idea to change default configuration values for different flavors instead of doing different changing of original yaml file :+1: Reworking all script in this way will help in avoiding copy/paste errors.

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
```diff
@@ -108,14 +108,24 @@ else
...
 CHE_IMAGE_TAG=${CHE_IMAGE_TAG:-${DEFAULT_CHE_IMAGE_TAG}}
 DEFAULT_CHE_LOG_LEVEL="INFO"
 CHE_LOG_LEVEL=${CHE_LOG_LEVEL:-${DEFAULT_CHE_LOG_LEVEL}}
-DEFAULT_ENABLE_SSL="true"
-ENABLE_SSL=${ENABLE_SSL:-${DEFAULT_ENABLE_SSL}}
 DEFAULT_K8S_VERSION_PRIOR_TO_1_6="true"
 K8S_VERSION_PRIOR_TO_1_6=${K8S_VERSION_PRIOR_TO_1_6:-${DEFAULT_K8S_VERSION_PRIOR_TO_1_6}}
+# OPENSHIFT_FLAVOR can be minishift or openshift
```
sleshchenko: fix this comment. Actually `OPENSHIFT_FLAVOR` can be minishift or openshift or **OCP**

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
benoitf: yes but AFAIK for now che6 scripts should use che6 binaries by default

l0rd (approved):

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
riuvshin: we can't change that for each branch, and to not forget change it before merge to master I would prefer to keep at as in master and use ENV var to override. Because I'm 100% it will be merged in master with che6 tags

benoitf: we don't change that for each branch, only "master" and "che6"

User running scripts with CLI 6.0.0-M1 will expect that it's running the correct "branch"/image by default

garagatyi:
@riuvshin @benoitf can we decide what behavior should be by default?

garagatyi:
@riuvshin are you ok with che6 tag in the script?

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
sleshchenko: It would be more convenient for development process to have here the same default value as docker images build scripts have. And it is `nightly`.

benoitf: yeah but I think there is something wrong because local and `dockerhub` values should match the same targeted branch.
you use the cli to install the scripts in a folder, so you expect the cli install the scripts for the cli version you've used.

as it's a puppet module, couldn't we have a property value that is automatically setting the correct value based on the image that we're building ?

garagatyi: We agreed to set che6 tag

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
garagatyi: ok

garagatyi:
@benoitf @riuvshin @sleshchenko I added more changes , can you take a look again

sleshchenko (approved):

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
```diff
@@ -438,7 +443,7 @@ cat "${CHE_DEPLOYMENT_FILE_PATH}" | \
     oc apply --force=true -f -
 else
   echo "[CHE] Deploying Che on OpenShift Container Platform (image ${CHE_IMAGE})"
-  DEFAULT_CHE_DEPLOYMENT_FILE_PATH=./che-spi-openshift.yml
+  DEFAULT_CHE_DEPLOYMENT_FILE_PATH=./che-che6-openshift.yml
```
benoitf: maybe we could rename file to be `che6-openshift.yml` instead of `che-che6-`

benoitf (approved):

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
garagatyi: :) yep

garagatyi:
@riuvshin @benoitf @l0rd @davidfestal I changed some stuff in the script. Now it behaves better in my opinion. Apart from the problem described [here](https://github.com/eclipse/che/issues/7087) Che starts without failure of builds. 
I also changed:
- increased timeout of waiting for Postgres start
- removed second check of Postgres after the first one (didn't get why there were 2 of them)
- fixed bugs in checking for Postgres status

Big thanks to @riuvshin for his assistance in the investigation of problems with deploy script.

benoitf (approved):
thanks @garagatyi for the update. I still validate the changes.

FYI I think the shortname for PostreSQL is Postgres, not potsgre

garagatyi:
yep, you are right. I've changed it.

benoitf:
@garagatyi do you will backport some fixes on master branch ?
like fixes on https://github.com/eclipse/che/blob/master/dockerfiles/init/modules/openshift/files/scripts/multi-user/wait_until_postgres_is_available.sh

dockerfiles/init/modules/openshift/files/scripts/multi-user/wait_until_postgres_is_available.sh
```diff
@@ -16,10 +16,10 @@ command -v jq >/dev/null 2>&1 || { echo >&2 "[CHE] [ERROR] Command line tool jq
 available=$(oc get dc postgres -o json | jq '.status.conditions[] | select(.type == "Available") | .status')
 progressing=$(oc get dc postgres -o json | jq '.status.conditions[] | select(.type == "Progressing") | .status')
 
-DEPLOYMENT_TIMEOUT_SEC=120
+DEPLOYMENT_TIMEOUT_SEC=1200
 POLLING_INTERVAL_SEC=5
 end=$((SECONDS+DEPLOYMENT_TIMEOUT_SEC))
-while [ "${available}" != "\"True\"" ] && [ ${SECONDS} -lt ${end} ]; do
+while [[ "${available}" != "\"True\"" || "${progressing}" != "\"True\"" ]] && [ ${SECONDS} -lt ${end} ]; do
```
davidfestal, line 22: The second part that you're adding: `|| "${progressing}" != "\"True\"" ` was removed some commits ago, because, in some use cases, the `progressing` status never comes to `True` even when the `available` status becomes `True`.  

Could you precise what is the issue this addition is trying to fix ?

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
```diff
@@ -288,8 +293,6 @@ if [ "${CHE_MULTI_USER}" == "true" ]; then
     else
         "${COMMAND_DIR}"/multi-user/deploy_postgres_only.sh
     fi
-
-    "${COMMAND_DIR}"/multi-user/wait_until_postgres_is_available.sh
```
davidfestal, line 292: I don't see why you remove this line. Allowing the Che server to start before either the Keycloak server or the che server leads to Che startup error, and the resuling pods do not work correctly.

afaik , by removing this line, you allow the Che server to start before the Postgres server is up and running, at least in the use-case when the dedicated Keycloak server is not used.

davidfestal (changes requested):

garagatyi:
@benoitf I was not going to do that since I work with the che6 branch and doesn't know the state of the script for master implementation of Che6

dockerfiles/init/modules/openshift/files/scripts/multi-user/wait_until_postgres_is_available.sh
garagatyi: @riuvshin helped me here and he said that when Postgres is available but not progressing is not enough to consider Postgres is fully online.

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
garagatyi: I see what you mean. In my case check of Postgres was broken and was called twice in `deploy_postgres_only.sh` and here. So I removed second call, but now I see that it was needed for `deploy_postgres_only.sh`.
I'm going to add `"${COMMAND_DIR}"/multi-user/wait_until_postgres_is_available.sh` right after `"${COMMAND_DIR}"/multi-user/deploy_postgres_only.sh`
Are you OK with this solution?

davidfestal: yes

dockerfiles/init/modules/openshift/files/scripts/multi-user/wait_until_postgres_is_available.sh
davidfestal: I'm really surprised about that. @riuvshin could you give more details about that ? If I remember correctly, we had encountered cases when `processing` *never* comes to `True`, even when the Postgres server is correctly up and running.

riuvshin: @davidfestal hm, no... As I remember our issue was that progressing was in 3 states:
 **empty**, **unknown**, **true**;
 while script was expected only 2 states **emtpy** and **true** and it was exiting when it was not **empty** _but_ also it wasn't yet **true**

l0rd (approved):

dockerfiles/init/modules/openshift/files/scripts/multi-user/wait_until_postgres_is_available.sh
davidfestal: OK, then

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
davidfestal: That would be great to also correctly support the case when an external Keycloak server is used.

garagatyi: done

garagatyi:
@davidfestal is it OK for you now?

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
davidfestal: Thanks.

davidfestal:
> @davidfestal is it OK for you now?

Sure

davidfestal (approved):

Issue #6891, opened 2017-10-22
[multi-tenant] can not create workspace if user name contains '@' e.g. (ibuziuk@redhat.com)


ibuziuk:
### Description
Failure when creating workspace with multi-tenant che-server if username  contains '@' e.g. (ibuziuk@redhat.com).

custom server evaluation strategy template used for multi-tenant is using the user
template variable which can be email, though the route creation mechanism in the `OpenshiftConnector`used the `*-che` namespace of the user (rh-che implementation of the multi-tenancy) but the assumption that` <keycloak preferred-user>-che == OSIO namespace` gets broken
for names that were created in Keycloak as email addresses

### Reproduction Steps
Try to create workspace with multi-tenat che-server against prod clustrer:

More info: 

> Could not start workspace springo-bt9bt. Reason: Start of environment 'default' failed. Error: Timeout. The Che server is unable to ping your workspace. This implies a network configuration issue, workspace boot failure, or an unusually slow workspace boot.

workspace url - https://che-ibuziuk-che.glusterpoc37aws.devshift.net/dashboard/#/ide/ibuziuk@redhat.com/wksp-48p5

Routes:
![image](https://user-images.githubusercontent.com/1461122/31860519-2bd59256-b71b-11e7-82d9-b01d68b6540d.png)

**OS and version:**
Tested against `dfestal/che-server-multiuser:nightly-fabric8` [1] deployed on aws against prod osio cluster

[1] https://hub.docker.com/r/dfestal/che-server-multiuser/tags/

ibuziuk:
cc: @davidfestal

musienko-maxim:
Is it still actual?

ibuziuk:
This one was fixed some time ago. Closing

musienko-maxim:
Thx!

PR #7017, opened 2017-10-26
Fix for https://github.com/eclipse/che/issues/6891

davidfestal:
Make the `OpenshiftConnector` workspace route naming rule consistent
with custom server evaluation strategy in all cases (especially in case
of unexpected openshift.io user name).

### What does this PR do?

Make the `OpenshiftConnector` workspace route naming rule consistent
with custom server evaluation strategy in all cases (especially in case
of unexpected openshift.io user name).

This change is only relevant when using multi-tenancy (currently implemented in the `rh-che` distribution), and doesn't impact other use-cases.

https://github.com/eclipse/che/issues/6891


riuvshin (approved):

benoitf (approved):

benoitf:
ci-build

ibuziuk (approved):

l0rd (approved):
LGTM

PR #7019, opened 2017-10-26
Fix multiuser network with custom port

### LINKED ISSUE #6885: Can´t start workspace in multiuser mode
Then of install and run eclipse che in multiuser mode, i can create a new workspace but i cant run it.
I run a server in a vm virtualbox with the next command:
`docker run -it -v /var/run/docker.sock:/var/run/docker.sock -v /data/example:/data -e CHE_PORT=9000 -e CHE_CONTAINERT=example -e CHE_MULTIUSER=true eclipse/che start`

Then i create a new workspace for the user, but when i run it this error appears:

> Could not start workspace wksp-q0qx. Reason: Start of environment 'default' failed. Error: Error response from docker API, status: 404, message: No such network: che_che-network

captura

**OS and version:**  
Ubuntu 16.04.3

riuvshin:
Fix multiuser network with custom port

fixes: https://github.com/eclipse/che/issues/6885

Issue was due to hardcoded network name for multiuser, but network name in comose depends on compose project name


dockerfiles/base/scripts/base/startup_01_init.sh
```diff
@@ -158,12 +158,8 @@ init_global_vars() {
   DEFAULT_CHE_LICENSE=false
   CHE_LICENSE=${CHE_LICENSE:-${DEFAULT_CHE_LICENSE}}
 
-  if [[ "${CHE_CONTAINER_NAME}" = "${CHE_MINI_PRODUCT_NAME}" ]]; then   
-    if [[ "${CHE_PORT}" != "${DEFAULT_CHE_PORT}" ]]; then
+  if [[ "${CHE_CONTAINER_NAME}" = "${CHE_MINI_PRODUCT_NAME}" ]]; then
       CHE_CONTAINER_NAME="${CHE_CONTAINER_PREFIX}-${CHE_PORT}"
-    else 
-      CHE_CONTAINER_NAME="${CHE_CONTAINER_PREFIX}"
```
benoitf: AFAIK this is changing the behavior

From what I'm seing with these changes che server will be named `che-8080` by default instead of only `che` as before

ghost (approved):

dockerfiles/base/scripts/base/startup_01_init.sh
riuvshin: that's right! 
I was waiting for this question :)
I just wanted to have less if statements and more clear flow.
So how do you think is it that critical for us to keep 'default' `che` name if port is default?
The issue isthat I don't have access in erb template to `$DEFAULT_CHE_PORT` so I will use 8080 to check which is not that good...

benoitf: yes, it's horrible to see `che-8080` for end user instead of `che`
(and if there are some user custom scripts, it will make them fail as it will expect to find a `che`container)

I see the issue but current fix is very odd to me.
AFAIK you can have all properties that you want inside erb templates.
BTW, can't we set the network name with docker compose file, so we can keep `che`container name but network will have the port prefix

riuvshin: Ok so I will handle default values 
compose will always prepend network name with compose project name to avoid conflicts, and if specify network name in compose file compose will not try to create  that network it will expect that it is already exist.

riuvshin:
@benoitf updated. now defaults are respected :)

skabashnyuk (approved):

benoitf (approved):

PR #7027, opened 2017-10-27
Fix DockerInternalRuntime test that is not throwing an expected checked exception

benoitf:
### What does this PR do?

RuntimeStartInterruptedException is a checked exception that can't be thrown by createNetwork method
It is reported with new version of Mockito 

Replacing then the checked exception by the exception that createNetwork method can throw (which is InternalInfrastructureException)

note: I can change the code to what you'll expect, it's just that current code is not possible.

#5326 

#### Release Notes
N/A

#### Docs PR
N/A

Change-Id: Icccf451ef5570494441c583495f41382246e10f1


garagatyi (approved):

benoitf:
ci-build

akorneta (approved):

Issue #6972, opened 2017-10-25
[Che-6] Factory by direct urls feature works wrong


musienko-maxim:
### Reproduction Steps
Consume factory like: ```your_host/f?url=https%3A%2F%2Fgithub.com%2FTarrantella%2FgitPullTest%2Ftree%2Fmaster%2Fmy-lib```
Expected result: factory should be consumed, workspace and prject should be created.
Actual result: We get responce message: ```message:Workspace default environment configuration required```. ```environments``` values are not set in the factory.

**OS and version:**    
Fedora 25, chrome browser, Che-6 M1
**Diagnostics:** 
The selenium tests depended on this:
* DirectUrlFactoryWithKeepDirectory
* DirectUrlFactoryWithRootFolder
* DirectUrlFactoryWithSpecificBranch

PR #7032, opened 2017-10-27
Fix URLFactoryResolver

sleshchenko:
### What does this PR do?
Uncomments and adapts URLFactoryBuilder.

https://github.com/eclipse/che/issues/6972

#### Release Notes
Fixed URLFactoryResolver

### DIFF
--- a/plugins/plugin-urlfactory/src/main/java/org/eclipse/che/plugin/urlfactory/URLFactoryBuilder.java
+++ b/plugins/plugin-urlfactory/src/main/java/org/eclipse/che/plugin/urlfactory/URLFactoryBuilder.java
@@ -11,6 +11,9 @@
 package org.eclipse.che.plugin.urlfactory;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.Collections.singletonList;
+import static java.util.Collections.singletonMap;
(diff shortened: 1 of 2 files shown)


skabashnyuk (approved):

garagatyi (approved):

plugins/plugin-urlfactory/src/main/java/org/eclipse/che/plugin/urlfactory/URLFactoryBuilder.java
```diff
@@ -40,17 +44,21 @@
...
-  @Inject private URLChecker URLChecker;
-
-  /** Grab content of URLs */
-  @Inject private URLFetcher URLFetcher;
+  private final URLChecker urlChecker;
+  private final URLFetcher urlFetcher;
+  private final FactoryMessageBodyAdapter factoryAdapter;
 
-  @Inject private FactoryMessageBodyAdapter factoryAdapter;
+  @Inject
```
benoitf, line 55: is there a reason to move to constructor based injection ?

sleshchenko: I like this way. At least because IDE doesn't show warnings that fields are not assigned

benoitf (approved):

Issue #7008, opened 2017-10-26
Command output widget performance


skabashnyuk:
### Description
When I've tried to build che in che I've noticed that overall performance of browser tab become slower and slower. Something like starting from 1/3 or 1/4 of total build. I can assume that it happens because of a big amount of output produced by build command. I can propose that in one moment of time we can show only limited number of rows(let's say 100) and then use our exec agent api to get remaining logs https://github.com/eclipse/che/blob/che6/agents/go-agents/docs/ws_api.md#get-process-logs
![2017-10-26 11 36 54 2](https://user-images.githubusercontent.com/1614429/32042984-14a2cb1a-ba42-11e7-8c87-698cbb32ebb3.png)

### Reproduction Steps
Run build che in che.

**OS and version:**    
OSX
**Diagnostics:** 
Browser tab become unresponsive

gazarenkov:
Can we try not sending the output to IDE to check if it is the real reason first?

vitaliy-guliy:
Browser places all output in console in processes panel. When amount of lines become more 3000-5000 browser become working slower and slower. The reason is browser renders the console widget on each adding a new line and when amount of lines too much browser hangs.
We need to improve console widget and show only part of output. Other part of output we can easily keep in the memory and show when user scrolls up the console.

gazarenkov:
OK, solution proposed by @skabashnyuk looks good, moreover I think (and @eivantsov remember this too) we had this back in version 3.x or so.
Definitely it is blocker for more-less big projects like Che

AndrienkoAleksandr:
If problem is slow rendering, like variant to resolve problem we can use for rendering command output terminal widget (instead of our custom GWT widget): xterm.js in read only mode. This widget is faster renders command output and for this widget we can use "attach" plugin for bufferization data to render. And very often build tools use some progressors with ANSI escape sequences for some pretty graphic effects(loaders, pseudo graphic animations...). Our GWT command output widget doesn't has ANSI escape sequences parser and display such output incorrectly. xterm.js contains embedded control sequences parser. Or we should really rework current command output widget.

gazarenkov:
@AndrienkoAleksandr your comment is great, thanks a lot!
I would say we need to optimize our output channels too.
Would be great if you can make more detailed analysis of xterm.js vs ours output processing and share your ideas how we can optimize it. This is really important and high priority for us.
Meantime (I consider implementing it takes time), I would use something "fast and dirty".

[...]

Issue #7029, opened 2017-10-27
Compare rendering performance terminal widget vs CommandOutputConsole.


AndrienkoAleksandr:
Description
Compare rendering performance terminal widget vs CommandOutputConsole
What should be done:
- [x]  terminal widget should be used for rendering command output in read only mode.
- [x] compare performance terminal rendering command output vs CommandOutputConsole
- [x] check if it is possible to use terminal "attach" plugin for bufferization output and check performance again.
- [x] create Orion based widget to render command output and check performance.

AndrienkoAleksandr:
Conclusions
==========
I completed comparing the performance of current CommandOutput widget and widget based on xterm.js and widget based on Orion. During investigation I created widgets based on xterm.js and Orion. For all widgets I set up scroll-back size 5000 lines. For test I used rendering of an output from a file and received this output with the help of a custom command. This command reads ***n*** lines from a huge file (40148 lines). Each step I changed the amount of lines (100, 200, 500, 1000, 2500, 5000, 10000) and measured the performance (full rendering time) with the help of Google Chrome Development tools. You can see the comparison in the table and the graphic (time in seconds):

|   Quantity lines               |  100   |  200   |   500  |  1000 | 2500 |   5000   |  10000 |
| --------------------------------|:---------|:--------|:--------|:---------|:-------|:-----------|:----------|
|  Console output widget  | 0.464 | 0.724 | 1.456 | 3.158 | 9.7   | 26.3138 |   49.2   |
|  Xterm.js based widget  | 0.384 | 0.514 | 0.756 | 1.198 | 2.67 |   3.974   | 5.684   |    
| Orion based widget       | 0.532 | 0.916 | 1.068 |  1.86  | 7.39 |     9.3     | 16.646 |

![comparison_perfomance](https://user-images.githubusercontent.com/6873095/32785349-06374ac0-c95a-11e7-8698-9fb641dc3881.png)

On each step the fastest was xterm.js based widget, the second was Orion based widget, and the last one was current CommandOutputWidget. On the big output xterm.js was faster than current widget in 3-9 times. Attach plugin for terminal was designed for using with web-sockets so we can't use it to speed up terminal more.

Xterm.js (2.4.0)
--------------------
Pluses: rendering performance, xterm.js contains control sequence parser (to display text colors, progressors, text-based user interface), contains auto scroll feature, linkify url feature.
[...]

PR #7051, opened 2017-10-27
Limit command output to fix problem with crashed browser tab

skabashnyuk:
### What does this PR do?
With help of @vitaliy-guliy  Limit command output to fix problem with crashed browser tab

Fast fix discussed here https://github.com/eclipse/che/issues/7008#issuecomment-339650502
Now we can build che in che

#### Release Notes
Command output limited to 500 lines

### DIFF
--- a/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/console/OutputConsoleViewImpl.java
+++ b/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/console/OutputConsoleViewImpl.java
@@ -317,6 +317,11 @@ public class OutputConsoleViewImpl extends Composite implements OutputConsoleVie
 
   @Override
   public void print(final String text, boolean carriageReturn, String color) {
+
+    if (consoleLines.getElement().getChildCount() > 500) {
+      consoleLines.getElement().getFirstChild().removeFromParent();
+    }
+
     if (this.carriageReturn) {
       Node lastChild = consoleLines.getElement().getLastChild();
       if (lastChild != null) {


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/console/OutputConsoleViewImpl.java
```diff
@@ -317,6 +317,11 @@ public void print(String text, boolean carriageReturn) {
 
   @Override
   public void print(final String text, boolean carriageReturn, String color) {
+
+    if (consoleLines.getElement().getChildCount() > 500) {
```
tolusha, line 321: It is better to have a constant but not a magic number ^)

skabashnyuk: I would say this change would not make this fast and dirty fix less dirty.

vitaliy-guliy:
Looks good for start

vitaliy-guliy (approved):

vparfonov (approved):

benoitf (dismissed):
it needs to be parametrized (or use buffering)
we shouldn't do "quick and dirty" code

benoitf:
note that you can use `quiet`flag as well for maven

skabashnyuk:
What kind of parametrization? 
@benoitf are you ready to implement that until next Thursday? because https://github.com/eclipse/che/issues/7008 is a real blocker.
FYI @AndrienkoAleksandr is going to make some test here https://github.com/eclipse/che/issues/7029 but it would not be very quick

benoitf:
parametrization is to not hardcode limits in a product

gazarenkov:
makes no sense to improve workaround,
might be the choice if we will not be able to make something better after @AndrienkoAleksandr investigation

Issue #6953, opened 2017-10-24
[REGRESSION] che workspaces fail to start on mac os


riuvshin:
on Mac OS workspace start fails with:

```
Could not start workspace wksp-qboq. Reason: Start of environment 'default' failed. Error: Timeout. The Che server is unable to ping your workspace. This implies a network configuration issue, workspace boot failure, or an unusually slow workspace boot.
```

che run command:
```
docker run --rm -it -v /var/run/docker.sock:/var/run/docker.sock \
                    -v /Users/roman/development/codenvy_projects/.che_multi:/data \
                    -e CHE_MULTIUSER=true \
                    eclipse/che-cli:nightly $@
```

this regression appear after: https://github.com/eclipse/che/pull/6738
@davidfestal can you please check this.

on tag `5.19.0` works as expected

riuvshin:
update, workspaces does not start even for single user CHE

riuvshin:
update:
regression caused by this: https://github.com/eclipse/che/pull/6738/files#diff-405a4d0485b220b257fa4d05ded17040R73
```
-    String wsAgentPingUrl = wsAgentServer.getProperties().getInternalUrl();
+    String wsAgentPingUrl = wsAgentServer.getUrl();
```
in case if external url is localhost which is default for mac and windows OS ws master won't be able to connect to ws agent on localhost because it is running in a container with own localhost

riuvshin:
fixed! Thanks @davidfestal

Issue #7058, opened 2017-10-27
[multi-tenant] Can not create workpace against multi-tenant che-server when ssl is enabled


ibuziuk:
### Description
Error while creating workspaces with mt che-sever deployed on https://glusterpoc37aws.devshift.net:8443 :

> 2017-10-27 17:07:27,631[aceSharedPool-0]  [ERROR] [o.e.c.a.agent.WsAgentLauncher 137]   - Fail pinging ws agent with http://wsagent-egrqz99hlzu1byha-ibuziuk-che.8a09.starter-us-east-2.openshiftapps.com/api/ url in workspaceegrqz99hlzu1byha workspace in machinenkwzppnuq6kfmwjd machine on localhost node.

> 2017-10-27 17:07:29,461[aceSharedPool-0]  [ERROR] [o.e.c.a.w.s.WorkspaceManager 683]    - Timeout. The Che server is unable to ping your workspace. This implies a network configuration issue, workspace boot failure, or an unusually slow workspace boot.

### Reproduction Steps
Create a workpace on mt che-server deployed on https://glusterpoc37aws.devshift.net:8443  with ssl enabled: 

**OS and version:**    
https://glusterpoc37aws.devshift.net:8443 
image `rhche/che-server-multiuser/903ed59-fabric8-4ee7187`

**Diagnostics:** 
For some reason protocol for `WsAgentPingRequestFactory` returns an http url even with ssl enabled

ibuziuk:
cc: @davidfestal

eivantsov:
Why is ws-agent route http but not https?

PR #7004, opened 2017-10-25
Fix issue #6953 ...

davidfestal:
... by restoring the previous `WsAgentPingRequestFactory` behavior, and
providing a new custom server evaluation template strategy when both
internal *and* external addresses should be based on the custom
template. 

### What does this PR do?

This PR restores the previous `WsAgentPingRequestFactory` behavior, and
provides a new custom server evaluation template strategy when both
internal *and* external addresses should be based on the custom
template. 

### What issues does this PR fix or reference?

issue #6953

PR #7071, opened 2017-10-30
Fix for #7058 ...

davidfestal:
### What does this PR do?

It makes the new `always-external-custom` server evaluation strategy
also support `https`. The previous PR #7004 to fix issue #6953 was not
complete and didn't support the secured use-case correctly.

#7058

### DIFF
--- a/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/AlwaysExternalCustomServerEvaluationStrategy.java
+++ b/plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/AlwaysExternalCustomServerEvaluationStrategy.java
@@ -14,6 +14,7 @@ import com.google.inject.Inject;
 import com.google.inject.name.Named;
 import java.util.Map;
 import org.eclipse.che.api.machine.server.model.impl.ServerImpl;
+import org.eclipse.che.api.machine.server.model.impl.ServerPropertiesImpl;
 import org.eclipse.che.commons.annotation.Nullable;
 import org.eclipse.che.plugin.docker.client.WorkspacesRoutingSuffixProvider;
 import org.eclipse.che.plugin.docker.client.json.ContainerInfo;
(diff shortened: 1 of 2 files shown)


plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/AlwaysExternalCustomServerEvaluationStrategy.java
```diff
@@ -49,11 +53,26 @@ public AlwaysExternalCustomServerEvaluationStrategy(
...
     return super.getExternalAddressesAndPorts(containerInfo, internalHost);
   }
+
+  @Override
+  protected ServerImpl updateServer(ServerImpl server) {
+    super.updateServer(server);
+    if (!Strings.isNullOrEmpty(cheDockerCustomExternalProtocol)) {
+    ServerPropertiesImpl properties = (ServerPropertiesImpl) server.getProperties();
+    String url = properties.getInternalUrl();
+    if(url != null && url.startsWith("http:")) {
```
ibuziuk: @davidfestal formatting in `if(`.  Probably it would be better to extract it to some method e.g. `isHttp(String url)`
Could you please clarify what exactly is done in `updateServer` and why internal URL is updated only when it is http based ?

plugins/plugin-docker/che-plugin-docker-machine/src/main/java/org/eclipse/che/plugin/docker/machine/AlwaysExternalCustomServerEvaluationStrategy.java
```diff
@@ -56,4 +57,19 @@ public AlwaysExternalCustomServerEvaluationStrategy(
       ContainerInfo containerInfo, String internalHost) {
     return super.getExternalAddressesAndPorts(containerInfo, internalHost);
   }
+
+  @Override
+  protected ServerImpl updateServer(ServerImpl server) {
+    super.updateServer(server);
+    ServerPropertiesImpl properties = (ServerPropertiesImpl) server.getProperties();
```
ibuziuk: +1 to merge, but I would probably extract all the logic to a separate method with meaningful name  / javadoc e.g.`updateProtocolForInternalUrl`

ibuziuk (approved):

garagatyi (approved):

Issue #6368, opened 2017-09-20
jdt.ls Use one LS per workspace


tsmaeder:
We want to use only one instance of jdt.ls per workspace. We need have proxy language servers per project to work with the current 1 LS/project strategy in Che.

tsmaeder:
Committed to branch

PR #7083, opened 2017-10-30
Allow for per workspace/per project launching strategies

tsmaeder:
This PR introduces launching strategies for language servers. Currently, a 1-server-per-workspace and a 1-server-per-project strategy are implemented.

This is an improved fix for https://github.com/eclipse/che/issues/6368

#### Release Notes
Support for single language server per workspace.


wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/launcher/LanguageServerLauncherTemplate.java
```diff
@@ -1,12 +1,9 @@
 /*
- * Copyright (c) 2012-2017 Red Hat, Inc.
- * All rights reserved. This program and the accompanying materials
- * are made available under the terms of the Eclipse Public License v1.0
- * which accompanies this distribution, and is available at
- * http://www.eclipse.org/legal/epl-v10.html
+ * Copyright (c) 2012-2017 Red Hat, Inc. All rights reserved. This program and the accompanying
```
benoitf, line 2: I think it will break the formatter check

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/registry/LanguageServerRegistryImpl.java
```diff
@@ -52,31 +75,47 @@
...
+  private final EventService eventService;
+  private final CheLanguageClientFactory clientFactory;
+
+  private static ClientCapabilities CLIENT_CAPABILITIES = initClientCapabilities();
+  private static final int PROCESS_ID = getProcessId();
+
+  private final List<ServerInitializerObserver> observers = new ArrayList<>();
+
+  private static int getProcessId() {
+    String name = ManagementFactory.getRuntimeMXBean().getName();
```
benoitf, line 91: it looks like very plaftorm/OS dependent to grab a process id so I'm not sure it's a good choice but it was there previously. we may add a TODO change as with DefaultProcessManager.java and CLibrary or with Java9 the getPid() is coming as well

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/registry/LanguageServerRegistryImpl.java
```diff
@@ -109,56 +148,83 @@ public ServerCapabilities getCapabilities(String fileUri) throws LanguageServerE
...
+                      LOG.info(
+                          "Initialized language server {} on thread {} and fileUri {}",
+                          launcherId,
+                          threadId,
+                          fileUri);
+                      fireServerInitialized(launcher, server, res.getCapabilities(), fileUri);
+                      synchronized (initializedServers) {
+                        initializedServers.add(
+                            new InitializedLanguageServer(id, server, res, launcher, key));
+                        LOG.info("launched for  " + thread + ": " + key);
```
benoitf: log is info ? pls use "{}" , format

benoitf (approved):
I like the strategies

Maybe we could drop the "Per" prefix
ProjectLaunchingSrategy and WorkspaceLaunchingStrategy

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/registry/LanguageServerRegistryImpl.java
tsmaeder: I agree that this looks strange, but I just movee this code, so I'd like to leave it as is in this PR.

benoitf: yes

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/launcher/LanguageServerLauncherTemplate.java
tsmaeder: I run mvn fmt:format on it and the file is not reformatted.

PR #7103, opened 2017-10-31
Remove outdated package old from Docker infra

garagatyi:
### What does this PR do?
Remove outdated package old from Docker infra


benoitf:
how is managed single port if you remove that ?

garagatyi:
@benoitf in case of OpenShift infra OpenShift implements single port functionality. In case of Docker infra it is not available in Che6 at the moment

benoitf:
@garagatyi I was referencing docker infra. Wouldn't it better to keep code until single port is there ? Also why it's not already there ?

garagatyi:
> Wouldn't it better to keep code until single port is there ?

Single port doesn't work right now and because of massive refactoring code removed in this PR is not compatible with the SPI. So we would have to rewrite it completely to recover single port feature. So I see no reason to keep this code while it is unused and can not be used directly later.

> Also why it's not already there ?

When we started SPI we broke a lot of the components in the system. Then we recovered parts one  by one. There are couple of components/features that are not recovered yet and single port is one of them

sleshchenko (approved):

skabashnyuk (approved):

akorneta (approved):

PR #7123, opened 2017-11-01
Cleanup and simplify script that deploys Che to OpenShift

garagatyi:
### What does this PR do?
We had a lot of issues because of complex logic in deploy script before. Each time someone changed it to fix something in one flavor of OpenShift he had to fix other flavors too, but it didn't happen. So we had different branches of if clauses where something was done for each flavor. 
This PR cleans up this and tries to unify code to have single flow for all the flavors. 
With this PR we have different areas:
- where defaults for each flavor are set
- where variables common for all flavors are  set (may depend on flavor dependent defaults)
- where Che deployment YAML file is customized with sed


dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
```diff
@@ -154,71 +105,132 @@ if [ "${OPENSHIFT_FLAVOR}" == "minishift" ]; then
...
+OPENSHIFT_ENDPOINT=${OPENSHIFT_ENDPOINT:-${DEFAULT_OPENSHIFT_ENDPOINT}}
+OPENSHIFT_USERNAME=${OPENSHIFT_USERNAME:-${DEFAULT_OPENSHIFT_USERNAME}}
+OPENSHIFT_PASSWORD=${OPENSHIFT_PASSWORD:-${DEFAULT_OPENSHIFT_PASSWORD}}
+CHE_OPENSHIFT_PROJECT=${CHE_OPENSHIFT_PROJECT:-${DEFAULT_CHE_OPENSHIFT_PROJECT}}
+CHE_KEYCLOAK_DISABLED=${CHE_KEYCLOAK_DISABLED:-${DEFAULT_CHE_KEYCLOAK_DISABLED}}
+CHE_DEBUGGING_ENABLED=${CHE_DEBUGGING_ENABLED:-${DEFAULT_CHE_DEBUGGING_ENABLED}}
+OC_SKIP_TLS=${OC_SKIP_TLS:-${DEFAULT_OC_SKIP_TLS}}
+CHE_APPLY_RESOURCE_QUOTAS=${CHE_APPLY_RESOURCE_QUOTAS:-${DEFAULT_CHE_APPLY_RESOURCE_QUOTAS}}
+IMAGE_PULL_POLICY=${IMAGE_PULL_POLICY:-${DEFAULT_IMAGE_PULL_POLICY}}
+DEFAULT_CHE_INFRA_OPENSHIFT_PVC_STRATEGY="common"
```
sleshchenko, line 209: What do you think about grouping setting of common defaults?.

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
```diff
@@ -389,73 +399,18 @@ MULTI_USER_REPLACEMENT_STRING="          - name: \"CHE_WORKSPACE_LOGS\"
...
-  CHE_DEPLOYMENT_FILE_PATH=${CHE_DEPLOYMENT_FILE_PATH:-${DEFAULT_CHE_DEPLOYMENT_FILE_PATH}}
-
-cat "${CHE_DEPLOYMENT_FILE_PATH}" | \
-    if [ ! -z "${OPENSHIFT_NAMESPACE_URL+x}" ]; then sed "s/    hostname-http:.*/    hostname-http: ${OPENSHIFT_NAMESPACE_URL}/" ; else cat -; fi | \
-    sed "s/          image:.*/          image: \"${CHE_IMAGE_SANITIZED}\"/" | \
-    sed "s/          imagePullPolicy:.*/          imagePullPolicy: \"${IMAGE_PULL_POLICY}\"/" | \
-    sed "s|    keycloak-oso-endpoint:.*|    keycloak-oso-endpoint: ${KEYCLOAK_OSO_ENDPOINT}|" | \
-    sed "s|    keycloak-github-endpoint:.*|    keycloak-github-endpoint: ${KEYCLOAK_GITHUB_ENDPOINT}|" | \
-    sed "s/    keycloak-disabled:.*/    keycloak-disabled: \"${CHE_KEYCLOAK_DISABLED}\"/" | \
+    if [ ! -z "${WORKSPACE_MEMORY_LIMIT+x}" ]; then sed "s/    workspaces-memory-limit: 2300Mi/    workspaces-memory-limit: ${WORKSPACE_MEMORY_LIMIT}/" ; else cat -; fi | \
```
sleshchenko: WDYT about rewriting all `sed`s command as the `sed` command for settings of the CHE_INFRA_OPENSHIFT_PASSWORD variables. I mean using of `:.*` instead of hardcoding default value that should be present in yml file. Because now script will work incorrectly if default value in yml file is changes.

sleshchenko (approved):
Thanks for good refactoring. Please take a look my proposals.

riuvshin:
that makes this script much more readable and easy to understand. But to merge it we have to test all the usecases

garagatyi:
@benoitf @l0rd @davidfestal can you review my PR?

benoitf (approved):
I've noticed a warning from shellcheck at

Line 435:
  ${COMMAND_DIR}/multi-user/configure_and_start_keycloak.sh
  ^-- SC2086: Double quote to prevent globbing and word splitting.


it's not related to this PR but I think that bats tests should be added as it's become tricker to handle all usecases only by reviewing the script.

l0rd (approved):
Cool!

garagatyi:
@sleshchenko I would do that next time.
@benoitf fixed.

riuvshin (approved):

garagatyi:
@benoitf I agree that tests would help a lot in keeping deployment to different flavors save

Issue #5579, opened 2017-07-10
Reject machine configs with cyclic dependencies of installers


akorneta:
Since we can store workspace config with a machine which contains cyclic dependencies between installers(without the ability to run it [log](https://user-images.githubusercontent.com/12269156/28018343-95508a2a-6585-11e7-9533-b20f3dec8b81.png)) we need to consider reworking the [logic](https://github.com/eclipse/che/blob/9d7a2d998c45ae2083499945a55525edadee9151/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/InternalMachineConfig.java#L79-L79) of installers dependency resolving and move it into workspace validator.

Issue #6274, opened 2017-09-14
[CHE6] Check machine entries in workspace config on estimation in OpenShift infra implementation


garagatyi:
Related to SPI branch only.
Now OpenShift infrastructure doesn't check that machines in workspace config are present in OpenShift recipe.
Right now we get NPE in certain cases because of lack of this check.

sleshchenko:
Will be fixed in https://github.com/eclipse/che/pull/7126

PR #7126, opened 2017-11-01
Moved logic of Environment estimation from RuntimeInfrastructure to InternalEnvironmentFactory

gazarenkov:
### What does this PR do?
Changes in Runtime Infrastructure SPI
Move logic of Environment estimation (producing InternalEnvironment) from RuntimeInfrastructure to InternalEnvironmentFactory

To provide clear way for associate different Recipes to different Runtime Environment (such as "dockerimage" recipe with "openshift" infrastructure)

### Changes guide
Here is the short guide how it is will work for more details see commits one by one.

Workspace Runtimes class creates a valid InternalRuntime instance with InternalRuntimeFactory#create. See [WorkspaceRuntime](https://github.com/eclipse/che/pull/7126/files#diff-40d98dfd00ecb1740ab9e35b6c89a5e9)

Abstract InternalEnvironmentFactory perform recipe downloading, retrieving information about configured installers, invoke validation of machine configs implemented in MachineConfigsValidator. See [InternalEnvironmentFactory](https://github.com/eclipse/che/pull/7126/files#diff-1a2c6bbf294c58325e591454a4a928cd)

MachineConfigsValidator checks that at least one machine is configured, that machines names, services names, ports and protocols are valid and that workspace agent is present. See [InternalEnvironmentValidator](https://github.com/eclipse/che/pull/7126/files#diff-546befd67ab2d5bd876fa9fa762b2132)

Compose, DockerImage, Dockerfile and OpenShift InternalEnvironments are POJO objects that add environment specific data to `InternalEnvironment`. See [ComposeEnvironment](https://github.com/eclipse/che/pull/7126/files#diff-02991805abcff0551a9983e3711e638a), [DockerImageEnvironment](https://github.com/eclipse/che/pull/7126/files#diff-d1927f9955a9ad5e97a2af326e8ffc77), [DockerfileEnvironment](https://github.com/eclipse/che/pull/7126/files#diff-fb221b781dc214295fc4b301db77c1ca), [OpenShiftEnvironment](https://github.com/eclipse/che/pull/7126/files#diff-300a71eee44903fdf26da12a8f23bc7a).

Each implementation of `InternalEnvironment` has own EnvironmentFactory which should parse, validate recipe and return valid specific internal environment. See the corresponding EnvironmentFactories.

Then the implementation of RuntimeInfrastructure will check which type of InternalRuntime, cast to the corresponding type, convert it to a supported one if it is needed and use it. 
See:
- for docker: [DockerRuntimeInfrastructure](https://github.com/eclipse/che/pull/7126/files#diff-6bea78ae0d9957eb2ab6d6657db40134)
 and [DockerEnvironmentConverter](https://github.com/eclipse/che/pull/7126/files#diff-587a65c2682419693ef3b151ce8e4aee), [ComposeEnvironmentConverter](https://github.com/eclipse/che/pull/7126/files#diff-557b2d466368dc025c6c1e9cf4e7120a) etc,
- for OpenShift: [OpenShiftInfrastructure](https://github.com/eclipse/che/pull/7126/files#diff-af3fecda517f34f356a133111f159465)

It also fixes https://github.com/eclipse/che/issues/5579 and https://github.com/eclipse/che/issues/6274
#### Docs PR
To introduce new type of Recipe you have to implement 
subclass of 

`org.eclipse.che.api.workspace.server.spi.InternalEnvironment and org.eclipse.che.api.workspace.server.spi.InternalEnvironmentFactory `

for creating this InternalEnvironment.
To make it accessible for RuntimeInfrastructures bound the factory via MapBinder with unique equal to Recipe Type ID like:
```
MapBinder<String, InternalEnvironmentFactory> environmentFactories =
        MapBinder.newMapBinder(binder(), String.class, InternalEnvironmentFactory.class);
environmentFactories.addBinding("my-recipe-type-ID").to(MyInternalEnvironmentFactory.class); 
 
```


infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenShiftInternalEnvironment.java
```diff
@@ -0,0 +1,43 @@
...
+import org.eclipse.che.api.installer.server.InstallerRegistry;
+import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
+import org.eclipse.che.api.workspace.server.spi.InternalEnvironment;
+import org.eclipse.che.api.workspace.server.spi.RecipeRetriever;
+
+/** @author Sergii Leshchenko */
+public class OpenShiftInternalEnvironment extends InternalEnvironment {
+
+  private final OpenShiftEnvironmentParser openShiftEnvironmentParser;
+  private OpenShiftEnvironment osEnv;
```
sleshchenko: I don't see any reason to have both `OpenShiftInternalEnvironment` and `OpenShiftEnvironment`. I think `OpenShiftInternalEnvironment` should be POJO object and I propose the following changes
```
public class OpenShiftInternalEnvironment extends InternalEnvironment {
  private final Map<String, Pod> pods;
  private final Map<String, Service> services;
  private final Map<String, Route> routes;
  private final Map<String, PersistentVolumeClaim> persistentVolumeClaims;

  // getters are here
}
```
Then OpenShiftInternalEnvironmentFactory can parse it with OpenShiftEnvironmentParser.
WDYT?

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenShiftInternalEnvironment.java
```diff
@@ -0,0 +1,43 @@
...
+package org.eclipse.che.workspace.infrastructure.openshift.environment;
+
+import org.eclipse.che.api.core.ValidationException;
+import org.eclipse.che.api.core.model.workspace.config.Environment;
+import org.eclipse.che.api.installer.server.InstallerRegistry;
+import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
+import org.eclipse.che.api.workspace.server.spi.InternalEnvironment;
+import org.eclipse.che.api.workspace.server.spi.RecipeRetriever;
+
+/** @author Sergii Leshchenko */
```
sleshchenko: I think, My authorship here is a copy-paste error, isn't it?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/RuntimeInfrastructure.java
```diff
@@ -83,55 +57,6 @@ public final EventService getEventService() {
...
-   * with {@link Beta} annotation to hint that. Do not override this method. <br>
-   * Workaround should be removed after resolution of https://github.com/eclipse/che/issues/6006
-   *
-   * @param environment incoming Environment to estimate
-   * @return calculated internal environment.
-   * @throws ValidationException if incoming Environment is not valid
-   * @throws InfrastructureException if any other error occurred
-   */
-  @Beta
-  public InternalEnvironment estimate(Environment environment)
```
sleshchenko: There is no any abstract logic in this class anymore. I'd change abstract class here to interface.

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/compose/ComposeInternalEnvironment.java
```diff
@@ -0,0 +1,142 @@
...
+import org.eclipse.che.workspace.infrastructure.docker.environment.compose.model.ComposeService;
+import org.eclipse.che.workspace.infrastructure.docker.model.DockerBuildContext;
+import org.eclipse.che.workspace.infrastructure.docker.model.DockerContainerConfig;
+import org.eclipse.che.workspace.infrastructure.docker.model.DockerEnvironment;
+
+/**
+ * @author Alexander Garagatyi
+ * @author Alexander Andrienko
+ */
+public class ComposeInternalEnvironment extends InternalEnvironment {
```
garagatyi: I think that main idea of this PR is to provide an ability to use parsed environments with different infrastructures. Here compose environment factory produces Docker infra's own DockerEnvironment instead of ComposeEnvironment which is native to the environment of type 'compose'. Can you elaborate on how this fits into redesigned SPI?

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/compose/ComposeInternalEnvironmentFactory.java
```diff
@@ -0,0 +1,39 @@
...
+  @Inject
+  public ComposeInternalEnvironmentFactory(
+      InstallerRegistry installerRegistry, RecipeRetriever recipeRetriever) {
+    super(installerRegistry, recipeRetriever);
+  }
+
+  @Override
+  public InternalEnvironment create(Environment environment)
+      throws InfrastructureException, ValidationException {
+    return new ComposeInternalEnvironment(environment, installerRegistry, recipeRetriever);
```
garagatyi: Please, consider putting business logic here instead of the constructor of `ComposeInternalEnvironment`. I believe that complex logic in the constructor (possibly with private methods used in it) may lead to complex bugs because of usage of incomplete state of the object.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -470,7 +475,8 @@ void recoverOne(RuntimeInfrastructure infra, RuntimeIdentity identity) throws Se
 
     InternalRuntime runtime;
     try {
-      InternalEnvironment internalEnvironment = infra.estimate(environment);
+      InternalEnvironment internalEnvironment =
```
garagatyi: Looks like previously created `InternalEnvironment` should be used instead of the creation of the new one.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/RuntimeInfrastructure.java
```diff
@@ -149,79 +74,14 @@ protected abstract void internalEstimate(InternalEnvironment env)
...
    * @param id the RuntimeIdentity
    * @param environment incoming internal environment
    * @return new RuntimeContext object
    * @throws ValidationException if incoming environment is not valid
    * @throws InfrastructureException if any other error occurred
    */
-  public RuntimeContext prepare(RuntimeIdentity id, InternalEnvironment environment)
-      throws ValidationException, InfrastructureException {
-    for (InternalEnvironmentProvisioner provisioner : internalEnvironmentProvisioners) {
-      provisioner.provision(id, environment);
```
garagatyi: Can you elaborate on where this code is performed in this PR (I can't find it)?

garagatyi (commented):
I think that the idea of this PR is clear. I see how it prevents duplication of code for different infrastructures. Can you elaborate on places I commented?

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenShiftInternalEnvironment.java
gazarenkov: Agree, moreover, I think OpenShiftEnvironmentParser's functionality can be a part of OpenShiftInternalEnvironmentFactory?

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenShiftInternalEnvironment.java
gazarenkov: Nope, I think it will have even more sense when we refactor it according to your comment above :) , my role here is just restructuring according to the new model, all of real functionality yours.

sleshchenko: OK :smile:

.mvn/jvm.config
```diff
@@ -1 +0,0 @@
--Xmx2048m -Xms1024m -Djava.awt.headless=true -Djava.net.preferIPv4Stack=true -Dgwt.compiler.localWorkers=1
```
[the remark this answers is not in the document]
gazarenkov: yes, you are right, sorry for that... can not imagine how happened, repairing

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/compose/ComposeInternalEnvironment.java
gazarenkov: This reused of what you initially did (to not to produce a lot of changes at ones and make it discussion happen). I agree with you it seems no needs to make the extra step of transforming ComposeEnv to DockerEnv

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
gazarenkov: Not sure I understand it, could you please rephrase

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -130,7 +134,7 @@ public void validate(Environment environment)
       throw new NotFoundException("Infrastructure not found for type: " + type);
     }
 
-    infraByRecipe.get(type).estimate(environment);
+    environmentFactories.get(environment.getRecipe().getType()).create(environment);
```
sleshchenko: Potentially NPE can be thrown here. It's needed to check that the corresponding environment parser is present before using it.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/RuntimeInfrastructure.java
```diff
@@ -37,31 +26,16 @@
...
-  public RuntimeInfrastructure(
-      String name,
-      Collection<String> types,
-      EventService eventService,
-      InstallerRegistry installerRegistry,
-      RecipeRetriever recipeRetriever,
-      Set<InternalEnvironmentProvisioner> internalEnvironmentProvisioners,
-      ServersNormalizer serversNormalizer) {
-    this.serversNormalizer = serversNormalizer;
+  public RuntimeInfrastructure(String name, Collection<String> types, EventService eventService) {
```
sleshchenko: `eventService` is unused. Can we remove it?

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerRuntimeInfrastructure.java
```diff
@@ -52,70 +45,68 @@
...
-      // move image from location to content
-      EnvironmentImpl envCopy = new EnvironmentImpl(environment);
-      envCopy.getRecipe().setContent(environment.getRecipe().getLocation());
-      envCopy.getRecipe().setLocation(null);
-      return super.estimate(envCopy);
-    }
-    return super.estimate(environment);
-  }
+  //  @Override
+  //  public InternalEnvironment estimate(Environment environment)
```
sleshchenko: Let's remove commented code

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerRuntimeInfrastructure.java
```diff
@@ -52,70 +45,68 @@
...
-        environmentParsers.keySet(),
-        eventService,
-        installerRegistry,
-        recipeRetriever,
-        internalEnvironmentProvisioners,
-        serversNormalizer);
-    this.dockerEnvironmentValidator = dockerEnvironmentValidator;
-    this.dockerEnvironmentParser = dockerEnvironmentParser;
+      EventService eventService) {
+    super("docker", environmentParsers.keySet(), eventService);
```
sleshchenko: I don't think that `environmentParsers.keySet()` should be here. Actually, it supports only `dockerfile`, `dockerimage`, `compose`:
```
 switch (type) {
      case "dockerfile":
          ...
      case "dockerimage":
         ...
      case "compose":
        ...
      default:
        throw new InfrastructureException("Recipe type is not allowed " + type);
```
It's better to use ImmutableMap of supported recipes types

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/DockerEnvironmentTypeModule.java
```diff
@@ -12,24 +12,20 @@
...
-    envParserMapBinder.addBinding("compose").to(ComposeEnvironmentParser.class);
-    envParserMapBinder
-        .addBinding(DockerfileEnvironmentParser.TYPE)
-        .to(DockerfileEnvironmentParser.class);
-    envParserMapBinder
-        .addBinding(DockerImageEnvironmentParser.TYPE)
-        .to(DockerImageEnvironmentParser.class);
+    MapBinder<String, InternalEnvironmentFactory> environmentFactories =
+        MapBinder.newMapBinder(binder(), String.class, InternalEnvironmentFactory.class);
+    environmentFactories.addBinding("compose").to(ComposeInternalEnvironmentFactory.class);
```
sleshchenko: can we move `compose`, `dockerfile`, `dockerimage` somewhere to constant fields?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/InternalEnvironmentFactory.java
```diff
@@ -0,0 +1,129 @@
...
+   */
+  public final InternalEnvironment create(final Environment environment)
+      throws InfrastructureException, ValidationException {
+
+    Map<String, InternalMachineConfig> machines = new HashMap<>();
+    List<Warning> warnings = new ArrayList<>();
+
+    // Workaround related to written docker image id into location instead of content
+    EnvironmentImpl envCopy = new EnvironmentImpl(environment);
+    if ("dockerimage".equals(environment.getRecipe().getType())
```
sleshchenko: I think this logic should be moved to `DockerimageInternalEnvironmentFactory`. But to do it this method must not be final. I mean the following approach:
```
public class DockerimageInternalEnvironmentFactory {
 public InternalEnvironment create(final Environment environment) {
    if (environment.getRecipe().getLocation() != null) {
      EnvironmentImpl envCopy = new EnvironmentImpl(environment);
      // move image from location to content
      envCopy.getRecipe().setContent(environment.getRecipe().getLocation());
      envCopy.getRecipe().setLocation(null);
      return super(envCopy);
    }
    return super(environment);
 }
}
```

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
garagatyi: Nevermind - I missed lines collapsed by Github and misunderstood your changes because of that. Sorry

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/DockerEnvironmentTypeModule.java
garagatyi: +1

infrastructures/openshift/src/test/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenShiftEnvironmentParserTest.java
```diff
@@ -48,48 +50,53 @@
 import org.testng.annotations.Test;
 
 /**
- * Tests {@link OpenShiftEnvironmentParser}.
+ * Tests {@link OpenShiftEnvironmentFactory}.
  *
  * @author Anton Korneta
  */
 public class OpenShiftEnvironmentParserTest {
```
sleshchenko: Let's rename test class according to the renaming of class that is tested here. 
`OpenShiftEnvironmentParserTest` -> `OpenShiftEnvironmentFactoryTest`

infrastructures/openshift/src/test/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenShiftEnvironmentParserTest.java
```diff
@@ -48,48 +50,53 @@
...
+    when(recipe.getContent()).thenReturn("recipe content");
   }
 
   @Test
   public void ignoreRoutesWhenRecipeContainsThem() throws Exception {
     final List<HasMetadata> objects = asList(new Route(), new Route());
     when(validatedObjects.getItems()).thenReturn(objects);
 
-    final OpenShiftEnvironment parsed = osEnvironmentParser.parse(internalEnvironment);
+    parsed = (OpenShiftInternalEnvironment) osEnvironmentParser.create(environment);
```
sleshchenko: I think `#create(Map<String, InternalMachineConfig>, InternalRecipe, List<Warning>)` should be used here. As `#create(InternalEnvironment` is a functional part of AbstractInternalEnvironmentParser and should be tested in the corresponding test and should be used while implementation testing

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/dockerimage/DockerimageInternalEnvironment.java
```diff
@@ -0,0 +1,44 @@
...
+import org.eclipse.che.api.core.ValidationException;
+import org.eclipse.che.api.core.model.workspace.Warning;
+import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
+import org.eclipse.che.api.workspace.server.spi.InternalEnvironment;
+import org.eclipse.che.api.workspace.server.spi.InternalMachineConfig;
+import org.eclipse.che.api.workspace.server.spi.InternalRecipe;
+import org.eclipse.che.workspace.infrastructure.docker.model.DockerEnvironment;
+
+/**
+ * @author Alexander Garagatyi
```
garagatyi: Can you change authors - current ones are definitely not the real one

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/MachineNormalizer.java
```diff
@@ -0,0 +1,63 @@
...
+
+  private MachineNormalizer() {}
+
+  public static void normalizeMachine(
+      String name, DockerContainerConfig container, InternalMachineConfig machineConfig)
+      throws ValidationException {
+
+    if (machineConfig.getAttributes().containsKey("memoryLimitBytes")) {
+      try {
+        container.setMemLimit(
```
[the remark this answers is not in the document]
sleshchenko: Sounds good to me

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/InternalEnvironmentFactory.java
gazarenkov: wanted to have create(final Environment environment) final but OK, will change, not a big deal to me

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/RuntimeInfrastructure.java
gazarenkov: Let's discuss it separately, not related to this PR

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerRuntimeInfrastructure.java
gazarenkov: done

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
gazarenkov: sure, fixed

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/DockerEnvironmentTypeModule.java
gazarenkov: ok

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/MachineNormalizer.java
```diff
@@ -0,0 +1,63 @@
...
+ * Utility for putting stuffs to machine
+ *
+ * @author Alexander Garagatyi
+ * @author Alexander Andrienko
+ */
+public class MachineNormalizer {
+
+  private MachineNormalizer() {}
+
+  public static void normalizeMachine(
```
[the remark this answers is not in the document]
gazarenkov: Sure, to me it looks like pure utility method. 
I think we should not overuse DI for classes which hardly will have other implementation to switch to. IMO it just makes code overloaded with huge number of constructor parameters for practically nothing.

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/MachineNormalizer.java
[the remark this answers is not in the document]
gazarenkov: maybe, how we use it after?

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/dockerfile/DockerfileInternalEnvironmentFactory.java
```diff
@@ -0,0 +1,96 @@
...
+
+  @Override
+  protected InternalEnvironment create(
+      Map<String, InternalMachineConfig> machines, InternalRecipe recipe, List<Warning> warnings)
+      throws InfrastructureException, ValidationException {
+
+    DockerEnvironment dockerEnvironment = dockerEnv(machines, recipe);
+    validator.validate(machines, dockerEnvironment);
+    // check that order can be resolved
+    startStrategy.order(dockerEnvironment);
```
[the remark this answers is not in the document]
gazarenkov: removed
actually, I would consider not to use "unified" DockerEnvironment (withmulticintainer and lots of fields inside) object as a data holder in such a simple case.

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/dockerimage/DockerimageInternalEnvironmentFactory.java
```diff
@@ -0,0 +1,77 @@
...
+  private DockerEnvironment dockerEnv(
+      Map<String, InternalMachineConfig> machines, InternalRecipe recipe)
+      throws ValidationException {
+
+    Map.Entry<String, InternalMachineConfig> entry = machines.entrySet().iterator().next();
+    DockerEnvironment dockerEnv = new DockerEnvironment();
+    DockerContainerConfig container = new DockerContainerConfig();
+    dockerEnv.getContainers().put(entry.getKey(), container);
+    container.setImage(recipe.getContent());
+    MachineNormalizer.normalizeMachine(entry.getKey(), container, entry.getValue());
```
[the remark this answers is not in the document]
gazarenkov: what do you mean?

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/MachineNormalizer.java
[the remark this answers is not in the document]
sleshchenko: Using of static methods which contain complicated logic makes testing so hard. We aren't able to stub invocation of static methods and you should prepare all that is needed for method execution without exception in each test where this method is used. So, you should prepare objects in tests for `DockerfileInternalEnvironmentFactory`, `DockerimageInternalEnvironmentFactory`.

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/MachineNormalizer.java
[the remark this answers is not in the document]
garagatyi: it is used as a memory limit for a machine

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/dockerfile/DockerfileInternalEnvironmentFactory.java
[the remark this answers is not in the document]
garagatyi: Looks like we already agreed on that [here](https://github.com/eclipse/che/pull/7126#discussion_r148524576). Are you going to do that in this PR or leave it as is for now?

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/dockerimage/DockerimageInternalEnvironmentFactory.java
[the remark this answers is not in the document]
garagatyi: I mean that I don't see call of `MachineNormalizer.normalizeMachine` in `ComposeInternalEnvironmentFactory`. I'm not sure whether I missed something or it should be added there

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/MachineNormalizer.java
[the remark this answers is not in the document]
gazarenkov: well, it is always a matter of compromise...there are no black/white decision
I am answering on question :)

for me, clear structure and readability prevales over being able to use the favorite testing framework and I would disagree that mocking always mean easier testing code.

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/MachineNormalizer.java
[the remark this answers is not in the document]
gazarenkov: ok for me

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/MachineNormalizer.java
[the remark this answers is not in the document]
benoitf: I would say it's not about using a favorite testing framework, it's the way of having extensibility/freedom.

If the only worry is about constructor parameters, you can use field injection or setter injection to not overload constructor parameters as well. (Using @InjectMocks is handling that when testing)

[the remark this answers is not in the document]
gazarenkov: Do you have an example of what kind of extensibility we can expect from the method under discussion? (for my understanding, maybe I underestimate its value, that is not my code inside)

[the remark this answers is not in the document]
benoitf: well, you may not know how your code will be used as part of an open source. People may want to extend some classes, replace some logical stuff, etc. 
So, following extensibility model (class instead of static code) you ease the derivative work that may be done
Sometimes it's surprising to see what people do as workaround only because upstream project is not so extensible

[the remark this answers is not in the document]
gazarenkov: I do not think it is a good place to discuss it in general, this particular case shows that there are nothing to extend normally.

Anyway I removed this class at all, reworking a bit Docker/Compose config instead

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -129,8 +133,7 @@ public void validate(Environment environment)
     if (!infraByRecipe.containsKey(type)) {
       throw new NotFoundException("Infrastructure not found for type: " + type);
     }
-
-    infraByRecipe.get(type).estimate(environment);
+    createInternalEnvironment(environment);
```
garagatyi: Can you add comment that explains that purpose of this call is validation which is performed in it

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/environment/InternalEnvironmentFactory.java
```diff
@@ -0,0 +1,135 @@
...
+      InstallerRegistry installerRegistry,
+      RecipeRetriever recipeRetriever,
+      InternalEnvironmentValidator<T> envValidator) {
+    this.installerRegistry = installerRegistry;
+    this.recipeRetriever = recipeRetriever;
+    this.envValidator = envValidator;
+  }
+
+  /**
+   * Creates a valid instance of InternalEnvironment.
```
garagatyi: Can you add more details on what is performed inside of this method? Looks like it is much more than just an instance creation

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/environment/InternalEnvironmentFactory.java
```diff
@@ -0,0 +1,135 @@
...
+   * @return InternalEnvironment
+   * @throws InfrastructureException if infrastructure specific error occurs
+   * @throws ValidationException if validation fails
+   */
+  protected abstract T doCreate(
+      InternalRecipe recipe, Map<String, InternalMachineConfig> machines, List<Warning> warnings)
+      throws InfrastructureException, ValidationException;
+
+  @VisibleForTesting
+  Map<String, ServerConfig> normalizeServers(Map<String, ? extends ServerConfig> servers) {
```
garagatyi, line 129: This method might be useful for concrete impls of this class, so I would make it protected.

garagatyi: Or method `normalizeServer`

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/dockerfile/DockerfileEnvironment.java
```diff
@@ -0,0 +1,74 @@
...
+import org.eclipse.che.api.core.model.workspace.Warning;
+import org.eclipse.che.api.workspace.server.spi.environment.InternalEnvironment;
+import org.eclipse.che.api.workspace.server.spi.environment.InternalMachineConfig;
+import org.eclipse.che.api.workspace.server.spi.environment.InternalRecipe;
+
+/** @author Sergii Leshchenko */
+public class DockerfileEnvironment extends InternalEnvironment {
+  public static final String TYPE = "dockerfile";
+
+  private final String dockerfile;
```
garagatyi: Is it the content of a Dockerfile? If so, please, specify it in comment or the name

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/compose/ComposeEnvironment.java
```diff
@@ -0,0 +1,92 @@
...
+    this.version = version;
+
+    this.services = services;
+  }
+
+  public String getVersion() {
+    return version;
+  }
+
+  /** Returns ordered services. */
```
garagatyi, line 46: I'm afraid that it might be not enough to mention order in the javadocs. Please, consider returning LinkedHashMap instead of Map

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/compose/ComposeEnvironmentFactory.java
```diff
@@ -0,0 +1,113 @@
...
+      throw new ValidationException(
+          format(
+              "Provided environment recipe content type '%s' is "
+                  + "unsupported. Supported values are: %s",
+              contentType, YAML_CONTENT_TYPES.stream().collect(joining(", "))));
+    }
+    ComposeRecipe composeRecipe = doParse(recipeContent);
+
+    // check that containers start order can be resolved
+    // NOTE: it should be performed before dockerEnvironmentNormalizer.normalize because
```
garagatyi: Looks like this comment is irrelevant now. WDYT?

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenShiftEnvironmentValidator.java
```diff
@@ -0,0 +1,54 @@
...
+        for (Container container : pod.getSpec().getContainers()) {
+          missingMachines.remove(Names.machineName(pod, container));
+        }
+      }
+    }
+    checkArgument(
+        missingMachines.isEmpty(),
+        "Environment contains machines that are missing in recipe: %s",
+        Joiner.on(", ").join(missingMachines));
+    // TODO Implement validation OpenShift objects https://github.com/eclipse/che/issues/7381
```
garagatyi, line 51: If there an issue for that do we need this todo?

sleshchenko: Why not? It shows when this part will be implemented

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/compose/ComposeEnvironmentFactory.java
sleshchenko: You're right. I'll remove this comment

garagatyi (approved):
Thank you for such a detailed explanation of what this PR does and what are the main parts. I believe it speeded up my review several times at least! This PR will be an example for me on how to write a PR description.
And PR itself looks good! A lot of simplifications. I might miss something but it is unavoidable for such a big PR.

gazarenkov (commented):
I do not think we need InternalEnvironmentValidator abstraction. It is not expected to have "invalid" InternalEnvironment object so all the validation should be performed in factory create() method. It is up to implementor when and how to do validation (before or after construction InternalEnvironment). Making it as a standalone "phase" just adds confusion and difficulties.

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/convert/ComposeEnvironmentConverter.java
```diff
@@ -0,0 +1,75 @@
...
+import org.eclipse.che.workspace.infrastructure.docker.model.DockerContainerConfig;
+import org.eclipse.che.workspace.infrastructure.docker.model.DockerEnvironment;
+
+/**
+ * Converts {@link ComposeEnvironment} to {@link DockerEnvironment}.
+ *
+ * @author Alexander Garagatyi
+ * @author Alexander Andrienko
+ */
+public class ComposeEnvironmentConverter implements DockerEnvironmentConverter {
```
gazarenkov, line 30: Why we need that? 
Make the object you need for runtime inside InternalEnvironment with factory.

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/convert/DockerEnvironmentConverter.java
```diff
@@ -0,0 +1,31 @@
...
+ * @author Alexander Andrienko
+ */
+public interface DockerEnvironmentConverter {
+
+  /**
+   * Returns {@link DockerEnvironment} that is converted from the specified {@link
+   * InternalEnvironment}.
+   *
+   * @param environment environment to converting
+   */
```
gazarenkov, line 29: Why we need that?
Make the object you need for runtime inside InternalEnvironment with factory.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/RuntimeInfrastructure.java
```diff
@@ -162,7 +94,6 @@ public RuntimeContext prepare(RuntimeIdentity id, InternalEnvironment environmen
     for (InternalEnvironmentProvisioner provisioner : internalEnvironmentProvisioners) {
```
gazarenkov, line 94: IMO it is too abstract to understand what's going on
Why not to add environment variables needed for all the machines just on creation InternalEnvironment with factory?

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenShiftEnvironmentValidator.java
garagatyi: ok

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/convert/ComposeEnvironmentConverter.java
garagatyi: The current approach doesn't link environments representations to a specific infra. Then, say, OpenShift infra can convert it how it prefers and doesn't have to convert Docker infra's internal representation into its own but instead converts recipe specific representation into its own. 
Personally, I like the current approach more.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/environment/InternalEnvironmentFactory.java
sleshchenko: We removed functionality when servers can be added depending on the recipe. So I don't think that it should be protected for now. Let's make it protected when if it will be needed.

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/compose/ComposeEnvironment.java
sleshchenko: Makes sense. I'll try

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/RuntimeInfrastructure.java
garagatyi: It also adds environment variables from installers to the machine where they are. And moves servers from installers to the machine. This is why it is a bit abstract. 
+1 to move into Factory for now. In case we decide to add estimation into workspace API later we would be willing to move this back since it is not a part of environment recipe processing.

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/convert/ComposeEnvironmentConverter.java
sleshchenko: DockerEnvironment is not a common environment that can be used by any infra because it doesn't have the corresponding recipe type.
As Oleksandr said It gives an ability for implementation for infrastructure to convert environment as it wants.
OpenShift infrastructure will support DockerImageEnvironments and to implement such functionality it requires the only value of docker image but not converted DockerImageEnvironments into DockerEnvironments with containers configs and docker specific information.
Hope I described clearly why we need DockerEnvironmentConverter inside docker infra.

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/convert/DockerEnvironmentConverter.java
sleshchenko: The same topic is discussed under the following [comment]( https://github.com/eclipse/che/pull/7126#discussion_r151604958)

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/RuntimeInfrastructure.java
sleshchenko: I agree to move it into InternalEnvironmentFactory for now

sleshchenko:
> I do not think we need InternalEnvironmentValidator abstraction. It is not expected to have "invalid" InternalEnvironment object so all the validation should be performed in factory create() method. It is up to implementor when and how to do validation (before or after construction InternalEnvironment). Making it as a standalone "phase" just adds confusion and difficulties.

@gazarenkov I added it because of it obliges internal environment factory to provide validation of parsed environment.
But I'm ok with making it up to InternalEnvironmentFactory implementor when and how to do validation. I'll rework it

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
sleshchenko: Added comment

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/dockerfile/DockerfileEnvironment.java
sleshchenko: Changed variable name and the corresponding methods to use `dockerfileContent`

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/compose/ComposeEnvironmentFactory.java
sleshchenko: Removed

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/compose/ComposeEnvironment.java
sleshchenko: Changed to returning LinkedHashMap instead of Map in `ComposeEnvironment` and `DockerEnvironment`

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/environment/InternalEnvironmentFactory.java
sleshchenko: Added describing of logic that is performed while instance creation. Please take a look if it is clear enough.

akorneta (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/RuntimeInfrastructure.java
sleshchenko: @gazarenkov @garagatyi I have a problem with such refactoring. WorkpaceIdEnvVar provisioner and MachineTokenEnvVarProvisioner need workspace id to provide the corresponding environment variable. It requires a refactoring and this PR already has a ton of changes. Are you OK to fix it separately in another issue?
P.S. I think It's OK since this PR doesn't introduce this functionality.

garagatyi:
@sleshchenko I'm ok with this approach

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/RuntimeInfrastructure.java
garagatyi: +1 to avoid adding new changes not related to this PR

sleshchenko:
@gazarenkov I've removed abstract `InternalEnvironmentValidator`. Please take a look this PR again.

sleshchenko:
ci-test

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/convert/ComposeEnvironmentConverter.java
gazarenkov: I think you missed the point. I am not talking about universal object, sure OpenShift environment will have to adopt the stuffs came from Docker Environment.
What I mean is that concrete InternalEnvironment should contain object good enough for Infrastructure to prepare RuntimeContext w/o any intermediate abstractions/processing.
I.e. if ComposeEnvironment is what prepare() method expects - use it. If "current" implementations due to historical reason need DockerEnvironment - convert it inside RuntimeContext.
It will do the same job with less code and more consistency.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/RuntimeInfrastructure.java
gazarenkov: OK, let's do it in next PR.
Just to make sure - I am talking about removing "provisioner" abstraction on SPI level at all (before GA of course).
Choose for it whatever tactic you want.

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/convert/ComposeEnvironmentConverter.java
garagatyi: Docker env accepts Dockerimage, Dockerfile and Compose internal environments. All of them have different models. To have simpler code we need some way to use them in the same way - convert into the same format. 
Both Docker and OpenShift infras need that. But both of them have their own universal format. 
I don't see that way to use all of the environments in both infras and decrease the amount of converting and simplify the code by removing some converters. 
@gazarenkov Do you have any particular example that can demonstrate how to simplify the code by removing converters? This might help to understand your idea of simplification.

gazarenkov: @garagatyi I perfectly understand what you did but still think you have enough abstractions to not to create one more.
Options:
- Option number 1 (what I did to not to change your code): produce DockerContainers in every Docker*InternalEnv. Pass it to OpenshiftInfra.prepare() - what is the problem? You have all you need.
- Option number 2: leave as it is for Docker*InternalEnv and perform what you did  in Converter inside Infra.prepare 
- I also have option 3 but it requires more reworking so to discuss in future, maybe.

skabashnyuk (approved):

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/convert/ComposeEnvironmentConverter.java
garagatyi: I think option 2 is already in this PR.
With option 1 we would still need to convert compose to DockerInternalEnvironment and then to OpenShiftInternalEnvironment. So I don't see simplification here.
Maybe the option you have in mind and requires refactoring would simplify code but as you said it requires a discussion.

artaleks9:
Selenium test report is OK

Issue #7018, opened 2017-10-26
Selenium tests of Git functionality failed when run simultaneously 


dmytro-ndp:
### Description
There is [org.eclipse.che.selenium.git](https://github.com/eclipse/che/tree/master/selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/git)  tests which influence each other when run simulteniously in several threads, for example for execution _[che-integration-tests:85](https://ci.codenvycorp.com/view/qa/job/che-integration-tests/85/consoleText)_:
```
2017-10-25 22:44:27,221[pool-1-thread-1]  [INFO ] [.e.c.s.c.i.SeleniumTestHandler 177]  - Dependencies injection in org.eclipse.che.selenium.git.CheckoutReferenceTest
2017-10-25 22:44:27,472[onPool-worker-0]  [INFO ] [c.c.TestWorkspaceServiceClient 204]  - Workspace name='workspace615es7' and id='workspaceuhx1ft0mptyzzu31' created
2017-10-25 22:44:34,622[pool-1-thread-3]  [INFO ] [.e.c.s.c.i.SeleniumTestHandler 222]  - Processing @PreDestroy annotation in org.eclipse.che.selenium.git.AmendCommitTest
2017-10-25 22:44:34,645[pool-1-thread-3]  [INFO ] [c.c.TestWorkspaceServiceClient 151]  - Workspace name='workspacewjcj4f', id='workspaceieu4zjy5ebrnlg85', username='che' removed
2017-10-25 22:44:34,728[pool-1-thread-3]  [INFO ] [.e.c.s.c.i.SeleniumTestHandler 177]  - Dependencies injection in org.eclipse.che.selenium.git.CheckoutToRemoteBranchTest
2017-10-25 22:44:35,312[onPool-worker-1]  [INFO ] [c.c.TestWorkspaceServiceClient 204]  - Workspace name='workspace2gfqcn' and id='workspaces1aoi1wegacrgtia' created
2017-10-25 22:44:52,498[onPool-worker-0]  [INFO ] [o.e.c.s.c.w.TestWorkspaceImpl 64]    - Workspace name='workspace615es7' id='workspaceuhx1ft0mptyzzu31' started.
Oct 25, 2017 10:44:52 PM org.openqa.selenium.remote.ProtocolHandshake createSession
INFO: Attempting bi-dialect session, assuming Postel's Law holds true on the remote end
[...]

PR #7148, opened 2017-11-02
Don't run git selenium tests simultaneously

dmytro-ndp:
Signed-off-by: Dmytro Nochevnov <dnochevnov@codenvy.com>

### What does this PR do?
This request is aimed to avoid failures of git-related selenium tests because of simultaneous run in parallel threads.

#7018


selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/client/TestSshServiceClient.java
```diff
@@ -53,7 +69,20 @@ public void deleteMachineKeyByName(String name) throws Exception {
         .request();
   }
 
-  public String generateGithubKey() throws Exception {
+  public synchronized void updateGithubKey() throws Exception {
```
tolusha: It is better to move this method to TestGitHubServiceClient

selenium/che-selenium-test/src/test/resources/suites/CheSuite.xml
```diff
@@ -165,6 +171,7 @@
             <class name="org.eclipse.che.selenium.factory.CheckFactoryWithUntilPolicyTest"/>
             <class name="org.eclipse.che.selenium.factory.CheckOpenFileFeatureTest"/>
             <class name="org.eclipse.che.selenium.factory.CheckRunCommandFeatureTest"/>
+            <class name="org.eclipse.che.selenium.git.CommitFilesByMultiSelectTest"/>
```
tolusha: I guess this is won't help if I try to run all test from the org.eclipse.che.selenium.git package

dmytro-ndp: Yes, but in that case user should run them with `--thread=1`.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/client/TestSshServiceClient.java
dmytro-ndp: **TestGitHubServiceClient** class is going to be replaced by original GitHub client in near future.

selenium/che-selenium-test/src/test/resources/suites/CheSuite.xml
tolusha: Any way, why do git tests fail if they run simultaneously?

dmytro-ndp: Because of setup methods operate with ssh-keys and application grants for github.com

tolusha: So, since method is synchronized no issues occur then?

dmytro-ndp: Probably yes, but it's better to avoid parallel execution of these tests anyway so as github.com and git-related properties are the same for all test workspaces and potentially git-tests can influence each other when they run in parallel.

tolusha: These changes confuse a lot. Shouldn't be done.

dmytro-ndp: That is the limitation of having all tests in one **\<test>** section of _suite-xml_. We do need start git-related tests in one thread. Alternatively,  we can set change _parallel_-mode of test execution to run all tests within the one **\<test>** section one-by-one, but it requires extracting all other tests into the separate **\<test>** sections as well, and it also will increase the time of all tests execution.

dmytro-ndp:
ci-test

selenium/che-selenium-test/src/test/resources/suites/CheSuite.xml
dmytro-ndp: We can try to extract tests of git into separate suite file as described [here](https://stackoverflow.com/a/12300270), but it looks like undocumented feature so as such ability isn't mentioned on [official site of testng](http://testng.org/doc/documentation-main.html).

dmytro-ndp:
ci-test

dmytro-ndp:
ci-test

dmytro-ndp:
ci-test

dmytro-ndp:
ci-test

dmytro-ndp:
ci-test

tolusha (approved):

Issue #7132, opened 2017-11-01
Improve DtoGenerator to annotate overridden methods


azatsarynnyy:
There are a lot of messages outputted during Che build about missed `@Override` annotations on the generated DTO implementations.
![image](https://user-images.githubusercontent.com/1636395/32280226-d282adb8-bf23-11e7-8e77-9539cd5920e7.png)

DtoGenerator should be fixed to annotate overridden methods with `@Override` or annotate generated classes with `@SuppressWarnings("MissingOverride")`.

I've tried to build Che with disabled Error Prone plugin (to avoid outputting these compilation warnings) and reverted [fix](https://github.com/eclipse/che/commit/da1895f0c25815d1c658e160df4857f71c9dc83f) that limits the amount of output in IDE. As a result, browser's tab with IDE was frozen much less time during Che build.

gazarenkov:
Good point and investigation, thanks!
I'd qualify it as a bug of P1 severity b/c it looks really bad and impacts on performance (as you mentioned)

PR #7169, opened 2017-11-02
Make compiler:compile less talkative while compiling the generated DTO impls

azatsarynnyy:
Signed-off-by: Artem Zatsarynnyi <azatsary@redhat.com>

### What does this PR do?
Make `compiler:compile` less talkative while compiling the generated DTO impls in order to make browser's tab with IDE more responsive during Che build.

For example, building of `che-core-api-factory-shared` module outputs ~500 lines less.

#7132

### DIFF
--- a/core/che-core-api-dto/src/main/java/org/eclipse/che/dto/generator/DtoTemplate.java
+++ b/core/che-core-api-dto/src/main/java/org/eclipse/che/dto/generator/DtoTemplate.java
@@ -216,7 +216,7 @@ public class DtoTemplate {
       builder.append("import com.google.gwt.json.client.*;\n");
       builder.append("import com.google.inject.Singleton;\n");
     }
-    builder.append("\n\n@SuppressWarnings({\"unchecked\", \"cast\"})\n");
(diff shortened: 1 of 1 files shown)


sleshchenko:
I'd say that it'd be better to add `@Override` annotations to templates instead of suppressing them. As far as I see [getters](https://github.com/eclipse/che/blob/6a95b6d74a7a592b082180f5e343e29de2519532/core/che-core-api-dto/src/main/java/org/eclipse/che/dto/generator/DtoImplClientTemplate.java#L209) have this annotation but [setters](https://github.com/eclipse/che/blob/6a95b6d74a7a592b082180f5e343e29de2519532/core/che-core-api-dto/src/main/java/org/eclipse/che/dto/generator/DtoImplClientTemplate.java#L1061) and [with...](https://github.com/eclipse/che/blob/6a95b6d74a7a592b082180f5e343e29de2519532/core/che-core-api-dto/src/main/java/org/eclipse/che/dto/generator/DtoImplClientTemplate.java#L1093) don't have.
But it's OK for me to merge it because this issue is not critical for generated DTO sources.

vparfonov (approved):

Issue #7180, opened 2017-11-03
Stabilize ProjectStateAfterRefreshTest and EditorValidationTest selenium tests


Ohrimenko1988:
We have random unexpected fails in the ProjectStateAfterRefreshTest and EditorValidationTest. Especialy this is actual for tests which runs on the Jenkins. So we need investigate reasons and fix this problem.

Ohrimenko1988:
Done https://github.com/eclipse/che/pull/7195

PR #7195, opened 2017-11-06
Stabilize ProjectStateAfterRefreshTest and EditorValidationTest selenium tests

Ohrimenko1988:
### What does this PR do?
Rework several steps in the ProjectStateAfterRefreshTest and EditorValidationTest selenium tests for more stable work.

Issue: https://github.com/eclipse/che/issues/7180

### DIFF
--- a/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
+++ b/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
@@ -703,7 +703,7 @@ public class CodenvyEditor {
    * @param markerType is type of the marker
    */
   public void waitAllMarkersDisappear(String markerType) {
-    loaderDriverWait.until(
+    elemDriverWait.until(
         (ExpectedCondition<Boolean>)
             driver -> driver.findElements(By.xpath(markerType)).size() == 0);
   }
--- a/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
+++ b/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
@@ -703,7 +703,7 @@ public class CodenvyEditor {
    * @param markerType is type of the marker
    */
   public void waitAllMarkersDisappear(String markerType) {
-    elemDriverWait.until(
+    loaderDriverWait.until(
         (ExpectedCondition<Boolean>)
             driver -> driver.findElements(By.xpath(markerType)).size() == 0);
   }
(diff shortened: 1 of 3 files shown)


tolusha (approved):

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/editor/autocomplete/EditorValidationTest.java
```diff
@@ -71,10 +71,10 @@ public void autoCompleteClassInTheSamePackage() {
         PROJECT_NAME + "/src/main/java/org/eclipse/qa/examples/AppController.java");
     editor.waitActiveEditor();
     // validation warnings
-    editor.waitAllMarkersDisappear(ERROR_MARKER);
```
dmytro-ndp: Why it was removed? It was an important check - ensure that there are no error markers just after the **AppController.java** has opened, wasn't it?

Ohrimenko1988: As you see this check is not removed, it is moved to another place, this is related to test logic

dmytro-ndp: Another place - is after the start of editing file, and previous check of absence of warning just after the file opened IMHO is also important thing and also ought to be checked somewhere. Do we have similar check in another test?

dmytro-ndp: By the way - I reviewed last 10 test executions here https://ci.codenvycorp.com/view/qa/job/che-integration-tests-che6/ and can't find failures because of method **waitAllMarkersDisappear()** in which you removed to make the test stable.
Could you, please, write a link to the report of the error in the method **waitAllMarkersDisappear** during the test execution on CI?

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
```diff
@@ -703,7 +703,7 @@ public void waitMarkerDisappears(String markerType, int position) {
    * @param markerType is type of the marker
    */
   public void waitAllMarkersDisappear(String markerType) {
-    loaderDriverWait.until(
+    elemDriverWait.until(
```
dmytro-ndp: **elemDriverWait** is 3 times less timeout compare to **loaderDriverWait**. So it's hardly ever to help to stabilize the tests.

dmytro-ndp (approved):

PR #7209, opened 2017-11-06
Removed scl enable rh-nodejs dependency (centos-nodejs)

JPinkney:
### What does this PR do?
Removes the scl enable rh-nodejs dependency in centos-nodejs stack

As of eclipse/che-dockerfiles@8e70296cb5a9bd79add294b3ce1bd55daf414798 we no longer have to preface commands with scl enable rh-nodejs

#### Release Notes

scl enable rh-nodejs removed from run command in centos-nodejs stack

### DIFF
--- a/ide/che-core-ide-stacks/src/main/resources/stacks.json
+++ b/ide/che-core-ide-stacks/src/main/resources/stacks.json
@@ -2241,7 +2241,7 @@
   },
   {
     "name": "CentOS nodejs",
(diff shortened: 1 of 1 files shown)


l0rd (approved):

l0rd:
cc @dharmit

benoitf:
a PR will be required for che6 branch as well

PR #7214, opened 2017-11-06
Fixed replace_stacks.json after addition of dotnet-centos

JPinkney:
### What does this PR do?
This PR proposes a fix for https://github.com/redhat-developer/rh-che/issues/408. 

replace_stacks.json was throwing a parse error because it was removing the comma after "org.eclipse.che.ws-agent" on this line: https://github.com/redhat-developer/rh-che/blob/master/assembly/fabric8-stacks/src/main/resources/stacks.json#L295 due to the sed at https://github.com/eclipse/che/blob/master/dockerfiles/init/modules/openshift/files/scripts/replace_stacks.sh#L47.

https://github.com/redhat-developer/rh-che/issues/408

#### Release Notes

N/A


riuvshin (approved):
LGTM

benoitf (approved):

l0rd (approved):

PR #7222, opened 2017-11-07
Use extension of jdt.ls

svor:
The PR proposes to use jdt.ls extension (https://github.com/eclipse/che-ls-jdt) instead of jdt.ls.
Also provides a service ( `ExecuteCommandService` ) which helps to execute custom commands on jdt.ls side.

### DIFF
--- a/agents/ls-java/src/main/resources/installers/1.0.1/org.eclipse.che.ls.java.script.sh
+++ b/agents/ls-java/src/main/resources/installers/1.0.1/org.eclipse.che.ls.java.script.sh
@@ -26,7 +26,8 @@ unset PACKAGES
 command -v tar >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" tar"; }
 command -v curl >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" curl"; }
 
-AGENT_BINARIES_URI=http://download.eclipse.org/jdtls/snapshots/jdt-language-server-latest.tar.gz
+DOWNLOAD_AGENT_BINARIES_URI='${WORKSPACE_MASTER_URI}/agent-binaries/jdt.ls.tar.gz'
+
 CHE_DIR=$HOME/che
 LS_DIR=${CHE_DIR}/ls-java
 LS_LAUNCHER=${LS_DIR}/launch.sh
@@ -121,10 +122,16 @@ fi
 
 
 #######################
-### Install Json LS ###
+### Install JDT LS ###
 #######################
 
-curl -sL ${AGENT_BINARIES_URI} | tar xzf - -C ${LS_DIR}
+# Compute URI of workspace master
+WORKSPACE_MASTER_URI=$(echo $CHE_API | cut -d / -f 1-3)
+
+## Evaluate variables now that prefix is defined
+eval "DOWNLOAD_AGENT_BINARIES_URI=${DOWNLOAD_AGENT_BINARIES_URI}"
+
+curl -sL ${DOWNLOAD_AGENT_BINARIES_URI} | tar xzf - -C ${LS_DIR}
 
 touch ${LS_LAUNCHER}
 chmod +x ${LS_LAUNCHER}
--- a/wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/service/ExecuteCommandService.java
+++ b/wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/service/ExecuteCommandService.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright (c) 2012-2017 Red Hat, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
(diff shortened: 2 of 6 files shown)


wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/service/ExecuteCommandService.java
```diff
@@ -0,0 +1,88 @@
...
+import org.eclipse.che.api.languageserver.registry.LanguageServerRegistry;
+import org.eclipse.che.api.languageserver.shared.model.ExtendedExecuteCommandParams;
+import org.eclipse.che.api.languageserver.util.LSOperation;
+import org.eclipse.che.api.languageserver.util.OperationUtil;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Json RPC API for the textDoc to send workspace/executeCommand requests. */
+@Singleton
+public class ExecuteCommandService {
```
tsmaeder, line 31: I'm not sure it's a good idea to expose a general command executor service. The commands we're interested in are specific extensions for jdt.ls.
Also, if we did expose this functionality, WorkspaceService would be a more natural place for it.

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/service/ExecuteCommandService.java
```diff
@@ -0,0 +1,88 @@
...
+
+    final Object[] result = new Object[1];
+    try {
+      List<InitializedLanguageServer> servers =
+          registry
+              .getApplicableLanguageServers(extendedExecuteCommandParams.getFileUri())
+              .stream()
+              .flatMap(Collection::stream)
+              .collect(Collectors.toList());
+      OperationUtil.doInParallel(
```
tsmaeder: ```doInParallel``` will ask ALL language servers to execute the command, however, you are throwing away all results except the last. ```doInSequence()" is probably what you want.

tsmaeder (changes requested):

agents/ls-java/src/main/resources/installers/1.0.1/org.eclipse.che.ls.java.script.sh
```diff
@@ -26,7 +26,8 @@ unset PACKAGES
 command -v tar >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" tar"; }
 command -v curl >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" curl"; }
 
-AGENT_BINARIES_URI=http://download.eclipse.org/jdtls/snapshots/jdt-language-server-latest.tar.gz
+DOWNLOAD_AGENT_BINARIES_URI='${WORKSPACE_MASTER_URI}/agent-binaries/jdt.ls.tar.gz'
```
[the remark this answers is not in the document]
benoitf: it's the black magic of single quotes.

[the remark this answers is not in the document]
svor: WORKSPACE_MASTER_URI can be changed before using DOWNLOAD_AGENT_BINARIES_URI. 
I think is not license issue here because I take already prepared bundle. @skabashnyuk @vparfonov Am I right?

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/service/ExecuteCommandService.java
svor: It's for now we need execute commands just for jdt.ls, maybe this functionality will be useful for other language servers. About `WorkspaceService` it is more like handler of events which were sent from che-client, that's because I moved executeCommand to different service.

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/service/ExecuteCommandService.java
svor: Yes, for now we can change it to `doInSequence()`. Will do it. Thank you

agents/ls-java/src/main/resources/installers/1.0.1/org.eclipse.che.ls.java.script.sh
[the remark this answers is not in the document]
vparfonov: @tsmaeder both projects are Eclipse project, what licensing issue do you mean? Who can help us with this questions?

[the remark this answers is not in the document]
tsmaeder: I talked to @gorkem and I guess anything that is not from the Eclipse foundation itself has been approved for use in Eclpse projects. We should be fine!

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/service/ExecuteCommandService.java
tsmaeder: Nope, part of workspace service: https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#workspace_executeCommand

tsmaeder (approved):

PR #7226, opened 2017-11-07
Fix replace_stacks.json after addition of dotnet-centos (backport from master)

benoitf:
backport https://github.com/eclipse/che/pull/7214 from master to che6

### What does this PR do?
This PR proposes a fix for https://github.com/redhat-developer/rh-che/issues/408. 

replace_stacks.json was throwing a parse error because it was removing the comma after "org.eclipse.che.ws-agent" on this line: https://github.com/redhat-developer/rh-che/blob/master/assembly/fabric8-stacks/src/main/resources/stacks.json#L295 due to the sed at https://github.com/eclipse/che/blob/master/dockerfiles/init/modules/openshift/files/scripts/replace_stacks.sh#L47.

https://github.com/redhat-developer/rh-che/issues/408

#### Release Notes

N/A


riuvshin (approved):

JPinkney:
I think we can remove this since https://github.com/eclipse/che/pull/7289 is the che6 fix for this issue and a bug that appeared in the stacks outlined in https://github.com/eclipse/che/pull/7286

PR #7240, opened 2017-11-08
Improved Deploy script

sleshchenko:
### What does this PR do?
Remove setting of properties that are unused by OpenShift infrastructure.
Remove unused Che properties and `DockerConnectorProvider`.

Rework `sed` commands not to expect default values from yaml file but setting always defaults that are needed for the corresponding flavour.
```
-    if [ "${ENABLE_SSL}" == "false" ]; then sed "s/    CHE_INFRA_OPENSHIFT_TLS__ENABLED: \"true\"/    CHE_INFRA_OPENSHIFT_TLS__ENABLED: \"false\"/" ; else cat -; fi | \
+    sed "s/    CHE_INFRA_OPENSHIFT_TLS__ENABLED:.*/    CHE_INFRA_OPENSHIFT_TLS__ENABLED: \"${ENABLE_SSL}\"/" | \
```
-

#### Release Notes
N/A


sleshchenko:
I don't see any usage of `remote-debugging-enabled` and `keycloak-disabled` values from config map. @davidfestal @ibuziuk Can you confirm that I can remove them in che6 branch?

dockerfiles/init/modules/openshift/files/scripts/che-openshift.yml
```diff
@@ -95,35 +95,19 @@ items:
...
     che.predefined.stacks.reload_on_start: "true"
     log-level: INFO
-    docker-connector: openshift
     port: "8080"
     remote-debugging-enabled: "false"
     che-oauth-github-forceactivation: "true"
-    workspaces-memory-limit: 2300Mi
-    workspaces-memory-request: 1500Mi
     enable-workspaces-autostart: "false"
     keycloak-oso-endpoint: ${KEYCLOAK_OSO_ENDPOINT}
```
garagatyi: Would be awesome to have all these values in the same format, e.g. KEYCLOAK_GITHUB_ENDPOINT

sleshchenko:
@davidfestal @ibuziuk Also Can we remove annotations and labels that are not used in this yaml like https://github.com/eclipse/che/blob/9e7386e726011bf52b64389c492d274bd68fea5f/dockerfiles/init/modules/openshift/files/scripts/che-openshift.yml#L18-L25 https://github.com/eclipse/che/blob/9e7386e726011bf52b64389c492d274bd68fea5f/dockerfiles/init/modules/openshift/files/scripts/che-openshift.yml#L28-L31

dockerfiles/init/modules/openshift/files/scripts/che-openshift.yml
sleshchenko: I'll rework it

davidfestal:
> I don't see any usage of `remote-debugging-enabled` and `keycloak-disabled` values from config map. @davidfestal @ibuziuk Can you confirm that I can remove them in che6 branch?

I can confirm that `keycloak-disabled` is not used anymore. Comes from the initial `rh-che` implementation.

On the contrary for the `remote-debugging-enabled`, it was used in the initial `deploy_che.sh` and yaml deployment descriptor to set the `CHE_DEBUG_SERVER` (see https://github.com/ibuziuk/docs/blob/master/che_remote_debugging.adoc) So I assume this could still be useful also now.

davidfestal:
> @davidfestal @ibuziuk Also Can we remove annotations and labels that are not used in this yaml like

I assume you could remove quite a bunch of them, since the yaml from which this file was initially copied was a concatenated yaml generated by fabric8 for the fabric8 context. But I think that @l0rd and @ibuziuk would know better than me what can be removed.

sleshchenko:
@davidfestal Thanks for your response.
As far as I understand Ilya used this `CHE_DEBUG_SERVER` env var for putting it as env var of Che CLI. But since deploy script doesn't use CLI for the launching of Che it can be removed.

sleshchenko:
@davidfestal @ibuziuk @l0rd I've removed all annotations and labels except `app: che`. Notify me if they should be reverted.

benoitf (approved):

garagatyi (approved):
Looks good! Fewer lines better!

riuvshin (approved):

sleshchenko:
I've move configuration properties from `CHE_SERVER_CONFIGURATION` env var to config map.
@akorneta Also removed the corresponding configuration properties from che-server configuration.
@riuvshin @garagatyi Please take a look one more time

garagatyi (approved):

riuvshin (approved):

l0rd (approved):

skabashnyuk (approved):

sleshchenko:
I was wrong about CHE_DEBUG_SERVER because it is used in the `entrypoint.sh` of `Che` container https://github.com/eclipse/che/blob/980a378c1482ad60ed322046251df8663a587574/dockerfiles/che/entrypoint.sh#L157
@akorneta Will fix the launching of Che server in debug mode deployed by deploy script

PR #7243, opened 2017-11-08
Multi tenant workspace cleaning

davidfestal:
### What does this PR do?

This PR introduces changes required to support some multi-tenancy use-cases that were not supported until now:
- clean openshift resources associated to a workspace when the workspace is idled
- clean openshift resources associated to all started workspaces on che-server shutdown.

These 2 use-cases are critical to allow OSIO multi-tenancy to go to production.

https://github.com/redhat-developer/rh-che/issues/370

### DIFF
--- a/dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
+++ b/dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
@@ -266,6 +266,8 @@ echo "done!"
 # If command == clean up then delete all openshift objects
 # -------------------------------------------------------------
 if [ "${COMMAND}" == "cleanup" ]; then
+  echo "[CHE] Stopping the Che server..."
+  oc scale --replicas=0 --timeout=3h dc che
   echo "[CHE] Deleting all OpenShift objects..."
   oc delete all --all 
   echo "[CHE] Cleanup successfully started. Use \"oc get all\" to verify that all resources have been deleted."
--- a/dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
+++ b/dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
@@ -267,7 +267,7 @@ echo "done!"
 # -------------------------------------------------------------
 if [ "${COMMAND}" == "cleanup" ]; then
   echo "[CHE] Stopping the Che server..."
-  oc scale --replicas=0 --timeout=3h dc che
+  oc scale --replicas=0 --timeout=3m dc che
   echo "[CHE] Deleting all OpenShift objects..."
   oc delete all --all 
   echo "[CHE] Cleanup successfully started. Use \"oc get all\" to verify that all resources have been deleted."
--- a/multiuser/machine-auth/che-multiuser-machine-authentication/src/main/java/org/eclipse/che/multiuser/machine/authentication/server/MachineTokenRegistry.java
+++ b/multiuser/machine-auth/che-multiuser-machine-authentication/src/main/java/org/eclipse/che/multiuser/machine/authentication/server/MachineTokenRegistry.java
@@ -96,6 +96,26 @@ public class MachineTokenRegistry {
   }
 
   /**
+   * Gets workspaceId by machine token
+   *
+   * @return workspace identifier
+   * @throws NotFoundException when no such machine token exists
+   */
+  public String getWorkspaceId(String token) throws NotFoundException {
+    lock.readLock().lock();
+    try {
+      for (Table.Cell<String, String, String> tokenCell : tokens.cellSet()) {
+        if (tokenCell.getValue().equals(token)) {
+          return tokenCell.getRowKey();
+        }
+      }
+      throw new NotFoundException("Workspace not found for token " + token);
+    } finally {
+      lock.readLock().unlock();
+    }
+  }
+
+  /**
    * Invalidates machine security tokens for all users of given workspace.
    *
    * @param workspaceId workspace to invalidate tokens
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
@@ -47,6 +47,7 @@ import io.fabric8.kubernetes.api.model.VolumeMount;
 import io.fabric8.kubernetes.api.model.VolumeMountBuilder;
 import io.fabric8.kubernetes.api.model.extensions.Deployment;
 import io.fabric8.kubernetes.api.model.extensions.DeploymentBuilder;
+import io.fabric8.kubernetes.client.Config;
 import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.Watcher;
 import io.fabric8.kubernetes.client.dsl.ExecWatch;
@@ -96,8 +97,11 @@ import javax.inject.Singleton;
 import okhttp3.OkHttpClient;
 import org.eclipse.che.api.core.notification.EventService;
 import org.eclipse.che.api.core.notification.EventSubscriber;
+import org.eclipse.che.api.workspace.server.WorkspaceSubjectRegistry;
 import org.eclipse.che.api.workspace.server.event.ServerIdleEvent;
 import org.eclipse.che.commons.annotation.Nullable;
+import org.eclipse.che.commons.env.EnvironmentContext;
+import org.eclipse.che.commons.subject.Subject;
 import org.eclipse.che.plugin.docker.client.DockerApiVersionPathPrefixProvider;
 import org.eclipse.che.plugin.docker.client.DockerConnector;
 import org.eclipse.che.plugin.docker.client.DockerConnectorConfiguration;
@@ -235,6 +239,10 @@ public class OpenShiftConnector extends DockerConnector {
   private final OpenshiftWorkspaceEnvironmentProvider openshiftWorkspaceEnvironmentProvider;
   private String apiEndpoint;
   private Boolean apiEndpointRetrieved = false;
+  private final WorkspaceSubjectRegistry workspaceSubjectRegistry;
+
+  private Map<String, String> containerIdToWorkspaceId = new HashMap<>();
+  private Map<String, String> imageIdToWorkspaceId = new HashMap<>();
 
   @Inject
   public OpenShiftConnector(
@@ -247,6 +255,7 @@ public class OpenShiftConnector extends DockerConnector {
       OpenShiftRouteCreator openShiftRouteCreator,
       OpenShiftDeploymentCleaner openShiftDeploymentCleaner,
       EventService eventService,
+      WorkspaceSubjectRegistry workspaceSubjectRegistry,
       @Nullable @Named("che.docker.ip.external") String cheServerExternalAddress,
       WorkspacesRoutingSuffixProvider cheWorkspacesRoutingSuffixProvider,
       @Named("che.openshift.project") String openShiftCheProjectName,
@@ -283,6 +292,7 @@ public class OpenShiftConnector extends DockerConnector {
     this.openShiftPvcHelper = openShiftPvcHelper;
     this.openShiftRouteCreator = openShiftRouteCreator;
     this.openShiftDeploymentCleaner = openShiftDeploymentCleaner;
+    this.workspaceSubjectRegistry = workspaceSubjectRegistry;
     eventService.subscribe(
         new EventSubscriber<ServerIdleEvent>() {
 
@@ -524,6 +534,42 @@ public class OpenShiftConnector extends DockerConnector {
     return portsToRefName;
   }
 
+  private Subject subjectForContainerId(String containerId) {
+    String workspaceId = containerIdToWorkspaceId.get(containerId);
+    if (workspaceId != null) {
+      Subject subject = workspaceSubjectRegistry.getWorkspaceStarter(workspaceId);
+      if (subject != null) {
+        return subject;
+      }
+      Subject envSubject = EnvironmentContext.getCurrent().getSubject();
+      LOG.info(
+          "Didn't find the user that started workspace '{}' => using environment context subject: '{}'",
+          workspaceId,
+          envSubject.getUserName());
+      return envSubject;
+    }
+    LOG.warn("Didn't find container '{}' in the map of created containers", containerId);
+    return null;
+  }
+
+  private Subject subjectForImageId(String imageId) {
+    String workspaceId = imageIdToWorkspaceId.get(imageId);
+    if (workspaceId != null) {
+      Subject subject = workspaceSubjectRegistry.getWorkspaceStarter(workspaceId);
+      if (subject != null) {
+        return subject;
+      }
+      Subject envSubject = EnvironmentContext.getCurrent().getSubject();
+      LOG.info(
+          "Didn't find the user that started workspace '{}' => using environment context subject: '{}'",
+          workspaceId,
+          envSubject.getUserName());
+      return envSubject;
+    }
+    LOG.warn("Didn't find image '{}' in the map of created images", imageId);
+    return null;
+  }
+
   /**
    * @param createContainerParams
    * @return
@@ -539,6 +585,8 @@ public class OpenShiftConnector extends DockerConnector {
     // imageForDocker is the docker version of the image repository. It's needed for other
     // OpenShiftConnector API methods, but is not acceptable as an OpenShift name
     String imageForDocker = createContainerParams.getContainerConfig().getImage();
+    imageIdToWorkspaceId.put(imageForDocker, getOriginalCheWorkspaceId(createContainerParams));
+
     // imageStreamTagName is imageForDocker converted into a form that can be used
     // in OpenShift
     String imageStreamTagName = KubernetesStringUtils.convertPullSpecToTagName(imageForDocker);
@@ -546,7 +594,11 @@ public class OpenShiftConnector extends DockerConnector {
     // imageStreamTagName is not enough to fill out a pull spec; it is only the tag, so we
     // have to get the ImageStreamTag from the tag, and then get the full ImageStreamTag name
     // from that tag. This works because the tags used in Che are unique.
-    ImageStreamTag imageStreamTag = getImageStreamTagFromRepo(imageStreamTagName);
+    ImageStreamTag imageStreamTag =
+        getImageStreamTagFromRepo(
+            imageStreamTagName,
+            openshiftWorkspaceEnvironmentProvider.getWorkspacesOpenshiftConfig(),
+            openshiftWorkspaceEnvironmentProvider.getWorkspacesOpenshiftNamespace());
     String imageStreamTagPullSpec = imageStreamTag.getMetadata().getName();
 
     // Next we need to get the address of the registry where the ImageStreamTag is stored
@@ -662,14 +714,15 @@ public class OpenShiftConnector extends DockerConnector {
         throw new OpenShiftException(
             "Failed to get the ID of the container running in the OpenShift pod");
       }
-    } catch (IOException | KubernetesClientException e) {
+      containerIdToWorkspaceId.put(containerID, getOriginalCheWorkspaceId(createContainerParams));
+    } catch (Exception e) {
       // Make sure we clean up deployment and service in case of an error -- otherwise Che can end
       // up
       // in an inconsistent state.
       LOG.info("Error while creating Pod, removing deployment");
       LOG.info(e.getMessage());
-      openShiftDeploymentCleaner.cleanDeploymentResources(
-          deploymentName, openshiftWorkspaceEnvironmentProvider.getWorkspacesOpenshiftNamespace());
+      imageIdToWorkspaceId.remove(imageForDocker);
+      openShiftDeploymentCleaner.cleanDeploymentResources(deploymentName);
       openShiftClient.resource(imageStreamTag).delete();
       throw e;
     } finally {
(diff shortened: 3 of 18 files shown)


benoitf:
ci-build

davidfestal:
any idea why the CI build failed ? it seems from the console output that the maven build was successful but the the Jenkins job failed.

benoitf:
image

wsmaster/che-core-api-system/src/main/java/org/eclipse/che/api/system/server/SystemManager.java
```diff
@@ -106,6 +106,7 @@ private void doStopServices() {
   @PreDestroy
   @VisibleForTesting
   void shutdown() throws InterruptedException {
+    LOG.info("Synchronous shutdown requested");
```
benoitf: it should be cleanup, no ?

davidfestal: why ? is it a problem to have this log ? it's only called once on shutdown, but might be very useful to check that the post-destroy method is called on SIGTERM, which wasn't the case last week (for an unknown reason, this was fixed when updating to the last master branch).

benoitf: it's more the level of the logger

davidfestal: well, the log level used in line [91](https://github.com/davidfestal/che/blob/f39184e40f737d1eb786aa246b82ee261c06c8af/wsmaster/che-core-api-system/src/main/java/org/eclipse/che/api/system/server/SystemManager.java#L91) is also `INFO`, and the type of event is clearly the same, no ?

benoitf:
ci-build

wsmaster/che-core-api-system/src/main/java/org/eclipse/che/api/system/server/SystemManager.java
benoitf: It's my view on log events (there are other views ;-) I still think these events are not at INFO level.
People that are interested in a specific part want always to push to INFO level but many times it's more DEBUG info. I'm not sure it's interesting for end-user to see that by default.

davidfestal: Fixed [here](https://github.com/eclipse/che/pull/7243/commits/15642ff2b93258e06534cf71adb2de746e4809e2)

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
```diff
@@ -266,6 +266,8 @@ echo "done!"
 # If command == clean up then delete all openshift objects
 # -------------------------------------------------------------
 if [ "${COMMAND}" == "cleanup" ]; then
+  echo "[CHE] Stopping the Che server..."
+  oc scale --replicas=0 --timeout=3h dc che
```
benoitf: as a side note, do we really want to wait 3h there ?

multiuser/machine-auth/che-multiuser-machine-authentication/src/main/java/org/eclipse/che/multiuser/machine/authentication/server/MachineTokenRegistry.java
```diff
@@ -96,6 +96,26 @@ public String getUserId(String token) throws NotFoundException {
   }
 
   /**
+   * Gets workspaceId by machine token
+   *
+   * @return workspace identifier
+   * @throws NotFoundException when no such machine token exists
+   */
+  public String getWorkspaceId(String token) throws NotFoundException {
```
benoitf, line 104: this method should be tested in multiuser/machine-auth/che-multiuser-machine-authentication/src/test/java/org/eclipse/che/multiuser/machine/authentication/server/MachineTokenRegistryTest.java

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceSubjectRegistry.java
```diff
@@ -0,0 +1,101 @@
...
+import javax.inject.Singleton;
+import org.eclipse.che.api.core.notification.EventService;
+import org.eclipse.che.api.core.notification.EventSubscriber;
+import org.eclipse.che.api.workspace.shared.dto.event.WorkspaceStatusEvent;
+import org.eclipse.che.commons.env.EnvironmentContext;
+import org.eclipse.che.commons.subject.Subject;
+import org.slf4j.Logger;
+
+@Singleton
+public class WorkspaceSubjectRegistry implements EventSubscriber<WorkspaceStatusEvent> {
```
benoitf: it is missing unit tests and class comment like What is the meaning of this class, etc.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceSubjectRegistry.java
```diff
@@ -0,0 +1,101 @@
...
+        LOG.warn("Workspace {} is being started by the 'anonymous' user.", workspaceId);
+        return;
+      }
+      userIdToWorkspaces.put(subject.getUserId(), workspaceId);
+      updateSubject(subject);
+    }
+  }
+
+  @PostConstruct
+  @VisibleForTesting
```
benoitf, line 93: I don't see the associated test ?

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
davidfestal: probably not :-) The idea though is to give enough time to stop all the workspaces started on the various external clusters / namespaces associated to users in the multi-tenant scenario.

l0rd (approved):

multiuser/machine-auth/che-multiuser-machine-authentication/src/main/java/org/eclipse/che/multiuser/machine/authentication/server/MachineTokenRegistry.java
skabashnyuk: Can't find where this method is used

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceSubjectRegistry.java
```diff
@@ -0,0 +1,101 @@
...
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import org.eclipse.che.api.core.notification.EventService;
+import org.eclipse.che.api.core.notification.EventSubscriber;
+import org.eclipse.che.api.workspace.shared.dto.event.WorkspaceStatusEvent;
+import org.eclipse.che.commons.env.EnvironmentContext;
+import org.eclipse.che.commons.subject.Subject;
+import org.slf4j.Logger;
+
+@Singleton
```
skabashnyuk: Contract of this class is unclear. Looks like a cache of information that can be retrieved from other API or managers. Can you clarify a contract?

wsmaster/wsmaster-local/src/main/java/org/eclipse/che/api/local/filters/EnvironmentInitializationFilter.java
```diff
@@ -45,6 +49,7 @@ public final void doFilter(
     Subject subject = new SubjectImpl("che", "che", "dummy_token", false);
     HttpSession session = httpRequest.getSession();
     session.setAttribute("codenvy_user", subject);
+    workspaceSubjectRegistry.updateSubject(subject);
```
skabashnyuk, line 52: Why do we need to cache subject?

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -524,6 +534,42 @@ private void idleCheServer(ServerIdleEvent event) {
     return portsToRefName;
   }
 
+  private Subject subjectForContainerId(String containerId) {
+    String workspaceId = containerIdToWorkspaceId.get(containerId);
+    if (workspaceId != null) {
+      Subject subject = workspaceSubjectRegistry.getWorkspaceStarter(workspaceId);
```
skabashnyuk, line 540: Why we need to get Subject from cache instead of ```EnvironmentContext.getCurrent().getSubject();```

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceSubjectRegistry.java
```diff
@@ -0,0 +1,101 @@
...
+    String workspaceId = event.getWorkspaceId();
+    if (WorkspaceStatusEvent.EventType.STOPPED.equals(event.getEventType())) {
+      workspaceOwners.remove(workspaceId);
+      while (userIdToWorkspaces.values().remove(workspaceId)) {}
+    }
+
+    if (WorkspaceStatusEvent.EventType.STARTING.equals(event.getEventType())) {
+      Subject subject = EnvironmentContext.getCurrent().getSubject();
+      if (subject == Subject.ANONYMOUS) {
+        LOG.warn("Workspace {} is being started by the 'anonymous' user.", workspaceId);
```
skabashnyuk: This case is impossible AFAIK. Shall we throw an error instead?

multiuser/machine-auth/che-multiuser-machine-authentication/src/main/java/org/eclipse/che/multiuser/machine/authentication/server/MachineTokenRegistry.java
davidfestal: This method is used by the `rh-che` assembly, in order to get an *up-to-date* Keycloak token that we can inject into the Bayesian Language server when starting it.

See PR https://github.com/redhat-developer/rh-che/pull/417, and especially [this line](https://github.com/redhat-developer/rh-che/pull/417/files#diff-7b667d7b1bc0a0182258f5095b7d3411R43)

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceSubjectRegistry.java
davidfestal: Sure, sorry about that. I'll add it.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
davidfestal: to manage the cases when `OpenshiftConnector` has to to some action on a workspace, triggered by a batch, non-UI event:
1. start/stop of the che-server (which will stop ll workspaces during the shutdown procedure, which in turn will call `OpenshiftConnector.removeImage()` and `OpenshiftConnector.removeContainer()`.
2. stop of the current workspace by the `activity-checker` due to workspace idling, which will also call `OpenshiftConnector.removeImage()` and `OpenshiftConnector.removeContainer()`.

In cases 1 and 2, the `EnvironmentContext.getCurrent().getSubject()` doesn't contain a valid subject that can be used to connect the Openshift cluster/namespace hosting the workspace.

3. Start of the Bayesian language server: in this case, calling `EnvironmentContext.getCurrent().getSubject()` on the wsmaster during a request coming from the wsagent would return the *machine token* instead of the Keycloak token in `subject.getToken()`, which is also wrong.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceSubjectRegistry.java
davidfestal: It's a sort of cache yes, but I'll add Javadoc to the class, as well as tests, which I omitted.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceSubjectRegistry.java
davidfestal: if you confirm that workspace start *always* happens *as a direct result of a user action* in the client-side Javascript application, then I'm not against it.

wsmaster/wsmaster-local/src/main/java/org/eclipse/che/api/local/filters/EnvironmentInitializationFilter.java
davidfestal: for the reasons explained [here](https://github.com/eclipse/che/pull/7243#discussion_r149786402)

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
davidfestal: @l0rd What value would you give to this timeout ?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceSubjectRegistry.java
```diff
@@ -0,0 +1,128 @@
...
+ */
+@Singleton
+public class WorkspaceSubjectRegistry implements EventSubscriber<WorkspaceStatusEvent> {
+
+  private static final Logger LOG = getLogger(WorkspaceSubjectRegistry.class);
+
+  private final EventService eventService;
+  private final ReadWriteLock lock = new ReentrantReadWriteLock();
+  private final Map<String, Subject> workspaceOwners = new HashMap<>();
+  private final Multimap<String, String> userIdToWorkspaces =
```
akorneta, line 61: @davidfestal I propose to replace Multimap and Map to next mapping:

``` Java
Map<String, String> workspaceId2UserId
Map<String, Subject> userId2Subject
```
cases:

**1. when the workspace is stopped**
``` Java
...
String userId = workspaceId2UserId.remove(workspaceId);
if(userId != null) {
    userId2Subject.remove(userId);
}
...
```

**2. when the workspace is started**
``` Java
...
workspaceId2UserId.put(workspaceId, subject.getUserId());
updateSubject(Subject subject);
...

public void updateSubject(Subject subject) {
    String token = subject != null ? subject.getToken() : null;
    if (token != null && token.startsWith("machine")) {      
      return;
    }
    ...
    userId2Subject.put(subject.getUserId(), subject);
    ...
  }
```

**3. get subject by workpsace id:**

``` Java
public Subject getWorkspaceStarter(String workspaceId) {
  ...
  String userId = workspaceId2UserId.get(workspaceId);
  if(userId != null) {
    return userId2Subject.get(userId);     
  }
  return null;
  ...
}
```
WDYT?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceSubjectRegistry.java
davidfestal: Done in commit 9d94514a09a4f13fcd2e2d472acddbdda66150f0

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceSubjectRegistry.java
davidfestal: Javadoc added in commit 3b911356925dabd7c02b5fdd4e96b8c62dae8b3d

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceSubjectRegistry.java
davidfestal: Javadoc added in commit 3b911356925dabd7c02b5fdd4e96b8c62dae8b3d
Unit tests added in commit 4a4a4f0495af4446a50bcd59e57896975b0b80bb

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceSubjectRegistry.java
davidfestal: see commit 4a4a4f0495af4446a50bcd59e57896975b0b80bb

benoitf:
ci-build

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
benoitf: it would be great if subject is not cached
why batch jobs can't send the required data ? or why associated subject is not propagated ? on removeContainer ?

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
l0rd: 3 minutes?

davidfestal: Fixed in commit cc969613ece31042561699261281c7813498dc0d

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceSubjectRegistry.java
davidfestal: It seems to me that this proposal doesn't support starting 2 workspaces with the same user and then stopping only one of both. The user Subject would be lost though it is still useful for the second workspace.

Or did I miss something ?

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
davidfestal: @benoitf in those 3 scenarios, currently (before this PR) the required user information (contained in the subject) is available nowhere in the wsmaster application. And since these 3 scenarios are triggered by some non-interactive event (idling, tomcat shutdown hook, internal event indirectly triggered by the language server registry), we have no way to send or propagate the user information, afaik.

To be clear, on the other hand, workspace stop initiated by the user himself from the Javascript UIs (GWT or Dashboard) was already working great.

davidfestal:
@skabashnyuk @benoitf @akorneta I tried to answer all your comments, but I might have missed some. Could you tell me if you still have pending comments / questions that still didn't get a satisfying answer ?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceSubjectRegistry.java
akorneta: yes, I missed that and the first case when the workspace is stopped should look like this:
``` Java
String userId = workspaceId2UserId.remove(workspaceId);
if(userId != null && !workspaceId2UserId.values().contains(userId)) {
    userId2Subject.remove(userId);
}
```

benoitf (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceSubjectRegistry.java
davidfestal: also it seems to me that in this proposal, `updateSubject(Subject subject)` would allow adding in the `userId2Subject` map a `userId` that wasn't previously associated to a workspace. In such a case, it will never be removed from the map and we have a memory leak, no ?

Could you explain what seems like a problem for you in the current implementation ?

skabashnyuk (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceSubjectRegistry.java
akorneta: Basically leak that you described is possible, but I thought that the start of the workspace it is the only case when this method is used and it can be solved by adding a check like:
``` Java
if (workspaceId2UserId.values().contains(userId)) {
   userId2Subject.put(subject.getUserId(), subject);
}
```

What I'm trying to improve it is the mapping of:
1 user to n workspaces and n workspaces to 1 subject
to
n workspaces to 1 user and 1 user to 1 subject.

Do you see what I mean?

davidfestal: Yes, I see, but I'm not sure performance would be better in the `updateSubject()` method, knowing that you potentially need to iterate the whole list of started workspaces.

And the `updateSubject()` method should be as fast as possible, since it's called each time a client request comes to the wsmaster.

Another point: with your proposal, `getWorkspaceStarter()` requires 2 hash map `get` calls instead  of one.

In fact the idea of the current implementation was to optimize performance in those 2 methods that are called much more often than the workspace start or stop methods.

akorneta: @davidfestal I see, so let it be as it is.

akorneta (approved):

garagatyi (approved):

benoitf:
ci-build

vparfonov (approved):

riuvshin (approved):

Issue #7128, opened 2017-11-01
Folders are not visible in the Project Explorer if in the Editor opened a file from root of a project folder


SkorikSergey:
**Reproduction Steps:**

1. Create a workspace from the **Java** stack with the **web-java-spring** project.
2. Open in the **Editor** README.md or pom.xml file.
3. Close the folder
4. Refresh the web page

**Observed behavior:**
After refreshing in **the Project Explorer** visible only files which opened in the **Editor**.
This issue is similar to https://github.com/eclipse/che/issues/4733 and https://github.com/eclipse/che/issues/5986, except that the **web-java-spring** folder should be closed before updating.
This issue is also reproduced if open a file from the **Navigate To File** feature.

**Che version:** 5.21.0-SNAPSHOT 
**OS and version:** Fedora 25

**Additional information:**
![teams-11-67](https://user-images.githubusercontent.com/7760565/32276305-4748a132-bf17-11e7-8da6-6326bd9a959e.gif)
Reproducing this issue with the **Navigate To File** feature:
![teams-11-63](https://user-images.githubusercontent.com/7760565/32275908-b1be0c84-bf15-11e7-9197-2943d904c3dc.gif)

PR #7250, opened 2017-11-08
findResource method refactoring related to hierarchical load resource tree

vzhukovs:
### What does this PR do?
This changes proposal modifies the load resources from the server. As Resource Manager uses internal storage for caching loaded file resources we need to take in account that situation, when user request from the server file for example with path `/project/a/b/file.ext` then resource manager should get single tree to the file path. It means that resource manager should call and cache (if need) tree `/project`, then tree `/project/a`, then tree `/project/a/b` and find `file.ext` inside `b` directory.

#7128 

#### Changelog 
findResource method refactoring related to hierarchical load resource tree

#### Release Notes
N/A

#### Docs PR
N/A


vzhukovs:
ci-build

ashumilova (approved):

vzhukovs:
ci-test

Issue #7095, opened 2017-10-30
[CHE6 branch] All content of panel menu is not displayed  


artaleks9:
**Reproduction Steps:**
* Open a project in the IDE (e.g. web-java-spring)
* Split 'process' panel and create several processes
* Open panel menu

**Expected behavior:**
* all content of panel menu should be displayed

**Observed behavior:**
* part of content is not displayed

**Che version:** 6.0.0-M1
**OS and version:** Fedora 25
**Docker version:** 17.06
**API docker version:** 1.30
**Che install:** Docker container

**Additional information:**
* Related selenium test: _WorkingWithSplitPanelTest_
* Problem can be reliably reproduced, doesn't happen randomly: [Yes]
* See the attachment:
![panel-menu](https://user-images.githubusercontent.com/5408906/32183458-04a90162-bda2-11e7-9488-244af4f2a6e8.png)

PR #7283, opened 2017-11-09
Fix menu position: if not enough space under button will be shown above 

vparfonov:

### What does this PR do?
Fix menu position: if not enough space under button will be shown above.
Should fix _WorkingWithSplitPanelTest_
Before:
not-dispay
After:
dispay

#7095

### DIFF
--- a/ide/che-core-ide-ui/src/main/java/org/eclipse/che/ide/ui/multisplitpanel/menu/MenuWidget.java
+++ b/ide/che-core-ide-ui/src/main/java/org/eclipse/che/ide/ui/multisplitpanel/menu/MenuWidget.java
@@ -46,7 +46,6 @@ public class MenuWidget extends Composite implements Menu {
 
   private ActionDelegate delegate;
   private long closeTime;
-  private int itemsInMenu;
 
   private MenuItem.ActionDelegate itemDelegate =
       new MenuItem.ActionDelegate() {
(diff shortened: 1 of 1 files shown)


vparfonov:
ci-test

ide/che-core-ide-ui/src/main/java/org/eclipse/che/ide/ui/multisplitpanel/menu/MenuWidget.java
```diff
@@ -106,24 +108,41 @@ public void onClose(CloseEvent<PopupPanel> event) {
...
+    itemsInMenu--;
     listPanel.remove(menuItem);
+    fixPositionIfNeed();
+  }
+
+  /**
+   * Fix top value if need. Need in case if bottom part of menu not display
+   */
+  private void fixPositionIfNeed() {
+    int totalHeight = itemsInMenu * 19; //19 height of menu item
```
azatsarynnyy: It seems that variable `itemsInMenu` stores the same value as `listPanel.getWidgetCount()` returns.
So you can avoid using an additional variable.

ide/che-core-ide-ui/src/main/java/org/eclipse/che/ide/ui/multisplitpanel/menu/MenuWidget.java
```diff
@@ -106,24 +108,41 @@ public void onClose(CloseEvent<PopupPanel> event) {
...
   public void removeListItem(MenuItem menuItem) {
+    itemsInMenu--;
     listPanel.remove(menuItem);
+    fixPositionIfNeed();
+  }
+
+  /**
+   * Fix top value if need. Need in case if bottom part of menu not display
+   */
+  private void fixPositionIfNeed() {
```
azatsarynnyy: Let's name it as `adjustTopPosition`? I think this name is clearer and better reflects what the method does.

vparfonov: +1 thx, I tried to found this name

ide/che-core-ide-ui/src/main/java/org/eclipse/che/ide/ui/multisplitpanel/menu/MenuWidget.java
vparfonov: Ok, I will try

azatsarynnyy (approved):

PR #7286, opened 2017-11-09
Fix for replace stacks not working when agents was on multiline

JPinkney:
### What does this PR do?
This PR fixes replace_stacks to work with agents that are in a flat list in the stacks json as well as multi line. The regression was introduced in 2e90ad67225e31b23e320fa7f2f536509af2abe2 due to a misunderstanding of sed as well as a misunderstanding of the use case.

Fixes replace_stacks to work with agents that are a flat list in the stacks json as well as multi line.

Previously it would not remove com.redhat.bayesian.lsp from something like https://github.com/redhat-developer/rh-che/blob/master/assembly/fabric8-stacks/src/main/resources/stacks.json#L542 but would work for something like https://github.com/redhat-developer/rh-che/blob/master/assembly/fabric8-stacks/src/main/resources/stacks.json#L184. This PR makes it work for both cases.

#### Release Notes

N/A


benoitf:
there is a conflict on the file

JPinkney:
I believe the conflict should be fixed now

benoitf (approved):

l0rd (approved):

riuvshin (approved):

PR #7289, opened 2017-11-09
Fix for replace_stacks agents on multiline case

JPinkney:

### What does this PR do?
This PR fixes replace_stacks to work with agents that are in a flat list in the stacks json as well as multi line.

Fixes replace_stacks to work with agents that are a flat list in the stacks json as well as multi line.

#### Release Notes

N/A

### DIFF
--- a/dockerfiles/init/modules/openshift/files/scripts/replace_stacks.sh
+++ b/dockerfiles/init/modules/openshift/files/scripts/replace_stacks.sh
@@ -44,7 +44,15 @@ echo ""
 
 echo -n "[CHE] Fetching the list of new Che stacks..."
 #new_stacks_json=$(curl -X GET -s --header 'Accept: application/json' "${NEW_STACKS_URL}" | sed 's/\\\"//g' | sed 's/\"com\.redhat\.bayesian\.lsp\"//g' | sed 's/ws-agent\",/ws-agent\"/g')
-new_stacks_json=$(curl -X GET -s --header 'Accept: application/json' "${NEW_STACKS_URL}" | sed 's/\"com\.redhat\.bayesian\.lsp\"//g' | sed 's/ws-agent\",/ws-agent\"/g')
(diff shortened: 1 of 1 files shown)


benoitf:
there is a conflict on the file

JPinkney:
I believe the conflict should be fixed now

benoitf (approved):

l0rd (approved):

riuvshin (approved):

matskiv:
I was about to send PR for this bug and noticed this one. This solution should be more reliable.
I did just a small enhancement on original sed :)

`new_stacks_json=$(curl -X GET -s --header 'Accept: application/json' "${NEW_STACKS_URL}" | sed 's/\"com\.redhat\.bayesian\.lsp\"//g' | sed 's/ws-agent\",\s*$/ws-agent\"/g')`

l0rd:
@matskiv +1, @JPinkney could you give it a try?

JPinkney:
@l0rd That solution will work in this case but it can break if agents are seperated onto multiple lines and ws-agent is not followed by bayesian due to a parse error. E.g. if the stacks ever end up like: https://gist.github.com/JPinkney/8593eeff5d3aa5a62e76f0ea9314dd85#file-test2-json-L297 where agents are on multiple lines and ws-agent isn't followed by bayesian then the comma after ws-agent will be removed and the parse will fail.

matskiv:
@JPinkney Yes, you are right :+1: That's why I was saying that your solution is more reliable (and I just realized that it looks like I am saying that about my fix).

l0rd:
Thanks @JPinkney and sorry for the confusion. ~Can you apply this same fix to branch `che6` too please~. Sorry @JPinkney this is che6 :-(

benoitf:
@l0rd this PR was for che6 branch

benoitf:
@l0rd for master branch, PR is there: https://github.com/eclipse/che/pull/7286

l0rd:
Thanks @benoitf I made some confusion ;-)

PR #7297, opened 2017-11-10
Allow to use Eclipse OpenJ9 JVM to run che-server

benoitf:
### What does this PR do?
Eclipse Che can be run on Java9 runtime. So adding Dockerfile for Eclipse JVM OpenJ9

#5326 

#### Release Notes
N/A

#### Docs PR
N/A

Change-Id: I309e6eb1af4d0166758481bfd462c074de384ad6

### DIFF
--- a/dockerfiles/che/Dockerfile.openj9
+++ b/dockerfiles/che/Dockerfile.openj9
@@ -1,0 +1,41 @@
+# Copyright (c) 2017 Red Hat, Inc.
+# All rights reserved. This program and the accompanying materials
+# are made available under the terms of the Eclipse Public License v1.0
+# which accompanies this distribution, and is available at
+# http://www.eclipse.org/legal/epl-v10.html
+#
+# Contributors:
+#   Florent Benoit - Initial Implementation
+#
+# To build it, run in the repository root:
+#  `docker build -t eclipse/che-server:openj9 -f Dockerfile.openj9 .`
+#
+# To run it:
+#  docker run -e IMAGE_CHE=eclipse/che-server:openj9 \
+#             -it \
+#             --rm \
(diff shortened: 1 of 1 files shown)


skabashnyuk:
How Dockerfile.openj9 will be used?

benoitf:
ppl can use it (like centos one)

ghost (approved):

skabashnyuk:
Imagine I'm that man who wants to use it.
Are you sure this is a right way?
```
+# To run it:
 +#  docker run --net=host \
 +#             --name che \
 +#             -v /var/run/docker.sock:/var/run/docker.sock \
 +#             -v /home/user/che/lib:/home/user/che/lib-copy \
 +#             -v /home/user/che/workspaces:/home/user/che/workspaces \
 +#             -v /home/user/che/storage:/home/user/che/storage \
 +#             codenvy/che-server
```
Onother question. Licensing.
@riuvshin  afaik we adding some notes about licenses for OpenJDK. WDYT do we need to do the same here?

benoitf:
@skabashnyuk the run instruction is a duplicate of Dockerfile (I've updated to remove codenvy from image name)

about licensing AFAIK OpenJ9 is EPL 2.0 and Apache 2.0 licensed

benoitf:
@skabashnyuk also with CLI you can override name of the images, so you could use `-e IMAGE_CHE="eclipse/che-server:nightly-openj9"`
I may update the launch part with that flag

skabashnyuk:
ok. Thx for clarification.

benoitf:
@skabashnyuk I've updated build and run instructions

riuvshin (approved):

Issue #7094, opened 2017-10-30
Unify the way how che-server image is configured


skabashnyuk:
### Description
In Che 6 we introducing infrastructures as a first level citizens. 
It happened that easiest and most widely way to provide configuration become an environment variables. In our configuration framework, we have this support.
The sequence are following :
1. che.properties inside of war.
2. che.properties in CHE_CONF dir
3. System variables
4. Environment variables.

In times of Che5 where docker was our primary target we used che.properties
https://github.com/eclipse/che/blob/master/assembly/assembly-main/src/assembly/assembly.xml#L162-L165
and put it outside of war. Also, we doing some other actions with this file in 
https://github.com/eclipse/che/blob/master/dockerfiles/che/entrypoint.sh#L247-L292

this way make a couple of difficulties.
1. Not all configuration parameters of che-dev image is configured with environment variables
2. Changes default values in che.properties inside of war have no effect of existed installation.
3. It brings difficulties for developers in understating where actually properties are located.
4. It revert's upside down original concept of default values packaged in war and managed by authors of code.

skabashnyuk:
After some discussions with @riuvshin  we have such proposal.
Remove default che.properties extraction from CLI in Che6. 
All properties have to be configured from che.env. If user wants to configure some additional 
properties he can do that by adding additional values to che.env.

@benoitf please validate this idea and let me know if you have some concerns.

benoitf:
@skabashnyuk I'm perfectly fine

davidwindell:
Will env vars still work as overrides?

benoitf:
@davidwindell I would say yes

skabashnyuk:
@davidwindell I confirm - yes env vars still work as overrides
The schema will be
1 .defaults che.properties inside of war.
2. che.properties in CHE_CONF dir
3. System variables
4. Environment variables.

Issue #6750, opened 2017-10-16
[Multiuser] Instructions to run without CLI


davidwindell:
In the docs, instructions are provided to run Multi-user Che with the CLI. 

We use a orchestrator (Rancher) to deploy our servers so we need to replicate how `CHE_MULTIUSER=true` works manually (i.e. by launching `che-server-multiuser`, etc).

Could someone update the docs or provide a link here to follow?

ghost:
@davidwindell currently there are no docs as it's automated in CLI. Even though the process is not too complicated, there are things to take care of:

* configuration for Keycloak and Postgres
* volumes
* containers boot order (postgres-keycloak-che)

Here's composefile https://github.com/eclipse/che/blob/master/dockerfiles/init/modules/compose/templates/docker-compose.yml.erb

And this is how keycloak and postgres are configured https://github.com/eclipse/che/tree/master/dockerfiles/init/modules

I'll try to document it later

davidwindell:
Fantastic, thanks!

davidwindell:
After a bit of a push and shove, I've been able to get this to work in Rancher/Compose :+1: This is roughly what our compose definition now looks like:

```yml
version: '2'
services:
  keycloak:
    image: jboss/keycloak:3.3.0.CR2-3
    volumes:
    - /host/path/to/che/theme:/opt/jboss/keycloak/themes/che
    links:
    - postgres:postgres
    ports:
    - 5050:8080/tcp=
  che-server:
    image: eclipse/che-server-multiuser:5.19.0
    environment:
      CHE_PORT: '80'
      CHE_DOCKER_IP_EXTERNAL: our.che.host
      CHE_DOCKER_SERVER__EVALUATION__STRATEGY: docker-local
      CHE_HOST: our.che.host
      CHE_WORKSPACE_VOLUME: /var/run/docker.sock:/var/run/docker.sock
      CHE_MULTIUSER: 'true'
    volumes:
    - /var/run/docker.sock:/var/run/docker.sock
    links:
    - postgres:postgres
  postgres:
    image: postgres:9.6-alpine
    environment:
      POSTGRES_USER: keycloak
      POSTGRES_PASSWORD: password
    volumes:
    - /host/path/to/postgres/data:/var/lib/postgresql/data
```

Anyone who takes this route will have to create their own che and keycloak databases and keycloak realm (I just imported the erb template into keycloak). We also mount the Che keycloak theme as a manual process.

davidwindell:
[...]

PR #7299, opened 2017-11-10
Configure che-server only from environment variables

skabashnyuk:
### What does this PR do?
- Removed default properties extraction from war.
- Removed all usage of ```CHE_LOCAL_CONF_DIR```
- Forced unset of ```CHE_LOCAL_CONF_DIR``` to make sure that Eclipse Che server can be configured only with environment variables.  
- Changed location of default properties from ```/codenvy/che.properties``` to ```/che/che.properties```

Unify the way how che-server image is configured https://github.com/eclipse/che/issues/7094

#### Release Notes
che-server image can be configured only with environment variables.

#### Docs PR
https://github.com/eclipse/che-docs/pull/315


skabashnyuk:
Test docker
- assembly mode
- repo mode
- image mode

OpenShift in progress

skabashnyuk:
OCP - tested

dockerfiles/che/entrypoint.sh
```diff
@@ -99,10 +98,8 @@ set_environment_variables () {
     CHE_HOME=$(echo /"${CHE_HOME}" | sed  's|\\|/|g' | sed 's|:||g')
   fi
 
-  # Che configuration directory - where che.properties lives
-  if [ -z "${CHE_LOCAL_CONF_DIR}" ]; then
-    export CHE_LOCAL_CONF_DIR="${CHE_HOME}/conf/"
-  fi
+  # All Eclipse Che variables has to be set as environment variables.
+  unset CHE_LOCAL_CONF_DIR
```
benoitf: question why do we need to unset if it's not used ?

skabashnyuk: To make sure nobody trying to work around it and we have perfect portability.
BTW If it's blocker for you I can remove it.

riuvshin: Im ok to remove that unset

skabashnyuk: ok. No problem

ghost (approved):

benoitf:
@skabashnyuk I would remove https://github.com/eclipse/che/blob/che6/core/commons/che-core-commons-inject/src/main/java/org/eclipse/che/inject/CheBootstrap.java#L108
and https://github.com/eclipse/che/blob/che6/core/commons/che-core-commons-inject/src/main/java/org/eclipse/che/inject/CheBootstrap.java#L222-L226
(so no need to unset CHE_LOCAL_CONF_DIR, it will be not be used even if defined)

benoitf:
so even agent should not use that pattern if possible
https://github.com/eclipse/che/blob/che6/assembly/assembly-wsagent-server/src/assembly/setenv.sh#L13 and https://github.com/eclipse/che/blob/che6/assembly/assembly-wsagent-server/src/assembly/setenv.sh#L31

skabashnyuk:
I've talked with @riuvshin  and he suggested not to hurry up. Maybe we may have the situation where it might be useful.  For instance for users who don't use cli or che-server image and natively runs tomcat.

benoitf:
who are these users ?

benoitf:
we could add warning at https://github.com/eclipse/che/blob/che6/core/commons/che-core-commons-inject/src/main/java/org/eclipse/che/inject/CheBootstrap.java#L224 that it will be removed before 6.0 GA

skabashnyuk:
For example https://github.com/eclipse/che/issues/6750

skabashnyuk:
I'm not 100% sure that removing at all CHE_LOCAL_CONF at all is a good idea. 
I'm sure that is not needed for che-server image. But it might be some cases when it can be useful. 
@riuvshin @eivantsov  wdyt?

benoitf:
@skabashnyuk on #6750 it's using the docker image

skabashnyuk:
I mean that it might be other contributors who use che in a different way then we are and we should carefully think before removing some ability completely.  For instance, this pr still give the ability to define CHE_LOCAL_CONF directory and mount it in a container if needed, for urgent cases.
If you don't mind I would decouple this issue https://github.com/eclipse/che/issues/7094 from removing completely  ```CHE_LOCAL_CONF```

benoitf:
I do not say that it should not be done carefully but che6 is allowing to do that. so we may deprecated in 5.x and remove it in 6.x

on contributors, I think that env file is allowing to do the extension, also with custom assemblies you can override default value with properties file as well.

And yes I'm fine to decouple issue but as this `local conf folder` way is the "old way" of configuring I would say it's time to get rid of it definitely.

riuvshin (approved):

benoitf (approved):

benoitf:
@skabashnyuk should I create an issue for getting rid of conf folder for all instances ?

skabashnyuk:
@benoitf If you think so - then yes, why no. I'm now 40/60(more no, than yes) about that. An issue will be a good place to elaborate about that.

benoitf:
Hello, I've created https://github.com/eclipse/che/issues/7324, it will be easier to discuss in this issue.

Issue #7340, opened 2017-11-13
[CHE6-OCP] (multiuser) Randomly failing tests on the login page


Ohrimenko1988:
### Description
After run full selenium tests suite on the CHE6-OCP (multiuser) we have 20 - 30 random tests which can not log in to the che by login page. If compare with CHE6-OCP (singleuser) we have close 10 % regression. So we need investigate and resolve this problem.

**Report with reproduced bug:**
https://ci.codenvycorp.com/view/qa/job/che-multiuser-integration-tests-che6-ocp/6/Selenium_tests_report/

**Report for the same day in the single user mode for compare:**
https://ci.codenvycorp.com/view/qa/job/che-integration-tests-che6-ocp/38/Selenium_tests_report/

**Attachments:**

![aaa](https://user-images.githubusercontent.com/22724619/32734335-6b585fb8-c89a-11e7-8063-d9bae0f62116.png)

**OS and version:**    
CHE6-OCP (multiuser) 6.0.0-M2 ; Chrome 59.0 ; Fedora 25

**Diagnostics:**

Ohrimenko1988:
Done https://github.com/eclipse/che/pull/7455

PR #7341, opened 2017-11-13
Resolve problem with selenium random fails related to login page

Ohrimenko1988:
### What does this PR do?
Resolve the problem with selenium random fails related to login page. Added input checks of username and password fields and wait for "Login" button visibility.
 
Issue: https://github.com/eclipse/che/issues/7340


dmytro-ndp:
Good catch! It is the approach which we need to use around the most sendKeys().
Also next code would be enclosed in the separate method, say **"rewrite(WebElement field, String content)"**:
```javascript
usernameInput.clear();
usernameInput.sendKeys(username);
waitTextIsPresent(usernameInput, username);
```

dmytro-ndp (approved):

dmytro-ndp:
We need to merge it to check the result on Multi-User Che so as **ci-test** starts Single-user Che.

PR #7350, opened 2017-11-14
Fix deploy script

riuvshin:
Fix deploy script in case deploy on ocp


sleshchenko (approved):
Thanks for quick fix!

garagatyi (approved):

Issue #7327, opened 2017-11-13
Rework testing of Find Text Feature presentation


SkorikSergey:
We need rework the **FindTextFeatureTest** selenium test:
- Because **checkTextResultsPagination** test method of **FindTextFeatureTest** selenium test fails by https://github.com/eclipse/che/issues/7055 in start of the method and don't check other part of feature we need rework it to check count of a page occurrences in the end of the test method.
- Add checking count of found occurrences from all pages.

SkorikSergey:
Merged.

Issue #7055, opened 2017-10-27
Results of the Find Text feature are displayed in different order in CHE and OCP


SkorikSergey:
**Reproduction Steps:**
- Create a workspace from the **Java-MySql** stack with the **web-java-petclinic** project
- **Menu** -> **Edit** -> **Find**
- Type in form 'Str' and press the **Search** button

Do the same steps in **Codenvy**

**Observed behavior:**
In **CHE** there are 125 occurrences on the first page but in **Codenvy** there are 144 occurrences.

**Che version:** 5.20.0-SNAPSHOT 
**OS and version:** Fedora 25

**Additional information:**
![teams-11-55](https://user-images.githubusercontent.com/7760565/32107832-6f261640-bb38-11e7-91cb-8ed0b80778c0.gif)

CHE
![eclipse che wksp-ptyq - google chrome_025](https://user-images.githubusercontent.com/7760565/32107652-fa9054d0-bb37-11e7-8b8e-5d26da6cc972.png)

Codenvy(https://nightly.codenvy-stg.com)
![codenvy wksp-6t0c - google chrome_024](https://user-images.githubusercontent.com/7760565/32107653-fd3a3458-bb37-11e7-9bf6-a8c59fa13203.png)

SkorikSergey:
Not actual.

PR #7355, opened 2017-11-14
Rework testing of Find Text Feature results presentation

SkorikSergey:
### What does this PR do?
We need rework the **FindTextFeatureTest** selenium test:
- Because **checkTextResultsPagination** test method of **FindTextFeatureTest** selenium test fails by https://github.com/eclipse/che/issues/7055 in start of the method and don't check other part of feature we need rework it to check count of a page occurrences in the end of the test method.
- Checking that sums of found occurrences and files from all pages are correct.

https://github.com/eclipse/che/issues/7327


selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/FindText.java
```diff
@@ -506,4 +511,21 @@ public void openFileNodeByDoubleClick(String pathToFile) {
         .click();
     actionsFactory.createAction(seleniumWebDriver).doubleClick().perform();
   }
+
+  /** get number of found occurrences in a page */
```
dmytro-ndp: Please, add an example into the method docs.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/FindText.java
```diff
@@ -506,4 +511,21 @@ public void openFileNodeByDoubleClick(String pathToFile) {
...
+  public int getFoundOccurrencesNumberOnPage() {
+    List<String> results = new ArrayList<>(Arrays.asList(getResults().split(" ")));
+    return Integer.parseInt(results.get(0));
+  }
+
+  /** get number of found files in a page */
```
dmytro-ndp: Please, add an example into the method docs.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/FindText.java
```diff
@@ -491,10 +494,12 @@ public void clickOnNextPageButton() {
   }
 
   public Boolean checkNextPageButtonIsEnabled() {
+    WaitUtils.sleepQuietly(1);
```
tolusha: If possible it is better to avoid `WaitUtils.sleepQuietly(1);`

tolusha: `visibilityOfElementLocated` would work well here

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/FindText.java
dmytro-ndp: I suggest you to create data value class to encapsulate each number:
```
SearchFileResult(String result)
-----------------------
private final int foundOnPage;
private final int occurrenceOnPage;
private final int totalFound;
-----------------------
public int getFoundOnPage()
public int getOccurranceOnPage()
public int getTotalFound()
-----------------------

File: selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/FindText.java [hunk 4]
```diff
@@ -506,4 +511,21 @@ public void openFileNodeByDoubleClick(String pathToFile) {
         .click();
     actionsFactory.createAction(seleniumWebDriver).doubleClick().perform();
   }
+
+  /** get number of found occurrences in a page */
+  public int getFoundOccurrencesNumberOnPage() {
+    List<String> results = new ArrayList<>(Arrays.asList(getResults().split(" ")));
```
dmytro-ndp: We don't need to convert from array to list and back here, so as split() command returns String[].
You can use this command here instead:
```javascript
return Integer.parseInt(getResults().split(" ")[0]);
```

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/miscellaneous/FindTextFeatureTest.java
```diff
@@ -387,54 +424,64 @@ public void checkTextResultsPagination() {
...
+    Assert.assertTrue(findText.checkNextPageButtonIsEnabled());
     Assert.assertFalse(findText.checkPreviousPageButtonIsEnabled());
+    sumOfFoundFiles += findText.getFoundFilesNumberOnPage();
+    sumOfFoundOccurrences += findText.getFoundOccurrencesNumberOnPage();
     findText.clickOnNextPageButton();
 
-    // check results on second page and the previous page button is enabled
-    Assert.assertEquals(findText.getResults(), resultsOnSecondPage);
+    // Check move page buttons status on the second page
+    Assert.assertTrue(findText.checkNextPageButtonIsEnabled());
```
dmytro-ndp: It's better to use static import here to improve readability.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/FindText.java
SkorikSergey: The **next** and **previous** page buttons are always visible in the **Find Text Info** panel and timeout is needed for waiting that the buttons can change their status after clicking(enabled or disabled).

tolusha: I guess there is a way :)

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/FindText.java
dmytro-ndp: This object **SearchFileResult** being returned from the dedicated method of **FindText** class allows to get all required data by one method call, and to have logic to parse the search results string in one most suitable place, that is to comply the [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single_responsibility_principle).

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/FindText.java
```diff
@@ -507,17 +507,36 @@ public void openFileNodeByDoubleClick(String pathToFile) {
     actionsFactory.createAction(seleniumWebDriver).doubleClick().perform();
   }
 
-  /** get number of found occurrences in a page */
-  public int getFoundOccurrencesNumberOnPage() {
-    return Integer.parseInt(getResults().split(" ")[0]);
-  }
+  public static class SearchFileResult {
+    private int occurrencesFoundOnPage;
```
dmytro-ndp: don't forget to make them final

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/miscellaneous/FindTextFeatureTest.java
```diff
@@ -442,51 +440,51 @@ public void checkTextResultsPagination() {
 
     // Check move page buttons status on the first page
     findText.sendCommandByKeyboardInFindInfoPanel(ARROW_DOWN.toString());
-    assertTrue(findText.checkNextPageButtonIsEnabled());
-    assertFalse(findText.checkPreviousPageButtonIsEnabled());
-    sumOfFoundFiles += findText.getFoundFilesNumberOnPage();
-    sumOfFoundOccurrences += findText.getFoundOccurrencesNumberOnPage();
+    Assert.assertTrue(findText.checkNextPageButtonIsEnabled());
+    Assert.assertFalse(findText.checkPreviousPageButtonIsEnabled());
+    sumOfFoundFiles += findText.getResults().getFoundFilesOnPage();
```
dmytro-ndp: It's better to set the returning by `findText.getResults()` object to the variable and then use it to get the numbers, to avoid multiple execution of commands inside the constructor `SearchFileResult()`.

dmytro-ndp: What about renaming **findText** variable to the more suitable one like **"testFinder"**, **"findTextPage"**?

dmytro-ndp (approved):

dmytro-ndp:
ci-build

tolusha (approved):

Issue #7067, opened 2017-10-30
Make bootstrapping of OpenShift machines parallel


sleshchenko:
OpenShift infrastructure starts machines parallel but bootstraps them one by one. Make bootstrapping parallel should speed up the start of multimachine workspaces.

skabashnyuk:
won't fix.  Current solution is good enough.

skabashnyuk:
The current level of parallelism  - is workspace. We can't just simple multiple numbers of running threads by a number of machines because this wouldn't give us any profit. Too many threads will concurrently fight with other resources like a disk or network io.  Making some limits below workspaces *  number of machines we are considering as too complex to implement.

gorkem:
@skabashnyuk what do you mean by `The current level of parallelism - is workspace.`. I thought this one was about starting containers on a multi-container workspace without waiting for each other.

PR #7379, opened 2017-11-15
Make bootstrapping of openshift machines parallel

mshaposhnik:
### What does this PR do?
Makes bootstrapping of openshift machines run in  parallel.

#7067


infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
```diff
@@ -139,6 +128,54 @@ protected void internalStart(Map<String, String> startOptions) throws Infrastruc
...
+                  }));
+    }
+
+    CompletableFuture<Void> allAvailable =
+        CompletableFuture.allOf(checkMachines.toArray(new CompletableFuture[0]));
+    try {
+      CompletableFuture.anyOf(allAvailable, firstFailed).join();
+    } catch (CompletionException e) {
+      if (firstFailed.isDone()) {
+        throw firstFailed.join();
```
sleshchenko: You expect here get InfrastructureException but you won't get it here. Because this completable future is complete exceptionally [here](https://github.com/eclipse/che/pull/7379/files#diff-fd5e0f37b880417aed392d5dc7020dd6R150).
And then as far as I understand `firstFailed.join();` will throw CompletionException caused by InfrastructureException.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
```diff
@@ -139,6 +128,54 @@ protected void internalStart(Map<String, String> startOptions) throws Infrastruc
...
+   */
+  private void bootstrapMachines() throws InfrastructureException, InterruptedException {
+    CompletableFuture<InfrastructureException> firstFailed = new CompletableFuture<>();
+    List<CompletableFuture> checkMachines = new ArrayList<>(machines.size());
+    for (OpenShiftMachine machine : machines.values()) {
+      checkMachines.add(
+          CompletableFuture.runAsync(
+                  () -> {
+                    try {
+                      machine.waitRunning(machineStartTimeoutMin);
```
sleshchenko: It's needed to rethrow `InterruptedException` while waiting of machine instead of wrapping it into `InfrastructureException`.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
```diff
@@ -139,6 +128,54 @@ protected void internalStart(Map<String, String> startOptions) throws Infrastruc
...
+                      sendRunningEvent(machine.getName());
+                    }
+                  }));
+    }
+
+    CompletableFuture<Void> allAvailable =
+        CompletableFuture.allOf(checkMachines.toArray(new CompletableFuture[0]));
+    try {
+      CompletableFuture.anyOf(allAvailable, firstFailed).join();
+    } catch (CompletionException e) {
```
sleshchenko: It's needed to cancel all start tasks when first is failed.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
```diff
@@ -139,6 +128,54 @@ protected void internalStart(Map<String, String> startOptions) throws Infrastruc
...
+   *
+   * @throws InfrastructureException when any error occurs while bootstrapping machines
+   * @throws InterruptedException when machines bootstrapping was interrupted
+   */
+  private void bootstrapMachines() throws InfrastructureException, InterruptedException {
+    CompletableFuture<InfrastructureException> firstFailed = new CompletableFuture<>();
+    List<CompletableFuture> checkMachines = new ArrayList<>(machines.size());
+    for (OpenShiftMachine machine : machines.values()) {
+      checkMachines.add(
+          CompletableFuture.runAsync(
```
sleshchenko: Mb it would be better to use separated ThreadPool for bootstrapping of OpenShift machines

sleshchenko (changes requested):
It should be reworked to process Interrupted and Infrastructure exceptions which can occur while bootstrapping of machines.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
mshaposhnik: Can you show where is wrapped? Test shows it rethrown correctly;

sleshchenko: Here https://github.com/eclipse/che/blob/f9381a6533fba905b73e311802a4a1802761bfd0/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/project/OpenShiftPods.java#L184-L186

mshaposhnik: How it's related to my PR ?

sleshchenko: If `Thread` will be interrupted while waiting running you will rethrow the wrong exception.

mshaposhnik: That is the problem of `wait()` method i think.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
```diff
@@ -139,6 +128,54 @@ protected void internalStart(Map<String, String> startOptions) throws Infrastruc
...
+                  () -> {
+                    try {
+                      machine.waitRunning(machineStartTimeoutMin);
+                      bootstrapMachine(machine);
+                      checkMachineServers(machine);
+                    } catch (InfrastructureException ex) {
+                      sendFailedEvent(machine.getName(), ex.getMessage());
```
sleshchenko: I think so. And you can fix it in your PR, it's so easy I think. Or create a new issue for that.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
sleshchenko: I think so. And you can fix it in your PR, it's so easy I think. Or create a new issue for that.

mshaposhnik: Uff. You're confused me strongly. I thought you mean that  _MY_ code changes the ex.

mshaposhnik:
@sleshchenko pls take a look, i have removed `firstFailed` at all and do running via pool which is shutdown in case of any error.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
```diff
@@ -139,6 +130,48 @@ protected void internalStart(Map<String, String> startOptions) throws Infrastruc
...
+                    try {
+                      machine.waitRunning(machineStartTimeoutMin);
+                      bootstrapMachine(machine);
+                      checkMachineServers(machine);
+                    } catch (InfrastructureException ex) {
+                      executorService.shutdownNow();
+                      sendFailedEvent(machine.getName(), ex.getMessage());
+                      throw new CompletionException(ex);
+                    } catch (InterruptedException ei) {
+                      executorService.shutdownNow();
```
sleshchenko: Make sure that you don't have to call `Thread.currentThread().interrupt();` here

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
```diff
@@ -139,6 +130,48 @@ protected void internalStart(Map<String, String> startOptions) throws Infrastruc
...
 
+  /**
+   * Bootstraps machines in parallel.
+   *
+   * @throws InfrastructureException when any error occurs while bootstrapping machines
+   * @throws InterruptedException when machines bootstrapping was interrupted
+   */
+  private void bootstrapMachines() throws InfrastructureException, InterruptedException {
+    List<CompletableFuture> checkMachines = new ArrayList<>(machines.size());
+    final ExecutorService executorService = Executors.newCachedThreadPool();
```
sleshchenko: I see that using ExecutorService is maybe the easiest way to implement task interruption, but I'm not sure that it's a good idea to create new executor for each workspace start.
BTW it should have the naming format.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
```diff
@@ -139,6 +130,48 @@ protected void internalStart(Map<String, String> startOptions) throws Infrastruc
...
+                      executorService.shutdownNow();
+                      sendFailedEvent(machine.getName(), ex.getMessage());
+                      throw new CompletionException(ex);
+                    } catch (InterruptedException ei) {
+                      executorService.shutdownNow();
+                      throw new CompletionException(ei);
+                    }
+                  },
+                  executorService)
+              .thenAccept(a -> sendRunningEvent(machine.getName())));
```
sleshchenko: Are you sure that you need setting `thenAccept` here? Looks like you can put sending running event just after servers checking line. I mean 
```
try {
  machine.waitRunning(machineStartTimeoutMin);
  bootstrapMachine(machine);
  checkMachineServers(machine);
  sendRunningEvent(machine.getName());
} catch ...
```

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
mshaposhnik: Yes it will call `interrupt()` for all workers.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
mshaposhnik: Makes sense

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
sleshchenko: Are you going to throw `InterruptedException` from `OpenShiftPods#wait` method?

mshaposhnik: I don't know yet. I'm not familiar with that code, & don't see selation to the current subject.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
mshaposhnik: So why than you suggested this in your [comment](https://github.com/eclipse/che/pull/7379#discussion_r151351793)?

sleshchenko: I meant using one Executor to OpenShiftInternalRuntime class (like WorkspaceSharedPool) but not one for each workspace start. Sorry that it wasn't clear enough.

mshaposhnik:
@sleshchenko renewed

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
```diff
@@ -139,6 +132,66 @@ protected void internalStart(Map<String, String> startOptions) throws Infrastruc
...
+    List<CompletableFuture> checkMachines = new ArrayList<>(machines.size());
+    List<BootstrapMachineTask> tasks = new ArrayList<>(machines.size());
+
+    for (OpenShiftMachine machine : machines.values()) {
+      BootstrapMachineTask task = new BootstrapMachineTask(machine);
+      tasks.add(task);
+      checkMachines.add(CompletableFuture.runAsync(task, pool.getExecutor()));
+    }
+    try {
+      CompletableFuture.allOf(checkMachines.toArray(new CompletableFuture[0])).join();
```
garagatyi: Consider using exact array size since it is easy to get it, e.g. 
```
new CompletableFuture[checkMachines.size()]
```

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
```diff
@@ -139,6 +132,66 @@ protected void internalStart(Map<String, String> startOptions) throws Infrastruc
...
+    } catch (CompletionException e) {
+      tasks.forEach(BootstrapMachineTask::interrupt);
+      if (e.getCause() instanceof InterruptedException) {
+        throw (InterruptedException) e.getCause();
+      } else if (e.getCause() instanceof InfrastructureException) {
+        throw (InfrastructureException) e.getCause();
+      } else {
+        throw new InfrastructureException(e.getCause().getMessage(), e.getCause());
```
garagatyi: I would say that in this situation InternalRuntimeException is better. @mshaposhnik @sleshchenko WDYT?

mshaposhnik: Didi you mean InternalInfrastructureException?

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
```diff
@@ -143,6 +138,67 @@ protected void internalStart(Map<String, String> startOptions) throws Infrastruc
...
+   * @throws InterruptedException when machines bootstrapping was interrupted
+   */
+  private void bootstrapMachines() throws InfrastructureException, InterruptedException {
+    List<CompletableFuture> checkMachines = new ArrayList<>(machines.size());
+    List<BootstrapMachineTask> tasks = new ArrayList<>(machines.size());
+
+    for (OpenShiftMachine machine : machines.values()) {
+      BootstrapMachineTask task = new BootstrapMachineTask(machine);
+      tasks.add(task);
+      checkMachines.add(CompletableFuture.runAsync(task, pool.getExecutor()));
```
akorneta: use pool.[runAsync(...)](https://github.com/eclipse/che/blob/che6/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceSharedPool.java#L105) instead

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
garagatyi: Yes, sorry for confusing.

akorneta (approved):

mshaposhnik:
Delete branch ?

PR #7401, opened 2017-11-16
Declare factory.find as Deprecated, print warning if different than creator.userId  or name

skabashnyuk:
### What does this PR do?
Declare factory.find on unlimited number of  attributes as Deprecated

This feature has an unpredictable performance there is no guarantee that for a given number of attributes will be a database index. I would like to declare this method as Deprecated.Then collect feedback on which attributes search is useful. Then we either create new api for that or limit the scope of the current method but make sure performance is optimized for them.
For now I saw that used search by two attributes ```creator.userId``` and ```name```

#### Release Notes
Declare factory.find on unlimited number of  attributes as Deprecated

#### Docs PR
n/a


benoitf:
it looks like strange to deprecate find while it's still used with userId
deprecation policy implies that you can only deprecate when there is a replacement

could you make deprecate only if attribute is not including `creator.userId` or `name` ?

skabashnyuk:
> could you make deprecate only if attribute is not including creator.userId or name 

yes. But I'm not sure that we are not using something else. My motivation was to declare and collect feedback.  The limitation I was going to make in next release.

benoitf:
my concern is that it's strange if when UD is making requests on factories we see warnings on console. IMHO warnings should appear only for cases that you don't know.

skabashnyuk:
You think it's a bad idea to:
1. 6.0.0-M2 - add a warning for all parameter. It explains that this method is going to be refactored.
2. 6.0.0-M3 - Keep warn for creator.userId or name, all others throw an error.
2. 6.0.0-M4 - Add new api for creator or name search. Or decided to keep current wiht M3 limitation.

vparfonov (approved):

benoitf:
@skabashnyuk yes because for me it's bad to print warnings when we know some of them will be printed when using Che normally.

I would better
- M2 : print warning if different than `creator.userId` or `name`: so you can really track the behavior that you don't know.
- M3 : error instead of warning
- M4 : based on feedback you only allow search by userid/name or another way

skabashnyuk:
```
2017-11-16 20:51:14,387[nio-8080-exec-7]  [WARN ] [o.e.c.a.f.s.FactoryService 223]      - Method factory.find is going to be removed or limited in scope in 6.0 GA. Requested attributes [{first=version, second=che}], skipCount [token, skipCount, maxItems], maxItems 30

```

benoitf:
thanks for the output @skabashnyuk 

Looking at UD source code
we have
https://github.com/eclipse/che/blob/a22f44294b7bc349d99f97c5c4c17a2dfcd92201/dashboard/src/components/api/che-factory.factory.ts#L79
https://github.com/eclipse/che/blob/a22f44294b7bc349d99f97c5c4c17a2dfcd92201/dashboard/src/components/api/che-factory.factory.ts#L103-L109

so I suppose pagination elements (like `skipCount` and `maxItems`) should not be included as part of the warning as well

And there is the `token` attribute which is not targeting this method directly so it should be ignored.

skabashnyuk:
`so I suppose pagination elements (like skipCount and maxItems) should not be included as part of the warning as well`
@benoitf 
They are not included since they are skipped in a https://github.com/eclipse/che/pull/7401/files#diff-fc34b5aa12e2c9b6e3b005a8987613ccR214

benoitf (approved):
thx

Issue #4636, opened 2017-03-30
Variable values can't be reviewed when IDE steps into JRE class for debug


dmytro-ndp:
**Reproduction Steps:**
1. Create and open workspace from java stack.
1. Import project https://github.com/dmytro-ndp/java-with-external-libs.git and configure it as maven project.
1. Add command "**build and debug**" to the IDE: `mvn -f ${current.project.path} clean install && java -jar -Xdebug -Xnoagent -Djava.compiler=NONE -Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=8000 ${current.project.path}/target/java-with-external-libs-1.0-SNAPSHOT-jar-with-dependencies.jar`
1. Open **SimpleLogger.java** file in editor.
1. Set debug breakpoint to the line 13 with command `javaLogger.log(Level.INFO, "Info from java logger");`.
1. Execute command "**build and debug**".
1. Call top-menu command "_Run > Debug > Remote Java_".
1. Wait until execution is stopped at the breakpoint.
1. Go to _Debug Panel_ and call "**Step Into**" command.

**Expected behavior**
IDE opens class **Logger** and there are values of variables "**msg**" and "**level**" in _Variables Panel_ list.
_Evaluate Expression_ popup displays values of "**msg**" and "**level**" variables.

**Actual wrong behavior**
There were no values of variables "**msg**" and "**level**" in _Variables Panel_ list.
_Evaluate Expression_ popup displays next error message when requested values of "**msg**" and "**level**" variables:
> Error: Failed to evaluate an expression.
> Reason: {}

Server returns _500 Internal Server Error_ on address http://172.19.20.32:33739/api/debugger/6/evaluation?expression=msg

![screenshot from 2017-03-30 12 06 11](https://cloud.githubusercontent.com/assets/1197777/24496970/8ccc45c6-1542-11e7-9b61-7a1df0429279.png)

**Screencast**
![out](https://cloud.githubusercontent.com/assets/1197777/24497017/b5cced90-1542-11e7-9943-1e7708d4fc67.gif)

**Error reason**
There was **org.eclipse.che.plugin.jdb.server.expression.ExpressionException** at the server caused by **com.sun.jdi.AbsentInformationException**.

[...]

PR #7410, opened 2017-11-16
CHE-4636: Remove usecase for unexisted feature

tolusha:

### What does this PR do?
Removes selenium test case for unexisted feature.

https://github.com/eclipse/che/issues/4636

### DIFF
--- a/selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/debugger/DebugExternalClassTest.java
+++ b/selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/debugger/DebugExternalClassTest.java
@@ -121,35 +121,6 @@ public class DebugExternalClassTest {
   }
 
   @Test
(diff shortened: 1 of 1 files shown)


tolusha:
ci-build

vparfonov (approved):

dmytro-ndp:
@tolusha: can you, please, explain which feature you treated as inexistent? Because of title and description of the PR say nothing about it.

tolusha:
We can't figure out debug information if class isn't compiled with `-g` flag.

dmytro-ndp:
I see. If user still have an ability to go inside the JRE class during the debug session, debug of JRE class, IMHO, is existed use case which can and ought to be tested, at least, to check that server is returning meaningful error message, instead of:
> Error: Failed to evaluate an expression.
> Reason: {}

tolusha:
Any way no need to test this in selenium. I has to be a pure server side test.

PR #7417, opened 2017-11-16
Do not pull all CHE images on CLI init if --skip:pull specified

riuvshin:
Do not pull all CHE images on CLI `init` if `--skip:pull` specified
this is very annoying when user wants perform ONLY cli `config` action or `init` it always pull all the images if they are missing. but in fact it does not need it.

### DIFF
--- a/dockerfiles/base/scripts/base/commands/cmd_init.sh
+++ b/dockerfiles/base/scripts/base/commands/cmd_init.sh
@@ -73,7 +73,9 @@ cmd_init() {
     warning "($CHE_MINI_PRODUCT_NAME init): 'nightly' installations cannot be upgraded to non-nightly versions"
   fi
 
-  cmd_lifecycle download $FORCE_UPDATE
+  if ! skip_pull && ! is_fast; then
+      cmd_lifecycle download $FORCE_UPDATE
+  fi
 
   if require_license; then
     if [[ "${AUTO_ACCEPT_LICENSE}" = "false" ]]; then
--- a/dockerfiles/base/scripts/base/commands/cmd_init.sh
+++ b/dockerfiles/base/scripts/base/commands/cmd_init.sh
@@ -73,7 +73,7 @@ cmd_init() {
     warning "($CHE_MINI_PRODUCT_NAME init): 'nightly' installations cannot be upgraded to non-nightly versions"
   fi
 
-  if ! skip_pull && ! is_fast; then
+  if ! skip_pull; then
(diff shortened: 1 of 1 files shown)


benoitf (commented):
What is happening on the first launch if fast is specified ?

riuvshin:
@benoitf first launch of what? 
that were added I think to pre pull everything before launch and this allows you to skip pre pull. So docker needed image will be pulled automatically on docker run where it is needed. 
for example if I want only generate configurations it will pull only init image.

benoitf:
@riuvshin 

I'm a fresh user, no docker images
I do : 
`$ docker run .... eclipse/che start --fast`

it will fail (because start call init)

also there is already some conditional check there
https://github.com/eclipse/che/blob/30a03af5fd3b9e8f5eff73f950473d666a427ade/dockerfiles/base/scripts/base/commands/cmd_init.sh#L38-L42

I would say you're more interested in a `--skip-download` and gives this flag to download command which will do a no-op

garagatyi (approved):

sleshchenko (approved):
I'm not very familiar with CLI source code so I can't review your changes well.
It looks good to me. But up to other reviewers to merge it or not :smile:

riuvshin:
@benoitf are you sure that it will fail? it can't fail due to not pulled image. If image is not there docker will pull it automatically. And actually `download` does `docker pull` so skip pull is exactly what I  want here :)

riuvshin:
@benoitf I've just tested it works well for your use case of new user with no images. It works even better with `--fast` because user don't need to wait while unnecessary images will be pulled for his action. Like in my case --fast worked really faster because I didn't spend time on waiting for pulling che multi user as im running single user. 
So why don't we remove that `pre pull`  at all? I believe that it is absolutely unnecessary and redundant

benoitf:
@riuvshin it will fail as we capture output of docker run. if image is not there previously we will get traces about pulling images so parsing output will fail.

riuvshin:
> @riuvshin it will fail as we capture output of docker run. if image is not there previously we will get traces about pulling images so parsing output will fail. 

Can you please show that code? 
It perfectly works, I've just tested.

benoitf:
I will test later on that.
But coming back to the cmd_download, AFAIK it should still be applied if `--fast` is used

It's changing too many stuff in the logic to bypass from init phase as it's used by most of the commands

riuvshin:
@benoitf I;ve removed --fast from that

benoitf (approved):

Issue #4211, opened 2017-02-22
Debugger doesn't activate breakpoints in inner classes


dmytro-ndp:
Java console project can be found [here](https://github.com/dmytro-ndp/java-inner.git). There are varouse type of inner classes with methods where breakpoints aren't activated in runtime after start of debug session.

**Reproduction Steps:**
1. Open workspace and import test project https://github.com/dmytro-ndp/java-inner as maven project.
1. Add maven command "**build-and-debug**": `clean install && java -jar -Xdebug -Xrunjdwp:transport=dt_socket,address=8000,server=y,suspend=y ${current.project.path}/target/*.jar`
1. Open "**src/main/java/test/App.java**" file inside editor and set next breakpoints:
\- _anonymous class_: at line **26** in the method _getAnonymValue()_
\- _method local inner class_: at line **42** in the method _App.getAppValue().MethodLocalInner.getValue()_;
\- _inner class_: at line **53** in the method _App.Inner.getInnerValue()_;
\- _static inner class_: at line **61** in the method _App.StaticInner.getStaticInnerValue()_.
1. Call top menu command "_Run > Edit Debug Configurations..._" and create _Java Remote configuration_ with port=8000.
1. Invoke "**build-and-debug**" command and wait until there is "Listening for transport dt_socket at address: 8000" message in console.
1. Call top menu command "_Run > Debug > Remote java_"

***Expected behavior:***
All breakpoints become an active, and Che is break executing of application at the next line one-by-one: 26, 42, 53, 61

***Actual wrong behaviour:***
Breakpoints didn't become an active, and Che executed an application without stops.

***Screencast:***
![out](https://cloud.githubusercontent.com/assets/1197777/23223321/6eeed9ea-f933-11e6-842c-d01626f2c49d.gif)

***Selenium tests:***
There is couple selenium tests which should be enabled after this bug is fixed:
- InnerClassAndLambdaDebuggingTest.**shouldDebugMethodLocalInnerClass()**
- InnerClassAndLambdaDebuggingTest.**shouldDebugInnerClass()**
- InnerClassAndLambdaDebuggingTest.**shouldDebugStaticInnerClass()**

**OS and version:**    
Ubuntu 14.04 LTS
Browser: Chrome Version 55.0.2883.87 (64-bit)
Che version: 5.3.1

How to fix:
Analyse parents of `iMember`
[...]

PR #7431, opened 2017-11-17
CHE-4211: Fix InnerClassAndLambdaDebuggingTest test

tolusha:

### What does this PR do?
* Fixes java debugger server side part concerning adding breakpoints into local classes
* Removes checking of unexisted functionality of the eclipse che debugger

https://github.com/eclipse/che/issues/4211

### DIFF
--- a/selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/debugger/InnerClassAndLambdaDebuggingTest.java
+++ b/selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/debugger/InnerClassAndLambdaDebuggingTest.java
@@ -140,7 +140,6 @@ public class InnerClassAndLambdaDebuggingTest {
 
     // then
     editor.waitActiveBreakpoint(53);
-    debugPanel.waitTextInVariablesPanel("methodValue=\"App method local inner test\"");
   }
 
   @Test(priority = 2)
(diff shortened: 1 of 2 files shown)


vparfonov (approved):

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/debugger/InnerClassAndLambdaDebuggingTest.java
```diff
@@ -140,7 +140,6 @@ public void shouldDebugMethodLocalInnerClass() {
 
     // then
     editor.waitActiveBreakpoint(53);
-    debugPanel.waitTextInVariablesPanel("methodValue=\"App method local inner test\"");
```
dmytro-ndp, line 143: Could you, please, explain, why did you remove this check? Does it mean that there will be no values inside the _Variables Panel_?

tolusha: Why did you decide that no variables will be inside the Variable Panel?
It means that `methodValue` is unavailable only.

dmytro-ndp: But **methodValue** variable should be there in _Variable Panel_ so as it is accessible from inside the **MethodLocalInner::getValue()** method and is returning by this method.
```javascript
    public String getAppValue() {
        final String methodValue = "App method local inner test";

        class MethodLocalInner {
            String getValue() {
line 53:      return methodValue;
            }
        }

        return new MethodLocalInner().getValue();
    }
```

tolusha: Sure, but the reference Eclipse IDE doesn't show that variable too. I think that intelejii does some trick to show that.

dmytro-ndp: You are right :-(
screenshot from 2017-11-17 16-57-34

Tested on Eclipse IDE for Java Developers, Version: Neon Release (4.6.0)

PR #7440, opened 2017-11-17
Update version of openjdk base image to 8u131

l0rd:
### What does this PR do?
Update the base Docker image to `openjdk:8u131` (was `8u111` before)

Version `8u111` was a quite old version (1yr old)

### DIFF
--- a/dockerfiles/che/Dockerfile
+++ b/dockerfiles/che/Dockerfile
@@ -19,7 +19,7 @@
 #             -v /home/user/che/storage:/home/user/che/storage \
 #             codenvy/che-server
 #           
-FROM openjdk:8u111-jre-alpine
+FROM openjdk:8u131-jre-alpine
 
 ENV LANG=C.UTF-8 \
     DOCKER_VERSION=1.6.0 \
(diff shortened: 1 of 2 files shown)


riuvshin (changes requested):
unfortunately we can't update openJDK version just like that. Take a look on https://github.com/eclipse/che/blob/che6/dockerfiles/che/open-jdk-source-file-location

That file describes how OpenJDK was built on parent image which was requested by RH legal department. 
So that file must be updated to represent how newer openJDK version was built

l0rd:
@riuvshin I have updated both PRs (this one and https://github.com/eclipse/che/pull/7439)

riuvshin (approved):

riuvshin:
@l0rd AFAIK we also used exactly same version of JDK for workspaces images and agents that will install JDK if it is missing to keep ws master and agent on same version of JDK. Im not sure if we should update it also... @skabashnyuk WDYT?

l0rd:
@riuvshin it doesn't look like it's the same version https://github.com/eclipse/che-dockerfiles/blob/master/recipes/stack-base/ubuntu/Dockerfile#L39. 

Said that I think we should update ubuntu stacks with the latest version of openjdk available for that distribution (i.e. `8u151-b12-0ubuntu0.16.04.2`).

PR #7446, opened 2017-11-20
Add an ability to use additional volumes in machines

### LINKED ISSUE #7010: Add volumes to machine configuration
Since in Che6 we move to multicontainer/sidecar way it become more important to be able to describe "interesting for workspace" persistence storage as a part of workspace configuration.

Currently, there are two ways to describe in a case of attaching (mounting, syncing etc) volumes (folders) with persisted data:
- /projects folder attached to "ws-agent" machine is a sign to attach identified persistence storage to it
- everything else (like an arbitrary PVC) described in an infrastructure specific recipe (docker volume mount,  PVC etc).

It is proposed to add "volumes" section into workspace->environment->machine config, like:

```
 volumes:
    <name>:
        path:</path/to/machine's/folder>
        properties:{} - do we need that as option?
```
 
@garagatyi @sleshchenko @skabashnyuk

garagatyi:
### What does this PR do?
I recommend reviewing this PR commit by commit. 
Adds an ability to specify additional volumes in Machine configuration (inside of Workspace configuration).
Che6 model of MachineConfig: 
```
{
    "servers" : ...
    "attributes" : ...
    "env" : ...
    "installers" : ...
}
```
Model of MachineConfig in this PR: 
```
{
    "servers" : ...
    "attributes" : ...
    "env" : ...
    "installers" : ...
    "volumes" : {
        "myvolume" : {
            "path" : "/some/path/in/container"
        }
    }
}
```
where field `volumes` contains volumes names as keys and volume object as value. 
Volume object contains the path of volume in the machine.
Here are some additional details that may be useful to write docs and implement UI for volumes:
- field 'path' in volume object contains the absolute path in the machine, so should start from '/'
- folders that don’t exist in the container before machine start get root access level which may lead to some errors, issues for a user. A user might want to create appropriate folders in the image before adding mounts for them.
- this PR doesn't provide UI for adding volumes, but they can be added using API or editing workspace config on UD.
- when volume with the same name is used in different machines it is treated as a volume shared between those machines. This provides sharing data between machines.
- volumes in the environment recipe are still not allowed.
- there is no sharing of volumes between workspaces since they are workspace-scoped. 

Further improvements:
- this PR doesn't add volumes to installers but it might be useful because in this case, we won't need code that adds projects volume to a machine with the ws-agent server - an installer will request volumes it needs. Should we create an issue for this change?
- maybe we should declare volumes on the environment level to declare these volumes and use it in machines. This would more explicitly show sharing of volumes between machines. Should we do that?
- there is no ability to mount a particular folder on Docker host to a machine volume for now. It is not clear whether it is needed and how this fits cloud-native tooling approach. 
- do we want to add support for usage of volumes from the recipe? What pros and cons?
- add UI for volumes management.
- add warnings if volumes conf may prevent runtime start. This is interesting for the OpenShift environment in particular. For example, volumes declared shared between machines from different pods may lead to runtime strat failure in most cases. Whether start will fail or not depends on read-write access settings of OpenShift volumes infrastructure, pods scheduling. 

Fixes #7010

### DIFF
--- a/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/workspace/model/MachineConfigImpl.java
+++ b/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/workspace/model/MachineConfigImpl.java
@@ -17,6 +17,7 @@ import java.util.Map;
 import java.util.stream.Collectors;
 import org.eclipse.che.api.core.model.workspace.config.MachineConfig;
 import org.eclipse.che.api.core.model.workspace.config.ServerConfig;
+import org.eclipse.che.api.core.model.workspace.config.Volume;
 
 public class MachineConfigImpl implements MachineConfig {
 
@@ -24,12 +25,14 @@ public class MachineConfigImpl implements MachineConfig {
   private Map<String, String> env;
   private Map<String, String> attributes;
   private Map<String, ServerConfigImpl> servers;
+  private Map<String, VolumeImpl> volumes;
 
   public MachineConfigImpl(
       List<String> installers,
       Map<String, ? extends ServerConfig> servers,
       Map<String, String> env,
-      Map<String, String> attributes) {
+      Map<String, String> attributes,
+      Map<String, ? extends Volume> volumes) {
     if (installers != null) {
       this.installers = new ArrayList<>(installers);
     }
@@ -48,10 +51,23 @@ public class MachineConfigImpl implements MachineConfig {
     if (attributes != null) {
       this.attributes = new HashMap<>(attributes);
     }
+    if (volumes != null) {
+      this.volumes =
+          volumes
+              .entrySet()
+              .stream()
+              .collect(
+                  Collectors.toMap(Map.Entry::getKey, entry -> new VolumeImpl(entry.getValue())));
+    }
   }
 
   public MachineConfigImpl(MachineConfig machine) {
-    this(machine.getInstallers(), machine.getServers(), machine.getEnv(), machine.getAttributes());
+    this(
+        machine.getInstallers(),
+        machine.getServers(),
+        machine.getEnv(),
+        machine.getAttributes(),
+        machine.getVolumes());
   }
 
   @Override
@@ -87,6 +103,14 @@ public class MachineConfigImpl implements MachineConfig {
   }
 
   @Override
+  public Map<String, VolumeImpl> getVolumes() {
+    if (volumes == null) {
+      volumes = new HashMap<>();
+    }
+    return volumes;
+  }
+
+  @Override
   public boolean equals(Object obj) {
     if (this == obj) {
       return true;
@@ -98,7 +122,8 @@ public class MachineConfigImpl implements MachineConfig {
     return getInstallers().equals(that.getInstallers())
         && getEnv().equals(that.getEnv())
         && getAttributes().equals(that.getAttributes())
-        && getServers().equals(that.getServers());
+        && getServers().equals(that.getServers())
+        && getVolumes().equals(that.getVolumes());
   }
 
   @Override
@@ -108,6 +133,7 @@ public class MachineConfigImpl implements MachineConfig {
     hash = 31 * hash + getEnv().hashCode();
     hash = 31 * hash + getAttributes().hashCode();
     hash = 31 * hash + getServers().hashCode();
+    hash = 31 * hash + getVolumes().hashCode();
     return hash;
   }
 
@@ -122,6 +148,8 @@ public class MachineConfigImpl implements MachineConfig {
         + attributes
         + ", servers="
         + servers
+        + ", volumes="
+        + volumes
         + '}';
   }
 }
--- a/infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/OnWorkspaceRemoveDataVolumeRemover.java
+++ b/infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/OnWorkspaceRemoveDataVolumeRemover.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright (c) 2012-2017 Red Hat, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.workspace.infrastructure.docker;
+
+import static java.lang.String.format;
+import static org.slf4j.LoggerFactory.getLogger;
+
+import java.io.IOException;
+import java.util.List;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import org.eclipse.che.api.core.notification.EventService;
+import org.eclipse.che.api.core.notification.EventSubscriber;
+import org.eclipse.che.api.workspace.shared.event.WorkspaceRemovedEvent;
+import org.eclipse.che.infrastructure.docker.client.DockerConnector;
+import org.eclipse.che.infrastructure.docker.client.json.volume.Volume;
+import org.eclipse.che.workspace.infrastructure.docker.provisioner.volume.VolumeNames;
+import org.slf4j.Logger;
+
+/** @author Alexander Garagatyi */
+@Singleton
+public class OnWorkspaceRemoveDataVolumeRemover implements EventSubscriber<WorkspaceRemovedEvent> {
+
+  private static final Logger LOG = getLogger(OnWorkspaceRemoveDataVolumeRemover.class);
+
+  private final DockerConnector docker;
+
+  @Inject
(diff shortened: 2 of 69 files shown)


garagatyi:
ci-test

riuvshin (approved):

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/workspace/model/MachineConfigImpl.java
```diff
@@ -17,19 +17,22 @@
...
 import org.eclipse.che.api.core.model.workspace.config.ServerConfig;
+import org.eclipse.che.api.core.model.workspace.config.Volume;
 
 public class MachineConfigImpl implements MachineConfig {
 
   private List<String> installers;
   private Map<String, String> env;
   private Map<String, String> attributes;
   private Map<String, ServerConfigImpl> servers;
+  private Map<String, Volume> volumes;
```
sleshchenko: I think `VolumeImpl` should be here

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/provisioner/volume/VolumeNameGenerator.java
```diff
@@ -0,0 +1,39 @@
...
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.workspace.infrastructure.docker.provisioner.volume;
+
+/**
+ * Helps to generate volume names and matches existing volumes to workspace.
+ *
+ * @author Alexander Garagatyi
+ */
+public class VolumeNameGenerator {
```
sleshchenko: I'd propose to rework this class in the following way:
```java
...
public class VolumeNames {

  public static String generate(String workspaceId, String originVolumeName)...

  public static boolean matches(String volumeName, String workspaceId)...

  public static boolean matches(String volumeName, String originVolumeName, String workspaceId)...
  
  ....
}
```
WDYT?

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/workspace/model/MachineConfigImpl.java
garagatyi: +1

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/provisioner/volume/VolumeNameGenerator.java
garagatyi: +1

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/project/pvc/UniqueWorkspacePVCStrategy.java
```diff
@@ -59,41 +61,50 @@ public UniqueWorkspacePVCStrategy(
...
+    if (machineConfig == null || machineConfig.getVolumes().isEmpty()) {
+      return;
+    }
+    for (Entry<String, Volume> volumeEntry : machineConfig.getVolumes().entrySet()) {
+      String volumeName = volumeEntry.getKey();
+      String volumePath = volumeEntry.getValue().getPath();
+      String subPath = workspaceId + '/' + volumeName;
+      String pvcUniqueName = pvcName + '-' + workspaceId + '-' + volumeName;
+
+      container.getVolumeMounts().add(newVolumeMount(pvcUniqueName, volumePath, subPath));
```
sleshchenko: Should you add volume mount only if it doesn't exist? Or mb rewrite existing one

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/provisioner/volume/VolumeNameGenerator.java
garagatyi: done

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/workspace/model/MachineConfigImpl.java
garagatyi: done

garagatyi:
I significantly extended the description of this PR. @eivantsov @ashumilova feel free to ask for more details if anything additional is needed for users docs, support or UI.

garagatyi:
@gazarenkov FYI

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/local/projects/BindMountProjectsVolumeProvisioner.java
```diff
@@ -65,24 +69,26 @@ public ProjectsVolumeProvisioner(
...
 
-    DockerContainerConfig devMachineConfig = internalEnv.getContainers().get(devMachineName);
-    if (devMachineConfig == null) {
-      throw new InternalInfrastructureException(
-          format("Docker configuration for machine '%s' not found", devMachineName));
+      DockerContainerConfig value = containerEntry.getValue();
+      List<String> newVolumes = new ArrayList<>();
+      for (String volume : value.getVolumes()) {
+        String[] volumeSourceTarget = volume.split(":");
+        if (VolumeNames.matches(volumeSourceTarget[0], "projects", identity.getWorkspaceId())) {
```
sleshchenko: Maybe It'd be better to make `projects` as public constant of ProjectsVolumeForWsAgentProvisioner class

sleshchenko (approved):
Looks good to me! :+1: 
Thanks for providing changes in different commits!

ghost (approved):
+1 very important feature

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/local/projects/BindMountProjectsVolumeProvisioner.java
garagatyi: ok

ashumilova (approved):

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/local/projects/BindMountProjectsVolumeProvisioner.java
garagatyi: done

gazarenkov:
Description looks cool and corresponds to what we discussed, thanks

I think we need to support volumes declared in recipe (can definitely start with your things as it is).
Like (taking OpenShift based example):
- we try to identify if machine -> volume corresponds some of PVC from the recipe and (try to) provide it physically the way declared in PVC if so 
- otherwise we should try to add recipe specific volumes (PVC) as standalone volume, considering it as a detail which make environment work but not related to developer flow specifics .

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/OnWorkspaceRemoveDataVolumeRemover.java
```diff
@@ -0,0 +1,78 @@
...
+  @Inject
+  public OnWorkspaceRemoveDataVolumeRemover(DockerConnector docker) {
+    this.docker = docker;
+  }
+
+  @Override
+  public void onEvent(WorkspaceRemovedEvent event) {
+    String wsId = event.getWorkspace().getId();
+    try {
+      List<Volume> volumes = docker.getVolumes().getVolumes();
```
akorneta, line 45: call `docker.getVolumes().getVolumes()` looks strange for me 
consider to change it to something like: `docker.getVolumes().asList()` / `docker.getVolumes().values()` / `docker.getVolumes().items()`

garagatyi: I don't like it either. But in this case, I would have to customize POJO `Volumes` with annotations to customize parsing logic. And I don't want to add this complexity, so for me, the current code is an acceptable tradeoff.

garagatyi: in docker API:
get volumes returns
```
{
    "volumes": []
}
```

akorneta (approved):
gj 👍

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/OnWorkspaceRemoveDataVolumeRemover.java
akorneta: I see, let it be as it is

PR #7449, opened 2017-11-20
findResource method refactoring related to hierarchical load resourcetree

vzhukovs:
### What does this PR do?
This changes proposal modifies the load resources from the server. As Resource Manager uses internal storage for caching loaded file resources we need to take in account that situation, when user request from the server file for example with path `/project/a/b/file.ext` then resource manager should get single tree to the file path. It means that resource manager should call and cache (if need) tree `/project`, then tree `/project/a`, then tree `/project/a/b` and find `file.ext` inside `b` directory.

Back port of https://github.com/eclipse/che/pull/7250

#7128 

#### Changelog
findResource method refactoring related to hierarchical load resource tree

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/resources/impl/ResourceManager.java
+++ b/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/resources/impl/ResourceManager.java
@@ -11,7 +11,6 @@
 package org.eclipse.che.ide.resources.impl;
 
 import static com.google.common.base.Optional.absent;
-import static com.google.common.base.Optional.fromNullable;
 import static com.google.common.base.Optional.of;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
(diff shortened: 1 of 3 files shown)


ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/resources/impl/ResourceManager.java
```diff
@@ -944,28 +941,42 @@ public void visit(Resource resource) {
...
-              return promises.resolve(absent());
-            });
+  private Promise<Optional<Resource>> findResource(final Path absolutePath, boolean quiet) {
+    String[] segments = absolutePath.segments();
+
+    Promise<Optional<Resource>> chain = promises.resolve(null);
+
+    for (int i = 0; i <= segments.length; i++) {
+      Path pathToRetrieve = absolutePath.removeLastSegments(segments.length - i);
+      chain = chain.thenPromise(__ -> doFindResource(pathToRetrieve));
```
vparfonov: Try to find more informative name than "__"

vzhukovs: It's ignored value, `ignored` or `arg` name also looks not so good.

ashumilova (approved):

PR #7452, opened 2017-11-20
Fix samples.json loading

### LINKED ISSUE #7057: Question : Regarding dotnet-web-simple not appear in Che 5.19.0
OS : 16.04.3
Che : 5.19.0
Mode : Multi-User and Single-user

**First Case**
i been using che from the first version till now, meaning that upgrade repeatly till latest version 5.19.0, but i still can't find the dotnet-web-simple.

**Second Case**
I try first done the installation on version 5.18.0 and upgrade to 5.19.0, sadly i still can't find the 
dotnet-web-simple.

**Third Case**
I deleted docker and all related files for Che, meaning that i done an installation from fresh, in this case, i can find dotnet-web-simple.

screenshot from 2017-10-28 00-18-32

screenshot from 2017-10-28 00-19-39

**so, what wrong with Che?**

riuvshin:
fixes: https://github.com/eclipse/che/issues/7057

We have an issue that new samples added to default samples.json in newversion won't be loaded after update. That PR makes sure that on each run of che-server it will use accurate samples right from the assembly.
Also it allows users to add their own samples in `templates` dir as separate files as we are now not deleting everything `rm -rf "${CHE_DATA}"/templates/*`

### DIFF
--- a/dockerfiles/che/entrypoint.sh
+++ b/dockerfiles/che/entrypoint.sh
@@ -311,11 +311,11 @@ init() {
     rm -rf "${CHE_DATA}"/stacks
   fi
 
-  if [[ ! -f "${CHE_DATA}"/templates/samples.json ]];then
-    rm -rf "${CHE_DATA}"/templates/*
-    mkdir -p "${CHE_DATA}"/templates
-    cp -rf "${CHE_HOME}"/templates/* "${CHE_DATA}"/templates
-  fi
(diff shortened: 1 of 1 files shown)


ghost (approved):

ghost:
@riuvshin should be issue the same PR against che6 branch?

benoitf (approved):

Issue #7451, opened 2017-11-20
Errors with indexing and editing files with Ü and other similar symbols


dkulieshov:
We have exception each time we try to index text file with mentioned symbol. We have blanks instead of mentioned symbols when we open such files in text editor.

dkulieshov:
Further investigation shows that the issue is caused by improper file encoding. Right now we support only `utf-8` and for files with that encoding such symbols are shown properly.

However if mentioned symbol Ü is contained in _litain_1_ encoded file (for example https://raw.githubusercontent.com/che-samples/web-java-spring-petclinic/master/src/main/resources/messages/messages_de.properties) we have errors.

PR #7456, opened 2017-11-20
Errors with indexing files with Ü and other similar symbols

dkulieshov:
Related issue: https://github.com/eclipse/che/issues/7451

This fixes errors when trying to index files with special characters like Ü.


vparfonov (approved):

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/search/server/impl/LuceneSearcher.java
```diff
@@ -437,9 +453,19 @@ public final void update(Path fsPath) throws ServerException {
   }
 
   private void doUpdate(Term deleteTerm, Path fsPath) throws ServerException {
+    if (!fsPath.toFile().exists()) {
+      return;
```
skabashnyuk, line 457: can you make one ```if``` instead of two?

skabashnyuk (approved):

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/search/server/impl/LuceneSearcher.java
dkulieshov: It is done on purpose because those conditions are not logically related.

Issue #4979, opened 2017-05-03
Inner packages are lost after rename (after refresh page they are restored )


Ohrimenko1988:
**Reproduction Steps:**
- Go to the spring project with 
- Create some packages (see attachment)
- Choose package and press "shift + F6" 
- **Important** In the opened window **clear** input field and type package name as in attachment
- Click "Ok" button
- Pay attention to result 

**Expected behavior:**
![anim](https://cloud.githubusercontent.com/assets/22724619/25653063/ecbbd926-2ff3-11e7-966f-abf6c4e067e1.gif)

**Observed behavior:**
![anim](https://cloud.githubusercontent.com/assets/22724619/25653088/09cc6eea-2ff4-11e7-88bf-d28baf8eff35.gif)

**Additional information:**
* Related selenium test is the _RenamePackageTest_

ghost:
Still an issue in 5.15-SNAPSHOT

riuvshin:
@vparfonov we have pretty high regression due to this, can we take this to current sprint?

PR #7458, opened 2017-11-20
Prevent NPE after applying the refactoring session

vzhukovs:
### What does this PR do?
Fix an issue with apply refactoring session. After applying the session server responses with bunch of changes (type, old path, new path), but in changes there a may be en nullable change, so this PR adds additional check for `null` in refactoring updater component.

#4979 

#### Changelog
Prevent NPE after applying the refactoring session

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/resources/impl/ResourceManager.java
+++ b/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/resources/impl/ResourceManager.java
@@ -39,7 +39,6 @@ import com.google.common.base.Optional;
 import com.google.inject.Inject;
 import com.google.inject.assistedinject.Assisted;
 import com.google.web.bindery.event.shared.EventBus;
-import elemental.util.ArrayOf;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


vzhukovs:
ci-build

vzhukovs:
ci-test

vzhukovs:
ci-build

vzhukovs:
ci-test

ashumilova (approved):

musienko-maxim (approved):

ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/resources/impl/ResourceManager.java
```diff
@@ -1171,45 +1170,38 @@ private Resource newResourceFrom(final ItemReference reference) {
   }
 
   protected Promise<ResourceDelta[]> synchronize(final ResourceDelta[] deltas) {
-    List<Promise<Void>> promisesToResolve = new ArrayList<>(deltas.length);
+    Promise<Void> chain = promises.resolve(null);
```
vparfonov, line 1173: I am not follow the logic here. 
`Promise<Void> chain = promises.resolve(null); `
then `chain = chain.thenPromise(something)`;
What you want to solve here?

vzhukovs: Okay, I'll explain. If we'll have the code like this:

```javascript
new Promise(function(resolve, reject) {
  setTimeout(() => resolve(1), 1000); // (*)
}).then(function(result) { // (**)
  alert(result); // 1
  return result * 2;
}).then(function(result) { // (***)
  alert(result); // 2
  return result * 2;
}).then(function(result) {
  alert(result); // 4
  return result * 2;
});
```

the `result` will sequentially pass to the next promise object. And chain will be strictly executed one by one. The same we have in current PR, we have a variable with promise and on the next iteration of the cycle we're creating a new promise and call then on it.

Visually it'll look like in the picture:



But if we'll have smth like this:

```javascript
let promise = new Promise(function(resolve, reject) {
  setTimeout(() => resolve(1), 1000);
});

promise.then(function(result) {
  alert(result); // 1
  return result * 2;
});

promise.then(function(result) {
  alert(result); // 1
  return result * 2;
});

promise.then(function(result) {
  alert(result); // 1
  return result * 2;
});
```

We'll have a situation, when the first promise is resolved ignoring other passed promises.



What about replacing collection with promises with the chain, there is a simple reason, we can call synchronize twice on the same folder and we can avoid simultaneous call on the same folder, we need to process two deltas one by one.

PR #7461, opened 2017-11-20
Add the ability to configure logback (MDC) to show `identity_id` and `X-Request-ID` in logs files when available.

sunix:
### What does this PR do?
Add the ability to configure logback to show `identity_id` and `X-Request-ID` in logs files when available.

- Add `X-Request-ID` http header in the logs through MDC
- Add identity_id from keycloack when available in the logs through MDC

https://github.com/redhat-developer/rh-che/issues/312

#### Release Notes

Add the ability to configure logback to show `identity_id` and `X-Request-ID` in logs files when available.

#### Docs PR

TODO

### DIFF
--- a/core/commons/che-core-commons-logback/pom.xml
+++ b/core/commons/che-core-commons-logback/pom.xml
@@ -27,6 +27,10 @@
             <artifactId>javax.inject</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-core</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
         </dependency>
--- a/core/commons/che-core-commons-logback/pom.xml
+++ b/core/commons/che-core-commons-logback/pom.xml
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) 2012-2017 Red Hat, Inc.
+    All rights reserved. This program and the accompanying materials
+    are made available under the terms of the Eclipse Public License v1.0
+    which accompanies this distribution, and is available at
+    http://www.eclipse.org/legal/epl-v10.html
+
+    Contributors:
+      Red Hat, Inc. - initial API and implementation
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>che-core-commons-parent</artifactId>
+        <groupId>org.eclipse.che.core</groupId>
+        <version>5.22.0-SNAPSHOT</version>
+    </parent>
+    <artifactId>che-core-commons-logback</artifactId>
+    <packaging>jar</packaging>
+    <name>Che Core :: Commons :: Commons Logback</name>
+    <dependencies>
+        <dependency>
+            <groupId>javax.inject</groupId>
+            <artifactId>javax.inject</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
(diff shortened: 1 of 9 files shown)


skabashnyuk:
identity_id and X-Request-ID - what is the usecase for this  value?
Shall we move this filter in separate artifact and reuse it on the agent as well?

sunix:
- `Identity_id` : to track the user related logs in a multi-user configuration
- `request_id` : to track a request that goes through multiple sub system

I'm going to move the filter for request id somewhere else

skabashnyuk:
Identity_id  - ok. What does this value ```X-Request-Id``` mean for Che?

sunix:
For Che it means that wherever the cluster environnement it is installed and where routers are using `X-Request-Id`, Che would be able to log these request-Ids out of the box

skabashnyuk:
What about the separate artifact and reuse it on the agent?
I thought about adding web application name to MDC too. It would be good place to do that. WDYT?

sunix:
OK I'm OK to create a separate artifact, I was looking at an existing one and https://github.com/eclipse/che/tree/master/core/commons/che-core-commons-j2ee/src/main/java/org/eclipse/che/filter was looking good by its name but it seems to be included only in `assembly-ide-war`. Let me know if you have a prefer place and name for this artifact.

sunix:
About the web application name, I'm not sure MDC is the right place: MDC works per thread and pretty much like a threadLocal would work.

skabashnyuk:
what if core/commons/che-core-commons-logback ?

sunix:
I've updated the PR but for some reasons i can't get into the multi-user app, will investigate tomorow

core/commons/che-core-commons-logback/pom.xml
```diff
@@ -0,0 +1,36 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>javax.servlet</groupId>
+            <artifactId>javax.servlet-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <scope>provided</scope>
```
skabashnyuk: this artifact is not provided. please remove this scope.

core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/filter/RequestIdLoggerFilter.java
```diff
@@ -0,0 +1,45 @@
...
+import java.io.IOException;
+import javax.servlet.Filter;
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import org.slf4j.MDC;
+
```
skabashnyuk, line 23: javadoc

skabashnyuk: What about test?

multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/KeycloakEnvironmentInitalizationFilter.java
```diff
@@ -103,6 +106,8 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
       }
     }
 
+    putIdentityIdToMDC(subject);
```
skabashnyuk: I think you can replace it with just ```MDC.put(IDENTITY_ID_MDC_KEY, subject.getUserId());```

skabashnyuk: wait. It should not be here. 
I propose to make a separate filter in core/commons/che-core-commons-logback/ that will put subject id in mdc.

sunix:
@skabashnyuk I've updated this PR can you check and approve if ok ? thanks

skabashnyuk (approved):

sunix:
thanks

skabashnyuk:
@sunix will you provide che6 pr?

PR #7471, opened 2017-11-20
Avoid executing chown command if the user is not a sudoer

l0rd:
### What does this PR do?
Avoid changing the ownership of folder `/projects/` in script `org.eclipse.che.exec.script.sh` if the user that is executing the script is not a sudoer. 

That is the last of a list of errors that have been reported [here](https://github.com/openshiftio/openshift.io/issues/531)


l0rd:
che6 PR created too https://github.com/eclipse/che/pull/7472

tolusha (approved):

Issue #4991, opened 2017-05-03
Investigate integration of clangd language server


slemeur:
To provide a better support for C/C++ project, we should study the integration of the clangd language server.

https://www.phoronix.com/scan.php?page=news_item&px=LLVM-Clang-Clangd-Server

maximilianodendahl:
great to see that progressing! 

I guess latest code is here: https://github.com/llvm-mirror/clang-tools-extra/tree/master/clangd

tolusha:
@evidolob 
WDYT?

evidolob:
Yes, we should look at this, but clangd is very limited for now, it's has on formatting stuff

maximilianodendahl:
I dont think this is true. I believe it already supports

* code completion: https://github.com/llvm-mirror/clang-tools-extra/commit/ca5f112a2e2f05d9a5cc8662ac740ff5a931d13c
* FixIts: https://github.com/llvm-mirror/clang-tools-extra/commit/6fbfc8451e5c1ce11298e2ec5b0f80cf02f78913

which is quite powerful already

evidolob:
Yes, my bad. I check only initial commit :(

AndrienkoAleksandr:
Result of investigation https://docs.google.com/document/d/1Hr1kYhJsPQoaNJXWkGfrtLG0RzvRXmkEUqD2V-fakM0/edit

TylerJewell:
This is great and thorough.  @bmicklea - perhaps send this link to our friends over at Silexica. They'll have comments for the analysis of @AndrienkoAleksandr

bmicklea:
Agreed, nice write up @AndrienkoAleksandr!  I've sent it to the Silexica team on Slack.

gazarenkov:
@vkuznyetsov so, investigation is done, the issue supposed to be closed?

hassec:
Hi, 
didn't want to open a new issue. 
Are there any plans to revisit this sometime soon?

Even an experimental C++ language server would be much better than none.

maximilianodendahl:
@hassec Silexica will start looking into this next month as part of our slx.cloud offering, hoping to have a first version ready quite soon. I'll keep you updated

@hkolvenbach FYI

hassec:
Hi @maximilianodendahl, 

just wondering if there has been any progess on this? :)

hkolvenbach:
[...]

PR #7516, opened 2017-11-22
ClangD support for Eclipse Che

hkolvenbach:
### What does this PR do?
This PR adds initial support for C/C++ Intellisense by using clangd as a language server. It supports starting the language server and connecting to the clangd in a hardcoded path within the docker image

There is still work to do, but I would like to use this PR to discuss how to best continue.
Open Questions:
- Which clangd binary to integrate? Stable release (5.0.1) or latest svn (6.x)
- How to provide the clangd binary? Who is maintaining the `eclipse/cpp_gcc` docker image and should we integrate clangd into it?
- How should we handle the compile_commands.json? ClangD expects this to do proper "gotoDefinition" handling, however this file needs to be generated e.g. by cmake and passed as an argument to the language server

ToDo:
- Che6 Integration

#4991


hkolvenbach:
@AndrienkoAleksandr should i squash or rewrite your commits to fix the ip validation error?

AndrienkoAleksandr:
@hkolvenbach  ok. You can do all necessary
.

tolusha:
Hello. Here is my humble opinions. 

> Which clangd binary to integrate? Stable release (5.0.1) or latest svn (6.x)

It depends. If the latest version provides considerably more functionality then it make sense, otherwise I would prefer the stable version.

> How to provide the clangd binary? 

We can host them as we host other binaries for language servers.

> Who is maintaining the eclipse/cpp_gcc docker image and should we integrate clangd into it?

I would say no.

> How should we handle the compile_commands.json? ClangD expects this to do proper "gotoDefinition" handling, however this file needs to be generated e.g. by cmake and passed as an argument to the language server

@evidolob may be could help here.

evidolob:
> How should we handle the compile_commands.json? ClangD expects this to do proper "gotoDefinition" handling, however this file needs to be generated e.g. by cmake and passed as an argument to the language server

I guess, we need to generate that file after project creation/clone.

AndrienkoAleksandr:
@hkolvenbach About generation json, we can try workaround. I think we can install cmake with help installation script org.eclipse.che.ls.clangd.sh.  And than we can run command for generation json like we do it for csharp-ls: ["dotnet restore on start csharp-ls"](https://github.com/eclipse/che/blob/master/plugins/plugin-csharp/che-plugin-csharp-lang-server/src/main/java/org/eclipse/che/plugin/csharp/languageserver/CSharpLanguageServerLauncher.java#L48) . For csharp we run command "dotnet restore" to restore dotnet dependencies, without them cshape intellisense doesn't work. For clangD we can run cmake command to generate compilation data-source json. 
And as I know cmake support generation such json from some version. We should check installation cmake in the container and version cmake. In case if cmake was not installed than install this third-party program.
What do you think @tolusha ?

tolusha:
@AndrienkoAleksandr I don't mind.

hkolvenbach:
@tolusha @AndrienkoAleksandr @evidolob thanks for your responses, we are now continuing to work on it. just another question: what would be the most appropriate way to add the `--compile-commands-dir=<compile_commands.json>` parameter to the agent launch script? just write a new `launch.sh` script with the specific `compile_commands.json` parameter before https://github.com/eclipse/che/pull/7516/files#diff-b0a97402fc623235c7707c20684cfe61R75 for each project?

FYI @thierryk12

tolusha:
IMHO it is better choice:
>  just write a new launch.sh script with the specific compile_commands.json parameter before https://github.com/eclipse/che/pull/7516/files#diff-b0a97402fc623235c7707c20684cfe61R75 for each project?

hkolvenbach:
just a quick update: we have a working compile_commands.json integration internally, but noticed that clangd (or the communication from eclipse che to clangd) is crashing regularly after a few minutes. currently investigating if this is a clangd or eclipse che problem.

hkolvenbach:
by the way: am i correct that eclipse che is not restarting a crashed language server? the VSCode plugin for clangd restarts it once it crashes.

tolusha:
@hkolvenbach 
You are right, LS won't be restarted in case of crashing.

hkolvenbach:
Quick update (as already told on the community call yesterday):
I will soon update the PR with a Che6 version of the clangd integration (we won't continue on che5)

What is currently working:
- go to definition
find_definition
- autodetect compile_commands.json (clangd feature)
- refactoring
- code completion & templates like "for" loops
for_autocomplete
- warnings and errors 

clangd very much depends on the compile_commands.json file, but if it is present, the results are generally very good. regarding #4991, a quick comparison to `cquery` also didn't show much differences (comparing latest clangd 6 svn).

There are some open issues regarding the languageserver implemention of che though, which I would like to discuss before we propose changes that break in other languages:
- code completion appends instead of replacing the text (see the for loop example, `for` becomes `forfor`) We think the issue is here https://github.com/eclipse/che/blob/af0376de7f0d0aaadc90647fe4e54e9ad6040938/plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/codeassist/CompletionItemBasedCompletionProposal.java#L230-L239. @gorkem @tsmaeder  is there any reason to have this else case? where is offset coming from? if we use this https://github.com/eclipse/che/blob/af0376de7f0d0aaadc90647fe4e54e9ad6040938/plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/codeassist/CompletionItemBasedCompletionProposal.java#L226 instead, it works correctly
- the autocompletion uses a fuzzy search https://github.com/eclipse/che/blob/af0376de7f0d0aaadc90647fe4e54e9ad6040938/plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/codeassist/LanguageServerCodeAssistProcessor.java#L136, compared to VScode this generates a lot of noise and `printf`, which actually starts with `pri` is not even in the first few results (see screenshots)
**Eclipse Che:**
fuzzy_autocomplete
**VSCode:**
fuzzy_autocomplete_vscode

CC @thierryk12

hkolvenbach:
the PR is updated, to test it you need to enable the clangd language server under installers and use `silexica/tools:nightly` as a base image (it has an integrated clangd binary)

agents/ls-clang/src/main/resources/installers/1.0.0/org.eclipse.che.ls.clang.script.sh
```diff
@@ -0,0 +1,157 @@
...
+#        ${SUDO} yum install ${PACKAGES};
+#    }
+#
+#    command -v nodejs >/dev/null 2>&1 || {
+#        curl --silent --location https://rpm.nodesource.com/setup_6.x | ${SUDO} bash -;
+#        ${SUDO} yum -y install nodejs;
+#    }
+#
+## Red Hat Enterprise Linux 6
+#############################
```
tolusha: Ideally we have to support all possible linux distributions

tolusha:
@hkolvenbach
I've tested the PR. It seems working well.

agents/ls-clang/src/main/resources/installers/1.0.0/org.eclipse.che.ls.clang.script.sh
hkolvenbach: yes, i will think about a way to install it through the script

assembly/assembly-wsmaster-war/pom.xml
```diff
@@ -415,6 +419,22 @@
                             <outputDirectory>${project.build.directory}/classes</outputDirectory>
                         </configuration>
                     </execution>
+                    <!-- <execution>
```
dkuleshov: Hi, I guess this could be either removed or uncommented?

plugins/plugin-clangd/che-plugin-clangd-lang-server/src/main/java/org/eclipse/plugin/clangd/inject/ClangModule.java
```diff
@@ -0,0 +1,64 @@
...
+public class ClangModule extends AbstractModule {
+  public static final String LANGUAGE_ID = "clangd";
+  private static final String[] EXTENSIONS =
+      new String[] {
+        "c", "h", "cpp", "hpp", "cc", "hh", "hxx", "cxx", "C", "H", "CPP", "HPP", "CC", "HH", "CXX",
+        "HXX"
+      };
+  private static final String MIME_TYPE = "text/x-cpp";
+
+  // @Override
```
dkulieshov: I just would like to remind that we're trying to avoid having comments of such kind when we do merge to master, do you think this section could be removed?

plugins/plugin-clangd/che-plugin-clangd-lang-server/src/main/java/org/eclipse/plugin/clangd/inject/ClangModule.java
```diff
@@ -0,0 +1,64 @@
...
+import com.google.inject.multibindings.Multibinder;
+import org.eclipse.che.api.languageserver.launcher.LanguageServerLauncher;
+import org.eclipse.che.api.languageserver.shared.model.LanguageDescription;
+import org.eclipse.che.api.project.server.type.ProjectTypeDef;
+import org.eclipse.che.inject.DynaModule;
+import org.eclipse.che.plugin.cpp.projecttype.CProjectType;
+import org.eclipse.che.plugin.cpp.projecttype.CppProjectType;
+import org.eclipse.plugin.clangd.languageserver.ClangDLanguageServerLauncher;
+
+/** @author Alexander Andrienko */
```
dkulieshov, line 25: Seems like copy/pasted author?

AndrienkoAleksandr: No :)

plugins/plugin-clangd/che-plugin-clangd-lang-server/src/main/java/org/eclipse/plugin/clangd/inject/ClangModule.java
hkolvenbach: good point, i will go through it and remove those

hkolvenbach:
@dkuleshov I removed the comments. I think the PR is ready to be merged, if there are no more suggestions from you or the Che team.

BTW: I picked clang 6 (will be released soon anyway), because clangD 5 was not working properly with the Che language protocol client

agents/ls-clang/src/main/resources/installers/1.0.0/org.eclipse.che.ls.clang.json
```diff
@@ -0,0 +1,8 @@
+{
+    "id": "org.eclipse.che.ls.clangd",
+    "version": "1.0.0",
+    "name": "Clangd language server",
+    "description": "Clangd intellisense for C/C++ projects.",
+    "dependencies": [],
+    "properties": {}
+  }
```
tolusha: It is supposed to be a new line.

agents/ls-clang/src/main/resources/installers/1.0.0/org.eclipse.che.ls.clang.script.sh
```diff
@@ -0,0 +1,162 @@
...
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+#
+
+unset PACKAGES
+unset SUDO
+command -v tar >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" tar"; }
+command -v curl >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" curl"; }
+command -v wget >/dev/null 2>&1 || { PACKAGES=${PACKAGES}" wget"; }
+test "$(id -u)" = 0 || SUDO="sudo -E"
```
tolusha: Have a look at 
https://github.com/eclipse/che/blob/2c2bd6637c8169018e2b33397f9d738bbf82a0b1/agents/ls-csharp/src/main/resources/installers/1.0.1/org.eclipse.che.ls.csharp.script.sh#L12

The scripts have been modified. For instance
```
is_current_user_root() {
    test "$(id -u)" = 0
}

is_current_user_sudoer() {
    sudo -n true > /dev/null 2>&1
}

set_sudo_command() {
    if is_current_user_sudoer && ! is_current_user_root; then SUDO="sudo -E"; else unset SUDO; fi
}
```
have been added to test if user is sudo or isn't

hkolvenbach: thanks, i will take a look!

agents/ls-clang/src/main/resources/installers/1.0.0/org.eclipse.che.ls.clang.json
hkolvenbach: fixed

vparfonov (approved):

tolusha (approved):
Pls, update pom.xml to set 6.1.0-SNAPSHOT version.

hkolvenbach:
done!

tolusha:
@vparfonov What are the next steps here?

dkulieshov (approved):

pom.xml
```diff
@@ -1028,6 +1033,12 @@
                 <artifactId>che-plugin-camel-server</artifactId>
                 <version>${che.version}</version>
             </dependency>
+            <dependency>
+                <groupId>org.eclipse.che.plugin</groupId>
+                <artifactId>che-plugin-clangd-lang-server</artifactId>
+
```
skabashnyuk: redundant line

hkolvenbach: fixed

skabashnyuk (approved):

vparfonov:
ci-build

skabashnyuk:
ci-test

musienko-maxim:
The selenium tests passed without regressions

Issue #7543, opened 2017-11-24
Fix unexpected fail CreateNewNotJavaFilesFromContextMenuTest selenium test


Ohrimenko1988:
### Description
We have unexpected fail CreateNewNotJavaFilesFromContextMenuTest selenium test. The test has a problem with fill the AskForWalueDialog form. We need to investigate and fix this problem.

Report with a reproduced problem: 
https://ci.codenvycorp.com/view/qa/job/che-multiuser-integration-tests-che6/31/Selenium_tests_report/

### Reproduction Steps

**OS and version:** 
Che 6.0.0-M2 ; Chrome 59.0 ; Fedora 25   

**Diagnostics:**

Ohrimenko1988:
Done https://github.com/eclipse/che/pull/7545

PR #7545, opened 2017-11-24
Fix bug in AskForValueDialog in typeAndWaitText method

Ohrimenko1988:
### What does this PR do?
Fix bug in AskForValueDialog in typeAndWaitText method

Issue https://github.com/eclipse/che/issues/7543

### DIFF
--- a/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/AskForValueDialog.java
+++ b/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/AskForValueDialog.java
@@ -10,13 +10,15 @@
  */
 package org.eclipse.che.selenium.pageobject;
 
+import static org.eclipse.che.selenium.core.constant.TestTimeoutsConstants.LOAD_PAGE_TIMEOUT_SEC;
 import static org.eclipse.che.selenium.core.constant.TestTimeoutsConstants.REDRAW_UI_ELEMENTS_TIMEOUT_SEC;
+import static org.openqa.selenium.support.ui.ExpectedConditions.invisibilityOfElementLocated;
+import static org.openqa.selenium.support.ui.ExpectedConditions.visibilityOf;
 
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
 import org.eclipse.che.selenium.core.SeleniumWebDriver;
 import org.eclipse.che.selenium.core.action.ActionsFactory;
-import org.eclipse.che.selenium.core.constant.TestTimeoutsConstants;
 import org.openqa.selenium.By;
 import org.openqa.selenium.Keys;
 import org.openqa.selenium.NoSuchElementException;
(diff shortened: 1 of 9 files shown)


selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/AskForValueDialog.java
```diff
@@ -137,7 +138,10 @@ public void clickCancelButtonJava() {
   }
 
   public void typeAndWaitText(String text) {
-    input.sendKeys(text);
+    waitFormToOpen();
```
dmytro-ndp: It should be checked before the call of the method.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/AskForValueDialog.java
```diff
@@ -137,7 +138,10 @@ public void clickCancelButtonJava() {
...
+    new WebDriverWait(seleniumWebDriver, TestTimeoutsConstants.LOAD_PAGE_TIMEOUT_SEC)
```
dmytro-ndp: Good catch!
Lets extract `new WebDriverWait(seleniumWebDriver, TestTimeoutsConstants.LOAD_PAGE_TIMEOUT_SEC)` into the separate constant and add similar verification in other methods.

tolusha (approved):

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/AskForValueDialog.java
```diff
@@ -105,90 +108,88 @@ public AskForValueDialog(
...
+    loadPageWait.until(visibilityOf(okBtnnewJavaClass)).click();
     loader.waitOnClosed();
   }
 
   public void typeTextInFieldName(String text) {
-    inputNameJavaClass.sendKeys(text);
+    loadPageWait.until(visibilityOf(inputNameJavaClass)).sendKeys(text);
+  }
+
+  public void waitTextInFieldName(String expectedText) {
```
dmytro-ndp: IMHO it's better to enclose call of this method into the **typeTextInFieldName()** in similar way, as it was done for **waitInputNameContains()** method.

selenium/che-selenium-test/src/test/resources/conf/selenium.properties
```diff
@@ -31,11 +31,11 @@ che.selenium.infrastructure=docker
 # Define configuration for che default test user.
 # For che single user assembly, the default values will be enough.
 # For che multiuser assembly, user must be already created on identity provider service.
-che.test_user.email=che@eclipse.org
-che.test_user.password=secret
+che.test_user.email=admin@admin.com
```
dmytro-ndp: Is it really what you wanted to change? :-)

dmytro-ndp (approved):

Issue #7155, opened 2017-11-02
Check whether we can run rh-che6 multi-user on OSIO in a multi-cluster manner


garagatyi:
### Description
Che5 is supposed to run on OSIO in a multi-cluster way in multi-user mode. 
We need to check whether Che6 may be run in the same way and if not create issues to accomplish this.

sleshchenko:
#### What does rh-che assembly contain?
1. Workspace Master
Replaced default che stacks, added bayesian-agent
2. Workspace Agent
Added bayesian language server
3. GWT IDE
Replaced Che product info plugin
4. Dashboard
There are two packaging of Che Server - with and without dashboard. It is needed because OSIO doesn't provide and ability to remove workspaces. Packaging with Che Server with dashboard will be removed when OSIO will has such functionality.

#### Discovered tasks

- [x] Create new branch che6 in [rh-che](https://github.com/redhat-developer/rh-che) repository based on  [multi-tenant-rh-che](https://github.com/redhat-developer/rh-che/tree/multi-tenant-rh-che) branch.
-  [x] Adapt gwt packaging to new gwt assembling approach that was introduced in che6 branch.
- [x] Remove deprecated components that were needed to extending OpenShiftConnector. Extend OpenShift infrastructure to use OpenShift project and connection config(master url, token) which are provided by Fabric8 user service.
- [x] Remove excluding of svn plugin as it is removed in che6 branch
- [ ] Override logs folders for agents https://github.com/eclipse/che/issues/7807
- [x] Implement caching of token per workspace or investigate another way for interaction with OpenShift cluster without user (idle timeout stop, Bayesian plugin).
- [x] Update `deploy_script.sh`
- [x] BuildInfo in not packaged into che-ide war archive. So it is needed to take a look `assembly/aseembly-build-info` which unpacked BuildInfo.properties for ci scripts
- [ ] Prepare migration for workspace files https://github.com/redhat-developer/rh-che/issues/477

sleshchenko:
Most of the required changes are made and will be merged soon. 
Then it will be possible to run workspace with Che6 on OSIO. See Demo:
![rh-che6](https://user-images.githubusercontent.com/5887312/33375453-9e98b72c-d512-11e7-8de6-6d90d2855e13.gif)

Few of tasks will be done in separated issues.

PR #7604, opened 2017-11-28
Minor changes for extending che by rh-che

sleshchenko:
### What does this PR do?
Adds an ability to throw an exception while OpenShift client creation. It is needed while other OpenShift client factory implementations should be able to do some operations before construction of OpenShift client instance and throw an exception if an error occurs and it is impossible to create OpenShift client. 

Replace stacks has an outdated link. So this PR fixes it.

The following change is not related to rh-che: Workspace agent components use AgentHttpJsonRequestFactory that is responsible for the setting of machine token to requests to workspace master API. So such logic is removed in WorkspaceProjectSynchronizer because:
- it uses the outdated environment variable `USER_TOKEN`, it should be `CHE_MACHINE_TOKEN`.
- AgentHttpJsonRequestFactory does the same thing.

https://github.com/eclipse/che/issues/7155

### DIFF
--- a/dockerfiles/init/modules/openshift/files/scripts/replace_stacks.sh
+++ b/dockerfiles/init/modules/openshift/files/scripts/replace_stacks.sh
@@ -22,7 +22,7 @@ if [ -z "${CHE_API_ENDPOINT+x}" ]; then
     echo "done (${CHE_API_ENDPOINT})"
 fi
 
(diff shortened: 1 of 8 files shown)


sleshchenko:
@dkuleshov Please review https://github.com/eclipse/che/pull/7604/commits/82dd1a733b770030cf9cce74b227805a53ddda68

garagatyi (approved):
Looks good

akorneta (approved):

dkulieshov (approved):

Issue #7313, opened 2017-11-10
New Status bar / Activity monitor


slemeur:
### Description
The status bar in its current form is used only for maven project and it is rarely active. The information displayed in is important, but we must always look at how we can optimize the use of the space on the screen.

For all users who are not doing Maven, the status bar is stealing 20px which could be better use. 

The goal of this issue is to discuss and define what other solutions we could put in place. 

### Options

1- Display the status bar only for maven project
2- Hide the bar, when no information to be displayed (would be annoying for end-user)
3- Allowing the command controller to display IDE activity information (similar to XCode and Visual Studio)
4- Display the information in a toast popup

AndrienkoAleksandr:
Related to this discussion "Make IDE more informative": https://github.com/eclipse/che/issues/6929

skabashnyuk:
@Stevan
You are right showing project type specific information that is rarely(arguable topic but still) active in 20px is a crime. We must think about the reasonable use of space.

Let's quickly summarise world best-practices.
![2017-11-10 21 14 43](https://user-images.githubusercontent.com/1614429/32676122-1af12f1c-c662-11e7-8ae0-20895a114b1d.png)

On this image, we can see three IDE. They all show at the bottom the static panel, that probably 
can be disabled, but by default, it's visible 100% of the time.
What kind of information we see.
1. Line and column position 
2. Line feed, carriage return mode-. 
3. Tabulation size
4. Codepage of file
5. Opened file type.
6. Ready only or not.
7. Long-running process progress. Git clone, git push, maven import, refactoring, etc.
8. Event log.
9. IDE memory consumption.
10. VCS information, like branch name.

Most if not all these indicators are projected agnostic

What do we have
<img width="832" alt="2017-11-10 21 31 54" src="https://user-images.githubusercontent.com/1614429/32676128-261f04ea-c662-11e7-8e84-3cff7dd561c5.png">

This is a panel inside editor that always shown. Not sure it's  can be disabled.
It's located above bottom panels and show:
1. Line and column position 
2. Codepage of file
3. Opened file type.
Not much, yeh.

I agree if we put the status panel on the bottom like other IDEs we will have two panels instead of one.
And it's quite bad.


What do I propose? 
1. Have one single panel on the bottom. https://github.com/eclipse/che/issues/6929
2. Revise list of best practices and summarise what we can show on this panel. I think major part should be a project or file independent.
[...]

PR #7606, opened 2017-11-28
Displays current project name with vcs head name in status bar if present

vzhukovs:
### What does this PR do?
This PR proposes to display in status bar the current project name from the application context with vcs head name if the last one exists.

git_branches

N/A

#### Release Notes
Displays current project name with vcs head name in status bar if present

#### Docs PR
Doesn't need to be documented.

### DIFF
--- a/plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/GitExtension.java
+++ b/plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/GitExtension.java
@@ -10,6 +10,7 @@
  */
 package org.eclipse.che.ide.ext.git.client;
 
+import static org.eclipse.che.ide.api.action.IdeActions.GROUP_LEFT_STATUS_PANEL;
 import static org.eclipse.che.ide.api.action.IdeActions.GROUP_MAIN_MENU;
 import static org.eclipse.che.ide.api.action.IdeActions.GROUP_PROFILE;
 import static org.eclipse.che.ide.api.constraints.Anchor.BEFORE;
@@ -44,6 +45,7 @@ import org.eclipse.che.ide.ext.git.client.action.ShowBranchesAction;
 import org.eclipse.che.ide.ext.git.client.action.ShowMergeAction;
 import org.eclipse.che.ide.ext.git.client.action.ShowRemoteAction;
 import org.eclipse.che.ide.ext.git.client.action.ShowStatusAction;
+import org.eclipse.che.ide.ext.git.client.action.StatusBarBranchPointer;
 import org.eclipse.che.ide.ext.git.client.action.ToggleGitPanelAction;
 import org.vectomatic.dom.svg.ui.SVGImage;
 
@@ -100,7 +102,8 @@ public class GitExtension {
       PreviousDiffAction previousDiffAction,
       KeyBindingAgent keyBinding,
       ToggleGitPanelAction gitPanelAction,
-      GitNotificationsSubscriber gitNotificationsSubscriber) {
+      GitNotificationsSubscriber gitNotificationsSubscriber,
+      StatusBarBranchPointer statusBarBranchPointer) {
     gitNotificationsSubscriber.subscribe();
 
     resources.gitCSS().ensureInjected();
@@ -235,5 +238,9 @@ public class GitExtension {
     keyBinding
         .getGlobal()
         .addKey(new KeyBuilder().alt().charCode('g').build(), GIT_PANEL_ACTION_ID);
+
+    DefaultActionGroup rightStatusPanelGroup =
+        (DefaultActionGroup) actionManager.getAction(GROUP_LEFT_STATUS_PANEL);
+    rightStatusPanelGroup.add(statusBarBranchPointer);
   }
 }
--- a/plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/action/StatusBarBranchPointer.java
+++ b/plugins/plugin-git/che-plugin-git-ext-git/src/main/java/org/eclipse/che/ide/ext/git/client/action/StatusBarBranchPointer.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright (c) 2012-2017 Red Hat, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.ide.ext.git.client.action;
+
+import com.google.common.annotations.Beta;
+import com.google.gwt.dom.client.Style;
+import com.google.gwt.dom.client.Style.Cursor;
+import com.google.gwt.dom.client.Style.Unit;
+import com.google.gwt.user.client.ui.HorizontalPanel;
+import com.google.gwt.user.client.ui.Label;
+import com.google.gwt.user.client.ui.Widget;
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+import org.eclipse.che.ide.api.action.ActionEvent;
+import org.eclipse.che.ide.api.action.BaseAction;
+import org.eclipse.che.ide.api.action.CustomComponentAction;
+import org.eclipse.che.ide.api.action.Presentation;
+import org.eclipse.che.ide.api.app.AppContext;
+import org.eclipse.che.ide.api.resources.Project;
+import org.eclipse.che.ide.ext.git.client.GitResources;
+import org.eclipse.che.ide.ext.git.client.branch.BranchPresenter;
+import org.vectomatic.dom.svg.ui.SVGImage;
+
+/**
+ * Displays in status bar current project name with vcs branch if present.
+ *
+ * @author Vlad Zhukovskyi
+ * @since 6.0.0
+ */
+@Singleton
+@Beta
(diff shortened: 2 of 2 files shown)


vinokurig (approved):

vzhukovs:
ci-build

vparfonov:
This example of feature that easy implement but very useful.
Thanks

vparfonov (approved):

vzhukovs:
ci-build

skabashnyuk (approved):

ashumilova (approved):

slemeur:
@vzhukovskii : 
1- What's supposed to be shown when user is having the focus in the terminal, or in a command editor ?
2- Why "documentation is not needed" ? 
3- We do have an information bar in the editor. In #5128, it was specified to display the branch information in that editor status bar. [see](https://user-images.githubusercontent.com/1636769/29118684-4149e206-7d03-11e7-961d-283f7ca99ba5.png). The way you choose to display the information force the user to go "outside" of the editor area. When you have a single file and no terminal open, that's fine. But if a user has multiple files opened plus a terminal, that information is then displayed out of any context.

vzhukovs:
@slemeur :
1 - The last project selected in application context, it doesn't need to be related with editor, terminal, project explorer, etc. It works only with currently selected project.
2 - Because its too obvious for the user, widget displays the project name with the branch like in Project Explorer and changes with the selection change.
3 - As for displaying the branch in editor status panel, and
>The way you choose to display the information force the user to go "outside" of the editor area.

Look how it implemented in Intellij, it works independently and it will be updated when selection changes.

>But if a user has multiple files opened plus a terminal, that information is then displayed out of any context.

It works using the current context. When part become active, it puts the current project into the context.

slemeur:
I don't think this information is relevant in the status bar - even if others are doing it this way. 

We are basically duplicating an information that is already present in our UI, in a more useful way. 

The branch information is displayed in the project explorer - so giving a look at the project explorer allows the user to get the branch currently opened at a glance. He can also get this information for all the projects currently in the workspace. The closer an information is displayed to the relative source, the better it is. When I'm browsing the list of my project, I don't want to go into another area of my tool, to know the branch I'm actually working on.
We could and should allow fast switching from the project explorer, as you implemented in your widget. 

Similarly, in #5128 ([here](https://user-images.githubusercontent.com/1636769/29118684-4149e206-7d03-11e7-961d-283f7ca99ba5.png)) it was specified to display the branch information in the editor info bar. Again, the purpose is simple. The closer an information is displayed to the relative source, the better it is :
- When I'm editing a file, I don't want to go in another area of the tool, to know on which branch I'm currently working on.
- I can have multiple files opened - in this case, I should be able to know the branch for each files opened.
Here again, we should allow fast switching from the project explorer, as you implemented in your widget. 

It would be so much simpler if we try to think and analyze about how we can make things better than IntelliJ and others instead of just copying it.

sunix:
I also don't think it's a good idea to show that in the a global status bar. What if i'm editing another file that is not the one selected in the explorer ... this only one example that may confuse but i'm sure we can find more. why not showing in the editor's status bar ?

vzhukovs:
@sunix if you switch to editor, editor panel will set into application context current opened file and related root project, and widget will update its state.

sunix:
sorry @vzhukovskii  this is the current behavior you have implemented but you will never solve the fact that a user may be confused how your application will behave. We have to try to keep things obvious to the user. So why not in the editor bar ? it would make sense ... no ?
And what if I run a command or open the terminal in full screen and have this status bar not showing the project where my command is running .... as I said we can find many example to show that this is not straight forward ... the user would have to think ... try and guess the behaviour to understand what the information you display is about.
I guess you don't think like this because you are replicating a behaviour that is in an IDE you are using everyday but just think that other users may have other habits and may think differently.

skabashnyuk:
@sunix the opposite statement is also correct. It might be confusing to not to see branch name where the majority of our contributors or users (users of modern ides like: IntelliJ, Atom, VsCode) may expect. And maybe we should adjust to their expectation.

vzhukovs:
@sunix don't forget that project name can be very long and git branch name as well (for example, if you checkout to specific tag):

codenvy _ wksp-011c 2017-11-29 17-50-25

User won't understand what a big hash is doing near project name especially if user open factory.
But if user works for a long time in different IDEs, he'll expect to find such switcher in familiar place.

sunix:
Guys, what about having this in the editor status bar ? don't you think it's the best place ? this is not the opposite statement ... but to find a way that works for everyone.

@vzhukovskii sorry i don't understand your point if you have the git hash, only the first one are enough to be used ... and if I see a big hash in the status bar I won't understand neither ....

we are trying to improve things here maybe we should have a community vote for this kind of things

skabashnyuk:
>Guys, what about having this in the editor status bar? 

I personally think this duplication is redundant since we have one general status bar. Perhaps we may remove editor's status bar to save some space.
Looks like my proposal https://github.com/eclipse/che/issues/7313#issuecomment-343571099 is quite popular in the community. And @vzhukovskii changes looks pretty consistent with this proposal.

sunix:
@skabashnyuk 

> I personally think this duplication is redundant since we have one general status bar. Perhaps we may remove editor's status bar to save some space.

IMO, if we need to remove one thing, it should be the global status bar. The editor bar is good because it gives the information of what you are currently editing. If you want to see the line number you are, you don't look at the bar below the terminal ... you look at the bar just below the editor first.
To me, activity bar should be hidden and shown when there is something to display.

Here you are not giving your feeling about the editor bar, you are just talking about duplication, but if you talk about duplication, let's not display these one in the global status bar because it is display already in the project explorer ...

So what do you think about displaying in the editor bar instead of the global status bar? (you didn't really answerd) it's already here, we solve the problem when you have multiple editor display at the same time ... etc ... and developers would know what it is about and not get confused.

> Looks like my proposal #7313 (comment) is quite popular in the community. And @vzhukovskii changes looks pretty consistent with this proposal.

;) pehaps but the community is not only located in Ukraine ;)

vzhukovs:
@sunix 
>because it is display already in the project explorer

If you have a big expanded project in explorer, you have to scroll up to see in what branch you perform work, this is 90% situation (for example if you develop Che in Che). But meantime, the project name can be very long and branch name may be invisible, and again, have to scroll. This is two use cases from many others...

>So what do you think about displaying in the editor bar instead of the global status bar?

Meantime, the editor might be closed and you'll have to do something to view in what branch you are. 

>and developers would know what it is about and not get confused.

I think we have to try to solve the problem with proper focus recognizing, because if you'll have tree of four opened editors (I mean in split view), you'll have the same count of status bars, do you really think that this is a good idea, instead of make a proper mechanism to highlight zone you work with and display important information in one specific for this place?

The main idea of this PR, is to display the branch with project name in bottom status bar and to be always in field-of-view.

As @skabashnyuk mentioned, almost of our users are developers and developers who works with popular IDEs. So why we have to force them to learn new behavior instead of making something familiar to them?

skabashnyuk:
I think that is the question that comes to all popular ide. global vs editors. And I think they choose global because you may need to display information that is not related to the editor. The only drawback I see that information is not that close as it can be. But that's acceptable to me and looks like for IntelliJ, Atom, VsCode users too.

>;) pehaps but the community is not only located in Ukraine ;)

Pr is open to anyone. I'm glad that our Ukrainian community uses  Che more and more so you can see engineers are more active recently.


As a small sum up of this discussion.  Since I don't see strong disagree on current PR I would like to give a go for @vzhukovskii to merge it and make the community happy. As far as I can see the majority of them are +1.

sunix:
I think we have all given our opinion about this one. It wont go anywhere continuing discussing about this. Then I feel like it's a shame to have arguments "because the others are doing it, we should do the same" ... i would prefer real arguments why it should be there, why it is better. I would like to see Che the best IDE because we are bringing something new and really useful. Then if you feel like Stevan and mine are not strong disagree, feel free to merge :)
Honestly, I'm very happy to see you guys using che in che so I don't feel like I'm the only one doing that. Let's put this information in both global status bar AND editor bar for some times and try maybe ? have feedback from users ... we will have more and more users with OSIO giving feedback on usuability and maybe they would love having this in the editor bar or maybe not.

vzhukovs:
>i would prefer real arguments why it should be there, why it is better

just compare two imagine variants, for example:
eclipse che 2017-12-01 10-44-24
eclipse che 2017-12-01 11-13-10

second variant is not overloaded with the information given to user and it changes with the current context

AndrienkoAleksandr:
I think if most IDEs use some classic way it's not the shame to do the same, it's a good practice... I mean, yes we really should not copy and paste some solutions,  but we need to learn experience from another good projects and use it. And customer during migration to us should not spent a lot of time to learn and get used to our UI... There are a lot cases when we should not invent new vehicle.

sunix:
About scrolling to always show the project root folder, we could do like it is done for the header of this page: http://www.jugsummercamp.org/edition/8

sunix:
@vzhukovskii i prefere the one with the information on each editor: I don't have to think "which editor is this information related to ?"
Let's do like the most popular editor ever: emacs http://ergoemacs.org/emacs/emacs_split_windows.html

benoitf:
it looks odd to have this PR merged while people disagree on it.

skabashnyuk:
@benoitf I'm a bit disagree with your last statement.
1. From @sunix  message  ````Then if you feel like Stevan and mine are not strongly disagree, feel free to merge :)```` https://github.com/eclipse/che/pull/7606#issuecomment-348349181
2. Second, what do you propose? Ignore opinion of 18 Eclipse Che users.

benoitf:
@skabashnyuk reading the comment you linked is that you considered their remarks as not useful

2 -->  IMHO It would have be nice to converge a common agreement. I would also hope to have heard comments from worldwide. (you're saying that you don't want to ignore 18 users (but AFAIK there are more than 18 ppl using the product))

PR #7625, opened 2017-11-29
Add attributes field to ServerConfig and Server model objects

### LINKED ISSUE #7560: Add attributes field into ServerConfig and Server model objects
### Description
To provide an [ability to use Language server as a separate container](https://github.com/eclipse/che/issues/7554) we need a way to identify it in workspace runtime and provide some metadata for wsagent.
1. Add Map<String, String> field with name `attributes` into ServerConfig model
1. Add the same field into Server object model (representation of ServerConfig at runtime)
1. Copy attributes from ServerConfig into the corresponding Server

### Reproduction Steps

**OS and version:**    

**Diagnostics:**

garagatyi:
### What does this PR do?
Note: this PR is designed for commit-by-commit review. 
This PR adds field `attributes` represented by map<string, string> to the model of objects ServerConfig and Server. This allows us to store some metadata in a server and use it at runtime for different purposes, e.g. recognizing LS servers started in separate containers.

Also, this PR enables DockerAbandonedResourceCleaner which was disabled some time ago, probably, by mistake.

Fixes #7560

### DIFF
--- a/infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Labels.java
+++ b/infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Labels.java
@@ -10,8 +10,9 @@
  */
 package org.eclipse.che.workspace.infrastructure.docker;
 
-import static java.util.Collections.emptyMap;
-
+import com.google.common.reflect.TypeToken;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.Map;
@@ -39,10 +40,16 @@ public final class Labels {
   private static final String SERVER_PORT_LABEL_FMT = LABEL_PREFIX + "server.%s.port";
   private static final String SERVER_PROTOCOL_LABEL_FMT = LABEL_PREFIX + "server.%s.protocol";
   private static final String SERVER_PATH_LABEL_FMT = LABEL_PREFIX + "server.%s.path";
+  private static final String SERVER_ATTR_LABEL_FMT = LABEL_PREFIX + "server.%s.attributes";
 
   /** Pattern that matches server labels e.g. "org.eclipse.che.server.exec-agent.port". */
   private static final Pattern SERVER_LABEL_PATTERN =
       Pattern.compile("org\\.eclipse\\.che\\.server\\.(?<ref>[\\w-/.]+)\\..+");
+
+  private static final Gson GSON = new GsonBuilder().disableHtmlEscaping().create();
+  // used to avoid frequent creations of the object at runtime
+  private static final java.lang.reflect.Type mapTypeToken =
+      new TypeToken<Map<String, String>>() {}.getType();
 
   /** Creates new label serializer. */
   public static Serializer newSerializer() {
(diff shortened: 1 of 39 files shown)


garagatyi:
ci-build

gazarenkov (approved):

garagatyi:
ci-build

skabashnyuk (approved):

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Labels.java
```diff
@@ -99,6 +106,9 @@ public Serializer server(String ref, ServerConfig server) {
       if (server.getPath() != null) {
         labels.put(String.format(SERVER_PATH_LABEL_FMT, ref), server.getPath());
       }
+      if (server.getAttributes() != null) {
```
sleshchenko, line 109: May `server.getAttributes` return null value?

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Labels.java
```diff
@@ -156,7 +166,8 @@ public RuntimeIdentity runtimeId() {
                     labels.get(String.format(SERVER_PORT_LABEL_FMT, ref)),
                     labels.get(String.format(SERVER_PROTOCOL_LABEL_FMT, ref)),
                     labels.get(String.format(SERVER_PATH_LABEL_FMT, ref)),
-                    emptyMap()));
+                    GSON.fromJson(
```
sleshchenko: Will `GSON.fromJson(null);` work without exception?

sleshchenko (approved):
LGTM

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Labels.java
garagatyi: Why not? It is just an interface, we don't know which impl is there

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Labels.java
garagatyi: yes

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Labels.java
sleshchenko: Ok, I'd propose document that it must not return null value like we do for fields in WorkspaceConfig. As far as I remember there was some declaration that model interfaces must not return null values for collections. If they can we should rework ton of places where we didn't check returned value on null where a collection is returned. For example you can take a look usage of org.eclipse.che.api.core.model.workspace.runtime.Machine#getServers.
Personally, I like such declaration because it makes developers life easier and code cleaner =)

akorneta (approved):

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/Labels.java
sleshchenko: Ok, thx.

vparfonov (approved):

tolusha (approved):

PR #7627, opened 2017-11-29
Add ability to wait defined count of splitted tabs in the editor

Ohrimenko1988:
### What does this PR do?
 - Remove displaying user`s password in the log
 - Change "LOG" to "Assert.fail()" in the "getElementSrcLink" method
 - add waitCountTabsWithProvidedName method 
 - add wait for open splitted tabs in the editor to the "SplitEditorFeatureTest"


selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/user/TestUserImpl.java
```diff
@@ -91,7 +91,7 @@ public TestUserImpl(
     this.userServiceClient.create(name, email, password);
     this.authToken = authServiceClient.login(name, password);
     this.id = userServiceClient.findByEmail(email).getId();
-    LOG.info("User name='{}', password='{}', id='{}' has been created", name, password, id);
+    LOG.info("User name='{}', id='{}' has been created", name, id);
```
dmytro-ndp, line 94: We need a password to be able to log into Che as a test user and verify the status of account.

dmytro-ndp:
We need a user's password to be able to log into Che as a test user and verify the status of account, for example, in time of debugging.
Also. it shouldn't be a problem for as so as test user is been removed after the test execution.

dmytro-ndp:
Please, explain what type of tab this request is about in title of PR.

skabashnyuk:
We do not need the password in logs. Please merge this change ASAP

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/user/TestUserImpl.java
tolusha: If password is configured why do you need to log it?

dmytro-ndp: Moreover, in context of testing of Che no test user has been created. Tests use pre-exist user which credentials are described in the test properties. So, the message is completely out of scope here.

But we need it for Codenvy Onprem and SaaS, which are using the same class **TestUserImpl**.

tolusha (approved):

Issue #7418, opened 2017-11-16
Che 6 on OCP: Failed Workspace Start


ghost:
### Description

Automated QA tests show an enormous number of failed workspace starts. As a result, most of tests fail because workspace has never started, or a workspace somehow depends on the workspaces that should have started before but never did so.

Server says that:

```
2017-11-16 14:10:02,861[aceSharedPool-3]  [ERROR] [w.i.o.OpenShiftInternalRuntime 123]  - Failed to start of OpenShift runtime. An error has occurred.
```
where `An error has occurred` is `KubernetesClientException Caused by: java.net.SocketTimeoutException: connect timed out` 

It happens both when starting and deleting workspaces.

Server logs are also full of:

```
WARNING: A connection to https://kubernetes.default.svc/ was leaked. Did you forget to close a response body? To see where this was allocated, set the OkHttpClient logger level to FINE: Logger.getLogger(OkHttpClient.class.getName()).setLevel(Level.FINE);

```
which is described in this issue - https://github.com/eclipse/che/issues/5902

This is how exception looks like:

```
2017-11-16 15:57:19,190[nio-8080-exec-7]  [INFO ] [o.e.c.a.w.s.WorkspaceRuntimes 238]   - Starting workspace 'admin/workspace8pnbim' with id 'workspace0p8c84wq7r55599f' by user 'admin'
Exception in thread "OkHttp Dispatcher" io.fabric8.kubernetes.client.KubernetesClientException: An error has occurred.
	at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:61)
	at io.fabric8.kubernetes.client.KubernetesClientException.launderThrowable(KubernetesClientException.java:52)
	at io.fabric8.openshift.client.internal.OpenShiftOAuthInterceptor.authorize(OpenShiftOAuthInterceptor.java:122)
	at io.fabric8.openshift.client.internal.OpenShiftOAuthInterceptor.intercept(OpenShiftOAuthInterceptor.java:71)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:92)
	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.java:67)
	at okhttp3.RealCall.getResponseWithInterceptorChain(RealCall.java:179)
	at okhttp3.RealCall$AsyncCall.execute(RealCall.java:129)
[...]

Issue #5902, opened 2017-08-04
Fix leak of connection to OpenShift


sleshchenko:
When OpenShift workspace is started we have following output in logs of tomcat:
```
2017-08-03 15:44:15,559[aceSharedPool-0]  [INFO ] [o.e.c.a.w.s.WorkspaceManager 411]    - Workspace 'che:testx19' with id 'workspace8go3un4udnwr1pfd' started by user 'undefined'
Aug 03, 2017 3:49:01 PM okhttp3.internal.platform.Platform log
WARNING: A connection to https://192.168.42.246:8443/ was leaked. Did you forget to close a response body? To see where this was allocated, set the OkHttpClient logger level to FINE: Logger.getLogger(OkHttpClient.class.getName()).setLevel(Level.FINE);
```
Creation of all OpenShift clients placed in try with resources section.
So it is needed to investigate why leak is done and fix it.

l0rd:
[This issue](https://issues.jboss.org/browse/CHE-180?focusedCommentId=13396715&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13396715) has some analysis about various problems (including memory leaks) in the OpenShift client. And since we don't use JIRA anymore we are tracking that [here](https://github.com/redhat-developer/rh-che/issues/198).

sleshchenko:
@l0rd As far as I understand described issues mostly are not in OpenShift connector but in fabric8 kubernetes client. Some of them have the corresponding issue for kubernetes-client, some no. Also, there is no described problem that is described here. 
Do you think this is our incorrect usage of kubernetes client or they have a bug? If they have, then should I create issue for kubernetes-client and link it with this issue?

l0rd:
@sleshchenko I have no idea. @amisevsk @ibuziuk and @snjeza have worked on that and may know if it's a kubernetes client bug or how we use it in OpenShiftConnector.

sleshchenko:
@amisevsk @ibuziuk and @snjeza Do you have any thought about it?

snjeza:
@sleshchenko I have described the issue in https://issues.jboss.org/browse/CHE-180?focusedCommentId=13396715&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13396715

sleshchenko:
@snjeza I see. Thanks

akorneta:
[...]

PR #7629, opened 2017-11-29
Fixes various problems usage of OpenShift client

akorneta:
### What does this PR do?
Replaces the usage of one `OpenShiftClient` per action to one client per che-server. The reason why we do that is the `OpenShiftClient` holds an instance of `OkHttpClient` inside and each time when we create new `OpenShiftClient` we create new `OkHttpClient` with all the components inside,  on close, the `OpenShiftClient` prevent cancel all calls currently that are executed by `OkHttpClient`, close all the connections and stop the dispatcher executor service (that undoubtedly is overhead).
Adds closing of watcher on pod removing.
Disables abnormal stop and container events watching.
Launches bootstrapper as a detached process.

#7418
This problem is not reproduced in case of using one OpenShift client per che server: #5902

### DIFF
--- a/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftClientFactory.java
+++ b/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftClientFactory.java
@@ -16,14 +16,25 @@ import io.fabric8.openshift.client.DefaultOpenShiftClient;
 import io.fabric8.openshift.client.OpenShiftClient;
 import io.fabric8.openshift.client.OpenShiftConfig;
 import io.fabric8.openshift.client.OpenShiftConfigBuilder;
+import javax.annotation.PreDestroy;
 import javax.inject.Inject;
 import javax.inject.Named;
+import javax.inject.Singleton;
 import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
 import org.eclipse.che.commons.annotation.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-/** @author Sergii Leshchenko */
+/**
+ * @author Sergii Leshchenko
+ * @author Anton Korneta
+ */
+@Singleton
 public class OpenShiftClientFactory {
-  private final OpenShiftConfig config;
+
+  private static final Logger LOG = LoggerFactory.getLogger(OpenShiftClientFactory.class);
+
+  private final UnclosableOpenShiftClient client;
 
   @Inject
   public OpenShiftClientFactory(
@@ -52,7 +63,7 @@ public class OpenShiftClientFactory {
     if (doTrustCerts != null) {
       configBuilder.withTrustCerts(doTrustCerts);
     }
-    config = configBuilder.build();
+    this.client = new UnclosableOpenShiftClient(configBuilder.build());
   }
 
   /**
@@ -61,6 +72,30 @@ public class OpenShiftClientFactory {
    * @throws InfrastructureException if any error occurs on client instance creation.
    */
   public OpenShiftClient create() throws InfrastructureException {
-    return new DefaultOpenShiftClient(config);
+    return client;
+  }
+
+  @PreDestroy
+  private void cleanup() {
+    try {
+      client.doClose();
+    } catch (RuntimeException ex) {
+      LOG.error(ex.getMessage());
+    }
+  }
+
+  /** Decorates the {@link DefaultOpenShiftClient} so that it can not be closed from the outside. */
+  private static class UnclosableOpenShiftClient extends DefaultOpenShiftClient {
+
+    public UnclosableOpenShiftClient(OpenShiftConfig config) {
+      super(config);
+    }
+
+    @Override
+    public void close() {}
+
+    void doClose() {
+      super.close();
+    }
   }
 }
--- a/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
+++ b/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
@@ -107,8 +107,9 @@ public class OpenShiftInternalRuntime extends InternalRuntime<OpenShiftRuntimeCo
       for (Route route : osEnv.getRoutes().values()) {
         createdRoutes.add(project.routes().create(route));
       }
-      project.pods().watch(new AbnormalStopHandler());
-      project.pods().watchContainers(new MachineLogsPublisher());
+      // TODO https://github.com/eclipse/che/issues/7653
+      // project.pods().watch(new AbnormalStopHandler());
+      // project.pods().watchContainers(new MachineLogsPublisher());
 
       createPods(createdServices, createdRoutes);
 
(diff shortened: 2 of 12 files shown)


infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/bootstrapper/OpenShiftBootstrapper.java
```diff
@@ -103,7 +103,8 @@ protected void doBootstrapAsync(String installerWebsocketEndpoint, String output
             + Integer.toString(installerTimeoutSeconds)
             + " -file "
             + BOOTSTRAPPER_DIR
-            + CONFIG_FILE);
+            + CONFIG_FILE
+            + " &>/dev/null &");
```
garagatyi: Please add a comment why you've added this. It is not obvious that it is a way to make process detached.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
```diff
@@ -107,8 +107,9 @@ protected void internalStart(Map<String, String> startOptions) throws Infrastruc
       for (Route route : osEnv.getRoutes().values()) {
         createdRoutes.add(project.routes().create(route));
       }
-      project.pods().watch(new AbnormalStopHandler());
-      project.pods().watchContainers(new MachineLogsPublisher());
+      // TODO link with issue
```
garagatyi: Please add the corresponding issue.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftClientFactory.java
```diff
@@ -52,7 +59,7 @@ public OpenShiftClientFactory(
     if (doTrustCerts != null) {
       configBuilder.withTrustCerts(doTrustCerts);
     }
-    config = configBuilder.build();
+    this.client = new DefaultOpenShiftClient(configBuilder.build());
```
garagatyi: Consider adding a decorator that would eliminate the risk of calling close on the client.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/project/OpenShiftPods.java
```diff
@@ -65,6 +64,9 @@
   private static final Pattern CONTAINER_FIELD_PATH_PATTERN =
       Pattern.compile("spec.containers\\{(?<" + CONTAINER_NAME_GROUP + ">.*)}");
 
+  // TODO make it configurable
+  public static final int POD_REMOVAL_TIMEOUT_MIN = 5;
```
garagatyi: Please create an issue for this todo or remove it

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/project/OpenShiftPods.java
```diff
@@ -381,28 +389,38 @@ public void delete() throws InfrastructureException {
...
+        removed.get(POD_REMOVAL_TIMEOUT_MIN, TimeUnit.MINUTES);
       } catch (InterruptedException e) {
         Thread.currentThread().interrupt();
         throw new InfrastructureException(
             "Interrupted while waiting for pod removal. " + e.getMessage());
       } catch (ExecutionException e) {
         throw new InfrastructureException(
             "Error occurred while waiting for pod removing. " + e.getMessage());
+      } catch (TimeoutException ex) {
+        throw new InfrastructureException("Pods removal timeout reached " + ex.getMessage());
```
garagatyi, line 403: Can you elaborate on what will happen in this situation with cleanup of other resources and with the possible following start of the same workspace?

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/project/OpenShiftPods.java
```diff
@@ -381,28 +389,38 @@ public void delete() throws InfrastructureException {
...
-      podResource.watch(new DeleteWatcher(deleteFuture));
+      final Watch watch = podResource.watch(new DeleteWatcher(deleteFuture));
+
       podResource.delete();
-      return deleteFuture;
+
+      return deleteFuture
+          .thenAccept((v) -> watch.close())
+          .exceptionally(
+              e -> {
```
garagatyi: Shouldn't we also log the exception?

garagatyi (commented):
Looks good. The investigation is cool!

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/project/OpenShiftPods.java
```diff
@@ -381,28 +389,38 @@ public void delete() throws InfrastructureException {
...
+          clientFactory.create().pods().inNamespace(namespace).withName(name);
       final CompletableFuture<Void> deleteFuture = new CompletableFuture<>();
-      podResource.watch(new DeleteWatcher(deleteFuture));
+      final Watch watch = podResource.watch(new DeleteWatcher(deleteFuture));
+
       podResource.delete();
-      return deleteFuture;
+
+      return deleteFuture
+          .thenAccept((v) -> watch.close())
```
sleshchenko: Looks like it's the same as 
```
      return deleteFuture
          .handle((v, e) -> {
            watch.close();
            return null;
          });
```
But it will supress rethrowing of exception from original delete future. Is it OK?

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/project/OpenShiftPods.java
akorneta: added exception logging

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/project/OpenShiftPods.java
akorneta: It is similar to other exception that is possible on cleanup, I would say that we should revise cleanup in general. Because if any step of cleanup is failed then, other resources won't be cleaned. I'll create an issue for that.

akorneta: https://github.com/eclipse/che/issues/7660

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftClientFactory.java
```diff
@@ -61,6 +69,30 @@ public OpenShiftClientFactory(
...
+      client.doClose();
+    } catch (RuntimeException ex) {
+      LOG.error(ex.getMessage());
+    }
+  }
+
+  /** Decorates the {@link DefaultOpenShiftClient} so that it can not be closed from the outside. */
+  private static class UnclosableOpenShiftClient extends DefaultOpenShiftClient {
+
+    public UnclosableOpenShiftClient(OpenShiftConfig build) {
```
sleshchenko: Let's rename it to config =)

sleshchenko (approved):

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftClientFactory.java
```diff
@@ -16,14 +16,22 @@
...
 import javax.inject.Inject;
 import javax.inject.Named;
+import javax.inject.Singleton;
 import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
 import org.eclipse.che.commons.annotation.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** @author Sergii Leshchenko */
+@Singleton
```
sleshchenko: Don't be shy to add yourself as an author :wink:

garagatyi (approved):
Good job!

ahmadmiz (commented):
Ok

PR #7636, opened 2017-11-29
feat: Integration with Smart Testing tool & introduced experimental features settings

MatousJobanek:
### What does this PR do?

In this PR I'm introducing two new features.

#### 1. Smart Testing integration
This is the main feature and also the reason for this PR. The code contains all necessary bits to provide [Smart Testing](https://github.com/arquillian/smart-testing/) features in Che. Smart Testing is a tool that speeds up the test running phase by reordering test execution plan to increase a probability of fail-fast execution and thus give you faster feedback about your project’s health. This means that if a user uses the optimized test execution

optimized-test-item

then based on the set strategies and mode (in a [configuration](https://github.com/MatousJobanek/che/blob/96f22829ca1360af567d7421e8cc896bdf40a7d3/plugins/plugin-optimized-testing/che-plugin-optimized-testing-server/src/test/resources/configuration/smart-testing.yaml) file), the runner will select/order the most important tests (`failed`/`changed`/`new`/`affected`)
To make it happen, I had to change a few classes in the current implementation (created some templates, opened implementation using abstract methods, added additional parameters etc.) 
   * One of the key points is a new abstraction of JavaTestFinder. This is actually responsible for finding/ordering/selecting the test classes to be executed. 
   * To make it clear which mode and strategies have been applied, I added a new information label in the "Test Results Panel"

test-result-panel

#### 2. Experimental features settings
As this Smart Testing integration can be considered as an experimental feature, I've created a new category in the Preferences window. There is a possibility to enable/disable the Smart Testing feature. 

experimental-feature-settings

When it is disabled, the item "Optimized Test" is not visible and the statistics about failures (for failure strategy) are not stored.


tsmaeder:
Just a couple of comments:

1. There is no associated feature request for this, can you open one, please?
2. Should this go into che or rh-che?
3. If che, the target should be che6, not master, IMO.

Generally, I think we should discuss if we want to include this in the base che distribution or not. What does @slemeur think?

slemeur:
We have initiated a discussion on the mailing list about this particular PR: 
https://dev.eclipse.org/mhonarc/lists/che-dev/msg02383.html

We will be opening a community vote during the week. 

Few more notes:
- Agree we should target che 6. I don't believe there is too much changes. Adding @svor to comment and provide guidance on that.

skabashnyuk:
@tsmaeder @slemeur @vparfonov  are we still interested in this PR?

slemeur:
We are still definitely interested into this PR, but this would need to be reworked as a plugin for the next generation of Che.

PR #7639, opened 2017-11-29
Fix for part of redhat-developer/rh-che#431

JPinkney:

### What does this PR do?
This PR fixes an issue in OpenShiftConnector that happens when the tag part of repository:tag is less then 20 characters due to long repository name.

Part of redhat-developer/rh-che/issues/431
Part of redhat-developer/rh-che/issues/334

#### Release Notes

N/A


l0rd:
Thanks @JPinkney! How have you tested this PR? With which stacks and on which platform?

JPinkney:
I tested it on minishift with centos-nodejs (One of the ones previously throwing the error that this fixes), centos-go and various other non centos stacks as well

l0rd (approved):

PR #7649, opened 2017-11-30
Unification between Docker and Openshift infrastructures

skabashnyuk:
### What does this PR do?
We would like to make sure that all our components and configuration working consistently with all our major infrastructure implementation.  One of the components that is not working like that is ```sshd ``` functionality. With this pr we want's to drop off support of org.eclipse.che.ssh installer. Main reasons are:
1. It doesn't work on OpenShift or infrastructure that is not allowing ```sudo```.
2. It's quite hard to make universal installer for sshd
3. We have a lot of ```if``` statement that is quite hard to support.
4. There is a replacement for this functionality with build-in terminal
5. There are quite simple steps to restore this functionality for individuals who still want's it.

#### Release Notes
```org.eclipse.che.ssh``` installer has been removed

#### Docs PR
TODO


tolusha (approved):

skabashnyuk:
ci-test

benoitf (changes requested):
I haven't see any discussion prior to this PR

AFAIK I don't see why we should remove SSH feature.

skabashnyuk:
>I haven't see any discussion prior to this PR

We  can discuss here if you want

>AFAIK I don't see why we should remove SSH feature.

See description above. https://github.com/eclipse/che/pull/7649#issue-278108593

benoitf:
@skabashnyuk AFAIK This is not why I'm calling a reason. It's not because I have a tyre that is not working that I should remove all wheels.

slemeur:
If I'm not mistaken, this PR will:
- remove `ssh` capabilities that were there out-of-the box for a user who is using Che on Docker
- break the `sync` capability that is available with the CLI

Instead of removing features (sic...), why not keeping the agent and enabling it only when Che is running on Docker?

benoitf:
I think it breaks 
- automatic ssh keys for a workspace
- `che dir` feature as well

skabashnyuk:
Will be replaced with less invasive things. As a result, user will have to explicitly enable sshd if he wants https://github.com/eclipse/che/issues/7685

Issue #7673, opened 2017-12-01
Unexpected fail of AutocompleteProposalJavaDocTest selenium test


Ohrimenko1988:
### Description
 
We have unexpected fail of AutocompleteProposalJavaDocTest selenium test. It looks like problem in getElementSrcLink method in the CodenvyEditor page-object, which get src link from element, we need to add wait for src link to this method.

*Exception:*
java.lang.AssertionError: src link in the context java doc window does not attached

*Report with reproduced problem:*
https://ci.codenvycorp.com/view/qa/job/che-multiuser-integration-tests-che6/39/Selenium_tests_report/

Ohrimenko1988:
Done https://github.com/eclipse/che/pull/7675

PR #7675, opened 2017-12-01
Add wait src link text in the getElementSrcLink method in the CodenvyEditor page-object

Ohrimenko1988:
### What does this PR do?
Add wait src link text in the getElementSrcLink method in the CodenvyEditor page-object
In the current state this method try to get src link only once and if something went wrong it throws exception. But with this changes, method will send requests every 500 millisecond until it gets link or timeout will over.

Issue https://github.com/eclipse/che/issues/7673


dmytro-ndp (approved):
Can you, please, explain, what is the reason of the changes?

Ohrimenko1988:
I updated a description

tolusha (approved):

PR #7677, opened 2017-12-01
Add id for cursor view and apply this for the tests

musienko-maxim:
### What does this PR do?
* Apply the id for cursor position view for editor. Before we've had a locator which build searching from active editor to cursor position by xpath. But this in some specific cases brought unstable into a caouple tests. This PR set id for an unambiguous choice of the webElement.


musienko-maxim:
ci-build

vparfonov (approved):

Issue #7571, opened 2017-11-27
Investigate how CHE server can be configured in case of deploy on OCP


riuvshin:
Today we have many issues with `deploy_che.sh` and `che-openshift.yml` due to lots of `sed` commands that really hard to support and also we are not able to change all CHE configuration parameters on OCP. For example just to add ability to configure oAuth for CHE I used to extend deploy script https://github.com/eclipse/che/pull/7212/files which is really bad thing to do if we are talking about whole set of CHE configurations that way does not work at all.

~~So the idea is try to generate `che-openshift.yml` from a puppet template and make deploy script really simple.~~

We decided that using CHE CLI for that will make this very complex and non trivial in support, So I will try to get rid of seds and allow to set any config for CHE server with minimum effort for a user

gorkem:
What is the advantage of using a puppet template over any other? Would openshift templates be enough here?

riuvshin:
~~We use CHE CLI to manage CHE configuration, and in CHE CLI we use puppet. So we have all CHE configurations in the CLI and we just need to generate `che-openshift.yml` based on that configuration instead of doing lots of `sed`. Btw `che-openshift.yml` is already under puppet but as a file not as a template: https://github.com/eclipse/che/tree/che6/dockerfiles/init/modules/openshift/files/scripts
With that approach using CHE CLI we should be able to run CHE on docker OR on OS with exactly same set of configuration~~

After some investigation it becomes clear that CHE CLI is overkill and makes it really complicated. I was not familiar with `openshift templates` and it seems it should be something better... And I should investigate that.

PR #7692, opened 2017-12-01
 Allow add che configurations in case of deploy on Openshift

riuvshin:
Allow add che configurations in case of deploy on Openshift

- removed alot of error prone `sed` commands
- introduced `che-config`  file which should be used to add CHE configurations while deploy CHE master

needed for: https://github.com/eclipse/che/issues/7571

tested on che single / multi user on OCP


benoitf (approved):

ghost (approved):

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
```diff
@@ -361,33 +387,14 @@ fi
 echo
 echo "[CHE] Deploying Che on ${OPENSHIFT_FLAVOR} (image ${CHE_IMAGE})"
 
-
 DEFAULT_CHE_DEPLOYMENT_FILE_PATH=./che-openshift.yml
 CHE_DEPLOYMENT_FILE_PATH=${CHE_DEPLOYMENT_FILE_PATH:-${DEFAULT_CHE_DEPLOYMENT_FILE_PATH}}
+DEFAULT_CHE_CONFIG_MAP_FILE_PATH="$(pwd)/che-config-map"
```
garagatyi: Can you check whether this fails in case user call deploy script, not from the current directory?

garagatyi (approved):

ghost (approved):

dockerfiles/init/modules/openshift/files/scripts/che-config-map
```diff
@@ -0,0 +1,34 @@
...
+CHE_KEYCLOAK_AUTH__SERVER__URL: ${CHE_KEYCLOAK_AUTH__SERVER__URL}
+CHE_KEYCLOAK_CLIENT__ID: ${CHE_KEYCLOAK_CLIENT__ID}
+CHE_KEYCLOAK_GITHUB_ENDPOINT: ${KEYCLOAK_GITHUB_ENDPOINT}
+CHE_KEYCLOAK_OSO_ENDPOINT: ${CHE_KEYCLOAK_OSO_ENDPOINT}
+CHE_KEYCLOAK_REALM: ${CHE_KEYCLOAK_REALM}
+CHE_LOCAL_CONF_DIR: /etc/conf
+CHE_LOGS_DIR: /data/logs
+CHE_LOG_LEVEL: ${CHE_LOG_LEVEL}
+CHE_MULTIUSER: ${CHE_MULTIUSER}
+CHE_OAUTH_GITHUB_CLIENTID:
```
sleshchenko: Should it be empty value here or `${CHE_OAUTH_GITHUB_CLIENTID}` env var?

dockerfiles/init/modules/openshift/files/scripts/che-config-map
```diff
@@ -0,0 +1,34 @@
...
+CHE_OAUTH_GITHUB_CLIENTSECRET:
+CHE_OAUTH_GITHUB_FORCEACTIVATION: "true"
+CHE_PORT: "8080"
+CHE_PREDEFINED_STACKS_RELOAD__ON__START: ${CHE_PREDEFINED_STACKS_RELOAD}
+CHE_SERVER_JAVA_OPTS: "-XX:+UseG1GC -XX:+UseStringDeduplication -XX:MinHeapFreeRatio=20 -XX:MaxHeapFreeRatio=40 -XX:MaxRAM=700m -Xms256m"
+CHE_WEBSOCKET_ENDPOINT:  ${WS_PROTOCOL}://che-${OPENSHIFT_NAMESPACE_URL}/api/websocket
```
sleshchenko: I'd move declaration of this env var somewhere near [CHE_API](https://github.com/eclipse/che/pull/7692/files#diff-1e554eb231a824102dbb255a1ee83d0fR3) because they are related

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
sleshchenko: Actually, it is not config map but just env variables list. Maybe it'd better rename it.

dockerfiles/init/modules/openshift/files/scripts/che-config-map
```diff
@@ -0,0 +1,34 @@
+# To put some extra CHE server configurations just add here CHE ENV vars following format:
+# KEY: "value"
+CHE_API: ${HTTP_PROTOCOL}://che-${OPENSHIFT_NAMESPACE_URL}/api
+CHE_DEBUG_SERVER: ${CHE_DEBUG_SERVER}
+CHE_HOST: che-${OPENSHIFT_NAMESPACE_URL}
+CHE_INFRASTRUCTURE_ACTIVE: openshift
+CHE_INFRA_OPENSHIFT_BOOTSTRAPPER_BINARY__URL: ${HTTP_PROTOCOL}://che-${OPENSHIFT_NAMESPACE_URL}/agent-binaries/linux_amd64/bootstrapper/bootstrapper
+CHE_INFRA_OPENSHIFT_MACHINE__START__TIMEOUT__MIN: "5"
+CHE_INFRA_OPENSHIFT_MASTER__URL: ""
```
sleshchenko: Let's set value as env var to make it possible to override it

sleshchenko (approved):
Good job. :+1:

dockerfiles/init/modules/openshift/files/scripts/che-config-map
riuvshin: I thought by default yes, it should be empty and I probably have to remove it from deploy script. as this file is supposed to be edited to set this configuration. But if we will set value to `${CHE_OAUTH_GITHUB_CLIENTID}` that would respect global ENVs and will pick it if it is set. So I think we could do same for all ENVs that are empty by default.

dockerfiles/init/modules/openshift/files/scripts/che-config-map
riuvshin: which exactly value?

dockerfiles/init/modules/openshift/files/scripts/che-config-map
riuvshin: can you please specify which exactly ENV you are talking about

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
riuvshin: done

riuvshin: done

dockerfiles/init/modules/openshift/files/scripts/che-config-map
sleshchenko: I'd say 
`CHE_INFRA_OPENSHIFT_MASTER__URL: ${CHE_INFRA_OPENSHIFT_MASTER__URL}`

As far as I understand it will be empty if `CHE_INFRA_OPENSHIFT_MASTER__URL` env var is missed, isn't it?

dockerfiles/init/modules/openshift/files/scripts/che-config-map
sleshchenko: I'm talking about `CHE_WEBSOCKET_ENDPOINT`. I think it'd better to group env var like it was before
```
CHE_HOST: che-${OPENSHIFT_NAMESPACE_URL}
CHE_API: ${HTTP_PROTOCOL}://che-${OPENSHIFT_NAMESPACE_URL}/api
CHE_WEBSOCKET_ENDPOINT:  ${WS_PROTOCOL}://che-${OPENSHIFT_NAMESPACE_URL}/api/websocket
```

riuvshin: Done

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
riuvshin: done

PR #7703, opened 2017-12-04
Fix configuring of OPENSHIFT_NAMESPACE_URL in deploy script

sleshchenko:
### What does this PR do?
Fixes configuring of `OPENSHIFT_NAMESPACE_URL` in deploy script.
Namespace URL should be configured by a user and it equals to `{CHE_OPENSHIFT_PROJECT}.{ROUTING_SUFFIX}`. But in this case, a user should know default value of `CHE_OPENSHIFT_PROJECT` or use custom value if he changed it. So it would be more convenient to configure `ROUTING_SUFFIX` only and construct namespace URL automatically.

Fixes checking of `IMAGE_PULL_POLICY` in `ocp.sh` script.

N/A


benoitf (approved):

ghost (approved):

riuvshin (approved):
|_,00|<5 [¬ 0 0 [) †0 ][\\/][∑

garagatyi (approved):

Issue #7704, opened 2017-12-04
Make it possible to update che server deploymentConfig 


ghost:
Currently, OpenShift deploy script does not handle situations when Che server configuration needs to be updated.

The script should be able to apply new envs to deployment config and trigger a new deployment

riuvshin:
Done.
to apply just a config change use:
`deploy_che.sh -c rollupdate`
`ocp.sh --update`

PR #7718, opened 2017-12-04
Allow to update che server deploymentConfig

riuvshin:
Allow to update che server deploymentConfig

here I replaced `oc rollout latest che` with `oc apply` as we have 
```
    triggers:
    - type: ConfigChange
```
But I'm not 100% sure that it is a equivalent so **dear** @reviewers if you have any concerns about that please feel free to share your opinion.

fixes: https://github.com/eclipse/che/issues/7704

### DIFF
--- a/dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
+++ b/dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
@@ -294,7 +294,7 @@ echo "done!"
 
 COMMAND_DIR=$(dirname "$0")
 
-if [[ "${CHE_MULTIUSER}" == "true" ]] && [[ "${COMMAND}" != *"rollupdate"* ]]; then
+if [[ "${CHE_MULTIUSER}" == "true" ]] && [[ "${COMMAND}" == "deploy" ]]; then
     if [ "${CHE_DEDICATED_KEYCLOAK}" == "true" ]; then
         "${COMMAND_DIR}"/multi-user/deploy_postgres_and_keycloak.sh
     else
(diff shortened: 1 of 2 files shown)


ghost (approved):

benoitf (approved):

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
```diff
@@ -275,38 +287,14 @@ echo -n "[CHE] Switching to \"${CHE_OPENSHIFT_PROJECT}\"..."
...
-
 # -------------------------------------------------------------
 # Deploying secondary servers
 # for postgres and optionally Keycloak
 # -------------------------------------------------------------
 
 COMMAND_DIR=$(dirname "$0")
 
-if [ "${CHE_MULTIUSER}" == "true" ]; then
+if [[ "${CHE_MULTIUSER}" == "true" ]] && [[ "${COMMAND}" != *"rollupdate"* ]]; then
```
sleshchenko: Will be this statement true if the command is `cleanup`?

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
```diff
@@ -344,6 +332,29 @@ else
   CHE_KEYCLOAK_CLIENT__ID=${CHE_KEYCLOAK_CLIENT__ID:-"openshiftio-public"}
 fi
 
+# -------------------------------------------------------------
+# If command == cleanup then delete all openshift objects
+# -------------------------------------------------------------
+if [ "${COMMAND}" == "cleanup" ]; then
```
sleshchenko, line 338: Can you explain why did you move these lines?

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
riuvshin: good catch Sergio! Fixed! Thanks

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
riuvshin: Thanks for the question Sergio! By my nature I love moving lines around even without any reasons, but not in that case... Here I moved them (with a pleasure I must admit) because in case of `rollupdate ` we call `get_che_pod_config | oc apply -f -` and here before we applied configuration update we need all ENV vars used in `DEFAULT_CHE_CONFIG_FILE_PATH=./che-config` to be inited. And as you can see here before my changes these lines were before `# Setting Keycloak-related environment variables
` section where we init KC related ENVs so that's why.

sleshchenko (approved):

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
sleshchenko: Thanks for the explanation!

garagatyi (approved):

Issue #7705, opened 2017-12-04
Unexpected failing of Swagger test on CI for single user assembly


musienko-maxim:
### Description
After fix Swagger test on multiuser assembly - the test passes as well. But on single user assembly this one became unstable. We should clarify reason and fix it
**OS and version:**  Chrome browser, CHE-6.0.0.-M3

musienko-maxim:
ci-test

PR #7747, opened 2017-12-06
Add methods for getting cursor position from split editor

musienko-maxim:
### What does this PR do?
* Add methods for correct getting  cursor position from split editors
* Fix CheckRestoringSplitEditorTest selenium test
* Rename Editor class (Codenvy preffix is not actual)
https://github.com/eclipse/che/issues/7705
@tolusha, @vparfonov, @dmytro-ndp

### DIFF
--- a/selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/editor/CheckRestoringSplitEditorTest.java
+++ b/selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/editor/CheckRestoringSplitEditorTest.java
@@ -10,6 +10,10 @@
  */
 package org.eclipse.che.selenium.editor;
 
+import static org.eclipse.che.selenium.core.constant.TestTimeoutsConstants.LOAD_PAGE_TIMEOUT_SEC;
+import static org.eclipse.che.selenium.core.project.ProjectTemplates.MAVEN_JAVA_MULTIMODULE;
+import static org.eclipse.che.selenium.pageobject.CodenvyEditor.TabAction.SPIT_HORISONTALLY;
+import static org.eclipse.che.selenium.pageobject.CodenvyEditor.TabAction.SPLIT_VERTICALLY;
 import static org.testng.Assert.fail;
 
 import com.google.common.base.Joiner;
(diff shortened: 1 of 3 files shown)


musienko-maxim:
ci-test

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/Editor.java
```diff
@@ -0,0 +1,1914 @@
...
+
+    /**
+     * set cursor in specified position for current visible editor
+     *
+     * @param positionLine
+     *         is the specified number line
+     * @param positionChar
+     *         is the specified number char
+     */
+    public void setCursorToDefinedLineAndCharForCurrentVisibleEditor(int positionLine, int positionChar) {
```
dmytro-ndp: Method name **setCursorToDefinedLineAndCharForCurrentVisibleEditor()** looks to long, IMHO.
Can we use here **goToPosition()**?

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/Editor.java
```diff
@@ -0,0 +1,1914 @@
...
+
+    /**
+     * wait specified values for Line and Char cursor positions in the Codenvy editor
+     *
+     * @param linePosition
+     *         expected line position
+     * @param charPosition
+     *         expected char position
+     */
+    public void waitLineAndCharForCurrentVisibleEditor(final int linePosition, final int charPosition) {
```
dmytro-ndp: IMHO, it's better to rename the method to have a shorter name **waitCursorInPosition()** .

dmytro-ndp:
@musienko-maxim: if **Editor** page object operates with the current active visible editor only, the words "_VisibleEditor_", "_CurrentVisibleEditor_", "_ActiveEditor_" look redundant in the methods' names and so could be omitted with saving the meaning of methods, IMHO.
If it linked to split editors, we should select the only one word between the `{Active, Visible, Current}`. As for me, **Active** is the most suitable for this case.

tolusha:
It could be nice of to split commit into several ones not to spend time browsing through refactoring stuff.

PR #7775, opened 2017-12-07
Introduce a concept of internal machine server which is accessible from the inside of a workspace only

### LINKED ISSUE #7561: Add an ability to make Servers of LSs non-accessible for non-authenticated access from external network
### Description
While we want to provide an ability to run language server as a [separate container](https://github.com/eclipse/che/issues/7554) and identify them as servers in workspace runtime [via attributes](https://github.com/eclipse/che/issues/7560), we need to have an ability to limit access to the LS. 
Without that LS that listens TCP socket would allow non-authenticated access to files that are accessible from the LS's container. Because of that, we want to add a concept of internal server. This server should be treated by workspace infrastructure in a different way than the regular server.
URL of such a servers must be non-accessible outside of workspace's containers.
Since these servers are useless for UD, IDE clients we can hide them from workspace runtime returned by Workspace API by default. To get runtime with internal servers client should add some query parameter, e.g. `?internal=true` (feel free to provide your suggestions).
Tasks:
1. Create an attribute that identifies a server as internal.
1. Make workspace infras use URLs that are accessible from workspace only for servers identified as internal.
1. Show internal servers in the runtime of a workspace only when a special query parameter is provided by a client.

### Reproduction Steps

**OS and version:**    

**Diagnostics:**

garagatyi:
### What does this PR do?
Note: you can review this PR commit-by-commit.
Adds a concept of an "internal server" that use infrastructure-specific workspace-scoped communication mechanism to allow communication from workspace machines only.
This allows defining a server that is accessible from other machines(containers) of the same workspace but not accessible from the internet.
This server doesn't have to be secured by any authentication since alien users cannot access it.
This allows using workspace runtime information about private workspace ports for a service discovery.
Example of flow:
1. A user adds a machine with an LS to a workspace and specifies some port as an internal server. 
1. Workspace API starts workspace and doesn't open that port to the world because it is marked as internal.
1. Another service started as part of the workspace(ws-agent) wants to discover this LS. It fetches workspace runtime information and figures out how to connect to the LS.

Supported infras:
- Docker
- OpenShift

How to use it:
1. Add ServerConfig object to a machine.
1. Add attributes field to the server config.
1. Add attribute `internal` with value `true`.
1. This server will be opened as internal. It won't be available at runtime on GET Workspace API by default
1. To get this server from runtime use GET Workspace API with query parameter `?includeInternalServers=true` or `?includeInternalServers=` or `?includeInternalServers`
1. GET Workspace API will return runtime that includes internal servers.

How ServerConfig and Server objects look:
ServerConfig:
```
"my-internal-server" : {
    "port" : "4488",
    "protocol" : "http",
    "path": null,
    "attributes" : {
        "internal" : "true"
    }
}
```
Server:
```
"my-internal-server" : {
    "status" : "UNKNOWN",
    "url" : "http://ls-ts:4488",
    "attributes" : {
        "internal" : "true"
    }
}
```
Where `ls-ts` is some infra-specific value that identifies machine where server is running. 
For the compose recipe it is a name of service/Che machine, for OpenShift it will be generated OpenShift service.

Fixes #7561

### DIFF
--- a/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/config/ServerConfig.java
+++ b/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/config/ServerConfig.java
@@ -11,6 +11,7 @@
 package org.eclipse.che.api.core.model.workspace.config;
 
 import java.util.Map;
+import org.eclipse.che.api.core.model.workspace.runtime.Server;
 import org.eclipse.che.commons.annotation.Nullable;
 
 /**
@@ -19,6 +20,13 @@ import org.eclipse.che.commons.annotation.Nullable;
  * @author Alexander Garagatyi
  */
 public interface ServerConfig {
+
+  /**
+   * {@link ServerConfig} and {@link Server} attribute name which can identify server as internal or
+   * external. Attribute value {@code true} makes a server internal, any other value or lack of the
+   * attribute makes the server external.
+   */
+  String INTERNAL_SERVER_ATTRIBUTE = "internal";
 
   /**
    * Port used by server.
--- a/infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/server/mapping/ServersMapper.java
+++ b/infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/server/mapping/ServersMapper.java
@@ -10,6 +10,8 @@
  */
 package org.eclipse.che.workspace.infrastructure.docker.server.mapping;
 
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -17,21 +19,31 @@ import java.util.List;
 import java.util.Map;
 import org.eclipse.che.api.core.model.workspace.config.ServerConfig;
 import org.eclipse.che.api.workspace.server.model.impl.ServerImpl;
+import org.eclipse.che.api.workspace.server.spi.InternalInfrastructureException;
+import org.eclipse.che.commons.annotation.Nullable;
 import org.eclipse.che.infrastructure.docker.client.json.ContainerPort;
 import org.eclipse.che.infrastructure.docker.client.json.NetworkSettings;
 import org.eclipse.che.infrastructure.docker.client.json.PortBinding;
 
-/** Maps container ports bindings to machine servers. */
+/**
+ * Maps container ports bindings to machine servers.
+ *
+ * @author Yevhenii Voevodin
+ * @author Alexander Garagatyi
+ */
 public class ServersMapper {
 
   private final String hostname;
+  private final String machineName;
 
   /**
-   * Creates mapper using given {@code hostname} as hostname for all the servers urls produced by
-   * mapper.
+   * Creates mapper using given {@code hostname} as hostname for all the servers urls which are
+   * publicly published. For workspace-wide available servers provided {@code machineName} is used
+   * as a hostname.
    */
-  public ServersMapper(String hostname) {
+  public ServersMapper(String hostname, String machineName) {
     this.hostname = hostname;
+    this.machineName = machineName;
   }
 
   /**
@@ -40,9 +52,10 @@ public class ServersMapper {
    * @param ports container ports to map
    * @param configs servers configuration map used to resolve server references
    * @return server reference -> server map. Note that if there is no server configuration for
-   *     container bound port, port+type itself(like 4022/tpc) will be used as a reference
+   *     container bound port, port+type itself(like 4022/tcp) will be used as a reference
    */
-  public Map<String, ServerImpl> map(ContainerPort[] ports, Map<String, ServerConfig> configs) {
+  public Map<String, ServerImpl> map(ContainerPort[] ports, Map<String, ServerConfig> configs)
+      throws InternalInfrastructureException {
     if (ports == null || ports.length == 0) {
       return Collections.emptyMap();
     }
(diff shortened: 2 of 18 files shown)


garagatyi:
ci-build

garagatyi:
ci-test

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -797,4 +805,36 @@ private WorkspaceDto asDtoWithLinksAndToken(Workspace workspace) throws ServerEx
...
+
+  private WorkspaceDto filterVisible(WorkspaceDto workspace, Visibility visibility) {
+    // no runtime - nothing to filter
+    if (workspace.getRuntime() == null) {
+      return workspace;
+    }
+    boolean includeInternal = Visibility.internal.equals(visibility);
+    String internalVisibilityAttribute = Visibility.internal.name();
+    for (MachineDto machine : workspace.getRuntime().getMachines().values()) {
+      HashMap<String, ServerDto> collect =
```
skabashnyuk: shall we make this with old style loops?

artaleks9 (approved):

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/server/mapping/ServersMapper.java
```diff
@@ -151,12 +167,20 @@ private String makeUrl(ContainerPort port, String protocol, String path) {
       path = '/' + path;
     }
 
-    return protocol + "://" + hostname + ':' + port.getPublicPort() + path;
+    int serverPort = port.getPublicPort() != 0 ? port.getPublicPort() : port.getPrivatePort();
+
+    return protocol + "://" + hostname + ':' + serverPort + path;
```
vparfonov: Maybe UriBuilder would be better here?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -184,10 +188,14 @@ public WorkspaceDto getByKey(
...
-          String key)
+          String key,
+      @ApiParam("Servers with which visibility to show")
+          @DefaultValue("external")
+          @QueryParam("visibility")
+          Visibility visibility)
       throws NotFoundException, ServerException, ForbiddenException, BadRequestException {
     validateKey(key);
-    return asDtoWithLinksAndToken(workspaceManager.getWorkspace(key));
+    return filterVisible(asDtoWithLinksAndToken(workspaceManager.getWorkspace(key)), visibility);
```
sleshchenko: When I see `visibility=internal` I expect to get servers which have internal visibility. But in fact, a response will contain all servers (including internal servers). Maybe it would be more logic to use here another parameter for including internal servers

garagatyi: After discussion with @sleshchenko and @gazarenkov we decided to rename query parameter to `includeInternalServers` with no value or boolean. ServerConfig attribute should be renamed to `internal`:"<true|false>". If it is missing it is supposed that server is external.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -797,4 +805,36 @@ private WorkspaceDto asDtoWithLinksAndToken(Workspace workspace) throws ServerEx
...
+    for (MachineDto machine : workspace.getRuntime().getMachines().values()) {
+      HashMap<String, ServerDto> collect =
+          machine
+              .getServers()
+              .entrySet()
+              .stream()
+              .collect(
+                  HashMap::new,
+                  (map, entry) -> {
+                    if (includeInternal
```
sleshchenko: if `includeInternal` is true you can avoid modifying machine servers at all

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/server/mapping/ServersMapper.java
garagatyi: I don't think I can use it, whereas I'm not sure whether I can use UriBuilderImpl because we don't use it in non-tests code.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
garagatyi: tnx

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
garagatyi: done

garagatyi:
@sleshchenko @gazarenkov updated PR in accordance with our discussion

skabashnyuk (approved):

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/config/ServerConfig.java
```diff
@@ -20,6 +21,12 @@
  */
 public interface ServerConfig {
 
+  /**
+   * {@link ServerConfig} and {@link Server} attribute name which can identify server as internal or
+   * external.
```
sleshchenko: I think it could be useful to specify the format of value for this attribute here.

sleshchenko (approved):

vparfonov (approved):

benoitf (approved):

gazarenkov (approved):
TBH, I would consider more isolation of "internal" vs "external" environment.
Up to making different serialization for External vs Internal Environment and making different entry points to access it remotely. But, it seems it requires more things to consider, so I am ok with current solution, it is definitely step ahead, will see then, approved

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/server/mapping/ServersMapper.java
gazarenkov: agree with Vitaliy, this code does not look nice (even if there is very small probability of changing url format in near future ) )

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/config/ServerConfig.java
garagatyi: ok

garagatyi:
@gazarenkov Not sure I understand what you meant. Can you elaborate?

Issue #7601, opened 2017-11-28
[CHE 6] Move creation of folders in PVCs into start runtime phase


sleshchenko:
### Description
OpenShift Infrastructure creates folders (inits subpathes) in PVCs while preparing of an environment. But prepare operation should be quick while now it takes about 15 seconds on Che server that is deployed on OCP and start workspaces on `openshift.io` cluster.

It is needed to move performing such logic into start runtime phase.

PR #7777, opened 2017-12-07
Move PVC creation into OpenShiftInternalRuntime

akorneta:
### What does this PR do?
Moves the PVC creation and preparation of subpaths into `OpenShiftInternalRuntime`.
The reason why we do that:
The process of preparing workspaces for the start must not take much time and give a response to a client as soon as possible, but an operation that creates workspace's subpaths is not instantaneous due to pull an image, mounting PVC e.t.c, so we should not perform it on a preparing stage.
Adds ability to configure use pre-creation of subpaths or not.

#7601


skabashnyuk (approved):

riuvshin (approved):

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/Constants.java
```diff
@@ -37,5 +37,11 @@
   public static final String MACHINE_NAME_ANNOTATION_FMT =
       "org.eclipse.che.container.%s.machine_name";
 
+  /**
+   * The additional property name with the wildcard reserved for workspace id. Formatted property
+   * with the real workspace id is used to get workspace subpaths directories.
```
sleshchenko: Please document which format value has, and that this value will be removed before PVC creation

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
```diff
@@ -182,6 +193,34 @@ private void bootstrapMachine(OpenShiftMachine machine)
...
+   */
+  @VisibleForTesting
+  void preparePVCs(OpenShiftEnvironment osEnv) throws InfrastructureException {
+    final Collection<PersistentVolumeClaim> claims = osEnv.getPersistentVolumeClaims().values();
+    final OpenShiftPersistentVolumeClaims pvcs = project.persistentVolumeClaims();
+    final Set<String> existing =
+        pvcs.get().stream().map(p -> p.getMetadata().getName()).collect(toSet());
+    final String workspaceId = getContext().getIdentity().getWorkspaceId();
+    for (PersistentVolumeClaim pvc : claims) {
+      final String[] subpaths =
```
sleshchenko: I'd say that it should be performed by `WorkspaceVolumesStrategy` but not by Runtime. We could rework it in the following way:
- add a method to WorkspaceVolumesStrategy interface, something like `provision`, that would just add PVC, volumes and volumeMounts;
- then Provisioner could invoke this method instead of `prepare` method;
- Runtime would invoke prepare method which would do PVC creation and pre-creation of subpaths if it is needed.

I think it'd be better because you would be able to encapsulate the logic of subpaths pre-creation into CommonPVCStrategy.

It's just proposal and I'm OK to merge current implementation as soon as possible.

sleshchenko (approved):

benoitf (approved):

sleshchenko (approved):

garagatyi (approved):

Issue #6499, opened 2017-09-29
Investigate problem of failing selenium tests for Find Usage and Navigate File Structure features


SkorikSergey:
**FileStructureNodesTest** and **FindUsagesBaseOperationTest** selenium tests fail while trying to open Node by double click action. If do double click action twice then Node will open. We should clarify of reason why the tests do not work properly.

riuvshin:
seems this is related to: https://github.com/eclipse/che/issues/6524

PR #7783, opened 2017-12-08
Selenium: Add changes to unstable selenium tests

SkorikSergey:
### What does this PR do?
This PR:
- remove try/catch blocks from **FileStructureNodesTest** and **FindUsagesBaseOperationTest** selenium tests;
- add node selection before double click on it in **FindUsages** and **FileStructure** selenium page objects;
- in **waitExpectedResultInEvaluateExpression** method of  **DebugPanel** selenium page object add finding web element before checking that its text is equals to expected .

https://github.com/eclipse/che/issues/6499


dmytro-ndp (approved):
looks good to me

tolusha (approved):

PR #7793, opened 2017-12-08
Fix the OpenshiftClient connection leaks

davidfestal:
### What does this PR do?

Fix the OpenshiftClient connection leaks ... while reusing a common `OkHttpClient` instance as much as possible (not possible in some cases, like `startExec()`, that involve
`InputStreamPumper` and broken pipes).

This PR is related to the issue discussed [here](https://github.com/eclipse/che/issues/5902#issuecomment-346858757) and [here](https://github.com/eclipse/che/issues/5902#issuecomment-348516145), as well as this initial issue of the `OpenshiftConnector` : https://issues.jboss.org/browse/CHE-180

It fixes these issues in a way that is compatible with the CHE-5-based OSIO multi-tenant server.

### DIFF
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftClientFactory.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftClientFactory.java
@@ -23,7 +23,6 @@ import javax.inject.Inject;
 import javax.inject.Singleton;
 import okhttp3.Authenticator;
 import okhttp3.Interceptor;
-import okhttp3.Interceptor.Chain;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Response;
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftClientFactory.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftClientFactory.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright (c) 2012-2017 Red Hat, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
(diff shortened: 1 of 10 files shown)


plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftClientFactory.java
```diff
@@ -0,0 +1,105 @@
...
+@Singleton
+public class OpenShiftClientFactory {
+
+  private OkHttpClient httpClient;
+
+  @Inject
+  public OpenShiftClientFactory(
+      OpenshiftWorkspaceEnvironmentProvider workspaceEnvironmentProvider) {
+    this.httpClient =
+        HttpClientUtils.createHttpClient(workspaceEnvironmentProvider.getDefaultOpenshiftConfig());
```
ibuziuk, line 39: @davidfestal not sure, but formatting does not look right from the first glance

davidfestal: well, it's formatted by the maven formatter plugin afaik

ibuziuk: ok, cool - sorry than ;-)

benoitf:
ci-build

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftClientFactory.java
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  @PreDestroy
+  public void cleanup() {
+    if (httpClient.connectionPool() != null) {
```
sleshchenko: As far as I see checking `httpClient.connectionPool()` on null is redundant at least for 3.6.0 http client https://github.com/square/okhttp/blob/parent-3.6.0/okhttp/src/main/java/okhttp3/OkHttpClient.java#L738. The same for `httpClient.dispatcher()` and `httpClient.dispatcher().executorService()`
Of course up to you to leave these checks or not.

davidfestal:
ci-build

benoitf:
ci-build

benoitf:
ci-build

l0rd (approved):

ibuziuk (approved):

PR #7802, opened 2017-12-10
Add handling for avoid unstable after getting data from the Swagger page

musienko-maxim:
### What does this PR do?
* Sometimes when we get text from swagger page the JSON may be in rendering state. In this case we get invalid data for further DTO converting. For avoiding this problem in the page object has been added loop, with 2 attempts. Each attempt has 500 msec delay for end of rendering page

https://github.com/eclipse/che/issues/7705
@dmytro-ndp, @tolusha, @vparfonov


tolusha (approved):

vparfonov (approved):

dmytro-ndp (approved):
Such workaround looks ugly and we should find more elegant way after we get 100% success rate.

musienko-maxim:
ci-build

PR #7813, opened 2017-12-11
Fix internal servers URL rewriting on OSX

garagatyi:
### What does this PR do?
Fix internal servers URL rewriting on OSX

Related to #7561


sleshchenko (approved):

skabashnyuk (approved):

benoitf:
question: do we have a way to still get internal url instead of localhost for remote clients ?

garagatyi:
@benoitf you have probably got a bug on OSX. Try to re-pull the repo and build Che again - I've already merged a fix.

garagatyi:
@benoitf Probably I misunderstood you. If you ask whether we have internal vs external URLs then answer is no, we don't.

benoitf:
@garagatyi OK thanks for your answer (and yes my question was about internal vs external)

Issue #7163, opened 2017-11-02
[CHE 6] Add an ability to scroll to a file that is opened in the editor


sleshchenko:
### Description
Add an ability to scroll to a file that is opened in the editor.
There is an ability to link editor and project tree. But I don't like it because of unnecessary switching of selection in the project tree which distracts my attention. It also requires some resource to perform refreshing of project tree each time when an active file in the editor is changed.
It would be better to provide an ability to scroll to a file that is opened in the editor once when a user needs it.

mmorhun:
I also miss this feature. +1
I would say navigate instead of scroll because it could be folded and noting to scroll to. Sometimes it is needed to expand folders before scroll.

RomanNikitenko:
Some part of discussion this feature here: https://github.com/eclipse/che/issues/6932

PR #7818, opened 2017-12-11
Add ability to fast reveal and collapse resources

RomanNikitenko:
### What does this PR do?
- Rename 'Reveal resource' to 'Reveal in project explorer' action title
- Add hotkeys:
Ctrl+↓ for Reveal in project explorer action
Ctrl+↑ for Collapse All action
- Move 'Reveal resource' action from 'Edit' menu in 'Options'

options
- Add 'Reveal resource' action to editor context menu

editorcontext
- Add 'Collapse All' action to resource context menu

projecttreecontext

#7163 

### DIFF
--- a/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/CoreLocalizationConstant.java
+++ b/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/CoreLocalizationConstant.java
@@ -26,6 +26,12 @@ public interface CoreLocalizationConstant extends Messages {
   @Key("extension.category")
   String extensionCategory();
 
+  @Key("action.revealResource.text")
+  String actionRevealResourceText();
+
+  @Key("action.revealResource.description")
+  String actionRevealResourceDescription();
+
   @Key("action.navigateToFile.text")
   String actionNavigateToFileText();
 
--- a/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/actions/CollapseAllAction.java
+++ b/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/actions/CollapseAllAction.java
@@ -12,13 +12,9 @@ package org.eclipse.che.ide.actions;
 
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
-import com.google.web.bindery.event.shared.EventBus;
 import org.eclipse.che.ide.CoreLocalizationConstant;
 import org.eclipse.che.ide.api.action.ActionEvent;
 import org.eclipse.che.ide.api.action.BaseAction;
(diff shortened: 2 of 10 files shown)


RomanNikitenko:
ci-build

RomanNikitenko:
ci-test

vparfonov (approved):

dkulieshov:
Hi,

Do we have plans to have macOS hotkeys  for those actions as well?

RomanNikitenko:
@dkuleshov What macOS hotkeys do you prefer for those actions?

dkulieshov:
I personally have no preference over hotkeys for mentioned actions, perhaps you could check if there are hotkeys for such actions in other macOS applications?

RomanNikitenko:
@dkuleshov  As far as I know, for example, Intellij Idea, does not have hotkeys for these actions. So could you, as macOS device owner, say what hotkeys you use or will be convenient for the case

dkulieshov:
I personally don't use those actions and probably I am not that good at UX to give such advices. Moreover I'm not sure if that is indeed needed as at the moment, as far as I know, we do not properly support hotkeys for macOS. I'm just wondering if there are any plans to have those hotkeys for macOS.

RomanNikitenko:
@dkuleshov I think we should investigate what hotkeys will be convenient for macOS users.
Moreover we need to review existed actions and hotkeys for macOS.
In light of the above I think it should be another issue.

dkulieshov:
Also, if my understanding of the action is correct, on my macOS IntelliJ Idea it is called 'Scroll from source' 
scroll_from_source
 and the hotkey is not available by default, moreover in order to set a hotkey for that action there is a [specific](https://plugins.jetbrains.com/plugin/7606-scroll-from-source) plugin.

RomanNikitenko:
@dkuleshov yes, you are right. FYI: this comment https://github.com/eclipse/che/issues/7163#issuecomment-350674862 has reference on discussion about "Scroll from source" feature

dkulieshov:
@RomanNikitenko Understood, so currently we correctly support only hotkeys for Linux class operation systems, while macOS hotkeys need profound revisal, which obviously goes out of the scope of this issue. Thank you 👍

dkulieshov:
> @dkuleshov yes, you are right. FYI: this comment #7163 (comment) has reference on discussion about "Scroll from source" feature

@RomanNikitenko I'm not arguing, mentioned just in case someone is not in a context of this question here.

RomanNikitenko:
@dkuleshov the same from my side, just FYI and 

> in case someone is not in a context of this question here.

RomanNikitenko:
ci-build

artaleks9 (approved):

musienko-maxim (approved):

PR #7822, opened 2017-12-11
Selenium: Change timeout in the page object 'Preferences'

artaleks9:
### What does this PR do?
* Change timeout in the page object `Preferences` related to running on the CHE6-ocp platform


dmytro-ndp:
Could you, please, add comment into the code with explanation that so big timeout is needed for testing Che 6 on OCP?

tolusha (approved):

artaleks9:
The comment is added

Issue #7512, opened 2017-11-22
Add an ability to set exact number of rerun of selenium tests


dmytro-ndp:
We need to have an ability to rewrite default rerun number = 3 when execute selenium tests.
Number should be accepted by **webdriver.sh** parameter `--rerun=NN`.

Issue #7511, opened 2017-11-22
Adapt webdriver.sh script to run on Multi User Che


dmytro-ndp:
We need to have next features when running selenium tests on Multi User Che with **--multiuser** parameter:
1. To have **--multiuser** parameter in sample commands in test output like the follow:
```
[TEST] Try rerun all tests:
[TEST] 	/home/ndp/projects/che/selenium/che-selenium-test/selenium-tests.sh   --threads=2 -Mlocal --failed-tests
[TEST] 	/home/ndp/projects/che/selenium/che-selenium-test/selenium-tests.sh   --threads=2 -Mgrid --failed-tests
[TEST]
[TEST] Or run them one by one:
[TEST] 	/home/ndp/projects/che/selenium/che-selenium-test/selenium-tests.sh   -Mlocal --test=org.eclipse.che.selenium.assistant.CheckFindActionFeatureInCheTest
[TEST] 	/home/ndp/projects/che/selenium/che-selenium-test/selenium-tests.sh   -Mlocal --test=org.eclipse.che.selenium.dashboard.CreateAndDeleteProjectsTest
```
2. Compare results with proper job on CI https://ci.codenvycorp.com/view/qa/job/che-multiuser-integration-tests/

PR #7823, opened 2017-12-11
Adapt webdriver.sh to Multiuser Che; add ability to set rerun attempts

dmytro-ndp:
### What does this PR do?
- moves `--multiuser` parameter from **selenium-test.sh** to **webdriver.sh**;
- fixed **webdriver.sh** script to compare selenium testing results with multiuser CI job;
- makes it clear the difference between 'CI job' from 'build' in terms of comparison of test results;
- adds an ability to set exact number of rerun of selenium tests.

- #7511
- #7512

#### Docs PR
https://github.com/eclipse/che-docs/pull/323

### DIFF
--- a/selenium/che-selenium-core/bin/webdriver.sh
+++ b/selenium/che-selenium-core/bin/webdriver.sh
@@ -53,13 +53,14 @@ initVariables() {
 
     readonly MAX_RERUN=2
 
+    export CHE_MULTIUSER=${CHE_MULTIUSER:-false}
+
     # CALLER variable contains parent caller script name
     # CUR_DIR variable contains the current directory where CALLER is executed
     [[ -z ${CALLER+x} ]] && { CALLER=$(basename $0); }
     [[ -z ${CUR_DIR+x} ]] && { CUR_DIR=$(cd "$(dirname "$0")"; pwd); }
 
     [[ -z ${API_SUFFIX+x} ]] && { API_SUFFIX="/api/"; }
-    [[ -z ${BASE_ACTUAL_RESULTS_URL+x} ]] && { BASE_ACTUAL_RESULTS_URL="https://ci.codenvycorp.com/view/qa/job/che-integration-tests-che6/"; }
 
     MODE="grid"
     GRID_OPTIONS="-Dgrid.mode=true"
@@ -78,7 +79,6 @@ initVariables() {
     PRODUCT_PROTOCOL="http"
     PRODUCT_HOST=$(detectDockerInterfaceIp)
     PRODUCT_PORT=8080
-    export CHE_MULTIUSER=${CHE_MULTIUSER:-false}
 
     unset DEBUG_OPTIONS
     unset MAVEN_OPTIONS
@@ -506,6 +506,15 @@ detectLatestResultsUrl() {
 fetchActualResults() {
     unset ACTUAL_RESULTS
     unset ACTUAL_RESULTS_URL
+
+    # define the URL of CI job to compare result with it
+    if [[ ${CHE_MULTIUSER} == true ]]; then
+      local nameOfCIJob=che-multiuser-integration-tests-che6
+    else
+      local nameOfCIJob=che-integration-tests-che6
+    fi
+
+    [[ -z ${BASE_ACTUAL_RESULTS_URL+x} ]] && { BASE_ACTUAL_RESULTS_URL="https://ci.codenvycorp.com/view/qa/job/$nameOfCIJob/"; }
 
     local job=$(echo $@ | sed 's/.*--compare-with-ci\W\+\([0-9]\+\).*/\1/')
     if [[ ! ${job} =~ ^[0-9]+$ ]]; then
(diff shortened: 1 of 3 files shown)


selenium/che-selenium-core/bin/webdriver.sh
```diff
@@ -50,21 +50,18 @@ initVariables() {
...
     [[ -z ${CUR_DIR+x} ]] && { CUR_DIR=$(cd "$(dirname "$0")"; pwd); }
 
     [[ -z ${API_SUFFIX+x} ]] && { API_SUFFIX="/api/"; }
-    [[ -z ${BASE_ACTUAL_RESULTS_URL+x} ]] && { BASE_ACTUAL_RESULTS_URL="https://ci.codenvycorp.com/view/qa/job/che-integration-tests-che6/"; }
 
     MODE="grid"
     GRID_OPTIONS="-Dgrid.mode=true"
-    RERUN=false
-
+    RERUN_ATTEMPTS=0
```
tolusha, line 64: Why did you change the default value to `0` ?

selenium/che-selenium-core/bin/webdriver.sh
```diff
@@ -412,13 +425,13 @@ HOW TO of usage:
...
         ${CALLER} <...> --failed-tests
-        ${CALLER} <...> --failed-tests --rerun
+        ${CALLER} <...> --failed-tests --rerun [ATTEMPTS]
 
     Debug selenium test:
         ${CALLER} -Mlocal --test=<TEST> --debug
 
     Analyse tests results:
-        ${CALLER} --compare-with-ci [CI job number]
+        ${CALLER} --compare-with-ci [BUILD NUMBER]
```
tolusha, line 434: `BUILD NUMBER` collocation usually references to the product build number but `CI job number` indicates the number of the job on the CI.

selenium/che-selenium-core/bin/webdriver.sh
```diff
@@ -428,7 +441,7 @@ printRunOptions() {
     echo "[TEST]"
     echo "[TEST] =========== RUN OPTIONS ==========================="
     echo "[TEST] Mode                : "${MODE}
-    echo "[TEST] Rerun failing tests : "${RERUN}
+    echo "[TEST] Rerun attempts      : "${RERUN_ATTEMPTS}
```
tolusha, line 444: Does that mean all tests will be rerun ?

selenium/che-selenium-core/bin/webdriver.sh
dmytro-ndp: It means we DO NOT rerun tests if `--rerun` parameter hasn't been set.

selenium/che-selenium-core/bin/webdriver.sh
dmytro-ndp: You might be right. From this point, `[JOB BUILD]` sounds better, doesn't it?

selenium/che-selenium-core/bin/webdriver.sh
dmytro-ndp: Not, it doesn't. I have just realised that the code will be simpler if we replace `rerun = false` by identical by meaning `rerun attempts = 0`.

selenium/che-selenium-core/bin/webdriver.sh
tolusha: ok. So, let's automatically set `RERUN_ATTEMPTS` to `1` if `rerun` is set.
If a developer needs more attempts then he has to set it explicitly.

selenium/che-selenium-core/bin/webdriver.sh
tolusha: ok

tolusha: Actually it is a job number.

selenium/che-selenium-core/bin/webdriver.sh
tolusha: ok

selenium/che-selenium-core/bin/webdriver.sh
dmytro-ndp: Job doesn't have a number. It has a name which is the name of the project to build https://wiki.jenkins.io/display/JENKINS/Building+a+software+project. And each build has unique number. 
For example, URL of page with test report https://ci.codenvycorp.com/view/qa/job/che-integration-tests-che6/100/testReport/ has:
- **che-integration-tests-che6**: name of project which job is aimed to build; 
- **100**: number of build of project. 

Therefore, we could name this number **100** as a `[PROJECT BUILD NUMBER]` and it can be confused with Che build namber, or as `[JOB BUILD NUMBER]` which is correct, but IMHO can be shorten to `[JOB BUILD]`.

tolusha: ok, `BUILD_NUMBER` makes sense then.

selenium/che-selenium-core/bin/webdriver.sh
dmytro-ndp: Good advice, thank you!

tolusha (approved):

PR #7876, opened 2017-12-14
Allow getting list of supported environment recipe types from Workspace API

### LINKED ISSUE #7530: Provide the data about supported environments
There is need for the clients to know the types of supported environments - to proper validate the workspace creation from recipe and the environment itself. Example - dockerfile, dockerimage, openshift, compose.

garagatyi:
### What does this PR do?
Adds a list of environment recipe types supported in a running Che instance. 
To get it cal `GET /api/workspace/settings`. The response is an object with next field:
`supported_environments` and value such as `"dockerimage,dockerfile"`.
Recipe types are separated by the comma sign.

Fixes #7530

### DIFF
--- a/wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
+++ b/wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
@@ -61,5 +61,7 @@ public final class Constants {
 
   public static final String WS_MACHINE_NAME = "default";
 
+  public static final String SUPPORTED_ENVIRONMENTS = "supported_environments";
+
   private Constants() {}
 }
--- a/wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
+++ b/wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
@@ -61,7 +61,7 @@ public final class Constants {
 
   public static final String WS_MACHINE_NAME = "default";
 
-  public static final String SUPPORTED_ENVIRONMENTS = "supportedEnvironments";
+  public static final String SUPPORTED_RECIPE_TYPES = "supportedRecipeTypes";
 
   private Constants() {}
 }
--- a/wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
+++ b/wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
@@ -61,7 +61,7 @@ public final class Constants {
 
   public static final String WS_MACHINE_NAME = "default";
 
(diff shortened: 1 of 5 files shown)


wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceService.java
```diff
@@ -675,7 +678,9 @@ public void deleteProject(
   @ApiOperation(value = "Get workspace server configuration values")
   @ApiResponses({@ApiResponse(code = 200, message = "The response contains server settings")})
   public Map<String, String> getSettings() {
-    return new HashMap<>();
+    return ImmutableMap.of(
+        Constants.SUPPORTED_ENVIRONMENTS,
```
sleshchenko: I'd change constant according to value which will be there -> `supportedRecipes`

wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
```diff
@@ -61,5 +61,7 @@
 
   public static final String WS_MACHINE_NAME = "default";
 
+  public static final String SUPPORTED_ENVIRONMENTS = "supported_environment_recipe_types";
```
sleshchenko: Mb it would be better to use a simpler value like `supportedRecipes` or `supportedEnvironments`?

garagatyi: Forget to apply the latest fix. Changed to `supported_environments`

garagatyi:
@sleshchenko @gazarenkov I'm not sure whether we should use camel case or underscores in an attribute name. Feel free to express your POV.

sleshchenko:
@garagatyi in attributes of user and profile we use camel case for attributes names. I think we should do in the same way whole all APIs. Is there any API which uses underscores in attributes names?

garagatyi:
@sleshchenko I see. I agree that we should do the same for this API. I'll rework the PR

wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
```diff
@@ -61,5 +61,7 @@
 
   public static final String WS_MACHINE_NAME = "default";
 
+  public static final String SUPPORTED_ENVIRONMENTS = "supported_environments";
```
gazarenkov: I'd prefer something like supportedRecipeTypes instead

gazarenkov (approved):
I'd prefer something like supportedRecipeTypes instead.
Otherwise I am OK

wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
garagatyi: ok

sleshchenko (approved):

PR #7884, opened 2017-12-14
Remove unnecessary item from the check list in the ServerRuntimeInfoTest

musienko-maxim:
### Description
After changing servers list in the IDE the 'terminal' item in runtime info widget is not visible on this moment. (We should resize or scroll this one to make it visible ). 
@vparfonov, @tolusha


musienko-maxim:
ci-build

tolusha (approved):

tolusha (approved):

PR #7893, opened 2017-12-15
Added project removal and enhanced checking of running OpenShift che projects in ocp.sh script

JamesDrummond:
### What does this PR do?
1. Added ability to replace che project started from ocp.sh
2. Added additional environment variables often set by user with ocp.sh
3. Modified deploy function to only generate config files if deploy_che.sh is not in current directory
4. Added additional checking using oc command if project really doesn't exist. After removing a project, there is a short period where `get` will not find project but a new project with the same name cannot be created.
5. Added a default value for CHE_INFRA_OPENSHIFT_MASTER__URL="${OPENSHIFT_ENDPOINT}/auth/"

NA

#### Release Notes
NA because ocp.sh should already be in release notes for che 6 release.

#### Docs PR
Docs should be added for che6 for this new ocp.sh script usage. An example of command line I run.

```
#Deploy che project and fails if project already exists
CHE_IMAGE_TAG=che6 OPENSHIFT_TOKEN=$(oc whoami -t) OPENSHIFT_ROUTING_SUFFIX="example.com" OC_PUBLIC_HOSTNAME=example.com OC_PUBLIC_IP=example DNS_PROVIDER=com ./ocp.sh --deploy-che --replace-stacks

#Deploy che project and delete if existing
CHE_IMAGE_TAG=che6 OPENSHIFT_TOKEN=$(oc whoami -t) OPENSHIFT_ROUTING_SUFFIX="example.com" OC_PUBLIC_HOSTNAME=example.com OC_PUBLIC_IP=example DNS_PROVIDER=com ./ocp.sh --deploy-che --remove-che --replace-stacks

#Remove che project without deploying
CHE_IMAGE_TAG=che6 OPENSHIFT_TOKEN=$(oc whoami -t) OPENSHIFT_ROUTING_SUFFIX="example.com" OC_PUBLIC_HOSTNAME=example.com OC_PUBLIC_IP=example DNS_PROVIDER=com ./ocp.sh --remove-che
```


dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
```diff
@@ -199,6 +199,7 @@ if [ -z "${OPENSHIFT_TOKEN+x}" ]; then
   OPENSHIFT_USERNAME=${OPENSHIFT_USERNAME:-${DEFAULT_OPENSHIFT_USERNAME}}
   OPENSHIFT_PASSWORD=${OPENSHIFT_PASSWORD:-${DEFAULT_OPENSHIFT_PASSWORD}}
 fi
+DEFAULT_CHE_INFRA_OPENSHIFT_MASTER__URL="${OPENSHIFT_ENDPOINT}/auth/"
```
sleshchenko: I don't think that `/auth` is really should be here. It can make an impression that it is auth endpoint, but actually, this URL is a parameter for creating fabric8 OpenShift client. Maybe it will work with `auth` in the path, but `"${OPENSHIFT_ENDPOINT}"` should be enough. 
Another note about what will happen if we don't specify this parameter -> fabric8 OpenShift client will use data from mounted service account token. I think it's kind of the same if we specify `"${OPENSHIFT_ENDPOINT}"` for master URL. So I'm OK to add default value for it to make it more obvious.

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
```diff
@@ -34,6 +34,16 @@ export JQ_BINARY_DOWNLOAD_URL=${JQ_BINARY_DOWNLOAD_URL:-${DEFAULT_JQ_BINARY_DOWN
 DEFAULT_CHE_MULTIUSER="false"
 export CHE_MULTIUSER=${CHE_MULTIUSER:-${DEFAULT_CHE_MULTIUSER}}
 
+DEFAULT_CHE_DEPLOY=false
+export CHE_DEPLOY=${CHE_DEPLOY:-${DEFAULT_CHE_DEPLOY}}
+CHE_DEPLOYED=false
+
+DEFAULT_CHE_REMOVE_PROJECT=false
+export CHE_REMOVE_PROJECT=${CHE_REMOVE_PROJECT:-${DEFAULT_CHE_MULTIUSER}}
```
sleshchenko: Are you sure that `DEFAULT_CHE_MULTIUSER` should be here as default value?

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
```diff
@@ -219,6 +303,27 @@ parse_args() {
...
+    fi
+
+    if [[ "$@" == *"--remove-che"* ]]; then
+      CHE_REMOVE_PROJECT=true
+    fi
+
+    if [[ "$@" == *"--replace-stacks"* ]]; then
+      CHE_REPLACE_STACKS=true
+    fi
+    for i in "${@}"
```
garagatyi: This doesn't look consistent with other places where arguments are parsed. I believe that 3 ways of parsing arguments is a thing that we should avoid.

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
```diff
@@ -171,12 +212,14 @@ server_is_booted() {
 
 wait_until_server_is_booted() {
   SERVER_BOOT_TIMEOUT=300
-  echo "[CHE] wait CHE pod booting..."
+  echo -n "[CHE] wait CHE pod booting..."
```
garagatyi, line 187: Can you elaborate on why you did change the output here and how it would look like after the change?

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
```diff
@@ -148,15 +158,46 @@ run_ocp() {
...
-    cd "${CONFIG_DIR}/instance/config/openshift/scripts/"
     bash deploy_che.sh ${DEPLOY_SCRIPT_ARGS}
     wait_until_server_is_booted
+    CHE_DEPLOYED=true
+    if $CHE_REPLACE_STACKS; then
+      replace_che_stacks_on_ocp
+    fi
+}
+
+replace_che_stacks_on_ocp() {
```
riuvshin: afaik we don't need to replace any stacks on OCP

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
```diff
@@ -148,15 +158,46 @@ run_ocp() {
 }
 
 deploy_che_to_ocp() {
-    #Repull init image only if IMAGE_PULL_POLICY is set to Always
-    if [ $IMAGE_PULL_POLICY == "Always" ]; then
-        docker pull "$IMAGE_INIT"
+    #Only generate scripts and config files if deploy_che.sh does not exist in same folder
+    if [ ! -f deploy_che.sh ]; then
+      #Repull init image only if IMAGE_PULL_POLICY is set to Always
+      if [ $IMAGE_PULL_POLICY == "Always" ]; then
```
garagatyi, line 161: This code gets too complicated. Can you refactor it to have less nesting?

garagatyi (changes requested):
This PR makes code too complex. I would recommend to refactor it with the usage of functions and less nesting. 
It is also not clear the reason why these changes are needed in Che, please elaborate. 
FYI we now use PR titles as changelog entries, so, please, make sure that title of this PR provides enough information for the changelog.

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
```diff
@@ -34,6 +34,16 @@ export JQ_BINARY_DOWNLOAD_URL=${JQ_BINARY_DOWNLOAD_URL:-${DEFAULT_JQ_BINARY_DOWN
 DEFAULT_CHE_MULTIUSER="false"
 export CHE_MULTIUSER=${CHE_MULTIUSER:-${DEFAULT_CHE_MULTIUSER}}
 
+DEFAULT_CHE_DEPLOY=false
+export CHE_DEPLOY=${CHE_DEPLOY:-${DEFAULT_CHE_DEPLOY}}
+CHE_DEPLOYED=false
```
riuvshin: what is the difference between
`CHE_DEPLOY` and `CHE_DEPLOYED`

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
```diff
@@ -148,15 +158,46 @@ run_ocp() {
 }
 
 deploy_che_to_ocp() {
-    #Repull init image only if IMAGE_PULL_POLICY is set to Always
-    if [ $IMAGE_PULL_POLICY == "Always" ]; then
-        docker pull "$IMAGE_INIT"
+    #Only generate scripts and config files if deploy_che.sh does not exist in same folder
+    if [ ! -f deploy_che.sh ]; then
```
riuvshin, line 159: the idea was to never use local scripts and always use generated because some files required to deploy may be generated only with no defaults. technically local script works for now but it should not be used in that way. especially taking in account fact that Mario works on integration this to CLI

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
```diff
@@ -148,15 +158,46 @@ run_ocp() {
...
+      replace_che_stacks_on_ocp
+    fi
+}
+
+replace_che_stacks_on_ocp() {
+  if ! $CHE_DEPLOY || $CHE_DEPLOYED; then
+    echo "[CHE] Checking if project \"${CHE_OPENSHIFT_PROJECT}\" exists before replacing stacks..."
+    if $OC_BINARY get project "${CHE_OPENSHIFT_PROJECT}" &> /dev/null; then
+      echo "[CHE] Replacing Che Stacks."
+      if [ ! -f replace_stacks.sh ]; then
```
riuvshin: again we should not use local scripts.

JamesDrummond: How would you propose to debug scripts without having the ability to make changes locally. Also, there is no way to change where the replace_stack.sh pulls it's stack definitions from when using the generated scripts. I could redo this script so that there is a command line argument `--local-scripts` that will use local script in lieu of generated. This means that unless user specifically wants to use local scripts it will default to generated scripts as it does currently.

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
JamesDrummond: CHE_DEPLOYED is used so that if replace stacks is run before deploy che it will not run. Example  is when CLI arguments are `--replace-stacks --deploy-che`. However it could run twice with  `--deploy-che --replace-stacks` so I will correct this by adding a variable `STACKS_REPLACED` to ensure the replace stack is only ran once.

CHE_DEPLOY is used so that replace stacks will now if CLI argument `--deploy-che` is being used. This allows replace stack to be run without deploying.

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
JamesDrummond: Not sure I understand the comment. I add a separate for loop to catch --help if the user does something like `--replace-stacks --help`. However I can put --help back into the other for loop as this case is rare. Let me know.

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
JamesDrummond: I allows the use of periods to indicate the script is working. It's similar to what has been done in other parts of the script.

```
[CHE] wait CHE pod booting...
[CHE] wait CHE pod booting....
[CHE] wait CHE pod booting......
```

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
JamesDrummond: Not sure how to do this. I could provide a separate function which would make things more DRY as I repeat this twice in `deploy_che_to_ocp` and `replace_che_stacks_on_ocp`.

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
riuvshin: build image -> define pull policy  that is how you debug
replace stacks it is obsolete script which does not work in multiuser and lost any sense in che6 and will be deleted soon.

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
garagatyi: Are you sure that output will be 
`[CHE] wait CHE pod booting......`
not 
```
[CHE] wait CHE pod booting...
.
.
.
```
?

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
garagatyi: Yes, make it simpler if possible.

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
garagatyi: Oh, I see that I missed what `-n` does. Sorry for confusing.

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
garagatyi: To parse arguments we use if clauses (lines 298-316), loop for checking help argument, loop for starting commends. This is too complex for such a small and simple thing as parsing arguments.
-1 to add that help loop. I'd rather refactor this to use a simpler approach such as parsing arguments in a loop, starting commands afterward.

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
JamesDrummond: Didn't know the defaults now work. Thought it was still a work in progress. Will I can remove all replace stacks from this pull request.

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
JamesDrummond: Well based on Roman's comment the replace stacks is no longer needed. If someone wants to replace this with custom stacks they can do this via assembly as described in the docs.

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
JamesDrummond: Will revise. Thanks.

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
JamesDrummond: If you make modifications to a script in "${CONFIG_DIR}/instance/config/openshift/scripts/" it will replace any scripts I modify within that folder such as deploy_che.sh . I think the ability to use your own scripts instead of the ones in the docker should be an option as users might need to patch the scripts to get things running on their infrastructure. I would be ok with having a variable that indicates to the docker run not to replace existing files if they exist. This would not be the default though but would allow something like the following.

export CONFIG_DIR=$(pwd)
export REPLACE_EXISTING_CONF=false #default true

```
docker run -t --rm -v /var/run/docker.sock:/var/run/docker.sock -v "${CONFIG_DIR}":/data -e IMAGE_INIT="$IMAGE_INIT" -e CHE_MULTIUSER="$CHE_MULTIUSER" -e REPLACE_EXISTING_CONF="$REPLACE_EXISTING_CONF" eclipse/che-cli:${CHE_IMAGE_TAG} destroy --quiet --skip:pull --skip:nightly
docker run -t --rm -v /var/run/docker.sock:/var/run/docker.sock -v "${CONFIG_DIR}":/data -e IMAGE_INIT="$IMAGE_INIT" -e CHE_MULTIUSER="$CHE_MULTIUSER" -e REPLACE_EXISTING_CONF="$REPLACE_EXISTING_CONF" eclipse/che-cli:${CHE_IMAGE_TAG} config --skip:pull --skip:nightly
```

JamesDrummond: Are you stating that the script files should be edited in the image and then commit the image to override the existing? This doesn't seem like the easiest solution as I would need to ssh into the image to get an editor that is not terminal to edit the file.

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
JamesDrummond: Will do.

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
JamesDrummond: Updated so that `export CHE_INFRA_OPENSHIFT_MASTER__URL=${CHE_INFRA_OPENSHIFT_MASTER__URL:-${OPENSHIFT_ENDPOINT}}` in the ocp.sh script. Removed from deploy_che.sh.

JamesDrummond:
Added ability in commit 3 to set or not set CHE_INFRA_OPENSHIFT_USERNAME/PASSWORD by using a new environment variable CHE_INFRA_ENABLED which defaults true when using ocp but false with osio and minishift. If these are set to any value and are not null/"", I cannot get workspaces to run which throws an error about "location" missing on server. I assume I need to setup a service account with developer through openshift to get this to work. I don't know if this is a bug or how it is intended to function. 

I would think that if CHE_INFRA_OPENSHIFT_MASTER__URL was not set it wouldn't attempt to use https://(hostname):8443/auth/ to make api calls to openshift cluster but it does if CHE_INFRA_OPENSHIFT_USERNAME/PASSWORD are set.

dockerfiles/init/modules/openshift/files/scripts/token
```diff
@@ -0,0 +1 @@
+eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJrdWJlcm5ldGVzL3NlcnZpY2VhY2NvdW50Iiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9uYW1lc3BhY2UiOiJlY2xpcHNlLWNoZSIsImt1YmVybmV0ZXMuaW8vc2VydmljZWFjY291bnQvc2VjcmV0Lm5hbWUiOiJkZWZhdWx0LXRva2VuLWYzNWRsIiwia3ViZXJuZXRlcy5pby9zZXJ2aWNlYWNjb3VudC9zZXJ2aWNlLWFjY291bnQubmFtZSI6ImRlZmF1bHQiLCJrdWJlcm5ldGVzLmlvL3NlcnZpY2VhY2NvdW50L3NlcnZpY2UtYWNjb3VudC51aWQiOiIxOTIyYmMyOS1lMzZhLTExZTctODk0MC0wMDA3Y2IwYjk2NzgiLCJzdWIiOiJzeXN0ZW06c2VydmljZWFjY291bnQ6ZWNsaXBzZS1jaGU6ZGVmYXVsdCJ9.qzcDTqQ1VVLl4kHiOqTpF9NDOWQoixrjzk7pBvM6ZoRufy_0s2LmKGqADflQcrVKyuZ1J72YLu6gkZv9naa2fvZFUgovYknumdjN3Uu3pyXB0_l7EFH14DxmiO_mpCDwjNRR8Dt3ww1n6u8NR8nY8v9szLil_-4uF7SC4cRRLF8O58pTHofOO86rYVb1rOkD0QrNl5TSV3d5VVY-Tz3Xyo9hy9kwfvswelcdWasfDzFZk1NrJJRW7k0pzMtIHOEVVhEaSmpbdvDsaiBVvDH6jiHc_QjkZ4mW0Mq_HRKepCcWa2Dkc1sbPoE3nYYdnRIZJG5iMkBXC_dzU3qlYCDS2w
```
sleshchenko: Should it be removed?

sleshchenko:
@JamesDrummond There is an ability not set CHE_INFRA_OPENSHIFT_USERNAME/PASSWORD by setting them explicitly to `""`. I'd avoid adding new parameter if it is possible.
Can you describe more about the issue that you faced? Which OpenShift instance did you use? How did you deploy Che Server there? And which Che Server configuration was there?

dockerfiles/init/modules/openshift/files/scripts/token
JamesDrummond: Removed.

sleshchenko:
@JamesDrummond I see. Thanks for the explanation. Then I can propose you to use the following construction:
```
if [ -z ${CHE_INFRA_OPENSHIFT_USERNAME+x} ]; then CHE_INFRA_OPENSHIFT_USERNAME=$OPENSHIFT_USERNAME; fi
```

skabashnyuk:
Please change target branch of this pr to master branch.

PR #7898, opened 2017-12-15
Support of headers to 'HttpJsonRequest'. Adding 'Connection: close' header while pinging workspace agent

ibuziuk:
### What does this PR do?
Adding support of headers to 'HttpJsonRequest'
Adding 'Connection: close' header while pinging workspace agent
Details in https://github.com/redhat-developer/rh-che/issues/479#issuecomment-351482507

Image used for verification on osio - `ibuziuk/che-server:connection-close`

https://github.com/redhat-developer/rh-che/issues/479

Adding support of headers to 'HttpJsonRequest'

#### Release Notes
Adding support of headers to 'HttpJsonRequest'

### DIFF
--- a/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequest.java
+++ b/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequest.java
@@ -68,6 +68,7 @@ public class DefaultHttpJsonRequest implements HttpJsonRequest {
   private String method;
   private Object body;
   private List<Pair<String, ?>> queryParams;
+  private List<Pair<String, String>> headers;
   private String authorizationHeaderValue;
 
   protected DefaultHttpJsonRequest(String url, String method) {
@@ -118,6 +119,16 @@ public class DefaultHttpJsonRequest implements HttpJsonRequest {
     return this;
   }
 
+  public HttpJsonRequest addHeader(@NotNull String name, @NotNull String value) {
+    requireNonNull(name, "Required non-null header name");
+    requireNonNull(value, "Required non-null header value");
+    if (headers == null) {
+      headers = new ArrayList<>();
+    }
+    headers.add(Pair.of(name, value));
+    return this;
+  }
+
   @Override
   public HttpJsonRequest setAuthorizationHeader(@NotNull String value) {
     requireNonNull(value, "Required non-null header value");
@@ -149,7 +160,7 @@ public class DefaultHttpJsonRequest implements HttpJsonRequest {
     if (method == null) {
       throw new IllegalStateException("Could not perform request, request method wasn't set");
     }
-    return doRequest(timeout, url, method, body, queryParams, authorizationHeaderValue);
+    return doRequest(timeout, url, method, body, queryParams, authorizationHeaderValue, headers);
   }
 
   /**
@@ -182,7 +193,8 @@ public class DefaultHttpJsonRequest implements HttpJsonRequest {
       String method,
       Object body,
       List<Pair<String, ?>> parameters,
-      String authorizationHeaderValue)
+      String authorizationHeaderValue,
+      List<Pair<String, String>> headers)
       throws IOException, ServerException, ForbiddenException, NotFoundException,
           UnauthorizedException, ConflictException, BadRequestException {
     final String authToken = EnvironmentContext.getCurrent().getSubject().getToken();
@@ -202,6 +214,15 @@ public class DefaultHttpJsonRequest implements HttpJsonRequest {
     final HttpURLConnection conn = (HttpURLConnection) new URL(url).openConnection();
     conn.setConnectTimeout(timeout > 0 ? timeout : 60000);
     conn.setReadTimeout(timeout > 0 ? timeout : 60000);
+
+    final boolean hasHeaders = headers != null && !headers.isEmpty();
+
+    if (hasHeaders) {
+      for (Pair<String, String> header : headers) {
+        conn.setRequestProperty(header.first, header.second);
+      }
+    }
+
     try {
       conn.setRequestMethod(method);
       // drop a hint for server side that we want to receive application/json
@@ -225,7 +246,6 @@ public class DefaultHttpJsonRequest implements HttpJsonRequest {
           output.write(DtoFactory.getInstance().toJson(body).getBytes());
         }
       }
-
       final int responseCode = conn.getResponseCode();
       if ((responseCode / 100) != 2) {
         InputStream in = conn.getErrorStream();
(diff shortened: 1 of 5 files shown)


ibuziuk:
ci-test

davidfestal (approved):

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequest.java
```diff
@@ -118,6 +121,16 @@ public HttpJsonRequest addQueryParam(@NotNull String name, @NotNull Object value
     return this;
   }
 
+  public HttpJsonRequest addRequestHeader(@NotNull String name, @NotNull String value) {
```
garagatyi: Consider renaming to `addHeader`. We can not add response header to a request.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequest.java
```diff
@@ -118,6 +121,16 @@ public HttpJsonRequest addQueryParam(@NotNull String name, @NotNull Object value
...
+    requireNonNull(name, "Required non-null header name");
+    requireNonNull(value, "Required non-null header value");
+    if (headers == null) {
+      headers = new ArrayList<>(DEFAULT_HEADERS_LIST_SIZE);
```
garagatyi: Can you elaborate on why you have decided to have this 5 as a default value instead of letting default constructor do his job?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequest.java
```diff
@@ -202,6 +216,15 @@ protected DefaultHttpJsonResponse doRequest(
     final HttpURLConnection conn = (HttpURLConnection) new URL(url).openConnection();
     conn.setConnectTimeout(timeout > 0 ? timeout : 60000);
     conn.setReadTimeout(timeout > 0 ? timeout : 60000);
+
+    final boolean hasHeaders = headers != null && !headers.isEmpty();
```
garagatyi, line 218: Can you inline this into if clause?

garagatyi (approved):
Looks good. Can you take a look at my comments?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequest.java
ibuziuk: ok, lemme do it

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequest.java
ibuziuk: I have decided not to break existing design of the class (see impl. of query params), I think we can safely remove this param for both lists. WDYT?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequest.java
ibuziuk: I have decided not to break existing design of the class (see processing of query params above). I can rework both if needed ?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequest.java
ibuziuk: as I understand it was mainly done in order to reduce the initial capacity and avoid recreating array in most cases. Maybe also typical number of query params used in che = 5 ? Anyway, probably the performance impact seems to be insufficient and we might just opt for using default capacity for both query params / headers?

garagatyi: Yeah, it was definitely an optimization. And author might add a comment to that value to clarify that. 
I opt for leaving it with default settings.

garagatyi: I mean header related one. Let's leave query params optimisation as is

ibuziuk: sounds good - PR has been updated

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequest.java
garagatyi: you are free to use the style of code you want as long as it doesn't make the code inefficient or too complex (probably empirical for maintainers values ;) ). 
So my comment about this line is just matter of my taste )
Feel free to do as you want here.
And since I approved the PR in the initial state I was OK to merge it at that stage, my comments were just insignificant details.

benoitf:
ci-build

l0rd (approved):

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequest.java
skabashnyuk: can we initialise ```headers = new ArrayList<>()``` in constructor to not to have ``` if (headers == null) {``` all other the code.

ibuziuk: @skabashnyuk I'm quite sure that the initial designer of the class who seems to be really keen on performance and even used custom capacity[1] for query list initialization would like this idea. Basically, the rule of thumb for such cases is either do a massive refactoring or follow the design that is already in place e.g [2] (even if you think it is not smth.  you would normally do if implementing from scratch)  


[1] https://github.com/eclipse/che/pull/7898/files#diff-570b40ab3f5e69b575585ffb160ef3b5L115
[2] https://github.com/eclipse/che/pull/7898/files#diff-570b40ab3f5e69b575585ffb160ef3b5R201

skabashnyuk (approved):

skabashnyuk:
ci-test

skabashnyuk:
@ibuziuk can you make che6 pr aswell?

ibuziuk:
@skabashnyuk sure I will ;-)

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequest.java
garagatyi: +1 to what @ibuziuk said

ibuziuk:
PR for che6 https://github.com/eclipse/che/pull/7949

Issue #6736, opened 2017-10-13
Port implementors to jdt.ls extension


tsmaeder:
Move the code that finds implementers into jdt.ls extension and call from "implementers" through custom command.

gazarenkov:
@tsmaeder labels?
https://github.com/eclipse/che/wiki/Labels

tolusha:
@bdshadow is working on this issue.

PR #7904, opened 2017-12-15
che-6736: Port implementors to jdt.ls extension

bdshadow:
Signed-off-by: Dmitrii Bocharov <dbocharo@redhat.com>

### What does this PR do?
ports finding implementors to the language server together with https://github.com/eclipse/che-ls-jdt/pull/11

https://github.com/eclipse/che/issues/6736


bdshadow:
@tsmaeder please, take a look.
I haven't removed some things here:
1) https://github.com/eclipse/che/pull/7904/files#diff-e53b584767405bf8374b3e48a7f30335R129. `SymbolKind` doesn't have an `Annotation` value
2) i also left the processing of the binary types commented here: https://github.com/eclipse/che/pull/7904/files#diff-40b085755395803301441347022db8caR136. How should `symbolInformation` be detected as  a binary (external library) and be dealt with?

dkulieshov (approved):

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/navigation/openimplementation/NoImplementationWidget.java
```diff
@@ -59,17 +60,16 @@ public String getEmptyMessage() {
...
 
   @Override
-  public Element createItem(final Type itemModel) {
+  public Element createItem(final SymbolInformation itemModel) {
     final Element element = Elements.createLiElement(popupResources.popupStyle().item());
     final Element iconElement = Elements.createDivElement(popupResources.popupStyle().icon());
 
-    int flag = itemModel.getFlags();
-    if (flag == -1) {
+    if (itemModel.getKind() == null) {
```
tsmaeder: Is this a case that happens? If not don't check.

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/navigation/openimplementation/OpenImplementationPresenter.java
```diff
@@ -96,161 +91,149 @@ public void show(final EditorPartPresenter editorPartPresenter) {
...
-      if (!srcFolder.isPresent()) {
+      if (project == null || !srcFolder.isPresent()) {
         return;
       }
 
       final String fqn = JavaUtil.resolveFQN((Container) srcFolder.get(), (Resource) file);
 
       service
-          .getImplementations(project.get().getLocation(), fqn, activeEditor.getCursorOffset())
+          .findImplementations(
```
tsmaeder: I think it would make more sense to pass the current location in the editor as a TextDocumentPositionParameter and to resolve the position to the innermost enclosing interface in the jdt.ls extension. That way, we can also find implementations of interfaces other than the main type. For example, we could select "Runnable" in "Runnable r= ...." and find implementers of that interface. Also, it would be more in line with general LSP conventions and we could easily extend it to find implementations of methods in the future.

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/navigation/openimplementation/OpenImplementationPresenter.java
```diff
@@ -96,161 +91,149 @@ public void show(final EditorPartPresenter editorPartPresenter) {
...
       NoImplementationWidget noImplementationWidget, TextEditor editorPartPresenter) {
     int offset = editorPartPresenter.getCursorOffset();
     PositionConverter.PixelCoordinates coordinates =
         editorPartPresenter.getPositionConverter().offsetToPixel(offset);
-    Type type = dtoFactory.createDto(Type.class);
-    type.setFlags(-1);
-    noImplementationWidget.addItem(type);
+    SymbolInformation symbolInformation = new SymbolInformation();
+    symbolInformation.setKind(null);
+    noImplementationWidget.addItem(symbolInformation);
```
tsmaeder: PopupWidget handles the case of no elements. Why do we complicate our lives here?

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/navigation/openimplementation/OpenImplementationPresenter.java
```diff
@@ -96,161 +91,149 @@ public void show(final EditorPartPresenter editorPartPresenter) {
...
+                    new NoImplementationWidget(
+                        popupResources,
+                        javaResources,
+                        locale,
+                        OpenImplementationPresenter.this,
+                        title);
+                if (overridingSize == 1) {
+                  actionPerformed(impls.getImplementations().get(0));
+                } else if (overridingSize > 1) {
+                  openOneImplementation(
```
tsmaeder: If we have more than one implementation we call a method that is call "openONEImplementation? Somethings seems misnamed here.

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/navigation/openimplementation/OpenImplementationPresenter.java
```diff
@@ -96,161 +91,149 @@ public void show(final EditorPartPresenter editorPartPresenter) {
...
-      implementationWidget.addItem(type);
+    for (SymbolInformation symbolInformation : implementationsDescriptor.getImplementations()) {
+      implementationWidget.addItem(symbolInformation);
     }
     implementationWidget.show(coordinates.getX(), coordinates.getY());
     implementationWidget.asElement().getStyle().setWidth(600 + "px");
   }
 
-  private void setCursor(final Region region) {
+  private void setCursor(final Range region) {
```
tsmaeder: Methinks we can delete this if you use OpenFileInEditorHelper

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/navigation/openimplementation/OpenImplementationPresenter.java
```diff
@@ -96,161 +91,149 @@ public void show(final EditorPartPresenter editorPartPresenter) {
...
-              public void execute() {
-                TextEditor editor = (TextEditor) editorAgent.getActiveEditor();
-                editor.setFocus();
-                editor
-                    .getDocument()
-                    .setSelectedRange(
-                        LinearRange.createWithStart(region.getOffset()).andLength(0), true);
-              }
+            () -> {
+              ((TextEditor) editorAgent.getActiveEditor())
```
tsmaeder: This is bogus: the active editor may have changed in the meantime and the cast will crash.

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/navigation/service/JavaNavigationServiceImpl.java
```diff
@@ -237,23 +247,39 @@ public void getContent(
...
-            + fqn
-            + "&offset="
-            + offset;
-
-    return requestFactory
-        .createGetRequest(url)
-        .header(ACCEPT, APPLICATION_JSON)
-        .loader(loaderFactory.newLoader())
-        .send(unmarshallerFactory.newUnmarshaller(ImplementationsDescriptorDTO.class));
+  public Promise<ImplementationsDescriptor> findImplementations(
```
tsmaeder: So far, we've added jdt.ls extension services to JavaLanguageServiceExtensionClient. Unless there is a good reason, I would suggest to move this method ther.

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/navigation/service/JavaNavigationServiceImpl.java
```diff
@@ -237,23 +247,39 @@ public void getContent(
...
+      FindImplementationsCommandParameters params) {
+    return Promises.create(
+        (resolve, reject) -> {
+          requestTransmitter
+              .newRequest()
+              .endpointId(WS_AGENT_JSON_RPC_ENDPOINT_ID)
+              .methodName("java/navigation")
```
tsmaeder: "java/navigation" is not a good message name. "java/implementers" or the likes would be better.

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/navigation/service/JavaNavigationServiceImpl.java
```diff
@@ -237,23 +247,39 @@ public void getContent(
...
+              .paramsAsDto(params)
+              .sendAndReceiveResultAsDto(ImplementationsDescriptor.class, 10000)
+              .onSuccess(resolve::apply)
+              .onTimeout(
+                  () -> {
+                    reject.apply(
+                        new PromiseError() {
```
tsmaeder: JSPromisError.create() makes this simpler.

plugins/plugin-java/che-plugin-java-ext-lang-client/src/test/java/org/eclipse/che/ide/ext/java/client/navigation/openimplementation/OpenImplementationPresenterTest.java
```diff
@@ -123,8 +123,9 @@ public void testShouldDisplayOneImplementationIsRealFile() throws Exception {
     when(relatedProject.getLocation()).thenReturn(Path.valueOf("/a"));
     when(editor.getCursorOffset()).thenReturn(123);
     when(implementationsPromise.then(any(Operation.class))).thenReturn(implementationsPromise);
-    when(javaNavigationService.getImplementations(eq(Path.valueOf("/a")), eq("c.d.file"), eq(123)))
-        .thenReturn(implementationsPromise);
+    //    when(javaNavigationService.getImplementations(eq(Path.valueOf("/a")), eq("c.d.file"),
```
tsmaeder: Why comment this out? Either we need this or not. We have a repository for history.

plugins/plugin-java/che-plugin-java-ext-lang-client/src/test/java/org/eclipse/che/ide/ext/java/client/navigation/openimplementation/OpenImplementationPresenterTest.java
```diff
@@ -162,8 +163,9 @@ public void testShouldDisplayNoImplementations() throws Exception {
...
```
tsmaeder: As above.

plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/JavaLanguageServerExtensionService.java
```diff
@@ -169,6 +175,12 @@ public void configureMethods() {
         .paramsAsString()
         .resultAsListOfDto(ClasspathEntryDto.class)
         .withFunction(this::getClasspathTree);
+    requestHandler
+        .newConfiguration()
+        .methodName("java/navigation")
```
tsmaeder: As already mentioned: message name.

plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/JavaLanguageServerExtensionService.java
```diff
@@ -346,6 +358,26 @@ public String getOutputDir(String projectUri) {
...
+    CompletableFuture<Object> result =
+        executeCommand(FIND_IMPLEMENTORS_COMMAND, singletonList(params));
+
+    Type targetClassType = new TypeToken<ImplementationsDescriptor>() {}.getType();
+    try {
+      ImplementationsDescriptor implementationsDescriptor =
+          gson.fromJson(gson.toJson(result.get(10, TimeUnit.SECONDS)), targetClassType);
+      for (SymbolInformation symbolInformation : implementationsDescriptor.getImplementations()) {
+        Location location = symbolInformation.getLocation();
+        location.setUri(LanguageServiceUtils.removePrefixUri(location.getUri()));
```
tsmaeder: This will not work when the implementations are in jars. Have a look at LanguageServiceUtils#fixLocation():

tsmaeder (changes requested):

plugins/plugin-java/che-plugin-java-ext-lang-client/src/test/java/org/eclipse/che/ide/ext/java/client/navigation/openimplementation/OpenImplementationPresenterTest.java
bdshadow: Just commented the failing tests by now to be able to build it, check and push for the review. Definitely, i'll fix it

plugins/plugin-java/che-plugin-java-ext-lang-client/src/test/java/org/eclipse/che/ide/ext/java/client/navigation/openimplementation/OpenImplementationPresenterTest.java
bdshadow: as above :)

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/navigation/openimplementation/NoImplementationWidget.java
```diff
@@ -117,16 +117,18 @@ private native CustomEvent createValidateEvent(String eventType) /*-{
...
-    } else {
-      icon = javaResources.javaFile();
+    switch (symbolKind) {
+      case Interface:
+        icon = javaResources.interfaceItem();
+        break;
+      case Enum:
+        icon = javaResources.enumItem();
+        break;
+        // case Annotation: icon = javaResources.annotationItem();  break;
```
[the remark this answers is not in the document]
tsmaeder: I don't think you can extend annotations in Java. At least Eclipse complains about it. Right now, on the Che side, annotations are not handled. I don't know how what jdt.ls returns for annotations. Use the the source, luke? Workspace symbols request returns SymbolInfo.

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/navigation/openimplementation/OpenImplementationPresenter.java
tsmaeder: You can easily have two types called "Foo" in a Java project. Which one are you going to use?

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/navigation/openimplementation/OpenImplementationPresenter.java
tsmaeder: PopupWidget seems to show a settable text when no elements are present (sth. like "No implementations"), so no need to invent an artificial element.

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/navigation/openimplementation/OpenImplementationPresenter.java
bdshadow: 2 types can't have the same fully qualified name in one project

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/navigation/openimplementation/NoImplementationWidget.java
```diff
@@ -11,45 +11,29 @@
 package org.eclipse.che.ide.ext.java.client.navigation.openimplementation;
```
tsmaeder: I still don't understand why we need a special widget to show "no implementations."

tsmaeder (changes requested):

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/navigation/openimplementation/NoImplementationWidget.java
bdshadow: @tsmaeder what do you propose to use instead? sorry if you already said this somewhere and i missed it

tsmaeder: If a PopupWidget has no elements, it shows the string returned from "getEmtpyMessage()". I don't understand why that is not what we want.

tsmaeder: Yes, but we have a "ImplementationWidget" and a "NoImplementationWidget". Why?

tsmaeder (approved):

PR #7948, opened 2017-12-19
Allow to use JSInterop on export

benoitf:
### What does this PR do?
Add the arguments for GWT compiler to enable JSInterop export.

Change-Id: I37664554064cbe87ec68a8e062e78f34b45d745f

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1877,7 +1877,11 @@
                     <codeserverArgs>
                         <arg>-noprecompile</arg>
                         <arg>-noincremental</arg>
+                        <arg>-generateJsInteropExports</arg>
                     </codeserverArgs>
+                    <compilerArgs>
+                        <arg>-generateJsInteropExports</arg>
+                    </compilerArgs>
                     <jvmArgs>
                         <arg>${gwt.compiler.jvmArgs.Xss}</arg>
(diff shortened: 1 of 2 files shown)


vparfonov (approved):

evidolob (approved):

benoitf:
ci-build

pom.xml
```diff
@@ -1877,7 +1877,11 @@
                     <codeserverArgs>
                         <arg>-noprecompile</arg>
                         <arg>-noincremental</arg>
+                        <arg>-generateJsInteropExports</arg>
                     </codeserverArgs>
+                    <compilerArgs>
+                        <arg>-generateJsInteropExports</arg>
```
azatsarynnyy, line 1883: Florent is it really needed to pass `-generateJsInteropExports` option to GWT compiler here to work with Super DevMode?
Or it's enough to pass this option to CodeServer only? Since GWT compilation is performed in `che-ide-gwt-app`

azatsarynnyy (approved):

pom.xml
benoitf: @azatsarynnyy when I was using super dev mode without that flag all jsinterop export code was missing so I would say it's needed

azatsarynnyy: hmm... Strange behavior. But ok if it works )

Issue #7715, opened 2017-12-04
Improve UX to delete a command


slemeur:
Following: https://github.com/openshiftio/openshift.io/issues/1502 

We should be improving the UX to delete a command.
- The icon does not look intuitive
- We are missing a right click action on commands (to duplicate, rename, delete, move)
- We are also missing a delete button in the command editor. 
- Also please consider handling the case when "delete" on the keyboard is pressed.

slemeur:
Also please consider handling the case when "delete" on the keyboard is pressed.

vzhukovs:
![2018-1-24 15-39](https://user-images.githubusercontent.com/1968177/35378851-e6ced298-01bc-11e8-9534-5cb514972eff.png)

I don't know if it is a good idea to duplicate buttons, wdyt?

ashumilova:
@slemeur do we need both context menu and delete&copy buttons?

slemeur:
Can the contextual menu have actions for:
- rename: it will open the command if not opened, and gives the focus into the name's textfield.
- move: it will also open the command if not opened, and gives the focus on the goal dropdown.

Can we add the icons in the contextual menu?

Then:
- I think the colors for the quick actions should not be in blue - it seems overwhelming.
- The spacing between the icons also seems to be too important.

vzhukovs:
During implementing this issue, two moments where discovered:

- [ ] Don't set up focus forcibly after editor initialization https://github.com/eclipse/che/issues/8573
- [ ] Make an ability to handle focus and key press event on CustomListBox https://github.com/eclipse/che/issues/8574

Created a PR: https://github.com/eclipse/che/pull/8581

PR #7951, opened 2017-12-19
Use trash icon from FontAwesome for delete command operation

vzhukovs:
### What does this PR do?
This changes proposal replaces current icon for delete command button with the new one used from Font Awesome:

image

https://github.com/openshiftio/openshift.io/issues/1502

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/command/explorer/CommandsTreeRenderer.java
+++ b/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/command/explorer/CommandsTreeRenderer.java
@@ -151,7 +151,7 @@ class CommandsTreeRenderer extends DefaultPresentationRenderer<Node> {
       SpanElement element = Document.get().createSpanElement();
       element.getStyle().setFontSize(11., Style.Unit.PT);
       element.getStyle().setMarginTop(2., Style.Unit.PT);
-      element.setInnerHTML((String)icon);
+      element.setInnerHTML((String) icon);
       return element;
     }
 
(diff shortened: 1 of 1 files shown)


vzhukovskii:
ci-build

vzhukovs:
ci-build

dkulieshov:
Hi,

Everything looks good, but I wonder what you think about the next thing. I'd rather added a separate method `createButton(String)` instead of making common method `getIconElement`, because though technically it is permissible, it seems to me a bit weird when you do `icon instanceof String`, because essentially icon and string are quite different entities. Or, by the way, we can have `createButton(Element)` and `createButtonElement(SVGResource resource)` and `createButtonElement(String iHtmlTag)`?

vzhukovs:
@dkuleshov it's not a problem to refactor this PR. The main reason why I'm comparing with String is that we have `org.eclipse.che.ide.FontAwesome` class where icons described as html strings. There isn't any impediments to create two separate private methods to process two types of icon.

azatsarynnyy (approved):

ashumilova (approved):

vparfonov (approved):

slemeur:
@vzhukovskii : Have you handle "deleting" using the keyboard ?

slemeur:
This does not fix: https://github.com/eclipse/che/issues/7715

vzhukovs:
@slemeur no, this PR just changes the icon with the new one.

slemeur:
so where are the other improvementS?

ashumilova:
The PR handled the description of the following issue only - https://github.com/openshiftio/openshift.io/issues/1502
this one will take into account - https://github.com/eclipse/che/issues/7715

PR #7976, opened 2017-12-20
Allow deploying multi-tenant rh-che on Minishift in single-tenant mode

davidfestal:
### What does this PR do?

Allow deploying multi-tenant rh-che on Minishift in single-tenant mode, thus usng the default mechanism of creating workspace PODs in the Che server OS namespace.

This is part of the fix for issue
https://github.com/redhat-developer/rh-che/issues/489


benoitf:
@davidfestal it seems that now due to merge of che6 into master you need to change target branch

l0rd (approved):

benoitf (approved):

riuvshin (approved):

Issue #7971, opened 2017-12-20
clean up CI after merge CHE6 to master


riuvshin:
as soon as we will merge che6 to master we will need to clean up some jobs on CI, remove `che6` on docker hub images. and get rid of che6 specific qa-ci jobs

riuvshin:
Done

PR #7979, opened 2017-12-20
Clean up after merge che6 to master

riuvshin:
Clean up after merge che6 to master
part of: https://github.com/eclipse/che/issues/7971

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/resources/che-in-che.json
+++ b/assembly/assembly-wsmaster-war/src/main/resources/che-in-che.json
@@ -24,7 +24,7 @@
     ],
     "source": {
       "type": "image",
-      "origin": "eclipse/che-dev:che6"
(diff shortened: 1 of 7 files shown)


benoitf (approved):
do we need to provide "nightlies" for 5.x branch ?

riuvshin:
@benoitf che 5.x and codenvy will be supported only in maintenance mode. we will not have whole CI for that.
I going to remove everything 5.x related from CI soon

benoitf:
well I would say that it's the case only once master is GA

riuvshin:
not sure about that, we don't expect any changes in 5.x branches except blocker bugfixes
it means we will never have 5.23.0, 5.24 and so on only 5.22.1-n

skabashnyuk (approved):

garagatyi (approved):

Issue #7154, opened 2017-11-02
IDE and Dashboard don't show any error message when docker workspace fails to start because of InternalInfrastructureException


sleshchenko:
### Description
IDE and Dashboard don't show any error message when docker workspace fails to start because of InternalInfrastructureException.

See how it looks like on IDE 
![ide](https://user-images.githubusercontent.com/5887312/32328244-087dc09c-bfe2-11e7-8a38-9629568e8974.gif)

See how it looks like on Dashboard
![dashboard](https://user-images.githubusercontent.com/5887312/32328245-08ab46b6-bfe2-11e7-8fa5-d51f39f8a3fe.gif)

<details>
<summary>Workspace master logs</summary>

```
2017-11-02 13:23:21,203[nio-8080-exec-1]  [INFO ] [o.e.c.a.w.s.WorkspaceRuntimes 238]   - Starting workspace 'che/wksp-122t' with id 'workspacejbvytuks783dql14' by user 'che'
2017-11-02 13:23:21,209[aceSharedPool-1]  [INFO ] [o.e.c.a.w.s.WorkspaceRuntimes 291]   - Workspace 'che:wksp-122t' with id 'workspacejbvytuks783dql14' start failed
2017-11-02 13:23:21,209[aceSharedPool-1]  [ERROR] [o.e.c.a.w.s.WorkspaceRuntimes 301]   - Error response from docker API, status: 500, message: could not find an available, non-overlapping IPv4 address pool among the defaults to assign to the network

org.eclipse.che.api.workspace.server.spi.InternalInfrastructureException: Error response from docker API, status: 500, message: could not find an available, non-overlapping IPv4 address pool among the defaults to assign to the network

        at org.eclipse.che.workspace.infrastructure.docker.network.NetworkLifecycle.createNetwork(NetworkLifecycle.java:46)
        at org.eclipse.che.workspace.infrastructure.docker.DockerInternalRuntime.internalStart(DockerInternalRuntime.java:188)
        at org.eclipse.che.api.workspace.server.spi.InternalRuntime.start(InternalRuntime.java:103)
        at org.eclipse.che.api.workspace.server.WorkspaceRuntimes$StartRuntimeTask.run(WorkspaceRuntimes.java:275)
        at org.eclipse.che.commons.lang.concurrent.CopyThreadLocalRunnable.run(CopyThreadLocalRunnable.java:29)
[...]

PR #7988, opened 2017-12-21
Store last workspace startup or crash error status and message in workspace attributes.

mshaposhnik:
### What does this PR do?
To allow to recognize on client fast happening workspace startup errors, we store them in the WS attributes and cleanup after successful startup.

https://github.com/eclipse/che/issues/7154

Store last workspace startup error status and message in workspace attributes.

#### Release Notes
N/A

#### Docs PR
N/A


wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -58,9 +58,15 @@
   private static final Logger LOG = LoggerFactory.getLogger(WorkspaceManager.class);
 
   /** This attribute describes time when workspace was created. */
-  public static final String CREATED_ATTRIBUTE_NAME = "created";
-  /** This attribute describes time when workspace was last updated or started/stopped/recovered. */
-  public static final String UPDATED_ATTRIBUTE_NAME = "updated";
+  protected static final String CREATED_ATTRIBUTE_NAME = "created";
```
sleshchenko: I think such constants should be placed in `org.eclipse.che.api.workspace.shared.Constants` class. Then [here](https://github.com/eclipse/che/pull/7988/files#diff-1340504bd34c91a6a4c1666d91511f01R192) you will be able to use constant instead of hardcoded string

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -58,9 +58,15 @@
...
+  /** This attribute describes time when workspace was last updated or started */
```
sleshchenko: Why is it still used for storing the time when a workspace is started?

sleshchenko:
Looks like there are not your commits in this PR. Please rebase to `master` branch to fix it

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
mshaposhnik: why it should not ?

sleshchenko: Because it is `UPDATED` attribute but not `STARTED`? Earlier `UPDATED` was used for all actions with a workspace. Now it is used only for registering of update and start, but not for a stop. IMHO It's a quite unclear behavior.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -58,9 +58,15 @@
...
+  protected static final String UPDATED_ATTRIBUTE_NAME = "updated";
+  /** This attribute describes time when workspace was last stopped */
+  protected static final String STOPPED_ATTRIBUTE_NAME = "stopped";
+  /** This attribute indicates is last stop was abnormal */
+  protected static final String STOPPED_ABNORMALLY_ATTRIBUTE_NAME = "stoppedAbnormally";
```
sleshchenko: Is it used anywhere?
As for me abnormal stop is the case when workspace was running and abnormally stopped (like crashed).
Also when a workspace is running and becomes stopped is failing to start but not abnormal stopping.
WDYT?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -58,9 +58,15 @@
...
+  /** This attribute describes latest startup error message */
+  protected static final String ERROR_MESSAGE_ATTRIBUTE_NAME = "errorMessage";
```
sleshchenko: It would be better to rename this attribute to make it clear that it is startup error message.

wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
```diff
@@ -39,6 +39,9 @@
 
   public static final String CHE_WORKSPACE_AUTO_START = "che.workspace.auto_start";
 
+  /** This attribute describes latest startup error message */
+  public static final String ERROR_MESSAGE_ATTRIBUTE_NAME = "errorMessage";
```
sleshchenko: It would be better to rename this attribute to make it clear that it is startup error message.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -58,9 +59,13 @@
   private static final Logger LOG = LoggerFactory.getLogger(WorkspaceManager.class);
 
   /** This attribute describes time when workspace was created. */
-  public static final String CREATED_ATTRIBUTE_NAME = "created";
-  /** This attribute describes time when workspace was last updated or started/stopped/recovered. */
-  public static final String UPDATED_ATTRIBUTE_NAME = "updated";
+  protected static final String CREATED_ATTRIBUTE_NAME = "created";
+  /** This attribute describes time when workspace was last updated or started */
+  protected static final String UPDATED_ATTRIBUTE_NAME = "updated";
```
sleshchenko: I see you moved only `ERROR_MESSAGE_ATTRIBUTE_NAME` to Constants class. I'd say that all these constants should be moved there to make clients able to use them.

wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
mshaposhnik: It probably can be used not only for startup errors

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
mshaposhnik: doesn't get the idea., sorry. We probably can use it when WS is crashed, too.  (which is basically the same)

wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
sleshchenko: Then please fix java doc for it

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
sleshchenko: We can but we don't. Let's do it or provide the corresponding comment. Now it's not clear enough.

benoitf:
IMHO the logic is not so obvious. It looks like a workaround.

Why can't we subscribe to events before the start of a workspace ? Is it only just because we don't have the value of the channel before we start the workspace ? AFAIK we should have these values before starting it.
Besides, the JSON RPC bus is global to the workspace master so there should be no issue

mshaposhnik:
@benoitf We discussed this here locally before start coding . Yes, you didn't know the link to the status channel before you started the workspace. (yes, you can estimate it but it is not official way :) ). And there is also theoretically possible some boundary usecases when client who starting and client who using are different. (e.g start with dashb and switch to standalone IDE etc). So having reserve place for storing latest error seems to be useful.

benoitf:
@mshaposhnik I see no reason why the links are not available before the start as well.

sleshchenko (approved):

wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
```diff
@@ -39,6 +39,17 @@
 
   public static final String CHE_WORKSPACE_AUTO_START = "che.workspace.auto_start";
 
+  /** This attribute describes time when workspace was created. */
```
garagatyi: It is not clear these are workspace attributes. Can you add `{@link Workspace#getAttributes}` link to the javadocs?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -576,20 +580,36 @@ private String sessionUserNameOr(String nameIfNoUser) {
...
   }
 
-  private class CleanupRuntimeOnAbnormalRuntimeStop implements EventSubscriber<RuntimeStatusEvent> {
+  private class AbnormalRuntimeStopListener implements EventSubscriber<RuntimeStatusEvent> {
     @Override
     public void onEvent(RuntimeStatusEvent event) {
       if (event.isFailed()) {
         RuntimeState state = runtimes.remove(event.getIdentity().getWorkspaceId());
         if (state != null) {
+          String workspaceId = state.runtime.getContext().getIdentity().getWorkspaceId();
```
garagatyi: 2 lines up workspace ID is fetched too. Can you move this assignment higher and reuse the result?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -405,6 +409,38 @@ private WorkspaceImpl normalizeState(WorkspaceImpl workspace, boolean includeRun
...
+      updateWorkspace(workspace.getId(), workspace);
+    } catch (NotFoundException | ServerException | ValidationException | ConflictException e) {
+      LOG.warn(
+          String.format(
+              "Cannot set error status of the workspace %s. Error is: %s",
+              workspace.getId(), e.getMessage()));
+    }
+  }
+
+  private void handleStartupSuccess(Workspace workspace) {
```
garagatyi, line 432: Is this covered by unit tests?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -576,20 +580,36 @@ private String sessionUserNameOr(String nameIfNoUser) {
...
+              workspaceId,
               STOPPED,
               RUNNING,
               "Error occurs on workspace runtime stop. Error: " + event.getError());
+          setAbnormalStopAttributes(workspaceId, event.getError());
         }
       }
     }
+
+    private void setAbnormalStopAttributes(String workspaceId, String error) {
```
garagatyi, line 600: Is this change covered by unit tests?

garagatyi (commented):
Looks good. Please, take a look at my comments about some improvements you may add to the PR.

vparfonov (approved):

skabashnyuk (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
mshaposhnik: Hell..i have no idea how to test this inner class yet..

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
mshaposhnik: Added extra test

skabashnyuk:
ci-test

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
garagatyi: 1. Call WorkspaceRunitmes#init to subscribe abnormal stop listener to the EventService
1. Call EventService#publish to fire RuntimeStatusEvent about failed runtime
1. Verify event workspace event publishing, an update of workspace attributes.

Please, do not hesitate to comment back if I missed something

garagatyi (approved):

PR #8002, opened 2017-12-21
Make it possible to update Che on OCP, including rolling update

ghost:
1. Use oc apply instead of oc create to make it possible to re-run deployment script, for example, to update to a new version.
2. Use Rolling update strategy, which does not interrupt Che server


riuvshin (approved):
:rainbow_flag:

benoitf (approved):

PR #8014, opened 2017-12-22
OCP deployment: more ram for keycloak and housepeeking

ghost:
### What does this PR do?

1. If swap is turned off on an OCP instance, 1GB isn't enough for Keycloak to start. It is killed by OOM. So, increasing the limit up to 1.5GB fixes it.

2. Remove unused env. It's heritage from 5.x OpenShiftConnector.

3. Use the right protocol in Che route.


dockerfiles/init/modules/openshift/files/scripts/multi-user/keycloak/deployment-config.yaml
```diff
@@ -60,9 +60,9 @@ spec:
         - containerPort: 8080
         resources:
           limits:
-            memory: 1024Mi
+            memory: 1430Mi
           requests:
```
benoitf: For me 1.5 x 1024 = 1536

benoitf (approved):

dockerfiles/init/modules/openshift/files/scripts/multi-user/keycloak/deployment-config.yaml
ghost: I used a lame converter then :) Fixed

benoitf (approved):

Issue #8022, opened 2017-12-22
Fix Oauth login link on Dashboard on multiuser mode


mshaposhnik:
Since we merged OAuth logins management via keycloak, there  is two things that must be changed on dashb:

1) Remove check if OAuth enabled via our wsmaster API (it is always disabled in M/U mode now);
2) GItHub Login button must point to URL Like:
`http://172.19.20.29:8080/api/oauth/authenticate?oauth_provider=github&token=<keycloak_token>&redirect_after_login=<where_to_redirect>`

PR #8027, opened 2017-12-22
Fix oauth login to GitHub with keycloak

ashumilova:

### What does this PR do?
This PR fixes the detection - when to display message about not configured GitHub provider on project importing. If it's multi-user che and keycloak is configured - no need to show that message. Also the login to Gihub in multi-user Che was fixed.

https://github.com/eclipse/che/issues/8022

#### Release Notes
N/A

#### Docs PR
N/A


riuvshin:
does this will work for both che single user (w/o KC) and che multi user (with KC) ?

benoitf (approved):

olexii4 (approved):

ashumilova:
@riuvshin yes, the goal is to support both single and multi user.

PR #8030, opened 2017-12-22
Add FluentWait for avoiding unexpected fails for animated web element

musienko-maxim:
### What does this PR do?
This PR fix potential problems with the Pull Request widget during animation (Not Clickable Element exception). That will make the test more stable when animation will be 'slow'.
@riuvshin, @vparfonov, @dmytro-ndp


riuvshin (approved):

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/PullRequestPanel.java
```diff
@@ -231,8 +237,11 @@ public void waitRepoUrl(String url) {
   }
 
   public void openPullRequestOnGitHub() {
-    loadPageDriverWait
-        .until(elementToBeClickable(By.xpath(PullRequestLocators.OPEN_GITHUB_BTN)))
-        .click();
+    Wait<WebDriver> wait =
```
dmytro-ndp: I suggest you to use existed method [waitElementIsRendered(WebElement webElement)](https://github.com/eclipse/che/blob/master/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/TestWebElementRenderChecker.java#L54), like it was done [here](https://github.com/eclipse/che/blob/master/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java#L602).

dmytro-ndp (approved):

Issue #8043, opened 2017-12-26
Add the 'Assert.fail()' to the 'CommitFilesTest' related to known issue


artaleks9:
### Description
* Need to add the 'Assert.fail()' to the selenium test 'CommitFilesTest' related to known issue

**Related issue:** #8042

Issue #8042, opened 2017-12-26
The rename of package does not work properly


artaleks9:
### Description
* After refactoring a name of package in the project tree the name of package is not displayed properly

**Reproduction Steps:**
* Open a project in the IDE (e.g. `web-java-spring` from template)
* Select the package and perform refactoring name it (e.g. `org.eclipse.che.examples` to `org.eclipse.dev.examples`)

**Expected behavior:**
* should be displayed a new name of package in correct view  in the project tree  

**Observed behavior:**
* the structure of package is broken
* part of previous name is present in the project tree
* sometimes in new name of package absents the last symbol

**Che version:** 6.0.0-M4
**OS and version:** Fedora 25
**Docker version:** 17.06
**API docker version:** 1.30
**Che install:** Docker container

**Additional information:**
* Related selenium test: _CommitFilesTest_
* Problem can be reliably reproduced, doesn't happen randomly: [Yes]
* See the attachment:

_**screenshot from CI report:**_
![rename-pack](https://user-images.githubusercontent.com/5408906/34352444-30fd1860-ea2b-11e7-94d0-be1e0ef36097.png)

**_video:_**
![rename-package](https://user-images.githubusercontent.com/5408906/34354557-c74e0688-ea37-11e7-9ba1-dafe4d2801b9.gif)

PR #8047, opened 2017-12-26
Selenium: Add the 'Assert.fail() to the 'CommitFilesTest' related to known issue

artaleks9:
### What does this PR do?
* Add the 'Assert.fail() to the selenium test 'CommitFilesTest' related to known issue
* Add the deletion of commit form in the test 'CommitFilesTest'
* Add the 'isWidgetOpen' method in the page objects 'GitCommit' and 'Git'
* Correct the name of method 'clickOnCancelBtnComitForm' in the page object 'Git'

#8043
#8042

### DIFF
--- a/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/git/Git.java
+++ b/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/git/Git.java
@@ -363,7 +363,7 @@ public class Git {
   }
 
   /**
-   * check open state of commit form
+   * check opened state of commit form
    *
    * @return true if widget is open
    */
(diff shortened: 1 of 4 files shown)


selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/git/Git.java
```diff
@@ -362,8 +362,17 @@ public void waitCommitMainFormIsOpened() {
     gitCommit.waitMainFormCommit();
   }
 
+  /**
+   * check open state of commit form
+   *
+   * @return true if widget is open
```
dmytro-ndp: open > opened

dmytro-ndp (approved):

tolusha (approved):

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/git/Git.java
artaleks9: Thanks, it is corrected.

PR #8073, opened 2017-12-27
Fix Che using different database path in different versions

### LINKED ISSUE #8068: Che cannot access old database after it's upgrade
### Description
 
After upgrading existing Che instance to 6.0.0-M3 version, it doesn't work with existing database but new one is being created and used instead.

Old database location: /<path_to_che_instance/data/db/che.mv.db
New database location: /<path_to_che_instance/data/storage/db/che.mv.db

### Reproduction Steps
1) Run any Che prior to 6.0.0-M3 version.
2) Stop it, and upgrade that Che instance to 6.0.0-M3
3) Run upgraded Che

**OS and version:**    
Fedora 27

mkuznyetsov:
### What does this PR do?
Export Che environment variables only if there were not defined
Move database files, that were stored in different paths until and after Che 6.0.0-M3 version:
- Database will be stored at path "/data/storage/db/che.mv.db"

- If there is a database at old path present ("data/db/che.mv.db"), a warning will be shown in Che container logs, and, if no database is present by the new path, it will be moved there

should fix https://github.com/eclipse/che/issues/8068

Fix Che using different database path in different versions.
entypoint.sh will not overwrite Che environment variables, if they were already defined

#### Docs PR

N/A


dockerfiles/che/entrypoint.sh
```diff
@@ -47,10 +47,10 @@ Variables:
 
   # Must be exported as this will be needed by Tomcat's JVM
   DEFAULT_CHE_REGISTRY_HOST=localhost
-  export CHE_REGISTRY_HOST=${CHE_REGISTRY_HOST:-${DEFAULT_CHE_REGISTRY_HOST}}
```
riuvshin: in that case we don;t need any change because it already pick value fom ENV VAR

dockerfiles/che/entrypoint.sh
```diff
@@ -47,10 +47,10 @@ Variables:
...
+  [ -z "$CHE_REGISTRY_HOST" ] && export CHE_REGISTRY_HOST=${CHE_REGISTRY_HOST:-${DEFAULT_CHE_REGISTRY_HOST}}
 
   DEFAULT_CHE_PORT=8080
-  export CHE_PORT=${CHE_PORT:-${DEFAULT_CHE_PORT}}
```
riuvshin: in that case we don;t need any change because it already pick value fom ENV VAR

dockerfiles/che/entrypoint.sh
```diff
@@ -59,7 +59,7 @@ Variables:
   CHE_LOG_LEVEL=${CHE_LOG_LEVEL:-${DEFAULT_CHE_LOG_LEVEL}}
 
   DEFAULT_CHE_LOGS_DIR="${CATALINA_HOME}/logs/"
-  export CHE_LOGS_DIR=${CHE_LOGS_DIR:-${DEFAULT_CHE_LOGS_DIR}}
```
riuvshin: in that case we don;t need any change because it already pick value fom ENV VAR

riuvshin (changes requested):
please read comments, we need to handle ONLY vars that set value, vars like
```
export VAR=${VAR:-${DEFAULT_VAR}}
```
are OK as they already pick vars from ENV

dockerfiles/che/entrypoint.sh
```diff
@@ -205,48 +205,48 @@ init() {
...
   fi
 
   ### We need to discover the host mount provided by the user for `/data`
-  export CHE_DATA="/data"
+  [ -z "$CHE_DATA" ] && export CHE_DATA="/data"
   CHE_DATA_HOST=$(get_che_data_from_host)
 
   CHE_USER=${CHE_USER:-root}
-  export CHE_USER=$CHE_USER
+  [ -z "$CHE_USER" ] && export CHE_USER=$CHE_USER
```
riuvshin: why you changed that?

dockerfiles/che/entrypoint.sh
```diff
@@ -205,48 +205,48 @@ init() {
...
   if [ "$CHE_USER" != "root" ]; then
     if [ ! $(getent group docker) ]; then
       echo "!!!"
       echo "!!! Error: The docker group doesn't exist."
       echo "!!!"
       exit 1
     fi
-    export CHE_USER_ID=${CHE_USER}
+    [ -z "$CHE_USER_ID" ] && export CHE_USER_ID=${CHE_USER}
```
riuvshin: why you changed that?

dockerfiles/che/entrypoint.sh
```diff
@@ -205,48 +205,48 @@ init() {
...
+  [ -z "$CHE_WORKSPACE_TERMINAL__LINUX__AMD64" ] && export CHE_WORKSPACE_TERMINAL__LINUX__AMD64=${CHE_DATA_HOST}/lib/linux_amd64/terminal
+  [ -z "$CHE_WORKSPACE_TERMINAL__LINUX__ARM7" ] && export CHE_WORKSPACE_TERMINAL__LINUX__ARM7=${CHE_DATA_HOST}/lib/linux_arm7/terminal
+  [ -z "$CHE_WORKSPACE_EXEC__LINUX__AMD64" ] && export CHE_WORKSPACE_EXEC__LINUX__AMD64=${CHE_DATA_HOST}/lib/linux_amd64/exec
 
   # CHE_DOCKER_IP_EXTERNAL must be set if you are in a VM.
   HOSTNAME=${CHE_DOCKER_IP_EXTERNAL:-$(get_docker_external_hostname)}
   if has_external_hostname; then
     # Internal property used by Che to set hostname.
-    export CHE_DOCKER_IP_EXTERNAL=${HOSTNAME}
+    [ -z "$CHE_DOCKER_IP_EXTERNAL" ] && export CHE_DOCKER_IP_EXTERNAL=${HOSTNAME}
```
riuvshin: why you changed that?

dockerfiles/che/entrypoint.sh
```diff
@@ -199,6 +199,24 @@ launch_docker_registry () {
...
+  CHE_DATA=/data
+  # Move database to original location
+
+  if [ -f ${CHE_DATA}/storage/db/che.mv.db ]; then
+    echo "Performing migration of the database file, to it's original old path"
+    echo "In case if there was already a database file in old path, it will be renamed to 'che.mv.db.old'"
+    echo "See issue https://github.com/eclipse/che/issues/8068 for details"
+    #in case if there is existing database in old path, back it up
+    if [ -f ${CHE_DATA}/db/che.mv.db ]; then
+      mv ${CHE_DATA}/db/che.mv.db ${CHE_DATA}/db/che.mv.db.old
```
sleshchenko: It can be useful to add logs that `old` database is backed up

mkuznyetsov: ok

riuvshin (approved):

skabashnyuk (approved):

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
```diff
@@ -11,7 +11,7 @@
 
 ### CHE SERVER
 # Folder where Che will store internal data objects
-che.database=${che.home}/storage
+che.database=${che.home}
```
benoitf: why it's not anymore in a subfolder ?

mkuznyetsov: It was like so in Che instances, that were running through CLI for a long while because of this che.env.erb:
https://github.com/eclipse/che/blob/56866159228bf4ba1d3b7fc8d2206c60b129579b/dockerfiles/init/modules/che/templates/che.env.erb#L36
So, the values from the che.properties or entrypoint.sh, which would point to path with "storage" directory had no effect, until recently (which was the issue cause)

I decided to consolidate everything by old path, though if you think extra "storage" directory is better, it can be changed to that

benoitf: I would say that it looks odd to have a "root" directory as path for che database so yes using a subfolder would be better IMHO


https://github.com/eclipse/che/blob/93b974d1e0ec7836635f3d7da196b7651da8ad1b/dockerfiles/che/entrypoint.sh#L233

skabashnyuk: @benoitf Am I understand you correctly. You want to change db file location.
Add extra path "storage" to what we have in Che5?

benoitf: yes. From what I see there is a 'db' subfolder added
It's not directly stored inside the che.home folder

mkuznyetsov:
@benoitf I changed the PR, and updated the description, describing how it works now
- database will be stored by path "/data/storage/db/che.mv.db"
- if upgrading Che from versions, that were using different path, it will leave a warning in the container logs, and would place it in the proper path, unless there is already DB there.

benoitf (approved):

dockerfiles/che/entrypoint.sh
```diff
@@ -84,7 +84,7 @@ set_environment_variables () {
   # CHE_DOCKER_IP is used internally by Che to set its IP address
   if [[ -z "${CHE_DOCKER_IP}" ]]; then
     if [[ -n "${CHE_IP}" ]]; then
-        export CHE_DOCKER_IP="${CHE_IP}"
+      export CHE_DOCKER_IP="${CHE_IP}"
```
benoitf: maybe you can skip this change from this PR's commit as it's unrelated (it's more a indent fix)

PR #8088, opened 2017-12-28
Add continuous servers liveness checks

### LINKED ISSUE #7550: Add continuous checking of servers statuses for wsagent, exec, terminal
### Description
We have servers checking for wsagent, exec, terminal. Now it is used on servers starts only whereas the server can fail at any point in time, e.g. OOM. This leads to a bad UX of workspaces usage.
We want to make these checks continuous to identify when a problem occurs and WS should be stopped due to the error.

### Reproduction Steps

**OS and version:**    

**Diagnostics:**

garagatyi:
### What does this PR do?
Adds continuous servers liveness checks for the same servers that are checked at a start of a workspace - terminal, exec, wsagent.
To add this adds:
- liveness endpoints to the mentioned agents which are not protected by authentication
- a framework that should be called to schedule servers checks or to cancel those checks
- schedule servers checks after server become available on workspace startup server check
- call servers check cancellation in various places to guarantee that servers checks don't get left when workspace is stopped, start failed or unexpectedly failed.

Fixes #7550

### DIFF
--- a/agents/go-agents/exec-agent/main.go
+++ b/agents/go-agents/exec-agent/main.go
@@ -17,6 +17,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"regexp"
 	"time"
 
 	"strconv"
@@ -83,6 +84,17 @@ func main() {
 				},
 			},
 		},
+		{
+			Name: "Exec-Agent liveness route",
+			Items: []rest.Route{
+				{
+					Method:     "GET",
+					Path:       "/liveness",
+					Name:       "Check Exec-Agent liveness",
+					HandleFunc: restutil.OKRespondingFunc,
+				},
+			},
+		},
 	}
 
 	appOpRoutes := []jsonrpc.RoutesGroup{
@@ -95,9 +107,8 @@ func main() {
 	jsonrpc.RegRoutesGroups(appOpRoutes)
 	jsonrpc.PrintRoutes(appOpRoutes)
 
-	var handler = wrapWithAuth(r)
-	http.Handle("/", handler)
-	http.Handle("/liveness", restutil.LivenessCheckingHandler())
+	// do not protect liveness check endpoint
+	var handler = wrapWithAuth(r, "/liveness")
 
 	server := &http.Server{
 		Handler:      handler,
@@ -108,14 +119,15 @@ func main() {
 	log.Fatal(server.ListenAndServe())
 }
 
-func wrapWithAuth(h http.Handler) http.Handler {
-	// required authentication for all the requests, if it is configured
-	if config.authEnabled {
-		cache := auth.NewCache(time.Minute*time.Duration(config.tokensExpirationTimeoutInMinutes), time.Minute*5)
-		return auth.NewCachingHandler(h, config.apiEndpoint, droppingRPCChannelsUnauthorizedHandler, cache)
+func wrapWithAuth(h http.Handler, mapping string) http.Handler {
+	// required authentication for all the requests that match mappings, if auth is configured
+	if !config.authEnabled {
+		return h
 	}
 
-	return h
+	pattern := regexp.MustCompile(mapping)
+	cache := auth.NewCache(time.Minute*time.Duration(config.tokensExpirationTimeoutInMinutes), time.Minute*5)
+	return auth.NewCachingHandler(h, config.apiEndpoint, droppingRPCChannelsUnauthorizedHandler, cache, pattern)
 }
 
 func droppingRPCChannelsUnauthorizedHandler(w http.ResponseWriter, req *http.Request, err error) {
--- a/agents/go-agents/exec-agent/main.go
+++ b/agents/go-agents/exec-agent/main.go
@@ -18,9 +18,8 @@ import (
 	"net/url"
 	"os"
 	"regexp"
-	"time"
-
 	"strconv"
+	"time"
 
 	"github.com/eclipse/che/agents/go-agents/core/auth"
 	"github.com/eclipse/che/agents/go-agents/core/jsonrpc"
--- a/agents/go-agents/exec-agent/main.go
+++ b/agents/go-agents/exec-agent/main.go
@@ -118,15 +118,15 @@ func main() {
 	log.Fatal(server.ListenAndServe())
 }
 
-func wrapWithAuth(h http.Handler, mapping string) http.Handler {
+func wrapWithAuth(h http.Handler, ignoreMapping string) http.Handler {
 	// required authentication for all the requests that match mappings, if auth is configured
 	if !config.authEnabled {
 		return h
 	}
 
-	pattern := regexp.MustCompile(mapping)
+	ignorePattern := regexp.MustCompile(ignoreMapping)
 	cache := auth.NewCache(time.Minute*time.Duration(config.tokensExpirationTimeoutInMinutes), time.Minute*5)
-	return auth.NewCachingHandler(h, config.apiEndpoint, droppingRPCChannelsUnauthorizedHandler, cache, pattern)
+	return auth.NewCachingHandler(h, config.apiEndpoint, droppingRPCChannelsUnauthorizedHandler, cache, ignorePattern)
 }
 
 func droppingRPCChannelsUnauthorizedHandler(w http.ResponseWriter, req *http.Request, err error) {
--- a/agents/go-agents/exec-agent/main.go
+++ b/agents/go-agents/exec-agent/main.go
@@ -19,13 +19,15 @@ import (
 	"os"
 	"time"
 
+	"strconv"
+
 	"github.com/eclipse/che/agents/go-agents/core/auth"
 	"github.com/eclipse/che/agents/go-agents/core/jsonrpc"
 	"github.com/eclipse/che/agents/go-agents/core/jsonrpc/jsonrpcws"
 	"github.com/eclipse/che/agents/go-agents/core/process"
 	"github.com/eclipse/che/agents/go-agents/core/rest"
+	"github.com/eclipse/che/agents/go-agents/core/rest/restutil"
 	"github.com/eclipse/che/agents/go-agents/exec-agent/exec"
-	"strconv"
 )
 
 var (
@@ -93,8 +95,9 @@ func main() {
 	jsonrpc.RegRoutesGroups(appOpRoutes)
 	jsonrpc.PrintRoutes(appOpRoutes)
 
-	var handler = getHandler(r)
+	var handler = wrapWithAuth(r)
 	http.Handle("/", handler)
+	http.Handle("/liveness", restutil.LivenessCheckingHandler())
 
 	server := &http.Server{
 		Handler:      handler,
@@ -105,7 +108,7 @@ func main() {
 	log.Fatal(server.ListenAndServe())
 }
 
-func getHandler(h http.Handler) http.Handler {
+func wrapWithAuth(h http.Handler) http.Handler {
 	// required authentication for all the requests, if it is configured
 	if config.authEnabled {
 		cache := auth.NewCache(time.Minute*time.Duration(config.tokensExpirationTimeoutInMinutes), time.Minute*5)
--- a/agents/go-agents/terminal-agent/main.go
+++ b/agents/go-agents/terminal-agent/main.go
@@ -91,15 +91,15 @@ func main() {
 	log.Fatal(server.ListenAndServe())
 }
 
-func wrapWithAuth(h http.Handler, mapping string) http.Handler {
+func wrapWithAuth(h http.Handler, ignoreMapping string) http.Handler {
 	// required authentication for all the requests that match mappings, if auth is configured
 	if !config.authEnabled {
 		return h
 	}
 
-	pattern := regexp.MustCompile(mapping)
+	ignorePattern := regexp.MustCompile(ignoreMapping)
 	cache := auth.NewCache(time.Minute*time.Duration(config.tokensExpirationTimeoutInMinutes), time.Minute*5)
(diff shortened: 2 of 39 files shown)


garagatyi:
ci-test

garagatyi:
ci-build

garagatyi:
ci-test

garagatyi:
ci-test

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerInternalRuntime.java
```diff
@@ -322,6 +354,37 @@ public void accept(String serverRef) {
...
+      String serverName = probeResult.getServerName();
+      ProbeStatus probeStatus = probeResult.getStatus();
+      Server server = machine.getServers().get(serverName);
+      ServerStatus oldServerStatus = server.getStatus();
+      ServerStatus serverStatus;
+
+      if (probeStatus == ProbeStatus.FAILED && oldServerStatus == ServerStatus.RUNNING) {
+        serverStatus = ServerStatus.STOPPED;
+      } else if (probeStatus == ProbeStatus.PASSED
+          && (oldServerStatus == null
```
akorneta: Consider simplifying this check to  ` oldServerStatus != ServerStatus.RUNNING`.

garagatyi: tnx

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/probe/ProbeScheduler.java
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    tasks.forEach(task -> task.cancel(true));
+  }
+
+  private void schedule(
+      String workspaceId, ProbeFactory probeFactory, Consumer<ProbeResult> probeResultConsumer) {
+    ProbeConfig probeConfig = probeFactory.getProbeConfig();
+    Task task = new Task(probeFactory, probeResultConsumer);
+    // scheduleWithFixedDelay is used in favor of scheduleAtFixedRate because in case of bug amount
```
akorneta: Looks like '_bug amount_' is typo, am I right?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/probe/Probe.java
```diff
@@ -0,0 +1,52 @@
...
+      // clear interrupted state
+      Thread.interrupted();
+    }
+  }
+
+  /**
+   * Returns {@code true} if probe finishes successfully, {@code false} otherwise. Must return false
+   * when probe is interrupted even if interruption is not respected by probe implementation.
+   */
+  public abstract boolean doProbe();
```
sleshchenko: Should it be `protected`?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/probe/ProbeConfig.java
```diff
@@ -0,0 +1,82 @@
...
+  private int failureThreshold = 3;
+  private int timeoutSeconds = 1;
+  private int periodSeconds = 10;
+  private int initialDelaySeconds;
+
+  /**
+   * Creates probe configuration.
+   *
+   * @param successThreshold minimum consecutive successes for the probe to be considered successful
+   *     after having failed. Minimum value is 1. Defaults to 1.
```
sleshchenko: Doc says about default values but for me, it's not clear how to create `ProbeConfig` with default values

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/probe/ProbeScheduler.java
```diff
@@ -0,0 +1,203 @@
...
+        if (successes >= probeConfig.getSuccessThreshold()) {
+          // TODO make them completable futures? Then we should ensure that
+          // consecutive calls won't use executors thread time but rather use common thread
+          // pool to perform processing of passed probe result
+          if (cancelled.get()) {
+            return;
+          }
+          // Health check satisfies probeConfig health conditions
+          probeResultConsumer.accept(
+              new ProbeResult(
```
sleshchenko, line 160: What makes me confused is that `ProbeConfig` and `ProbeResult` are not config and result of `Probe` but result and config of scheduling.
Maybe we can encapsulate such logic in the `Probe#probe` method?
Then it will be able to return `ProbeResult` and you will not have to use `ProbeFactory` to construct `ProbeResult`.
If we can't do it in this way please review these objects names.

agents/go-agents/exec-agent/main.go
```diff
@@ -17,15 +17,18 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"regexp"
 	"time"
 
+	"strconv"
```
sleshchenko: Should this import really be in a separate section? Maybe we can append it to previous?

agents/go-agents/exec-agent/main.go
```diff
@@ -105,14 +119,15 @@ func main() {
 	log.Fatal(server.ListenAndServe())
 }
 
-func getHandler(h http.Handler) http.Handler {
-	// required authentication for all the requests, if it is configured
-	if config.authEnabled {
-		cache := auth.NewCache(time.Minute*time.Duration(config.tokensExpirationTimeoutInMinutes), time.Minute*5)
-		return auth.NewCachingHandler(h, config.apiEndpoint, droppingRPCChannelsUnauthorizedHandler, cache)
+func wrapWithAuth(h http.Handler, mapping string) http.Handler {
```
sleshchenko: I expect that `mapping` value will say which should be wrapped, but actually, it is `noAuthPattern`. Please consider to renaming this variable.

agents/go-agents/exec-agent/main.go
```diff
@@ -105,14 +119,15 @@ func main() {
...
+	// required authentication for all the requests that match mappings, if auth is configured
+	if !config.authEnabled {
```
sleshchenko: Maybe it would be better to place here checking of `noAuthPattern` instead of putting it into `NewCachingHandler`

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/probe/ProbeScheduler.java
garagatyi: tnx

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/probe/Probe.java
garagatyi: +1

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/probe/ProbeConfig.java
garagatyi: Was done with config deserialization in mind. Will remove for the time being.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/probe/ProbeScheduler.java
garagatyi: Probe itself is a one-time object since it needs cancelation in some cases. Because of that ProbeConfig needs to be outside of the Probe. 
As to the ProbeResult it is generated not on every probe result but only if it satisfies success/failure thresholds of a probe configuration. Because of that, it cannot be generated by the probe itself, since probe doesn't know how many times it succeeded/failed.

agents/go-agents/exec-agent/main.go
garagatyi: +1

agents/go-agents/exec-agent/main.go
garagatyi: Go default library doesn't have negative lookahead syntax in regexp. Because of that, I don't know how to do that without complicating the code and significantly rewriting it. Feel free to do a suggestion if you have any idea.

agents/go-agents/exec-agent/main.go
garagatyi: I didn't get what you mean. Can you elaborate on that?

agents/go-agents/exec-agent/main.go
sleshchenko: `ignoreMapping`?

agents/go-agents/exec-agent/main.go
sleshchenko: I mean rewriting it in the following way:
```go

func wrapWithAuth(h http.Handler, ignoreMapping string) http.Handler {
	// required authentication for all the requests that match mappings, if auth is configured
	noAuthPattern := regexp.MustCompile(ignoreMapping)
	if !config.authEnabled || handler.noAuthPattern.MatchString(req.URL.Path){
		return h
	}

	cache := auth.NewCache(time.Minute*time.Duration(config.tokensExpirationTimeoutInMinutes), time.Minute*5)
	return auth.NewCachingHandler(h, config.apiEndpoint, droppingRPCChannelsUnauthorizedHandler, cache)
}
```
It's just proposal. Maybe it's not better. Why I think about it because your comments says
```
	// required authentication for all the requests that match mappings, if auth is configured
```
but actually, it doesn't check mapping at all.
The second thing is if CachingHandler checks mapping, why it doesn't check config?

agents/go-agents/exec-agent/main.go
sleshchenko: Is this `+1` to the first or second comment? =)

agents/go-agents/exec-agent/main.go
garagatyi: +1

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/probe/ProbeScheduler.java
```diff
@@ -0,0 +1,203 @@
...
+    // seem a good thing
+    ScheduledFuture scheduledFuture =
+        probesExecutor.scheduleWithFixedDelay(
+            task,
+            probeConfig.getInitialDelaySeconds(),
+            probeConfig.getPeriodSeconds(),
+            TimeUnit.SECONDS);
+
+    List<ScheduledFuture> workspaceProbes =
+        probesFutures.computeIfPresent(
```
akorneta, line 105: since you are using `HashMap` you have no guarantee that `computeIfPresent` would be performed atomically, so potentially it can lead to non-deterministic behaviour.

garagatyi: +1 thanks

agents/go-agents/exec-agent/main.go
garagatyi: fixed

agents/go-agents/exec-agent/main.go
garagatyi: We can check mapping on a per-request basis only. Because of that, we have to check it as part of the auth mechanism. To avoid that we would need a way how to map auth filtering to certain paths, but we have an only single mapping. 
To make it prettier it is needed to rewrite routing mechanism to add auth there. I believe that it is not the right time to do it now. 
checking that auth is enabled is needed here because in the case it is not we don't wrap routing with auth at all.

sleshchenko:
Is it still `in-progress`?

agents/go-agents/exec-agent/main.go
sleshchenko: OK then

akorneta (approved):

agents/go-agents/terminal-agent/main.go
```diff
@@ -26,13 +26,16 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"regexp"
 	"time"
 
+	"strconv"
```
sleshchenko: the same as here https://github.com/eclipse/che/pull/8088#discussion_r159417582

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/probe/ProbeScheduler.java
sleshchenko: I understand. 
Then please add java doc to `Probe` that it should not be used directly but `ProbeScheduler` should be used instead. Am I  wrong?

sleshchenko (approved):

agents/go-agents/terminal-agent/main.go
garagatyi: good catch!

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/probe/ProbeScheduler.java
garagatyi: ok

garagatyi:
ci-build

skabashnyuk (approved):

garagatyi:
ci-test

PR #8110, opened 2018-01-01
Fix internal policy of event.Bus events publishing

voievodin:
Hi there :hand:

This pull request fixed the behaviour of events publishing in Golang based [event bus](https://github.com/eclipse/che/blob/83d25535a92c71820e4283eb70cc4b9c98f913e8/agents/go-agents/core/event/event.go). 
The current version of `Bus.Pub` func:
```go
func (bus *Bus) Pub(e E) {
	bus.RLock()
	defer bus.RUnlock()
	cons, ok := bus.consumers[e.Type()]
	if ok {
		for idx, v := range cons {
			v.Accept(e)
			if tmpCons, ok := v.(TmpConsumer); ok && tmpCons.IsDone() {
				bus.rm(e.Type(), idx) // <= oops
			}
		}
	}
}
```
The main problem here is modification of bus state within the read lock in case consumer is instance of `TmpConsumer` and call to `isDone` returns true. One of the commits introduces a test-case that breaks the behaviour of this function and ends up throwing one of the following errors:
```
panic: runtime error: slice bounds out of range
fatal error: concurrent map writes
```

To solve the problem copy-on-write approach was used. So basically each time there is a new subscription or consumer removal the original array of consumers is replaced with a new one. It allows concurrently publishing go-routines to complete without any errors. This approach is similar to the one used by [EventService](https://github.com/eclipse/che/blob/83d25535a92c71820e4283eb70cc4b9c98f913e8/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/notification/EventService.java).

### DIFF
--- a/agents/go-agents/core/event/event.go
+++ b/agents/go-agents/core/event/event.go
@@ -64,13 +64,14 @@ func NewBus() *Bus {
 // Pub publishes an event to the interested consumers.
 func (bus *Bus) Pub(e E) {
 	bus.RLock()
-	defer bus.RUnlock()
-	cons, ok := bus.consumers[e.Type()]
+	consumers, ok := bus.consumers[e.Type()]
+	bus.RUnlock()
+
 	if ok {
-		for idx, v := range cons {
+		for _, v := range consumers {
 			v.Accept(e)
 			if tmpCons, ok := v.(TmpConsumer); ok && tmpCons.IsDone() {
-				bus.rm(e.Type(), idx)
+				bus.Rm(tmpCons)
 			}
 		}
 	}
@@ -81,7 +82,7 @@ func (bus *Bus) Pub(e E) {
 func (bus *Bus) Sub(consumer Consumer, eType string) {
 	bus.Lock()
 	defer bus.Unlock()
-	bus.consumers[eType] = append(bus.consumers[eType], consumer)
+	bus.consumers[eType] = append(bus.copyConsumers(eType), consumer)
 }
 
 // SubAny does the same as Sub func, but for multiple event types.
@@ -89,11 +90,12 @@ func (bus *Bus) SubAny(consumer Consumer, types ...string) {
 	bus.Lock()
 	defer bus.Unlock()
 	for _, t := range types {
-		bus.consumers[t] = append(bus.consumers[t], consumer)
+		bus.consumers[t] = append(bus.copyConsumers(t), consumer)
 	}
 }
 
-// Rm removes given consumer from the bus, == operator is used to determine
+// Rm removes given consumer from the bus, == operator is used
+// to determine what elements should be removed.
 // Returns true if consumer is removed.
 func (bus *Bus) Rm(consumer Consumer) bool {
 	return bus.RmIf(func(c Consumer) bool { return c == consumer })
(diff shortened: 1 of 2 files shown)


skabashnyuk:
ci-test

garagatyi:
Hi @voievodin! Nice to see you again. Can you sign your commits to satisfy eclipse ip check?

akorneta (approved):

agents/go-agents/core/event/event.go
```diff
@@ -81,19 +82,20 @@ func (bus *Bus) Pub(e E) {
...
 }
 
 // SubAny does the same as Sub func, but for multiple event types.
 func (bus *Bus) SubAny(consumer Consumer, types ...string) {
 	bus.Lock()
 	defer bus.Unlock()
 	for _, t := range types {
-		bus.consumers[t] = append(bus.consumers[t], consumer)
+		bus.consumers[t] = append(bus.copyConsumers(t), consumer)
 	}
```
garagatyi, line 94: Since the whole function is covered by a lock do we really need to copy consumers before appending the new one?

agents/go-agents/core/event/event.go
```diff
@@ -81,19 +82,20 @@ func (bus *Bus) Pub(e E) {
 func (bus *Bus) Sub(consumer Consumer, eType string) {
 	bus.Lock()
 	defer bus.Unlock()
-	bus.consumers[eType] = append(bus.consumers[eType], consumer)
+	bus.consumers[eType] = append(bus.copyConsumers(eType), consumer)
```
garagatyi, line 85: Since the whole function is covered by a lock do we really need to copy consumers before appending the new one?

voievodin: Technically it will work without copying, as a new element will be added to the underlining array while concurrently reading(publishing) go routines will be working with slices referencing the same array but different elements.

The reason i changed this code was to achieve the same copy-on-write pattern concept throughout all bus state modification functions. For example `Bus.rm` copies values because otherwise it would change the original values of underlining array, including those being concurrently referenced by reading routines, which may produce unexpected results. _Fail-safe_ traversing, simplicity are the main reasons. 

If we take a look at `EventService` java implementation it carries `ConcurrentMap`of `CopyOnWriteArraySets`, so each write produces a new instance of underlining array, which is the same approach in terms of modification.

garagatyi: I thought that it always creates a new array, but now I see that there are cases when the slice is just resliced. 
Thank you for the explanation.

garagatyi (approved):

voievodin:
I'm not sure i can pass CLA check as i have no access to my previous email.
I would appreciate if some of you guys commit these changes and merge them. 
Thank you in advance!

benoitf:
@voievodin just register your new gmail through Eclipse ECA

here is the link https://accounts.eclipse.org/user/login?destination=user/eca

voievodin:
@benoitf thank you, i will register and pass the validation =)

voievodin:
@garagatyi, @akorneta all checks passed, feel free to merge when necessary

benoitf (commented):
ci-build

skabashnyuk:
@voievodin thx

PR #8142, opened 2018-01-03
Selenium: add checking that the Authorize button exists in the ImportProjectFromGitHubTest test

SkorikSergey:
### What does this PR do?
This PR adds checking that the **Authorize** button exists in the **ImportProjectFromGitHubTest** test!

### DIFF
--- a/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/GitHub.java
+++ b/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/GitHub.java
@@ -10,6 +10,8 @@
  */
 package org.eclipse.che.selenium.pageobject;
 
+import static org.eclipse.che.selenium.core.constant.TestTimeoutsConstants.LOAD_PAGE_TIMEOUT_SEC;
+
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
 import org.eclipse.che.selenium.core.SeleniumWebDriver;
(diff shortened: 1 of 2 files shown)


selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/dashboard/ImportProjectFromGitHubTest.java
```diff
@@ -116,11 +117,10 @@ private void connectGithubAccount() {
     gitHub.typeLogin(gitHubUsername);
     gitHub.typePass(gitHubPassword);
     gitHub.clickOnSignInButton();
-    loader.waitOnClosed();
+    WaitUtils.sleepQuietly(5);
```
tolusha: It is better to find another solution.

tolusha (changes requested):

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/GitHub.java
```diff
@@ -247,6 +247,8 @@ public void waitAuthorizeBtn() {
   /** click on authorize button */
   public void clickOnAuthorizeBtn() {
     waitAuthorizeBtn();
+    new WebDriverWait(seleniumWebDriver, 10)
```
tolusha: Use timeout constant

tolusha (approved):

PR #8154, opened 2018-01-03
User can specify project name using openshift ocp.sh.

JamesDrummond:
### What does this PR do?
User can specify project name using ocp.sh using environment variable CHE_OPENSHIFT_PROJECT.

NA

#### Release Notes
User can specify project name using openshift ocp.sh script.

Signed-off-by: James Drummond james@devcomb.com


riuvshin:
The reason why we didn't allow to change project name was that there are some other places where it is hardcoded. To merge that you have to find all impacted places and make sure it will respect ENV

JamesDrummond:
@riuvshin I have tested it and successfully created a workspace after setting the project name to something else. Will there be conflicts if two are running at the same time by the same user? Is there additional tests I can do?

JamesDrummond:
@riuvshin I created two different projects named "che-eclipse-che-test" and "che-eclipse-che-test2" and created two different workspaces with the same user. I don't have any issues. The only thing I can think of being an issue is workspace project clean up which I don't know how to do through CLI. I usually have to delete one by one if the che server is deleted improperly. 

I also have a different user that has it's own che project. I am not using multiuser though so I have to specify different project names.

I think another solution is to create a hash in the default project name values for the che server similar to workspaces. Something like `DEFAULT_CHE_OPENSHIFT_PROJECT="eclipse-che-${hash}"` for all but osio.

image

riuvshin:
@JamesDrummond have you tested this in multiuser mode? It may affect multi user

JamesDrummond:
Will test today.

sleshchenko (approved):
I believe it should work fine. What is hardcoded and can't be changed easily is Che route name like here https://github.com/eclipse/che/blob/c46b3e7748276b808f2fe6d55bfcca44bc0a1d58/dockerfiles/init/modules/openshift/files/scripts/che-config#L9
where `che` is a route name.

JamesDrummond:
@sleshchenko I don't have an issue keeping the route name which give the url a suffix of `che-`. The project name just allows for a unique url like `http://che-myprojectname.example.com/` for example. This is important for ocp multiuser setup where users need a way to setup a unique url address.

JamesDrummond:
@riuvshin Tested in multiuser mode. Works great :)

JamesDrummond:
@riuvshin Can you please review/approve?

riuvshin (approved):

PR #8158, opened 2018-01-03
Allow removal of existing Openshift projects using ocp.sh script.

JamesDrummond:
### What does this PR do?
Allow removal of existing Openshift projects using ocp.sh script.

NA

#### Release Notes
Allow removal of existing Openshift projects using ocp.sh script.

### DIFF
--- a/dockerfiles/init/modules/openshift/files/scripts/ocp.sh
+++ b/dockerfiles/init/modules/openshift/files/scripts/ocp.sh
@@ -70,6 +70,9 @@ export IMAGE_INIT=${IMAGE_INIT:-${DEFAULT_IMAGE_INIT}}:${CHE_IMAGE_TAG}
 DEFAULT_CONFIG_DIR="/tmp/che-config"
 export CONFIG_DIR=${CONFIG_DIR:-${DEFAULT_CONFIG_DIR}}
 
+DEFAULT_CHE_REMOVE_PROJECT=false
+export CHE_REMOVE_PROJECT=${CHE_REMOVE_PROJECT:-${DEFAULT_CHE_REMOVE_PROJECT}}
+
 }
 
 get_tools() {
@@ -156,6 +159,9 @@ deploy_che_to_ocp() {
     docker run -v "${CONFIG_DIR}":/to_remove alpine sh -c "rm -rf /to_remove/" || true
     docker run -t --rm -v /var/run/docker.sock:/var/run/docker.sock -v "${CONFIG_DIR}":/data -e IMAGE_INIT="$IMAGE_INIT" -e CHE_MULTIUSER="$CHE_MULTIUSER" eclipse/che-cli:${CHE_IMAGE_TAG} config --skip:pull --skip:nightly
     cd "${CONFIG_DIR}/instance/config/openshift/scripts/"
+    if $CHE_REMOVE_PROJECT; then
+      remove_che_from_ocp
+    fi
     bash deploy_che.sh ${DEPLOY_SCRIPT_ARGS}
     wait_until_server_is_booted
     if [ $CHE_MULTIUSER == true ]; then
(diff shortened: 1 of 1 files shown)


dockerfiles/init/modules/openshift/files/scripts/ocp.sh
```diff
@@ -156,6 +159,9 @@ deploy_che_to_ocp() {
     docker run -v "${CONFIG_DIR}":/to_remove alpine sh -c "rm -rf /to_remove/" || true
     docker run -t --rm -v /var/run/docker.sock:/var/run/docker.sock -v "${CONFIG_DIR}":/data -e IMAGE_INIT="$IMAGE_INIT" -e CHE_MULTIUSER="$CHE_MULTIUSER" eclipse/che-cli:${CHE_IMAGE_TAG} config --skip:pull --skip:nightly
     cd "${CONFIG_DIR}/instance/config/openshift/scripts/"
+    if $CHE_REMOVE_PROJECT; then
+      remove_che_from_ocp
```
riuvshin: I don;t think that remove che from ocp should be there

JamesDrummond: Needs to be there the way in the way I have programmed. The way that cli parameters are currently parsed I cannot be sure that remove will run before deploy if user puts `--deploy-che` before `--remove-che` in the cli command. This makes sure that remove is run before deploy no matter the order of the parameters. There is a way to search parameters in order but I tried not to change the existing code too much.

riuvshin: that could be invoked right in `parse_args()` instead of setting VAR just call the method.

JamesDrummond: That is the issue though. If deploy is executed before removal it will fail. `parse_args()` does not order parameters so I cannot be sure the removal function will run before deploy function.

Would you like me to rewrite `parse_args()` to search in order?

riuvshin: I meant just change this:

```
parse_args()
...

 +    if [[ "$@" == *"--remove-che"* ]]; then
 +      CHE_REMOVE_PROJECT=true
 +    fi
```

to this:

```
parse_args()
...

 +    if [[ "$@" == *"--remove-che"* ]]; then
 +      remove_che_from_ocp
 +    fi
```

it will be executed before `case`

JamesDrummond: @riuvshin Thanks. Yes that makes sense. Didn't think about doing it that way :) . Fixed please review.

riuvshin (approved):
lgtm

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
```diff
@@ -216,6 +218,38 @@ destroy_ocp() {
...
 }
 
+remove_che_from_ocp() {
+	echo "[CHE] Checking if project \"${CHE_OPENSHIFT_PROJECT}\" exists before removing..."
+	WAIT_FOR_PROJECT_TO_DELETE=true
+	CHE_REMOVE_PROJECT=true
+	DELETE_OPENSHIFT_PROJECT_MESSAGE="[CHE] Removing Project \"${CHE_OPENSHIFT_PROJECT}\"."
+	if $OC_BINARY get project "${CHE_OPENSHIFT_PROJECT}" &> /dev/null; then
+		echo "[CHE] Project \"${CHE_OPENSHIFT_PROJECT}\" exists."
+		while $WAIT_FOR_PROJECT_TO_EXIT
```
sleshchenko: Where initial value assigned to this variable? Maybe `WAIT_FOR_PROJECT_TO_DELETE` should be here

riuvshin: good catch! I guess `WAIT_FOR_PROJECT_TO_DELETE` should be renamed to `WAIT_FOR_PROJECT_TO_EXIT` or vise versa

JamesDrummond: @sleshchenko Thanks. Hard to test for problems like this. Fixed.

JamesDrummond:
@sleshchenko Please review.

JamesDrummond:
@sleshchenko Please review. Was a small change per your comment.

sleshchenko (approved):

PR #8159, opened 2018-01-03
Allow Openshift deploy_che.sh script to detect if project deletion is completed and new project successfully created.

JamesDrummond:
### What does this PR do?
Allow Openshift deploy_che.sh script to detect if project deletion is completed and new project successfully created.

N/A

#### Release Notes
Allow Openshift deploy_che.sh script to detect if project deletion is completed and new project successfully created.

### DIFF
--- a/dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
+++ b/dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
@@ -282,13 +282,39 @@ if ! oc get project "${CHE_OPENSHIFT_PROJECT}" &> /dev/null; then
   if [ "${COMMAND}" == "cleanup" ] || [ "${COMMAND}" == "rollupdate" ]; then echo "**ERROR** project doesn't exist. Aborting"; exit 1; fi
   if [ "${OPENSHIFT_FLAVOR}" == "osio" ]; then echo "**ERROR** project doesn't exist on OSIO. Aborting"; exit 1; fi
 
-  echo -n "Project does not exist...creating it..."
-  oc new-project "${CHE_OPENSHIFT_PROJECT}" &> /dev/null
+  # OpenShift will not get project but project still exists for for a period after being deleted.
+  # The following will loop until it can create successfully.
+
+  WAIT_FOR_PROJECT_TO_DELETE=true
+  WAIT_FOR_PROJECT_TO_DELETE_MESSAGE="Waiting for project to be deleted fully(~15 seconds)..."
+
+  echo "Project \"${CHE_OPENSHIFT_PROJECT}\" does not exist...trying to creating it."
+  while $WAIT_FOR_PROJECT_TO_DELETE
+  do
+  { # try
+
+      oc new-project "${CHE_OPENSHIFT_PROJECT}" &> /dev/null && \
+      WAIT_FOR_PROJECT_TO_DELETE=false # Only excutes if project creation is successfully
+  } || { # catch
+      echo -n $WAIT_FOR_PROJECT_TO_DELETE_MESSAGE
+      WAIT_FOR_PROJECT_TO_DELETE_MESSAGE="."
+      sleep 2
+  }
+  done
 fi
-echo "done!"
+echo "Project \"${CHE_OPENSHIFT_PROJECT}\" creation done!"
 
 echo -n "[CHE] Switching to \"${CHE_OPENSHIFT_PROJECT}\"..."
-oc project "${CHE_OPENSHIFT_PROJECT}" &> /dev/null
+WAIT_TO_SWITCH_TO_PROJECT=true
+while $WAIT_TO_SWITCH_TO_PROJECT
+do
+{ # try
+    oc project "${CHE_OPENSHIFT_PROJECT}" &> /dev/null && \
+    WAIT_TO_SWITCH_TO_PROJECT=false # Only excutes after project creation is successfully
+} || { # catch
+    echo -n "."
+}
+done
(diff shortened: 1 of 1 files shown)


JamesDrummond:
@riuvshin @benoitf @sleshchenko Please review.

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
```diff
@@ -282,13 +282,39 @@ if ! oc get project "${CHE_OPENSHIFT_PROJECT}" &> /dev/null; then
   if [ "${COMMAND}" == "cleanup" ] || [ "${COMMAND}" == "rollupdate" ]; then echo "**ERROR** project doesn't exist. Aborting"; exit 1; fi
   if [ "${OPENSHIFT_FLAVOR}" == "osio" ]; then echo "**ERROR** project doesn't exist on OSIO. Aborting"; exit 1; fi
 
-  echo -n "Project does not exist...creating it..."
-  oc new-project "${CHE_OPENSHIFT_PROJECT}" &> /dev/null
+  # OpenShift will not get project but project still exists for for a period after being deleted.
```
sleshchenko: typo double **for**

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
```diff
@@ -282,13 +282,39 @@ if ! oc get project "${CHE_OPENSHIFT_PROJECT}" &> /dev/null; then
...
+
+  WAIT_FOR_PROJECT_TO_DELETE=true
+  WAIT_FOR_PROJECT_TO_DELETE_MESSAGE="Waiting for project to be deleted fully(~15 seconds)..."
+
+  echo "Project \"${CHE_OPENSHIFT_PROJECT}\" does not exist...trying to creating it."
+  while $WAIT_FOR_PROJECT_TO_DELETE
+  do
+  { # try
+
+      oc new-project "${CHE_OPENSHIFT_PROJECT}" &> /dev/null && \
```
sleshchenko: It can fail if another user has a project with the same name. So infinitive loop will be in this case. Please consider adding timeout for waiting.

JamesDrummond: @sleshchenko I don't think you can have two projects with the same name in OpenShift can you? This is one of the reason I created PR [https://github.com/eclipse/che/pull/8154](https://github.com/eclipse/che/pull/8154). I can add a timeout though anyways.

sleshchenko: @JamesDrummond You can't. So script expects that error is temporary (like a project is marked to remove and removing is in progress) and it will disappear when a project will be deleted. 
But error can be permanent - like another user has a project with the same name. Then infinitive loop will be here. 
P.S. Let me know if my explanation is not clear enough. I'll try to reword it.

JamesDrummond: @sleshchenko In OpenShift I don't think it is possible to have two projects with the same name. Even seperate users can't use a project name if it's already taken. I actually remember when I used osio this was even an issue. Took some screen captures below where james@devcomb.com has eclipse-che project and developer user cannot create a project with same name.

image

image

image

If another user just happens to create a project with the same name right after you delete yours, the user can kill the script manually. This is not a likely event. If the project has issues being deleted then this is really another problem that is not likely to happen. I use this script often and rarely have any issues.

riuvshin: everybody agrees that you can't have 2 projects with the same name, but what @sleshchenko meant is that if this error will happen script will go into endless loop which is no OK and there is should be a timer or some kind of interruptor.

JamesDrummond: @riuvshin Sorry didn't understand. Will add timer. Thanks.

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
```diff
@@ -279,16 +279,55 @@ echo "done!"
...
+        sleep $POLLING_INTERVAL_SEC
+    }
+    done
 fi
-echo "done!"
+echo "Project \"${CHE_OPENSHIFT_PROJECT}\" creation done!"
 
 echo -n "[CHE] Switching to \"${CHE_OPENSHIFT_PROJECT}\"..."
-oc project "${CHE_OPENSHIFT_PROJECT}" &> /dev/null
+WAIT_TO_SWITCH_TO_PROJECT=true
```
sleshchenko: Do we really need this waiting? As far as I see a current project is automatically switched to newly created after `oc new-project`. When there can be a delay between project creation (`oc new-project`) and ability to switch to it?

JamesDrummond: @sleshchenko I think the intent was to allow switch if project already existed but I cannot see the reason to do this anymore. Removed.

JamesDrummond:
@sleshchenko Please review.

sleshchenko (approved):

riuvshin (approved):

sleshchenko:
@JamesDrummond Have you tested `rollupdate` command after your changes?

It doesn't work for me.

screenshot_20180118_101016

Looks like this else section brokes `rollupdate` command https://github.com/eclipse/che/pull/8159/files#diff-c656525c29bc2779734318b21b1b983dR313

PR #8160, opened 2018-01-04
Automatically set the CHE_INFRA_OPENSHIFT_OAUTH__TOKEN.

JamesDrummond:
### What does this PR do?
Automatically set the CHE_INFRA_OPENSHIFT_OAUTH__TOKEN to the project's service account for minishift and ocp. 

CHE_INFRA_OPENSHIFT_[USERNAME|PASSWORD] are set to "" for minishift and ocp. 

_Additional conversation is now needed to determine if CHE_INFRA_OPENSHIFT_[USERNAME|PASSWORD] variables are even needed. 

N/A

#### Release Notes
Automatically set Openshift's deployment configuration environment variable  CHE_INFRA_OPENSHIFT_OAUTH__TOKEN to project's service account.

### DIFF
--- a/dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
+++ b/dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
@@ -377,6 +377,10 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: che" | oc apply -f -
+  CHE_INFRA_OPENSHIFT_USERNAME=""
+  CHE_INFRA_OPENSHIFT_PASSWORD=""
+  oc create serviceaccount che -n ${CHE_OPENSHIFT_PROJECT}
+  CHE_INFRA_OPENSHIFT_OAUTH__TOKEN=$(oc sa get-token che -n ${CHE_OPENSHIFT_PROJECT})  
 fi
 
 # ----------------------------------------------


JamesDrummond:
@sleshchenko Let me know if you can review this. This is a really good way for me not to have to set the token for infra or use user token that expires. The service account token is longer lived. Not sure if it expires at all. Based on the use of a service account I would think infra username/password would almost never be used/needed. There might be certain setups that would not let che service account have admin permission to create projects/workspaces but I cannot see a case of this. Let me know your thoughts.

sleshchenko:
@JamesDrummond Are you sure that che service account has permissions to create projects? When I tried it, it was able to manage OpenShift objects in its namespace. But when I tried to create projects I got `permissions dinied` exception. But I used mounted service account token(DefaultOpenshiftClient uses it by default) instead of manually fetching it.
Let me know if I did something wrong

JamesDrummond:
@sleshchenko You are right. I guess I granted permissions to the service account before I forgot about it as I ran into the same issues you did before. Thought I magically got it working ;)

I was thinking about adding this roles to the minishift's che sa as default since the user will most likely have system:admin access. However, it is not for sure that a user on ocp will have system:admin access. 

Do you think we could try with ocp to add role to sa with system:admin account and catch any errors which would then keep the infra token value set to the openshift_token? If you don't know what I mean I can code it up for review.

This is how to add project creation access to sa. Requires system:admin access.
```
oc login -u system:admin
oc create sa che -n eclipse-che
oadm policy add-cluster-role-to-user self-provisioner system:serviceaccount:eclipse-che:che
#test
oc login --token=$(oc sa get-token che -n eclipse-che)
oc new-project test
```

System admins can grant all service accounts permissions to create accounts with the following.
```
oc login -u system:admin
oc create sa che -n eclipse-che
oadm policy add-cluster-role-to-user self-provisioner system:serviceaccount
```

I am wondering which option is better for ocp. Try to gain access to system:admin in the script through the first method or require system admin to set this up for all service accounts. I think the first option would be best as I don't think that service accounts have quotas of resource constraints put on them...

Do you know if user account tokens can be made to last longer? Maybe that could be a solution.

JamesDrummond:
@sleshchenko Do you know why things got changed from creating workspace pods in the same project to creating workspace pods in a separate project? I know sa can create pods.

Update: I think it's because project resources can be limited so having all workspaces in the same project will not function well. Openshift can limit user resources by limiting the amount of projects the user can create. Allowing service accounts to create project when the user is not the system admin would therefore not be good thing. OK I am going to scrap this PR.

I think a better way if you have system:admin rights is to set longer token timeouts for their users.

[https://docs.openshift.com/enterprise/3.0/admin_guide/configuring_authentication.html#token-options](https://docs.openshift.com/enterprise/3.0/admin_guide/configuring_authentication.html#token-options)

sleshchenko:
@JamesDrummond 

> Do you know why things got changed from creating workspace pods in the same project to creating workspace pods in a separate project?

The Resources quotas issue is only one of the issues with creating all workspace objects in one OpenShift Project.
The another is one is OpenShift objects conflicting like two workspaces can't have service with the same name but sometimes they need it.
The one more issue is about restriction access from objects of one workspace to another. In one OpenShift project, all services are accessible for all containers. And it is a problem for the multiuser use case.

But, if you want to use ONE OpenShift project you are able to configure Che Service in this way.
So, please export the following variables
```
export CHE_INFRA_OPENSHIFT_USERNAME=""
export CHE_INFRA_OPENSHIFT_PASSWORD=""
export CHE_INFRA_OPENSHIFT_OAUTH__TOKEN=""
export CHE_INFRA_OPENSHIFT_PROJECT={TARGET_PROJECT_NAME_HERE}
```
Then one OpenShift project will be used and `Che` service account token will be used while OpenShift objects managing.

Issue #7922, opened 2017-12-18
OpenShift multi-container workspace fails to start


sleshchenko:
### Description
OpenShift multi-container workspace fails to start. It happens because of wrong resolving machine servers when containers are located in one Pod.

### Reproduction Steps
1. Create a workspace with the following OpenShift multi-container workspace configuration:
<details>
<summary>Workspace Config</summary>

```json
{
  "environments": {
    "default": {
      "machines": {
        "ws/typescript-ls": {
          "servers": {
            "ls": {
              "attributes": {
                "internal": "true",
                "type": "ls",
                "config": "{\"id\":\"org.eclipse.che.plugin.web.typescript\", \"documentFilters\":[  {  \"languageId\":\"typescript\",  \"pathRegex\":\".*\\\\.(ts|tsx)\"} ]}"
              },
              "port": "4417",
              "protocol": "tcp"
            }
          },
          "volumes": {
            "projects": {
              "path": "/projects"
            }
          },
          "installers": [],
          "env": {}
        },
        "ws/dev": {
          "attributes": {},
          "servers": {
            "codeserver": {
              "attributes": {},
              "port": "9876",
              "protocol": "http"
            }
          },
          "volumes": {
            "projects": {
              "path": "/projects"
            }
          },
          "installers": [
            "org.eclipse.che.exec",
            "org.eclipse.che.terminal",
            "org.eclipse.che.ws-agent"
          ],
          "env": {}
        }
      },
      "recipe": {
        "type": "openshift",
        "content": "---\nkind: List\nitems:\n-\n  apiVersion: v1\n  kind: Pod\n  metadata:\n    name: ws\n  spec:\n    containers:\n      -\n        image: eivantsov/tsls\n        name: typescript-ls\n      -\n        image: eclipse/ubuntu_jdk8\n        name: dev",
        "contentType": "application/x-yaml"
      }
    }
  },
  "defaultEnv": "default",
  "projects": [
    {
      "links": [],
      "name": "greeter",
      "attributes": {},
      "type": "typescript",
      "source": {
        "location": "https://github.com/Microsoft/TypeScriptSamples.git",
        "type": "git",
        "parameters": {
          "convertToTopLevelProject": "true",
          "keepVcs": "false",
          "keepDir": "greeter"
        }
      },
      "path": "/greeter",
      "problems": [],
      "mixins": []
    }
  ],
  "name": "TypeScript_LS_Demo",
  "commands": [],
  "links": []
}
```
</details>
[...]

PR #8182, opened 2018-01-04
Fixed resolving of servers of an OpenShift multi-container workspace

sleshchenko:
### What does this PR do?
Previously ServerResolver analyzed servers of a machine (container) by real OpenShift object connections. But it works wrong while Service is linked with Pod but not a particular container (service expose all containers in one Pod at the same time).
While OpenShift infrastructure creates Services and Routes for Che Servers, it would be simpler to serialize machine name into Route/Service annotations and use it while servers resolving instead of analyzing real OpenShift objects connections. So in this PR, it is implemented in the described way and it fixes resolving of servers of an OpenShift multi-container workspace.

This PR also contains the following minor changes:
- Added more information to bootstrapper debug logging;
- Since RuntimeContext **MUST** return non-null value checking Runtime on null is redundant. So it is removed;
- Fixed java doc for RuntimeInfrastructure#prepare.

https://github.com/eclipse/che/issues/7922

### DIFF
--- a/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/Annotations.java
+++ b/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/Annotations.java
@@ -22,8 +22,8 @@ import org.eclipse.che.api.core.model.workspace.config.ServerConfig;
 import org.eclipse.che.api.workspace.server.model.impl.ServerConfigImpl;
 
 /**
- * Helps to convert {@link ServerConfig} objects to OpenShift infrastructure annotations and
- * vise-versa.
+ * Helps to convert servers related entities (like {@link ServerConfig} and machine name) to
+ * OpenShift annotations and vise-versa.
  *
  * @author Sergii Leshchenko
  */
@@ -36,6 +36,8 @@ public class Annotations {
   public static final String SERVER_PATH_ANNOTATION_FMT = ANNOTATION_PREFIX + "server.%s.path";
   public static final String SERVER_ATTR_ANNOTATION_FMT =
       ANNOTATION_PREFIX + "server.%s.attributes";
+
+  public static final String MACHINE_NAME_ANNOTATION = ANNOTATION_PREFIX + "machine.name";
 
   /** Pattern that matches server annotations e.g. "org.eclipse.che.server.exec-agent.port". */
   private static final Pattern SERVER_ANNOTATION_PATTERN =
@@ -86,6 +88,11 @@ public class Annotations {
       return this;
     }
 
+    public Serializer machineName(String machineName) {
+      annotations.put(MACHINE_NAME_ANNOTATION, machineName);
+      return this;
+    }
+
     public Map<String, String> annotations() {
       return annotations;
     }
@@ -120,6 +127,10 @@ public class Annotations {
         }
       }
       return servers;
+    }
+
+    public String machineName() {
+      return annotations.get(MACHINE_NAME_ANNOTATION);
     }
   }
 
--- a/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/bootstrapper/OpenShiftBootstrapper.java
+++ b/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/bootstrapper/OpenShiftBootstrapper.java
@@ -114,14 +114,16 @@ public class OpenShiftBootstrapper extends AbstractBootstrapper {
   }
 
   private void injectBootstrapper() throws InfrastructureException {
-    LOG.debug("Creating folder for bootstrapper");
+    String machineName = openShiftMachine.getName();
+    LOG.debug(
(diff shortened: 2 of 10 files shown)


skabashnyuk (approved):

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/bootstrapper/OpenShiftBootstrapper.java
```diff
@@ -114,14 +114,16 @@ protected void doBootstrapAsync(String installerWebsocketEndpoint, String output
   }
 
   private void injectBootstrapper() throws InfrastructureException {
-    LOG.debug("Creating folder for bootstrapper");
+    LOG.debug(
+        "Bootstrapping {}:{}. Creating folder for bootstrapper", runtimeIdentity, openShiftMachine);
```
garagatyi: Does OpenShift machine have meaningful `toString()` to use it here?

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/bootstrapper/OpenShiftBootstrapper.java
```diff
@@ -114,14 +114,16 @@ protected void doBootstrapAsync(String installerWebsocketEndpoint, String output
...
     openShiftMachine.exec("mkdir", "-p", BOOTSTRAPPER_DIR);
-    LOG.debug("Downloading bootstrapper binary");
+    LOG.debug(
+        "Bootstrapping {}:{}. Downloading bootstrapper binary", runtimeIdentity, openShiftMachine);
```
garagatyi: openShiftMachine is already logged in line 117. We can avoid it here. WDYT?

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/bootstrapper/OpenShiftBootstrapper.java
```diff
@@ -114,14 +114,16 @@ protected void doBootstrapAsync(String installerWebsocketEndpoint, String output
...
     openShiftMachine.exec(
         "curl", "-o", BOOTSTRAPPER_DIR + BOOTSTRAPPER_FILE, bootstrapperBinaryUrl);
     openShiftMachine.exec("chmod", "+x", BOOTSTRAPPER_DIR + BOOTSTRAPPER_FILE);
 
-    LOG.debug("Creating bootstrapper config file");
+    LOG.debug("Bootstrapping {}:{}. Creating config file", runtimeIdentity, openShiftMachine);
```
garagatyi: Se above

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/Annotations.java
```diff
@@ -37,6 +37,8 @@
   public static final String SERVER_ATTR_ANNOTATION_FMT =
       ANNOTATION_PREFIX + "server.%s.attributes";
 
+  public static final String EXPOSED_MACHINE_ANNOTATION = ANNOTATION_PREFIX + "exposed_machine";
```
garagatyi: Can we avoid adding a new term `exposed machine` and use something like `machine_name`

garagatyi (commented):
Looks good and the new approach to resolve servers is simpler.  
Looks like there are much fewer changes in tests than in code. Is it on-purpose?

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/bootstrapper/OpenShiftBootstrapper.java
sleshchenko: Good catch. `OpenShiftMachine.getName()` should be there

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/bootstrapper/OpenShiftBootstrapper.java
sleshchenko: I think OpenShift machine name can be useful here. Because infrastructure can implement bootstrapping in a parallel way, then logging won't be sequential and it won't be clear to which machine these logs belong to. Does it make sense?

garagatyi: Yes, with the name it's OK, I was concerned about the whole object

sleshchenko:
@garagatyi Nice catch. There were missed tests for `Annotations` and `ServerExposer` classes. So I added them.

akorneta (approved):

garagatyi (approved):

Issue #8056, opened 2017-12-26
Set machine status to RUNNING when it is started


sleshchenko:
### Description
Now OpenShift and Docker infrastructure set RUNNING status for the machine only when servers checkings are finished. 

There was a discussion with @gazarenkov and  @garagatyi where was decided to declare that machine status is independent of server status. It does infrastructure freer while servers checking implementing. And clients are able to keep track off when a machine is started (for example image is pulled and a container is started) but servers are not ready yet.

So, it is needed to rework OpenShift and Docker infrastructure to set RUNNING status when it is started. And declare somewhere in java doc that machines statuses don't depend on servers statuses.

PR #8192, opened 2018-01-05
Restore functionality of uploading SSH keys to machine on startup.

### LINKED ISSUE #6778: Fix disabled code from plugin-machine
### Description
In Che6 branch we have unused code in plugin-machine. Looks like at least some of the functionality of this code is reimplemented in wsagent ssh-key ssh-server modules. 
We need to check what we can remove and what we should fix.

Missing functionality is ssh key injector. We need to reimplement it with reimplementing it with help of exec agent. I think it will be a mandatory dependency for this functionality.

### Reproduction Steps

**OS and version:**    

**Diagnostics:**

mshaposhnik:
### What does this PR do?
Restores uploading SSH keys to machine on startup using exec agents.

Fixes #6778 

#### Release Notes
Restored uploading SSH keys to machine on startup using exec agents.

#### Docs PR
N/A


plugins/plugin-machine/che-plugin-machine-ext-server/src/main/java/org/eclipse/che/ide/ext/machine/server/ssh/KeysInjector.java
```diff
@@ -54,72 +75,100 @@ public void start() {
         new EventSubscriber<MachineStatusEvent>() {
           @Override
           public void onEvent(MachineStatusEvent event) {
-            if (event.getEventType() == MachineStatus.RUNNING) {
-              /*final Instance machine;
+            if (event.getEventType() != MachineStatus.RUNNING) {
```
sleshchenko: I think you should rework it to ServerStatusEvent. Because when Machine is running it doesn't mean that exec agent is RUNNING. See https://github.com/eclipse/che/issues/8056

mshaposhnik: Its a problem than. Because startup of sinlge agent is not guarantee that all other agents will start successfully. Startup still may be failed later. So only way it to do that later, on whole `workspace` startup.

sleshchenko: I'd say that it is not related to listening server statuses or machine statuses. Because if one machine is started it is not guaranteed that all machines will be started.
But I don't think that it is a problem. You inject ssh key as soon as possible. Can you elaborate what is the problem of failing a workspace when ssh key is already injected?

mshaposhnik: because:
1. This is not logical (inject SSH key each time when exec agent started? how they are related?)
2. This is useless action when startup failed.
3. This is highly error risky (you submitiing a task to workspace which is begin to shutdown and state is changing rapidly, you can loss your actual state of command etc)

plugins/plugin-machine/che-plugin-machine-ext-server/src/main/java/org/eclipse/che/ide/ext/machine/server/ssh/KeysInjector.java
```diff
@@ -54,72 +75,100 @@ public void start() {
...
+              SshPairImpl sshWorkspacePair = null;
               try {
-                  machine = environmentEngine.getMachine(event.getWorkspaceId(),
-                                                         event.getMachineId());
+                sshWorkspacePair = sshManager.getPair(user.getId(), "workspace", workspaceId);
               } catch (NotFoundException e) {
-                  LOG.error("Unable to find machine: " + e.getLocalizedMessage(), e);
-                  return;
-              }*/
+                LOG.debug("No ssh key associated to the workspace", e);
```
sleshchenko: Please add workspace related info here, like workspace id.

plugins/plugin-machine/che-plugin-machine-ext-server/src/main/java/org/eclipse/che/ide/ext/machine/server/ssh/KeysInjector.java
```diff
@@ -54,72 +75,100 @@ public void start() {
...
-                      command.append("&& echo '")
-                             .append(publicKey)
-                             .append("' >> ~/.ssh/authorized_keys");
-                  }
+              GetProcessResponseDto process = client.getProcess(workspaceId, startProcess.getPid());
+              int tryNumber = 1;
+              while (process.isAlive()) {
+                process = client.getProcess(workspaceId, startProcess.getPid());
+                tryNumber++;
+                if (tryNumber > 10) {
```
sleshchenko: Are you sure that killing of the process after 10 failed tries is OK? I would say no while there are not any delays between tries.

plugins/plugin-machine/che-plugin-machine-ext-server/src/main/java/org/eclipse/che/ide/ext/machine/server/ssh/KeysInjector.java
sleshchenko: Makes sense. But 2 and 3 points are also related MachineStatusEvent too. So `WorkspaceStatus` is OK for me.

ide/che-core-ide-api/src/main/resources/org/eclipse/che/agent/exec/Exec.gwt.xml
```diff
@@ -0,0 +1,19 @@
...
+    http://www.eclipse.org/legal/epl-v10.html
+
+    Contributors:
+      Red Hat, Inc. - initial API and implementation
+
+-->
+<!DOCTYPE module PUBLIC "-//Google Inc.//DTD Google Web Toolkit 2.7.0//EN" "http://gwtproject.org/doctype/2.7.0/gwt-module.dtd">
+<module>
+    <inherits name="com.google.gwt.user.User"/>
+    <inherits name="com.google.gwt.json.JSON"/>
```
azatsarynnyy: Those two `inherits` are useless there so you can remove it.

azatsarynnyy (approved):

agents/exec/che-exec-agent-client/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-dto</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-workspace</artifactId>
+        </dependency>
+    </dependencies>
+    <build />
```
sleshchenko: Please remove this empty section

agents/exec/che-exec-agent-client/src/main/java/org/eclipse/che/agent/exec/client/ExecAgentClient.java
```diff
@@ -0,0 +1,142 @@
...
+  public ProcessKillResponseDto killProcess(String workspaceId, int pid) throws ServerException {
+    try {
+      return requestFactory
+          .fromUrl(serverEndpoint + "/" + pid)
+          .useDeleteMethod()
+          .addQueryParam("token", machineTokenProvider.getToken(workspaceId))
+          .setAuthorizationHeader("none") // to prevent sending KC token
+          .request()
+          .asDto(ProcessKillResponseDto.class);
+    } catch (IOException
```
sleshchenko: I think the following variant would be simpler
```
} catch (IOException | ApiException | MachineTokenException e)
  ...
}
```

agents/exec/che-exec-agent-shared/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+        <artifactId>exec-agent-parent</artifactId>
+        <groupId>org.eclipse.che</groupId>
+        <version>6.0.0-M5-SNAPSHOT</version>
+    </parent>
+    <artifactId>che-exec-agent-shared</artifactId>
+    <packaging>jar</packaging>
+    <name>Agent :: Exec :: Shared</name>
+    <properties>
+        <dto-generator-out-directory>${project.build.directory}/generated-sources/dto/</dto-generator-out-directory>
+        <findbugs.failonerror>false</findbugs.failonerror>
```
sleshchenko: Please elaborate why we can't enable `findbugs` here.

plugins/plugin-machine/che-plugin-machine-ext-server/src/main/java/org/eclipse/che/ide/ext/machine/server/ssh/KeysInjector.java
```diff
@@ -29,97 +46,120 @@
...
-              } catch (IOException | ServerException e) {
-                  LOG.error(e.getLocalizedMessage(), e);
-              }*/
+                }
+                if (process.getExitCode() != 0) {
+                  LOG.warn("Uploading key failed with exit code " + process.getExitCode());
+                }
+              }
+            } catch (ServerException | InterruptedException e) {
+              LOG.error(e.getLocalizedMessage(), e);
```
sleshchenko: Maybe it would be useful to add a message that injection of ssh keys failed.

plugins/plugin-machine/che-plugin-machine-ext-server/src/main/java/org/eclipse/che/ide/ext/machine/server/ssh/KeysInjector.java
```diff
@@ -29,97 +46,120 @@
...
-          public void onEvent(MachineStatusEvent event) {
-            if (event.getEventType() == MachineStatus.RUNNING) {
-              /*final Instance machine;
+          public void onEvent(WorkspaceStatusEvent event) {
+            if (event.getStatus() != WorkspaceStatus.RUNNING) {
+              return;
+            }
+            final String workspaceId = event.getWorkspaceId();
+            final String userId = EnvironmentContext.getCurrent().getSubject().getUserId();
+            final List<String> serverURLs = getExecServerUrls(workspaceId);
```
sleshchenko: Let's rename this variable to `execServersURL`?

plugins/plugin-machine/che-plugin-machine-ext-server/src/main/java/org/eclipse/che/ide/ext/machine/server/ssh/KeysInjector.java
```diff
@@ -29,97 +46,120 @@
...
   @PostConstruct
   public void start() {
     eventService.subscribe(
-        new EventSubscriber<MachineStatusEvent>() {
+        new EventSubscriber<WorkspaceStatusEvent>() {
           @Override
-          public void onEvent(MachineStatusEvent event) {
-            if (event.getEventType() == MachineStatus.RUNNING) {
-              /*final Instance machine;
+          public void onEvent(WorkspaceStatusEvent event) {
```
sleshchenko: This method is too long. I think it would be better to rewrite it using private methods. Then body of this method will show common logic, like:
```
@Override
public void onEvent(WorkspaceStatusEvent event) {
  final List<String> execServerURLs = getExecServerUrls(workspaceId);
  if (serverURLs.isEmpty()) {
    return; // no exec servers installed, so not possible to execute commands
  }
  
  List<String> publicKeys = getPublicKeys(workspaceId);
  if (publicKeys.isEmpty() {
    //nothing to inject
   return;
  }

  for (String execServerURL : execServerURLs) {
    doInjectPublicKeys(publicKeys);
  }
}
```
Does it make sense?

plugins/plugin-machine/che-plugin-machine-ext-server/src/test/java/org/eclipse/che/ide/ext/machine/server/ssh/KeysInjectorTest.java
```diff
@@ -10,284 +10,265 @@
...
-  //
-  //        verify(lineConsumer).writeLine(eq("Error of injection public ssh keys. FAILED"));
-  //    }
+
+  private static final String WORKSPACE_ID = "workspace123";
+  private static final String OWNER = "user123";
+
+  @Captor private ArgumentCaptor<EventSubscriber<WorkspaceStatusEvent>> subscriberCaptor;
+
+  @Mock Exec exec;
```
sleshchenko: As far as I see `exec` and `logMessage` fields are not used anywhere. Please review tests.

sleshchenko (approved):

plugins/plugin-machine/che-plugin-machine-ext-server/src/main/java/org/eclipse/che/ide/ext/machine/server/ssh/KeysInjector.java
```diff
@@ -29,97 +45,130 @@
...
-                          try {
-                              machine.getLogger().writeLine("Error of injection public ssh keys. " + logMessage.getContent());
-                          } catch (IOException ignore) {
-                          }
-                      }
-                  });
-              } catch (IOException | ServerException e) {
-                  LOG.error(e.getLocalizedMessage(), e);
-              }*/
+            StringBuilder commandLine = new StringBuilder("mkdir ~/.ssh/ -p");
```
garagatyi: Maybe add a space to the end of this line?

plugins/plugin-machine/che-plugin-machine-ext-server/src/main/java/org/eclipse/che/ide/ext/machine/server/ssh/KeysInjector.java
```diff
@@ -29,97 +45,130 @@
...
+            for (String publicKey : publicKeys) {
+              commandLine
+                  .append("&& echo '")
+                  .append(publicKey)
+                  .append("' >> ~/.ssh/authorized_keys");
```
garagatyi: Maybe add a space to the end of this line?

plugins/plugin-machine/che-plugin-machine-ext-server/src/main/java/org/eclipse/che/ide/ext/machine/server/ssh/KeysInjector.java
```diff
@@ -29,97 +45,130 @@
...
+            }
+            for (String serverUrl : execServerUrls) {
+              doInjectPublicKeys(serverUrl, workspaceId, commandLine.toString());
```
garagatyi: Can we perform `toString` call once?

plugins/plugin-machine/che-plugin-machine-ext-server/src/main/java/org/eclipse/che/ide/ext/machine/server/ssh/KeysInjector.java
```diff
@@ -29,97 +45,130 @@
...
-    //        this.environmentEngine = environmentEngine;
+    this.workspaceManager = workspaceManager;
+    this.execAgentClientFactory = execAgentClientFactory;
   }
 
   @PostConstruct
   public void start() {
     eventService.subscribe(
-        new EventSubscriber<MachineStatusEvent>() {
+        new EventSubscriber<WorkspaceStatusEvent>() {
```
garagatyi: This subscriber looks quite heavy. I think it would be better not to hold event service thread such a long time on such a common event as workspace running event.

skabashnyuk (approved):

garagatyi (approved):
Looks good!

PR #8205, opened 2018-01-09
Make user home dir writable for OpenShift arbitrary user

ghost:
This change makes home dir writable for an arbitrary OpenShift user and is required for JS plugin development flow using eclipse/che-dev:nightly image


riuvshin (approved):
+

PR #8210, opened 2018-01-09
Make machine present in workspace runtime when starting event is sent

### LINKED ISSUE #7900: Machine is absent in workspace runtime after receiving an event that machine become starting
### Description
 

Client subscribes on `machine/statusChanged` events and listens changing states of machines.
After machine status has been changed to STARTING client must send internal `MachineStartingEvent`.
But before sending the event IDE updates the workspace by sending GET request to `/api/workspace`.
As it turns out the workspace runtime doesn't contain machine which status has changed to STARTING and it appears in the runtime after few seconds.
This leads to internal `MachineStartingEvent` is not sent and IDE does not react on changing of the machine state.

### LINKED ISSUE #8056: Set machine status to RUNNING when it is started

garagatyi:
### What does this PR do?
Make machine present in workspace runtime when starting event is sent.
Also set machine RUNNING status before bootstrapping installers and
checking servers statuses.

Fixes #7900 
Fixes #8056

### DIFF
--- a/infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerInternalRuntime.java
+++ b/infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerInternalRuntime.java
@@ -149,7 +149,7 @@ public class DockerInternalRuntime extends InternalRuntime<DockerRuntimeContext>
       DockerMachine machine = machineCreator.create(container);
       String name = Labels.newDeserializer(container.getLabels()).machineName();
 
-      runtimeMachines.addMachine(name, machine);
+      runtimeMachines.putMachine(name, machine);
       stopDetector.startDetection(container.getId(), name, new AbnormalMachineStopHandlerImpl());
       streamLogsAsync(name, container.getId());
     }
@@ -193,7 +193,7 @@ public class DockerInternalRuntime extends InternalRuntime<DockerRuntimeContext>
         checkInterruption();
         String machineName = containerEntry.getKey();
 
-        runtimeMachines.addMachine(machineName, new DockerMachine.StartingDockerMachine());
+        runtimeMachines.putMachine(machineName, new DockerMachine.StartingDockerMachine());
         sendStartingEvent(machineName);
 
         try {
@@ -302,7 +302,7 @@ public class DockerInternalRuntime extends InternalRuntime<DockerRuntimeContext>
             identity,
             new AbnormalMachineStopHandlerImpl());
     try {
-      runtimeMachines.replaceMachine(name, machine);
+      runtimeMachines.putMachine(name, machine);
 
       return machine;
     } catch (InfrastructureException e) {
--- a/infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerInternalRuntime.java
+++ b/infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerInternalRuntime.java
@@ -21,7 +21,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.CountDownLatch;
 import java.util.function.Consumer;
 import org.eclipse.che.api.core.model.workspace.Warning;
 import org.eclipse.che.api.core.model.workspace.runtime.Machine;
@@ -58,11 +57,16 @@ import org.eclipse.che.workspace.infrastructure.docker.network.NetworkLifecycle;
 import org.eclipse.che.workspace.infrastructure.docker.server.mapping.ExternalIpURLRewriter;
 import org.slf4j.Logger;
 
-/** @author Alexander Garagatyi */
+/**
+ * Represents {@link InternalRuntime} for Docker infrastructure.
+ *
+ * @author Alexander Garagatyi
+ */
 public class DockerInternalRuntime extends InternalRuntime<DockerRuntimeContext> {
 
   private static final Logger LOG = getLogger(DockerInternalRuntime.class);
 
+  private final RuntimeMachines runtimeMachines;
   private final StartSynchronizer startSynchronizer;
   private final Map<String, String> properties;
   private final NetworkLifecycle networks;
@@ -145,7 +149,7 @@ public class DockerInternalRuntime extends InternalRuntime<DockerRuntimeContext>
       DockerMachine machine = machineCreator.create(container);
       String name = Labels.newDeserializer(container.getLabels()).machineName();
 
-      startSynchronizer.addMachine(name, machine);
+      runtimeMachines.addMachine(name, machine);
       stopDetector.startDetection(container.getId(), name, new AbnormalMachineStopHandlerImpl());
       streamLogsAsync(name, container.getId());
     }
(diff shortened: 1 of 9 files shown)


garagatyi:
ci-build

garagatyi:
ci-test

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerInternalRuntime.java
```diff
@@ -533,6 +557,18 @@ public synchronized void addMachine(String name, DockerMachine machine)
       }
     }
 
+    public synchronized void replaceMachine(String name, RunningDockerMachine machine)
+        throws InternalInfrastructureException {
+      if (machines != null) {
+        if (!machines.containsKey(name)) {
+          throw new InternalInfrastructureException("Machine is not in a STARTING state");
```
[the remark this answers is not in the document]
sleshchenko: `RunningDockerMachine` everywhere looks not so good. And `StaringDockerMachine` is a stub that cannot be used outside of `DockerInternalRuntime`.
I propose to consider the following way:
1. Create `StartingDockerMachine` internal class of DockerInternalRuntime that will extend `Machine`.
2. Have two methods in `StartSynchronizer`: one for getting `Machine` (public representation) and another one for getting DockerMachine (internal representation).

Then we won't need a refactoring and having an interface for `DockerMachine`.
Let me know if I didn't explain good enough.
WDYT?

garagatyi:
@sleshchenko I reworked the PR in accordance with our discussion. Please, take a look

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerMachine.java
```diff
@@ -37,16 +37,15 @@
...
-
   /**
    * Default HOSTNAME that will be added in all docker containers that are started. This host will
    * container the Docker host's ip reachable inside the container.
    */
   public static final String CHE_HOST = "che-host";
+  /** Name of the latest tag used in Docker image. */
+  public static final String LATEST_TAG = "latest";
+
+  private static final Logger LOG = getLogger(DockerMachine.class);
```
sleshchenko: I'd revert this moving

garagatyi:
ci-test

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/RuntimeMachines.java
```diff
@@ -0,0 +1,107 @@
...
+   * state of machine with the real one that returns all the information about machine.
+   *
+   * @param name name of the machine to replace
+   * @param machine replacement
+   * @throws InternalInfrastructureException if runtime start is cancelled which led to removal of
+   *     machines from this storage
+   * @throws InternalInfrastructureException if machine is not present in this storage which means
+   *     that correct runtime start flow is broken
+   */
+  public synchronized void replaceMachine(String name, DockerMachine machine)
```
sleshchenko: method name says nothing about `STARTING` machines. Why do we need this method? 
Maybe it would be good enough to have only `putMachine(String name, DockerMachine machine)` instead of `addMachine` and `replaceMachine`.

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerMachine.java
garagatyi: +1

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/RuntimeMachines.java
garagatyi: I was thinking about preventing misuse of the method

garagatyi: Agree to simplify this

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerInternalRuntime.java
```diff
@@ -187,11 +192,17 @@ protected void internalStart(Map<String, String> startOptions) throws Infrastruc
...
         checkInterruption();
         String machineName = containerEntry.getKey();
-        final DockerContainerConfig config = containerEntry.getValue();
+
+        runtimeMachines.putMachine(machineName, new DockerMachine.StartingDockerMachine());
         sendStartingEvent(machineName);
+
         try {
-          startMachine(machineName, config);
+          DockerMachine machine = startMachine(machineName, containerEntry.getValue());
```
sleshchenko, line 200: Let's put running machine into `runtimeMachines` here instead of in `startMachine` method. I think it will make it more clear that starting machine will be updated.

sleshchenko (approved):
LGTM. 
Separated thanks for refactoring of StartSynchonizer.

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/RuntimeMachines.java
```diff
@@ -0,0 +1,83 @@
...
+
+  private Map<String, DockerMachine> machines;
+
+  public RuntimeMachines() {
+    this.machines = new HashMap<>();
+  }
+
+  /** Returns machines map. If runtime start is canceled returns empty map. */
+  public synchronized Map<String, DockerMachine> getMachines() {
+    return machines != null ? machines : emptyMap();
```
akorneta: As I understand here it is better to return the copy of original machines, because of guarantee that it won't be modified by `RuntimeMachines` and you can traverse on it(like [here](https://github.com/eclipse/che/pull/8210/files#diff-284cec01b9604144df613a5ffbbeeb7eL254)) without risking of concurrent modification.

skabashnyuk (approved):

akorneta (approved):
It's a cool that you decoupled the `DockerInternalRuntime`, Good Job.

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/RuntimeMachines.java
garagatyi: The copy is done [here](https://github.com/garagatyi/che/blob/647f982342945f2fa83cf41c03db7d4f04a75697/infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerInternalRuntime.java#L268-L268), whereas this component is kinda internal, so I would leave it as is, for now

garagatyi:
@akorneta you are right. I've fixed that

Issue #5652, opened 2017-07-14
Need to move OutputCustomizers to the according plugins and make them to be loaded according to workspace's settings


vrubezhny:
Currently all the existing Output Customizers are placed into `che-core-ide-app`. However it  doesn't appear to be a correct to tCore IDE App to hold the output customizers for the different runtimes...

We need to move OutputCustomizers to the according runtime plugins and make them to be loaded according to workspace's settings (or a list of runtimes used in the workspace projects).

gorkem:
@vrubezhny Is this still a valid issue?

vrubezhny:
@gorkem As we moved to use more or less native VSCode extensions (with no additions/changes)  I don't think so. 
Anyway, the current PR cannot be easily rebased and probably needs to be re-developed from the very beginning

PR #8235, opened 2018-01-10
Need to move OutputCustomizers to the according plugins and make them…

vrubezhny:
… to be loaded according to workspace's settings

Issue: #5652

### What does this PR do?

The enhancement provides the following changes:
 - 'Output Customizers' are replaced by 'Output Renderers' for better representation of their meaning.
 - The renderers are moved into according plug-ins according to their goals and to optimize the dependencies.
- Command editor is added with 'Renderers' page that allows a user to select which output renderers to use along with the given command. This allows to run/execute only the required renderers against the console output of the command.
- When a user creates a command (of some kind, like, for instance, the Java one) the according renderer will be selected by default (however, user may deselect it in case it's not needed to be used)
- JUnit tests are also reorganized by moving into according plug-ins along with their renderers.

Issue #5652

The 'Output Customizers' are turned to be 'Output Renderers' and moved into according plug-ins. Command editor is added with a possibility to specify the list of renderers to be invoked with a command.

#### Release Notes

che5652

This version introduces the changes in 'Output Customizers' and Command Editor. 'Customizers' turned into 'Renderers' (which better fits their meaning) and moved to according plug-ins to optimize their use and dependencies. Command editor is added with a page to select a list of renderers, so a user can select which of renderers are to be used when running a specified command.


riuvshin:
@vparfonov what is the state of this PR?

musienko-maxim:

sunix:
I think we should close this PR as this is not a priority for a while + moving to Theia

PR #8239, opened 2018-01-11
Changed back recreate update in che deployment config

sleshchenko:
### What does this PR do?
Fixes updating of Che application by deployment config.
The `rolling` update strategy doesn't work
- for the single-user Che on local OCP where `H2` database is used: 
*because* file database is occupied by previous Che pod;
- for both (single and multi-user) on OCP instances where nodes number is more than 1: 
*because* while we're using `ReadWriteOnce` PVC access mode new Che Pod can't be started because PVC is already mounted to old Che Pod.

-

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/dockerfiles/init/modules/openshift/files/scripts/che-openshift.yml
+++ b/dockerfiles/init/modules/openshift/files/scripts/che-openshift.yml
@@ -58,7 +58,9 @@ items:
     selector:
       app: che
     strategy:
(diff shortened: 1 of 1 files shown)


ghost (approved):

riuvshin (approved):
yabadabadoo

Issue #8186, opened 2018-01-05
Add a possibility to run functional tests on CI on the OpenShift infrastructure instead of Docker


garagatyi:
### Description
Now we have a possibility to start functional tests in Github PRs on Che with the Docker infra enabled. But we also need to test PRs that change or impact OpenShift infra. 

### Reproduction Steps

**OS and version:**    

**Diagnostics:**

riuvshin:
done, to launch tests against che multi user on ocp trigger phrase is: `ci-test-ocp`
ci job: https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/

PR #8255, opened 2018-01-11
Allow to override cli-image in ocp.sh

riuvshin:
Allow to override cli-image in ocp.sh
needed for: https://github.com/eclipse/che/issues/8186


dockerfiles/init/modules/openshift/files/scripts/ocp.sh
```diff
@@ -67,6 +67,9 @@ export CHE_IMAGE_REPO=${CHE_IMAGE_REPO:-${DEFAULT_CHE_IMAGE_REPO}}
 DEFAULT_IMAGE_INIT="eclipse/che-init"
 export IMAGE_INIT=${IMAGE_INIT:-${DEFAULT_IMAGE_INIT}}:${CHE_IMAGE_TAG}
 
+DEFAULT_CHE_CLI_IMAGE="eclipse/che-cli"
+export CHE_CLI_IMAGE=${CHE_CLI_IMAGE:-${DEFAULT_CHE_CLI_IMAGE}}:${CHE_IMAGE_TAG}
```
sleshchenko: Should `CHE_CLI_IMAGE_TAG` be here?

benoitf: also, is default image is not "eclipse/che" (-cli being like an alias that we might drop)

riuvshin: we use same tag for che init and cli by default it is nightlym if you want custom you have to build all 3 images

riuvshin: @benoitf I would prefer to use che-cli because it is more obvious, usually guys think that eclipse/che is che-server not cli :)

riuvshin: @sleshchenko ok, I;ve updated changes in such a way that allows to override INIT and CLI images with tags

benoitf (approved):

sleshchenko (approved):

Issue #7799, opened 2017-12-10
Multiuser Keycloak


JamesDrummond:
### Description
Keycloak not working when I use multiuser with OpenShift Origin.

### Reproduction Steps
mkdir ~/che-openshift-commands
cd ~/che-openshift-commands
  DEPLOY_SCRIPT_URL=https://raw.githubusercontent.com/eclipse/che/master/dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh 
WAIT_SCRIPT_URL=https://raw.githubusercontent.com/eclipse/che/master/dockerfiles/init/modules/openshift/files/scripts/wait_until_che_is_available.sh 
STACKS_SCRIPT_URL=https://raw.githubusercontent.com/eclipse/che/master/dockerfiles/init/modules/openshift/files/scripts/replace_stacks.sh 
curl -fsSL ${DEPLOY_SCRIPT_URL} -o ./get-che.sh 
curl -fsSL ${WAIT_SCRIPT_URL} -o ./wait-che.sh 
curl -fsSL ${STACKS_SCRIPT_URL} -o ./stacks-che.sh
oc login -u developer
export OPENSHIFT_TOKEN=$(oc whoami -t)
export EXTERNAL_IP=$(curl -s https://4.ifcfg.me/)
export MINISHIFT_IP=${EXTERNAL_IP}
CHE_MULTIUSER="true"
bash ./get-che.sh && bash ./wait-che.sh
bash ./stacks-che.sh

**OS and version:**    
Scaleway's Server using OpenShift Origin
Step by Step at https://medium.com/@james_devcomb/openshift-origin-on-vps-like-scaleways-2197294cf4fe

**Diagnostics:** 

## Logs from Keycloak Build:
```
warning: Image sha256:a03a21d7b825d1d95f9e54bdfed47c28b60744fa97a468a1d599db324da7e88e does not contain a value for the io.openshift.s2i.scripts-url label
--
  | WARNING: Error getting save-artifacts from : script "save-artifacts" not installed
  | Current directory: /opt/jboss
  | Directory contents of /tmp/src:
  | .
  | ..
  | .s2i
  | realms
  | themes
  | Directory contents of /tmp/src/themes:
  | total 0
  | drwxr-xr-x 3 jboss jboss 17 Dec 10 05:14 .
  | drwxr-xr-x 5 jboss jboss 46 Dec 10 05:14 ..
[...]

PR #8263, opened 2018-01-11
Openshift ocp fails due to livenessProbe.

JamesDrummond:
### What does this PR do?
The livenessProbe's failureThreshold is set to low. This causes the livenessProbe having to short a period of probing before it fails and removes the container. The failureThreshold is set to 4, the initialDelaySeconds is set to 5 and the periodSeconds is 5. This means 5s+4x5s=25s before the container fails. This patch set failureThreshold higher to 11 allowing for a more reasonable overall probe time of 60s (5s+11x5s=60s) before fails and removes the container.

https://github.com/eclipse/che/issues/7799

#### Release Notes
N/A

### DIFF
--- a/dockerfiles/init/modules/openshift/files/scripts/multi-user/keycloak/deployment-config.yaml
+++ b/dockerfiles/init/modules/openshift/files/scripts/multi-user/keycloak/deployment-config.yaml
@@ -39,7 +39,7 @@ spec:
         image: ' '
         name: keycloak
         livenessProbe:
-          failureThreshold: 4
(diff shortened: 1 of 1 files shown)


JamesDrummond:
Put high severity as keycloak on multiuser will fail unless using a pod with high CPU that can startup keycloak under 25 seconds.

sleshchenko (approved):
OK for me

riuvshin (approved):

benoitf (approved):

ghost (approved):

PR #8264, opened 2018-01-11
Recover from an error when old workspace OS resources already exist

davidfestal:
### What does this PR do?

This PR allows recovers from errors that arise when old workspace OpenShift resources already exist in the user namespace.

This is a fix that will make https://github.com/redhat-developer/rh-che/issues/507 a non-issue

### DIFF
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
@@ -567,6 +567,8 @@ public class OpenShiftConnector extends DockerConnector {
     return null;
   }
 
+  private ThreadLocal<Boolean> retryingAfterCleanup = ThreadLocal.withInitial(()->false);
+  
   /**
    * @param createContainerParams
    * @return
@@ -715,6 +717,11 @@ public class OpenShiftConnector extends DockerConnector {
       LOG.info(e.getMessage());
       imageIdToWorkspaceId.remove(imageForDocker);
       openShiftDeploymentCleaner.cleanDeploymentResources(deploymentName);
+      if (! retryingAfterCleanup.get()) {
+          LOG.info("Retry the creation of the workspace container after cleaning pending deployment resources");
+          retryingAfterCleanup.set(true);
+          return createContainer(createContainerParams);
+      }
       try {
         openShiftClient =
             ocFactory.newOcClient(
@@ -727,6 +734,7 @@ public class OpenShiftConnector extends DockerConnector {
       throw e;
     }
 
+    retryingAfterCleanup.set(false);
     return new ContainerCreated(containerID, null);
   }
 
--- a/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
+++ b/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
@@ -567,8 +567,8 @@ public class OpenShiftConnector extends DockerConnector {
     return null;
   }
 
-  private ThreadLocal<Boolean> retryingAfterCleanup = ThreadLocal.withInitial(()->false);
-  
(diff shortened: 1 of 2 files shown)


plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -715,6 +717,11 @@ public ContainerCreated createContainer(CreateContainerParams createContainerPar
       LOG.info(e.getMessage());
       imageIdToWorkspaceId.remove(imageForDocker);
       openShiftDeploymentCleaner.cleanDeploymentResources(deploymentName);
+      if (! retryingAfterCleanup.get()) {
```
ibuziuk: @davidfestal could you please explain why you initially set `retryingAfterCleanup` = false, and than checks it as negative `!retryingAfterCleanup.get()`?  maybe I missing smth. but wouldn't it be more straight-forward to set it initially to true, and check without "!'

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
```diff
@@ -567,6 +567,8 @@ private Subject subjectForImageId(String imageId) {
     return null;
   }
 
+  private ThreadLocal<Boolean> retryingAfterCleanup = ThreadLocal.withInitial(()->false);
```
ibuziuk: maybe, it would make sense to have this setting defined in `che.properties`, so that it would be possible to chnage this behaviour on osio by simply updating cm ?

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
davidfestal: Then I would have to name the TLS `canRetryOnce` :-) I can do it sure.

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
davidfestal: why would we need to change the behaviour ?

ibuziuk: I do not insist, if was more like thinking outloud ;-) If does not make sense let's keep it as non-tunable option

ibuziuk (approved):

plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java
ibuziuk: approving, just thinking that probably better naming for `retryingAfterCleanup` can be done

l0rd:
From what I see in this PR code I understand that you are introducing the following logic: 

if `createContainer()` fails (for any reason, not just "when old workspace OpenShift resources already exist") you call `createContainer()` again recursively. And to avoid that the recursive call is made more than once you use a sort of counter (`retryingAfterCleanup`). 

Is it correct? I am asking because I think that's not easy to understand reading the code and I would like to propose you some changes to make it easier for "readers" to understand this logic.

davidfestal:
@l0rd that's exactly what it does. I'm currently about to test slight naming changes (use `canRecurseInCreateContainer` initialized to `true` at start) as well as a filter on the excepiton error message containing "Already exists".

davidfestal:
@l0rd @ibuziuk do you think my last commit makes it easier to understand ?

davidfestal:
@l0rd is it OK for you like that ?

riuvshin:
@davidfestal @l0rd is it still actual?

davidfestal:
Well, we probably won't work on this anymore, since it was a Che5 OpenshiftConnector.

l0rd:
@riuvshin @davidfestal yes let's close this PR

Issue #8326, opened 2018-01-17
ocp.sh need to fallback to xip.io or some other if nip.io is down


riuvshin:
nip.io is not quite stable, so we have to check if it online before we deploy che to openshift and if not we have to use alternatives

PR #8332, opened 2018-01-17
add test for DNS providers in ocp.sh and use backup if default doesn't work

riuvshin:
nip.io is down again... and we have failed tests on CI due to that.
So this PR adds the check for DNS provider and if default `nip.io` will fail test,  backup DNS provider will be used.

needed for: https://github.com/eclipse/che/issues/8326

### DIFF
--- a/dockerfiles/init/modules/openshift/files/scripts/ocp.sh
+++ b/dockerfiles/init/modules/openshift/files/scripts/ocp.sh
@@ -40,7 +40,7 @@ export OPENSHIFT_USERNAME=${OPENSHIFT_USERNAME:-${DEFAULT_OPENSHIFT_USERNAME}}
 DEFAULT_OPENSHIFT_PASSWORD="developer"
 export OPENSHIFT_PASSWORD=${OPENSHIFT_PASSWORD:-${DEFAULT_OPENSHIFT_PASSWORD}}
 
-BACKUP_DNS_PROVIDERS=(
+DNS_PROVIDERS=(
 xip.io
 nip.codenvy-stg.com
 )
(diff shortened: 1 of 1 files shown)


dockerfiles/init/modules/openshift/files/scripts/ocp.sh
```diff
@@ -77,6 +81,25 @@ export CONFIG_DIR=${CONFIG_DIR:-${DEFAULT_CONFIG_DIR}}
 
 }
 
+test_dns_provider() {
+    #Test  DNS_PROVIDER
+    if [[ $(dig +short 10.0.0.1.$DNS_PROVIDER) = "10.0.0.1" ]]; then
+        echo "$DNS_PROVIDER works - OK"
+    else
+        echo "$DNS_PROVIDER seems is not working, trying to use BACKUP_DNS_PROVIDERS"
```
garagatyi: Should `BACKUP_DNS_PROVIDERS` be a variable expansion?

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
```diff
@@ -77,6 +81,25 @@ export CONFIG_DIR=${CONFIG_DIR:-${DEFAULT_CONFIG_DIR}}
...
+        for i in ${BACKUP_DNS_PROVIDERS[@]}
```
garagatyi: I would start with this code and omit first if clause at all

garagatyi (approved):
Looks OK, but take a look at my comments - there is a space to make PR even better!

riuvshin:
@garagatyi thanks for your feedback, PR updated.

sleshchenko (approved):

dockerfiles/init/modules/openshift/files/scripts/ocp.sh
```diff
@@ -77,6 +81,21 @@ export CONFIG_DIR=${CONFIG_DIR:-${DEFAULT_CONFIG_DIR}}
 
 }
 
+test_dns_provider() {
+    #add current $DNS_PROVIDER to the providers list to respect environment settings
+    DNS_PROVIDERS=("$DNS_PROVIDER" "${DNS_PROVIDERS[@]}")
+    for i in ${DNS_PROVIDERS[@]}
+        do
+        if [[ $(dig +short 10.0.0.1.$i) = "10.0.0.1" ]]; then
```
benoitf: should we use dig or only "host" command ?
also should it have some timeout ? if one of DNS provider is not replying we might have delays

riuvshin: @benoitf
1)  I feel more comfortable with `dig` host output should be parsed to get the IP while dig allows to get only IP
2) The reason why I didn't add timeout is because `dig` has own very short timeout.
real host resolve time:
```
time dig +short 10.0.0.1.nip.io
10.0.0.1

real	0m0.010s
user	0m0.002s
sys	0m0.003s
```

fake host resolve time:
```
time dig +short thisisfakehostnamepiupiu.com

real	0m0.055s
user	0m0.003s
sys	0m0.004s
```

benoitf: ok for dig.

By timeout, I was talking about when DNS provider is not responding/timeouting, not with an invalid domain name as in that case there is a DNS that reply that it is not existing.

garagatyi: +1 to what @benoitf said. In your example DNS is resolved fast but IP is not responding whereas the problem might be in DNS

riuvshin:
@garagatyi @benoitf added timeout

benoitf:
@riuvshin you don't want to use dig parameters like 
```
+time=T
Sets the timeout for a query to T seconds. The default time out is 5 seconds. An attempt to set T to less than 1 will result in a query timeout of 1 second being applied.

+tries=T
Sets the number of times to retry UDP queries to server to T instead of the default, 3. If T is less than or equal to zero, the number of retries is silently rounded up to 1.
```

riuvshin:
@benoitf that is exactly what I've said that dig has a default timeout (my examples with unknown dns didn't showed that timeout) that is why initially no timeout was specified, maybe I misunderstand but I thought you guys want something external to make sure that it will timeout no matter what

should I rever changes to original state?

benoitf:
@riuvshin my question was only about the "default timeout" when dig ask for a DNS server that is timeouting on reply

riuvshin:
@benoitf specified timeout to 5s(default) and try count to `1` (3 is default)

benoitf (approved):

Issue #8305, opened 2018-01-16
Add ability to use more than one test user in the selenium tests


Ohrimenko1988:
We need to add ability to use more than one test user in the selenium tests. It needs for usecases in the "organizations" tests package, and in "AccountTest" for safe updating and deleting user in the test suite.

PR #8342, opened 2018-01-18
Add ability to add several tests users to selenium tests

dmytro-ndp:
### What does this PR do?
1) Removes redundant legacy code to inject test users by email;
2) Fixes issue of injecting test user which has name which differs from email prefix;
3) Adds second test user which credentials is passed through the environment variables: _CHE_SECOND_TESTUSER_NAME_, _CHE_SECOND_TESTUSER_EMAIL_, _CHE_SECOND_TESTUSER_PASSWORD_;
Example of usage of _second test user_:
```
@Inject CheSecondTestUser secondTestUser;
```
4) Removes creation and removal of test user from the **TestUserImpl** class;
5) Makes it possible to use **TestUserServiceClient** class for non-default user.

#8305


selenium/che-selenium-core/bin/webdriver.sh
```diff
@@ -886,36 +889,81 @@ prepareTestUserForMultiuserChe() {
...
-           echo -e "${RED}[WARN] There is a problem with creation of test user in Keycloak server: '${response}'.${NO_COLOUR}"
-           echo -e "Admin user will be used as a test user."
+           echo -e "${RED}[WARN] There is a problem with creation of default test user in Keycloak server: '${response}'.${NO_COLOUR}"
+           echo -e "Admin user will be used as default test user."
            CHE_TESTUSER_NAME=${CHE_ADMIN_NAME}
            CHE_TESTUSER_EMAIL=${CHE_ADMIN_EMAIL}
            CHE_TESTUSER_PASSWORD=${CHE_ADMIN_PASSWORD}
         fi
+        
+        # create second test user
```
tolusha, line 898: It is possible to use the same function to create both users?

dmytro-ndp: yes, it is possible

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/user/TestUserImpl.java
```diff
@@ -138,13 +106,6 @@ public void delete() {
             format("User name='%s' failed to remove workspace name='%s'", workspace, name), e);
       }
     }
-
-    try {
-      userServiceClient.remove(id);
```
tolusha, line 143: Is there a good reason to remove this?

dmytro-ndp: Che 6 doesn't allow to remove user by using user service so as user management is under control of Keycloak.

dmytro-ndp: Selenium tests are creating and removing test users by launcher webdriver.sh.

selenium/che-selenium-core/bin/webdriver.sh
dmytro-ndp: see fixup in new commit https://github.com/eclipse/che/pull/8342/commits/0eb82c942060c7c2d5863d0daf9559514a4cae94

tolusha (approved):

PR #8366, opened 2018-01-19
Fix kc deploy on openshift

### LINKED ISSUE #8320: fix che multi user deployment on ocp
today we have deployment issue while we deploy che multiuser to ocp, due to this issue we have first boot of KC pod failed and treat it as "normal behavior"  which not good.
This issue happen because to start KC pod it need CHE pod route while CHE pod also depends on KC.

riuvshin:
Makes deploy of CHE multi-user to openshift linear and stable. Before these changes we had situation where KC pod was always failing for the first time due to circular dependency between CHE and KC pods.

fixes: https://github.com/eclipse/che/issues/8320


sleshchenko (approved):
Good job!

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
```diff
@@ -429,10 +471,6 @@ echo "[CHE] Deploying Che on ${OPENSHIFT_FLAVOR} (image ${CHE_IMAGE})"
 get_che_pod_config | oc apply --force=true -f -
 echo
 
-if [ "${CHE_DEDICATED_KEYCLOAK}" == "true" ]; then
-  "${COMMAND_DIR}"/multi-user/configure_and_start_keycloak.sh
```
garagatyi, line 433: Is it possible not to run keycloak now?

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
```diff
@@ -312,6 +311,49 @@ else
     exit 1
 fi
 
+# -------------------------------------------------------------
+# create CHE service and route
+# -------------------------------------------------------------
+echo "[CHE] Creating serviceaccount, service and route for CHE pod"
+ echo "apiVersion: v1
```
benoitf, line 318: it looks strange to have yaml content in a shell script and not in its own yml file

riuvshin: we have a lot of that already, but yeah a I completely agree with that. Some times it is need to be able to use ENV vars for that yaml

benoitf: but there are some sed commands for that no ?

riuvshin: no, no sed at all.
https://github.com/eclipse/che/blob/master/dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh#L391-L400
https://github.com/eclipse/che/blob/master/dockerfiles/init/modules/openshift/files/scripts/multi-user/deploy_postgres_and_keycloak.sh#L25-L38
https://github.com/eclipse/che/blob/master/dockerfiles/init/modules/openshift/files/scripts/multi-user/deploy_postgres_only.sh

we have a lot of that, I'm not saying that this is good, but nothing new.

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
riuvshin: not sure I got the question, that section that I;ve removed was used to run KC after it failed first time now we don;t need that at all as it will start up normally with first attempt

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
```diff
@@ -245,8 +245,7 @@ CHE_CONFIG_FILE_PATH=${CHE_CONFIG_FILE_PATH:-${DEFAULT_CHE_CONFIG_FILE_PATH}}
 cat "${CHE_DEPLOYMENT_FILE_PATH}" | \
     sed "s/          image:.*/          image: \"${CHE_IMAGE_SANITIZED}\"/" | \
     sed "s/          imagePullPolicy:.*/          imagePullPolicy: \"${IMAGE_PULL_POLICY}\"/" | \
-    inject_che_config "#CHE_MASTER_CONFIG" "${CHE_CONFIG_FILE_PATH}" | \
-    if [ "${ENABLE_SSL}" == "false" ]; then grep -v -e "tls:" -e "insecureEdgeTerminationPolicy: Redirect" -e "termination: edge" ; else cat -; fi #| \
+    inject_che_config "#CHE_MASTER_CONFIG" "${CHE_CONFIG_FILE_PATH}"
```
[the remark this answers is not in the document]
riuvshin: sorry, but I don;t see that to align here...

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
riuvshin: so yes it is possible not to run KC here because KC is already started at this moment, and even CHE deployed. that part was needed to trigger KC run after it failed for the first time and CHE is already deployed but now KC is already fully started before CHE deploy

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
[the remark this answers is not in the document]
garagatyi: I mean the function you are changing. It is not aligned properly.

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
garagatyi: I think that the true way is to put everything into one deployment config, but since we are not there I'm OK to do this approach.
@riuvshin please, consider reworking everything to deployment config file(s) in the further improvements.

garagatyi (approved):

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
riuvshin: yeah, sigle DC that is definitely a proper way, but yes we are far from there and we will go that way.

riuvshin: @benoitf WDYT?

benoitf (approved):

PR #8401, opened 2018-01-22
Fix servers probes on https powered deployment

### LINKED ISSUE #8378: Server liveness probes don't work when server is exposed as https URL without port
### Description
We have a bug when a port is not set in a server URL and https is used. We should use 443 instead of 80 in this case as a port. Or we can rework servers to use null as a port in such cases to be able to use default port when building URI.

### Reproduction Steps

**OS and version:**    

**Diagnostics:**

garagatyi:
### What does this PR do?
Fixes possibility to use server probes on an OpenShift deployment where https is enabled for routes.
Also, disallow server probes with port 0.

Fixes #8378

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/probe/server/ExecServerLivenessProbeConfigFactory.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/probe/server/ExecServerLivenessProbeConfigFactory.java
@@ -29,7 +29,7 @@ public class ExecServerLivenessProbeConfigFactory implements HttpProbeConfigFact
     try {
       URL url = new URL(server.getUrl());
       return new HttpProbeConfig(
-          url.getPort() == -1 ? 80 : url.getPort(),
+          url.getPort() == -1 ? url.getDefaultPort() : url.getPort(),
           url.getHost(),
           url.getProtocol(),
           "/liveness",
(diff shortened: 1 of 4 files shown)


garagatyi:
ci-build

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/probe/server/ExecServerLivenessProbeConfigFactory.java
```diff
@@ -29,7 +29,7 @@ public HttpProbeConfig get(String workspaceId, Server server)
     try {
       URL url = new URL(server.getUrl());
       return new HttpProbeConfig(
-          url.getPort() == -1 ? 80 : url.getPort(),
+          url.getPort() == -1 ? url.getDefaultPort() : url.getPort(),
```
sleshchenko, line 32: Maybe it would be better to check `default port` here and throw `InternalInfrastructureException`. Otherwise, as far as I understand `IllegalArgumentException` may be thrown later.

sleshchenko (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/probe/server/ExecServerLivenessProbeConfigFactory.java
garagatyi: getDefaultPort will return 80 or 443 in case of https/http. And since we know that this server is http(s) it is OK I believe.

sleshchenko: Then it's OK to me.

akorneta (approved):

garagatyi:
ci-build

PR #8406, opened 2018-01-22
Updated policy of installers update

sleshchenko:
### What does this PR do?
Updates policy of installers update.
Now existing Che installers ([exec](https://github.com/eclipse/che/blob/master/agents/exec/installer/src/main/resources/installers/1.0.1/org.eclipse.che.exec.script.sh), [wsagent](https://github.com/eclipse/che/blob/master/wsagent/agent/src/main/resources/installers/1.0.0/org.eclipse.che.ws-agent.script.sh), etc.) must not be modified because they won't be applied to existing instances. Instead of updating new version must be added and them all workspace will pick it up if they use `latest` version.

With these changes, existing installers can be modified and they will be picked up by existing Che instances when it will be upgraded.
So, InstallerRegistry will do the following action when tomcat will be started:
<pre>
Check if installer from resources is already present in database
| +  if yes, check if installer in database and in resources are the same:
      | + if yes, log a message that latest version of an installer is already present.
      | - if no, update an installer in a database and log a message that an installer is updated.
| -  if no, insert an installer in a database and log a message that an installer is created.
</pre>

### Why it is important?
Now there is no a centralized ability to update installers that are already initialized. But IMHO it is so useful in the following cases:
- when it is needed to fix API incompatibility in an existing installer;
- when it is needed to fix minor bug or typo in an existing installer;

Now in the described cases, we have to create new installer version even when old one becomes broken.
And Dashboard will propose a user to installer broken installer, see example:
screenshot_20180122_175332
The actual difference between these two installers is - new installer is able to use environment variable with a free port for its server while old one always uses default port and will fail if any ports conflicts occur.
For End Users, it's not clear enough what is really a difference between installers and which one they should use. Actually, there is no sense to use an older version.

One of the use cases when It makes sense to have different installers version if they really provide different agents or install different software.
screenshot_20180122_180046

As additional benefit developer life become easier:
- because now PR doesn't show really changes, because always new file is added;
- if a developer wants to check his changes - he doesn't have an ability to build new binaries and check it on an existing instance because his changes won't be applied. He has to test it on new instance after destroying existing one or update installer manually via swagger (it is not so easy while original installer script should be processed each time to be JSON String compatible, like escaping all new lines and `"` characters).

#### Release Notes
Updated policy of installers update.

#### Docs PR
N/A

### DIFF
--- a/wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/impl/LocalInstallerRegistry.java
+++ b/wsmaster/che-core-api-installer/src/main/java/org/eclipse/che/api/installer/server/impl/LocalInstallerRegistry.java
@@ -24,7 +24,6 @@ import java.util.Set;
 import org.apache.maven.artifact.versioning.ComparableVersion;
 import org.eclipse.che.api.core.Page;
 import org.eclipse.che.api.installer.server.InstallerRegistry;
-import org.eclipse.che.api.installer.server.exception.InstallerAlreadyExistsException;
 import org.eclipse.che.api.installer.server.exception.InstallerException;
 import org.eclipse.che.api.installer.server.exception.InstallerNotFoundException;
 import org.eclipse.che.api.installer.server.model.impl.InstallerImpl;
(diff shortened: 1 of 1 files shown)


tolusha (approved):

sleshchenko:
Dear Reviewers, please review my PR and share your opinion about it.

garagatyi (approved):
Looks good!

akorneta (approved):
:+1:

skabashnyuk (approved):

gazarenkov (approved):

PR #8422, opened 2018-01-23
Rework default memory limit setting

### LINKED ISSUE #8377: Extract default machine ram limit from Internal Environment Factory
### Description
We need to get rid of the [defaultMachineMemorySizeMB](https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/environment/InternalEnvironmentFactory.java#L59) parameter because not all the implementation of infrastructure may need this and actually not all of them may support the limits that is defined in Che, so the proposed way do to it:
- Make infrastructure responsible for the setting of limits;
- Make attribute [memoryLimitBytes](https://github.com/eclipse/che/blob/master/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/config/MachineConfig.java#L28) optional;
- Adapt resource API components to the optional value of the attribute.

garagatyi:
### What does this PR do?
Extract default machine memory limit setting from InternalEnvironmentFactory to recipe specific environment factories.
Make memory limit attribute optional by respecting it by resource API subsystem.

Fixes #8377

### DIFF
--- a/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/config/MachineConfig.java
+++ b/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/config/MachineConfig.java
@@ -21,9 +21,9 @@ import java.util.Map;
 public interface MachineConfig {
 
   /**
-   * Name of the attribute from {@link #getAttributes()} which if present sets memory limit of the
-   * machine in bytes. If memory limit is set in environment specific recipe this attribute should
-   * override value from recipe.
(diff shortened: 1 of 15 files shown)


garagatyi:
ci-test

akorneta (approved):

sleshchenko (approved):
LGTM.
One suggestion: please consider adding mention of memory limit in `InternalEnvironmentFactory#doCreate`

skabashnyuk (approved):

PR #8453, opened 2018-01-25
Change PVC name provisioning to prefix + generating 8 chars

### LINKED ISSUE #8439: Create workspace failed on OpenShift: Name for persistent volume too long
I have successfully deployed Che 6.0.0-M5-SNAPSHOT on Red Hat OpenShift Container Platform 3.7, but creating a workspace always fails in my setup.

Root cause is that the persistent volume for the storage of the dev machine can not be provisioned. I am using GlusterFS (Container Native Storage) with dynamic provisioning as storage backend for OpenShift. I can see the following event in OpenShift:

> _Failed to provision volume with StorageClass "gluster-container": glusterfs: create volume err: failed to create endpoint/service error creating service: Service "glusterfs-dynamic-claim-che-workspace-workspaceb2jggx3x4a0138v8-projects" is invalid: metadata.name: Invalid value: "glusterfs-dynamic-claim-che-workspace-workspaceb2jggx3x4a0138v8-projects": must be no more than 63 characters_

Unfortunately the prefix "glusterfs-dynamic-" is always added in this kind of setup which leads to 73characters total (which is too long for kubernetes).

Is there a way to modify the template for deploying a new workspace, to make the persistent volume name shorter (e.g. drop "claim-che-workspace" or "-workspace" )? Workspace ID plus "-projects" might be enough for a unique name?

akorneta:
### What does this PR do?
Some of the OpenShift storage backends can use prefixes for objects names but the name length limit is equal to 63 characters, in such cases, the composite name(`prefix-workspaceId-volumeName`) that is used for `unique` PVC strategy does not fit the limitation. 
We proposing to change the naming to: `prefix-generated_chars_length_8` and put the volume name into PVC labels.

fix #8439

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
@@ -338,7 +338,7 @@ che.infra.openshift.pvc.enabled=true
 #        Existing PVC will be used or new one will be created if it doesn't exist.
 #
 # - 'unique'
-#        PVC with name that is evaluated as '{che.infra.openshift.pvc.name} + '-' + {WORKSPACE_ID}'
+#        PVC with name that is evaluated as '{che.infra.openshift.pvc.name} + '-' + {generated_8_chars}'
 #        will be used for storing of workspaces data.
 #        Existing PVC will be used or a new one will be created if it doesn't exist.
 che.infra.openshift.pvc.strategy=common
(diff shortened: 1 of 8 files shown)


skabashnyuk (approved):

akorneta:
ci-build

akorneta:
ci-test-ocp

riuvshin:
ci-test-ocp

garagatyi (approved):

sleshchenko (approved):

PR #8474, opened 2018-01-26
init and cli images should not use tag from CHE

riuvshin:
init and cli images should not use tag from CHE, that is needed if user wants to set only custom CHE and not build init and CLI images.


garagatyi (approved):

sleshchenko (changes requested):
Is `CHE_IMAGE_TAG` used somewhere?
https://github.com/eclipse/che/blob/d6bd53abc21499e6919add8ee8d0c84075a9908a/dockerfiles/init/modules/openshift/files/scripts/ocp.sh#L69-L70

Please fix help info. Now it is not clear that `CHE_IMAGE_TAG`  will not be used for `che-cli` and `che-init` images https://github.com/eclipse/che/blob/d6bd53abc21499e6919add8ee8d0c84075a9908a/dockerfiles/init/modules/openshift/files/scripts/ocp.sh#L281

riuvshin:
@sleshchenko thanks for your review.
- `CHE_IMAGE_TAG` is used in deploy scrip.
- help fixed

sleshchenko (approved):
LGTM

Issue #6829, opened 2017-10-19
Don't remove non-test workspaces at the cleanup environment stage of selenium testing


dmytro-ndp:
### Description
In selenium tests we remove all workspace of test user at the final clean up stage [before removing test user](https://github.com/eclipse/che/blob/5.19.0/selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/user/TestUserImpl.java#L133-L140). It was done so to remove test workspaces created during the tests from factory or in similar way. This removal can affect non-test workspaces of single user of Che which is being tested.
So as we had added proper tear down methods to remove factory workspaces, we could avoid removal of all workspaces before removal of test users. 
At the same time, test workspaces from the workspace pool is removed by the [TestWorkspaceProviderImpl](https://github.com/eclipse/che/blob/5.19.0/selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/workspace/TestWorkspaceProviderImpl.java#L145) class.

Also, we need to find the way to remove test workspaces in case of unexpected interrupting of test execution by Ctrl-C key stroke.

dmytro-ndp:
There are test workspaces which tests don't remove after execution:
- test workspace linked to field [testWorkspace](https://github.com/eclipse/che/blob/6.0.0-M4/selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/gwt/CheckSimpleGwtAppTest.java#L58) which CheckSimpleGwtAppTest test substitutes in **prepare()** method and it courses initial workspace remains exists after test execution;
- test workspace linked to some test which being executed in the _CheOneThreadTestsSuite.xml_ suite when _CheSuite.xml_ suite starts execution, removes that test from [SeleniumTestHandler::runningTests](https://github.com/eclipse/che/blob/6.0.0-M4/selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/inject/SeleniumTestHandler.java#L173) list and so prevents removal of test workspace which belongs to that test when [SeleniumTestHandler::shutdown()](https://github.com/eclipse/che/blob/6.0.0-M4/selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/inject/SeleniumTestHandler.java#L418) method is executed
- test workspaces which startup failed, and they weren't removed after that because of removal were hidden by [Assert.fail()](https://github.com/eclipse/che/commit/2c299254768e49f6082106b7331dc67a25831e28) command.

PR #8481, opened 2018-01-26
Avoid removal of non-test workspaces at the cleanup environment stage of selenium testing

dmytro-ndp:
### What does this PR do?
This PR get rid of code which removes all user's worskspaces after the all test executions. 
Also it restores removal of test workspaces in the next cases:
- removal of test workspace linked to field [CheckSimpleGwtAppTest::testWorkspace](https://github.com/eclipse/che/blob/6.0.0-M4/selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/gwt/CheckSimpleGwtAppTest.java#L58);
- removal of test workspace linked to some test which being executed in the _CheOneThreadTestsSuite.xml_ suite when _CheSuite.xml_ suite starts execution, removes that test from [SeleniumTestHandler::runningTests](https://github.com/eclipse/che/blob/6.0.0-M4/selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/inject/SeleniumTestHandler.java#L173) list and so prevents removal of test workspace which belongs to that test when [SeleniumTestHandler::shutdown()](https://github.com/eclipse/che/blob/6.0.0-M4/selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/inject/SeleniumTestHandler.java#L418) method is executed;
- removal of test workspaces which startup failed, and they weren't removed after that because of removal were hidden by [Assert.fail()](https://github.com/eclipse/che/commit/2c299254768e49f6082106b7331dc67a25831e28) command.

#6829

### DIFF
--- a/selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/inject/SeleniumTestHandler.java
+++ b/selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/inject/SeleniumTestHandler.java
@@ -82,6 +82,7 @@ public abstract class SeleniumTestHandler
 
   private static final Logger LOG = LoggerFactory.getLogger(SeleniumTestHandler.class);
   private static final AtomicBoolean isCleanUpCompleted = new AtomicBoolean();
+  private static final AtomicBoolean isWebDriverSessionCreationChecked = new AtomicBoolean();
 
   @Inject
   @Named("tests.screenshot_dir")
@@ -170,18 +171,23 @@ public abstract class SeleniumTestHandler
 
   @Override
   public void onStart(ISuite suite) {
-    runningTests.clear();
     suite.setParentInjector(injector);
   }
 
   /** Check if webdriver session can be created without errors. */
   private void checkWebDriverSessionCreation() {
+    if (isWebDriverSessionCreationChecked.get()) {
+      return;
+    }
+
     SeleniumWebDriver seleniumWebDriver = null;
     try {
       seleniumWebDriver = new SeleniumWebDriver(browser, webDriverPort, gridMode, webDriverVersion);
     } finally {
       Optional.ofNullable(seleniumWebDriver)
           .ifPresent(SeleniumWebDriver::quit); // finish webdriver session
+
+      isWebDriverSessionCreationChecked.set(true);
     }
   }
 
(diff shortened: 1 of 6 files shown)


Ohrimenko1988 (approved):

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/inject/SeleniumTestHandler.java
```diff
@@ -170,18 +171,23 @@ public void onFinish(ITestContext context) {}
 
   @Override
   public void onStart(ISuite suite) {
-    runningTests.clear();
```
tolusha, line 173: I guess it was added for a good reason. 
I would like to keep it here unless it brings issue.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/inject/SeleniumTestHandler.java
```diff
@@ -82,6 +82,7 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(SeleniumTestHandler.class);
   private static final AtomicBoolean isCleanUpCompleted = new AtomicBoolean();
+  private static final AtomicBoolean isWebDriverSessionCreationChecked = new AtomicBoolean();
```
tolusha: If we invoke `checkWebDriverSessionCreation` in constructor then we won't needed this flag.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/inject/SeleniumTestHandler.java
dmytro-ndp: I was added it before to solve the problem which had gone since that time and now it causes another problem: missing clean up step for some test which being executed in the **CheOneThreadTestsSuite.xml** suite bacause when **CheSuite.xml** suite in turn starts execution, this command **runningTests.clear()** removes that test from **SeleniumTestHandler::runningTests** list and so prevents removal of test workspace which belongs to that test when **SeleniumTestHandler::shutdown()** method is executed;

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/inject/SeleniumTestHandler.java
dmytro-ndp: Good catch. Thank you!

tolusha (approved):

Issue #7807, opened 2017-12-11
Add an ability to configure logs dir for agents


sleshchenko:
### Description
Che 5 on OSIO have an ability to configure logs dir for agents (like workspace agent and exec agent). 
But agents in Che 6 hold own logs in their default logs folders and there is no ability to configure that.
So it is needed to have an ability to configure logs dir for agents https://github.com/redhat-developer/rh-che/pull/461#issuecomment-347820591

garagatyi:
The idea I'm going to implement is following:
1. Che master adds an env var to machines which says base dir for all logs of any software interested in persisting logs between workspace restarts. Let's call it CHE_LOGS_BASE_DIR
1. Che may mount this folder somewhere to persist its content between workspace restarts
1. Any agent or installer that is running inside of a workspace machine can read this env variable and store its logs there. It is better if each interested software creates its own subfolder of folder provided by env var and stores logs files in this subfolder.
1. If folder specified by that env variable doesn't exist or agent, for example,  doesn't have rights to write logs there it can fail or write logs into another directory (should we choose single strategy here?)
1. If environment variable is missing or empty then agent, for example, should use another folder for storing logs

@sleshchenko @skabashnyuk @l0rd Do you think guys this strategy is OK?

garagatyi:
One thing suggested by @sleshchenko is that I'm not going to add an obligation for the infrastructure implementation to persist those logs. But infrastructure may do that if it will.

l0rd:
@garagatyi that looks of for me. We should probably also define a default value for `CHE_LOGS_BASE_DIR` that would be used in case the var is empty or missing (point 5.). For point 4 the simpler the better. 

Ideally agents should log on stdout and the [infrastructure should take care of the logs delivery/backup](https://docs.docker.com/engine/admin/logging/overview/). But that's not possible now because we run multiple agents per container. As a workaround we need to log to a volume.

Issue #7386, opened 2017-11-15
CHE-6: Persist workspace logs


ghost:
Currently, workspace logs are not saved anywhere which makes debugging and investigating issues quite problematic.

skabashnyuk:
Your proposals?

ghost:
@skabashnyuk If a common strategy is used, create a subpath with mkdir pod and mount ~/che/ws-agent/logs. So, there will be /projects and /logs in this directory.

In case of a unique strategy just mount an additional path.

@akorneta may want to comment too :)

skabashnyuk:
Will be fixed by https://github.com/eclipse/che/issues/7807

akorneta:
for the OpenShift flavours, the log backup was done in the scope of this PR: https://github.com/eclipse/che/pull/8535

eivantsov:
Closing as implemented. WS logs are in workspace PV.

PR #8535, opened 2018-01-31
Provide machine logs volume and configuration of logs root path

akorneta:
### What does this PR do?
Adds configuration of workspace logs root directory;
Provides environment variable for each OpenShift machine that contains value of logs root directory;
Adds logs volume into all OpenShift machines and make distinct subpaths for logs with machine name.
Modifies installation scripts of ws-agent and exec-agent with a logic that defines the logs folder for this agents and redirects bootstrapper logs to workspace logs root.

file structure on host for different strategies:

One PV will be used for storing logs of machines that are placed in the same pod.
**unique:** 
unique_strategy

One PV will be used for storing logs and projects data of workspace with two machines.
**common:**
common_strategy

#7807 
partially solves this issue(for OpenShift only): #7386 

#### Docs PR
n/a

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
@@ -28,6 +28,11 @@ che.workspace.storage=${che.home}/workspaces
 # workspace. This is the directory in the machine where your projects are placed.
 che.workspace.projects.storage=/projects
 
+#Defines the directory inside the machine where all the workspace logs are placed.
+#The value of this folder should be provided into machine e.g. like environment variable
+#so agents developers can use this directory for backup agents logs.
+che.workspace.logs.root_dir=/workspace_logs
+
 # Configures proxies used by runtimes powering workspaces
 che.workspace.http_proxy=
 che.workspace.https_proxy=
--- a/dockerfiles/init/manifests/che.env
+++ b/dockerfiles/init/manifests/che.env
@@ -122,6 +122,11 @@
 #     agents within the workspace when it boots, then the workspace will not be started.
 #CHE_WORKSPACE_AGENT_DEV_MAX__START__TIME__MS=300000
 
+# Workspace logs root
+#     Defines the directory inside the machine where all the workspace logs are placed.
+#     The value of this folder should be provided into machine e.g. like environment variable
+#     so agents developers can use this directory for backup agents logs.
+#CHE_WORKSPACE_LOGS_ROOT__DIR=/workspace_logs
 
 
 # Workspace Java Options
--- a/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/provision/env/LogsRootEnvVariableProvider.java
+++ b/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/provision/env/LogsRootEnvVariableProvider.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2012-2018 Red Hat, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.workspace.infrastructure.openshift.provision.env;
+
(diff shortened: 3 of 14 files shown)


akorneta:
ci-build

akorneta:
ci-test-ocp

riuvshin:
will we have ws logs on docker?

akorneta:
@riuvshin maybe, but not in the scope of this pr

riuvshin:
@akorneta sure, but we need ws logs on docker same as on openshift, so that should be considered I think in order to close ws logs issue

akorneta:
@riuvshin I understand that. 
This PR will resolve this issue: https://github.com/eclipse/che/issues/7807
and for OpenShift only, this one: https://github.com/eclipse/che/issues/7386 (updated the pr summary)

dmytro-ndp:
@akorneta: isn't #7386 for logs of ws run on docker?

akorneta:
@dmytro-ndp no(see comment above), I'll not close that issue, just will comment that for OS infra it works.

dmytro-ndp:
@akorneta: I see, thank you.

dockerfiles/init/manifests/che.env
```diff
@@ -122,6 +122,11 @@
 #     agents within the workspace when it boots, then the workspace will not be started.
 #CHE_WORKSPACE_AGENT_DEV_MAX__START__TIME__MS=300000
 
+# Workspace logs root
+#     Defines the directory inside the machine where all the workspace logs are placed.
+#     The value of this folder should be provided into machine e.g. like environment variable
+#     so agents developers can use this directory for backup agents logs.
+#CHE_WORKSPACE_LOGS_ROOT_DIR=/workspace_logs
```
riuvshin: maybe default value should be set right here in `che.env` instead of defining in each agent installer script? 
This way it will be more obvious to  end user where are logs stored by default

akorneta: By default value from `che.properties` will be used, and it equals to `/workspace_logs`(we won't define it in each script). This value is commented to be the same as all properties in this file.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/provision/env/LogsRootEnvVariableProvider.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Add env variable to machines with path to root folder of workspace logs.
+ *
+ * @author Anton Korneta
+ */
+public class LogsRootEnvVariableProvider implements EnvVarProvider {
+
+  /** Environment variable that points to root folder of projects inside machine */
+  public static final String WORKSPACE_LOGS_ROOT_VARIABLE = "CHE_WORKSPACE_LOGS_ROOT_DIR";
```
sleshchenko: I'd used `ENV_VAR` instead of `VARIABLE`

sleshchenko (approved):
LGTM.

Only one question: Did you consider configuring log dir for the terminal agent?

akorneta:
@sleshchenko 
Yes, I thought about that. Actually, to be able to store terminal logs into the file we need to do some work on that because currently they are streamed to stdout and by the way, terminal logs do not contain any useful information(just start/end of the session) so, that's why I did not work on that.

dockerfiles/init/manifests/che.env
riuvshin: My point was just to let user know what default is. So if default defined in che.props can you at least mention about that here in `che.env`

riuvshin: ok, thx

riuvshin (approved):

benoitf (approved):

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
```diff
@@ -28,6 +28,11 @@ che.workspace.storage=${che.home}/workspaces
 # workspace. This is the directory in the machine where your projects are placed.
 che.workspace.projects.storage=/projects
 
+#Defines the directory inside the machine where all the workspace logs are placed.
+#The value of this folder should be provided into machine e.g. like environment variable
+#so agents developers can use this directory for backup agents logs.
+che.workspace.logs.root.dir=/workspace_logs
```
garagatyi: Consider usage of underscore between `root` and `dir` words

skabashnyuk (approved):

garagatyi (approved):
Looks good, nice job!

l0rd:
👍 thanks @akorneta

PR #8558, opened 2018-02-01
Remove obsolete script

riuvshin:
As any CHE stack should work on openshift we don't need `replace_stacks.sh` anymore

related to: https://github.com/eclipse/che/pull/6618


sleshchenko (approved):

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
```diff
@@ -12,9 +12,7 @@
 #   curl -fsSL ${DEPLOY_SCRIPT_URL} -o get-che.sh
 #   WAIT_SCRIPT_URL=https://raw.githubusercontent.com/eclipse/che/master/dockerfiles/cli/scripts/openshift/wait_until_che_is_available.sh
 #   curl -fsSL ${WAIT_SCRIPT_URL} -o wait-che.sh
-#   STACKS_SCRIPT_URL=https://raw.githubusercontent.com/eclipse/che/master/dockerfiles/cli/scripts/openshift/replace_stacks.sh
-#   curl -fsSL ${STACKS_SCRIPT_URL} -o stacks-che.sh
-#   bash get-che.sh && wait-che.sh && stacks-che.sh
+#   bash get-che.sh && wait-che.sh
```
l0rd: I think you can remove script `wait-che.sh` too. WDYT?

benoitf (commented):
Up to @l0rd

dockerfiles/init/modules/openshift/files/scripts/deploy_che.sh
riuvshin: I think that script still could be useful... Myabe it is better to move it to deploy_che as a func and remove wait CHE funcs from the ocp.sh 
@sleshchenko @l0rd WDYT?

sleshchenko: @riuvshin I like this idea. Do you think it should always wait for Che or make it an option?

Also `get-che.sh` file name makes me a bit confused. Maybe it is better to rename it to `deploy_che.sh` as the original file.

riuvshin: that could be an option for sure, just to not break current approach

l0rd: +1 to add it as a function in `deploy_che.sh`, making it an option may be interesting but also part of another PR. It's up to you.

riuvshin: @sleshchenko @l0rd PR updated

sleshchenko: Should `bash get-che.sh --wait-che` be here?

riuvshin: @sleshchenko nice catch, ty, updated.

garagatyi (approved):

l0rd (approved):
nice :-)

Issue #8583, opened 2018-02-02
Unexpected fail of CheckErrorsWarningsTabTest in the selenium tests


Ohrimenko1988:
### Description
 
We have unexpected fail of CheckErrorsWarningsTabTest in the selenium tests, it looks like something went wrong with a method which does click to the item in the "Preferences" form. We need to investigate and fix this problem.

**Log:**
```
org.openqa.selenium.ElementNotVisibleException: 
element not visible
  (Session info: chrome=56.0.2924.87)
  (Driver info: chromedriver=2.27.440175 (9bc1d90b8bfa4dd181fbbf769a5eb5e575574320),platform=Linux 3.10.0-514.16.1.el7.x86_64 x86_64) (WARNING: The server did not provide any stacktrace information)
Command duration or timeout: 189 milliseconds
Build info: version: '3.1.0', revision: '86a5d70', time: '2017-02-16 07:57:44 -0800'
System info: host: '661cb80d0add', ip: '172.20.0.3', os.name: 'Linux', os.arch: 'amd64', os.version: '3.10.0-514.16.1.el7.x86_64', java.version: '1.8.0_121'
Driver info: org.openqa.selenium.chrome.ChromeDriver
Capabilities [{applicationCacheEnabled=false, rotatable=false, mobileEmulationEnabled=false, networkConnectionEnabled=false, chrome={chromedriverVersion=2.27.440175 (9bc1d90b8bfa4dd181fbbf769a5eb5e575574320), userDataDir=/tmp/.org.chromium.Chromium.cTVNJ3}, takesHeapSnapshot=true, pageLoadStrategy=normal, databaseEnabled=false, handlesAlerts=true, hasTouchScreen=false, version=56.0.2924.87, platform=LINUX, browserConnectionEnabled=false, nativeEvents=true, acceptSslCerts=true, locationContextEnabled=true, webStorageEnabled=true, browserName=chrome, takesScreenshot=true, javascriptEnabled=true, cssSelectorsEnabled=true, unexpectedAlertBehaviour=}]
Session ID: 2c4180fae34bc330547f7fc2c92292ad
Command duration or timeout: 956 milliseconds
Build info: version: '3.0.1', revision: '1969d75', time: '2016-10-18 09:49:13 -0700'
[...]

PR #8591, opened 2018-02-02
Selenium: Fix unexpected fail of CheckErrorsWarningsTabTest in the selenium tests

Ohrimenko1988:
### What does this PR do?
Fix unexpected fail of CheckErrorsWarningsTabTest in the selenium tests

Issue: https://github.com/eclipse/che/issues/8583


SkorikSergey (approved):

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/Preferences.java
```diff
@@ -592,4 +598,10 @@ public void clickOnShowArtifactCheckBox() {
     Actions actions = new Actions(seleniumWebDriver);
     actions.click(showArtifactCheckBox).build().perform();
   }
+
+  private void waitAndClickOnElement(WebElement element) {
```
dmytro-ndp: Let's reuse [SeleniumWebDriverHelper::waitAndClickOnElement() method](https://github.com/eclipse/che/blob/6.0.0/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/SeleniumWebDriverHelper.java#L60) instead of this one to avoid duplication.

dmytro-ndp (approved):

Issue #8288, opened 2018-01-13
[Multiuser] Authorize an application on GitHub from the 'Import GiitHub' does not work


artaleks9:
### Description
* When I try to authorize an application on GitHub from the 'Import Project GitHub' form    I have error notification and nothing more

**Reproduction Steps:**
* Open the IDE, go to the 'Workspace' in the Menu
* Select the 'Import Project' -> 'GitHub'
* Click the 'Load Repo'
 
**Expected behavior:**
* should be appeared web page with a login form to GitHub 

**Observed behavior:**
there is message: 'Failed to authorize application on GitHub'

**Che version:** 6.0.0-M5
**OS and version:** Fedora 25
**Che install:** Docker/OCP (multiuser)

**Additional information:**
* Problem can be reliably reproduced, doesn't happen randomly: [Yes]
* See the attachment:
![auth-github-import](https://user-images.githubusercontent.com/5408906/34907156-eea7d584-f882-11e7-9afc-5933ccab781e.gif)

ghost:
@artaleks9 I think adding steps on configuring GitHub oAuth are necessary too. Please, update the issue.

mshaposhnik:
Fixed;

PR #8604, opened 2018-02-05
Handle cases when connection is failed for keycloak related requests

RomanNikitenko:
### What does this PR do?
Do not wrap UnauthorizedException in ServerException at connecting to GitHub API using OAuth

#8288 

### DIFF
--- a/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequest.java
+++ b/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/rest/DefaultHttpJsonRequest.java
@@ -12,6 +12,7 @@ package org.eclipse.che.api.core.rest;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
 import static java.util.Objects.requireNonNull;
+import static javax.ws.rs.core.UriBuilder.fromUri;
 
 import com.google.common.io.CharStreams;
 import java.io.IOException;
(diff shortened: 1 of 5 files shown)


RomanNikitenko:
ci-test

vparfonov (approved):

RomanNikitenko:
ci-test

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/HttpJsonRequestUtil.java
```diff
@@ -0,0 +1,123 @@
...
+import com.google.gson.JsonSyntaxException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import org.eclipse.che.commons.annotation.Nullable;
+
+/**
+ * Provides useful methods to simplify work with {@link
+ * org.eclipse.che.api.core.rest.HttpJsonRequest} instances.
+ *
+ * @author Roman Nikitenko
```
skabashnyuk: Look like specific to KeycloakOAuthAuthenticationService.java class. Plese move it close to service

skabashnyuk (approved):

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/HttpJsonRequestUtil.java
RomanNikitenko: @skabashnyuk 
In DefaultHttpJsonRequest class we create formatted error message with corresponding info when request is failed.
We should have ability to extract this info from formatted message. 
At the moment I use this class to extract and analyze code and error message of response in KeycloakOAuthAuthenticationService.java class only.
But we can face the need to extract formatted in DefaultHttpJsonRequest class info in another services.
So I think this util class more related to DefaultHttpJsonRequest class, but up to you.
WDYT?

RomanNikitenko:
@vparfonov @skabashnyuk I've updated the PR, please review.

PR #8614, opened 2018-02-05
Support alternate OIDC providers, to prepare for the switch from Keycloak to `fabric8_auth`  

davidfestal:
### What does this PR do?

This PR allows switching to an alternate OIDC provider (provided that it emits
access tokens as JWT tokens).

This is the implementation required in upstream Che, for issues
redhat-developer/rh-che#502 and
redhat-developer/rh-che#525

### Changes in dependencies

This PR depends on PR https://github.com/eclipse/che-dependencies/pull/95 and the associated CQ.

### DIFF
--- a/assembly-multiuser/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/multiuser.properties
+++ b/assembly-multiuser/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/multiuser.properties
@@ -102,9 +102,13 @@ che.organization.email.org_renamed_template=st-html-templates/organization_renam
 #####                             KEYCLOACK CONFIGURATION                  #####
 
 #     Url to keycloak identity provider server
+#     Can be set to NULL only if `che.keycloak.oidcProvider`
+#     is used 
 che.keycloak.auth_server_url=http://${CHE_HOST}:5050/auth
 
 #     Keycloak realm is used to authenticate users
+#     Can be set to NULL only if `che.keycloak.oidcProvider`
+#     is used 
 che.keycloak.realm=che
 
 #     Keycloak client id in che.keycloak.realm that is used by dashboard, ide and cli to authenticate users
@@ -120,3 +124,18 @@ che.keycloak.github.endpoint=NULL
 
 #     The number of seconds to tolerate for clock skew when verifying exp or nbf claims.
 che.keycloak.allowed_clock_skew_sec=3
+
+#     Use the OIDC optional `nonce` feature to increase security.
+che.keycloak.use_nonce=true
+
+#     URL to the Keycloak Javascript adapter we want to use.
+#     if set to NULL, then the default used value is  
+#     `${che.keycloak.auth_server_url}/js/keycloak.js`,
+#     or `<che-server>/wsmaster/keycloak/OIDCKeycloak.js`
+#     if an alternate `oidc_provider` is used 
+che.keycloak.js_adapter_url=NULL
+
+#     Base URL of an alternate OIDC provider that provides
+#     a discovery endpoint as detailed in the following specification
+#     https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig
+che.keycloak.oidc_provider=NULL
--- a/assembly-multiuser/assembly-wsmaster-war/src/main/webapp/keycloak/OIDCKeycloak.js
+++ b/assembly-multiuser/assembly-wsmaster-war/src/main/webapp/keycloak/OIDCKeycloak.js
@@ -1,0 +1,1360 @@
+(function( window, undefined ) {
+
+    var Keycloak = function (config) {
+        if (!(this instanceof Keycloak)) {
+            return new Keycloak(config);
+        }
+
+        var kc = this;
+        var adapter;
+        var refreshQueue = [];
+        var callbackStorage;
+
+        var loginIframe = {
+            enable: true,
+            callbackList: [],
+            interval: 5
+        };
+
+        var useNonce = true;
+        
+        kc.init = function (initOptions) {
+            kc.authenticated = false;
+
+            callbackStorage = createCallbackStorage();
+
+            if (initOptions && initOptions.adapter === 'cordova') {
+                adapter = loadAdapter('cordova');
+            } else if (initOptions && initOptions.adapter === 'default') {
+                adapter = loadAdapter();
+            } else {
+                if (window.Cordova) {
+                    adapter = loadAdapter('cordova');
+                } else {
+                    adapter = loadAdapter();
+                }
+            }
+
+            if (initOptions) {
+                if (typeof initOptions.useNonce !== 'undefined') {
+                    useNonce = initOptions.useNonce;
+                }
+
+                if (typeof initOptions.checkLoginIframe !== 'undefined') {
+                    loginIframe.enable = initOptions.checkLoginIframe;
+                }
+
+                if (initOptions.checkLoginIframeInterval) {
+                    loginIframe.interval = initOptions.checkLoginIframeInterval;
+                }
+
+                if (initOptions.onLoad === 'login-required') {
+                    kc.loginRequired = true;
+                }
+
+                if (initOptions.responseMode) {
+                    if (initOptions.responseMode === 'query' || initOptions.responseMode === 'fragment') {
+                        kc.responseMode = initOptions.responseMode;
+                    } else {
+                        throw 'Invalid value for responseMode';
+                    }
+                }
+
+                if (initOptions.flow) {
+                    switch (initOptions.flow) {
+                        case 'standard':
+                            kc.responseType = 'code';
+                            break;
+                        case 'implicit':
+                            kc.responseType = 'id_token token';
+                            break;
+                        case 'hybrid':
+                            kc.responseType = 'code id_token token';
+                            break;
+                        default:
+                            throw 'Invalid value for flow';
+                    }
+                    kc.flow = initOptions.flow;
+                }
+
+                if (initOptions.timeSkew != null) {
+                    kc.timeSkew = initOptions.timeSkew;
+                }
+            }
+
+            if (!kc.responseMode) {
+                kc.responseMode = 'fragment';
+            }
+            if (!kc.responseType) {
+                kc.responseType = 'code';
+                kc.flow = 'standard';
+            }
+
+            var promise = createPromise();
+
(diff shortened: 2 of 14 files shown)


assembly-multiuser/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/multiuser.properties
```diff
@@ -120,3 +124,18 @@ che.keycloak.github.endpoint=NULL
 
 #     The number of seconds to tolerate for clock skew when verifying exp or nbf claims.
 che.keycloak.allowed_clock_skew_sec=3
+
+#     Use the OIDC optional `nonce` feature to increase security.
+che.keycloak.use_nonce=true
```
skabashnyuk, line 129: shall we provide some implementation agnostic name such as  ```che.openid.use_nonce``` - ?

assembly-multiuser/assembly-wsmaster-war/src/main/webapp/keycloak/OIDCKeycloak.js
```diff
@@ -0,0 +1,1371 @@
...
+    if ( typeof module === "object" && module && typeof module.exports === "object" ) {
+        module.exports = Keycloak;
+    } else {
+        window.Keycloak = Keycloak;
+
+        if ( typeof define === "function" && define.amd ) {
+            define( "keycloak", [], function () { return Keycloak; } );
+        }
+    }
+})( window );
```
benoitf, line 1380: how is unit tested that JS ?

davidfestal: That JS is not unit-tested for now TBH since it's mainly a modification of the `keycloak.js` Javascript adapter file. The next step (not a critical priority though) is to open a PR to the Keycloak repo to propose these changes, and have them tested in the context of the Keycloak project. I should be able to open the PR this week, and should be able to discuss with Keycloak team member about this at the end of the week.

skabashnyuk:
Documentation? How to do that?

assembly-multiuser/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/multiuser.properties
davidfestal: Well, for now this idea was to do as few modifications as possible.

Additionally, it's an option related to the *Keycloak Javascript Adapter* client library.
So having it called `che.keycloak.use_nonce` seems consistent to me.

We are still using the Keycloak Javascript adapter, but I simply made it more friendly to act as client of alternate OIDC servers.

davidfestal:
> Documentation? How to do that?

Simply by using properties documented in the properties file here: https://github.com/eclipse/che/pull/8614/files#diff-9392b7fdcb2c51c819ba7cdd95321768

But if there are some other places where doc has to be updated, could you point me to it ?

assembly-multiuser/assembly-wsmaster-war/src/main/webapp/keycloak/OIDCKeycloak.js
benoitf: @davidfestal maybe add a comment on the top of the file that it's a change from upstream keycloak and/or if possible to have a diff to maintain it over new versions of keycloak ?

davidfestal: well, I think that for now I can add a comment at the top of the file. Not sure about the diff, which would make it hard to maintain IMO.

The idea in fact is that, once it is merged on the Keycloak side, Che could simply retrieve the keycloak.js through NPM since the Keycloak JS adapter is available as a standalone NPM package: https://www.npmjs.com/package/keycloak-js.

Then we won't have to manage a diff at all.

skabashnyuk:
@eivantsov can you help us with docs?

assembly-multiuser/assembly-wsmaster-war/src/main/webapp/keycloak/OIDCKeycloak.js
davidfestal: @benoitf would this comment be sufficient for you : https://github.com/eclipse/che/pull/8614/commits/8237cd3b4115eb2355f0ae9ee707c2b1072a87e8 ?

benoitf (approved):
ok on my side for the part I could maintain

dashboard/src/app/index.module.ts
```diff
@@ -49,11 +49,19 @@ window.name = 'NG_DEFER_BOOTSTRAP!';
 
 declare const Keycloak: Function;
 function buildKeycloakConfig(keycloakSettings: any) {
-  return {
-    url: keycloakSettings['che.keycloak.auth_server_url'],
-    realm: keycloakSettings['che.keycloak.realm'],
-    clientId: keycloakSettings['che.keycloak.client_id']
-  };
+  const theOidcProvider = keycloakSettings['che.keycloak.oidc_provider'];
+  if(!theOidcProvider) {
```
benoitf: I think there is a missing space between if and parenthesis
Also could you confirm that there is no lint report from dashboard build around the UD changes ?

dashboard/src/app/index.module.ts
```diff
@@ -65,18 +73,18 @@ function keycloakLoad(keycloakSettings: any) {
...
-    script.src = keycloakSettings['che.keycloak.auth_server_url'] + '/js/keycloak.js';
+    script.src = keycloakSettings['che.keycloak.js_adapter_url'];
     script.addEventListener('load', resolve);
     script.addEventListener('error', () => reject('Error loading script.'));
     script.addEventListener('abort', () => reject('Script loading aborted.'));
     document.head.appendChild(script);
   });
 }
-function keycloakInit(keycloakConfig: any) {
+function keycloakInit(keycloakConfig: any, theUseNonce: boolean) {
```
benoitf, line 83: useNonce instead of theUseNonce ?

dashboard/src/app/index.module.ts
```diff
@@ -102,7 +110,11 @@ angular.element(document).ready(() => {
     // load Keycloak
     return keycloakLoad(keycloakSettings).then(() => {
       // init Keycloak
-      return keycloakInit(keycloakAuth.config);
+      var useNonce: boolean;
+      if (typeof keycloakSettings['che.keycloak.use_nonce'] == 'string') {
```
benoitf: shouldn't be === instead of == ?

dashboard/src/app/index.module.ts
```diff
@@ -102,7 +110,11 @@ angular.element(document).ready(() => {
...
+        useNonce = keycloakSettings['che.keycloak.use_nonce'].toLowerCase() == 'true';
```
benoitf: shouldn't be === instead of ==

dashboard/src/app/index.module.ts
davidfestal: it was to avoid a duplicate with the `useNonce` named parameter of the `keycloak.init()` method.

ibuziuk (approved):

assembly-multiuser/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/multiuser.properties
```diff
@@ -120,3 +124,18 @@ che.keycloak.github.endpoint=NULL
...
+
+#     URL to the Keycloak Javascript adapter we want to use.
+#     if set to NULL, then the default used value is  
+#     `${che.keycloak.auth_server_url}/js/keycloak.js`,
+#     or `<che-server>/wsmaster/keycloak/OIDCKeycloak.js`
+#     if an alternate `oidc_provider` is used 
+che.keycloak.js_adapter_url=NULL
```
riuvshin, line 136: should we add these new props to the `che.env` ?

davidfestal: not sure, because its main use, for now, will be to be used on the `rh-che` side (OSIO Che) to use the `fabric8_auth` OIDC provider instead of raw Keycloak.

dashboard/src/app/index.module.ts
davidfestal: fixed in commit https://github.com/eclipse/che/pull/8614/commits/b4604216a28c11a500d754e81df000213827762f

dashboard/src/app/index.module.ts
davidfestal: fixed in commit https://github.com/eclipse/che/pull/8614/commits/b4604216a28c11a500d754e81df000213827762f

dashboard/src/app/index.module.ts
davidfestal: fixed in commit https://github.com/eclipse/che/pull/8614/commits/b4604216a28c11a500d754e81df000213827762f

skabashnyuk:
@davidfestal what is your plan about master?

benoitf:
AFAIK a PR will be submitted for master by David

davidfestal:
@skabashnyuk @benoitf yes, I'm starting on the port to Che 6 on master right now, and will create a separate PR for it.

skabashnyuk (approved):

ghost:
@davidfestal will you please issue a PR to docs?

I think it fits this section well https://github.com/eclipse/che-docs/blob/master/src/main/pages/setup-openshift/openshift-config.md#multi-user-using-own-keycloak-and-psql

davidfestal:
@eivantsov I assume PR for docs should be synchronized with the PR I will open after porting this to Che 6 (master) ?

@skabashnyuk wdyt ? can we merge this PR to the `5.22.x` branch, and postpone the doc update for when the PR on master is opened ?

skabashnyuk:
@davidfestal ok for me

ghost:
Fine with me too. This new change isn't for all users, so we can document it later.

l0rd (approved):

assembly-multiuser/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/multiuser.properties
l0rd: I think that's fine for che 5 but you probably should add the new properties in `che.env` too for che 6.

PR #8615, opened 2018-02-05
Fix singleport mode in multiuser

mshaposhnik:
### What does this PR do?
Fixes problems launching singleport in multiuser mode.
 - Added proxy routes to keycloak server;
 - Added container network labels in multiuser mode for traefik and WS containers;
 - Some property injection fixes prevent NPE's.
#### Release Notes
N/A

#### Docs PR
N/A


benoitf (approved):

skabashnyuk (approved):

dockerfiles/init/modules/keycloak/templates/che-realm.json.erb
```diff
@@ -58,8 +58,18 @@
     "enabled" : true,
     "clientAuthenticatorType" : "client-secret",
     "secret" : "56abc663-1c41-4ab2-81d1-d4b704249221",
+<% if scope.lookupvar('che::che_single_port') == "true" -%>
```
riuvshin: I'm sorry to say that but, unfortunately, these changes require some changes here: https://github.com/eclipse/che/blob/master/dockerfiles/init/s2i/keycloak/assemble#L42 
as you can see there for some reason we are removing all template injections because file `che-realm.json.erb` will be just renamed to `che-realm.json` and used in KC as is... 
otherwise, importing `che-realm.json` will fail for CHE multiuser deployed to openshift

riuvshin:
I think we need to add some docs related to single port.
for example, in which case che_docker_ip_external should be used?
how to set up to make this work with DNS names, etc.

dockerfiles/init/modules/keycloak/templates/che-realm.json.erb
mshaposhnik: Ok, yes, i forgot about this problem. I think we can remove this, since it is will be applied only if the first keycloak run is made with singleport option. This will require the redirect and webOrigin URL's to be set manually in keycloak GUI.

mshaposhnik:
It is discussed with @eivantsov that we will provide full information support when the sinleport doc will be rewritten. But basically there is not much changes since previous version of sinlgle port. Just a few more props added to allow custom wildcard DNS like nip.io etc

garagatyi (approved):

riuvshin (approved):

PR #8636, opened 2018-02-06
Property rename: old value to new support

skabashnyuk:
### What does this PR do?
During code evolution might be the case then someone will want to rename some property. This brings a couple of problems like support of old property name in external plugins and support old configuration values in code with the new property name. To cover these cases there is a file  che_aliases.properties that contains old names of all existed properties.  It has such format current_name =old_name, very_old_name. In this case will be such binding.
1. Always current_name = current_value
2. if old_name property exist it will be binded to old_value, and current_name = old_value and very_old_name = old_value
3. if very_old_name property exist it will be binded to very_old_value, and current_name = very_old_value and old_name = very_old_value

NOTE: its prohibited to use a different name for same property on the same level. From the example above - you can use environment property CHE_CURRENT_NAME and CHE_OLD_NAME. But you can use it on a different level, for instance, environment property and system property.

#### Release Notes
Extended support for property renaming.

#### Docs PR
n/a

### DIFF
--- a/core/commons/che-core-commons-inject/src/main/java/org/eclipse/che/inject/CheBootstrap.java
+++ b/core/commons/che-core-commons-inject/src/main/java/org/eclipse/che/inject/CheBootstrap.java
@@ -32,8 +32,10 @@ import java.nio.charset.Charset;
 import java.util.AbstractMap.SimpleEntry;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Properties;
 import java.util.Set;
 import java.util.function.Function;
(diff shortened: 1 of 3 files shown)


skabashnyuk:
ci-test

skabashnyuk:
ci-test

core/commons/che-core-commons-inject/src/main/java/org/eclipse/che/inject/CheBootstrap.java
```diff
@@ -98,9 +100,24 @@
  * <tr><td>${root_data}/input/</td><td>&nbsp;</td><td>&nbsp;</td><td>${root_data}/input/</td></tr>
  * </table>
  *
+ * During code evolution might be the case then someone will want to rename some property. This
```
sleshchenko: Looks like typo is here:
be the case ~t~when someone

sleshchenko (approved):
OK for me

Issue #7889, opened 2017-12-14
Save ws-agent logs of test workspaces which were used by failed selenium tests


dmytro-ndp:
### Description
Sometimes we need to have _ws-agent_'s logs to investigate the reasons of selenium tests failures.
There are parameters of logging which could be used to setup storing of ws-agents log to the host where the testing Eclipse Che is run: https://github.com/eclipse/che/blob/che6/assembly/assembly-wsagent-server/src/assembly/setenv.sh#L41

We can also borrow the _ws-agent's_ logs from the workspace container before removal of it and save it into the tests target directory at the end of failed tests execution by the command like the follow:
```
docker exec -it $(docker ps | grep dev-machine | cut -d ' ' -f1) sh -c "cat /home/user/che/ws-agent/logs/archive/$(date +%Y/%m/%d)/catalina.log"
```

Alternative command to save all logs:
```
docker cp $(docker ps -q -f name=<ws-id>):/home/user/che/ws-agent/logs/archive/ <local-destination>

PR #8652, opened 2018-02-07
Save meaningful logs of failed tests workspaces in the target directory

dmytro-ndp:
### What does this PR do?
Sometimes it is needed to have workspace logs to investigate the reasons of selenium tests failures.
This request add functionality to store non-empty workspace logs of failed tests into the test **target** directory taking into account _Docker_ and _OpenShift_ infrastructure.

#7889


core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
```diff
@@ -42,7 +44,7 @@
    * @param stderr a consumer where stderr will be redirected
    * @throws IOException
    */
-  public static void process(Process p, LineConsumer stdout, LineConsumer stderr)
+  public static void readOutput(Process p, LineConsumer stdout, LineConsumer stderr)
```
tolusha: It isn't a good idea to refactor core util class

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
```diff
@@ -126,6 +129,73 @@ public static Process executeAndWait(
...
+      throw new TimeoutException(
+          format(
+              "Process '%s' was terminated by timeout %s %s.",
+              Joiner.on(" ").join(commandLine), timeout, timeUnit.name().toLowerCase()));
+    }
+
+    future.join(); // wait until process output is read
+
+    String errorMessage = stderr.getText();
+    if (!errorMessage.isEmpty()) {
```
tolusha: Some processes write into stderr. 
You'd have better to check result code.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/utils/FileUtil.java
```diff
@@ -0,0 +1,30 @@
...
+
+/** @author Dmytro Nochevnov */
+public class FileUtil {
+
+  public static void removeEmptyDirectory(Path directory) throws IOException {
+    if (Files.exists(directory)
+        && Files.isDirectory(directory)
+        && Files.list(directory).collect(toList()).isEmpty()) {
+
+      Files.delete(directory);
```
tolusha: This method deletes only empty directories.

dmytro-ndp: Do you think the method name is not good?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
skabashnyuk: -1 to change method name. It can be used by someone else.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
```diff
@@ -66,7 +68,7 @@ public static void process(Process p, LineConsumer stdout, LineConsumer stderr)
    * @param stdout a consumer where stdout will be redirected
    * @throws IOException
    */
```
skabashnyuk: -1 to change method name. It can be used by someone else.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
dmytro-ndp: Agree

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
dmytro-ndp: Agree

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
dmytro-ndp: I see. Will restore the previous name.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
dmytro-ndp: Good catch. Will check if `process.exitCode()` is suitable for the case of execution of docker command which fails.

skabashnyuk: what will happen if stderr contains 2GB of data? 10GB of data?

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/utils/FileUtil.java
skabashnyuk: org.eclipse.che.commons.lang.IoUtil.deleteRecursive

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
dmytro-ndp: @riuvshin said that from the point of selenium tests workspace logs hardly ever can have size more than hundred Mb, so we may not to check the size additionally.

dmytro-ndp: Thank you for advice. We can limit size of text which we are putting into the exception, say, to hundreds symbols.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/utils/FileUtil.java
dmytro-ndp: Thank you for advice.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
dmytro-ndp: @tolusha:
> Some processes write into stderr.
> You'd have better to check result code.

In that case `ProcessBuilder#start()` throws **IOException**, and for that case we have already had another method [org.eclipse.che.api.core.util.ProcessUtil#executeAndWait(java.lang.String[], int, java.util.concurrent.TimeUnit, org.eclipse.che.api.core.util.LineConsumer)](https://github.com/eclipse/che/blob/b90cc2dfd86af12ea680f9ce6177410a5f946224/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java#L96).

at the same time proposed method is dedicated to handle the case when exit code = 0, but **stderr** is not empty - that is what we sometimes have as a result of docker command execution.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/utils/FileUtil.java
dmytro-ndp: In this case `Files.delete(directory)` is safe enough so as we are verifying if directory is empty itself before removal.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
dmytro-ndp: I have found out that `ProcessBuilder#start()` doesn't throw **IOException** in case of running error-prone command as following `bash -c '<error-prone_command>'`, despite of non-zero exit code. So, additional check of non-zero exit code will be enough.

dmytro-ndp:
ci-build

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
```diff
@@ -126,6 +130,83 @@ public static Process executeAndWait(
...
+              Joiner.on(" ").join(commandLine), timeout, timeUnit.name().toLowerCase()));
+    }
+
+    future.join(); // wait until process output is read
+
+    if (process.exitValue() == 0) {
+      return process;
+    }
+
+    String errorMessage = stderr.getText();
```
skabashnyuk: How are you going to handle 10GB of error text here?

dmytro-ndp: I propose to cut error text which we is setting as error message of exception, to 1000 symbols.

skabashnyuk: As far as I can see code that you proposed is not enough. Because OOM will happen earlier. Do you see that?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
```diff
@@ -126,6 +130,83 @@ public static Process executeAndWait(
...
+   *     the command execution
+   * @throws TimeoutException if process gets more time then defined by {@code timeout}
+   */
+  public static Process executeAndWait(
+      String[] commandLine,
+      int timeout,
+      TimeUnit timeUnit,
+      ListLineConsumer stdout,
+      ListLineConsumer stderr)
+      throws TimeoutException, IOException, InterruptedException {
```
skabashnyuk: All methods of this class are using LineConsumer why you limit it with ListLineConsumer?

dmytro-ndp: The problem is [LineConsumer](https://github.com/eclipse/che/blob/master/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LineConsumer.java) interface doesn't have methods to read consumed line, whereas [ListLineConsumer](https://github.com/eclipse/che/blob/master/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ListLineConsumer.java) has. So, usage of **ListLineConsumer** simplifies reading **stderr** after execution of command in code `String errorMessage = stderr.getText();`.
Of course, the method can be re-wrote to use **LineConsumer** arguments, but it will add potentially unnecessary complexity to it, IMHO.

skabashnyuk: What is the main purpose of ```executeAndWait``` to execute and wait or to show the error message? Are you sure you need to read ALL message from stderror to send part of it in an exception?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
dmytro-ndp: It could happen when we are writing into the ListLineConsumer's **stdout** and **stderr** here:
- https://github.com/eclipse/che/blob/61c6247e26d6ee38cc8f029a2eb41135eeb4b216/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java#L56
- https://github.com/eclipse/che/blob/61c6247e26d6ee38cc8f029a2eb41135eeb4b216/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java#L59

skabashnyuk: So this method designed for commands that for sure are not producing some large output?
What if you change ListLineConsumer's to LineConsumer and remove part of the stderr from output.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
skabashnyuk: One more question. What should caller method do with Process in case if it's not completed during the desired timeout?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
dmytro-ndp: It's possible with some remarks https://github.com/eclipse/che/pull/8652#discussion_r167246019

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
skabashnyuk: In case if you change ListLineConsumer  to LineConsumer I see no reason to limit it.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
skabashnyuk: Not clear enough.  Is this method designed to handle large output for stderror or stdout?

skabashnyuk:
@dmytro-ndp  would you consider as an option moving new methods from ProcessUtil.java to some util package in selenium modules or you still want to make it general?

dmytro-ndp:
I would prefer to make it general, or to fix existed method org.eclipse.che.api.core.util.ProcessUtil#executeAndWait(java.lang.String[], int, java.util.concurrent.TimeUnit, org.eclipse.che.api.core.util.LineConsumer) to throw an exception if command returns exit code > 0. 
Handling such situation could be useful not just for selenium tests, but also in other cases.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
dmytro-ndp: Okay, I will change it to avoid OOM.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
dmytro-ndp: ListLineConsumer is not designed to avoid OOM when something write 10 GB by using method writeLine().

dmytro-ndp:
ci-test-ocp

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ListLineConsumer.java
```diff
@@ -24,21 +24,52 @@
...
+    if (textSizeLimit < 0) {
+      lines.add(line);
+      return;
+    }
+
+    if (consumedTextSize >= textSizeLimit) {
+      return;
+    }
+
+    if (line != null && (line.length() + consumedTextSize > textSizeLimit)) {
```
tolusha: Does it consume only first `testSizeLimit` characters? Will it be more relevant to consume the last  `testSizeLimit` characters? I would personally introduce a new class for that.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/StubLineConsumer.java
```diff
@@ -0,0 +1,22 @@
...
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.api.core.util;
+
+/**
+ * Line consumer which does nothing.
+ *
+ * @author Dmytro Nochevnov
+ */
+public class StubLineConsumer implements LineConsumer {
```
tolusha: There is https://github.com/eclipse/che/blob/master/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/AbstractLineConsumer.java

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ProcessUtil.java
```diff
@@ -27,11 +27,13 @@
  *
  * @author andrew00x
  * @author Alexander Garagatyi
+ * @author Dmytro Nochevnov
  */
 public final class ProcessUtil {
 
   private static final Logger LOG = LoggerFactory.getLogger(ProcessUtil.class);
   private static final ProcessManager PROCESS_MANAGER = ProcessManager.newInstance();
+  private static final int MAX_SIZE_OF_ERROR_MESSAGE = 1000;
```
skabashnyuk: Not used?

dmytro-ndp: My fault. Forgot to remove.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/StubLineConsumer.java
dmytro-ndp: I know, but AbstractLineConsumer requires implementation and here it is.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/ListLineConsumer.java
dmytro-ndp: 1. `Does it consume only first testSizeLimit characters?`
Yes

1. `Will it be more relevant to consume the last testSizeLimit characters?`
I proceeded from the fact that usually important information about an error is being placed at the start of message.

1. `I would personally introduce a new class for that.`
Agree

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/StubLineConsumer.java
tolusha: https://github.com/eclipse/che/blob/2c2bd6637c8169018e2b33397f9d738bbf82a0b1/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LineConsumer.java#L26
:)

dmytro-ndp:
@skabashnyuk, @tolusha: please, review fresh fix up,

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LimitedListLineConsumer.java
```diff
@@ -0,0 +1,64 @@
...
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.api.core.util;
+
+/**
+ * Extension of {@link ListLineConsumer} which consumes limited amount of text {@link #writeLine}.
```
skabashnyuk: Not clear what this class limiting. Characters or lines?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LimitedListLineConsumer.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  @Override
+  public void writeLine(String line) {
+    if (textSizeLimit < 0) {
+      lines.add(line);
+      return;
+    }
+
+    if (consumedTextSize >= textSizeLimit) {
```
skabashnyuk: Very strange algorithm. Can you describe in general how writeLine of this class is working?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LimitedListLineConsumer.java
dmytro-ndp: It limits size of text which is consuming. 
Will fix class description.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LimitedListLineConsumer.java
dmytro-ndp: - _writeLine()_ ensures that size of consuming text, that `([line1 size] + [line2 size] + ... + [lineN size]) <= [textSizeLimit value]` if **textSizeLimit** > 0
- _writeLine()_ doesn't limit size if **textSizeLimit** < 0
- _writeLine()_ ignores lines if **textSizeLimit** = 0
- _writeLine()_ writes substring of _lineN_ if `(consumedTextSize + [lineN size]) > [textSizeLimit value]` so that `(consumedTextSize + [substring of lineN size]) = [textSizeLimit value]`

Will add this explanation to the method description.

tolusha: Personally, It should look like this:
```
public class LimitedSizeLineConsumer implements LineConsumer {
  private final LineConsumer lineConsumer;
  private final int capacity;
  private int length;

  public LimitedSizeLineConsumer(LineConsumer lineConsumer, int capacity) {
    this.lineConsumer = lineConsumer;
    this.capacity = capacity;
  }

  @Override
  public void writeLine(String line) throws IOException {
    if (line == null) {
      return;
    }

    int maxLineLength = Math.min(capacity - length, line.length());
    lineConsumer.writeLine(line.substring(0, maxLineLength));
    length += maxLineLength;
  }

  @Override
  public void close() throws IOException {
    lineConsumer.close();
  }
}
```

Take into account that, `substring` method doesn't create a new string, so the original string still will be in memory.

skabashnyuk: I agree with @tolusha. His implementation looks better. @dmytro-ndp your implementation looks overcomplicated. The thing that I would add is a condition if we know that we already wrote the maximum number of characters then we simply exit from ```writeLine``` method.

@dmytro-ndp Please consider moving LimitedSizeLineConsumer to selenium framework.

dmytro-ndp: @tolusha: your implementation looks simpler, and I like an idea to have a kind of wrapper for **LineConsumer** classes. Despite that I see some issues there:
1. It doesn't check that capacity is positive number. But it could be easily fixed.
1. It doesn't work properly in case of usage **org.eclipse.che.api.core.util.ListLineConsumer** after **org.eclipse.che.api.core.util.ListLineConsumer#clear()** method is called. 
1. Some of the LineConsumer implementations, for example, ListLineConsumer, consume `line = null` without any exceptions. So, next code will break their behaviour:
```
    if (line == null) {
      return;
    }
```

It is also kind of wrapper, so it better to name it properly, for example: **LimitedSizeLineConsumerWrapper**.

dmytro-ndp: @skabashnyuk: 
> The thing that I would add is a condition if we know that we already wrote the maximum number of characters then we simply exit from `writeLine` method.

Good catch. Thank you!

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LimitedSizeLineConsumerWrapper.java
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Wrapper of {@link LineConsumer} which limits total size of consuming lines.
+ *
+ * @author Dmytro Nochevnov
+ * @author Anatolii Bazko
+ */
+public class LimitedSizeLineConsumerWrapper implements LineConsumer {
+  private final int limit;
+  private final LineConsumer lineConsumer;
```
skabashnyuk: Why you chouse wrapping instead of extending?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LimitedSizeLineConsumerWrapper.java
```diff
@@ -0,0 +1,66 @@
...
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.api.core.util;
+
+import java.io.IOException;
+
+/**
+ * Wrapper of {@link LineConsumer} which limits total size of consuming lines.
+ *
```
skabashnyuk: Are you sure about lines?
Maybe characters?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LimitedSizeLineConsumerWrapper.java
```diff
@@ -0,0 +1,66 @@
...
+  private final LineConsumer lineConsumer;
+  private int totalLength;
+
+  /**
+   * @param limit maximum total size of consuming lines.
+   * @throws IllegalArgumentException if the {@code limit} is negative.
+   */
+  public LimitedSizeLineConsumerWrapper(LineConsumer lineConsumer, int limit) {
+    if (limit < 0) {
+      throw new IllegalArgumentException(Integer.toString(limit));
```
skabashnyuk: Not very clear exception message. Please add more details.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LimitedSizeLineConsumerWrapper.java
dmytro-ndp: Because @tolusha suggested wrapping [here](https://github.com/eclipse/che/pull/8652/files/1f18703c2ee9bcc393593ffc5861542425f58534#r167779494) and it actually looks better than my variant, and you agreed with it [here](https://github.com/eclipse/che/pull/8652/files/1f18703c2ee9bcc393593ffc5861542425f58534#r167779494).

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LimitedSizeLineConsumerWrapper.java
dmytro-ndp: Okay.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LimitedSizeLineConsumerWrapper.java
dmytro-ndp: As for me, `size of consuming characters` sounds strange so as each character has fixed size which can't be limited.

dmytro-ndp:
ci-test-ocp

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LimitedSizeLineConsumerWrapper.java
tolusha: Could you explain why did you add `Wrapper` suffix ?
https://en.wikipedia.org/wiki/Chain-of-responsibility_pattern

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LimitedSizeLineConsumerWrapper.java
```diff
@@ -0,0 +1,66 @@
...
+public class LimitedSizeLineConsumerWrapper implements LineConsumer {
+  private final int limit;
+  private final LineConsumer lineConsumer;
+  private int totalLength;
+
+  /**
+   * @param limit maximum total size of consuming lines.
+   * @throws IllegalArgumentException if the {@code limit} is negative.
+   */
+  public LimitedSizeLineConsumerWrapper(LineConsumer lineConsumer, int limit) {
```
tolusha: If we have `LimitedSize` class prefix then the parameter should be `size` but not `limit`

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LimitedSizeLineConsumerWrapper.java
skabashnyuk: Ok. you may rephrase as you want. But ```size of consuming lines``` actually is not true.

``` size of consuming characters``` - a number, not size.

File: core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LimitedSizeLineConsumerWrapper.java (hunk 12 above)
Comment (dmytro-ndp) [reply core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LimitedSizeLineConsumerWrapper.java]: Because it looks and behaves like a **structural** pattern _Wrapper_ kind of _Decorator_ which extends functionality of wrapping object of type **LineConsumer** dynamically - in our case is limiting total size of consuming lines and sending all messages to wrapping object being a type of  wrapping object.
Whereas _Chain of Responsibility_ pattern is **behavioral** pattern which chains the receiving objects and pass the request along the chain until an object handles it. So, it could extends interface of receivers of object only, not the interface of object to send.
More details about difference between _Decorator_ and _Chain of Responsibility_ patterns can be found here: https://coderanch.com/t/668194/certification/Difference-Decorator-Chain-Responsibility-patterns
An examples of _Wrappers_ in Eclipse Che codebase:
- [IndentWrapperLineConsumer](https://github.com/eclipse/che/blob/2c2bd6637c8169018e2b33397f9d738bbf82a0b1/core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/IndentWrapperLineConsumer.java#L20)
- [WebSocketJsoWrapper.java](https://github.com/eclipse/che/blob/2c2bd6637c8169018e2b33397f9d738bbf82a0b1/ide/commons-gwt/src/main/java/org/eclipse/che/ide/websocket/impl/WebSocketJsoWrapper.java#L20)

File: core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LimitedSizeLineConsumerWrapper.java (hunk 13 above)
Comment (dmytro-ndp) [reply core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LimitedSizeLineConsumerWrapper.java]: I see. What about the next variant:
`Wrapper of {@link LineConsumer} object which is limiting total number of characters which this object can consume`

Comment (skabashnyuk) [reply core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LimitedSizeLineConsumerWrapper.java]: ok

File: core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LimitedSizeLineConsumerWrapper.java (hunk 15 above)
Comment (dmytro-ndp) [reply core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LimitedSizeLineConsumerWrapper.java]: Frankly speaking, as for me, class name **LimitedLineConsumerWrapper** with **limit** field inside looks more obvious, similar to the class [LimitedInputStream](https://github.com/eclipse/che/blob/2c2bd6637c8169018e2b33397f9d738bbf82a0b1/infrastructures/docker/docker-client/src/main/java/org/eclipse/che/infrastructure/docker/client/connection/LimitedInputStream.java#L17).

**Comment by tolusha** [comment]
To stop endless discussion I would recommend you to move the new class into `selenium` module

**Comment by dmytro-ndp** [comment]
Why do you call it endless? I have fixed a lot of things you had suggested, haven't it?

**Comment by dmytro-ndp** [comment]
Reading stderr stream from process is not selenium-test specific part, and proposed classes would be useful for someone else.

File: core/che-core-api-core/src/main/java/org/eclipse/che/api/core/util/LimitedLineConsumerWrapper.java [hunk 16]
```diff
@@ -13,12 +13,13 @@
...
- * Wrapper of {@link LineConsumer} object which is limiting total number of characters which this object can consume.
+ * Wrapper of {@link LineConsumer} object which is limiting total number of characters which this
+ * object can consume.
  *
  * @author Dmytro Nochevnov
  * @author Anatolii Bazko
  */
-public class LimitedSizeLineConsumerWrapper implements LineConsumer {
+public class LimitedLineConsumerWrapper implements LineConsumer {
   private final int limit;
```
skabashnyuk: We are walking by the circle. Can you guarantee that this class limiting number of lines?

skabashnyuk:
I'm sorry @dmytro-ndp there are 73 messages in this PR and it's going to be hard to review again and again. Plese follow @tolusha suggestion https://github.com/eclipse/che/pull/8652#issuecomment-365594772

dmytro-ndp:
I closed pull requests because of reviewers seems just kidding me.

Issue #8465, opened 2018-01-26
Unexpected failing of NavigateToFileTest selenium test on CI 


SkorikSergey:
### Description
We have unexpected fail of **NavigateToFileTest** selenium test. The test sometimes fails on CI when waiting suggestion list with found files from the **Navigate To File** feature. We should clarify reason and fix it.

**Log:**
```
org.openqa.selenium.WebDriverException: 
unknown error: Element <td style="opacity: 0.6;">...</td> is not clickable at point (895, 300). Other element would receive the click: <div class="GIWAMGHBPOB" style="right: 0px; bottom: 0px; position: absolute; left: 0px; top: 0px;">...</div>
  (Session info: chrome=56.0.2924.87)
  (Driver info: chromedriver=2.27.440175 (9bc1d90b8bfa4dd181fbbf769a5eb5e575574320),platform=Linux 3.10.0-514.16.1.el7.x86_64 x86_64) (WARNING: The server did not provide any stacktrace information)
Command duration or timeout: 129 milliseconds
Build info: version: '3.1.0', revision: '86a5d70', time: '2017-02-16 07:57:44 -0800'
System info: host: 'bf642ff2277b', ip: '172.18.0.3', os.name: 'Linux', os.arch: 'amd64', os.version: '3.10.0-514.16.1.el7.x86_64', java.version: '1.8.0_121'
Driver info: org.openqa.selenium.chrome.ChromeDriver
Capabilities [{applicationCacheEnabled=false, rotatable=false, mobileEmulationEnabled=false, networkConnectionEnabled=false, chrome={chromedriverVersion=2.27.440175 (9bc1d90b8bfa4dd181fbbf769a5eb5e575574320), userDataDir=/tmp/.org.chromium.Chromium.zvskk5}, takesHeapSnapshot=true, pageLoadStrategy=normal, databaseEnabled=false, handlesAlerts=true, hasTouchScreen=false, version=56.0.2924.87, platform=LINUX, browserConnectionEnabled=false, nativeEvents=true, acceptSslCerts=true, locationContextEnabled=true, webStorageEnabled=true, browserName=chrome, takesScreenshot=true, javascriptEnabled=true, cssSelectorsEnabled=true, unexpectedAlertBehaviour=}]
Session ID: 33994dcb-678a-4b12-adc2-a997bfb381c4
[...]

PR #8654, opened 2018-02-07
Selenium: wait that a found files list in the Navigate To File widget is changed

SkorikSergey:
### What does this PR do?
This PR adds timeout for waiting for a found files list in the **Navigate To File** widget is changed. 

https://github.com/eclipse/che/issues/8465

### DIFF
--- a/selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/miscellaneous/NavigateToFileTest.java
+++ b/selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/miscellaneous/NavigateToFileTest.java
@@ -27,7 +27,6 @@ import java.util.Map;
 import java.util.concurrent.ThreadLocalRandom;
 import org.eclipse.che.selenium.core.client.TestCommandServiceClient;
 import org.eclipse.che.selenium.core.client.TestProjectServiceClient;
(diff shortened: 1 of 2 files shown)


selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/miscellaneous/NavigateToFileTest.java
```diff
@@ -147,6 +148,9 @@ private void launchNavigateToFileAndCheckResults(
     String openedFileNameInTheTab = openedFileWithExtension.replace(".java", "");
     launchNavigateToFileFromUIAndTypeValue(navigatingValue);
     navigateToFile.waitSuggestedPanelIsDisplayed();
```
dmytro-ndp: According to [scrrenshot](https://user-images.githubusercontent.com/7760565/34935870-055e2b46-f9e8-11e7-986b-e07d584c9f16.png) from the target issue the problem seems with method *waitSuggestedPanelIsDisplayed()* which has succeed before suggested panel is actually displayed. Let's fix this method instead of adding `WaitUtils.sleepQuietly(3)` pause.

dmytro-ndp (approved):

Issue #8580, opened 2018-02-02
Use URI for Classfiles in Java Debugger


tsmaeder:
Currently, we are using a combination of <library id>/<fully qualified typename> to identify a debugger location in the Java debug support for jdt.ls. 
We should use the uri's that jdt.ls uses when identifying a class file. See for example ExternalResourceLocationHandler.java.

mmorhun:
Done

PR #8656, opened 2018-02-07
CHE-8580: Use URI for Classfiles in Java Debugger

mmorhun:
### What does this PR do?
Fix jar file location path.

https://github.com/eclipse/che/issues/8580

### DIFF
--- a/plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/tree/library/JarFileNode.java
+++ b/plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/tree/library/JarFileNode.java
@@ -145,7 +145,12 @@ public class JarFileNode extends SyntheticNode<JarEntry>
 
   @Override
   public Path getLocation() {
-    return Path.valueOf(getData().getPath());
+    // We have to use uri instead of path here despite it looks a bit weird.
+    // We are allowed to do this because this path is used as virtual file ID in the editor only.
+    // First, sometimes we don't have path field filled in (like in case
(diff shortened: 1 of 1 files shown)


tolusha (approved):

svor (approved):

evidolob (approved):

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/tree/library/JarFileNode.java
```diff
@@ -145,7 +145,7 @@ public boolean isLeaf() {
 
   @Override
   public Path getLocation() {
-    return Path.valueOf(getData().getUri());
+    return Path.valueOf(getData().getPath());
```
tsmaeder: Che IDE uses VirtualFile#getPath() as an identifier for the virtual file. However, when we get things via LSP (like "textDocument/references"), we only get an URI. So the URI is the only unique identifier we have for a file (possibly a classfile). So we need to get the URI into the Path somehow. 
There seems to have been a concrete case where the old version did not work. Can you say what that case was?

tsmaeder (changes requested):

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/tree/library/JarFileNode.java
mmorhun: Yes, I missed case with "textDocument/references". Thanks for spotting!

tsmaeder (approved):

Issue #8468, opened 2018-01-26
Sometimes  Selenium cannot open workspace page after invocation of the straight URL and login.


Ohrimenko1988:
### Description
Selenium cannot open workspace page after invocation of the straight URL and login. It looks like something went wrong with user credentials which were used for workspace creation. We need to investigate and 

**Log:**
```
org.openqa.selenium.TimeoutException: Expected condition failed: waiting for visibility of element located by By.id: gwt-debug-projectTree (tried for 120 second(s) with 500 MILLISECONDS interval)
Build info: version: '3.0.1', revision: '1969d75', time: '2016-10-18 09:49:13 -0700'
System info: host: 'slave6.codenvycorp.com', ip: '127.0.0.1', os.name: 'Linux', os.arch: 'amd64', os.version: '3.10.0-514.16.1.el7.x86_64', java.version: '1.8.0_102'
Driver info: org.eclipse.che.selenium.core.SeleniumWebDriver
	at org.eclipse.che.selenium.editor.autocomplete.FormatterTest.formatterTest(FormatterTest.java:102)
Caused by: org.openqa.selenium.NoSuchElementException: 
no such element: Unable to locate element: {"method":"id","selector":"gwt-debug-projectTree"}
  (Session info: chrome=56.0.2924.87)
  (Driver info: chromedriver=2.27.440175 (9bc1d90b8bfa4dd181fbbf769a5eb5e575574320),platform=Linux 3.10.0-514.16.1.el7.x86_64 x86_64) (WARNING: The server did not provide any stacktrace information)
Command duration or timeout: 22 milliseconds
For documentation on this error, please visit: http://seleniumhq.org/exceptions/no_such_element.html
Build info: version: '3.1.0', revision: '86a5d70', time: '2017-02-16 07:57:44 -0800'
System info: host: '82fc916612ef', ip: '172.18.0.4', os.name: 'Linux', os.arch: 'amd64', os.version: '3.10.0-514.16.1.el7.x86_64', java.version: '1.8.0_121'
Driver info: org.openqa.selenium.chrome.ChromeDriver
[...]

PR #8680, opened 2018-02-08
Catch issue #8468 in the ProjectExplorer page-object

Ohrimenko1988:
### What does this PR do?
Catch issue #8468 in the ProjectExplorer page-object

Issue: https://github.com/eclipse/che/issues/8468


dmytro-ndp (approved):
`page-object` looks slightly inappropriate as for me.
Let's use well-known term `page object`.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/SeleniumWebDriverHelper.java
```diff
@@ -65,4 +70,13 @@ public void waitAndClickOnElement(By elementLocator) {
...
+
+  public boolean elementIsVisible(By locator) {
+    try {
+      attachingElementToDomWait.until(visibilityOfElementLocated(locator));
+      return true;
+    } catch (TimeoutException ex) {
+      return false;
+    }
+  }
 }
```
musienko-maxim, line 82: I have not quiet understood why we need SeleniumWebDriverHelper, why not SeleniumWebDriverUtils class? The classes very similar as for me.

dmytro-ndp: **SeleniumWebDriverUtils** - means utility to expand **SeleniumWebDriver** functionality itself, while **SeleniumWebDriverHelper** provides high-level functionality which just consume **SeleniumWebDriver**

musienko-maxim (approved):

PR #8709, opened 2018-02-09
Increase postgres claim for OpenShift to be 1gb

gshipley:
For compatibility and portability between OSO, OSD, and OCP this needs to be 1Gi as anything under that is not allowed in Online environments.  Having it set to 100Mi results in errors and the deployment fails.

### DIFF
--- a/dockerfiles/init/modules/openshift/files/scripts/multi-user/keycloak/keycloak-data-claim.yaml
+++ b/dockerfiles/init/modules/openshift/files/scripts/multi-user/keycloak/keycloak-data-claim.yaml
@@ -16,5 +16,5 @@ spec:
   - ReadWriteOnce
   resources:
     requests:
-      storage: 100Mi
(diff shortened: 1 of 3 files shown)


riuvshin (approved):

riuvshin:
@gshipley could you please also do same for 
https://github.com/eclipse/che/blob/master/dockerfiles/init/modules/openshift/files/scripts/multi-user/keycloak/keycloak-data-claim.yaml#L19
https://github.com/eclipse/che/blob/master/dockerfiles/init/modules/openshift/files/scripts/multi-user/keycloak/keycloak-log-claim.yaml#L19

gshipley:
@riuvshin I updated the two additional files for keycloak.  I just didn't see the error on those yet since I am working through the docker build issue that isn't supported on OpenShift Online.

riuvshin:
@gshipley thanks!

Issue #8682, opened 2018-02-08
UI for workspace loading application


vitaliy-guliy:
#8556 needs a proper UI containing a simple loader and field with log messages.

benoitf:
hello, do we have a mockup ?

ashumilova:
no, we do not

vitaliy-guliy:
First steps. Animating loader with output area at the bottom.

![wsloader 01](https://user-images.githubusercontent.com/1655894/36032495-e0fd8962-0db6-11e8-9ba9-7e8778561eae.png)

Output can be maximized by simple clicking.
![wsloader 02](https://user-images.githubusercontent.com/1655894/36032639-532c154e-0db7-11e8-9842-23e5649ea328.png)

slemeur:
@vitaliy-guliy : Great progress.

Can you confirm: This loader is displayed while the container for the new IDE is being started and is displaying only the outputs of that container. Correct?

vitaliy-guliy:
@slemeur Yes, this loader is displayed while the workspace is loading and displaying machine logs.

benoitf:
@vitaliy-guliy what is displayed with openshift ?

vitaliy-guliy:
@benoitf I think only the loader will be displayed and no outputs. Openshift doesn't send any output to the client.

benoitf:
@vitaliy-guliy I have access to machine logs through openshift console so I suppose output are somewhere

benoitf:
@vitaliy-guliy but it's more a platform issue remark

PR #8714, opened 2018-02-12
UI for workspace-loader

vitaliy-guliy:
### What does this PR do?
Add UI for worspace loader app.

https://github.com/eclipse/che/issues/8682


wokspace-loader/src/index.ts
```diff
@@ -14,64 +14,77 @@ import "./style.css";
...
 
+    // for (let machineName in machines) {
+    //     let servers = machines[machineName].servers;
+    //     for (let serverId in servers) {
+    //         let attributes = servers[serverId].attributes;
+    //         if (attributes['type'] === 'ide') {
+    //             console.log('Open IDE: ', servers[serverId].url);
+    //         }
+    //     }
+    // }
```
evidolob: Can you remove or uncomment this?

wokspace-loader/src/index.ts
```diff
@@ -83,29 +96,22 @@ async function openIdeAfterRunning(ws: che.IWorkspace) {
...
 
     if (!ideUrl) {
-        ideUrl = ws.links.ide;
+        ideUrl = workspace.links.ide;
     }
 
-    console.log('IDE url is: ', ideUrl);
-    document.location.assign(ideUrl);
+    console.log('IDE URL: ', ideUrl);
+    // document.location.assign(ideUrl);
```
evidolob: Loader will never redirect to Theia or GWT IDE, remove comment on this line

evidolob (changes requested):

wokspace-loader/src/index.ts
vitaliy-guliy: Done

wokspace-loader/src/index.ts
vitaliy-guliy: Done

wokspace-loader/src/index.ts
```diff
@@ -14,64 +14,67 @@ const css = require('./style.css');
...
-        document.getElementById("output").appendChild(document.createTextNode(message.text));
-    };
-    let workspaceStatusHandler = (message: any) => {
-        console.log('workspace status', message);
-        let workspaceStatus = message.status;
-
-        if (workspaceStatus === 'RUNNING') {
-            openIdeAfterRunning(ws);
-        }
+var loader: Loader = new Loader();
```
evidolob: use `let` for variable or `const` if it never changes, type `Loader` is redundant there

vitaliy-guliy: Done

evidolob (approved):

wokspace-loader/src/index.html
```diff
@@ -18,13 +18,21 @@
...
     <% } %>
-    <script src="<%= htmlWebpackPlugin.options.urlPrefix %><%= htmlWebpackPlugin.files.js %>" type="text/javascript"></script>
+    <script src="<%= htmlWebpackPlugin.options.urlPrefix %><%= htmlWebpackPlugin.files.js %>" type="text/javascript" async="true"></script>
 </head>
 
-<body>
-    <main>
-        <div id="output"></div>
-    </main>
+<body style="background-color: #21252b; transition: background-color 0.5s ease;">
```
benoitf, line 24: why the style is in the html and not in the css ?

wokspace-loader/src/index.ts
```diff
@@ -14,64 +14,66 @@ const css = require('./style.css');
...
-        document.getElementById("output").appendChild(document.createTextNode(message.text));
-    };
-    let workspaceStatusHandler = (message: any) => {
-        console.log('workspace status', message);
-        let workspaceStatus = message.status;
-
-        if (workspaceStatus === 'RUNNING') {
-            openIdeAfterRunning(ws);
-        }
+function handleWorkspace(workspace: che.IWorkspace) {
```
benoitf, line 31: it's not related to this PR but it would be great to have such functions moved classes

wokspace-loader/src/index.html
vitaliy-guliy: It's for prevent blinking of the background. The page has grey background and new background color is used only after css file has been loaded. I played too much with this, only moving styles inside body resolves the problem.

wokspace-loader/src/index.ts
vitaliy-guliy: I don't understand pretty sure, where we should move those functions.

wokspace-loader/src/index.html
benoitf: I would say that the main issue is that the JS script should be loaded at the last item of body document, not in head
while css should be kept in `<head>` element

wokspace-loader/src/index.ts
benoitf: it looks like a WsMaster connection so it could be a WsMaster class or WebScoketWsMaster

wokspace-loader/src/loader/loader.ts
```diff
@@ -0,0 +1,53 @@
...
+     */
+    log(message: string): void {
+        let element = document.createElement("pre");
+        element.innerHTML = message;
+        document.getElementById("workspace-console-container").appendChild(element);
+        element.scrollIntoView();
+    }
+
+    onclick(event: Event): void {
+        console.log(">> click");
```
benoitf, line 42: I'm not sure we should keep this message

vitaliy-guliy: Removed, thanks.

Issue #8701, opened 2018-02-09
Launch Theia IDE inside Che workspace based on the provided list of plugins


ashumilova:
The aim is to define Theia IDE launcher, that will read the environment variables inside container - generate proper package.json with the list of defined plugins, build and run the application.

PR #8732, opened 2018-02-12
Theia docker file with launcher script 

evidolob:
### What does this PR do?
add Docker file with Theia and launcher script that can rebuild Theia with provided set of the plugin.

#8701


dockerfiles/theia/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM node:8-alpine
```
benoitf, line 10: could you add Copyright headers .

dockerfiles/theia/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+# build dependencies requireed to compile a custom Theia
+RUN apk add --no-cache make gcc g++ python git openssh bash
+WORKDIR /home/theia
+# build Theia with all extensions to persist yarn cache in the image and
+# have default Theia build in the workspace in case no plugins are requested
+ADD theia-full-package.json /home/theia/package.json
```
benoitf: it looks like there is a fork of upstream package.json, how to be in sync with upstream changes automatically ?

dockerfiles/theia/Dockerfile
evidolob: Done

dockerfiles/theia/Dockerfile
ghost: I think we can have two options here:

1. curl original Theia package.json as a Dockerfile instruction
2. Rely on our CI to do this after a triggered webhook.

benoitf: having the github URL of theia package.json will allow to only re-run the RUN instruction only if the remote side is updated

benoitf: (so we can reuse layer cache if there is no remote change)

dockerfiles/theia/theia-default-package.json
```diff
@@ -0,0 +1,17 @@
...
+        "@theia/preferences": "latest",
+        "@theia/git": "latest",
+        "@theia/file-search": "latest",
+        "@theia/markers": "latest",
+        "@theia/extension-manager": "latest"
+    },
+    "devDependencies": {
+        "@theia/cli": "latest"
+    }
+}
```
benoitf: could you please add missing eof line

evidolob: done

dockerfiles/theia/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN yarn && \
+    yarn theia build && \
+    rm -rf * && \
+    cd /home/default/theia && \
+    yarn && \
+    yarn theia build
+EXPOSE 3000
+ENV USE_LOCAL_GIT=true \
+    GITHUB_TOKEN=957257352ad57b7853c8b8714d2201bf4d337b5e
+#    ENV THEIA_PLUGINS=[\"@theia\/extension-manager\",\"\@theia\/python\"]
```
benoitf: should we remove this comment ?

benoitf (approved):

dockerfiles/theia/Dockerfile
evidolob: we leave it for testing purpose, but we also can use  `-e THEIA_PLUGINS=...` for this, so I remove it

benoitf: @evidolob you may add a README or in the header of Dockerfile on how to use that image

dockerfiles/theia/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+ADD src/main.js /theia_launcher/theia_launcher.js
+RUN yarn && \
+    yarn theia build && \
+    rm -rf * && \
+    cd /home/default/theia && \
+    yarn && \
+    yarn theia build
+EXPOSE 3000
+ENV USE_LOCAL_GIT=true \
+    GITHUB_TOKEN=957257352ad57b7853c8b8714d2201bf4d337b5e
```
riuvshin: I do not think that this is a good idea to put someone's token in dockerfile, I think it user should pass it's own GH token if that is necessary

benoitf: yes

ghost: Sure it's not good. But at this point we don't have any way to get user's token

benoitf:
based on demo, could it be possible to give 
```
THEIA_PLUGINS="@theia/extension-manager,@theia/python"
```
instead of
```
THEIA_PLUGINS=[\"@theia\/extension-manager\",\"\@theia\/python\"]
```

benoitf:
@eivantsov but humans are not good parsers :-)

eivantsov:
Easier for node script. So, we agreed to start with `["",""]`

evidolob:
@benoitf We never consider that user will manually edit that property, we should get good UI in dashboard for that

benoitf:
```
let myPlugins="@theia/extension-manager,@theia/python"
var array = myPlugins.split(',');
var jsonString = JSON.stringify(array);
console.log('json array is', jsonString);

==> json array is ["@theia/extension-manager","@theia/python"]


```

dockerfiles/theia/Dockerfile
ghost: https://github.com/roblourens/vscode-ripgrep/issues/4

l0rd: What about https://docs.docker.com/engine/reference/builder/#environment-replacement?

ghost: @l0rd how does that help? The scenario is as follows:

1. Theia image starts, we have a running container
2. A node script checks for Theia plugins env and once found runs `yarn` and `yarn theia build` against a generated or existing package.json.  It is during this build that a token is required.

Maybe I don't understand smth about env replacement.

l0rd: @eivantsov sorry I meant https://docs.docker.com/engine/reference/commandline/run/#set-environment-variables--e-env-env-file

riuvshin: @eivantsov we don't  need to get token, user should pass it 
if you need token in runtime  just pass it with `docker run -e GITHUB_TOKEN=`, if you need token for `docker build` you can do 
```
+ARG GITHUB_TOKEN
+ENV GITHUB_TOKEN ${GITHUB_TOKEN}
```

`docker build --build-arg GITHUB_TOKEN=<token>`

ghost: A user will just launch a workspace. It is us who do docker run, Che server I mean.

evidolob:
Now plugin list should provided as @benoitf proposed.
Example: `-e THEIA_PLUGINS=@theia/python,@theia/go`

dockerfiles/theia/Dockerfile
riuvshin: ahh, I thought it is image for CI, because all images we have in `dockerfiles/` are for CLI and they are supposed to be built on CI. if that is image for WS maybe it should be moved to github.com/eclipse/che-dockerfiles ?

eivantsov:
Even though this is a stack image, we will probably need it to be versioned and updated on a regular manner.

riuvshin (approved):

ashumilova (approved):

l0rd:
By the way @eivantsov it would be nice if all Che runtime stacks images could be versioned too. But that's another story :-)

PR #8736, opened 2018-02-12
Selenium: change a way for selecting an element from suggestions list

SkorikSergey:
### What does this PR do?
This PR modifies the **selectFileByName** method to send double-click on the found web element(not just double click).

https://github.com/eclipse/che/issues/8465

### DIFF
--- a/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/NavigateToFile.java
+++ b/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/NavigateToFile.java
@@ -11,6 +11,7 @@
 package org.eclipse.che.selenium.pageobject;
 
 import static java.lang.String.format;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static org.eclipse.che.selenium.core.constant.TestTimeoutsConstants.LOAD_PAGE_TIMEOUT_SEC;
 import static org.eclipse.che.selenium.core.constant.TestTimeoutsConstants.REDRAW_UI_ELEMENTS_TIMEOUT_SEC;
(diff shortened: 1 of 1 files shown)


selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/NavigateToFile.java
```diff
@@ -153,12 +154,15 @@ public void selectFileByFullName(String pathName) {
...
-        .click();
-    actionsFactory.createAction(seleniumWebDriver).doubleClick().perform();
+    WebElement webElement =
+        new WebDriverWait(seleniumWebDriver, REDRAW_UI_ELEMENTS_TIMEOUT_SEC)
+            .until(
+                visibilityOfElementLocated(
+                    By.xpath(format(Locators.FILE_NAME_LIST_SELECT, nameOfFile))));
+
+    webElement.click();
+    WaitUtils.sleepQuietly(500, MILLISECONDS);
```
dmytro-ndp, line 164: Please, add comment to explain why we should wait so long here.

dmytro-ndp:
ci-test

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/NavigateToFile.java
SkorikSergey: It is timeout between selecting element and double click on it. 0.5 sec is not so long.

dmytro-ndp: Yes, you are right.

dmytro-ndp (approved):

PR #8753, opened 2018-02-13
Fix orion document changing event getting text from native orion event instead of guessing it

sunix:
### What does this PR do?
Fix orion document changing event getting text from native orion event instead of guessing it.
"Changing" events were getting their new changed text from the document based on the new text range positions. While this way of doing works fine for "changed" events, it doesn't work for "changing" events as the change has not been performed yet to the document yet. The result is that the event send the previous text instead of the new added text. This fix is getting the new text from native orion event data object.


vparfonov (approved):

sunix:
@benoitf this has been reverted fyi

Issue #8557, opened 2018-02-01
Dto Generation is broken for Formatter Options


svor:
After applying code format (Edit -> Format or Alt+Shift+L) have an error

````
SEVERE: Unknown DTO type class org.eclipse.lsp4j.FormattingOptions
java.lang.IllegalArgumentException: Unknown DTO type class org.eclipse.lsp4j.FormattingOptions
	at Unknown.Xo(Throwable.java:66)
	at Unknown.ep(Exception.java:29)
	at Unknown.ip(RuntimeException.java:29)
	at Unknown.yYd(IllegalArgumentException.java:29)
	at Unknown.T2g(DtoFactory.java:113)
	at Unknown.Yfl(DtoFactory.java:38)
	at Unknown._fl(LanguageServerFormatter.java:137)
	at Unknown.O9g(OrionEditorPresenter.java:949)
	at Unknown.x5e(FormatterAction.java:55)
	at Unknown.Wvk(PopupMenu.java:562)
	at Unknown.jwk(PopupMenu.java:719)
	at Unknown.FBd(DOM.java:1481)
	at Unknown.iDd(DOMImplStandard.java:317)
	at Unknown._p(Impl.java:309)
	at Unknown.cq(Impl.java:361)
	at Unknown.anonymous(Impl.java:78)
	at Unknown.bC(StringValidator.java:66)
	at Unknown.aC(StringValidator.java:49)
	at Unknown.OB(RequestBuilder.java:292)
	at Unknown.VMk(AsyncRequest.java:72)
	at Unknown.ZMk(MachineAsyncRequestFactory.java:129)
	at Unknown.anonymous(JsPromise.java:116)
	at Unknown.Vll(OccurrencesProvider.java:92)
	at Unknown.Xll(OccurrencesProvider.java:85)
	at Unknown.anonymous(JsPromise.java:68)
	at Unknown.bC(StringValidator.java:66)
	at Unknown.aC(StringValidator.java:49)
	at Unknown.OB(RequestBuilder.java:292)
	at Unknown.VMk(AsyncRequest.java:72)
	at Unknown.ZMk(MachineAsyncRequestFactory.java:129)
	at Unknown.anonymous(JsPromise.java:116)
	at Unknown.Vll(OccurrencesProvider.java:92)
	at Unknown.Xll(OccurrencesProvider.java:85)
	at Unknown.anonymous(JsPromise.java:68)
	at Unknown.bC(StringValidator.java:66)
[...]

PR #8318, opened 2018-01-16
CHE-8304: Support DTO generating for new LSP4J constructs

jonahgraham:
Support for classes that extend a Map type and Either3.
See org.eclipse.lsp4j.FormattingOptions

### What does this PR do?

Fixes #8304: Adds support for Dto generation in new cases exposed by LSP4J 0.3.0+. Two specific cases are catered for, both exposed by the rewriting of [FormattingOptions](https://github.com/eclipse/lsp4j/blob/026cb14c2218f6e8a9fb1d8464f38b49a9d7177a/org.eclipse.lsp4j/src/main/xtend-gen/org/eclipse/lsp4j/FormattingOptions.java#L22) in https://github.com/eclipse/lsp4j/issues/99#issuecomment-313330347.

The new signature for the class is:
```java
public class FormattingOptions extends LinkedHashMap<String, Either3<String, Number, Boolean>> {}
```
The first issue is that FormattingOptions is of type Map, but does not have generics itself. This is resolved with a little heuristic that attempts to get the generic types from the superclass.

The second issue is the new Either3 type. New support has been written for Either3.

This allows, but also requires, Lsp4j to be upgraded to 0.3.0 or later (when Either3 was introduced). The upgrade is useful for new features, but more importantly, Debug Protocol implementation in #5508 requires it. As such a PR for che-dependencies is also included, see https://github.com/eclipse/che-dependencies/pull/93

### What issues does this PR fix or reference?

Fixes #8304 

#### Release Notes

The version of LSP4J has been upgraded to 0.3.0.

Issue #5508, opened 2017-06-30
VS Code debug protocol support on Che


ibuziuk:
Investigate what are the requirements, tasks areas and impediments to integrate VS Code debug protocol [1] to Che. 
Create more detailed tasks on how to continue.
[1] https://github.com/Microsoft/vscode-debugadapter-node

jonahgraham:
At EclipseCon Europe we (Kichwa Coders) demonstrated debug servers running in Che based on VS Code debug protocol. We will soon be providing a PR for Che to add this support.

jonahgraham:
I posted a question on [che-dev](https://dev.eclipse.org/mhonarc/lists/che-dev/msg02454.html) asking about upgrading to 0.4.0 of LSP4J. That upgrade is a dep of this bug. Hopefully it is on someone's todo list :-)

skabashnyuk:
@tolusha @tsmaeder are you interested in this issue?

gorkem:
Closing as this is done for Che 7 and Che 6 is not active anymore

PR #5292, opened 2017-06-07
Update lsp4J lib to newer version.

AndrienkoAleksandr:
### What does this PR do?
Update lsp4j to 0.2.0 version

PR #8784, opened 2018-02-14
CHE-8557: No Dto available for FormattingOptions

jonahgraham:
The Dto isn't needed for FormattingOptions as it is really a specialized
Map and the types that contain a FormattingOptions field handle
the field as a Map during JSON serialize/deserialize

Signed-off-by: Jonah Graham <jonah@kichwacoders.com>

### What does this PR do?

Fixes formatting when using Language server

#8557

### DIFF
--- a/core/che-core-api-dto/pom.xml
+++ b/core/che-core-api-dto/pom.xml
@@ -47,6 +47,10 @@
             <artifactId>che-core-commons-lang</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.eclipse.lsp4j</groupId>
+            <artifactId>org.eclipse.lsp4j.jsonrpc</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.reflections</groupId>
             <artifactId>reflections</artifactId>
         </dependency>
--- a/core/che-core-api-dto/pom.xml
+++ b/core/che-core-api-dto/pom.xml
@@ -47,10 +47,6 @@
             <artifactId>che-core-commons-lang</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.eclipse.lsp4j</groupId>
-            <artifactId>org.eclipse.lsp4j.jsonrpc</artifactId>
-        </dependency>
-        <dependency>
             <groupId>org.reflections</groupId>
             <artifactId>reflections</artifactId>
         </dependency>
--- a/core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java
+++ b/core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java
@@ -33,6 +33,7 @@ import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.ServiceLoader;
@@ -41,7 +42,6 @@ import org.eclipse.che.commons.lang.reflect.ParameterizedTypeImpl;
 import org.eclipse.che.dto.shared.DTO;
 import org.eclipse.che.dto.shared.JsonArray;
 import org.eclipse.che.dto.shared.JsonStringMap;
-import org.eclipse.lsp4j.jsonrpc.json.adapters.EitherTypeAdapterFactory;
 
 /**
  * Provides implementations of DTO interfaces.
@@ -102,13 +102,11 @@ public final class DtoFactory {
   // It helps avoid reflection when need create copy of exited DTO instance.
   private final Map<Class<?>, DtoProvider<?>> dtoImpl2Providers = new ConcurrentHashMap<>();
   private final Gson dtoGson =
-      new GsonBuilder()
-          .registerTypeAdapterFactory(
-              new NullAsEmptyTAF<>(Collection.class, Collections.emptyList()))
-          .registerTypeAdapterFactory(new NullAsEmptyTAF<>(Map.class, Collections.emptyMap()))
-          .registerTypeAdapterFactory(new DtoInterfaceTAF())
-          .registerTypeAdapterFactory(new EitherTypeAdapterFactory())
-          .create();
+      buildDtoParser(
+          ServiceLoader.load(TypeAdapterFactory.class).iterator(),
+          new NullAsEmptyTAF<>(Collection.class, Collections.emptyList()),
+          new NullAsEmptyTAF<>(Map.class, Collections.emptyMap()),
+          new DtoInterfaceTAF());
 
   /**
    * Created deep copy of DTO object.
@@ -492,4 +490,19 @@ public final class DtoFactory {
   }
 
   private DtoFactory() {}
+
+  private static Gson buildDtoParser(
+      Iterator<TypeAdapterFactory> factoryIterator, TypeAdapterFactory... factories) {
+    GsonBuilder builder = new GsonBuilder();
+
+    for (Iterator<TypeAdapterFactory> it = factoryIterator; it.hasNext(); ) {
+      TypeAdapterFactory factory = it.next();
+      builder.registerTypeAdapterFactory(factory);
+    }
+
+    for (TypeAdapterFactory factory : factories) {
+      builder.registerTypeAdapterFactory(factory);
+    }
+    return builder.create();
+  }
 }
--- a/core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java
+++ b/core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java
@@ -41,6 +41,7 @@ import org.eclipse.che.commons.lang.reflect.ParameterizedTypeImpl;
 import org.eclipse.che.dto.shared.DTO;
 import org.eclipse.che.dto.shared.JsonArray;
 import org.eclipse.che.dto.shared.JsonStringMap;
+import org.eclipse.lsp4j.jsonrpc.json.adapters.EitherTypeAdapterFactory;
 
 /**
  * Provides implementations of DTO interfaces.
@@ -106,6 +107,7 @@ public final class DtoFactory {
               new NullAsEmptyTAF<>(Collection.class, Collections.emptyList()))
           .registerTypeAdapterFactory(new NullAsEmptyTAF<>(Map.class, Collections.emptyMap()))
           .registerTypeAdapterFactory(new DtoInterfaceTAF())
+          .registerTypeAdapterFactory(new EitherTypeAdapterFactory())
           .create();
 
   /**
--- a/plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/LanguageServerFormatter.java
+++ b/plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/LanguageServerFormatter.java
@@ -188,7 +188,7 @@ public class LanguageServerFormatter implements ContentFormatter {
   }
 
   private FormattingOptions getFormattingOptions() {
-    FormattingOptions options = dtoFactory.createDto(FormattingOptions.class);
+    FormattingOptions options = new FormattingOptions();
     options.setInsertSpaces(Boolean.parseBoolean(getEditorProperty(EditorProperties.EXPAND_TAB)));
     options.setTabSize(Integer.parseInt(getEditorProperty(EditorProperties.TAB_SIZE)));
     return options;
--- a/wsagent/che-core-api-languageserver-maven-plugin/src/main/java/org/eclipse/che/api/languageserver/generator/FromJsonGenerator.java
+++ b/wsagent/che-core-api-languageserver-maven-plugin/src/main/java/org/eclipse/che/api/languageserver/generator/FromJsonGenerator.java
@@ -87,7 +87,8 @@ public class FromJsonGenerator extends ConversionGenerator {
   }
 
   private void generateMapConversion(
-      String indent, PrintWriter out, String varName, String jsonValName, Type paramType) {
+      String indent, PrintWriter out, String varName, String jsonValName, Type inputParamType) {
+    Type paramType = inputParamType;
     if (!(paramType instanceof ParameterizedType)) {
       paramType = ((Class<?>) paramType).getGenericSuperclass();
     }
@@ -95,11 +96,18 @@ public class FromJsonGenerator extends ConversionGenerator {
     Type containedType = genericType.getActualTypeArguments()[1];
     String objectName = varName + "o";
     String containedName = objectName + "X";
+    String typeName = inputParamType.getTypeName();
+    String instantiateTypeName = typeName;
+    if (getRawClass(inputParamType).isInterface()) {
+      if (inputParamType instanceof ParameterizedType) {
+        instantiateTypeName = String.format("HashMap<String, %1$s>", containedType.getTypeName());
+      } else {
+        throw new RuntimeException(
+            "Unsupported Map Conversion. Generator needs to be updated for new LSP4J construct");
+      }
+    }
     out.println(
-        indent
-            + String.format(
-                "HashMap<String, %1$s> %2$s= new HashMap<String, %3$s>();",
-                containedType.getTypeName(), varName, containedType.getTypeName()));
+        indent + String.format("%1$s %2$s= new %3$s();", typeName, varName, instantiateTypeName));
     out.println(
         indent
             + String.format(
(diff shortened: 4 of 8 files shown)


jonahkichwacoders:
@svor Can you review/test please?

svor:
@jonahkichwacoders 
Thank you for quick response! These changes fix the first problem which you describe [here](https://github.com/eclipse/che/issues/8557#issuecomment-365731778). 
But after that I got the second problem :-)
selection_021

jonahkichwacoders:
@svor Can you provide me some instructions to test? I don't seem to have any failing tests at the moment and I can't get the error message you are getting. Which language server is it? Thanks!

dkuleshov:
Hi,

@jonahkichwacoders Could you please dispel my concerns about something that I see in this line: https://github.com/eclipse/che/blob/master/ide/che-core-orion-editor/src/main/java/org/eclipse/che/ide/editor/orion/client/OrionEditorBuilder.java#L36

If I understand correctly, we use `org.eclipse.che.ide.api.editor.editorconfig.DefaultTextEditorConfiguration` for the editor while your changes impact `org.eclipse.che.plugin.languageserver.ide.editor.LanguageServerFormatter` which is used in cases where we initialize `org.eclipse.che.plugin.languageserver.ide.editor.LanguageServerEditorConfiguration`. If that's true I guess your changes and the code related to formatting that we're dealing here with is never used. How can you be sure that it works well, do we have tests for that or something that can prove it's consistency? 

@svor please correct me if my concerns are in vain.

Thanks.

svor:
@jonahkichwacoders 
The json which is sent from the client to the server is like
```
{
  "options": {
    "insertSpaces": true,
    "tabSize": 4
  },
  "textDocument": {
    "uri": "/console-java-simple/src/main/java/org/eclipse/che/examples/A.java"
  }
}
```
I've created a test to see the error. This is the patch:
```
Index: wsagent/che-core-api-languageserver/src/test/java/org/eclipse/che/api/languageserver/dto/DtoConversionTest.java
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- wsagent/che-core-api-languageserver/src/test/java/org/eclipse/che/api/languageserver/dto/DtoConversionTest.java	(revision 2b9c7413a3e8555b627399b547a5bb1df02fecbf)
+++ wsagent/che-core-api-languageserver/src/test/java/org/eclipse/che/api/languageserver/dto/DtoConversionTest.java	(revision )
@@ -10,6 +10,8 @@
  */
 package org.eclipse.che.api.languageserver.dto;
 
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
 import java.lang.reflect.Array;
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
@@ -25,7 +27,9 @@
 import org.eclipse.che.api.languageserver.server.dto.DtoServerImpls.WorkspaceEditDto;
 import org.eclipse.che.api.languageserver.shared.model.ExtendedCompletionItem;
 import org.eclipse.che.api.languageserver.shared.model.ExtendedCompletionList;
 import org.eclipse.lsp4j.CompletionItem;
+import org.eclipse.lsp4j.DocumentFormattingParams;
 import org.eclipse.lsp4j.Hover;
 import org.eclipse.lsp4j.MarkedString;
 import org.eclipse.lsp4j.ParameterInformation;
@@ -83,6 +87,17 @@
     Assert.assertTrue(reflectionEquals(originalDto, convertedDto));
   }
 
+  @Test
+  public void testDocumentFormattingParamsDeserializer() throws Exception {
+    JsonObject json = (JsonObject) new JsonParser().parse(
+        "{\"options\": {\"insertSpaces\": true,\"tabSize\": 4},\"textDocument\": {\"uri\": \"/console-java-simple/src/main/java/org/eclipse/che/examples/A.java\"}}");
+
+    DocumentFormattingParams params = DtoFactory.getInstance()
+        .createDtoFromJson(json.toString(), DocumentFormattingParams.class);
+
+    Assert.assertTrue(params.getOptions().isInsertSpaces());
+  }
+
   @Test
   public void testEitherConversion() {
     Either<String, MarkedString> either1 = Either.forLeft("foobar");
```

jonahkichwacoders:
@dkuleshov my original patch, #8318 was to support using LSP4J 0.4.0 so that Debug Sever Protocol  could be supported, part of #5508. I can't comment/don't know about the Orion editor.

@svor Let me look into it.

dkuleshov:
@jonahkichwacoders I see, thank you for the quick response.

By a mistake (and I am ashamed for that) I must have missed that for https://github.com/eclipse/che/pull/8318 - I see that there is a bunch of a new functionality, but I don't see the way we can appropriately test it. That is why I'm eager to know how we can ensure consistency and quality here. Could you please comment if you have any unit or integration tests for that purpose, or perhaps you can advise another way how we can make sure that solutions presented in actual pull request and in https://github.com/eclipse/che/pull/8318 work well?

jonahkichwacoders:
@dkuleshov I too am sorry. I didn't realize this change was not covered by existing tests. The Debug's need for newer LSP4J did not really need other parts of Che updated, but with a single classloader there was no option. (I have been spoiled by years of Eclipse IDE development where independent changes can be truly independent thanks to OSGi.)

So, how to test? I don't know, I guess we are talking about retrofitting tests at the moment.

jonahkichwacoders:
@svor A few things all going wrong together here. 

- #3103 changed deserializing to stop using the Dto generator and use Gson to deserialize. Problem is Gson and generated Dtos don't create the exact same JSON. This means that calling `org.eclipse.che.dto.server.DtoFactory.createDtoFromJson(String, Class<T>)` vs `org.eclipse.che.dto.server.DtoFactory.createDtoFromJson(JsonElement, Class<T>)` uses completely different deserializing. The former uses Gson, the latter uses the Dto's implementation (`org.eclipse.che.dto.server.DtoProvider.fromJson(JsonElement)`). This is ok for most everything, but stuff like lsp4j has special encoding/decoding for handling fields of `Either` type

That means this test works (but see problem 2):

```
  @Test
  public void testHover() throws Exception {
    Hover hover = DtoFactory.getInstance().createDto(Hover.class);
    MarkedString markedString = DtoFactory.getInstance().createDto(MarkedString.class);
    markedString.setLanguage("Language");
    markedString.setValue("Value");
    List<Either<String, MarkedString>> list = new ArrayList<>();
    list.add(Either.forRight(markedString));
    list.add(Either.forLeft("normal String"));
    hover.setContents(list);

    JsonElement json2 = DtoFactory.getInstance().toJsonElement(hover);

//    Hover params =
//        DtoFactory.getInstance().createDtoFromJson(json2.toString(), Hover.class);
    Hover params =
        DtoFactory.getInstance().createDtoFromJson(json2, Hover.class);

    Assert.assertTrue(params.getContents().get(0).isRight());
    Assert.assertEquals("Value", params.getContents().get(0).getRight().getValue());
    Assert.assertEquals("Language", params.getContents().get(0).getRight().getLanguage());
    Assert.assertTrue(params.getContents().get(1).isLeft());
    Assert.assertEquals("normal String", params.getContents().get(1).getLeft());
  }

```

but this one does not:

```
  @Test
  public void testHover() throws Exception {
    Hover hover = DtoFactory.getInstance().createDto(Hover.class);
    MarkedString markedString = DtoFactory.getInstance().createDto(MarkedString.class);
    markedString.setLanguage("Language");
    markedString.setValue("Value");
    List<Either<String, MarkedString>> list = new ArrayList<>();
    list.add(Either.forRight(markedString));
    list.add(Either.forLeft("normal String"));
    hover.setContents(list);

    JsonElement json2 = DtoFactory.getInstance().toJsonElement(hover);

    Hover params =
        DtoFactory.getInstance().createDtoFromJson(json2.toString(), Hover.class);
//    Hover params =
//        DtoFactory.getInstance().createDtoFromJson(json2, Hover.class);

    Assert.assertTrue(params.getContents().get(0).isRight());
    Assert.assertEquals("Value", params.getContents().get(0).getRight().getValue());
    Assert.assertEquals("Language", params.getContents().get(0).getRight().getLanguage());
    Assert.assertTrue(params.getContents().get(1).isLeft());
    Assert.assertEquals("normal String", params.getContents().get(1).getLeft());
  }
```

The fix for this is to configure the Gson object to understand LSP4J serializing/deserializing (by registering the appropriate type adapters, see `org.eclipse.lsp4j.jsonrpc.json.MessageJsonHandler.getDefaultGsonBuilder()`) or by ensuring the dto generation creates JSON in the format Gson is configured to see.

- The `org.eclipse.che.api.languageserver.util.EitherUtil.matches(JsonElement, JsonDecision[])` is broken by #5292. Some code cleanup was done and it now returns true instead of false when JsonDecision is string but the element is actually an object. There are two copies of this file in the code base, one has the mistake, the other is in the state prior to #5292 so does not.

```
$ find . -name EitherUtil.java | grep org/eclipse/che/api/languageserver/util/EitherUtil.java
./wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/util/EitherUtil.java
./plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/api/languageserver/util/EitherUtil.java
```

- The third problem is that we generate some mistaken code cause a class cast exception when passing the json element rather than that json string to `DtoFactory.createDtoFromJson`. This was caused by #8318.

---

For problem 1 I am not sure of the general solution, but in this one use case I can provide a fix. For 2 and 3 I'll provide an update soon.

jonahkichwacoders:
> For problem 1 I am not sure of the general solution, but in this one use case I can provide a fix. For 2 and 3 I'll provide an update soon.

Actually take that back. I can provide fix for 2 and 3.

The other alternative is to roll back #8318 as that exposed these problems. Once you need the extra functionality you can reapply and fix the inconsistent json handling around lsp4j.

jonahkichwacoders:
Please don't push this! The last commit will cause compiler errors (as opposed to runtime exceptions). 

@svor I am not around tomorrow (Friday 16/Feb) so I won't be able to look at this again for a few days. Feel free to revert #8318 (and https://github.com/eclipse/che-dependencies/pull/93) until a correct fix for upgrading LSP4J can be obtained.

dkuleshov:
@jonahkichwacoders Understood, thanks. I guess we can't do a lot for current problem, but I guess we eventually will have to develop more convenient and transparent testing workflows to avoid unnecessary problems in the future. However that is obviously out of the scope of this pull request.

tolusha:
@jonahkichwacoders Any news

jonahkichwacoders:
@tolusha Thanks for the ping.

Rebased and pushed my set of fixes. This does not fix all LSP4J dto issues, but fixes all the formatting ones. The issues with hover that was there is still there. I recommend reviewing the use of casts in the dto generated code as the casts are converting compile time errors to runtime ones still. But AFAICT those parts of LSP are not yet in use (e.g. org.eclipse.lsp4j.TextDocumentRegistrationOptions)

tolusha:
@jonahkichwacoders 
Thank you, it works

@dkuleshov 
@skabashnyuk 
Review that patch pls

tolusha (approved):

tolusha:
ci-build

core/che-core-api-dto/pom.xml
```diff
@@ -46,6 +46,10 @@
             <groupId>org.eclipse.che.core</groupId>
             <artifactId>che-core-commons-lang</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.lsp4j</groupId>
+            <artifactId>org.eclipse.lsp4j.jsonrpc</artifactId>
```
skabashnyuk: I'm kind of -1 for this dependency. DTO is supposed to be feature agnostic.
Can we avoid to use it here? Maybe we should make dto mechanism extensible so you can override it?

core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java
```diff
@@ -106,6 +107,7 @@ public Gson getGson() {
               new NullAsEmptyTAF<>(Collection.class, Collections.emptyList()))
           .registerTypeAdapterFactory(new NullAsEmptyTAF<>(Map.class, Collections.emptyMap()))
           .registerTypeAdapterFactory(new DtoInterfaceTAF())
+          .registerTypeAdapterFactory(new EitherTypeAdapterFactory())
```
skabashnyuk: Why you need this change?

core/che-core-api-dto/pom.xml
jonahgraham: (Answering this and next comment together.)

We need this because at the moment Dto generation for LSP4J handles the Either class with special handling. It "looks inside" Eithers and stores the active one. This is compatible with how LSP4J serializes and deserializes, but only if EitherTypeAdapterFactory is used. 

AFAICT, based on the contents of #3103, originally Dto handled the serialize and deserialize all the time, i.e. whatever Dto wrote to the stream, Dto would read from the stream so Dto only had to be consistent with itself. But #3103 added using Gson as an option to doing the deserialize (depending on which variant of DtoFactory.createDtoFromJson is called). 

The alternative solutions to this problem:

- Don't use Gson for deserialize when doing LSP4J (is there the ability to provide this fine grained control of which createDtoFromJson is called?)
- Change LSP4J Dto creation to stop handling LSP4J's Either in a special way (I don't know if this is possible, can Gson handle the semantics of Either without an adapter?)
- Make it extensible (E.g. a method to register additional adapter factories). for example register adapter factory at the same time as registering providers. If new factories are added the gson instance has to be replaced, or the gson instance creation has to be delayed until all factories are registered

core/che-core-api-dto/src/main/java/org/eclipse/che/dto/server/DtoFactory.java
jonahgraham: See above answer.

skabashnyuk:
@tolusha @svor @vparfonov  as discussed I can propose to do something like this

```
@@ -101,12 +102,11 @@ public final class DtoFactory {
   // It helps avoid reflection when need create copy of exited DTO instance.
   private final Map<Class<?>, DtoProvider<?>> dtoImpl2Providers = new ConcurrentHashMap<>();
   private final Gson dtoGson =
-      new GsonBuilder()
-          .registerTypeAdapterFactory(
-              new NullAsEmptyTAF<>(Collection.class, Collections.emptyList()))
-          .registerTypeAdapterFactory(new NullAsEmptyTAF<>(Map.class, Collections.emptyMap()))
-          .registerTypeAdapterFactory(new DtoInterfaceTAF())
-          .create();
+      buildDtoParser(
+          ServiceLoader.load(TypeAdapterFactory.class).iterator(),
+          new NullAsEmptyTAF<>(Collection.class, Collections.emptyList()),
+          new NullAsEmptyTAF<>(Map.class, Collections.emptyMap()),
+          new DtoInterfaceTAF());

   /**
    * Created deep copy of DTO object.
@@ -490,4 +490,19 @@ public final class DtoFactory {
   }

   private DtoFactory() {}
+
+  private static Gson buildDtoParser(
+      Iterator<TypeAdapterFactory> factoryIterator, TypeAdapterFactory... factories) {
+    GsonBuilder builder = new GsonBuilder();
+
+    for (Iterator<TypeAdapterFactory> it = factoryIterator; it.hasNext(); ) {
+      TypeAdapterFactory factory = it.next();
+      builder.registerTypeAdapterFactory(factory);
+    }
+
+    for (TypeAdapterFactory factory : factories) {
+      builder.registerTypeAdapterFactory(factory);
+    }
+    return builder.create();
+  }
 }

```

wdyt?
In this way we can avoid adding lsp specific dependency to core

tolusha:
+1

core/che-core-api-dto/pom.xml
skabashnyuk: >Make it extensible

See my comment below.

svor:
agree

svor:
@skabashnyuk Are you going to merge your changes to master branch?

skabashnyuk:
No. I hope authors of this pr will test it. it was just a prototype.

svor:
@jonahkichwacoders 
Could you please try [this](https://github.com/eclipse/che/pull/8784#issuecomment-369213686) way of registering `TypeAdapterFactory` to avoid adding `org.eclipse.lsp4j.jsonrpc` dependency?

jonahkichwacoders:
@svor Sure, I can have a look.

jonahkichwacoders:
@svor I have been very busy this week, I will try to get to this in the next few days. Feel free to apply the proposed [solution](https://github.com/eclipse/che/pull/8784#issuecomment-369213686) to the branch if this is more urgent than I can get to it.

svor:
Hello @jonahkichwacoders 
I've made a PR to your branch https://github.com/jonahkichwacoders/che/pull/1
Please take a look.

svor:
Thanks @jonahkichwacoders 
@skabashnyuk Can you take another look at this PR
@dkuleshov @vparfonov Please review these changes

svor (approved):

skabashnyuk (approved):

tolusha:
ci-test

vparfonov (approved):

evidolob (approved):

dkulieshov (approved):

tsmaeder (approved):
lgtm

svor:
@musienko-maxim Can you help to analyse test report please?

musienko-maxim:
The selenium session did not find new bugs

svor:
@jonahkichwacoders so we are ready to merge. Maybe you want something to add?

jonahkichwacoders:
I'm good. Thanks.

Issue #8792, opened 2018-02-15
Unexpected fail of CreateAndDeleteProjectsTest selenium test 


SkorikSergey:
### Description
We have unexpected fail of **CreateAndDeleteProjectsTest** selenium test.  The test, sometimes, fails while trying to delete a project but projects settings page is empty. We should clarify reason and fix it.

**Screenshot**
![org eclipse che selenium dashboard createanddeleteprojectstest createanddeleteprojecttest_db8py4hr](https://user-images.githubusercontent.com/7760565/36253205-90127ede-124f-11e8-9d8d-351d4638f6e7.png)

**Log:**
```
org.openqa.selenium.TimeoutException: 
Expected condition failed: waiting for visibility of element located by By.xpath: //button/span[text()='Delete'] (tried for 5 second(s) with 500 MILLISECONDS interval)
Build info: version: '3.0.1', revision: '1969d75', time: '2016-10-18 09:49:13 -0700'
System info: host: 'slave6.codenvycorp.com', ip: '127.0.0.1', os.name: 'Linux', os.arch: 'amd64', os.version: '3.10.0-514.16.1.el7.x86_64', java.version: '1.8.0_102'
Driver info: org.eclipse.che.selenium.core.SeleniumWebDriver
	at org.eclipse.che.selenium.dashboard.CreateAndDeleteProjectsTest.createAndDeleteProjectTest(CreateAndDeleteProjectsTest.java:103)
```

**Report with reproduced problem:**
https://ci.codenvycorp.com/view/qa/job/che-integration-tests-multiuser-master-ocp/118/Selenium_tests_report/

SkorikSergey:
It seems that this problem is similar to https://github.com/eclipse/che/issues/6761 but reproduced randomly.

SkorikSergey:
Created issue for this bug https://github.com/eclipse/che/issues/8931.

PR #8794, opened 2018-02-15
Selenium: add info about known issue

SkorikSergey:
### What does this PR do?
This PR adds info about known issue to **CreateAndDeleteProjectsTest** selenium test and **WorkspaceProjects** page object.


dmytro-ndp (approved):
It would be quite useful to see in commit history and description of PR the exact id #8792 of issue information about which is being added to test

PR #8805, opened 2018-02-15
Make CHE 6 multi-tenant compatible for OSIO

davidfestal:
### What does this PR do?

This PR makes a number or required changes to have Che6 be used in OSIO :

1. Copy the boostrapper config file to POD *installer by installer* to avoid executing a huge command, which in turn tries to send a huge GET request (73kb-long URL). This long GET request was previously not supported by the fabric8 OSO proxy.
See openshiftio/openshift.io#2254 for mor details.
This change is implemented by commit https://github.com/eclipse/che/pull/8805/commits/f992e247f4699503dd731b7c53506c306aef4847

2. Rework the `KubernetesClientFactory` and `OpenshiftClientFactory` objects to:
- share a single `OkHttpClient` that is the basis of all created
Kubernetes or Openshift clients potentially pointing to
different.cluster URL with different authorization tokens
- provide the workspaceId in the `create` methods whenever it's
available (workspace start / stop, idling, etc ...)
- have distinct methods for creating the Kubernetes client and the
Openshift client (required since the Openshift client creates connection
leaks on some calls such as exec of POD watch)
This is the implementation of issue
redhat-developer/rh-che#516

3. Add the `userId` into the SPI `RuntimeIdentity` object. Currently,
only the `userName` is available in this object that gathers information
about the identity attached to a workspace runtime. This change is
required because the `userId` should be accessible from the
`workspaceId` for implementing multi-tenancy and creating workspace
resources in a user-specific location.
This is the implementation of issue
redhat-developer/rh-che#501

Changes 2 and 3 are implemented by commit eclipse/che@31156f7

openshiftio/openshift.io#2254
redhat-developer/rh-che#516
redhat-developer/rh-che#516

### DIFF
--- a/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/runtime/RuntimeIdentity.java
+++ b/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/runtime/RuntimeIdentity.java
@@ -16,5 +16,7 @@ public interface RuntimeIdentity {
 
   String getEnvName();
 
-  String getOwner();
+  String getOwnerName();
+
+  String getOwnerId();
 }
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
@@ -148,6 +148,35 @@ public class KubernetesClientFactory {
     return defaultConfig;
   }
 
+  protected Interceptor buildKubernetesInterceptor(Config config) {
+    return new Interceptor() {
+      @Override
+      public Response intercept(Chain chain) throws IOException {
+        Request request = chain.request();
+        if (isNotNullOrEmpty(config.getUsername()) && isNotNullOrEmpty(config.getPassword())) {
+          Request authReq =
+              chain
+                  .request()
+                  .newBuilder()
+                  .addHeader(
+                      "Authorization",
+                      Credentials.basic(config.getUsername(), config.getPassword()))
+                  .build();
+          return chain.proceed(authReq);
+        } else if (isNotNullOrEmpty(config.getOauthToken())) {
+          Request authReq =
+              chain
+                  .request()
+                  .newBuilder()
+                  .addHeader("Authorization", "Bearer " + config.getOauthToken())
+                  .build();
+          return chain.proceed(authReq);
+        }
+        return chain.proceed(request);
+      }
+    };
+  }
+
   protected void doCleanup() {
     ConnectionPool connectionPool = httpClient.connectionPool();
     Dispatcher dispatcher = httpClient.dispatcher();
@@ -177,37 +206,7 @@ public class KubernetesClientFactory {
         httpClient.newBuilder().authenticator(Authenticator.NONE).build();
     OkHttpClient.Builder builder = clientHttpClient.newBuilder();
     builder.interceptors().clear();
-    clientHttpClient =
-        builder
-            .addInterceptor(
-                new Interceptor() {
-                  @Override
-                  public Response intercept(Chain chain) throws IOException {
-                    Request request = chain.request();
-                    if (isNotNullOrEmpty(config.getUsername())
-                        && isNotNullOrEmpty(config.getPassword())) {
-                      Request authReq =
-                          chain
-                              .request()
-                              .newBuilder()
-                              .addHeader(
-                                  "Authorization",
-                                  Credentials.basic(config.getUsername(), config.getPassword()))
-                              .build();
-                      return chain.proceed(authReq);
-                    } else if (isNotNullOrEmpty(config.getOauthToken())) {
-                      Request authReq =
-                          chain
-                              .request()
-                              .newBuilder()
-                              .addHeader("Authorization", "Bearer " + config.getOauthToken())
-                              .build();
-                      return chain.proceed(authReq);
-                    }
-                    return chain.proceed(request);
-                  }
-                })
-            .build();
+    clientHttpClient = builder.addInterceptor(buildKubernetesInterceptor(config)).build();
 
     return new UnclosableKubernetesClient(clientHttpClient, config);
   }
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
@@ -69,8 +69,12 @@ public class KubernetesClientFactory {
   }
 
   /**
-   * Creates instance of {@link KubernetesClient}.
+   * Creates an instance of {@link KubernetesClient} that can be used to perform any operation
+   * related to a given workspace. </br> For all operations performed in the context of a given
+   * workspace (workspace start, workspace stop, etc ...), this method should be used to retrieve a
+   * Kubernetes client.
    *
+   * @param workspaceId Identifier of the workspace on which Kubernetes operations will be performed
    * @throws InfrastructureException if any error occurs on client instance creation.
    */
   public KubernetesClient create(String workspaceId) throws InfrastructureException {
@@ -80,7 +84,10 @@ public class KubernetesClientFactory {
   }
 
   /**
-   * Creates instance of {@link KubernetesClient}.
+   * Creates an instance of {@link KubernetesClient} that can be used to perform any operation
+   * <strong>that is not related to a given workspace</strong>. </br> For all operations performed
+   * in the context of a given workspace (workspace start, workspace stop, etc ...), the {@code
+   * create(String workspaceId)} method should be used to retrieve a Kubernetes client.
    *
    * @throws InfrastructureException if any error occurs on client instance creation.
    */
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
@@ -18,13 +18,17 @@ import io.fabric8.kubernetes.client.ConfigBuilder;
 import io.fabric8.kubernetes.client.DefaultKubernetesClient;
 import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.kubernetes.client.utils.HttpClientUtils;
+import io.fabric8.kubernetes.client.utils.Utils;
 import java.io.IOException;
+import java.util.concurrent.ExecutorService;
 import javax.annotation.PreDestroy;
 import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
 import okhttp3.Authenticator;
+import okhttp3.ConnectionPool;
 import okhttp3.Credentials;
+import okhttp3.Dispatcher;
 import okhttp3.Interceptor;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
@@ -43,7 +47,13 @@ public class KubernetesClientFactory {
 
   private static final Logger LOG = LoggerFactory.getLogger(KubernetesClientFactory.class);
 
+  /** {@link OkHttpClient} instance shared by all Kubernetes clients. */
   private OkHttpClient httpClient;
+
+  /**
+   * Default Kubernetes {@link Config} that will be the base configuration to create per-workspace
+   * configurations.
+   */
   private Config defaultConfig;
 
   @Inject
@@ -56,15 +66,6 @@ public class KubernetesClientFactory {
     this.defaultConfig =
         buildDefaultConfig(masterUrl, username, password, oauthToken, doTrustCerts);
     this.httpClient = HttpClientUtils.createHttpClient(defaultConfig);
-  }
-
-  /**
-   * Builds the Kubernetes {@link Config} object based on a default {@link Config} object and an
-   * optional workspace Id.
-   */
-  public Config buildConfig(Config defaultConfig, @Nullable String workspaceId)
-      throws InfrastructureException {
-    return defaultConfig;
   }
 
   /**
@@ -132,7 +133,35 @@ public class KubernetesClientFactory {
   }
 
   /**
-   * Creates instance of {@link KubernetesClient}.
+   * Builds the Kubernetes {@link Config} object based on a default {@link Config} object and an
+   * optional workspace Id.
+   */
+  protected Config buildConfig(Config defaultConfig, @Nullable String workspaceId)
+      throws InfrastructureException {
+    return defaultConfig;
+  }
+
+  protected void doCleanup() {
+    ConnectionPool connectionPool = httpClient.connectionPool();
+    Dispatcher dispatcher = httpClient.dispatcher();
+    ExecutorService executorService =
+        httpClient.dispatcher() != null ? httpClient.dispatcher().executorService() : null;
+
+    if (dispatcher != null) {
+      dispatcher.cancelAll();
+    }
+
+    if (connectionPool != null) {
+      connectionPool.evictAll();
+    }
+
+    Utils.shutdownExecutorService(executorService);
+  }
+
+  /**
+   * Creates instance of {@link KubernetesClient} that uses an {@link OkHttpClient} instance derived
+   * from the shared {@code httpClient} instance in which interceptors are overriden to authenticate
+   * with the credentials (user/password or Oauth token) contained in the {@code config} parameter.
    *
    * @throws InfrastructureException if any error occurs on client instance creation.
    */
@@ -179,14 +208,7 @@ public class KubernetesClientFactory {
   @PreDestroy
   private void cleanup() {
     try {
-      if (httpClient.connectionPool() != null) {
-        httpClient.connectionPool().evictAll();
-      }
-      if (httpClient.dispatcher() != null
-          && httpClient.dispatcher().executorService() != null
-          && !httpClient.dispatcher().executorService().isShutdown()) {
-        httpClient.dispatcher().executorService().shutdown();
-      }
+      doCleanup();
     } catch (RuntimeException ex) {
       LOG.error(ex.getMessage());
     }
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
@@ -11,15 +11,24 @@
 package org.eclipse.che.workspace.infrastructure.kubernetes;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
+import static io.fabric8.kubernetes.client.utils.Utils.isNotNullOrEmpty;
 
 import io.fabric8.kubernetes.client.Config;
 import io.fabric8.kubernetes.client.ConfigBuilder;
 import io.fabric8.kubernetes.client.DefaultKubernetesClient;
 import io.fabric8.kubernetes.client.KubernetesClient;
+import io.fabric8.kubernetes.client.utils.HttpClientUtils;
+import java.io.IOException;
 import javax.annotation.PreDestroy;
 import javax.inject.Inject;
 import javax.inject.Named;
 import javax.inject.Singleton;
+import okhttp3.Authenticator;
+import okhttp3.Credentials;
+import okhttp3.Interceptor;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
 import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
 import org.eclipse.che.commons.annotation.Nullable;
 import org.slf4j.Logger;
@@ -34,7 +43,8 @@ public class KubernetesClientFactory {
 
   private static final Logger LOG = LoggerFactory.getLogger(KubernetesClientFactory.class);
 
-  private final UnclosableKubernetesClient client;
+  private OkHttpClient httpClient;
+  private Config defaultConfig;
 
   @Inject
   public KubernetesClientFactory(
@@ -43,6 +53,13 @@ public class KubernetesClientFactory {
       @Nullable @Named("che.infra.kubernetes.password") String password,
       @Nullable @Named("che.infra.kubernetes.oauth_token") String oauthToken,
       @Nullable @Named("che.infra.kubernetes.trust_certs") Boolean doTrustCerts) {
+    this.defaultConfig =
+        buildDefaultConfig(masterUrl, username, password, oauthToken, doTrustCerts);
+    this.httpClient = HttpClientUtils.createHttpClient(defaultConfig);
+  }
+
+  protected Config buildDefaultConfig(
+      String masterUrl, String username, String password, String oauthToken, Boolean doTrustCerts) {
     ConfigBuilder configBuilder = new ConfigBuilder();
     if (!isNullOrEmpty(masterUrl)) {
       configBuilder.withMasterUrl(masterUrl);
@@ -63,7 +80,78 @@ public class KubernetesClientFactory {
     if (doTrustCerts != null) {
       configBuilder.withTrustCerts(doTrustCerts);
     }
-    this.client = new UnclosableKubernetesClient(configBuilder.build());
+
+    Config config = configBuilder.build();
+    return config;
+  }
+
+  protected OkHttpClient getHttpClient() {
+    return httpClient;
+  }
+
+  protected Config getDefaultConfig() {
+    return defaultConfig;
+  }
+
+  public Config buildConfig(Config defaultConfig, @Nullable String workspaceId)
+      throws InfrastructureException {
+    return defaultConfig;
+  }
+
+  /**
+   * Creates instance of {@link KubernetesClient}.
+   *
+   * @throws InfrastructureException if any error occurs on client instance creation.
+   */
+  private KubernetesClient create(Config config) throws InfrastructureException {
+    OkHttpClient clientHttpClient =
+        httpClient.newBuilder().authenticator(Authenticator.NONE).build();
+    OkHttpClient.Builder builder = clientHttpClient.newBuilder();
+    builder.interceptors().clear();
+    clientHttpClient =
+        builder
+            .addInterceptor(
+                new Interceptor() {
+                  @Override
+                  public Response intercept(Chain chain) throws IOException {
+                    Request request = chain.request();
+                    if (isNotNullOrEmpty(config.getUsername())
+                        && isNotNullOrEmpty(config.getPassword())) {
+                      Request authReq =
+                          chain
+                              .request()
+                              .newBuilder()
+                              .addHeader(
+                                  "Authorization",
+                                  Credentials.basic(config.getUsername(), config.getPassword()))
+                              .build();
+                      return chain.proceed(authReq);
+                    } else if (isNotNullOrEmpty(config.getOauthToken())) {
+                      Request authReq =
+                          chain
+                              .request()
+                              .newBuilder()
+                              .addHeader("Authorization", "Bearer " + config.getOauthToken())
+                              .build();
+                      return chain.proceed(authReq);
+                    }
+                    return chain.proceed(request);
+                  }
+                })
+            .build();
+
+    return new UnclosableKubernetesClient(clientHttpClient, config);
+  }
+
+  /**
+   * Creates instance of {@link KubernetesClient}.
+   *
+   * @throws InfrastructureException if any error occurs on client instance creation.
+   */
+  public KubernetesClient create(String workspaceId) throws InfrastructureException {
+    Config configForWorkspace = buildConfig(defaultConfig, workspaceId);
+
+    return create(configForWorkspace);
   }
 
   /**
@@ -72,13 +160,20 @@ public class KubernetesClientFactory {
    * @throws InfrastructureException if any error occurs on client instance creation.
    */
   public KubernetesClient create() throws InfrastructureException {
-    return client;
+    return create(buildConfig(defaultConfig, null));
   }
 
   @PreDestroy
   private void cleanup() {
     try {
-      client.doClose();
+      if (httpClient.connectionPool() != null) {
+        httpClient.connectionPool().evictAll();
+      }
+      if (httpClient.dispatcher() != null
+          && httpClient.dispatcher().executorService() != null
+          && !httpClient.dispatcher().executorService().isShutdown()) {
+        httpClient.dispatcher().executorService().shutdown();
+      }
     } catch (RuntimeException ex) {
       LOG.error(ex.getMessage());
     }
@@ -89,15 +184,11 @@ public class KubernetesClientFactory {
    */
   private static class UnclosableKubernetesClient extends DefaultKubernetesClient {
 
-    public UnclosableKubernetesClient(Config config) {
-      super(config);
+    public UnclosableKubernetesClient(OkHttpClient httpClient, Config config) {
+      super(httpClient, config);
     }
 
     @Override
     public void close() {}
-
-    void doClose() {
-      super.close();
-    }
   }
 }
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
@@ -58,6 +58,52 @@ public class KubernetesClientFactory {
     this.httpClient = HttpClientUtils.createHttpClient(defaultConfig);
   }
 
+  /**
+   * Builds the Kubernetes {@link Config} object based on a default {@link Config} object and an
+   * optional workspace Id.
+   */
+  public Config buildConfig(Config defaultConfig, @Nullable String workspaceId)
+      throws InfrastructureException {
+    return defaultConfig;
+  }
+
+  /**
+   * Creates instance of {@link KubernetesClient}.
+   *
+   * @throws InfrastructureException if any error occurs on client instance creation.
+   */
+  public KubernetesClient create(String workspaceId) throws InfrastructureException {
+    Config configForWorkspace = buildConfig(defaultConfig, workspaceId);
+
+    return create(configForWorkspace);
+  }
+
+  /**
+   * Creates instance of {@link KubernetesClient}.
+   *
+   * @throws InfrastructureException if any error occurs on client instance creation.
+   */
+  public KubernetesClient create() throws InfrastructureException {
+    return create(buildConfig(defaultConfig, null));
+  }
+
+  /** Retrieves the {@link OkHttpClient} instance shared by all Kubernetes clients. */
+  protected OkHttpClient getHttpClient() {
+    return httpClient;
+  }
+
+  /**
+   * Retrieves the default Kubernetes {@link Config} that will be the base configuration to create
+   * per-workspace configurations.
+   */
+  protected Config getDefaultConfig() {
+    return defaultConfig;
+  }
+
+  /**
+   * Builds the default Kubernetes {@link Config} that will be the base configuration to create
+   * per-workspace configurations.
+   */
   protected Config buildDefaultConfig(
       String masterUrl, String username, String password, String oauthToken, Boolean doTrustCerts) {
     ConfigBuilder configBuilder = new ConfigBuilder();
@@ -83,19 +129,6 @@ public class KubernetesClientFactory {
 
     Config config = configBuilder.build();
     return config;
-  }
-
-  protected OkHttpClient getHttpClient() {
-    return httpClient;
-  }
-
-  protected Config getDefaultConfig() {
-    return defaultConfig;
-  }
-
-  public Config buildConfig(Config defaultConfig, @Nullable String workspaceId)
-      throws InfrastructureException {
-    return defaultConfig;
   }
 
   /**
@@ -141,26 +174,6 @@ public class KubernetesClientFactory {
             .build();
 
     return new UnclosableKubernetesClient(clientHttpClient, config);
-  }
-
-  /**
-   * Creates instance of {@link KubernetesClient}.
-   *
-   * @throws InfrastructureException if any error occurs on client instance creation.
-   */
-  public KubernetesClient create(String workspaceId) throws InfrastructureException {
-    Config configForWorkspace = buildConfig(defaultConfig, workspaceId);
-
-    return create(configForWorkspace);
-  }
-
-  /**
-   * Creates instance of {@link KubernetesClient}.
-   *
-   * @throws InfrastructureException if any error occurs on client instance creation.
-   */
-  public KubernetesClient create() throws InfrastructureException {
-    return create(buildConfig(defaultConfig, null));
   }
 
   @PreDestroy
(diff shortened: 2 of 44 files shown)


infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/bootstrapper/KubernetesBootstrapper.java
```diff
@@ -128,14 +129,44 @@ private void injectBootstrapper() throws InfrastructureException {
...
+      }
+      filesToContatenate.add(GSON.toJson(installer));
+    }
+    filesToContatenate.add("]");
+    int counter = 1;
+    for (String fileToConcatenate : filesToContatenate) {
+      kubernetesMachine.exec(
+          "sh",
+          "-c",
+          "cat > "
```
sleshchenko: Why do you need created n files? It'd better use `>>` instead of `>` and cat all content into one file part by part.

sleshchenko (commented):
Please update links to comment in PR the description. Not it's pointing directly commits and there is no an ability to submit review comment. I'd better to paste link like that [PR diff by commit](https://github.com/eclipse/che/pull/8805/commits/f992e247f4699503dd731b7c53506c306aef4847)

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/runtime/RuntimeIdentity.java
```diff
@@ -16,5 +16,7 @@
 
   String getEnvName();
 
-  String getOwner();
+  String getOwnerName();
+
+  String getOwnerId();
```
sleshchenko, line 21: Do we really need name and id at the same time? I would consider replacing username with user id. 
For me, it is better because:
1. We already have workspace id here but not workspace name (it's a weak reason but still it is better to do in the same way).
2. Username is potentially modifiable value, while user id is unmodifiable. So if username from `RuntimeIdentity` will be stored somewhere like DB, logs then it can be false.

@gazarenkov WDYT?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/bootstrapper/KubernetesBootstrapper.java
davidfestal: I'll do the change

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/runtime/RuntimeIdentity.java
davidfestal: I agree that having both `userId` and `userName` might seem redundant, but it seems that best way to make this PR non-breaking and avoid making it even bigger and more complicated. For example I assume that the owner was used to build the name of the created workspace in the dashboard, and in certainly in some other places, and I'm not sure we want to have this replaced with the `userId`.
Maybe using only the `userId`, and replace the use of the `ownerName` field by calculated / retrieved values at use sites might be done in a distinct PR, no ?

sleshchenko: @davidfestal I'd way that your changes are already a bit breaking because you rename the field.
I'm OK to replace name value with id in another PR but then let's do not add owner id in this PR.
If you need it in rh-che, you can retrieve user id by username from RuntimeIdentity using UserManager class. I think it can be used as a temporary solution.
WDYT?

P.S. 
> For example I assume that the owner was used to build the name of the created workspace in the dashboard 

RuntimeIdentity is not exposed by REST API, so it can not be used by clients like a dashboard.

davidfestal: well, it seems to me it's a bit just going backwards. But if you all want me to do that, I can do it. @l0rd wdyt ?

sleshchenko: I think it's better to get agreement how it should look like and only then rework it. Anyway, your job can be saved in another branch and it could be used later.

davidfestal: Sure. I'll wait for other thoughts then.

skabashnyuk: What if we just add ```ownerId``` field?

skabashnyuk: I mean left method ```String getOwner();``` as is and not rename it and also add method ```String getOwnerId();```

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
```diff
@@ -63,7 +80,78 @@ public KubernetesClientFactory(
...
+
+    return new UnclosableKubernetesClient(clientHttpClient, config);
+  }
+
+  /**
+   * Creates instance of {@link KubernetesClient}.
+   *
+   * @throws InfrastructureException if any error occurs on client instance creation.
+   */
+  public KubernetesClient create(String workspaceId) throws InfrastructureException {
```
sleshchenko: I prefer to make classes have the following structure
```
class Class {
  // constants
  
  // fields
  
  // constructur
  
  // public methods
  
  // protected methods

  // private methods
}
```
I'd sort methods here.

It's not a requirement but just my personal taste for code =) 
So you can leave without changes if you think if it is better.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftClientFactory.java
```diff
@@ -65,22 +71,56 @@ public OpenShiftClientFactory(
...
   }
 
   /**
    * Creates instance of {@link OpenShiftClient}.
    *
    * @throws InfrastructureException if any error occurs on client instance creation.
    */
-  public OpenShiftClient create() throws InfrastructureException {
-    return client;
+  public OpenShiftClient createOC(String workspaceId) throws InfrastructureException {
```
sleshchenko: Can you elaborate why did you add new method instead of overring parent's method?

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/project/OpenShiftProject.java
```diff
@@ -48,12 +49,13 @@ public OpenShiftProject(OpenShiftClientFactory clientFactory, String name, Strin
       throws InfrastructureException {
     super(clientFactory, name, workspaceId, false);
     this.routes = new OpenShiftRoutes(name, workspaceId, clientFactory);
-    doPrepare(name, clientFactory.create());
+    doPrepare(name, clientFactory.create(), clientFactory.createOC());
```
sleshchenko: Should `clientFactory.create()` be invoked here with `workspaceId`?

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/runtime/RuntimeIdentity.java
davidfestal: I don't see what it would have simplified, and things would be less clear IMO.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/project/OpenShiftProject.java
davidfestal: you're right, it should be invoked with the `workspaceId` it seems

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
```diff
@@ -72,13 +160,20 @@ public KubernetesClientFactory(
    * @throws InfrastructureException if any error occurs on client instance creation.
    */
   public KubernetesClient create() throws InfrastructureException {
```
sleshchenko: I think this method should be removed at all. It may lead to an error because default client is not supposed to be used for performing an operation with OS cluster, while it can be used. 
As far as I see it is used only for parsing Kubernetes list. Then it makes sense to create default kubernetes client there or use `io.fabric8.kubernetes.client.utils.Serialization#unmarshal` method directly.

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/runtime/RuntimeIdentity.java
skabashnyuk: 1. It's backward compatible. Nothing has been removed or renamed.
2. It's adding a method to get owner id.

akorneta:
Since we have no restriction on installer script size, the current approach does not guarantee that this problem won't occurs at all, so I think it is possible to do something like this:
https://gist.github.com/akorneta/54617f52ce320bd503bceeb432d10830
In this case, we would be able to adjust the buffer size to our needs, (to get the most effective number of requests / no problem with installer scripts that are bigger than proxy limits)
@davidfestal what do you think?

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/runtime/RuntimeIdentity.java
garagatyi: +1 to what @gazarenkov said. But I think that there is a confusion:
- @gazarenkov right now username is used in this field, and @davidfestal needs the user ID
- @davidfestal my vote is for setting the user ID in the owner field, neither renaming nor adding any field. I believe it is what @sleshchenko and @gazarenkov was talking about

davidfestal: Until now the Keycloak `preferred_name` (=> user name such as `dfestal` => login name afaik) was used in `RuntimeIdentity`, instead of the real `userId` that is the real user *primary key* afaik. In this PR I add the `userId` in the `RuntimeIdentity` because:
1. We need the `userId` in OSIO
2. I think it has its natural place in the `RuntimeIdentity` object (as you just explained)

However, since I don't fully understand why the user name was used previously, and what precise requirements is used to meet, I kept it in the `RuntimeIdentity` object for now: it has the advantage of making it directly available in every place where it is used, instead of having to retrieve it from the user ID.

My proposal is to merge this PR with the 2 fields, since it doesn't hurt and makes sure there's no regression, and then study the removal of the `userName` part in a distinct Pull Request.

davidfestal: @garagatyi I'm OK with this, but I still think it would be better to do this is a distinct PR.

davidfestal:
> so I think it is possible to do something like this:
> https://gist.github.com/akorneta/54617f52ce320bd503bceeb432d10830
> In this case, we would be able to adjust the buffer size to our needs, (to get the most effective number of requests / no problem with installer scripts that are bigger than proxy limits)

@akorneta I'm not against it, but remember that the goal of my fix, which was meant to be as simple and lightweight as possible, is to have Che 6 work on OSIO with the OSO proxy, which was not possible due to the huge GET request.

Sure we could make this even more robust and have better control on the length of `exec` commands, but IMO that's not the purpose of this PR. I'd prefer doing this in a distinct PR in order to avoid postponing the availability of Che6 on OSIO, which is the main goal of this Sprint afaik.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftClientFactory.java
davidfestal: In some use-cases (mainly the use-cases involving `Exec` or `Watch` commands with underlying long-runing web-socket connections), the `OpenshiftClient` introduces connection leaks as soon as we create / use / manage several `OpenshiftClient` instances. This seems to be due to the `OpenShiftOAuthInterceptor`, and the structure of the `OpenshiftClient` doesn't allow us overriding this error-prone behavior.
The previous fix for this in Che6 was to use a single `OpenshiftClient` instance for all the calls, which is obviously not possible in multi-tenant scenarios.
 
However the good news is that for such use-cases (`Exec` and `Watch`), the `KubernetesClient` is fully sufficient. In fact the `KubernetesClient` is sufficient for most of the API calls in the Openshift insfastructure (which now derives from the kubernetes infrastructure BTW).

So the idea of this change is that, even in the Openshift infrastructure, we use the `KubernetesClient` in all the places where it is possible, and only use the `OpenshiftClient` for Openshift-specific API calls: routes and projects.

This is the way it has been fixed in the Che5-based OSIO currently running in production.

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/runtime/RuntimeIdentity.java
garagatyi: -1 to merge with an additional field. I think we should avoid extending code in different places and rework it in the following PRs. The best code is not written code. And using userId instead of username in owner field would remove a lot of code from this PR and keep things simple. 
About the reasoning for initial putting username into the owner field, I believe that it was done unintentionally. AFAIK Che 6 is not GA yet, so backward compatibility should not be an issue in this case.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/bootstrapper/KubernetesBootstrapper.java
davidfestal: Changed in commit https://github.com/eclipse/che/pull/8805/commits/1fe3e1403394efecc391f7cbaeb749fc8467e5a8

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
davidfestal: Done in commit https://github.com/eclipse/che/pull/8805/commits/c1d9dfcbdb7f856774e1d7cf24b4069d6134bf98

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/project/OpenShiftProject.java
davidfestal: Done in commit https://github.com/eclipse/che/pull/8805/commits/112e995392f1066e94d6708f5cf2761e38b312eb

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/runtime/RuntimeIdentity.java
gazarenkov: We just need to fill owner with unique user identifier (whatever it is called in user identity system: id, name, key, etc). Since this class is not intended to be persisted (in normal case), I see no risk changing it, especially before GA.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
```diff
@@ -43,6 +53,59 @@ public KubernetesClientFactory(
...
+
+    return create(configForWorkspace);
+  }
+
+  /**
+   * Creates instance of {@link KubernetesClient}.
+   *
+   * @throws InfrastructureException if any error occurs on client instance creation.
+   */
+  public KubernetesClient create() throws InfrastructureException {
```
sleshchenko: I think this method should be removed at all. It may lead to an error because default client is not supposed to be used for performing an operation with OS cluster, while it can be used.
As far as I see it is used only for parsing Kubernetes list. Then it makes sense to create default Kubernetes client there or use io.fabric8.kubernetes.client.utils.Serialization#unmarshal method directly.

davidfestal: > I think this method should be removed at all. It may lead to an error because default client is not supposed to be used for performing an operation with OS cluster, while it can be used.

I don't understand the problem or the error this method could lead to: it's still calling the default configuration, it's still calling `buildConfig()` method, though without a workspace ID.

davidfestal: Functionally it's the exactly the same method as the previous `create()` method.
Once again, maybe we can remove it and change the `list()` method to something else, but I'm not sure it has its place in this PR.

sleshchenko: > Functionally it's the exactly the same method as the previous create() method.

Yeah, but previously there was only one method.
As far as I understand, Now clients **must** be created with specified `workspaceId`.

> I don't understand the problem or the error this method could lead to: it's still calling the default configuration, it's still calling buildConfig() method, though without a workspace ID.

Within Che it will work even if all components of Kubernetes/OpenShift infrastructure will use default client without specifying`workspaceId`. But will RhChe work in this case? You'll update all places to create clients with workspaceId. But it's not clear enough where workspace id must be specified and where it is optional. 
So new components may create default clients and I guess it will lead to failing RHChe packaging. WDYT? Does it make sense?

davidfestal: > Now clients *must* be created with specified workspaceId

Well, not exactly. That's more: *when the `workspaceId` is known then it __should__ be provided* to help inferring the right OpenShift cluster / namespace to connect to.

However if no `workspaceId` is available then `rh-che` would simply use the current user information to infer the right Openshift cluster / namespace. I don't see the problem here: it's mainly the same behavior as on Che 5, but implemented in a cleaner way IMO.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesPods.java
```diff
@@ -475,10 +476,17 @@ private ExecWatchdog() {
...
-    public void onOpen(Response response) {}
+    public void onOpen(Response response) {
+      if (response != null && response.body() != null) {
+        response.body().close();
+      }
+    }
 
     @Override
     public void onFailure(Throwable t, Response response) {
+      if (response != null && response.body() != null) {
```
garagatyi: I think these workarounds are needed to avoid connection leaks. Am I right? If so, please, add comments into the code because it is not obvious at all.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
```diff
@@ -63,22 +126,67 @@ public KubernetesClientFactory(
...
+
+    return new UnclosableKubernetesClient(clientHttpClient, config);
   }
 
   @PreDestroy
   private void cleanup() {
     try {
-      client.doClose();
+      if (httpClient.connectionPool() != null) {
+        httpClient.connectionPool().evictAll();
```
garagatyi: Please, take a look at my comment to OpenshiftClientFactory

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
```diff
@@ -63,22 +126,67 @@ public KubernetesClientFactory(
...
+    OkHttpClient.Builder builder = clientHttpClient.newBuilder();
+    builder.interceptors().clear();
+    clientHttpClient =
+        builder
+            .addInterceptor(
+                new Interceptor() {
+                  @Override
+                  public Response intercept(Chain chain) throws IOException {
+                    Request request = chain.request();
+                    if (isNotNullOrEmpty(config.getUsername())
```
garagatyi: If I'm not mistaken this code is copied from HttpClientUtils. If so, it is an incomplete copy. Can you elaborate why we don't use HttpClientUtils instead of this code? If there is no specific reason I would suggest using original code. It is less error-prone in case of any further changes in k8s API.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
```diff
@@ -43,6 +53,59 @@ public KubernetesClientFactory(
...
+    this.defaultConfig =
+        buildDefaultConfig(masterUrl, username, password, oauthToken, doTrustCerts);
+    this.httpClient = HttpClientUtils.createHttpClient(defaultConfig);
+  }
+
+  /**
+   * Builds the Kubernetes {@link Config} object based on a default {@link Config} object and an
+   * optional workspace Id.
+   */
+  public Config buildConfig(Config defaultConfig, @Nullable String workspaceId)
```
garagatyi: Can we show that this method is not supposed to be used directly by consumers of this class by making the method protected?

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/runtime/RuntimeIdentity.java
l0rd: Iterating with 2 distinct PR should not to be considered bad by itself. I prefer multiple small PRs instead of one-month-long PR that is more likely to break something, more difficult to test and review.

I think we all agree that `ownerName` should be replace by `ownerID` but if David wants to split that in 2 steps (1. add `userId` 2. remove `userName`) that make sense to me.

The value of this PR is not the refactoring of `RuntimeIdentity` but to make Che work in multi tenant scenario like OSIO.

@gorkem I'm afraid that's one of the cases where we need your help to take a decision without spending too much time discussing.

garagatyi: @l0rd as far as I know the idea is to avoid adding changes to the class rather than changing it twice.
What is suggested is to change the content of the field which should be 1 line change, and remove dozens of lines of changes.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
garagatyi: Actually looks like you can make this method protected and call it from the OpenshiftClientFactory to avoid duplicated code. WDYT?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/pvc/UniqueWorkspacePVCStrategy.java
```diff
@@ -174,7 +174,7 @@ else if (provisionedVolumeName2PVC.containsKey(volumeName)) {
   @Override
   public void cleanup(String workspaceId) throws InfrastructureException {
     clientFactory
-        .create()
+        .create(workspaceId)
         .persistentVolumeClaims()
         .inNamespace(namespaceName)
```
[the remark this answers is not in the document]
sleshchenko: @garagatyi I'm not sure that we can do it while configuring a client. Now we do such configuring by using `KubernetesNamespace`/`OpenShiftProject` clients wrappers. Maybe we should use `KubernetesPersistentVolumeClaims` from `KubernetesNamespace` here instead of directly using of kubernetes client.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
sleshchenko: > Now clients must be created with specified workspaceId

Oh, It was my wrong assumption. It's quite confusing. Maybe a good java doc will make it more clear.

But is default client really needed? If no, please consider removing default client at all to make it more straightforward. As far as I understand `workspaceId` is unknown only if this operation is not really related to a particular workspace. But then why should we interact with Kubernetes/OpenShift API? It could be a case like fetch cluster information, quotas or something like that. But I think it'd better remove such method while we don't need them. So, I don't have strong arguments to do it, so **up to you** remove it or not

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftClientFactory.java
sleshchenko: I thought that caching of OkHttpClient resolves an issue with connection leaks. Doesn't it?

If not then please consider adding java doc that using OpenShift client may lead to connection leaks and it is better to avoid using it if Kubernetes client is able to do the same action.

But it looks like a workaround it's better to register an issue for fabric8 OS Client.

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/runtime/RuntimeIdentity.java
gazarenkov: @l0rd it is definitely not the case you have in mind, so procedure you propose is hardly applicable. I do not even think it looks like a situation (yet at least) where code maintainer should be involved (as it stated here https://github.com/eclipse/che/wiki/Development-Process ). Let engineering discussion go and the best/simplest way win :)

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftClientFactory.java
davidfestal: > I thought that caching of `OkHttpClient` resolves an issue with connection leaks. Doesn't it?

That's not precisely the case. In order to fix connection leaks, Che 6 was initially caching the `OpenshiftClient` instance itself in a global way. But now we cannot cache it anymore since multi-tenancy involves using different `OpenshiftClient` instances for starting workspaces for different users.

> If not then please consider adding java doc that using OpenShift client may lead to connection leaks and it is better to avoid using it if Kubernetes client is able to do the same action.

Yes, that's something I will do

> But it looks like a workaround it's better to register an issue for fabric8 OS Client.

Sure, we can open an issue on the OS client, but once again, I'm not sure we want to wait for this bug / limitation to be clearly analyzed / fixed on the OS client side, a new OS client release issued, and a new CQ approved before being able to run Che6 on OSIO ...

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
davidfestal: We only want to apply the part that is related to the Authentication interceptor. All the other common configuration steps have already been applied when creating the shared base `OkHttpClient` by calling precisely the `HttpClientUtils.createHttpClient()` method. But here we only want to override the authentication interceptor.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
davidfestal: Good idea.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
davidfestal: yes, I'll change it

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/runtime/RuntimeIdentity.java
l0rd: @garagatyi yeah but [good refactoring practices](https://www.amazon.com/Refactoring-Improving-Design-Existing-Code/dp/0201485672) (an excellent book by the way) suggest not to take shortcuts. When you need to replace some code the advice is to split it in 2 simpler steps (addition of new code and removal of existing code).

@gazarenkov unfortunately it looks like there is no consensus. This is an urgent PR: it allows to deploy Che 6 on OSIO. Given the importance I would like to move forward as soon as possible. I don't want this PR to be subject to a multiple weeks review process because of a trivial refactoring when the problem it solves has been afflicting us for months.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
garagatyi: I think it is very unclear that this takes place. Can you add comments to the code that explains this behavior?

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/runtime/RuntimeIdentity.java
gazarenkov: backing to the discussion in substance.
As I understand all agreed that we need only one field (owner) which currently contains user name but it is better if it contain userID.

As I understand the multistep procedure ((1) add OwnerID; (2) remove Owner(Name)) proposed due to potential backward compatibility risk.

My questions are:
What is the backward compatibility risk you, guys, (@davidfestal @l0rd @skabashnyuk) see if we just filling Owner with userID (one line code change as @garagatyi mentioned) taking into account that: this class's content is not exposed externally (not used in API etc)?

@davidfestal just for my understanding,  when you said "We need the userId in OSIO" what exactly purpose/case you have in mind in this regard?

thanks

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftClientFactory.java
```diff
@@ -65,22 +71,56 @@ public OpenShiftClientFactory(
...
 
   @PreDestroy
   private void cleanup() {
+    OkHttpClient httpClient = getHttpClient();
     try {
-      client.doClose();
+      if (httpClient.connectionPool() != null) {
+        httpClient.connectionPool().evictAll();
+      }
+      if (httpClient.dispatcher() != null
```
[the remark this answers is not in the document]
davidfestal: I fixed this in commit https://github.com/eclipse/che/pull/8805/commits/2f9d7cf6e0fc28b099f36180b4c4303569e9ca43 by simply using the same code as `KubernetesClient.close()`, inside a `doCleanup()` protected  method that is used in both `KubernetesClientFactory` and `OpenshiftClientFactory`.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftClientFactory.java
```diff
@@ -65,22 +71,56 @@ public OpenShiftClientFactory(
...
+          && httpClient.dispatcher().executorService() != null
```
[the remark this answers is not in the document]
davidfestal: I fixed this in commit https://github.com/eclipse/che/pull/8805/commits/2f9d7cf6e0fc28b099f36180b4c4303569e9ca43 by simply using the same code as `KubernetesClient.close()`, inside a `doCleanup()` protected  method that is used in both `KubernetesClientFactory` and `OpenshiftClientFactory`.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesPods.java
davidfestal: I added comments in commit https://github.com/eclipse/che/pull/8805/commits/2f9d7cf6e0fc28b099f36180b4c4303569e9ca43

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
davidfestal: I fixed this in commit https://github.com/eclipse/che/pull/8805/commits/2f9d7cf6e0fc28b099f36180b4c4303569e9ca43 by simply using the same code as `KubernetesClient.close()`, inside a `doCleanup()` protected  method that is used in both `KubernetesClientFactory` and `OpenshiftClientFactory`.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
davidfestal: I added comments in commit https://github.com/eclipse/che/pull/8805/commits/2f9d7cf6e0fc28b099f36180b4c4303569e9ca43

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
davidfestal: I fixed it commit https://github.com/eclipse/che/pull/8805/commits/2f9d7cf6e0fc28b099f36180b4c4303569e9ca43

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/runtime/RuntimeIdentity.java
l0rd: @gazarenkov the fact that `owner` was not used in API doesn't mean that it's not used at all and that we can remove it and be sure that there will be no regression. That's just one more modification that need attention and testing and @davidfestal prefer to do it in a separate PR. And that's absolutely legitimate taken into account the complexity of the actual PR.

Your second question (why we need the userID) is actually answered at point 3 of this PR description.

I am sure that there are still tons of questions that can be asked on this PR and that there are a lot of refactoring that could still be done (is code ever perfect?) but considered this PR importance and that @davidfestal has answered and addressed almost all comments I think that we should merge and move forward. @sleshchenko @garagatyi wdyt?

l0rd (approved):

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesPods.java
davidfestal: After testing again without these changes, it appears that connection leaks are still fixed (obviously by the use of the kubernetes client in API calls that use web-sockets).
So I rolled-back those changes in commit https://github.com/eclipse/che/pull/8805/commits/ddc2beb4288df376eed9c4566e923a6154d5dfbf since they are not required anymore to fix the connection leaks.

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/runtime/RuntimeIdentity.java
gazarenkov: @l0rd thanks for answering second question

still confusing for first...
Should I read it as: there are no visible risk but you prefer not to change the current PR (due to complexity etc). Which means once merged this PR will be immediately followed by another which removes userName?
If it is the case - I let you, guys, decide on tactic.
Otherwise - I still have conceptual concerns I voiced initially and ask for describing the risk you can see in details.

ibuziuk (approved):

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/runtime/RuntimeIdentity.java
l0rd: @gazarenkov yes this PR will be followed by another one which removes `userName`

garagatyi (approved):
I'm OK to merge the PR in this state.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
sleshchenko: @davidfestal Please add java docs that make clear when a client should be created with specified workspace id and when default one without parameters.

sleshchenko (approved):
OK to me to merge it when the mentioned javadoc will be added.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
davidfestal: Done in commit https://github.com/eclipse/che/pull/8805/commits/5ae5218d17217e07557baa62f6523b3964262f8a

davidfestal:
> OK to me to merge it when the mentioned javadoc will be added.

@sleshchenko done in commit https://github.com/eclipse/che/pull/8805/commits/5ae5218d17217e07557baa62f6523b3964262f8a. Does it correspond to what you were expecting ?

sleshchenko:
@davidfestal yes, thx.

davidfestal:
ci-build

benoitf:
add to whitelist

benoitf:
ci-build

davidfestal:
ci-test

skabashnyuk (approved):

akorneta (approved):

davidfestal:
ci-test

davidfestal:
ci-test-ocp

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/runtime/RuntimeIdentity.java
davidfestal: I created issue #8859 as a placeholder for further discussion on the userName - userId replacement.

davidfestal:
ci-test-ocp

davidfestal:
ci-test-ocp

davidfestal:
ci-test

davidfestal:
ci-test-ocp

davidfestal:
ci-test-ocp

davidfestal:
The last commit https://github.com/eclipse/che/pull/8805/commits/c3c718cf0ee2464fa3005122edab72afc9693d01 fixed the regression shown in the following [tests](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/48/Selenium_tests_report/) from build [48](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/48/Selenium_tests_report/) that was failing.

The [last OCP integration tests](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/49//Selenium_tests_report) show these errors are fixed in build [49](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/49/).

FTR the regression described in the commit is explained also explained hereafter:

> In the Openshift infrastructure, the authentication interceptor should always convert userName/password-based authentication to oauth-token-based authentication, even when using the
`KubernetesClient`.

ibuziuk:
LGTM

Issue #7768, opened 2017-12-07
Refactoring method does not work properly in certain case


musienko-maxim:
### Reproduction Steps
I've used che during dog fooding session. I opened the CodenvyEditor class and try to rename method ``setCursorToDefinedLineAndCharForVisible``. But refactoring did not complete and i had 500 error from server without clarified details in the response.
![ezgif com-video-to-gif2](https://user-images.githubusercontent.com/1640058/33716151-c2e447c6-db5d-11e7-8d44-fe8eb5ce2abb.gif)

**OS and version:**    
Che 6.0.0-M3, Chrome browser, Eclipse Che stack.
**Diagnostics:**

mmorhun:
Seems this is quite good reprodusable on some methods in `CodenvyEditor`

PR #8814, opened 2018-02-16
Set up default file encoding to utf8 in workspace agent configuration

vzhukovs:
### What does this PR do?
This changes proposal sets up default file encoding in ws agent to utf8.

The problem is in code:

https://github.com/eclipse/che/blob/master/plugins/plugin-java/che-plugin-java-ext-jdt/org-eclipse-core-filebuffers/src/main/java/org/eclipse/core/internal/filebuffers/FileStoreTextFileBuffer.java#L419

```java
protected void commitFileBufferContent(IProgressMonitor monitor, boolean overwrite)
      throws CoreException {
    //		if (!isSynchronized() && !overwrite)
    //			throw new CoreException(new Status(IStatus.WARNING, FileBuffersPlugin.PLUGIN_ID,
    // IResourceStatus.OUT_OF_SYNC_LOCAL, FileBuffersMessages.FileBuffer_error_outOfSync, null));

    String encoding = computeEncoding();  <<<<-----

    ....
```

which calls:

https://github.com/eclipse/che/blob/master/plugins/plugin-java/che-plugin-java-ext-jdt/org-eclipse-core-filebuffers/src/main/java/org/eclipse/core/internal/filebuffers/TextFileBufferManager.java#L443

```java
  /*
   * @see org.eclipse.core.buffer.text.IBufferedFileManager#getDefaultEncoding()
   */
  public String getDefaultEncoding() {
    return System.getProperty("file.encoding"); // $NON-NLS-1$;
  }
```

and on che-in-che stacks it returns `ANSI_X3.4-1968`. If user project sources contain some characters which are not supported by this encoding then during refactoring session we'll get `JavaModelException` with message: 

>Some characters cannot be mapped using "ANSI_X3.4-1968" character encoding.
Either change the encoding or remove the characters which are not supported by the "ANSI_X3.4-1968" character encoding.

This issue can be solved by:

* Setting up default locale into stack _(which is not preferable)_
* Setting up default file encoding in java options via `-Dfile.encoding=UTF8`

Second variant is more preferable because our users source code is stored in UTF8 locale.

#7768 

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
@@ -75,7 +75,7 @@ che.workspace.java_options=-XX:MaxRAM=150m -XX:MaxRAMFraction=2 -XX:+UseParallel
 che.workspace.maven_options=-XX:MaxRAM=150m -XX:MaxRAMFraction=2 -XX:+UseParallelGC -XX:MinHeapFreeRatio=10 -XX:MaxHeapFreeRatio=20 -XX:GCTimeRatio=4 -XX:AdaptiveSizePolicyWeight=90 -Dsun.zip.disableMemoryMapping=true -Xms20m -Djava.security.egd=file:/dev/./urandom
 
 # Default java command line options to be added to JVM that run workspace agent.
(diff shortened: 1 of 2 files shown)


vzhukovskii:
ci-test

skabashnyuk (approved):

vparfonov (approved):

musienko-maxim (approved):

Issue #8436, opened 2018-01-24
Open Symbol is Broken in jdt.ls Branch


tsmaeder:
Open Symbol does open files anymore in the jdt.ls branch. The problem is that I missed the conversion to using URI's instead of paths.
To reproduce, try to open the type java.lang.String via "Open Symbol".

PR #8827, opened 2018-02-19
Fix for workspace symbols not opening the file

JPinkney:

### What does this PR do?
This PR fixes an issue where workspace symbols would not open the file in the editor

https://github.com/eclipse/che/issues/8436


wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/service/LanguageServiceUtils.java
```diff
@@ -35,6 +35,10 @@ public static String removeUriScheme(String uri) {
     return uri.startsWith(FILE_PROJECTS) ? uri.substring("file://".length()) : uri;
   }
 
+  public static String removeFilePrefix(String uri) {
+    return uri.startsWith("file:/projects") ? uri.substring("file:/projects".length()) : uri;
```
tolusha: Where does `file:/project` come from?

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/service/WorkspaceService.java
```diff
@@ -162,7 +162,7 @@ public boolean handleResult(
                 InitializedLanguageServer element, List<? extends SymbolInformation> locations) {
               locations.forEach(
                   o -> {
-                    o.getLocation().setUri(removePrefixUri(o.getLocation().getUri()));
+                    o.getLocation().setUri(removeFilePrefix(o.getLocation().getUri()));
```
tolusha: `setUri` method suppose to set uri but here we rather set relative path.

dkuleshov:
Hi, 

I see no unit tests for the new methods, did I miss something?

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/service/LanguageServiceUtils.java
JPinkney: It is from the workspace symbols when the file is in your project e.g. "file:/projects/test/src/main/java/tester.java"

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/service/WorkspaceService.java
JPinkney: I must have misunderstood removePrefixUri, it looks like it is trying to remove "file:///projects" from the beginning of the response but I did not find any cases in which this appeared. My understanding was that removing "file:/projects" would not matter because it was supposed to be removing "file:///projects" to begin with? In which case the original code would have produced a relative path.

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/service/LanguageServiceUtils.java
tolusha: Could you point out the method?

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/service/WorkspaceService.java
tolusha: I see.

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/service/LanguageServiceUtils.java
JPinkney: In [this part](https://github.com/eclipse/che/blob/5730_java_ls_poc/wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/service/WorkspaceService.java#L165) o.getLocation().getUri() is being set to something like "file:/projects/test/src/main/java/tester.java" when the file is in your workspace

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/util/OpenFileInEditorHelper.java
```diff
@@ -176,6 +176,15 @@ public void openLocation(String uri, LinearRange range) {
     }
   }
 
+  public void openLocation(String uri, TextRange range) {
```
tsmaeder: We get the uri and range from a LSP SymbolInformation object. There already is a method on this class that can open a lsp Location object, so let's not add a new method here.

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/quickopen/EditorQuickOpenEntry.java
```diff
@@ -38,7 +38,7 @@ protected TextRange getTextRange() {
   public boolean run(Mode mode) {
     if (mode == Mode.OPEN) {
       String filePath = getFilePath();
-      editorHelper.openPath(filePath, getTextRange());
+      editorHelper.openLocation(filePath, getTextRange());
```
tsmaeder: I would just scrap the EditorQuickOpenEntry (it is only extended once) and instead make a QuickOpenEntry derived class that holds a SymbolInformation object directly. Then just use the location to open the editor.

tsmaeder (changes requested):
I think we need to a bit more refactoring here.
Don't hesitate to change code when the requirements change. Your reviewers will tell you if they think you're going too far.

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/service/WorkspaceService.java
tsmaeder: You don't need to do this, IMO. Just remove this code and the utility method

tsmaeder (changes requested):

JPinkney:
Waiting for https://github.com/eclipse/eclipse.jdt.ls/pull/564 to go through

tsmaeder (approved):

tolusha (approved):

svor (approved):

Issue #8802, opened 2018-02-15
Method "ProjectExplorer#openItemByPath()" is closing items which are have been already opened before


Ohrimenko1988:
### Description
In current state, this method does not work with already expanded tree. 
For example:
result of execution "expandPathInProjectTree("web-java-spring/src/main/java/org/eclipse/che/examples")" :
![wwww](https://user-images.githubusercontent.com/22724619/36264998-e7c0149e-1276-11e8-9fa9-39cf1004c007.png)

but if run "expandPathInProjectTree("web-java-spring/src/main/webapp/WEB_INF")" in this expanded project tree :
![cccc](https://user-images.githubusercontent.com/22724619/36265190-63879d4a-1277-11e8-8f4d-4da0fddd9afb.png)

Ohrimenko1988:
Done https://github.com/eclipse/che/pull/8835

PR #8835, opened 2018-02-19
Selenium: Rework "ProjectExplorer#openItemByPath()" method for working with already expanded project explorer tree and fix related selenium tests

Ohrimenko1988:
### What does this PR do?
Rework "ProjectExplorer#openItemByPath()" method for working with already expanded project explorer tree and fix related selenium tests

Issue: https://github.com/eclipse/che/issues/8802

### DIFF
--- a/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java
+++ b/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java
@@ -866,7 +866,7 @@ public class ProjectExplorer {
    * @param path path for expand
    */
   public void expandPathInProjectExplorer(String path) {
-    List<String> itemsPaths = getItemsPathsWithReplaceDots(path);
+    List<String> itemsPaths = splitPathOnSubitems(path);
     ListIterator<String> iterator = itemsPaths.listIterator();
 
     while (iterator.hasNext()) {
@@ -893,7 +893,7 @@ public class ProjectExplorer {
    * @return Paths in format ["RootFolder", "RootFolder/src", "RootFolder/src/main",
    *     "RootFolder/src/main/java", "RootFolder/src/main/java/org/package"]
    */
-  private List<String> getItemsPathsWithReplaceDots(String path) {
+  private List<String> splitPathOnSubitems(String path) {
     Path fullPath = Paths.get(path);
     List<String> itemsPaths = new ArrayList<>();
 
--- a/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java
+++ b/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java
@@ -865,7 +865,7 @@ public class ProjectExplorer {
    * @param path path for expand
    */
   public void expandPathInProjectExplorer(String path) {
-    List<String> itemsPaths = splitPathOnSubitems(path);
+    List<String> itemsPaths = splitPathWithReplaceDotsToSlashes(path);
 
     for (int i = 0; i < itemsPaths.size(); i++) {
       String currentItem = itemsPaths.get(i);
@@ -893,13 +893,14 @@ public class ProjectExplorer {
   }
 
   /**
-   * Parses {@code path} and makes full path for each element which separated by "/".
+   * Parses {@code path} and makes full path for each element which separated by "/". Items like
+   * 'org.package.sub' will be treated as 'org/package/sub'.
    *
-   * @param path Path in format "RootFolder/src/main/java/org.package".
+   * @param path Path in format "RootFolder/src/main/java/org.package.sub".
    * @return Paths in format ["RootFolder", "RootFolder/src", "RootFolder/src/main",
-   *     "RootFolder/src/main/java", "RootFolder/src/main/java/org/package"]
+   *     "RootFolder/src/main/java", "RootFolder/src/main/java/org/package/sub"]
    */
-  private List<String> splitPathOnSubitems(String path) {
+  private List<String> splitPathWithReplaceDotsToSlashes(String path) {
     Path fullPath = Paths.get(path);
     List<String> itemsPaths = new ArrayList<>();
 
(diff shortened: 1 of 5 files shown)


artaleks9 (approved):

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/SeleniumWebDriverHelper.java
```diff
@@ -451,33 +453,35 @@ public void moveCursorTo(WebElement webElement) {
...
-      webDriverWaitFactory
-          .get(ATTACHING_ELEM_TO_DOM_SEC)
-          .until(visibilityOfElementLocated(elementLocator));
-      return true;
-    } catch (TimeoutException ex) {
+      return seleniumWebDriver.findElement(elementLocator).isDisplayed();
+    } catch (NoSuchElementException ex) {
+      return false;
+    } catch (Exception ex) {
+      LOG.info("Unhandled {} in \"isVisible()\" method", ex.getClass().getSimpleName());
```
dmytro-ndp: We can omit `in \"isVisible()\" method` string so as log4j prints reference itself, but logging this info is useless IMHO.

dmytro-ndp: If we are talking about very rare exception which should be took into account, it's better not to handle another type of exceptions here and let client code to work around the problem.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java
```diff
@@ -860,50 +860,49 @@ public void scrollToItemByPath(String path) {
...
-    }
-    String currentPath = currentItem.substring(1);
-    openItemByPath(currentPath);
-    for (int i = numItem + 1; i < pathItems.length; i++) {
-      currentItem = pathItems[i].replace('.', '/');
-      currentPath += "/" + currentItem;
-      waitItem(currentPath);
-      loader.waitOnClosed();
-      openItemByPath(currentPath);
+  private List<String> getItemsPathsWithReplaceDots(String path) {
```
dmytro-ndp: `getItemsPathsWithReplaceDots` doesn't clear explain the propose of method. 
What about another name: `splitPathOnSubitems`?

Ohrimenko1988: :+1:

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java
```diff
@@ -860,50 +860,49 @@ public void scrollToItemByPath(String path) {
...
-      currentPath += "/" + currentItem;
-      waitItem(currentPath);
-      loader.waitOnClosed();
-      openItemByPath(currentPath);
+  private List<String> splitPathOnSubitems(String path) {
+    Path fullPath = Paths.get(path);
+    List<String> itemsPaths = new ArrayList<>();
+
+    for (int i = 0; i < fullPath.getNameCount(); i++) {
+      String currentItemPath = fullPath.subpath(0, i + 1).toString().replace(".", "/");
```
dmytro-ndp: What is the propose of replacing "." on "/"?
It could end up having error so as there could be filename with extension at the end of path.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java
```diff
@@ -860,50 +860,49 @@ public void scrollToItemByPath(String path) {
...
-      waitItem(currentPath);
-      loader.waitOnClosed();
-      openItemByPath(currentPath);
+    List<String> itemsPaths = splitPathOnSubitems(path);
+    ListIterator<String> iterator = itemsPaths.listIterator();
+
+    while (iterator.hasNext()) {
+      String currentItem = iterator.next();
+      if (!iterator.hasNext()) {
+        waitItem(currentItem);
```
dmytro-ndp: Code inside `if` looks useless so as we are opening this item as next one at the previous iteration.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java
Ohrimenko1988: No, it should not. For your described case uses "ProjectExplorer#expandPathInProjectExplorerAndOpenFile()" . Replace "."  on '"/" was applied in the previous versions for correct work with packages.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java
Ohrimenko1988: Without this `if` we will have `IndexOutOfBounds` exception in all cases

dmytro-ndp: I see.
What about simplifying method a little bit:
```java
    for (int i = 0; i < itemsPaths.size(); i++) {
      String currentItem = itemsPaths.get(i);
      int nextItemIndex = i + 1;
      
      // don't open if next item has already been visible
      if (nextItemIndex < itemsPaths.size()
          && seleniumWebDriverHelper.isVisible(By.xpath(format(PROJECT_EXPORER_ITEM_TEMPLATE, itemsPaths.get(nextItemIndex))))) {
        continue;
      }

      waitItem(currentItem);
      openItemByPath(currentItem);
    }
```

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java
dmytro-ndp: I see. But someone can try to reuse the method later and to get unexpected side effect if there is file name with extension at the end of path. 
Lets make it's obvious that the path is file less:
`splitPathOnSubitems(String path)` > `splitPathOnSubitems(String fileLessPath)`

`@param path Path in format "RootFolder/src/main/java/org.package".` 
\> 
`@param fileLessPath Path in format "RootFolder/src/main/java/org.package" without ending file name.`

dmytro-ndp (approved):

dmytro-ndp:
ci-build

artaleks9 (approved):

PR #8836, opened 2018-02-19
Make start of OpenShift machines parallel

### LINKED ISSUE #7067: Make bootstrapping of OpenShift machines parallel

akorneta:
### What does this PR do?
Parallelizes the start of OpenShift machines.

Schematically the process of machine boot described in the following picture:

parallel start of machines_

_failure context_ its a state that shared overall machine threads in the scope of one workspace, it's used as an indicator of error that might appear in process of the start. This context holds only one exception that was first thrown(and its is visible for all the threads).

**s1.** Creates OpenShift pod watcher that listens to pod events and if the state satisfies predicate then this step is considered as finished. If the connection is closed or the listening is cancelled then this step is considered as failed;
**s2, s4, s6** These steps are used to break the chain of invocation if there is an exception in the failure context(so if one of machine is failed we need to prevent the launching of other machines);
**s3.** Sets machine running status and propagates machine running event through event service.
**s5.** Asynchronously starts bootstrapping of installers and listen to bootstrapper status events, when an event with status `done` received then this step is considered as finished if the received event has status `failed` or listening is cancelled then this step is considered as failed.
**s7.** Asynchronously starts checking the readiness of servers of a machine, when servers are available then this step is considered as finished when this step is cancelled then it is considered as failed.

par_boot

fixes #7067 

#### Release Notes
n/a


akorneta:
ci-test-ocp

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/bootstrap/AbstractBootstrapper.java
```diff
@@ -28,6 +28,7 @@
  * @author Sergii Leshchenko
  */
 public abstract class AbstractBootstrapper {
+
   private final String machineName;
   private final int bootstrappingTimeoutMinutes;
```
sleshchenko: Since it's used only by one bootstrapping method (#bootstrap but not #bootstrapAsync), I would move it to parameters of the #bootstrap method. It'd make clear that fact that bootstrap async is not limited by time out of the box.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
```diff
@@ -151,16 +163,164 @@ protected void internalStart(Map<String, String> startOptions) throws Infrastruc
...
+  private void waitMachines(
+      List<CompletableFuture<Void>> machinesFutures,
+      List<CompletableFuture<?>> toCancelFutures,
+      CompletableFuture<Void> failure)
+      throws InfrastructureException, InterruptedException {
+    try {
+      final CompletableFuture<Void> allDone =
+          CompletableFuture.allOf(
+              machinesFutures.toArray(new CompletableFuture[machinesFutures.size()]));
+      CompletableFuture.anyOf(allDone, failure).get(machineStartTimeoutMin, TimeUnit.MINUTES);
```
sleshchenko: As far as I understand previously it was used for setting a timeout of waiting for one machine to become running (Pod should be running but installers and servers are not launched yet). Now within this time all machines must become running, installers must be launched and servers must become RUNNING. Please consider renaming configuration parameter. Now it looks like Kubernetes workspace start timeout.

garagatyi (commented):
Looks very interesting, but requires careful review. Good job!

akorneta:
ci-test-ocp

akorneta:
ci-test-ocp

akorneta:
ci-test-ocp

akorneta:
ci-test-ocp

riuvshin:
"ПУШКА"

akorneta:
ci-test-ocp

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesPods.java
```diff
@@ -199,6 +199,61 @@ public void onClose(KubernetesClientException cause) {
...
+   *       <li>connection problem occurred
+   *     </ul>
+   *     otherwise, it must be explicitly closed
+   */
+  public CompletableFuture<Void> watchReadinessAsync(String name, Predicate<Pod> predicate) {
+    final CompletableFuture<Void> podRunningFuture = new CompletableFuture<>();
+    try {
+      final PodResource<Pod, DoneablePod> podResource =
+          clientFactory.create().pods().inNamespace(namespace).withName(name);
+      final Watch watch =
```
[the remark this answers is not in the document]
akorneta: I think that there should not be any problems with WebSocket connection leaks because the method documents indicate that the future must be explicitly cancelled if it is not needed and has not reached the final state.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesMachine.java
```diff
@@ -113,4 +114,11 @@ public void waitRunning(int timeoutMin) throws InfrastructureException {
             timeoutMin,
             p -> (KUBERNETES_POD_STATUS_RUNNING.equals(p.getStatus().getPhase())));
   }
+
+  public CompletableFuture<Void> watchReadinessAsync() {
```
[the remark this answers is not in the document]
akorneta: I'll write doc for this method and will describe how this method might be used, to prevent leaks.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
```diff
@@ -120,22 +126,28 @@ protected void internalStart(Map<String, String> startOptions) throws Infrastruc
...
-          checkMachineServers(machine);
-        } catch (InfrastructureException rethrow) {
-          sendFailedEvent(machine.getName(), rethrow.getMessage());
-          throw rethrow;
-        }
+        final CompletableFuture<Void> machineRunningFuture = machine.watchReadinessAsync();
+        toCancelFutures.add(machineRunningFuture);
+        final CompletableFuture<Void> machineBootChain =
+            machineRunningFuture
+                .thenComposeAsync(checkFailure(failure), executor)
```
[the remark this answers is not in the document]
akorneta: makes sense, I'll change that.

multiuser/machine-auth/che-multiuser-machine-authentication/src/main/java/org/eclipse/che/multiuser/machine/authentication/server/MachineTokenProviderImpl.java
```diff
@@ -33,8 +32,7 @@ public MachineTokenProviderImpl(MachineTokenRegistry tokenRegistry) {
   }
 
   @Override
-  public String getToken(String workspaceId) {
```
mshaposhnik: I would rather left this method here and call getToken() with two params within it instead of adding 
`EnvironmentContext.getCurrent()...` in 15 other places

akorneta: yes, it makes sense. The reason why the method was removed is to force developers to explicitly call `EnvironmentContext.getCurrent()` in cases where they sure that user won't be anonymous, but I think it might be reworked in a better way, so I'll consider to back this method and adding a check(that user is not anonymous) before token retrieval.

akorneta:
ci-test

skabashnyuk (approved):

sleshchenko (approved):
Good job! :+1: 
It's mega cool feature, especially for agents that are launched as parallel machines.

mshaposhnik (approved):

akorneta:
ci-test

akorneta:
ci-test-ocp

garagatyi (approved):
Cool PR! Looks good.

Issue #7824, opened 2017-12-11
Remove properties and env vars with triggers that are not used in Che6 anymore


garagatyi:
### Description
In Che6 we have rewritten a lot of code. We haven't removed properties that are no longer in use. This may lead to a situation when a user configures something that won't work and contacts support or chats to figure out what is wrong. 
We need to cleanup `che.properties`, `che.env` template and docs regarding those properties. 

### Reproduction Steps

**OS and version:**    

**Diagnostics:**

PR #8872, opened 2018-02-22
Remove declaration of unused properties & environment variables

mkuznyetsov:
### What does this PR do?
Remove declaration of unused properties & environment variables

https://github.com/eclipse/che/issues/7824

### DIFF
--- a/dockerfiles/init/manifests/che.env
+++ b/dockerfiles/init/manifests/che.env
@@ -116,12 +116,6 @@
 #     Example: che.docker.securityopt=seccomp:unconfined,apparmor:unconfined
 #CHE_DOCKER_SECURITYOPT=seccomp:unconfined
 
-# Agent Start Timeout
-#     The length of time that a workspace will be allowed to boot before the system terminates the
-#     boot process. If the Che container cannot establish two way communications with the
-#     agents within the workspace when it boots, then the workspace will not be started.
-#CHE_WORKSPACE_AGENT_DEV_MAX__START__TIME__MS=300000
-
 # Workspace logs root
 #     Defines the directory inside the machine where all the workspace logs are placed.
 #     The value of this folder should be provided into machine e.g. like environment variable
@@ -170,9 +164,8 @@
 #    2. Else use the value of CHE_API
 #
 # Che Server --> Docker Daemon Progression:
-#    1. Use the value of CHE_DOCKER_DAEMON__URL
-#    2. Else, use the value of DOCKER_HOST system variable
-#    3. Else, use Unix socket over unix:///var/run/docker.sock
+#    1. Use the value of DOCKER_HOST system variable
+#    2. Else, use Unix socket over unix:///var/run/docker.sock
 #
 # Che Server --> Workspace Connection:
 
@@ -201,15 +194,8 @@
 #     Provides basic communication endpoint where clients can get/push statuses/output.
 #CHE_WEBSOCKET_ENDPOINT=ws://${CHE_HOST}:${CHE_PORT}/api/websocket
 
-# Docker Daemon URL
-#     How the Che server will discover the location of the Docker daemon. If this is
-#     not set, then Che will use DOCKER_HOST or the default unix:///var/run/docker.sock.
-#     It would be rare to need to set this as most clients set DOCKER_HOST or volume
-#     mount a Docker socket when starting the Che CLI.
-#CHE_DOCKER_DAEMON__URL=NULL
-
 # Docker Host
-#     How Che will connect to the Docker host if CHE_DOCKER_DAEMON__URL is not set. This
+#     How Che will connect to the Docker host. This
 #     value can be provided on the command line as well. If this value is not set, then
 #     Che will use the volume mount of unix:///var/run/docker.sock.
 #DOCKER_HOST=tcp://localhost:2375


benoitf:
@mkuznyetsov could you refine the title of the PR as it's used to generate changelog ?

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
```diff
@@ -41,10 +41,6 @@ che.workspace.no_proxy=
 # Configures dns resolving servers. May contain several values separated by comma sign.
 che.dns.resolvers=
```
sleshchenko: I think It's a good time to prefix everything related to docker infra with `che.infra.docker`. @garagatyi WDYT? Shall we create an issue for that?

garagatyi: I would love to. But what about backward compatibility? I think we've got something for that recently.

sleshchenko: Aliases mechanism can provide backward compatibility? Here is an PR where it was done https://github.com/eclipse/che/pull/8636

garagatyi:
@l0rd @ibuziuk in this PR @mkuznyetsov removes CHE_WORKSPACE_AUTO_START env var which was added by @sunix . Can you confirm that it is not needed anymore?

garagatyi (approved):
I'm OK with these changes but would like to hear from @ibuziuk or @l0rd whether this change is OK for Che on OSIO

benoitf (approved):

ibuziuk:
@garagatyi good catch this property is used on openshift.io / rh-che so it should be preserved [1] / [2]

[1] https://github.com/redhat-developer/rh-che/blob/master/openshift/rh-che.config.yaml#L24
[2] https://github.com/redhat-developer/rh-che/blob/master/openshift/rh-che.app.yaml#L151

garagatyi:
@ibuziuk the problem which @mkuznyetsov faced is that it seems that this property doesn't work anymore. So he decided to remove it. 
Whereas I was not sure whether we should remove it or restore. Do you know whether it is needed for any particular flow on OSIO?

l0rd:
On OSIO every workspace should be started by che-starter (that makes some checks, inject tokens, variables etc...). We should avoid the automatic creation of workspace hijacking che-starter. So that should be fixed (in this PR or in another one but please open an issue on rh-che github repo).

garagatyi (changes requested):
@mkuznyetsov Looks like the part about workspace auto start has to be left. Sorry for the inconvenience.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
```diff
@@ -41,10 +41,6 @@ che.workspace.no_proxy=
...
 
-# By default, when users access to a workspace with its URL the workspace
-# automatically starts if it is stopped. You can set this to false to disable this.
-che.workspace.auto_start=true
```
slemeur: Why have we removed that capability.
If removed it should have been deprecated !
cc @skabashnyuk

dockerfiles/init/manifests/che.env
```diff
@@ -116,12 +116,6 @@
 #     Example: che.docker.securityopt=seccomp:unconfined,apparmor:unconfined
 #CHE_DOCKER_SECURITYOPT=seccomp:unconfined
 
-# Agent Start Timeout
-#     The length of time that a workspace will be allowed to boot before the system terminates the
-#     boot process. If the Che container cannot establish two way communications with the
-#     agents within the workspace when it boots, then the workspace will not be started.
-#CHE_WORKSPACE_AGENT_DEV_MAX__START__TIME__MS=300000
```
slemeur, line 123: similar question what is the reason for removing this?

slemeur:
Docs?

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
skabashnyuk: Not clear what do you want? 
Make investigation why it was removed? Restore it in GWT ide?

dockerfiles/init/manifests/che.env
skabashnyuk: @mkuznyetsov didn't found any code that is using it?

skabashnyuk:
@slemeur what kind of docs you expect. 
Like we found that these X properties are not used and we are removing it?

dockerfiles/init/manifests/che.env
mkuznyetsov: this property has never been used in Che 6, it was used in WsAgentLauncher.class in Che 5.

I guess it can be removed from here as well, for consistency
https://github.com/eclipse/che-docs/blob/master/src/main/pages/setup-docker/docker-config.md

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
mkuznyetsov: Ok, this property is going to stay, and I fixed it's missing functionality

mkuznyetsov:
`che.workspace.auto_start=true` will remain, and it's functionality has been fixed

garagatyi (approved):
@mkuznyetsov can you correct the PRs title - since you restored some functionality it should be reflected in the title which will be used for the changelog.

mkuznyetsov:
@garagatyi I'll move the "restoring" to the other PR for clarity: https://github.com/eclipse/che/pull/8913

garagatyi:
@mkuznyetsov nice idea!

mkuznyetsov:
@slemeur I'll make Pr to Che Docs and remove mentions of unused properties

What I'm removing so far:

CHE_WORKSPACE_AGENT_DEV_MAX__START__TIME__MS - no longer used
CHE_DOCKER_DAEMON__URL - this does nothing, looks like it was supposed to override this property, but it has different (old-styled) name, and we've got no aliases to that:
https://github.com/eclipse/che/blob/master/infrastructures/docker/docker-client/src/main/java/org/eclipse/che/infrastructure/docker/client/DockerConnectorConfiguration.java#L88

mkuznyetsov:
Although, "CHE_DOCKER_DAEMON__URL" sounds like a bug, would rather have it returned as another issue https://github.com/eclipse/che/issues/8935

slemeur:
@mkuznyetsov : Thanks that's actually what I was expecting for the docs.

mkuznyetsov:
https://github.com/eclipse/che-docs/pull/364
https://github.com/eclipse/che-docs/pull/363
That should be all necesarry changes to the docs regarding all of these properties

skabashnyuk (approved):

PR #8892, opened 2018-02-23
Allow configuring the dispatcher and connection pool of the `KubernetesClient` shared underlying http client.

davidfestal:
### What does this PR do?

Allow configuring the dispatcher and connection pool of
the `KubernetesClient` shared underlying http client.

For example, default values are 64 concurrent asynchronous
requests, and **only 5 concurrent asynchronous requests per-host**,
which obviously doesn't seem correct for multi-user scenarios,
knowing that Che keeps a number of connections opened
(e.g. for commands or ws-agent logs)

Leaving these wrong default values can lead to workspace
starts being stuck in the middle of the process while waiting 
for an available request, as soon as several users start
a workspace on the same che-server.

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
@@ -431,7 +431,7 @@ che.infra.kubernetes.client.http.connection_pool.max_idle=5
 # Keep-alive timeout of the connection pool
 # of the Kubernetes-client shared http client
 # in minutes
-che.infra.kubernetes.client.http.connection_pool.keep_alive.mins=5
+che.infra.kubernetes.client.http.connection_pool.keep_alive_min=5
 
 ### OpenShift Infra parameters
 #
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
@@ -413,6 +413,25 @@ che.infra.kubernetes.ingress.annotations_json=NULL
 che.infra.kubernetes.pod.security_context.run_as_user=NULL
 che.infra.kubernetes.pod.security_context.fs_group=NULL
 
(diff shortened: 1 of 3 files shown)


davidfestal:
ci-test-ocp

skabashnyuk:
Maybe we should move ```httpClient``` creation in dedicated class ```HttpClientProvider``` it will reduce the size of KubernetesClientFactory constructor?

davidfestal:
> Maybe we should move httpClient creation in dedicated class HttpClientProvider it will reduce the size of KubernetesClientFactory constructor?

Why not in a next PR ;-)

ibuziuk (approved):

l0rd (approved):

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
```diff
@@ -413,6 +413,25 @@ che.infra.kubernetes.ingress.annotations_json=NULL
...
+che.infra.kubernetes.client.http.async_requests.max_per_host=1000
+
+# Max number of idle connections in the connection pool
+# of the Kubernetes-client shared http client 
+che.infra.kubernetes.client.http.connection_pool.max_idle=5
+
+# Keep-alive timeout of the connection pool
+# of the Kubernetes-client shared http client
+# in minutes
+che.infra.kubernetes.client.http.connection_pool.keep_alive.mins=5
```
garagatyi: We practice separation of measurement unit by underscore instead of a dot

garagatyi (approved):

akorneta (approved):

sleshchenko (approved):

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
davidfestal: Fixed in commit https://github.com/eclipse/che/pull/8892/commits/e88a2649bb2f9387b40739f7bcc52941ec2a8fac

l0rd (approved):

PR #8899, opened 2018-02-23
Add fmt:format and sortpom:sort

### LINKED ISSUE #8897: Default add fmt:format & sortpom:sort to Deploy IDE
### Description
Took me about 2 Hours to figure out as a Maven newbe.

As the full build Process aka Deploy IDE takes about 15Minutes on a strong machine and the Errors are realy showing up close to end.
It would be nice to by default add fmt:format & sortpom:sort to the  `Deploy IDE` Command. As such Errors then get corrected automaticly and don't let the Build fail after 10 Minutes in the running

```
cd ${current.project.eldest.parent.path}
 mvn -am -pl :assembly-ide-war  sortpom:sort fmt:format install 
 cp ${current.project.eldest.parent.path}/assembly/assembly-ide-war/target/*.war /home/user/tomcat8/webapps/ide.war
```

Additional:
Does it really take to always add clean?
Is it not enugh to build all changed files?

**OS and version:**    
Che in Che

gausa:
Let Maven automaticly fix Pom sorting and File Formatting
Prevents Buildfail on autofixables

### What does this PR do?

Fixes #8897

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/resources/che-in-che.json
+++ b/assembly/assembly-wsmaster-war/src/main/resources/che-in-che.json
@@ -83,7 +83,7 @@
           }
         },
         {
-          "commandLine": "cd ${current.project.eldest.parent.path}\n mvn gwt:codeserver -pl :assembly-ide-war -am -Psdm-in-che -Dskip-enforce -Dskip-validate-sources -Dche.dto.skip -Dmaven.main.skip -Dmaven.resources.skip -Dmaven.buildNumber.skip",
+          "commandLine": "cd ${current.project.eldest.parent.path} && mvn gwt:codeserver -pl :assembly-ide-war -am -Psdm-in-che -Dskip-enforce -Dskip-validate-sources -Dche.dto.skip -Dmaven.main.skip -Dmaven.resources.skip -Dmaven.buildNumber.skip",
           "name": "GWT SDM",
           "type": "gwt_sdm_che",
           "attributes": {
@@ -92,7 +92,7 @@
           }
         },
         {
-          "commandLine": "cd /home/user/tomcat8/bin\n export SERVER_PORT=8080\n export JPDA_ADDRESS=9000\n ./catalina.sh jpda run",
+          "commandLine": "cd /home/user/tomcat8/bin && export SERVER_PORT=8080 && export JPDA_ADDRESS=9000 && ./catalina.sh jpda run",
           "name": "Tomcat8-IDE Start",
           "type": "custom",
           "attributes": {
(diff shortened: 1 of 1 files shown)


skabashnyuk:
can you explain a motivation of this pr?
```fmt:format and sortpom:sort``` this kind of actions is modifying sources behind the scene. As for me it's not very good to execute this command as part of other command not related to the subject of this pr. It may confuse users. What if it will be separate command? Like "Sort&LicenseCheck&Format"?

**Comment by gausa** [comment]
@skabashnyuk Yeah I know. But it should only sort the Code and not really change the essence of it, does it?
Nicest thing I could think of is that it does the Fixes it can do by it self first. Instead of failing the Build after 10Minutes in the running.
But if we can't find a nice "doing it without knowing to have to do it" solution I'd be fine to setup a seperate Command.

(I just think there will be lots of Dev's unintentionally skipping that explicit Command and then hit head on desk ;-) )

Open for oppinions and thanks!

**Comment by skabashnyuk** [comment]
+1 for separate command.

**Comment by riuvshin** [comment]
yeah, +1 for separate command

**Comment by gausa** [comment]
Just to mention:
If I Create an Command in Che as:
```cd ${current.project.eldest.parent.path}
 mvn sortpom:sort fmt:format```
Which is equal to the in Config used:
```cd ${current.project.eldest.parent.path}\n mvn sortpom:sort fmt:format```

I get:
```
/bin/bash: line 0: cd: /projects/che
: No such file or directory
```

But the already by Config created Commands work fine.

I always have to use 
```cd ${current.project.eldest.parent.path} && mvn sortpom:sort fmt:format```
while creating a new Command in Che to get the desired results.

Could that be a Bug?

**Comment by skabashnyuk** [comment]
I think this is an expected behavior. Try to execute this command 
```cd / \n mvn sortpom:sort fmt:format```
 in your terminal

File: assembly/assembly-wsmaster-war/src/main/resources/che-in-che.json [hunk 1]
```diff
@@ -135,6 +135,15 @@
             "goal": "Run",
             "previewUrl": ""
           }
+        },
+        {
+          "commandLine": "cd ${current.project.eldest.parent.path}\n mvn sortpom:sort fmt:format",
```
skabashnyuk: cd ${current.project.eldest.parent.path} && mvn sortpom:sort fmt:format

gausa:
Yes, +1 I'd expect `cd / \n mvn sortpom:sort fmt:format` to fail badly but as example, in the Commanddefinition for this Workspace it's:
```
"commandLine": "cd ${current.project.eldest.parent.path}\n mvn clean install -pl :assembly-wsagent-war -am\n cp ${current.project.eldest.parent.path}/assembly/assembly-wsagent-war/target/*.war /home/user/tomcat8/webapps/ROOT.war",
"name": "Deploy Workspace Agent",
```

where `\n` as Separator works fine

skabashnyuk:
strange. It's suppose to fail

gausa:
If the Interpreter correctly sees this as a separator and executes the Commands in a Row it's perfectly fine.
Which it obviously does.

If you create a Command via the Webinterface but not the Configfile it look like it's getting confused by the Windows'ish \r

gausa:
So, how do we want to proceed here?

Both variants are valid in my opinion.
differ only in that && is more safe in stopping on error while \n or would just run each command no matter if fail or not

skabashnyuk:
I prefer ```&&```

skabashnyuk (approved):

skabashnyuk:
ci-test

PR #8924, opened 2018-02-27
Fix context propagation for bootstrap and server checks

akorneta:
### What does this PR do?
After the merge of https://github.com/eclipse/che/pull/8836, we got a problem with start of workspaces in **rh-che**. The reason for these failures is there are no token in environment context while a machine is bootstrapping. The absence of the token in context causes the kubernetes client to attempt to make requests on behalf of an anonymous user.

### DIFF
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
@@ -56,6 +56,7 @@ import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
 import org.eclipse.che.api.workspace.server.spi.InternalInfrastructureException;
 import org.eclipse.che.api.workspace.server.spi.InternalRuntime;
 import org.eclipse.che.api.workspace.server.spi.environment.InternalMachineConfig;
+import org.eclipse.che.commons.env.EnvironmentContext;
 import org.eclipse.che.workspace.infrastructure.kubernetes.bootstrapper.KubernetesBootstrapperFactory;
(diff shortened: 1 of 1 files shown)


davidfestal (approved):
With the change it seems to provide the right environment context to the threads that effectively perform the workspace start.
And multi-tenant `rh-che` on Che6 is not broken anymore.
So approving even if I didn't catch fully how environment context propagation works.

sleshchenko (approved):
LGTM

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
```diff
@@ -225,40 +228,47 @@ private void waitMachines(
...
-      final ServersChecker serverCheck =
-          serverCheckerFactory.create(runtimeId, machineName, machine.getServers());
-      final CompletableFuture<?> serversReadyFuture;
       try {
-        serversReadyFuture = serverCheck.startAsync(new ServerReadinessHandler(machineName));
-        toCancelFutures.add(serversReadyFuture);
-        serversAndProbesFuture.whenComplete((ok, ex) -> serversReadyFuture.cancel(true));
-      } catch (InfrastructureException ex) {
-        serversAndProbesFuture.completeExceptionally(ex);
+        EnvironmentContext.setCurrent(context);
```
garagatyi: What if there would be a method that wraps anything with context setting. Would the code be simpler with less indentation (which leads to complexity of code reading)?
Example:
```java
.thenCompose(setContext(context, bootstrap(toCancelFutures, machine)))
```

akorneta: sure, that makes sense, I'll think about that

skabashnyuk (approved):

garagatyi (approved):

PR #8823, opened 2018-02-17
Revert "Fix orion document changing event getting text from native orion event instead of guessing it"

vparfonov:
### What does this PR do?
This reverts commit 523b823dc991e6b7d4b955f61f29adf4c86923ad.

### What issues does this PR fix or reference?
#8812

PR #8942, opened 2018-02-27
Fix orion document changing event getting text from native orion event instead of guessing it

sunix:
### What does this PR do?
Fix orion document changing event getting text from native orion event instead of guessing it.
"Changing" events were getting their new changed text from the document based on the new text range positions. While this way of doing works fine for "changed" events, it doesn't work for "changing" events as the change has not been performed yet to the document yet. The result is that the event send the previous text instead of the new added text. This fix is getting the new text from native orion event data object.

Reapplying https://github.com/eclipse/che/pull/8823 without altering model changed event. (just use getText in model changing event).

### DIFF
--- a/ide/che-core-orion-editor/src/main/java/org/eclipse/che/ide/editor/orion/client/OrionDocument.java
+++ b/ide/che-core-orion-editor/src/main/java/org/eclipse/che/ide/editor/orion/client/OrionDocument.java
@@ -104,7 +104,7 @@ public class OrionDocument extends AbstractDocument {
     int addedCharCount = param.addedCharCount();
     int removedCharCount = param.removedCharCount();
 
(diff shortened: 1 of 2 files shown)


benoitf:
ci-test

sunix:
tests look right ?

musienko-maxim:
Tests look good. But some fails look like the PR has not been merged with last changes in the master branch. If you have time would be nice merge PR and launch tests again. But if it urgent i think it can be merged

sunix:
@musienko-maxim I've rebased the branch, could we rerun the test ?

sunix:
ci-test

benoitf:
ci-test

benoitf (approved):

Issue #8950, opened 2018-02-28
Transform locators from "CodenvyEditor" page object to enumerations, and remove redundant locators


Ohrimenko1988:
### Description
We have redundant locators in the "CodenvyEditor" page object. We have to remove these locators. Besides, locators which define items from editor's context menu and editor's markers types are grouped in abstract classes like "String" constants. They should be moved to the enumerations. And related to this constants methods also should be changed according to this changes.

Ohrimenko1988:
Done https://github.com/eclipse/che/pull/8961

PR #8961, opened 2018-02-28
Selenium: Transform locators from "CodenvyEditor" page object to enumerations and rework related methods

Ohrimenko1988:
### What does this PR do?
Transform locators from "CodenvyEditor" page object to enumerations and rework related methods

Issue: https://github.com/eclipse/che/issues/8950

### DIFF
--- a/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
+++ b/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
@@ -186,7 +186,7 @@ public class CodenvyEditor {
   }
 
   /** Editor`s markers types */
-  public enum Marker {
+  public enum MarkerLocator {
     ERROR_OVERVIEW("//div[@class='ruler annotations']/div[@class='annotation error']"),
     WARNING_OVERVIEW("//div[@class='ruler annotations']/div[@class='annotation warning']"),
     TASK_OVERVIEW("//div[@class='ruler annotations']/div[@class='annotation task']"),
@@ -196,7 +196,7 @@ public class CodenvyEditor {
 
     private final String itemLocator;
 
-    Marker(String itemLocator) {
+    MarkerLocator(String itemLocator) {
       this.itemLocator = itemLocator;
     }
 
@@ -613,10 +613,10 @@ public class CodenvyEditor {
   /**
    * Waits a marker with specified {@code markerType} on the defined {@code position}
    *
-   * @param markerType marker's type, defined in {@link Marker}
+   * @param markerType marker's type, defined in {@link MarkerLocator}
    * @param position line`s number where marker is expected
    */
-  public void waitMarkerInPosition(Marker markerType, int position) {
+  public void waitMarkerInPosition(MarkerLocator markerType, int position) {
     elemDriverWait.until(visibilityOfElementLocated(By.xpath(format(markerType.get(), position))));
     setCursorToLine(position);
     expectedNumberOfActiveLine(position);
@@ -708,10 +708,10 @@ public class CodenvyEditor {
   /**
    * Waits marker with specified {@code markerType} on the defined {@code position} and click on it
    *
-   * @param markerType marker's type, defined in {@link Marker}
+   * @param markerType marker's type, defined in {@link MarkerLocator}
    * @param position line's number, where marker is expected
    */
-  public void waitMarkerInPositionAndClick(Marker markerType, int position) {
+  public void waitMarkerInPositionAndClick(MarkerLocator markerType, int position) {
     loadPageDriverWait
         .until(visibilityOfElementLocated(By.xpath(format(markerType.get(), position))))
         .click();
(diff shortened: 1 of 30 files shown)


selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
```diff
@@ -237,18 +181,58 @@ private Locators() {}
...
+  /** Editor`s markers types */
+  public enum MarkersType {
+    ERROR_MARKER_OVERVIEW("//div[@class='ruler annotations']/div[@class='annotation error']"),
+    WARNING_MARKER_OVERVIEW("//div[@class='ruler annotations']/div[@class='annotation warning']"),
+    TASK_MARKER_OVERVIEW("//div[@class='ruler annotations']/div[@class='annotation task']"),
+    ERROR_MARKER("//div[@class='ruler overview']/div[@class='annotationOverview error']"),
+    WARNING_MARKER("//div[@class='ruler overview']/div[@class='annotationOverview warning']"),
+    INFO_MARKER("//div[@class='annotationHTML info']");
+
+    private final String itemLocator;
```
dmytro-ndp: What about returning `By locator`?

Ohrimenko1988: In my opinion, it should be "String". It gives us flexible. Because locator can be of the different types (By.id, By.name, By.xpath, By.css and other). And "String" can contain all of them.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
```diff
@@ -237,18 +181,58 @@ private Locators() {}
     }
   }
 
-  @FindBy(id = Locators.EDITOR_TABS_PANEL)
-  private WebElement editorTabsPanel;
+  /** Editor`s markers types */
+  public enum MarkersType {
```
dmytro-ndp: It seems type of ordinary type, not a bunch of types. So, let's name it `MarkerLocator`.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
```diff
@@ -237,18 +181,58 @@ private Locators() {}
...
+    ERROR_MARKER_OVERVIEW("//div[@class='ruler annotations']/div[@class='annotation error']"),
```
dmytro-ndp: `MARKER` is redundant here so as it is has already been a type of Marker.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
```diff
@@ -237,18 +181,58 @@ private Locators() {}
...
+      this.itemLocator = itemLocator;
+    }
+
+    private String get() {
+      return this.itemLocator;
+    }
+  }
+
+  /** Editor's context menu items */
+  public enum EditorContextMenu {
```
dmytro-ndp: `Editor` could be omitted in _enum_ name so as it's part of `CodenvyEditor` class.

dmytro-ndp:
I suggest to add prefix "_Locator_" to the name of enums which hold locators to make it obviouse what that enums are about, for example: `MarkerLocator`.

artaleks9:
good, 
it will be nice to run test suite

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
Ohrimenko1988: No, this is exactly different marker's types. But I agree that name is not so very accurate. I'll change it to "MarkerTypes".

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
```diff
@@ -613,10 +613,10 @@ public void waitTextIntoAutocompleteContainer(final String value) {
   /**
    * Waits a marker with specified {@code markerType} on the defined {@code position}
    *
-   * @param markerType marker's type, defined in {@link Marker}
+   * @param markerType marker's type, defined in {@link MarkerLocator}
    * @param position line`s number where marker is expected
    */
-  public void waitMarkerInPosition(Marker markerType, int position) {
+  public void waitMarkerInPosition(MarkerLocator markerType, int position) {
```
dmytro-ndp: `markerType` > `markerLocator`

dmytro-ndp:
ci-test

dmytro-ndp (approved):

artaleks9 (approved):

Issue #8962, opened 2018-02-28
Rework methods from "CodenvyEditor" page object with using "SeleniumWebDriverHelper" and "WebDriverWaitFactory" classes


Ohrimenko1988:
### Description
In the "CodenvyEditor" page object we have a lot of code like this:
```
new WebDriverWait(seleniumWebDriver, customTimeout)
```

and 

```
loadPageDriverWait
        .until(visibilityOfElementLocated(By.xpath(format(Locators.TAB_FILE_CLOSE_ICON, fileName))))
        .click();
```
They should be replaced by methods wich defined in "SeleniumWebDriverHelper" and "WebDriverWaitFactory" classes. 

for the first example:

```
webDriverWaitFactory.get(customTimeout)
```

for the second example:

```
seleniumWebDriverHelper.waitAndClick(By.xpath(format(Locators.TAB_FILE_CLOSE_ICON, fileName)));

```

This is the second iteration of changes, the first iteration was #8854, and next iteration will be check methods logic, optimize it and remove unnecessary methods.

Ohrimenko1988:
Done https://github.com/eclipse/che/pull/8980

PR #8980, opened 2018-03-01
Selenium: Rework methods from "CodenvyEditor" page object with using "SeleniumWebDriverHelper" and "WebDriverWaitFactory" classes

Ohrimenko1988:
### What does this PR do?
Rework methods from "CodenvyEditor" page object with using "SeleniumWebDriverHelper" and "WebDriverWaitFactory" classes

Issue: https://github.com/eclipse/che/issues/8962

### DIFF
--- a/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
+++ b/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
@@ -186,7 +186,7 @@ public class CodenvyEditor {
   }
 
   /** Editor`s markers types */
-  public enum Marker {
+  public enum MarkerLocator {
     ERROR_OVERVIEW("//div[@class='ruler annotations']/div[@class='annotation error']"),
     WARNING_OVERVIEW("//div[@class='ruler annotations']/div[@class='annotation warning']"),
     TASK_OVERVIEW("//div[@class='ruler annotations']/div[@class='annotation task']"),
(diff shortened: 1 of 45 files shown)


selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
```diff
@@ -617,16 +656,16 @@ public void waitTextIntoAutocompleteContainer(final String value) {
...
     setCursorToLine(position);
     expectedNumberOfActiveLine(position);
   }
 
   /** Wait for no Git change markers in the opened editor. */
   public void waitNoGitChangeMarkers() {
-
-    new WebDriverWait(seleniumWebDriver, LOAD_PAGE_TIMEOUT_SEC)
+    webDriverWaitFactory
+        .get(DEFAULT_TIMEOUT)
```
dmytro-ndp: We can omit _DEFAULT_TIMEOUT_ parameter by creation `WebDriverWaitFactory#get()` method which uses `DEFAULT_TIMEOUT` inside.

Ohrimenko1988:
ci-test

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
Ohrimenko1988: Good point, thanks for the idea.

dmytro-ndp (approved):

SkorikSergey (approved):

dmytro-ndp (approved):

Issue #8922, opened 2018-02-27
Study possible ways to replicate JPA cache (EclipseLink based) betwee master instances


mshaposhnik:
### Description
To enable parallel run of several master instances we need to synchonize state of the JPA caches to avoid data inconsistency.

Docs:
[1] https://wiki.eclipse.org/EclipseLink/Examples/JPA/CacheCoordination
[2] http://www.eclipse.org/eclipselink/documentation/2.5/concepts/cache011.htm

PR #8982, opened 2018-03-02
Configure JPA L2 cache coordination between instances on Openshift®  / Kubernetes

mshaposhnik:
### What does this PR do?
Adds JGroups based coordination for EclipseLink L2 caches for multiple master instances on openshift or K8S infra.

https://github.com/eclipse/che/issues/8922

#### Release Notes
Added JGroups based coordination for EclipseLink L2 caches for multiple master instances on openshift or K8S infra.

#### Docs PR
N/A


assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/jgroups/che-tcp.xml
```diff
@@ -0,0 +1,54 @@
...
+    max_bundle_size="64K"
+    enable_diagnostics="true"
+    thread_naming_pattern="cl"
+
+    thread_pool.min_threads="0"
+    thread_pool.max_threads="500"
+    thread_pool.keep_alive_time="30000" />
+
+  <kubernetes.KUBE_PING
+    namespace="${OPENSHIFT_KUBE_PING_NAMESPACE:eclipse-che}"
```
sleshchenko: As far as I understand it is used for Kubernetes and OpenShift infrastructures. So let's rename env var to `KUBERNETES_KUBE_PING_NAMESPACE`, or maybe even `CHE_INFRA_KUBERNETES_KUBE__PING__NAMESPACE` to make it clear that it is infrastructure specific configuration property.

dockerfiles/init/modules/openshift/files/scripts/che-openshift.yml
```diff
@@ -64,6 +68,8 @@ items:
             name: http
           - containerPort: 8000
             name: http-debug
+          - containerPort: 8888
```
sleshchenko: Is it default value? Is it a different property from [TCP_PORT](https://github.com/eclipse/che/pull/8982/files#diff-be9e46711b86b2abcc7f8a90858249b3R18)?

mshaposhnik: Yes, thats different one needed  for KUBE_PING.

sleshchenko: Ok, thanks

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/jgroups/che-tcp.xml
mshaposhnik: I think we can remove it here at all. But we still should pass it to container as is:
See
https://github.com/jgroups-extras/jgroups-kubernetes/blob/0.9.3/kube/src/main/java/org/jgroups/ping/kube/KubePing.java#L96
+
https://github.com/jgroups-extras/jgroups-kubernetes/blob/0.9.3/kube/src/main/java/org/jgroups/ping/kube/KubePing.java#L131
I will try to remove it from here.

sleshchenko: Up to you, but I'd define our own documented property into che.properties and che.env.

dockerfiles/init/modules/openshift/files/scripts/che-openshift.yml
```diff
@@ -48,6 +48,10 @@ items:
       spec:
         containers:
         - env:
+          - name: OPENSHIFT_KUBE_PING_NAMESPACE
```
sleshchenko: Shouldn't be added the same changes to https://github.com/eclipse/che/blob/master/dockerfiles/init/modules/kubernetes/files/che-kubernetes.yaml? 
Also maybe kubernetes helm plugin should be updated accordingly.

@garagatyi @l0rd I assume it may affect minishift addon, should it be updated or the corresponding issue created?

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/jgroups/che-tcp.xml
mshaposhnik: For me it seems we can't use own prop-s here.

sleshchenko: Can you describe why we can't? Is not it just name of environment property that should be respected?

mshaposhnik: Checked the jgroups code. We can use custom system property. But env.variable should be named as in the links above.

assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterModule.java
```diff
@@ -264,7 +264,15 @@ private void configureSingleUserMode() {
     bind(org.eclipse.che.security.oauth.OAuthAuthenticationService.class);
   }
 
-  private void configureMultiUserMode() {
+  private void configureMultiUserMode(
+      Map<String, String> persistenceProperties, String infrastructure) {
+    if (OpenShiftInfrastructure.NAME.equals(infrastructure)
```
garagatyi: This hardcoding of infrastructure names makes Che less extensibility friendly. I would recommend using another approach where Che deployment can specify environment variable to configure JGROUPS. @mshaposhnik @skabashnyuk @sleshchenko WDYT?

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/jgroups/che-tcp.xml
```diff
@@ -0,0 +1,53 @@
...
+    which accompanies this distribution, and is available at
+    http://www.eclipse.org/legal/epl-v10.html
+
+    Contributors:
+      Red Hat, Inc. - initial API and implementation
+
+-->
+<config xmlns="urn:org:jgroups"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="urn:org:jgroups http://www.jgroups.org/schema/JGroups-3.1.xsd">
```
garagatyi, line 15: Can you elaborate on whether there is a way to tweak this configuration in case service delivery team need to change something?

dockerfiles/init/modules/openshift/files/scripts/che-openshift.yml
garagatyi: I can't find where this env var is used. @mshaposhnik @sleshchenko can you guys point me to it?

dockerfiles/init/modules/openshift/files/scripts/che-openshift.yml
garagatyi: @mshaposhnik can you elaborate on how it is used? It is not clear for me.

mshaposhnik: Pls read https://github.com/jgroups-extras/jgroups-kubernetes/tree/0.9.3

dockerfiles/init/modules/openshift/files/scripts/che-openshift.yml
mshaposhnik: Pls read https://github.com/jgroups-extras/jgroups-kubernetes/tree/0.9.3

assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterModule.java
mshaposhnik: There is a lot of ohter places where it is configured in a such way.  It is not easy to refactor them all. Probably it's another big issue.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/jgroups/che-tcp.xml
mshaposhnik: We can probably move filename into property. Then, changing conf will be as simple as put new xml conf file under classpath and change that property. That's a good point.

mshaposhnik: Or no... that's already module. Need to find out way. Maybe discuss with @riuvshin

assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterModule.java
skabashnyuk: @garagatyi how do you propose to implement in another way?

dockerfiles/init/modules/openshift/files/scripts/che-openshift.yml
garagatyi: @mshaposhnik  If there is no service for this port how this port can be accessed from other pods?

dockerfiles/init/modules/openshift/files/scripts/che-openshift.yml
garagatyi: Thank you

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/jgroups/che-tcp.xml
garagatyi: I see, thank you.

assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterModule.java
garagatyi: @skabashnyuk having an environment variable that enables what now is enabled by the name of the active infrastructure.

garagatyi (commented):
@mshaposhnik Is this already tested on a real cluster of it is just a POC?

dockerfiles/init/modules/openshift/files/scripts/che-openshift.yml
mshaposhnik: Have you read the doc? there IS a service...
```
On a discovery request, Kubernetes returns list of 3 IP addresses. 
JGroups will use embedded HTTP Server exposed on port 8888 by default 
(see below for configuration) and will send 
HTTP based requests to each od them.
```

garagatyi: Yes, I read the docs. But I mean kubernetes service. Without it there is no access to a port of a pod.

sleshchenko: I think there is access if you know IP for a container. I think Kubernetes JGroups discovers pods/containers IPs itself.
If you access by service, actually you don't know which pod are you request if replication is more than 1. So, it's impossible to use service at all.
Maybe I'm wrong somewhere, but I imagine it in this way.

garagatyi: Yeah, agree about service. 

But how the lib accesses k8s/OS API? which token or SA is used?

mshaposhnik: You can set it into ENV or default will be used from here:
https://github.com/jgroups-extras/jgroups-kubernetes/blob/0.9.3/kube/src/main/java/org/jgroups/ping/kube/KubePing.java#L88

garagatyi: Then we need to be sure that this approach works in both kubernetes and openshift before the merge.

mshaposhnik:
Checked it works on K8S:
1. JPA set connection between nodes:
kubernetes_jgroups
2. Cache propagation :
https://drive.google.com/file/d/1aXkfi7LxR0pkYsQhOson6OaCguri_-w6/view?usp=sharing

garagatyi:
@mshaposhnik Can you also check on Openshift?

mshaposhnik:
It was developed on Openshift, because it is a primary goal of the issue. Kubernetes is just  like an free bonus :)

garagatyi:
@mshaposhnik thank you for the clarification!

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/jgroups/che-tcp.xml
```diff
@@ -0,0 +1,53 @@
...
+    max_bytes="4M"/>
+  <pbcast.GMS print_local_addr="true" join_timeout="3000"
+
+    view_bundling="true"/>
+  <UFC max_credits="2M"
+    min_threshold="0.4"/>
+  <MFC max_credits="2M"
+    min_threshold="0.4"/>
+  <FRAG2 frag_size="60K"/>
+  <!--RSVP resend_interval="2000" timeout="10000"/-->
```
sleshchenko: Do we need this commented configuration line?

sleshchenko (approved):
LGTM

It would be better to name environment configuration as Infrastructure specific property, but I'm OK to merge it as it is because looks like there is no a way to define custom env variable name for `kubernetes-jgroups` library.

riuvshin (approved):

garagatyi:
@mshaposhnik I'm sorry but I've merged PR that moves deployment files to another folder. This has affected your PR. I think that the easiest way to adapt your PR is to rebase it against master and then move files that would be left from `dockerfiles/init/modules/che-kubernetes-helm` to `deploy/kubernetes/helm/che`. I hope you won't get any rebase conflicts.

skabashnyuk (approved):

ghost (approved):

mshaposhnik:
ci-test

PR #8983, opened 2018-03-02
Enable single-level DNS domains (*.domain.tld) in SINGLE_PORT mode to simplify HTTPS setup

hkolvenbach:
### What does this PR do?
This PR changes the behavior how DNS names for `SINGLE_PORT` mode are generated, if the new ENV variable `CHE_SINGLEPORT_CUSTOM_DNS` is set to `true`. It will then generate DNS names with underscores instead of dots, such that all subdomains are conforming to `*.domain.tld`, which is important for a HTTPS setup using a wildcard certificate (usually only `*.domain.tld` is supported by wildcard certificates, not `*.*.domain.tld` and so on). This requires a wildcard DNS entry for your domain in the form of `*.domain.tld` and **won't** use any service like `nip.io`.

`CHE_SINGLEPORT_WILDCARD__DOMAIN_HOST` needs to be set to `domain.tld` in that case.

Work to do:
- change the Java part to enable this only if the  `CHE_SINGLEPORT_CUSTOM_DNS` flag is set (currently hardcoded)

I would like to get feedback if this is something upstream che would want. If yes, I will solve the outstanding issues and write documentation for it. It would certainly make HTTPS setup easier for everyone and I would not need to maintain a patched fork of che.

#### Release Notes

SINGLE_PORT mode now supports custom single level DNS domains `*.domain.tld`

#### Docs PR

https://github.com/eclipse/che-docs/pull/371

### DIFF
--- a/dockerfiles/init/manifests/che.pp
+++ b/dockerfiles/init/manifests/che.pp
@@ -40,6 +40,8 @@ node default {
   #
   $che_single_port = getValue("CHE_SINGLE_PORT","false")
   $che_single_port_wildcard_domain_host = getValue("CHE_SINGLEPORT_WILDCARD__DOMAIN_HOST","nip.io")
+  $che_single_port_custom_dns = getValue("CHE_SINGLEPORT_CUSTOM_DNS","false")
+
 
   ###############################
   # Che multiuser
@@ -66,6 +68,7 @@ node default {
   $system_super_privileged_mode=getValue("SYSTEM_SUPER__PRIVILEGED__MODE", "false")
 
   $che_keycloak_admin_require_update_password=getValue("CHE_KEYCLOAK_ADMIN_REQUIRE_UPDATE_PASSWORD", "true")
+  $che_keycloak_auth__server__url=getValue("CHE_KEYCLOAK_AUTH__SERVER__URL","")
 
   ###############################
   # Include base module
--- a/dockerfiles/init/manifests/che.pp
+++ b/dockerfiles/init/manifests/che.pp
@@ -67,7 +67,6 @@ node default {
   $system_super_privileged_mode=getValue("SYSTEM_SUPER__PRIVILEGED__MODE", "false")
 
   $che_keycloak_admin_require_update_password=getValue("CHE_KEYCLOAK_ADMIN_REQUIRE_UPDATE_PASSWORD", "true")
-  $che_keycloak_auth__server__url=getValue("CHE_KEYCLOAK_AUTH__SERVER__URL","")
 
   ###############################
   # Include base module
--- a/dockerfiles/init/manifests/che.pp
+++ b/dockerfiles/init/manifests/che.pp
@@ -40,8 +40,7 @@ node default {
   #
   $che_single_port = getValue("CHE_SINGLE_PORT","false")
   $che_single_port_wildcard_domain_host = getValue("CHE_SINGLEPORT_WILDCARD__DOMAIN_HOST","nip.io")
-  $che_single_port_custom_dns = getValue("CHE_SINGLEPORT_CUSTOM_DNS","false")
-
+  $che_single_port_wildcard_domain_ipless = getValue("CHE_SINGLEPORT_WILDCARD__DOMAIN_IPLESS","false")
 
   ###############################
   # Che multiuser
--- a/dockerfiles/init/modules/compose/templates/docker-compose.yml.erb
+++ b/dockerfiles/init/modules/compose/templates/docker-compose.yml.erb
@@ -98,7 +98,7 @@ services:
 <% end -%>
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
-      - '<%= scope.lookupvar('che::che_instance') -%>/config/traefik/traefik.toml:/etc/traefik/traefik.toml'
+      - '<%= scope.lookupvar('che::che_instance') -%>/config/traefik:/etc/traefik'
     restart: always
 <% end -%>
 
--- a/dockerfiles/init/modules/compose/templates/docker-compose.yml.erb
+++ b/dockerfiles/init/modules/compose/templates/docker-compose.yml.erb
@@ -162,7 +162,7 @@ services:
     labels:
       traefik.keycloak.frontend.entryPoints: "http"
       traefik.keycloak.port: "8080"
-<% if scope.lookupvar('che::che_single_port') == 'true' and scope.lookupvar('che::che_single_port_custom_dns') == 'true' -%>
+<% if scope.lookupvar('che::che_single_port') == 'true' and scope.lookupvar('che::che_single_port_wildcard_domain_ipless') == 'true' -%>
       traefik.keycloak.frontend.rule: "Host:keycloak.<%= scope.lookupvar('che::che_single_port_wildcard_domain_host') -%>"
 <% elsif ! @che_docker_ip_external.empty? -%>
       traefik.keycloak.frontend.rule: "Host:keycloak.<%= scope.lookupvar('che::che_docker_ip_external') -%>.<%= scope.lookupvar('che::che_single_port_wildcard_domain_host') -%>"
--- a/dockerfiles/init/modules/compose/templates/docker-compose.yml.erb
+++ b/dockerfiles/init/modules/compose/templates/docker-compose.yml.erb
@@ -162,7 +162,9 @@ services:
     labels:
       traefik.keycloak.frontend.entryPoints: "http"
       traefik.keycloak.port: "8080"
-<% if ! @che_docker_ip_external.empty? -%>
+<% if scope.lookupvar('che::che_single_port') == 'true' and scope.lookupvar('che::che_single_port_custom_dns') == 'true' -%>
+      traefik.keycloak.frontend.rule: "Host:keycloak.<%= scope.lookupvar('che::che_single_port_wildcard_domain_host') -%>"
+<% elsif ! @che_docker_ip_external.empty? -%>
       traefik.keycloak.frontend.rule: "Host:keycloak.<%= scope.lookupvar('che::che_docker_ip_external') -%>.<%= scope.lookupvar('che::che_single_port_wildcard_domain_host') -%>"
 <% else -%>
       traefik.keycloak.frontend.rule: "Host:keycloak.<%= scope.lookupvar('che::docker_ip') -%>.<%= scope.lookupvar('che::che_single_port_wildcard_domain_host') -%>"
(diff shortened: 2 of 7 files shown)


hkolvenbach:
@JamesDrummond this might be interesting for you. Did you solve the HTTPS setup already?

benoitf:
in che 5.x we had a template so users could specify through a property the behavior on how subdomains were concatenated (like if some ppl wanted `-`or `_`)
Adding a property is fine for me but I would give a greater choice of possibility to che6 as for che5
@mshaposhnik

hkolvenbach:
@benoitf like the idea, this would be even easier to implement. by the way: is there a good reason to use dots as default or should we just switch to `_` or `-`?

benoitf:
@hkolvenbach I would say that default could be `-`or `_` it was just easy to use `.`as separators

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/provisioner/labels/SinglePortLabelsProvisioner.java
```diff
@@ -94,10 +95,11 @@ public void provision(DockerEnvironment internalEnv, RuntimeIdentity identity)
...
-        (externalIpOfContainers != null && host.contains(externalIpOfContainers))
-            ? host.indexOf(externalIpOfContainers)
-            : host.indexOf(internalIpOfContainers);
-    return host.substring(0, idx - 1).replaceAll("\\.", "-");
+    // int idx =
+    //     (externalIpOfContainers != null && host.contains(externalIpOfContainers))
+    //         ? host.indexOf(externalIpOfContainers)
+    //         : host.indexOf(internalIpOfContainers);
+    // return host.substring(0, idx - 1).replaceAll("\\.", "-");
+    return host.replaceAll("\\.", "-");
```
mshaposhnik: We don't want hostnames\IP's in the labels. So need to cut URL either until any IP or first dot in case of custom DNS.

dockerfiles/init/manifests/che.pp
```diff
@@ -40,6 +40,8 @@
   #
   $che_single_port = getValue("CHE_SINGLE_PORT","false")
   $che_single_port_wildcard_domain_host = getValue("CHE_SINGLEPORT_WILDCARD__DOMAIN_HOST","nip.io")
+  $che_single_port_custom_dns = getValue("CHE_SINGLEPORT_CUSTOM_DNS","false")
```
mshaposhnik: Property name doesnt looks very self-explained to me.. They're all is a DNS. The only difference is it requires explicit IP or not. So i propose something like `CHE_SINGLEPORT_WILDCARD__DOMAIN_IPLESS`

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/provisioner/labels/SinglePortLabelsProvisioner.java
```diff
@@ -94,10 +95,11 @@ public void provision(DockerEnvironment internalEnv, RuntimeIdentity identity)
    * exec-agent-http-dev-machine-workspaceao6k83hkdav975g5
```
mshaposhnik, line 94: Now it will be `exec-agent-http_dev-machine_workspaceao6k83hkdav975g5`

dockerfiles/init/manifests/che.pp
hkolvenbach: good suggestion!

garagatyi (commented):
Changing wildcard approach to something more widespread than multi-level wildcard certificates seems a good idea to me. Including default behavior. 

On the other hand, adding more properties and options to it will lead, eventually, to a rule-based approach that we had in Che5. So I would opt for implementing something similar to it in Che 6. Since our code has significantly changed in Che 6 it is not easy to move that code from Che 5 and I even think it is better to reimplement it from scratch to avoid legacy baggage from the previous architecture.

hkolvenbach:
@garagatyi I was thinking about porting https://github.com/codenvy/codenvy/blob/61c0113b00eaea22cd20381be6ecbe9dd0da56e0/dockerfiles/init/modules/codenvy/templates/machine.properties.erb#L67 and https://github.com/codenvy/codenvy/blob/fa4422ec95acc09f9daad591fd2f6edc08614028/plugins/plugin-hosted/codenvy-machine-hosted/src/main/java/com/codenvy/machine/UriTemplateServerProxyTransformer.java to Che 6, to be able to rewrite URLs to something like workspace_wsagent_xyz.domain.tld/node_port. Do you think this is a good idea?

Regarding the docker infrastructure, it would probably be better to do this through labeling the exposed ports for traefik, but I couldn't find a starting point in the source code for that. Do you know how exposed ports for wsagent etc are configured?

garagatyi:
@benoitf You have previously created another approach of URLs rewriting. WDYT about the suggested idea?

@mshaposhnik could you help with labeling ports in traefik here?

benoitf:
@garagatyi yes I think than `//%5$s/%3$s_%2$s/%4$s` is hard to make it evolve like introducing new  arguments while with ${server-reference} etc, it's easier to understand what are the new variables without breaking any existing users settings.
So I would prefer using specific variable pattern instead of 1, 2, 3

garagatyi:
@hkolvenbach here are docs about previous (Che 5 specific) configuration of URLs templates https://www.eclipse.org/che/docs/setup/configuration/index.html#workspace-address-resolution-strategy.
An example of configuration is: `che.docker.server_evaluation_strategy.custom.template=<serverName>.<machineName>.<workspaceId>.<wildcardNipDomain>:<chePort>`
The code that implemented it is outdated since we reworked that area completely. But you can implement something like it. 
You could ping us if you have any questions about it. BTW @benoitf was the author of that code

dockerfiles/init/modules/compose/templates/docker-compose.yml.erb
```diff
@@ -98,7 +98,7 @@ services:
 <% end -%>
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
-      - '<%= scope.lookupvar('che::che_instance') -%>/config/traefik/traefik.toml:/etc/traefik/traefik.toml'
+      - '<%= scope.lookupvar('che::che_instance') -%>/config/traefik:/etc/traefik'
```
hkolvenbach, line 101: this is needed to place https certifactes in the `config/traefik` folder. not strictly necessary for single-level-subdomains

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/provisioner/labels/SinglePortLabelsProvisioner.java
hkolvenbach: fixed

dockerfiles/init/manifests/che.pp
hkolvenbach: @benoitf i tried to do it like you described, but i am not sure how to elegantly get the configuration into the `SinglePortHostnameBuilder` (without passing it as an argument to the  `SinglePortHostnameBuilder` each time), so I opted for the easier method.

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/provisioner/labels/SinglePortLabelsProvisioner.java
hkolvenbach: fixed

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/provisioner/labels/SinglePortLabelsProvisioner.java
```diff
@@ -91,13 +91,18 @@ public void provision(DockerEnvironment internalEnv, RuntimeIdentity identity)
...
+   * exec-agent-http_dev-machine_workspaceao6k83hkdav975g5
    */
   private String getServiceName(String host) {
     int idx =
         (externalIpOfContainers != null && host.contains(externalIpOfContainers))
             ? host.indexOf(externalIpOfContainers)
             : host.indexOf(internalIpOfContainers);
-    return host.substring(0, idx - 1).replaceAll("\\.", "-");
+    if (idx > 1) {
+      return host.substring(0, idx - 1).replaceAll("\\.", "-");
```
mshaposhnik: Please, review logic here again. We dont need to replace dots anymore. Only cut until first dot and throw out any optional IP  and wildcard domain.

mshaposhnik:
Other is ok for me.

skabashnyuk:
pr for docs?

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/provisioner/labels/SinglePortLabelsProvisioner.java
hkolvenbach: good point!

hkolvenbach:
created the docs pr here: https://github.com/eclipse/che-docs/pull/371

mshaposhnik (approved):

skabashnyuk (approved):

dockerfiles/init/manifests/che.pp
```diff
@@ -66,6 +67,7 @@
   $system_super_privileged_mode=getValue("SYSTEM_SUPER__PRIVILEGED__MODE", "false")
 
   $che_keycloak_admin_require_update_password=getValue("CHE_KEYCLOAK_ADMIN_REQUIRE_UPDATE_PASSWORD", "true")
+  $che_keycloak_auth__server__url=getValue("CHE_KEYCLOAK_AUTH__SERVER__URL","")
```
riuvshin: I do not see where $che_keycloak_auth__server__url is used, do we need that?

hkolvenbach: yes, i think this is a leftover

hkolvenbach:
i needed to change the joiner from `_` to `-` because it would generate invalid hostnames otherwise (this becomes a problem if you try to generate a certificate from letsencrypt for it, browsers seem to not care about `_` in hostnames

hkolvenbach:
new docs pr: https://github.com/eclipse/che-docs/pull/373/files

garagatyi (approved):

riuvshin (approved):

garagatyi:
@hkolvenbach is there something that needs to be added to this PR?
@benoitf Are you OK with the PR?

I would want to proceed with the PR to avoid possible issues because of changes from other PRs.
@hkolvenbach can you it one more time?

benoitf (approved):

hkolvenbach:
@garagatyi from my perspective, the PR is ready to be merged. or is there anything else i should do?

garagatyi:
@hkolvenbach I've merged PR. Thank you for contributing this functionality!

PR #8998, opened 2018-03-03
Allow to configure logback logger with env variables

### LINKED ISSUE #8997: Allow to configure logback logger with env variables
### Description

Now in most cases, we have two regimes of logger INFO or DEBUG.
But during development, especially when Che is working inside of containers,
when it's not easy to reconfigure logback with xml configuration or JMX,
it might be useful to be able to configure particular logger. For example like 

```
CHE_LOGGER_CONFIG=org.eclipse.che=DEBUG,org.eclipse.che.api.installer.server.impl.LocalInstallerRegistry=OFF 
```
To be able to configure ws-master  - a user will have to add this variable to che.env
In case of ws-agent - a user will have to add this variable to env configuration of the desired machine.

skabashnyuk:
### What does this PR do?
Allow to configure logback logger with env variables
Example.
```
CHE_LOGGER_CONFIG=org.eclipse.che=DEBUG,org.eclipse.che.api.installer.server.impl.LocalInstallerRegistry=OFF 
```
To be able to configure ws-master  - a user will have to add this variable to che.env

In case of ws-agent - a user will have to add this variable to env configuration of the desired machine.

Fixes https://github.com/eclipse/che/issues/8997
Depend on https://github.com/eclipse/che-lib/pull/79

#### Release Notes
CHE_LOGGER_CONFIG environment variable is used to configure logback logger
#### Docs PR
https://github.com/eclipse/che-docs/pull/369

### DIFF
--- a/assembly/assembly-ide-war/pom.xml
+++ b/assembly/assembly-ide-war/pom.xml
@@ -57,6 +57,16 @@
             <artifactId>che-core-commons-j2ee</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-commons-logback</artifactId>
+            <exclusions>
+                <exclusion>
+                    <artifactId>che-core-api-core</artifactId>
+                    <groupId>org.eclipse.che.core</groupId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
             <groupId>javax.servlet</groupId>
             <artifactId>javax.servlet-api</artifactId>
             <scope>provided</scope>
@@ -112,6 +122,7 @@
                                 <!-- dependency is required just to overlay it's content -->
                                 <dep>org.eclipse.che:che-ide-gwt-app</dep>
                                 <dep>org.eclipse.che.core:che-core-commons-j2ee</dep>
+                                <dep>org.eclipse.che.core:che-core-commons-logback</dep>
                                 <dep>ch.qos.logback:logback-classic</dep>
                             </ignoredUnusedDeclaredDependencies>
                         </configuration>
--- a/core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
+++ b/core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
@@ -10,8 +10,6 @@
  */
 package org.eclipse.che.commons.logback;
 
-import static com.google.common.base.Strings.isNullOrEmpty;
-
 import ch.qos.logback.classic.Level;
 import ch.qos.logback.classic.Logger;
 import ch.qos.logback.classic.LoggerContext;
@@ -52,7 +50,7 @@ public class EnvironmentVariablesLogLevelPropagator extends ContextAwareBase
     String loggerName = parts[0];
     String levelStr = parts[1];
 
-    if (isNullOrEmpty(loggerName) || isNullOrEmpty(levelStr)) {
+    if (levelStr.isEmpty() || loggerName.isEmpty()) {
       return;
     }
 
--- a/core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
+++ b/core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
@@ -71,10 +71,12 @@ public class EnvironmentVariablesLogLevelPropagator extends ContextAwareBase
     }
   }
 
+  @Override
   public void stop() {
     isStarted = false;
   }
 
+  @Override
   public boolean isStarted() {
     return isStarted;
   }
--- a/core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
+++ b/core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
@@ -42,19 +42,18 @@ public class EnvironmentVariablesLogLevelPropagator extends ContextAwareBase
   }
 
   private void setLoggerLevel(String loggerConfig) {
-    String[] parts = loggerConfig.split("=");
+    String[] parts = loggerConfig.split("=", 2);
 
     if (parts.length < 2) {
       return;
     }
     String loggerName = parts[0];
     String levelStr = parts[1];
-    if (loggerName == null) {
+
+    if (loggerName == null || levelStr == null || levelStr.isEmpty() || loggerName.isEmpty()) {
       return;
     }
-    if (levelStr == null) {
-      return;
-    }
+
     loggerName = loggerName.trim();
     levelStr = levelStr.trim();
 
--- a/core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
+++ b/core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright (c) 2012-2018 Red Hat, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.commons.logback;
+
+import ch.qos.logback.classic.Level;
+import ch.qos.logback.classic.Logger;
+import ch.qos.logback.classic.LoggerContext;
+import ch.qos.logback.classic.spi.LoggerContextListener;
+import ch.qos.logback.core.spi.ContextAwareBase;
+import ch.qos.logback.core.spi.LifeCycle;
+import java.util.Arrays;
+
+public class EnvironmentVariablesLogLevelPropagator extends ContextAwareBase
+    implements LoggerContextListener, LifeCycle {
+
+  private boolean isStarted;
+
+  private void setLoggerLevel(String loggerConfig) {
+    int i = loggerConfig.indexOf('=');
+    if (i < 0) {
+      return;
+    }
+    String loggerName = loggerConfig.substring(0, i);
+    String levelStr = loggerConfig.substring(i + 1);
+    if (loggerName == null) {
+      return;
+    }
+    if (levelStr == null) {
+      return;
+    }
+    loggerName = loggerName.trim();
+    levelStr = levelStr.trim();
+
+    addInfo("Trying to set level " + levelStr + " to logger " + loggerName);
+    LoggerContext lc = (LoggerContext) context;
+
+    Logger logger = lc.getLogger(loggerName);
+    if ("null".equalsIgnoreCase(levelStr)) {
+      logger.setLevel(null);
+    } else {
+      Level level = Level.toLevel(levelStr, null);
+      if (level != null) {
+        logger.setLevel(level);
+      }
+    }
+  }
+
+  @Override
+  public void start() {
+
+    String config = System.getenv("CHE_LOGGER_CONFIG");
+    if (config != null && !config.isEmpty()) {
+      Arrays.stream(config.split(",")).map(String::trim).forEach(this::setLoggerLevel);
+    }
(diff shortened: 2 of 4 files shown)


core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
```diff
@@ -0,0 +1,90 @@
...
+      }
+    }
+  }
+
+  @Override
+  public void start() {
+
+    String config = System.getenv("CHE_LOGGER_CONFIG");
+    if (config != null && !config.isEmpty()) {
+      Arrays.stream(config.split(",")).map(String::trim).forEach(this::setLoggerLevel);
```
sleshchenko: As far as I understand RuntimeException will be thrown in a case when `CHE_LOGGER_CONFIG` value has the wrong format. Am I right? If yes, maybe it would be better to catch it and log pretty message like '`CHE_LOGGER_CONFIG` env variable has the wrong format. It should be ...'

core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
```diff
@@ -0,0 +1,90 @@
...
+
+import ch.qos.logback.classic.Level;
+import ch.qos.logback.classic.Logger;
+import ch.qos.logback.classic.LoggerContext;
+import ch.qos.logback.classic.spi.LoggerContextListener;
+import ch.qos.logback.core.spi.ContextAwareBase;
+import ch.qos.logback.core.spi.LifeCycle;
+import java.util.Arrays;
+
+public class EnvironmentVariablesLogLevelPropagator extends ContextAwareBase
```
sleshchenko: Please add java doc

core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
```diff
@@ -0,0 +1,90 @@
...
+    implements LoggerContextListener, LifeCycle {
+
+  private boolean isStarted;
+
+  private void setLoggerLevel(String loggerConfig) {
```
sleshchenko: It would a bit better if a private method is placed under that one who invokes it

assembly/assembly-ide-war/pom.xml
```diff
@@ -56,6 +56,16 @@
             <groupId>org.eclipse.che.core</groupId>
             <artifactId>che-core-commons-j2ee</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-commons-logback</artifactId>
+            <exclusions>
+                <exclusion>
+                    <artifactId>che-core-api-core</artifactId>
```
sleshchenko, line 64: Why is it excluded?

skabashnyuk: to make sure ide war doesn't contain che-core-api-core and all it unnecessary dependencies.

core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
skabashnyuk: No, AFAIK it would be just silently ignored. In fact, the same way if you put not existed classes or packages in logback.xml

core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  private void setLoggerLevel(String loggerConfig) {
+    int i = loggerConfig.indexOf('=');
+    if (i < 0) {
+      return;
+    }
+    String loggerName = loggerConfig.substring(0, i);
+    String levelStr = loggerConfig.substring(i + 1);
+    if (loggerName == null) {
```
sleshchenko: `substring` method never returns null, these null checks look like redundant. After removing them you'll be able to inline `String loggerName = loggerConfig.substring(0, i).trim();`

core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
```diff
@@ -0,0 +1,98 @@
...
+ * format:
+ *
+ * <p>CHE_LOGGER_CONFIG=logger1_name=logger1_level,logger2_name=logger2_level
+ *
+ * <p>In case if some logger name or logger level are omitted this pair will be silently ignored.
+ */
+public class EnvironmentVariablesLogLevelPropagator extends ContextAwareBase
+    implements LoggerContextListener, LifeCycle {
+
+  private boolean isStarted;
```
garagatyi, line 32: Can you elaborate on how it is used and why it should not be thread safe?

core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
```diff
@@ -0,0 +1,98 @@
...
+
+    String config = System.getenv("CHE_LOGGER_CONFIG");
+    if (config != null && !config.isEmpty()) {
+      Arrays.stream(config.split(",")).map(String::trim).forEach(this::setLoggerLevel);
+    }
+    isStarted = true;
+  }
+
+  private void setLoggerLevel(String loggerConfig) {
+    int i = loggerConfig.indexOf('=');
```
garagatyi: I believe that call of split() would be clearer.

core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
```diff
@@ -0,0 +1,98 @@
...
+    if (loggerName == null) {
+      return;
+    }
+    if (levelStr == null) {
+      return;
+    }
+    loggerName = loggerName.trim();
+    levelStr = levelStr.trim();
+
+    addInfo("Trying to set level " + levelStr + " to logger " + loggerName);
```
sleshchenko: Maybe it makes sense to check levelStr and loggerName variable on empty here? Or at least wrap them into `'` to make it clear that they are empty.

core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
skabashnyuk: It is implemented in similar to ch.qos.logback.classic.jul.LevelChangePropagator way
I guess life cycle of this component is not multi-threaded.

assembly/assembly-ide-war/pom.xml
sleshchenko: Looks like a bit strange. It means that you're using classes from this dependency which don't use `api-core` classes.
But somebody can easily use it, and then Runtime exception will occur.
It's just thought. I'm OK to leave it as it is.

garagatyi: Why we need logback in IDE war at all?

core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
```diff
@@ -0,0 +1,98 @@
...
+      logger.setLevel(null);
+    } else {
+      Level level = Level.toLevel(levelStr, null);
+      if (level != null) {
+        logger.setLevel(level);
+      }
+    }
+  }
+
+  public void stop() {
```
garagatyi: `stop` and `isStarted` methods are not used. Can we just delete them?

assembly/assembly-ide-war/pom.xml
skabashnyuk: @sleshchenko about exclusion.
We already have che-core-commons-logback module. To avoid exclusion we need another module with only one single class. When I selecting this approach I thought it would be less confusing then to have two modules.

core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
skabashnyuk: They come from the interface. I'm going to restore Override annotations.

assembly/assembly-ide-war/pom.xml
skabashnyuk: >>Why we need logback in IDE war at all?

@garagatyi 
Some filters are using slf4j-> logback logging.

core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
```diff
@@ -0,0 +1,100 @@
...
+
+    String config = System.getenv("CHE_LOGGER_CONFIG");
+    if (config != null && !config.isEmpty()) {
+      Arrays.stream(config.split(",")).map(String::trim).forEach(this::setLoggerLevel);
+    }
+    isStarted = true;
+  }
+
+  private void setLoggerLevel(String loggerConfig) {
+    String[] parts = loggerConfig.split("=");
```
sleshchenko: I think it would be better to specify limit here

core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
```diff
@@ -0,0 +1,100 @@
...
+
+    if (parts.length < 2) {
+      return;
+    }
+    String loggerName = parts[0];
+    String levelStr = parts[1];
+    if (loggerName == null) {
```
sleshchenko: When is it possible that parts array contains null values?

assembly/assembly-ide-war/pom.xml
garagatyi: Well, it is more convenient right now, but we will have to remember that this exclusion is needed. And other projects that would extend Che also would need to know about this exclusion.

core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
garagatyi: Oh, then it would be clearer

garagatyi (approved):

assembly/assembly-ide-war/pom.xml
skabashnyuk: That would not be a problem soon. Since that is not a case for Theia.

core/commons/che-core-commons-logback/src/main/java/org/eclipse/che/commons/logback/EnvironmentVariablesLogLevelPropagator.java
```diff
@@ -0,0 +1,99 @@
...
+    isStarted = true;
+  }
+
+  private void setLoggerLevel(String loggerConfig) {
+    String[] parts = loggerConfig.split("=", 2);
+
+    if (parts.length < 2) {
+      return;
+    }
+    String loggerName = parts[0];
```
sleshchenko, line 50: it makes sense to trim values here

assembly/assembly-ide-war/pom.xml
garagatyi: I'm ok with this approach, just thinking out loud

sleshchenko (approved):
Thanks for contributing such feature.
Previously setting the log level to DEBUG just enabled chaos mode for logs. :smiley: 
When it is possible to set the log level for a particular component, debug logs become useful.

skabashnyuk:
ci-test

gazarenkov (approved):
Useful, make sure we add this info to documentation.

gazarenkov (approved):

vparfonov (approved):

PR #9047, opened 2018-03-07
Allow to use NULL in JDBC env vars

garagatyi:
### What does this PR do?
Allows to use NULL as a value of JDBC config to simplify usage of
environment variables for JDBC settings in k8s/Openshift
deployment files. NULL is treated as unset variable.


benoitf (approved):

skabashnyuk (approved):

riuvshin (approved):

garagatyi:
ci-build

Issue #7161, opened 2017-11-02
Unexpectedly absent the warning marker in the editor


artaleks9:
* Sometimes selenium tests from 'opendeclaration' package are failed by reason not present warning markers
* Another case of failing test: after perform the 'open declaration' (F4) the expected file is absent in the editor
* As a rule it happen when tests were running on CI, but on local host the tests are passed
* We need to fix up it
* Related selenium tests: _Eclipse0093Test_, _Eclipse0115Test_, _Eclipse0121Test_ , _AutocompleteProposalJavaDocTest_ , _AutocompleteWithInheritTest_ , _ShowHintsCommandTest_

artaleks9:
After merging the pull request #7164 the problem is not reproduced on CI

dmytro-ndp:
We have the same test failure again on CI https://ci.codenvycorp.com/view/qa/job/che-integration-tests-che6/113/testReport/junit/org.eclipse.che.selenium.opendeclaration/Eclipse0115Test/test0115/:
```
org.openqa.selenium.TimeoutException: Expected condition failed: waiting for visibility of element located by By.xpath: //div[@class='ruler overview']/div[@class='annotationOverview warning'] (tried for 20 second(s) with 500 MILLISECONDS interval)
Build info: version: '3.0.1', revision: '1969d75', time: '2016-10-18 09:49:13 -0700'
System info: host: 'slave6.codenvycorp.com', ip: '127.0.0.1', os.name: 'Linux', os.arch: 'amd64', os.version: '3.10.0-514.16.1.el7.x86_64', java.version: '1.8.0_102'
Driver info: org.eclipse.che.selenium.core.SeleniumWebDriver
	at org.openqa.selenium.support.ui.WebDriverWait.timeoutException(WebDriverWait.java:80)
	at org.openqa.selenium.support.ui.FluentWait.until(FluentWait.java:265)
	at org.eclipse.che.selenium.pageobject.CodenvyEditor.waitMarkerInPosition(CodenvyEditor.java:627)
```
![org eclipse che selenium opendeclaration eclipse0115test test0115_qqldkvfe](https://user-images.githubusercontent.com/1197777/33932335-61a70794-dffb-11e7-9b22-44133d0ba3a2.png)

- https://ci.codenvycorp.com/view/pr-builds/job/che-pullrequests-test/279/testReport/org.eclipse.che.selenium.opendeclaration/Eclipse0115Test/test0115/

dmytro-ndp:
[...]

Issue #9058, opened 2018-03-12
Apply expand the project tree by native clicks in the 'Eclipse0120Test' 


artaleks9:
### Description
* Need to replace expanding the project by the method `quickExpandWithJavaScript()` on the native clicks to increase stability
* Related issue: #7161

Issue #9057, opened 2018-03-12
Delete the 'try/catch' from the tests in the 'opendeclaration' package


artaleks9:
### Description
* Need to delete the `try/catch` blocks for the selenium tests in the `opendeclaration` package related to resolve and close issue #7161
* Related selenium test: _Eclipse0093Test_, _Eclipse0115Test_, _Eclipse0121Test_

PR #9059, opened 2018-03-12
Selenium: Delete the 'try/catch' from selenium tests in the 'opendeclaration' package

artaleks9:
### What does this PR do?
* Delete the 'try/catch' blocks from tests related to resolve and close issue #7161
* Replace the 'quickExpandWithJavaScript()' method on the native clicks to expand the project tree in the 'Eclipse0120Test'
* Related selenium tests: _Eclipse0093Test_, _Eclipse0115Test_, _Eclipse0121Test_; _Eclipse0120Test_

#9057, #9058

### DIFF
--- a/selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/opendeclaration/Eclipse0120Test.java
+++ b/selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/opendeclaration/Eclipse0120Test.java
@@ -31,7 +31,7 @@ import org.testng.annotations.Test;
 public class Eclipse0120Test {
   private static final String PROJECT_NAME = NameGenerator.generate("Eclipse0120Test-", 4);
   private static final String PATH_TO_EXPAND =
-      String.format("%s/%s", PROJECT_NAME, "src/main/java/org.eclipse.qa.examples");
+      PROJECT_NAME + "/src/main/java/org.eclipse.qa.examples";
 
   @Inject private TestWorkspace ws;
   @Inject private Ide ide;
--- a/selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/opendeclaration/Eclipse0120Test.java
+++ b/selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/opendeclaration/Eclipse0120Test.java
@@ -29,9 +29,9 @@ import org.testng.annotations.Test;
 
 /** @author Aleksandr Shmaraev */
 public class Eclipse0120Test {
-  private static final String PATH_TO_PACKAGE_PREFIX = "/src/main/java/org/eclipse/qa/examples/";
-  private static final String PROJECT_NAME =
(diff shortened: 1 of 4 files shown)


SkorikSergey (approved):

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/opendeclaration/Eclipse0120Test.java
```diff
@@ -29,9 +29,9 @@
 
 /** @author Aleksandr Shmaraev */
 public class Eclipse0120Test {
-  private static final String PATH_TO_PACKAGE_PREFIX = "/src/main/java/org/eclipse/qa/examples/";
-  private static final String PROJECT_NAME =
-      NameGenerator.generate(Eclipse0120Test.class.getSimpleName(), 4);
+  private static final String PROJECT_NAME = NameGenerator.generate("Eclipse0120Test-", 4);
+  private static final String PATH_TO_EXPAND =
+      String.format("%s/%s", PROJECT_NAME, "src/main/java/org.eclipse.qa.examples");
```
Ohrimenko1988: In my opinion, concatenation is better for this case.
```
private static final String PATH_TO_EXPAND = PROJECT_NAME + "/src/main/java/org.eclipse.qa.examples";
```

artaleks9: Why do you think so?

Ohrimenko1988: It looks cleaner and simpler.

artaleks9: @dmytro-ndp 
As I see, you agree with the opinion of the @Ohrimenko1988.
Couldn't  you explain, why the concatenation is better in this case?

dmytro-ndp: I like you use **String.format()** to make code more readable. And it works well in case of more than one substitutions at the middle of line, or long strings, or the complex format of substitute.
But in this case, where we have one substitution at the start of line, concatenation looks really clearer, IMHO.

dmytro-ndp (approved):

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/opendeclaration/Eclipse0120Test.java
artaleks9: Thank you for your explanation

artaleks9:
ci-test

Ohrimenko1988 (approved):

PR #9073, opened 2018-03-13
Unable to open Theia IDE

vitaliy-guliy:
### What does this PR do?

Adds Theia server to servers checkers.

Prevents hanging the browser when workspace loader displays huge amount of outputs.
Browser become working slower and slower, so setting limitation in 500 lines resolves the problem.

https://github.com/eclipse/che/issues/9079


garagatyi:
I'm ok with the adding theia as yet another hardcoded server that defines whether workspace is started or not. But would want to hear opinions from @skabashnyuk @gazarenkov

garagatyi (approved):

ashumilova (approved):

Issue #9086, opened 2018-03-14
Unexpected fail of RenameWorkspaceTest selenium test


SkorikSergey:
### Description
We have unexpected fail of **RenameWorkspaceTest** selenium test. The test fails while checking that workspace name is valid after renaming.  We should clarify reason and fix it.

**Log:**
```
java.lang.AssertionError: expected [false] but found [true]
	at org.eclipse.che.selenium.dashboard.RenameWorkspaceTest.renameWorkspace(RenameWorkspaceTest.java:97)
	at org.eclipse.che.selenium.dashboard.RenameWorkspaceTest.renameNameWorkspaceTest(RenameWorkspaceTest.java:89)
```

**Report with reproduced problem:**
https://ci.codenvycorp.com/view/qa/job/che-integration-tests-multiuser-master-ocp/145/Selenium_tests_report/

SkorikSergey:
Done https://github.com/eclipse/che/pull/9087.

PR #9087, opened 2018-03-14
Selenium: add checking that a workspace name is valid in RenameWorkspaceTest

SkorikSergey:
### What does this PR do?
This PR adds method for checking that a workspace name is valid. It fix unstable **RenameWorkspaceTest** selenium test.

https://github.com/eclipse/che/issues/9086


artaleks9 (approved):

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/dashboard/workspaces/WorkspaceOverview.java
```diff
@@ -88,6 +90,11 @@ public Boolean isWorkspaceNameErrorMessageEquals(String message) {
     return errorMessages.getText().equals(message);
   }
 
+  public void waitWorkspaceNameIsValid() {
```
dmytro-ndp: Method name is not clear enough, IMHO.
May we name it as: _checkWorkspaceNameErrorAbsence()_?

dmytro-ndp (approved):

Issue #9100, opened 2018-03-15
Unexpected failure of selenium tests which are create workspaces from Dashboard


SkorikSergey:
### Description
We have unexpected fail of selenium tests that create workspaces from **Dashboard**. The tests, sometimes, fail while waiting that **Project Explorer** visibility. We should clarify reason and fix it.

**Log:**
```
org.openqa.selenium.TimeoutException:  Expected condition failed: waiting for visibility of element located by By.id: gwt-debug-projectTree (tried for 240 second(s) with 500 MILLISECONDS interval) Build info: version: '3.0.1', revision: '1969d75', time: '2016-10-18 09:49:13 -0700' System info: host: 'slave6.codenvycorp.com', ip: '127.0.0.1', os.name: 'Linux', os.arch: 'amd64', os.version: '3.10.0-514.16.1.el7.x86_64', java.version: '1.8.0_102' Driver info: org.eclipse.che.selenium.core.SeleniumWebDriver 	at org.eclipse.che.selenium.dashboard.ImportProjectFromZipTest.importProjectFromZipTest(ImportProjectFromZipTest.java:83) Caused by: org.openqa.selenium.NoSuchElementException:  no such element: Unable to locate element: {"method":"id","selector":"gwt-debug-projectTree"}
```

**Report with reproduced problem:**
https://ci.codenvycorp.com/view/qa/job/che-integration-tests-multiuser-master-ocp/146/Selenium_tests_report/

SkorikSergey:
Check is this issue will not reproduce after https://github.com/eclipse/che/pull/9117.

SkorikSergey:
Not reproduced after https://github.com/eclipse/che/pull/9262.

SkorikSergey:
Reproduced on CI https://ci.codenvycorp.com/view/qa/job/che-integration-tests-multiuser-master-ocp/177/Selenium_tests_report/.
**Failed test:** DirectUrlFactoryWithKeepDirectoryTest.

dmytro-ndp:
IMHO it's worth to mention in merge commit that https://github.com/eclipse/che/pull/9375 fixes selenium tests DirectUrlFactoryWithKeepDirectoryTest.

SkorikSergey:
Not reproduced.

PR #9117, opened 2018-03-16
Selenium: check that the IDE is loaded before switching to IDE application frame

SkorikSergey:
### What does this PR do?
This PR adds to **switchFromDashboardIframeToIde** method checking that the **IDE** is loaded before switching to IDE application frame.

https://github.com/eclipse/che/issues/9100

### DIFF
--- a/selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/SeleniumWebDriver.java
+++ b/selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/SeleniumWebDriver.java
@@ -323,9 +323,9 @@ public class SeleniumWebDriver
   }
 
   public void switchFromDashboardIframeToIde(int timeout) {
-    new WebDriverWait(this, timeout)
-        .until(visibilityOfElementLocated(By.id("ide-application-iframe")));
-    new WebDriverWait(this, ATTACHING_ELEM_TO_DOM_SEC)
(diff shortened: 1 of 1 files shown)


musienko-maxim (approved):

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/SeleniumWebDriver.java
```diff
@@ -317,13 +319,22 @@ public void switchToNoneCurrentWindow(String currentWindowHandler) {
...
   public void switchFromDashboardIframeToIde() {
-    new WebDriverWait(this, LOADER_TIMEOUT_SEC * 6)
-        .until(ExpectedConditions.frameToBeAvailableAndSwitchToIt(By.id("ide-application-iframe")));
+    switchFromDashboardIframeToIde(APPLICATION_START_TIMEOUT_SEC);
   }
 
   public void switchFromDashboardIframeToIde(int timeout) {
     new WebDriverWait(this, timeout)
-        .until(ExpectedConditions.frameToBeAvailableAndSwitchToIt(By.id("ide-application-iframe")));
+        .until(visibilityOfElementLocated(By.id("ide-application-iframe")));
```
dmytro-ndp: There is method `wait(timeout)` below in the class. You can use it to simplify the code.

Issue #9062, opened 2018-03-12
K8S / OpenShift infrastrucrures should support grace termination period of pods


ibuziuk:
By default grace shutdown period of pods on k8s / openshift defaults to 30 seconds. In Che 5 pods are created with `0` grace period which allows to kill a pod almost instantly and free the resources. This becomes vital for osio where it is only allowed to have single running workspace at a time. We need to support `0` grace termination period either via default property or hardcoded (not sure if this property should be configurable actually)

[1] https://github.com/eclipse/che/blob/5.22.x/plugins/plugin-docker/che-plugin-openshift-client/src/main/java/org/eclipse/che/plugin/openshift/client/OpenShiftConnector.java#L197

ibuziuk:
@garagatyi @l0rd WDYT ?

garagatyi:
I opt for a property that configures or enables/disables zero time shutdown.
@sleshchenko WDYT?

l0rd:
@ibuziuk it makes sense to set grace period to zero but having a property to disable it is even better

ibuziuk:
sounds good, so looks like we need a property for setting workspace pod termination period which would defaults to 0

garagatyi:
@ibuziuk do we have a tracking issue in rh-che repo? What priority should be set on this issue?

ibuziuk:
There is no issue in rh-che for that, in terms of priority IMO it's P1

ibuziuk:
PR is sent https://github.com/eclipse/che/pull/9118

ibuziuk:
PR for doc

ibuziuk:
PR & PR for docs have been merged. Closing

PR #9118, opened 2018-03-16
che #9062: Adding property for setting workspace pod termination grace period for Kubernetes / OpenShift infrastructures

ibuziuk:
### What does this PR do?
che #9062: Adding property for setting workspace pod termination grace period for Kubernetes / OpenShift infrastructures

#9062

#### Release Notes
Adding property for setting workspace pod termination grace period for Kubernetes / OpenShift infrastructures

#### Docs PR
https://github.com/eclipse/che-docs/pull/376

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
@@ -416,6 +416,13 @@ che.infra.kubernetes.ingress.annotations_json=NULL
 che.infra.kubernetes.pod.security_context.run_as_user=NULL
 che.infra.kubernetes.pod.security_context.fs_group=NULL
 
+# Defines grace termination period for pods that will be created by Kubernetes / OpenShift infrastructures
+#
+# Grace termination period of Kubernetes / OpenShift workspace's pods defaults '0', which allows to terminate
+# pods almost instantly and significantly decrease the time required for stopping a workspace.
+# Note: if `terminationGracePeriodSeconds` have been explicitly set in Kubernetes / OpenShift recipe it will not be overridden.
+che.infra.kubernetes.pod.termination_grace_period_sec=0
+
 # Number of maximum concurrent async web requests
 # (http requests or ongoing  web socket calls)
 # supported in the underlying shared http client
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesEnvironmentProvisioner.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesEnvironmentProvisioner.java
@@ -19,6 +19,7 @@ import org.eclipse.che.workspace.infrastructure.kubernetes.environment.Kubernete
 import org.eclipse.che.workspace.infrastructure.kubernetes.namespace.pvc.WorkspaceVolumesStrategy;
 import org.eclipse.che.workspace.infrastructure.kubernetes.provision.InstallerServersPortProvisioner;
 import org.eclipse.che.workspace.infrastructure.kubernetes.provision.LogsVolumeMachineProvisioner;
+import org.eclipse.che.workspace.infrastructure.kubernetes.provision.PodTerminationGracePeriodProvisioner;
 import org.eclipse.che.workspace.infrastructure.kubernetes.provision.SecurityContextProvisioner;
 import org.eclipse.che.workspace.infrastructure.kubernetes.provision.UniqueNamesProvisioner;
 import org.eclipse.che.workspace.infrastructure.kubernetes.provision.env.EnvVarsConverter;
@@ -46,6 +47,7 @@ public class KubernetesEnvironmentProvisioner {
   private final InstallerServersPortProvisioner installerServersPortProvisioner;
   private final LogsVolumeMachineProvisioner logsVolumeMachineProvisioner;
   private final SecurityContextProvisioner securityContextProvisioner;
+  private final PodTerminationGracePeriodProvisioner podTerminationGracePeriodProvisioner;
 
   @Inject
   public KubernetesEnvironmentProvisioner(
@@ -58,7 +60,8 @@ public class KubernetesEnvironmentProvisioner {
       RamLimitProvisioner ramLimitProvisioner,
       InstallerServersPortProvisioner installerServersPortProvisioner,
       LogsVolumeMachineProvisioner logsVolumeMachineProvisioner,
-      SecurityContextProvisioner securityContextProvisioner) {
+      SecurityContextProvisioner securityContextProvisioner,
+      PodTerminationGracePeriodProvisioner podTerminationGracePeriodProvisioner) {
     this.pvcEnabled = pvcEnabled;
     this.volumesStrategy = volumesStrategy;
     this.uniqueNamesProvisioner = uniqueNamesProvisioner;
@@ -69,6 +72,7 @@ public class KubernetesEnvironmentProvisioner {
     this.installerServersPortProvisioner = installerServersPortProvisioner;
     this.logsVolumeMachineProvisioner = logsVolumeMachineProvisioner;
     this.securityContextProvisioner = securityContextProvisioner;
+    this.podTerminationGracePeriodProvisioner = podTerminationGracePeriodProvisioner;
   }
 
   public void provision(KubernetesEnvironment k8sEnv, RuntimeIdentity identity)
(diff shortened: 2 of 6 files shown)


ibuziuk:
ci-test

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesEnvironmentProvisioner.java
```diff
@@ -91,5 +95,8 @@ public void provision(KubernetesEnvironment k8sEnv, RuntimeIdentity identity)
     uniqueNamesProvisioner.provision(k8sEnv, identity);
     ramLimitProvisioner.provision(k8sEnv, identity);
     securityContextProvisioner.provision(k8sEnv, identity);
+
+    // 4 stage - set Kubernetes pod termination grace period
+    podTerminationGracePeriodProvisioner.provision(k8sEnv, identity);
```
sleshchenko: I'd say that it should be in 3 stage like restartPolicyRewriter and securityContextProvisioner.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/provision/PodTerminationGracePeriodProvisioner.java
```diff
@@ -0,0 +1,48 @@
...
+      @Named("che.infra.kubernetes.pod.termination_grace_period_sec")
+          long graceTerminationPeriodSec) {
+    this.graceTerminationPeriodSec = graceTerminationPeriodSec;
+  }
+
+  @Override
+  public void provision(KubernetesEnvironment k8sEnv, RuntimeIdentity identity)
+      throws InfrastructureException {
+    for (Pod pod : k8sEnv.getPods().values()) {
+      if (!isTerminationGracePeriodSet(pod)) {
```
sleshchenko, line 42: Maybe it would be better to override user-defined grace period. And do not override only if the grace period is configured to `0`.
If not then it makes sense to document it and maybe rename configuration property like `default_termination_grace_period_sec`.

I vote for overriding user-defined grace period.
@garagatyi @ibuziuk WDYT?

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
```diff
@@ -416,6 +416,13 @@ che.infra.kubernetes.ingress.annotations_json=NULL
 che.infra.kubernetes.pod.security_context.run_as_user=NULL
 che.infra.kubernetes.pod.security_context.fs_group=NULL
 
+# Defines grace termination period for pods that will be created by Kubernetes / OpenShift infrastructures
+#
+# Grace termination period of Kubernetes / OpenShift pods defaults to 30 seconds, which results in
```
sleshchenko: Doc says `defaults to 30 seconds` but default value is
```
che.infra.kubernetes.pod.termination_grace_period_sec=0
```
Please check and sync these values

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesEnvironmentProvisioner.java
garagatyi: +1

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
garagatyi: +1

skabashnyuk:
@eivantsov  Should we add some changes in docs? WDYT?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/provision/PodTerminationGracePeriodProvisioner.java
```diff
@@ -0,0 +1,48 @@
...
+package org.eclipse.che.workspace.infrastructure.kubernetes.provision;
+
+import io.fabric8.kubernetes.api.model.Pod;
+import javax.inject.Inject;
+import javax.inject.Named;
+import org.eclipse.che.api.core.model.workspace.runtime.RuntimeIdentity;
+import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
+import org.eclipse.che.workspace.infrastructure.kubernetes.environment.KubernetesEnvironment;
+
+public class PodTerminationGracePeriodProvisioner implements ConfigurationProvisioner {
```
garagatyi: Please, add javadocs

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/provision/PodTerminationGracePeriodProvisioner.java
garagatyi: Not sure we should override the value specified by a user. What for? I opt for overriding only values that are not specified. About renaming I think current name is OK, and name with prefix default would be OK too.
In case we need to set it to 0, in any case, we will be able to add it as a separate flag later.

garagatyi (approved):

eivantsov:
Yes, a few lines here https://github.com/eclipse/che-docs/blob/master/src/main/pages/setup-openshift/openshift-config.md won't hurt

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/provision/PodTerminationGracePeriodProvisioner.java
sleshchenko: > What for?

To be able to guarantee that workspace will be stopped(or start will be interrupted) in predictable time.

garagatyi: And what if user's app needs more time, e.g. 1 minute?

sleshchenko: The situation is similar to the following: a user's container needs more time to start than workspace start timeout is configured.
Since removing of pods is the last phase of workspace stop, and we remove all pods asynchronously, I'm OK not to override a value defined by a user. But let's document it in che.properties.

garagatyi: +1 to document it. Let's start without rewriting user's timeouts and then we will figure out whether we need to change it

akorneta (approved):

davidfestal (approved):

l0rd (approved):

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesEnvironmentProvisioner.java
ibuziuk: thanks for the pointer will fix

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
ibuziuk: well, default to `0` is done deliberately and I think we should stick to it  (which would significantly decrease time required for workspace stop on k8s / openshift). Probably I should just remove info about default k8s config, which you find misleading

ibuziuk:
@eivantsov ok, thanks for the pointer - will add some docs

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/provision/PodTerminationGracePeriodProvisioner.java
ibuziuk: ok, will do

ibuziuk:
@eivantsov @skabashnyuk @sleshchenko PR has been updated. PR for docs https://github.com/eclipse/che-docs/pull/376

sleshchenko (approved):
LGTM :+1:

PR #9161, opened 2018-03-21
Introduce logger REST API to manage log level at runtime

benoitf:
### What does this PR do?

following this previous comment https://github.com/eclipse/che/issues/8997#issuecomment-370127156

with this Rest API, it's possible to create new logger with log level, list all loggers defined with loglevel, change level on a given logger, etc

On multi-user, `System#MANAGE_SYSTEM` permission is required.

Change-Id: I1c105aca33cc88f90270ade4d792d3a75191740a

https://github.com/eclipse/che/issues/8997

### DIFF
--- a/wsmaster/che-core-api-logger-shared/pom.xml
+++ b/wsmaster/che-core-api-logger-shared/pom.xml
@@ -1,0 +1,115 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright (c) 2012-2018 Red Hat, Inc.
+    All rights reserved. This program and the accompanying materials
+    are made available under the terms of the Eclipse Public License v1.0
+    which accompanies this distribution, and is available at
+    http://www.eclipse.org/legal/epl-v10.html
+
+    Contributors:
+      Red Hat, Inc. - initial API and implementation
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <artifactId>che-master-parent</artifactId>
+        <groupId>org.eclipse.che.core</groupId>
+        <version>6.3.0-SNAPSHOT</version>
+    </parent>
+    <artifactId>che-core-api-logger-shared</artifactId>
+    <packaging>jar</packaging>
+    <name>Che Core :: API :: Logger :: Shared</name>
+    <properties>
+        <dto-generator-out-directory>${project.build.directory}/generated-sources/dto/</dto-generator-out-directory>
+    </properties>
+    <dependencies>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-dto</artifactId>
+        </dependency>
+    </dependencies>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.eclipse.che.core</groupId>
+                <artifactId>che-core-api-dto-maven-plugin</artifactId>
+                <version>${project.version}</version>
+                <executions>
+                    <execution>
+                        <id>generate-server-dto</id>
+                        <phase>process-sources</phase>
+                        <goals>
+                            <goal>generate</goal>
+                        </goals>
+                        <configuration>
+                            <dtoPackages>
+                                <package>org.eclipse.che.api.logger.shared.dto</package>
+                            </dtoPackages>
+                            <outputDirectory>${dto-generator-out-directory}</outputDirectory>
+                            <genClassName>org.eclipse.che.api.logger.server.dto.DtoServerImpls</genClassName>
+                            <impl>server</impl>
+                        </configuration>
+                    </execution>
+                </executions>
+                <dependencies>
+                    <dependency>
+                        <groupId>org.eclipse.che.core</groupId>
+                        <artifactId>che-core-api-logger-shared</artifactId>
+                        <version>${project.version}</version>
+                    </dependency>
+                </dependencies>
+            </plugin>
+            <plugin>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>pre-compile</id>
+                        <phase>generate-sources</phase>
+                        <goals>
+                            <goal>compile</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>build-helper-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>add-resource</id>
+                        <phase>process-sources</phase>
+                        <goals>
+                            <goal>add-resource</goal>
+                        </goals>
+                        <configuration>
+                            <resources>
+                                <resource>
+                                    <directory>${dto-generator-out-directory}/META-INF</directory>
+                                    <targetPath>META-INF</targetPath>
+                                </resource>
+                            </resources>
+                        </configuration>
+                    </execution>
+                    <execution>
+                        <id>add-source</id>
+                        <phase>process-sources</phase>
+                        <goals>
+                            <goal>add-source</goal>
+                        </goals>
+                        <configuration>
+                            <sources>
+                                <source>${dto-generator-out-directory}</source>
+                            </sources>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+</project>
--- a/wsmaster/che-core-api-logger/pom.xml
+++ b/wsmaster/che-core-api-logger/pom.xml
@@ -54,34 +54,9 @@
             <artifactId>slf4j-api</artifactId>
         </dependency>
         <dependency>
-            <groupId>com.google.inject.extensions</groupId>
-            <artifactId>guice-persist</artifactId>
-            <scope>provided</scope>
-        </dependency>
-        <dependency>
-            <groupId>javax.servlet</groupId>
-            <artifactId>javax.servlet-api</artifactId>
-            <scope>provided</scope>
-        </dependency>
-        <dependency>
-            <groupId>javax.websocket</groupId>
-            <artifactId>javax.websocket-api</artifactId>
-            <scope>provided</scope>
-        </dependency>
-        <dependency>
             <groupId>javax.ws.rs</groupId>
             <artifactId>javax.ws.rs-api</artifactId>
             <scope>provided</scope>
-        </dependency>
-        <dependency>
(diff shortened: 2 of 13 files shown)


wsmaster/che-core-api-logger/pom.xml
```diff
@@ -0,0 +1,150 @@
...
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>javax.ws.rs</groupId>
+            <artifactId>javax.ws.rs-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-db</artifactId>
```
skabashnyuk: why it's provided?

wsmaster/che-core-api-logger/pom.xml
```diff
@@ -0,0 +1,150 @@
...
+            <artifactId>everrest-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.google.inject.extensions</groupId>
+            <artifactId>guice-persist</artifactId>
+            <scope>provided</scope>
```
skabashnyuk: why its provided?

wsmaster/che-core-api-logger/src/main/java/org/eclipse/che/api/logger/LoggerService.java
```diff
@@ -0,0 +1,136 @@
...
+import org.eclipse.che.api.logger.shared.dto.LoggerDto;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Defines Logger REST API. It allows to manage the loggers (with log level) dynamically.
+ *
+ * @author Florent Benoit
+ */
+@Api(value = "/logger", description = "Logger REST API")
+@Path("/logger")
```
skabashnyuk, line 46: can you make sure this service is protected with authentification in multi-user mode?

wsmaster/che-core-api-logger/pom.xml
```diff
@@ -0,0 +1,150 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-commons-test</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-db-vendor-h2</artifactId>
```
sleshchenko: Please revise dependencies list, I guess you don't need che-core-db-vendor-h2, org.eclipse.persistence.jpa dependencies, do you?

wsmaster/che-core-api-logger/src/main/java/org/eclipse/che/api/logger/LoggerService.java
```diff
@@ -0,0 +1,136 @@
...
+  @ApiOperation(
+    value = "Get loggers which are configured",
+    notes = "This operation can be performed only by authorized user",
+    response = LoggerDto.class,
+    responseContainer = "List"
+  )
+  @ApiResponses({
+    @ApiResponse(code = 200, message = "The loggers successfully fetched"),
+  })
+  public List<LoggerDto> getLoggers() {
```
skabashnyuk: should we add pagination here? with limit 30 by default

wsmaster/che-core-api-logger/src/main/java/org/eclipse/che/api/logger/LoggerService.java
```diff
@@ -0,0 +1,136 @@
...
+
+  @POST
+  @Path("/{name}")
+  @Consumes(APPLICATION_JSON)
+  @Produces(APPLICATION_JSON)
+  @ApiOperation(value = "Create a new logger level")
+  @ApiResponses({
+    @ApiResponse(code = 200, message = "The logger successfully created"),
+  })
+  public LoggerDto createLogger(
```
sleshchenko, line 120: I think this method should be covered with `System#MANAGE_SYSTEM` permission actions. Or maybe even all methods of this service. WDYT?

sleshchenko (commented):
Cool and useful feature :+1: 
But It looks like authorization is missed here. I guess It's not OK to this API should is available for all authenticated users. Or I missed something?

wsmaster/che-core-api-logger/pom.xml
benoitf: copied from https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-workspace/pom.xml#L119

wsmaster/che-core-api-logger/pom.xml
benoitf: copied from https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-workspace/pom.xml#L139

wsmaster/che-core-api-logger/pom.xml
benoitf: yes I need to cleanup

benoitf:
I should have specified that for now there is now permission check on this service, mainly because I added only this service on the single user and not the multi user.

https://github.com/eclipse/che/blob/fc3fd1dc83ba7d7305e54c956353ebdae7b2aedb/assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterModule.java#L249-L269

But yes I wanted to include it as well for multi-user, just wanted to discuss on how you were seeing the permission that would be required to call it. It seems `System#MANAGE_SYSTEM` is a choice ?

wsmaster/che-core-api-logger/src/main/java/org/eclipse/che/api/logger/LoggerService.java
benoitf: We could yes but for now there is only 3 loggers
https://github.com/eclipse/che-lib/blob/master/che-tomcat8-slf4j-logback/src/assembly/conf/logback.xml#L47-L49

wsmaster/che-core-api-logger/pom.xml
benoitf: removed the dependency

wsmaster/che-core-api-logger/pom.xml
benoitf: removed the dependency

wsmaster/che-core-api-logger/src/main/java/org/eclipse/che/api/logger/LoggerService.java
benoitf: for now it was not enabled in multi user mode

wsmaster/che-core-api-logger/pom.xml
benoitf: done

wsmaster/che-core-api-logger/src/main/java/org/eclipse/che/api/logger/LoggerService.java
skabashnyuk: What is your plan for that?

skabashnyuk:
> It seems System#MANAGE_SYSTEM is a choice

I think yes.

wsmaster/che-core-api-logger/src/main/java/org/eclipse/che/api/logger/LoggerService.java
benoitf: permission mechanism added

wsmaster/che-core-api-logger/src/main/java/org/eclipse/che/api/logger/LoggerService.java
benoitf: permission mechanism added

benoitf:
ok I have added permission stuff and then enabled as well logger api for multi user mode

wsmaster/che-core-api-logger/src/main/java/org/eclipse/che/api/logger/LoggerService.java
benoitf: added pagination

wsmaster/che-core-api-logger-shared/pom.xml
```diff
@@ -0,0 +1,115 @@
...
+                        <goals>
+                            <goal>generate</goal>
+                        </goals>
+                        <configuration>
+                            <dtoPackages>
+                                <package>org.eclipse.che.api.logger.shared.dto</package>
+                            </dtoPackages>
+                            <outputDirectory>${dto-generator-out-directory}</outputDirectory>
+                            <genClassName>org.eclipse.che.api.logger.server.dto.DtoServerImpls</genClassName>
+                            <impl>server</impl>
```
sleshchenko, line 56: I'd say that it is better to generate DTO server implementations in server module. Because a shared module is supposed to contain only shared things which are compatible with dependent modules.
So GWT module will get classes that can't be used there.
WDYT? Does it make sense?

sleshchenko (approved):
LGTM

wsmaster/che-core-api-logger-shared/pom.xml
benoitf: I copied content from workspace-shared module so maybe if I'm doing it wrong there is an issue with this module as well ?

it's up to you, let me know what is the better approach.
maybe client or server should be additional classifier artifacts

main artifacts = interfaces
client classifier = for gwt
server classifier = for server

sleshchenko: I like the idea with classifiers. 

> Is an issue with this module as well 

I think yes.

I think now we can leave it as it is. And later rework it in all places.
@skabashnyuk WDYT?

skabashnyuk (approved):

wsmaster/che-core-api-logger-shared/pom.xml
```diff
@@ -0,0 +1,115 @@
...
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-dto</artifactId>
+        </dependency>
+    </dependencies>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.eclipse.che.core</groupId>
+                <artifactId>che-core-api-dto-maven-plugin</artifactId>
+                <version>${project.version}</version>
```
riuvshin, line 42: should there `${che.version}` ?

mshaposhnik (approved):

riuvshin (approved):

wsmaster/che-core-api-logger-shared/pom.xml
benoitf: @riuvshin 
it's the same than here
https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-user-shared/pom.xml#L54
or here https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-workspace-shared/pom.xml#L59

what should I do ?

riuvshin: ah, ok, in theory both are fine, but I thought we use che.version in such places so it seems it is ok

wsmaster/che-core-api-logger-shared/pom.xml
skabashnyuk: I think we can revise it later.

vparfonov (approved):

PR #9162, opened 2018-03-21
generate workspace dto impls in same package

tareksha:
Generate implementations of ``che-core-api-workspace-shared`` DTOs in same package for allow simpler use externally.

### DIFF
--- a/wsmaster/che-core-api-workspace-shared/pom.xml
+++ b/wsmaster/che-core-api-workspace-shared/pom.xml
@@ -77,7 +77,7 @@
                         <package>org.eclipse.che.api.workspace.shared.dto</package>
                     </dtoPackages>
                     <outputDirectory>${dto-generator-out-directory}</outputDirectory>
-                    <genClassName>org.eclipse.che.api.workspace.server.dto.DtoServerImpls</genClassName>
+                    <genClassName>org.eclipse.che.api.workspace.shared.dto.DtoServerImpls</genClassName>
                     <impl>server</impl>
                 </configuration>
             </plugin>


skabashnyuk:
Can you explain more your motivation?

tareqhs:
Hi,

this is slightly strange but I am trying to communicate with Che6 environment from a Che3 environment. everything works except for when I try to use ``che-core-api-workspcae-shared`` to reuse the Che6 DTOs without having to re-model the JSON responses. ``org.eclipse.che.api.workspace.server.dto.DtoServerImpls`` appears in Che6 although the DTO package is ``org.eclipse.che.api.workspace.shared.dto`` which conflicts with Che3's ... though this tiny trick will solve everything.

skabashnyuk:
I doubt that your motivation is enough to do such change. Because
- it's not aligned with other modules with the same structure.
- it can cause some backward compatibility issues.

@garagatyi @sleshchenko  thoughts?

tareqhs:
Hi,

It does not break existing code because the DTO APIs are untouched neither are the workspace DTOs themselves. This merely puts the same generated classes in a different package. The package itself is loaded dynamically via service loading so we do not have any lookup issues or class conflicts.

garagatyi:
It is definitely a trick that introduces differences between the configuration of different modules. I believe that we should not do those tricks unless we know who/how/when will fix it in a better way that doesn't confuse developers.

skabashnyuk:
@tareqhs if it's loaded dynamically why it is an issue for you?

tareqhs:
As I explained earlier, the class ``org.eclipse.che.api.workspace.server.dto.DtoServerImpls`` is identical to that in the legacy workspace API , changing the packaging solves it

skabashnyuk:
Are you using generated class directly by full FQN or with help of Che DTO factory? I still don't get why you have an issue with that.

tareqhs:
Of course, but still the class has an identical FQN with ``DtoServerImpls`` in legacy code.
Actually in a second thought I can simply change it in legacy code ..

PR #9163, opened 2018-03-21
CHE-9095: Improve Theia build stability on plugin list changes

### LINKED ISSUE #9095: Sometimes rebuild Theia with plugins applied by env THEIA_PLUGINS fails
### Description
Sometimes rebuild Theia with plugins applied by env THEIA_PLUGINS fails:
yarn install v1.3.2
info No lockfile found.
[1/4] Resolving packages...
warning @theia/typescript > @theia/core > electron > electron-download > nugget > progress-stream > through2 > xtend > object-keys@0.4.0: 
[2/4] Fetching packages...
warning vscode-base-languageclient@0.0.1-alpha.3: The engine "vscode" appears to be invalid.
info fsevents@1.1.3: The platform "linux" is incompatible with this module.
info "fsevents@1.1.3" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
warning "@theia/extension-manager > @theia/application-package > font-awesome-webpack@0.0.5-beta.2" has unmet peer dependency "font-awesome@>=4.3.0".
[4/4] Building fresh packages...
[6/7] ⠐ find-git-repositories: CXX(target) Release/obj.target/findGitRepos/src/FindGitRepos.o
[2/7] ⠐ electron
[3/7] ⠐ node-pty:    ^~~~
[4/7] ⠐ nsfw:    ^~~~
error /home/user/default/theia/node_modules/vscode-ripgrep: Command failed.
Exit code: 1
Command: node ./lib/postinstall.js
Arguments: 

More: 
yarn manifest: 
  No manifest

Lockfile: 
  No lockfile

Trace: 
  Error: /home/theia/node_modules/vscode-ripgrep: Command failed.
  Exit code: 1
  Command: sh
  Arguments: -c node ./lib/postinstall.js
  Directory: /home/theia/node_modules/vscode-ripgrep
  Output:
  Downloading ripgrep failed: Error: Request failed with code 401
      at ProcessTermError.MessageError (/opt/yarn/lib/cli.js:186:110)
      at new ProcessTermError (/opt/yarn/lib/cli.js:226:113)
      at ChildProcess.<anonymous> (/opt/yarn/lib/cli.js:30281:17)
      at emitTwo (events.js:126:13)
      at ChildProcess.emit (events.js:214:7)
      at maybeClose (internal/child_process.js:925:16)
      at Socket.stream.socket.on (internal/child_process.js:346:11)
      at emitOne (events.js:116:13)
      at Socket.emit (events.js:211:7)
      at Pipe._handle.close [as _onclose] (net.js:567:12)

mmorhun:
### What does this PR do?
Adds support for Theia native plugins from filesystem
Adds support for Theia native plugins which is hosted in git repository
Improves build stability. Skips Invalid plugins (where possible).

### Plugin format
To set custom Theia plugins one should list them in `THEIA_PLUGINS` environment variable, each plugin should be separated by `,`. For example: `THEIA_PLUGINS=@theia/go,@theia/python`

Plugin versions are supported and could be specified after `:`. Example: `plugin-name:5.6.7`

 - if a plugin is native Theia plugin, name and optionally version should be specified. For example: `@theia/python:0.3.7`
- If a plugin is located on local file system (useful for developing and testing purposes), one should specify absolute path to the plugin folder itself (not parent folder with `lerna.json`) instead of version. Example: `plugin-name:/path/my-plugin`
- If a plugin is located in git repository, the following format should be used: 
`plugin-name:git://github.com/username/plugin-repository`
Note, field `name` in plugin's `package.json` should be the same as `plugin-name` at the beggining of plugin defining.

Partially fixes https://github.com/eclipse/che/issues/9095

### DIFF
--- a/dockerfiles/theia/src/main.js
+++ b/dockerfiles/theia/src/main.js
@@ -11,46 +11,151 @@
  */
 const fs = require("fs");
 const cp = require("child_process");
-const theiaRoot = '/home/theia';
-const theiaPath = theiaRoot + '/package.json';
 
-const defaultTheiaPath = `/home/default/theia`;
-const defaultConfig = require(`${defaultTheiaPath}/package.json`);
+const theiaRoot = '/home/theia';
+const defaultTheiaRoot = '/home/default/theia';
+const gitPluginsRoot = '/tmp/theia-plugins-from-git-repos';
+
+cp.execSync(`rm -rf ${gitPluginsRoot}; mkdir -p ${gitPluginsRoot}`);
 
 process.chdir(theiaRoot);
+prepareTheia();
 
-let theiaConfig;
-if (fs.existsSync(theiaPath)) {
-    handlePromise(callRun());
-}
-else {
-    let pluginString = process.env.THEIA_PLUGINS;
-    let pluginList = [];
-    if (pluginString && pluginString.length !== 0) {
-        let arr = pluginString.split(',');
-        pluginList = [...arr];
-        theiaConfig = defaultConfig;
-        let dep = theiaConfig.dependencies;
-        for (let d of pluginList) {
-            // check if plugin has a version using format pluginName:versionNumber
-            if (d.indexOf(":") > -1) {
-                let newDep = d.split(":");
-                let depName = newDep[0].trim();
-                let depVersion = newDep[1].trim();
-                dep[depName] = depVersion;
-                continue;
-            }
-            if (!dep.hasOwnProperty(d)) {
-                dep[d] = "latest";
-            }
-        }
-        fs.writeFileSync(theiaPath, JSON.stringify(theiaConfig));
-        handlePromise(callYarn().then(callBuild).then(callRun));
-    } else {
-        cp.execSync(`rsync -rv ${defaultTheiaPath}/ ${theiaRoot} --exclude 'node_modules' --exclude 'yarn.lock'`);
-        handlePromise(callRun());
+/** Rebuilds if needed and runs Theia */
+function prepareTheia() {
+    const extraPlagins = getExtraTheiaPlugins();
+    let currentPlugins = getTheiaPlugins();
+    let newPlugins = getDefaultTheiaPlugins();
+
+    if (Object.keys(currentPlugins).length === 0) {
+        // first run
+        copyDefaultTheiaBuild();
+        currentPlugins = getTheiaPlugins();
     }
 
+    for (let pluginName in extraPlagins) {
+        newPlugins[pluginName] = extraPlagins[pluginName];
+    }
+
+    if (!isPluginsEqual(currentPlugins, newPlugins)) {
+        console.log('Plugins set changed. Rebuilding Theia...');
+        rebuildTheiaWithNewPluginsAndRun(newPlugins);
+    } else {
+        handlePromise(callRun());
+    }
+}
+
+function isPluginsEqual(pls1, pls2) {
+    if (Object.keys(pls1).length !== Object.keys(pls2).length) {
+        return false;
+    }
+
+    for (let dep in pls1) {
+        if (pls1[dep] !== pls2[dep]) {
+            return false;
+        }
+    }
+
+    return true;
+}
+
+function copyDefaultTheiaBuild() {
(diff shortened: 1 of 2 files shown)


dockerfiles/theia/src/main.js
```diff
@@ -11,46 +11,159 @@
...
+            cp.execSync(`git clone --depth=1 --quiet ${gitRepository} ${pluginPath}`);
+        } catch (error) {
+            // failed again, skip plugin
+            console.error('Failed to get plugin: ' + pluginName + '. Skipping... Is the url specified properly?');
+            return;
+        }
     }
 
+    const rootPackageJson = require(pluginPath + 'package.json');
+    if (rootPackageJson['name'] === pluginName && !fs.existsSync(pluginPath + 'lerna.json')) {
```
benoitf, line 141: question: do we need to use all synchronous API or we could use async ? and then install in parallel ?

mmorhun: @benoitf what we can do in parallel that's cloning of git repositories and building them if needed.
But this makes the logic of the script more complicated. We have to deal with paralleling and barrier then, which is not supported in nodejs natively. So, some kind of emulation is needed. We'll end up with batch of promises wrapped in timeouts and async await chain up to the 'main' function. More, not all the plugins are git-hosted, but to have similar code we'll have to use promises for other plugins where they aren't needed actually. But the main problem is stability of that code. Having parallelism emulation and all that stuff is error prone. Also it could be not such efficient as it looks like with this approach.
Bringing real parallelism with child process library also isn't trivial. Threads may interact via IPC or some libs for shared memory, which is require new dependency and again much more complicated and error prone code.
And taking into account, that we are going to have new plugins model - this script will be removed then. So, I think, that it is not needed for now.

AndrienkoAleksandr, line 141: Seems in this case we need check if node_modules and lib exists and build plugin if needed.  We can split  build logic to separated method and reuse it here. @mmorhun  What do you think?

dockerfiles/theia/src/main.js
```diff
@@ -11,46 +11,159 @@
...
+        }
+    }
+
+    return true;
+}
 
-const defaultTheiaPath = `/home/default/theia`;
-const defaultConfig = require(`${defaultTheiaPath}/package.json`);
+function copyDefaultTheiaBuild() {
+    cp.execSync(`cp -r ${defaultTheiaRoot}/node_modules ${theiaRoot} && cp ${defaultTheiaRoot}/yarn.lock ${defaultTheiaRoot}/package.json ${theiaRoot}`);
```
AndrienkoAleksandr, line 61: Maybe split this commad:
`cp.execSync(`cp -r ${defaultTheiaRoot}/node_modules ${theiaRoot}`);`
 `cp.execSync(`cp ${defaultTheiaRoot}/yarn.lock ${defaultTheiaRoot}/package.json ${theiaRoot}`);`
It should be a bit more readable. But up to you.

mmorhun: I just don't want to spawn one more child process for this

AndrienkoAleksandr (approved):

dockerfiles/theia/src/main.js
mmorhun: Well, we do not know how to build this plugin in general case. And most of the plagins (at least generated with Theai plugin generator, which is majority) use lerna (i.e. else block will be executed). I use this just to be able to include kinda 'ready' plugin (as you know, will work with the plugin's subdirectory). This can be used to speed up Theia run.
But if you think that we should do it I'll add the build here too.

evidolob (approved):

dockerfiles/theia/src/main.js
```diff
@@ -11,46 +11,159 @@
...
+const defaultTheiaRoot = '/home/default/theia';
+const gitPluginsRoot = '/tmp/theia-plugins-from-git-repos';
+
+prepareTheia();
+
+/** Rebuilds if needed and runs Theia */
+function prepareTheia() {
+    process.chdir(theiaRoot);
+
+    const extraPlagins = getExtraTheiaPlugins();
```
ashumilova: extraPlagins -> extraPlugins

ashumilova (approved):

benoitf (approved):

PR #9170, opened 2018-03-21
CHE-8675: fix reconnection to a WebSocket entrypoint

### LINKED ISSUE #8675: Dashboard doesn't reconnect to a WebSocket endpoint if the previous connection failed
### Description
Dashboard doesn't reconnect to a WebSocket endpoint if the previous connection failed. As result Dashboard doesn't receive any event about workspaces status changes.

### Reproduction Steps
It happens on all infrastructures Docker, OpenShift, Kubernetes. But it can be reproduced stably only on Kubernetes infrastructure, on other, it happens quite rarely.

I suppose it happens because of `nginx` ingress controller. K8s infra adds new ingress for each workspace server and `nginx` controller restarts itself when new ingress rule is added. So, there is an ability to configure ingress controller for created workspaces ingresses. So only Dashboard should be fixed, and WebSocket connections can be made stable with another ingress controller (I suppose traefik ingress controller can be used for such purposes)

It can be reproduced in the following way:
1. Deploy Che on Kubernetes.
2. Create workspace.
3. Open Developer Console and watch WebSocket connections.
4. Run workspace in IDE.
5. Check a WebSocket connection to wsmaster from the dashboard.
Actual: Dashboard websocket connection is closed, as result Dashboard shows that workspace is STARTING when it is STARTED.
Expected: Dashboard websocket connection is pending, Dashboard receives events that workspace is started and updates Workspace Status Icon.

<details>
<summary>More see screencast</summary>

dashboard_websocket

</details>

**OS and version:**    

**Diagnostics:**

akurinnoy:
Signed-off-by: Oleksii Kurinnyi <okurinny@redhat.com>

### What does this PR do?
This PR adds the ability to handle the 'close' event of JSON-RPC client and tries to re-connect the client if the previous connection fails.

fix https://github.com/eclipse/che/issues/8675

#### Release Notes

N/A - bugfix

#### Docs PR

N/A - bugfix

### DIFF
--- a/dashboard/src/components/api/json-rpc/che-json-rpc-master-api.ts
+++ b/dashboard/src/components/api/json-rpc/che-json-rpc-master-api.ts
@@ -36,12 +36,52 @@ const UNSUBSCRIBE: string = 'unsubscribe';
  * @author Ann Shumilova
  */
 export class CheJsonRpcMasterApi {
+  private $log: ng.ILogService;
+  private $timeout: ng.ITimeoutService;
   private cheJsonRpcApi: CheJsonRpcApiClient;
   private clientId: string;
 
-  constructor (client: ICommunicationClient, entrypoint: string) {
+  private maxReconnectionAttempts = 5;
+  private reconnectionAttemptNumber = 0;
+  private reconnectionDelay = 30000;
+
+  constructor (client: ICommunicationClient,
+               entrypoint: string,
+               $log: ng.ILogService,
+               $timeout: ng.ITimeoutService) {
+    this.$log = $log;
+    this.$timeout = $timeout;
+
+    client.addListener('open', () => this.onConnectionOpen());
+    client.addListener('close', () => this.onConnectionClose(entrypoint));
+
     this.cheJsonRpcApi = new CheJsonRpcApiClient(client);
     this.connect(entrypoint);
+  }
+
+  onConnectionOpen(): void {
+    this.$log.log('WebSocket connection is opened.');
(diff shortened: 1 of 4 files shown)


olexii4 (approved):

dashboard/src/components/api/json-rpc/che-json-rpc-master-api.ts
```diff
@@ -36,21 +36,73 @@ const UNSUBSCRIBE: string = 'unsubscribe';
...
+
+    client.addListener('open', () => this.onConnectionOpen());
+    client.addListener('close', () => this.onConnectionClose(entrypoint));
+
     this.cheJsonRpcApi = new CheJsonRpcApiClient(client);
     this.connect(entrypoint);
   }
 
+  onConnectionOpen(): void {
+    this.$log.log('WebSocket connection is opened.');
```
benoitf: should we keep this log ?

benoitf (approved):

dashboard/src/components/api/json-rpc/che-json-rpc-master-api.ts
akurinnoy: We could keep this log in case if it is re-connection
```
if (this.reconnectionAttemptNumber !== 0) {
  this.$log.log('...');
}
```
wdyt?

benoitf: yep

ashumilova (approved):

Issue #6444, opened 2017-09-26
Publish Maven artifacts to central repository


benoitf:
Eclipse Che artifacts should be available/synchronized on maven central repository instead of having only a private repository.

Example:
https://mvnrepository.com/search?q=eclipse+che
users are not finding eclipse che

while end users would expect to find it at http://central.maven.org/maven2/org/eclipse/ (with che folder)

Also it would help ppl by having local mirrors.

https://maven.apache.org/guides/mini/guide-central-repository-upload.html

Moved to central:
- [x] che-dev
- [x] che-parent
- [x] che-lib
- [x] che-docs
- [x] che
- [x] che-ide-extension
- [x] che-ide-server-extension
- [x] che-plugin-embedjs
- [x] che-plugin-json
- [x] che-plugin-menu
- [x] che-plugin-wizard

benoitf:
it should fix as well issues when codenvy maven repository is not reliable

```
[ERROR] Failed to execute goal on project che-keycloak-plugin-server: Could not resolve dependencies for project com.redhat.che:che-keycloak-plugin-server:jar:1.0.0-SNAPSHOT: Failed to collect dependencies at org.eclipse.che.core:che-core-api-core:jar:5.18.0 -> 
org.eclipse.che.core:che-core-api-dto:jar:5.18.0: Failed to read artifact descriptor for org.eclipse.che.core:che-core-api-dto:jar:5.18.0: Could not transfer artifact org.eclipse.che.core:che-core-api-dto:pom:5.18.0 from/to codenvy-public-repo (https://maven.codenvycorp.com/content/groups/public/): 
Failed to transfer file: https://maven.codenvycorp.com/content/groups/public/org/eclipse/che/core/che-core-api-dto/5.18.0/che-core-api-dto-5.18.0.pom.
 Return code is: 502 , ReasonPhrase:Proxy Error. -> [Help 1]
```

riuvshin:
we have some issues with che-lib
```
failureMessage    Missing: no sources jar found in folder '/org/eclipse/che/lib/org-eclipse-jdt-core-repack/6.100.0
failureMessage    Missing: no javadoc jar found in folder '/org/eclipse/che/lib/che-tomcat8-slf4j-logback/6.100.0'
failureMessage    Missing: no javadoc jar found in folder '/org/eclipse/che/lib/org-eclipse-jdt-core-repack/6.100.0'
```
this should be fixed, otherwise we wont be able to deploy on central

[...]

PR #9190, opened 2018-03-22
Use templates only to deploy Che to OpenShift

ghost:
## The purpose of this PR is to:

1. Eventually get rid of deployment scripts to avoid deployment complexity
2. Use OpenShift way of creating objects when user input is required

## What is done

1. OpenShift templates and other objects are introduced
2. All env vars are parameterized and a user can provide own values
3. Readme answers some questions and covers deployment of all Che flavors in all possible configurations.

## Why not a single template?

There's no universal template that will fit all use cases. It will be easier to use upstream che-server template for those who depend on it.

## Is it tested?

Yes. MiniShift and OSO. Both single and multi user. http and https

## Will existing scripts go away?

Eventually, yes. But not immediately. As soon as those who depend on current deployment scripts are ready to migrate, scripts will be deleted.

## Does Che deployment have all envs as parameters?

No, only those that make sense at a first glance. Further PRs are welcome.

## Will these changes break CI tests?

Even though there are changes in eclipse/che-keycloak Dockerfile, a different entrypoint script is used in Keycloak deployment. So, **both** deployment script **and** templates works well, even though they use the same images.

## What about Keycloak configuration pod?

Master and Che realm, as well as admin user for Keycloak and Che are created in keycloak image entrypoint.

## Will docs be provided?

There's README.md to start with, but eventually docs will be updated to reference template based deployment method.

## How to test?

Follow instructions in README. Use eivantsov/keycloak as Keycloak image for Keycloak deployment:

`oc process -f multi/keycloak-template.yaml -p ROUTING_SUFFIX=$(minishift ip).nip.io -p IMAGE_KEYCLOAK=eivantsov/keycloak | oc apply -f -`


deploy/openshift/templates/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Why many templates?
+
+Since there are several Che flavors and configuration options, it is not possible to list all objects in one template, while duplicating templates makes it more difficult to maintain them.
+
+## Che server logs persistence
+
+All of the below commands create a PVC for Che server and Che deployment uses this PVC.
+
+It is possible to avoid using PVC for Che server - just skip creating pvc `oc apply -f pvc/che-server-pvc.yaml` and `oc set volume dc/che --add -m /data --name=che-data-volume --claim-name=che-data-volume` commands.
```
sleshchenko, line 56: Actually, PVC is used by Single-User Che to store H2 Database file. Please, add a note that without PVC Single-User Che will lose all data after Pod restarting.

ghost: Good point will do. For single user PVC is created by default - this is what I suggest in an example command.

But once again - good point.

deploy/openshift/templates/che-server-template.yaml
```diff
@@ -0,0 +1,279 @@
...
+            value: "${WS_PROTOCOL}://che-${NAMESPACE}.${ROUTING_SUFFIX}/api/websocket"
+          - name: CHE_DEBUG_SERVER
+            value: 'false'
+          - name: CHE_INFRASTRUCTURE_ACTIVE
+            value: openshift
+          - name: CHE_INFRA_KUBERNETES_BOOTSTRAPPER_BINARY__URL
+            value: "${PROTOCOL}://che-${NAMESPACE}.${ROUTING_SUFFIX}/agent-binaries/linux_amd64/bootstrapper/bootstrapper"
+          - name: CHE_INFRA_KUBERNETES_MACHINE__START__TIMEOUT__MIN
+            value: '5'
+          - name: CHE_INFRA_KUBERNETES_MASTER__URL
```
sleshchenko: Why doesn't this parameter have value? It makes sense to configure it for multi-cluster mode (When Che Server is started on one OS cluster while Workspace are created on another one)

deploy/openshift/templates/che-server-template.yaml
```diff
@@ -0,0 +1,279 @@
...
+          - name: CHE_INFRA_KUBERNETES_OAUTH__TOKEN
+            value: "${TOKEN}"
+          - name: CHE_INFRA_KUBERNETES_PASSWORD
+            value: "${OPENSHIFT_PASSWORD}"
+          - name: CHE_INFRA_OPENSHIFT_PROJECT
+            value: "${CHE_INFRA_OPENSHIFT_PROJECT}"
+          - name: CHE_INFRA_KUBERNETES_PVC_STRATEGY
+            value: "${CHE_INFRA_KUBERNETES_PVC_STRATEGY}"
+          - name: CHE_INFRA_KUBERNETES_PVC_PRECREATE__SUBPATHS
+            value: 'true'
```
sleshchenko: Please add a variable here. It makes sense to configure it. On newest OpenShift instances it may be disabled, and workspace start may speed up.

deploy/openshift/templates/che-server-template.yaml
```diff
@@ -0,0 +1,279 @@
...
+            value: 'false'
+          - name: CHE_INFRASTRUCTURE_ACTIVE
+            value: openshift
+          - name: CHE_INFRA_KUBERNETES_BOOTSTRAPPER_BINARY__URL
+            value: "${PROTOCOL}://che-${NAMESPACE}.${ROUTING_SUFFIX}/agent-binaries/linux_amd64/bootstrapper/bootstrapper"
+          - name: CHE_INFRA_KUBERNETES_MACHINE__START__TIMEOUT__MIN
+            value: '5'
+          - name: CHE_INFRA_KUBERNETES_MASTER__URL
+          - name: CHE_INFRA_KUBERNETES_OAUTH__TOKEN
+            value: "${TOKEN}"
```
sleshchenko: Why `OPENSHIFT_USERNAME`, `OPENSHIFT_PASSWORD` but `TOKEN`? Maybe it would be better to name them in the same way.

deploy/openshift/templates/che-server-template.yaml
```diff
@@ -0,0 +1,279 @@
...
+          - name: CHE_INFRA_KUBERNETES_PASSWORD
```
sleshchenko: It makes sense to place `CHE_INFRA_KUBERNETES_PASSWORD` and `CHE_INFRA_KUBERNETES_USERNAME` together.

deploy/openshift/templates/che-server-template.yaml
```diff
@@ -0,0 +1,279 @@
...
+            value: "/data/logs"
+          - name: CHE_LOG_LEVEL
+            value: INFO
+          - name: CHE_MULTIUSER
+            value: "${CHE_MULTIUSER}"
+          - name: CHE_KEYCLOAK_AUTH__SERVER__URL
+            value: "${CHE_KEYCLOAK_AUTH__SERVER__URL}"
+          - name: CHE_JDBC_URL
+            value: "${CHE_JDBC_URL}"
+          - name: CHE_OAUTH_GITHUB_CLIENTID
```
sleshchenko: How can it be configured if there is no specified value?

deploy/openshift/templates/che-server-template.yaml
```diff
@@ -0,0 +1,279 @@
...
+        restartPolicy: Always
+        schedulerName: default-scheduler
+        securityContext: {}
+        serviceAccount: che
+        serviceAccountName: che
+        terminationGracePeriodSeconds: 30
+    test: false
+    triggers:
+    - type: ConfigChange
+  status:
```
sleshchenko: Are you sure that it makes sense to specify status here?

deploy/openshift/templates/multi/postgres-template.yaml
```diff
@@ -0,0 +1,126 @@
...
+apiVersion: v1
+metadata:
+  name: che
+  annotations:
+    description: Che
+objects:
+- apiVersion: v1
+  kind: DeploymentConfig
+  metadata:
+    creationTimestamp: null
```
sleshchenko: Why do we need it?

dockerfiles/keycloak/Dockerfile
```diff
@@ -11,3 +11,4 @@ ADD . /scripts/
 USER root
 RUN chgrp -R 0 /scripts && \
     chmod -R g+rwX /scripts
+
```
riuvshin: is that mandatory change? :)

ghost: New line?

deploy/openshift/templates/README.md
```diff
@@ -0,0 +1,155 @@
...
+Without PVC Che deployment can use Rolling update strategy: `-p STRATEGY=Rolling` that causes zero downtime when updating Che server.
+However, in this case, if you need to persist logs you need to use solutions like Kibana.
+
+## Creating Workspace Objects
+
+By default, che SA is used to create workspace objects in the same namespace with Che server. This is configurable - you may provide login and password or token, and set
```
benoitf: FYI sometimes Che is written `che` sometimes `Che`

benoitf: is there any word expected after `and set` ?

deploy/openshift/templates/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Upgrade from http to https
+
+1. Update Che deployment with `PROTOCOL=https, WS_PROTOCOL=wss, TLS=true`
+2. Manually edit or recreate routes for Che and Keycloak `oc apply -f https`
+3. Once done, go to `https://keycloak.${NAMESPACE}.${ROUTING_SUFFIX}`, log in to admin console.
+Default credentials are `admin:admin`.
+Go to Clients, `che-public` client and edit **Valid Redirect URIs** and **Web Origins** URLs so that they use **https** protocol.
+You do not need to do that if you initially deploy Che with 	https support.
```
benoitf: like https here have spaces around

deploy/openshift/templates/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Deploy single user Che (https)
+
+```
+oc new-project che
+
+oc apply -f pvc/che-server-pvc.yaml; \
+oc process -f che-server-template.yaml -p ROUTING_SUFFIX=$(minishift ip).nip.io \
+		-p PROTOCOL=https \
+		-p WS_PROTOCOL=wss \
```
benoitf: could WS_PROTOCOL/TLS be automatically computed based on the fact that it's https ?

deploy/openshift/templates/README.md
```diff
@@ -0,0 +1,155 @@
+## Context
+
+All commands are executed in the directory where this README is located.
+
+## Templates and Parameters
+
+Templates are provided with a set of predefined params which you can override with with `-p key=value`. If you miss envs and parameters, you can add them to your template.
```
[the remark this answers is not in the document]
garagatyi: +1

deploy/openshift/templates/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+IMPORTANT! **ROUTING_SUFFIX** value in the below commands `$(minishift ip).nip.io` works for MiniShift only. You MUST provide **own routing suffix** when deploying to OCP cluster.
```
garagatyi: It would be useful to have a command to get this value if there is a command for that.

ghost: I wish I knew such a command :)

deploy/openshift/templates/README.md
ghost: I wish a template can handle some ifs. I failed to find a good universal solution. Any help is appreciated.

deploy/openshift/templates/README.md
```diff
@@ -0,0 +1,155 @@
...
+Default credentials are `admin:admin`.
+Go to Clients, `che-public` client and edit **Valid Redirect URIs** and **Web Origins** URLs so that they use **https** protocol.
+You do not need to do that if you initially deploy Che with 	https support.
+
+## Deploy single user Che (http)
+
+```
+oc new-project che
+
+oc process -f che-server-template.yaml -p ROUTING_SUFFIX=$(minishift ip).nip.io | oc apply -f -; \
```
garagatyi: Interestingly `oc process` supports golang templates with `-t` option and looks like they support `if` clauses https://golang.org/pkg/text/template/
With this approach, we would have something similar to helm I suppose.

deploy/openshift/templates/README.md
benoitf: could we create a dummy service then we got the default routing suffix and then we can display/reuse it ?

deploy/openshift/templates/README.md
```diff
@@ -0,0 +1,155 @@
...
+oc apply -f pvc/che-server-pvc.yaml; \
+oc set volume dc/che --add -m /data --name=che-data-volume --claim-name=che-data-volume
```
garagatyi: I would suggest using set before applying template

ghost: We don't have dc/che at that point. I need to create Che deploymentConfig and then set volume for it.

The spec says `oc set volumes RESOURCE/NAME --add|--remove|--list [options]` so I need an object to modify, while process just produces output.

deploy/openshift/templates/README.md
ghost: Not sure I understand. Dummy service for what?

benoitf: @eivantsov If I create a dummy application on openshift like nginx app and I create a service nginx and expose port 80, openshift will provide me an URL. From this URL I can guess the "routing suffix" so user has a way to find it (through a command that can try to create app, create service and route and then display the url and delete project)

It's just to have a way to provide to users "hey we can give you some hints around this routing suffix"

ghost: A lame workaround may be updating Che dc after routes for Che and Keycloak are created. This will add complexity to commands, but this way, no input is required from a user

benoitf: in my approach, it was a way to see what the route could be.
in your approach, we create using defaults, grab values and update again.

in both case it would help "newcomers" so I'm fine.

deploy/openshift/templates/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+It is possible to avoid using PVC for Che server - just skip creating pvc `oc apply -f pvc/che-server-pvc.yaml` and `oc set volume dc/che --add -m /data --name=che-data-volume --claim-name=che-data-volume` commands.
+Without PVC Che deployment can use Rolling update strategy: `-p STRATEGY=Rolling` that causes zero downtime when updating Che server.
+However, in this case, if you need to persist logs you need to use solutions like Kibana.
+
+## Creating Workspace Objects
+
+By default, che SA is used to create workspace objects in the same namespace with Che server. This is configurable - you may provide login and password or token, and set
+
+## Upgrade from http to https
```
garagatyi: I would put this section somewhere in the end - more advanced stuff

deploy/openshift/templates/README.md
```diff
@@ -0,0 +1,155 @@
...
+oc process -f multi/postgres-template.yaml | oc apply -f -; \
+oc process -f multi/keycloak-template.yaml -p ROUTING_SUFFIX=$(minishift ip).nip.io | oc apply -f -; \
+oc apply -f pvc/che-server-pvc.yaml; \
+oc process -f che-server-template.yaml -p ROUTING_SUFFIX=$(minishift ip).nip.io -p CHE_MULTIUSER=true | oc apply -f -; \
+oc set volume dc/che --add -m /data --name=che-data-volume --claim-name=che-data-volume
+```
+
+## Deploy multi user Che with bundled Keycloak and Postgres (https)
+
+```
```
garagatyi, line 106: Looks like step with creation of the project is missing here

deploy/openshift/templates/che-server-template.yaml
```diff
@@ -0,0 +1,279 @@
...
+          readinessProbe:
+            httpGet:
+              path: /api/system/state
+              port: 8080
+              scheme: HTTP
+            initialDelaySeconds: 15
+            timeoutSeconds: 60
+          name: che
+          ports:
+          - containerPort: 8080
```
garagatyi: Looks like ports are declared twice

deploy/openshift/templates/README.md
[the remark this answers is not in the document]
ghost: Good point. Examples updated

deploy/openshift/templates/README.md
ghost: Added a sub-section on How to get routing suffix.

deploy/openshift/templates/README.md
ghost: Fixed

deploy/openshift/templates/README.md
ghost: Fixed

deploy/openshift/templates/README.md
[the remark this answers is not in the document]
ghost: Added a note in readme

dockerfiles/keycloak/Dockerfile
ghost: Fixed

deploy/openshift/templates/multi/postgres-template.yaml
ghost: Removed

deploy/openshift/templates/che-server-template.yaml
ghost: Removed

deploy/openshift/templates/che-server-template.yaml
ghost: Fixed

deploy/openshift/templates/che-server-template.yaml
ghost: Fixed

deploy/openshift/templates/che-server-template.yaml
ghost: Fixed

deploy/openshift/templates/che-server-template.yaml
ghost: FIxed

deploy/openshift/templates/README.md
ghost: Fixed

deploy/openshift/templates/multi/keycloak-template.yaml
```diff
@@ -0,0 +1,172 @@
...
+        restartPolicy: Always
+        volumes:
+        - name: keycloak-data
+          persistentVolumeClaim:
+            claimName: keycloak-data
+        - name: keycloak-log
+          persistentVolumeClaim:
+            claimName: keycloak-log
+    test: false
+  status: {}
```
garagatyi: Is it needed?

deploy/openshift/templates/multi/keycloak-template.yaml
```diff
@@ -0,0 +1,172 @@
...
+    labels:
+      io.kompose.service: keycloak-data
+    name: keycloak-data
+  spec:
+    accessModes:
+    - ReadWriteOnce
+    resources:
+      requests:
+        storage: 1Gi
+  status: {}
```
garagatyi: Is it needed?

deploy/openshift/templates/multi/keycloak-template.yaml
```diff
@@ -0,0 +1,172 @@
...
+    labels:
+      io.kompose.service: keycloak-log
+    name: keycloak-log
+  spec:
+    accessModes:
+    - ReadWriteOnce
+    resources:
+      requests:
+        storage: 1Gi
+  status: {}
```
garagatyi: Is it needed?

deploy/openshift/templates/multi/postgres-template.yaml
```diff
@@ -0,0 +1,126 @@
...
+          volumeMounts:
+          - mountPath: /var/lib/pgsql/data
+            name: postgres-data
+        restartPolicy: Always
+        volumes:
+        - name: postgres-data
+          persistentVolumeClaim:
+            claimName: postgres-data
+    test: false
+  status: {}
```
garagatyi: Is it needed?

deploy/openshift/templates/multi/postgres-template.yaml
```diff
@@ -0,0 +1,126 @@
...
+      io.kompose.service: postgres
+    name: postgres
+  spec:
+    ports:
+    - name: "5432"
+      port: 5432
+      targetPort: 5432
+    selector:
+      io.kompose.service: postgres
+  status:
```
garagatyi: Is it needed?

deploy/openshift/templates/README.md
garagatyi: Looks like it is not.

garagatyi (approved):

deploy/openshift/templates/README.md
ghost: I can see it

riuvshin:
@eivantsov  Can you please add examples of how to run CHE with some non-default configuration for instance with a higher timeout of something. And also an example of how to update the configuration of already deployed CHE (if that is different from the first example I've mentioned)

deploy/openshift/templates/che-server-template.yaml
```diff
@@ -0,0 +1,274 @@
...
+  displayName: OpenShift Username
+  description: OpenShift username that will be used to create workspace objects
+- name: OPENSHIFT_PASSWORD
+  displayName: OpenShift Password
+  description: OpenShift Password that will be used to create workspace objects
+- name: OPENSHIFT_TOKEN
+  displayName: OpenShift token
+  description: OpenShift token to be used to create workspace objects. Must be set if username/password are impossible to use. Do not set if usrname/pass are provided!
+- name: PROTOCOL
+  displayName: htpps or http protocol
```
benoitf: testing script I noticed a typo on `https`

deploy/openshift/templates/che-server-template.yaml
```diff
@@ -0,0 +1,274 @@
...
+  value: 'false'
+- name: OPENSHIFT_USERNAME
+  displayName: OpenShift Username
+  description: OpenShift username that will be used to create workspace objects
+- name: OPENSHIFT_PASSWORD
+  displayName: OpenShift Password
+  description: OpenShift Password that will be used to create workspace objects
+- name: OPENSHIFT_TOKEN
+  displayName: OpenShift token
+  description: OpenShift token to be used to create workspace objects. Must be set if username/password are impossible to use. Do not set if usrname/pass are provided!
```
benoitf: typo on username

eivantsov:
@riuvshin good point. Yes, I will do that: both envs and env file approach.

eivantsov:
Updated README with instructions on how to update Che deployment and pass env variables when deploying. Links to OpenShift docs included.

riuvshin (approved):

deploy/openshift/templates/multi/postgres-template.yaml
sleshchenko: Where is it removed? I still see few `creationTimestamp` fields in this template. Do I have an issue with browser cache?

sleshchenko (approved):
LGTM

deploy/openshift/templates/multi/keycloak-template.yaml
```diff
@@ -0,0 +1,165 @@
...
+          - name: KEYCLOAK_PASSWORD
+            value: admin
+          - name: PROTOCOL
+            value: "${PROTOCOL}"
+          - name: ROUTING_SUFFIX
+            value: "${ROUTING_SUFFIX}"
+          - name: CHE_KEYCLOAK_ADMIN_REQUIRE_UPDATE_PASSWORD
+            value: "${CHE_KEYCLOAK_ADMIN_REQUIRE_UPDATE_PASSWORD}"
+          image: '${IMAGE_KEYCLOAK}:${CHE_VERSION}'
+          command: ["/scripts/kc_realm_user.sh"]
```
l0rd, line 57: While testing I got `stat /scripts/kc_realm_user.sh: no such file or directory`. Isn't the right script `/scripts/keycloak_config.sh`?

ghost: You need to test with `-p KEYCLOAK_IMAGE=eivantsov/keycloak`

I added a new entrypoint and kept the old one for backward compatibility. But since the PR isn't merged yet, there's no new Keycloak image with the new entrypoint, hence - there's my image

l0rd: Ok got it thanks. I was able to start keycloak successfully.

l0rd:
While testing I noticed that the service `che-host` is associated with the 3 pods (che, postgre and keycloak). That may have weird side effects like that route `che-che.$(minishift ip).nip.io` is linked to keycloak.

image

I think that this can be related to the fact that the 3 templates have the same `.metadata.name`

l0rd:
Another thing that I have notice are some `Readiness probe failed` that appears systematically in the openshift events when keycloak and postgre are started. These do no harm but may be confusing for a user that is installing Che and may be fixed increasing the `initialDelaySeconds` of the Readiness probes.

l0rd:
Anyway the `README.md` is brilliant and would be even more brilliant if you could provide instructions to delete che, postgre and keycloak objects.

eivantsov:
@l0rd all good points. Thanks!

Updates:

1. Readiness probes for Postgres and Keycloak are optimized. On MiniShift I now do not see failed readiness probes for those two. Of course, everything is possible, but I checked multiple scenarios when some or all of the images are unavailable.

2. Liveness and readiness probes for Che are optimized. Che server fails to start if Keycloak is unavailable. Liveness probe takes that into account and restarts a container rather than tries all over again (it makes no sense to retry or wait).

3. Templates cleanup.

4. README update: how to delete resources using labels selector.

deploy/openshift/templates/che-server-template.yaml
```diff
@@ -0,0 +1,262 @@
...
+  description: Force an admin to update password after 1st login. True by default
+  value: 'true'
+- name: CHE_KEYCLOAK_AUTH__SERVER__URL
+  displayName: Identity provider URL
+  description: URL of a remote identity provider. Defaults to Keycloak bundled with Che multi user
+  value: '${PROTOCOL}://keycloak-${NAMESPACE}.${ROUTING_SUFFIX}/auth'
+- name: CHE_JDBC_URL
+  displayName: JDBC URL
+  description: URL Che server will use to connect to JDBC. Defaults to Postgres service bundled with Che multi user
+  value: 'jdbc:postgresql://postgres:5432/dbche'
```
benoitf: it seems this URL is for multi user che
the command of README using template with a single user che is failing then as it tries to connect on postgresql
did I miss something ?

benoitf: (single user default url being jdbc:h2:che)

ghost: Fixed. DB URL can be passed as env not a param, while for default single and multi user uses cases, server has got props

benoitf:
it says that at some point script could be removed
But could we still get a cli on top of these files so it could be easy for a dumb user like me ?
so no need to clone Eclipse Che repository, or grab many files from repository, check if it's up-to-date or not, etc ?

just execute a binary.

cli

eivantsov:
Looks like you have got it already )

benoitf:
@eivantsov yes I like the power/advanced mode/maintenability and flexibility of these new templates (over lot of duplicated lines) but I kinda like "default/simplicity" for onboarding

eivantsov:
@benoitf it does not necessarily mean that existing scripts go away entirely. Maybe they will be just dramatically simplified and there will be just one simple deploy script with a new basic uses cases: single vs multi user, http vs https etc.

As to CLI you are suggesting, yes it would be nice to have one, however, it is out of this issue's scope. If you are talking about CLI written in Golang, we need to spec it out first, then decide how to release and upload binaries etc.

garagatyi:
I think what @eivantsov said is a very good tradeoff between convenience and maintainability and code clarity.

benoitf:
@eivantsov on your cli example, is it really working or it's just a list of println ?
Also I noticed a typo on Congrats :D written contgrats ;-)
FYI my gif was not a go cli but it was in rust and working.

@garagatyi I said I liked these new templates so I'm not sure to see what you meant by this comment ?

benoitf:
@garagatyi I agree with you but this is also what I said (when I said `build CLI on top of these files`)
it's to leverage these nice templates.

eivantsov:
@benoitf I am not THAT advanced with Println to build a CLI for Che :) Probably Golang and Cobra is the right path here. I made one for myself - can deploy single and multi user Che in http and https mode:

```
he deploy single/multi --secure -r routing_suffix
```
Of course, MiniShift ip is detected and there are defaults, so a local http single user deployment will be just `che deploy` 

But looking at all the options when deploying to Che, I realized that such a CLI will duplicate oc client functionality and will either call oc binaries as exec or import oc go client directly.

I think such a CLI can be ussful for 1st Che deploy - we cannot be better than oc client when it comes to managing deployments and all related OS and K8S objects. But once again, this is a totally different story and can be discussed in a dedicated issue (i've already got a pile of lame golan code :) )

benoitf:
@eivantsov for me the key is the abstraction

let say I want to deploy che on k8s

```
$ che deploy k8s single
```
or on docker
```
$ che deploy docker single
```

if I understand all the OC tool commands then yes probably I may use oc commands at some points but just to give a try I would use cli.

Over language I tend to prefer `rust` over `golang` right now. (speed) but yes let's discuss on a new issue

eivantsov:
I just looked at what OpenShift and K8s clis use and Cobra was there.

Anyway, a universal CLI that deploys to 3 supported infras...that's quite an effort.

l0rd (approved):

l0rd:
I have tested on minishift and works fine. Great job @eivantsov

l0rd:
@benoitf @garagatyi @benoitf by the way we all agree that we don't need a CLI integration as part of this PR right?

benoitf:
yes

garagatyi:
yes

gorkem:
Do we want to retire the minishift add-on eventually as well? @l0rd @eivantsov

eivantsov:
@gorkem good question. That's two sets of yamls to maintain. Though from user's perspective, applying an addon is somewhat easier.

I tried to craft a simple CLI that deploys to MiniShift and OCP https://github.com/eivantsov/che-cli, lame code but works well, and all a user needs to do is:

```
oc login
oc new-project che
./eclipse-che deploy
```
to deploy to MiniShift. Works with OSO Pro too. I know Florent has some rust cli too.

Maybe we should look in this direction and start with an OpenShift CLI that will make it easy to deploy to any OpenShift cluster?

l0rd:
@gorkem that's a good point. A copy of these `yaml` files should be in rh-che repo too. So there will be 3 copies in 3 different repositories. Even if we retire the minishift addon we will still have the problem with rh-che. What about moving the yaml in a distinct github repository and referring to this repository? Then yaml could be versioned and published on maven central as part of the build. In this case the deployment scripts (and minishift addons) should download the yaml from the maven central as part of the deployment.

l0rd:
> @l0rd can't rh use templates for upsteam repo?
How? I mean we could add instructions for rh-che developers to clone upstream repo and use these `yaml` but that's not ideal. But the real problem is the openshift.io CI/CD that is not so flexible. Currently that's how [rh-che deployment config for osio is specified](https://github.com/openshiftio/saas-openshiftio/blob/master/dsaas-services/rh-che6.yaml).

eivantsov:
@benoitf do you mean maven central or Che nexus?

benoitf:
@eivantsov if https://github.com/eclipse/che/issues/6444 is done it should be maven central as well :-)

garagatyi:
@benoitf 👍 Yeah, we don't really need an additional repo to publish the artifact

l0rd:
@eivantsov this PR looks like Penelope's shroud :-) What is missing to merge it? I am asking because I have opened a similar [PR](https://github.com/redhat-developer/rh-che/pull/620) and I am waiting impatiently for this one to be merged.

eivantsov:
@l0rd thanks for your patience and a wonderful metaphor. I have been waiting for all interested parties to have their say on templates themselves, cross platform cli that deploys to all infras and related stuff. Besides, I have introduced some changes to `ocp.sh` that QE uses to run daily tests, and while testing discovered a bug where PVCs were not deleted upon workspace deletion if workspace objects are created in a shared dedicated namespace (https://github.com/eclipse/che/issues/9431) (default behavior in new templates).

Now that that QA tests show no regression , with a light heart, I merge this PR. Thanks for all suggestions and helpful remarks - they did help make templates and deployment docs better.

P.S. Old scripts and deployment yamls will stay here for a while and will be deleted in a separate PR.

Related docs PR: https://github.com/eclipse/che-docs/pull/391

Issue #9192, opened 2018-03-22
Incorrect editor tab name for pom.xml file 


RomanNikitenko:
### Reproduction Steps
1. Go Profile -> Preferences -> Code Formatter -> Java
2. Import Code Formatter for Workspace
3. Open pom.xml file
4. Rename artifactId field

### Expected behavior:
Editor tab will be renamed

### Observed behavior: 
Editor tab is not renamed corresponding to artifactId

![pom_bug](https://user-images.githubusercontent.com/5676062/37781946-697a05f0-2dfa-11e8-816a-6ec7d35ee7bb.gif)

**OS and version:**    
![che_ver](https://user-images.githubusercontent.com/5676062/37780882-55ac0854-2df8-11e8-8daf-9c3654883481.png)

PR #9193, opened 2018-03-22
Do not throw Exception when not project item is found in root

RomanNikitenko:
### What does this PR do?
We store some settings for workspace in .che folder at the same level as the projects
So, we should not throw Exception when not project item is found in root

#9192 


RomanNikitenko:
ci-test

SkorikSergey (approved):

dkulieshov (approved):

vparfonov (approved):

Issue #9202, opened 2018-03-23
Unexpected fail of CheckRestoringWorkspaceAfterStoppingWsAgentProcess selenium test on OCP


SkorikSergey:
### Description
We have unexpected fail of **CheckRestoringWorkspaceAfterStoppingWsAgentProcess** selenium test on OCP. The test fails while checking that workspace has **STOPPING** status before starting. We should clarify reason and fix it.

**Log:**
```
Workspace workspace7hkdh8, status=RUNNING, expected status=STOPPING
```

**Report with reproduced problem:**
https://ci.codenvycorp.com/view/qa/job/che-integration-tests-master-ocp/188/Selenium_tests_report/

SkorikSergey:
Not reproduced.

PR #9205, opened 2018-03-23
Selenium: fix unstable selenium tests

SkorikSergey:
### What does this PR do?
This PR:
- don't check that workspace has **STOPPING** status before removing in **DeleteRunningWorkspaceTest** test;
- check that Project Explorer is visible before switching to IDE frame;
- don't check workspace STOPPING status before restarting it in **CheckRestoringWorkspaceAfterStoppingWsAgentProcess**.

https://github.com/eclipse/che/issues/9100, https://github.com/eclipse/che/issues/9202

### DIFF
--- a/selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/miscellaneous/CheckRestoringWorkspaceAfterStoppingWsAgentProcess.java
+++ b/selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/miscellaneous/CheckRestoringWorkspaceAfterStoppingWsAgentProcess.java
@@ -11,7 +11,6 @@
 package org.eclipse.che.selenium.miscellaneous;
 
 import static org.eclipse.che.api.core.model.workspace.WorkspaceStatus.RUNNING;
-import static org.eclipse.che.api.core.model.workspace.WorkspaceStatus.STOPPING;
 import static org.eclipse.che.selenium.core.constant.TestCommandsConstants.CUSTOM;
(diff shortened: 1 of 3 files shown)


selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/miscellaneous/CheckRestoringWorkspaceAfterStoppingWsAgentProcess.java
```diff
@@ -71,7 +68,6 @@ public void checkRestoreWsAgentByApi() throws Exception {
 
     notificationsPopupPanel.waitWorkspaceAgentIsNotRunning();
```
dmytro-ndp, line 69: I would check that workspace has status "STOPPED" after `notificationsPopupPanel.waitWorkspaceAgentIsNotRunning();`

dmytro-ndp (approved):

artaleks9 (approved):

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/miscellaneous/CheckRestoringWorkspaceAfterStoppingWsAgentProcess.java
SkorikSergey: Actually workspace has status "RUNNING" even when ws-agent is stopped.

Issue #9125, opened 2018-03-19
Make sure GitHub test repo has been created in selenium tests


dmytro-ndp:
### Description
We are using *org.kohsuke.github* library to create test repository, but doesn't ensure that it is ready consume content. As a result we have sometimes had test failure because of [error](https://ci.codenvycorp.com/view/qa/job/che-integration-tests-master-docker/216/testReport/org.eclipse.che.selenium.git/GitPullTest/pullTest/):
```
java.io.UncheckedIOException: org.kohsuke.github.GHFileNotFoundException: {"message":"Not Found","documentation_url":"https://developer.github.com/v3/repos/contents/#update-a-file"}
	at org.eclipse.che.selenium.core.client.TestGitHubServiceClient.lambda$addContentToRepository$5(TestGitHubServiceClient.java:244)
	at java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)
```

As a workaround I propose to check if new repository is really ready for usage before creating its content.

dmytro-ndp:
Target error [has been reproduced again](https://ci.codenvycorp.com/view/qa/job/che-integration-tests-multiuser-master-docker/153/testReport/junit/org.eclipse.che.selenium.git/GitPullTest/pullTest/):
<details>
 <summary>org.kohsuke.github.GHFileNotFoundException: {"message":"Not Found","documentation_url":"https://developer.github.com/v3/repos/contents/#update-a-file"}</summary>
 <pre>
      java.io.UncheckedIOException: org.kohsuke.github.GHFileNotFoundException: {"message":"Not Found","documentation_url":"https://developer.github.com/v3/repos/contents/#update-a-file"}
	at org.eclipse.che.selenium.core.client.TestGitHubRepository.lambda$addContent$1(TestGitHubRepository.java:91)
	at java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:184)
	at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:175)
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
	at java.util.Iterator.forEachRemaining(Iterator.java:116)
	at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
[...]

PR #9219, opened 2018-03-23
Improve interaction between selenium tests and GitHub

dmytro-ndp:
### What does this PR do?
- removes github repo when selenium test failed; 
- tries to create file on GitHub once again in case of _GHFileNotFoundException_ error.

#9125

### DIFF
--- a/selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/client/TestGitHubRepository.java
+++ b/selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/client/TestGitHubRepository.java
@@ -51,16 +51,15 @@ public class TestGitHubRepository {
 
   private GHRepository create() throws IOException, InterruptedException {
     GHRepository repo = gitHub.createRepository(repoName).create();
-    ensureRepositoryCreated(repo);
+    ensureRepositoryCreated(repo, System.currentTimeMillis());
 
     LOG.info("GitHub repo {} has been created", repo.getHtmlUrl());
     return repo;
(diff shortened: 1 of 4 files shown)


selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/client/TestGitHubRepository.java
```diff
@@ -56,63 +60,80 @@ private GHRepository create() throws IOException, InterruptedException {
...
-        sleepQuietly(1); // sleep one second
+        sleepQuietly(GITHUB_OPERATION_TIMEOUT_SEC); // sleep one second
       }
     }
 
     throw new IOException(
         format(
             "GitHub repo %s hasn't been created in %s seconds",
-            repo.getHtmlUrl(), REPO_CREATION_TIMEOUT_SEC),
+            repo.getHtmlUrl(), (System.currentTimeMillis() - startTime) / 1000),
```
Ohrimenko1988: Could you encapsulate this code part:
```
(System.currentTimeMillis() - startTime) / 1000
```
to the private method for more readibility, something like:
```
int getStartingTime(long startTime)
```

dmytro-ndp: If it is not clear enough what that statement does, I would extract it into separate variable.

Ohrimenko1988 (approved):

artaleks9 (approved):

PR #9143, opened 2018-03-20
Move deployment files out of dockerfiles folder

garagatyi:
### What does this PR do?
Move openshift/k8s deployment files from dockerfiles/init folder
to deploy folder with a better structure because they are not
related to init dockerfile.
Add readme files in each deployment target folder with links to
Che docs where the process of deployment is described.
Remove delivery of deployment files by init image which is
useless.
Remove empty docs files from the repo and init image.
Fix some minor typos and trailing spaces.
Tested on next deployments:
- docker
- kubernetes (deployment file)
- kubernetes (helm)
- openshift (ocp via scripts)

### What issues does this PR fix or reference?
Related to https://github.com/redhat-developer/rh-che/issues/528, https://github.com/eclipse/che/pull/8822, https://github.com/eclipse/che/pull/8982

PR #9230, opened 2018-03-25
Fix K8s Helm Chart

perspectivus1:
### What does this PR do?
Fix broken K8s Helm chart

Helm chart cannot be packaged because of mismatch between directory name and chart name (in Chart.yaml). Cause is movement of deployment files (#9143).


ghost (approved):

riuvshin (approved):

sleshchenko (approved):

garagatyi (approved):
Sorry for breaking Helm installation. 
BTW I checked it using instruction in the readme before the merge. @perspectivus1 can you help me figuring out what I had to do to find out that something was broken?

perspectivus1:
No problem @garagatyi.

The Helm installation still works, if performing it directly from the Git repo. However, the preferred method for installing a Helm chart is by referencing a Helm repository. The prerequisite for uploading a Helm chart to a Helm repo is packing it into a .tgz file using ```helm package```. This command will fail if the directory name is different from the chart's name.

We maintain an internal Jenkins job that run periodically, clones the eclipse/che git repo, packages the Helm chart, and then uploads it to a our own Helm repository. This job failed after the change.

l0rd:
@perspectivus1 why `eclipse-che` and not `che` as the name of the chart? I think I saw some eclipse projects in the helm charts repository that are not prefixed with `eclispe`.

perspectivus1:
@l0rd I named the chart ```eclipse-che``` so it would be easy to find when exploring a Helm repo.

Helm chart naming conventions are [here](https://github.com/kubernetes/helm/blob/master/docs/chart_best_practices/conventions.md#chart-names). There are no naming conventions for charts in Helm repos, other than uniqueness.

I can rename it to ```che``` if you think it's better.

l0rd:
@perspectivus1 I understand you point. I personally prefer `che` to avoid confusions with the eclipse desktop IDE but I don't want to take this decision. 

@slemeur @gorkem how should we call Che in an kubernetes application catalog like [kubeapps](https://hub.kubeapps.com/charts): `che` or `eclipse-che`?

perspectivus1:
@l0rd, I followed your advice and changed the chart name to ```che```.

l0rd:
@perspectivus1 thank you. If you are ok I would merge this PR right away.

l0rd (approved):

perspectivus1:
Thanks @l0rd. Go ahead with the merge.

Issue #9015, opened 2018-03-05
Provide some meaningful error message during user manipulation operation in case if Che is configured to work with Keycloak


skabashnyuk:
### Description
When Che is configured to work with Keycloak user can be created only on the first user connection to che from information provide by keycloak's JWT token. Also, some part of profile data is also taken from Keycloak.   That means that some operation like user creation, removal, profile update has no sense and has to be made in keycloak first.

Related issue https://github.com/eclipse/che/issues/9009

PR #9237, opened 2018-03-26
 Set correct response for non-working services in Multi User Che

mkuznyetsov:
### What does this PR do?
Add Filter, that will be mapped for resources, that are not meant to be available in Multi User Che.
Filter will return 403 status and appropriate message

https://github.com/eclipse/che/issues/9015


multiuser/keycloak/che-multiuser-keycloak-server/src/test/java/org/eclipse/che/multiuser/keycloak/server/UnavailableReourceInMultiUserFilterTest.java
```diff
@@ -0,0 +1,113 @@
...
+  public void shouldReturnForbiddenResponseForUserCreation() {
+
+    final Response response =
+        given()
+            .auth()
+            .basic(ADMIN_USER_NAME, ADMIN_USER_PASSWORD)
+            .when()
+            .post("/user/");
+
+    assertEquals(response.getStatusCode(), 403);
```
sleshchenko: Please add also few tests that UserService and ProfileService GET methods are available.

multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/UnavailableResourceInMultiUserFilter.java
```diff
@@ -0,0 +1,54 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * Filter that will return HTTP status 403. Used for resources, that are not meant to be available
+ * in multi-user Che. Filter omits GET requests.
+ */
+@Singleton
+public class UnavailableResourceInMultiUserFilter implements Filter {
+  protected static final String ERROR_RESPONSE_JSON_MESSAGE =
+      "{\"error\" : \"This resource is unavailable in multi-user Che\" }";
```
sleshchenko: I think the cause why it's forbidden is a Keycloak but not multiuser =) 
So, I think it makes sense to rewrite this message to make it clear that user is able to perform users management via Keycloak dashboard?

mkuznyetsov:
ci-test

multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/UnavailableResourceInMultiUserFilter.java
mkuznyetsov: ok

multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/UnavailableResourceInMultiUserFilter.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Filter that will return HTTP status 403. Used for resources, that are not meant to be available
+ * in multi-user Che. Filter omits GET requests.
+ */
+@Singleton
+public class UnavailableResourceInMultiUserFilter implements Filter {
+  protected static final String ERROR_RESPONSE_JSON_MESSAGE =
+      "{\"error\" : \"This resource is disabled."
+          + " User and Profile management operations are done through Keycloak\" }";
```
skabashnyuk: ```This operation is not allowed since third-party user management service is configured```

sleshchenko (approved):

skabashnyuk (approved):

mkuznyetsov:
ci-test

Issue #9130, opened 2018-03-19
Make it possible to specify non-default timeout for building docker image


skryzhny:
Now we have no ability to specify non-default timeout for building docker image from dockerfile.
It would be very useful to set this timeout in `che.env` to larger or smaller value depending of user's need.

**OS and version:**    Che 6.3.0

PR #9259, opened 2018-03-27
Provide configuration property to be able to set docker image build timeout

akorneta:
### What does this PR do?
Provides configuration property which allows setting a limitation for docker image build timeout.
property:
`che.infra.docker.build_timeout_sec`
default value(experimental):
`480 seconds`

#9130

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
@@ -275,6 +275,9 @@ che.infra.docker.bootstrapper.server_check_period_sec=3
 # in parallel on workspace startups.
 che.infra.docker.max_pull_threads=10
 
+# Time(in seconds) that limits the docker build process.
+# The default value is 8 minutes, after which the build will be considered as failed.
+che.infra.docker.build_timeout_sec=480
 
 # Single port mode
 che.single.port=false
--- a/infrastructures/docker/docker-client/src/main/java/org/eclipse/che/infrastructure/docker/client/DockerConnector.java
+++ b/infrastructures/docker/docker-client/src/main/java/org/eclipse/che/infrastructure/docker/client/DockerConnector.java
@@ -12,6 +12,7 @@ package org.eclipse.che.infrastructure.docker.client;
 
 import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.net.UrlEscapers.urlPathSegmentEscaper;
+import static java.lang.String.format;
 import static javax.ws.rs.core.Response.Status.CREATED;
 import static javax.ws.rs.core.Response.Status.NOT_FOUND;
 import static javax.ws.rs.core.Response.Status.NOT_MODIFIED;
@@ -43,7 +44,10 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 import javax.inject.Inject;
+import javax.inject.Named;
 import javax.inject.Singleton;
 import javax.ws.rs.core.MediaType;
 import org.eclipse.che.api.core.util.FileCleaner;
@@ -139,6 +143,7 @@ public class DockerConnector {
           .setFieldNamingPolicy(FieldNamingPolicy.UPPER_CAMEL_CASE)
           .create();
 
+  private final long imageBuildTimeoutSec;
   private final URI dockerDaemonUri;
   private final DockerRegistryAuthResolver authResolver;
   private final ExecutorService executor;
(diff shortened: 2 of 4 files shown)


infrastructures/docker/docker-client/src/test/java/org/eclipse/che/infrastructure/docker/client/DockerConnectorTest.java
```diff
@@ -1341,6 +1332,34 @@ public void shouldThrowIOExceptionWhenBuildImageButNoSuccessMessageInResponse()
     verify(dockerResponse).getInputStream();
   }
 
+  @Test(
+    expectedExceptions = DockerException.class,
+    expectedExceptionsMessageRegExp = "Docker image build exceed timeout.*"
```
sleshchenko: Would it better to rewrite it in the following way "Docker image build exceed timeout .* seconds." ?

infrastructures/docker/docker-client/src/main/java/org/eclipse/che/infrastructure/docker/client/DockerConnector.java
```diff
@@ -900,7 +907,7 @@ private String buildImage(
                       "Docker image build failed. Image id not found in build output.", 500);
                 });
 
-        return imageIdFuture.get();
+        return imageIdFuture.get(imageBuildTimeoutSec, TimeUnit.SECONDS);
```
sleshchenko, line 910: Do we interrupt Thread which builds images? Will be building really interrupted?

riuvshin: :O good point, @akorneta are you sure that docker build process will be interrupted?

garagatyi: @sleshchenko Do you know a better approach? 
AFAIK from recent versions of Docker, it is a way to perform interruption of the build.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
```diff
@@ -267,6 +267,9 @@ che.infra.docker.bootstrapper.server_check_period_sec=3
 # in parallel on workspace startups.
 che.infra.docker.max_pull_threads=10
 
+# Time(in seconds) that limits the docker build process.
+# The default value is 8 minutes, after which the build will be considered as failed.
+che.infra.docker.build_timeout_sec=480
```
garagatyi, line 280: How much time does it take to build dockerfile of Che master in some good case?

infrastructures/docker/docker-client/src/main/java/org/eclipse/che/infrastructure/docker/client/DockerConnector.java
sleshchenko: @garagatyi No, I don't. I just ask if it will solve the original issue :smiley_cat:

garagatyi, line 910: Can we add an ability to disable timeout? It might be useful for some users.

garagatyi (commented):
Looks good!
Can you take a look at my inlined comment?

infrastructures/docker/docker-client/src/main/java/org/eclipse/che/infrastructure/docker/client/DockerConnector.java
garagatyi: AFAIK Docker CLI does the same

sleshchenko: Ok, thanks for explanation

akorneta: @sleshchenko , @riuvshin so if you carefully look at the implementation of docker image build you will see that in the case when exceptions occur in main(waiting) thread there no any additional work for interruption of building thread the explanation why is placed [here](https://github.com/eclipse/che/blob/19465ec962c4ec4676a1c7aa7d39393c3e24f4a2/infrastructures/docker/docker-client/src/main/java/org/eclipse/che/infrastructure/docker/client/DockerConnector.java#L881-L887).

sleshchenko (approved):

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
akorneta: in master, we do not have any stacks that contain workspace config with dockerfile recipe type but form my custom workspace config docker image build does not take more than 3 minutes. The build process depends on many factors so it is hard to estimate the default value, any suggestions are welcome.

garagatyi (approved):

skabashnyuk (approved):

akorneta:
ci-test

akorneta:
ci-test

PR #9264, opened 2018-03-28
Add ability to configure Theia port

mmorhun:
### What does this PR do?
Adds ability to configure port on which Theia should be run.
To configure port set environment variable `THEIA_PORT`. If it is not configured or invalid default 3000 port will be used.

### DIFF
--- a/dockerfiles/theia/src/main.js
+++ b/dockerfiles/theia/src/main.js
@@ -16,6 +16,8 @@ const theiaRoot = '/home/theia';
 const defaultTheiaRoot = '/home/default/theia';
 const gitPluginsRoot = '/tmp/theia-plugins-from-git-repos';
 
+const THEIA_PORT = getTheiaPort();
+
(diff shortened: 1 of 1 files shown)


dockerfiles/theia/src/main.js
```diff
@@ -188,11 +190,23 @@ function callBuild() {
...
+/**
+ * Returns port on which Theia should be run.
+ * If invalid or not specified, then default 3000 port will be returned.
+ */
+function getTheiaPort() {
+    const port = Number(process.env.THEIA_PORT);
+    if (port) {
+        return port;
+    }
+    return 3000;
```
benoitf: maybe add on top of the file 
```typescript
const DEFAULT_THEIA_PORT = 3000;
``` and use the constant here

benoitf (approved):

AndrienkoAleksandr (approved):

Issue #8859, opened 2018-02-21
Consider using user Id instead of user name in the workspace `RuntimeIdentity`


davidfestal:
### Description

Following the work done in PR #8805, and as a follow-up of the review discussion that can be found [here](https://github.com/eclipse/che/pull/8805#pullrequestreview-97104680), it was decided that the `RuntimeIdentity` class should now store only the `userId`, instead of having both `userId` and `userName` (which will be a result of PR #8805), knowing that initially only the `userName` was available from the `RuntimeIdentity.getOwner()` method.

PR #9274, opened 2018-03-28
Use the owner Id only in the workspace `RuntimeIdentity` (#8859)

davidfestal:
### What does this PR do?

This PR removes the `ownerName` attribute of the `RuntimeIdentity` class, to keep only the `ownerId` according to what was requested in [the following discussion](https://github.com/eclipse/che/pull/8805#pullrequestreview-97104680).
This is done while taking care of still using the owner name when it was previously used (in the container name, etc...)

All the tests pass, and the Che server + workspace start works correctly on Openshift DevCluster.

This fixes issue https://github.com/eclipse/che/issues/8859

#### Release Notes

Nothing to add in the release notes, since it's internal refactoring

#### Docs PR

Nothing to add in the docs, since it's internal refactoring.

### DIFF
--- a/infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/DockerEnvironmentNormalizer.java
+++ b/infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/DockerEnvironmentNormalizer.java
@@ -16,11 +16,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import javax.inject.Inject;
-import org.eclipse.che.api.core.NotFoundException;
-import org.eclipse.che.api.core.ServerException;
-import org.eclipse.che.api.core.model.user.User;
 import org.eclipse.che.api.core.model.workspace.runtime.RuntimeIdentity;
-import org.eclipse.che.api.user.server.spi.UserDao;
 import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
 import org.eclipse.che.workspace.infrastructure.docker.container.ContainerNameGenerator;
 import org.eclipse.che.workspace.infrastructure.docker.model.DockerContainerConfig;
@@ -29,13 +25,10 @@ import org.eclipse.che.workspace.infrastructure.docker.model.DockerEnvironment;
 /** @author Alexander Garagatyi */
 public class DockerEnvironmentNormalizer {
   private final ContainerNameGenerator containerNameGenerator;
-  private final UserDao userDao;
 
   @Inject
-  public DockerEnvironmentNormalizer(
-      ContainerNameGenerator containerNameGenerator, UserDao userDao) {
+  public DockerEnvironmentNormalizer(ContainerNameGenerator containerNameGenerator) {
     this.containerNameGenerator = containerNameGenerator;
-    this.userDao = userDao;
   }
 
   public void normalize(DockerEnvironment dockerEnvironment, RuntimeIdentity identity)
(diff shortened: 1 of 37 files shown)


davidfestal:
ci-test

l0rd (approved):

garagatyi (commented):
I think we should not add UserDao and instead should use user ID as owner value.

@skabashnyuk @sleshchenko WDYT?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/InternalRuntime.java
```diff
@@ -60,7 +70,12 @@ public String getActiveEnv() {
 
   @Override
   public String getOwner() {
-    return context.getIdentity().getOwnerName();
+    try {
+      User user = userDao.getById(context.getIdentity().getOwnerId());
```
sleshchenko: I don't think it is good to return owner name or owner id here. Let's document that owner id will be returned in `Runtime` class and just return it.

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/DockerEnvironmentNormalizer.java
```diff
@@ -40,11 +47,20 @@ public void normalize(DockerEnvironment dockerEnvironment, RuntimeIdentity ident
...
+      } catch (NotFoundException | ServerException e) {
+        userName = identity.getOwnerId();
+      }
+
       containerConfig.setContainerName(
           containerNameGenerator.generateContainerName(
               identity.getWorkspaceId(),
               containerConfig.getId(),
-              identity.getOwnerName(),
+              userName,
```
sleshchenko: I'd say that using owner id here is even better. @garagatyi WDYT?

davidfestal: so the container naming strategy will change ? Doesn't it have any impact on UI for the end-user ?

sleshchenko: I think no. The only one person who sees it is Che Admin using docker CLI. @garagatyi Am I right?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/InternalRuntime.java
sleshchenko: > Maybe we could just throw an exception instead.

You can try to do it but I think it might be difficult with current classes hierarchy

> Change the meaning of value from a plain user name to an ID (opaque to the user), in a location that is used in many places (even in the RuntimeDto).

Makes sense. But I think using username there is a mistake that was done because of docs lack.
@gazarenkov should be right person who knows what is supposed to be there.

If user id then I think we should try to change it and run QA tests. I guess clients don't use it and everything will be OK :wink:

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/DockerEnvironmentNormalizer.java
garagatyi: I agree that container name is the only place. And it doesn't make much sense to keep username there if it costs us something.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/InternalRuntime.java
garagatyi: I think we agreed in your previous PR that we should do this change. I don't think that clients use it somehow now. But running tests, indeed, should verify this.

davidfestal:
ci-test

davidfestal:
@sleshchenko @garagatyi I made the 2 changes you requested to remove the use of the `UserDao`. I've ran the full build locally with tests and everything was OK.

Let's see what integration tests say about this change from the user name to the user Id in the runtime definition...

davidfestal:
@sleshchenko @garagatyi In the second build (after `UserDao` removal), there are slightly more selenium tests failing, but I don't see how new failures are obviously related to this change.

Could you have a look please and approve this PR is it's really OK ?

sleshchenko:
I think there are some unstable tests. @eclipse/eclipse-che-qa Could you please check them, maybe it is needed to run selenium tests one more time?

garagatyi (approved):
Looks good

sleshchenko (approved):
I believe that your changes are not the reason of the failed tests. 
LGTM

Issue #9275, opened 2018-03-28
Add an ability to manage Che master logs with ELK stack


garagatyi:
### Description
Che master is a Docker container application on k8s/docker/openshift infrastructures. But it doesn't provide an ability to manage its logs with popular in container world ELK stack or similar. 
Without that maintenance of Che on enterprise-grade installations of k8s/openshift looks odd.
On the other hand, I think that there should be an ability to leave plain text logs too since they are more appropriate for developers of Che itself.
So tasks are:
- Add an ability to encode Che master logs with JSON format
- Add an ability to switch between plain text and JSON formats of logs

### Reproduction Steps

**OS and version:**    

**Diagnostics:**

garagatyi:
@l0rd @skabashnyuk FYI

PR #9296, opened 2018-03-30
Add setting of an environment variable to configure master logs encoding

garagatyi:
### What does this PR do?
Add setting of master logs encoding to che.env.

Related to #9275, eclipse/che-lib#83

#### Docs PR
https://github.com/eclipse/che-docs/pull/384


garagatyi:
@perspectivus1 @guydaichs FYI

dockerfiles/init/manifests/che.env
```diff
@@ -56,6 +56,13 @@
 #CHE_DEBUG_SUSPEND=false
 #CHE_DEBUG_PORT=8000
 
+# Che master logs
+#
+# Choose variant of Che master logs encoding and storing. Default is `plaintext`.
+# Another supported value is `json` which makes Che master produce JSON encoded logs
+# in system output of PID 1 process of container. Useful for storing logs in systems such as Logstash.
+# CHE_LOGS_APPENDERS_IMPL=plaintext
```
sleshchenko, line 64: Shouldn't default value be added into che.properties?

riuvshin, line 64: if that is commented, the default value will be used, do we have default value declared in a code? because I do not see it in the `che.props`

l0rd: I think the default value is declared [here](https://github.com/eclipse/che-lib/pull/83/files#diff-64fb0b389654f324dfc3abfe151c0b74R26)

l0rd (commented):
I think this PR deserves a Doc PR too

garagatyi:
@l0rd here is a PR for the docs https://github.com/eclipse/che-docs/pull/384

dockerfiles/init/manifests/che.env
garagatyi: No, it is not read from properties

garagatyi: @l0rd is right this value is not read by the code and default value is in xml file of logback

garagatyi:
@l0rd @riuvshin @sleshchenko @eivantsov I've merged the PR in che-lib so we can proceed with this one. Can you review the PR?

benoitf (approved):

sleshchenko (approved):
LGTM

l0rd (approved):

Issue #8695, opened 2018-02-09
Manage the list of IDE plugins from dashboard


ashumilova:
Need to add ability to manage particular workspace's "tooling" data in workspace details page. 
For the start, we will add tooling page to workspace details - with defined one - Theia and will add layout for managing it's plugins. The initial list of plugins could be requested from npm registry.

PR #9300, opened 2018-03-30
Add ability to manage the list of Theia plugins in UD

ashumilova:

### What does this PR do?
Adds the ability to manage the list of Theia plugins in UD and add custom plugins. 

https://github.com/eclipse/che/issues/8695

### DIFF
--- a/dashboard/src/app/ide/ide.controller.ts
+++ b/dashboard/src/app/ide/ide.controller.ts
@@ -152,11 +152,15 @@ class IdeCtrl {
       return;
     }
 
-    if (this.ideSvc.openedWorkspace && this.selectedWorkspace.id === this.ideSvc.openedWorkspace.id) {
+    // todo commented for testing purposes - with IDE running inside container - after workspace restart - location changes,
+    // and thus start facing connection issues.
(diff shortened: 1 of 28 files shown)


ashumilova:
ci-test

dashboard/src/components/api/npm-registry.factory.ts
```diff
@@ -0,0 +1,65 @@
+/*
```
benoitf, line 1: I would assume the managing of plugins would be server side instead of client side (like searching all extensions).
So it doesn't use client connection to download plugins but use server side.

But it may be for step 2

dashboard/src/app/ide/ide.controller.ts
```diff
@@ -152,11 +152,15 @@ class IdeCtrl {
       return;
     }
 
-    if (this.ideSvc.openedWorkspace && this.selectedWorkspace.id === this.ideSvc.openedWorkspace.id) {
+    // todo commented for testing purposes - with IDE running inside container - after workspace restart - location changes,
```
benoitf: before merging, need to remove comments or put it back

benoitf (approved):

ashumilova:
ci-test

olexii4 (approved):

akurinnoy (approved):

Issue #9014, opened 2018-03-05
Cluster mode for periodical jobs on ws-master


skabashnyuk:
### Description
 In this issue https://github.com/eclipse/che/issues/8870, we identified some stateful components of ws-master.  The goal if this issue is to make them compatible with cluster mode. One of the ways to do that is to run them only if the current node is a coordinator see more https://docs.jboss.org/infinispan/9.0/apidocs/org/infinispan/manager/EmbeddedCacheManager.html#isCoordinator--
or use http://infinispan.org/docs/stable/user_guide/user_guide.html#single_node_submission

skabashnyuk:
out of date

PR #9311, opened 2018-04-02
Add replicated storage for remote websocket subscriptions

mshaposhnik:
### What does this PR do?
Add JGroups-based replicated storage for remote websocket subscriptions.

#9014 (partially)

#### Release Notes
Added JGroups-based replicated storage for remote websocket subscriptions.

#### Docs PR
N/A


skabashnyuk (commented):
In general Ok. @riuvshin wdyt?

mshaposhnik:
ci-test

mshaposhnik:
ci-test

dkulieshov (approved):
I see no unit tests, is that intentional?

multiuser/api/che-multiuser-api-remote-subscription/src/main/java/org/eclipse/che/multiuser/api/subscription/DistributedRemoteSubscriptionStorage.java
```diff
@@ -0,0 +1,93 @@
...
+  private final LockService lockService;
+
+  @Inject
+  public DistributedRemoteSubscriptionStorage(@Named("jgroups.config.file") String confFile)
+      throws Exception {
+    try {
+      JChannel channel = new JChannel(confFile);
+      this.lockService = new LockService(channel);
+      channel.connect(CHANNEL_NAME);
+      subscriptions = new ReplicatedHashMap<>(channel);
```
akorneta, line 51: looks like we are trying to achieve the consistent state of subscribers in each period of time, and so updates should be performed across the cluster synchronously, to achieve that you can use `blocking_updates` property, if we won't need that we can manage local concurrent operations using read/write locks instead of `LockService`.

mshaposhnik: Makes sense, added.

PR #9301, opened 2018-03-30
Implemented recovery functionality for Kubernetes/OpenShift infrastructures

sleshchenko:
### What does this PR do?
Implements recovery functionality for Kubernetes/OpenShift infrastructures.

It's done by usage of persistent cache for storing Kuberentes/OpenShift runtimes states including machine and servers information.

Jpa is used as persistent cache backend. To make it possible to have two Running Che Server (like during Rolling Update) it is needed to synchronize JPA L2 cache with JGroups, it is done in https://github.com/eclipse/che/pull/8982.

### What issues does this PR fix or reference?
https://github.com/eclipse/che/issues/5919

#### Release Notes
Implemented recovery functionality for Kubernetes/OpenShift infrastructures.

#### Docs PR
N/A

Issue #5919, opened 2017-08-07
Implement recovery for Kubernetes/OpenShift infrastructures


sleshchenko:
Kubernetes/OpenShift workspaces are considered as stopped when workspace master is restarted.
It is needed to implement recovery for Kubernetes/OpenShift workspaces, so workspaces will be considered as running after the restart of the master.
Recovering should be adapted to Rolling Update of a workpace master.  So, recovery workflow should look like:
1. Pod CheServer is running and Service redirects to it. There may be running, starting or stopping workspaces.
2. Pod CheServer* is starting. After the start, it should know about all workspaces, it should be possible to interact with running workspace (requests its servers, stop it).
3. Pod CheServer* is running Service routes all traffic to it.
4. CheServer is stopping. It should finish all operation with workspaces (stopping and starting workspaces).
4. CheServer* should know about finished operations and pick up RUNNING workspaces.

sleshchenko:
Depends on https://github.com/eclipse/che/issues/7785

sleshchenko:
Today we have the following status on this issue:

Looks like there is no an ability restore all running workspaces when tomcat is booted using only Kubernetes/OpenShift client and checking create objects on a cluster like a recovery is implemented in Docker Infrastructure.

Another proposed way to recovery workspaces was implementing lazy recovery when each workspace will be recovered only when a workspace is requested by a client. In this case request for getting workspaces list (GET /api/workspace) would initiate several requests to K8s/OS cluster and it would increase response time. Because of that, it was decided not to implement it.

So, it's needed to persist somewhere (like a database) metainformation of running workspaces for further recovery of them.

Also, the scope of this issue was extended and it is required to make Kubernetes/OpenShift infrastructure ready for Rolling Update (Issue description is updated). In this case, recovery should be implemented in the following way:
1. Pod CheServer is running and Service redirects to it. There may be running, starting or stopping workspaces.
2. Pod CheServer* is starting. After the start, it should know about all workspaces, it should be possible to interact with running workspace (requests its servers, stop it).
3. Pod CheServer* is running Service routes all traffic to it.
4. CheServer is stopping. It should finish all operation with workspaces (stopping and starting workspaces).
4. CheServer* should know about finished operations and pick up RUNNING workspaces.

Since there may be two running Che Server instances at the same time, it's not enough to rework infrastructure, because of Workspace API has own local cache. So Workspace API should be reworked to use local/(persistent or distributed) depending on configuration.

[...]

PR #9345, opened 2018-04-05
Improve abstract InternalRuntime to be more flexible for recovery functionality

sleshchenko:
### What does this PR do?
Improve abstract InternalRuntime to be more flexible for recovery functionality. It includes:

- Add an ability to store status by an implementation of InternalRuntime. It allows infrastructure provide an actual status after runtime recovery. Also, it adds an ability to deny stopping of STARTING workspaces if it is not implemented like in Kubernetes/OpenShift infrastructures.
Unfortunately, denying stop of STARTING workspace will not work correctly. Here is a registered bug for that https://github.com/eclipse/che/issues/9354.

- Add an ability to throw InfrastructureException by InternalRuntimes#getMachines method. It is needed because Runtime may store machines information in storage that may throw some exceptions (like database as storage may throw connection related exceptions).

https://github.com/eclipse/che/issues/5919
It's needed for OpenShift recovery functionality changes https://github.com/eclipse/che/pull/9301

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
@@ -144,16 +144,25 @@ public class WorkspaceRuntimes {
    *
    * @param workspace the workspace to inject runtime into
    */
-  public void injectRuntime(WorkspaceImpl workspace) {
+  public void injectRuntime(WorkspaceImpl workspace) throws ServerException {
     RuntimeState runtimeState = runtimes.get(workspace.getId());
     if (runtimeState != null) {
-      workspace.setRuntime(new RuntimeImpl(runtimeState.runtime));
+      try {
+        workspace.setRuntime(asRuntime(runtimeState));
+      } catch (InfrastructureException e) {
+        throw new ServerException(
+            "Error occurred while runtime state inspection. " + e.getMessage());
+      }
       workspace.setStatus(runtimeState.status);
     } else {
       workspace.setStatus(STOPPED);
     }
   }
 
+  private RuntimeImpl asRuntime(RuntimeState runtimeState) throws InfrastructureException {
+    InternalRuntime<?> runtime = runtimeState.runtime;
+    return new RuntimeImpl(runtime.getActiveEnv(), runtime.getMachines(), runtime.getOwner());
+  }
   /**
    * Gets workspace status by its identifier.
    *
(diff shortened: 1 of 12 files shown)


wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -144,16 +144,25 @@ public void validate(Environment environment)
    *
    * @param workspace the workspace to inject runtime into
    */
-  public void injectRuntime(WorkspaceImpl workspace) {
+  public void injectRuntime(WorkspaceImpl workspace) throws ServerException {
```
garagatyi, line 147: I'd rather use InfrastructureException. We have discussed that ServerException if rather REST exception.

akorneta (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/InternalRuntime.java
```diff
@@ -86,9 +112,21 @@ public String getOwner() {
...
-   * <p>Implementation should not return null
+   * <p>Implementation must not return null
+   *
+   * @throws InfrastructureException when any error occurs
+   */
+  protected abstract Map<String, ? extends Machine> getInternalMachines()
+      throws InfrastructureException;
+
+  /**
+   * Returns workspace status.
```
garagatyi: Runtime status or workspace status?

garagatyi (approved):
LGTM

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
sleshchenko: I agree that `ServerException` is REST exception, but `InfrastructureException` is an exception that is supposed to be thrown by Infrastructure. As you can see, now `WorkspaceRuntimes` throw only REST exceptions (Conflict, NotFound, Server).
I'd say that WorkspaceRuntimes should not throw REST exception but and Infrastructure exception isn't the right one here too. I guess there is an issue for reworking Che Master exception approach and this issue could be solved there.
@garagatyi Does it makes sense?

sleshchenko:
ci-test

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
garagatyi: Yep, you are completely right. I missed that.

sleshchenko:
ci-test

skabashnyuk (approved):

sleshchenko:
ci-test

PR #9350, opened 2018-04-06
Selenium: not check Workspace API installer changeability in WorkspaceDetailsSingleMachineTest test

SkorikSergey:
### What does this PR do?
This PR removes checking that **Workspace API** installer is not changeable because for now all installer can be disabled.

selection_225


dmytro-ndp (commented):
Do we need to hold **isInstallerStateNotChangeable()** method at all?

SkorikSergey:
As all installers can be disabled this method is redundant.

dmytro-ndp (approved):

PR #9363, opened 2018-04-10
Update content from the resource for CheckGeneratingMavenArchetypeTest file to actual state

musienko-maxim:
### What does this PR do?
The CheckGeneratingMavenArchetypeTest generate the maven-archetype-quickstart project from the wizard. After that, the test checks that artifact, snapshot and content of the pom.xml have been created properly. But default content  pom.xml of maven-archetype-quickstart project has been changed on maven central repo. side. We should also update this one in the resources for correct checking

https://github.com/eclipse/che/issues/9365


dmytro-ndp (commented):
I think it worth to add comment into the test with suggestion to check on maven archetype version in case of the same error occur.

musienko-maxim:
So i've just removed the resource file with a big part of content for checking. In this file was written full content from pom.xml file of the maven-archetype-quickstart project. As we clarified this content may be changed on Maven central repo and we cannot control this. I've changed the test this way - we generate unique artifact id and grup id of the pom.xml each run, and check this part in the test.

vparfonov (approved):

dmytro-ndp (approved):

musienko-maxim:
ci-build

Issue #9397, opened 2018-04-12
Exported folder packs with absolute paths


xcariba:
### Description
When I try to download folder as Zip archive Che zips it with absolute paths.
I expect it to zip with paths, relative to selected folder.

### Reproduction Steps
Download folder as zip
![image](https://user-images.githubusercontent.com/7170514/38682531-fd09a79c-3e73-11e8-8d66-e12f2153ab82.png)
It contains full path in zip
![image](https://user-images.githubusercontent.com/7170514/38682647-526851ac-3e74-11e8-9f59-deafe0bdfae0.png)

**OS and version:**    
Fedora 27
Che 6.3.0
**Diagnostics:**

dkuleshov:
Hi,

This issue should be fixed by the following pull request: https://github.com/eclipse/che/pull/9400

eivantsov:
Closing the issue in favor of https://github.com/eclipse/che/pull/9400

PR #9400, opened 2018-04-12
Changed absolute paths to relative paths for zipped content

dkulieshov:
Relative issue: https://github.com/eclipse/che/issues/9397

Changed absolute paths to relative paths for zipped content

### DIFF
--- a/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/fs/server/impl/ZipArchiver.java
+++ b/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/fs/server/impl/ZipArchiver.java
@@ -64,12 +64,7 @@ class ZipArchiver {
 
       try (FileOutputStream fos = new FileOutputStream(outFile);
           ZipOutputStream zos = new ZipOutputStream(fos)) {
-        Path parent = fsPath.getParent();
-        if (root.startsWith(parent)) {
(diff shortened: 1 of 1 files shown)


dkuleshov:
ci-build

vparfonov (approved):

dkuleshov:
ci-build

dkuleshov:
ci-test

xcariba:
Project import does not work if zip file does not contain folders, only paths.
Example (zipinfo): 
Does not work:
```
src/main/Example.txt
```
Does work:
```
src/
src/main/
src/main/Example.txt
```
ZipArchiver fails with FileNotFound exception.

xcariba:
This change adds extra folder in zip archive. (ex. in 5.21 if I download folder "src" zip archive would contain only contents of the folder. After this changes zip archive contains "src" folder and it`s contents inside.
This is probably due to fsManager API, zip method has no "skip root" parameter.

dkuleshov:
ci-test

musienko-maxim:
The selenium tests passed without new regressions.
Also was checked next features manually: 
* download folder with files by context menu with check saving relative paths
* download projects with menu Workspace -> Download as Zip
* download a project with menu Project -> Download as Zip
We have not created autotests for checking this functionality. But i create issue for current sprint for developing tests:

Issue #8122, opened 2018-01-02
Unexpected fail of CreateNewPackagesWithHelpCreationJavaClassTest selenium test


Ohrimenko1988:
We have unexpected fail of CreateNewPackagesWithHelpCreationJavaClassTest selenium test. It looks like sometimes, after creating java class with a package (for example "org.example.Example.java"), the newly created package is not expanded and inner java class is not displayed.

**Screenshot:**

![hhhh](https://user-images.githubusercontent.com/22724619/34488344-cdbbae9a-efe0-11e7-9801-970e255070ac.png)

**Report with reproduced problem:**

https://ci.codenvycorp.com/view/qa/job/che-integration-tests-multiuser-master-docker/73/Selenium_tests_report/

**Screencast:**

![out-2](https://user-images.githubusercontent.com/22724619/37275242-4b80bf8c-25e7-11e8-97b9-e621b8e2ed90.gif)

Ohrimenko1988:
The bug is actual again 
**Report with reproduced problem:**
https://ci.codenvycorp.com/view/qa/job/che-integration-tests-multiuser-master-ocp/233/Selenium_tests_report/

PR #9417, opened 2018-04-13
CHE-8122: Expand added package if it contains child nodes

vinokurig:
### What does this PR do?
Expand package node in Project Explorer if it contains child node.
We have `TreeResourceRevealer` that handle's nodes loading, witch can expand such nodes, but it catches event with children nodes that received from local storage. It means that children can be not actual and added package can be not expanded in this case.
This handler expands added package with child node anyway.

#8122 

#### Release Notes

Fix bug when Project Explorer doesn't expand new Java class with packages e.g. folder.nestedFolder.Class.java

#### Docs PR

N/A


vzhukovs (approved):

vparfonov (approved):

vinokurig:
ci-test

vinokurig:
ci-test

Issue #7653, opened 2017-11-30
Alternative implementation of abnormal machine stop and container events propagation


akorneta:
### Description
Since we have changed the usage of `OpenShiftClient` from one instance per action to one instance per Che server in the scope of [PR](https://github.com/eclipse/che/pull/7629), we have one connection pool which has limits for a number of idled connections (it equals to 5 connections by default but it can be overridden). Based on this we have no way to watch pod actions and container events for each workspace with using 2 separate WebSocket connections because at some point we overload the connection pool. 
It is possible to use one websocket connection and listen to events of all workspaces if the Che server is configured in the following ways:
- If Che server is configured to use system admin credentials (username/password or token);
- If Che server is configured to use a service account.

But there is no clear way how to implement this when Che server uses users' token (like OSIO scenario) or credentials of a user which doesn't have system admin rights.
So it is needed to investigate possible ways how to listen to OpenShift events which will work for all configuration combinations.

PR #9426, opened 2018-04-13
rh-che #557: Adding property and handler for stopping k8s / openshift workspace if unrecoverable event occurs during workspace startup

ibuziuk:
### What does this PR do?
Adding property and handler for stopping k8s / openshift workspace if unrecoverable event occurs during workspace startup

Demo - https://youtu.be/J_DfNZXidE0

https://github.com/redhat-developer/rh-che/issues/557

Adding property `che.infra.kubernetes.workspace_unrecoverable_events`

if during workspace startup an unrecoverable event occurs (e.g. "Failed Mount" "Failed Scheduling") terminate workspace immediately instead of waiting util timeout

#### Release Notes
Adding property `che.infra.kubernetes.workspace_unrecoverable_events`

if during workspace startup an unrecoverable event occurs (e.g. "Failed Mount" "Failed Scheduling") terminate workspace immediately instead of waiting util timeout

#### Docs PR
https://github.com/eclipse/che-docs/pull/395

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
@@ -353,6 +353,10 @@ che.infra.kubernetes.workspace_start_timeout_min=8
 # Defines the timeout in minutes that limits the period for which Kubernetes Ingress become ready
 che.infra.kubernetes.ingress_start_timeout_min=5
 
+# If during workspace startup an unrecoverable event defined in the property occurs,
+# terminate workspace immediately instead of waiting until timeout
+che.infra.kubernetes.workspace_unrecoverable_events=Failed Mount,Failed Scheduling,Failed to pull image
+
 che.infra.kubernetes.bootstrapper.binary_url=http://${CHE_HOST}:${CHE_PORT}/agent-binaries/linux_amd64/bootstrapper/bootstrapper
 che.infra.kubernetes.bootstrapper.installer_timeout_sec=180
 che.infra.kubernetes.bootstrapper.server_check_period_sec=3
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
@@ -13,6 +13,7 @@ package org.eclipse.che.workspace.infrastructure.kubernetes;
 import static java.lang.String.format;
 import static java.util.Collections.emptyMap;
 
+import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMap;
 import com.google.inject.assistedinject.Assisted;
 import io.fabric8.kubernetes.api.model.Container;
@@ -22,7 +23,9 @@ import io.fabric8.kubernetes.api.model.Service;
 import io.fabric8.kubernetes.api.model.extensions.Ingress;
 import io.fabric8.kubernetes.client.Watcher.Action;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -89,6 +92,7 @@ public class KubernetesInternalRuntime<
 
   private final int workspaceStartTimeout;
   private final int ingressStartTimeout;
+  private final String unrecoverableEvents;
   private final ServersCheckerFactory serverCheckerFactory;
   private final KubernetesBootstrapperFactory bootstrapperFactory;
   private final ProbeScheduler probeScheduler;
@@ -104,6 +108,7 @@ public class KubernetesInternalRuntime<
   public KubernetesInternalRuntime(
       @Named("che.infra.kubernetes.workspace_start_timeout_min") int workspaceStartTimeout,
       @Named("che.infra.kubernetes.ingress_start_timeout_min") int ingressStartTimeout,
+      @Named("che.infra.kubernetes.workspace_unrecoverable_events") String unrecoverableEvents,
       NoOpURLRewriter urlRewriter,
       KubernetesBootstrapperFactory bootstrapperFactory,
       ServersCheckerFactory serverCheckerFactory,
@@ -123,6 +128,7 @@ public class KubernetesInternalRuntime<
     this.volumesStrategy = volumesStrategy;
     this.workspaceStartTimeout = workspaceStartTimeout;
     this.ingressStartTimeout = ingressStartTimeout;
+    this.unrecoverableEvents = unrecoverableEvents;
     this.probeScheduler = probeScheduler;
     this.probesFactory = probesFactory;
     this.namespace = namespace;
@@ -183,6 +189,8 @@ public class KubernetesInternalRuntime<
         throw new InfrastructureException("Kubernetes environment start was interrupted");
       }
       wrapAndRethrow(e);
+    } finally {
+      namespace.pods().stopWatch();
     }
   }
 
@@ -418,7 +426,10 @@ public class KubernetesInternalRuntime<
 
     // TODO https://github.com/eclipse/che/issues/7653
     // namespace.pods().watch(new AbnormalStopHandler());
-    // namespace.pods().watchContainers(new MachineLogsPublisher());
+    namespace.pods().watchContainers(new MachineLogsPublisher());
+    if (!Strings.isNullOrEmpty(unrecoverableEvents)) {
+      namespace.pods().watchContainers(new UnrecoverableEventHanler());
+    }
 
     final KubernetesServerResolver serverResolver =
         new KubernetesServerResolver(createdServices, readyIngresses);
@@ -600,8 +611,69 @@ public class KubernetesInternalRuntime<
     }
   }
 
+  /** Listens container's events and terminates workspace if unrecoverable event occurs. */
+  public class UnrecoverableEventHanler implements ContainerEventHandler {
+    private List<String> events;
+
+    public UnrecoverableEventHanler() {
+      this.events =
+          Strings.isNullOrEmpty(unrecoverableEvents)
+              ? Collections.EMPTY_LIST
+              : Arrays.asList(unrecoverableEvents.split(","));
+    }
+
+    @Override
+    public void handle(ContainerEvent event) {
+      if (isUnrecoverable(event)) {
+        String reason = event.getReason();
+        String message = event.getMessage();
+        String podName = event.getPodName();
+        try {
+          internalStop(emptyMap());
+        } catch (InfrastructureException e) {
+          String workspaceId = getContext().getIdentity().getWorkspaceId();
+          LOG.error(
+              "Unrecoverable event occured during workspace '{}' startup: {}, {}, {}",
+              workspaceId,
+              reason,
+              message,
+              podName);
+        } finally {
+          eventPublisher.sendRuntimeStoppedEvent(
+              format("Unrecoverable event occured: '%s', '%s', '%s'", reason, message, podName),
+              getContext().getIdentity());
+        }
+      }
+    }
+
+    /**
+     * @param ContainerEvent
+     * @return true if event reason or message matches one of the comma separated values defined in
+     *     'che.infra.kubernetes.workspace_unrecoverable_events',false otherwise
+     */
+    private boolean isUnrecoverable(ContainerEvent event) {
+      boolean isUnrecoverable = false;
+      String reason = event.getReason();
+      String message = event.getMessage();
+      // Consider unrecoverable if event reason 'equals' one of the property values e.g. "Failed
+      // Mount"
+      if (events.contains(reason)) {
+        isUnrecoverable = true;
+      } else {
+        for (String e : events) {
+          // Consider unrecoverable if event message 'startsWith' one of the property values e.g.
+          // "Failed to pull image"
+          if (message != null && message.startsWith(e)) {
+            isUnrecoverable = true;
+          }
+        }
+      }
+      return isUnrecoverable;
+    }
+  }
+
   /** Listens container's events and publish them as machine logs. */
-  class MachineLogsPublisher implements ContainerEventHandler {
+  public class MachineLogsPublisher implements ContainerEventHandler {
 
     @Override
     public void handle(ContainerEvent event) {
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/event/ContainerEvent.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/event/ContainerEvent.java
@@ -21,12 +21,15 @@ import java.util.Objects;
 public class ContainerEvent {
   private final String podName;
   private final String containerName;
+  private final String reason;
   private final String message;
   private final String time;
 
-  public ContainerEvent(String podName, String containerName, String message, String time) {
+  public ContainerEvent(
+      String podName, String containerName, String reason, String message, String time) {
     this.podName = podName;
     this.containerName = containerName;
+    this.reason = reason;
     this.message = message;
     this.time = time;
   }
@@ -39,6 +42,11 @@ public class ContainerEvent {
   /** Returns container name which produced event. */
   public String getContainerName() {
     return containerName;
+  }
+
+  /** Returns the reason of the event. */
+  public String getReason() {
+    return reason;
   }
 
   /** Returns the contents of the event. */
@@ -62,13 +70,14 @@ public class ContainerEvent {
     ContainerEvent that = (ContainerEvent) o;
     return Objects.equals(podName, that.podName)
         && Objects.equals(containerName, that.containerName)
+        && Objects.equals(reason, that.reason)
         && Objects.equals(message, that.message)
         && Objects.equals(time, that.time);
   }
 
   @Override
   public int hashCode() {
-    return Objects.hash(podName, containerName, message, time);
+    return Objects.hash(podName, containerName, reason, message, time);
   }
 
   @Override
@@ -79,6 +88,9 @@ public class ContainerEvent {
         + '\''
         + ", containerName='"
         + containerName
+        + '\''
+        + ", reason='"
+        + reason
         + '\''
         + ", message='"
         + message
--- a/infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntimeTest.java
+++ b/infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntimeTest.java
@@ -92,6 +92,7 @@ import org.eclipse.che.api.workspace.server.spi.environment.InternalMachineConfi
 import org.eclipse.che.api.workspace.shared.dto.event.MachineLogEvent;
 import org.eclipse.che.api.workspace.shared.dto.event.MachineStatusEvent;
 import org.eclipse.che.workspace.infrastructure.kubernetes.KubernetesInternalRuntime.MachineLogsPublisher;
+import org.eclipse.che.workspace.infrastructure.kubernetes.KubernetesInternalRuntime.UnrecoverableEventHanler;
 import org.eclipse.che.workspace.infrastructure.kubernetes.bootstrapper.KubernetesBootstrapper;
 import org.eclipse.che.workspace.infrastructure.kubernetes.bootstrapper.KubernetesBootstrapperFactory;
 import org.eclipse.che.workspace.infrastructure.kubernetes.cache.KubernetesMachineCache;
@@ -142,6 +143,10 @@ public class KubernetesInternalRuntimeTest {
   private static final String M1_NAME = POD_NAME + '/' + CONTAINER_NAME_1;
   private static final String M2_NAME = POD_NAME + '/' + CONTAINER_NAME_2;
 
+  private static final String EMPTY_UNRECOVERABLE_EVENTS = "";
+  private static final String UNRECOVERABLE_EVENTS =
+      "Failed Mount,Failed Scheduling,Failed to pull image";
+
   private static final RuntimeIdentity IDENTITY =
       new RuntimeIdentityImpl(WORKSPACE_ID, "env1", "id1");
 
@@ -169,6 +174,9 @@ public class KubernetesInternalRuntimeTest {
   private KubernetesInternalRuntime<KubernetesRuntimeContext<KubernetesEnvironment>>
       internalRuntime;
 
+  private KubernetesInternalRuntime<KubernetesRuntimeContext<KubernetesEnvironment>>
+      internalRuntimeWithoutUnrecoverableHandler;
+
   @BeforeMethod
   public void setup() throws Exception {
     MockitoAnnotations.initMocks(this);
@@ -180,6 +188,26 @@ public class KubernetesInternalRuntimeTest {
         new KubernetesInternalRuntime<>(
             13,
             5,
+            UNRECOVERABLE_EVENTS,
+            new URLRewriter.NoOpURLRewriter(),
+            bootstrapperFactory,
+            serverCheckerFactory,
+            volumesStrategy,
+            probesScheduler,
+            workspaceProbesFactory,
+            new RuntimeEventsPublisher(eventService),
+            new KubernetesSharedPool(),
+            runtimeStatesCache,
+            machinesCache,
+            context,
+            namespace,
+            emptyList());
+
+    internalRuntimeWithoutUnrecoverableHandler =
+        new KubernetesInternalRuntime<>(
+            13,
+            5,
+            EMPTY_UNRECOVERABLE_EVENTS,
             new URLRewriter.NoOpURLRewriter(),
             bootstrapperFactory,
             serverCheckerFactory,
@@ -244,6 +272,7 @@ public class KubernetesInternalRuntimeTest {
     verify(pods).create(any());
     verify(ingresses).create(any());
     verify(services).create(any());
+    verify(namespace.pods(), times(2)).watchContainers(any());
     verify(bootstrapper, times(2)).bootstrapAsync();
     verify(eventService, times(4)).publish(any());
     verifyOrderedEventsChains(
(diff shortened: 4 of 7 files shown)


infrastructures/openshift/src/test/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntimeTest.java
```diff
@@ -193,6 +195,7 @@ public void shouldStartMachines() throws Exception {
     verify(routes).create(any());
     verify(services).create(any());
 
+    verify(project.pods(), times(1)).watchContainers(any());
```
garagatyi: Can you also check for the cancellation of watching?

infrastructures/openshift/src/test/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntimeTest.java
```diff
@@ -137,6 +138,7 @@ public void setup() throws Exception {
         new OpenShiftInternalRuntime(
             13,
             5,
+            FAIL_ON_UNRECOVERABLE_EVENT,
```
garagatyi: Can you add a case when the value is false?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
```diff
@@ -420,6 +426,10 @@ protected void startMachines() throws InfrastructureException {
     // namespace.pods().watch(new AbnormalStopHandler());
     // namespace.pods().watchContainers(new MachineLogsPublisher());
 
+    if (failFastOnUnrecoverableEvents) {
+      namespace.pods().watchContainers(new UnrecoverableEventHanler());
```
sleshchenko: I think it could be useful add `MachineLogsPublisher` here. WDYT?

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntimeTest.java
```diff
@@ -180,6 +182,7 @@ public void setup() throws Exception {
         new KubernetesInternalRuntime<>(
             13,
             5,
+            FAIL_ON_UNRECOVERABLE_EVENT,
```
garagatyi: Can you add a case when the value is false?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
```diff
@@ -600,6 +610,26 @@ public void accept(ProbeResult probeResult) {
...
+      if (isUnrecoverable(reason)) {
+        LOG.error("Unrecoverable event occured {}, {}", reason, message);
+        eventPublisher.sendRuntimeStoppedEvent(
+            format("Unrecoverable event occured: '%s', '%s'", reason, message),
+            getContext().getIdentity());
+      }
+    }
+
+    private boolean isUnrecoverable(String reason) {
+      return reason.equals("Failed Mount") || reason.equals("Failed Scheduling");
```
sleshchenko: Maybe it makes sense to make these values configurable?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
```diff
@@ -600,6 +610,26 @@ public void accept(ProbeResult probeResult) {
     }
   }
 
+  /** Listens container's events and terminates runtime if unrecoverable event occurs. */
```
garagatyi: AFAIK we don't have runtime interruption in place for k8s runtime yet. Can you elaborate how this handler stops start of a runtime?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
garagatyi: +1

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
garagatyi: @sleshchenko can you elaborate on what exactly to do?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
ibuziuk: @sleshchenko do you mean property with comma separated reasons ?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/event/ContainerEvent.java
```diff
@@ -80,6 +88,9 @@ public String toString() {
         + ", containerName='"
         + containerName
         + '\''
+        + ", reason='"
```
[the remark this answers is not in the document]
ibuziuk: yeah, good point

infrastructures/openshift/src/test/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntimeTest.java
ibuziuk: @garagatyi do you think it is ok to check one infra with `false` & other with `true`, or you think there should be both cases handled in both infras ?

garagatyi: Since they might have their own bugs we should have tests for both

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
```diff
@@ -600,6 +610,26 @@ public void accept(ProbeResult probeResult) {
...
+  public class UnrecoverableEventHanler implements ContainerEventHandler {
+
+    @Override
+    public void handle(ContainerEvent event) {
+      String reason = event.getReason();
+      String message = event.getMessage();
+      if (isUnrecoverable(reason)) {
+        LOG.error("Unrecoverable event occured {}, {}", reason, message);
```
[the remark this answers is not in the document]
garagatyi: Maybe workspace ID, maybe some pod/container info if any

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
```diff
@@ -240,6 +244,8 @@ private void waitMachines(
       failure.completeExceptionally(ex);
       cancelAll(toCancelFutures);
       wrapAndRethrow(ex.getCause());
+    } finally {
+      namespace.pods().stopWatch();
```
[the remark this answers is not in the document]
akorneta: +1

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
sleshchenko: I think yes.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
sleshchenko: I think it should be something like the following:
```
--> start
try {
  namespace.pods().watchContainers(new MachineLogsPublisher());
  if (failFastOnUnrecoverableEvents) {
    namespace.pods().watchContainers(new UnrecoverableEventHanler());
  }

  //doStart here
} catch ....
... 
} finally {
  namespace.pods().stopWatch();
}
```
In this case clients would receive K8s/OS events text as machine's output, and it would be easier to understand why workspace start fails.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/event/ContainerEvent.java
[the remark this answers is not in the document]
ibuziuk: fixed

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
ibuziuk: @sleshchenko Adding `MachineLogsPublisher` seems to be unrelated to this issue but rather to https://github.com/eclipse/che/issues/7653 IMO. Could this be addressed in another PR ?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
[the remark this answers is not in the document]
ibuziuk: have added pod info

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
ibuziuk: s/runtime/workspace/

akorneta (approved):

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
sleshchenko: @ibuziuk For me, it looks like it's easy to add it here (I guess described changes should be enough) and it will be useful (machine logs panel will have all events). It's why I write it here. 
But I agree that it is not related to a solved issue and it can be added in another PR.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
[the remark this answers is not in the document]
ibuziuk: sounds good, let me fix

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
ibuziuk: @sleshchenko ok, sounds relevant - let me add it to this PR

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
ibuziuk: done

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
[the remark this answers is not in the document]
ibuziuk: moved to `internalStart`

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
```diff
@@ -347,6 +347,11 @@ che.infra.kubernetes.workspace_start_timeout_min=8
 # Defines the timeout in minutes that limits the period for which Kubernetes Ingress become ready
 che.infra.kubernetes.ingress_start_timeout_min=5
 
+# If during workspace startup an unrecoverable event defined in 'che.infra.kubernetes.workspace_unrecoverable_events' occurs
+# terminate workspace immediately instead of waiting util timeout
```
sleshchenko: Should `util` be `until`?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
```diff
@@ -600,8 +612,35 @@ public void accept(ProbeResult probeResult) {
...
+  public class UnrecoverableEventHanler implements ContainerEventHandler {
+
+    @Override
+    public void handle(ContainerEvent event) {
+      if (isUnrecoverable(event)) {
+        String reason = event.getReason();
+        String message = event.getMessage();
+        String podName = event.getPodName();
+        try {
+          internalStop(emptyMap());
```
sleshchenko, line 632: I think it doesn't work correctly until k8s runtime start interruption will be implemented. 
I guess client will receive two workspace status events:
- one that workspace failed to start because of Unrecoverable;
- a second one that start timeout exceeded.

Let me check that. BTW I think we can merge these changes with the described issue and it will be fixed with implemented interruption of k8s runtime start.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
```diff
@@ -347,6 +347,11 @@ che.infra.kubernetes.workspace_start_timeout_min=8
...
+che.infra.kubernetes.workspace_fail_on_unrecoverable_events=true
```
sleshchenko: Can you describe why it is necessary to have separate configuration property for disabling listening to unrecoverable events. Maybe it makes sense to configure empty `che.infra.kubernetes.workspace_unrecoverable_events` for disabling instead of having a separate one?

sleshchenko (approved):
Your changes will make life easier during an investigation of workspace start failing :+1:

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
ibuziuk: good catch

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
ibuziuk: good point, less config we have the better

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
sleshchenko: I've checked. So here is the Che Server logs
```

  | 2018-04-17 08:02:45,912[io-8080-exec-10]  [INFO ] [o.e.c.a.w.s.WorkspaceRuntimes 236]   - Starting workspace 'che/qwe' with id 'workspacea75oxtgryzdjq5iq' by user 'che'
  /////////////////// Here Client received that workspace failed to start but Che Server doesn't contain the corresponding message ///////////////////////////
  /////////////////// ~3 minutes later reached workspace start timeout ///////////////////////////
  | 2018-04-17 08:05:18,006[aceSharedPool-0]  [WARN ] [.i.k.KubernetesInternalRuntime 177]  - Failed to start Kubernetes runtime of workspace workspacelhu15ifhded6jrxf. Cause: Server 'terminal' in machine 'dev-machine' not available.
  | 2018-04-17 08:05:18,006[ServersChecker]   [ERROR] [.i.k.KubernetesInternalRuntime 335]  - Unable to update status of the machine '%s:%s'. Cause: %s
  | 2018-04-17 08:05:18,066[aceSharedPool-0]  [INFO ] [o.e.c.a.w.s.WorkspaceRuntimes 292]   - Workspace 'che:wksp-g9f7' with id 'workspacelhu15ifhded6jrxf' start failed
  | 2018-04-17 08:05:18,070[aceSharedPool-0]  [WARN ] [o.e.c.a.w.s.WorkspaceManager 423]    - Cannot set error status of the workspace workspacelhu15ifhded6jrxf. Error is: Workspace with id 'workspacelhu15ifhded6jrxf' doesn't exist
```

infrastructures/kubernetes/pom.xml
```diff
@@ -66,6 +66,10 @@
             <groupId>javax.inject</groupId>
             <artifactId>javax.inject</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.codehaus.plexus</groupId>
```
skabashnyuk: I guess we don't have CQ for this dependency. Can you use some alternatives?

ibuziuk: @skabashnyuk seems to be defined in parent https://github.com/eclipse/che-parent/blob/master/pom.xml#L252

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
```diff
@@ -347,6 +347,10 @@ che.infra.kubernetes.workspace_start_timeout_min=8
 # Defines the timeout in minutes that limits the period for which Kubernetes Ingress become ready
 che.infra.kubernetes.ingress_start_timeout_min=5
 
+# If during workspace startup an unrecoverable event defined in the property occurs,
+# terminate workspace immediately instead of waiting until timeout
+che.infra.kubernetes.workspace_unrecoverable_events=Failed Mount,Failed Scheduling,Failed to pull image
```
l0rd, line 358: For users it may be interesting to have a list (or better a link) of valid kubernetes events.

l0rd (approved):

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
ibuziuk: @l0rd could you please clarify a bit ? is it about changing the property format

infrastructures/kubernetes/pom.xml
skabashnyuk: It's not an issue where it's defined. the problem is that it becomes compile level and we have no cq for it. https://dev.eclipse.org/ipzilla/buglist.cgi?query_format=advanced&short_desc_type=allwordssubstr&short_desc=&component=ecd.che&long_desc_type=substring&long_desc=&bug_file_loc_type=allwordssubstr&bug_file_loc=&keywords_type=allwords&keywords=&emailassigned_to1=1&emailtype1=substring&email1=&emailassigned_to2=1&emailreporter2=1&emailcc2=1&emailtype2=substring&email2=&bugidtype=include&bug_id=&chfieldfrom=&chfieldto=Now&chfieldvalue=&cmdtype=doit&order=Reuse+same+sort+as+last+time&field0-0-0=noop&type0-0-0=noop&value0-0-0=

AFAIK you can find the same helper method on other artifacts like guava

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
garagatyi: I think Mario means a link to docs where all the event types are listed. 
And :+1:  to add it here.

l0rd: If I am a Che user and I am editing che.properties I don't know what are the valid values for this new property. You may want to say in the comment that this is a comma separated list of kubernetes events. And it would be great if we could add a link to a list of valid kubernetes events (but I am not sure we can easily find one so don't lose too much time on this)

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
```diff
@@ -600,8 +611,42 @@ public void accept(ProbeResult probeResult) {
...
+        } catch (InfrastructureException e) {
+          LOG.error("Unrecoverable event occured {}, {}, {}", reason, message, podName);
```
garagatyi: It seems that without workspace ID it won't be possible to match the incident with a workspace/user. I think we need WSID her.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/event/ContainerEvent.java
```diff
@@ -62,6 +70,7 @@ public boolean equals(Object o) {
     ContainerEvent that = (ContainerEvent) o;
     return Objects.equals(podName, that.podName)
         && Objects.equals(containerName, that.containerName)
+        && Objects.equals(reason, that.reason)
```
garagatyi, line 73: Can you also add the reason to the hashcode method? It is common to have them aligned.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
```diff
@@ -92,7 +97,11 @@ protected void startMachines() throws InfrastructureException {
     }
     // TODO https://github.com/eclipse/che/issues/7653
     // project.pods().watch(new AbnormalStopHandler());
-    // project.pods().watchContainers(new MachineLogsPublisher());
+
+    project.pods().watchContainers(new MachineLogsPublisher());
```
garagatyi, line 101: TLDR please, add finally here as well
Since we set watchers here we should also ensure that all the watchers are properly deleted. We can not guarantee that there is no bug which prevents the execution of finally block from k8s infra.

garagatyi:
@ibuziuk please, do not push to the branch with force. Each force push make reviewers review whole PR from scratch

garagatyi:
BTW looks like you need to address @skabashnyuk 's comment before merging since we can't add new dependency without a CQ

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
ibuziuk: sorry, not following - `startMachines` is supposed to be called from `internalStart` where watchers closed / deleted in finally. Could you describe potential bug that could prevent closing watchers currently ?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
ibuziuk: ok, will also add workspace id

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/event/ContainerEvent.java
ibuziuk: done, thanks for the pointer

infrastructures/kubernetes/pom.xml
ibuziuk: @skabashnyuk ok, good to know - changed to `com.google.common.base.Strings`

ibuziuk:
@garagatyi fixed dep CQ problem -  changed to `com.google.common.base.Strings`

ibuziuk:
@l0rd have added PR for docs https://github.com/eclipse/che-docs/pull/395

garagatyi:
@ibuziuk is there any other cases that you are going to cover in unit tests?

ibuziuk:
@garagatyi yeah, working on adding more tests

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
```diff
@@ -347,6 +347,10 @@ che.infra.kubernetes.workspace_start_timeout_min=8
 # Defines the timeout in minutes that limits the period for which Kubernetes Ingress become ready
 che.infra.kubernetes.ingress_start_timeout_min=5
 
+# If during workspace startup an unrecoverable event defined in the property occurs,
+# terminate workspace immediately instead of waiting until timeout
+che.infra.kubernetes.workspace_unrecoverable_events=Failed Mount,Failed Scheduling,Failed
```
l0rd: `Failed to pull image` has become `Failed`

l0rd (changes requested):

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
ibuziuk: it was actually done deliberate, but I can rework

l0rd: Yes please revert that. Having `Failed` in a list of kubernetes events is confusing.

skabashnyuk (approved):

ibuziuk:
@l0rd @garagatyi changed property value, added tests. Please, review

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftInternalRuntime.java
garagatyi: Sorry, my bad. We don't need finally here.

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntimeTest.java
```diff
@@ -352,8 +412,9 @@ public void throwsInfrastructureExceptionWhenKubernetesNamespaceCleanupFailed()
   @Test
   public void testRepublishContainerOutputAsMachineLogEvents() throws Exception {
     final MachineLogsPublisher logsPublisher = internalRuntime.new MachineLogsPublisher();
-    final ContainerEvent out1 = mockContainerEvent("pulling image", "07/07/2007 19:01:22");
-    final ContainerEvent out2 = mockContainerEvent("image pulled", "07/07/2007 19:08:53");
+    final ContainerEvent out1 =
```
garagatyi, line 415: You have added logs publishing in the runtime start process and this is very cool. Can you also add a unit test for this case?

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntimeTest.java
```diff
@@ -366,10 +427,34 @@ public void testRepublishContainerOutputAsMachineLogEvents() throws Exception {
     assertTrue(captor.getAllValues().containsAll(machineLogs));
   }
 
+  @Test
+  public void testHandleUnrecoverableEvent() throws Exception {
+    final UnrecoverableEventHanler unrecoverableEventHanler =
```
garagatyi: Please, test unrecoverable events matching/non-matching in the whole flow instead of with a separated component. We don't have any guarantees that handler is called as it is supposed in this test case.
Please cover cases:
non-matching event
matching by the reason field
matching by the message field

It would be nice to verify that stop of the runtime is really called after appearing of the matching event.

l0rd (approved):
LGTM. Thanks @ibuziuk!

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntimeTest.java
ibuziuk: isn't it already handled by https://github.com/eclipse/che/pull/9426/files#diff-08f3ea6fea2790aa766edf8d57d37897R275 e.g. handler for publishing events is added during runtime start?

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntimeTest.java
ibuziuk: @garagatyi was not able to figure out how exactly to test unrecoverable events in starting flow, but have added separate tests for matching by the reason / matching by the message. Also, not sure how exactly verifying that stop of the runtime e.g. `internalStop` should be tested - currently done by verifying that namespace cleanup was called

Are you ok to merge it in the current state, since I do not really want to face merge conflicts, and we could discuss more tests cases which I would add later in separate PR ?

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntimeTest.java
garagatyi: Well, it just checks that some subscribers are watching it. We have a separate test for logs publisher, so maybe we can also check that one of the watchers is logs publisher and another one is unrecoverable events listener? But it is not critical, so feel free to leave it as is if you want.

garagatyi (approved):
Looks good! I believe that tests that we were discussing are not mandatory for the PR.

ibuziuk:
@garagatyi ok, thanks I will try to add more tests for those event handlers during the sprint

Issue #9431, opened 2018-04-16
Workspace PVCs aren't deleted when pods get created in one namespace


ghost:
### Description

When workspace objects are created in one shared namespace (either in a namespace with server itself or their own dedicated namespace), workspace deletion does not delete associated pvcs. As a result, each deleted workspace leaves 2 pvcs hanging out there. They are bound to PVs, thus making them unavailable for new workspaces.

### Reproduction Steps

1. Deploy Che to OpenShift with `CHE_INFRA_OPENSHIFT_PROJECT=eclipse-che` or any other namespace

2. Create and start a workspace

3. Stop & delete a workspace.

PVCs have not been deleted.

![image](https://user-images.githubusercontent.com/5337267/38789578-d83f1942-4143-11e8-8372-4b39720dd519.png)

```
eugene@ivantsoft ~/tools/gopath/src/github.com/eivantsov/che-cli (master) $ oc get pv
NAME      CAPACITY   ACCESSMODES   RECLAIMPOLICY   STATUS      CLAIM                                      STORAGECLASS   REASON    AGE
pv0042    100Gi      RWO,ROX,RWX   Recycle         Bound       eclipse-che/claim-che-workspace-sfq5ghh4                            2d                                                            2d
pv0064    100Gi      RWO,ROX,RWX   Recycle         Bound       eclipse-che/claim-che-workspace-ahw4bm2p                            2d                                                          2d
pv0070    100Gi      RWO,ROX,RWX   Recycle         Bound       eclipse-che/che-data-volume                                         2d
```

I've run into a problem with lack of pvs when testing https://github.com/eclipse/che/tree/templates where templates define the following behavior:

1. No username/password or token is used to create workspace objects (che SA is used), and thus
2. All workspace objects are created in the same namespace with Che

QA tests failed at some point, because workspaces are created and deleted during QA cycle, and in a short while OpenShift cluster was unable to satisfy more claims.

PR #9437, opened 2018-04-16
Fixed removing of PVCs when OpenShift project name is predefined

sleshchenko:
### What does this PR do?
The bug appeared on OpenShift infrastructure because of `che.infra.kubernetes.namespace` configuration property. It is used by few components which are bound for both Kubernetes and OpenShift infrastructure. But OpenShift infrastructure introduces a new property called `che.infra.openshift.project` and ignore old one. So components used non-configured property, it was a reason why WorkspacePVCCleaner doesn't subscribe to WorkspaceRemoved events.

This PR introduces changes to avoiding injection of `che.infra.kubernetes.namespace` configuration property. Instead of it, components should use other components like KubernetesNamespaceFactory, KubernetesNamespace, which are overridden by OpenShift infrastructure (OpenShiftProjectFactory, OpenShiftProject).

https://github.com/eclipse/che/issues/9431

#### Release Notes
N/A

#### Docs PR
N/A


garagatyi (approved):
Looks good!

akorneta (approved):

PR #9446, opened 2018-04-16
CHE-9428: fix kubernetes infra after a merge of not checked PR

garagatyi:
### What does this PR do?
Fix kubernetes infra after a merge of not checked PR.
Set correct bindings of generics, fix generics;
Add missing properties in che.properties and kubectl deployment.

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
@@ -335,6 +335,12 @@ che.infra.kubernetes.password=
 che.infra.kubernetes.oauth_token=
 che.infra.kubernetes.trust_certs=
 
+# Defines the way how servers are exposed to the world in k8s infra.
+# List of  strategies implemented in Che: default-host, multi-host, single-host
+che.infra.kubernetes.server_strategy=default-host
+# Used to generate domain for a server in a workspace
+che.infra.kubernetes.ingress.domain=unspecified-ingress-domain
+
 # Defines Kubernetes namespace in which all workspaces will be created.
 # If not set, every workspace will be created in a new namespace, where namespace = workspace id
 #
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
@@ -339,7 +339,7 @@ che.infra.kubernetes.trust_certs=
 # List of  strategies implemented in Che: default-host, multi-host, single-host
 che.infra.kubernetes.server_strategy=default-host
 # Used to generate domain for a server in a workspace in case property `che.infra.kubernetes.server_strategy` is set to `multi-host`
-che.infra.kubernetes.ingress.domain=unspecified-ingress-domain
+che.infra.kubernetes.ingress.domain=
 
 # Defines Kubernetes namespace in which all workspaces will be created.
 # If not set, every workspace will be created in a new namespace, where namespace = workspace id
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
@@ -338,7 +338,7 @@ che.infra.kubernetes.trust_certs=
 # Defines the way how servers are exposed to the world in k8s infra.
 # List of  strategies implemented in Che: default-host, multi-host, single-host
 che.infra.kubernetes.server_strategy=default-host
-# Used to generate domain for a server in a workspace
+# Used to generate domain for a server in a workspace in case property `che.infra.kubernetes.server_strategy` is set to `multi-host`
 che.infra.kubernetes.ingress.domain=unspecified-ingress-domain
 
 # Defines Kubernetes namespace in which all workspaces will be created.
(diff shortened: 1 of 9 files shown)


ghost (approved):

garagatyi:
ci-test

skabashnyuk (approved):

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
```diff
@@ -335,6 +335,12 @@ che.infra.kubernetes.password=
 che.infra.kubernetes.oauth_token=
 che.infra.kubernetes.trust_certs=
 
+# Defines the way how servers are exposed to the world in k8s infra.
+# List of  strategies implemented in Che: default-host, multi-host, single-host
+che.infra.kubernetes.server_strategy=default-host
+# Used to generate domain for a server in a workspace
+che.infra.kubernetes.ingress.domain=unspecified-ingress-domain
```
sleshchenko: Is it used for each of server strategies?

akorneta (approved):

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
l0rd: And what are the possible values?

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
```diff
@@ -335,6 +335,12 @@ che.infra.kubernetes.password=
 che.infra.kubernetes.oauth_token=
 che.infra.kubernetes.trust_certs=
 
+# Defines the way how servers are exposed to the world in k8s infra.
+# List of  strategies implemented in Che: default-host, multi-host, single-host
+che.infra.kubernetes.server_strategy=default-host
```
l0rd, line 340: How is this related with `che.single.port`? Does this property makes sense if single port is set to `false`?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInfraModule.java
```diff
@@ -77,8 +78,11 @@ protected void configure() {
     volumesStrategies.addBinding(UNIQUE_STRATEGY).to(UniqueWorkspacePVCStrategy.class);
     bind(WorkspaceVolumesStrategy.class).toProvider(WorkspaceVolumeStrategyProvider.class);
 
-    MapBinder<String, ExternalServerExposerStrategy> ingressStrategies =
-        MapBinder.newMapBinder(binder(), String.class, ExternalServerExposerStrategy.class);
+    MapBinder<String, ExternalServerExposerStrategy<KubernetesEnvironment>> ingressStrategies =
+        MapBinder.newMapBinder(
+            binder(),
+            new TypeLiteral<String>() {},
```
l0rd, line 84: I am curious: do you really need to use `TypeLiteral` for String (that is not a generic type)?

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
garagatyi: > Is it used for each of server strategies?

no, it is used for `multi-host` strategy only

> And what are the possible values?

Something like `192.168.99.101.nip.io`. Really depends on DNS provider, public multi-level DNS, IP, etc.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
garagatyi: It is not related to a single port AFAIK

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInfraModule.java
garagatyi: I need TypeLiteral for `ExternalServerExposerStrategy<KubernetesEnvironment>` and there is no binding method with mixed class and type literal arguments

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
sleshchenko: > no, it is used for multi-host strategy only

Then maybe it makes sense to make it more clear in doc or even property name?
Is it possible that default value `unspecified-ingress-domain` will work? If not, maybe it'd better to leave it empty and fail tomcat start if it is not configured?

garagatyi: I didn't add this property in code, so not sure whether I should change the name now. But I can add a comment about that.
In a case when multi-host is not chosen strategy failure on start when the value is not set would break Che in different valid cases. Value `unspecified-ingress-domain` won't work but hints a user that he needs to address this issue.

sleshchenko: > In a case when multi-host is not chosen strategy failure on start when the value is not set would break Che in different valid cases.

I mean failing start only if multi-host is chosen and `che.infra.kubernetes.ingress.domain` is not configured

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInfraModule.java
l0rd: Ok makes sense thanks

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
l0rd: Ok just wondering how `multi-host` works with single port set to `false`. But in any case this is not related to this PR so ok to merge for me.

l0rd (approved):

sleshchenko (approved):

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
garagatyi: AFAIK Single host mode is used by docker infrastructure only. So neither k8s nor openshift installations of Che use it

Issue #9432, opened 2018-04-16
Unexpected fail of "CheckRestoringSplitEditorTest" selenium test


Ohrimenko1988:
### Description
We have unexpected fail of "CheckRestoringSplitEditorTest" selenium test. It looks like something went wrong with the method which compares specified text template and existing text in the defined split editor tab. 

**Attachment:**

![rrrr](https://user-images.githubusercontent.com/22724619/38795034-a78ce1be-415f-11e8-9e62-0a8c36e57fe1.png)

**Log:**
```
org.openqa.selenium.TimeoutException: 
Expected condition failed: waiting for org.eclipse.che.selenium.pageobject.CodenvyEditor$$Lambda$135/1509260156@3977aeff (tried for 10 second(s) with 500 MILLISECONDS interval)
Build info: version: '3.0.1', revision: '1969d75', time: '2016-10-18 09:49:13 -0700'
System info: host: 'slave7.codenvycorp.com', ip: '127.0.0.1', os.name: 'Linux', os.arch: 'amd64', os.version: '3.10.0-514.16.1.el7.x86_64', java.version: '1.8.0_102'
Driver info: org.eclipse.che.selenium.core.SeleniumWebDriver
	at org.eclipse.che.selenium.editor.CheckRestoringSplitEditorTest.checkSplitdEditorAfterRefreshing(CheckRestoringSplitEditorTest.java:121)
	at org.eclipse.che.selenium.editor.CheckRestoringSplitEditorTest.checkRestoringStateSplittedEditor(CheckRestoringSplitEditorTest.java:107)
```

**Report with reproduced problem:**
https://ci.codenvycorp.com/view/qa/job/che-integration-tests-master-ocp/215/Selenium_tests_report/

Ohrimenko1988:
Investigate the problem is over and bug issue was created 
https://github.com/eclipse/che/issues/9456

PR #9451, opened 2018-04-17
Selenium: Add logs storing for investigate issue #9432

Ohrimenko1988:
### What does this PR do?
Add logs storing for investigate issue #9432

Issue: https://github.com/eclipse/che/issues/9432


SkorikSergey (approved):

dmytro-ndp:
All webdriver logs are stored automatically in target directory in case of exception thrown. Why do you need to store webdriver logs again?

Issue #9452, opened 2018-04-17
Workspace Start/Stop failed while tomcat shutdown


akorneta:
### Description
All the service terminations are performed on container `@PreDestroy` and since the shutdown calls are not [ordered](https://github.com/eclipse/che/blob/2c2bd6637c8169018e2b33397f9d738bbf82a0b1/core/commons/che-core-commons-inject/src/main/java/org/eclipse/che/inject/lifecycle/Destroyer.java#L21) there is no guarantee that workspace service terminator will be the first one. So there are possible cases when e.g. [KubernetesClientFactory.cleanup()](https://github.com/eclipse/che/blob/master/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java#L232) will be called first then workspaces might fail because all the connection to `OpenShift` will be closed.
As result, there is no guarantee that  user workspaces will be started or normally stopped while rolling updates of Che master.

mshaposhnik:
Done, moved KubernetesClientFactory.cleanup() into separate termination instead of `@PreDestroy`.

PR #9463, opened 2018-04-17
Kubernetes client shutdown via termination

mshaposhnik:
### What does this PR do?
Kubernetes client shutdown via termination

https://github.com/eclipse/che/issues/9452


infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientFactory.java
```diff
@@ -192,7 +191,7 @@ public Response intercept(Chain chain) throws IOException {
     };
   }
 
-  protected void doCleanup() {
+  public void shutdownClient() {
```
sleshchenko: Please add java doc since it is public method

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientTermination.java
```diff
@@ -0,0 +1,51 @@
...
+import org.eclipse.che.api.system.server.ServiceTermination;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Termination for Kubernetes HTTP client.
+ *
+ * @author Max Shaposhnik (mshaposh@redhat.com)
+ */
+public class KubernetesClientTermination implements ServiceTermination {
```
sleshchenko, line 26: Is there any guarantee that `WorkspaceServiceTermination` will be invoked before `KubernetesClientTermination`?

mshaposhnik: No :) But at least we can change binding order if something will go wrong

mshaposhnik: It fits our needs better, but i see very weak logic how K8S client stop extends workspace shutdowning...

akorneta (approved):

sleshchenko (approved):
Now LGTM

PR #9483, opened 2018-04-18
Add dependencies for terminations

mshaposhnik:
### What does this PR do?
Add dependencies for terminations

### DIFF
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientTermination.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientTermination.java
@@ -14,6 +14,7 @@ import com.google.common.collect.ImmutableSet;
 import java.util.Set;
 import javax.inject.Inject;
 import org.eclipse.che.api.system.server.ServiceTermination;
+import org.eclipse.che.api.workspace.server.WorkspaceServiceTermination;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -53,6 +54,6 @@ public class KubernetesClientTermination implements ServiceTermination {
 
   @Override
   public Set<String> getDependencies() {
-    return ImmutableSet.of("workspace");
+    return ImmutableSet.of(WorkspaceServiceTermination.SERVICE_NAME);
   }
 }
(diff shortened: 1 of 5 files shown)


wsmaster/che-core-api-system/src/main/java/org/eclipse/che/api/system/server/ServiceTerminator.java
```diff
@@ -30,12 +38,14 @@
...
-  private final Set<ServiceTermination> terminations;
+  private final List<ServiceTermination> terminations;
 
   @Inject
   ServiceTerminator(EventService eventService, Set<ServiceTermination> terminations) {
     this.eventService = eventService;
-    this.terminations = terminations;
+    ArrayList<ServiceTermination> terminationsList = new ArrayList<>(terminations);
+    Collections.sort(terminationsList, new ServiceTerminationComparator(terminations));
+    this.terminations = terminationsList;
```
skabashnyuk: can we do that in one shot? With stream for example?

wsmaster/che-core-api-system/src/main/java/org/eclipse/che/api/system/server/ServiceTerminator.java
```diff
@@ -88,4 +99,45 @@ private void doTerminate(ServiceTermination termination) throws InterruptedExcep
...
+    }
+
+    private final Map<String, Set<String>> dependencies;
+
+    @Override
+    public int compare(ServiceTermination o1, ServiceTermination o2) {
+
+      return checkTransitiveDependency(o1.getServiceName(), o2.getServiceName(), new HashSet<>());
+    }
+
```
skabashnyuk: can you add some info what is going on here?

wsmaster/che-core-api-system/src/main/java/org/eclipse/che/api/system/server/ServiceTerminator.java
akorneta: +1 consider to do it like `this.terminations = ImmutableSortedSet.copyOf(new Comparator<..>(), terminations);`

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesClientTermination.java
```diff
@@ -48,4 +50,9 @@ public void suspend() throws InterruptedException {
   public String getServiceName() {
     return "KubernetesClient";
   }
+
+  @Override
+  public Set<String> getDependencies() {
+    return ImmutableSet.of("workspace");
```
sleshchenko: Please consider using constant from WorkspaceServiceTermination here

akorneta (approved):

sleshchenko (approved):

skabashnyuk (approved):

Issue #9489, opened 2018-04-18
Parametrize keycloak admin user in templates


siamaksade:
The KeyCloak admin user and password is hard-coded to `admin` in the templates:
https://github.com/eclipse/che/tree/master/deploy/openshift/templates

They should become template parameters.

eivantsov:
@siamaksade they are hardcoded but in jsons here - https://github.com/eclipse/che/tree/master/dockerfiles/init/modules/keycloak/templates

Env variables are ignored in this case. I'll issue a PR to fix it. Thanks for the pointer.

siamaksade:
Slight relevant to this, the `CHE_KEYCLOAK_ADMIN_REQUIRE_UPDATE_PASSWORD` template param should only be in the keycloak template. It's not used in [Che template](https://github.com/eclipse/che/blob/master/deploy/openshift/templates/che-server-template.yaml)

eivantsov:
@siamaksade it is not use din Che template. Not sure I understand your comment.

eivantsov:
Closing this issue as the original request was addressed in a merged PR. @siamaksade feel free to open a separate issue related to your another concern.

PR #9490, opened 2018-04-19
Make Keycloak Admin configurable

ghost:
### What does this PR do?

Makes Keycloak admin user name and password configurable

https://github.com/eclipse/che/issues/9489


dockerfiles/keycloak/kc_realm_user.sh
```diff
@@ -24,6 +21,12 @@ cat /scripts/che-realm.json.erb | \
                                 sed -e "s@<%= scope\.lookupvar('che::che_server_url') %>@${PROTOCOL}://che-${NAMESPACE}.${ROUTING_SUFFIX}@" \
                                 > /scripts/che-realm.json
 
+echo "Creating Admin user..."
+
+if [ $KEYCLOAK_USER ] && [ $KEYCLOAK_PASSWORD ]; then
```
riuvshin, line 26: `! -z` ?

ghost: Basically reused this https://github.com/jboss-dockerfiles/keycloak/blob/3.3.0.CR2-3/server/docker-entrypoint.sh

In fact, even a check isn't really required since ENVS are set in deploymentConfig.

deploy/openshift/templates/multi/keycloak-template.yaml
```diff
@@ -140,10 +140,20 @@ objects:
...
   value: http
+- name: KEYCLOAK_USER
+  displayName: Keycloak admin user
+  description: Default Keycloak admin user
+  value: "admin"
+- name: KEYCLOAK_PASSWORD
+  displayName: Keycloak admin password
+  description: Default Keycloak admin password. Can be changed after login
+  value: "admin"
+  value: http
```
garagatyi: Looks like this line is a typo

deploy/openshift/templates/multi/keycloak-template.yaml
```diff
@@ -140,10 +140,20 @@ objects:
         storage: 1Gi
 
 parameters:
+
```
garagatyi: Do we need this empty line?

garagatyi (approved):
Looks good, but I wonder why we haven't used command line params instead of those files beforehand.

deploy/openshift/templates/multi/keycloak-template.yaml
ghost: Fixed. Thanks

deploy/openshift/templates/multi/keycloak-template.yaml
ghost: Fixed. Thanks

riuvshin (approved):

PR #9504, opened 2018-04-20
Fix bootstrapping fast failed when URL is invalid

akorneta:
### What does this PR do?
Adds `-f` flag to `curl` while bootstrapping agents, so if HTTP error occurs on downloading binaries then workspace start will fail.


sleshchenko (approved):
:+1: Thanks for fixing it. It will help recognize configuration issues easily.
Thank @eivantsov for reporting such an issue

ghost (approved):

garagatyi (approved):

Issue #8467, opened 2018-01-26
Quick documentation is broken when use java language server


mmorhun:
### Reproduction Steps
1. Open a workspace with a java project
2. Set cursor into some class name
3. Invoke Quick Documentation (`Ctrl + q`)

![screenshot from 2018-01-26 10-55-08](https://user-images.githubusercontent.com/15607393/35432032-74449a2a-0287-11e8-81a3-e3ca19cc9ce0.png)

<details>
 <summary>Stack trace in dev-machine log</summary>

```
java.lang.NullPointerException: null
	at org.eclipse.che.jdt.JavadocFinder.findJavadoc(JavadocFinder.java:161)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at org.eclipse.che.plugin.java.server.rest.JavadocService.findJavadoc(JavadocService.java:50)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at org.everrest.core.impl.method.DefaultMethodInvoker.invokeMethod(DefaultMethodInvoker.java:140)
	at org.everrest.core.impl.method.DefaultMethodInvoker.invokeMethod(DefaultMethodInvoker.java:60)
	at org.everrest.core.impl.RequestDispatcher.dispatch(RequestDispatcher.java:233)
	at org.everrest.core.impl.RequestDispatcher.dispatch(RequestDispatcher.java:128)
	at org.everrest.core.impl.RequestDispatcher.doInvokeResource(RequestDispatcher.java:306)
	at org.everrest.core.impl.EverrestProcessor.process(EverrestProcessor.java:120)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:742)
	at org.everrest.core.servlet.EverrestServlet.service(EverrestServlet.java:61)
	at com.google.inject.servlet.ManagedServletPipeline.service(ManagedServletPipeline.java:91)
	at com.google.inject.servlet.ServletDefinition.service(ServletDefinition.java:181)
	at com.google.inject.servlet.ServletDefinition.doServiceImpl(ServletDefinition.java:286)
	at org.everrest.core.impl.RequestHandlerImpl.handleRequest(RequestHandlerImpl.java:62)
[...]

PR #9508, opened 2018-04-20
Fixed quick documentation

JPinkney:
### What does this PR do?
Fixed quick documentation by binding the quick documentation action to hover. It uses the Orion built in command orion.edit.showTooltip to perform the hover at the current location.

https://github.com/eclipse/che/issues/8467


selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/editor/autocomplete/JavaDocPopupTest.java
```diff
@@ -140,7 +137,7 @@ public void javaDocPopupTest() throws Exception {
 
     editor.openJavaDocPopUp();
     editor.waitJavaDocPopUpOpened();
-    editor.checkTextToBePresentInJavaDocPopUp(ANNOTATION_TEXT);
+    // editor.checkTextToBePresentInJavaDocPopUp(ANNOTATION_TEXT);
```
dkulieshov, line 140: I guess you accidentally forgot to remove this line :-)

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/hover/HoverProvider.java
```diff
@@ -105,4 +127,89 @@ private String renderContent(Hover hover) {
...
+   */
+  private String createWorkableURI(String uri) {
+    String fixedURI = uri;
+    if (uri.indexOf("#") != -1) {
+      fixedURI = fixedURI.substring(0, uri.lastIndexOf("#"));
+    }
+
+    // Because of the markdown2html library in jdt.ls file:///projects gets stripped to
+    // file:/projects when converting the markdown to html
+    if (uri.startsWith("file:/projects")) {
```
dkulieshov: Here I guess we should consider both cases ("file:///" and "file:/") in order to make the code to process correctly both variants.

dkulieshov (approved):

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/hover/HoverProvider.java
tsmaeder: No, if it's file:///, it's already correct and doesn't need to be changed.

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/hover/HoverProvider.java
```diff
@@ -105,4 +127,89 @@ private String renderContent(Hover hover) {
...
+   * @param uri The URI to fix
+   * @return A URI that can be opened by the editor
+   */
+  private String createWorkableURI(String uri) {
+    String fixedURI = uri;
+    if (uri.indexOf("#") != -1) {
+      fixedURI = fixedURI.substring(0, uri.lastIndexOf("#"));
+    }
+
+    // Because of the markdown2html library in jdt.ls file:///projects gets stripped to
```
tsmaeder: markdown2html? Shouldn't that be javadoc2markdown? If so, why don't we fix it in jdt.ls for all clients instead of handling it here?

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/hover/HoverProvider.java
```diff
@@ -105,4 +127,89 @@ private String renderContent(Hover hover) {
...
+  }
+
+  /**
+   * Gets the location from a URI that is in the form of URI#LineNumber
+   *
+   * @param uri The URI to get the location from
+   * @return Location object from the URI
+   */
+  private Location getLocationFromURI(String uri) {
+    String pattern = "\\d+$";
```
tsmaeder: If it's of the form uri#lineNumber, wouldn't a simple uri.lastIndexOf('#') do the same trick?

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/hover/HoverProvider.java
```diff
@@ -105,4 +127,89 @@ private String renderContent(Hover hover) {
...
+
+    Position uriPos = new Position();
+
+    // We have line information from the URI
+    if (m != null && m.getGroupCount() > 0) {
+      String lastMatch = m.getGroup(m.getGroupCount() - 1);
+      Integer lineNumber = Integer.parseInt(lastMatch);
+
+      if (lineNumber > 0) {
+        uriPos.setLine(lineNumber - 1);
```
tsmaeder: Why do we do that?

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
```diff
@@ -1433,27 +1426,16 @@ private void waitTextInJavaDoc(String expectedText) {
...
-   * <p>Please use {@link CodenvyEditor#waitAvailabilityAndSwitchToJavaDocFrame()} method for
-   * switching to javadoc frame.
-   *
    * @param expectedText text which should be present in javadoc body
    * @return true - if {@code expectedText} is present in javadoc body, false - if not
    */
   public boolean waitAndCheckTextPresenceInJavaDoc(String expectedText) {
     return seleniumWebDriverHelper
-        .waitVisibility(By.tagName("body"))
+        .waitVisibility(By.xpath(JAVA_DOC_POPUP))
```
tsmaeder, line 1434: I guess this is a bugfix?

tsmaeder (changes requested):
A couple of questions.

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/hover/HoverProvider.java
JPinkney: Yes I think the comment should say javadoc2markdown. The problem is that the library for javadoc2markdown always removes the two trailing slashes. So file:/// always gets changed to file:/. 

It looks like the javadoc2markdown library hasn't been worked on in years so i'm not sure how fast that change would be integrated/released in that and the only other option that I can think of is a do a string replace of all file:/ with file:///  in jdt.ls.

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/hover/HoverProvider.java
JPinkney: Yes I will change this, I just didn't know if we'd run into some weird case where uri.lastIndexOf('#') might break

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/hover/HoverProvider.java
JPinkney: IIRC VSCode and Che have different behaviours on how they interpret the line they should jump to. In VSCode file://somefile#20 would link right to line 20 whereas with Che file://somefile#20 would link to line 21

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
JPinkney: Yes, before the quick documentation opened javadoc in an iframe but now quick documentation uses the built in hover which isn't an iframe so I had to change this

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/hover/HoverProvider.java
tsmaeder: Ok, I can live with that for now, but I think we should fix this, or all clients will have to work around it. Can you open a but against jdt.ls describing the problem?

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/hover/HoverProvider.java
tsmaeder: 0-based vs. 1-based? But shouldn't we get the line number in the for LSP encodes it and shouldn't "OpenFileInEditorHelper" interpret that correctly?

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
tsmaeder: ack

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/hover/HoverProvider.java
JPinkney: Currently [this](https://github.com/eclipse/eclipse.jdt.ls/blob/master/org.eclipse.jdt.ls.core/src/org/eclipse/jdt/ls/core/internal/javadoc/JavaElementLinks.java#L180) is how its implemented in JDT.LS. 

> But shouldn't we get the line number in the for LSP encodes it 

I think you're actually correct. In the code I linked above the LSP encodes it correctly but for VSCode to actually jump to the correct line you have to add one because of their indexing.

> shouldn't "OpenFileInEditorHelper" interpret that correctly

I think the problem is that VSCode has their links being 1-based whereas in Che its correct which creates unfortunate consequences (like this one).

tsmaeder (approved):

PR #9513, opened 2018-04-22
Update deploy_che script

ghost:
### What does this PR do?

After https://github.com/eclipse/che/pull/9190 has been merged, existing deploy_che.sh and everything not in `deploy/openshift/templates` directory should go away (helper scripts and yamls) and start using new templates (deploy_che.sh).

This PR:

1. Removes all old yamls and scripts. Only `ocp.sh` and `deploy_che.sh` are left. Both have help menus and are quite clear to follow the deployment logic
2. Makes `deploy_che.sh` templates friendly, i.e. deployment happens by applying templates.
3. Introduces args to `deploy_che.sh` so that once can override most commonly used ENVS and params like update strategy, pull policy, deployment image, https support, multi user flavor, etc.
4. Makes it possible for `deploy_che.sh` to grab all local env variables beginning with `CHE_` and pass them to Che deployment. Envs and params are printed before templates are applied.
5. NEW! Auto-computes routing suffix by creating and exposing (and of course deleting afterwards) a test service. If `ROUTING_SUFFIX` is exported (for example, `ocp.sh` does that), it is used as is.
6. Removes the concept of flavor in `deploy_che.sh`. The script deploys Che with a set of envs and params to any OpenShift cluster be it MiniShift, a local Origin, OSD, OCP or OSO. There's just one pre-req: one needs to be logged in oc, i.e. have an active session (`oc whoami` is the test command in the script). So, when deploying, say, to OSO or OCP/OSD, all one needs to do is to login to cluster using `oc` and run `./deploy_che.sh` with a set of commands. It is possible that the desired namespace name already exists (in someone else's account) - it is possible to provide own with `-p | --project` parameter.

Tested with Origin (oc cluster up locally) and OSO Pro.


deploy/openshift/deploy_che.sh
```diff
@@ -14,563 +14,376 @@
...
+--no-pull - IfNotPresent pull policy for Che server deployment
+--rolling - Rolling update strategy (Recreate is the default one). With Rolling strategy Che server pvc and volume aren't created
+--wait-che - Track Che deployment progress until pod is healthy
+--debug - Deploy Che in a debug mode, create and expose debug route
+--image-che= - Override default Che image. Example --image-che=org/repo:tag. Tag is mandatory!
+===================================
+ENV vars: this script automatically detect envs vars beginning with "CHE_" and passes them to Che deployments:
+CHE_IMAGE_REPO - Che server Docker image, defaults to "eclipse-che-server"
+CHE_IMAGE_TAG - Set che-server image tag, defaults to "nightly"
+CHE_INFRA_KUBERNETES_PROJECT - namespace for workspace objects (defaults to eclipse-che). A separate ws namespace can be used only if username/password or token is provided
```
sleshchenko: `CHE_INFRA_KUBERNETES_NAMESPACE` is used for Kubernetes Infra.
`CHE_INFRA_OPENSHIFT_PROJECT` is used for OpenShift Infra.

deploy/openshift/deploy_che.sh
```diff
@@ -14,563 +14,376 @@
...
+                         -p PULL_POLICY=${IMAGE_PULL_POLICY} \
+                         -p STRATEGY=${UPDATE_STRATEGY} \
+                         -p CHE_MULTIUSER=${CHE_MULTIUSER} \
+                         -p PROTOCOL=${HTTP_PROTOCOL} \
+                         -p WS_PROTOCOL=${WS_PROTOCOL} \
+                         -p TLS=${TLS} \
+                         ${ENV}
+
+    if [ ${UPDATE_STRATEGY} == "Recreate" ]; then
+    ${OC_BINARY} apply -f ${BASE_DIR}/templates/pvc/che-server-pvc.yaml
```
sleshchenko: Looks like the formatting is broken here. Please check

deploy/openshift/ocp.sh
```diff
@@ -328,6 +205,11 @@ parse_args() {
     --destroy - destroy ocp cluster
     --deploy-che - deploy che to ocp
     --multiuser - deploy che in multiuser mode
+    --no-pull - IfNotPresent pull policy for Che server deployment
+    --rolling - rolling update strategy (Recreate is the default one)
+    --wait-che - track Che deployment progress until pod is healthy
+    --debug - deploy Che in a debug mode, create and expose debug route
+    --image-che= - override default Che image. Example --image-che=org/repo:tag. Tag is mandatory!
```
sleshchenko: looks like `=` symbol is redundant

ghost: Yes, makes sense - the format is in the example anyway

sleshchenko (approved):
LGTM :+1: 

riuvshin (approved):

Issue #9429, opened 2018-04-15
Unable to config keycloak if use https


LarryZhangy:
### Description
If config https to run che by use nightly image, it can't access keycloak  by run /scripts/keycloak_config.sh in keycloak pod， Because there is no https config in keycloak_config.sh

I can access https://keycloak-domaim/auth on chrome.

### Reproduction Steps
config keycloak by run job
```
apiVersion: batch/v1
kind: Job
metadata:
  name: keycloak-configure-job
spec:
  template:
    spec:
      initContainers:
      - name: wait-for-keycloak
        image: <my>/alphine:3.5
        command: ["sh", "-c", " adresses_length=0; until [ $adresses_length -gt 0 ]; do echo waiting for keycloak to be ready...; sleep 2; endpoints=`curl -s --cacert /var/run/secrets/kubernetes.io/serviceaccount/ca.crt -H \"Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)\"     https://kubernetes.default.svc/api/v1/namespaces/$POD_NAMESPACE/endpoints/keycloak`; adresses_length=`echo $endpoints | jq -r \".subsets[]?.addresses // [] | length\"`; done;"]
        env:
          - name: POD_NAMESPACE
            valueFrom:
              fieldRef:
                fieldPath: metadata.namespace
      containers:
      - name: keycloak-configure-job
        image: eclipse/che-keycloak:nightly
        command: ["/scripts/keycloak_config.sh"]
        env:
          - name: HTTP_PROTOCOL
            value: "https"
          - name: KC_HOST
            value: <domain>
          - name: CHE_KEYCLOAK_ADMIN_REQUIRE_UPDATE_PASSWORD
            value: "true"
          - name: CHE_HOST
            value: <domain>
      serviceAccountName: che-keycloak
      restartPolicy: Never
  #backoffLimit: 4
```

**OS and version:**    

all image of che:   nightly

**Diagnostics:**

LarryZhangy:
@guydaichs

guydaichs:
There are a couple of issues with running KeyCloak with TLS:
1. Initial Che realm configuration (keycloak_config.sh) might fail due to [SSL enforcement on master realm](https://www.keycloak.org/docs/3.2/server_admin/topics/realms/ssl.html#_ssl_modes), if the pod's cluster IP is not in a certain range (10.0.x.x, 192.168.x.x, and 172..16.x.x). 
[...]

PR #9533, opened 2018-04-24
Remove Keycloak job, use the right entrypoint, add envs the script re…

ghost:
### What does this PR do?

1. Removes Keycloak job to configure realm and user
2. Adds a new entrypoint to Keycloak container spec
3. Provides Keycloak pod with required ENVs used by entrypoint script

https://github.com/eclipse/che/issues/9429


deploy/kubernetes/helm/che/charts/che-keycloak/templates/deployment.yaml
```diff
@@ -45,7 +45,19 @@ spec:
           value: admin
         - name: CHE_HOST
           value: {{ template "cheHost" . }}
+        - name: ROUTING_SUFFIX
```
guydc, line 48: Is it possible to use the [Che HostHelper template](https://github.com/eclipse/che/blob/master/deploy/kubernetes/helm/che/templates/_hostHelper.tpl) instead of passing the ingress domain, namespace, and constructing the hostname in the configuration script?
This will be more consistent with the way we configure ingresses, config maps, etc.

ghost: Yes, i have thought about it, but I will then need to change the entrypoint script that expects PROTOCOL and ROUTING_SUFFIX envs.

garagatyi (approved):

sleshchenko (approved):

Issue #8293, opened 2018-01-15
Unexpected fail of NavigateToFileTest selenium test


SkorikSergey:
### Description
We have unexpected fail of **NavigateToFileTest** selenium test. The test sometimes fails when waiting suggestion list with found files from the **Navigate To File** feature.

**Log:**
```
org.openqa.selenium.WebDriverException: 
unknown error: Element <td style="opacity: 0.6;">...</td> is not clickable at point (895, 300). Other element would receive the click: <div class="GIWAMGHBPOB" style="right: 0px; bottom: 0px; position: absolute; left: 0px; top: 0px;">...</div>
  (Session info: chrome=56.0.2924.87)
  (Driver info: chromedriver=2.27.440175 (9bc1d90b8bfa4dd181fbbf769a5eb5e575574320),platform=Linux 3.10.0-514.16.1.el7.x86_64 x86_64) (WARNING: The server did not provide any stacktrace information)
Command duration or timeout: 129 milliseconds
Build info: version: '3.1.0', revision: '86a5d70', time: '2017-02-16 07:57:44 -0800'
System info: host: 'bf642ff2277b', ip: '172.18.0.3', os.name: 'Linux', os.arch: 'amd64', os.version: '3.10.0-514.16.1.el7.x86_64', java.version: '1.8.0_121'
Driver info: org.openqa.selenium.chrome.ChromeDriver
Capabilities [{applicationCacheEnabled=false, rotatable=false, mobileEmulationEnabled=false, networkConnectionEnabled=false, chrome={chromedriverVersion=2.27.440175 (9bc1d90b8bfa4dd181fbbf769a5eb5e575574320), userDataDir=/tmp/.org.chromium.Chromium.zvskk5}, takesHeapSnapshot=true, pageLoadStrategy=normal, databaseEnabled=false, handlesAlerts=true, hasTouchScreen=false, version=56.0.2924.87, platform=LINUX, browserConnectionEnabled=false, nativeEvents=true, acceptSslCerts=true, locationContextEnabled=true, webStorageEnabled=true, browserName=chrome, takesScreenshot=true, javascriptEnabled=true, cssSelectorsEnabled=true, unexpectedAlertBehaviour=}]
Session ID: 33994dcb-678a-4b12-adc2-a997bfb381c4
	at org.eclipse.che.selenium.miscellaneous.NavigateToFileTest.launchNavigateToFileAndCheckResults(NavigateToFileTest.java:149)
[...]

PR #9585, opened 2018-05-02
 Selenium: change url of known issue to #8465 instead of #8293 in NavigateToFileTest test 

SkorikSergey:
**What does this PR do?**
- Change the url of the known issue in the try/catch block in the **NavigateToFileTest**.


dmytro-ndp:
According to convention commit message should be kind of infinitive: _To `change known issue from #8293 to #8465 in NavigateToFileTest`_

dmytro-ndp (approved):

PR #9598, opened 2018-05-03
Build any Theia extensions which are downloaded from a git provider

mmorhun:
### What does this PR do?
Makes build of all Theia plugins which are downloaded form a git provider
Fixes typos

### DIFF
--- a/dockerfiles/theia/src/main.js
+++ b/dockerfiles/theia/src/main.js
@@ -140,7 +140,7 @@ function addPluginFromGitRepository(plugins, pluginName, gitRepository) {
         }
     }
 
-    if (!fs.existsSync(pluginPath + 'node_mudules') || !fs.existsSync(pluginPath + 'lib')) {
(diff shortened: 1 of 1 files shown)


evidolob (approved):

dockerfiles/theia/src/main.js
```diff
@@ -123,7 +123,7 @@ function getExtraTheiaPlugins() {
 
 /**
  * Clones git repository and adds the plugin as local.
- * This is done becouse Theia requires npm package.json (not lerna) which is located, in most cases, in subdirectory of repository.
+ * This is done becouse Theia requires npm package.json (not lerna) which is located, in some cases, in subdirectory of repository.
```
benoitf: there is a typo on "because/becouse"

benoitf (approved):

dockerfiles/theia/src/main.js
mmorhun: Thanks

PR #9608, opened 2018-05-04
Allow adding Theia extension on image build from within container

mmorhun:
### What does this PR do?
Improves script which adds extensions into Theia on Theia IDE image build to be able to use extensions from container (before we supported only git provided extensions). So now we can use:
```
extension-name:git://github.com/user/extension.git
extension-name:file:///path/to/extension-root
```
as arguments to the `add-extensions` script in the Dockerfile.

Related to
https://github.com/eclipse/che/pull/9216


benoitf (approved):

benoitf:
we may change as well THEIA_PLUGINS to THEIA_EXTENSIONS to not confuse users

PR #9684, opened 2018-05-11
Remove Theia Dockerfile from Che repo

ghost:
### What does this PR do?

It makes sense to move Theia Dockerfile and associated scripts to dockerfiles repo, since dockerfiles here is used for CLI images and are subjected to CI builds. Also, the image is tagged with Che version which does not make any sense.
 
PR to Dockerfiles: https://github.com/eclipse/che-dockerfiles/pull/183

Once merged, CI will build and tag Theia image with Theia release tag.


benoitf:
you're saying that lifecycle of theia is not linked to the lifecycle of Che.
but I assume that using 'latest' for any Che version will probably fail as theia plugins may depend on a specific version of the che about the API.

eivantsov:
We will use tagged Theia images only.

benoitf:
so theia image will have a new tag for each theia release ?

I was talking about the che-extensions developed on top of theia that may work only for a given version of che.

eivantsov:
Yes, correct! eclipse/che-theia:0.3.7 is the first we will have.

ashumilova (approved):

benoitf (changes requested):
I definitely still don't get why it should be removed from che as it will include some che extensions that are bound to the version of che (and not only theia) so it needs to be part of che.

riuvshin (approved):

eivantsov:
1. dockerfiles directory contains Che CLI images
2. eclipse/che-theia:nightly has been broken for over a month and it was obviously ok (I have not see any issues related to that)
3. ssh extension has been removed since it does not work with 0.3.7 and we cannot yet use 0.3.8 as far as I understand.
4. what Che extensions in this **broken** Theia image are bound to Che server version?

benoitf:
1. I'm not against moving theia stuff to another directory than `dockerfiles` if we aim to keep only CLI stuff.

2. I guess nobody using it also there is another temporary image based upon work in progress for plugins on top of theia.

3+4: If I recap correctly, Some che extensions need to be part of the docker image so it is not waiting a couple or more minutes when starting a workspace.
Some of these extensions will use `@eclipse-che/workspace-client` library which is a library that is communicating with Che workspaces. So somehow it's matching a specific che version.

Dependency graph is then
Che --> depends on the `Che/theia` image
And this `Che/theia`image will also depends on some extensions that are depending on a specific theia version but also a specific che version.

So Che/theia image is linked to the che version and not only a theia version

And if che/theia image would not depend on any Che specific version stuff (which is not the case), then it would mean it could be part of theia repositories as there would have no need to have some specific for che (so it wouldn't go to `che-dockerfiles` repository either)

eivantsov:
Strongly disagree - Che tag in Theia image is useless and it actually resulted in the situation we are now in: we DO NOT KNOW what Theia core is currently in the image and if or not Theia and Che extensions that we decided to use for default Theia works fine with that version of core.

benoitf:
> Che tag in Theia image is useless and it actually resulted in the situation we are now in: we DO NOT KNOW what Theia core is currently in the image 

why you write it in uppercase ? BTW you know the version as it's part of the image
Could you describe why you need to have a tag which is the theia version ?

> if or not Theia and Che extensions that we decided to use for default Theia works fine with that version of core.

Che extensions would need to be packaged upon build time.
The dynamic extensions are called plugins and then it's not need "build operation" as it can be loaded dynamically

ashumilova:
At the moment as far as I know - we do not have Che specific stuff in che/theia image. The script works with container's environment variables and package.json, nothing with workspaces and che itself. @mmorhun rechecking.

benoitf:
"at the moment" yes but what is coming is that some extensions will be precompiled during the build of the docker image.
https://github.com/theia-demo-plugins/wiptheia-docker-image/blob/master/Dockerfile#L25-L28

or what you're suggesting is that we'll wait 2 or 3 minutes before having a workspace ready ? (and maybe not if npm registry is offline or if there is network issue) ?

eivantsov:
OK. I then let you guys decide and fix it. Feel free to close.

eivantsov:
@benoitf @ashumilova I am closing the PR then as there're no reasons to keep it if Theia Dockerfiles and associated scripts are to remain here in this repo.

You may grab the actual Dockerfile and scripts from https://github.com/eclipse/che-dockerfiles/pull/183 and then open a new PR against this repo.

Don't forget to notify @riuvshin about changes so that CI tags Theia image the way @benoitf considers appropriate for current situation and any hypothetical situations with changing workspace model etc.

Also, I highly recommend not just exercise your committer powers when you think something is wrong but actually make sure that this image is in a good state - perhaps setup the right CI for it, or even contribute to selenium tests to make sure Theia isn't just built but can actually start. Thank you.

benoitf:
thanks @eivantsov for your detailed comment

Issue #9542, opened 2018-04-25
Workspaces can not be started after change image in the "Machine" section


Ohrimenko1988:
### Reproduction Steps
 - Create workspace from default "Java" stack 
 - Open "Workspaces" page
 - In the workspaces list select just created workspace and click on it
 - In the opened "workspace details" page click on the "Machine" button
 - Click on the "Edit" button 
 - In the "Recipe" text field delete several letters in the image name, it should be "eclipse/ubuntu_jd"
 - Click on the "Save" button
 - Click on the "Run" button
 - Pay attention to the error message
 - Try to create and run workspace with any stack
 - Pay attention to the errors

**Expected conditions:**
  Workspace with wrong image is not started but test workspace is started successfully 

**Observed behavior:**
Workspace with wrong image is not started and test workspace has the same error  

**Attachment:**
![screencast-6_1](https://user-images.githubusercontent.com/22724619/39252751-01f3123a-48af-11e8-82d3-fd3d93263e7a.gif)

**OS and version:**    
Che 6.5.0 ; Fedora 25 ; Chrome 63.0

sleshchenko:
I think it is a server side issue. It is related to https://github.com/eclipse/che/pull/9426.
There is an issue with container watching when all workspaces' objects are created in one namespace: In such case, it is required to filter pods and propagate only events which are related a particular workspace.

A similar scenario but with one workspace has two issues:
<details>
<summary>Reproduction Steps</summary>

1. Create workspace from default "Java" stack.
2. Open "Workspaces" page.
3. In the workspaces list select just created workspace and click on it.
4. In the opened "workspace details" page click on the "Machine" button.
5. Click on the "Edit" button.
6. In the "Recipe" text field delete several letters in the image name, it should be "eclipse/ubuntu_jd".
7. Click on the "Save" button.
8. Click on the "Run" button.
9. Pay attention to the error message.
10. Fix machine's image to a correct one.
11. Rerun the workspace.
12. Pay attention to the errors.
</details>

1. OpenShift Infrastructure doesn't support start interruption, because of that the second start try fails with error message: `Runtime is already started`. Also because of that, it is not possible to delete failed workspace. Will be fixed by https://github.com/eclipse/che/issues/5918.
[...]

PR #9703, opened 2018-05-15
#9542 Proper handling of unrecoverable events. Processing only events which are related to current workspace pods

ibuziuk:
### What does this PR do?
Proper handling of unrecoverable events. Processing only events which are related to current workspace pods

#9542

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
@@ -22,10 +22,12 @@ import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.api.model.Service;
 import io.fabric8.kubernetes.api.model.extensions.Ingress;
 import io.fabric8.kubernetes.client.Watcher.Action;
+import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -72,6 +74,7 @@ import org.eclipse.che.workspace.infrastructure.kubernetes.namespace.event.Conta
 import org.eclipse.che.workspace.infrastructure.kubernetes.namespace.event.PodActionHandler;
 import org.eclipse.che.workspace.infrastructure.kubernetes.namespace.pvc.WorkspaceVolumesStrategy;
 import org.eclipse.che.workspace.infrastructure.kubernetes.server.KubernetesServerResolver;
+import org.eclipse.che.workspace.infrastructure.kubernetes.util.ContainerEvents;
 import org.eclipse.che.workspace.infrastructure.kubernetes.util.KubernetesSharedPool;
 import org.eclipse.che.workspace.infrastructure.kubernetes.util.RuntimeEventsPublisher;
 import org.slf4j.Logger;
@@ -432,10 +435,8 @@ public class KubernetesInternalRuntime<
     // namespace.pods().watch(new AbnormalStopHandler());
     namespace.pods().watchContainers(new MachineLogsPublisher());
     if (!Strings.isNullOrEmpty(unrecoverableEvents)) {
-      // The handler for unrecoverable events is disabled because it has a bug that prevents start
-      // of workspaces after previous workspace start failure.
-      // See https://github.com/eclipse/che/issues/9542
-      // namespace.pods().watchContainers(new UnrecoverableEventHanler());
+      Map<String, Pod> pods = getContext().getEnvironment().getPods();
+      namespace.pods().watchContainers(new UnrecoverableEventHanler(pods));
     }
 
     final KubernetesServerResolver serverResolver =
@@ -654,37 +655,74 @@ public class KubernetesInternalRuntime<
   /** Listens container's events and terminates workspace if unrecoverable event occurs. */
   public class UnrecoverableEventHanler implements ContainerEventHandler {
     private List<String> events;
+    private Map<String, Pod> workspacePods;
+    private Date handlerInitialization;
 
-    public UnrecoverableEventHanler() {
+    public UnrecoverableEventHanler(Map<String, Pod> workspacePods) {
       this.events =
           Strings.isNullOrEmpty(unrecoverableEvents)
               ? Collections.EMPTY_LIST
               : Arrays.asList(unrecoverableEvents.split(","));
+      this.workspacePods = workspacePods;
+      this.handlerInitialization = new Date();
     }
 
+    /*
+     * Event is considered to be unrecoverable if it belongs to one of the workspace pods
+     * and 'lastTimestamp' of the event is *after* the time of handler initialization
+     */
     @Override
     public void handle(ContainerEvent event) {
-      if (isUnrecoverable(event)) {
+      if (isWorkspaceEvent(event)
+          && happenedAfterHandlerInitialization(event)
+          && isUnrecoverable(event)) {
         String reason = event.getReason();
         String message = event.getMessage();
-        String podName = event.getPodName();
         String workspaceId = getContext().getIdentity().getWorkspaceId();
         LOG.error(
             "Unrecoverable event occurred during workspace '{}' startup: {}, {}, {}",
             workspaceId,
             reason,
             message,
-            podName);
+            event.getPodName());
         try {
           internalStop(emptyMap());
         } catch (InfrastructureException e) {
           LOG.error("Error occurred during runtime '{}' stopping. {}", workspaceId, e.getMessage());
         } finally {
           eventPublisher.sendRuntimeStoppedEvent(
-              format("Unrecoverable event occurred: '%s', '%s', '%s'", reason, message, podName),
+              format(
+                  "Unrecoverable event occurred: '%s', '%s', '%s'",
+                  reason, message, event.getPodName()),
               getContext().getIdentity());
         }
       }
+    }
+
+    /** Returns true if event belongs to one of the workspace pods, false otherwise */
+    private boolean isWorkspaceEvent(ContainerEvent event) {
+      String podName = event.getPodName();
+      return workspacePods.containsKey(podName);
+    }
+
+    /**
+     * Returns true if 'lastTimestamp' of the event is *after* the time of the handler
+     * initialization
+     */
+    private boolean happenedAfterHandlerInitialization(ContainerEvent event) {
+      String eventLastTimestamp = event.getLastTimestamp();
+      Date eventLastTimestampDate = convertEventTimestampToDate(eventLastTimestamp);
+      return eventLastTimestampDate != null && eventLastTimestampDate.after(handlerInitialization);
+    }
+
+    private Date convertEventTimestampToDate(String timestamp) {
+      Date date = null;
+      try {
+        date = ContainerEvents.convertEventTimestampToDate(timestamp);
+      } catch (ParseException e) {
+        LOG.error("Error occurred during parsing the event timestamp '{}'", timestamp);
+      }
+      return date;
     }
 
     /**
@@ -728,7 +766,10 @@ public class KubernetesInternalRuntime<
           if (machine.getPodName().equals(podName)
               && machine.getContainerName().equals(containerName)) {
             eventPublisher.sendMachineLogEnvent(
-                entry.getKey(), event.getMessage(), event.getTime(), getContext().getIdentity());
+                entry.getKey(),
+                event.getMessage(),
+                event.getCreationTimeStamp(),
+                getContext().getIdentity());
             return;
           }
         }
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/util/ContainerEvents.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/util/ContainerEvents.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) 2012-2018 Red Hat, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.workspace.infrastructure.kubernetes.util;
+
+import java.text.DateFormat;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import org.eclipse.che.workspace.infrastructure.kubernetes.namespace.event.ContainerEvent;
+
+/**
+ * Helps to simplify the interaction with the {@link ContainerEvent}.
+ *
+ * @author Ilya Buziuk
+ */
+public final class ContainerEvents {
+  private static DateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssX");
+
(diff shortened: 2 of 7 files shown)


sleshchenko:
Looks like the issue with old events is not fixed https://github.com/eclipse/che/issues/9542#issuecomment-384575071

> 2. OpenShift sends events that happened before (like 20 minutes ago). It means that even when image reference will be fixed workspace start will fails because of an old event from history. In this way, Che Server considers workspace as failed to start when it receives old events. REST API allows specifying resource version or time (I'm not sure what exactly) for filtering old events. But Fabric8 client has deprecated method watch(String resourceVerion, W watcher). Need to investigate how to fix it.

Is it actual? If yes, should we fix it before enabling unrecoverable events functionality?

ibuziuk:
@sleshchenko good point. So currently events (even old) would be processed only for concrete workspace. e.g. the following scenario would work just fine:

- start workspace with stack referring to non-existing image
- startup failed
- start another workspace with stack refering to existing image
- startup succeed 

However, it looks like the following scenario still does not work properly:
- start workspace with stack refering to non-existing image
- startup failed
- fix image in workspace config to point to the existing image
- ERROR: startup failed due to previous unrecoverable event.

In terms of fixing it upstream: the plan is to create a fix with filtering events on Che side and also create an issue for proper filtering of the events for k8s client upstream (from the previous experience it might take months to get the fix, so it would be relevant to do filtering on wsmaster side). WDYT ?

sleshchenko:
@ibuziuk 
> it would be relevant to do filtering on wsmaster side

Makes sense. :+1: 
Maybe it would be better to merge your changes along with filtering by time.
Or your changes can be merged separately but with disabled unrecoverable events functionality.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
```diff
@@ -654,20 +652,22 @@ public void accept(ProbeResult probeResult) {
...
               ? Collections.EMPTY_LIST
               : Arrays.asList(unrecoverableEvents.split(","));
+      this.workspacePods = workspacePods;
     }
 
     @Override
     public void handle(ContainerEvent event) {
-      if (isUnrecoverable(event)) {
+      String podName = event.getPodName();
+      if (isUnrecoverable(event) && workspacePods.containsKey(podName)) {
```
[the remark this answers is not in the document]
ibuziuk: this is what I'm trying to do now ;-)

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntimeTest.java
```diff
@@ -432,7 +432,7 @@ public void testRepublishContainerOutputAsMachineLogEvents() throws Exception {
   public void testHandleUnrecoverableEventByReason() throws Exception {
     final String unrecoverableEventReason = "Failed Mount";
     final UnrecoverableEventHanler unrecoverableEventHanler =
-        internalRuntime.new UnrecoverableEventHanler();
+        internalRuntime.new UnrecoverableEventHanler(k8sEnv.getPods());
```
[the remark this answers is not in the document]
ibuziuk: sure

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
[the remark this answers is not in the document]
ibuziuk: well, passing `k8sEnv.getPods()` is more straight-forward IMO than manipulations with `keySet` IMO, also params in tests would be reused. Do not have strong opinion here though. If you think it is important I could change

ibuziuk:
@sleshchenko yeah, I plan to provide time filtering in this PR also ;-)

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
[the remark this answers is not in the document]
garagatyi: I'm OK with either too, so just leave it as it is now to proceed with the PR faster

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/util/ContainerEvents.java
```diff
@@ -0,0 +1,46 @@
...
+    timestamp = timestamp.replaceAll("[^\\d-:]", " ").trim();
+    Date date = dateFormat.parse(timestamp);
+    return date;
+  }
+
+  /**
+   * Converts the {@link Date} to {@link ContainerEvent} timestamp format e.g.
+   * '2018-05-15T16:17:54Z'
+   */
+  public static String convertDateToEventTimestamp(Date date) throws ParseException {
```
sleshchenko: Looks like ParseException can't be thrown here

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
```diff
@@ -654,20 +655,34 @@ public void accept(ProbeResult probeResult) {
...
+     * and 'lastTimestamp' of the event is *after* the the time of handler initialization
+     */
     @Override
     public void handle(ContainerEvent event) {
-      if (isUnrecoverable(event)) {
+      String podName = event.getPodName();
+      String eventLastTimestamp = event.getLastTimestamp();
+      Date eventLastTimestampDate = convertEventTimestampToDate(eventLastTimestamp);
+
+      if (isUnrecoverable(event)
```
sleshchenko: I think it may sense to reorder conditions here. 
```java
 if (workspacePods.containsKey(podName) // if event is not related to workspaces pods it doesn't make sense to check other conditions
   && eventLastTimestampDate != null
   && eventLastTimestampDate.after(handlerInitialization) // if event happened before initialization it doesn't make sense to check is it recoverable or not
   && isUnrecoverable(event)) {
```
It doesn't bring so much performance benefits but still :smiley:

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
```diff
@@ -654,20 +655,34 @@ public void accept(ProbeResult probeResult) {
...
+          && workspacePods.containsKey(podName)
+          && eventLastTimestampDate != null
+          && eventLastTimestampDate.after(handlerInitialization)) {
```
sleshchenko: I think it makes sense to perform this check in `KubernetesPods#watchContainers`. Then you'll solve even more issues like publishing to client old events by MachineLogsPublisher.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/util/ContainerEvents.java
```diff
@@ -0,0 +1,46 @@
...
+  private static DateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
+
+  private ContainerEvents() {}
+
+  /**
+   * Converts the time of {@link ContainerEvent} e.g. '2018-05-15T16:17:54Z' to the {@link Date}
+   * format
+   */
+  public static Date convertEventTimestampToDate(String timestamp) throws ParseException {
+    timestamp = timestamp.replaceAll("[^\\d-:]", " ").trim();
```
garagatyi: I think it is possible to simplify parsing by using date format `"yyyy-MM-dd'T'HH:mm:ssX"` and not replacing anything.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
garagatyi: Since null appears only when a date is not parsable I doubt it makes much sense to optimize )

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
garagatyi: +1

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
sleshchenko: I'd say that main improvement (main but little one) is putting `workspacePods.containsKey(podName)` at the begging. Because containsKey in hash map is cheaper operation than sequential check list items with start with.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
ibuziuk: @sleshchenko @garagatyi true, are you ok to add `Date getInitializationDate();` to `ContainerEventHandler` interface ?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
ibuziuk: good point

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
sleshchenko: Sorry, but I don't understand why do you need it. I think you can save initialization date here https://github.com/eclipse/che/blob/ac3b4016492fcfb9361243bc708f1191d06c1adf/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesPods.java#L309 and do not publish events which happened before.

ibuziuk: so, you mean to check `Watcher` not `Handler` initialization date? or you mean to save Handler initialization date when adding handler to the `containerEventsHandlers` queue [1] ?

[1] https://github.com/eclipse/che/blob/ac3b4016492fcfb9361243bc708f1191d06c1adf/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesPods.java#L346

sleshchenko: > you mean to check Watcher not Handler initialization date?

yes. I think it would be good enough. We don't need any initialization data on Handler level. Handler only needs to handle **new** events. 

Offtop :smile: 
It is strange that we don't face the similar issue with `KubernetesPods#waitAsync` method. Maybe events are clear after pod removing or something like that.
Hmmm. I see. Everything is fine with `KubernetesPods#waitAsync` because of using a resourceVersion in BaseOperation.java 
```
    public Watch watch(Watcher<T> watcher) throws KubernetesClientException {
        return this.watch(this.resourceVersion, watcher);
    }
```
Looks like it doesn't work for all Events because we don't have a particular object with specified resourceVersion while subscribing.

ibuziuk: @sleshchenko well, I'm not sure that comparing `lastTimestamp` of event with Watcher initialization is a good idea. It would work currently since 2 handlers we have & watcher are initialized at roughly the same time. But imagine the situation that some new handler would be initialized differently - this would results in subtle bugs & need of rewriting watcher logic

sleshchenko: @ibuziuk 
Well, I expect that all old events will be processed and skipped immediately after establishing the WebSocket connection. So then:
1. The first handler is registered.
2. Watcher is initialized and subscribed, he receives all events that already happens and skips them. He publishes only events that happened after initialization.
3. The second handler is registered. It does not receive old events because a WebSocket connection is already established and all old events are already processed / skipped.

So nobody does not get an event that happened before handler initialization.

Do you mean an issue that there is a probability that the second handler will receive an event that occurred between watcher initialization and handler initialization? But why can it happen? Shouldn't a watcher receive an event just when it happened if a WebSocket connection is established?

ibuziuk: @sleshchenko ah, ok it makes total sense. let me rework to support this

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntimeTest.java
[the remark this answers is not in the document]
ibuziuk: done

ibuziuk:
@garagatyi @sleshchenko guys, are you ok to review/merge current PR since it is ready with tests and I will provide fix for filtering events on the watcher level in another one ?

sleshchenko (approved):
I'm OK to merge current PR if there will be another PR with filtering events on the watcher level.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
```diff
@@ -654,39 +655,80 @@ public void accept(ProbeResult probeResult) {
...
+                  "Unrecoverable event occurred: '%s', '%s', '%s'",
+                  reason, message, event.getPodName()),
               getContext().getIdentity());
         }
       }
     }
 
+    /**
+     * @param event
+     * @return true if event belongs to one of the workspace pods, false otherwise
```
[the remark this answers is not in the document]
ibuziuk: @sleshchenko lemme rewrite, but this is the doc Eclipse generates for me ;-)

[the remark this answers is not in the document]
ibuziuk: version with `@return` is better parsed in doc pop-up IMO
image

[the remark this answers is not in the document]
ibuziuk: I have fixed since it is commonly used java doc format in project ;-)

garagatyi (approved):
Looks good. Nice fix!

eivantsov:
@ibuziuk and you have got a docs PR waiting to be merged https://github.com/eclipse/che-docs/pull/397

Is it still relevant?

ibuziuk:
@eivantsov thanks for the pointer, I thought this one has been already merged ;-)

davidfestal (approved):

ibuziuk:
ci-test

ibuziuk:
ci-test

Issue #9126, opened 2018-03-19
Rework ws-agent activity notification mechanism


akorneta:
### Description
In cases when notification scheduler does not notify about activity till the _current time_ >= _last update time_ + _threshold_ then the activity notification will be sent in the same thread in which this activity occurred. Such behaviour may lead to problems because the request for activity has default timeout 1 minute, and if the master is not available for this period it will block the activity thread for one minute. The activity thread its tomcat thread where `LastAccessTimeFilter` call `WorkspaceActivityNotifier.onActivity()`, and during this period the client won't receive a response from the agent.

The problem is not abstract, it occurs while recovery on docker infrastructure.
- on post-construct ws-master recovers runtime
- then checks servers with request timeout 3 sec
- server checks produce activity and wait for 3 sec till agent notifies about activity master(master) in the same thread after timeout is exceeded server checks fail and workspaces considered as stopped.

artaleks9:
Still actual for ver.6.5.0

PR #9708, opened 2018-05-16
Send request with activity on ws-agent in a separate thread

mkuznyetsov:
### What does this PR do?
On WS-Agent, sending requests will be performed in separate threads, to prevent blocking the tomcat thread if request takes too long to complete.

https://github.com/eclipse/che/issues/9126

### DIFF
--- a/wsagent/che-core-workspace-activity-server/src/main/java/org/eclipse/che/workspace/activity/WorkspaceActivityNotifier.java
+++ b/wsagent/che-core-workspace-activity-server/src/main/java/org/eclipse/che/workspace/activity/WorkspaceActivityNotifier.java
@@ -53,9 +53,14 @@ public class WorkspaceActivityNotifier {
   /**
    * Notify workspace master about activity in this workspace.
    *
-   * <p>After last notification, any consecutive activities that come within specific amount of time
-   * - {@code threshold}, will not notify immediately, but trigger notification in scheduler method
-   * {@link WorkspaceActivityNotifier#scheduleActivityNotification}
+   * <p>There is a {@code threshold} that limits request sending frequency - if no activity has
+   * occurred within the duration of threshold, then the request will be sent immediately (in a
+   * separate thread, to prevent it's possible long execution from blocking of the Tomcat thread,
+   * from which it may be called).
+   *
+   * <p>In case there was any recent activity within threshold time, any consecutive activities will
+   * not send request immediately, but trigger it's sending in scheduler method {@link
+   * WorkspaceActivityNotifier#scheduleActivityNotification}
    */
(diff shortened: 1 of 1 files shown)


wsagent/che-core-workspace-activity-server/src/main/java/org/eclipse/che/workspace/activity/WorkspaceActivityNotifier.java
```diff
@@ -75,10 +75,17 @@ private void scheduleActivityNotification() {
   }
 
   private void notifyActivity() {
-    try {
-      httpJsonRequestFactory.fromUrl(apiEndpoint + "/activity/" + wsId).usePutMethod().request();
-    } catch (Exception e) {
-      LOG.error("Cannot notify master about workspace " + wsId + " activity", e);
-    }
+    new Thread(
```
sleshchenko: Should not be this thread daemon?

sleshchenko: We need new Thread only when `onActivity` method is called but not when `scheduleActivityNotification`. Then maybe it would be better to have two methods: `notifyActivity` and `notifyActivityAsync`?

sleshchenko (approved):

skabashnyuk:
@mkuznyetsov  could you update ```onActivity``` javadoc according to the new changes?

mkuznyetsov:
ok

wsagent/che-core-workspace-activity-server/src/main/java/org/eclipse/che/workspace/activity/WorkspaceActivityNotifier.java
```diff
@@ -53,16 +53,24 @@ public WorkspaceActivityNotifier(
...
+   * WorkspaceActivityNotifier#scheduleActivityNotification}
    */
   public void onActivity() {
     long currentTime = System.currentTimeMillis();
     if (currentTime < (lastUpdateTime + threshold)) {
       activeDuringThreshold.set(true);
     } else {
-      notifyActivity();
+      Thread activityRequestThread = new Thread(this::notifyActivity);
+      activityRequestThread.setDaemon(true);
```
skabashnyuk: give some meaningful name to this thread.

mkuznyetsov: ok

mkuznyetsov:
ci-test

skabashnyuk (approved):

Issue #9292, opened 2018-03-29
Workspace.Next/Phase 1 - CheService/CheFeature model


skabashnyuk:
### Description
In this issue, I would like to discuss CheService/CheFeature model before we start to implement it.
![tl4zjygm3dtzapa3iz9lfvdeoc51oc1smwj4vaeodjaw---4j4thyhigl_ozfzklpr5aps73u574l2v-b1a7cbrt-q2dqybywp6f1bnpzlb9u0e0cxn8dwarcssoy84gm8z5flynnvvyhjdo4mzxjb4vjdw85pvsdrcswrkmwrvojnn9itgehmzeh5ahaat14n_jnjfujtthfqptkbhqatwkgq5tm8bwfqtwxoyluc9-wv9y](https://user-images.githubusercontent.com/1614429/38088672-45adae0e-3365-11e8-829b-6db2f0e52885.png)

### Service Model
CheServices combines kubernetes Pods and Services to define servers running inside of containers. It can be referenced, for instance in features by name and version.  A pod can define environment variables. Value of variables can be templated with ${VARIABLE_NAME} syntax. Value of variable can be combined in feature level. See more in feature description. 
Service may specify Volumes in PodSpec and volumeMounts in ContainerSpec then infrastructure will provide PVC.
PodSpec#Volumes#persistentVolumeClaim#ClaimName will be used of Che volume identifier, and if two different services specify the same claim names, it means that they will use shared Che volume (PVC).
So, services should specify `projects` as claimName to mount Projects Sources. Another example is naming claim as m2 to mount ~/.m2 folder.

```
apiVersion: stable.che.eclipse.org/v1
kind: CheService
metadata:
 name: io.typefox.theia-ide.che-service
spec:
 services:
   - apiVersion: v1
     kind: Service
     metadata:
       name: io.typefox.theia-ide.che-service.mainservice
     spec:
       selector:
         app: theia
       ports:
       - protocol: TCP
         port: 80
         targetPort: 9376
 pods:
   - apiVersion: v1
     kind: Pod
     metadata:
       name: io.typefox.theia-ide.che-service.mainpod
     spec:
       selector:
         app: theia
       containers:
         - image: eclipse/che-theia:nightly
           labels:
             che.eclipse.org/container-name: theia-core
           name: theia
           env:
             - name: THEIA_PLUGINS
               value: ${THEIA_PLUGINS}
           volumeMounts:
             - mountPath: "/projects"
               name: projects   
[...]

PR #9774, opened 2018-05-22
Add an ability to try Workspace.Next features

garagatyi:
This PR is designed to be reviewed in a commit by commit manner. 

It is in WIP state because not all the unit tests are finished. 
### What does this PR do?
Add an ability to try Workspace.Next features. 
To try it out user needs to add an attribute to a workspace with name `features` and a list of features as a value.
An example of value: `"org.eclipse.che.che-theia-hello/0.0.1,org.eclipse.che.che-theia-ssh/0.0.1"`
This PR contains code to use Workspace.Next with the kubernetes recipe only. 
Support of other recipes should be added additionally.

### DIFF
--- a/deploy/kubernetes/kubectl/Deploy Che.md
+++ b/deploy/kubernetes/kubectl/Deploy Che.md
@@ -33,3 +33,7 @@ and clean redeploy of Che.
 - Create namespace `che`: `kubectl create namespace che`
 - Deploy Che: `kubectl --namespace=che apply -f che-kubernetes.yaml`
 - Check Che pod status until it become `Running`: `kubectl get --namespace=che pods`
+
+### Workspace Next:
+There is a file `wsnext/feature-api.yaml` which contains kubesrv service needed to test Workspace Next on kubernetes infrastructure.
+To use it call `kubectl apply --namespace=che -f wsnext/feature-api.yaml` and Che will be able to use Workspace Next flow.
--- a/deploy/kubernetes/kubectl/che-kubernetes.yaml
+++ b/deploy/kubernetes/kubectl/che-kubernetes.yaml
@@ -72,6 +72,7 @@ items:
     CHE_LOGS_APPENDERS_IMPL: "plaintext"
     CHE_INFRA_KUBERNETES_INGRESS_DOMAIN: "192.168.99.100.nip.io"
     CHE_INFRA_KUBERNETES_SERVER__STRATEGY: "default-host"
+    CHE_WORKSPACE_FEATURE_API: "http://feature-api-service:3000"
 - apiVersion: extensions/v1beta1
   kind: Ingress
   metadata:
@@ -269,6 +270,11 @@ items:
             valueFrom:
               configMapKeyRef:
                 key: CHE_INFRA_KUBERNETES_SERVER__STRATEGY
+                name: che
+          - name: CHE_WORKSPACE_FEATURE_API
+            valueFrom:
+              configMapKeyRef:
+                key: CHE_WORKSPACE_FEATURE_API
                 name: che
           image: eclipse/che-server:nightly
           imagePullPolicy: Always
--- a/deploy/kubernetes/kubectl/che-kubernetes.yaml
+++ b/deploy/kubernetes/kubectl/che-kubernetes.yaml
@@ -70,7 +70,7 @@ items:
     CHE_WORKSPACE_AUTO_START: "false"
     CHE_INFRA_KUBERNETES_INGRESS_ANNOTATIONS__JSON: '{"nginx.ingress.kubernetes.io/rewrite-target": "/","nginx.ingress.kubernetes.io/ssl-redirect": "false","nginx.ingress.kubernetes.io/proxy-connect-timeout": "3600","nginx.ingress.kubernetes.io/proxy-read-timeout": "3600"}'
     CHE_LOGS_APPENDERS_IMPL: "plaintext"
-    CHE_INFRA_KUBERNETES_INGRESS_DOMAIN: "192.168.99.101.nip.io"
+    CHE_INFRA_KUBERNETES_INGRESS_DOMAIN: "192.168.99.100.nip.io"
     CHE_INFRA_KUBERNETES_SERVER__STRATEGY: "default-host"
 - apiVersion: extensions/v1beta1
   kind: Ingress
--- a/deploy/kubernetes/kubectl/wsnext/feature-api.yaml
+++ b/deploy/kubernetes/kubectl/wsnext/feature-api.yaml
@@ -1,0 +1,85 @@
+---
+apiVersion: v1
+kind: List
+items:
+- apiVersion: v1
+  kind: Service
+  metadata:
+    labels:
+      app: feature-api
+    name: feature-api-service
+  spec:
+    ports:
+    - name: feature-api
+      port: 3000
+      protocol: TCP
+      targetPort: 3000
+    selector:
+      app: feature-api
+- apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    labels:
+      app: feature-api
+    name: feature-api
+  data:
+    CHE_REGISTRY_UPDATE_INTERVAL: "60"
+    CHE_REGISTRY_GITHUB_URL: "https://github.com/garagatyi/che-registry.git"
+- apiVersion: extensions/v1beta1
+  kind: Ingress
+  metadata:
+    name: feature-api-ingress
+    annotations:
+      nginx.ingress.kubernetes.io/rewrite-target: /
+      nginx.ingress.kubernetes.io/proxy-read-timeout: "3600"
+      nginx.ingress.kubernetes.io/proxy-connect-timeout: "3600"
+  spec:
+    rules:
+    - host: feature.192.168.99.100.nip.io
+      http:
+        paths:
+        - backend:
+            serviceName: feature-api-service
+            servicePort: 3000
+- apiVersion: extensions/v1beta1
+  kind: Deployment
+  metadata:
+    labels:
+      app: feature-api
+    name: feature-api
+  spec:
+    replicas: 1
+    revisionHistoryLimit: 2
+    selector:
+      matchLabels:
+        app: feature-api
+    strategy:
+      type: Recreate
+    template:
+      metadata:
+        labels:
+          app: feature-api
+      spec:
+        containers:
+        - env:
+          - name: CHE_REGISTRY_UPDATE_INTERVAL
+            valueFrom:
+              configMapKeyRef:
+                key: CHE_REGISTRY_UPDATE_INTERVAL
+                name: feature-api
+          - name: CHE_REGISTRY_GITHUB_URL
+            valueFrom:
+              configMapKeyRef:
+                key: CHE_REGISTRY_GITHUB_URL
+                name: feature-api
+          image: garagatyi/kubesrv:latest
+          imagePullPolicy: Always
+          name: feature-api
+          ports:
+          - containerPort: 3000
+            name: feature-api
+          resources:
+            limits:
+              memory: 600Mi
+            requests:
+              memory: 256Mi
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsnext/KubernetesWorkspaceNextApplier.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsnext/KubernetesWorkspaceNextApplier.java
@@ -1,0 +1,182 @@
+/*
+ * Copyright (c) 2012-2018 Red Hat, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.workspace.infrastructure.kubernetes.wsnext;
+
+import static java.util.Collections.singletonMap;
+import static org.eclipse.che.api.core.model.workspace.config.MachineConfig.MEMORY_LIMIT_ATTRIBUTE;
+
+import com.google.common.annotations.Beta;
+import io.fabric8.kubernetes.api.model.ContainerBuilder;
+import io.fabric8.kubernetes.api.model.Pod;
+import io.fabric8.kubernetes.api.model.Quantity;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import javax.inject.Inject;
+import javax.inject.Named;
+import org.eclipse.che.api.core.model.workspace.config.ServerConfig;
+import org.eclipse.che.api.workspace.server.model.impl.ServerConfigImpl;
+import org.eclipse.che.api.workspace.server.model.impl.VolumeImpl;
+import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
+import org.eclipse.che.api.workspace.server.spi.environment.InternalEnvironment;
+import org.eclipse.che.api.workspace.server.spi.environment.InternalMachineConfig;
+import org.eclipse.che.api.workspace.server.wsnext.WorkspaceNextApplier;
+import org.eclipse.che.api.workspace.server.wsnext.model.CheService;
+import org.eclipse.che.api.workspace.server.wsnext.model.Container;
+import org.eclipse.che.api.workspace.server.wsnext.model.EnvVar;
+import org.eclipse.che.api.workspace.server.wsnext.model.ResourceRequirements;
+import org.eclipse.che.api.workspace.server.wsnext.model.Server;
+import org.eclipse.che.api.workspace.server.wsnext.model.Volume;
+import org.eclipse.che.workspace.infrastructure.kubernetes.Names;
+import org.eclipse.che.workspace.infrastructure.kubernetes.environment.KubernetesEnvironment;
+import org.eclipse.che.workspace.infrastructure.kubernetes.util.Containers;
+
+/**
+ * Applies Workspace.Next configuration to a kubernetes internal runtime object.
+ *
+ * @author Oleksander Garagatyi
+ */
+@Beta
+public class KubernetesWorkspaceNextApplier implements WorkspaceNextApplier {
+
+  private final String defaultMachineMemorySizeAttribute;
+
+  @Inject
+  public KubernetesWorkspaceNextApplier(
+      @Named("che.workspace.default_memory_mb") long defaultMachineMemorySizeMB) {
+    this.defaultMachineMemorySizeAttribute =
+        String.valueOf(defaultMachineMemorySizeMB * 1024 * 1024);
+  }
+
+  @Override
+  public void apply(InternalEnvironment internalEnvironment, Collection<CheService> cheServices)
+      throws InfrastructureException {
+    if (cheServices.isEmpty()) {
+      return;
+    }
+    KubernetesEnvironment kubernetesEnvironment = (KubernetesEnvironment) internalEnvironment;
+    Map<String, Pod> pods = kubernetesEnvironment.getPods();
+    if (pods.size() != 1) {
+      throw new InfrastructureException(
+          "Workspace.Next configuration can be applied to a workspace with one pod only");
+    }
+    Pod pod = pods.values().iterator().next();
+    for (CheService cheService : cheServices) {
+      for (Container container : cheService.getSpec().getContainers()) {
+        io.fabric8.kubernetes.api.model.Container k8sContainer =
+            addContainer(pod, container.getImage(), container.getEnv(), container.getResources());
+
+        String machineName = Names.machineName(pod, k8sContainer);
+
+        InternalMachineConfig machineConfig =
+            addMachine(
+                kubernetesEnvironment, machineName, container.getServers(), container.getVolumes());
+
+        normalizeMemory(k8sContainer, machineConfig);
+      }
+    }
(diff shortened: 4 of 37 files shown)


wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
```diff
@@ -68,6 +68,18 @@
    * Workspace#getAttributes}
    */
   public static final String ERROR_MESSAGE_ATTRIBUTE_NAME = "errorMessage";
+  /**
+   * Contains Workspace.Next features list that should be added to a workspace.
+   * Should be set/read from {@link Workspace#getAttributes}.
+   *
+   * <p/>Value is comma separated list of features in a format:
+   * '< feature1Name >/< feature1Version >,< feature2Name >/< feature2Version >'
```
sleshchenko: Why is `/` symbol chosen as a separator of name and version? Wouldn't `:` symbol more conventional?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/wsnext/WorkspaceNextObjectsRetriever.java
```diff
@@ -0,0 +1,299 @@
...
+
+  protected <T> T parseResponseStreamAndClose(InputStream inputStream, Class<T> clazz)
+      throws IOException {
+    try (InputStreamReader reader = new InputStreamReader(inputStream)) {
+      T objectFromJson = GSON.fromJson(reader, clazz);
+      if (objectFromJson == null) {
+        LOG.error(
+            "Response doesn't contain any data, though it was expected to contain some.");
+        throw new IOException(
+            "Internal server error. Unexpected response body received.");
```
sleshchenko: I would say that `empty response body` would be more accurate message.

deploy/kubernetes/kubectl/che-kubernetes.yaml
```diff
@@ -270,6 +270,8 @@ items:
               configMapKeyRef:
                 key: CHE_INFRA_KUBERNETES_SERVER__STRATEGY
                 name: che
+          - name: CHE_WORKSPACE_FEATURE_API
+            value: "http://feature-api-service:3000"
```
sleshchenko: Wouldn't it better to move a value to the config map?

sleshchenko (commented):
In general, looks good to me :+1: 
I'll clone your changes locally and review in details.

wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
garagatyi: It is better to ask @skabashnyuk - it is Feature API that declares format.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/wsnext/WorkspaceNextObjectsRetriever.java
garagatyi: Does message `Response contains an empty body, though it was expected to contain data` looks better to you?

deploy/kubernetes/kubectl/che-kubernetes.yaml
garagatyi: Agree, will update this

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/wsnext/WorkspaceNextObjectsRetriever.java
sleshchenko: If the message will contain that this response is from "Feature API" then it would be even more clear and helpful.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/wsnext/WorkspaceNextObjectsRetriever.java
```diff
@@ -0,0 +1,299 @@
...
+      // for now we match whole env variable value against '${<parameter name>}'
+      service.getSpec()
+          .getContainers()
+          .forEach(container -> container.getEnv()
+              .forEach(envVar -> {
+                if (isParameter(envVar)) {
+                  String envVarValue = envVar.getValue();
+                  String parameterKey = serviceKey + '/' + envVarValue;
+                  List<String> remove = parameters.remove(parameterKey);
+                  if (remove != null) {
```
sleshchenko: Just idea to discuss, maybe we should set empty value if there is not the corresponding parameter value in a feature?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/wsnext/WorkspaceNextObjectsRetriever.java
```diff
@@ -0,0 +1,299 @@
...
+
+    for (CheService service : services) {
+      String serviceName = service.getMetadata().getName();
+      String serviceVersion = service.getSpec().getVersion();
+      String serviceKey = serviceName + '/' + serviceVersion;
+
+      // for now we match whole env variable value against '${<parameter name>}'
+      service.getSpec()
+          .getContainers()
+          .forEach(container -> container.getEnv()
```
sleshchenko: Maybe it would be better(maybe not =)) to rewrite it in the following manner
```
      service.getSpec()
          .getContainers()
          .stream()
          .flatMap(c -> c.getEnv().stream())
          .forEach(envVar -> {...});
```

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/wsnext/WorkspaceNextApplier.java
```diff
@@ -0,0 +1,37 @@
...
+import com.google.common.annotations.Beta;
+import java.util.Collection;
+import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
+import org.eclipse.che.api.workspace.server.spi.environment.InternalEnvironment;
+import org.eclipse.che.api.workspace.server.wsnext.model.CheService;
+
+/**
+ * Applies Workspace.Next configuration to an internal runtime object that represents workspace
+ * runtime configuration on an infrastructure level.
+ *
```
sleshchenko, line 22: I think it may be helpful to add a note that implementations must be bound with map binder and maybe add an example of binding.

deploy/kubernetes/kubectl/che-kubernetes.yaml
```diff
@@ -270,6 +270,8 @@ items:
               configMapKeyRef:
                 key: CHE_INFRA_KUBERNETES_SERVER__STRATEGY
                 name: che
+          - name: CHE_WORKSPACE_FEATURE_API
```
sleshchenko, line 274: Maybe it makes sense to add the same configuration into OS deployment

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsnext/KubernetesWorkspaceNextApplier.java
```diff
@@ -0,0 +1,178 @@
...
+  @Inject
+  public KubernetesWorkspaceNextApplier(@Named("che.workspace.default_memory_mb") long defaultMachineMemorySizeMB) {
+    this.defaultMachineMemorySizeAttribute =
+        String.valueOf(defaultMachineMemorySizeMB * 1024 * 1024);
+  }
+
+  @Override
+  public void apply(InternalEnvironment internalEnvironment, Collection<CheService> cheServices)
+      throws InfrastructureException {
+    KubernetesEnvironment kubernetesEnvironment = (KubernetesEnvironment) internalEnvironment;
```
sleshchenko: It makes sense to move this line lower since it is not needed to cast environment if Che services list is empty

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsnext/KubernetesWorkspaceNextApplier.java
```diff
@@ -0,0 +1,178 @@
...
+
+  private io.fabric8.kubernetes.api.model.Container addContainer(
+      Pod toolingPod,
+      String image,
+      List<EnvVar> env,
+      ResourceRequirements resources) {
+    io.fabric8.kubernetes.api.model.Container container =
+        new ContainerBuilder()
+            .withImage(image)
+            .withName(Names.generateName(""))
```
sleshchenko: Maybe it makes sense to add some meaningful prefix like "tooling"

sleshchenko (commented):
Good job. Please ping me one more time when PR won't be WIP.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/wsnext/WorkspaceNextObjectsRetriever.java
```diff
@@ -0,0 +1,299 @@
...
+          services.put(key, getService(serviceName, serviceVersion));
+        }
+      }
+    }
+    return services.values();
+  }
+
+  /**
+   * Returns map where key is concatenation of service name, version and parameter name separated by slash symbol.
+   * < serviceName/serviceVersion/parameter > to < parameterValue >
```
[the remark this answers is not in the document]
garagatyi: yes, thanks for pointing it out

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/wsnext/WorkspaceNextObjectsRetriever.java
garagatyi: maybe, I'm not sure here

garagatyi:
@sleshchenko I've updated PR with fixes, unit tests and rebased it against master to resolve conflicts. Please, take a look one more time. 

@l0rd @skabashnyuk Looks like this PR is not successful in acquiring new code reviewers, so I'll merge it as soon as @sleshchenko approves it unless other reviews appear.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/wsnext/WorkspaceNextObjectsRetriever.java
```diff
@@ -0,0 +1,403 @@
...
+          format("Service %s/%s retrieval failed", serviceName, serviceVersion));
+    } catch (IOException e) {
+      throw new InfrastructureException(
+          format(
+              "Error occurred on retrieval of service %s. Error: %s",
+              serviceName + '/' + serviceVersion, e.getMessage()));
+    }
+  }
+
+  private void validateFeature(CheFeature feature, String name, String version)
```
sleshchenko, line 271: Just proposal. Maybe it would be better to move validation private methods to new `Validator` class to make this class a bit smaller.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/wsnext/WorkspaceNextObjectsRetriever.java
```diff
@@ -0,0 +1,403 @@
...
+  }
+
+  private void requireNotNull(Object object, String error, String... errorArgs)
+      throws InfrastructureException {
+    if (object == null) {
+      throw new InfrastructureException(format(error, (Object[]) errorArgs));
+    }
+  }
+
+  private void requireEqual(String version, String version1, String error, String... errorArgs)
```
sleshchenko, line 397: I guess you used `Refactor` functionality and it's why the arguments are named `version` and `version1`. Am I right?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsnext/KubernetesWorkspaceNextApplier.java
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  @Override
+  public void apply(InternalEnvironment internalEnvironment, Collection<CheService> cheServices)
+      throws InfrastructureException {
+    if (cheServices.isEmpty()) {
+      return;
+    }
+    KubernetesEnvironment kubernetesEnvironment = (KubernetesEnvironment) internalEnvironment;
+    Map<String, Pod> pods = ((KubernetesEnvironment) internalEnvironment).getPods();
```
sleshchenko: Looks like you don't need to cast it for the second time and you can reuse `kubernetesEnvironment`

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsnext/KubernetesWorkspaceNextApplierTest.java
```diff
@@ -0,0 +1,177 @@
...
+  private static final String ENV_VAR = "PLUGINS_ENV_VAR";
+  private static final String ENV_VAR_VALUE = "PLUGINS_ENV_VAR_VALUE";
+  private static final String MEMORY_KEY = "memory";
+  private static final String MEMORY_VALUE = "100Mi";
+  private static final String POD_NAME = "pod12";
+  private static final Map<String, String> RESOURCES_REQUEST = ImmutableMap.of(MEMORY_KEY, MEMORY_VALUE);
+
+
+  @Mock
+  Pod pod;
```
sleshchenko: Looks like unformatted. Please don't forget to execute `mvn fmt:format` before merge.
Also, I would make these variable private. It doesn't really matter while it is test class but still.

deploy/kubernetes/kubectl/wsnext/feature-api.yaml
```diff
@@ -0,0 +1,85 @@
...
+    selector:
+      app: feature-api
+- apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    labels:
+      app: feature-api
+    name: feature-api
+  data:
+    CHE_REGISTRY_UPDATE_INTERVAL: "60"
```
sleshchenko, line 26: Is it value in seconds?

wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
skabashnyuk: The format is still under discussion. I do like ```:```  Fill free to comment. https://github.com/eclipse/che/issues/9895

deploy/kubernetes/kubectl/wsnext/feature-api.yaml
```diff
@@ -0,0 +1,85 @@
+---
```
sleshchenko, line 1: It would be nice to have few words in `Deploy Che.md` about deploying `feature-api`

sleshchenko (approved):
Good job :+1:

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/wsnext/WorkspaceNextObjectsRetriever.java
garagatyi: Looks like, I don't really remember, to be honest.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsnext/KubernetesWorkspaceNextApplier.java
garagatyi: thanks

deploy/kubernetes/kubectl/wsnext/feature-api.yaml
garagatyi: yes

deploy/kubernetes/kubectl/wsnext/feature-api.yaml
garagatyi: ok

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsnext/KubernetesWorkspaceNextApplierTest.java
garagatyi: ok

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/wsnext/WorkspaceNextObjectsRetriever.java
garagatyi: I would not want to create a separate class for that. At least for now.

garagatyi:
ci-build

deploy/kubernetes/kubectl/Deploy Che.md
```diff
@@ -33,3 +33,7 @@ and clean redeploy of Che.
 - Create namespace `che`: `kubectl create namespace che`
 - Deploy Che: `kubectl --namespace=che apply -f che-kubernetes.yaml`
 - Check Che pod status until it become `Running`: `kubectl get --namespace=che pods`
+
+### Workspace Next:
+There is a file `wsnext/feature-api.yaml` which contains kubesrb service needed to test Workspace Next on kubernetes infrastructure.
```
l0rd: You may want to `s/kubesrb/kubesrv/`

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/wsnext/model/Container.java
```diff
@@ -0,0 +1,156 @@
...
+import java.util.List;
+import java.util.Objects;
+
+public class Container {
+
+  private String image = null;
+  private List<EnvVar> env = new ArrayList<EnvVar>();
+  private ResourceRequirements resources = null;
+  private List<Command> commands = new ArrayList<Command>();
+  private List<Server> servers = new ArrayList<Server>();
```
l0rd, line 23: I don't remember that `Container` had a list of servers as an attribute in the original model. And I don't think that we should do that: 
- A `CheService` that has a list of `Container` and each `Container` has a list of `Servers` is something difficult to understand (what's the difference between a `CheService` and a `Server`?)
- Moreover one of workspace.next goals was to extract the list servers that originally ran in on container into multiple containers (sidecars). So it doesn't make sense to have a list of `Server` for each `Container`.

I would suggest to move higher (at `CheService` level) the attributes of the `Server`.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/wsnext/model/Container.java
```diff
@@ -0,0 +1,156 @@
...
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.api.workspace.server.wsnext.model;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+public class Container {
```
l0rd, line 17: Wouldn't it be more consistent to call it `CheContainer` (as `CheService` or `CheFeature`)? In the issue where the model is described it is called `Container` but that's weird. WDYT?

l0rd (commented):
Sorry my review comes really late. That's a great PR and the code is clean and easy to read/review but I think we should still discuss if a `Container` should have a list of `Servers`.

deploy/kubernetes/kubectl/Deploy Che.md
garagatyi: thanks

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/wsnext/model/Container.java
garagatyi: Maybe. I reused already generated files by Swagger model because I don't want to bring new changes that I would have to redo after regeneration of the model classes.
So, I believe, it is rather a suggestion for @skabashnyuk

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/wsnext/model/Container.java
garagatyi: I clearly see servers field in one of the model versions, in particular, https://github.com/eclipse/che/issues/9292#issuecomment-383866761.

> what's the difference between a CheService and a Server

As we discussed before usage of Service word is confusing and brings misunderstanding like this one. 

> Moreover one of workspace.next goals was to extract the list servers that originally ran in on container into multiple containers (sidecars). So it doesn't make sense to have a list of Server for each Container.

It make sense. We should, probably, discuss it with @skabashnyuk. 

> I would suggest to move higher (at CheService level) the attributes of the Server.

I don't own Workspace.Next model, so I would discuss it with Sergii first. In any case I think it makes sense to do that in a separate PR.

garagatyi:
ci-build

riuvshin:
ci-build

Issue #9351, opened 2018-04-06
Newer Go and C# language servers do not work with Che LSP client


ghost:
## The problem

Having tried to launch Go and new C# language server, I faced the same problem:

1. Language server is initialized ok
2. Language server capabilities do not work, and ws agent spits the following log: https://gist.github.com/eivantsov/45ba02bec3d7112a500a3b94e7c80186

## Steps to reproduce

1. Create and launch a workspace using the following config https://gist.github.com/eivantsov/6b52bafaca1b7a4180229ae289a46d0a
2. Open .go file

1. Create and launch workspace using the following config
https://gist.github.com/eivantsov/caae9c485402c2e91ec3372b4d634918
2. Open .cs file

## Language servers used

https://github.com/theia-ide/go-language-server
https://github.com/OmniSharp/omnisharp-roslyn/releases

gorkem:
Do we need the new version of LSP4J @vparfonov probably good candidate for next sprint. @tsmaeder are we going to update LSP4J together with jdt.ls

eivantsov:
Language servers do not necessarily have to be in sidecars. You can install C# with an installer and launch with a launcher script as usual (`./run --lsp --verbose`)

PR #9776, opened 2018-05-22
Send initialized event, after initialization

vparfonov:

### What does this PR do?
According to the specification client must send `initialized` message jaust after getting server capabilities. Looks like not all LS support this method we met this problem with LS's what we got from VS Code. Like GoLang ang C# 

#9351


vparfonov:
ci-test

musienko-maxim:
After first running the selenium tests, some tests fell (there were some problems on github side). But after relaunching they passed. Also were checked LS servers by manual test plan. The LS like TypeScripr, Clang were checked on Docker infrastructure, because on Openshift infrastructure this LS demand root privilegs. This will be fixed in near time.

musienko-maxim (approved):

dkulieshov (approved):

Issue #9484, opened 2018-04-18
Pull Request panel does not open when select to  the project


artaleks9:
### Description
_I have got unexpected behaviour of the pull request panel when select the project. The PR panel does not open._  

**Reproduction Steps:**
* Add workspace and open in the IDE
* Import a project (e.g. https://github.com/artaleks9/pull-request-plugin-test.git)
* Select the project in the project explorer after import

**Expected behavior:** 
* the PR panel should be open

**Observed behavior:**
* the PR panel does not open

**Che version**: 6.4.0
**OS and version:** Fedora 25
**Che install:** multi-user Docker/OCP 

**Additional information:**
* Related selenium test: _PullRequestPluginTest_
* Problem can be reliably reproduced, doesn't happen randomly: [Yes]
* See the attachment:
![open-pr-panel_1](https://user-images.githubusercontent.com/5408906/38944058-92da5908-433b-11e8-9e1c-da00c246fd5b.gif)

PR #9806, opened 2018-05-24
Fix contribute part displaying at project selection

RomanNikitenko:
### What does this PR do?
- Fix bug related to restore IDE state when one of the components can not be restored correctly
- Add Preferences page for Contribute Part
- Control the Contribute Part displaying corresponding to preference property.

Preference property "Activate Contribute Panel by clicking on a project" defined as "true" by default - so we show the Contribute Part at project selection, but once it is closed - the preference property gets "false" value and it must stay closed.  

User can display the panel:
- change the preference property to "true" and select a project
- using Panel selector
- from menu Assistant ---> Tool Windows ---> Contribute action
- by hot key

Preference property value is changed automatically to "false":
- user clicks "Hide" button of the Tooling Part Stack panel 
- user hit "Options --> Hide" action of the Tooling Part Stack panel
- user hit action from menu Assistant --> Tool Windows --> Contribute when Contribute Part is active

To inform user how change preference property notification is displayed:
 
notification

#9484  

contrpart

### DIFF
--- a/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/parts/PartStack.java
+++ b/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/parts/PartStack.java
@@ -104,8 +104,21 @@ public interface PartStack extends Presenter {
   /** Shows the part stack. */
   void show();
 
-  /** Hides the part stack. */
+  /**
+   * Hides the Part Stack. Use {@link #hide(boolean)} when hiding of the Part Stack is caused by
+   * user action
+   */
   void hide();
+
+  /**
+   * Hides the Part Stack.
+   *
+   * @param isUserInteraction pass {@code true} when hiding of the Part Stack is caused by user
+   *     action (user clicked 'Hide' button, for example) or {@code false} otherwise
+   */
+  default void hide(boolean isUserInteraction) {
+    hide();
+  }
 
   /** Maximizes the part stack. */
   void maximize();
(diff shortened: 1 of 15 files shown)


RomanNikitenko:
ci-test

RomanNikitenko:
ci-test

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/parts/PartStack.java
```diff
@@ -145,6 +146,9 @@
    */
   void setDelegate(ActionDelegate delegate);
 
+  /** Returns related to Part Stack properties */
+  PartStackProperties getProperties();
```
vparfonov: try to add  Default Implementation here for better backward compatibility

vparfonov: Also would be more simple return here just a Map<String, String>

RomanNikitenko:
ci-test

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/parts/PartStack.java
RomanNikitenko: @vparfonov 
I see your point, thank you, Vitalii!
We can have some issue related to manage and control properties at this case. Also we should notify consumers  when some property is changed.
I have changed implementation so we can avoid extra objects.
Please review.

RomanNikitenko: @vparfonov 
I have changed implementation so we can avoid extra objects.
Please review.

RomanNikitenko:
ci-test

RomanNikitenko:
ci-test

vparfonov (approved):

artaleks9:
Test report does not have regression.
Manual checking of the new feature 'Contribute' is OK, and it will be covered by selenium test.

Issue #9718, opened 2018-05-16
Prepare build for eclipse/che-theia image on CI


ashumilova:
This task has the following requirements:
- [x] the build of the image from the following [dockerfile](https://github.com/eclipse/che/blob/master/dockerfiles/theia/Dockerfile) should be handled on daily basis, same as for other nightly images we have in Сhe, and published on DockerHub
- [x] on Che release the eclipse/che-theia should have complex version number - [che version]-[theia version] (example, eclipse/che-theia:6.11.0-0.3.8), in case of Che release only Che version part changes, theia one will change manually on theia releases
- [x] stacks.json should be updated as well with the tagged version

riuvshin:
@ashumilova just a few questions:
1) `eclipse/che-theia:nightly ` is this right theia image name which should be generated on daily basis?

2) which of stacks.json should be updated 
```
ide/che-core-ide-stacks/src/main/resources/stacks.json
```
OR
```
wsmaster/che-core-api-workspace/src/test/resources/stacks.json
```
3) where can I get actual theia tag to use?

ashumilova:
1. eclipse/che-theia:nightly-0.3.8 or maybe change the order to have - eclipse/che-theia:0.3.8-nightly
@benoitf wdyt?

2. The stacks are the following:
https://github.com/eclipse/che/blob/master/ide/che-core-ide-stacks/src/main/resources/stacks.json#L56
https://github.com/eclipse/che/blob/master/ide/che-core-ide-stacks/src/main/resources/stacks.json#L126

3. https://github.com/eclipse/che/tree/master/dockerfiles/theia
https://hub.docker.com/r/eclipse/che-theia/
https://hub.docker.com/r/theiaide/theia/

ashumilova:
FYI - 0.3.10 - for Theia current version

riuvshin:
nightly tag is already in place: https://hub.docker.com/r/eclipse/che-theia/tags/

PR #9850, opened 2018-05-29
fix build.sh for theia image

riuvshin:
- add an ability to support multiple build args in `build.include`
- add an ability to override theia version using `build.sh`
needed for: https://github.com/eclipse/che/issues/9718#issuecomment-392681224

Please not that with this PR we set concrete default Theia version in dockerfile.  It will be possible to override it locally with build arg but to update theia version we will need change Dockerfile.

### DIFF
--- a/dockerfiles/build.include
+++ b/dockerfiles/build.include
@@ -77,7 +77,7 @@ init() {
       --dockerfile:*)
         DOCKERFILE="${1#*:}"
         shift ;;
-      --build-args:*)
+      --build-arg*:*)
         BUILD_ARGS_CSV="${1#*:}"
         prepare_build_args $BUILD_ARGS_CSV
         shift ;;
--- a/dockerfiles/build.include
+++ b/dockerfiles/build.include
@@ -21,6 +21,13 @@ skip_tests() {
   fi
 }
 
+prepare_build_args() {
+    IFS=',' read -r -a BUILD_ARGS_ARRAY <<< "$@"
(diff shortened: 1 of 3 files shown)


dockerfiles/build.include
```diff
@@ -70,8 +77,9 @@ init() {
       --dockerfile:*)
         DOCKERFILE="${1#*:}"
         shift ;;
-      --build-arg:*)
-        BUILD_ARG="--build-arg ${1#*:}"
+      --build-args:*)
```
benoitf: should we keep backward compliance with --build-arg ?

AndrienkoAleksandr (approved):

dockerfiles/build.include
riuvshin: yeah, I thought about this... but not sure how to do it except keep old arg name... any idea?

ashumilova (approved):

dockerfiles/build.include
AndrienkoAleksandr: I guess, users quite seldom use this argument by manually. In most cases I think they use original images...

riuvshin: we are talking not only about users here, for instance after merge of this PR I will need to fix several CI jobs... but I do not see how to make it backward compatible without renaming OR duplication args

benoitf: @riuvshin maybe in your case it's handling more usecase than expected but I'm fine (like --build-argument:foo)

benoitf (approved):

PR #9869, opened 2018-05-30
Automate adding self signed cert to Java trust store in Che server pod

ghost:
### What does this PR do?

This PR adds a simple instruction in entrypoint.sh to add self signed cert to Java trust store of Che server pod. The cert content is retrieved as env variable.
 
### Why not automate for workspace pods?

It is possible to do that by injecting this env into workspace containers, however,  entrypoints in all images need to be changed as this should happen in the entrypoint, ie before a bootstrapper is curl'ed into a pod. 

Also, when it comes to workspaces, Java trust store isn't enough in this case, and the cert should be added to root certs so that curl and bootstrapper do not yell at https Che route that has a self signed cert.

Even if we disable ssl check for curl exec, the bootstrapper itself fails to start:

```
2018/05/30 05:21:45 Couldn't connect to endpoint 'wss://che-eclipse-che.192.168.1.79.nip.io/api/websocket', due to error 'x509: certificate signed by unknown authority'
```

This requires sudo privileges. There might be some workarounds like changing permissions for certain files in base images, however, this is not a universal solution.

So, a this point, rebuilding an image with the cert and creating a custom stack sounds like the most reasonable solution for this non-standard use case.

Related docs PR with instructions: https://github.com/eclipse/che-docs/pull/412

A local Origin with Che multi user in https mode:

image

image

image

### DIFF
--- a/dockerfiles/che/entrypoint.sh
+++ b/dockerfiles/che/entrypoint.sh
@@ -288,7 +288,7 @@ add_cert_to_truststore() {
             echo "${OPENSHIFT_IDENTITY_PROVIDER_CERTIFICATE}" > /data/openshift.crt
             echo yes | keytool -keystore /data/openshift.jks -importcert -alias HOSTDOMAIN -file /data/openshift.crt -storepass minishift
         fi
-            export JAVA_OPTS="${JAVA_OPTS} -Djavax.net.ssl.trustStore=/data/openshift.jks -Djavax.net.ssl.trustStorePassword=minishift"
+        export JAVA_OPTS="${JAVA_OPTS} -Djavax.net.ssl.trustStore=/data/openshift.jks -Djavax.net.ssl.trustStorePassword=minishift"
     fi
 }
 
--- a/dockerfiles/che/entrypoint.sh
+++ b/dockerfiles/che/entrypoint.sh
@@ -283,12 +283,10 @@ init() {
 add_cert_to_truststore() {
 
     if [ "${OPENSHIFT_IDENTITY_PROVIDER_CERTIFICATE}" != "" ]; then
-        if [ ! -f /data/openshift.crt ]; then
-            echo "Found a custom cert. Adding it to java trust store..."
-            echo "${OPENSHIFT_IDENTITY_PROVIDER_CERTIFICATE}" > /data/openshift.crt
-            echo yes | keytool -keystore /data/openshift.jks -importcert -alias HOSTDOMAIN -file /data/openshift.crt -storepass minishift
-        fi
-        export JAVA_OPTS="${JAVA_OPTS} -Djavax.net.ssl.trustStore=/data/openshift.jks -Djavax.net.ssl.trustStorePassword=minishift"
+        echo "Found a custom cert. Adding it to java trust store..."
+        echo "${OPENSHIFT_IDENTITY_PROVIDER_CERTIFICATE}" > /home/user/openshift.crt
+        echo yes | keytool -keystore /home/user/openshift.jks -importcert -alias HOSTDOMAIN -file /home/user/openshift.crt -storepass minishift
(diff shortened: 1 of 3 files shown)


dockerfiles/che/entrypoint.sh
```diff
@@ -280,6 +280,18 @@ init() {
   export JAVA_OPTS="${JAVA_OPTS} -Dche.docker.network=$NETWORK_NAME"
 }
 
+add_cert_to_truststore() {
+
+    if [ "${OPENSHIFT_IDENTITY_PROVIDER_CERTIFICATE}" != "" ]; then
+        if [ ! -f /data/openshift.crt ]; then
```
sleshchenko: If `OPENSHIFT_IDENTITY_PROVIDER_CERTIFICATE` is set and `/data/openshift.crt` already exists then existing file will be added to java store. But there no guarantee that environment variable and file have the same values. Am I correct?

dockerfiles/che/entrypoint.sh
```diff
@@ -280,6 +280,18 @@ init() {
...
+            echo "Found a custom cert. Adding it to java trust store..."
+            echo "${OPENSHIFT_IDENTITY_PROVIDER_CERTIFICATE}" > /data/openshift.crt
+            echo yes | keytool -keystore /data/openshift.jks -importcert -alias HOSTDOMAIN -file /data/openshift.crt -storepass minishift
+        fi
+            export JAVA_OPTS="${JAVA_OPTS} -Djavax.net.ssl.trustStore=/data/openshift.jks -Djavax.net.ssl.trustStorePassword=minishift"
```
sleshchenko: Because of formatting looks like this line is related to previous conditional section. Please reformat,

dockerfiles/che/entrypoint.sh
ghost: Good point. So, there will be problems updating the cert.  I think it's better not to persist it but save elsewhere, so that the file is recreated each time, which I have just done.

benoitf:
@eivantsov is it possible to use let's encrypt certificates automatically ?

eivantsov:
@benoitf only for Che server route and only if you have a public DNS.

The best way to handle this is to use Letsencrypt certs for OpenShift router - but once again, it has to be a valid public DNS.

eivantsov:
ci-test

l0rd (commented):
@eivantsov I am ok to approve because is a step forward and may unblock some customers but I think that it's still complicated and annoying for users to get through all these steps (find the openshift cert, create a secret, write a dockerfile then build and push the image in the right registry, create a custom stack based on that image, configure chrome...).

I would have preferred other options like:
- add a property to skip certificate verification in wsmaster, wsagent and stacks (a kind of what you have with `curl --insecure` or `chromium --ignore-certificate-errors`)
- add a property to use let's encrypt to generate new certificates for wsmaster and workspaces routes (using [cert-manager](https://github.com/jetstack/cert-manager) under the cover, and yes it would work fine for valid public DNS only but still a brilliant solution to the problem)

l0rd (approved):

eivantsov:
@l0rd I do understand what it takes to do this. Well, I can make it simpler for our stacks so that the cert gets injected into workspace containers and the right things happen in the entrypoint. To make it happen, some chown and chmod need to happen in those images so that update certs command can be executed without sudo.

There are 4 clients that do not like self signed certs:

* wsmaster
* wsagent
* curl and bootstrapper
* browser (you need to run Chrome with the flag you mentioned)

As to 

> add a property to use let's encrypt to generate new certificates for wsmaster and workspaces routes

it does not make sense to me, since wound't it be easier just to install letsencrypt cert for OpenShift router and forget about ssl issues once and forever (well, until the cert expires). This way all routes in an OpenShift clusters are good for any clients.

Thanks for approving :)

sleshchenko (approved):

Issue #9781, opened 2018-05-23
Go language server: no hovers and docs


ghost:
When LS (https://github.com/theia-ide/go-language-server) starts in a workspace, there are no hovers,  even though this LS has this particular capability.

The same LS in VS Code and Theia shows type and docs if any.

![vscode](https://user-images.githubusercontent.com/5337267/40426189-1703d7cc-5ea3-11e8-864e-afa2d1759dd9.png)

Exceptions in wsagent logs, though they are not cause by hovering on keywords.

```
2018-05-23 09:42:57,741[ool-10-thread-1]  [ERROR] [o.e.l.jsonrpc.RemoteEndpoint 50]     - Internal error: java.lang.reflect.InvocationTargetException
java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
        at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.lambda$null$0(GenericEndpoint.java:53)
        at org.eclipse.lsp4j.jsonrpc.services.GenericEndpoint.request(GenericEndpoint.java:105)
        at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.handleRequest(RemoteEndpoint.java:210)
        at org.eclipse.lsp4j.jsonrpc.RemoteEndpoint.consume(RemoteEndpoint.java:146)
        at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.handleMessage(StreamMessageProducer.java:149)
        at org.eclipse.lsp4j.jsonrpc.json.StreamMessageProducer.listen(StreamMessageProducer.java:77)
        at org.eclipse.lsp4j.jsonrpc.json.ConcurrentMessageProcessor.run(ConcurrentMessageProcessor.java:84)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.reflect.InvocationTargetException: null
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
[...]

PR #9874, opened 2018-05-30
Fixed golang language server hover bug

dkulieshov:
Related issue https://github.com/eclipse/che/issues/9781

Added default hover and occurrences provider initialization, moved default extension to mime type mappings to a separate class.

### DIFF
--- a/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/filetype/DefaultExtensionToMimeTypeMappings.java
+++ b/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/filetype/DefaultExtensionToMimeTypeMappings.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright (c) 2012-2018 Red Hat, Inc.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.ide.api.editor.filetype;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import java.util.Collection;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import javax.inject.Singleton;
+
+/** @author Dmytro Kulieshov */
+@Singleton
+public class DefaultExtensionToMimeTypeMappings {
+  private final Map<String, Set<String>> mappings =
+      ImmutableMap.<String, Set<String>>builder()
+          .put("c", ImmutableSet.of("text/x-csrc"))
+          .put("C", ImmutableSet.of("text/x-c++src"))
+          .put("cc", ImmutableSet.of("text/x-c++src"))
+          .put("cpp", ImmutableSet.of("text/x-c++src"))
+          .put("ino", ImmutableSet.of("text/x-c++src"))
+          .put("h", ImmutableSet.of("text/x-chdr"))
+          .put("hh", ImmutableSet.of("text/x-c++hdr"))
+          .put("c++", ImmutableSet.of("text/x-c++src"))
+          .put("cs", ImmutableSet.of("text/x-csharp"))
+          .put("m", ImmutableSet.of("text/x-objective-c"))
+          .put("java", ImmutableSet.of("text/x-java"))
+          .put("class", ImmutableSet.of("text/x-java"))
+          .put("scala", ImmutableSet.of("text/x-scala"))
+          .put("sbt", ImmutableSet.of("text/x-scala"))
+          .put("clj", ImmutableSet.of("text/x-clojure"))
+          .put("groovy", ImmutableSet.of("text/x-groovy"))
+          .put("gvy", ImmutableSet.of("text/x-groovy"))
+          .put("gy", ImmutableSet.of("text/x-groovy"))
+          .put("gradle", ImmutableSet.of("text/x-groovy"))
+          .put("kt", ImmutableSet.of("text/x-kotlin"))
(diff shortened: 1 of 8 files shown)


dkuleshov:
ci-test

ghost (approved):

eivantsov:
Everything works as expected - hovers do appear and display info they should display - type, docs etc

tsmaeder:
Could you please state what the problem was and how this PR fixes it?

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/filetype/DefaultExtensionToMimeTypeMappings.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.Collection;
+import java.util.Map;
+import java.util.Set;
+import java.util.stream.Collectors;
+import javax.inject.Singleton;
+
+/** @author Dmytro Kulieshov */
+@Singleton
+public class DefaultExtensionToMimeTypeMappings {
+  private final Map<String, Set<String>> mappings =
```
tsmaeder, line 24: We can make this static, it's immutable for the lifetime of the class. Lazy initialization/dependency injection gives no advantage here.

tsmaeder (changes requested):
I think this PR uses the wrong approach. We should configure hovers, etc. based on contributions. We have the possibility to contribute through configuration and through injection. Let's get rid of the default mappings.

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/filetype/ExtensionFileTypeIdentifier.java
```diff
@@ -10,21 +10,29 @@
...
 import org.eclipse.che.ide.api.resources.VirtualFile;
 import org.eclipse.che.ide.util.loging.Log;
 
+@Singleton
 public class ExtensionFileTypeIdentifier implements FileTypeIdentifier {
 
   /** The known extensions registry. */
   Map<String, List<String>> mappings = new HashMap<>();
 
-  public ExtensionFileTypeIdentifier() {
```
tsmaeder, line 26: It would have been a much simpler fix to simply register the mime types for hover here instead of refactoring stuff.

tsmaeder (dismissed):

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/filetype/DefaultExtensionToMimeTypeMappings.java
dkulieshov: >We can make this static, it's immutable for the lifetime of the class. Lazy initialization/dependency injection gives no advantage here.

Besides the fact that it preserves architectural homogeneity of the system and simplifies unit tests composition.

ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/editor/filetype/ExtensionFileTypeIdentifier.java
dkulieshov: As I remember we have already discussed this in another pull request, in general my attitude stays the same - my preference is to do it the most appropriate and simple way, not just simple way. If you experience difficulties in understanding I would recommend you to become more familiar with such concepts as [decomposition](https://en.wikipedia.org/wiki/Decomposition_(computer_science)), [code refactoring](https://en.wikipedia.org/wiki/Code_refactoring) and [single responsibility principle](https://en.wikipedia.org/wiki/Single_responsibility_principle). In our case there are several quite different components and their corresponding responsibilities: file type identification, default mime types registration, hover configuration initialization. Please just ping me if you need any other assistance.

dkuleshov:
> I think this PR uses the wrong approach. We should configure hovers, etc. based on contributions. We have the possibility to contribute through configuration and through injection. Let's get rid of the default mappings.

@tsmaeder I find it strange that you've missed that during your review, but actually this pull request does not take away ability to configure hovers, it simply adds the default set of configurations for hovers, which may be (and for most supported languages is) overriden by contributed configurations (e.g. through injection). So when you say "we should configure hovers" and imply that this pull request disables that feature I find myself confused and puzzled.

vparfonov:
@tsmaeder I think this PR doesn't have any influence to the jdt.ls. Proposal provided by @dkuleshov just work and I don't see any technical mistake on it. We discus this solution internal before start implementation.  If you have idea how to implement it in better way you can provide own PR

vparfonov (approved):

musienko-maxim:
We did not find regression in the  functionality that is covered by selenium tests. Also we checked LS features according to manual testplan: https://github.com/eclipse/che/wiki/Test-plan-for-checking-Language-servers. The main features worked properly.

musienko-maxim (approved):

tsmaeder:
Thinking about it overnight, I am wondering why we have a special entity for the default mime types. Why are we not simply contributing default LanguageDescription instances? Then all the hover-, etc. registration would be in one place. Methinks that would make the code simpler.

dkuleshov:
@tsmaeder 
> Thinking about it overnight, I am wondering why we have a special entity for the default mime types. Why are we not simply contributing default LanguageDescription instances? Then all the hover-, etc. registration would be in one place. Methinks that would make the code simpler.

That's a reasonable suggestion, my only concern is how we will distinguish the default `LanguageDescription` and the custom one. Actually @vparfonov and I discussed exactly the same idea to be the first viable solution. However we assumed that addition of a dozen of default language descriptions along with the mechanism that will allow us to set priorities for language descriptions (to make possible to override default descriptions by custom ones) would have larger impact on the code base. So we decided just to add a hover provider initializer for default mime types and keep the rest of the algorithms the untouched.

Ideal solution I believe is to move language description to some configuration (like workspace configuration) so we could configure languages on demand w/o recompilation, however I have not clear vision on how we can do that now.

vparfonov:
For now i prefer current proposal in future will se maybe we will found better solution. Anyone can propose new PR. So goto merge this

Issue #9885, opened 2018-05-31
Docker infrastructure must report that it doesn't supports workspaces recovery


mshaposhnik:
### Description
As it was supposed to support workspaces recovery on Docker ifrastucture, in fact it doesn't at the moment due to networking problems. That causes problems on graceful shutdown like [this](https://github.com/eclipse/che/issues/9877#issuecomment-393394021) because infra reports that it is supports recovery and workspaces didn't stop at shutdown. 
Need to fix infra to make it correctly claim that recovery is unsupported.

mshaposhnik:
Fixed in https://github.com/eclipse/che/pull/9887

PR #9887, opened 2018-05-31
 Does not claim recovery is supported on docker infra

mshaposhnik:
### What does this PR do?
Make Docker infra does not claim recovery is supported;
Fix in WS termination to allow execute termination after unsuccessful suspending;

https://github.com/eclipse/che/issues/9885

#### Release Notes
N/A (bug)

#### Docs PR
N/A (bug)

### DIFF
--- a/infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerRuntimeInfrastructure.java
+++ b/infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerRuntimeInfrastructure.java
@@ -80,6 +80,9 @@ public class DockerRuntimeInfrastructure extends RuntimeInfrastructure {
 
   @Override
   public Set<RuntimeIdentity> getIdentities() throws InfrastructureException {
+    // Due to https://github.com/eclipse/che/issues/5814, recovering is not fully possible
+    // so infrastructure should not claim support of it.
+    // return containers.findIdentities();
     throw new UnsupportedOperationException("Runtimes tracking currently does not supported.");
   }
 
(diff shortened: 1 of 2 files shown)


infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerRuntimeInfrastructure.java
```diff
@@ -80,7 +80,7 @@ protected RuntimeContext internalPrepare(
 
   @Override
   public Set<RuntimeIdentity> getIdentities() throws InfrastructureException {
-    return containers.findIdentities();
```
sleshchenko, line 83: I think it would be better to comment old line and add a link to issue why docker recovery is disabled https://github.com/eclipse/che/issues/5814.

sleshchenko (approved):
LGTM. Please take a look my advice

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerRuntimeInfrastructure.java
mshaposhnik: Ok

skabashnyuk (approved):

infrastructures/docker/infrastructure/src/main/java/org/eclipse/che/workspace/infrastructure/docker/DockerRuntimeInfrastructure.java
```diff
@@ -80,7 +80,10 @@ protected RuntimeContext internalPrepare(
...
+    // Due to https://github.com/eclipse/che/issues/5814, recovering is not fully possible
+    // so infrastructure should not claim support of it.
+    // return containers.findIdentities();
+    throw new UnsupportedOperationException("Runtimes tracking currently does not supported.");
```
gorkem, line 86: "Runtimes tracking currently does not supported." should be "Runtimes tracking is not supported."

PR #9932, opened 2018-06-04
Watch connection manager never closed when trying to delete a non-existing POD

davidfestal:
### What does this PR do?

This PR fixes a nasty bug that occurs in some conditions when trying to delete a non-existing command POD:  the corresponding watch connection manager is never closed and this leads to an infinite loop of regular attempts to reconnect web-sockets to a non-existing POD.
  
This is the root cause of https://github.com/redhat-developer/rh-che/issues/672.

### DIFF
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesPods.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesPods.java
@@ -554,22 +554,36 @@ public class KubernetesPods {
   }
 
   private CompletableFuture<Void> doDelete(String name) throws InfrastructureException {
+    Watch toCloseOnException = null;
     try {
       final PodResource<Pod, DoneablePod> podResource =
           clientFactory.create(workspaceId).pods().inNamespace(namespace).withName(name);
-      final CompletableFuture<Void> deleteFuture = new CompletableFuture<>();
+      CompletableFuture<Void> deleteFuture = new CompletableFuture<>();
       final Watch watch = podResource.watch(new DeleteWatcher(deleteFuture));
-
-      podResource.delete();
-      return deleteFuture.whenComplete(
-          (v, e) -> {
-            if (e != null) {
-              LOG.warn("Failed to remove pod {} cause {}", name, e.getMessage());
-            }
-            watch.close();
-          });
+      toCloseOnException = watch;
+      deleteFuture =
+          deleteFuture.whenComplete(
+              (v, e) -> {
+                if (e != null) {
+                  LOG.warn("Failed to remove pod {} cause {}", name, e.getMessage());
+                }
+                watch.close();
+              });
+      Boolean deleteSucceeded = podResource.delete();
+      if (deleteSucceeded == null || !deleteSucceeded) {
+        watch.close();
+      }
+      return deleteFuture;
     } catch (KubernetesClientException ex) {
+      if (toCloseOnException != null) {
+        toCloseOnException.close();
+      }
       throw new KubernetesInfrastructureException(ex);
+    } catch (Exception e) {
+      if (toCloseOnException != null) {
+        toCloseOnException.close();
+      }
+      throw e;
     }
   }
 
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesPods.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesPods.java
@@ -15,6 +15,7 @@ import static java.util.concurrent.CompletableFuture.allOf;
 import static org.eclipse.che.workspace.infrastructure.kubernetes.Constants.CHE_WORKSPACE_ID_LABEL;
 import static org.eclipse.che.workspace.infrastructure.kubernetes.namespace.KubernetesObjectUtil.putLabel;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.fabric8.kubernetes.api.model.DoneablePod;
 import io.fabric8.kubernetes.api.model.Event;
 import io.fabric8.kubernetes.api.model.ObjectReference;
@@ -553,27 +554,26 @@ public class KubernetesPods {
     }
   }
 
-  private CompletableFuture<Void> doDelete(String name) throws InfrastructureException {
+  @VisibleForTesting
+  CompletableFuture<Void> doDelete(String name) throws InfrastructureException {
     Watch toCloseOnException = null;
     try {
       final PodResource<Pod, DoneablePod> podResource =
           clientFactory.create(workspaceId).pods().inNamespace(namespace).withName(name);
-      CompletableFuture<Void> deleteFuture = new CompletableFuture<>();
+      final CompletableFuture<Void> deleteFuture = new CompletableFuture<>();
       final Watch watch = podResource.watch(new DeleteWatcher(deleteFuture));
       toCloseOnException = watch;
-      deleteFuture =
-          deleteFuture.whenComplete(
-              (v, e) -> {
-                if (e != null) {
-                  LOG.warn("Failed to remove pod {} cause {}", name, e.getMessage());
-                }
-                watch.close();
-              });
       Boolean deleteSucceeded = podResource.delete();
       if (deleteSucceeded == null || !deleteSucceeded) {
-        watch.close();
+        deleteFuture.complete(null);
       }
-      return deleteFuture;
+      return deleteFuture.whenComplete(
+          (v, e) -> {
+            if (e != null) {
+              LOG.warn("Failed to remove pod {} cause {}", name, e.getMessage());
+            }
+            watch.close();
+          });
     } catch (KubernetesClientException ex) {
       if (toCloseOnException != null) {
         toCloseOnException.close();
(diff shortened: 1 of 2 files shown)


davidfestal:
ci-test

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesPods.java
```diff
@@ -554,22 +554,36 @@ public void delete() throws InfrastructureException {
...
+          deleteFuture.whenComplete(
+              (v, e) -> {
+                if (e != null) {
+                  LOG.warn("Failed to remove pod {} cause {}", name, e.getMessage());
+                }
+                watch.close();
+              });
+      Boolean deleteSucceeded = podResource.delete();
+      if (deleteSucceeded == null || !deleteSucceeded) {
+        watch.close();
```
sleshchenko: I think, in a case when a pod doesn't exist, delete future will be completed exceptionally with message `Webscoket connection is closed. But event about removing is not received.`. And I would say that it is not an exceptional situation. Then maybe `deleteFuture.complete()` would be better instead of closing a `Watch`

davidfestal: @sleshchenko  It appears (also through debugging) that it's clearly not the case: when the POD doesn't exist, the `DeleteWatcher` isn't called at all by the `WatchConnectionManager`  and finally the `WatchConnectionManager` is never closed.

This is precisely the bug I'm fixing here.

This is why I have to explicitly close the `WatchConnectionManager` when the `delete()` call returns `false` because no such POD exists.

sleshchenko: I understand the fixed bug and it's nice catch =)
And what I understand is (as far as I understand) if you completed future then onComplete will be called and watch will be closed. And completing the future without exception would be more correctly here.

davidfestal: well, we would still have to close the watch in case an exception occurs in the `podResource.delete()` call and no `DeleteWatcher` method is called. So why not do the same action in the 2 cases where deletion didn't occur ?

davidfestal:
@l0rd I added a test

@sleshchenko I finally changed to use `deleteFuture.complete()` as you proposed. Please could you confirm this is what you were expecting ?

davidfestal:
ci-test

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesPods.java
```diff
@@ -569,7 +575,15 @@ public void delete() throws InfrastructureException {
...
     } catch (KubernetesClientException ex) {
+      if (toCloseOnException != null) {
+        toCloseOnException.close();
+      }
       throw new KubernetesInfrastructureException(ex);
+    } catch (Exception e) {
+      if (toCloseOnException != null) {
+        toCloseOnException.close();
+      }
+      throw e;
```
sleshchenko, line 586: Maybe it would be better to wrap this exception into `InternalInfastructureException`

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesPods.java
```diff
@@ -553,14 +554,19 @@ public void delete() throws InfrastructureException {
     }
   }
 
-  private CompletableFuture<Void> doDelete(String name) throws InfrastructureException {
+  @VisibleForTesting
+  CompletableFuture<Void> doDelete(String name) throws InfrastructureException {
+    Watch toCloseOnException = null;
```
sleshchenko, line 559: I can propose you to rewrite this method in the following way:
```java
@VisibleForTesting
  CompletableFuture<Void> doDelete(String name) {
    final CompletableFuture<Void> deleteFuture = new CompletableFuture<>();
    try {
      final PodResource<Pod, DoneablePod> podResource =
          clientFactory.create(workspaceId).pods().inNamespace(namespace).withName(name);

      Watch watch = podResource.watch(new DeleteWatcher(deleteFuture));
      //watch is opened - register callback to close it
      deleteFuture.whenComplete((v, e) -> watch.close());

      Boolean deleteSucceeded = podResource.delete();
      if (deleteSucceeded == null || !deleteSucceeded) {
        deleteFuture.complete(null);
      }
    } catch (KubernetesClientException ex) {
      deleteFuture.completeExceptionally(new KubernetesInfrastructureException(ex));
    } catch (Exception e) {
      deleteFuture.completeExceptionally(new InternalInfrastructureException(e.getMessage(), e));
    }

    return deleteFuture;
  }
```
Why it can be better - because in a case when an exception occurs during removing many pods - then there will be trying to remove all pods. With current approach - if an exception occurs then no more pods will be removed. And I think it would be better to try to clean up resources as many as we can.

sleshchenko (approved):
Good fix :+1: 

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesNamespaceTest.java
```diff
@@ -203,6 +207,33 @@ public void testStopsWaitingServiceAccountEventJustAfterEventReceived() throws E
     verify(serviceAccountResource).watch(any());
   }
 
+  @Test
+  public void testDeleteNonExistingPodBeforeWatch() {
+    System.out.println(
```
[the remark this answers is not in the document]
davidfestal: It's debug info that should have been removed, sure !

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesPods.java
davidfestal: well, honestly I prefer re-throwing exactly the same exception, so that there is no behavior change compared to the previous implementation that used to only catch `KubernetesClientException` exceptions.

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesNamespaceTest.java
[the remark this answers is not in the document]
davidfestal: fixed

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesNamespaceTest.java
```diff
@@ -203,6 +207,33 @@ public void testStopsWaitingServiceAccountEventJustAfterEventReceived() throws E
     verify(serviceAccountResource).watch(any());
   }
 
+  @Test
+  public void testDeleteNonExistingPodBeforeWatch() {
```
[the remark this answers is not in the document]
davidfestal: I added 2 new tests

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesNamespaceTest.java
```diff
@@ -203,6 +207,33 @@ public void testStopsWaitingServiceAccountEventJustAfterEventReceived() throws E
...
+    Watch watch = mock(Watch.class);
+    doReturn(watch).when(podResource).watch(any());
+
+    try {
+      new KubernetesPods("", "", clientFactory).doDelete("nonExistingPod").get(5, TimeUnit.SECONDS);
+    } catch (InfrastructureException
+        | InterruptedException
+        | ExecutionException
+        | TimeoutException e) {
+      e.printStackTrace();
```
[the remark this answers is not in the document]
davidfestal: fixed

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesPods.java
davidfestal: Do you mind if we do this change in another PR ?
I'd prefer keep the behavior as much identical as possible to what was previously, and only fix the bug of the non-closed watch.

davidfestal:
ci-test

davidfestal:
@l0rd Are you OK with the 3 tests I added ?

l0rd (approved):
LGTM thanks @davidfestal

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesPods.java
sleshchenko: @davidfestal Makes sense. I'm OK with that

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesPods.java
sleshchenko: I agree that it's the same as was before. But only InfrastructureException is declared as throwing and it would be safer to wrap other exception in `InternalInfastructureException`.
`InternalInfastructureException` is special kind of exception that should wrap unexpected exceptions like NullPointerException which happen because of developer fault.
I would recommend wrapping but you can leave it as it is since it doesn't change current approach.

PR #9939, opened 2018-06-05
[WIP] Upgrade Lsp4j to the 0.4.1 version

vparfonov:


plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/codeassist/CompletionItemBasedCompletionProposal.java
```diff
@@ -103,12 +106,22 @@ public void getAdditionalProposalInfo(final AsyncCallback<Widget> callback) {
   }
 
   private Widget createAdditionalInfoWidget() {
-    String documentation = completionItem.getItem().getDocumentation();
-    if (documentation == null || documentation.trim().isEmpty()) {
```
tsmaeder, line 107: Why don't we need this handling of empty/null anymore?

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/signature/ParamterInfoImpl.java
```diff
@@ -30,6 +30,10 @@ public String getLabel() {
 
   @Override
   public Optional<String> getDocumentation() {
-    return Optional.fromNullable(dto.getDocumentation());
+    if (dto.getDocumentation().isLeft()) {
+      return Optional.fromNullable(dto.getDocumentation().getLeft());
+    }
+    // TODO: need to find solution for using markdown content @see org.eclipse.lsp4j.MarkupContent
+    return Optional.fromNullable(dto.getDocumentation().getRight().getValue());
```
tsmaeder, line 37: Can dto.getDocumentation() not return null? According to https://microsoft.github.io/language-server-protocol/specification#textDocument_signatureHelp the documentation field is optional, so could be null.

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/CheLanguageClient.java
```diff
@@ -84,4 +89,17 @@ public void logMessage(MessageParams message) {
         break;
     }
   }
+
+  @Override
+  public CompletableFuture<Void> registerCapability(RegistrationParams params) {
```
tsmaeder, line 94: I don't think it's a good idea to pretend we support dynamic registration when we really don't. Language servers that would correctly report their capabilities upon startup, might actually decide to use dynamic registration instead. In particular, this might break jdt.ls

wsagent/che-core-api-languageserver/src/test/java/org/eclipse/che/api/languageserver/dto/DtoConversionTest.java
```diff
@@ -140,11 +140,11 @@ public void testEitherDeserializerWithDto() throws Exception {
 
     Hover params = DtoFactory.getInstance().createDtoFromJson(json, Hover.class);
 
-    Assert.assertTrue(params.getContents().get(0).isRight());
-    Assert.assertEquals("Value", params.getContents().get(0).getRight().getValue());
-    Assert.assertEquals("Language", params.getContents().get(0).getRight().getLanguage());
-    Assert.assertTrue(params.getContents().get(1).isLeft());
-    Assert.assertEquals("normal String", params.getContents().get(1).getLeft());
+    //    Assert.assertTrue(params.getContents().get(0).isRight());
```
tsmaeder, line 143: I guess this is because of WIP.

tsmaeder (changes requested):
I know it's WIP, but couldn't resist :-)

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/CheLanguageClient.java
dkulieshov: There is a dedicated class `ClientCapabilitiesProvider` that defines if we support dynamic registration or not and according to it we do not support dynamic registration and don't pretend we do, so there is no reason for your concern. 

I assume this stub is added to avoid `UnsupportedOperationException ` for cases when language server (by mistake) ignores our definition of dynamic registration support and still tries to register capability.

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/CheLanguageClient.java
```diff
@@ -84,4 +89,17 @@ public void logMessage(MessageParams message) {
...
+    return CompletableFuture.completedFuture(null); // runAsync(
```
dkulieshov, line 95: I think we should either remove this comment or add proper LOG message.

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/TextDocumentService.java
```diff
@@ -403,8 +401,14 @@ public boolean canDo(ExtendedLanguageServer element) {
           @Override
           public boolean handleResult(ExtendedLanguageServer element, Hover hover) {
             if (hover != null) {
-              HoverDto hoverDto = new HoverDto(hover);
-              result.getContents().addAll(hoverDto.getContents());
+              if (hover.getContents().isLeft()) {
+                List<Either<String, MarkedString>> left = hover.getContents().getLeft();
+                result.getContents().getLeft().addAll(left);
+              } else {
+                // TODO: here we can't assume response from several servers MarkupContent will be
```
dkulieshov, line 408: I think we should avoid adding new TODOs because they never get done, do you think we can change the comment or implement what's needed here?

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/CheLanguageClient.java
vparfonov: yes, thx. Will do

wsagent/che-core-api-languageserver/src/test/java/org/eclipse/che/api/languageserver/dto/DtoConversionTest.java
vparfonov: yes, you are right i will fix

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/TextDocumentService.java
vparfonov: I added TODO here because i don't find solution, what will be best to do here

wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/CheLanguageClient.java
tsmaeder: But the error message in the log is correct, we get it because the implementation of the language server is broken. How will we detect that a language server is broken in this respect? At least we would have to put a warning in the log.

tsmaeder:
For me, the build fails on che-core-api-languageserver when compiling dtos for MarkupKind, for example, since the class is final now. We'll have to adjust the LSP dto generator in order to update (at least in the jdt.ls branch).

Issue #9924, opened 2018-06-04
Consistent Lifecycle Events for Projects


tsmaeder:
I have been analysing selenium failures in the jdt.ls branch, and one of the reasons stuff fails is that we rely on Project created events. The idea is to make jdt.ls become aware of all projects that che is aware of.
Looking at the Events in org.eclipse.che.api.project.server.notification, I have noticed that they are not sent from all code paths creating projects. 
For example, ProjectCreatedEvent is sent from ProjectServiceAPI#createProject(), but when going through ProjectServiceAPI#importProject(), no such event is sent.
I see no reliable way to implement the needed functionality for jdt.ls with the current lifecycle event story. If I am wrong on this count, I would be grateful for guidance on how to do it.
In general, I think our code would benefit from making lifecycle events more consistent. As I read the code, org.eclipse.che.api.project.server.impl.ProjectConfigRegistry holds the ultimate truth of whether a project exists in Che, so I think events should be sent from there.

tsmaeder:
Maybe it makes sense to make the interface IProjectConfigRegistry package private since it's not really supposed to be implemented by someone else.

dkuleshov:
I think it's a reasonable suggestion by @tsmaeder to make project related events consistent. I agree with @gazarenkov that we should keep project registry as simple as possible, it has it's own humble responsibilities. 

I'm also concerned that `ProjectManager` and current project lifecycle are not quite ready to support **consistent** project related events, I mean if we're talking about real consistency then it is not about just addition of events in corresponding places, it is a more challenge than it may seem at first. Most obvious problem - we yet don't have clear vision on how we should treat (in context of events) projects created outside of our API (vis SSH). In actual implementation we support creation of a project by simply running `mkdir /projects/che` in a terminal (project API is not engaged).

We catch a corresponding FS event then we rise a corresponding project lifecycle event. And here I'm asking myself several questions:
- How will I distinguish FS event that caused by project API and the one that caused by outer sources? FS event don't care if a directory was created via SSH or project API, so I need to exclude  by myself FS events in such cases or I will have duplicated events (from `ProjectManager` and from FS).
- If I rely only on FS events how can I know when a project creation is finished not to rise an event too early, how I distinguish project removal and project renaming, etc.?

So before starting the work we need to clarify if we either need to develop an understanding of how we will further handle project life cycle events along with file system events, or we sacrifice full consistency for the sake of a simpler solution, which will be enough for what @tsmaeder initially asks.

[...]

PR #9954, opened 2018-06-06
Fire ProjectCreatedEvent on importProject

vparfonov:

### What does this PR do?
Fire ProjectCreatedEvent on importProject and during setting ProjectConfiguration firstly 

#9924

### DIFF
--- a/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/impl/ProjectServiceApi.java
+++ b/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/impl/ProjectServiceApi.java
@@ -289,12 +289,13 @@ public class ProjectServiceApi {
   /** Import project from specified source storage to specified location */
   public void importProject(
       String wsPath, boolean force, String clientId, SourceStorageDto sourceStorage)
-      throws ConflictException, ForbiddenException, UnauthorizedException, IOException,
-          ServerException, NotFoundException, BadRequestException {
+      throws ConflictException, ForbiddenException, UnauthorizedException, ServerException,
+          NotFoundException {
 
     wsPath = absolutize(wsPath);
 
     projectManager.doImport(wsPath, sourceStorage, force, jsonRpcImportConsumer(clientId));
+    eventService.publish(new ProjectCreatedEvent(wsPath));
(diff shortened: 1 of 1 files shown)


vparfonov:
ci-test

tsmaeder (changes requested):
This PR does not cover the case of importZip, which is used for Selenium tests.

dkulieshov (approved):

vparfonov:
@tsmaeder In case importZip we don't create Project in ProjectRegistry so fire event in this case no reason

dkuleshov:
@tsmaeder 

> This PR does not cover the case of importZip, which is used for Selenium tests.

It is expected behavior, according to [this comment](https://github.com/eclipse/che/issues/9924#issuecomment-395003908), specifically
> In case of just creating folder in the projects root - configuration is not assigned explicitly => Project even is not fired.

tsmaeder:
Yes, but a project appears in the navigator without a ProjectCreatedEvent ever being sent. Without the event being sent, I cannot tell jdt.ls to look at the project.
Perhaps just listening to file system events is the way to go for jdt.ls and to completely ignore che projects, but then we'll get into situations where what che thinks is a project and what jdt.ls thinks is a project will be different. The will be no fun to debug.

vparfonov:
ok, got it. Looks like you are right

dkuleshov:
@tsmaeder I think I understand what you mean. Just to be sure that we're on the same wave. We consider that the project is created **only** when corresponding configuration is added to project config registry. In other words when a folder is created/archive is unpacked in the projects' root we don't consider it as a project until its configuration is added.

For the case that you mentioned (selenium tests) they basically do two things: they unzip a folder and add a project configuration via `updateProject` API call. So you're right that the problem exists, however the problem is not that `importZip` API call does not publish an event, the problem is that when we add a project config with `updateProject` API call we do not publish project created event. 

So, eventually there are two bugs: 
- no `ProjectCreatedEvent` on project import 
- no `ProjectCreatedEvent` on project update, when we add new configuration

I also think that your suggestion to move project related events on a separate abstraction level to avoid event micromanagement and all related errors sounds good.

tsmaeder (approved):

PR #9975, opened 2018-06-08
Adding pony 0.22.5 lang default stack for che workspaces

philcali:
### What does this PR do?

This PR adds a pony lange image with a working run command to compile + run a pony binary built within the IDE. Longer term, I would like to provide a pony agent that can perform "on save compilation" among other things. Providing a default stack with a compile + run command is good for a community lacking an IDE presently.

N/A

#### Release Notes

- Provide a default pony lang support with compilation + binary execution.

^^ More than happy to. I was curious if something like this would be consider for merging before I went down that road.


skabashnyuk:
Hello, @philcali sorry for late response. 
Unfotenetly we are not interested in new lang stack for GWT-based ide at this moment.

PR #9981, opened 2018-06-08
Add support of Workspace Next to Openshift  infrastructure 

garagatyi:
### What does this PR do?
Add support of Workspace Next to Openshift recipe, infrastructure. 
Adds a template to deploy Feature API to Openshift.

### DIFF
--- a/deploy/openshift/templates/che-server-template.yaml
+++ b/deploy/openshift/templates/che-server-template.yaml
@@ -142,6 +142,8 @@ objects:
                 key: ca.crt
                 name: openshift-identity-provider
                 optional: true
+          - name: CHE_WORKSPACE_FEATURE_API
+            value: "http://feature-api-service:3000"
           image: ${IMAGE_CHE}:${CHE_VERSION}
           imagePullPolicy: "${PULL_POLICY}"
           livenessProbe:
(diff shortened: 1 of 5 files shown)


ghost (approved):

deploy/openshift/templates/che-server-template.yaml
```diff
@@ -142,6 +142,8 @@ objects:
                 key: ca.crt
                 name: openshift-identity-provider
                 optional: true
+          - name: CHE_WORKSPACE_FEATURE_API
+            value: "http://feature-api-service:3000"
```
sleshchenko, line 146: Does it mean that WS next functionality will be enabled even when feature API is not deployed?

sleshchenko (approved):
LGTM.

Just proposal to add an option to deploy che script to deploy and configure feature-api.

deploy/openshift/templates/che-server-template.yaml
garagatyi: Yes, it will be configured but it won't work unless someone adds features property in his workspace. So, I think it is OK to set it by default to simplify deployment in case Workspace Next functionality is needed.

PR #10003, opened 2018-06-12
Increase remove node-loader handler timeout

vinokurig:
### What does this PR do?

We have a bug that sometimes happens on CI: java package does not expand when creating a new java class with sub-directories e.g. `test1.test2.test.3.NewClass.java`. This problem was already fixed by adding a delay before removing the node-loader handler, but it still sometimes happens on CI. Increasing a delay should solve the problem on CI.

#8122 

#### Release Notes

Increase remove node-loader handler timeout

#### Docs PR

N/A


vparfonov (approved):

musienko-maxim:
The existed tests did not find new regression

vinokurig:
ci-build

PR #10021, opened 2018-06-12
Use Deployments instead of Pods in Kubernetes/OpenShift infra

amisevsk:
### What does this PR do?
Changes OpenShift infrastucture to use `DeploymentConfig`s instead of `Pod`s for workspace machines.

This is related to an issue ([rh-che#668](https://github.com/redhat-developer/rh-che/issues/668)).

In OpenShift, using a `DeploymentConfig` instead of a bare `Pod` for workspaces makes more sense, as loose `Pod`s are intended to be used for short-running jobs. As we intend workspaces to run until terminated by a user, a `DeploymentConfig` makes more sense in this context.

Further, using `DeploymentConfigs` could enable us to use some of OpenShift's features for managing running workspaces later. 

#### Release Notes

Use Deployments instead of Pods for managing workspaces when using Kubernetes/Openshift infrastructure.

#### Docs PR

N/A

### DIFF
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
@@ -18,18 +18,24 @@ import static org.eclipse.che.workspace.infrastructure.kubernetes.Constants.POD_
 import static org.eclipse.che.workspace.infrastructure.kubernetes.Constants.POD_STATUS_PHASE_SUCCEEDED;
 import static org.eclipse.che.workspace.infrastructure.kubernetes.namespace.KubernetesObjectUtil.putLabel;
 
-import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 import io.fabric8.kubernetes.api.model.DoneablePod;
 import io.fabric8.kubernetes.api.model.Event;
+import io.fabric8.kubernetes.api.model.ObjectMeta;
 import io.fabric8.kubernetes.api.model.ObjectReference;
+import io.fabric8.kubernetes.api.model.OwnerReference;
 import io.fabric8.kubernetes.api.model.Pod;
+import io.fabric8.kubernetes.api.model.PodSpec;
+import io.fabric8.kubernetes.api.model.extensions.Deployment;
+import io.fabric8.kubernetes.api.model.extensions.DoneableDeployment;
+import io.fabric8.kubernetes.api.model.extensions.ReplicaSet;
 import io.fabric8.kubernetes.client.KubernetesClientException;
 import io.fabric8.kubernetes.client.Watch;
 import io.fabric8.kubernetes.client.Watcher;
 import io.fabric8.kubernetes.client.dsl.ExecListener;
 import io.fabric8.kubernetes.client.dsl.ExecWatch;
 import io.fabric8.kubernetes.client.dsl.PodResource;
+import io.fabric8.kubernetes.client.dsl.ScalableResource;
 import java.io.ByteArrayOutputStream;
 import java.io.UnsupportedEncodingException;
 import java.net.URLEncoder;
@@ -38,6 +44,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ConcurrentLinkedQueue;
@@ -61,15 +68,16 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * Defines an internal API for managing {@link Pod} instances in {@link KubernetesPods#namespace
- * predefined namespace}.
+ * Defines an internal API for managing {@link Pod} and {@link Deployment} instances in {@link
+ * KubernetesDeployments#namespace predefined namespace}.
  *
  * @author Sergii Leshchenko
  * @author Anton Korneta
+ * @author Angel Misevski
  */
-public class KubernetesPods {
+public class KubernetesDeployments {
 
-  private static final Logger LOG = LoggerFactory.getLogger(KubernetesPods.class);
+  private static final Logger LOG = LoggerFactory.getLogger(KubernetesDeployments.class);
 
   private static final String CONTAINER_NAME_GROUP = "name";
   // when event is related to container `fieldPath` field contains
@@ -79,6 +87,7 @@ public class KubernetesPods {
 
   // TODO https://github.com/eclipse/che/issues/7656
   public static final int POD_REMOVAL_TIMEOUT_MIN = 5;
+  public static final int POD_CREATION_TIMEOUT_MIN = 5;
 
   private static final String POD_OBJECT_KIND = "Pod";
   // error stream data initial capacity
@@ -86,16 +95,17 @@ public class KubernetesPods {
   public static final String STDOUT = "stdout";
   public static final String STDERR = "stderr";
 
-  private final String namespace;
+  protected final String namespace;
+  protected final String workspaceId;
   private final KubernetesClientFactory clientFactory;
   private final ConcurrentLinkedQueue<PodActionHandler> podActionHandlers;
   private final ConcurrentLinkedQueue<PodEventHandler> containerEventsHandlers;
-  private final String workspaceId;
   private Watch podWatch;
   private Watch containerWatch;
   private Date watcherInitializationDate;
 
-  KubernetesPods(String namespace, String workspaceId, KubernetesClientFactory clientFactory) {
+  protected KubernetesDeployments(
+      String namespace, String workspaceId, KubernetesClientFactory clientFactory) {
     this.namespace = namespace;
     this.workspaceId = workspaceId;
     this.clientFactory = clientFactory;
@@ -104,11 +114,45 @@ public class KubernetesPods {
   }
 
   /**
-   * Creates specified pod.
+   * Starts the specified Pod via a Deployment.
    *
-   * @param pod pod to create
+   * @param pod pod to deploy
    * @return created pod
    * @throws InfrastructureException when any exception occurs
+   */
+  public Pod deploy(Pod pod) throws InfrastructureException {
+    putLabel(pod, CHE_WORKSPACE_ID_LABEL, workspaceId);
+    ObjectMeta metadata = pod.getMetadata();
+    PodSpec podSpec = pod.getSpec();
+    podSpec.setRestartPolicy("Always"); // Only allowable value
+    try {
+      Deployment deployment =
+          clientFactory
+              .create(workspaceId)
+              .extensions()
+              .deployments()
+              .createNew()
+              .withMetadata(metadata)
+              .withNewSpec()
+              .withReplicas(1)
+              .withNewTemplate()
+              .withMetadata(metadata)
+              .withSpec(podSpec)
+              .endTemplate()
+              .endSpec()
+              .done();
+      return awaitAndGetPod(deployment);
+    } catch (KubernetesClientException e) {
+      throw new KubernetesInfrastructureException(e);
+    }
+  }
+
+  /**
+   * Create a terminating pod that is not part of a Deployment.
+   *
+   * @param pod the Pod to create
+   * @return the created pod
+   * @throws InfrastructureException when any error occurs
    */
   public Pod create(Pod pod) throws InfrastructureException {
     putLabel(pod, CHE_WORKSPACE_ID_LABEL, workspaceId);
@@ -139,14 +183,17 @@ public class KubernetesPods {
   }
 
   /**
-   * Returns optional with pod that have specified name.
+   * Returns optional with pod that either has specified name or is controlled by Deployment with
+   * specified name.
    *
+   * @param name name of the Pod or Deployment
    * @throws InfrastructureException when any exception occurs
    */
   public Optional<Pod> get(String name) throws InfrastructureException {
+    String podName = getPodName(name);
     try {
       return Optional.ofNullable(
-          clientFactory.create(workspaceId).pods().inNamespace(namespace).withName(name).get());
+          clientFactory.create(workspaceId).pods().inNamespace(namespace).withName(podName).get());
     } catch (KubernetesClientException e) {
       throw new KubernetesInfrastructureException(e);
     }
@@ -155,22 +202,23 @@ public class KubernetesPods {
   /**
    * Waits until pod state will suit for specified predicate.
    *
-   * @param name name of pod that should be watched
+   * @param name name of pod or deployment containing pod that should be watched
    * @param timeoutMin waiting timeout in minutes
    * @param predicate predicate to perform state check
-   * @return pod that suit for specified predicate
+   * @return pod that satisifes the specified predicate
    * @throws InfrastructureException when specified timeout is reached
    * @throws InfrastructureException when {@link Thread} is interrupted while waiting
    * @throws InfrastructureException when any other exception occurs
    */
   public Pod wait(String name, int timeoutMin, Predicate<Pod> predicate)
       throws InfrastructureException {
+    String podName = getPodName(name);
     CompletableFuture<Pod> future = new CompletableFuture<>();
     Watch watch = null;
     try {
 
       PodResource<Pod, DoneablePod> podResource =
-          clientFactory.create(workspaceId).pods().inNamespace(namespace).withName(name);
+          clientFactory.create(workspaceId).pods().inNamespace(namespace).withName(podName);
 
       watch =
           podResource.watch(
@@ -186,13 +234,17 @@ public class KubernetesPods {
                 public void onClose(KubernetesClientException cause) {
                   future.completeExceptionally(
                       new InfrastructureException(
-                          "Waiting for pod '" + name + "' was interrupted"));
+                          "Waiting for pod '" + podName + "' was interrupted"));
                 }
               });
 
       Pod actualPod = podResource.get();
       if (actualPod == null) {
-        throw new InfrastructureException("Specified pod " + name + " doesn't exist");
+        if (name.equals(podName)) { // `name` refers to a bare pod
+          throw new InfrastructureException("Specified pod " + podName + " doesn't exist");
+        } else { // `name` refers to a deployment
+          throw new InfrastructureException("No pod in deployment " + name + " found.");
+        }
       }
       if (predicate.test(actualPod)) {
         return actualPod;
@@ -202,10 +254,10 @@ public class KubernetesPods {
       } catch (ExecutionException e) {
         throw new InfrastructureException(e.getCause().getMessage(), e);
       } catch (TimeoutException e) {
-        throw new InfrastructureException("Waiting for pod '" + name + "' reached timeout");
+        throw new InfrastructureException("Waiting for pod '" + podName + "' reached timeout");
       } catch (InterruptedException e) {
         Thread.currentThread().interrupt();
-        throw new InfrastructureException("Waiting for pod '" + name + "' was interrupted");
+        throw new InfrastructureException("Waiting for pod '" + podName + "' was interrupted");
       }
     } catch (KubernetesClientException e) {
       throw new KubernetesInfrastructureException(e);
@@ -223,7 +275,8 @@ public class KubernetesPods {
    * <p>Note that the resulting future must be explicitly cancelled when its completion no longer
    * important because of finalization allocated resources.
    *
-   * @param name the pod name that should be watched
+   * @param name the pod or deployment (that contains pod) name that should be watched
+   * @param predicate a function that performs pod state check
    * @return completable future that is completed when one of the following conditions is met:
    *     <ul>
    *       <li>complete successfully in case of "Running" pod state.
@@ -239,8 +292,9 @@ public class KubernetesPods {
   public CompletableFuture<Void> waitRunningAsync(String name) {
     final CompletableFuture<Void> podRunningFuture = new CompletableFuture<>();
     try {
+      final String podName = getPodName(name);
       final PodResource<Pod, DoneablePod> podResource =
-          clientFactory.create().pods().inNamespace(namespace).withName(name);
+          clientFactory.create().pods().inNamespace(namespace).withName(podName);
       final Watch watch =
           podResource.watch(
               new Watcher<Pod>() {
@@ -253,15 +307,20 @@ public class KubernetesPods {
                 public void onClose(KubernetesClientException cause) {
                   podRunningFuture.completeExceptionally(
                       new InfrastructureException(
-                          "Waiting for pod '" + name + "' was interrupted"));
+                          "Waiting for pod '" + podName + "' was interrupted"));
                 }
               });
 
       podRunningFuture.whenComplete((ok, ex) -> watch.close());
       final Pod pod = podResource.get();
       if (pod == null) {
-        podRunningFuture.completeExceptionally(
-            new InfrastructureException("Specified pod " + name + " doesn't exist"));
+        InfrastructureException ex;
+        if (name.equals(podName)) { // `name` refers to bare pod
+          ex = new InfrastructureException("Specified pod " + podName + " doesn't exist");
+        } else {
+          ex = new InfrastructureException("No pod in deployment " + name + " found.");
+        }
+        podRunningFuture.completeExceptionally(ex);
       } else {
         handleStartingPodStatus(podRunningFuture, pod);
       }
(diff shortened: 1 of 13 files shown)


amisevsk:
@garagatyi, @sleshchenko I'm very open to suggestions on improving how this is done -- I'm still relatively unfamiliar with this section of the code base. 

Also, note that there are a few caveats with this PR in its current state:
- I've only been able to test locally (via minishift). Our dev cluster seems to not be working
- Using `OpenShiftClient` (which is necessary to access `DeploymentConfigs` causes the return of an old issue: leaking connections to `https://kubernetes.default.svc/`. I don't see a way to avoid this at the moment.
- Rarely, an error is logged during workspace termination, with message 
   ```
   Error occurs on workspace 'che/wksp-o9mk' with id 'workspace2885e7jfjos1jw6d' stopped by user 'che'. Error: {}
   org.eclipse.che.api.workspace.server.spi.InfrastructureException: Error(s) occurs while cleaning up the namespace. Failure executing: PATCH at: https://kubernetes.default.svc/api/v1/namespaces/mini-che/replicationcontrollers/workspace2885e7jfjos1jw6d.dockerimage-1. Message: Forbidden!Configured service account doesn't have access. Service account may have been revoked. replicationcontrollers "workspace2885e7jfjos1jw6d.dockerimage-1" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: User "developer" cannot update deploymentconfigs/finalizers.apps.openshift.io in project "mini-che", <nil>. The error may be caused by an expired token or changed password. Update Che server deployment with a new token or password.
   ```
   This is not a token issue; it appears to be related to how kubernetes-client deals with deployment configuration. The workspace is terminated successfully even when it is logged, and I cannot reproduce it consistently. Any insights here?

@eivantsov Regarding documentation, I believe that the way I've implemented this means that all changes should be internal. We are basically just wrapping `Pods` in `DeploymentConfigs` at this point. I'd appreciate your comments on the matter regardless.

amisevsk:
I've added a commit to use `Deployments` instead of `DeploymentConfigs`. This seems to have solved the issues mentioned in my last comment. However, it seems like the version of kubernetes-client we are using is still using the [`/v1beta1`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#deployment-v1beta1-apps) API (`Deployments` are no longer in beta on Kubernetes).

@l0rd I'm a little unclear on the discussion in the issue linked -- replica sets are created by deployments to manage the pods. With the just added commit, I suppose we are back on replica sets as a side effect of using deployments. However, the suggestion to move to deployments seems strange, since that will result in the creation of replicasets anyways.

The general structure, as I understand it, is 
`DeploymentConfig -> ReplicationController -> Pods`
and
`Deployment -> ReplicaSet -> Pods`

DeploymentConfigs are what seems to be causing us issues given the version of kubernetes-client we are using.

sleshchenko (commented):
Generally looks good :+1:

I think using of DeploymentConfig is a bit overhead in Che use case because Che Server does deployment management function and using Deployments doesn't give many benefits =) But looks like we have not a choice. Maybe there is another way to mark Che Servers Pods as long-living?

So, I would say that extending KubernetesPods by OpenShiftDeployments is not the best way (maybe easiest but not the best :) ) to solve this issue.
I want to propose to you consider the following architecture:
You can add new abstraction like MachinePods, to create pods for machines (by creation pods or using deployment configs). It would look like:
```
MachinePods:
-create
-watch
-exec
-delete
...
```
`KubernetesPods` and `OpenshiftDeploymentsConfigs` would encapsulate operations with fabric8 k8s/os clients.
Then K8s/OS infrastructure will use `MachinePods` to work with machine pods and there will be two implementations - Pods based and DeploymentsConfigs based.
What could be better in this approach - deployment configs operation with fabric8 client would be separated from logic with adaption pods by DeploymentsConfig.
P.S. I would say that `MachinePods` is just topic to discuss but not requirement or recommendation.

l0rd:
@amisevsk 

> I'm a little unclear on the discussion in the issue linked 

Me too honestly :-) Let's try to figure this out with the issue author in the next days

> it seems like the version of kubernetes-client we are using is still using the /v1beta1 API 

Have you been using v3.1.12? https://github.com/eclipse/che-dependencies/commit/14a8533be22f3d21d1f1e28af70c0e3067463510

@sleshchenko 

> But looks like we have not a choice

If that's not the good choice we should not do that.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/project/OpenShiftDeploymentConfigs.java
```diff
@@ -0,0 +1,223 @@
...
+            .create(workspaceId)
+            .pods()
+            .inNamespace(namespace)
+            .withLabels(selector)
+            .list()
+            .getItems();
+    if (pods.isEmpty()) {
+      throw new InfrastructureException(String.format("Failed to find pod with name %s", name));
+    } else if (pods.size() > 1) {
+      throw new InfrastructureException(
```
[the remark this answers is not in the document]
amisevsk: Could you go into this scenario a little more? As I see it, there shouldn't be a situation in which multiple pods are within the same deployment, and so if that does occur it suggests an error state somewhere along the way. 

I'm trying to work something out where we check for the rolling state, but it seems like it would add quite a bit of logic, so for clarity's sake I want to understand the motivation.

amisevsk:
@sleshchenko I agree that at this moment Deployments are somewhat of an unnecessary overhead from a Che server perspective, but I still think it makes more sense than using bare pods, as on OpenShift (from my understanding, at least) bare Pods are generally intended for short-running (i.e. terminating) jobs. The solution to making Pods long living would be to set `activeDeadlineSeconds` to zero; however, the issue with this approach is that it would require the cluster admin to allow this to happen -- I believe on OSIO Pods are forced into a one hour timeout. I'm not sure if we want to introduce such a OS setting as a requirement to running Che. 

Regardless, that's not a settled matter and I believe we're still discussing it, so there may be another way or workaround I haven't considered.

Regarding the MachinePods abstraction, I'm happy to try to implement something like this, but I worry that it will only add confusion. I want to keep the underlying implementation details as Pods to avoid separating the OpenShift infra from the Kubernetes infra too much, so inserting this abstraction feels like it would be somewhat like just renaming KubernetesPods to WorkspacePods. Alternatively we could separate KubernetesPods and OpenShiftDeploymentConfigs entirely and connect them only via interface WorkspacePods, but then there is a lot of duplicated code between the two -- much of the functionality in OpenShiftDeploymentConfigs is just a simple indirection to translate Deployment name into Pod name (which to me makes sense, because fundamentally workspaces are Pods at the end of the day). I'll investigate it more when I have a bit of time.

@l0rd I agree, I'd like to get this right so that we don't have a nightmare on our hands later. Regarding kubernetes-client, this is all with 3.1.12 at the moment; 3.2.0 was released a few days ago but I haven't checked if that changes anything. If a later version of kubernetes-client does change to using the `v1` version of Deployments, then we would have to modify those uses (hopefully just remove the `extensions()` call), but I don't think it's a huge issue. Kubernetes [documentation](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#api-versioning) has this to say about the beta apis:
> - Code is well tested. Enabling the feature is considered safe. Enabled by default.
> - Support for the overall feature will not be dropped, though details may change
> - The schema and/or semantics of objects may change in incompatible ways in a subsequent beta or stable release. When this happens, we will provide instructions for migrating to the next version. This may require deleting, editing, and re-creating API objects. The editing process may require some thought. This may require downtime for applications that rely on the feature.
> - Recommended for only non-business-critical uses because of potential for incompatible changes in subsequent releases. If you have multiple clusters which can be upgraded independently, you may be able to relax this restriction.

Assuming kubernetes-client would maintain its internal API upon the switch, the last two points do not seem to be large issues. From eyeballing the API docs for the two versions of Deployment, it does not appear we are using deployments in any way that is not supported in `v1`

sleshchenko:
@amisevsk 
> but then there is a lot of duplicated code between the two -- much of the functionality in OpenShiftDeploymentConfigs is just a simple indirection to translate Deployment name into Pod name 

I guess it's not the problem to implement KubernetesPods into OpenShiftDeploymentConfigs, no need to duplicate code. Deployment based implementation of exec method for WorkspacePods may look like
```
public CompletableFuture<Void> exec(String podName, String[] command, ....) {
    Pod pod = osDeployments.findPod(podName);
    return kubernetesPods.exec(pod.getMetadata().getName(), command ...);
}
```
So, this class would contain only logic to mapping Pod <> DeploymentConfig.
To make sure which way is the clearest - need to try and compare. So, up to you which way is the best. I'm OK even to leave extending KubernetesPods by OpenShiftDeploymentConfigs while it looks like the simplest way. Only it would be  better to add note that OpenShiftDeploymentConfigs is not really KubernetesPods extender in OOP term but wrapper, which wraps KubernetesPods and redirects calls to right pod, something like that.

While I was writing this message I've remembered that we use KubernetesPods in PVCSubpathHelper, I think we really don't need deployment config for precreate/remove jobs. So, with extending it is impossible to create job anymore, with MachinePods/WorkspacePods is still possible to inject KubernetesPods directly and create job pod.

garagatyi (commented):
Overall looks good. I believe that this PR needs to be carefully tested since it changes very basic operations and something like what @sleshchenko described with PVC might be broken. 
We should also think whether new WSMaster can pick up old workspace pods after update of Che on OSIO with this changes. As far as I know, we don't want to stop workspaces on each update of Che, so we should be sure that this change either doesn't incorporate incompatibility or require workspaces stopping. 
I also don't like complexity and duplication that this PR adds, but don't see any proper alternative. Looks like we reached the point when we can't keep the code more or less simple. 
Regarding the last comment from @sleshchenko about PVC: since Angel used k8s deployments instead of Openshift DeploymentConfigs (recent change that might not be reflected in the PR description) we can consider moving this code to k8s infrastructure and add an ability to use Deployments, bare pods, jobs for different actions, like PVC subpath creation. It might do Openshift infrastructure code that simple again but would extend k8s infrastructure instead.
Apart from that maybe we should think what we might do if support of all the entities from k8s model is needed and how we would handle that. In that case, we would have a direct way of creation of Jobs, bare pods, deployments, etc.

amisevsk:
So after discussion today, two significant issues have come up:

1. I missed that this will cause `PVCSubpathHelper` to use Deployments instead of Pods incorrectly. This could potentially be resolved by including two methods, e.g. `createLongRunning()` and `createJob()` or something along those lines. 
2. Updating Che to include this PR will require manual termination of all running workspaces, as attempting to delete a workspace hosted in a bare Pod will fail (we'll try to delete a Deployment). I'm still trying to think out a solution to this issue. It'd be great if the transition could occur transparently.

Also, the fact that we are now using Deployments instead of Pods (i.e. still completely Kubernetes-side implementations) makes the structure of this PR make significantly less sense. I think something along the lines of what @sleshchenko suggested (introducing a new `machinePods` abstraction) is more appealing now. Ideally I would like an abstract implementation for workspace container creation, that specifies an internal API for creating, deleting, etc., with a shared implementation for Pod specific methods (e.g. `watch`, `exec`). As such, all implementations of the abstraction would only be responsible for creation and deletion, with a way of getting the actual `Pods` they are responsible for. With this, it might even be possible to allow users to choose whichever method of deploying workspaces they want via a property -- `deployments`, `pods`, or (if we can get it to work) `depoymentconfigs`. These are all still early thoughts though, so nothing is concretely formed yet. 

@garagatyi @sleshchenko WDYT?

garagatyi:
I think we don't need an ability to configure the type of pods scheduling (bare, deployment, etc). Let's stick to one solution unless we actually need to support several options. 
It seems that having `machinePods` abstraction would not help with PVC stuff, so I'm not sure we need this right now while we can stick to just one implementation of pods scheduling.  My opinion on that is not strong, so I'm OK with the opposite approach as well.  BTW we match a pod to a machine, so maybe `workspacePods` would be better wording. 
Not sure I got the solution with `createLongRunning` and `createJob` methods. 
@sleshchenko @l0rd @amisevsk how do you think should we consider a way to gracefully migrating workspaces to pods in deployments? Maybe someone has any ideas how to do that without a big chunk of code?

amisevsk:
@garagatyi  @sleshchenko I've redone most of this PR. The changes as of now are:
- Move deployments functionality into `KubernetesPods` class, and rename to `KubernetesDeployments` to reflect this change.
- Add method "createTemporary(Pod pod)` for creating bare pods, since that is still useful for things like PVC cleanup jobs.
- Merge deletion code so that `KubernetesDeployments` can cleanup workspaces whether they have started via deployments or pods. This means that we can do a hot update of che-server, and it will correctly clean up pod-based workspaces.
  - This code does not need to be deprecated as we had discussed earlier, as the pod cleanup aspect is still useful for cleaning up 'temporary' jobs. 

Any thoughts / suggestions?

garagatyi:
@amisevsk can you resolve conflicts in the PR?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
```diff
@@ -104,13 +114,47 @@
...
-   * @param pod pod to create
+   * @param pod pod to deploy
    * @return created pod
    * @throws InfrastructureException when any exception occurs
    */
   public Pod create(Pod pod) throws InfrastructureException {
+    putLabel(pod, CHE_WORKSPACE_ID_LABEL, workspaceId);
+    ObjectMeta metadata = pod.getMetadata();
+    PodSpec podSpec = pod.getSpec();
+    podSpec.setRestartPolicy("Always"); // Only allowable value
```
garagatyi: Should we restart a pod if a container exits with the exit code 0? I'm not sure. It is kinda a situation that should not happen, right? So maybe it makes sense to use OnFailure? I don't have a strong opinion on that, just thinking about such a use case.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
```diff
@@ -104,13 +114,47 @@
...
+  }
+
+  /**
+   * Create a terminating pod that is not part of a Deployment.
+   *
+   * @param pod the Pod to create
+   * @return the created pod
+   * @throws InfrastructureException when any error occurs
+   */
+  public Pod createTemporary(Pod pod) throws InfrastructureException {
```
garagatyi: Maybe we should use some kubernetes native vocabulary not to confuse with a new concept of a temporary pod?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
```diff
@@ -104,13 +114,47 @@
...
+              .endSpec()
+              .done();
+      return awaitAndGetPod(deployment);
+    } catch (KubernetesClientException e) {
+      throw new KubernetesInfrastructureException(e);
+    }
+  }
+
+  /**
+   * Create a terminating pod that is not part of a Deployment.
```
garagatyi, line 151: The class is called Deployment but methods create a pod that is not a part of the deployment. I think it may be a bit confusing.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
```diff
@@ -186,13 +234,13 @@ public void eventReceived(Action action, Pod pod) {
...
                       new InfrastructureException(
-                          "Waiting for pod '" + name + "' was interrupted"));
+                          "Waiting for pod '" + podName + "' was interrupted"));
                 }
               });
 
       Pod actualPod = podResource.get();
       if (actualPod == null) {
-        throw new InfrastructureException("Specified pod " + name + " doesn't exist");
+        throw new InfrastructureException("Specified pod " + podName + " doesn't exist");
```
garagatyi: There is a situation when deployment was specified and the message may be confusing because it says that podName was specified while it wasn't.

garagatyi (commented):
@amisevsk you did a great job in making this change small and convenient for the Che master update! And the drawback of this is that code maintainability might be significantly affected. 
It seems we need to think about better separation/abstraction not to make code too complex.
@amisevsk @sleshchenko WDYT?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
amisevsk: Sadly, the only supported option in Deployments (at least on OpenShift) seems to be "Always". Setting it to "OnFailure" causes the `POST` request to be rejected. For what it's worth however, I manually failed a pod, and the workspace was detected as not running by the server (due to wsagent crashing) quickly and it showed a fairly understandable prompt to restart the workspace.

While the deployment does recreate a Pod that does nothing, it doesn't seem to have a significant negative impact. Hopefully in the future we'll be able to leverage the built in restart functionality to deal with failing workspaces through Kubernetes/OpenShift directly.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
amisevsk: I agree, the naming in this class has become quite confusing. There's now somewhat of a split role in KubernetesDeployments, as its becoming more of a utility class for managing Pod resources (both through bare pods and deployments) than a simple wrapper around Pods. 

In picking this name (which I agree is not very good), I was trying to choose something that would at least not add to the confusion -- calling it `createJob` is deceptive as it currently doesn't create jobs, and calling it `createPod` is confusing since both create methods result in the creation of pods. I'm starting to lean towards finding a sensible non-kubernetes vocabulary way of describing what this class does, and renaming it away from `KubernetesDeployments`

Perhaps
```
KubernetesDeployments -> KubernetesComputeResources
    .create() -> .createNotTerminating()
    .createTemporary() -> .createTerminating()
```
This would at least bring it in line with the language used in describing [resource quotas](https://kubernetes.io/docs/concepts/policy/resource-quotas/). WDYT?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
amisevsk: I agree, see the comment above about renaming to `create[Not]Terminating()` to better reflect the intended purpose of this class.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
amisevsk: Yeah this is a tricky one. One of the things I'm not quite happy with in this PR is that as a result of trying to not affect the rest of the internal workspaces representation (as Pods are used extensively), I've had to keep the underlying way of running pods invisible to che-server. While this allows most of Che to work without any modification, within this class, it makes it confusing as you point out, as it's possible to call this method with a deployment name. 

If this message were to be logged, the two causes are 
1. The method is called with a deployment name, and `getPodName()` succeeds, returning the name of a running pod controlled by the specified deployment, *and* that running pod terminates between the call to `getPodName()` and this `get()` call. In this case, the log message should be something along the lines of `No pods in Deployment [name] found`
2. The method is called with a pod name directly (i.e. there is no deployment), and the pod that we `get()` in the `getPodName()` method terminates before we reach this point. In this case ideally we would log the message currently there. 

Does this make sense, or can we make it more understandable? This check should really never fail, as `getPodName` always calls `get()` on the same `PodResource` and throws an error earlier if `null` is encountered.

amisevsk: I've modified the logging to reflect the above, leveraging the fact that if `name.equals(podName)` then `name` refers to a bare pod. It does make the already-long error handling block in these methods longer, however. Let me know what you think.

amisevsk:
@garagatyi Regarding the abstraction and separation question, I'm definitely open to moving things around if it makes more sense. For me, since so much of the functionality is shared between the two ways of launching pods (deployments vs. bare pods), it makes sense to keep everything in one class rather than have imports and calls across multiple classes. In reality, the only methods that care whether we are running in a bare pod or a deployment are the create() and delete() methods, as they have to do something different. Most of the other classes operate / interact with the pods themselves directly. Additionally, keeping everything together like this ensures that we can pick up old workspaces running in bare pods and handle them transparently if che server is hot updated.

Do you have any ideas on a cleaner separation between the two?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
sleshchenko: I would say that will be created bare pod terminating or not depends only on infrastructure configuration.

Maybe `#deploy(Pod)` and `create(Pod)` will make more clear that one method creates pod directly while another one wraps pod into Deployment. Then we have to document which limitations have creation bare pod.

Another way to make the clear purpose of `create(Pod)` method may be using something like that `create(Pod pod, String[] command)`. So it would be clear that this method should be used only if a user wants to execute a one-time task. We would even be able to move `PVCSubPathHelper#execute(String, String[], String...)` to KubernetesDeployment, so not only PVCSubPathHelper would be able to easily execute a command (submit task).
Guys. WHYT? Does it make sense?

sleshchenko (commented):
About separation, I think it may do code cleaner if we would have two separate classes for `KubernetesDeployments` and `KubernetesPods`, then `KubernetesDeployments` would have only logic related to managing deployments, mapping deployment to Pod and reuse method of `KubernetesPods` to work with pods.

>  Additionally, keeping everything together like this ensures that we can pick up old workspaces running in bare pods and handle them transparently if che server is hot updated.

I think it's possible to do the same with two classes. The only one thing that should be implemented in `KubernetesDeployment` - try to pick up bare pod if pod managed by deployment was not found while pod to deployment mapping. And it may be removed after few releases.

garagatyi:
Right now I would go with the next solution:
- rename creation of bare pods and deployment. create+deploy seems a good wording. 
- do not separate classes since it seems we don't have a strong vision on how it must look like
- do not move exec functionality from PVCSubPathHelper since it is not the goal of this task and in any case, those are just non-critical ideas how to improve the code. 

Seems that we don't have enough certainty about the naming, so let's strip down the amount of work to apply to this PR and continue improving code in future when that certainty or necessity comes.

amisevsk:
Alright I have updated the PR with the method name changes suggested. I think that the documentation still makes sense but I may have missed something in the refactor. Let me know if there is anything else that needs to be changed before a rebase.

@sleshchenko I like your idea of splitting it into two classes, so I would like to create an issue after this is merged where we can discuss it further and settle on a design choice before starting implementation. Does that sound good to you? I'm mostly concerned around increased confusion since a `KubernetesDeployments` class would have a lot of shared code and methods with `KubernetesPods`, so there would ideally be some inheritance between the two. In fact, the interface for both classes would probably be identical.

garagatyi (approved):
PR looks good. Please, run CI tests before merging it

amisevsk:
ci-test

benoitf:
ci-test

sleshchenko (approved):

mmclanerh:
In the case where a pod has tens of thousands of files, it can take several minutes for the chown to happen on mount. The deploymentconfig here will need to take this into account and hopefully not terminate the pod spin up prematurely.

ibuziuk:
@mmclanerh does `tens of thousands of files` is expected or abnormal workspace state ? Probably, the issue you are raising in the last comment is smth. reported in rh-che already (old workspace files were not cleaned from PVs) https://github.com/redhat-developer/rh-che/issues/730 ?

ibuziuk:
@amisevsk when you are going to merge do not forget to remove [WIP] from description since PR title would be used for the changes log

eivantsov:
I'd be great to update docs too.

amisevsk:
@eivantsov I'm happy to take a pass over the update. Are there any sections of the documentation you think I should focus on (e.g. Kube / OS admin guides?)

eivantsov:
@amisevsk I think I will start a branch and update deployment images that explain what objects we create, and we can go from there.

Issue #10008, opened 2018-06-12
Selenium: add option not to start test workspaces after creation


SkorikSergey:
### Description
When we create test workspaces in selenium tests, it always automatically starts after creation. It works fine for **IDE** tests, but unnecessary in tests when we just need to create some workspaces to work with them on the **Dashboard**(change workspace details, deleting, sharing).

PR #10022, opened 2018-06-12
Add option not to start test workspaces after creation

dmytro-ndp:
## What does this PR do?
It allows to define test workspace which should be created without starting just after creation, for example:
```
@InjectTestWorkspace(startAfterCreation = false)
TestWorkspace testWorkspace;
```

#10008


dmytro-ndp:
ci-test

SkorikSergey (approved):

Issue #10018, opened 2018-06-12
Che Server has wrong value for ROUTING_SUFFIX when non-default DNS provider is used


sleshchenko:
### Description
Che Server has wrong value for ROUTING_SUFFIX when non-default DNS provider is used.

<details>
<summary>Konsole output </summary>

```
[serg@localhost projects]$ ocp --run-ocp --multiuser --no-pull --debug --deploy-che
download oc client v3.9.0
download jq...
Test nip.io DNS provider failed, trying next one.
Test xip.io DNS provider failed, trying next one.
Test nip.codenvy-stg.com - works OK, using it as DNS provider
Deleted existing OpenShift container
Using nsenter mounter for OpenShift volumes
Using public hostname IP 172.17.0.1 as the host IP
Using 172.17.0.1 as the server IP
Starting OpenShift using openshift/origin:v3.9.0 ...
OpenShift server started.

The server is accessible via web console at:
    https://172.17.0.1:8443

You are logged in as:
    User:     developer
    Password: <any value>

To login as administrator:
    oc login -u system:admin
                                                                                                                                                                                                                                             
[OCP] wait for ocp full boot...                                                                                                                                                                                                              
Logging in to OpenShift cluster...                                                                                                                                                                                                           
                                                                                                                                                                                                                                             
                10100000                                                                                                                                                                                                                     
            0110011001101101                                                                                                                                                                                                                 
       01001001001001011011010110                                                                                                                                                                                                            
   1100110101010001000001011101000001                                                                                                                                                                                                        
0111001001111010111011111000111100010001                                                                                                                                                                                                     
1110111011010111000001100000111001110                                                                                                                                                                                                        
101010011001001001    10011101000                                                                                                                                                                                                            
1011000101000              0                                                                                                                                                                                                                 
100101011                           0110                                                                                                                                                                                                     
10100001                        11011111                                                                                                                                                                                                     
111101111011                111011110001                                                                                                                                                                                                     
10111001000110101      00000111111011111                                                                                                                                                                                                     
0110111111011100000001011001010101111111                                                                                                                                                                                                     
1101000001101001000010010111101011011011                                                                                                                                                                                                     
    00000001111001000001110111100000                                                                                                                                                                                                         
        000110101101000100011110                                                                                                                                                                                                             
            1101011100001001                                                                                                                                                                                                                 
                 000101                                                                                                                                                                                                                      
                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                             
[...]

PR #10024, opened 2018-06-13
Fix ocp.sh

ghost:
### What does this PR do?

Removes unnecessary env OPENSHIFT_ROUTING_SUFFIX

Wraps set volume command in `" "` to make it possible to use deploy_che.sh in Git bash 

https://github.com/eclipse/che/issues/10018


sleshchenko (approved):
Thanks for quick fix

l0rd (approved):

PR #10028, opened 2018-06-13
Simple Java project is not recognized #9526

### LINKED ISSUE #9526: Simple Java project is not recognized
### Description
Existed plain java project is not recognized as a project when the workspace is restarted.

anim

vrubezhny:
### What does this PR do?

The fix listens for the Language Server initialization StatusReport events from eclipse.jdt.ls and requests to update existing projects upon receiving LS's `Started` status.

Fixes #9526 

#### Release Notes

ls9526

Project Explorer tree now get's updated during the start up when `Started` StatusReport event is received for the Language Server.

### DIFF
--- a/plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/JavaLanguageServerLauncher.java
+++ b/plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/JavaLanguageServerLauncher.java
@@ -29,10 +29,18 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
+import org.eclipse.che.api.core.BadRequestException;
+import org.eclipse.che.api.core.ConflictException;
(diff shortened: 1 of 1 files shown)


tolusha (approved):
Don't forget to remove extra logging before merging.

plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/JavaLanguageServerLauncher.java
```diff
@@ -52,18 +55,47 @@
...
+  private void updateWorkspaceOnLSStarted() {
+    projectManager
+        .getAll()
+        .forEach(
+            registeredProject -> {
+              if (!registeredProject.getProblems().isEmpty()) {
+                try {
+                  projectManager.update(registeredProject);
+                  eventService.publish(new ProjectUpdatedEvent(registeredProject.getPath()));
+                } catch (Exception e) {
```
tsmaeder: This would catch runtime exceptions, etc. I this what we want? I would rather use multi-catch to catch expected Exceptions.

Issue #10034, opened 2018-06-14
Fix Rename Selenium Tests Waiting for OK


tsmaeder:
Currently, Selenium tests fail on the jdt.ls branch because they seem to wait for a confirmation dialog of some sort, for example: org.eclipse.che.selenium.refactor.types.RenameTypeTest

org.openqa.selenium.TimeoutException: 
Expected condition failed: waiting for visibility of Proxy element for: DefaultElementLocator 'By.xpath: //button[@id='ask-dialog-ok']/ancestor::table' (tried for 10 second(s) with 500 MILLISECONDS interval)
Build info: version: '3.0.1', revision: '1969d75', time: '2016-10-18 09:49:13 -0700'
System info: host: 'slave4.codenvycorp.com', ip: '127.0.0.1', os.name: 'Linux', os.arch: 'amd64', os.version: '3.10.0-514.16.1.el7.x86_64', java.version: '1.8.0_102'
Driver info: org.eclipse.che.selenium.core.SeleniumWebDriver
	at org.openqa.selenium.support.ui.WebDriverWait.timeoutException(WebDriverWait.java:80)
	at org.openqa.selenium.support.ui.FluentWait.until(FluentWait.java:265)
	at org.eclipse.che.selenium.core.webdriver.SeleniumWebDriverHelper.waitVisibility(SeleniumWebDriverHelper.java:154)
	at org.eclipse.che.selenium.core.webdriver.SeleniumWebDriverHelper.waitVisibility(SeleniumWebDriverHelper.java:164)
	at org.eclipse.che.selenium.pageobject.AskDialog.waitFormToOpen(AskDialog.java:74)
	at org.eclipse.che.selenium.refactor.types.RenameTypeTest.testCase(RenameTypeTest.java:196)
	at org.eclipse.che.selenium.refactor.types.RenameTypeTest.test1(RenameTypeTest.java:111)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:85)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:696)
[...]

PR #10080, opened 2018-06-18
Show warning dialog after refactoring operation if something isn't valid

svor:

### What does this PR do?
After preparing changes of refactoring operation it may have some error/warning status. So the main idea of current PR is to show warning dialog before applying changes.

https://github.com/eclipse/che/issues/10034


tolusha (approved):

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/refactoring/rename/JavaRefactoringRename.java
```diff
@@ -307,10 +309,13 @@ private void performRename(String newName) {
     extensionServiceClient
         .rename(settings)
         .then(
-            edits -> {
+            result -> {
               enableAutoSave();
               undoChanges();
-              applyWorkspaceEditAction.applyWorkspaceEdit(edits);
+              RefactoringStatus refactoringStatus = result.getRefactoringStatus();
+              if (!FATAL.equals(refactoringStatus.getRefactoringSeverity())) {
```
tsmaeder: How do we inform the user that something went wrong in this case?

tsmaeder (approved):

plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/refactoring/rename/JavaRefactoringRename.java
svor: added a way to show notification with an error message

Issue #9978, opened 2018-06-08
Make theia Docker image be able to use yarn cache on Openshift


garagatyi:
### Description
Current Theia image when is started on Openshift can't use yarn cache folder from docker image build `/usr/local/share/.cache/yarn/v1`. Because of that usage of `THEIA_PLUGINS` env var in Theia container makes it fetch all the dependencies from scratch and workspace start fails because of timeout.
On Docker cache is used and start is fast enough. 
So, we need to fix it somehow. 
`yarn cache dir` returns this on Openshift:
```
/home/theia $ yarn cache dir
warning Skipping preferred cache folder "/.cache/yarn" because it is not writable.
warning Selected the next writable cache folder in the list, will be "/tmp/.yarn-cache-1000090000".
/tmp/.yarn-cache-1000090000/v1
/home/theia $
```

### Reproduction Steps

**OS and version:**    

**Diagnostics:**

garagatyi:
@eivantsov It prevents us from being able to use Theia extensions on Openshift, so I will be very grateful if you know how to fix that sooner than later.

benoitf:
this docker image https://hub.docker.com/r/wiptheia/wip/ includes all the stuff to handle openshift (and openshift.io)

eivantsov:
@benoitf but you do not run theia build as you use this one for the new plugin model, correct?

benoitf:
@eivantsov I used yarn inside, in order to create from scratch a plugin with a yeoman generator so I would say we're able download npm dependencies, etc. Before it was not possible due to "invalid permission" or 'directory not writeable' errors, etc.

eivantsov:
@benoitf I only see these dirs writable for root group https://github.com/theia-demo-plugins/wiptheia-docker-image/blob/master/Dockerfile#L50

What is the secret then? 

With wiptheia you are not running yarn build against theia itself, right? Just building the plugin? Maybe I am missing smth, but looking at the Dockerfile does not give me a clear answer.

benoitf:
https://github.com/theia-demo-plugins/wiptheia-docker-image/blob/master/Dockerfile#L23 ?

eivantsov:
Well, that's user with id 1001 - it won't be writable unless you do what you do on line 50. Probably you do that to make jsons and js files have user permissions as they are probably ADDed with root ones.

Can you please clarify, if this image is used for Theia using the new plugin model, ie when Theia isn't rebuilt to include a new plugin?

[...]

PR #10116, opened 2018-06-20
Make theia Docker image be able to use yarn cache on Openshift

mmorhun:
### What does this PR do?
Adds settings and corresponding permissions into docker image with Theia to be able to use yarn cache on openshift

https://github.com/eclipse/che/issues/9978

### DIFF
--- a/dockerfiles/theia/Dockerfile
+++ b/dockerfiles/theia/Dockerfile
@@ -13,7 +13,8 @@ ARG GITHUB_TOKEN
 ARG THEIA_VERSION=0.3.10
 ENV USE_LOCAL_GIT=true \
     GITHUB_TOKEN=${GITHUB_TOKEN} \
-    THEIA_VERSION=${THEIA_VERSION}
+    THEIA_VERSION=${THEIA_VERSION} \
+    YARN_CACHE_FOLDER=/usr/local/share/.cache/yarn
 
 # build dependencies requireed to compile a custom Theia
 RUN apk add --no-cache make gcc g++ python git openssh bash supervisor jq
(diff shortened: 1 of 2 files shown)


ghost (approved):

dockerfiles/theia/Dockerfile
```diff
@@ -23,16 +24,18 @@ WORKDIR /home/theia
...
-RUN cd /home/theia && \
+RUN cd /home/default && ./versions.sh && \
+    cd /home/theia && \
     yarn && \
     yarn theia build --mode development && \
     rm -rf * && \
     cd /home/default/theia && \
     yarn && \
     yarn theia build
+RUN find ${YARN_CACHE_FOLDER} -exec sh -c "chgrp 0 {}; chmod g+rwX {}" \;
```
benoitf, line 36: should we group all RUN instructions together ?

benoitf:
I think it doesn't solve issue on write access on some folder
some of fixes seem not be backported from https://github.com/theia-demo-plugins/wiptheia-docker-image/blob/master/Dockerfile

dockerfiles/theia/Dockerfile
mmorhun: Do not think so. It could be helpful if we add something and then delete it at the end of `RUN`. But in this case it still remains we cannot reduce image size. What we will have - a single layer with all the changes, which is not so good, because it makes using caches harder.

mmorhun:
@benoitf it is not backport of https://github.com/theia-demo-plugins/wiptheia-docker-image/blob/master/Dockerfile. This PR solves problems with caches on openshift.

AndrienkoAleksandr (approved):

mmorhun:
We shouldn't replace this dockerfile with wiptheia one. Otherwise we'll lose fix for caches problem. I think we should merge them but I would like to do it in a separate PR.

benoitf:
@mmorhun yes when I said "replaced" it was more "merge"
ok in a separate PR

benoitf:
but my meaning was more that we could have fixed wip dockerimage and then drop this one and use the new one. As this image is more broken than the wip-dockerimage

mmorhun:
Yes, I agree that we have better image in wiptheia than here in Che and it defenetly should be considered.

benoitf (approved):
would be better by merging RUN but can be reworked with the upcoming merge

garagatyi (approved):
Looks good!

dockerfiles/theia/Dockerfile
mmorhun: But it will happen anyway and doesn't depend on number of layers...

Issue #10124, opened 2018-06-21
Merge wiptheia dockerfile with Che Theia IDE dockerfile


mmorhun:
### Description
After merginig wiptheia into upstream we need to move all stuff from wiptheia organization.
There we have a dockerfile for wiptheia nightly image and it should be merged into Che's one.

mmorhun:
Done

PR #10148, opened 2018-06-22
Merge wiptheia dockerfile with Che Theia IDE dockerfile

mmorhun:
Signed-off-by: Mykola Morhun <mmorhun@redhat.com>

### What does this PR do?
Combines features of witheia docker image with Che one.
Removes unneeded user and `sudo` from the image.
Drops support for Theia extensions vie env variable. Now thay must be added on docker image build phase.

https://github.com/eclipse/che/issues/10124


mmorhun:
This PR uses `next` version of Theia in the image because we don't have released Theia modules with `plugin-ext` extension. Also patched (for the `next` version) extensions were used. I added `==TODO==`s what should be fixed after the module will be released.

It is open for discussions.

benoitf:
hello, could you clarify the 'uneeded user' ?
with the changes, this image is working on che.openshift.io ?

AndrienkoAleksandr:
Maybe we should use next with hash commit instead of next version?

mmorhun:
Yes. I see no pint in having it in the image. I've just checked it on docker and openshift with machines extension - and it works fine (with `next` versions of course)

benoitf:
when you say "openshift", is it openshift.io ?

mmorhun:
It is local ocp

dockerfiles/theia/Dockerfile
```diff
@@ -10,33 +10,46 @@
...
-    rm -rf * && \
-    cd /home/default/theia && \
+    USE_LOCAL_GIT=true \
+    HOME=/home/theia
+
+RUN apk update && apk add --no-cache make gcc g++ python git openssh bash supervisor jq && \
+    rm -rf /tmp/* /var/cache/apk/*
+
+# Packages which will be included into default build of Theia
+ADD theia-default-package.json ${HOME}/package.json
```
benoitf: as we're adding all files into ${HOME}
maybe we could use a single ADD instruction for everything that goes into /home/theia
(instead of one for package.json, one for versions.sh, one for add-extensions.js and one for start.sh

we could have folders like
```
src/home/theia --> /home/theia
src/etc/ ---> /etc/
```

mmorhun:
I updated this PR. Tested on docker and openshift.io.
Review again please.

AndrienkoAleksandr (approved):

dockerfiles/theia/Dockerfile
mmorhun: To me it is better to keep it as is because that `ADD` instructions appears when target file is needed. For example:
```
ADD src/versions.sh ${HOME}/versions.sh
RUN ${HOME}/versions.sh && rm ${HOME}/versions.sh
```
We add, use and finally remove file. No need to search whole dockerfile and think about context (who else could use or modify or require it).

evidolob (approved):

ghost (approved):

benoitf (changes requested):
image is almost 1GB, IMHO there are layers that need to be skipped

eivantsov:
@mmorhun I'd combine all RUN instructions if possible and have one ADD instruction that adds all files and scripts to one directory from which you can then consume them in a Dockerfile

dockerfiles/theia/Dockerfile
```diff
@@ -10,33 +10,46 @@
...
+    # ==TODO== add after generator-theia-plugin release
+    # npm install -g yo @wiptheia/generator-theia-plugin
+# Change permissions to allow editing of files for openshift user
+RUN find ${HOME} -exec sh -c "chgrp 0 {}; chmod g+rwX {}" \;
 
+EXPOSE 3000 3030
+WORKDIR ${HOME}
+ADD supervisord.conf /etc/
+ADD src/start.sh ${HOME}/start.sh
+RUN touch /var/log/supervisord.log && chmod g+rwX /var/log/supervisord.log && chgrp 0 /var/log/supervisord.log
```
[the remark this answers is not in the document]
mmorhun: I think it will impact how easy to read this dockerfile is. But ok, I'll move it.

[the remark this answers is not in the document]
benoitf: @mmorhun I would say that workspace should start as fast as possible (minimal set of layers, best size as possible)

dockerfiles/theia/Dockerfile
```diff
@@ -10,33 +10,46 @@
...
-RUN cd /home/default && ./versions.sh && \
-    cd /home/theia && \
-    yarn && \
-    yarn theia build --mode development && \
-    rm -rf * && \
-    cd /home/default/theia && \
+    USE_LOCAL_GIT=true \
+    HOME=/home/theia
+
+RUN apk update && apk add --no-cache make gcc g++ python git openssh bash supervisor jq && \
```
benoitf: on a side note, I notice that you're installing openssh but can we really connect through ssh ?

mmorhun: Yeah, good catch. It is result of merge. Will remove it.

benoitf:
on a side note, is it possible to make GITHUB_TOKEN optional ?
it seems it is mandatory to build the docker image while it should only be used if defined

mmorhun:
No, it is not mandatory. One could provide it as a build arg.

benoitf:
@mmorhun 
If I don't export GITHUB_TOKEN to a valid token before launching `./build.sh`
I got 
```
error An unexpected error occurred: "/home/theia/node_modules/vscode-ripgrep: Command failed.
Exit code: 1
Command: sh
Arguments: -c node ./lib/postinstall.js
Directory: /home/theia/node_modules/vscode-ripgrep
Output:
Downloading ripgrep failed: Error: Request failed with code 401".
```

(it's like it provides an empty value token)

eivantsov:
@benoitf this is a known limitation of Theia

benoitf:
@eivantsov I don't see how it's a limitation of theia ?

I can clone theia, build theia and run theia without defining GITHUB_TOKEN

eivantsov:
Yes, Theia will use an empty token and build may be a success unless you hit access rate on Github.

Everyone had such an issue when building Theia.

Btw, do we still need GITHUB_TOKEN after https://github.com/theia-ide/theia/issues/2091?

mmorhun:
@eivantsov seems yes. I've tried `next` version and the problem was there.

eivantsov:
@benoitf I don't know a good way to know in advance if or not `vscode-ripgrep` fails that's why GITHUB_TOKEN was added

As the image is built mostly by CI it wasn't a problem.

If you have a better way to build Theia and make one of its components not to download anything from GitHub, we will all appreciate it.

benoitf:
`curl https://api.github.com/rate_limit` ?

benoitf:
BTW I'm not trying to argue that GITHUB_TOKEN is a bad idea and that it doesn't solve an issue. just that it seems it's forcing ppl to give that parameter even if they don't require it.

with `curl https://api.github.com/rate_limit` you can check the limit rate and probably you can see if it will fail or not. In that case we could say : Please define GITHUB_TOKEN to make the build OK.

eivantsov:
Thanks so much. This will of course add complexity to a Dockerfile. An additional script?

benoitf:
yes maybe we can split long RUN instructions to small shell scripts to be more readable.
RUN fork-theia && compile-theia && update-extensions

I'm fine with all solutions as long as I may not be forced to give GITHUB_TOKEN if not required

mmorhun:
So, we can split it, but I added comments to the sections and finally it doesn't look huge and unreadable.

dockerfiles/theia/Dockerfile
```diff
@@ -10,33 +10,44 @@
...
+RUN \
+    # Install needed software
+    apk update && apk add --no-cache make gcc g++ python git bash supervisor jq && \
+    rm -rf /tmp/* /var/cache/apk/* && \
+    # Change version of Theia to specified in THEIA_VERSION
+    ${HOME}/versions.sh && rm ${HOME}/versions.sh && \
+    # Add Theia extensions
+    #==TODO== replace with Eclipse official version after Theia release
+    #git clone https://github.com/eclipse/che-theia-hosted-plugin-manager-extension /tmp/hosted-plugin-extension && \
+    git clone https://github.com/mmorhun/che-theia-hosted-plugin-manager-extension /tmp/hosted-plugin-extension && \
```
benoitf: what is the difference between mmorhun/che-theia-hosted-plugin-manager-extension and https://github.com/eclipse/che-theia-hosted-plugin-manager-extension ? could it just be a branch of main repo ?

dockerfiles/theia/Dockerfile
```diff
@@ -10,33 +10,44 @@
...
-    yarn theia build
-RUN find ${YARN_CACHE_FOLDER} -exec sh -c "chgrp 0 {}; chmod g+rwX {}" \;
-RUN mkdir -p /.yarn && chgrp 0 /.yarn && chmod g+rwX /.yarn && \
-    touch /.yarnrc && chgrp 0 /.yarnrc && chmod g+rwX /.yarnrc
-ADD src/main.js /theia_launcher/theia_launcher.js
-EXPOSE 3000
+    yarn theia build && \
+    # Install Theia plugin generator
+    #==TODO== add after generator-theia-plugin release
+    #npm install -g yo @wiptheia/generator-theia-plugin && \
```
benoitf, line 47: you think we shouldn't include the wip package for now ?

dockerfiles/theia/Dockerfile
mmorhun: Yes, sure.

dockerfiles/theia/Dockerfile
```diff
@@ -10,33 +10,44 @@
...
+    # Change permissions to allow editing of files for openshift user
+    find ${HOME} -exec sh -c "chgrp 0 {}; chmod g+rwX {}" \; && \
+    # Grant permissions for modifying supervisor log file
+    touch /var/log/supervisord.log && chmod g+rwX /var/log/supervisord.log && chgrp 0 /var/log/supervisord.log
 
+WORKDIR ${HOME}
```
benoitf: (not important) could we move WORKDIR at the top of the file before RUN instruction so we can reuse that layer every time

dockerfiles/theia/Dockerfile
mmorhun: I think we need release it first. Do not want to download wiptheia core and stuff.

dockerfiles/theia/Dockerfile
```diff
@@ -10,33 +10,44 @@
 FROM node:8-alpine
 
 ARG GITHUB_TOKEN
-ARG THEIA_VERSION=0.3.10
-ENV USE_LOCAL_GIT=true \
-    GITHUB_TOKEN=${GITHUB_TOKEN} \
+ARG THEIA_VERSION=0.4.0-next.b17727c1
```
benoitf, line 13: why using a specific version of next instead of next ?
we may want to include all upstream theia fixes on nightlies

mmorhun: So you want to use `next`?

mmorhun: I do not mind

mmorhun: Just to be sure we have working image (I tested it on both docker and oso infras)

dockerfiles/theia/Dockerfile
benoitf: ah yes I forgot to push changes that use @theia/plugin
let's keep commented as you made 👍

dockerfiles/theia/Dockerfile
mmorhun: No problem

benoitf (approved):
we can merge and iterate later on the changes

dockerfiles/theia/Dockerfile
AndrienkoAleksandr: It will be confused when we move back to the strict Theia version, but dockerhub save "next" tag. @benoitf What do you think? Somebody can try to use it, but it won't be next any more. And if we have next tag we should update it every night.

benoitf: @AndrienkoAleksandr I would think to have a real "nightly" based on "next" and a fixed version as suggested by @mmorhun

dockerfiles/theia/Dockerfile
mmorhun: Done

benoitf:
FYI I created https://github.com/eclipse/che/issues/10205 for improvements

mmorhun:
Thanks

Issue #10052, opened 2018-06-14
Selenium: Source Folder Paths have wrong base


tsmaeder:
ConfigureClasspathBaseTest fails because of the value the source folder paths we show in the dialog:
expected is '/ConfigureClasspathBase_5fah/my-lib/src/main/java', but we show '/my-lib/src/main/java'. I suspect that we are using the eclipse project path instead of the locationURI.

org.openqa.selenium.TimeoutException:  Expected condition failed: waiting for org.eclipse.che.selenium.pageobject.ConfigureClasspath$$Lambda$194/1036591098@1b286f4d (tried for 5 second(s) with 500 MILLISECONDS interval) Build info: version: '3.0.1', revision: '1969d75', time: '2016-10-18 09:49:13 -0700' System info: host: 'slave4.codenvycorp.com', ip: '127.0.0.1', os.name: 'Linux', os.arch: 'amd64', os.version: '3.10.0-514.16.1.el7.x86_64', java.version: '1.8.0_102' Driver info: org.eclipse.che.selenium.core.SeleniumWebDriver 	at org.openqa.selenium.support.ui.WebDriverWait.timeoutException(WebDriverWait.java:80) 	at org.openqa.selenium.support.ui.FluentWait.until(FluentWait.java:265) 	at org.openqa.selenium.support.ui.FluentWait.until(FluentWait.java:205) 	at org.eclipse.che.selenium.pageobject.ConfigureClasspath.waitExpectedTextJarsAndFolderArea(ConfigureClasspath.java:162) 	at org.eclipse.che.selenium.plainjava.ConfigureClasspathBaseTest.checkConfigureClasspath(ConfigureClasspathBaseTest.java:83) 	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) 	at 
<snip>

PR #10153, opened 2018-06-22
Use location instead of Eclipse workspace path for classpath

tsmaeder:
### What does this PR do?
Uses location URI instead of eclipse workspace path. This is basically a bugfix

https://github.com/eclipse/che/issues/10052
#### Release Notes
n/a: this reestablishes the same behaviour as in master

#### Docs PR
None


tolusha (approved):

svor (approved):
@tsmaeder just a question why did you move dto maven plugin to `plugins/plugin-java/che-plugin-java-ext-jdt/org-eclipse-jdt-ui/pom.xml`? 
As for me we need to remove `org-eclipse-jdt-ui` module

tsmaeder:
@svor the code generation in the shared project is a left over from very old days and simply wrong. I have removed it in other places where it made sense before and now had to do it here. 
Removing jdt.ui is a next step.

Issue #9946, opened 2018-06-05
Better failure model when PVC fails during workspace creation


dneary:
When starting a new workspace, if the PVC fails on the underlying infrastructure (OpenShift), for whatever reason, there is no immediate feedback to the user. The workspace loader spins, the mount is retried and re-fails, until the operation times out (60s), and returns a generic error message ("Workspace failed to start". There are "Failed to mount PV" errors reports in the OpenShift pod events for the workspace, but this information does not seem to propagate back up to Che.

I believe that we should be able to detect that a PVC failed when creating the workspace, and then inform the user that there was a problem allocating storage for the workspace.

eivantsov:
@dneary are you certain you use the latest Che? According to https://github.com/eclipse/che/blob/master/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties#L358 Failed Mount should be considered an an unrecoverable event and it should be displayed in the IDE and server logs (not sure about server logs though)

dneary:
It's a recent deployment. It says "Che 6.7.0-SNAPSHOT".

An easy way to reproduce this using NFS is to define a PV without a valid NFS share - the NFS mount fails, resulting in the PVC claim failing, and it takes ~60s for this to propagate to the user.

eivantsov:
What exact error do you see in the IDE?

dneary:
Well, I fixed it... will need to re-break it to see.

dneary:
In the OpenShift workspace pod events, I see:

> 11:50:09 AM | Warning | Failed Mount | MountVolume.SetUp  failed for volume "nfs-pv16" : mount failed: exit status 32 Mounting command: systemd-run Mounting arguments: --description=Kubernetes transient mount for  /var/lib/origin/openshift.local.volumes/pods/23c9eba3-69a1-11e8-9a4b-c03fd56fd023/volumes/kubernetes.io~nfs/nfs-pv16  --scope -- mount -t nfs  dn2.desklab.eng.bos.redhat.com:/home/dneary/nfsroot/pv16  /var/lib/origin/openshift.local.volumes/pods/23c9eba3-69a1-11e8-9a4b-c03fd56fd023/volumes/kubernetes.io~nfs/nfs-pv16 Output: Running scope as unit run-9587.scope.…         
> -- | -- | -- | --
[...]

PR #10155, opened 2018-06-22
che #9946: Better handling of unrecoverable events for k8s & openshift infra (processing both 'container' and 'pod' events)

ibuziuk:
### What does this PR do?
Better handling of unrecoverable events for k8s & openshift infra (processing both 'container' and 'pod' events). Previously only containers (sub-set of pod events) were processed. However, some unrecoverable events happens on pod level and simply not taken into account during processing `Unrecoverable` events. So, after discussion with @sleshchenko it was decided to process all pod events and make more precise filtering on a handler level e.g. `MachineLogsPublisher` will still publish only container events and `UnrecoverableEventHandler` would process both container & pod events

[WIP] status is just because I want to test it a bit more before pushing and probably add some javadocs

che #9946 

#### Release Notes
N/A

#### Docs PR
N/A (yet)

### DIFF
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/event/PodEvent.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/event/PodEvent.java
@@ -11,22 +11,23 @@
 package org.eclipse.che.workspace.infrastructure.kubernetes.namespace.event;
 
 import java.util.Objects;
+import org.eclipse.che.commons.annotation.Nullable;
 
 /**
- * The event that should be published when container change occurs, e.g. image pulled, container
- * started.
+ * The event that should be published when the pod event occurs, e.g. pulling image, created
+ * container
  *
  * @author Sergii Leshchenko
  */
-public class ContainerEvent {
+public class PodEvent {
   private final String podName;
-  private final String containerName;
+  @Nullable private final String containerName;
   private final String reason;
   private final String message;
   private final String creationTimestamp;
   private final String lastTimestamp;
 
-  public ContainerEvent(
+  public PodEvent(
       String podName,
       String containerName,
       String reason,
@@ -41,12 +42,16 @@ public class ContainerEvent {
     this.lastTimestamp = lastTimestamp;
   }
 
-  /** Returns name of pod related to container. */
+  /** Returns name of pod related to the event. */
   public String getPodName() {
     return podName;
   }
 
-  /** Returns container name which produced event. */
+  /**
+   * Returns container name produced by the event. Could be null if the event is related to the pod
+   * but not any particular 'container'
+   */
+  @Nullable
   public String getContainerName() {
     return containerName;
   }
@@ -79,7 +84,7 @@ public class ContainerEvent {
     if (o == null || getClass() != o.getClass()) {
       return false;
     }
-    ContainerEvent that = (ContainerEvent) o;
+    PodEvent that = (PodEvent) o;
     return Objects.equals(podName, that.podName)
         && Objects.equals(containerName, that.containerName)
         && Objects.equals(reason, that.reason)
@@ -95,7 +100,7 @@ public class ContainerEvent {
 
   @Override
   public String toString() {
-    return "ContainerEvent{"
+    return "PodEvent{"
         + "podName='"
         + podName
         + '\''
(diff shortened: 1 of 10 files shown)


garagatyi:
@riuvshin Is it possible to include Ilya to the list of trusted github users for the CI?

garagatyi (approved):
Looks good! An additional thank you for simplifying the code! 
Can you add more details to the PR description since it is pretty hard to understand what was changed, why and why it wasn't done before? I see the reason and I have enough context to understand it fully but it might be a bit complex overall or when the context is lost. 
Can you also exaplins why it is in the WIP state?

ibuziuk:
@garagatyi thanks for your comment. WIP is just because I want to test it a bit more. It appeared to be that events itself changed a bit between versions of openshift (e.g. on 3.6 when there is a no pv event is "FailedSheduling" and on 3.9 it is "Failed Mount" with `MountVolume.SetUp failed for volume ...` message [1] we discussed it a bit with @eivantsov yesterday

[1] https://github.com/eclipse/che/issues/9946#issuecomment-398356719

garagatyi:
Would be awesome to decide which versions of Openshift/Kubernetes we support and test events on all supported versions not to miss rewording of the error messages in some variants. (Not a requirement for this PR).
@eivantsov do we have somewhere specified supported versions of Openshift/Kubernetes?

ibuziuk:
> @riuvshin Is it possible to include Ilya to the list of trusted github users for the CI?

I believe I'm already included and can trigger the ci via `ci-test` (it does not seem be triggered automatically on PR send)

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/event/EventHandler.java
```diff
@@ -11,12 +11,12 @@
 package org.eclipse.che.workspace.infrastructure.kubernetes.namespace.event;
 
 /**
- * Defines the handling mechanism for Kubernetes container events.
+ * Defines the handling mechanism for Kubernetes events.
  *
  * @author Sergii Leshchenko
  */
-public interface ContainerEventHandler {
+public interface EventHandler {
```
sleshchenko: Would it more specified if we rename `EventHandler` to `PodEventHandler` since it handles only pods events?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/event/PodEvent.java
```diff
@@ -11,22 +11,23 @@
 package org.eclipse.che.workspace.infrastructure.kubernetes.namespace.event;
 
 import java.util.Objects;
+import org.eclipse.che.commons.annotation.Nullable;
 
 /**
  * The event that should be published when container change occurs, e.g. image pulled, container
```
sleshchenko: Please fix java docs. Now it should be published even when no container changes occur. It would be useful to add an example of a case when pod event is published.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/event/PodEvent.java
```diff
@@ -11,22 +11,23 @@
...
  * started.
  *
  * @author Sergii Leshchenko
  */
-public class ContainerEvent {
+public class PodEvent {
   private final String podName;
-  private final String containerName;
+  @Nullable private final String containerName;
```
sleshchenko, line 24: It would be even more useful to have `@Nullable` annotation on `#getContainerName`. Also please note in javadoc of `#getContainerName` method that it may be nullable and when

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/util/ContainerEvents.java
```diff
@@ -27,17 +27,13 @@
   private ContainerEvents() {}
```
sleshchenko: Does it make sense to rename this class too?

sleshchenko (approved):
Looks good
Please take a look my comments

sleshchenko:
P.S. Don't forget to remove WIP and Issue number from PR title since it will be used for changes log

skabashnyuk (approved):

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/event/PodEvent.java
ibuziuk: fixed

ibuziuk:
ci-test

PR #10177, opened 2018-06-25
Fix a bug in the Che GWT maven plugin that prevents RhChe to build with upstream `master`

davidfestal:
### What does this PR do?

This PR fixes the root cause of the CI build failures that can be found here: https://ci.codenvycorp.com/job/rh-che-ci-master

It is a bug in the Che GWT maven plugin (in the `ProcessExcludesMojo`) that prevents `rh-che` to build correctly against the Che upstream `master` branch.

The fix consists in using the `baseVersion` (i.e. `6.8.0-SNAPSHOT` for example), instead of the precise version (i.e. `6.8.0-20180625.090508-19` for example) when trying to resolve excluded artifacts, since the precise version available in the Maven repositories can be a different between the main artifact and excluded artifacts.

It fixes CI build failures that can be found here: https://ci.codenvycorp.com/job/rh-che-ci-master


ibuziuk (approved):

l0rd (approved):

skabashnyuk (approved):

skabashnyuk:
ci-test

PR #10228, opened 2018-06-29
#10229 Publish all workspace pod events to the machine logs

ibuziuk:
### What does this PR do?
Publish all workspace pod events to the machine logs (not only container events).

image

This would enrich logs automatically with "Scheduled" & "Successful Mount Volume"

https://github.com/eclipse/che/issues/10229

N/A

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
@@ -772,19 +772,17 @@ public class KubernetesInternalRuntime<
     }
   }
 
-  /** Listens container's events and publish them as machine logs. */
+  /** Listens pod events and publish them as machine logs. */
   public class MachineLogsPublisher implements PodEventHandler {
 
     @Override
     public void handle(PodEvent event) {
       final String podName = event.getPodName();
-      final String containerName = event.getContainerName();
       try {
         for (Entry<String, KubernetesMachineImpl> entry :
             machines.getMachines(getContext().getIdentity()).entrySet()) {
           final KubernetesMachineImpl machine = entry.getValue();
-          if (machine.getPodName().equals(podName)
-              && machine.getContainerName().equals(containerName)) {
+          if (machine.getPodName().equals(podName)) {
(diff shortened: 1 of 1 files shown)


davidfestal (approved):

benoitf (approved):
IMHO it is very nice to grab other events

benoitf:
ci-test

amisevsk (approved):
Good to have, especially with recent PVC issues.

garagatyi (approved):

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
```diff
@@ -772,19 +772,17 @@ private boolean isUnrecoverable(PodEvent event) {
...
     @Override
     public void handle(PodEvent event) {
       final String podName = event.getPodName();
-      final String containerName = event.getContainerName();
       try {
         for (Entry<String, KubernetesMachineImpl> entry :
             machines.getMachines(getContext().getIdentity()).entrySet()) {
           final KubernetesMachineImpl machine = entry.getValue();
-          if (machine.getPodName().equals(podName)
-              && machine.getContainerName().equals(containerName)) {
```
sleshchenko, line 787: @ibuziuk Looks like in the case when there are two containers in one pod, containers related events will be published in both machines, something like the following example in case OpenShift Sql stack:
```
pulling image `eclipse-che/ubuntu`
pulling image `eclipse-che/ubuntu`
Successfully pulled image `eclipse-che/mysql`
Successfully pulled image `eclipse-che/mysql`
Created container
Created container
Started container
Started container
```
Is it true or I missed something? If it's true let's filter events by container name if it is not empty.

sleshchenko: Another fix that I think is needed for multi-container workspace is removing `return` statement [here](https://github.com/eclipse/che/pull/10228/files#diff-5cac1af7ada92b7a5309db88b7f49d81R791) since two or more machines may be related to the same pod and it's OK to duplicate logs in such cases, otherwise, the only first machine will have logs. WDYT?

sleshchenko (commented):
Please take my inlined comments

Issue #10062, opened 2018-06-15
Factory in Theia: optimizing che-theia-factory loading


sunix:
### Description
Part of import projects which is a subtask of #9074 

With https://github.com/eclipse/che-theia-factory-extension/pull/1 it is possible to have Theia cloning projects that are defined in the factory definition when coming from a factory URL. However it is taking time compiling Theia each time a factory is launched. The goal of this issue is to include che-theia-factory-extension to the default che-theia docker image and avoid this compilation step and go faster.

PR #10233, opened 2018-06-29
Optimizing che-theia-factory loading sequence by adding it to the def…

sunix:
…ault che-theia docker image (#10062)

Signed-off-by: Sun Seng David TAN <sutan@redhat.com>

### What does this PR do?
Adding che-theia-factory extension to the default che-theia docker image to avoid recompiling at startup.

#10062

#### Release Notes

Optimizing che-theia-factory loading sequence by adding it to the default che-theia docker image

### DIFF
--- a/dockerfiles/theia/src/package.json
+++ b/dockerfiles/theia/src/package.json
@@ -11,7 +11,7 @@
         "@theia/markers": "latest",
         "@theia/extension-manager": "latest",
         "@theia/messages": "latest",
-        "@eclipse-che/theia-factory-extension": "latest"
+        "@eclipse-che/theia-factory-extension": "0.0.1-1530189818"
     },
     "devDependencies": {
         "@theia/cli": "latest"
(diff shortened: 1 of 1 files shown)


benoitf:
@sunix shouldn't it in https://github.com/sunix/che/blob/0f5cf18f5edae03ef8d5eec67e85d45c8e7ff746/dockerfiles/theia/Dockerfile#L36-L38
@mmorhun ?

sunix:
@benoitf @mmorhun to me this addExtension.js part is a hack to add the extension from source (not from a npm repo)

benoitf:
@sunix IMHO we should use a fixed version as well (latest may be for another theia version)

sunix:
@benoitf you mean using theia 0.3.12 ?

benoitf:
@sunix I was referencing
```
@eclipse-che/theia-factory-extension": "latest"
```

as we should include only version of this extension compliant with the theia version that will be used

sunix:
yes sure but we want to use 0.3.12 don't we ? or wait for @mmorhun  ?

sunix:
@benoitf done

benoitf (approved):
we could change behavior after @mmorhun will come back

PR #10237, opened 2018-07-02
Make it possible for Go agents to use auth tokens from HTTP header

mshaposhnik:
### What does this PR do?
For now Go agents is able to extract aithentication tokens only from request query parameters, 
which is inconvenient for proxy authetication and has some other disadvantages.
This PR makes possible to pass tokens via `Authorization` header and refactors readiness/liveness probes to use them.

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/agents/go-agents/core/auth/auth.go
+++ b/agents/go-agents/core/auth/auth.go
@@ -18,6 +18,7 @@ import (
 	"regexp"
 
 	"fmt"
+	"strings"
 	"github.com/dgrijalva/jwt-go"
 	"os"
 )
@@ -111,6 +112,12 @@ func (handler handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 	token := req.URL.Query().Get("token")
+	if token == "" {
+	  header := req.Header.Get("Authorization")
+	  if header != "" && strings.HasPrefix(strings.ToUpper(header), strings.ToUpper("bearer")) {
+	   token = header[7:]
+	  }
+	}
 	if err := authenticate(token); err == nil {
 		handler.delegate.ServeHTTP(w, req)
 	} else {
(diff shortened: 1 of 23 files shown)


agents/go-agents/core/auth/auth.go
```diff
@@ -111,6 +112,12 @@ func (handler handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 	token := req.URL.Query().Get("token")
+	if token == "" {
+	  header := req.Header.Get("Authorization")
+	  if header != "" && strings.HasPrefix(strings.ToLower(header), "bearer") {
```
sleshchenko: it's impossible in our authentication implementation but still, you check `bearer` word and cut `bearer ` what if there will be a token that starts with `bearer` that is not type but just part of token like `bearer9SUcJXjsdenx`. Maybe it would be better to check if token is prefixed with `bearer `

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/hc/probe/server/ExecServerLivenessProbeConfigFactory.java
```diff
@@ -23,10 +29,15 @@
  * @author Alexander Garagatyi
  */
 public class ExecServerLivenessProbeConfigFactory implements HttpProbeConfigFactory {
+
   private final int successThreshold;
+  private final MachineTokenProvider machineTokenProvider;
 
-  public ExecServerLivenessProbeConfigFactory(int successThreshold) {
+  @Inject
+  public ExecServerLivenessProbeConfigFactory(
```
sleshchenko: I think we are able not to change current approach with public available liveness probes.
@skabashnyuk WDYT? Should we revert it here and create an issue for implementing exclude in JWT proxy?

agents/go-agents/core/auth/auth.go
mshaposhnik: `Bearer` word indicates type of token, there also can be `Basic` or `Digest` which is not supported. So we cut it away to have pure token. Or do i understood your question wrong ?

sleshchenko: Markdown helps me to confuse you =)
Let me rephrase, you check bearer word without space and cut bearer with space (7 characters) Maybe it would be better to check if token is prefixed with **bearer ** with space.

mshaposhnik: Ok

garagatyi (approved):
Looks good!

skabashnyuk (approved):

sleshchenko (approved):

Issue #10243, opened 2018-07-02
API requests from ws-agent to ws-master have the same authorization header


vinokurig:
### Description
 

If a workspace is shared with another user, all API requests from ws-agent to ws-master will be signed with the same machine token: https://github.com/eclipse/che/blob/5f4d4e2034b737df11a2c86d62f68677983ce825/wsagent/che-wsagent-core/src/main/java/org/eclipse/che/wsagent/server/AgentHttpJsonRequestFactory.java#L37
As `machineToken` is the same for all users in the workspace, all users authorize http-requests with the same token. So guest user makes requests as a workspace owner.

### Reproduction Steps

1. User A created an organization and added a workspace and shared it with User B
2. User A generated a key, uploaded it to GitHub and added git committer name at Profile > Preferences > Git Committer
3. User B doesn't have any SSH keys, goes to Git menu -> Remotes -> Push

Expected: User B should get an error because he doesn't have any Ssh keys.
Actual: User B receives an `up-to-date`  message that means that he made the request to Github with Ssh keys from User A
**OS and version:**    

**Diagnostics:**

PR #10249, opened 2018-07-03
Fix machine auth by using own user token instead of workspace owner token.

mshaposhnik:
### What does this PR do?
Fixes wrong behavior when in shared workspace all users authorize http-requests with the same token created on workspace startup. Now own user token will be used.

https://github.com/eclipse/che/issues/10243

#### Release Notes
N/A

#### Docs PR
N/A


mshaposhnik:
ci-test

sleshchenko (approved):

skabashnyuk (approved):

Issue #10081, opened 2018-06-18
Cover ws-agent/exec-agent/terminal with JWT proxy on K8s/OpenShift infrastuctures


sleshchenko:
### Description
Cover ws-agent/exec-agent/terminal with JWT proxy on K8s/OpenShift infrastuctures.

It includes the following operations:
- add `jwtproxy` container for one workspace;
- configure `jwtproxy` in the added container, the corresponding public key and config file should be injected into the container. The injection can be implemented with using secrets.
- rewrite public accessible servers to `jwtproxy` ones instead of original;

For implementing the same in Docker infrastructure there is an issue https://github.com/eclipse/che/issues/10242.
**OS and version:**    

**Diagnostics:**

sleshchenko:
Instead of hardcoding the described servers I'm going to introduce an abstract way for covering workspace servers with authentication. It may be done with attributes in servers configs, as example workspace agent installer
```json
{
  "id": "org.eclipse.che.ws-agent",
  "version": "1.0.3",
  "name": "Workspace API",
  "description": "Workspace API support",
  "dependencies": [],
  "properties": {},
  "servers": {
    "wsagent/http": {
      "port": "4401/tcp",
      "protocol": "http",
      "path" : "/api",
      "attributes": {
        "secure": "true"
      }
    },
    "wsagent/ws": {
      "port": "4401/tcp",
      "protocol": "ws",
      "path" : "/wsagent",
      "attributes": {
        "secure": "true"
      }
    },
    "wsagent-debug": {
      "port": "4403/tcp",
      "protocol": "tcp"
    }
  }
}
```
I'm open to suggestions attribute name or another way for abstracted covering servers with authentication.
@gazarenkov @garagatyi  @skabashnyuk WDYT?

gazarenkov:
According to "add jwtproxy container for one workspace;" - should not such a thing be on workspace level configuration instead?

sleshchenko:
@gazarenkov Do you mean that all servers should be covered with authentication or not according to the configuration on a workspace level? I think it may be redundant to cover user's applications servers, so we need somehow to add an ability for user/installer developer (later feature developer) to mark servers which should be covered with authentication and which shouldn't.
Please elaborate more what do you mean if you mean something else.

gazarenkov:
Got it, thanks.
Pls, be maximum clear documenting the policies (like how infrastructure behaves for each particular configuration), especially "default" case.

sleshchenko:
[...]

PR #10266, opened 2018-07-04
Refactored external servers exposition in K8s/OS infrastructures

sleshchenko:
### What does this PR do?
Refactors external servers exposition in K8s/OS infrastructures:
- moves external servers related classes to a separated package;
- fixes java doc in OpenShiftExternalServerExposer;
- makes ServerServiceBuilder non-internal class to make it possible to reuse it;
- refactors external server exposer strategy to expose service port one by one;
Previously expose server strategy had to match servers with services port, but sometimes it is impossible like in case of JWT proxy. Because in this case, server config contains port that is different to service port. Like `wsagent` server is exposed on 4401 original service port, but it will be publicly accessible via 4911 (random one) JWTProxy service port.

- fixes and improve tests

It is changes needed for https://github.com/eclipse/che/issues/10081.
Separated changes from https://github.com/eclipse/che/pull/10252 to make review easier.

#### Release Notes
N/A

#### Docs PR
N/A


garagatyi (approved):
Looks good! 
BTW each significant change we do in this area I realize that had to do the code simpler and cleaner in the first place.
But thank you for moving the code in the direction of simplicity!

sleshchenko:
ci-test

musienko-maxim:
QE team did not find new regressions after selenium session.

musienko-maxim (approved):

PR #10268, opened 2018-07-04
Different Quickfix Behaviour Java vs. LSP #10112

### LINKED ISSUE #10112: Different Quickfix Behaviour Java vs. LSP
### Description
The difference is that in CHE-master we don't need to set the cursor into the position with a problem it is enough to set it into the line with a problem.
  
**Master**
qfmaster

**LSP**
qf_lsp

vrubezhny:
The fix makes LS Java Editor's quick fixes to follow CHE Java editor behavior

Fixes: #10112

### What does this PR do?

The fix makes LS Java Editor's quick fixes to follow CHE Java editor behavior

Fixes: #10112

#### Release Notes

che10112

With this fix LSP Editor's quick fixes follow the CHE Java editor one's behavior

### DIFF
--- a/plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/quickassist/LanguageServerQuickAssistProcessor.java
+++ b/plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/quickassist/LanguageServerQuickAssistProcessor.java
@@ -12,7 +12,7 @@ package org.eclipse.che.plugin.languageserver.ide.editor.quickassist;
 
 import com.google.gwt.user.client.rpc.AsyncCallback;
 import com.google.gwt.user.client.ui.Widget;
-import java.util.ArrayList;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -33,6 +33,7 @@ import org.eclipse.che.ide.api.editor.quickfix.QuickAssistProcessor;
 import org.eclipse.che.ide.api.editor.text.LinearRange;
 import org.eclipse.che.ide.api.editor.text.TextPosition;
 import org.eclipse.che.ide.api.editor.text.annotation.Annotation;
+import org.eclipse.che.ide.api.editor.texteditor.TextEditor;
 import org.eclipse.che.ide.api.icon.Icon;
 import org.eclipse.che.plugin.languageserver.ide.editor.DiagnosticAnnotation;
 import org.eclipse.che.plugin.languageserver.ide.service.TextDocumentServiceClient;
@@ -66,6 +67,8 @@ public class LanguageServerQuickAssistProcessor implements QuickAssistProcessor 
       QuickAssistInvocationContext invocationContext, CodeAssistCallback callback) {
     LinearRange range = invocationContext.getTextEditor().getSelectedLinearRange();
     Document document = invocationContext.getTextEditor().getDocument();
+    TextEditor textEditor = invocationContext.getTextEditor();
+    boolean goToClosest = (range.getLength() == 0);
     QueryAnnotationsEvent.QueryCallback annotationCallback =
         new QueryAnnotationsEvent.QueryCallback() {
 
@@ -75,26 +78,25 @@ public class LanguageServerQuickAssistProcessor implements QuickAssistProcessor 
               Map<Annotation, org.eclipse.che.ide.api.editor.text.Position> annotations) {
             // iteration with range never returns anything; need to filter ourselves.
             // https://github.com/eclipse/che/issues/4338
-            List<Diagnostic> diagnostics =
-                annotations
-                    .entrySet()
-                    .stream()
-                    .filter(
-                        (e) -> e.getValue().overlapsWith(range.getStartOffset(), range.getLength()))
-                    .map(Entry::getKey)
-                    .map(a -> (DiagnosticAnnotation) a)
-                    .map(DiagnosticAnnotation::getDiagnostic)
-                    .collect(Collectors.toList());
+            List<Diagnostic> diagnostics = new LinkedList<>();
+            int offset =
+                collectQuickFixableAnnotations(
+                    range, document, annotations, goToClosest, diagnostics);
+            if (offset != range.getStartOffset()) {
+              TextEditor presenter = ((TextEditor) textEditor);
+              presenter.getCursorModel().setCursorPosition(offset);
+            }
 
             CodeActionContext context = new CodeActionContext(diagnostics);
 
-            TextPosition start = document.getPositionFromIndex(range.getStartOffset());
+            TextPosition start =
+                document.getPositionFromIndex(goToClosest ? offset : range.getStartOffset());
             TextPosition end =
-                document.getPositionFromIndex(range.getStartOffset() + range.getLength());
+                document.getPositionFromIndex(
+                    goToClosest ? offset : range.getStartOffset() + range.getLength());
             Position rangeStart = new Position(start.getLine(), start.getCharacter());
             Position rangeEnd = new Position(end.getLine(), end.getCharacter());
             Range rangeParam = new Range(rangeStart, rangeEnd);
-            rangeParam.setEnd(rangeEnd);
 
             TextDocumentIdentifier textDocumentIdentifier =
                 new TextDocumentIdentifier(document.getFile().getLocation().toString());
@@ -103,7 +105,7 @@ public class LanguageServerQuickAssistProcessor implements QuickAssistProcessor 
 
             Promise<List<Command>> codeAction =
                 textDocumentService.codeAction(new CodeActionParamsDto(params));
-            List<CompletionProposal> proposals = new ArrayList<>();
+            List<CompletionProposal> proposals = new LinkedList<>();
             codeAction.then(
                 (commands) -> {
                   for (Command command : commands) {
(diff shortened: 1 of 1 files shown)


plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/quickassist/LanguageServerQuickAssistProcessor.java
```diff
@@ -75,26 +78,26 @@ public void respond(
...
+                document.getPositionFromIndex(goToClosest ? offset : range.getStartOffset());
             TextPosition end =
-                document.getPositionFromIndex(range.getStartOffset() + range.getLength());
+                document.getPositionFromIndex(
+                    goToClosest ? offset : range.getStartOffset() + range.getLength());
             Position rangeStart = new Position(start.getLine(), start.getCharacter());
             Position rangeEnd = new Position(end.getLine(), end.getCharacter());
             Range rangeParam = new Range(rangeStart, rangeEnd);
-            rangeParam.setEnd(rangeEnd);
+            //   rangeParam.setEnd(rangeEnd);
```
dkulieshov: We should avoid leaving such comments.

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/quickassist/LanguageServerQuickAssistProcessor.java
```diff
@@ -125,6 +128,124 @@ public void respond(
...
+      List<Diagnostic> resultingDiagnostics) {
+    int invocationLocation = range.getStartOffset();
+    if (goToClosest) {
+      LinearRange line =
+          document.getLinearRangeForLine(
+              document.getPositionFromIndex(range.getStartOffset()).getLine());
+      int rangeStart = line.getStartOffset();
+      int rangeEnd = rangeStart + line.getLength();
+
+      ArrayList<org.eclipse.che.ide.api.editor.text.Position> allPositions = new ArrayList<>();
```
dkulieshov: Probably should be just `java.util.List`?

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/quickassist/LanguageServerQuickAssistProcessor.java
```diff
@@ -125,6 +128,124 @@ public void respond(
...
+      List<DiagnosticAnnotation> allAnnotations = new ArrayList<>();
+      int bestOffset = Integer.MAX_VALUE;
+      for (Annotation a : annotations.keySet()) {
+        org.eclipse.che.ide.api.editor.text.Position pos = annotations.get(a);
+        if (pos != null && isInside(pos.offset, rangeStart, rangeEnd)) { // inside our range?
```
dkulieshov: Instead of leaving a comment we can probably simply call the method `isInsideOurRange`. This will save us from unnecessary comment and at the same time will make the code declarative. What do you think?

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/quickassist/LanguageServerQuickAssistProcessor.java
```diff
@@ -125,6 +128,124 @@ public void respond(
...
+  /**
+   * Tells is the offset is inside the (inclusive) range defined by start-end.
+   *
+   * @param offset the offset
+   * @param start the start of the range
+   * @param end the end of the range
+   * @return true iff offset is inside
+   */
+  private boolean isInside(int offset, int start, int end) {
+    return offset == start
```
dkulieshov: I might be mistaken but why can't we have just 
```java
return start <= offset && offset <= end;
```
?

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/quickassist/LanguageServerQuickAssistProcessor.java
```diff
@@ -125,6 +128,124 @@ public void respond(
...
+  }
+
+  private int processAnnotation(
+      Annotation annot,
+      org.eclipse.che.ide.api.editor.text.Position pos,
+      int invocationLocation,
+      int bestOffset) {
+    final int posBegin = pos.offset;
+    final int posEnd = posBegin + pos.length;
+    if (isInside(invocationLocation, posBegin, posEnd)) { // covers invocation location?
```
dkulieshov, line 201: Now I see, you're using `isInside` method all over the pull request and give an explanation in comments to make it clearer for the reader. That is very nice of you however I would still recommend to avoid using comments in such cases and make the code itself more declarative instead.

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/quickassist/LanguageServerQuickAssistProcessor.java
```diff
@@ -125,6 +128,124 @@ public void respond(
...
+      return invocationLocation;
+    } else if (bestOffset != invocationLocation) {
+      final int newClosestPosition = computeBestOffset(posBegin, invocationLocation, bestOffset);
+      if (newClosestPosition != -1) {
+        if (newClosestPosition != bestOffset) { // new best
+          // Can't use JavaAnnotationUtil.hasCorrections() to see if there are
```
dkulieshov, line 207: However I admit that in some cases there is indeed no better way to explain the idea than to use a proper comment :-D

dkulieshov (approved):
Without taking into account minor remarks is an okay pull request.

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/quickassist/LanguageServerQuickAssistProcessor.java
```diff
@@ -75,26 +78,26 @@ public void respond(
...
-                annotations
-                    .entrySet()
-                    .stream()
-                    .filter(
-                        (e) -> e.getValue().overlapsWith(range.getStartOffset(), range.getLength()))
-                    .map(Entry::getKey)
-                    .map(a -> (DiagnosticAnnotation) a)
-                    .map(DiagnosticAnnotation::getDiagnostic)
-                    .collect(Collectors.toList());
+            List<Diagnostic> diagnostics = new ArrayList<>();
```
dkulieshov: Probably it would be better to use `java.util.LinkedList` here and further as we don't know the exact size of a list, however I don't think this is so critical.

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/quickassist/LanguageServerQuickAssistProcessor.java
vrubezhny: isInside() method is an almost 1-to-1 copy from https://github.com/eclipse/che/blob/master/plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/editor/JavaQuickAssistProcessor.java#L313, so I wouldn't change its name here in order to 'not loose' the root from where this code came from.
As well as collectQuickFixableAnnotations() method, which is changed a lot due to the specifics of LSP DiagnosticAnnotation-s, but still a kind of copy of its namesake method from https://github.com/eclipse/che/blob/master/plugins/plugin-java/che-plugin-java-ext-lang-client/src/main/java/org/eclipse/che/ide/ext/java/client/editor/JavaQuickAssistProcessor.java#L192 as well as most of comments used in its body.

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/quickassist/LanguageServerQuickAssistProcessor.java
vrubezhny: I have already commented on the 'source' of the method, but I believe you're right and your replacement would work and looks better

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/quickassist/LanguageServerQuickAssistProcessor.java
vrubezhny: Again. Already commented on the source of these comments.

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/quickassist/LanguageServerQuickAssistProcessor.java
vrubezhny: In this case it is my own comment on that I wasn't able to replace JDT realization with something more or less similar in the LSP one. And as such this potentially can be a reason of why LSP realization works differently in some cases (and i believe there will be such cases)

plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/quickassist/LanguageServerQuickAssistProcessor.java
vrubezhny: I don't see it as something that can give a big improvement as in reality this list doesn't consist like hundreds of elements - in my case it has one element while there are four error annotations in the same line. We just need to store "more than one" diagnostics here as some error annotations might overlap... Not more, so it's not about the performance, imho.

vrubezhny:
@dkuleshov I've updated PR according to your notes.
I've rebased it and, as such, pushed with force, so no fix-ups. Sorry about that.

dkuleshov:
@vrubezhny excellent work thank you, we can merge this after we get it approved by the QE (@musienko-maxim).

musienko-maxim:
We did not find regression in the E2E tests and during manual checking of LS features which is not covered by selenium tests

musienko-maxim (approved):

Issue #10361, opened 2018-07-10
Handling Wrong Events on Project Creation


tsmaeder:
We're still handling "ProjectInitializedEvent" instead of Project created

PR #10363, opened 2018-07-10
Send and handle ProjectCreatedEvent

tsmaeder:
### What does this PR do?
Send project created event on zip import and handles these events instead of "ProjectInitializedEvent".

https://github.com/eclipse/che/issues/10361

### DIFF
--- a/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/impl/ProjectServiceApi.java
+++ b/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/impl/ProjectServiceApi.java
@@ -534,14 +534,17 @@ public class ProjectServiceApi {
 
     fsManager.unzip(wsPath, zip, skipFirstLevel);
 
-    return Response.created(
-            serviceContext
-                .getBaseUriBuilder()
-                .clone()
-                .path(ProjectService.class)
-                .path(ProjectService.class, "getChildren")
-                .build(new String[] {wsPath.substring(1)}, false))
-        .build();
+    Response response =
+        Response.created(
+                serviceContext
+                    .getBaseUriBuilder()
+                    .clone()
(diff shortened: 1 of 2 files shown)


riuvshin:

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/impl/ProjectServiceApi.java
```diff
@@ -534,14 +534,17 @@ public Response importZip(String wsPath, InputStream zip, Boolean skipFirstLevel
...
+    Response response =
+        Response.created(
+                serviceContext
+                    .getBaseUriBuilder()
+                    .clone()
+                    .path(ProjectService.class)
+                    .path(ProjectService.class, "getChildren")
+                    .build(new String[] {wsPath.substring(1)}, false))
+            .build();
+    eventService.publish(new ProjectCreatedEvent(wsPath));
```
svor, line 546: i am not sure if it is a good place to publish `ProjectCreatedEvent`  @vparfonov WDYT ?

svor (approved):

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/impl/ProjectServiceApi.java
vparfonov: I am ok for firing event from this class we already do it in several places, but i think we need to fire it only if we create project in ROOT

tsmaeder: @vparfonov I don't really see the logic of that, but anyways...we need to have this work to continue fixing selenium tests. We can clear up the difference once we have a fix for https://github.com/eclipse/che/issues/9924

PR #10366, opened 2018-07-10
set default k8s namespace to che

riuvshin:
set default `CHE_INFRA_KUBERNETES_NAMESPACE` to `che`


ghost (approved):

deploy/kubernetes/helm/che/values.yaml
```diff
@@ -37,3 +37,4 @@ global:
   gitHubClientID: ""
   gitHubClientSecret: ""
   pvcClaim: "1Gi"
+  cheInfraKubernetesNamespace: ""
```
sleshchenko, line 40: Maybe it would be better to name it something like `cheWorkspacesNamespace` from helm chart perspective

eivantsov:
@sleshchenko maybe. The name is more self explaining

sleshchenko:
@riuvshin Could you prepare a PR if it makes sense for you?

qianlei90:
What's the difference between [line 34](https://github.com/eclipse/che/blob/master/deploy/kubernetes/helm/che/templates/configmap.yaml#L34) and [line 53](https://github.com/eclipse/che/blob/master/deploy/kubernetes/helm/che/templates/configmap.yaml#L53)? We already have `cheNamespace` in values.

sleshchenko:
@qianlei90 Thanks for pointing.
`Values.global.cheNamespace` is a namespace where Che objects should be created
`Values.global.cheInfraKubernetesNamespace` is a namespace where workspace related objects should be created
But definitly `CHE_INFRA_KUBERNETES_NAMESPACE` env var should be set once. @riuvshin Please take a look

PR #10385, opened 2018-07-11
Validate infrastructure related configuration only on start workspace

gazarenkov:
### Motivation: 

Currently, we validate recipe on both workspace create/update and start.
It is not convenient policy since in this case we consider workspace/stack storage to have only workspaces with ONLY the same supported type(s) of infrastructure as supported by runtime.
It is not convenient for portable workspaces since:

- it does not allow to have several environments of different recipe types inside a workspace 
- we can consider keeping shared registry among different infrastructures
- we can consider using environment of one type to be imported and used as a template for other, "supported" type on demand

### Side effect:

Workspace storage may contain incorrect recipes but that's a reasonable drawback since it will be validated before start

### What does this PR do?

Logic of recipe validating moved from "generic" validation part to start Workspace

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
@@ -41,6 +41,7 @@ import org.eclipse.che.api.core.model.workspace.WorkspaceStatus;
 import org.eclipse.che.api.core.notification.EventService;
 import org.eclipse.che.api.workspace.server.model.impl.WorkspaceConfigImpl;
 import org.eclipse.che.api.workspace.server.model.impl.WorkspaceImpl;
+import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
 import org.eclipse.che.api.workspace.server.spi.WorkspaceDao;
 import org.eclipse.che.api.workspace.shared.event.WorkspaceCreatedEvent;
 import org.eclipse.che.commons.annotation.Nullable;
@@ -343,7 +344,7 @@ public class WorkspaceManager {
             });
   }
 
-  public Set<String> getSupportedRecipes() {
+  Set<String> getSupportedRecipes() {
     return runtimes.getSupportedRecipes();
   }
 
@@ -357,9 +358,16 @@ public class WorkspaceManager {
               workspace.getNamespace(), workspace.getConfig().getName(), envName));
     }
     workspace.getAttributes().put(UPDATED_ATTRIBUTE_NAME, Long.toString(currentTimeMillis()));
-    workspaceDao.update(workspace);
     final String env = firstNonNull(envName, workspace.getConfig().getDefaultEnv());
 
+    // validate environment in advance
+    try {
+      runtimes.validate(workspace.getConfig().getEnvironments().get(env));
+    } catch (InfrastructureException | ValidationException e) {
+      throw new ServerException(e);
+    }
+
+    workspaceDao.update(workspace);
     runtimes
         .startAsync(workspace, env, firstNonNull(options, Collections.emptyMap()))
         .thenAccept(aVoid -> handleStartupSuccess(workspace))
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
@@ -344,7 +344,10 @@ public class WorkspaceManager {
             });
   }
 
-  Set<String> getSupportedRecipes() {
+  /**
+   * @return set of supported recipe types
+   */
+  public Set<String> getSupportedRecipes() {
     return runtimes.getSupportedRecipes();
   }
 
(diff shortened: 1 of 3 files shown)


wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -343,7 +344,7 @@ public void stopWorkspace(String workspaceId, Map<String, String> options)
             });
   }
 
-  public Set<String> getSupportedRecipes() {
+  Set<String> getSupportedRecipes() {
```
sleshchenko: Could you explain why this method should be package private but not public?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceValidator.java
```diff
@@ -94,12 +85,6 @@ public void validateConfig(WorkspaceConfig config)
           environment.getMachines().entrySet()) {
         validateMachine(machineEntry.getKey(), machineEntry.getValue());
       }
-
-      try {
-        runtimes.validate(environment);
```
sleshchenko, line 99: Java doc should be updated along with these changes https://github.com/eclipse/che/pull/10385/files#diff-5f46ddd221720706a3210208652a0f1eR57

sleshchenko (commented):
> it does not allow to have several environments of different recipe types inside a workspace

Let me clarify a bit. Now it is possible to have several environments of different recipe types inside a workspace. BUT it is not possible to have environments which have recipes that are currently not supported. At the same time, at least one supported environment is enough to be able to start a workspace.

> we can consider using environment of one type to be imported and used as a template for other, "supported" type on demand

Could you elaborate more on what do you mean using an environment as `template` for other?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
gazarenkov: Well, there are no other reasons than IntelliJ suggestion and absence of javadoc (I thought it is not important enough to be public )) ). If you think it is useful, I do not mind to revert and add javadocs as well.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceValidator.java
gazarenkov: done

gazarenkov:
> we can consider using environment of one type to be imported and used as a template for other, "supported" type on demand
> 
> Could you elaborate more on what do you mean using an environment as template for other?

Sure, nothing special in fact:
1/ You have workspace with recipe of type1 in one instance (supporting type1)
2/ You want to make it work in other instance which supports recipe of type2
3/ You export this workspace and import to other instance with no supportability problem (due to this update)
4/ You modify it as you want to make it support type2 and test

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceValidator.java
```diff
@@ -49,26 +46,16 @@
   private static final Pattern VOLUME_NAME = Pattern.compile("[a-z][a-z0-9]{1,18}");
   private static final Pattern VOLUME_PATH = Pattern.compile("/.+");
 
-  private final WorkspaceRuntimes runtimes;
```
sleshchenko, line 52: It does not affect anything but please remove `WorkspaceRuntimes` from `WorkspaceValidator` too
https://github.com/eclipse/che/blob/2fb979bbe1cf2fc5f4179c2440bfd4de9be73749/wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/workspace/server/WorkspaceValidatorTest.java#L46

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -343,6 +344,7 @@ public void stopWorkspace(String workspaceId, Map<String, String> options)
             });
   }
 
+  /** @return set of supported recipe types */
```
sleshchenko: According to [Google Java Code Style](https://google.github.io/styleguide/javaguide.html#s7.2-summary-fragment) it makes sense to reformat this java doc as 
```
/** Returns set of supported recipe types */
```
Up to you change it or not

sleshchenko (approved):
LGTM

skabashnyuk (approved):

riuvshin:
ci-test build report:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/471/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/471//Selenium_tests_report)
[selenium tests report data](http://tests.codenvycorp.com/selenium_test_report_data/10385)
docker image: eclipseche/che-server:10385

Ohrimenko1988:
In the functionality which is covered by selenium tests - new regression was not found.

PR #10397, opened 2018-07-12
Check when ws-edit changes were applied

svor:
The fix for some unstable tests from `org.eclipse.che.selenium.refactor.move`.
Sometimes tests check the state of the project tree before applying refactoring changes. Firstly we need to wait  when all WorkspaceEdit changes were applied.

It is needed for https://github.com/eclipse/che/issues/10399

### DIFF
--- a/selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/refactor/move/MoveItemsTest.java
+++ b/selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/refactor/move/MoveItemsTest.java
@@ -10,6 +10,8 @@
  */
 package org.eclipse.che.selenium.refactor.move;
 
+import static org.eclipse.che.selenium.core.constant.TestTimeoutsConstants.LOAD_PAGE_TIMEOUT_SEC;
+
 import com.google.inject.Inject;
 import java.net.URL;
 import java.nio.charset.Charset;
@@ -23,6 +25,7 @@ import org.eclipse.che.selenium.core.project.ProjectTemplates;
 import org.eclipse.che.selenium.core.workspace.TestWorkspace;
 import org.eclipse.che.selenium.pageobject.CodenvyEditor;
 import org.eclipse.che.selenium.pageobject.Consoles;
+import org.eclipse.che.selenium.pageobject.Events;
 import org.eclipse.che.selenium.pageobject.Ide;
 import org.eclipse.che.selenium.pageobject.Loader;
 import org.eclipse.che.selenium.pageobject.Menu;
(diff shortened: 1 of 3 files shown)


selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/refactor/move/MoveItemsTest.java
```diff
@@ -106,7 +118,10 @@ public void checkMoveItem0() throws Exception {
     refactor.clickOnExpandIconTree("/src/main/java");
     refactor.chooseDestinationForItem("p1");
     refactor.clickOkButtonRefactorForm();
+    loader.waitOnClosed();
```
tsmaeder, line 121: You put this these in various places. Do you understand when they are needed?

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/refactor/move/MoveItemsTest.java
```diff
@@ -106,7 +118,10 @@ public void checkMoveItem0() throws Exception {
...
     refactor.waitMoveItemFormIsClosed();
+    loader.waitOnClosed();
+    events.waitExpectedMessage(APPLY_WORKSPACE_CHANGES, LOAD_PAGE_TIMEOUT_SEC);
```
tsmaeder: Is load page timeout the right number here? What would be a reasonable timeout for those refactorings?

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/refactor/move/MoveJavaClassToSubpackageTest.java
```diff
@@ -10,6 +10,8 @@
  */
 package org.eclipse.che.selenium.refactor.move;
 
+import static org.eclipse.che.selenium.core.constant.TestTimeoutsConstants.LOAD_PAGE_TIMEOUT_SEC;
```
tsmaeder: Same as above, is this the right timeout?

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/refactor/move/MoveItemsTest.java
svor: it is needed in places where UI loader can be shown

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/refactor/move/MoveItemsTest.java
svor: I've changed to `DEFAULT_TIMEOUT`, the value is the same - 10 sec. It should be enough to apply changes.

tsmaeder (approved):

Issue #10424, opened 2018-07-16
Check set of key-bindings


svor:
Need to check set of existing key bindings in test resources. See `che/selenium/che-selenium-test/src/test/resources/org/eclipse/che/selenium/assistant/key-bindings.txt`

Related test: `org/eclipse/che/selenium/assistant/KeyBindingsTest.java`

PR #10426, opened 2018-07-16
Update set of existing key-bindings

svor:
### What does this PR do?
Fixes a test which checks set of key bindings.
Instead of **Open editor for selected item**  we have **Find Definition**
Instead of **Shows all parameters which constructor or method can accept** we have **Signature Help**

https://github.com/eclipse/che/issues/10424


tolusha (approved):

tsmaeder (approved):
lgtm

Issue #10417, opened 2018-07-13
Announce Capability for Javadoc Markup in Client Capabilities


tsmaeder:
Since jdt.ls updated to lsp4j 0.4.0, it expects that we announce the capability to receive marked up javadoc for completion items.
Need to update to 0.4.0 and send the capability.

PR #10436, opened 2018-07-16
Consume markup and snippets in completion doc

tsmaeder:
### What does this PR do?
Updates code using LSP4J to use 0.4.0 and announces the capability to consume snippets and markdown in completions.

#10417


wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/TextDocumentService.java
```diff
@@ -440,12 +440,28 @@ public boolean canDo(ExtendedLanguageServer element) {
           public boolean handleResult(ExtendedLanguageServer element, Hover hover) {
             if (hover != null) {
               HoverDto hoverDto = new HoverDto(hover);
-              result.getContents().addAll(hoverDto.getContents());
+              Either<List<Either<String, MarkedString>>, MarkupContent> contents =
+                  hoverDto.getContents();
+              if (contents.isLeft()) {
+
```
svor: do we need to make some rendering of the content here?

svor: it looks like need to add some rendering of the content here

tsmaeder: Yes we do...fixing the test does not make the code correct.

svor (approved):

Issue #10429, opened 2018-07-16
Update theia version to 0.3.12 in Che


ashumilova:
As far as Theia has released [0.3.12](https://github.com/theia-ide/theia/releases/tag/v0.3.12) and it contains a new plugin model - there is need to update Theia to the latest version in eclipse/che-theia image, in [hosted plugin manager](https://github.com/eclipse/che-theia-hosted-plugin-manager-extension).

mmorhun:
Won't fix

PR #10441, opened 2018-07-17
Use Theia 0.3.12 core

mmorhun:
### What does this PR do?
Use Theia 0.3.12 core version instead of next

https://github.com/eclipse/che/issues/10429

#### Release Notes
Switch to Theia 0.3.12 core

### DIFF
--- a/dockerfiles/theia/Dockerfile
+++ b/dockerfiles/theia/Dockerfile
@@ -10,7 +10,7 @@
 FROM node:8-alpine
 
 ARG GITHUB_TOKEN
-ARG THEIA_VERSION=0.4.0-next.b17727c1
+ARG THEIA_VERSION=0.3.12
 
 EXPOSE 3000 3030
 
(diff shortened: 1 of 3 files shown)


AndrienkoAleksandr (approved):

benoitf:
@mmorhun how it will work with https://github.com/theia-ide/theia/commit/e358acd5eb55bab331fed8cb18ea935b4c9a2f8e#diff-bba9137b90f183d0b3d5325d2547406a fix ?

sunix (approved):
LGTM

mmorhun:
@benoitf it won't. We need next release of the artifact. Or hashes again...

benoitf:
@mmorhun so we better should have a working image with plug-ins for Che/Theia

benoitf:
@mmorhun can we use hashes and use this Theia version for building each extension when building the docker image ? (independently of what has been pickup by each extension) ?

mmorhun:
@fbenoit yes, but:
- we have to build each extension from source (clone it on build) and do not use npmjs
- we have to alter each extension source with an additional script
- it is fragile

Or we c an try to use another approach with resolution. @AndrienkoAleksandr has PR with such changes but it is for Theia core. But I am not sure that it will work with children dependencies. Need to try it..

mmorhun:
@benoitf @AndrienkoAleksandr and I just tried way with resolutions and it seems it works. It has redundant dependencies, but still works as expected.

AndrienkoAleksandr:
Yep. We need some how live with current Theia release cycle and make our Theia images stable. I would prefer way with resolutions, because in this case we should not edit package.json files for all extensions and it should work with extensions released on the npm too. 
@mmorhun  maybe we can try this one release https://github.com/theia-ide/theia/releases/tag/eclipse_initial it's newer than 0.3.12?

mmorhun:
@AndrienkoAleksandr unfortunately there is not such release on npmjs. So the question is open.

benoitf:
I think it's interesting to have "nightly" theia image as well so we can try any Nightly version of theia with the "master" version of each extensions.

so we can detect failures that we will have, etc. (fragile is ok there as the goal is to check and use "nightly")

so we need both ways:
- use of stable theia
- use of nightly theia (and for this one we need to fix all resolutions/extensions to be the version of Theia we want)

mmorhun:
So, this issue goal is to update version of Theia to 0.3.12 but we cannot use it, so I ma going to close this PR.
For all the improvements let's create separate issue with clear description and DOD.

PR #10467, opened 2018-07-18
Simplify the theia stack from 2 machines to 1 with Theia only

ashumilova:

### What does this PR do?

Simplifies the Theia stack to contain only one Theia machine by removing dev-machine (which is not used in this case) and switching from compose to dockerimage recipe type. It works on Openshift as well.

### DIFF
--- a/ide/che-core-ide-stacks/src/main/resources/stacks.json
+++ b/ide/che-core-ide-stacks/src/main/resources/stacks.json
@@ -1,12 +1,11 @@
 [
   {
-    "id": "java-theia-docker",
+    "id": "simple-theia-docker",
     "creator": "ide",
-    "name": "Java Theia (docker)",
-    "description": "Che development + Theia in a sidecar container (docker)",
+    "name": "Theia IDE on docker",
+    "description": "Theia in a sidecar container (docker)",
     "scope": "general",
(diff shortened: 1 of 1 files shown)


benoitf (approved):

AndrienkoAleksandr (approved):

ide/che-core-ide-stacks/src/main/resources/stacks.json
```diff
@@ -1,12 +1,11 @@
 [
   {
-    "id": "java-theia-docker",
+    "id": "simple-theia-docker",
     "creator": "ide",
-    "name": "Java Theia (docker)",
-    "description": "Che development + Theia in a sidecar container (docker)",
+    "name": "Simple Theia (docker)",
```
mmorhun: I would prefer to remove brackets, because stack validator says that the field is invalid (painful on copying the stack).
What about `Theia on docker`?

mmorhun (approved):

ide/che-core-ide-stacks/src/main/resources/stacks.json
ashumilova: Agree with you, maybe Theia IDE on docker?

PR #10477, opened 2018-07-19
Add Bearer token prefix for project service requests

mshaposhnik:
### What does this PR do?
This PR adds "Bearer" token identifier for the all project service related requests, since it is required by jwtproxy. 
That can greatly reduce number of failing tests on jwtproxy auth.

#### Release Notes
N/A

#### Docs PR
N/A


selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/client/TestProjectServiceClient.java
```diff
@@ -38,6 +38,8 @@
  */
 @Singleton
 public class TestProjectServiceClient {
+
+  private final String BEARER_TOKEN_PREFIX = "Bearer ";
```
sleshchenko: Should not be constant marked with `static`?

mshaposhnik:
I'll merge that into my feature branch, and then will request QE review on it.

Issue #10295, opened 2018-07-05
Kubernetes/OpenShift machines are started sequentially


sleshchenko:
### Description
Kubernetes/OpenShift machines are started sequentially. It leads to several issues:
1. Machines should be started asynchronously, it was implemented to speed up images pulling and machines bootstrapping.
2. Machines logs are not published because Kubernetes/OpenShift machine is not considered as created until Pod is RUNNING.
3. Unrecoverable events functionality doesn't work because there is no Pod related machine until Pod is RUNNING.

To fix it we should not wait until deployment will be ready https://github.com/eclipse/che/blob/master/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java#L748

### Reproduction Steps
1. Start K8s/OS multipod environment with one broken pod, like container there is defined mount to non-existing config map.
2. Check Workspace status tab:
Expected: Both machines should be displayed there as STARTING
Actual: Broken pod related machine is not displayed in workspace status tab.

3. Check Kubernetes/OpenShift console for unrecoverable events, they should be there for broken Pod, then
Expected: Workspace failed to start because of unrecoverable event.
Actuals: Workspace failed to start because of timeout.

4. Check machines logs while machines are starting, like an image is pulling.
Expected: Something like the following logs should be present
```
Container image "eclipse/ubuntu_jdk8" is pulling
Container image is pulled
Created container
Started container
```
Actual: There are no container related events.

P.S. For 2nd and 3rd steps single container workspace would be fine.

sleshchenko:
I figure out that parallel image pulling in case of single-pod depends on Kubernetes/OpenShift instance configuration (more in [OCP doc](https://access.redhat.com/documentation/en-us/openshift_container_platform/3.3/html/installation_and_configuration/install-config-master-node-configuration#master-node-configuration-parallel-image-pulls-with-docker)). 
We should not really expect that images will be pulled in parallel.

Also, since workspace loader starts workspace second step should be adapted in QA tests.

So, QA tests have to check that both machine are STARTING in the same time, Workspace API may be called to check it.

PR #10545, opened 2018-07-25
Fix sequential start of workspace machines

amisevsk:
### What does this PR do?
Fixes an issue when running on Kubernetes/OpenShift where workspace are started sequentially. This blocking breaks some useful functionality and can cause issues if a Pod is unable to start fully.

Instead of waiting for the Deployment to be ready (meaning all Pods in running state), return the Pod as soon as it is created.

A docker image with these changes should be available at `amisevsk/che-server:issue-10295`.

https://github.com/eclipse/che/issues/10295

#### Docs PR
N/A (bugfix)

### DIFF
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
@@ -13,6 +13,7 @@ package org.eclipse.che.workspace.infrastructure.kubernetes.namespace;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.concurrent.CompletableFuture.allOf;
+import static org.eclipse.che.workspace.infrastructure.kubernetes.Constants.CHE_DEPLOYMENT_NAME_LABEL;
 import static org.eclipse.che.workspace.infrastructure.kubernetes.Constants.CHE_WORKSPACE_ID_LABEL;
 import static org.eclipse.che.workspace.infrastructure.kubernetes.Constants.POD_STATUS_PHASE_FAILED;
 import static org.eclipse.che.workspace.infrastructure.kubernetes.Constants.POD_STATUS_PHASE_RUNNING;
@@ -88,7 +89,7 @@ public class KubernetesDeployments {
 
   // TODO https://github.com/eclipse/che/issues/7656
   public static final int POD_REMOVAL_TIMEOUT_MIN = 5;
-  public static final int POD_CREATION_TIMEOUT_MIN = 5;
+  public static final int POD_CREATION_TIMEOUT_MIN = 1;
 
   private static final String POD_OBJECT_KIND = "Pod";
   // error stream data initial capacity
@@ -123,29 +124,58 @@ public class KubernetesDeployments {
    */
   public Pod deploy(Pod pod) throws InfrastructureException {
     putLabel(pod, CHE_WORKSPACE_ID_LABEL, workspaceId);
+
     ObjectMeta metadata = pod.getMetadata();
+    // Note: metadata.name will be changed as for pods it is set by the deployment.
+    String originalName = metadata.getName();
+    putLabel(pod, CHE_DEPLOYMENT_NAME_LABEL, originalName);
+
     PodSpec podSpec = pod.getSpec();
     podSpec.setRestartPolicy("Always"); // Only allowable value
+    final CompletableFuture<Pod> createFuture = new CompletableFuture<>();
+    final Watch createWatch =
+        clientFactory
+            .create(workspaceId)
+            .pods()
+            .inNamespace(namespace)
+            .watch(new CreateWatcher(createFuture, workspaceId, originalName));
     try {
-      Deployment deployment =
-          clientFactory
-              .create(workspaceId)
-              .extensions()
-              .deployments()
-              .inNamespace(namespace)
-              .createNew()
-              .withMetadata(metadata)
-              .withNewSpec()
-              .withReplicas(1)
-              .withNewTemplate()
-              .withMetadata(metadata)
-              .withSpec(podSpec)
-              .endTemplate()
-              .endSpec()
-              .done();
-      return awaitAndGetPod(deployment);
+      clientFactory
+          .create(workspaceId)
+          .extensions()
+          .deployments()
+          .inNamespace(namespace)
+          .createNew()
+          .withMetadata(metadata)
+          .withNewSpec()
+          .withReplicas(1)
+          .withNewTemplate()
+          .withMetadata(metadata)
+          .withSpec(podSpec)
+          .endTemplate()
+          .endSpec()
+          .done();
+      return createFuture.get(POD_CREATION_TIMEOUT_MIN, TimeUnit.MINUTES);
     } catch (KubernetesClientException e) {
       throw new KubernetesInfrastructureException(e);
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new InfrastructureException(
+          String.format(
+              "Interrupted while waiting for Pod creation. -id: %s -message: %s",
+              metadata.getName(), e.getMessage()));
+    } catch (ExecutionException e) {
+      throw new InfrastructureException(
+          String.format(
+              "Error occured while waiting for Pod creation. -id: %s -message: %s",
+              metadata.getName(), e.getCause().getMessage()));
+    } catch (TimeoutException e) {
+      throw new InfrastructureException(
+          String.format(
+              "Pod creation timeout exceeded. -id: %s -message: %s",
+              metadata.getName(), e.getMessage()));
+    } finally {
+      createWatch.close();
     }
   }
 
@@ -731,34 +761,6 @@ public class KubernetesDeployments {
   }
 
   /**
-   * Wait until deployment is ready and return created Pod
-   *
-   * @param deploy the deployment responsible for creating the pod
-   * @return the created Pod
-   * @throws InfrastructureException if timeout elapses while waiting.
-   */
-  private Pod awaitAndGetPod(Deployment deploy) throws InfrastructureException {
-    String deploymentName = deploy.getMetadata().getName();
-    try {
-      clientFactory
-          .create(workspaceId)
-          .extensions()
-          .deployments()
-          .inNamespace(namespace)
-          .withName(deploymentName)
-          .waitUntilReady(POD_CREATION_TIMEOUT_MIN, TimeUnit.MINUTES);
-      return get(deploymentName)
-          .orElseThrow(
-              () ->
-                  new InfrastructureException(
-                      String.format("Failed to create Deployment %s", deploymentName)));
-    } catch (InterruptedException e) {
-      throw new InfrastructureException(
-          String.format("Timeout while creating Deployment %s", deploymentName));
-    }
-  }
-
-  /**
    * Get the Deployment resource that owns a specified Pod via the Pod's owner references.
    *
    * @param pod the pod controlled by desired deployment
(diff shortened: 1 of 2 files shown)


riuvshin:

amisevsk:
I would also like to add some test cases to cover this issue to avoid it happening in the future, but I cannot see a good way to write an implementation independent unit test so it may have to be managed in functional tests.

ibuziuk:
@amisevsk could you please test if https://github.com/eclipse/che/issues/9946 is fixed by this PR ?

skabashnyuk:
ci-test

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
```diff
@@ -125,7 +125,16 @@ public Pod deploy(Pod pod) throws InfrastructureException {
...
+    Watch toCloseOnException = null;
     try {
+      final CompletableFuture<Pod> createFuture = new CompletableFuture<>();
+      final Watch createWatch =
+          clientFactory
+              .create(workspaceId)
+              .pods()
+              .inNamespace(namespace)
+              .watch(new CreateWatcher(createFuture, workspaceId));
+      toCloseOnException = createWatch;
```
garagatyi: Can't we move `createWatch` declaration before try block and not use an additional variable to close it in the finally block?

garagatyi (approved):
Looks good even though I'm not sure whether this PR brings back speed we had with bare pods creation. Back then request didn't require any watcher and now we have it. Would be cool to know whether we can improve the approach even further in the future. 
But the current solution is good too!
I'm also wondering whether @amisevsk you want to add some unit tests ;)

garagatyi:
Since @sleshchenko is on PTO would be nice to have some reviews from other members of the platform team. @skabashnyuk @mshaposhnik FYI

riuvshin:
ci-test build report:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/521/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/521//Selenium_tests_report)
[selenium tests report data](http://tests.codenvycorp.com/selenium_test_report_data/10545)
docker image: eclipseche/che-server:10545

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
amisevsk: Good catch, I've made the change.

amisevsk:
@ibuziuk This PR brings back logs to workspace launch, so when I launch a workspace without any PVs to bind a PVC to the error does appear in the logs, but the error I'm getting locally is still a timeout issue. Perhaps the switch to Deployments also broke unrecoverable events handling? Also for some reason today I'm not seeing the actual IDE load -- just a loading bar with error logs printed within the window [1]. After a timeout the server still spits logs as in https://github.com/eclipse/che/issues/9946#issuecomment-398356719

@garagatyi Regarding speed, the overhead here should be small apart from creating the watch -- in my testing the `get` call on the watch returns immediately most of the time. I don't know if there is any way around this though, as the deployment creation call can return before any pods are created, and the method requires returning the created Pod. We can look into this in the future.

[1]:

amisevsk:
@ibuziuk Looking into it more closely, the Deployments change seems to have broken some events functionality -- no events are being listed as "workspace events", likely due to the difference in naming (deployments name pods with a hash). I added a bit of logging and am getting this on failed launches:

```
2018-07-27 00:13:21,282[default.svc/...]  [ERROR] [.i.k.KubernetesInternalRuntime 730]  - IS WORKSPACE EVENT: false
2018-07-27 00:13:21,283[default.svc/...]  [ERROR] [.i.k.KubernetesInternalRuntime 731]  - IS UNRECOVERABLE:   true
2018-07-27 00:13:21,283[default.svc/...]  [ERROR] [.i.k.KubernetesInternalRuntime 732]  -  EVENT: PodEvent{podName='workspacemn1seqjgj6xgrugz.dockerimage-6db7db4d99-j5dmh', containerName='null', reason='FailedMount', message='Unable to mount volumes for pod "workspacemn1seqjgj6xgrugz.dockerimage-6db7db4d99-j5dmh_mini-che(94ab01eb-9131-11e8-9d9d-5254009779e3)": timeout expired waiting for volumes to attach/mount for pod "mini-che"/"workspacemn1seqjgj6xgrugz.dockerimage-6db7db4d99-j5dmh". list of unattached/unmounted volumes=[claim-che-workspace-5c7ny8op claim-che-workspace-x0xjyc0a default-token-nm8t4]', creationTimestamp='2018-07-27T00:13:21Z', lastTimestamp='2018-07-27T00:13:21Z'}
```

I'll open an issue.

garagatyi:
> I don't know if there is any way around this though, as the deployment creation call can return before any pods are created, and the method requires returning the created Pod. We can look into this in the future.

But if deployment is returned before the creation of the POD how our code would work? I mean the code that immediately after that uses fields of the returned pod object to create Che machine object.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
```diff
@@ -852,6 +841,31 @@ private String getPodName(String name) throws InfrastructureException {
...
+
+    private CreateWatcher(CompletableFuture<Pod> future, String workspaceId) {
+      this.future = future;
+      this.workspaceId = workspaceId;
+    }
+
+    @Override
+    public void eventReceived(Action action, Pod resource) {
+      Map<String, String> labels = resource.getMetadata().getLabels();
+      if (workspaceId.equals(labels.get(CHE_WORKSPACE_ID_LABEL))) {
```
sleshchenko: I afraid comparing workspace id label only is not enough. It may don't work for multi pods workspaces. Are we able to compare deployment specific label too to ensure that created pod really belongs to needed deployment?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
```diff
@@ -125,26 +125,43 @@ public Pod deploy(Pod pod) throws InfrastructureException {
...
       throw new KubernetesInfrastructureException(e);
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new InfrastructureException(
+          "Interrupted while waiting for Pod creation. " + e.getMessage());
+    } catch (ExecutionException e) {
+      throw new InfrastructureException(
+          "Error occured while waiting for Pod creation. " + e.getMessage());
+    } catch (TimeoutException e) {
+      throw new InfrastructureException("Pod creation timeout exceeded " + e.getMessage());
```
sleshchenko: Please consider adding `.` not to mix our message and exception message

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
```diff
@@ -125,26 +125,43 @@ public Pod deploy(Pod pod) throws InfrastructureException {
...
+          .withSpec(podSpec)
+          .endTemplate()
+          .endSpec()
+          .done();
+      return createFuture.get(POD_CREATION_TIMEOUT_MIN, TimeUnit.MINUTES);
     } catch (KubernetesClientException e) {
       throw new KubernetesInfrastructureException(e);
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
+      throw new InfrastructureException(
```
sleshchenko, line 163: I see that the same exception is thrown in other components with similar logic like `KubernetesNamespace`, `KubernetesIngresses` but `RuntimeStartInterruptedException` may be better here. 
I would say that it is just a note since it is not related to this PR. I'll try to take a look how it may influence workspace start interruption and I will create an issue if needed.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
```diff
@@ -125,26 +125,43 @@ public Pod deploy(Pod pod) throws InfrastructureException {
...
+          "Interrupted while waiting for Pod creation. " + e.getMessage());
+    } catch (ExecutionException e) {
+      throw new InfrastructureException(
+          "Error occured while waiting for Pod creation. " + e.getMessage());
```
sleshchenko: Does `e.getMessage` contain a meaningful message or it is needed to fetch the message of the cause exception? Like in [KubernetesNamespace](https://github.com/eclipse/che/blob/3ad17d60c9b5c5b09ab508296f1e0b15ff054d06/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesNamespace.java#L251). Please check which way is better.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
sleshchenko: It may be useful to have an identifier of pod or deployment in error message.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
amisevsk: ~~I'm not sure there's a good way to do this with the the infra is currently written.~~

- ~~One path would be to follow `OwnerReferences` on the pod up to the deployment that created it, *a la* `getDeploymentResource()`. This would work but could potentially be pretty slow (multiple calls to OS API).~~
- ~~Alternatively, we could generate a random label to go with the deployment and pass that in to the watcher, but this seems like a total hack.~~

~~The issue I'm having here is that multi-pod/deployment workspaces seem to be impossible on OS at the moment -- the API gives no option for multiple pods in a single deployment -- only one `podSpec` is permitted (although multiple containers in a single pod works perfectly fine), and attempting to make multiple pods/deployments through sequential calls to the `deploy()` method would fail due to a name collision (deployment names are e.g. `workspaceId.dockerfile`). Apart from that, the only unique label assigned to created pods (that we have access to when the watcher is created) is workspaceId.~~

~~In short, AFAICT, the current setup is that we require a 1-1 correspondance between `workspaceId`, a deployment name, and the label given to pods. Unless I'm missing something I feel like this issue should be resolved if/when multi-pod workspaces are implemented.~~

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
amisevsk: I've added identifier (pod name) to the errors thrown here. 

Regarding the `RuntimeStartInterruptedException`, I'm not sure what we should use for env or ownerId here. Also, an interrupted exception is handled in [`KubernetesInternalRuntime`](https://github.com/eclipse/che/blob/44588aa775e77f629eb0a0052ac20d949ab8c7c3/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java#L214), which throws a `RuntimeStartInterruptedException` as you suggest.

amisevsk:
@garagatyi I'm not sure I fully understand you. The issue I was referring to was that the `kubernetesClient.deployments()....create()` call returns as soon as the OpenShift API creates a deployment. Since a deployment requests pods be created in its name, the creation of a deployment is disconnected from the creation of the pod it controls, and so without the wait, we cannot guarantee that any pod exists (I don't think there's a way to go directly from a `deployment` to the `pod`(s) it controls except by label matching). 

What I'm trying to say is that pod creation is asynchronous with respect to deployment creation, so even something like 'get all pods with labels corresponding to the deployments `matchLabels()`' will fail sometimes (i.e. return a list of zero pods). This means that the watch, and associated overhead, is unavoidable in the deployments approach -- luckily the future seems to be complete by the time we need it anyways, so the overhead is negligible.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
amisevsk: Actually I thought about it more and realized you are right. I'll look into this, and likely add the deployment name as a label in the `deploy()` method.

amisevsk: Fixed -- pods now also get a label to refer to the deployment that created them.

garagatyi:
@amisevsk never mind, I misunderstood some part of the code, now I got it

sleshchenko (approved):
LGTM

garagatyi:
ci-test

riuvshin:
ci-tests

riuvshin:
ci-test

riuvshin:
ci-test build report:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/616/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/616//Selenium_20tests_20report/)
[selenium tests report data](http://tests.codenvycorp.com/selenium_test_report_data/10545)
docker image: eclipseche/che-server:10545

Issue #10476, opened 2018-07-19
Allow to apply patches for the Theia released version used in eclipse/che-theia image 


ashumilova:
In some cases the released version of Theia may contain some annoying bug(s) and the the fix of it is in the upstream. So for us to have some kind of hotfix and do not wait for the next Theia release, there is need to have mechanism of applying patches on top of the released version.

benoitf:
Fixed

Issue #10275, opened 2018-07-04
Che’s Theia extension-ready package


slemeur:
### Description
On our road to introduce the new IDE, based on Eclipse Theia, we have been building a set of specific extensions for Che.

| Project        | Description  |
| ----------- |-----|
| [GitHub Plugin](https://github.com/eclipse/che-theia-github-plugin) | GitHub VCS support.  |
| [Java Plugin](https://github.com/eclipse/che-theia-java-plugin) | Java language support. |
| [SSH Plugin](https://github.com/eclipse/che-theia-ssh-plugin) | SSH public/private key pairs management. |
| [Terminal Plugin](https://github.com/eclipse/che-theia-terminal-plugin) | Terminal plugin. |
| [Machines Plugin](https://github.com/eclipse/che-theia-machines-plugin) | Workspace's containers controller plugin. |
| [Hosted plugin manager](https://github.com/eclipse/che-theia-hosted-plugin-manager-extension) | IDE hosting handler for url changing and port forwarding. |
| [Commands plugin](https://github.com/eclipse/che-theia-task-plugin) | Commands support of Che workspaces. |
| [Factories plugin](https://github.com/eclipse/che-theia-factory-extension) | Factories extension to support Che's factories in Eclipse Theia. |

We need to:
- create a Che's Theia package with all the existing plugins (or extensions) we have
- identify from the list, the specific tools for che which needs to adopt the plugin model and migrate those
- make that Che's Theia package available for anyone looking to the new IDE from Che.

### Sub-Tasks

- [x] **STEP 1**: Create a `che-theia` Docker image with all plugins packaged in it (frontend and backend in the same container)
  - [x] Create a Dockerfile (and eventual scripts/metadata)
  - [x] Setup a CI to push the image on DockerHub
  - [x] Create a new stack with a Dockerimage recipe that point to the Che image

- [ ] Make extensions and plugins properly work on k8s infrastructure (TODO add depency issues)

AndrienkoAleksandr:
@slemeur I removed env-variables-plugin from the table this is deprecated stuff. For Theia extensions this functionality is available from theia-core, for plugins we have https://github.com/theia-ide/theia/blob/master/packages/plugin/src/theia.d.ts#L2405

slemeur:
Thanks @AndrienkoAleksandr !

l0rd:
@slemeur should this run on top of ws.next or on top of current workspace model?

l0rd:
[...]

PR #10552, opened 2018-07-26
[ Theia Docker image] Updates

benoitf:
### What does this PR do?

- Allow to make patches on our theia instance
- Move to 0.3.13
- upgraded to NodeJS 9
- introduce Java inside image to have java intellisense
- move all Dockerfile logic to separate setup file
- allow to apply patches on top of existing theia instance
- recompile all extensions to ensure they're compatible with current Theia version
- Then we publish internally a new version on a local registry, and change dependency version on all extensions to make compliant with this version
- install globally typescript so typescript lsp is working

Add 2 extensions :
 - https://github.com/eclipse/che-theia-github-plugin
- https://github.com/eclipse/che-theia-java-plugin

Two can't be added as not compliant with openshift:
 - https://github.com/eclipse/che-theia-terminal-plugin
-  https://github.com/eclipse/che-theia-task-plugin

tested on che.openshift.io https://che.openshift.io/f?id=factorybyszeyh3vowtyyxw

https://github.com/eclipse/che/issues/10476
https://github.com/eclipse/che/issues/10275

Change-Id: Ic6ba219e35b0e83783dc37b28f05dd54f2864cf2

### DIFF
--- a/dockerfiles/theia/src/add-extensions.js
+++ b/dockerfiles/theia/src/add-extensions.js
@@ -36,7 +36,7 @@ function addExtensions(extensions) {
     const extensionsToAdd = {};
     for (let extension of extensions) {
         const extensionName = extension["name"];
-        const extensionRootPath = EXTENSIONS_DIR + '/' + extensionName.replace(/\//g, '_');
+        let extensionRootPath = EXTENSIONS_DIR + '/' + extensionName.replace(/\//g, '_');
 
         const extensionType = extension["type"];
         switch(extensionType) {
@@ -49,8 +49,14 @@ function addExtensions(extensions) {
             default:
                 throw new Error(`Invalid extension type ${extensionType}.`);
         }
-        buildExtension(extensionRootPath);
-        extensionsToAdd[extensionName] = getBinaryPath(extensionRootPath, extensionName);
+        const buildFolder = extension["folder"];
+        const originalExtensionPath = extensionRootPath;
+        if (buildFolder) {
+            console.log('There is a build folder defined', buildFolder);
+            extensionRootPath = extensionRootPath + '/' + buildFolder;
+        }
+        buildExtension(extensionRootPath, originalExtensionPath);
+        extensionsToAdd[extensionName] = 'file://' + extensionRootPath;
     }
     console.log("Extension to add: ", extensionsToAdd);
     addExtensionsIntoDefaultPackageJson(extensionsToAdd);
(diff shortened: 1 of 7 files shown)


dockerfiles/theia/src/add-extensions.js
```diff
@@ -98,38 +104,28 @@ function checkoutRepo(path, checkoutTarget) {
...
+            fs.unlinkSync(rootPath + '/lerna.json');
+        }
+        //spawnSync(`node`, [`${DEFAULT_THEIA_ROOT}/resolutions-provider.js`, `${rootPath}/package.json`], {cwd: `${rootPath}`});
         console.log('Building extension: ', path);
         const nodeModulesPath = `${DEFAULT_THEIA_ROOT}/node_modules`;
         // build extension, but use Theia node_modules to reuse dependencies and prevent growing docker image.
-        spawnSync(`yarn`, ['--modules-folder', nodeModulesPath, '--global-folder', nodeModulesPath, '--cache-folder', nodeModulesPath], {cwd: `${path}`, stdio:[0,1,2]});
+        spawnSync(`yarn`, ['--modules-folder', nodeModulesPath, '--global-folder', nodeModulesPath], {cwd: `${path}`, stdio:[0,1,2]});
     } catch (error) {
+        console.log(error);
```
AndrienkoAleksandr: Maybe it would be better to use console.error here?

dockerfiles/theia/src/add-extensions.js
```diff
@@ -98,38 +104,28 @@ function checkoutRepo(path, checkoutTarget) {
     }
 }
 
-function buildExtension(path) {
+function buildExtension(path,rootPath) {
```
AndrienkoAleksandr: Format `path,rootPath` => `path, rootPath`

AndrienkoAleksandr:
Impressive!

AndrienkoAleksandr (approved):

dockerfiles/theia/src/add-extensions.js
benoitf: sure

dockerfiles/theia/src/add-extensions.js
benoitf: ok

dockerfiles/theia/src/add-extensions.js
benoitf: fixed

dockerfiles/theia/src/add-extensions.js
benoitf: fixed

mmorhun (approved):

dockerfiles/theia/src/extensions.json
```diff
@@ -26,6 +33,20 @@
             "source": "https://github.com/eclipse/che-theia-hosted-plugin-manager-extension.git",
             "checkoutTo": "latest-deps",
             "type": "git"
+        },
+        {
+            "name": "@eclipse-che/theia-task-extension",
+            "source": "https://github.com/eclipse/che-theia-task-plugin",
```
AndrienkoAleksandr: Task plugin(actually for now extension) use terminal server side, without this part it's compiles but doesn't work.

benoitf: removed

AndrienkoAleksandr (approved):

evidolob (approved):

riuvshin:
should this also be changed in here https://github.com/eclipse/che/blob/master/ide/che-core-ide-stacks/src/main/resources/stacks.json#L53 ? and other places where we have older version `0.4.0-next.b17727c1`

benoitf:
@riuvshin yes I just wanted to change when image will be published

riuvshin:
why not in this PR? I mean there is no risk to do it with this PR because it is only stacks.json so build will not be affected

benoitf:
AFAIK, if someone grab and compile che, image will not be there

riuvshin:
we will generate images right after merge :)

benoitf:
@riuvshin yes this why I will update after images are generated. Else you might have ppl with links to an unknown image. 
Also image build may fail due to randomly errors

riuvshin (approved):

benoitf:
@riuvshin could you trigger the build of theia image ?

Issue #10569, opened 2018-07-30
Fail of Che build


Ohrimenko1988:
### Description
The error appears during che build.

### Reproduction Steps
 - Open a terminal and go to Che root folder
 - Checkout to the master branche and pull the latest changes
 - Cleanup docker (perform next commands: "docker rm -f $(docker ps -aq)", "docker rmi -f $(docker images -aq)", "docker network prune -f")
 - Perfom next command for che building: "mvn clean install -DskipTests -Dfindbugs.skip -Dskip-enforce=true -Dskip-validate-sources=true -Dfindbugs.skip"
 - Wait while the module "Che Workspace Loader" is being built
 - Pay attention to the fail of the build.

**OS and version:**    
Che 6.9.0; Fedora 28; npm 5.6.0; Docker 17.12.0; Node 8.11.3

**Attachment:**
![screenshot from 2018-07-30 09-57-04](https://user-images.githubusercontent.com/22724619/43383512-e5d12d64-93e3-11e8-9178-6a541c06336e.png)

![screenshot from 2018-07-30 09-55-50](https://user-images.githubusercontent.com/22724619/43383515-e87c9e04-93e3-11e8-9722-263eebbbd00d.png)

PR #10572, opened 2018-07-30
Fix workspace-loader application build

akurinnoy:
### What does this PR do?
This PR fixes the issue when all existing Jest tests in workspace-loader application are failed. 
File `package-lock.json` is added to stick with current dependencies.

#10569

#### Release Notes

N/A

#### Docs PR

N/A

### DIFF
--- a/workspace-loader/.gitignore
+++ b/workspace-loader/.gitignore
@@ -1,3 +1,2 @@
 node_modules/
 dist/
-package-lock.json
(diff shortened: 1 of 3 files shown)


olexii4 (approved):

evidolob (approved):

benoitf (approved):
should we use yarn instead of npm ?

akurinnoy:
@benoitf I'm OK to use any of them, but could you explain your reason for using yarn here?

Issue #10292, opened 2018-07-05
Add method to the "WaitUtils" class for waiting specified condition happens without using of "WebDriverWait"


Ohrimenko1988:
### Description
We should create a method in the "WaitUtils" class which will take a condition as an argument and will wait for its true result. This method should use only standard sleep commands without using "WebDriverWait". It will allow to get rid of duplication and boilerplate code.
**Note!** After done of this issue the "HotUpdateUtil" class should be refactored with using of the created method.

Ohrimenko1988:
Done https://github.com/eclipse/che/pull/10576

PR #10576, opened 2018-07-30
Selenium: Add method to the "WaitUtils" class for waiting specified condition happens without using of "WebDriverWait"

Ohrimenko1988:
### What does this PR do?
Add method to the "WaitUtils" class for waiting specified condition happens without using of "WebDriverWait"

Issue: https://github.com/eclipse/che/issues/10292


selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/utils/WaitUtils.java
```diff
@@ -45,4 +52,41 @@ public static void sleepQuietly(int timeout, TimeUnit timeUnit) {
...
+   * @param condition expression which should be performed
+   */
+  public static void waitSuccessCondition(BooleanCondition condition) {
+    final int defaultTimeout = LOAD_PAGE_TIMEOUT_SEC;
+    waitSuccessCondition(condition, defaultTimeout);
+  }
+
+  @FunctionalInterface
+  public interface BooleanCondition {
+    boolean execution();
```
dmytro-ndp: Looks as redundant.

dmytro-ndp: You can use `BooleanSupplier` instead.

Ohrimenko1988: You are definitely right, thank you.

SkorikSergey (approved):

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/utils/WaitUtils.java
```diff
@@ -46,4 +54,67 @@ public static void sleepQuietly(int timeout, TimeUnit timeUnit) {
...
+   */
+  public static void waitSuccessCondition(
+      BooleanSupplier condition,
+      int timeout,
+      int delayBetweenAttemptsInMilliseconds,
+      TimeUnit timeUnit) {
+    final long waitingTime = timeUnit.toMillis(timeout);
+    final long startingTime = System.currentTimeMillis();
+
+    while (isTimeAvailable(startingTime, waitingTime)) {
```
dmytro-ndp: `isTimeAvailable()` method looks like kind of over engineering, IMHO.
I would involve `finishTime` instead of `startingTime` and then check it inside the `while` staement as follow:
```
while (System.currentTimeMillis() < finishTime) {
...
```

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/utils/WaitUtils.java
```diff
@@ -46,4 +54,67 @@ public static void sleepQuietly(int timeout, TimeUnit timeUnit) {
...
+      if (condition.getAsBoolean()) {
+        break;
```
dmytro-ndp: You need to return from the method immediately if condition succeeded. Otherwise, with `break`, program goes further and throw timeout exception.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/utils/WaitUtils.java
```diff
@@ -46,4 +54,67 @@ public static void sleepQuietly(int timeout, TimeUnit timeUnit) {
...
+   *
+   * @param condition expression which should be performed
+   * @param timeout waiting time
+   * @param delayBetweenAttemptsInMilliseconds delay between tries of {@code condition} execution in
+   *     milliseconds
+   */
+  public static void waitSuccessCondition(
+      BooleanSupplier condition,
+      int timeout,
+      int delayBetweenAttemptsInMilliseconds,
```
dmytro-ndp: `long` type is more suitable to present milliseconds in delay and timeout

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/utils/WaitUtils.java
```diff
@@ -46,4 +54,67 @@ public static void sleepQuietly(int timeout, TimeUnit timeUnit) {
...
+   * Waits during {@code timeout} until {@code condition} has a "true" state.
+   *
+   * @param condition expression which should be performed
+   * @param timeout waiting time
+   * @param delayBetweenAttemptsInMilliseconds delay between tries of {@code condition} execution in
+   *     milliseconds
+   */
+  public static void waitSuccessCondition(
+      BooleanSupplier condition,
+      int timeout,
```
dmytro-ndp: it's better to declare implicitely that timeout is in milliseconds itself, by naming - `timeoutInMillisec`

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/utils/WaitUtils.java
```diff
@@ -46,4 +54,67 @@ public static void sleepQuietly(int timeout, TimeUnit timeUnit) {
...
+      int delayBetweenAttemptsInMilliseconds,
+      TimeUnit timeUnit) {
```
dmytro-ndp: It's not clear time unit of what `timeUnit` argument is - _timeout_ or _delay_. 
To simplify the code I would better to define both _timeout_ and _delay_ in millisec without giving an ability to define time unit at all.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/utils/WaitUtils.java
```diff
@@ -46,4 +60,79 @@ public static void sleepQuietly(int timeout, TimeUnit timeUnit) {
...
+      long delayBetweenAttemptsInMilliseconds,
+      TimeUnit timeoutTimeUnit)
+      throws InterruptedException {
+    final long waitingTime = timeoutTimeUnit.toMillis(timeout);
+    final long startingTime = System.currentTimeMillis();
+    final long finishTime = startingTime + waitingTime;
+
+    Callable<Boolean> waitTrueState =
+        () -> {
+          while (System.currentTimeMillis() <= finishTime) {
```
dmytro-ndp: If `System.currentTimeMillis() = finishTime` it doesn't make sense to continue execution so as timeout exceeded.

dmytro-ndp (approved):

dmytro-ndp:
ci-build

Issue #10465, opened 2018-07-18
Upgrade project license from EPL-1.0 to EPL-2.0


skabashnyuk:
### Description
We want to upgrade project license from EPL-1.0 to EPL-2.0
list of repos https://github.com/orgs/eclipse/teams/eclipse-che/repositories
## che-platform team
- [x] https://github.com/eclipse/che
- [x] https://github.com/eclipse/che-dev
- [x] https://github.com/eclipse/che-dockerfiles
- [x] https://github.com/eclipse/che-docs
- [x] https://github.com/eclipse/che-jwtproxy
- [x] https://github.com/eclipse/che-lib
- [x] https://github.com/eclipse/che-parent

## che-ide team
- [x] https://github.com/eclipse/che-theia-emacs-plugin -> empty
- [x] https://github.com/eclipse/che-theia-vi-plugin -> empty
- [x] https://github.com/eclipse/che-theia-task-plugin

## che-ide2 team
- [x] https://github.com/eclipse/che-theia-env-variables-plugin -> deprecated
- [x] https://github.com/eclipse/che-theia-factory-extension
- [x] https://github.com/eclipse/che-theia-github-plugin
- [x] https://github.com/eclipse/che-theia-hosted-plugin-manager-extension
- [x] https://github.com/eclipse/che-theia-machines-plugin
- [x] https://github.com/eclipse/che-theia-oauth-authentication-plugin -> empty
- [x] https://github.com/eclipse/che-theia-ssh-plugin
- [x] https://github.com/eclipse/che-theia-terminal-plugin
- [x] https://github.com/eclipse/che-workspace-client

## Deprecated reamain on EPL-1.0
- [x] https://github.com/eclipse/che-plugin-svn
- [x] https://github.com/eclipse/che-archetypes
- [x] https://github.com/eclipse/che-dependencies
 

On 07/17/2018 05:10 PM, Richard Fontana wrote:
> EPL 1.0 code can be upgraded by the user to EPL 2.0 without further
> permission from the copyright holder:

skabashnyuk:
@evidolob @tsmaeder please take a look on your's repositories and change license to EPL-2.0

mkuznyetsov:
Che has remaining about 1300 files with EPL 1.0 license headers
these are mostly placed in following modules:
- /pluginsplugin-java/che-plugin-java-ext-lang-client
- /plugins/plugin-java/che-plugin-java-ext-lang-server
[...]

PR #10597, opened 2018-07-31
Selenium: Update license headers in 'selenium' module

Ohrimenko1988:
### What does this PR do?
Update license headers in 'selenium' module


dmytro-ndp:
Why did you decide to restore the headers? There is an issue https://github.com/eclipse/che/issues/10465

Issue #10612, opened 2018-08-01
Fix selenium tests after updating licenses headers


musienko-maxim:
### Description
We have many tests which depend on setting of cursor position (refactoring, debugging and etc.). So we should adapt all depended tests according to new licence headers to fix regression https://ci.codenvycorp.com/view/qa/job/che-integration-tests-multiuser-master-ocp/304

### Reference
https://github.com/eclipse/che/issues/10465

PR #10622, opened 2018-08-01
Set Che QE team members as owners of selenium tests sub-project

dmytro-ndp:
### What does this PR do?
It sets next members of Che QE team as an owners of _Eclipse Che_ E2E selenium tests situated in the sub-project [che/selenium](https://github.com/eclipse/che/tree/master/selenium): @musienko-maxim, @dmytro-ndp, @Ohrimenko1988 

It will prevent from merging unchecked changes into selenium tests which could cause regression like the follow #10612 into the **master** branch.


vparfonov (approved):

gazarenkov (approved):
I would prefer to have more fine grained codeownership
i.e. one person (2 max) per piece, but it's better then nothing anyways

skabashnyuk (approved):

Issue #10617, opened 2018-08-01
Migrate to mockito/mockito-testng


skabashnyuk:
### Description
Since we would like to use dependencies that are published on Maven Central we need to migrate to 
https://github.com/mockito/mockito-testng. We need to make sure that v0.0.1 artifacts and our test are compatible. 

Linked issue https://github.com/mockito/mockito/issues/962

PR #10639, opened 2018-08-02
move to central

riuvshin:
related to eclipse/che#6444

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2006,18 +2006,6 @@
             <url>https://maven.codenvycorp.com/content/repositories/codenvy-public-snapshots/</url>
         </repository>
     </repositories>
-    <pluginRepositories>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -2006,18 +2006,6 @@
             <url>https://maven.codenvycorp.com/content/repositories/codenvy-public-snapshots/</url>
```
riuvshin, line 2011: note we will still have this because we have some artifacts that exists only on codenvy nexus, but once we will have it on central we will get rid of it.

benoitf (approved):

skabashnyuk (commented):
You want to merge it before https://github.com/eclipse/che/issues/10617 or after?

skabashnyuk (approved):

PR #10673, opened 2018-08-06
Make Resource memory request configurable for k8s and OpenShift

i300543:
Signed-off-by: Sergey Kuperman <sergey.kuperman@sap.com>

### What does this PR do?
Defines memory request parameter when provisioning user workspaces for k8s and openshift, which would define memory reserved on creation, and not only the limit for max workspace memory.
without this reserved memory = limit memory which would not allow to create workspaces greedily 
(if workspacesCount*memoryLimit > availableMemory -> workspace creation fails, or containers are destroyed to free up memory)

#### Release notes
Adding additional attribute when creating workspace : 
      'memoryRequestBytes' - which will define memory requested to allocate on creation for the 
      workspace
      similarly to attribute 'memoryLimitBytes' which will define maximum memory workspace restriction
Those attributes can be passed to create workspace payload call (wsmaster endpoint (POST) /workspace)
If only the limit is present, requested memory = limit.

#### Changelog
 - rename che.workspace.default_memory_mb property to che.workspace.default_memory_limit_mb
 - add che.workspace.default_memory_request_mb property
 - fix all references to the limit property, add usage of the request property in all relevant places (hopefully)
 - did not touch the dashboard project or EnvironmentRamCalculator class

#### Docs PR
 will add later


skabashnyuk:
I guess you have a similar pr. No?

i300543:
No, the other PR is about shortening the hostname in the helm deployment scripts.
i pushed this PR over the helm PR by mistake today (fixed this already) 
this one is currently unique i hope

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
```diff
@@ -94,8 +94,10 @@ che.workspace.terminal_linux_arm7=${che.home}/lib/linux_arm7/terminal
 
 che.workspace.exec_linux_amd64=${che.home}/lib/linux_amd64/exec
 
-# RAM default for each machine that has no RAM settings in environment.
-che.workspace.default_memory_mb=1024
+# RAM limit default for each machine that has no RAM settings in environment.
```
skabashnyuk, line 97: is this a property rename?

skabashnyuk:
ci-test

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
i300543: yes

skabashnyuk: So we need.
1. Add an alias for backward compatibility. https://github.com/eclipse/che/blob/master/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che_aliases.properties#L13
2. Replace all existed values including ```CHE_WORKSPACE_DEFAULT__MEMORY__MB``` .   https://github.com/eclipse/che/blob/master/dockerfiles/init/manifests/che.env#L95
3. Write some release notes to explain what is going on.

skabashnyuk: 4. Explain why you need 2GB instead of all 1GB

i300543: i will check what is needed regarding 1,2
3. release notes in the PR itself right ? or the documentation ?
4. I dont actually need 2GB , i just wanted default requested memory be less than the limit,
but if the limit stays 1GB and requested is lets say 512Mb that seems too little no ? would you suggest request to be the same as the limit by default ?

skabashnyuk: 3. PR. for the docs that is another story. I think we need @eivantsov opinion here.
4. Analyzing.

riuvshin:
ci-test build report:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/554/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/554//Selenium_tests_report)
[selenium tests report data](http://tests.codenvycorp.com/selenium_test_report_data/10673)
docker image: eclipseche/che-server:10673

skabashnyuk:
@i300543 can you explain a bit more difference between ```che.workspace.default_memory_limit_mb``` and ```che.workspace.default_memory_request_mb``` it's not very clear from description.

i300543:
che.workspace.default_memory_limit_mb is translated into  container.resources.limits.memory
che.workspace.default_memory_limit_mb is translated into container.resources.requests.memory
those last 2 are basically parameters which are passed to k8s when provisioning a container (pod),
see more description here : 
https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/
so this means that the amount of memory "preallocated" for container is the requested memory,
and the amount of memory container is allowed to grow to is the limit.
if the containers grows bigger than the limit k8s might decide to kill it
allocated memory affects how many containers may share the same physical node, and allows you to create more workspaces, in cases when workspaces dont actually need amount of ram defined in their limit

skabashnyuk:
How will the docker runtime react on these properties?

i300543:
I have checked the code, and i think docker runtime is currently not affected by the additional parameter.
the machine config saves the passed attribute, but does not (currently) translate it to docker parameter.
If i am not mistaken this happens in MemoryAttributeConverter::provision-> DockerContainerConfig::setMemLimit,
so here i did not translate the new property to docker parameter.
i think you are working with docker version < 3:
https://docs.docker.com/compose/compose-file/compose-file-v2/#cpu-and-other-resources
so i could add method for setting that param in DockerContainerConfig (mem_reservation)
and call it in MemoryAttributeConverter right ?
or maybe do it in another pr ?

skabashnyuk:
I guess we need some assistance from @sleshchenko @garagatyi .

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
sleshchenko: 4. For me, 512mb as request and 1gb as limit looks OK. 
It should not change the current behavior, am I right?

sleshchenko (commented):
Overall LGTM

Please take a look my inlined comments/proposals

sleshchenko:
> I have checked the code, and i think docker runtime is currently not affected by the additional parameter.
the machine config saves the passed attribute, but does not (currently) translate it to docker parameter.
If i am not mistaken this happens in MemoryAttributeConverter::provision-> DockerContainerConfig::setMemLimit,
so here i did not translate the new property to docker parameter.
i think you are working with docker version < 3:
https://docs.docker.com/compose/compose-file/compose-file-v2/#cpu-and-other-resources
so i could add method for setting that param in DockerContainerConfig (mem_reservation)

AFAIS We use [1.23 docker api ](https://github.com/eclipse/che/blob/44588aa775e77f629eb0a0052ac20d949ab8c7c3/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties#L226)
According to [Docker Docs](https://docs.docker.com/engine/api/v1.23/) there is the corresponding field in `ContainerConfig#HostConfig#MemoryReservation`

> and call it in MemoryAttributeConverter right ?

As far as I understand, there few places that should be changed: DockerContainerConfig, ContainerConfig (add memory resarvation field), MemoryAttributeConverter(move memory reservation from MachineConfig to DockerContainerConfig), DockerMachineStarter (move memory reservation from DockerContainerConfig to ContainerConfig).

> or maybe do it in another pr ?

I would say that it's up to you do it in this PR or separate one. But it would be more consistent to do it in one PR, otherwise there will be some period when memoryRequest is introduced but not supported by Docker infra. In this case, it should be documented somewhere.

@garagatyi Please fix me if I am wrong or missed something

dockerfiles/init/manifests/che.env
```diff
@@ -92,7 +92,7 @@
 
 # Memory
 #     The recommended RAM size for new workspaces when created from the dashboard.
-#CHE_WORKSPACE_DEFAULT__MEMORY__MB=1024
+#CHE_WORKSPACE_DEFAULT__MEMORY__LIMIT__MB=1024
```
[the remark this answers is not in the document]
i300543: i did not make any changes in the dashboard with regards to the change, there is no UI possibility to config this from dashboard. Should i still do it ?

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/compose/ComposeEnvironmentFactory.java
```diff
@@ -116,14 +122,32 @@ void addRamLimitAttribute(
...
-          attributes.put(MEMORY_LIMIT_ATTRIBUTE, String.valueOf(ramLimit));
-        } else {
-          attributes.put(MEMORY_LIMIT_ATTRIBUTE, defaultMachineMemorySizeAttribute);
-        }
+
+      initIfEmpty(
+          () -> entry.getValue().getMemLimit(),
+          machineConfig,
+          MEMORY_LIMIT_ATTRIBUTE,
+          defaultMachineMaxMemorySizeAttribute);
```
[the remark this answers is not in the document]
i300543: yes good eye, missed it
will fix

dockerfiles/init/manifests/che.env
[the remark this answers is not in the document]
sleshchenko: Actually, I would say that it is a bit wrong comment. This file and configuration attributes are not really related to Dashboard. It is like a configuration template that shows which environment variable can be changed and what is the default value.

So, from Che Server Perspective
CHE_WORKSPACE_DEFAULT__MEMORY__LIMIT__MB - is a default value of memory limit that will be set into each machine if it is not specified by a user.

And that fact that Dashboard uses it as recommended RAM size it is just side effect.

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/compose/ComposeEnvironmentFactory.java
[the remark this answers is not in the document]
i300543: what should i do if the user requests such inconsistent requirements ? request > limit
InfrastructureException ?
i beleive k8s will not accept such command

i300543:
@sleshchenko 
ok i will update the PR with those changes

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/compose/ComposeEnvironmentFactory.java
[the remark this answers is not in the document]
i300543: just checked, k8s denies attempts to define inconsistent requirements:
image

i300543:
what should happen if the default values are inconsistent ?
che.workspace.default_memory_limit_mb < che.workspace.default_memory_request_mb
should i ignore the lesser value or throw exception and fail to create environment ?

sleshchenko:
@i300543 Good questions =)

> should i ignore the lesser value 

I think it is a quite good option if you will log warning message via Logger.

> throw exception and fail to create environment

As about this option, it is not good to deny for a user to create an environment because of admin mistake. And the user is not able to fix this issue. He is able to workaround it by specifying limits explicitly but I'm not sure that it is good UX.

Since these properties are used in Singleton components and they are created during DI container creation, I think you are able to throw an exception in components constructors to notify system admin about configuration inconsistency.

@garagatyi WDYT about throwing an exception in Factory components constructors?

i300543:
Can i move the default memory params initialization to InternalEnvironmentFactory to avoid duplicating this across all factories ?

garagatyi:
I'm OK to add this feature WITHOUT support in Docker. And the reason is that it is not mandatory, so it is up to @i300543 to add it to this PR or in another one or not implementing it at all. But since we might not have control over all the implementation of the infrastructure I DO think that we need a note in javadocs that would say that support of the attribute IS optional for the infrastructure implementation.

> @garagatyi WDYT about throwing an exception in Factory components constructors?

I would log an error in this case (not warning, but error) and set it to be equal to the limit.

i300543:
@garagatyi 
Do you mean request ? in case limit < request would not request value be more logical ?
or would you prefer to ignore the big request and set request = limit as you said ?

skabashnyuk:
ci-test

riuvshin:
ci-test build report:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/566/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/566//Selenium_20tests_20report/)
[selenium tests report data](http://tests.codenvycorp.com/selenium_test_report_data/10673)
docker image: eclipseche/che-server:10673

garagatyi:
@i300543 Yes, I would consider that limit is more important than the request. But it is just an opinion and we can set any priority we want. But I don't think that either of the variants is better than another one :)

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
```diff
@@ -94,8 +94,11 @@ che.workspace.terminal_linux_arm7=${che.home}/lib/linux_arm7/terminal
...
+che.workspace.default_memory_limit_mb=1024
+# RAM request default for each machine that has no explicit RAM settings in environment.
+# this amount will be allocated on workspace container creation
```
garagatyi: Can you add a comment that infra implementation might not respect this property?

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/config/MachineConfig.java
```diff
@@ -28,6 +28,13 @@
    */
   String MEMORY_LIMIT_ATTRIBUTE = "memoryLimitBytes";
 
+  /**
+   * Name of the attribute from {@link #getAttributes()} which if present defines memory limit of
+   * the machine in bytes. If memory limit is set in environment specific recipe this attribute used
+   * in {@code MachineConfig} should override value from recipe.
+   */
+  String MEMORY_REQUEST_ATTRIBUTE = "memoryRequestBytes";
```
garagatyi: Can you add a comment that infra implementation might not respect this attribute?

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/config/MachineConfig.java
```diff
@@ -28,6 +28,13 @@
    */
   String MEMORY_LIMIT_ATTRIBUTE = "memoryLimitBytes";
 
+  /**
+   * Name of the attribute from {@link #getAttributes()} which if present defines memory limit of
```
garagatyi: Javadocs of this attribute is the same as javadocs of `memoryLimitBytes`. Is it on purpose?

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/compose/ComposeEnvironmentFactory.java
```diff
@@ -116,14 +121,32 @@ void addRamLimitAttribute(
...
+      initIfEmpty(
+          () -> entry.getValue().getMemRequest(),
+          machineConfig,
+          MEMORY_REQUEST_ATTRIBUTE,
+          defaultMachineRequestMemorySizeAttribute);
+    }
+  }
+
+  private void initIfEmpty(
```
garagatyi: In a case when the memory limit is set in the recipe does this code respects it?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsnext/KubernetesWorkspaceNextApplier.java
```diff
@@ -15,6 +15,7 @@
 import static java.util.Collections.singletonMap;
 import static java.util.stream.Collectors.toMap;
 import static org.eclipse.che.api.core.model.workspace.config.MachineConfig.MEMORY_LIMIT_ATTRIBUTE;
+import static org.eclipse.che.api.core.model.workspace.config.MachineConfig.MEMORY_REQUEST_ATTRIBUTE;
```
garagatyi: Could you avoid changing this class? It is subject to change because we are experimenting with this stuff (it is not part of the regular user flow). So changing it will cause a lot of troubles in your PR

i300543:
Ok so i"m rewriting the setting of request and limit ,and here is what i think i might do : 
1. if (memLimit and memRequest) are both undefined => use the defaults
2. if only one of (memLimit or memRequest) is defined => both are set to  the defined value (limit=request) (this is the simplified version)
(this condition could be enhanced - 
if memLimit is defined, and default memRequest < memLimit 
=> memRequest = default memRequest, otherwise memRequest = memLimit
and vise versa: 
if memRequest is defined and default memLimit > memRequest 
=> memLimit = default memLimit, otherwise memLimit = memRequest )

3. if both memLimit and memRequest are defined:
  if they are consistent (memLimit>memRequest)
  => use them
  otherwise only use memLimit (memRequest = memLimit)

how does this sound ? should i use simple version, or the more complex one ?

garagatyi:
@i300543 Sorry, I didn't get the flow. Do you mean defined in the Che master properties or workspace recipe?

i300543:
memRequest and memLimit referenced above are values defined in workspace recipe
default memRequest , and default memLimit - are values defined in the master properties

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsnext/KubernetesWorkspaceNextApplier.java
i300543: should i rename the che.workspace.default_memory_mb property, or rely on alias to keep this code working ?

i300543:
Could someone please trigger the CI on this ?

sleshchenko:
ci-test

garagatyi (approved):
Looks good!

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/compose/ComposeEnvironmentFactory.java
```diff
@@ -116,14 +121,25 @@ void addRamLimitAttribute(
         machineConfig = new InternalMachineConfig();
         machines.put(entry.getKey(), machineConfig);
       }
+      Long memLimit = entry.getValue().getMemLimit();
+      Long memRequest = entry.getValue().getMemRequest();
+      // if both properties are not defined
+      if ((memLimit == null || memLimit <= 0) && (memRequest == null || memRequest <= 0)) {
```
sleshchenko: Should be the same be done for other environment factories?
Maybe it can be used in a separated component like
```java
class MemoryAttributesProvisioner {
  ...
   MemoryAttributesProvisioner(
      @Named("che.workspace.default_memory_limit_mb") long defaultMachineMaxMemorySizeMB,
      @Named("che.workspace.default_memory_request_mb") long defaultMachineRequestMemorySizeMB) {
    ....
    }

    void provision(MachineConfig machine, @Nullable Long memoryLimit, @Nullable Long memoryRequest) { // where memLimit and memRequest is provided by infrastructure and fetched from recipe
       // evaluate here memory settings configured or default and set correct one to machine attributes if needed
    }
}
```
@garagatyi WDYT?

i300543: you are talking about DockerImageEnvironmentFactory and DockerfileEnvironmentFactory right ?
the logic there currently only sets values if they are not defined in machineConfig.
so the values in the machine config at that point are the values requested by the user right ?
and they might not be valid correct ?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/environment/InternalEnvironmentFactory.java
```diff
@@ -42,22 +44,45 @@
  *   envFactories.addBinding("recipe_type_1").to(SubclassOfInternalEnvironmentFactory.class);
  * </pre>
  *
+ * defaultMachineMaxMemorySizeAttribute defines default machine memory limit
```
sleshchenko: I think it would be better to rewrite this doc like in the following way:
```
Implementation should add memory-related attributes into machine config if they are not specified. The values should be fetched from workspace recipe or default values from protected fields should be used.
There are two memory-related attributes: 
- memory limit - ....
- memory request - ...
```
It's not ready to copy/paste one but it shows the idea to define the contract of abstract class but not just protected fields.
WDYT?

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/compose/ComposeEnvironmentFactory.java
sleshchenko: > you are talking about DockerImageEnvironmentFactory and DockerfileEnvironmentFactory right ?

Yes, and KubernetesEnvironmentFactory and DockerfileEnvironmentFactory as well.

> the logic there currently only sets values if they are not defined in machineConfig.

Well, actually there can be a conflict situation which I described at the beginning. What if one attribute is configured and another one is missing, then default one should be used and conflict may happen. I hope the detailed explanation you can find here https://github.com/eclipse/che/pull/10673#discussion_r208473063.

> so the values in the machine config at that point are the values requested by the user right ?

right

> and they might not be valid correct?

Well, I missed this point, but it's a good one =)
It makes sense to add validation into `MachineConfigsValidator` to ensure that at that point provided values are correct.

riuvshin:
ci-test build report:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/595/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/595//Selenium_20tests_20report/)
[selenium tests report data](http://tests.codenvycorp.com/selenium_test_report_data/10673)
docker image: eclipseche/che-server:10673

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/compose/ComposeEnvironmentFactory.java
i300543: I see that on k8sEnvironmentFactory and OSEnvFactory you guys use InternalMachineConfig and it does not implement MachineConfig interface. Is there a reason for it ? 
@sleshchenko  should i pass InternalMachineConfig to the provision method in your suggestion above then ?

i300543:
Could someone please rerun ci on this ?

sleshchenko:
ci-test

riuvshin:
ci-test build report:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/623/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/623//Selenium_20tests_20report/)
[selenium tests report data](http://tests.codenvycorp.com/selenium_test_report_data/10673)
docker image: eclipseche/che-server:10673

i300543:
Fixed the startup error. 
Could someone retrigger CI please ?

sleshchenko:
ci-test

riuvshin:
ci-test build report:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/626/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/626//Selenium_20tests_20report/)
[selenium tests report data](http://tests.codenvycorp.com/selenium_test_report_data/10673)
docker image: eclipseche/che-server:10673

i300543:
Hi guys,
When you have some free time please review my latest changes:
https://github.com/eclipse/che/pull/10673/commits/01a56a961ff432ee31604f809f68cb6342eab7d7
and tell me what you think.
Thanks!

garagatyi:
@i300543 I like the changes you've made! :+1:

garagatyi:
ci-test

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/environment/MemoryAttributeProvisioner.java
```diff
@@ -0,0 +1,95 @@
...
+          "Requested default container memory limit is less than default memory request. Memory request parameter is ignored.");
+    }
+
+    this.defaultMachineMaxMemorySizeAttribute =
+        String.valueOf(defaultMachineMaxMemorySizeAttribute * 1024 * 1024);
+    this.defaultMachineRequestMemorySizeAttribute =
+        String.valueOf(defaultMachineRequestMemorySizeAttribute * 1024 * 1024);
+  }
+
+  public void provision(
```
sleshchenko: It makes sense to add java docs here and describe what is memoryLimit and memoryRequest and when they can be null or 0

sleshchenko (approved):
Well Done :+1:

garagatyi (approved):
@i300543 great job! And I like simplifications you've recently added!

garagatyi:
ci-test

garagatyi:
@skabashnyuk Do you want to do a review one more time or you are OK to go with a merge?

skabashnyuk:
@garagatyi I'm ok with pr, I'm ok to merge it. What about docs and release notes?

garagatyi:
@eivantsov can you help us to figure out what needs to be changed in docs and release notes and who is responsible for that in the PR review process?

i300543:
Pinging ,
Anything i can do about this ?

eivantsov:
@i300543 I think you may add a few things here https://github.com/eclipse/che-docs/blob/master/src/main/pages/setup-openshift/kubernetes-admin-guide.adoc

i300543:
anything else i can do to merge this into master ?

garagatyi:
@i300543 I'm sorry that the process of PR review is that long and complex. I think that we can postpone the release notes task since it is not mandatory to merge the PR. 
@dneary can you explain what is the process for a contributor? Does he have to contribute to the release notes? If yes, then has he do that before, after the merge?

garagatyi:
@skabashnyuk I think that it is OK to merge this PR. Are you OK with that?

skabashnyuk:
@garagatyi ok for me.

Issue #10620, opened 2018-08-01
Enable DELETE  /api/user api method


skabashnyuk:
### Description
Previously we thought we don't need this method.
Looks like it can be useful to delete all user-specific data.
The scope of this issue is to remove DELETE method from UnavailableResourceInMultiUserFilter binding

PR #10676, opened 2018-08-06
Added an ability to remove user when third-party user management service is configured

sleshchenko:
### What does this PR do?
Adds an ability to remove user when third-party user management service is configured.

Note that removing user via Che won't remove user on third-party user management service (like keycloak) and removed user still will be able to use Che with the cached token or after next login (user entity will be recreated).
Also, it is impossible to remove a user if he has running any workspaces. It is needed to stop all the workspaces before a user removing.

Also, it contains a fix non-related to issue:
Fix binding of UnavailableResourceInMultiUserFilter for ProfileService's methods. 
Previously, it banned only update profile attributes of the specified user (PUT /profile/{USER_ID}/attributes) but methods related to a current user were still available (DELETE /profile/attributes and PUT /profile/attributes)

https://github.com/eclipse/che/issues/10620

#### Docs PR
N/A

### DIFF
--- a/multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/deploy/KeycloakServletModule.java
+++ b/multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/deploy/KeycloakServletModule.java
@@ -36,6 +36,6 @@ public class KeycloakServletModule extends ServletModule {
     // but not remove user (DELETE /user/{USER_ID}
     filterRegex("^/user(/password/?|/)?$").through(UnavailableResourceInMultiUserFilter.class);
 
-    filterRegex("/profile/.*/attributes").through(UnavailableResourceInMultiUserFilter.class);
+    filterRegex("^/profile/(.*/)?attributes$").through(UnavailableResourceInMultiUserFilter.class);
   }
 }
--- a/multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/deploy/KeycloakServletModule.java
+++ b/multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/deploy/KeycloakServletModule.java
@@ -32,7 +32,10 @@ public class KeycloakServletModule extends ServletModule {
     filterRegex("^(?!.*(/docs/))(?!.*(/keycloak/settings/?|/api/oauth/callback/?)$).*")
         .through(IdentityIdLoggerFilter.class);
 
-    filterRegex("/user/?.*").through(UnavailableResourceInMultiUserFilter.class);
+    // Ban change password (POST /user/password) and create a user (POST /user/) methods
+    // but not remove user (DELETE /user/{USER_ID}
+    filterRegex("^/user(/password/?|/)?$").through(UnavailableResourceInMultiUserFilter.class);
+
     filterRegex("/profile/.*/attributes").through(UnavailableResourceInMultiUserFilter.class);
   }
 }


multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/deploy/KeycloakServletModule.java
```diff
@@ -32,7 +32,10 @@ protected void configureServlets() {
     filterRegex("^(?!.*(/docs/))(?!.*(/keycloak/settings/?|/api/oauth/callback/?)$).*")
         .through(IdentityIdLoggerFilter.class);
 
-    filterRegex("/user/?.*").through(UnavailableResourceInMultiUserFilter.class);
+    // Ban change password (POST /user/password) and create a user (POST /user/) methods
+    // but not remove user (DELETE /user/{USER_ID}
+    filterRegex("/user(/password|/)?").through(UnavailableResourceInMultiUserFilter.class);
```
mshaposhnik: Isn't pattern should be `/user(/password|/)?$` ? (add extra $ a the end)

mshaposhnik: Even more: `/user(/password/?|/)?$`

sleshchenko: I don't think that it will change a lot, but I agree that it makes sense to use `^` and `$` in regular expressions. So, do you think the following changes are OK
```
filterRegex("^/user(/password/?|/)?$")....
filterRegex("^/profile/.*/attributes$")....
```
?

mshaposhnik: If you didn't indicate end of string,  it matches `/user/anything`  then. So delete won't work.

skabashnyuk (approved):

multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/deploy/KeycloakServletModule.java
mshaposhnik: These are ok

sleshchenko: @mshaposhnik Your words make sense but I tested this PR and it works. So, it doesn't match `/user/anything`. And specifying `.*` the only way to make it work in this way.
let me double check this.

mshaposhnik (approved):

multiuser/keycloak/che-multiuser-keycloak-server/src/main/java/org/eclipse/che/multiuser/keycloak/server/deploy/KeycloakServletModule.java
sleshchenko: So, thanks for pointing to `/password/?`
As about `^` and `$` they are really optional here since `com.google.inject.servlet.RegexUriPatternMatcher` uses `Matcher#matches` that 
```
Returns <tt>true</tt> if, and only if, the entire region sequence matches this matcher's pattern
```
while `/user/anything` contains region that matches to `/user(/password/?|/)?` but it is not that one which matches to the specified pattern

I'm OK to specify `^` and `$` explicitly.

sleshchenko:
ci-test

riuvshin (approved):

riuvshin:
ci-test build report:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/556/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/556//Selenium_tests_report)
[selenium tests report data](http://tests.codenvycorp.com/selenium_test_report_data/10676)
docker image: eclipseche/che-server:10676

Issue #10687, opened 2018-08-07
Workspace loader hangs up when a user doesn't have permissions to use a workspace


sleshchenko:
### Description
Workspace loader hangs up when a user doesn't have permissions to use a workspace.

### Reproduction Steps
1. Log in as user A, and create workspace 1.
2. Grant user B with `read` permission for workspace 1.
3. Log in as user B.
4. Try to start the workspace.
Expected: The error is shown that the user doesn't have permissions to start the workspace.
Actual: Workspace loader hangs up.
<details>
<summary>Screenshot</summary>

![screenshot_20180807_133147](https://user-images.githubusercontent.com/5887312/43770894-48eb4606-9a46-11e8-9c89-d3ee512dd21d.png)
</details>

Another use case when a user tries to open started workspace:
1. Log in as user A, and create workspace 1.
2. Grant user B with `read` permission for workspace 1.
3. Start workspace 1.
4. Log in as user B.
5. Try to open the workspace.
Expected: The error is shown that the user doesn't have permissions to use the workspace.
Actual: Workspace loader hangs up.
<details>
<summary>Screenshot</summary>

![screenshot_20180807_133329](https://user-images.githubusercontent.com/5887312/43770974-804f2dec-9a46-11e8-926a-3176b78d2d64.png)
</details>

Note that now it is actual only in a branch where workspace permissions issue is fixed https://github.com/eclipse/che/pull/10684.

**OS and version:**    

**Diagnostics:**

sleshchenko:
https://github.com/eclipse/che/pull/10688 and https://github.com/eclipse/che/pull/10642 partially fix the issue but there are still two more issues:
1. WsLoader doesn't handle empty runtime when a workspace is already `RUNNING`. It checks only when workspace status changed event received.

<details>
<summary>Screenshot</summary>

![screenshot_20180807_133329](https://user-images.githubusercontent.com/5887312/43770974-804f2dec-9a46-11e8-926a-3176b78d2d64.png)
</details>

2. It would be better to improve an error message when a user doesn't have permissions to start a workspace. Predefined error message and error message from response are merged without any punctuation mark

<details>
<summary>Screenshot</summary>

![screenshot_20180808_153659](https://user-images.githubusercontent.com/5887312/43842065-328e0e38-9b2d-11e8-87b5-5d4f011d24b2.png)
</details>

PR #10642, opened 2018-08-03
CHE-10538: improves workspace loader

akurinnoy:
### What does this PR do?
When workspace loader starts a workspace and it fails, then an error appears and the progress bar hides.

restart

### What issues does this PR fix or reference?
fix #10538

#### Release Notes

N/A - bugfix

#### Docs PR

N/A - bugfix

PR #10688, opened 2018-08-07
CHE-10540: stop progress on websocket error

### LINKED ISSUE #10540: Make workspace-loader (or UD) show an error received from websocket on workspace start failure
### Description
When workspace start fails after some time workspace-loader and UD receive the message with the error, but provide UX that might be not clear to some users:
- There is a red circle near the workspace name
- There is no error shown
- The workspace loading progress bar doesn't disappear that might make a user think that workspace is still starting  

Instead of that, it would be nice to:
- stop showing the progress bar
- show a popup with the error message and a button to restart the workspace (the same as reloading the page) 

### Reproduction Steps

**OS and version:**    

**Diagnostics:**

akurinnoy:
### What does this PR do?
When workspace loader starts a workspace and receives an error on websocket, then error appears and the progress bar hides.
restart2

fix #10540 

#### Release Notes

N/A - bugfix

#### Docs PR

N/A - bugfix

### DIFF
--- a/workspace-loader/src/index.ts
+++ b/workspace-loader/src/index.ts
@@ -187,7 +187,7 @@ export class WorkspaceLoader {
                     if (xhr.readyState !== 4) { return; }
                     if (xhr.status !== 200) {
                         const errorMessage = 'Failed to get the workspace' + this.getRequestErrorMessage(xhr);
-                        reject(errorMessage);
+                        reject(new Error(errorMessage));
                         return;
                     }
                     resolve(JSON.parse(xhr.responseText));
@@ -209,7 +209,7 @@ export class WorkspaceLoader {
                     if (xhr.readyState !== 4) { return; }
                     if (xhr.status !== 200) {
                         const errorMessage = 'Failed to start the workspace'  + this.getRequestErrorMessage(xhr);
-                        reject(errorMessage);
+                        reject(new Error(errorMessage));
                         return;
                     }
                     resolve(JSON.parse(xhr.responseText));
@@ -258,13 +258,9 @@ export class WorkspaceLoader {
 
         const runningOnConnectionPromise = masterApiConnectionPromise
             .then((masterApi: CheJsonRpcMasterApi) => {
-                return new Promise((resolve) => {
+                return new Promise((resolve, reject) => {
                     masterApi.addListener('open', () => {
-                        this.getWorkspace(this.workspace.id).then((workspace) => {
-                            if (workspace.status === 'RUNNING') {
-                                resolve();
-                            }
-                        });
+                        this.checkWorkspaceRuntime().then(resolve, reject);
                     });
                 });
             });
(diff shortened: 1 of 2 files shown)


workspace-loader/src/index.ts
```diff
@@ -225,33 +208,34 @@ export class WorkspaceLoader {
...
-            return Promise.reject(errorMessage);
         });
     }
 
+    getRequestErrorMessage(xhr: XMLHttpRequest): string {
+        let errorMessage;
+        try {
+            const response = JSON.parse(xhr.responseText);
+            errorMessage = response.message;
+        } catch (e) { }
```
benoitf, line 226: should we log error message/response ?

workspace-loader/src/index.ts
```diff
@@ -322,7 +306,7 @@ export class WorkspaceLoader {
             masterApi.subscribeWorkspaceStatus(this.workspace.id,
                 (message: any) => {
                     if (message.error) {
-                        this.loader.error(message.error);
+                        reject(`Failed to run the workspace: "${message.error}"`);
```
benoitf: should we reject with new Error() ?

benoitf (approved):

workspace-loader/src/index.ts
akurinnoy: It is eventually logged here https://github.com/eclipse/che/blob/389e259d7672fe57300f10ad19657076ea7f19f9/workspace-loader/src/index.ts#L145-L146

Here I only tried to get a meaningful error message to reject a workspace getting or starting promise.

workspace-loader/src/index.ts
```diff
@@ -322,7 +306,7 @@ export class WorkspaceLoader {
...
                     } else if (message.status === 'RUNNING') {
                         resolve();
```
ashumilova: Can we, please, also check "runtime" is there (not null), when we have RUNNING status. If not - reject the promise. 
There is usecase, when user tries to open shared workspace, where he has no permissions on runtime.

workspace-loader/src/index.ts
akurinnoy: Yes, it may be useful, I'll fix this.

workspace-loader/src/index.ts
akurinnoy: sure, I'll add a fixup

sleshchenko: More details here https://github.com/eclipse/che/issues/10687
Maybe we can also handle second use case here when POST /workspace/WS_ID/runtime is failed

akurinnoy: That POST is handled here
https://github.com/eclipse/che/blob/389e259d7672fe57300f10ad19657076ea7f19f9/workspace-loader/src/index.ts#L211-L212

sleshchenko: Cool. I missed it because PR title and even description tell nothing about this change.

Looks like we will be able to close that issue after merge. I can recheck after merge.

akurinnoy: It was another PR https://github.com/eclipse/che/pull/10642 which is already merged.

sleshchenko: I see, thanks. 
I did find the given lines in master and you pointed commit from this PR, it's why I thought that it is introduced in this PR =)

olexii4 (approved):

ashumilova (approved):

vitaliy-guliy (approved):

PR #10690, opened 2018-08-07
Rework GitStatusProvider to use cached status storage

vinokurig:
### What does this PR do?
JGit status command on big projects (Eclipse Che) uses a lot of RAM.
screenshot from 2018-08-22 09-54-27_

When deep expand to some folder is performed, Git status is called on each folder expand:
screenshot-172 17 0 1-8080-2018 08 22-10-24-12

This causes a huge leap of RAM and finally OOM appears:
screenshot from 2018-08-22 09-58-43

In this PR `GitStatusProvider` is reworked to call Git status only if it is really needed e.g. file was modified. The provider is based on a cached Git Status Map that is updated from Git Events.

#10696

#### Release Notes

Fix `java.lang.OutOfMemoryError` on large projects.

#### Docs PR

N/A


wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitStatusProvider.java
```diff
@@ -1,119 +0,0 @@
-/*
```
vparfonov: Don't remove this class for now, just mark as Deprecated

wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/CachedGitStatusProvider.java
```diff
@@ -0,0 +1,290 @@
...
+    this.rootDirPathProvider = rootDirPathProvider;
+    this.eventService = eventService;
+    this.statusCache = new HashMap<>();
+    this.projectFiles = new HashMap<>();
+
+    subscribeToEvents();
+    collectProjectFiles(rootDirPathProvider.get());
+  }
+
+  private void collectProjectFiles(String root) {
```
vparfonov, line 98: I propose add @PostConstruct and remove call this method from constructor

vinokurig: Done

wsagent/che-core-api-git/src/main/java/org/eclipse/che/api/git/GitStatusProvider.java
vinokurig: Fixed

vparfonov (approved):

vinokurig:
ci-test

musienko-maxim (approved):

Issue #10764, opened 2018-08-14
Importing Simple Java Projects Broken in jdt.ls Selenium Tests


tsmaeder:
In org.eclipse.che.selenium.core.client.TestProjectServiceClient.importProject(String, Path, String, String), we write .classpath and .project files to the test resource location (i.e. where the tests class file is located). This is timing dependent when tests are run in parallel, because the resources are shared between different tests.

tsmaeder:
The above fix broke the path projects are imported as. Need to fix.

tsmaeder:
The problem was with handling root directories for zips: when adding the contents of a directory, all zip entries would be prefixed by the directory name, e.g. /foo/bar.txt would be "foo/bar.txt" in the zip, when it should have been just "bar.txt"

PR #10785, opened 2018-08-15
Fix parent path handling when importing test projects

tsmaeder:
### What does this PR do?
This is a fixup for breaking test project import via zipping directories. Streamlined adding files to the zip a bit in the process.

#10764

### DIFF
--- a/core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/ZipUtils.java
+++ b/core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/ZipUtils.java
@@ -1,13 +1,11 @@
 /*
- * Copyright (c) 2012-2018 Red Hat, Inc.
- * This program and the accompanying materials are made
- * available under the terms of the Eclipse Public License 2.0
- * which is available at https://www.eclipse.org/legal/epl-2.0/
+ * Copyright (c) 2012-2018 Red Hat, Inc. This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0 which is available at
+ * https://www.eclipse.org/legal/epl-2.0/
  *
  * SPDX-License-Identifier: EPL-2.0
  *
- * Contributors:
- *   Red Hat, Inc. - initial API and implementation
+ * Contributors: Red Hat, Inc. - initial API and implementation
  */
 package org.eclipse.che.commons.lang;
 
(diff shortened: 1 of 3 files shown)


core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/ZipUtils.java
```diff
@@ -38,91 +37,88 @@
...
+                } catch (IOException e) {
+                  throw new ExceptionWrapper(e);
+                }
+              });
+    } catch (ExceptionWrapper e) {
+      throw (IOException) e.getCause();
     }
   }
 
+  private static String relativePath(Path root, Path f) {
```
tolusha, line 120: AFAIK these changes won't be approved for the master branch.

tolusha: They just don't accept refactoring in those components.

tolusha: @skabashnyuk Am I right?

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/ZipUtils.java
```diff
@@ -1,27 +1,26 @@
 /*
- * Copyright (c) 2012-2018 Red Hat, Inc.
- * This program and the accompanying materials are made
- * available under the terms of the Eclipse Public License 2.0
- * which is available at https://www.eclipse.org/legal/epl-2.0/
+ * Copyright (c) 2012-2018 Red Hat, Inc. This program and the accompanying materials are made
```
skabashnyuk: I believe this changes in the license are made by mistake.

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/ZipUtils.java
skabashnyuk: I'm ok if unit and selenium test will be ok.

skabashnyuk (approved):
In general, I'm ok with this PR. Maybe it can be ported in master first?

tolusha (approved):

svor (approved):

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/ZipUtils.java
tsmaeder: Hmh...I might have had "skip license check" flag on when building this branch. Good catch, I'll fix it.

PR #10801, opened 2018-08-16
CHE_INFRASTRUCTURE and CHE_MULTIUSER vars can be now passed as system

rhopp:
### What does this PR do?
Before this PR, properties `CHE_MULTIUSER` and `CHE_INFRASTRUCTURE` had to be set as environmental variables. After this PR is merged, it will be possible to set them as system variables.

In other words - before this PR, it was not possible to start tests with `-Dche.infrastructure=openshift` for example... After this PR this should be possible.

From my testing, this PR should not break any flow (a.k.a. setting env variable `CHE_INFRASTRUCTURE` will be enough)

### DIFF
--- a/selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/configuration/SeleniumTestConfiguration.java
+++ b/selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/configuration/SeleniumTestConfiguration.java
@@ -128,12 +128,6 @@ public class SeleniumTestConfiguration extends InMemoryTestConfiguration {
       name = name.replace('_', '.');
       name = name.replace("=", "_");
 
-      // convert value of CHE_INFRASTRUCTURE to upper case to comply with Infrastructure
-      // enumeration;
-      if (name.equals("che.infrastructure")) {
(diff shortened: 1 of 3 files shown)


riuvshin:

riuvshin:

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/configuration/SeleniumTestConfiguration.java
```diff
@@ -128,12 +128,6 @@ public boolean test(Map.Entry<K, V> entry) {
       name = name.replace('_', '.');
       name = name.replace("=", "_");
 
-      // convert value of CHE_INFRASTRUCTURE to upper case to comply with Infrastructure
```
rhopp, line 131: As now, the `che.infrastructure` can be defined also as system property, I had to move this logic somewhere else (`InMemoryTestConfiguration`). If the new place is not the best one, please let me know.

dmytro-ndp (approved):
Well, it was impossible to run selenium tests on CI because of problem with running Eclipse Che on OCP, but locally they work as expected.

PR #10509, opened 2018-07-23
Make host calculation more short and concise for che master

i300543:
Signed-off-by: Sergey Kuperman <sergey.kuperman@sap.com>

### What does this PR do?
Shortens the che master hostname calculated in k8s helm scripts for single-host and default-host server strategies,
makes tls secret used in all ingresses configurable from values

PR #10802, opened 2018-08-16
Fix redirect URIs and webOrigins

ghost:
### What does this PR do?

Fixes regression introduced by https://github.com/eclipse/che/pull/10509/files

For some reason, sed commands was modified and as a result keycloak script never substitutes placeholders with actual urls.

### DIFF
--- a/dockerfiles/keycloak/kc_realm_user.sh
+++ b/dockerfiles/keycloak/kc_realm_user.sh
@@ -17,8 +17,11 @@ if [ "${CHE_KEYCLOAK_ADMIN_REQUIRE_UPDATE_PASSWORD}" == "false" ]; then
     sed -i -e "s#\"UPDATE_PASSWORD\"##" /scripts/che-users-0.json
 fi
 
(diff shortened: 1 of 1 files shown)


dockerfiles/keycloak/kc_realm_user.sh
```diff
@@ -18,7 +18,7 @@ if [ "${CHE_KEYCLOAK_ADMIN_REQUIRE_UPDATE_PASSWORD}" == "false" ]; then
 fi
 
 cat /scripts/che-realm.json.erb | \
-                                sed -e "s@<%= scope\.lookupvar('che::che_server_url') %>@${PROTOCOL}://${CHE_HOST}@" \
+				sed -e "s@<%= scope\.lookupvar('che::che_server_url') %>@${PROTOCOL}://che-${NAMESPACE}.${ROUTING_SUFFIX}@" \
```
sleshchenko: Please make sure that Che on kubernetes won't be broken after reverting changes introduced in https://github.com/eclipse/che/pull/10509/files#diff-48382692058266aaf66d0fc1105bb4bbR21

sleshchenko (approved):
Now it looks good to me :+1:

Issue #9148, opened 2018-03-20
A workspace item in an organization workspaces list is not clickable


SkorikSergey:
**Reproduction Steps:**

1. Create an organization.
2. Add a workspace to this organization.
3. Open the **Workspace** tab in organization details page.
4. Click on the workspace name in list to open its workspace details.

**Expected result:** workspace details page is opened.

**Actual result:** workspace item in list is not clickable.

**OS and version:**
Che 6.3.0-SNAPSHOT; Chrome 63.0; Fedora 25

**Failed test**: AddWorkspaceToOrganizationTest

**Additional information:**
![selection_218](https://user-images.githubusercontent.com/7760565/37658774-f6e44dd8-2c56-11e8-9e52-56613f8c873f.png)

![nimbus-record-video-2018-03-20-15-41-10](https://user-images.githubusercontent.com/7760565/37658523-673e6f2e-2c56-11e8-958e-51dafee11956.gif)

vkuznyetsov:
Still actual in Version: 6.10.0

PR #10814, opened 2018-08-17
Make clickable the whole workspace item row in workspaces list of an organization

akurinnoy:
### What does this PR do?
This PR adds a fix to make clickable the whole workspace item row in workspaces list of an organization. 
Excluding action buttons, previously clickable was only workspace name.

https://github.com/eclipse/che/issues/9148


sleshchenko:
Please fix PR title since it will be used for changes log and the current one is not self-explaining at all.

akurinnoy:
@sleshchenko Thanks, fixed

ashumilova (approved):

SkorikSergey (approved):

PR #10820, opened 2018-08-17
CHE-10769: improve loader.html page

### LINKED ISSUE #10769: Improve loader.html to be more user-friendly
### Description
There are few user-unfriendly things which could be improved in loader.html:
1. `body` of page is always hardcoded to `Loading a workspace...` but actually, it's not true, it makes sense to show simple loader there or print actual step.
2. When any exceptions occur then alert will be shown with the hardcoded message `alert('Error. Check logs in browser\'s console.');`. It would be much useful to show an original message for a user. Also, maybe there is a more elegant way to show an exception to the user instead of showing an alert.

### Reproduction Steps
1. Open `loader.html` without workspace id neither redirect URL parameter.
Actual: Alert with an error message is shown twice. After hiding an alert there is a blank page with `Loading a workspace...` content when there is no any action in background.

akurinnoy:
### What does this PR do?
This PR updates the `loader.html` page in the way it looks similar to workspace loader application:

**initial state**
selection_005

**after an error**
selection_003

fix #10769 

Signed-off-by: Oleksii Kurinnyi <okurinny@redhat.com>


sleshchenko (approved):
LGTM

The only thing that I want to note: please check how it looks like with real secure server to see whether the page is blink because of quick redirects, if yes maybe it makes sense to use white color for background

ashumilova (approved):

olexii4 (approved):

sleshchenko:
@akurinnoy I've checked that and here is how it works in normal case:
check

But if workspace id of a non-existent workspace is specified then `Error` is the only message that it shows:
screenshot_20180820_093944
Could you take a look on that?

akurinnoy:
@sleshchenko Thanks for the feedback, I'll fix it ASAP

akurinnoy:
@sleshchenko fixed!

sleshchenko (approved):
Well done

Issue #10115, opened 2018-06-20
New folders do not register in the jdt.ls workspace


svor:
### Description
New created folder from java project won't be registered in jdt.ls. 
Looks like file watcher doesn't send an event to jdt.ls e about creation of the folder.
### Steps to reproduce:
1. Create Plain java project;
2. Create new folder in the projects and mark it as source folder
3. Try to move some class to this folder by using Move Refactoring wizard 
4. In the Move wizard we don't see new source folder among possible directions 

It happens because jdt.ls checks if each element from the classpath exists on the file system but jdt.ls knows nothing about creation new folder.

_Selenium test which shows this problem:_ `org.eclipse.che.selenium.refactor.move.MoveJavaFileInNewSourceFolderTest`

PR #10893, opened 2018-08-23
New folders do not register in the jdt.ls workspace #10115

vrubezhny:
This adds the reporting creation/modification/deletion of source folders to jdt.ls

### What does this PR do?

Fix adds a listener for changes inside source folders (Create/Update/Delete) in order to report them to jds.ls.

ls10115-part2

Fixes eclipse/che#10115
Depends on eclipse/che-ls-jdt#70
Depends on eclipse/eclipse.jdt.ls#755

### DIFF
--- a/plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-server/pom.xml
+++ b/plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-server/pom.xml
@@ -39,6 +39,14 @@
             <artifactId>commons-io</artifactId>
         </dependency>
         <dependency>
+            <groupId>javax.annotation</groupId>
+            <artifactId>javax.annotation-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>javax.inject</groupId>
+            <artifactId>javax.inject</artifactId>
+        </dependency>
+        <dependency>
             <groupId>javax.ws.rs</groupId>
             <artifactId>javax.ws.rs-api</artifactId>
         </dependency>
@@ -85,6 +93,10 @@
         <dependency>
             <groupId>org.eclipse.che.plugin</groupId>
             <artifactId>che-plugin-java-server</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.lsp4j</groupId>
+            <artifactId>org.eclipse.lsp4j</artifactId>
         </dependency>
         <dependency>
             <groupId>org.slf4j</groupId>
--- a/plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-server/src/main/java/org/eclipse/che/plugin/java/plain/server/inject/PlainJavaProjectSourceFolderWatcher.java
+++ b/plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-server/src/main/java/org/eclipse/che/plugin/java/plain/server/inject/PlainJavaProjectSourceFolderWatcher.java
@@ -1,0 +1,135 @@
+/*
+ * Copyright (c) 2012-2018 Red Hat, Inc.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.plugin.java.plain.server.inject;
+
+import static java.nio.file.Files.isDirectory;
+import static java.util.Collections.singletonList;
+import static org.eclipse.che.api.languageserver.LanguageServiceUtils.prefixURI;
+import static org.eclipse.che.api.languageserver.LanguageServiceUtils.removeUriScheme;
+import static org.eclipse.che.jdt.ls.extension.api.Commands.GET_PROJECT_SOURCE_LOCATIONS_COMMAND;
+
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.reflect.TypeToken;
+import java.lang.reflect.Type;
+import java.nio.file.PathMatcher;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
+import javax.inject.Inject;
+import org.eclipse.che.api.core.notification.EventService;
+import org.eclipse.che.api.languageserver.ExtendedLanguageServer;
+import org.eclipse.che.api.languageserver.FindServer;
+import org.eclipse.che.api.project.server.notification.ProjectUpdatedEvent;
+import org.eclipse.che.api.watcher.server.FileWatcherManager;
+import org.eclipse.che.api.watcher.server.impl.FileWatcherByPathMatcher;
+import org.eclipse.che.plugin.java.inject.JavaModule;
+import org.eclipse.lsp4j.DidChangeWatchedFilesParams;
+import org.eclipse.lsp4j.ExecuteCommandParams;
+import org.eclipse.lsp4j.FileChangeType;
+import org.eclipse.lsp4j.FileEvent;
+import org.eclipse.lsp4j.services.LanguageServer;
+
+/**
+ * Reports the create/update/delete changes on project source folders to jdt.ls
+ *
+ * @author V. Rubezhny
+ */
+public class PlainJavaProjectSourceFolderWatcher {
+  private static final Gson gson =
+      new GsonBuilder().disableHtmlEscaping().serializeNulls().create();
+
+  private final FileWatcherManager manager;
+  private final FileWatcherByPathMatcher matcher;
+  private final FindServer lsRegistry;
+
+  private final EventService eventService;
+
+  private final CopyOnWriteArrayList<Integer> watcherIds = new CopyOnWriteArrayList<>();
+
+  @Inject
+  public PlainJavaProjectSourceFolderWatcher(
+      FileWatcherManager manager,
+      FileWatcherByPathMatcher matcher,
+      FindServer lsRegistry,
+      EventService eventService) {
+    this.manager = manager;
+    this.matcher = matcher;
+    this.lsRegistry = lsRegistry;
+    this.eventService = eventService;
+  }
+
+  @PostConstruct
+  protected void startWatchers() {
+    int watcherId =
+        manager.registerByMatcher(
+            folderMatcher(),
+            s -> report(s, FileChangeType.Created),
+            s -> {},
+            s -> report(s, FileChangeType.Deleted));
+
+    watcherIds.add(watcherId);
+    eventService.subscribe(this::onProjectUpdated, ProjectUpdatedEvent.class);
+  }
+
+  @PreDestroy
+  public void stopWatchers() {
+    watcherIds.stream().forEach(id -> manager.unRegisterByMatcher(id));
+  }
+
+  private void onProjectUpdated(ProjectUpdatedEvent event) {
+    ExecuteCommandParams params =
+        new ExecuteCommandParams(
+            GET_PROJECT_SOURCE_LOCATIONS_COMMAND, singletonList(prefixURI(event.getProjectPath())));
+
+    ExtendedLanguageServer languageServer = lsRegistry.byId(JavaModule.LS_ID);
+    if (languageServer == null) {
+      return;
+    }
+
+    languageServer
+        .getServer()
+        .getWorkspaceService()
+        .executeCommand(params)
+        .thenAccept(
+            result -> {
+              if (result == null) {
+                return;
+              }
+              Type type = new TypeToken<ArrayList<String>>() {}.getType();
+              List<String> paths = gson.fromJson(gson.toJson(result), type);
+              paths.stream().forEach(f -> matcher.accept(Paths.get(removeUriScheme(prefixURI(f)))));
+            });
+  }
+
+  private PathMatcher folderMatcher() {
+    return it -> isDirectory(it);
+  }
+
+  private void report(String path, FileChangeType changeType) {
+    ExtendedLanguageServer languageServer = lsRegistry.byId(JavaModule.LS_ID);
+    if (languageServer != null) {
+      send(languageServer.getServer(), path, changeType);
+    }
+  }
+
+  private void send(LanguageServer server, String path, FileChangeType changeType) {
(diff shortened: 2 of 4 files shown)


plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/JavaLanguageServerExtensionService.java
```diff
@@ -456,6 +457,20 @@ private ClasspathEntry fixEntry(ClasspathEntry e) {
     return result.stream().map(LanguageServiceUtils::removePrefixUri).collect(Collectors.toList());
   }
 
+  /**
```
tsmaeder: JavaLanguageServerExtensionService is made for communication with the browser. That's one reason why the methods are synchronous. This method here is NOT for that purpose, so should be somewhere else.

plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-server/src/main/java/org/eclipse/che/plugin/java/plain/server/inject/PlainJavaProjectSourceFolderWatcher.java
```diff
@@ -0,0 +1,197 @@
...
+            .stream()
+            .filter(p -> Paths.get(wsPath).startsWith(Paths.get(p)))
+            .count()
+        > 0);
+  }
+
+  private void setInitialPathsToWatch(String projectUri) {
+    List<String> sourceFolderLocations;
+    try {
+      sourceFolderLocations = extensionService.getAllSourceFoldersLocations(projectUri);
```
tsmaeder: We should not use a synchronous call here, rather use the promise we get from "WorkspaceService.executeCommand()". Then we can get rid of the executor service.

plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-server/src/main/java/org/eclipse/che/plugin/java/plain/server/inject/PlainJavaProjectSourceFolderWatcher.java
```diff
@@ -0,0 +1,197 @@
...
+    eventService.subscribe(this::onServerInitialized, LanguageServerInitializedEvent.class);
+    eventService.subscribe(this::onProjectUpdated, ProjectUpdatedEvent.class);
+  }
+
+  @PreDestroy
+  public void stopWatchers() {
+    watcherIds.stream().forEach(id -> manager.unRegisterByMatcher(id));
+  }
+
+  private void onServerInitialized(LanguageServerInitializedEvent event) {
```
tsmaeder: No need to collect all language servers here. We only want to notify the java language server, and we can find that one by id.

plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-server/src/main/java/org/eclipse/che/plugin/java/plain/server/inject/PlainJavaProjectSourceFolderWatcher.java
```diff
@@ -0,0 +1,197 @@
...
+  private PathMatcher folderMatcher() {
+    return it -> isSourceFolder(it);
+  }
+
+  private void report(String path, FileChangeType changeType) {
+    languageServers.stream().forEach(ls -> send(ls, path, changeType));
+  }
+
+  private void send(LanguageServer server, String path, FileChangeType changeType) {
+    RegisteredProject project = projectManager.getClosestOrNull(path);
```
tsmaeder: Why is this check necessary?

plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-server/src/main/java/org/eclipse/che/plugin/java/plain/server/inject/PlainJavaProjectSourceFolderWatcher.java
```diff
@@ -0,0 +1,197 @@
...
+
+  private void onProjectUpdated(ProjectUpdatedEvent event) {
+    executorService.submit(
+        () -> {
+          setInitialPathsToWatch(prefixURI(event.getProjectPath()));
+        });
+  }
+
+  private PathMatcher folderMatcher() {
+    return it -> isSourceFolder(it);
```
tsmaeder: This is causing us to communicate with jdt.ls for every folder change in the file system. I think it would be more efficient to send every folder change to jdt.ls and let jdt.ls do the filtering.
Also, are we interested in every change or only in create/delete?

tsmaeder (changes requested):

plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-server/pom.xml
```diff
@@ -38,6 +42,14 @@
             <groupId>commons-io</groupId>
             <artifactId>commons-io</artifactId>
         </dependency>
+				<dependency>
```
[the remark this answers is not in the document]
vrubezhny: 1. I don't have those tabs in my code... Looks like they were there for some time, but after I fixed them locally git ignores the fix. I'll try to fix
2. Probably because of:
```
+import javax.annotation.PreDestroy;
```
in `PlainJavaProjectSourceFolderWatcher`

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/refactor/move/MoveJavaFileInNewSourceFolderTest.java
```diff
@@ -90,6 +94,10 @@ public void checkMoveJavaClassInNewSourceFolder() {
     // move java file into new source folder
     projectExplorer.waitItem(PROJECT_NAME);
     projectExplorer.waitAndSelectItem(PATH_TO_FILE);
+
+    // As FileTreeWalker's period is 10 secs - wait for the folders to be finally reported to jdt.ls
+    WaitUtils.sleepQuietly(11);
```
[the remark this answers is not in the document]
vrubezhny: Right answer is `yes, we're waiting for a condition` but, during the test, the content of 'Move'-dialog's tree is not updated. This `sleepQuitely()` call allows all the required changes to happen before the dialog is invoked.

plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/JavaLanguageServerExtensionService.java
```diff
@@ -456,6 +457,20 @@ private ClasspathEntry fixEntry(ClasspathEntry e) {
...
+   * Gets all source folder locations of plain java project.
+   *
+   * @param projectPath project path
+   * @return source folder locations
+   */
+  public List<String> getAllSourceFoldersLocations(String projectPath) {
```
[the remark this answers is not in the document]
vrubezhny: Should read as "every location in source folder"...
If we have a project tree like:
```
Project
  +-- SourceFolder
       +-- org.pack.age
       +-- org.pack.house
```
With this method I'm going to get the list consisting of the following paths:
```
/projects/Project/SourceFolder/org/pack/age
/projects/Project/SourceFolder/org/pack/house
```
in order to make them accepted by `FileWatcherByPathMatcher`. 
If I don't do this, any file/folder that is not created during the working session (any existing path of a project) is never being reported as `Updated` or `Deleted` and, as such, will never be refreshed on jdt.ls workspace. As result, every such path (package) deleted by user, despite of its deletions, will appear in `Move`-dialog as an existing package.

plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-server/src/main/java/org/eclipse/che/plugin/java/plain/server/inject/PlainJavaProjectSourceFolderWatcher.java
vrubezhny: Not cleaned up code. sorry.

plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-server/src/main/java/org/eclipse/che/plugin/java/plain/server/inject/PlainJavaProjectSourceFolderWatcher.java
vrubezhny: OK

plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/JavaLanguageServerExtensionService.java
vrubezhny: Got it

plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-server/src/main/java/org/eclipse/che/plugin/java/plain/server/inject/PlainJavaProjectSourceFolderWatcher.java
vrubezhny: Re-implemented

vrubezhny:
@tsmaeder I've re-implemented the fix based on your comments.
As result `MoveJavaFileInNewSourceFolderTest` is also looks fixed. (However I still in doubt if I should have leaved that `sleepQuitely()` line in the test as the default timeout is equal to the File Tree Walker period and, in theory, test might fail by timeout under some circumstances)

Please review.

plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-server/src/main/java/org/eclipse/che/plugin/java/plain/server/inject/PlainJavaProjectSourceFolderWatcher.java
```diff
@@ -0,0 +1,144 @@
...
+  private void onServerInitialized(LanguageServerInitializedEvent event) {
+    if (!JavaModule.LS_ID.equals(event.getId())) return;
+
+    LOG.debug(
+        "Registering source folder watching operations for language server : {}", event.getId());
+
+    languageServer = event.getLanguageServer();
+  }
+
+  private void onProjectUpdated(ProjectUpdatedEvent event) {
```
tolusha: Do we have to check if project is a java project?

tolusha (approved):

tsmaeder (changes requested):

plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-server/src/main/java/org/eclipse/che/plugin/java/plain/server/inject/PlainJavaProjectSourceFolderWatcher.java
vrubezhny: It's done exactly the same way here:
https://github.com/eclipse/che/blob/master/wsagent/che-core-api-languageserver/src/main/java/org/eclipse/che/api/languageserver/LanguageServerFileWatcher.java#L82

If I won't wait for LS start, the obtaining of Java LS will be failed - not sure if it's a good trying to  do anything with LS before it's started...
I don't know... Maybe I can use someow `org.eclipse.che.api.languageserver.FindServer.byId(String)` - but again - it could return  null (if it is allowed t use it from `che-plugin-java-plain-server`)

plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-server/src/main/java/org/eclipse/che/plugin/java/plain/server/inject/PlainJavaProjectSourceFolderWatcher.java
```diff
@@ -0,0 +1,144 @@
...
+                return;
+              }
+              Type type = new TypeToken<ArrayList<String>>() {}.getType();
+              List<String> paths = gson.fromJson(gson.toJson(result), type);
+              paths.stream().forEach(f -> matcher.accept(Paths.get(removeUriScheme(prefixURI(f)))));
+            });
+  }
+
+  private PathMatcher folderMatcher() {
+    return it -> isDirectory(it);
```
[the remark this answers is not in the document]
vrubezhny: We discussed this on one of dailies... I would not add any other filters here, because to get the truly information on folder's project I have to ask jdt.ls to refresh the follder's project first, then ask the info from jdt.ls. Otherwise, the info on project might be incorrect. But if we do a refresh on project - why do we need to report a change on the folder? To refresh it once again?

For jdt.ls refreshing a workspace folder shouldn't be a costly operation in case the folder doesn't belong to any project. 

I tried to add here a filter based on project's source folders - it didn't work for some cases.

What if we're creating a folder (a package) inside a source folder in the moment when the project is not yet initialized properly? Your package would never be `visible` as a package from jds.ls point of view until it's parent is refreshed. So, you probably won't be able to move/copy a class into such a package. The problem is in very long project initialization - it takes more than a minute for me (when  https://github.com/eclipse/che/pull/11003 is applied)

vrubezhny:
@tsmaeder I've rebased and modified this PR - now it tries to find Java LS when it's needed by its ID (JavaModule.LS_ID)

plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-server/src/main/java/org/eclipse/che/plugin/java/plain/server/inject/PlainJavaProjectSourceFolderWatcher.java
```diff
@@ -0,0 +1,144 @@
...
+        .getWorkspaceService()
+        .executeCommand(params)
+        .thenAccept(
+            result -> {
+              if (result == null) {
+                return;
+              }
+              Type type = new TypeToken<ArrayList<String>>() {}.getType();
+              List<String> paths = gson.fromJson(gson.toJson(result), type);
+              paths.stream().forEach(f -> matcher.accept(Paths.get(removeUriScheme(prefixURI(f)))));
```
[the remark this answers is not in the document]
tsmaeder: Ok, that is obscure code :-(

tsmaeder (approved):

Issue #10919, opened 2018-08-27
Jdt.ls Fix selenium tests in factory package


tolusha:
### Description
 

Fix selenium tests in factory package

PR #10921, opened 2018-08-27
Fix selenium tests in factory package

tolusha:

### What does this PR do?
Fixes tests in `factory` package

https://github.com/eclipse/che/issues/10919

### DIFF
--- a/plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/JavaLanguageServerExtensionService.java
+++ b/plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/JavaLanguageServerExtensionService.java
@@ -79,6 +79,7 @@ import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -164,6 +165,8 @@ public class JavaLanguageServerExtensionService {
   private final EventService eventService;
   private final LanguageServerInitializer initializer;
   private final ExecuteClientCommandJsonRpcTransmitter clientCommandTransmitter;
+  private final AtomicReference<CompletableFuture<ServerCapabilities>> serverCapabilities =
+      new AtomicReference<>();
 
   @Inject
   public JavaLanguageServerExtensionService(
@@ -460,6 +463,18 @@ public class JavaLanguageServerExtensionService {
     if (findInitializedLanguageServer() == null) {
       throw new IllegalStateException("Language server isn't initialized");
     }
+  }
+
+  private CompletableFuture<ServerCapabilities> getOrInitializeLanguageServer() {
+    if (serverCapabilities.get() == null) {
+      synchronized (serverCapabilities) {
+        if (serverCapabilities.get() == null) {
+          this.serverCapabilities.set(initializer.initialize(prefixURI("/init.java")));
+        }
+      }
+    }
+
+    return serverCapabilities.get();
   }
 
   /**
@@ -1009,8 +1024,7 @@ public class JavaLanguageServerExtensionService {
 
   private CompletableFuture<Object> executeCommand(String commandId, List<Object> parameters) {
     ExecuteCommandParams params = new ExecuteCommandParams(commandId, parameters);
-    return initializer
-        .initialize(prefixURI("/init.java"))
+    return getOrInitializeLanguageServer()
         .thenCompose(
             (ServerCapabilities cap) -> {
               ExtendedLanguageServer ls = findInitializedLanguageServer();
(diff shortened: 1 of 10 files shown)


plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/JavaLanguageServerExtensionService.java
```diff
@@ -462,6 +465,18 @@ private void checkLanguageServerInitialized() {
     }
   }
 
+  private CompletableFuture<ServerCapabilities> getOrInitializeLanguageServer() {
+    if (serverCapabilities.get() == null) {
```
tsmaeder: It makes no sense to do an AtomicReference and synchronized protection.

plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/JavaLanguageServerExtensionService.java
```diff
@@ -462,6 +465,18 @@ private void checkLanguageServerInitialized() {
     }
   }
 
+  private CompletableFuture<ServerCapabilities> getOrInitializeLanguageServer() {
```
tsmaeder: We should not do this. Whether we decide to initialize a server or not should be done by the server initializer. Don't cache the result, it's not worth it.

plugins/plugin-maven/che-plugin-maven-server/src/main/java/org/eclipse/che/plugin/maven/server/projecttype/handler/MavenProjectInitHandler.java
```diff
@@ -11,43 +11,25 @@
...
 @Singleton
 public class MavenProjectInitHandler implements ProjectInitHandler {
 
-  private final WorkspaceSynchronizer workspaceSynchronizer;
-
   @Inject
-  public MavenProjectInitHandler(WorkspaceSynchronizer workspaceSynchronizer) {
-    this.workspaceSynchronizer = workspaceSynchronizer;
-  }
+  public MavenProjectInitHandler(WorkspaceSynchronizer workspaceSynchronizer) {}
```
tsmaeder: If we're not using it, remove the parameter.

plugins/plugin-maven/che-plugin-maven-server/src/main/java/org/eclipse/che/plugin/maven/server/projecttype/handler/MavenProjectInitHandler.java
```diff
@@ -11,43 +11,25 @@
...
   @Override
-  public void onProjectInitialized(String projectFolder)
-      throws ServerException, ForbiddenException, ConflictException, NotFoundException {
-
-    UpdateWorkspaceParameters params = new UpdateWorkspaceParameters();
-    params.setAddedProjectsUri(Collections.singletonList(prefixURI(projectFolder)));
-
-    workspaceSynchronizer.syncronizerWorkspaceAsync(params);
-  }
+  public void onProjectInitialized(String projectFolder) {}
```
tsmaeder, line 33: This looks like we're not using this project initializer anymore. Can you explain how this made a selenium test fail and why we need to change it?
Generally, I think this kind of "meaningful" change would warrant a separate bug report and PR.

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/impl/ProjectServiceApi.java
```diff
@@ -221,7 +221,8 @@ public ProjectConfigDto updateProject(String wsPath, ProjectConfigDto projectCon
       throws NotFoundException, ConflictException, ForbiddenException, ServerException, IOException,
           BadRequestException {
     if (wsPath != null) {
-      projectConfigDto.setPath(absolutize(wsPath));
+      wsPath = absolutize(wsPath);
```
tsmaeder, line 224: Again, what was the bug here?

tsmaeder (changes requested):

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/project/server/impl/ProjectServiceApi.java
tolusha: `absolutize` ensure that project path starts with `/`. 
Without that it leads it emitting `ProjectCreatedEvent` for already registered projects.

Have a look at code below:
```
    boolean registeredEarly = projectManager.isRegistered(wsPath);
    RegisteredProject updated = projectManager.update(projectConfigDto);
```
Without `absolutize` registration checking and updating will be performed for different project paths.

plugins/plugin-maven/che-plugin-maven-server/src/main/java/org/eclipse/che/plugin/maven/server/projecttype/handler/MavenProjectInitHandler.java
tolusha: Using `MavenProjectInitHandler` leads us to endless cycle:
1. Project with problems is imported
2. `JavaLanguageServerExtensionService.updateWorkspace` is invoked
3. `JavaLanguageServerExtensionService.updateProjectsWithProblems` is invoked
4. `MavenProjectInitHandler.onProjectInitialized` is invoked.
5. `WorkspaceSynchronizer.syncronizerWorkspaceAsync` is invoked
6. -> 2

svor (approved):
One point here. All existed factories which use java won't work after merging jdt.ls into master because they don't have `org.eclipse.che.ls.java` installer. It can be a problem.

tolusha:
@tsmaeder Could you have a look again?

tsmaeder (approved):

Issue #9591, opened 2018-05-03
Address Selenium Test Merge Blockers


tsmaeder:
Epic for addressing issues found in selenium tests with jdt.ls.

- [x] https://github.com/eclipse/che/issues/10431
- [x] https://github.com/eclipse/che/issues/10115
- [x] https://github.com/eclipse/che/issues/10425
- [x] https://github.com/eclipse/che/issues/10112
- [x] https://github.com/eclipse/che/issues/10095 
- [x] https://github.com/eclipse/che/issues/10094
- [x] https://github.com/eclipse/che/issues/10034
- [x] https://github.com/eclipse/che/issues/10451
- [x] https://github.com/eclipse/che/issues/10466

tsmaeder:
Package | Tests | Errors | Failures | Skipped | Success Rate | Time
-- | -- | -- | -- | -- | -- | --
org.eclipse.che.selenium.preferences | 7 | 0 | 2 | 0 | 71.429% | 222.404
org.eclipse.che.selenium.refactor.types | 17 | 0 | 1 | 0 | 94.118% | 328.927
org.eclipse.che.selenium.refactor.methods | 30 | 0 | 2 | 7 | 70% | 513.79
org.eclipse.che.selenium.dashboard | 25 | 0 | 1 | 0 | 96% | 277.458
org.eclipse.che.selenium.debugger | 45 | 0 | 3 | 5 | 82.222% | 537.671
org.eclipse.che.selenium.plainjava | 5 | 0 | 4 | 1 | 0% | 38.358
org.eclipse.che.selenium.dashboard.workspaces.details | 9 | 0 | 4 | 0 | 55.556% | 154.271
org.eclipse.che.selenium.testrunner | 6 | 0 | 0 | 0 | 100% | 144.4
org.eclipse.che.selenium.gwt | 1 | 0 | 0 | 0 | 100% | 143.316
org.eclipse.che.selenium.intelligencecommand | 17 | 0 | 5 | 9 | 17.647% | 420.408
org.eclipse.che.selenium.refactor.move | 13 | 0 | 4 | 0 | 69.231% | 443.595
org.eclipse.che.selenium.refactor.preview | 2 | 0 | 0 | 0 | 100% | 6.974
org.eclipse.che.selenium.workspaces | 9 | 0 | 3 | 0 | 66.667% | 324.146
org.eclipse.che.selenium.git | 61 | 0 | 4 | 15 | 68.852% | 1,619.099
[...]

PR #10938, opened 2018-08-27
Jdt.ls: fix selenium test PullRequestPluginWithForkTest and PullRequestPluginTest

vparfonov:
### What does this PR do?
Fixing PullRequestPluginWithForkTest and PullRequestPluginTest by adding waiting on jdt.ls initialization  

#9591


svor (approved):

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/Consoles.java
```diff
@@ -409,6 +411,11 @@ public void waitExpectedTextIntoConsole(String expectedText) {
     updateProjDriverWait.until(textToBePresentInElement(consoleContainer, expectedText));
   }
 
+  /** wait on staring jdt.ls */
+  public void waitUntilJdtLsStarted() {
```
dmytro-ndp, line 415: To be more precise, we are waiting on some message appearance only: `waitJDTLSStartedMessage`

vparfonov: As you wish @dmytro-ndp, i will fix it

Issue #10756, opened 2018-08-13
Yo generator doesn't work in Theia image


mmorhun:
### Description
Theia docker image has build-in yo generator. However it doesn't work and fall with the following error:
```
/projects # yo
/usr/local/lib/node_modules/yo/node_modules/write-file-atomic/index.js:197
    throw err
    ^

Error: EACCES: permission denied, open '/home/theia/.config/insight-nodejs/insight-yo.json.4135311943'
    at Object.fs.openSync (fs.js:646:18)
    at Function.writeFileSync [as sync] (/usr/local/lib/node_modules/yo/node_modules/write-file-atomic/index.js:180:17)
    at Conf.set store [as store] (/usr/local/lib/node_modules/yo/node_modules/conf/index.js:142:19)
    at Conf.set (/usr/local/lib/node_modules/yo/node_modules/conf/index.js:64:14)
    at Insight.set optOut [as optOut] (/usr/local/lib/node_modules/yo/node_modules/insight/lib/index.js:56:15)
    at Object.<anonymous> (/usr/local/lib/node_modules/yo/lib/cli.js:198:18)
    at Module._compile (module.js:652:30)
    at Object.Module._extensions..js (module.js:663:10)
    at Module.load (module.js:565:32)
    at tryModuleLoad (module.js:505:12)
```

### Reproduction Steps
Just run `yo` in workspace from standard stack (docker Theia image).

### Environment

Che version: `6.10.0-SNAPSHOT`
Theia image: `eclipse/che-theia:0.4.0-next.b17727c1-nightly`

sunix:
![eclipse che theia_factory_test_3 - google chrome_150](https://user-images.githubusercontent.com/650571/44036964-74977dde-9f13-11e8-8d51-77dbc5e91c07.png)
weird

benoitf:
@mmorhun AFAIK it's working with theia image 0.3.13-nightly ?

benoitf:
https://github.com/eclipse/che/blob/c5e92e7156439c9ed8f1475b50944572e337aa6f/ide/che-core-ide-stacks/src/main/resources/stacks.json is fine. Maybe your che data folder is old then stacks are not updated

mmorhun:
@benoitf yes you are right. I forgot to delete old config.

mmorhun:
More testing shows that it still has some troubles. Yo starts, user can choose options, but finally it failed on generation phase with the following error:
```
   create package.json
fs.js:646
[...]

PR #10960, opened 2018-08-28
Fix clone and yo in Theia image

ghost:
### What does this PR do?

Adds a little script to entrypoint that will write current uid into /etc/passwd and /etc/group

https://github.com/eclipse/che/issues/10756
https://github.com/eclipse/che-theia-factory-extension/issues/8

Factory to try https://che.openshift.io/f?id=factoryfv9dkqhzh5xnshvq

peek 2018-08-28 16-40

### DIFF
--- a/dockerfiles/theia/src/setup.sh
+++ b/dockerfiles/theia/src/setup.sh
@@ -5,7 +5,7 @@ set -u
 for f in "/etc/passwd" "/etc/group"; do
     chgrp -R 0 ${f}
     chmod -R g+rwX ${f};
-        done
+done
 # Generate passwd.template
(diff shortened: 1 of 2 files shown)


benoitf (approved):

dockerfiles/theia/src/setup.sh
```diff
@@ -2,6 +2,16 @@
 set -e
 set -u
 
+for f in "/etc/passwd" "/etc/group"; do
+    chgrp -R 0 ${f}
+    chmod -R g+rwX ${f};
+        done
```
benoitf: formatting looks strange there

ghost: fixed

benoitf: thx. already approved

Issue #10966, opened 2018-08-28
Unexpected fail of "	RollingUpdateStrategyWithWorkspacesStartStopTest" selenium test


Ohrimenko1988:
### Description
We have unexpected fail of "RollingUpdateStrategyWithWorkspacesStartStopTest" selenium test. It looks like something went wrong with a method which checks workspace status. We should investigate and fix this problem.

**Log:**
```
org.openqa.selenium.TimeoutException: 
Expected condition failed: waiting for org.eclipse.che.selenium.pageobject.dashboard.workspaces.Workspaces$1@56465f59 (tried for 240 second(s) with 500 MILLISECONDS interval)
Build info: version: '3.0.1', revision: '1969d75', time: '2016-10-18 09:49:13 -0700'
System info: host: 'slave6.codenvycorp.com', ip: '127.0.0.1', os.name: 'Linux', os.arch: 'amd64', os.version: '3.10.0-862.11.6.el7.x86_64', java.version: '1.8.0_102'
Driver info: org.eclipse.che.selenium.core.SeleniumWebDriver
	at org.openqa.selenium.support.ui.WebDriverWait.timeoutException(WebDriverWait.java:80)
	at org.openqa.selenium.support.ui.FluentWait.until(FluentWait.java:265)
	at org.eclipse.che.selenium.pageobject.dashboard.workspaces.Workspaces.waitWorkspaceStatus(Workspaces.java:125)
	at org.eclipse.che.selenium.hotupdate.rolling.RollingUpdateStrategyWithWorkspacesStartStopTest.startStopWorkspaceFunctionsShouldBeAvailableDuringRollingUpdate(RollingUpdateStrategyWithWorkspacesStartStopTest.java:62)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:85)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:696)
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:882)
	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1189)
[...]

PR #10967, opened 2018-08-28
[Selenium] Fix of unexpected fail of " RollingUpdateStrategyWithWorkspacesStartStopTest" selenium test

Ohrimenko1988:
### What does this PR do?
Fix of unexpected fail of " RollingUpdateStrategyWithWorkspacesStartStopTest" selenium test

Issue: https://github.com/eclipse/che/issues/10966

### DIFF
--- a/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/dashboard/workspaces/Workspaces.java
+++ b/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/dashboard/workspaces/Workspaces.java
@@ -296,7 +296,6 @@ public class Workspaces {
 
   public void clickOnWorkspaceStopStartButton(String workspaceName) {
     String buttonXpath = String.format(WORKSPACE_ITEM_STOP_START_WORKSPACE_BUTTON, workspaceName);
-    waitProgressBarInvisibility();
 
     // to avoid clicking on the tooltip
     moveCursorToWorkspaceRamSection(workspaceName);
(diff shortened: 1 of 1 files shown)


SkorikSergey (approved):

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/dashboard/workspaces/Workspaces.java
```diff
@@ -290,12 +296,15 @@ public void clickOnStackButton() {
 
   public void clickOnWorkspaceStopStartButton(String workspaceName) {
     String buttonXpath = String.format(WORKSPACE_ITEM_STOP_START_WORKSPACE_BUTTON, workspaceName);
+    waitProgressBarInvisibility();
```
dmytro-ndp: Do we really need to wait for progress bar invisibility at the start of method? 
It is expected that workspace is completely ready to perform an operation, isn't it?

Ohrimenko1988: By that case we ensuring that loader is not displayed because if it is present (because of some previous actions) click event may be lost.

musienko-maxim (approved):

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/dashboard/workspaces/Workspaces.java
dmytro-ndp: I see. But the method propose is just clicking on workspace stop/start button. It doesn't declare waiting on something before clicking. So, IMHO, it's client responsibility to ensure that there is nothing blocks clicking on the button.

dmytro-ndp (approved):

Issue #10974, opened 2018-08-28
Jdt.ls Fix selenium tests in autocomplete package


svor:
### Description
Fix selenium tests in autocomplete package

PR #10975, opened 2018-08-28
Fix tests in org.eclipse.che.selenium.editor.autocomplete

svor:

### What does this PR do?
Fixes tests in `autocomplete` package

https://github.com/eclipse/che/issues/10974

### DIFF
--- a/ide/che-core-orion-editor/src/main/java/org/eclipse/che/ide/editor/orion/client/signature/SignatureWidget.java
+++ b/ide/che-core-orion-editor/src/main/java/org/eclipse/che/ide/editor/orion/client/signature/SignatureWidget.java
@@ -58,6 +58,7 @@ import org.eclipse.che.ide.util.dom.Elements;
 public class SignatureWidget implements EventListener {
   private static final String DOCUMENTATION = "documentation";
   private static final int DOM_ITEMS_SIZE = 50;
+  private static final String SIGNATURES_CONTENT_ID = "signaturesContent";
 
   private final PopupResources popupResources;
   /** The related editor. */
@@ -107,6 +108,7 @@ public class SignatureWidget implements EventListener {
     popupElement.appendChild(headerElement);
 
     popupBodyElement = Elements.createDivElement(popupResources.popupStyle().body());
+    popupBodyElement.setId(SIGNATURES_CONTENT_ID);
     popupElement.appendChild(popupBodyElement);
 
     listElement = Elements.createUListElement();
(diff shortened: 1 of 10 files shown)


ide/che-core-orion-editor/src/main/java/org/eclipse/che/ide/editor/orion/client/signature/SignatureWidget.java
```diff
@@ -107,6 +107,7 @@ public SignatureWidget(
     popupElement.appendChild(headerElement);
 
     popupBodyElement = Elements.createDivElement(popupResources.popupStyle().body());
+    popupBodyElement.setClassName("signaturesContent");
```
tsmaeder: Isn't gwt-debug-id the way to go here?

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
```diff
@@ -1484,6 +1494,7 @@ private void waitTextInJavaDoc(String expectedText) {
    * @return true - if {@code expectedText} is present in javadoc body, false - if not
    */
   public boolean waitAndCheckTextPresenceInJavaDoc(String expectedText) {
+    seleniumWebDriverHelper.moveCursorTo(By.xpath(JAVA_DOC_POPUP));
```
tsmaeder: Is Javadoc still a thing? Shouldn't this be some generic language server popup?

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/editor/autocomplete/AutocompleteProposalJavaDocTest.java
```diff
@@ -206,41 +220,39 @@ public void shouldDisplayJavaDocOfJreClass() throws IOException {
...
     // when
     editor.waitActive();
     loader.waitOnClosed();
     editor.goToCursorPositionVisible(31, 23);
     editor.launchAutocompleteAndWaitContainer();
-    editor.selectAutocompleteProposal("info(String msg) : void Logger");
+    editor.selectCompositeAutocompleteProposal("info(String msg) : void Logger");
 
     // then
-    editor.waitProposalDocumentationHTML(".*No documentation found.*");
```
tsmaeder, line 221: This defeats the purpose of the test. We need an example that does not have javadoc or source.

tsmaeder (changes requested):

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
svor: renamed to `TEXT_VIEW_TOOLTIP`

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/editor/autocomplete/AutocompleteProposalJavaDocTest.java
svor: So if we don't have "Download source" feature i suggest to leave this test as I reworked (now it checks when jDoc is empty and popup window doesn't shown).
And I've also created one issue to write new test: https://github.com/eclipse/che/issues/10981

tsmaeder: Sorry, I misread the test here...If the removal of the slf4j artifacts works, it should show "No documentation found" afterwards, no?
I would think that the code jar is still there because the file is opened and cannot be removed.

svor: Now quick documentation shows in Orion tooltip and when Language Server returns empty string - tooltip isn't shown. That is why no "No documentation found" here, so i would say it is normal behavior

tsmaeder (approved):

Issue #10983, opened 2018-08-29
Simplify jdt.ls PR


tsmaeder:
Remove unrelated changes and some deletions from jdt.ls branch to make it better reviewable

PR #10984, opened 2018-08-29
Simplify jdtls pr

tsmaeder:
### What does this PR do?
Removes some changes from the jdt.ls branch (mostly removals) to reduce the number of changes in the jdt.ls branch

#10983


multiuser/sql-schema/src/main/resources/che-schema/5.19.0/0.1__add_permissions.sql
```diff
@@ -1,5 +1,5 @@
 --
--- Copyright (c) 2012-2018 Red Hat, Inc.
+-- Copyright (c) 2012-2017 Red Hat, Inc.
```
benoitf, line 2: it seems that there are some leftover 2017 dates on these sql files that should not have been changed

plugins/plugin-java/che-plugin-java-ext-jdt/org-eclipse-core-filebuffers/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+                <configuration>
+                    <excludes>
+                        <exclude>**/*.png</exclude>
+                        <exclude>**/*.gif</exclude>
+                        <exclude>**/*.rscjson</exclude>
+                        <exclude>**/*.zip</exclude>
+                        <exclude>**/*.txt</exclude>
+                        <exclude>**/*.jpg</exclude>
+                        <exclude>**/*.rsc</exclude>
+                        <exclude>**/*.json</exclude>
```
benoitf, line 85: do we need all these excludes ? it seems this module doesn't contain all these kind of files

multiuser/sql-schema/src/main/resources/che-schema/5.19.0/0.1__add_permissions.sql
sleshchenko: @benoitf you're right. These changes **must** be excluded from this PR since `flyway` evaluates hash of content and uses it to ensure that the right migration scripts are applied.

benoitf: in fact it's ok I read too fast :-) Thomas is putting back the dates to avoid master changes. please ignore my remark

plugins/plugin-java/che-plugin-java-ext-jdt/org-eclipse-core-filebuffers/pom.xml
tsmaeder: This is all dead code!

benoitf: yes pls ignore my comment :-)

multiuser/sql-schema/src/main/resources/che-schema/5.19.0/0.1__add_permissions.sql
sleshchenko: =) As well as my comment ) I saw that these sql files are modified and missed that this PR is not targetted to master branch =)
So, please mark this thread as resolved )

benoitf (approved):

PR #10990, opened 2018-08-30
Make tests stable in projectexplorer package

tolusha:

### What does this PR do?
Makes tests stable in projectexplorer package

### DIFF
--- a/selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/webdriver/SeleniumWebDriverHelper.java
+++ b/selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/webdriver/SeleniumWebDriverHelper.java
@@ -1401,4 +1401,9 @@ public class SeleniumWebDriverHelper {
   public void waitSuccessCondition(ExpectedCondition<Boolean> expression) {
     waitSuccessCondition(expression, DEFAULT_TIMEOUT);
   }
+
+  /** Hides context menu. */
+  public void hideContextMenu() {
+    actionsFactory.createAction(seleniumWebDriver).moveByOffset(-1, -1).click().build().perform();
+  }
 }
--- a/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java
+++ b/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java
@@ -68,6 +68,7 @@ import org.openqa.selenium.Keys;
 import org.openqa.selenium.NoSuchElementException;
 import org.openqa.selenium.StaleElementReferenceException;
 import org.openqa.selenium.TimeoutException;
+import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.interactions.Actions;
 import org.openqa.selenium.support.FindBy;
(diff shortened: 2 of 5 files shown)


selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/webdriver/SeleniumWebDriverHelper.java
```diff
@@ -1401,4 +1401,9 @@ public void waitSuccessCondition(ExpectedCondition<Boolean> expression, int time
   public void waitSuccessCondition(ExpectedCondition<Boolean> expression) {
     waitSuccessCondition(expression, DEFAULT_TIMEOUT);
   }
+
+  /** Hides context menu. */
+  public void hideContextMenu() {
+    actionsFactory.createAction(seleniumWebDriver).moveByOffset(-1, -1).click().build().perform();
```
tsmaeder, line 1407: What are we clicking here? Maybe <esc> would be better?

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java
```diff
@@ -655,14 +656,23 @@ public void waitRemoveItemsByPath(String pathToItem) {
    * @param path item's path in format: "Test/src/pom.xml".
    */
   public void openContextMenuByPathSelectedItem(String path) {
-    waitItem(path);
-    waitAndSelectItem(path);
-    waitItemIsSelected(path);
-
-    Actions action = actionsFactory.createAction(seleniumWebDriver);
-    action.moveToElement(waitAndGetItem(path)).contextClick().perform();
+    for (int i = 1; ; i++) {
```
tsmaeder, line 661: Could you explain what the problem was here and how the current code fixes it?

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java
```diff
@@ -1092,6 +1102,20 @@ public void waitItemIsSelected(String path) {
         ELEMENT_TIMEOUT_SEC);
   }
 
+  /**
+   * Waits until item with specified {@code path} be selected
+   *
+   * @param path item's path in format: "Test/src/pom.xml".
+   */
+  public void waitItemIsSelected(String path, int timeout) {
```
tsmaeder, line 1107: The version of this method without timeout parameter should call this one.

tsmaeder (changes requested):

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java
```diff
@@ -655,14 +656,23 @@ public void waitRemoveItemsByPath(String pathToItem) {
...
+      waitAndSelectItem(path);
 
-    waitContextMenu();
+      actionsFactory.createAction(seleniumWebDriver).contextClick().perform();
+      waitContextMenu();
+
+      try {
+        waitItemIsSelected(path, REDRAW_UI_ELEMENTS_TIMEOUT_SEC);
+        return;
+      } catch (WebDriverException e) {
```
dmytro-ndp: `WebDriverException` looks too general here. There is, for example, *UnreachableBrowserException* type of WebDriverException after which it doesn't make sense to repeat operation again.
`TimeoutException` would be enough in case of waiting for web element.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/webdriver/SeleniumWebDriverHelper.java
tolusha: To close context menu we have to click outside of it. So we move a mouse not far from menu and perform click.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/ProjectExplorer.java
tolusha: Selecting item and invoking context menu are two separate operations. 
 If some item retakes focus then context menu is invoked at wrong item. It might happen because project explorer updates asynchronously and new appeared items retake focus.
So, there are two tries to invoke context menu at correct item

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/webdriver/SeleniumWebDriverHelper.java
Ohrimenko1988: You can simplify this:
```
getAction().moveByOffset(-1, -1).click().perform();
```

tsmaeder: I was thinking of typing escape, but github took that as a tag :man_facepalming:

tsmaeder (approved):

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/webdriver/SeleniumWebDriverHelper.java
tolusha: Unfortunately escape does n't close context menu in che.

PR #11005, opened 2018-08-31
Fix selenium tests in languageserver package

tolusha:

### What does this PR do?
Fixes selenium tests in languageserver package

### DIFF
--- a/plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/ProjectsListener.java
+++ b/plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/ProjectsListener.java
@@ -94,7 +94,7 @@ public class ProjectsListener {
   private boolean isJavaProject(String path) {
     Optional<RegisteredProject> project = projectManager.get(path);
     return (project.isPresent()
-        && (project.get().getType().equals(MavenAttributes.MAVEN_ID)
-            || project.get().getType().equals(Constants.JAVAC)));
+        && (MavenAttributes.MAVEN_ID.equals(project.get().getType())
+            || Constants.JAVAC.equals(project.get().getType())));
   }
 }
--- a/plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/ProjectsListener.java
+++ b/plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/ProjectsListener.java
@@ -16,6 +16,7 @@ import static java.util.Collections.singletonList;
 
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
+import java.util.Optional;
 import javax.annotation.PostConstruct;
 import org.eclipse.che.api.core.notification.EventService;
 import org.eclipse.che.api.core.notification.EventSubscriber;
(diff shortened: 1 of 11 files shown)


selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
```diff
@@ -937,7 +937,7 @@ public void selectAutocompleteProposal(String item) {
    */
   public void selectCompositeAutocompleteProposal(String item) {
     seleniumWebDriverHelper.waitAndClick(
-        By.xpath(format(AUTOCOMPLETE_CONTAINER + "/li/span[.='%s']", item)));
+        By.xpath(format(AUTOCOMPLETE_CONTAINER + "/li/span[text()='%s']", item)));
```
svor: just reuse this method where you need
``` 
public void selectAutocompleteProposal(String item) {
    seleniumWebDriverHelper.waitAndClick(
        By.xpath(format(AUTOCOMPLETE_CONTAINER + "/li/span[text()='%s']", item)));
  }
```
https://github.com/eclipse/che/blob/0b2ea498eb6662d50e81c2bbb4b5a84a058e22ec/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java#L927

plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/ProjectsListener.java
```diff
@@ -87,7 +91,10 @@ private void onPreProjectDeleted(PreProjectDeletedEvent event) {
     workspaceSynchronizer.syncronizerWorkspaceAsync(params);
   }
 
-  private boolean isProjectRegistered(String path) {
-    return projectManager.isRegistered(path);
+  private boolean isJavaProject(String path) {
+    Optional<RegisteredProject> project = projectManager.get(path);
+    return (project.isPresent()
+        && (project.get().getType().equals(MavenAttributes.MAVEN_ID)
```
svor: it is better to call equals from constant to avoid NPE

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/CodenvyEditor.java
tsmaeder: I'm pretty sure you're going to break a couple of tests in the java autocomplete tests with this test. Why would including the text of enclosed spans break any test?

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/languageserver/PhpFileEditingTest.java
```diff
@@ -79,11 +79,8 @@ public void setup() throws Exception {
 
   @Test
   public void checkMainFeaturesPhpLS() {
-    String intitPhpLanguageServerMessage =
-        String.format(
-            "Finished language servers initialization, file path '/%s'", PATH_TO_INDEX_PHP);
-
-    consoles.waitExpectedTextIntoConsole(intitPhpLanguageServerMessage);
+    consoles.waitExpectedTextIntoConsole(
```
tsmaeder: It's odd to use a constant in the other tests, but not here. If you're fixing this line anyway, why not make it the same as the other tests?

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/languageserver/TypeScriptEditingTest.java
```diff
@@ -76,11 +76,8 @@ public void setUp() throws Exception {
 
   @Test
   public void checkMainFeaturesTypeScriptLS() {
-    String intitTypeScriptLanguageServerMessage =
-        String.format(
-            "Finished language servers initialization, file path '/%s'", PATH_TO_GREETER_FILE);
-
-    consoles.waitExpectedTextIntoConsole(intitTypeScriptLanguageServerMessage);
+    consoles.waitExpectedTextIntoConsole(
```
tsmaeder: Again, constant or not?

tsmaeder (changes requested):

svor (approved):

tsmaeder (approved):

Issue #11008, opened 2018-08-31
Unexpected error when authentication with identity provider in Eclipse Che Multi User on OCP


dmytro-ndp:
### Description
E2E selenium tests which verify working with _OCP_ authentication from _Eclipse Che_ [failed on CI](https://ci.codenvycorp.com/view/qa/job/che-multiuser-master-ocp-oauth-test/76) because of error `Unexpected error when authentication with identity provider`: 
![org eclipse che selenium site ocpoauth loginexisteduserwithopenshiftoauthtest checkexistedcheuserocpprojectcreationandremoval_xgzhyjoy](https://user-images.githubusercontent.com/1197777/44907068-78f95e00-ad1f-11e8-8291-11800f235410.png)

**_The reason_**: there is incorrect link to certificate in [deploy_che.sh](https://github.com/eclipse/che/blob/2b7ebd8f852c154f8ab5d423130ea8cbb64ce223/deploy/openshift/deploy_che.sh#L393):
```
--> Deploying template "eclipse-che/openshift-identity-provider-certificate" for "/home/codenvy/workspace/che-multiuser-master-ocp-oauth-test/deploy/openshift/templates/multi/openshift-certificate-secret.yaml" to project eclipse-che
```

PR #11010, opened 2018-08-31
Fix path to OCP oauth certificate

dmytro-ndp:
### What does this PR do?
It fixes path to OCP authentification certificate which wasn't restored after returning to OKD 3.9.0.

#11008


ghost (approved):
Thanks for fixing that. Leftovers from 3.10 upgrade

sleshchenko (approved):

Issue #10808, opened 2018-08-16
Clean up branches in Che Theia hosted plugin extension


sunix:
### Description
At the moment, https://github.com/eclipse/che-theia-hosted-plugin-manager-extension has 3 branches and it is very confusing which branch to use. The goals of this issue are

1. to have only one used branch: `master`.
2. use this one in the Che Theia docker image: https://github.com/eclipse/che/blob/master/dockerfiles/theia/src/extensions.json#L34
    ```
        ...
        },
        {
            "name": "@eclipse-che/che-theia-hosted-plugin-manager-extension",
            "source": "https://github.com/eclipse/che-theia-hosted-plugin-manager-extension.git",
            "checkoutTo": "latest-deps",
            "type": "git"
        },
        {
        ....
    ```
3. make sure that everything is working fine after theses changes

mmorhun:
@sunix this is problem of the extension, not Che. Please, next time let's create such issues in appropriate repository.

sunix:
@mmorhun 
AFAIK we create issue related to che in the che github issue.
- issues related to che to be done in https://github.com/eclipse/che-docs goes here
- issues related to che to be done in https://github.com/eclipse/che-dependencies used to go here as well, etc ...

mmorhun:
@sunix IMO, if the code is core part of Che it should be in Che repo, not somewhere else (obviously che parent, deps and docs are exceptions). If not - it is a separate repository with own issues.
Taking into account workspace next  features, Theia is just an option. And I have doubts about creating option's plugin(s) issues in main Che repo.

sunix:
What do you mean by "core" part ... it in Che by default or not. This is going to Che default packaging so it makes sense that it is in Che. Let's discuss about that during next Che dev meeting.

mmorhun:
By core I mean things which is required for Che and without them Che isn't usable.

sunix:
So none of the ui part part is part of core, as some of our customer use Che without our UI. We can use Che with openshift so docker support is not part of Che core. It doesn't make sense. Che is all the features we provide in Che product.

mmorhun:
No it makes sense. Just current approach is different. And I agree that statement with only core doesn't suite it. My point is that, issue should be near related code. In this very case it is not true. Why should one create an issue in one repository to fix something in another. Does it make sense?

sunix:
[...]

PR #11039, opened 2018-09-03
CHE-10808: Use master branch of Che-Theia-hosted-plugin-manager-extension

mmorhun:
### What does this PR do?
Before resolutions have been merged we used separate branch with specific npm packages versions to make che-theia-hosted-plugin-manager-extension works against latest Theia release.
Now, when resolutions are applied in our Theia image, we can use master branch of the extension without a need to patch the extension an each Theia release (unless API breaking changes introduced).

https://github.com/eclipse/che/issues/10808


AndrienkoAleksandr (approved):

mmorhun:
Tested on local docker infra and openshift.io

evidolob (approved):

benoitf (approved):

PR #11076, opened 2018-09-05
CHE-10756: Use regular user instead of root in theia image

mmorhun:
Signed-off-by: Mykola Morhun <mmorhun@redhat.com>

### What does this PR do?
Before we used root in Theia image. It is not good way to go and moreover caused some problems, for example, with `yo` generator.
This PR adds regular user `theia` and runs Theia IDE on behalf of the user.

https://github.com/eclipse/che/issues/10756

Tested on docker infrastructure and openshift.io

### DIFF
--- a/dockerfiles/theia/src/setup.sh
+++ b/dockerfiles/theia/src/setup.sh
@@ -14,16 +14,6 @@
 set -e
 set -u
 
-for f in "/etc/passwd" "/etc/group"; do
-    chgrp -R 0 ${f}
-    chmod -R g+rwX ${f};
-done
-# Generate passwd.template
-cat /etc/passwd | sed s#root:x.*#root:x:\${USER_ID}:\${GROUP_ID}::\${HOME}:/bin/bash#g > ${HOME}/passwd.template
-# Generate group.template
-cat /etc/group | sed s#root:x:0:#root:x:0:0,\${USER_ID}:#g > ${HOME}/group.template
-
-
 # Install basic software used for checking github API rate limit
 yum install -y epel-release
 yum -y install curl jq expect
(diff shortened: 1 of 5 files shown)


dockerfiles/theia/src/setup.sh
```diff
@@ -2,6 +2,12 @@
 set -e
 set -u
 
+yum install -y sudo
+# Add a regular user
+useradd -u 1000 -G users,wheel,root -d ${HOME} --shell /bin/bash theia
+usermod -p "*" theia
+echo "%wheel ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
+
```
benoitf: all this stuff

```
yum install -y sudo
# Add a regular user
useradd -u 1000 -G users,wheel,root -d ${HOME} --shell /bin/bash theia
usermod -p "*" theia
echo "%wheel ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers

for f in "/etc/passwd" "/etc/group"; do
    chgrp -R 0 ${f}
    chmod -R g+rwX ${f};
done
# Generate passwd.template
cat /etc/passwd | sed s#root:x.*#root:x:\${USER_ID}:\${GROUP_ID}::\${HOME}:/bin/bash#g > ${HOME}/passwd.template
# Generate group.template
cat /etc/group | sed s#root:x:0:#root:x:0:0,\${USER_ID}:#g > ${HOME}/group.template
```

should go after the check of github limit to fail fast.

--> where we have `# Add yarn repo`

dockerfiles/theia/src/setup.sh
```diff
@@ -2,6 +2,12 @@
 set -e
 set -u
 
+yum install -y sudo
```
benoitf: if possible, merge it with other yum install instruction post github-check

benoitf:
ci-build

benoitf:
@mmorhun could you rebase changes to check the e2e tests are passing ?

dockerfiles/theia/src/setup.sh
mmorhun: Yes, thanks.

dockerfiles/theia/src/setup.sh
mmorhun: To me it is clearer when we install components in its logical group/place.
Just curious, what is advantage of such merging?

benoitf: speed of build (like searching for mirrors again)

AndrienkoAleksandr (approved):

mmorhun:
@benoitf done
[
tests
](url)

ghost (approved):

benoitf:
ci-build

benoitf (approved):

benoitf (approved):
Please squash the commits

Issue #11091, opened 2018-09-06
Don't ask Java compiler options during workspace starting


svor:
### Description
 
When workspace has status RUNNING we ask jdt.ls to get Java compiler options, but at this time jdt.ls has not ready yet and we got some exceptions in browser console.
On the another way this request (to ask Java compiler options) initiate initialization of jdt.ls, so need to find another way to initialize jdt.ls when workspace has started.

### Reproduction Steps

**OS and version:**    

**Diagnostics:**

tsmaeder:
I'm really not sure we're handling this the right way. The problematic assumption here is that we have access to the user preferences as part of workspace startup. However, stuff managed by a language server is not supposed to be available.
Maybe compiler options should simply not be part of "preferences".

tsmaeder:
Just discussed this with @svor . The exception shown in the browser log comes from org.eclipse.che.api.languageserver.LanguageServerInitializer.initialize(String), line 136 (in master).
The problem is that it treats finding no language server to initialize as an error. However, this is wrong: if you turn off the installer for the php language server, you will not find a language server for PHP files. This is not an error, but an expected outcome of trying to initialize language servers for a particular file. We could handle this either by returning LS capabilities with everything turned off, or we could explicitly return null. I would chose the first option.

PR #11094, opened 2018-09-06
Don't throw an exception if none LS was found

svor:

### What does this PR do?
1. Add a way to check if jdt.ls installer exists in workspace's configuration;
2. Add initialization methods which is called when workspace is ready;
3. Don't ask for Java compiler options when workspace is running.

https://github.com/eclipse/che/issues/11091

### DIFF
--- a/plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/JavaLanguageServerExtensionService.java
+++ b/plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/JavaLanguageServerExtensionService.java
@@ -955,7 +955,9 @@ public class JavaLanguageServerExtensionService {
    */
   public JavaCoreOptions getJavaCoreOptions(List<String> filter) {
     Type type = new TypeToken<JavaCoreOptions>() {}.getType();
-    return doGetOne(Commands.GET_JAVA_CORE_OPTIONS_СOMMAND, new ArrayList<>(filter), type);
+    JavaCoreOptions result =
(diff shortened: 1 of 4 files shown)


tolusha (approved):

tsmaeder (changes requested):
Generally, I think we should do most of the work in the back end.

svor:
@tsmaeder I've updated the PR. Please take another look at it.

plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/JavaLanguageServerExtensionService.java
```diff
@@ -955,7 +955,9 @@ public OrganizeImportsResult organizeImports(OrganizeImportParams organizeImport
    */
   public JavaCoreOptions getJavaCoreOptions(List<String> filter) {
     Type type = new TypeToken<JavaCoreOptions>() {}.getType();
-    return doGetOne(Commands.GET_JAVA_CORE_OPTIONS_СOMMAND, new ArrayList<>(filter), type);
+    JavaCoreOptions result =
+        doGetOne(Commands.GET_JAVA_CORE_OPTIONS_СOMMAND, new ArrayList<>(filter), type);
+    return result == null ? new JavaCoreOptions() : result;
```
tsmaeder, line 960: Why do we need the null check? Shoudn't this return a result or throw an exception?

svor: It's better to return empty map of java compiler options than null, if jdt.ls wasn't initialized

tsmaeder (approved):

PR #11117, opened 2018-09-07
Set JWT proxy pulling strategy to Always

garagatyi:
### What does this PR do?
Now JWT Proxy uses image pull policy `IfNotPresent` which means that updates of a tag of the image are not populated on a node where this tag already exists.
I believe that pulling strategy `Always` would be a better fit for a security container.


skabashnyuk (approved):

sleshchenko (approved):
Good catch. Thanks for fixing it :+1:

PR #11118, opened 2018-09-07
Fix an ability to use locally built images of Che on minishift

garagatyi:
### What does this PR do?
Latest versions of minishift ignore Che master docker image that
was built locally inside the VM.
Removing `docker.io` prefix from the image name workarounds the issue.


sleshchenko (approved):
LGTM

ghost (approved):

Issue #11110, opened 2018-09-07
JDT.LS Configure lib folder for plain java project


svor:
### Description
We don't have a way to add all jars from some folder into classpath after plain java project configuration.

After plain java project creation try to configure folder with libraries in Project Configuration window
![selection_001](https://user-images.githubusercontent.com/1271546/45203427-dee86700-b284-11e8-9b7f-7b472563dda0.png)

Selenium test to check: `RunPlainJavaProjectTest`

PR #11133, opened 2018-09-10
Fix for RunPlainJavaProjectTest test

svor:

### What does this PR do?
Make it possible to configure folder with libraries for plain java project

https://github.com/eclipse/che/issues/11110

### DIFF
--- a/plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-server/src/main/java/org/eclipse/che/plugin/java/plain/server/projecttype/PlainJavaInitHandler.java
+++ b/plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-server/src/main/java/org/eclipse/che/plugin/java/plain/server/projecttype/PlainJavaInitHandler.java
@@ -11,13 +11,22 @@
  */
 package org.eclipse.che.plugin.java.plain.server.projecttype;
 
+import static org.eclipse.che.api.fs.server.WsPathUtils.absolutize;
 import static org.eclipse.che.ide.ext.java.shared.Constants.JAVAC;
+import static org.eclipse.che.plugin.java.plain.shared.PlainJavaProjectConstants.LIBRARY_FOLDER;
 
+import com.google.inject.Inject;
+import com.google.inject.Provider;
+import java.util.List;
 import org.eclipse.che.api.core.ConflictException;
 import org.eclipse.che.api.core.ForbiddenException;
 import org.eclipse.che.api.core.NotFoundException;
 import org.eclipse.che.api.core.ServerException;
+import org.eclipse.che.api.core.model.workspace.config.ProjectConfig;
+import org.eclipse.che.api.project.server.ProjectManager;
 import org.eclipse.che.api.project.server.handlers.ProjectInitHandler;
+import org.eclipse.che.plugin.java.languageserver.JavaLanguageServerExtensionService;
+import org.eclipse.che.plugin.java.languageserver.JavaLanguageServerLauncher;
 
 /**
  * Init handler for simple java project.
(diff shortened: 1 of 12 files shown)


tolusha (approved):

plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-server/src/main/java/org/eclipse/che/plugin/java/plain/server/projecttype/PlainJavaInitHandler.java
```diff
@@ -26,12 +35,41 @@
...
+      return;
+    }
+    String wsPath = absolutize(projectFolder);
+    ProjectConfig project =
+        projectRegistryProvider
+            .get()
+            .get(wsPath)
+            .orElseThrow(() -> new ServerException("Can't find a project: " + wsPath));
+
+    List<String> library = project.getAttributes().get(LIBRARY_FOLDER);
```
tsmaeder, line 71: If this property is set, shouldn't the classpath have been updated before? I'm not sure why we need this code.

plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/JavaLanguageServerLauncher.java
```diff
@@ -84,16 +87,24 @@ public JavaLanguageServerLauncher(
     this.notifyTransmitter = notifyTransmitter;
     this.eventService = eventService;
     this.projectManager = projectManager;
+
+    isStarted = new AtomicBoolean();
```
tsmaeder: Please use AtomicBoolean(false). While I would expect the default value to be false, I have to check the javadoc on AtomicBoolean to be sure.

tsmaeder (changes requested):

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/plainjava/RunPlainJavaProjectTest.java
```diff
@@ -53,7 +52,7 @@
 
 /** @author Aleksandr Shmaraev */
 public class RunPlainJavaProjectTest {
-  private static final String PROJECT_NAME = NameGenerator.generate("RunningPlainJavaProject", 4);
+  private static final String PROJECT_NAME = "run-plain-java-project";
```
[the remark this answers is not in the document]
svor: it resolves the problem about importing Simple project in jdt.ls workspace. Because `.project` file contains name of the project and here, when the name is generated, it isn't equal.

plugins/plugin-java/che-plugin-java-server/src/main/java/org/eclipse/che/plugin/java/languageserver/JavaLanguageServerLauncher.java
svor: ok

plugins/plugin-java/che-plugin-java-plain/che-plugin-java-plain-server/src/main/java/org/eclipse/che/plugin/java/plain/server/projecttype/PlainJavaInitHandler.java
svor: This code executes when the property was set and if this property exists we need to update classpath to add jars from this folder.

svor:
@tsmaeder PR was updated
@vparfonov pls take a look

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/plainjava/RunPlainJavaProjectTest.java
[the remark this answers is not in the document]
tsmaeder: Actually, the eclipse workspace project name should not make a difference, (i.e. we should not depend on it), but I guess since we're checking for log messages :-(

tsmaeder (approved):

Issue #11241, opened 2018-09-17
Remake the 'waitAndClick()' method from the 'SeleniumWebDriverHelper' according to the 'StaleElementReferenceException'


artaleks9:
### Description
* Sometimes selenium tests are failed (especially when they run on the CI) by the '_StaleElementReferenceException_'
* We need to remake the method _waitAndClick()_ in the _SeleniumWebDriverHelper_ to handle correctly  this exception and avoid fail the tests
* Also we need to remake two methods in the _Project Explorer_:
  - _waitAndSelectItem()_
  - _openItemByPath()_

using the handling of the _'StaleElementReferenceException'_ in the _SeleniumWebDriverHelper_:
* Related known issue: #11227

tsmaeder:
Do we understand why the stale element references occur?

Ohrimenko1988:
@tsmaeder The reason is - UI has been updated and web element reference is obsolete. So we are working on fixing this. You may see PR https://github.com/eclipse/che/pull/11246 which fixes this problem.

tsmaeder:
I understand the meaning of stale references. But do we understand why the references change when we expect them not to? That is the interesting questions, IMO.

Ohrimenko1988:
Done https://github.com/eclipse/che/pull/11246

PR #11246, opened 2018-09-17
[Selenium] Create method for performing specified functionality with ignoring provided exception

Ohrimenko1988:
### What does this PR do?
Create method for performing specified functionality with ignoring provided exception

For example: 
```
seleniumWebDriverHelper.waitPerformWithExceptionIgnoring(
()-> seleniumWebDriverHelper.waitAndClick(By.xpath("someXpath")),
StaleElementReferenceException.class)
```
method "waitAndClick" will be performed until success or timeout end. Besides if during performing the specified method will cause "StaleElementReferenceException" the method will be rerun, exception will be ignored and element will be found again.

Issue: https://github.com/eclipse/che/issues/11241


musienko-maxim (approved):

SkorikSergey (approved):

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/webdriver/WebDriverWaitFactory.java
```diff
@@ -43,6 +44,14 @@ public WebDriverWait get(int timeoutInSec) {
     return new WebDriverWait(seleniumWebDriver, timeoutInSec).withMessage(messageSupplier);
   }
 
+  public FluentWait<WebDriver> get(
+      int timeoutInSec, Class<? extends Throwable> ignoredExceptionType) {
+    return new FluentWait<WebDriver>(seleniumWebDriver)
+        .withTimeout(timeoutInSec, TimeUnit.SECONDS)
+        .pollingEvery(200, TimeUnit.MILLISECONDS)
```
dmytro-ndp: `.pollingEvery(200, TimeUnit.MILLISECONDS)` looks redundant so as default delay between the attempts is preset to 500 millisecs inside FluentWait itself.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/webdriver/WebDriverWaitFactory.java
```diff
@@ -43,6 +44,14 @@ public WebDriverWait get(int timeoutInSec) {
     return new WebDriverWait(seleniumWebDriver, timeoutInSec).withMessage(messageSupplier);
   }
 
+  public FluentWait<WebDriver> get(
```
dmytro-ndp: JavaDocs is needed.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/webdriver/SeleniumWebDriverHelper.java
```diff
@@ -69,6 +69,10 @@ public SeleniumWebDriverHelper(
     this.uploadUtil = uploadFileUtil;
   }
 
+  public interface VoidSupplier {
```
dmytro-ndp: We can get rid of `VoidSupplier` by using `Supplier<Void>`.

dmytro-ndp:
@Ohrimenko1988: please, take a look at small improvements here https://github.com/eclipse/che/pull/11246/commits/239ec30be60a2a1634943898b93251a183d90f59
Example of usage:
```
    seleniumWebDriverHelper.waitNoExceptions(
        new Supplier<Void>() {
          private int counter = 3;

          @Override
          public Void get() {
            if (counter-- > 0) {
              LoggerFactory.getLogger(DialogAboutTest.class).info("exception " + counter);
              WaitUtils.sleepQuietly(2);
              throw new StaleElementReferenceException("exception " + counter);
            }

            LoggerFactory.getLogger(DialogAboutTest.class).info("return null");
            return null;
          }
        },
        StaleElementReferenceException.class,
        20);
```

artaleks9 (approved):

tsmaeder:
To be honest, I don't think this is a good approach to make the ci tests more stable in the long run: we just paper over the fact that we don't understand why the StaleElementExceptions are happening. When wo do stuff the changes the elements in the page, we should wait after the change until the system has settled again. Not just continue and then wait the next time we want to work with the dom again.

Ohrimenko1988:
@tsmaeder The way which you described is only one approach, we can do by another way (like in this PR). And it will be used only in some places if compare with existing waits in this places, this PR approach works even faster.

dmytro-ndp:
ci-test

riuvshin:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/826/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/826//Selenium_20tests_20report/)
docker image: eclipseche/che-server:11246

dmytro-ndp (approved):

Issue #10881, opened 2018-08-22
Migrate Che Dashboard from bower to yarn


nickboldt:
### Description

Motivation:

* Bower is [deprecated](https://bower.io/blog/2017/how-to-migrate-away-from-bower/)/[dead](https://snyk.io/blog/bower-is-dead/)
* Bower has an https-only registry, which prevents us from building Che Dashboard on a server that's behind an http-only proxy (eg., #projectncl)
* [Yarn](https://yarnpkg.com/en/) is better, faster, newer, maintained.

### Reproduction Steps

When building behind an http-only proxy, bower install step fails like this:

```
    [exec] > che-dashboard@4.0.0 postinstall /tmp/eclipse-che-6.9-phantomjs-2.1.1-from-mirror/dashboard
     [exec] > npm run bower && npm run typings && npm run build
     [exec] 
     [exec] npm info it worked if it ends with ok
     [exec] npm verb cli [ '/opt/rh/rh-nodejs6/root/usr/bin/node',
     [exec] npm verb cli   '/opt/rh/rh-nodejs6/root/usr/bin/npm',
     [exec] npm verb cli   'run',
     [exec] npm verb cli   'bower' ]
     [exec] npm info using npm@3.10.9
     [exec] npm info using node@v6.11.3
     [exec] 
     [exec] > che-dashboard@4.0.0 bower /tmp/eclipse-che-6.9-phantomjs-2.1.1-from-mirror/dashboard
     [exec] > bower update
     [exec] 
     [exec] npm verb run-script [ 'prebower', 'bower', 'postbower' ]
     [exec] npm info lifecycle che-dashboard@4.0.0~prebower: che-dashboard@4.0.0
     [exec] npm info lifecycle che-dashboard@4.0.0~bower: che-dashboard@4.0.0

     [exec] bower                            retry Request to https://registry.bower.io/packages/angular-bootstrap failed with ECONNRESET, retrying in 1.5s
...
     [exec] bower jsonlint#1.6.0        ECONNRESET Request to https://registry.bower.io/packages/jsonlint failed: tunneling socket could not be established, cause=socket hang up


          [exec] 
     [exec] npm verb lifecycle che-dashboard@4.0.0~bower: unsafe-perm in lifecycle true
[...]

PR #11271, opened 2018-09-18
#10881 move dashboard to its own profile and assembly with dashboard to a separate file, so we can build Che Server w/o dashboard (until Bower -> Yarn migration done)

nickboldt:
https://github.com/eclipse/che/issues/10881 move dashboard to its own profile and assembly with dashboard to a separate file, so we can build Che Server w/o dashboard (until Bower -> Yarn migration done)

Change-Id: I2a035e8e0e1ca18fa9e2b65d512270b3e9939eb8
Signed-off-by: nickboldt <nboldt@redhat.com>

### DIFF
--- a/assembly/assembly-main/pom.xml
+++ b/assembly/assembly-main/pom.xml
@@ -72,11 +72,6 @@
             <type>jar</type>
         </dependency>
         <dependency>
-            <groupId>org.eclipse.che.dashboard</groupId>
-            <artifactId>che-dashboard-war</artifactId>
-            <type>war</type>
-        </dependency>
-        <dependency>
             <groupId>org.eclipse.che.docs</groupId>
             <artifactId>che-docs</artifactId>
             <type>war</type>
@@ -152,4 +147,41 @@
             </plugin>
         </plugins>
     </build>
+    <profiles>
+        <profile>
+            <id>dashboard</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-assembly-plugin</artifactId>
+                        <executions>
+                            <execution>
+                                <id>make-assembly</id>
+                                <phase>package</phase>
+                                <goals>
+                                    <goal>single</goal>
+                                </goals>
+                            </execution>
+                        </executions>
+                        <configuration>
+                            <descriptors>
+                                <descriptor>${project.basedir}/src/assembly/assembly-with-dashboard.xml</descriptor>
+                            </descriptors>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
+            <dependencies>
+                <dependency>
+                    <groupId>org.eclipse.che.dashboard</groupId>
+                    <artifactId>che-dashboard-war</artifactId>
+                    <type>war</type>
+                </dependency>
+            </dependencies>
+        </profile>
+    </profiles>
 </project>
--- a/assembly/assembly-main/src/assembly/assembly-with-dashboard.xml
+++ b/assembly/assembly-main/src/assembly/assembly-with-dashboard.xml
@@ -47,7 +47,7 @@
                 <include>org.eclipse.che:assembly-wsmaster-war</include>
             </includes>
         </dependencySet>
-        <!-- <dependencySet>
+        <dependencySet>
             <useProjectArtifact>false</useProjectArtifact>
             <unpack>false</unpack>
             <outputDirectory>tomcat/webapps</outputDirectory>
@@ -55,7 +55,7 @@
             <includes>
                 <include>org.eclipse.che.dashboard:che-dashboard-war</include>
             </includes>
-        </dependencySet> -->
+        </dependencySet>
         <dependencySet>
             <useProjectArtifact>false</useProjectArtifact>
             <unpack>false</unpack>
--- a/pom.xml
+++ b/pom.xml
@@ -36,7 +36,6 @@
         <module>ide/che-ide-core</module>
         <module>ide/che-ide-full</module>
         <module>ide/che-ide-gwt-app</module>
-        <module>dashboard</module>
         <module>workspace-loader</module>
         <module>assembly</module>
         <module>selenium</module>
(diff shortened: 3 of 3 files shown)


l0rd:
@nickboldt since it's something temporary why not using a separate branch? I mean you are doing this change just to verify if - dashboard excluded - everything builds fine, but you don't plan to permanently exclude the dashboard? 

Anyway don't you think that you will need a separate branch or even a separate repo sometime in the future?

nickboldt:
Well, it's been 4 weeks since I started the request to move the Dashboard to Yarn instead of Bower, with no end in sight. Including this now means that there's no future need to merge back from a branch. The profile is enabled by default. So for upstream codenvy Jenkins, no change should be seen. And for downstream productization build, I just need to add `-P!dashboard` to build.

So this is the cleanest option for me until @olexii4 can complete the migration away from Bower.

l0rd:
@nickboldt fair enough. I am +1 with that. Once it get merged it would be nice if you could update the wiki with a mention about this profile: https://github.com/eclipse/che/wiki/Development-Workflow

@benoitf @skabashnyuk @vparfonov any feedback?

benoitf:
well I'm more -1 on that. It is creating extra burden for nothing.
Also, if the build is run against a tag version of che, the artifacts are now available on central M2 so it means that dashboard will not be compiled but downloaded like any other maven dependency.

nickboldt:
@benoitf That's great if we only want to have a clean Rachel build once ever 3 weeks. But we want CI. And for CI I have to build against SNAPSHOT versions, which are not in Maven Central. 

So if we want engaged QE and frequent builds of the product, we need a way to build it from snapshot every day on every github commit. 

This profile allows that and provides ZERO cost to maintain (other than the fact that there are two asembly*.xml files now).

benoitf:
IMHO Rachel should be based on a tagged version of che, like Red Hat che.
Red Hat Che has snapshots but it is build on a tagged upstream release.
so you never build against snapshots.


you've written zero in uppercase but it has cost to maintain. You duplicated assembly XML file to a new one. So adding or removing an agent or anything from one assembly need to be ported over the other assembly. ==> it has cost.

Also you dropped dashboard from the main assembly instead of using another assembly "without" dashboard so it is breaking any tools relying on the current assembly.
==> it breaks

nickboldt:
> Also you dropped dashboard from the main assembly instead of using another assembly "without" dashboard so it is breaking any tools relying on the current assembly.
Yes, but the default build will enable the profile, so the main assembly will use the assembly-with-dashboard.xml as noted here: https://github.com/eclipse/che/pull/11271/files#diff-4c06d1de2892ec9f078d9036969ba103R172 Therefore nothing should break unless you have a test for "what's the filename used to do the assembly? ASSERT = assembly.xml, result = assembly-with-dashboard.xml"

If that's a real concern I can switch the files from assembly.xml (without) and assembly-with-dashboard.xml to assembly-without-dashboard.xml and assembly.xml (with).

> adding or removing an agent or anything from one assembly need to be ported over the other assembly. 
Yes, agreed - I did mention that having two files is a small cost. If having the two files is too costly, I could use a mojo to transform the assembly.xml rather than using a different file. Is that better? 

> so you never build against snapshots
If you want to get technical, the way NCL actually works is that it fetches from master, creates a new tag, and then builds from that tag so that it's reproduceable. I'm trying to enable us to do continuous productization, rather than being limited to only building against tags of the upstream Che repo, which only appear once every 3 weeks. One build every 3 weeks != CP, and that's impractical from a QE perspective.

benoitf:
is Rachel will include dashboard ? if yes, then this PR is useless.

nickboldt:
Yes, eventually, once @olexii4 successfully ports the Dashboard from Bower deps to use Yarn. 

But until then, being unable to build Che Server due to the Dashboard BLOCKS CREATION OF A PRODUCT. This UNBLOCKS that. 

If you don't like my harmless approach to creation of a profile, then please suggest another workaround. I've been waiting 1 month for a solution. This train is almost over. I need a fix ASAP. 

Calling my solution "useless" is not productive or helpful. Please suggest a better approach if you have one.

benoitf:
Product should be `complete` and then include the dashboard.
Mario suggested to use a custom branch. Also you can still use a tagged released where it will download dashboard from the central repository.
I don't see why "skipping" a module is better than "downloading it" as it is published like any other M2 dependency on central. 
I'm assuming that you're not recompiling all maven dependencies used in Che ?

nickboldt:
If this was a Brew build, then YES, I'd be recompiling everything. Using NCL bypasses that, but is limited in that we can't access 3rd party content over https via the Indy proxy. Support for https is coming early next year but that's too late for the Dec release. 

Gorkem has told me that for the first Alpha or Beta it's OK if the dashboard is missing as long as it comes back for GA. So temporarily removing it is a reasonable approach, considering it's been a month since I opened #10881 to migrate the Dashboard away from Bower and it's deprecated/security-hole dependencies.

As to building from a custom branch, that means that every time you push a new change to master, I will also have to **update my branch with the changes**. That introduces a MANUAL STEP, which is again the antithesis of Continuous (automated) Productization. If you merge this to master, then I get all the changes in master every time I rebuild in NCL. 

Why make it HARDER to automate? Why create more manual steps? 

As to the idea of building from master but with a tagged version of the dashboard, sure. That could work. But then I have to maintain either a new property for `-Dche.dashboard.version=6.11.0` (which is today not in che-parent's list of properties/dep version) while building 6.12.0-SNAPSHOT. And doing so could introduce bugs/incompatibilities which QE would have to explicitly check/test, as the product would now differ from the upstream. 

So if that's your best suggestion for how to work around this I can try that. IMHO it's less overhead than a custom branch for me, but more overhead for QE. I don't like it, but you're the gatekeeper here.

skabashnyuk:
@nickboldt do we have confirmation from @slemeur that Che has any value without dashboard in "your" case? How it would be used?

nickboldt:
Valid questions, @l0rd and @skabashnyuk. I'll wait for input from @slemeur. 

Also eager to see what's happened in https://github.com/eclipse/che/tree/CHE-10883 since I last looked. If we can merge THAT to master, we probably don't need this PR/workaround.

slemeur:
If there is no dashboard that can be only for internal purpose - that would not be a version we would distribute. 
I understand the goal here is to temporarily `unblock` the build process for the time being.
Correct?

nickboldt:
Yes, we can reject this suggestion and instead focus on the bower to yarn migration, and getting https proxy support in NCL/Indy.

Issue #11328, opened 2018-09-24
Don't fail when waiting for jdt.ls console messages


tsmaeder:
There is a suspicion that we have cases where the jdt.ls startup messages ("started", "workspace updated") are missed because the ide is starting too slowly and the message have been written before the IDE is showing the "workspace machine" console window.
Not failing the tests when we time out waiting for the would be more reliable. Writing out a message to the log would still allow to identify the case.

eivantsov:
Closing the issue since the PR is merged. Feel free to reopen though

PR #11334, opened 2018-09-24
Don't fail when jdt.ls messages are not found

tsmaeder:
### What does this PR do?
Instead of failing when initialization messages from jdt.ls are not found in the workspace machine console, just write out a warning message. The reaon is that we suspect that in the openshift runs of ci-tests, the ide is taking too long to start up and the console messages are never seen.

Don't fail when waiting for jdt.ls console messages #11328


tolusha (approved):

svor (approved):

dmytro-ndp:
FYI: warnings in the test logs are usually treated as annoying and being ignored.  You made verification useless so that they will not be reflected in test report, and they will be removed in near future.

tsmaeder:
@dmytro-ndp it's not verification, we need to wait until jdt.ls has indexed the project before running java related tests.

PR #11336, opened 2018-09-25
Selenium: update "YamlFileEditingTest" selenium test according to language servers test plan

SkorikSergey:
### What does this PR do?
Update "**YamlFileEditingTest**" selenium test according to language servers test plan.

https://github.com/eclipse/che/issues/11221


selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/languageserver/YamlFileEditingTest.java
```diff
@@ -60,6 +68,7 @@
           + "    app: che\n"
           + "    template: che\n"
           + "  name: che";
+  private List<String> symbols = Arrays.asList("apiVersionsymbols (194)", "kind", "metadata");
```
dmytro-ndp: `symbols` variable looks like a constant, and "apiVersionsymbols (194)" is more then one symbol.
So, I would recommend to remake variable `symbols` to constant `EXPECTED_GO_TO_SYMBOL_ALTERNATIVES`

dmytro-ndp (approved):

artaleks9 (approved):

PR #11360, opened 2018-09-26
Suspend handling of file changed notifications when applying ws edit

### LINKED ISSUE #11359: Renamed Open Files Are Closed
We have intermittent failures in Selenium tests when refactoring classes and then checking that the class is open in an editor (for example, in TestEnumerationsTest). The suspected reason for this is that the class EditorFileStatusNotificationOperation is closing the editor because it gets the "delete" notification from a rename or move and closes the editor before ApplyWorkspaceEditAction is able to adjust the editor to show the new File.
The proposed fix would be to suspend EditorFileStatusNotificationOperation while the workspace edit is being applied.

tsmaeder:
### What does this PR do?
Suspends EditorFileStatusNotificationOperation while applying a WorkspaceEdit.  See issue for why.

Renamed Open Files Are Closed #11359

### DIFF
--- a/plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/quickassist/ApplyWorkspaceEditAction.java
+++ b/plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/quickassist/ApplyWorkspaceEditAction.java
@@ -52,6 +52,7 @@ import org.eclipse.che.ide.api.resources.Container;
 import org.eclipse.che.ide.api.resources.Resource;
 import org.eclipse.che.ide.api.resources.VirtualFile;
 import org.eclipse.che.ide.dto.DtoFactory;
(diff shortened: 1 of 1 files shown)


plugins/plugin-languageserver/che-plugin-languageserver-ide/src/main/java/org/eclipse/che/plugin/languageserver/ide/editor/quickassist/ApplyWorkspaceEditAction.java
```diff
@@ -413,7 +429,8 @@ public static void applyTextEdits(Document document, List<TextEdit> edits) {
       Position end = r.getEnd();
       int startIndex =
           document.getIndexFromPosition(new TextPosition(start.getLine(), start.getCharacter()));
-      // python ls, for example shows as end position index 0 of the line after the change. If the
+      // python ls, for example shows as end position index 0 of the line after the
+      // change. If the
       // change is on the last line, we crash
```
svor: It is better to combine 433 and 434 lines

svor (approved):

PR #11366, opened 2018-09-26
Login to oc right after init

Katka92:
### What does this PR do?
Login to oc right after the class OpenShiftCliCommandExecutor is initialized. As far as this class is accessed in test via HotUpdateUtil, we faced cases that user was not logged in. There is no method for login via HotUpdateUtil. 
When the OpenShiftCliCommandExecutor is binded, it will need access to oc to provide execute and other methods so it should not be problem to log in right after init and then don't care about that later. 

https://github.com/redhat-developer/che-functional-tests/issues/365


riuvshin:

riuvshin:

riuvshin:

dmytro-ndp:
Log into _OpenShift_ is needed only for several tests of _Eclipse Che_, and `OpenShiftCliCommandExecutor` class doesn't need an access to oc just after it is bound. Even more - there are some commands which don't require being logged into OpenShift before execution, like `oc version`.

At the same time login operation is quite expensive because it requires downloading of _OpenShift ClI_ client into the system, and we would avoid it if we run separate test which is not bound to OpenShift, like `DialogAboutTest`.

If this PR is aimed to get rid of calling `OpenShiftCliCommandExecutor#login()` method just before `OpenShiftCliCommandExecutor#execute()`, I would propose just to rewrite `OpenShiftCliCommandExecutor#execute()` method as follow:
```
...
  private boolean isLoggedIn;
  private ReentrantLock loginLock = new ReentrantLock();
...

  @Override
  public String execute(String command) throws IOException {
    return execute(command, true);
  }

  private String execute(String command, boolean needToLogin) throws IOException {
    if (!PATH_TO_OPENSHIFT_CLI.toFile().exists()) {
      downloadOpenShiftCli();
    }

    if (needToLogin && !isLoggedIn) {
      loginLock.lock();
      try {
        login();
        isLoggedIn = true;
      } catch (IOException e) {
        isLoggedIn = false;
        throw e;
      } finally {
        loginLock.unlock();
      }
    }

    String openShiftCliCommand = format("%s %s", PATH_TO_OPENSHIFT_CLI, command);

    return processAgent.process(openShiftCliCommand);
  }

  /** Logs into OpenShift as a regular user */
  public void login() throws IOException {
    String loginToOpenShiftCliCommand;
    if (openShiftToken != null) {
      loginToOpenShiftCliCommand =
          format(
              "login --server=%s --token=%s --insecure-skip-tls-verify",
              openShiftWebConsoleUrlProvider.get(), openShiftToken);
    } else {
      loginToOpenShiftCliCommand =
          format(
              "login --server=%s -u=%s -p=%s --insecure-skip-tls-verify",
              openShiftWebConsoleUrlProvider.get(),
              openShiftUsername != null ? openShiftUsername : DEFAULT_OPENSHIFT_USERNAME,
              openShiftPassword != null ? openShiftPassword : DEFAULT_OPENSHIFT_PASSWORD);
    }

    execute(loginToOpenShiftCliCommand, false);
  }
```


Don't also forget to remove calling of `login()` method from the tests so as it will become unnecessary.

Katka92:
@dmytro-ndp Thank you very much for clarification! I agree that this would be better approach. Just a question - why should the default value of ```needToLogin``` be true? I would suggest it set to false - I think that it would be used much frequently as logged or without login at all - so I would say that false is better option.

```
@Override
  public String execute(String command) throws IOException {
    return execute(command, true);
  }
```
Regarding to your last sentence:
```
Don't also forget to remove calling of login() method from the tests so as it will become unnecessary.
``` 
This login is used only in one method ```OpenShiftKeycloakCliCommandExecutor#obtainKeycloakPodName()```. So I think if default value for login in execute() method is false, I can leave it here as it is.

Katka92:
Sorry I clicked wrong button by accident...

dmytro-ndp:
Looks good as for me. Let's execute the tests: ci-test

dmytro-ndp:
ci-test

dmytro-ndp (approved):
Local test execution against Eclipse Che Multiuser on OCP didn't show regression.

PR #11370, opened 2018-09-26
Provide a way to annotate rest endpoint functions to enable Machine Auth

davidfestal:
### What does this PR do?

This PR allows extends the newly-added mechanism that limits the scope of allowed operations with machine token to workspace related ones only.

We should be able to provide, in a custom assembly for example, additional wsmaster endpoints that would be considered as workspace-related.
This is a strong requirement to enable updating RhChe to upstrea Che 6.12.0 version.
For more details about the RhChe case, see [this comment](https://github.com/redhat-developer/rh-che/pull/902#issuecomment-424645862)

The proposed solution defines a `MachineAuthorized` annotation that can be added on an new REST endpoint method, so that it would support Machine-token-based authentication coming from agent requests. 

This is a pre-requisite for issue https://github.com/redhat-developer/rh-che/issues/927

### Was this tested ?

This was tested in the context of RhChe built on `6.12.0-SNAPSHOT` upstream version

### DIFF
--- a/multiuser/machine-auth/che-multiuser-machine-authentication/src/main/java/org/eclipse/che/multiuser/machine/authentication/server/MachineAuthorized.java
+++ b/multiuser/machine-auth/che-multiuser-machine-authentication/src/main/java/org/eclipse/che/multiuser/machine/authentication/server/MachineAuthorized.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright (c) 2012-2018 Red Hat, Inc.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.multiuser.machine.authentication.server;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Declares that a method can be invoked using machine token signed requests.
+ *
+ * @author David Festal (dfestal@redhat.com)
+ */
+@Target(ElementType.METHOD)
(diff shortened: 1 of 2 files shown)


ibuziuk:
[ci-test]

ibuziuk:
ci-test

ibuziuk (approved):

skabashnyuk:
@davidfestal please do not merge this pr.  We are discussing with @mshaposhnik @sleshchenko some alternative way to configure a list of services/methods. We would like to make services independent from authentification/authorization. I hope we will provide some details today.

davidfestal:
@skabashnyuk BTW what do you think of the annotation approach. That seems quite secured, no ? That means that no one can change the permissions of an upstream endpoint in a customized assembly, which seems rather important. wdyt ?

skabashnyuk:
@davidfestal not really. I do like the separation between services and filters and the fact that they should not know about each other. What I see is that we have low flexibility in a configuration of KeycloakAuthenticationFilter and MachineTokenAccessFilter. I think we should provide an ability to easily customize this filters without the need dig into MethodInterceptor, etc.

skabashnyuk:
@davidfestal FYI https://github.com/eclipse/che/issues/11374

skabashnyuk:
Yes sure it's easy from one hand from another hand it breaks distinctions between filter and service that I would like to keep. 
For example, WorkspaceService should know about machine tokens, authentification/authorization etc.

FYI @mshaposhnik has some prototypes for https://github.com/eclipse/che/issues/11374 I hope we will have some solution today/tomorrow. It's supposed to be based on configuration property/ environment variable.

skabashnyuk (changes requested):
Please wait for https://github.com/eclipse/che/issues/11374

riuvshin:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/875/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/875//Selenium_20tests_20report/)
docker image: eclipseche/che-server:11370

musienko-maxim:

vparfonov:
not actual

Issue #11374, opened 2018-09-27
 MachineTokenAccessFilter binding customization


skabashnyuk:
### Description
In this place https://github.com/eclipse/che/pull/11370#issuecomment-424984870 we met situation when developers of custom Che assembly are struggling with zero ability to customize binding of  
KeycloakAuthenticationFilter and MachineTokenAccessFilter. The goal of this issue is to provide a better way to configure this filters.

mshaposhnik:
Done for machine auth, will create separate issue for Keykloak Auth

PR #11377, opened 2018-09-27
Make methods accessible with machine token configurable via property.

mshaposhnik:
### What does this PR do?
Make methods accessible with machine token configurable via property instead of hardcoding.

https://github.com/eclipse/che/issues/11374

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/multiuser.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/multiuser.properties
@@ -164,8 +164,8 @@ che.keycloak.username_claim=NULL
 che.oauth.service_mode=delegated
 
 
-#####                             MACHINE AUTH CONFIGURATION                  #####
+#####                        MACHINE AUTH CONFIGURATION                  #####
 
 #     Describes set of methods which can be accessible using machine token.
-#     Format is semicolon separated tokens of form: `/service path=comma_separated_methods_list;`
+#     Format is semicolon separated tokens of form: `/service_path=comma_separated_methods_list;`
 che.multiuser.machine_authentication_allowed_methods=/workspace=getByKey,addProject,updateProject,deleteProject;/ssh=getPair,generatePair;/factory=getFactoryJson,getFactory,getFactoryByAttribute,resolveFactory;/preferences=find;/activity=active
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/multiuser.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/multiuser.properties
@@ -162,3 +162,10 @@ che.keycloak.username_claim=NULL
 #     If set to "delegated", then the service will use Keycloak IdentityProvider mechanism.
 #     Runtime Exception wii be thrown, in case if this property is not set properly.
 che.oauth.service_mode=delegated
+
+
+#####                             MACHINE AUTH CONFIGURATION                  #####
+
+#     Describes set of methods which can be accessible using machine token.
+#     Format is semicolon separated tokens of form: `/service path=comma_separated_methods_list;`
+che.multiuser.machine_authentication_allowed_methods=/workspace=getByKey,addProject,updateProject,deleteProject;/ssh=getPair,generatePair;/factory=getFactoryJson,getFactory,getFactoryByAttribute,resolveFactory;/preferences=find;/activity=active
--- a/multiuser/integration-tests/che-multiuser-cascade-removal/src/test/java/org/eclipse/che/multiuser/integration/jpa/cascaderemoval/JpaEntitiesCascadeRemovalTest.java
+++ b/multiuser/integration-tests/che-multiuser-cascade-removal/src/test/java/org/eclipse/che/multiuser/integration/jpa/cascaderemoval/JpaEntitiesCascadeRemovalTest.java
@@ -102,7 +102,7 @@ import org.eclipse.che.multiuser.api.permission.server.PermissionCheckerImpl;
 import org.eclipse.che.multiuser.api.permission.server.PermissionsManager;
 import org.eclipse.che.multiuser.api.permission.server.model.impl.AbstractPermissions;
 import org.eclipse.che.multiuser.api.permission.server.spi.PermissionsDao;
-import org.eclipse.che.multiuser.machine.authentication.server.MachineAuthModule;
+import org.eclipse.che.multiuser.machine.authentication.server.signature.jpa.JpaSignatureKeyDao;
 import org.eclipse.che.multiuser.machine.authentication.server.signature.spi.SignatureKeyDao;
 import org.eclipse.che.multiuser.organization.api.OrganizationJpaModule;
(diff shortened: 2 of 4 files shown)


skabashnyuk:
@davidfestal fyi

skabashnyuk (approved):

mshaposhnik:
ci-test

riuvshin:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/885/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/885//Selenium_20tests_20report/)
docker image: eclipseche/che-server:11377

mshaposhnik:
ci-test

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/multiuser.properties
```diff
@@ -162,3 +162,10 @@ che.keycloak.username_claim=NULL
 #     If set to "delegated", then the service will use Keycloak IdentityProvider mechanism.
 #     Runtime Exception wii be thrown, in case if this property is not set properly.
 che.oauth.service_mode=delegated
+
+
+#####                        MACHINE AUTH CONFIGURATION                  #####
+
+#     Describes set of methods which can be accessible using machine token.
+#     Format is semicolon separated tokens of form: `/service_path=comma_separated_methods_list;`
+che.multiuser.machine_authentication_allowed_methods=/workspace=getByKey,addProject,updateProject,deleteProject;/ssh=getPair,generatePair;/factory=getFactoryJson,getFactory,getFactoryByAttribute,resolveFactory;/preferences=find;/activity=active
```
davidfestal, line 171: So I assume that, in RhChe, we will have to override this property.

I'm a bit worried about this approach, since **what will happen when you add a new wsmaster endpoint in the upstream default property, and a customized assembly won't realize it and don't update its overriden property** ?
Custom assembly will somewhat fail, but possibly will start and then **behave erroneously only when accessing to the given endpoint,** which would be quite hard to troubleshoot...

2nd point is that it gives the ability to **enable specific upstream endpoints that were not intended to be enabled in upstream Che,** which seems a bit dangerous to me. The only need we have is to enable **new endoints that our custom assembly adds** 

On the contrary the annotation based approach proposed here https://github.com/eclipse/che/pull/11370 doesn't present these 2 drawbacks.
And we would not need to call these annotations according to filter names. This annotation is just a declarative / local configuration annotation as are the `@Path` or `@GET`.

@mario @ibuziuk wdyt about that ?

davidfestal (changes requested):

riuvshin:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/886/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/886//Selenium_20tests_20report/)
docker image: eclipseche/che-server:11377

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/multiuser.properties
skabashnyuk: I will try to explain my motivation. maybe it will allow us easier to find consensus.
1. All solutions have some cons.
2. "what will happen when you add a new wsmaster endpoint in the upstream default property, and a customized assembly won't realize it and don't update its overridden property" - that is not a problem for this PR. Because machine token will not work unless it's not configured.
3. This is a zero-coding solution. With common for all our code configuration approach.
4. It has no issue to add or remove anything in downstream assembly.
5. It's not mixing services and auth/autz. Like permissions decoupled from services.


The only problem I see (that is not exists in  @garagatyi's multi-binder approach ) that in case if an upstream configuration has been changed you need to change a downstream configuration too.  However, multi-binder has cons: you can't easily remove already configured value.

skabashnyuk: >It would help if @skabashnyuk could explain his concerns about that.

We are not managing authorization in services. Annotations are doing that indirectly.

Ok. If removing something is not a problem for downstream or even we would like to remove this ability - then I think @garagatyi 's approach is win-win for all cases.

garagatyi: @davidfestal Personally I don't believe in making project absolutely secure without allowing downstream to break something. When you need to implement something and framework doesn't allow you, you have to use hummer instead of a screwdriver. I don't want to make ppl use hummer to override what can't be overridden because authors of a framework wanted to forbid that.

mshaposhnik: OK there was my previous proposal here:
https://github.com/eclipse/che/compare/machine_auth_methods
is this approach ok fr you @davidfestal ?

davidfestal: @mshaposhnik I think it's better yes.

But I would even suggest adding adding directly the `Method.class` object into the `MachineAuthenticatedResource` instead of the method name, so that you wouldn't even need the service name, since you can get it from the method class `@Path` annotation.

I would also check on the method class package, and allow only additions by custom assemblies packages for methods that are not in `eclipse.che.*` packages.

This way we would have the best of both worlds...

mshaposhnik: Can you show example binding using `Method` ?

mshaposhnik: That's basically not very good idea because someone may override services and methods. And authentication may fail then. So that is why we choosed paths and method names (if you override the service you probably will not change it's path)

davidfestal: > someone may override services and methods

I understand your point, but would be interested in knowing how it is possible to override REST endpoint sevrice endpoints / methods, which I was not able to do until now according to how Service endpoints are bound in upstream generally.

But I don't want to pause this work any longer, so let's keep  your `MachineAuthenticatedResource`-based proposal as is :-)

davidfestal: > How will you bind methods?

`Class.getDeclaredMethod()` ? else use `Class` + method name.

But if you don't want to restrict overriding of the `eclipse.che.*` endpoints, then sure, it's not useful.

mshaposhnik: REST overriding example in Codenvy:
https://github.com/codenvy/codenvy/blob/master/wsmaster/codenvy-hosted-sso-oauth/src/main/java/com/codenvy/auth/sso/oauth/SsoOAuthAuthenticationService.java
Absolutely nothing special here.

Binding methods will require full parameter list, which can be huge:
`WorkspaceService.class.getDeclaredMethod("updateProject", String.class, String.class, ProjectConfigDto.class);`
and will fail with overrides. 
So I prefer to leave path and method name.

Issue #11223, opened 2018-09-14
Investigate the failing selenium test 'GitPullTest' on CI job


artaleks9:
### Description
* In the last time the selenium test _GitPullTest_ is failed when run on the CI job via the _'stale element reference'_
* Need to find the solution for it and fix

**See the report:**
https://ci.codenvycorp.com/view/qa-nightly/job/che-integration-tests-master-ocp/432/Selenium_20tests_20report/failsafe-report.html#org.eclipse.che.selenium.gitGitPullTest

artaleks9:
* The problem with a _'stale element reference'_ was resolved by the PR (#11389), but now we have another failure: _'Expected condition failed: waiting for element to no longer be visible'_.
* So need to continue investigate, find reason and fix it.
* See the report:
https://ci.codenvycorp.com/view/qa/job/che-integration-tests-multiuser-master-ocp/377/Selenium_20tests_20report/failsafe-report.html#org.eclipse.che.selenium.gitGitPullTest

artaleks9:
We have the bug in the product (see #11648), so the issue is closed.

PR #11389, opened 2018-09-27
Correct the 'GitPullTest' related to unexpected stale element reference

artaleks9:
### What does this PR do?
* Correct the selenium test _GitPullTest_ related to unexpected _'stale element reference'_ when the test runs on CI 
* We check that the expected text is not present in the editor after remove it on GitHub side, and performing git pull.
In fact we delete the folder with test files and as result after git pull these files are deleted from the editor and the project explorer.
On my opinion, we should check that test files are really delete from editor (the tab file names are not present), and to exclude the check of text unpresent in the editor.

#11223

### DIFF
--- a/selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/git/GitPullTest.java
+++ b/selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/git/GitPullTest.java
@@ -79,6 +79,7 @@ public class GitPullTest {
     prepareFilesForTest(jsFileName);
     prepareFilesForTest(htmlFileName);
     prepareFilesForTest(folderWithPlainFilesPath + "/" + readmeTxtFileName);
+    prepareFilesForTest(folderWithPlainFilesPath + "/README.md");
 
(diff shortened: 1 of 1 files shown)


SkorikSergey (approved):

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/git/GitPullTest.java
```diff
@@ -79,6 +79,7 @@ public void pullTest() throws Exception {
     prepareFilesForTest(jsFileName);
     prepareFilesForTest(htmlFileName);
     prepareFilesForTest(folderWithPlainFilesPath + "/" + readmeTxtFileName);
+    prepareFilesForTest(folderWithPlainFilesPath + "/README.md");
```
dmytro-ndp: Let's create the constant for that name as well:
```
private final static String README_MD_FILE_NAME = "README.md";
```

dmytro-ndp (approved):

Issue #11259, opened 2018-09-18
Do not allow the start of a workspace that has both installers and WS.NEXT plugins


garagatyi:
### Description
Che 6 and Che 7 workspace have a different design and should not be mixed. We need to throw an error when a user tries to start a workspace with such a configuration. 
We can also do this on workspace creation and editing, so we would not have a broken WS config stored.

### Reproduction Steps

**OS and version:**    

**Diagnostics:**

PR #11392, opened 2018-09-27
Fail workspace validation if config contains both ws.next plugins and installers

amisevsk:
### What does this PR do?
Checks if a workspace config contains both ws.next config options (editor, plugins) and installers during validation.

One potential shortcoming of the way this is currently implemented is that validation only occurs on creation/modification, so existing workspace with the issue will continue to launch. @garagatyi is this something we should also fix, or no?

https://github.com/eclipse/che/issues/11259

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceValidator.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceValidator.java
@@ -16,9 +16,13 @@ import static java.lang.String.format;
 import static org.eclipse.che.api.core.model.workspace.config.MachineConfig.MEMORY_LIMIT_ATTRIBUTE;
 import static org.eclipse.che.api.core.model.workspace.config.MachineConfig.MEMORY_REQUEST_ATTRIBUTE;
 
+import com.google.common.base.Strings;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.function.Function;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 import javax.inject.Singleton;
 import org.eclipse.che.api.core.ServerException;
 import org.eclipse.che.api.core.ValidationException;
@@ -47,6 +51,9 @@ public class WorkspaceValidator {
 
   private static final Pattern VOLUME_NAME = Pattern.compile("[a-z][a-z0-9]{1,18}");
   private static final Pattern VOLUME_PATH = Pattern.compile("/.+");
+
+  static final String PLUGINS_ATTRIBUTE = "plugins";
+  static final String EDITOR_ATTRIBUTE = "editor";
 
   /**
    * Checks whether given workspace configuration object is in application valid state, so it
@@ -100,6 +107,9 @@ public class WorkspaceValidator {
 
     // projects
     // TODO
+
+    // ensure using either plugins or installers but not both
+    validatePlugins(config);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


amisevsk:
ci-test

riuvshin:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/887/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/887//Selenium_20tests_20report/)
docker image: eclipseche/che-server:11392

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceValidator.java
```diff
@@ -166,6 +176,35 @@ private void validateLongAttribute(
...
+            .stream()
+            .map(
+                env ->
+                    env.getMachines()
+                        .values()
+                        .stream()
+                        .map(MachineConfig::getInstallers)
+                        .flatMap(List::stream))
+            .flatMap(Function.identity())
+            .collect(Collectors.toList());
```
sleshchenko: I think you don't need to collect all items and `findAny` will be enough here

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceValidator.java
```diff
@@ -166,6 +176,35 @@ private void validateLongAttribute(
     }
   }
 
+  /**
+   * Check that workspace has either plugins defined in attributes (Che 7) or installers defined in
+   * machines (Che 6).
+   *
+   * @throws ValidationException
```
sleshchenko: Please add few words when this exception can be thrown

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceValidator.java
garagatyi: Yeah, it can be simpler a bit. 
```
config
            .getEnvironments()
            .values()
            .stream()
            .flatMap(env -> env.getMachines().values().stream()) //convert stream of envs to stream of machines
            .flatMap(machine -> machine.getInstallers())// convert to a stream of installers
            .blahblah() // find if there is at least 1 entry 
```
And also:
```
wsNext = attributes.containsKey(PLUGINS_ATTRIBUTE) || attributes.containsKey(EDITOR_ATTRIBUTE) 
```

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceValidator.java
```diff
@@ -48,6 +52,9 @@
   private static final Pattern VOLUME_NAME = Pattern.compile("[a-z][a-z0-9]{1,18}");
   private static final Pattern VOLUME_PATH = Pattern.compile("/.+");
 
+  static final String PLUGINS_ATTRIBUTE = "plugins";
```
garagatyi: It might be better to use existing constants https://github.com/eclipse/che/blob/174e1aea9b5fc68815da7e7a485b6ff51b38ba15/wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java#L91

garagatyi (approved):
Looks good!

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceValidator.java
amisevsk: Thanks for the installers advice, I'll make that change. For the attributes, containsKey won't work because if e.g. a user toggles plugins on and off while creating a workspace it will add the attribute `"plugins"` with an empty key, and this can't be fixed during workspace creation.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceValidator.java
amisevsk: Ah didn't see, thanks. Will fix.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceValidator.java
amisevsk: Fixed

PR #11409, opened 2018-09-28
CHE-11408: Disable probes of Che master when debug is enabled

### LINKED ISSUE #11408: Disable liveness and readiness probes in Che deployment when deploy_che.sh is called with --debug option
### Description
`deploy_che.sh` script has an option: `--debug`. This option allows debugging Che master by setting needed ports, k8s services, Che env var, etc.
But when I do actually debug Che I hit a point of time when Che master gets restarted by k8s/OS because it doesn't respond to liveness/readiness probes for quite a long time. This is especially painful when I dig into some very hard and unclear problem with Che. This leads to more complicated debugging flow. Instead of that I just remove readiness/liveness probes from the deployment.
We can optimize this by doing that automatically when `--debug` option is used.

garagatyi:
### What does this PR do?
Remove readiness/liveness probes from Che deployment when debug is
enabled.
This allows debugging Che master for a long period of time without
Che container being killed because of unresponsive probes.

Fixes #11408

### DIFF
--- a/deploy/openshift/deploy_che.sh
+++ b/deploy/openshift/deploy_che.sh
@@ -336,6 +336,8 @@ if [ "${CHE_DEBUG_SERVER}" == "true" ]; then
   ${OC_BINARY}  expose service che-debug
   NodePort=$(oc get service che-debug -o jsonpath='{.spec.ports[0].nodePort}')
   printInfo "Remote wsmaster debugging URL: ${CLUSTER_IP}:${NodePort}"
+  printInfo "Removing liveness and readiness probes from Che deployment"
+  oc set probe dc/che --remove --readiness --liveness
 fi
 }
 


skabashnyuk (approved):

deploy/openshift/deploy_che.sh
```diff
@@ -336,6 +336,8 @@ if [ "${CHE_DEBUG_SERVER}" == "true" ]; then
   ${OC_BINARY}  expose service che-debug
   NodePort=$(oc get service che-debug -o jsonpath='{.spec.ports[0].nodePort}')
   printInfo "Remote wsmaster debugging URL: ${CLUSTER_IP}:${NodePort}"
+  printInfo "Removing liveness and rediness probes from Che deployment"
```
benoitf: typo on rediness

garagatyi: tnx

l0rd (approved):

sleshchenko (approved):
LGTM

Issue #11445, opened 2018-10-02
Runtimes recovery process stops on first error


amisevsk:
### Description
In [`WorkspaceRuntimes#recover()`](https://github.com/eclipse/che/blob/ab70c1485604c98ec5ba96faa57ce20c69f5f4ff/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java#L528), the loop over infra identities is inside the try block. If any exception occurs, an error will be logged and the method will return. This means that the recovery process can skip workspaces where there is no error.

Instead, the try/catch block should be inside the loop, to allow the recovery process to continue even if some workspaces cannot be recovered. This would also log *all* workspaces that cannot be recovered rather than the first.

Skipping workspaces after the first error also can have the effect of leaving workspace resources in an untracked and uncleaned state, which we've seen running rh-che on OpenShift (Deployments are left running while Che lists workspace as "stopped").

### Reproduction Steps
Error can occur if Che server is restarted for any reason; if any exception is thrown while recovering runtimes, remaining runtimes are not recovered. Server logs show e.g.
```
    October 2nd 2018, 04:36:35.628    Successfully recovered workspace runtime 'workspacetymylu68u6l6qwgp'
    October 2nd 2018, 04:36:35.751    Successfully recovered workspace runtime 'workspacegj73ipi3nn3jjm3e'
    October 2nd 2018, 04:36:35.903    Successfully recovered workspace runtime 'workspaceop853grq9fo01pb7'
    October 2nd 2018, 04:36:36.055    Successfully recovered workspace runtime 'workspace37t5fhzb9erijycf'
    October 2nd 2018, 04:36:36.132    An error occurred while attempted to recover runtimes using infrastructure 'openshift'. Reason: 'Couldn't recover runtime 'workspacefdz4vn59o1myqi1a:default'. Error: The anonymous subject is used, and won't be able to perform this action'
[no more recovered messages]
```

garagatyi:
good catch!

amisevsk:
Closing as PR is merged.

PR #11446, opened 2018-10-02
Don't stop workspace recovery when error is encountered

amisevsk:
Fix issue where encountering an exception while recovering workspaces on startup prevents recovery of further workspaces.

### What does this PR do?
Fix issue where encountering an exception while recovering workspaces on startup prevents recovery of further workspaces.

https://github.com/eclipse/che/issues/11445

### Testing
1. Start Che multiuser on minishift
2. Start a few workspaces
3. Scale Che server down to 0
4. modify postgres database to set `owner_id` to 0 for some workspace in table `che_k8s_runtime`
5. restart Che server
logs contain something like:
```
2018-10-02 20:41:56,358[ost-startStop-1]  [ERROR] [o.e.c.a.w.s.WorkspaceRuntimes 544]   - An error occurred while attempted to recover runtimes using infrastructure 'openshift'. Reason: 'Couldn't recover runtime 'workspace3ob15dnpujw9eamy:default'. Error: The user `0` doesn't have the required `use` permission for workspace `workspace3ob15dnpujw9eamy`'
2018-10-02 20:41:57,647[ost-startStop-1]  [INFO ] [o.e.c.a.w.s.WorkspaceRuntimes 590]   - Successfully recovered workspace runtime 'workspacekvgznl4cdef4u8r7'
2018-10-02 20:41:57,739[ost-startStop-1]  [INFO ] [o.e.c.a.w.s.WorkspaceRuntimes 590]   - Successfully recovered workspace runtime 'workspacec7tt9hc681m1uvr7'
```
and workspaces that can be recovered are. Workspace that had `owner_id` modified is listed as "not running" in Che, and deployment is not cleaned up, but there's not much we can do there.

Additionally, this will allow logging of all recovery errors, so this could be used to better track recovery bugs.

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
@@ -531,7 +531,7 @@ public class WorkspaceRuntimes {
       return;
     } catch (InfrastructureException e) {
       LOG.error(
-          "An error occurred while attempted to recover runtimes using infrastructure '{}'. Reason: '{}'",
+          "An error occurred while attempting to get runtime identities for infrastructure '{}'. Reason: '{}'",
           infrastructure.getName(),
           e.getMessage());
       return;
@@ -542,7 +542,7 @@ public class WorkspaceRuntimes {
         recoverOne(infrastructure, identity);
       } catch (ServerException | ConflictException e) {
         LOG.error(
-            "An error occurred while attempted to recover runtimes using infrastructure '{}'. Reason: '{}'",
+            "An error occurred while attempting to recover runtimes using infrastructure '{}'. Reason: '{}'",
             infrastructure.getName(),
             e.getMessage());
       }
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
@@ -523,23 +523,29 @@ public class WorkspaceRuntimes {
       LOG.warn("Recovery of the workspaces is rejected.");
       return;
     }
+    Set<RuntimeIdentity> identities;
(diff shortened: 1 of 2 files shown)


wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -523,23 +523,29 @@ void recover() {
...
-      LOG.error(
-          format(
-              "An error occurred while attempted to recover runtimes using infrastructure '%s'",
-              infrastructure.getName()),
-          x);
-    } catch (ConflictException | ServerException | InfrastructureException x) {
+      return;
+    } catch (InfrastructureException e) {
       LOG.error(
           "An error occurred while attempted to recover runtimes using infrastructure '{}'. Reason: '{}'",
```
sleshchenko: Le'ts rephrase this message to make it clear that error occurred during identities fetching.

sleshchenko (approved):
Good fix!

You can make this PR even better by adding a unit test that second workspace will be recovered if the first one failed.

skabashnyuk (approved):

garagatyi (commented):
Looks good! Please, add tests.

akorneta (approved):

sleshchenko:
ci-test

garagatyi:
@riuvshin I cannot express how I happy with `just` 3-hours long tests! It is so good to receive the feedback on the same day. THANK YOU!

riuvshin:
@garagatyi this particular run `Took 2 hr 42 min` :P I've enabled mvn parallel build which decreased execution time for 20mins

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
amisevsk: Done!

amisevsk:
I've added a test case to check that recovery continues past error.

amisevsk:
ci-test

garagatyi (approved):

PR #11586, opened 2018-10-11
Add abstraction for TestWorkspaceInjector.

Katka92:
### What does this PR do?
Adds an abstraction of TestWorkspaceInjector. Now we can reuse the most of code in Rh-che tests. 

https://github.com/redhat-developer/che-functional-tests/issues/340


riuvshin:

riuvshin:

riuvshin:

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/workspace/AbstractTestWorkspaceInjector.java
```diff
@@ -0,0 +1,86 @@
...
+import com.google.inject.name.Names;
+import java.lang.reflect.Field;
+import org.eclipse.che.selenium.core.user.DefaultTestUser;
+
+/**
+ * Injector for custom annotation {@link InjectTestWorkspace}.
+ *
+ * @author Katerina Foniok
+ */
+abstract class AbstractTestWorkspaceInjector<T> implements MembersInjector<T> {
```
dmytro-ndp: Let's make this class `public` to simplify the usage.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/workspace/AbstractTestWorkspaceInjector.java
```diff
@@ -0,0 +1,86 @@
...
+import com.google.inject.MembersInjector;
+import com.google.inject.Provider;
+import com.google.inject.name.Names;
+import java.lang.reflect.Field;
+import org.eclipse.che.selenium.core.user.DefaultTestUser;
+
+/**
+ * Injector for custom annotation {@link InjectTestWorkspace}.
+ *
+ * @author Katerina Foniok
```
dmytro-ndp: Please, add `Anatolii Bazko` as another author of code as well.

selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/workspace/TestWorkspaceInjector.java
```diff
@@ -11,57 +11,29 @@
...
-          isNullOrEmpty(injectTestWorkspace.user())
-              ? injector.getInstance(DefaultTestUser.class)
-              : findInjectedUser(instance, injectTestWorkspace.user());
-
-      int memoryGb =
-          injectTestWorkspace.memoryGb() <= 0
-              ? workspaceDefaultMemoryGb
-              : injectTestWorkspace.memoryGb();
-
       TestWorkspace testWorkspace =
```
dmytro-ndp: According the changes [here](https://github.com/Katka92/rh-che/commit/f36760ebad39a9a925a8de1053e6bef42776e26a)  the main difference is how workspace is created, if you replace `RhCheTestWorkspaceImpl` by `TestWorkspace`.
If so, I would propose to make just creation of workspace an abstract, instead of whole method `injectMembers()`:
```
AbstractTestWorkspaceInjector.java
...
  injectMembers(T instance) {
    ...
     TestWorkspace testWorkspace = createWorkspace();
      testWorkspace.await();
      field.setAccessible(true);
      field.set(instance, testWorkspace);
     ....
   }

  abstract TestWorkspace createWorkspace();
...

}

File: selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/workspace/AbstractTestWorkspaceInjector.java (hunk 1 above)
Comment (Katka92) [reply selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/workspace/AbstractTestWorkspaceInjector.java]: If I make this class public I won't be able to use abstract method. Or am I wrong?

Comment (dmytro-ndp) [reply selenium/che-selenium-core/src/main/java/org/eclipse/che/selenium/core/workspace/AbstractTestWorkspaceInjector.java]: _public_ class doesn't restrict usage of _public_ abstract method.

**Comment by Katka92** [comment]
@dmytro-ndp After some investigation I found out that the only difference is in used provider. So I suggested another change that you advised. Please review and let me know your opinion. Thanks!

**Comment by dmytro-ndp** [comment]
Thank you for investigation!

It's interesting to now if you tried to obtain instance of `TestWorkspaceProvider` when executing `injectMembers()` method? :
```
  @Override
  public void injectMembers(T instance) throws RuntimeException {
    try {
      TestWorkspaceProvider testWorkspaceProvider = injector.getInstance(TestWorkspaceProvider.class);
      TestWorkspace testWorkspace =
          testWorkspaceProvider.createWorkspace(
              getUser(instance),
              getMemory(),
              injectTestWorkspace.template(),
              injectTestWorkspace.startAfterCreation());
      testWorkspace.await();
      field.setAccessible(true);
      field.set(instance, testWorkspace);
    } catch (Exception e) {
      throw new RuntimeException("Could not inject workspace. ", e);
    }
  }
```

dmytro-ndp:
@Katka92: please, review [my fixup](https://github.com/eclipse/che/pull/11586/commits/a61e7d7e76844b4b5fafbba419727b6f90854324), where were:
1. Added abstract method `getTestWorkspaceProvider()` to ensure `testWorkspaceProvider` instance is not null when executing `injectMembers()` method.
1. Gotten rid of redundant **AbstractTestWorkspaceInjector.injectorProvider** field.

Katka92:
@dmytro-ndp I reviewed it and seems good to me. :) Thank you for your help!

dmytro-ndp (approved):

PR #11670, opened 2018-10-18
Fix git when it is being used from the Theia terminal

sunix:
when using git from the terminal within the theia container we get:

```
sh-4.2$ git clone https://github.com/eclipse/che-theia-factory-extension.git
Cloning into 'che-theia-factory-extension'...
git-remote-https: error while loading shared libraries: libcurl-httpd24.so.4: cannot open shared object file: No such file or directory
sh-4.2$
```
This patch is adding the library path to be used


benoitf:
complete fix at https://github.com/eclipse/che/pull/11671
(to use git in theia)

Issue #11683, opened 2018-10-19
Zip Import Broken for Some Zips


tsmaeder:
I crated a zip file from a selenium test sample project and uploaded it to a known location.
Importing the zip file failed because the first zip entry was not a file directly inside the root folder, but had a longer path /foo/bar/zoz/Main.java.
The problem is that intermediate directories are not created before creating a file output stream for the file. 
This throws a "FileNotFoundException" for the parent directory.

PR #11686, opened 2018-10-19
Create parent dirs in zip import

tsmaeder:
### What does this PR do?
Makes sure parent dirs are created before attempting to write to a file during zip import

Zip Import Broken for Some Zips #11683

### DIFF
--- a/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/fs/server/impl/ZipArchiver.java
+++ b/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/fs/server/impl/ZipArchiver.java
@@ -111,6 +111,7 @@ class ZipArchiver {
             }
           }
 
+          Files.createDirectories(path.getParent());
           if (zipEntry.isDirectory()) {
             Files.createDirectory(path);
           } else {


wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/fs/server/impl/ZipArchiver.java
```diff
@@ -111,6 +111,7 @@ void unzip(
             }
           }
 
+          Files.createDirectories(path.getParent());
```
vparfonov, line 114: Does this code respect `withParents` param?

tsmaeder: What is "withParents" supposed to do? The current code fails if the first entry in the zip file is _not_ an entry directly inside the root directory, which is clearly wrong. Unfortunately, I haven't found any documentation on "withParents".

azatsarynnyy (approved):

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/fs/server/impl/ZipArchiver.java
dkulieshov: I dare to assume that the answer is "no".

From the source code I see that `withParents` parameter should indicates if there is a need to create not existing parent directories in `fsPath`. However if I understand correctly, your solution ignores this parameter and always creates missing directories. (I can not immediately say in what cases this parameter is useful, but we are fixing the bug and not discussing the API here.) Thus I consider that provided solution breaks the API.

I have to admit that a small part of che code is not well documented so next time if you will have difficulties with understanding the source code or the API please do not hesitate to ask che community members (via mailing list or mattermost).

dkulieshov (changes requested):
Please respect `withParents` parameter in your fix.

musienko-maxim:

Issue #11249, opened 2018-09-17
Convert Che-theia factory extension to plugin


sunix:
### Description
Convert https://github.com/eclipse/che-theia-factory-extension to plugin:
Related to https://github.com/eclipse/che/issues/11050
Is required by https://github.com/eclipse/che/issues/11240

To convert, we need the following to be done from a plugin:
- Retrieve the url query string to get the factory id
- Execute a given task
- Open a file
- Some UI entry point for notifications
- git (clone, checkout)
- Trigger when app is ready

PR #11718, opened 2018-10-24
Factory plugin

sunix:
### What does this PR do?
- Removes the legacy Che-Theia factory extension.
- Includes the che-theia-factory plugin.

https://github.com/eclipse/che/issues/11249

#### Release Notes
Che-theia clones projects that are defined in the workspace definition.
Che-theia performs factory cloning and execute post loading actions.

### DIFF
--- a/dockerfiles/theia/Dockerfile
+++ b/dockerfiles/theia/Dockerfile
@@ -106,6 +106,7 @@ RUN useradd -u 1000 -G users,wheel,root -d ${HOME} --shell /bin/bash theia \
     && mkdir /default-theia-plugins \
     # Download yeoman generator plug-in
     && curl -L -o /default-theia-plugins/theia_yeoman_plugin.theia https://github.com/eclipse/theia-yeoman-plugin/releases/download/untagged-04f28ee329e479cc465b/theia_yeoman_plugin.theia \
+    && curl -L -o /default-theia-plugins/eclipse_che_theia_factory_plugin.theia https://github.com/eclipse/che-theia-factory-extension/releases/download/untagged-856bf351cc3b4bcc335a/eclipse_che_theia_factory_plugin.theia \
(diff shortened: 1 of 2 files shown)


AndrienkoAleksandr (approved):

dockerfiles/theia/Dockerfile
```diff
@@ -92,7 +92,7 @@ ADD src/builder/setup-nodejs.sh /tmp/builder/scripts/setup-nodejs.sh
 # Install supervisor for running theia
 # Install which tool in order to search git
 RUN yum install -y centos-release-scl
-RUN yum install -y rh-git29 scl-utils sudo java-1.8.0-openjdk-devel bzip2 which && \
+RUN yum install -y rh-git29 scl-utils sudo java-1.8.0-openjdk-devel bzip2 which make gcc-c++ && \
```
benoitf: the theia-ide image is only a runtime image so make, gcc and c++ should not be there

benoitf (changes requested):
theia-ide image should not contains any compilation stuff.
(probably it may go in https://github.com/ws-skeleton/che-plugin-dev-tooling which is the "development image for plug-ins, etc) which is in che7 stack

sunix:
@benoitf ok but
at the same time
- it is the image where we do self hosting
- the way we build in docker should be the closest to the way we build in our dev environment.
- why do we include node_modules in /node_modules in this case ?
- why do we have jdk stuff in this image ?

My point is that we are adding restriction that are not so important to me. We should give the choice to the developer to build image anywhere and if for some reason developer prefers to build in che-theia image, why restricting him ?

sunix:
@benoitf I have updated the PR without make and gcc-c++

benoitf (approved):

riuvshin:
@sunix no `ci-build` / `ci-test` ??? this broke master :)

PR #11761, opened 2018-10-26
[Keycloak] Fix bugs on the docker image.

monaka:
### What does this PR do?

Fix the container image for Keycloak to use Postgres instead of H2.

### DIFF
--- a/deploy/kubernetes/helm/che/custom-charts/che-keycloak/templates/deployment.yaml
+++ b/deploy/kubernetes/helm/che/custom-charts/che-keycloak/templates/deployment.yaml
@@ -31,6 +31,8 @@ spec:
                 fieldPath: metadata.namespace
       containers:
       - env:
+        - name: DB_VENDOR
+          value: POSTGRES
         - name: POSTGRES_PORT_5432_TCP_ADDR
           value: postgres
         - name: POSTGRES_PORT_5432_TCP_PORT
--- a/dockerfiles/keycloak/kc_realm_user.sh
+++ b/dockerfiles/keycloak/kc_realm_user.sh
@@ -10,11 +10,11 @@
 
 echo "Configuring Keycloak by modifying realm and user templates..."
 
-cat /scripts/che-users-0.json.erb | \
-                                  sed -e "/<% if scope.lookupvar('keycloak::che_keycloak_admin_require_update_password') == 'true' -%>/d" | \
-                                  sed -e "/<% else -%>/d" | \
-                                  sed -e "/<% end -%>/d" | \
-                                  sed -e "/\"requiredActions\" : \[ \],/d" > /scripts/che-users-0.json
+#cat /scripts/che-users-0.json.erb | \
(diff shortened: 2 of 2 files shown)


riuvshin:

riuvshin:

riuvshin:

monaka:
I've just tested on Helm. I guess there have the similar issue on another runtimes. But I didn't touch them as I'm not familiar with them.

monaka:
I believe this patch itself is sane. But I encountered the another issue around authentication with Che.
I suspect there doesn't have json.erb files required by Che in the Keycloak image.
Which is better, a or b?

(a) I send the additional PR after closed this as they are separated issues.
(b) I push additional patches to his PR because the aim of this PR is to fix deployment failure.

skabashnyuk:
@monaka can you explain how changes you've made related to the topic of this pr? Especially commented lines in kc_realm_user.sh

monaka:
Let me change the title and the aim for this PR. As I tweeted, I managed to run Che on AKS (managed vanilla Kubernetes). I'm going to add some more patches to here.

monaka:
> What do you want to do?

Enable to run Che on vanilla K8s clusters.

> On wich environment? Docker, OpenShift, Kubernetes?

Kubernetes.
(Possibly Docker doesn't. Because it will use `init` container for bootstrapping.) 

> Which version of infrastructure?

Azure AKS 1.11.3.
But the issue will be reproduced all versions of K8s based infrastructure.

> Which version of Che?

master branch, nightly builds.

> What issues do you have with current codebase? Traces, logs?

- I can't see any tables in `keycloak` database on Postgres container.
- I can't see any tables in H2 for the `che` realm.

After fixing above, 

- There is no content for the custom login page.

monaka:
Currently, `/opt/jboss/keycloak/bin/standalone.sh` is called directly on bootstrapping.
But it is required to use `/opt/jboss/docker-entrypoint.sh` if we want to use external DB.

eivantsov:
@monaka thanks for fixing it. However, I do not understand why you copied jsons? it's ok to use them from where they are

monaka:
Indeed it's not good to put same files into different places. But I couldn't resolve how to access JSONs in `init` container from `che-keycloak`. I'm afraid my patches cause side effects to Openshift and Docker deployments, even though it might resolve by some tricky patches to `init`.

eivantsov:
@monaka it is ok to reuse jsons from init.

Just changing start script in the entrypoint should be enough, and you did that already

monaka:
@eivantsov Who setups the realm for Che?

eivantsov:
@monaka it happens in the entrypoint

monaka:
Yes. In addition to patches here, we must add the patch to `depoyment/keycloak` like this.
I'm not sure I'm not familiar with Openshift. But I think this issue will be reproduced on it.

```
        - name: PROXY_ADDRESS_FORWARDING
          value: "true"
        - name: DB_VENDOR
          value: POSTGRES
```

musienko-maxim:

monaka:
I close this as it have not been updated.

Issue #11554, opened 2018-10-09
Che plugin broker should support ephemeral mode


l0rd:
### Description

It is now possible to configure workspace to run in ephemeral mode (i.e. PVs will be of type `EmptyDir`) setting `persistVolumes` attribute to false.

```json
"attributes": {
  "persistVolumes": "false"
}
```

But Che 7 plugin brokers are still started with PVCs (`broker-config-volume` and `claim-che-workspace`). Can we fix that?

garagatyi:
@l0rd Since broker is running in a pod separate from workspace pods how would we deliver plugins into the workspace? Looks like we would need to redesign the flow to achieve this. WDYT?

l0rd:
@garagatyi yes. That's what I am worried about. But in the new theia-plugin-broker we should not need to deliver plugins anymore don't we?

garagatyi:
Actually, we need it. Theia reads metadata from the plugins `package.json` files.

garagatyi:
@benoitf could you comment whether I am correct.

amisevsk:
From a brief look at this yesterday, one issue that has to be addressed before anything else is https://github.com/eclipse/che/issues/11653.

Apart from that, it seems like the task of getting the plugin broker to work with ephemeral workspaces will be rather difficult; since the broker runs separate from the workspace pod, it's impossible to share storage between the two at the moment. After discussion with @garagatyi I've got a few leads and will work towards a proof-of-concept PR to see if we can implement this.

amisevsk:
Opened PR #11657 to address #11653.

garagatyi:
@amisevsk can we close this issue?

Issue #11476, opened 2018-10-04
Support running Che 7 plugins and runtime in separate pods


l0rd:
### Description

Currently in Kube and OpenShift infrastructure all Che 7 plugins, editors and runtime are spawned in one unique fat pod:

![image](https://user-images.githubusercontent.com/606959/46469406-8fd61900-c7d3-11e8-913c-33130d5e26fb.png)

This was a limitation that we had to live with because all containers needed access to the same Persistent Volume and we needed to support RWO Persistent Volumes first. 

But this configuration has many drawbacks:

- ports collisions
- ~~cannot support deployment/deployment config based recipes~~ configurations with more than one pod
- cannot run short living sidecars
- cannot restart plugins
- cannot restart runtime

This epic is about supporting an alternative configuration with every container running in its own Pod:

![image](https://user-images.githubusercontent.com/606959/46470018-7cc44880-c7d5-11e8-845b-593af79baebd.png)

This configuration should be the default when the workspace persistent volume is RWX.

garagatyi:
We can separate sidecars that do not declare the dependency on the projects' volume even now!
Apart from that, we can implement different strategies of volume content syncing. RWX is not supported everywhere, but we can add also use sync sidecar to emulate it. 
I would start from separating non-volume dependent sidecars to separate pods.

ibuziuk:
@garagatyi @l0rd I have created an issue for supporting non-volume dependent sidecars as separate pods in rh-che [1]. Do you happen to have some ideas about other potential sub-tasks that we could add to the next sprint related to this epic ?

[1] https://github.com/redhat-developer/rh-che/issues/1005

l0rd:
@ibuziuk the goal of these 2 issues is to run containers in distinct pods. But besides that they are not related (e.g one is a subtask of the other). Let me clarify:
- this issue is about taking advantage of eventual RWX PVs
- https://github.com/redhat-developer/rh-che/issues/1005 is about containers that don't need persistence and can be run in a distinct POD

As mentioned in a call we had today I would prioritize the work in this order:
1. support fully ephemeral workspaces (che 7)
2. run containers that don't need access to the PV in a distinct Pod
3. take advantage of RWX PR when available

l0rd:
This issue is not current anymore. We are rather aiming to run all IDE services in the developer container.

PR #11786, opened 2018-10-31
Ensure plugin broker works with ephemeral workspaces

amisevsk:
### What does this PR do?
Changes necessary to make plugin broker work correctly when a workspace is marked as ephemeral. This is done by making the broker run as an init container on ephemeral workspaces, so that any volume-related actions are reflected in the workspace's emptyDir volumes.

Detailed changes:
- Slight changes to `EphemeralWorkspaceAdapter`: 1. Make ephemeral workspace check methods static to allow them to be used elsewhere, and 2. add method to "make" a workspace ephemeral, so that all attribute management happens in one place.
- Modify PVC strategies to account for init containers. For `EphemeralWorkspaceAdapter` this is necessary, as otherwise the `/plugins` directory cannot be mounted. I've also included this change for the common and unique strategies, although it doesn't do anything at the moment (since no init containers are added in the non-ephemeral case).
- Apply plugin broker as init container if workspace is ephemeral

#11554 

#### Release Notes
Allow Che 7 workspaces to run as ephermal (emptyDir) containers on Kubernetes/OpenShift.

### DIFF
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/pvc/CommonPVCStrategy.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/pvc/CommonPVCStrategy.java
@@ -23,8 +23,10 @@ import io.fabric8.kubernetes.api.model.Container;
 import io.fabric8.kubernetes.api.model.PersistentVolumeClaim;
 import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.api.model.PodSpec;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
@@ -102,7 +104,7 @@ public class CommonPVCStrategy implements WorkspaceVolumesStrategy {
   public void provision(KubernetesEnvironment k8sEnv, RuntimeIdentity identity)
       throws InfrastructureException {
     final String workspaceId = identity.getWorkspaceId();
-    if (ephemeralWorkspaceAdapter.isEphemeral(k8sEnv.getAttributes())) {
+    if (EphemeralWorkspaceAdapter.isEphemeral(k8sEnv.getAttributes())) {
       ephemeralWorkspaceAdapter.provision(k8sEnv, identity);
       return;
     }
@@ -112,6 +114,9 @@ public class CommonPVCStrategy implements WorkspaceVolumesStrategy {
     k8sEnv.getPersistentVolumeClaims().put(pvcName, pvc);
     for (Pod pod : k8sEnv.getPods().values()) {
       PodSpec podSpec = pod.getSpec();
+      List<Container> containers = new ArrayList<>();
+      containers.addAll(podSpec.getContainers());
+      containers.addAll(podSpec.getInitContainers());
       for (Container container : podSpec.getContainers()) {
         String machineName = Names.machineName(pod, container);
         InternalMachineConfig machineConfig = k8sEnv.getMachines().get(machineName);
@@ -129,7 +134,7 @@ public class CommonPVCStrategy implements WorkspaceVolumesStrategy {
   @Override
   public void prepare(KubernetesEnvironment k8sEnv, String workspaceId)
       throws InfrastructureException {
-    if (ephemeralWorkspaceAdapter.isEphemeral(k8sEnv.getAttributes())) {
+    if (EphemeralWorkspaceAdapter.isEphemeral(k8sEnv.getAttributes())) {
       return;
     }
     final Collection<PersistentVolumeClaim> claims = k8sEnv.getPersistentVolumeClaims().values();
@@ -156,7 +161,7 @@ public class CommonPVCStrategy implements WorkspaceVolumesStrategy {
 
   @Override
   public void cleanup(Workspace workspace) throws InfrastructureException {
-    if (ephemeralWorkspaceAdapter.isEphemeral(workspace)) {
+    if (EphemeralWorkspaceAdapter.isEphemeral(workspace)) {
       return;
     }
     String workspaceId = workspace.getId();
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/pvc/CommonPVCStrategy.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/pvc/CommonPVCStrategy.java
@@ -104,7 +104,7 @@ public class CommonPVCStrategy implements WorkspaceVolumesStrategy {
   public void provision(KubernetesEnvironment k8sEnv, RuntimeIdentity identity)
       throws InfrastructureException {
     final String workspaceId = identity.getWorkspaceId();
-    if (EphemeralWorkspaceAdapter.isEphemeral(k8sEnv.getAttributes())) {
+    if (EphemeralWorkspaceUtility.isEphemeral(k8sEnv.getAttributes())) {
       ephemeralWorkspaceAdapter.provision(k8sEnv, identity);
       return;
     }
@@ -117,7 +117,7 @@ public class CommonPVCStrategy implements WorkspaceVolumesStrategy {
       List<Container> containers = new ArrayList<>();
       containers.addAll(podSpec.getContainers());
       containers.addAll(podSpec.getInitContainers());
-      for (Container container : podSpec.getContainers()) {
+      for (Container container : containers) {
         String machineName = Names.machineName(pod, container);
         InternalMachineConfig machineConfig = k8sEnv.getMachines().get(machineName);
         addMachineVolumes(workspaceId, subPaths, pod, container, machineConfig.getVolumes());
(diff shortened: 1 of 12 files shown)


l0rd:
I have a couple of question with respect to the init container approach:
1. Will that work when plugins will run in distinct pods (c.f. https://github.com/eclipse/che/issues/11476 and https://github.com/redhat-developer/rh-che/issues/1005)? I mean ephemeral mode for containers running in distinct pods.
2. Will that work if we want some workspace containers to be loaded first than others (i.e. the editor container is deployed immediately to shorten the start up time for users and other containers are loaded in a lazy manner)?

amisevsk:
@l0rd I think ephemeral workspaces will make distinct pods a difficult issue -- an emptyDir volume can only be attached to one pod. It might be complicated but with some architecture work it might be possible. 

untitled diagram 1

As it stands, the diagram above shows what a Che 7 workspace looks like in terms of volumes. From this diagram, to make sure access to projects is shared, theia-ide, che-dev, and any plugins (e.g. che-hello) that needs access to /projects would need to be in the same pod. Further, for the broker to work, it needs to be an init container on the theia pod, so the only plugin that could be separated out into its own pod is the che-machine-exec container.

On the plugin broker side, however, since we only need to touch the plugins volume, the broker container needs only to run as a init container on the theia (or editor) container. If e.g. `/projects` was RWX, then the theia container could be its own pod that starts separately, with the broker as its init container. 

One thing Alex and I talked about briefly is potentially figuring out requirements for multiple brokers (e.g. a theia plugin-only broker would not need an initial run, and could run exclusively as an init container. We didn't really go far down this route though.

garagatyi:
@amisevsk FYI we were also considering an ability to create plugins not only for editors but for other items, such as plugins (yeah, plugins of plugins). For example JDT.LS can have its own plugins. But we haven't finished discussions regarding that.
And I would love to remove workspace logs volume in Che 7 since it doesn't make much sense now, here is the [issue](https://github.com/eclipse/che/issues/11806).

garagatyi (commented):
Overall looks good. Would be cool to have unit tests, though. 
I was thinking about the overall design and maybe we could have a code that would add broker as regular or init container depending on something. In this case, depending on the ephemeral property. 
Or later, depending on some config of a broker we would add. With such an approach the code might seem cleaner. But not sure whether we need to do this in this particular PR.

garagatyi:
@l0rd after discussing the brokering process with Angel and Serhii I'm thinking that we can do improvements described below to tackle your points.
We can set whether a broker changes workspace configuration or copies files. If it just changes configuration we don't need to add volumes to it (actually it can specify that it doesn't need any volume) and it can play with ephemeral workspaces nicely. If it just copies files it would go to the init containers and would probably be able to copy plugins without the usage of non-ephemeral volumes. And only in a case, it needs to do both steps it would be running 2 times in a case of the ephemeral workspace.
Maybe someone would also come up with other improvements for this flow.

l0rd:
@garagatyi @amisevsk @skabashnyuk I will try to provide a better description of the requirements. And I think we should have a call today or tomorrow to discuss it.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/pvc/CommonPVCStrategy.java
```diff
@@ -108,6 +110,9 @@ public void provision(KubernetesEnvironment k8sEnv, RuntimeIdentity identity)
     k8sEnv.getPersistentVolumeClaims().put(pvcName, pvc);
     for (Pod pod : k8sEnv.getPods().values()) {
       PodSpec podSpec = pod.getSpec();
+      List<Container> containers = new ArrayList<>();
```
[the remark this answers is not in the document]
amisevsk: Good catch -- thank you. I meant to loop over this list.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/pvc/EphemeralWorkspaceAdapter.java
```diff
@@ -61,11 +63,21 @@ public boolean isEphemeral(Workspace workspace) {
    *     of the PVC strategy, workspace volumes would be created as `emptyDir`. When a workspace Pod
    *     is removed for any reason, the data in the `emptyDir` volume is deleted forever
    */
-  public boolean isEphemeral(Map<String, String> workspaceAttributes) {
+  public static boolean isEphemeral(Map<String, String> workspaceAttributes) {
```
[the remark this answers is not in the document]
amisevsk: Done.

amisevsk:
I've addressed the review comments and added tests. I also had to rebase the PR to fix a merge conflict.

amisevsk:
ci-test

riuvshin:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1139/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1139//Selenium_20tests_20report/)
docker image: eclipseche/che-server:11786

garagatyi:

garagatyi (approved):

artaleks9:
Selenium tests execution on Eclipse Che Multiuser on OCP (https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1139/) doesn't show any regression against this Pull Request.

amisevsk:
Fixed codacy's complaining and rebased onto current master.

@garagatyi @ibuziuk @sleshchenko I'd like to merge tomorrow unless there's something else that needs to be improved.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsplugins/KubernetesBrokerInitContainerApplier.java
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  /**
+   * Apply plugin broker as init container to workspace environment. Workspace environment will have
+   * broker's configmap, machines, and volumes added in addition to the init container
+   *
+   * @param workspaceEnvironment
+   * @param runtimeID
+   * @param pluginsMeta
+   * @throws InfrastructureException
```
sleshchenko: I found it a bit useless to have listed parameters, exceptions without any description. WDYT?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsplugins/KubernetesBrokerInitContainerApplier.java
```diff
@@ -0,0 +1,100 @@
...
+   */
+  public void apply(
+      InternalEnvironment workspaceEnvironment,
```
sleshchenko: I think that we haven't used `workspaceEnvironment` word anywhere. Please consider changing it to `internalEnvironment`

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsplugins/KubernetesBrokerInitContainerApplier.java
```diff
@@ -0,0 +1,100 @@
...
+      RuntimeIdentity runtimeID,
+      Collection<PluginMeta> pluginsMeta)
+      throws InfrastructureException {
+
+    KubernetesEnvironment kubernetesEnvironment = (KubernetesEnvironment) workspaceEnvironment;
```
sleshchenko: Since your class is generified with `E extends KubernetesEnvironment` - you can avoid casting by putting E as type of `workspaceEnvironment`.
BTW it is not safe to cast without any check of type.

sleshchenko (approved):
LGTM
The only one bad side of current implementation - all plugin brokers are run twice and as result Plugins Binaries are downloaded twice as well, but I'm OK to merge it since I can't propose an alternative implementation with current Che 7 Model

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsplugins/KubernetesBrokerInitContainerApplier.java
amisevsk: Good point; I'll fix it.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsplugins/KubernetesBrokerInitContainerApplier.java
amisevsk: My intention in using `workspaceEnvironment` was to avoid confusion with `brokerEnvironment` and to support the intuition that we're basically copying environment info from the broker to the workspace. I do agree that immediately casting `workspaceEnvironment` to `kubernetesEnvironment` breaks that idea completely though. 

I've set the type of `workspaceEnvironment` to `E` and removed the cast. WDYT?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsplugins/KubernetesBrokerInitContainerApplier.java
amisevsk: Done.

amisevsk:
@sleshchenko Thanks for your review. I agree that running plugin broker twice is not ideal, but also don't know if it will be possible to avoid (at least, for the plugins that require e.g. services on the workspace Pod. One way the init container may be more convenient is for plugins (e.g. theia) that don't require modifying the OpenShift object, where the init container could be sufficient.

ibuziuk (approved):

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsplugins/KubernetesBrokerInitContainerApplier.java
sleshchenko: > My intention in using workspaceEnvironment was to avoid confusion with brokerEnvironment

Makes sense

PR #11794, opened 2018-10-31
Reuse the Dashboard Keycloak instance in IDE frames when possible

davidfestal:
### What does this PR do?

This PR allows reusing the Dashboard Keycloak instance in IDE frames when possible (=> when the IDE window is opened as a frame in the Dashboard window).

This avoids performing one or even (in some cases) 2 authentication flows before the real IDE loading,
and this decreases the overall IDE loading process by 5 seconds or more.

(5-6 seconds out of 10-12s in my case)

This is a first step in some more general work around optimizing the IDE loading in OSIO: https://github.com/redhat-developer/rh-che/issues/1013

### DIFF
--- a/ide/che-ide-gwt-app/src/main/resources/org/eclipse/che/ide/public/loader.js
+++ b/ide/che-ide-gwt-app/src/main/resources/org/eclipse/che/ide/public/loader.js
@@ -19,7 +19,7 @@ class KeycloakLoader {
         return new Promise((resolve, reject) => {
             if (window.parent && window.parent['_keycloak']) {
                 window['_keycloak'] = window.parent['_keycloak'];
-                resolve(window.parent['_keycloak']);
+                resolve(window['_keycloak']);
                 return;
             }
             try {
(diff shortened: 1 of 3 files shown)


ide/che-ide-gwt-app/src/main/resources/org/eclipse/che/ide/public/loader.js
```diff
@@ -17,6 +17,11 @@ class KeycloakLoader {
         const msg = "Cannot load keycloak settings. This is normal for single-user mode.";
 
         return new Promise((resolve, reject) => {
+            if (window.parent && window.parent['_keycloak']) {
+                window['_keycloak'] = window.parent['_keycloak'];
+                resolve(window.parent['_keycloak']);
```
benoitf: maybe we should resolve to `our` keycloak (window['_keycloak'])

davidfestal: yes, probably. In fact I added line 21 after line 22, which explains why it remained like that ;-)
Not sure that makes a real difference though since both point to the same object, and both are set only at page load.

ibuziuk (approved):

ide/che-ide-gwt-app/src/main/resources/org/eclipse/che/ide/public/loader.js
davidfestal: fixed

davidfestal:
@benoitf is it OK for you to approve now ?

benoitf (approved):

PR #11795, opened 2018-10-31
drop codenvy nexus repository

riuvshin:
since all che artifacts moved to maven central we should stop using Codenvy nexus.


riuvshin:
ci-build

skabashnyuk (approved):

PR #11809, opened 2018-11-02
Add Che 7 stack: Java + MySQL

mmorhun:
### What does this PR do?
Adds Che 7 stack with Java 8 and MySQL database.
Theia IDE added as a plugin by default.

https://github.com/redhat-developer/rh-che/issues/994

### DIFF
--- a/ide/che-core-ide-stacks/src/main/resources/stacks.json
+++ b/ide/che-core-ide-stacks/src/main/resources/stacks.json
@@ -164,146 +164,6 @@
     }
   },
   {
-    "id": "java-mysql-theia-openshift",
-    "creator": "ide",
-    "name": "Che 7 - Java + MySQL",
-    "description": "Multi Container Workspace -Java + MySQL in Theia IDE",
-    "scope": "general",
-    "tags": [
-      "Java 1.8",
-      "Theia",
-      "MySQL"
-    ],
-    "workspaceConfig": {
-      "environments": {
-        "default": {
-          "recipe": {
-            "contentType": "text/x-yaml",
-            "type": "openshift",
-            "content": "kind: List\nitems:\n - \n  apiVersion: v1\n  kind: Pod\n  metadata:\n   name: web\n   labels:\n    type: db\n  spec:\n   containers:\n    - \n     image: eclipse/centos_jdk8\n     name: dev\n     ports:\n      - \n       containerPort: 8080\n       protocol: TCP\n     resources:\n      limits:\n       memory: 512Mi\n    - \n     image: centos/mysql-57-centos7\n     name: mysql\n     ports:\n      - \n       name: mysql\n       containerPort: 3306\n       protocol: TCP\n     env:\n      - \n       name: MYSQL_USER\n       value: petclinic\n      - \n       name: MYSQL_ROOT_PASSWORD\n       value: password\n      - \n       name: MYSQL_PASSWORD\n       value: password\n      - \n       name: MYSQL_DATABASE\n       value: petclinic\n     resources:\n      limits:\n       memory: 256Mi\n - \n  apiVersion: v1\n  kind: Service\n  metadata:\n   name: db\n  spec:\n   selector:\n    type: db\n   ports:\n    - \n     name: mysql\n     port: 3306\n     protocol: TCP\n     targetPort: mysql\n"
-          },
-          "machines": {
-            "web/dev": {
-              "servers": {
-                "tomcat8": {
-                  "attributes": {},
-                  "protocol": "http",
-                  "port": "8080/tcp"
-                }
-              },
-              "volumes": {
-                "projects": {
-                  "path": "/projects"
-                }
-              },
-              "installers": [],
-              "env": {},
-              "attributes": {
-                "memoryLimitBytes": "536870912"
-              }
-            },
-            "web/mysql": {
-              "servers": {
-                "mysql": {
-                  "attributes": {},
-                  "port": "3306",
-                  "protocol": "tcp"
-                }
-              },
-              "volumes": {},
-              "installers": [],
-              "env": {},
-              "attributes": {
-                "memoryLimitBytes": "214748365"
-              }
-            }
-          }
-        }
-      },
-      "defaultEnv": "default",
-      "name": "default",
-      "projects": [],
-      "commands": [
-        {
-          "commandLine": "/opt/rh/rh-maven33/enable && mvn clean install -f ${current.project.path}",
-          "name": "build",
-          "type": "mvn",
-          "attributes": {
-            "machineName": "web/dev",
-            "previewUrl": "",
-            "goal": "Build"
-          }
-        },
-        {
-          "commandLine": "cd ${current.project.path}/target/ && cp *.war ${TOMCAT_HOME}/webapps/ROOT.war",
-          "name": "deploy",
-          "type": "custom",
-          "attributes": {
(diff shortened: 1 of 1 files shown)


mmorhun:
Because of problem with volumes on openshift.io I cannot fully test it, because build fails with errors like:
```
INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] Total time: 3.141 s
[INFO] Finished at: 2018-11-02T08:33:25+00:00
[INFO] Final Memory: 9M/20M
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-resources-plugin:2.6:resources (default-resources) on project spring-petclinic: Failed to copy full contents from /projects/web-java-spring-petclinic/src/main/resources/dandelion/datatables/datatables.properties to /projects/web-java-spring-petclinic/target/classes/dandelion/datatables/datatables.properties -> [Help 1]
```
However I managed to start petclinic sample project (checkout to `mysql` branch) and add value into database.

benoitf:
if it's in upstream che, shouldn't all the stacks be compliant with kubernetes and openshift ? here it seems it is only for openshift

ide/che-core-ide-stacks/src/main/resources/stacks.json
```diff
@@ -163,6 +163,146 @@
       "mediaType": "image/svg+xml"
     }
   },
+  {
+    "id": "java-mysql-theia-openshift",
+    "creator": "ide",
+    "name": "Che 7 - Java + MySQL",
+    "description": "Multi Container Workspace -Java + MySQL in Theia IDE",
```
sleshchenko: Please add space before java `- Java`

ide/che-core-ide-stacks/src/main/resources/stacks.json
```diff
@@ -163,6 +163,146 @@
...
+    "workspaceConfig": {
+      "environments": {
+        "default": {
+          "recipe": {
+            "contentType": "text/x-yaml",
+            "type": "openshift",
+            "content": "kind: List\nitems:\n - \n  apiVersion: v1\n  kind: Pod\n  metadata:\n   name: web\n   labels:\n    type: db\n  spec:\n   containers:\n    - \n     image: eclipse/centos_jdk8\n     name: dev\n     ports:\n      - \n       containerPort: 8080\n       protocol: TCP\n     resources:\n      limits:\n       memory: 512Mi\n    - \n     image: centos/mysql-57-centos7\n     name: mysql\n     ports:\n      - \n       name: mysql\n       containerPort: 3306\n       protocol: TCP\n     env:\n      - \n       name: MYSQL_USER\n       value: petclinic\n      - \n       name: MYSQL_ROOT_PASSWORD\n       value: password\n      - \n       name: MYSQL_PASSWORD\n       value: password\n      - \n       name: MYSQL_DATABASE\n       value: petclinic\n     resources:\n      limits:\n       memory: 256Mi\n - \n  apiVersion: v1\n  kind: Service\n  metadata:\n   name: db\n  spec:\n   selector:\n    type: db\n   ports:\n    - \n     name: mysql\n     port: 3306\n     protocol: TCP\n     targetPort: mysql\n"
```
sleshchenko, line 271: Stack description says nothing about `petclinic`, maybe it would be better to set some general values for Database env vars? Like `MYSQL_USER=user`, `MYSQL_PASSWORD=password` =)

sleshchenko: Also, you specified memory limit in the recipe and in machine configuration for `web/dev` machine. I think it would be better to choose one place for configuring memory.

sleshchenko:
> if it's in upstream che, shouldn't all the stacks be compliant with kubernetes and openshift ? here it seems it is only for openshift

Good point. @mmorhun You can change `openshift` type to `kubernetes` and then it will work for both ( K8s and OS infrastructures). P.S. Don't forget to remove `openshift` word from stack name as well.

ide/che-core-ide-stacks/src/main/resources/stacks.json
mmorhun: Ok, thanks

ide/che-core-ide-stacks/src/main/resources/stacks.json
mmorhun: @sleshchenko what is preferable place to configure memory? I found it quite confusing with both places.

sleshchenko: > @sleshchenko what is preferable place to configure memory?

It's hard to say what is preferable place in this particular case. A recipe is a place where a user defines his runtime (ideal case - the same recipe as he uses for running application on production) and he may define needed limits there.
Che gives an ability to override memory limit in a recipe by specifying a corresponding value in machine configuration.
Up to you which place you choose.

> I found it quite confusing with both places.

It's why I asked you to choose one place =) Because it is confusing that you configure and then override memory limit.

mmorhun: As about user that's should be arranges for specific project. Here we have command which clones petclinic sample. But it is easy to change the value in dashboard.

mmorhun:
@sleshchenko @benoitf I decoded to go with one stack of kubernetes recipe type because it doesn't contains any openshift specific features and therefore will work on both.

mmorhun:
@l0rd @slemeur it will not work until the problem with volumes resolved (unless workaround it in non user friendly way: clone the petclinic project outside volume, build and run).

ide/che-core-ide-stacks/src/main/resources/stacks.json
```diff
@@ -251,6 +251,152 @@
...
+          "name": "build",
+          "type": "mvn",
+          "attributes": {
+            "machineName": "web/dev",
+            "previewUrl": "",
+            "goal": "Build"
+          }
+        },
+        {
+          "commandLine": "cd ${current.project.path}/target/ && cp *.war ${TOMCAT_HOME}/webapps/ROOT.war",
```
sleshchenko, line 332: This command won't work with Theia, right?

ide/che-core-ide-stacks/src/main/resources/stacks.json
sleshchenko: @mmorhun 

> As about user that's should be arranges for specific project. Here we have command which clones petclinic sample. But it is easy to change the value in dashboard.

Sorry, I missed this command before. 
Well, I think stack's purpose is more than just petclinic application and it makes more sense to configure petclinic app (like with `sed` command after cloning) instead of configuring MYSQL container for petclinic sample. But up to you change or not

ide/che-core-ide-stacks/src/main/resources/stacks.json
mmorhun: It will if user selects a project in Files panel.

sleshchenko (approved):

AndrienkoAleksandr (approved):

l0rd (approved):

riuvshin:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1131/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1131//Selenium_20tests_20report/)
docker image: eclipseche/che-server:11809

musienko-maxim (approved):

benoitf (approved):

Issue #11810, opened 2018-11-02
Unexpected fail of "TheiaBuildPluginTest" and "Che7PreviewStackStartTest" selenium tests


Ohrimenko1988:
### Description
We have unexpected fail of "TheiaBuildPluginTest" and "Che7PreviewStackStartTest" selenium tests. It looks that something went wrong with a method which checks notifications textx. We should investigate and resolve this problem.

**Report with reproduced problem:**
https://ci.codenvycorp.com/view/qa-nightly/job/che-integration-tests-multiuser-master-ocp/414/Selenium_20tests_20report/

Ohrimenko1988:
Done https://github.com/eclipse/che/pull/11813

PR #11813, opened 2018-11-02
[Selenium] Fix unexpected fail of "TheiaBuildPluginTest" and "Che7PreviewStackStartTest" selenium tests

Ohrimenko1988:
### What does this PR do?
Fix unexpected fail of "TheiaBuildPluginTest" and "Che7PreviewStackStartTest" selenium tests

Issue: https://github.com/eclipse/che/issues/11810

### DIFF
--- a/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/theia/TheiaIde.java
+++ b/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/theia/TheiaIde.java
@@ -23,6 +23,7 @@ import com.google.inject.Inject;
 import com.google.inject.Singleton;
 import javax.annotation.PreDestroy;
 import org.eclipse.che.selenium.core.SeleniumWebDriver;
+import org.eclipse.che.selenium.core.utils.WaitUtils;
 import org.eclipse.che.selenium.core.webdriver.SeleniumWebDriverHelper;
 import org.openqa.selenium.By;
 import org.openqa.selenium.WebElement;
@@ -180,6 +181,9 @@ public class TheiaIde {
 
   public void pressKeyCombination(CharSequence... combination) {
     final String keyCombination = chord(asList(combination));
+
+    // for ensure that events are end before pressing keys combination
+    WaitUtils.sleepQuietly(1);
 
     seleniumWebDriverHelper.sendKeys(keyCombination);
   }
--- a/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/theia/TheiaIde.java
+++ b/selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/theia/TheiaIde.java
@@ -182,7 +182,7 @@ public class TheiaIde {
   public void pressKeyCombination(CharSequence... combination) {
     final String keyCombination = chord(asList(combination));
 
-    // for ensure that events are end before pressing keys combination
+    // to ensure that events are ended before pressing keys combination
     WaitUtils.sleepQuietly(1);
 
     seleniumWebDriverHelper.sendKeys(keyCombination);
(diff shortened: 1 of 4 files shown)


musienko-maxim (approved):

artaleks9 (approved):

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/theia/TheiaBuildPluginTest.java
```diff
@@ -125,7 +123,7 @@ public void pluginShouldBeBuilt() {
     }
   }
 
-  @Test(priority = 1)
+  @Test(priority = 1, groups = OPENSHIFT)
```
dmytro-ndp: If each public method in test class should be run on _OpenShift_ only, we can simplify the code by placing `@Test(groups = OPENSHIFT)` above the class declaration, and removing `groups = OPENSHIFT` from the test methods.

selenium/che-selenium-test/src/test/java/org/eclipse/che/selenium/theia/TheiaBuildPluginTest.java
```diff
@@ -217,23 +215,20 @@ private void switchToParentWindow(String parentWindowHandle) {
...
         driver -> {
-          String title = seleniumWebDriver.getTitle();
-
-          if (EXPECTED_DEVELOPMENT_HOST_TITLE.equals(title)) {
-            return true;
+          try {
+            theiaIde.waitTheiaIdeTopPanel();
+            theiaProjectTree.waitFilesTab();
+          } catch (TimeoutException ex) {
+            seleniumWebDriver.navigate().refresh();
```
dmytro-ndp, line 227: It's better to comment on why we need to have such kind of workaround.

selenium/che-selenium-test/src/test/resources/suites/CheOneThreadTestsSuite.xml
```diff
@@ -40,6 +40,8 @@
             <class name="org.eclipse.che.selenium.dashboard.ImportProjectFromGitHubTest"/>
             <class name="org.eclipse.che.selenium.factory.DirectUrlFactoryWithKeepDirectoryTest"/>
             <class name="org.eclipse.che.selenium.dashboard.workspaces.WorkspacesListTest"/>
+            <class name="org.eclipse.che.selenium.theia.TheiaBuildPluginTest"/>
```
dmytro-ndp: We have to support `CheOneThreadTestsSuite.xml` suite as short as possible, because these tests are nonscalable. 
What is the reason to move them from the `CheSuite.xml`? If they interfere each other, we can rearrange them inside the `CheSuite.xml`.

selenium/che-selenium-test/src/main/java/org/eclipse/che/selenium/pageobject/theia/TheiaIde.java
```diff
@@ -181,6 +182,9 @@ public void switchToIdeFrame() {
   public void pressKeyCombination(CharSequence... combination) {
     final String keyCombination = chord(asList(combination));
 
+    // for ensure that events are end before pressing keys combination
```
dmytro-ndp: slight correction `// to ensure that events are ended before pressing keys combination`

selenium/che-selenium-test/src/test/resources/suites/CheOneThreadTestsSuite.xml
Ohrimenko1988: The reason is if some theia stack is already run than each other workspace will be failed with errors

dmytro-ndp: In that case I would recommend to move second test far away (at the end) from first in CheSuite.xml list.

Ohrimenko1988: Great catch, thanks

dmytro-ndp (approved):

PR #11840, opened 2018-11-05
The PVC cleaning on workspace deletion is done in the wrong condition

davidfestal:
### What does this PR do?

The PVC cleaning on workspace deletion was done in the wrong condition: it was only done when each workspace is created in its own new Kubernetes namespace, through it seems it should be precisely the contrary.
This PR fixes this buggy behavior that also impacts the OSIO namespace PVC cleaning.

This is a fix for issue
https://github.com/redhat-developer/rh-che/issues/1026 (PVC is not
cleaned when workspace is deleted)


davidfestal:
ci-test

garagatyi (approved):
Looks good! It is strange that we have this behavior. Don't understand why we haven't figured out it before.

garagatyi:

artaleks9:
Selenium tests execution on Eclipse Che Multiuser on OCP (https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1136/) doesn't show any regression against this Pull Request.

sleshchenko (approved):
LGTM

PR #11888, opened 2018-11-09
Che-theia: adding @theia/cli to dependencies so it is available in production mode

sunix:
step 1 to avoid issue when running hosted instance:
```
yarn run v1.10.1
error Command "theia" not found.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```


benoitf (changes requested):
The global idea is not to add it into dependencies/production mode as the goal is to have the minimal image.
so having all cli dependencies is like having webpack dependencies, etc.

There was already some changes to leverage that when running theia as well

sunix:
ok, so using `node` command instead of yarn

PR #11908, opened 2018-11-12
Revert "CHE-10480: Fix NullPointerException (#11857)"

vinokurig:
This reverts commit 00f1afa2636335d6ef07ef47f6539fd3bc49035f.

PR #11857, opened 2018-11-07
CHE-10480: Fix NullPointerException

vinokurig:
### What does this PR do?
Fix NullPointerException

### What issues does this PR fix or reference?
#10480

#### Release Notes

N/A

#### Docs PR

N/A

PR #11919, opened 2018-11-13
Revert "Revert "CHE-10480: Fix NullPointerException (#11857)" (#11908)"

vinokurig:
### What does this PR do?
The revert #11857 was an experiment to fix the Jenkins job on CI. So this revert applies the changes back


riuvshin (approved):

sleshchenko:
Reverting of revert looks a bit strange. Why don't you do new commit which fixed NPE? IMHO It would looks much clear and better

vinokurig:
The revert of the commit was an experiment to fix the Jenkins job on CI. If I opened a PR with original title, it would look like a new PR. As it was approved earlier it has a revert title.

sleshchenko:
OK, I just say that it looks unclear and strange to me. If you want to make it clear that the same changed were merged and reverted before - I think PR description, commit details are good places for such information.

It's just a common recommendation and you do not have to fix it no this PR.

vinokurig:
Thank you for advice, I'll update the PR and commit description

PR #11975, opened 2018-11-17
Use `bash` instead of `sh`.

### LINKED ISSUE #10041: Unable to build Che images if default shell doesn't support `<<<` operator
### Description
Unable to build Che images if default shell doesn't support `<<<` operator.

### Reproduction Steps
1. Try to build Che Server image where default shell is configured to `dash`.
Expected: Image is built.
Actual: The following error appears:
```
serg@sergpad ~/projects/che/dockerfiles/che $ ./build.sh 
./build.sh: 25: /home/serg/projects/che/dockerfiles/che/../build.include: Syntax error: redirection unexpected
```
**OS and version:**    
```
$ uname -a
Linux sergpad 4.10.0-38-generic #42~16.04.1-Ubuntu SMP Tue Oct 10 16:32:20 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

$ lsb_release -a
No LSB modules are available.
Distributor ID:	LinuxMint
Description:	Linux Mint 18.3 Sylvia
Release:	18.3
Codename:	sylvia
```

**Diagnostics:**

monaka:
### What does this PR do?

Replaces some `sh` to `bash`.

Because `sh` is not linked to `bash` on some distributions like Ubuntu.

Will fix #10041 

#### Release Notes

N/A

#### Docs PR

N/A

### DIFF
--- a/dockerfiles/postgres/init-debug.sh
+++ b/dockerfiles/postgres/init-debug.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 # Copyright (c) 2012-2018 Red Hat, Inc.
 # This program and the accompanying materials are made
--- a/dockerfiles/action/build.sh
+++ b/dockerfiles/action/build.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 # Copyright (c) 2017 Red Hat, Inc.
 # This program and the accompanying materials are made
--- a/dockerfiles/base/build.sh
+++ b/dockerfiles/base/build.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 # Copyright (c) 2017 Red Hat, Inc.
 # This program and the accompanying materials are made
--- a/dockerfiles/bats/build.sh
+++ b/dockerfiles/bats/build.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 # Copyright (c) 2012-2017 Red Hat, Inc.
 # This program and the accompanying materials are made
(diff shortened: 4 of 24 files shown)


riuvshin:

riuvshin:

riuvshin:

monaka:
Actually, I got some build errors after applying this patch. I think they are caused even if I didn't fix this.

sleshchenko:
I'm OK with such a fix because otherwise, our scripts don't work properly if default shell doesn't support used operators. 

Now, the issue is caused by using `<<<` operator and we have to rewrite our scripts not to use it or declare bash as a default shell. As far as I remember @riuvshin know why `bash` as default shell may be a problem.

garagatyi:
Overall I agree that we need to stick to sh rather than bash. But this PR should be very well tested before the merge to be on a safe side. For example usage of `<<<` mentioned by Serhii is not fixed in this PR and this would break some functionality. So, we need someone assigning checking all the use-cases to himself.

riuvshin:
using `sh` was intentional for CLI scripts. However I think we can use `bash` for build scripts to fix `<<<` issue

dockerfiles/postgres/init-debug.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
sleshchenko, line 1: I think it also should be reverted since it is supposed to be executed in docker container.

sleshchenko:
@monaka Thanks for your contribution. I see that @vitaliy-guliy fixes the same issue in his PR https://github.com/eclipse/che/pull/12004
You can close your PR and Vitaliy will merge his after reviewers approvals and testing.

riuvshin:
#12004 is merged, so this PR is no longer relevant, anyway thanks for your efforts @monaka !

Issue #11832, opened 2018-11-05
Implementation of local part of devfile feature


mshaposhnik:
Implement local part of devfile feature (not related to remote repository fetching via git URL etc) which includes:
- model files  (jackson-annotated)
- REST endpoints (create workspace from yaml, export workspace into yaml), 
- converters of devfile to/from workspace config 
- devfile valitator, validation exceptions;
etc...

```java
Path("devfile")
public class DefFileService extends Service  {

// generate a workspace by sending a devfile to a rest API
// Initially this method will return empty workspace configuration.
// And will start che-devfile-broker on a background to clone sources and get devfile.
@POST
@Consumes("text/yml")
@Produces(APPLICATION_JSON)
public Response createFromYaml(DevFile defFile){
...
}
```
Example of yaml functionality
```yaml
---
specVersion: 0.0.1
name: petclinic-dev-environment
projects:
  - name: petclinic
    source:
      type: git
      location: 'git@github.com:spring-projects/spring-petclinic.git'
tools:
  - name: theia-ide
    definition:
      chePlugin:
        name: theia:0.0.3
  - name: jdt.ls
    definition:
      chePlugin:
        name: theia-jdtls:0.0.3
commands:
  - name: build
    toolsCommands:
      - tool: mvn-stack
        action:
          exec:
            command: mvn package
            workdir: /projects/spring-petclinic 
  - name: run
    toolsCommands:
      - tool: mysql
        action: start
      - tool: mvn-stack
        action:
          exec:
            command: mvn spring-boot:run
            workdir: /projects/spring-petclinic 
```

benoitf:
hello, do we already have complete examples of the new model ?

mshaposhnik:
@benoitf https://github.com/eclipse/che/issues/11719#issuecomment-433389404

sleshchenko:
> @benoitf #11719 (comment)

@mshaposhnik I think it makes sense to duplicate plan what should be done in PR description or at least specify the comment in the description. Otherwise, DoD of the issue is not clear enough.

skabashnyuk:
@mshaposhnik fixed

mshaposhnik:
Current Che7 command configuration is:
```
{ 
"commandLine": "mvn -f ${current.project.path} clean install", 
"name": "console-java-simple:build", 
"attributes": { 
          "goal": "Build", 
          "machineName": "...", 
          "previewUrl": "" 
          }
}
```
[...]

PR #11998, opened 2018-11-20
Devfile local features implenentation with schema validation and automated model build.

mshaposhnik:
### What does this PR do?
Initial implementation proposal for DevFile local part --  REST,  converter,  JSON schema prototype, validation and automated build of model POJO-s.

https://github.com/eclipse/che/issues/11832

#### Docs PR
N/A

<details>
  <summary>Yaml example:</summary><p>

```
version: 0.0.1
projects:
  - name: petclinic
    source:
      type: git
      location: 'git@github.com:spring-projects/spring-petclinic.git'
tools:
  - name: mvn-stack
    type: chePlugin
    id: eclipse/maven-jdk8:1.0.0
  - name: theia-ide
    type: cheEditor
    id: eclipse/theia:0.0.3
  - name: jdt.ls
    type: chePlugin
    id: eclipse/theia-jdtls:0.0.3
commands:
  - name: build
    actions:
      - type: exec
        tool: mvn-stack
        command: mvn package
        workdir: /projects/spring-petclinic
  - name: run
    attributes:
      runType: sequential
    actions:
      - type: exec
        tool: mvn-stack
        command: mvn spring-boot:run
        workdir: /projects/spring-petclinic
  - name: other
    actions:
      - type: exec
        tool: jdt.ls
        command: run.sh
```
</details>

<details>
  <summary>WS config after conversion example:</summary><p>

```
{
  "projects": [
    {
       "source": {
        "location": "git@github.com:spring-projects/spring-petclinic.git",
        "type": "git",
        "parameters": {}
      },
      "mixins": [],
      "name": "petclinic",
      "path": "/petclinic",
      "attributes": {}
    }
  ],
  "commands": [
    {
      "commandLine": "mvn package",
      "name": "build:mvn-stack",
      "type": "exec",
      "attributes": {
        "pluginId": "eclipse/maven-jdk8:1.0.0",
        "workingDir": "/projects/spring-petclinic"
      }
    },
    {
      "commandLine": "mvn spring-boot:run",
      "name": "run:mvn-stack",
      "type": "exec",
      "attributes": {
        "pluginId": "eclipse/maven-jdk8:1.0.0",
        "runType": "sequential",
        "workingDir": "/projects/spring-petclinic"
      }
    },
    {
      "commandLine": "run.sh",
      "name": "other:jdt.ls",
      "type": "exec",
      "attributes": {
        "pluginId": "eclipse/theia-jdtls:0.0.3"
      }
    }
  ],
  "environments": {
    "default": {
      "recipe": {
        "type": "dockerimage",
        "content": "eclipse/ubuntu_jdk8"
      },
      "machines": {
        "dev-machine": {
          "installers": [],
          "servers": {},
          "env": {},
          "volumes": {},
          "attributes": {
            "memoryLimitBytes": "2147483648"
          }
        }
      }
    }
  },
  "defaultEnv": "default",
  "name": "petclinic-dev-environment",
  "attributes": {
    "editor": "eclipse/theia:0.0.3",
    "plugins": "eclipse/maven-jdk8:1.0.0,eclipse/theia-jdtls:0.0.3",
    "toolsAliases": "eclipse/maven-jdk8:1.0.0=mvn-stack,eclipse/theia:0.0.3=theia-ide,eclipse/theia-jdtls:0.0.3=jdt.ls"
  }
}
```

</p>
</details>


wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevFileConverter.java
```diff
@@ -0,0 +1,191 @@
...
+import org.eclipse.che.api.workspace.server.model.impl.EnvironmentImpl;
+import org.eclipse.che.api.workspace.server.model.impl.MachineConfigImpl;
+import org.eclipse.che.api.workspace.server.model.impl.ProjectConfigImpl;
+import org.eclipse.che.api.workspace.server.model.impl.RecipeImpl;
+import org.eclipse.che.api.workspace.server.model.impl.SourceStorageImpl;
+import org.eclipse.che.api.workspace.server.model.impl.WorkspaceConfigImpl;
+
+public class DevFileConverter {
+
+  static Devfile workspaceToDevFile(WorkspaceConfigImpl wsConfig) {
```
sleshchenko: Please make methods of this class non-static. It would allow to bind any implementation and make easier testing of classes which use it.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
```diff
@@ -546,3 +546,6 @@ che.server.secure_exposer.jwtproxy.memory_limit=128mb
 # Maximum size of the json processing pool
 # in case if pool size would be exceeded message execution will be rejected
 che.core.jsonrpc.processor_max_pool_size=100
+
+# Devile JSON schema location (relative to api.war resources folder)
+che.devfile.schema.file_location=schema/devfile.json
```
skabashnyuk: new line

wsmaster/che-core-api-devfile/pom.xml
```diff
@@ -0,0 +1,149 @@
...
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.dataformat</groupId>
+            <artifactId>jackson-dataformat-yaml</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.github.java-json-tools</groupId>
+            <artifactId>json-schema-validator</artifactId>
+            <version>2.2.10</version>
```
skabashnyuk: move to parent

skabashnyuk (commented):
in general looks good to me

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevFileService.java
```diff
@@ -0,0 +1,125 @@
...
+import org.eclipse.che.api.core.rest.Service;
+import org.eclipse.che.api.devfile.model.Devfile;
+import org.eclipse.che.api.workspace.server.WorkspaceLinksGenerator;
+import org.eclipse.che.api.workspace.server.WorkspaceManager;
+import org.eclipse.che.api.workspace.server.model.impl.WorkspaceImpl;
+import org.eclipse.che.api.workspace.shared.dto.WorkspaceDto;
+import org.eclipse.che.commons.env.EnvironmentContext;
+
+@Path("/devfile")
+public class DevFileService extends Service {
```
metlos: Since the devfile schema location is configurable and therefore can actually be replaced with something different, maybe we should also add an endpoint to retrieve the schema, so that users can do the validation locally before submitting the file to Che.

wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json
```diff
@@ -0,0 +1,206 @@
...
+    "name",
+    "projects",
+    "tools",
+    "commands"
+  ],
+  "properties": {
+    "version": {
+      "$id": "#/properties/version",
+      "type": "string",
+      "title": "The Version Schema",
```
metlos: "schema version"?

wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json
```diff
@@ -0,0 +1,206 @@
+{
```
metlos, line 1: The `$id` and `title` are IMHO optional and bring no value in this form, which seems autogenerated. I'd instead add `description` to the individual schema elements with more detailed comments on their behavior/intended use.

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevFileService.java
skabashnyuk: +1

wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json
```diff
@@ -0,0 +1,206 @@
...
+              "petclinic"
+            ]
+          },
+          "source": {
+            "$id": "#/properties/projects/items/properties/source",
+            "type": "object",
+            "title": "The Project Source Schema",
+            "required": [
+              "type",
+              "location"
```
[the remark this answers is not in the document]
mshaposhnik: For now it is required in che project configuration, so we cant make it optional yet.

wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json
```diff
@@ -0,0 +1,206 @@
...
+            ],
+            "properties": {
+              "type": {
+                "$id": "#/properties/projects/items/properties/source/properties/type",
+                "type": "string",
+                "title": "The Project Source Type Schema",
+                "examples": [
```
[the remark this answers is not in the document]
mshaposhnik: Done

sleshchenko (commented):
I think placing `"pluginRef": "tollName"` in workspace attributes config is a wrong way to store tools aliases.
Instead of it, I think it would be better to have one attribute `"toolsAliases": "pluginRef1=toolName1,pluginRef2=toolName2,pluginRef3=toolName3"`. Why I think it would be better:
1. Now almost all of our attributes are listed and have documentation like `ServerConfig` attributes https://github.com/eclipse/che/blob/43066dcd220823562dbf3b7f6a50d375f1558924/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/config/ServerConfig.java#L26-L52 `WorkspaceConfigAttributes `https://github.com/eclipse/che/blob/aaa8f424cd198e68e128bca939d9ac058b9ea52e/wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java#L81-L120
With placing `"pluginRef": "tollName"` in attributes it looks like dump.
2. If we need to get rid or migrate these attributes it would be easier to do it when all of them are placed in one attributes otherwise, you check all attributes one by one and guess if attribute key is tool aliases or not by checking format, which does not guarantee that it is pluginRef.
I understand bad side of the proposed way: you should parse `toolsAliases` value each time during converting, but I think it is not an operation that will be performed often and it's OK to do it.

wsmaster/che-core-api-devfile/pom.xml
```diff
@@ -0,0 +1,156 @@
...
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-workspace-shared</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-commons-json</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.everrest</groupId>
+            <artifactId>everrest-core</artifactId>
```
sleshchenko, line 118: I think it is copied/pasted from another api module and you don't really need it, the same as ignoring this dependency here https://github.com/eclipse/che/pull/11998/files#diff-e6b2dd50f4fc644955b5ecd734663eefR148

wsmaster/che-core-api-devfile/pom.xml
```diff
@@ -0,0 +1,156 @@
...
+        </dependency>
+        <dependency>
+            <groupId>javax.servlet</groupId>
+            <artifactId>javax.servlet-api</artifactId>
```
sleshchenko: I don't see any usages of `javax.servlet-api`. I think you can just remote it.

wsmaster/che-core-api-devfile/pom.xml
```diff
@@ -0,0 +1,156 @@
...
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-workspace</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-workspace-shared</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-commons-json</artifactId>
```
sleshchenko, line 108: Classes of this dependency is used only in tests classes. So, please change its scope to `test`, otherwise `mvn clean install -Dmaven.test.skip=true` will fail with the following error 
```
[INFO] 
[INFO] --- maven-dependency-plugin:3.0.1:analyze-only (analyze) @ che-core-api-devfile ---
[WARNING] Unused declared dependencies found:
[WARNING]    org.eclipse.che.core:che-core-commons-json:jar:6.15.0-SNAPSHOT:compile
[INFO] BUILD FAILURE
[INFO] Total time: 7.285 s
[INFO] Finished at: 2018-11-29T12:15:38+02:00
```

@skabashnyuk It's an issue, because periodically build without tests is broken by someone. Can we configure dependency plugin not to use test compiled classes during analyzing?

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevFileService.java
```diff
@@ -0,0 +1,148 @@
...
+import org.eclipse.che.api.devfile.model.Devfile;
+import org.eclipse.che.api.workspace.server.WorkspaceLinksGenerator;
+import org.eclipse.che.api.workspace.server.WorkspaceManager;
+import org.eclipse.che.api.workspace.server.model.impl.WorkspaceConfigImpl;
+import org.eclipse.che.api.workspace.server.model.impl.WorkspaceImpl;
+import org.eclipse.che.api.workspace.shared.dto.WorkspaceDto;
+import org.eclipse.che.commons.env.EnvironmentContext;
+
+@Path("/devfile")
+public class DevFileService extends Service {
```
sleshchenko: Are you going to add tests for this class?

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevFileService.java
```diff
@@ -0,0 +1,148 @@
...
+   * namespace/workspace_name
+   *
+   * @see WorkspaceManager#getByKey(String)
+   */
+  @GET
+  @Path("/{key:.*}")
+  @Produces("text/yml")
+  public Response createFromWorkspace(@PathParam("key") String key)
+      throws NotFoundException, ServerException {
+    // TODO: validate key
```
sleshchenko: Just reminding: Don't forget to fix TODO before a merge

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevFileService.java
```diff
@@ -0,0 +1,148 @@
...
+  @Consumes({"text/yaml", "text/x-yaml", "application/yaml"})
+  @Produces(APPLICATION_JSON)
+  public WorkspaceDto createFromYaml(String data, @QueryParam("verbose") boolean verbose)
+      throws ServerException, ConflictException, NotFoundException, ValidationException,
+          BadRequestException {
+
+    Devfile devFile;
+    WorkspaceConfigImpl workspaceConfig;
+    try {
+      schemaValidator.validateBySchema(data, verbose);
```
sleshchenko: 1. You parse input data twice, the first time during validation and the second time during Parsing object to Devfile. Maybe there is a possible way to avoid doing it.
2. IMHO Service would look better if we move serialization/deserizalization logic to `MessageBodyWriter`, `MessageBodyReader` like we do with our DTO's and CheJsonProvider. You can take a look how it is implemented here https://github.com/eclipse/che/blob/07263f1e30089689d71b057f747a44a29283e3c4/wsagent/che-core-api-debug/src/main/java/org/eclipse/che/api/debugger/server/DebuggerActionProvider.java#L52

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevFileService.java
mshaposhnik: Added

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileSchemaCachedProvider.java
```diff
@@ -0,0 +1,56 @@
...
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.lang.ref.SoftReference;
+import java.util.stream.Collectors;
+import javax.inject.Singleton;
+import org.eclipse.che.api.core.ServerException;
+
+/** Loads a schema content and stores it in soft reference. */
+@Singleton
+public class DevfileSchemaCachedProvider {
```
skabashnyuk: What if we do like this ?
```java
import com.fasterxml.jackson.databind.JsonNode;
import com.google.common.io.Resources;
import com.google.inject.Singleton;

import java.io.IOException;
import java.lang.ref.SoftReference;
import java.nio.charset.Charset;

@Singleton
public class DefFileSchemaProvider {
  private final String SCHEMA_LOCATION = "";

  private SoftReference<String> schemaRef = new SoftReference(null);

  public String getString() throws IOException {
    String schema = schemaRef.get();
    if (schema == null) {
      synchronized (this) {
        if (schema == null) {
          schema =
              Resources.asCharSource(
                      Resources.getResource(SCHEMA_LOCATION), Charset.defaultCharset())
                  .read();
          schemaRef = new SoftReference(schema);
        }
      }
    }
    return schema;
  }

  public JsonNode getJsoneNode() throws IOException {
    return JsonLoader.fromString(getString());
  }
}
```

wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json
mshaposhnik: Removed id and reduntant titles, added descriptions

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileSchemaCachedProvider.java
mshaposhnik: 1. this is the strangest way to load file i've ever saw :)
2. don;t think double-checked-locking is needed here. there is no danger to load wrong/different data for the threads;
3. agree that this provider shold be used for validation also, nice idea.

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevFileService.java
mshaposhnik: Added

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevFileService.java
mshaposhnik: 1 st - fixed;
2 nd - dont think it's needed right now, since we constructing always the same object, and its creation is pretty simple. (in your example, different objects created depending of souce data).
We may probably need that when different versions of devfile will be introduced. (to know which object to create) so propose to review it later.

mshaposhnik:
ci-build

mshaposhnik:
ci-build

mshaposhnik:
ci-build

mshaposhnik:
ci-build

riuvshin:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1259/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1259//Selenium_20tests_20report/)
docker image: eclipseche/che-server:11998

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevFileSchemaProvider.java
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  public JsonNode getJsoneNode() throws IOException {
+    return JsonLoader.fromString(getSchemaContent());
+  }
+
+  private String loadFile() throws IOException {
+    URL schemaURL = getClass().getClassLoader().getResource(SCHEMA_LOCATION);
+    if (schemaURL != null) {
+      return new String(Files.readAllBytes(Paths.get(schemaURL.getFile())));
```
skabashnyuk: >new String(Files.readAllBytes(Paths.get(schemaURL.getFile())))
new String(Files.readAllBytes(Paths.get(schemaURL.getFile())), Charset.defaultCharset())

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevFileService.java
```diff
@@ -0,0 +1,189 @@
...
+        nameCandidate = config.getName() + "_" + ++counter;
+      } catch (NotFoundException nf) {
+        config.setName(nameCandidate);
+        break;
+      }
+    }
+    return config;
+  }
+
+  private void validateKey(String key) throws BadRequestException {
```
skabashnyuk: Extract this code to separate class WokspaceKeyValidator and create static method validateKey and reuse it here and in WorkspaceService

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevFileSchemaProvider.java
skabashnyuk: >new String(Files.readAllBytes(Paths.get(schemaURL.getFile())))

To make findbugs happy
new String(Files.readAllBytes(Paths.get(schemaURL.getFile())), Charset.defaultCharset())

wsmaster/che-core-api-devfile/src/test/resources/workspace_config.json
```diff
@@ -0,0 +1,51 @@
...
+      "type": "exec",
+      "attributes": {
+        "pluginId": "eclipse/theia-jdtls:0.0.3"
+      }
+    }
+  ],
+  "environments": {},
+  "name": "petclinic-dev-environment",
+  "attributes": {
+    "toolsAliases": "eclipse/maven-jdk8:1.0.0=mvn-stack,eclipse/theia:0.0.3=theia-ide,eclipse/theia-jdtls:0.0.3=jdt.ls",
```
skabashnyuk, line 47: Please update description example to reflect latest changes.

skabashnyuk (commented):
overall looks godd

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevFileConverter.java
```diff
@@ -0,0 +1,199 @@
...
+                .withId(editorId)
+                .withName(findToolName(wsConfig, editorId));
+        tools.add(editorTool);
+      } else if (entry.getKey().equals("plugins")) {
+        for (String pluginId : entry.getValue().split(",")) {
+          Tool pluginTool =
+              new Tool()
+                  .withId(pluginId)
+                  .withType("chePlugin")
+                  .withName(findToolName(wsConfig, pluginId));
```
sleshchenko: You parse tools aliases N times, where N is plugins number. Please consider parsing them once and save in Map.

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevFileConverter.java
```diff
@@ -0,0 +1,199 @@
...
+    projectConfig.setPath("/" + projectConfig.getName());
+    SourceStorageImpl sourceStorage = new SourceStorageImpl();
+    sourceStorage.setType(devProject.getSource().getType());
+    sourceStorage.setLocation(devProject.getSource().getLocation());
+    projectConfig.setSource(sourceStorage);
+    return projectConfig;
+  }
+
+  private String findToolName(WorkspaceConfigImpl wsConfig, String toolId) {
+    String aliasesString = firstNonNull(wsConfig.getAttributes().get("toolsAliases"), "");
```
sleshchenko: Please move `"toolsAliases"` somewhere to constant and describe where it is supposed to be used and describe format there. I think Constants in dev file module is good place

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevFileConverter.java
```diff
@@ -0,0 +1,199 @@
...
+import org.eclipse.che.api.devfile.model.Devfile;
+import org.eclipse.che.api.devfile.model.Project;
+import org.eclipse.che.api.devfile.model.Source;
+import org.eclipse.che.api.devfile.model.Tool;
+import org.eclipse.che.api.workspace.server.model.impl.CommandImpl;
+import org.eclipse.che.api.workspace.server.model.impl.ProjectConfigImpl;
+import org.eclipse.che.api.workspace.server.model.impl.SourceStorageImpl;
+import org.eclipse.che.api.workspace.server.model.impl.WorkspaceConfigImpl;
+
+/** Helps to convert Devfile into workspace config and back. */
```
sleshchenko: Java docs says `Devfile` but class is named `DevFile`, also model object is `Devfile`. Let's decide if we introduce `dev file` or `devfile`

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevFileConverter.java
```diff
@@ -0,0 +1,199 @@
...
+public class DevFileConverter {
+
+  public Devfile workspaceToDevFile(WorkspaceConfigImpl wsConfig) {
```
sleshchenko: What about converting workspace config which contains recipe? As far as I understand it should be unsupported for time being, but this method designed not to throw an exception and doesn't have the corresponding check.

mshaposhnik: Which exception do you propose to throw? `UnsupportedOperation` ? And in which case? If there is `defaultEnv` set and/or map of environments is not empty ?

sleshchenko: @mshaposhnik 
> Which exception do you propose to throw? `UnsupportedOperation` ?

`UnsupportedOperation` definitely not. I would propose to introduce some new checked exception. 

> `And in which case? If there is defaultEnv set and/or map of environments is not empty ?`

If there is any information in workspace config that can not be stored in Devfile. Like any `environment` is configured now,  `multiply-environments` later.
In other words if `WorkspaceConfig 1` -> `Devfile` -> `WorkspaceConfig 2` and there is no ability to provide `WorkspaceConfig 1` == `WorkspaceConfig 2`
I may be wrong here @skabashnyuk WDYT?

garagatyi (approved):
Looks good

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileConverter.java
```diff
@@ -0,0 +1,213 @@
...
+    List<ProjectConfigImpl> projects = new ArrayList<>();
+    devFile.getProjects().forEach(project -> projects.add(devProjectToProjectConfig(project)));
+    config.setProjects(projects);
+
+    // Manage tools
+    Map<String, String> attributes = new HashMap<>();
+    StringJoiner pluginsStringJoiner = new StringJoiner(",");
+    StringJoiner toolIdToNameMappingStringJoiner = new StringJoiner(",");
+    for (Tool tool : devFile.getTools()) {
+      if (tool.getType().equals("cheEditor")) {
```
sleshchenko: Should not an exception be thrown if a tool has an unknown type, for now, any not equals `cheEditor` nor `chePlugin`?

skabashnyuk: ```+```

mshaposhnik:
ci-build

mshaposhnik:
ci-build

mshaposhnik:
ci-test

riuvshin:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1274/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1274//Selenium_20tests_20report/)
docker image: eclipseche/che-server:11998

mshaposhnik:
ci-test

skabashnyuk (approved):

riuvshin:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1277/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1277//Selenium_20tests_20report/)
docker image: eclipseche/che-server:11998

PR #12004, opened 2018-11-20
build.sh become working on Ubuntu / Linux Mint

vitaliy-guliy:
Fixes problem with running build.sh script on Ubuntu / Linux Mint

Issue https://github.com/eclipse/che/issues/10041


.gitignore
```diff
@@ -65,6 +65,7 @@ Thumbs.db
 .che
 dockerfiles/che/eclipse-che.tar.gz
 dockerfiles/che/eclipse-che/
+dockerfiles/theia/e2e/.Dockerfile
```
sleshchenko: It is not related to `Use bash instead of sh`. Can you elaborate on why you did this change?

riuvshin: yeah, this file can be ignored as it is generated during the build and if build fails this file remain in the repo.

riuvshin (approved):

benoitf (commented):
-0 I still don't understand why turning to bash while it should just fix the issue to use sh
It's just performing workarounds

riuvshin:
ci-build

riuvshin (approved):

.gitignore
sleshchenko: It's kind of bug in `build.include`. And this issue appears during a build of all images - Che-Server, Postgres, Che-Dir, etc. So, it should be fixed for all. Possible solution - name a generated file like `.Dockerfile.genarated` and add it to `.gitignore` with a regular expression.
Please revert your change in this PR and create a separate issue for this bug - welcome to fix it in proposed way ;)

sleshchenko:
@benoitf The issue which I see with `sh` it's difficult to test it. When `>>>` was introduced - everything works just fine BUT only of most of the environment. 
What is the way to ensure that everything works fine after changing builds scripts with `sh`?

vitaliy-guliy:
.gitignore has been reverted

sleshchenko (approved):

benoitf:
@sleshchenko it's to use a sh implementation, not bash
AFAIK dash is one of them on linux.

benoitf:
note that using bash is adding 10MB just for some `>>>`
```
$ docker run --rm -it alpine
/ # du -s -h /
4.5M	/


/ # apk add --no-cache --update bash
fetch http://dl-cdn.alpinelinux.org/alpine/v3.7/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.7/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.7/community/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.7/community/x86_64/APKINDEX.tar.gz
(1/6) Installing pkgconf (1.3.10-r0)
(2/6) Installing ncurses-terminfo-base (6.0_p20171125-r1)
(3/6) Installing ncurses-terminfo (6.0_p20171125-r1)
(4/6) Installing ncurses-libs (6.0_p20171125-r1)
(5/6) Installing readline (7.0.003-r0)
(6/6) Installing bash (4.4.19-r1)
Executing bash-4.4.19-r1.post-install
Executing busybox-1.27.2-r6.trigger
OK: 13 MiB in 17 packages
/ # du -s -h /
14.8M	/
```

riuvshin:
but we are not installing bash in any container, these scripts are used by devs locally and on CI.
CLI scripts remain `sh`

PR #12029, opened 2018-11-22
rh-che #1075: Adding 'ImpersonatorInterceptor' during creation of KubernetesClient

ibuziuk:
### What does this PR do?
Adding 'ImpersonatorInterceptor' during creation of KubernetesClient
Currently `ImpersonatorInterceptor` is added only during creation of `OpenShiftClient`, but not `KubernetesClient`

https://github.com/eclipse/che/blob/44588aa775e77f629eb0a0052ac20d949ab8c7c3/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/OpenShiftClientFactory.java#L214

https://github.com/redhat-developer/rh-che/issues/1075

#### Release Notes
Adding 'ImpersonatorInterceptor' during creation of KubernetesClient

#### Docs PR
N/A


davidfestal (approved):

sleshchenko (approved):

sleshchenko:
@ibuziuk I noticed that Github automatically references issue/PR in the current repository, so it can be confusing
screenshot_20181123_111147
Maybe using rh-che-1075 would be better, maybe some another format. I it would be better to avoid such confusing references.

ibuziuk:
@sleshchenko good point. I will not refer issue outside the repo via # in commit messages in order to avoid confusion.

PR #12070, opened 2018-11-29
[dockerfile] [theia] Simplify the build

benoitf:
### What does this PR do?
By adding che packages in the current theia cloned folder and adding che assembly in examples folder

N/A

#### Release Notes
N/A

#### Docs PR
N/A


dockerfiles/theia/src/builder/production-mode.sh
```diff
@@ -13,14 +13,4 @@
...
-LAST_SIZE=${SIZE_BEFORE}
-
-cd ${HOME}
-yarn install --frozen-lockfile --no-cache --production
-
-# Size after
-SIZE_AFTER=$(du -s . | cut -f1)
-
-echo "Current gain $((${SIZE_BEFORE}-${SIZE_AFTER}))"
+cd ${HOME}/theia-source-code && che:theia production
```
AndrienkoAleksandr, line 16: New line

AndrienkoAleksandr (approved):

mmorhun (approved):

dockerfiles/theia/Dockerfile
```diff
@@ -33,50 +33,31 @@ ENV THEIA_VERSION=${THEIA_VERSION} \
...
 # Add patches
 ADD src/patches /home/theia-build/patches
 
 # Invalidate cache if there is a new commit
 ADD https://api.github.com/repos/theia-ide/theia/git/refs/tags/v${THEIA_VERSION} /tmp/sha1-repository
 # Build theia from source code and publish artifacts locally
-ADD src/builder/build-and-publish-theia.sh /tmp/builder/scripts/build-and-publish-theia.sh
-RUN /tmp/builder/scripts/build-and-publish-theia.sh
+ADD src/builder/clone.sh /tmp/builder/scripts/clone.sh
+RUN /tmp/builder/scripts/clone.sh
```
sunix, line 46: Would it be ok to (In this PR)
-  create a "dockerfiles/theia-builder" with the builder + theses scripts (somewhere else then /tmp/, in the $PATH maybe), without running them
- In here (multistage build) we would just be based from the previous image eclipse/che-theia-builder and call the scripts (clone.sh, yarn production.sh etc ...
WDYT ?

benoitf: it will be another PR (to have another image for the build tools)
but IMHO in any case it should be a multi-staged build to be sure to have the less layer than possible for the runtime (the smallest image)
it's not a multi-staged build, it means that all actions will be done again each time you change anything

benoitf: BTW I also think that we don't need sh scripts in builder image. (can be discussed there)
but the 'npm script' is handling it (or should handle other stuff)

benoitf:
cc @davidfestal

dockerfiles/theia/Dockerfile
sunix: yes but at some point you may also want to try a 'production' build in che and run it in a production container (no need to rebuild the production image,just share the right volume)

sunix (approved):
LGTM i have just rebuilt it locally and looks fine ... and faster

dockerfiles/theia/Dockerfile
sunix: > it will be another PR (to have another image for the build tools)

OK

> but IMHO in any case it should be a multi-staged build to be sure to have the less layer than possible for the runtime (the smallest image)
> it's not a multi-staged build, it means that all actions will be done again each time you change anything

Yes agree. We would keep the multistaged build:
```
FROM eclipse/che-theia-builder as builder
..... run scripts clone, etc ...
....

riuvshin (approved):

Issue #12091, opened 2018-12-02
Create REST method to get all workspaces with filter


skabashnyuk:
### Description
For some  diagnostics and admins task, it would be nice to have a method to get all workspaces with filter
- Filter by workspace status.
- Filter for how long workspace is running.
- etc?

metlos:
Note to self: 
* [ ] authz in multiuser

PR #12139, opened 2018-12-06
[WIP]Distintiguish between workspace start and update.

metlos:
### What does this PR do?
This makes the server store the timestamp of a workspace start as a separate attribute. Before start and update of a workspace were both tracked by a single attribute called "updated".

This PR will enable us to track the long running workspaces in #12091.

Not sure about the impact of this change. Is this part of the public API? It seems to have been part of the contract forever.

#12091 

#### Release Notes

The start time of a workspace is now stored as "started" attribute of a workspace. The "updated" attribute now only contains the timestamp of the last update of the workspace.

### DIFF
--- a/wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
+++ b/wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
@@ -57,9 +57,21 @@ public final class Constants {
    * Workspace#getAttributes}
    */
   public static final String CREATED_ATTRIBUTE_NAME = "created";
+
   /**
-   * Describes time when workspace was last updated or started. Should be set/read from {@link
-   * Workspace#getAttributes}
+   * Describes the time when the workspace was last started. Note that this is not reset when the
+   * workspace is stopped so that the last running time of a stopped workspace can be determined.
+   *
+   * <p>Should be set/read from {@link Workspace#getAttributes()}.
+   */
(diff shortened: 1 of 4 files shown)


riuvshin:

riuvshin:

riuvshin:

garagatyi (approved):
Looks good

garagatyi:
@davidfestal Please, check whether it affects Woopra in any way

benoitf:
I think it may affect all clients like dashboard that displays workspace in order of usage

metlos:
k, so what about splitting it like this:

* keep `updated` as is - tracks update or start
* add `started` to track the last started time
* add `modified` to track the last modification time (if this is needed at all)

skabashnyuk:
@metlos +1
we can 
 - for one or more releases, until we make sure other clients are not affected keep updated as is
 - not we can introduce  ```started``` to track the last started time
 - add modified to track the last modification time  - not sure 
- in one or two releases change or remove updated

garagatyi:
And who will be working on checking clients?

skabashnyuk:
>And who will be working on checking clients?

Work as usual with potentially breaking changes.
1. Notifying community about the change and a reason.
2. Giving some time to check and adapt.
3. In a reasonable time introduce change. 

Anyway, I'm not sure we have a proper proposal yet. I would like to work more on this task.

metlos:
I'm closing this PR because the proposed change is no longer needed. We found a better solution to the problem by extending the functionality around workspace expiration (now called workspace activity).

PR #12148, opened 2018-12-07
Add attributes to Che 7 plugin meta; make URL field not mandatory

garagatyi:
### What does this PR do?
Add attributes to Che 7 plugin meta; make URL field not mandatory.
This is needed for implementation of VS Code extension broker.

### DIFF
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/wsplugins/model/PluginMeta.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/wsplugins/model/PluginMeta.java
@@ -11,6 +11,8 @@
  */
 package org.eclipse.che.api.workspace.server.wsplugins.model;
 
+import java.util.HashMap;
+
 /** @author Oleksandr Garagatyi */
 public class PluginMeta {
   private String name = null;
@@ -21,6 +23,7 @@ public class PluginMeta {
   private String description = null;
   private String icon = null;
   private String url = null;
+  private HashMap<String, String> attributes = new HashMap<>();
 
   public PluginMeta name(String name) {
     this.name = name;
@@ -92,5 +95,17 @@ public class PluginMeta {
 
   public String getDescription() {
     return description;
+  }
+
+  public HashMap<String, String> getAttributes() {
+    if (attributes == null) {
+      attributes = new HashMap<>();
+    }
+    return attributes;
+  }
+
+  public PluginMeta attributes(HashMap<String, String> attributes) {
+    this.attributes = attributes;
+    return this;
   }
 }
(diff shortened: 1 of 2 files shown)


garagatyi:
ci-test

riuvshin:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1279/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1279//Selenium_20tests_20report/)
docker image: eclipseche/che-server:12148

skabashnyuk (approved):

benoitf (commented):
IMHO it looks strange to have check for null attributes on getter while it's initialized on class to a value.  Probably check on setter or not have hashmap initialized when defining the attributes ?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/wsplugins/model/PluginMeta.java
```diff
@@ -93,4 +96,16 @@ public PluginMeta description(String description) {
   public String getDescription() {
     return description;
   }
+
+  public HashMap<String, String> getAttributes() {
```
sleshchenko: Why this method returns `HashMap` but not `Map`?

garagatyi: fixed, 10x

garagatyi:
@benoitf we have this type of code all over the place in Che. I agree that approach suggested by you might be better. Would be better to discuss this as a separate topic with those who write server-side components and agree on what approach to use everywhere.

garagatyi:
@benoitf As far as I remember we had issues with equals on POJOs before. In a case, when one POJO instance is initialized with null or without value and another one has initialized with empty map these objects are not equal. But we never return null as a collection, so when such a POJO is used it should be an equal object.

So, we either need to ensure that collection is initialized on getter and use it in equals method or we need to initialize it with an empty map and reinitialize it on setting null as a map.
It would be better to have just one approach for such a situation because we have to use a lot of POJOs. Now we use this approach all over the place and you consider it strange. Imagine a situation when different POJOs would use different techniques - would be way more strange and error-prone.
I'm OK to change the approach in general but it is a thing that goes beyond purpose of this PR.

garagatyi:
ci-test

riuvshin:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1287/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1287//Selenium_20tests_20report/)
docker image: eclipseche/che-server:12148

PR #12171, opened 2018-12-11
[docker] [theia] Use of theia-dev image in theia image

benoitf:
### What does this PR do?
Re-use the theia-dev as builder image when building theia image
remove unused scripts

N/A

#### Release Notes
N/A

#### Docs PR
N/A

Change-Id: I0aec5d9649d4571045c36675496b7066b41de87e
Signed-off-by: Florent Benoit <fbenoit@redhat.com>


AndrienkoAleksandr (approved):

dockerfiles/theia/Dockerfile
```diff
@@ -11,54 +11,62 @@
...
 
-ENV THEIA_VERSION=${THEIA_VERSION} \
-    HOME=/home/theia
-WORKDIR ${HOME}
-
-# add eclipse che theia generator
-RUN yarn global add @eclipse-che/theia-generator@0.0.1-1543568125
-
+# Invalidate cache if there is a new commit
+ADD https://api.github.com/repos/theia-ide/theia/git/refs/tags/v${THEIA_VERSION} /tmp/.sha1-repository
```
mmorhun, line 25: Maybe we should invalidate cache after checking github limit?

dockerfiles/theia/Dockerfile
```diff
@@ -11,54 +11,62 @@
...
+      for file in $(find "${HOME}/patches/${THEIA_VERSION}" -name '*.patch'); do \
+        echo "Patching with ${file}"; \
+        cd ${HOME}/theia-source-code && patch -p1 < ${file}; \
+      done \
+    fi
 
 # Generate che-theia
-RUN cd ${HOME}/theia-source-code && che:theia init
+WORKDIR ${HOME}/theia-source-code
+RUN che:theia init
```
mmorhun, line 60: Doesn't it clone Theia?

mmorhun (approved):

dockerfiles/theia/Dockerfile
benoitf: the thing is that if we invalidate cache after checking limit, it means that we won't check limit on a second run (layers cached) so it'll probably fail when compiling. I don't know the best option

dockerfiles/theia/Dockerfile
benoitf: no, currently init  doesn't clone. It only configures directories / clone extensions

sunix (approved):

riuvshin (approved):
looks good, please let me know if it require any changes on ci

Issue #12191, opened 2018-12-13
Prometheus and Grafana ingress is created even if metrics are not enabled


tarioch:
### Description
The helm chart always creates the ingress for prometheus and grafana even in the cases where the metrics are not enabled.

sleshchenko:
cc @metlos

PR #12192, opened 2018-12-14
Only deploy metrics ingress if metrics are enabled.

metlos:
### What does this PR do?

The `che-metrics-ingress` is now only deployed using the Che Helm chart when metrics are enabled (by setting the `global.metricsEnabled` property to true).

#12191


sleshchenko (approved):

skabashnyuk:
ci-build

l0rd (approved):

Issue #12208, opened 2018-12-17
Store statuses in KubernetesRuntimeState as string value instead of ordinal number


sleshchenko:
### Description
Now, Kubernetes Infrastructure stores statuses in KubernetesRuntimeState as ordinal number but not string value since `@Enumerated` annotation is missing:
https://github.com/eclipse/che/blob/4edd9190bfae77a40eabea1baf24d25289025177/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/model/KubernetesRuntimeState.java#L52-L53
https://github.com/eclipse/che/blob/57c43ac8cdd7111649414545a6ee9d156392d249/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/model/KubernetesMachineImpl.java#L54-L55
https://github.com/eclipse/che/blob/57c43ac8cdd7111649414545a6ee9d156392d249/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/model/KubernetesServerImpl.java#L39-L40

It is needed to make Kubernetes/OpenShift infrastructure store statuses in KubernetesRuntimeState as a string value instead of ordinal number. Because using ordinal number is potentially risky since enum items order changing may cause DB inconsistency.

PR #12177, opened 2018-12-12
Workspace activity metrics

metlos:
### What does this PR do?

This enhances the `/activity` REST endpoint with a `GET` method that reports on activity of workspaces, optionally applying criteria on the activities.

* [x] business logic for fetching activity stats
* [x] unit tests
* [x] integration tests
* [x] apply authz in multiuser env

### What issues does this PR fix or reference?

#12091 

#### Release Notes
* [x] https://github.com/eclipse/che-docs/pull/649

#### Docs PR
* [x] https://github.com/eclipse/che-docs/pull/648

PR #12222, opened 2018-12-18
Use the names of the enums instead of their ordinals when storing them to DB

metlos:
### What does this PR do?

This PR changes the way enums listed in the issue #12208 are stored. Previously, the enums were stored using their ordinal values even though the columns were of VARCHAR type. It is generally a safer practice to store enums using their names because then they don't break when the enum is reordered or new enum values added in between existing ones.

#12208 

### Note

This should only be merged after #12177 because it would break its DB upgrade script.


riuvshin:

riuvshin:

riuvshin:

metlos:
ci-build

metlos:
ci-test

metlos:
thought I'd be able to run the `ci-*` now that I'm a committer :cry:

skabashnyuk:
ci-test

skabashnyuk (approved):

sleshchenko (approved):
LGTM

metlos:
ci-build

metlos:
I've marked this as WIP just so that it is not accidentally merged. We will wait with this PR until #12177 is merged, at which point this PR will need to be modified with the actual needed numbering of the upgrade script.

Issue #12217, opened 2018-12-18
Dashboard factory: Remove mandatory aspect from workspace.projects[x].type property when creating a factory from a workspace


sunix:
### Description
In Che dashboard, when creating a workspace, in some case we can end up with a project without any 'type' defined (create a workspace, and import project later).
When creating a factory from this kind of workspace, Dashboard is preventing user to create it because the fields `workspace.projects[x].type` are mandatory.

![capture d ecran 2018-12-17 a 15 55 09](https://user-images.githubusercontent.com/650571/50147902-7e22ee80-02b7-11e9-8e01-0c16c59cdf1e.png)

The type is not mandatory for the user who is not using the GWT based editor (for instance Theia Editor)
GWT based editor will still work (propose to select a type at start up of the factory)

The goal of the issue is to be able to create a Factory from a workspace with 'untyped' projects

sleshchenko:
@sunix Dashboard is not a guy who throws this exception
It is shown by Che Server because the type field is marked as `mandatory`
https://github.com/eclipse/che/blob/57c43ac8cdd7111649414545a6ee9d156392d249/wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/dto/ProjectConfigDto.java#L53
Or I missed something.

mshaposhnik:
Hey @sunix you can close this because it was done in #12232

sunix:
thanks :+1:

Issue #12186, opened 2018-12-13
Devfile support in factory


skabashnyuk:
### Description
We need to support factories like ```http://localhost:8080/f?url=https://github.com/benoitf/simple-project``` that can point to github repo that contains Devfile in the root. Also, we need to support a use case then link points to Devfile itself.

slemeur:
can you link it to the epic for devfile?

Issue #12227, opened 2018-12-19
Factories belongs to unexisted users may present in DB


mshaposhnik:
### Description
Due to bug in cascade removal listeners, ([1], [2]) there can be orphaned factories present in DB.
Need to write DB  mirgation script to cleanup them.

[1]https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/jpa/JpaFactoryDao.java#L201

[2]https://github.com/eclipse/che/blob/master/multiuser/integration-tests/che-multiuser-cascade-removal/src/test/java/org/eclipse/che/multiuser/integration/jpa/cascaderemoval/JpaEntitiesCascadeRemovalTest.java#L347

sleshchenko:
Also, it makes sense rework Factory object to store `creatorId` in a separate field with a foreign key configured instead of storing it as an attribute.

PR #12232, opened 2018-12-19
Factory support of devfile

mshaposhnik:
### What does this PR do?
 - Added support of making factories from devfile on Github repository or free-hosted devfile;
 - Removed support of creating factories from dockerfile feature;
 - Che7 style (plugins-based) of default factories created from Github repository (without devfile or factory json);
- Contains fix for https://github.com/eclipse/che/issues/12217 to make type attribute not required;
 - General factory code renewal and bugfix;

https://github.com/eclipse/che/issues/12186


plugins/plugin-github/che-plugin-github-factory-resolver/src/main/java/org/eclipse/che/plugin/github/factory/resolver/LegacyGithubURLParser.java
```diff
@@ -31,10 +31,6 @@ public LegacyGithubURLParser(URLChecker urlChecker) {
   @Override
   public GithubUrl parse(String url) {
     GithubUrl githubUrl = super.parse(url);
-    if (!urlChecker.exists(githubUrl.dockerFileLocation())) {
-      githubUrl.withDockerfileFilename(".codenvy.dockerfile");
-    }
-
     if (!urlChecker.exists(githubUrl.factoryJsonFileLocation())) {
       githubUrl.withFactoryFilename(".codenvy.json");
```
mshaposhnik: Maybe its time to drop support of such? WDYT @skabashnyuk @benoitf

benoitf: Agree for codenvy files it was only something useful for avoid any breaks but now I would say this time is over

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/urlfactory/URLFactoryBuilder.java
```diff
@@ -0,0 +1,117 @@
...
+
+  /**
+   * Help to generate default workspace configuration
+   *
+   * @param name the name of the workspace
+   * @return a workspace configuration
+   */
+  public WorkspaceConfigDto buildDefaultWorkspaceConfig(String name) {
+
+    Map<String, String> attributes = new HashMap<>();
```
mshaposhnik, line 120: Make this configurable somehow? (properties..?)

benoitf:
hello, side note: we had a gitlab integration as well on codenvy. It was forgotten during the move to eclipse che ? 
https://github.com/codenvy/codenvy/tree/master/plugins/plugin-gitlab/codenvy-plugin-gitlab-factory-resolver

mshaposhnik:
for gitlab: I propose to make issue and let it being prioritized by PM's

benoitf:
@mshaposhnik yes sure, was pure question if it was intentional or just something we forgot.

mshaposhnik:
cant say for sure.

benoitf (approved):
nice feature, eager to use it :-)

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/urlfactory/URLFactoryBuilder.java
```diff
@@ -0,0 +1,117 @@
...
+    attributes.put(WORKSPACE_TOOLING_EDITOR_ATTRIBUTE, "org.eclipse.che.editor.theia:1.0.0");
```
benoitf: hello, is it possible to grab the latest version from the registry instead of using 1.0.0 ?
the same for machine exec plug-in ?

sleshchenko: Well, maybe then it should be possible to have the ability to configure workspace to use the latest editor/plugin version instead of grabbing the latest version from the registry in Factory related code?
Do you think it makes sense for workspace configuration as well?

benoitf: I don't remember the ticket issue but I saw discussion around using `latest` instead of `1.0.0` or make version field optional as well

Also, there is still a remaining question, should it be like a che server env var (with default value) the list of all plug-ins set by default ? or to reuse another factory definition or something else.
Let say tomorrow a new plug-in is enabled on some stacks, we need to recompile Che to make the update

sleshchenko: It definitely must be configured somehow. I think, two properties - editor + plugins list would be enough.

plugins/plugin-github/che-plugin-github-factory-resolver/src/main/java/org/eclipse/che/plugin/github/factory/resolver/GithubFactoryParametersResolver.java
```diff
@@ -69,27 +83,34 @@ public FactoryDto createFactory(@NotNull final Map<String, String> factoryParame
...
-    return projectConfigDtoMerger.merge(factory, projectConfigDto);
+    // apply merging operation from existing and computed settings if needed
+    return projectConfigDtoMerger.merge(
+        factory,
+        () -> {
+          // Compute project configuration
+          return newDto(ProjectConfigDto.class)
+              .withSource(githubSourceStorageBuilder.build(githubUrl))
+              .withName(githubUrl.getRepository())
+              .withType("blank")
```
sleshchenko: Type field is not needed anymore according to Sun's description https://github.com/eclipse/che/issues/12217.
I think it would be better to leave it empty and make GWT IDE asking project type instead of hardcoding it. WDYT?

plugins/plugin-github/che-plugin-github-factory-resolver/src/main/java/org/eclipse/che/plugin/github/factory/resolver/GithubURLParserImpl.java
```diff
@@ -90,7 +92,7 @@ public GithubUrl parse(String url) {
         .withRepository(repoName)
         .withBranch(branchName)
         .withSubfolder(matcher.group("subFolder"))
-        .withDockerfileFilename(".factory.dockerfile")
+        .withDevfileFilename(".devfile")
```
sleshchenko: I think it should be `Devfile` or `.Devfile`. 
Need to decide which name we use and maybe rename [Devfile](https://github.com/redhat-developer/devfile/blob/master/petclinic.yaml) in [redhat-developer/devfile](https://github.com/redhat-developer/devfile) repository

benoitf: very good point @sleshchenko 
Maybe we can start with `Devfile` for iteration 1

mshaposhnik: There is no single opinion on this :). I was looking and testing on official @lord repo at 
https://github.com/che-incubator/chectl/  and there is `.devfile`

benoitf: :) @l0rd ? seems you're using .devfile so needs to clarify the final (or at least the starting name) :)

l0rd: `.devfile` is much better choice :-) 

Jokes apart can the match be case insensitive and support both `.devfile` and `devfile` (maybe in a separate PR)?

benoitf: i'm not sure case insensitive is easy. On github http request it may consist of several different requests

plugins/plugin-github/che-plugin-github-factory-resolver/src/main/java/org/eclipse/che/plugin/github/factory/resolver/GithubFactoryParametersResolver.java
mshaposhnik: Removed `blank` here and made as optional factory parameter

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/urlfactory/URLFactoryBuilder.java
mshaposhnik: @skabashnyuk  it is not used when devfile is present. it is used when there in _no_ devfile :-)
so i prefer to not use word "devfile" here

skabashnyuk: che.factory.default_editor che.factory.default_plugins ?

mshaposhnik: yes

mshaposhnik:
ci-test

sleshchenko (commented):
PR description says

> Removed support of dockerfile feature;

Did you really remove supporting `dockerfiles` whole system? If no - please consider rephrasing this item in the description.

As far as I see it also should add support of making factories from devfile on Github PRs. If yes - not be shy - declare that you implemented it.

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/ValueHelper.java
```diff
@@ -27,7 +27,7 @@
    */
   public static boolean isEmpty(Object value) {
     return (null == value)
-        || (value.getClass().equals(String.class) && isNullOrEmpty((String) value)
+        || ((value.getClass().equals(String.class) && isNullOrEmpty((String) value))
```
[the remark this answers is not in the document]
mshaposhnik: No - just compiler warning

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/jpa/JpaFactoryDao.java
```diff
@@ -108,41 +111,85 @@ public FactoryImpl getById(String id) throws NotFoundException, ServerException
...
       throw new ServerException(ex.getLocalizedMessage(), ex);
     }
   }
 
+  @Override
+  @Transactional
+  public List<FactoryImpl> getByUser(String userId) {
+    requireNonNull(userId);
+    final Pair<String, String> factoryCreator = Pair.of("creator.userId", userId);
+    return getFactoriesByAttributes(Integer.MAX_VALUE, 0, singletonList(factoryCreator));
```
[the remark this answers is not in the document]
mshaposhnik: Yes, it means - unlimited (see https://github.com/eclipse-ee4j/eclipselink/blob/master/jpa/org.eclipse.persistence.jpa/src/org/eclipse/persistence/internal/jpa/QueryImpl.java#L861-L862)
And  will be cnahed to separate field in #12227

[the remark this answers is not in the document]
sleshchenko: I understand that it is unlimited but it is good to load all factories in memory at the same time? See here what is alternative way https://github.com/eclipse/che/blob/07263f1e30089689d71b057f747a44a29283e3c4/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/jpa/JpaWorkspaceDao.java#L282-L288

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/ValueHelper.java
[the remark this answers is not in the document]
sleshchenko: I see, thanks.

riuvshin:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1345/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1345//Selenium_20tests_20report/)
docker image: eclipseche/che-server:12232

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryService.java
```diff
@@ -220,7 +223,7 @@ public FactoryDto getFactory(
     }
 
     final List<FactoryDto> factories = new ArrayList<>();
-    for (Factory factory : factoryManager.getByAttribute(maxItems, skipCount, query)) {
+    for (Factory factory : factoryManager.getByAttribute(maxItems, skipCount, query).getItems()) {
       factories.add(injectLinks(asDto(factory)));
     }
     return factories;
```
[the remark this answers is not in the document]
mshaposhnik: Example is not correct but i got the idea :-)

mshaposhnik:
ci-test

plugins/plugin-github/che-plugin-github-factory-resolver/src/main/java/org/eclipse/che/plugin/github/factory/resolver/GithubFactoryParametersResolver.java
```diff
@@ -66,30 +81,35 @@ public boolean accept(@NotNull final Map<String, String> factoryParameters) {
...
-    final GithubUrl githubUrl = githubUrlParser.parse(factoryParameters.get("url"));
+    final GithubUrl githubUrl = githubUrlParser.parse(factoryParameters.get(URL_PARAMETER_NAME));
 
     // create factory from the following location if location exists, else create default factory
-    FactoryDto factory = urlFactoryBuilder.createFactory(githubUrl.factoryJsonFileLocation());
-
+    FactoryDto factory =
+        urlFactoryBuilder
+            .createFactoryFromDevfile(githubUrl.devfileFileLocation())
+            .orElseGet(
```
sleshchenko, line 93: IMHO Hard to say that using double `orElseGet` makes code clearer.
Up to you revert it to if/else statements or use lambda expressions.

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileManager.java
```diff
@@ -0,0 +1,126 @@
...
+   * Creates workspace from given {@link WorkspaceConfigImpl} with available name search
+   *
+   * @param workspaceConfig initial workspace configuration
+   * @return created workspace instance
+   * @throws ValidationException when incoming configuration or attributes are not valid
+   * @throws ConflictException when any conflict occurs
+   * @throws NotFoundException when user account is not found
+   * @throws ServerException when other error occurs
+   */
+  public WorkspaceImpl createWorkspace(WorkspaceConfigImpl workspaceConfig)
```
sleshchenko: For me it looks strange to see such public method in `DevfileManager`. As far as I see it is used in Service in the following way:
```
  ....
 WorkspaceConfig config = manager.convert(devfileContent, verbose);
 manager.createWorkspace(config);
```
I see it more logical to have public method `WorkspaceImpl createWorkspace(String devfileContent)` in `DevfileManager`

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/jpa/JpaFactoryDao.java
```diff
@@ -107,42 +111,89 @@ public FactoryImpl getById(String id) throws NotFoundException, ServerException
...
       throw new ServerException(ex.getLocalizedMessage(), ex);
     }
   }
 
+  @Override
+  @Transactional(rollbackOn = {RuntimeException.class, ServerException.class})
+  public Page<FactoryImpl> getByUser(String userId, int maxItems, long skipCount) {
+    requireNonNull(userId);
+    final Pair<String, String> factoryCreator = Pair.of("creator.userId", userId);
+    long totalCount = countFactoriesByAttributes(singletonList(factoryCreator));
```
sleshchenko: As far as I see runtime exception may be thrown by an entity manager inside of `countFactoriesByAttributes` method. It must be mapped to `ServerException` but it IS NOT. Am I right?

wsmaster/che-core-api-factory/src/test/java/org/eclipse/che/api/factory/server/urlfactory/URLFactoryBuilderTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+  @BeforeClass
+  public void setUp() {
+    this.urlFactoryBuilder =
+        new URLFactoryBuilder(defaultEditor, defaultPlugin, urlFetcher, devfileManager);
+  }
+
+  @Test
+  public void checkDefaultConfiguration() throws Exception {
+
```
sleshchenko: Is it some new-fashion way to have an empty line before method body?

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/jpa/JpaFactoryDao.java
```diff
@@ -92,7 +96,7 @@ public void remove(String id) throws ServerException {
   }
 
   @Override
-  @Transactional
+  @Transactional(rollbackOn = {RuntimeException.class, ServerException.class})
```
sleshchenko: I think `@Transactional(rollbackOn = ServerException.class)` would be enough. But it is better to have more than needed but not less than needed in this case )

sleshchenko (commented):
Please be more careful with changes in DAO implementation

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/jpa/JpaFactoryDao.java
mshaposhnik: In this case, will transaction be rolled back on runtime exc ?

mshaposhnik: Ah you mean that runtime is catched here.ok

plugins/plugin-github/che-plugin-github-factory-resolver/src/main/java/org/eclipse/che/plugin/github/factory/resolver/GithubFactoryParametersResolver.java
mshaposhnik: if else with optionals and variable re-assigning looks messy too.this way looks not very complex for me

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileManager.java
mshaposhnik: Too much responsibility for single method and not flexible in usage IMO.
I would rather propose another schema:
1) Devfile content to devfile object - with schema validation but without intergrity validation;
2) Create WS from devfile - integrity validation, then convert to WsConfig and create WS. With this way, it will be easy to add later additional optional params for stuff like K8S list etc to be merged together into config. WDYT ?

mshaposhnik: yes, that's it.

sleshchenko (approved):

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileManager.java
```diff
@@ -0,0 +1,141 @@
...
+import org.eclipse.che.api.core.ValidationException;
+import org.eclipse.che.api.devfile.model.Devfile;
+import org.eclipse.che.api.devfile.server.validator.DevfileIntegrityValidator;
+import org.eclipse.che.api.devfile.server.validator.DevfileSchemaValidator;
+import org.eclipse.che.api.workspace.server.WorkspaceManager;
+import org.eclipse.che.api.workspace.server.model.impl.WorkspaceConfigImpl;
+import org.eclipse.che.api.workspace.server.model.impl.WorkspaceImpl;
+import org.eclipse.che.commons.env.EnvironmentContext;
+
+@Singleton
```
skabashnyuk, line 41: ```@Beta```
Javadoc?

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryService.java
```diff
@@ -219,11 +216,13 @@ public FactoryDto getFactory(
...
-    for (Factory factory : factoryManager.getByAttribute(maxItems, skipCount, query)) {
-      factories.add(injectLinks(asDto(factory)));
+    Page<? extends Factory> factoriesPage =
+        factoryManager.getByAttribute(maxItems, skipCount, query);
+    List<FactoryDto> list = new ArrayList<>();
+    for (Factory factory : factoriesPage.getItems()) {
+      list.add(injectLinks(asDto(factory)));
     }
-    return factories;
+    return Response.ok().entity(list).header("Link", createLinkHeader(factoriesPage)).build();
```
skabashnyuk, line 225:
```
    return Response.ok()
        .entity(
            workspacesPage
                .getItems()
                .stream()
                .filter(ws -> status == null || status.equalsIgnoreCase(ws.getStatus().toString()))
                .map(DtoConverter::asDto)
                .collect(toList()))
        .header("Link", createLinkHeader(workspacesPage))
        .build();
```

skabashnyuk (approved):

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/FactoryService.java
mshaposhnik: Not possible due to checked exceptions thrown in FactoryService::asDto

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileManager.java
mshaposhnik: Ok

Issue #11961, opened 2018-11-15
Che Server rejects JSON RPC messages which may lead to losing events


sleshchenko:
### Description
Che Server rejects JSON RPC messages which may lead to losing events.
There is a `ci-test` job that failed because of slow tests execution.
https://github.com/eclipse/che/pull/11920#issuecomment-439021314

Che Server Logs contain a lot of logged `wsagent` log events like the following:
```
2018-11-15 08:47:48,833[stProcessor-285]  [ERROR] [l.ErrorInstallerLogEventLogger 43]   - Installer `org.eclipse.che.ws-agent` error from machine=`dev-machine` owner=`1cf07453-4b2a-4226-8ca0-c22189c5da34` env=`workspacejm96ne` workspace=`workspaceii5wyhr367a3cmz3` text=`  "params": {` time=`2018-11-15T08:47:48.827536783Z`
2018-11-15 08:47:48,833[stProcessor-360]  [ERROR] [l.ErrorInstallerLogEventLogger 43]   - Installer `org.eclipse.che.ws-agent` error from machine=`dev-machine` owner=`1cf07453-4b2a-4226-8ca0-c22189c5da34` env=`workspacejm96ne` workspace=`workspaceii5wyhr367a3cmz3` text=`    "uri": "file:///projects/console-cpp-simple/iseven.cpp"` time=`2018-11-15T08:47:48.827616797Z`
2018-11-15 08:47:48,834[stProcessor-354]  [ERROR] [l.ErrorInstallerLogEventLogger 43]   - Installer `org.eclipse.che.ws-agent` error from machine=`dev-machine` owner=`1cf07453-4b2a-4226-8ca0-c22189c5da34` env=`workspacejm96ne` workspace=`workspaceii5wyhr367a3cmz3` text=`    "uri": "file:///projects/console-cpp-simple/iseven.cpp"` time=`2018-11-15T08:47:48.804885696Z`
```
After that there are errors that JSON RPC messages are dropped:
```
2018-11-15 08:49:19,326[nio-8080-exec-1]  [WARN ] [j.i.ServerSideRequestProcessor 62]   - Message org.eclipse.che.commons.lang.concurrent.CopyThreadLocalRunnable@5e59b14 rejected for execution
[...]

PR #12252, opened 2018-12-21
[WIP]Introduced two web-socket endpoints for workspace master to split JSON-RPC messages

dkulieshov:
Related issue: https://github.com/eclipse/che/issues/11961

Introduced two web-socket endpoints for workspace master to split JSON-RPC messages into minor and major. Minor is expected to be used for not important messages like logger messages etc., while major endpoint is for important ones like statuses.

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/CheMajorWebSocketEndpoint.java
+++ b/assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/CheMajorWebSocketEndpoint.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright (c) 2012-2018 Red Hat, Inc.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.api.deploy;
+
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.slf4j.LoggerFactory.getLogger;
+
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import java.util.concurrent.SynchronousQueue;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.ThreadPoolExecutor;
+import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.websocket.server.ServerEndpoint;
+import org.eclipse.che.api.core.jsonrpc.commons.RequestProcessorConfigurator;
+import org.eclipse.che.api.core.jsonrpc.impl.ServerSideRequestProcessor;
+import org.eclipse.che.api.core.websocket.commons.WebSocketMessageReceiver;
+import org.eclipse.che.api.core.websocket.impl.BasicWebSocketEndpoint;
+import org.eclipse.che.api.core.websocket.impl.GuiceInjectorEndpointConfigurator;
+import org.eclipse.che.api.core.websocket.impl.MessagesReSender;
+import org.eclipse.che.api.core.websocket.impl.WebSocketSessionRegistry;
+import org.eclipse.che.api.core.websocket.impl.WebsocketIdService;
+import org.eclipse.che.commons.lang.concurrent.LoggingUncaughtExceptionHandler;
+import org.slf4j.Logger;
+
+/**
+ * Implementation of {@link BasicWebSocketEndpoint} for Che packaging. Add only mapping
+ * "/websocket".
+ */
+@ServerEndpoint(value = "/websocket", configurator = GuiceInjectorEndpointConfigurator.class)
+public class CheMajorWebSocketEndpoint extends BasicWebSocketEndpoint {
+  private static final Logger LOG = getLogger(CheMajorWebSocketEndpoint.class);
+
+  private final int maxPoolSize;
+  private final RequestProcessorConfigurator requestProcessorConfigurator;
+
+  private ThreadPoolExecutor executor;
+
+  @Inject
+  public CheMajorWebSocketEndpoint(
+      WebSocketSessionRegistry registry,
+      MessagesReSender reSender,
+      WebSocketMessageReceiver receiver,
+      WebsocketIdService websocketIdService,
+      RequestProcessorConfigurator requestProcessorConfigurator,
+      @Named("che.core.jsonrpc.processor_max_pool_size") int maxPoolSize) {
+    super(registry, reSender, receiver, websocketIdService);
+    this.maxPoolSize = maxPoolSize;
+    this.requestProcessorConfigurator = requestProcessorConfigurator;
+  }
+
+  @Override
+  protected String getEndpointId() {
+    return "master-websocket-major-endpoint";
+  }
+
+  @PostConstruct
+  private void postConstruct() {
+    ThreadFactory factory =
+        new ThreadFactoryBuilder()
+            .setUncaughtExceptionHandler(LoggingUncaughtExceptionHandler.getInstance())
+            .setNameFormat(ServerSideRequestProcessor.class.getSimpleName() + "-%d")
+            .setDaemon(true)
+            .build();
+
+    executor =
+        new ThreadPoolExecutor(0, maxPoolSize, 60L, SECONDS, new SynchronousQueue<>(), factory);
+    executor.setRejectedExecutionHandler(
+        (r, __) -> LOG.warn("Message {} rejected for execution", r));
+    requestProcessorConfigurator.put(getEndpointId(), () -> executor);
+  }
+
+  @PreDestroy
+  private void preDestroy() {
+    executor.shutdown();
+    try {
+      if (executor.awaitTermination(3, SECONDS)) {
+        executor.shutdownNow();
+        executor.awaitTermination(3, SECONDS);
+      }
+    } catch (InterruptedException ie) {
+      executor.shutdownNow();
+      Thread.currentThread().interrupt();
+    }
+  }
+}
--- a/assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/CheMajorWebSocketEndpoint.java
+++ b/assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/CheMajorWebSocketEndpoint.java
@@ -19,12 +19,10 @@ import java.util.concurrent.SynchronousQueue;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
 import javax.annotation.PostConstruct;
-import javax.annotation.PreDestroy;
 import javax.inject.Inject;
 import javax.inject.Named;
 import javax.websocket.server.ServerEndpoint;
 import org.eclipse.che.api.core.jsonrpc.commons.RequestProcessorConfigurator;
-import org.eclipse.che.api.core.jsonrpc.impl.ServerSideRequestProcessor;
 import org.eclipse.che.api.core.websocket.commons.WebSocketMessageReceiver;
 import org.eclipse.che.api.core.websocket.impl.BasicWebSocketEndpoint;
 import org.eclipse.che.api.core.websocket.impl.GuiceInjectorEndpointConfigurator;
@@ -70,28 +68,14 @@ public class CheMajorWebSocketEndpoint extends BasicWebSocketEndpoint {
     ThreadFactory factory =
         new ThreadFactoryBuilder()
             .setUncaughtExceptionHandler(LoggingUncaughtExceptionHandler.getInstance())
-            .setNameFormat(ServerSideRequestProcessor.class.getSimpleName() + "-%d")
+            .setNameFormat(CheMajorWebSocketEndpoint.class.getSimpleName() + "-%d")
             .setDaemon(true)
             .build();
 
     executor =
         new ThreadPoolExecutor(0, maxPoolSize, 60L, SECONDS, new SynchronousQueue<>(), factory);
     executor.setRejectedExecutionHandler(
-        (r, __) -> LOG.warn("Message {} rejected for execution", r));
+        (r, __) -> LOG.error("Message {} rejected for execution", r));
     requestProcessorConfigurator.put(getEndpointId(), () -> executor);
-  }
-
-  @PreDestroy
-  private void preDestroy() {
-    executor.shutdown();
-    try {
-      if (executor.awaitTermination(3, SECONDS)) {
-        executor.shutdownNow();
-        executor.awaitTermination(3, SECONDS);
-      }
-    } catch (InterruptedException ie) {
-      executor.shutdownNow();
-      Thread.currentThread().interrupt();
-    }
   }
 }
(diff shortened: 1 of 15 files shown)


dkuleshov:
ci-test

assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/CheMajorWebSocketEndpoint.java
```diff
@@ -0,0 +1,97 @@
...
+        new ThreadFactoryBuilder()
+            .setUncaughtExceptionHandler(LoggingUncaughtExceptionHandler.getInstance())
+            .setNameFormat(ServerSideRequestProcessor.class.getSimpleName() + "-%d")
+            .setDaemon(true)
+            .build();
+
+    executor =
+        new ThreadPoolExecutor(0, maxPoolSize, 60L, SECONDS, new SynchronousQueue<>(), factory);
+    executor.setRejectedExecutionHandler(
+        (r, __) -> LOG.warn("Message {} rejected for execution", r));
```
sleshchenko: Since it is major websocket endpoint it should be logged as error

assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/CheMajorWebSocketEndpoint.java
```diff
@@ -0,0 +1,97 @@
...
+  protected String getEndpointId() {
+    return "master-websocket-major-endpoint";
+  }
+
+  @PostConstruct
+  private void postConstruct() {
+    ThreadFactory factory =
+        new ThreadFactoryBuilder()
+            .setUncaughtExceptionHandler(LoggingUncaughtExceptionHandler.getInstance())
+            .setNameFormat(ServerSideRequestProcessor.class.getSimpleName() + "-%d")
```
sleshchenko: Please consider changing name format to make it clear that it is `major websocket` related thread.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
```diff
@@ -277,6 +281,9 @@ che.infra.docker.master_api_endpoint=http://che-host:${CHE_PORT}/api
 # This is the webscoket base endpoint of the workspace master running within the core Che server.
 che.infra.docker.master_websocket_endpoint=ws://che-host:${CHE_PORT}/api/websocket
 
+# This is the minor webscoket base endpoint of the workspace master running within the core Che server.
+che.infra.docker.master_websocket_minor_endpoint=ws://che-host:${CHE_PORT}/api/websocket-minor
```
sleshchenko, line 285: It's not the best name to this property but I'm OK with that since docker infrastructure will be decommissioned

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
```diff
@@ -21,6 +21,10 @@ che.api=http://${CHE_HOST}:${CHE_PORT}/api
 # for websocket interaction/messaging.
 che.websocket.endpoint=ws://${CHE_HOST}:${CHE_PORT}/api/websocket
 
+# Che websocket minor endpoint. Provides basic communication endpoint
```
sleshchenko, line 24: What if we just introduce new endpoint for logs and will not use `major`, `minor` in endpoints names? For me, `minor`/`major` looks not self-explained at all.

dkulieshov: It's actually not about the logs, the naming comes from the degree of significance of messages that are transmitted over an endpoint. "Major" endpoint is expected to receive important messages, while "minor" is for the rest (e.i. logs). However I would like to know if it is possible to make it more clear, what is your suggestion ?

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/jsonrpc/impl/ServerSideRequestProcessorConfigurator.java
```diff
@@ -0,0 +1,96 @@
...
+        new ThreadPoolExecutor(
+            0, maxPoolSize, 60L, TimeUnit.SECONDS, new SynchronousQueue<>(), factory);
+    ((ThreadPoolExecutor) defaultExecutorService)
+        .setRejectedExecutionHandler(
+            (r, executor) -> LOG.warn("Message {} rejected for execution", r));
+  }
+
+  @PreDestroy
+  private void preDestroy() {
+    defaultExecutorService.shutdown();
```
garagatyi: The same question about shutting down executor and daemon threads.

garagatyi (commented):
Would be great to have an overall explanation of changes that PR brings and reasoning for them. This would significantly simplify reviewers their work.
I inlined some comments.

riuvshin:
[Build details](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1342/)
[Test report](https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1342//Selenium_20tests_20report/)
docker image: eclipseche/che-server:12252

assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/CheMajorWebSocketEndpoint.java
```diff
@@ -0,0 +1,97 @@
...
+    executor =
+        new ThreadPoolExecutor(0, maxPoolSize, 60L, SECONDS, new SynchronousQueue<>(), factory);
+    executor.setRejectedExecutionHandler(
+        (r, __) -> LOG.warn("Message {} rejected for execution", r));
+    requestProcessorConfigurator.put(getEndpointId(), () -> executor);
+  }
+
+  @PreDestroy
+  private void preDestroy() {
+    executor.shutdown();
```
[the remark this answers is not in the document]
dkulieshov: Not really, here and in other similar places the source code was just moved from `org.eclipse.che.api.core.jsonrpc.impl.ServerSideRequestProcessor#ServerSideRequestProcessor`. I'm going to review this in the nearest commit.

core/che-core-api-core/src/main/java/org/eclipse/che/api/core/jsonrpc/commons/JsonRpcMessageReceiver.java
```diff
@@ -55,24 +55,26 @@ public JsonRpcMessageReceiver(
...
       String error = "An error occurred on the server while parsing the JSON text";
-      errorTransmitter.transmit(endpointId, new JsonRpcException(-32700, error));
+      errorTransmitter.transmit(combinedEndpointId, new JsonRpcException(-32700, error));
     }
 
     List<String> messages = jsonRpcUnmarshaller.unmarshalArray(message);
     for (String innerMessage : messages) {
       if (jsonRpcQualifier.isJsonRpcRequest(innerMessage)) {
-        requestProcessor.process(new ProcessRequestTask(endpointId, innerMessage));
+        String endpointId = combinedEndpointId.split("<-:->")[1];
```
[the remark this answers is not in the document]
dkulieshov: The source is here `org.eclipse.che.api.core.websocket.impl.WebsocketIdService#SEPARATOR`. I'm still considering making the constant __`public`__, I see that it is necessary here.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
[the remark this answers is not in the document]
dkulieshov: I'm not sure I follow you, do you suggest to keep `che.infra.docker.master_websocket_endpoint` property in `che.properties` while move `che.infra.docker.master_websocket_minor_endpoint` to guice module?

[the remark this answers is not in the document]
garagatyi: Yes. But it is up to you. I don't have strong feelings about that.

dkuleshov:
ci-test

dkuleshov:
ci-test

garagatyi (dismissed):
Looks good

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
sleshchenko: > while "minor" is for the rest (e.i. logs)

Do we have something else for minor endpoint?

I suggest consider having `Che WebSocket Endpoint`(major) and `Che Logs WebSocket Endpoint`(minor). If we would have something else for minor endpoint - then it makes sense to introduce new endpoint and be able to configure `messages_pull_size` separately.

sleshchenko (dismissed):

dkuleshov:
@sleshchenko 

Actually I'm not sure if there is something else except logs for the minor endpoint right now, I'm okay with your suggestion however I would like to see @vparfonov and @garagatyi opinions.

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
garagatyi: I don't have strong feelings about that. So, up to the author of the PR if there are no objections from other sites

vparfonov:
I think "minor" it's more useful in this case, possibly we will send not only logs via it. For now don't have concrete examples but "minor" just more wider.

vparfonov (dismissed):

dmytro-ndp:
@dkuleshov, @musienko-maxim: [fresh ci-test execution](https://github.com/che-bot) because of test workspaces failed to start. There is [che-server-pod.log available](https://gist.github.com/dmytro-ndp/5d544fcd5790b59334cdcb30ac89f9f1) for review.

musienko-maxim:
ci-test

garagatyi:
> I think major and minor endpoint executors should be configured in two different ways separately
> 
> * Major (for statuses) - with an Int.MAX_VALUE queue and  individually configurable thread pool
> * Minor (for logs) - with SynchronousQueue queue and individually configurable thread pool
> 
> Please make sure we have ability to configure this values before the merge.

+1 to configure them separately
-1 to use MAX_INT - this leads to OOMs. Please use some meaningful value instead and make it configurable.

sleshchenko:
Test report shows that workspaces fails to start because start reached timeout:
```
2019-01-09 11:14:03,682[io-8080-exec-10]  [INFO ] [o.e.c.a.w.s.WorkspaceRuntimes 356]   - Starting workspace 'che/javaWS' with id 'workspacerzgm22nlvopj136o' by user 'che'
2019-01-09 11:22:03,701[aceSharedPool-1]  [WARN ] [.i.k.KubernetesInternalRuntime 245]  - Failed to start Kubernetes runtime of workspace workspacerzgm22nlvopj136o. Cause: Waiting for Kubernetes environment 'default' of the workspace'workspacerzgm22nlvopj136o' reached timeout
2019-01-09 11:22:03,991[aceSharedPool-1]  [INFO ] [o.e.c.a.w.s.WorkspaceRuntimes 417]   - Workspace 'che:javaWS' with id 'workspacerzgm22nlvopj136o' start failed
```
I assume that bootstrapper status event is not received by Che Server. To detect an issue debugging is needed. Settings debug log level for k8s infrastructure may help as well. It is needed to set `CHE_LOGGER_CONFIG=org.eclipse.che.workspace.infrastructure.kubernetes=DEBUG,org.eclipse.che.workspace.infrastructure.openshift=DEBUG,org.eclipse.che.api.workspace.server=DEBUG` to do it.

skabashnyuk:
>-1 to use MAX_INT - this leads to OOMs. Please use some meaningful value instead and make it configurable.

+1

sleshchenko:
I've figured out that WebSocket Methods are not linked with any of endpoints. So, it's up to clients to use `minor` or `major` endpoint to publish any of message Che Server receives. So, logs may be published to major endpoint as well.
@dkuleshov Am I correct? If yes - if there any plans/issue to bind WebSocket Method to a particular endpoint? Maybe it makes sense to describe this aspect somewhere at least in PR description.

dkuleshov:
> I've figured out that WebSocket Methods are not linked with any of endpoints. So, it's up to clients to use minor or major endpoint to publish any of message Che Server receives. So, logs may be published to major endpoint as well.

That is correct.

> if there any plans/issue to bind WebSocket Method to a particular endpoint? Maybe it makes sense to describe this aspect somewhere at least in PR description.

It was discussed several times in the past and there are no plans for that currently.

I also don't see how it can be useful to mention that in the description of this very pull request. It is not related to this specific issue but one of general principles of json-rpc/websocket transport implementation, so it would matter if we mention that in the description of the framework, which is out of the scope here, but I guess it is not a problem to create a separate issue for that if there is a concern.

garagatyi (commented):
Do you plan to accommodate changes proposed by @skabashnyuk in this PR? If thread pools are configured with the same value we can't configure dropping of log messages and keeping status messages in a flexible way

dkuleshov:
@garagatyi no worries, those changes were already introduced in an initial commit, you can configure any ExecutorService you need for each web-socket endpoint separately.

dkuleshov:
I'm saying that it is possible to configure separate executor service instance for each endpoint. You're concerned that for the endpoints executor services are configured with the same value and we can't treat them separately and that leads to inflexibility, while in fact the values are two similar but different instances that can be set separately if needed so it is flexible.

garagatyi:
@dkuleshov at the moment we have issues with OOMs in a case when the thread number is too big for the current container RAM limit. 
Since this PR creates 2 instances of the executor with the same thread pool settings we, potentially, can increase consumption of RAM and face OOM.
Since configuration is bound in the code of the upstream project I do not know how to change it in a downstream project.
Can you explain how to reconfigure this in a downstream project without asking for a new upstream release?

garagatyi:
We are already using `che.core.jsonrpc.processor_max_pool_size` in a way that it doesn't lead to a RAM consumption that exceeds container limit.
With this PR we will have 2 pool of that size which can, potentially, double the RAM consumption by the JSONRPC processing pools.

garagatyi:
So, with a bigger consumption by 2 pools RAM footprint differs and Che might be killed by OOM

dkuleshov:
Understood, so I suggest to have two separate properties for each endpoint

garagatyi:
@dkuleshov but shouldn't it be done upstream?

core/che-core-api-core/src/test/java/org/eclipse/che/api/core/jsonrpc/commons/JsonRpcMessageReceiverTest.java
```diff
@@ -30,7 +30,7 @@
 public class JsonRpcMessageReceiverTest {
 
   static final String MESSAGE = "message";
-  static final String ENDPOINT_ID = "endpoint-id";
+  static final String ENDPOINT_ID = "client-id<-:->endpoint-id";
```
ScrewTSW: Shouldn't this string use `org.eclipse.che.api.core.websocket.impl.WebsocketIdService.SEPARATOR` instead of a fixed value?

skabashnyuk:
ci-test

garagatyi:
@skabashnyuk can you elaborate on what is the plan regarding this https://github.com/eclipse/che/pull/12252#issuecomment-454015052?

skabashnyuk:
>@skabashnyuk can you elaborate on what is the plan regarding this #12252 (comment)?

It is on the  "to-do" list.

skabashnyuk:
closing. work would be finalized here https://github.com/eclipse/che/pull/12673

Issue #12207, opened 2018-12-17
Track workspace pod existence


skabashnyuk:
### Description
Sometimes it may happen that workspace Pod goes down when workspace master thinks that workspace should be alive. 
For example, the user stopped it from OpenShift/K8s console, admin can terminate (miner detection) pod or infrastructure node goes down.
Periodically (or once on startup) we should check the existence of the workspace pod and terminate workspace

PR #12257, opened 2018-12-21
Add periodical job that stops Runtimes without existing pods

sleshchenko:
#### The main change
Sometimes it may happen that workspace Pods goes down when Che Server considers that workspace is `RUNNING`.
For example, the user stopped it from OpenShift/K8s console, admin can terminate (miner detection) pod or infrastructure node goes down.
The main purpose of this PR is adding a periodical job that will stop such Runtimes without existing pods.
It is disabled by default because there is possible Che Server configuration when Che Server doesn't have an ability to interact with Kubernetes API when operation is not invoked by user. And it may be enabled by configuring `che.infra.kubernetes.runtimes_consistency_check_period_min` property with value more than `0` (the corresponding environment variable `CHE_INFRA_KUBERNETES_RUNTIMES__CONSISTENCY__CHECK__PERIOD__MIN`).

It DOES work on the following configurations:
- workspaces objects are created in the same namespace where Che Server is located;
- cluster-admin service account token is mount to Che Server pod;

It DOES NOT work on the following configurations(InconsistentRuntimesDetector will fail to check runtime consistency):
- Che Server communicates with Kubernetes API using token from OAuth provider;

#### Related changes
It also contains the following changes which are needed to make it possible to implement such job:
- Add an ability to get running internal runtimes via WorkspaceRuntimes;
- Change 'running' word to 'active' not to avoid confusing with RUNNING status;
- Fix get pod by name method to return empty optional if pod doesn't exist;
- Add an ability to get running internal runtimes via WorkspaceRuntimes;

#### Major fixes
- I've discovered that `RuntimeHangingDetector` does not publish an abnormal stopped event when an exception occurred on stopping hanging k8s runtimes. It may lead to hanging workspace in `stopping` state. So, fixed it by publishing the abnormal stopped event in `finally` statement.

#### Minor fixes
It also contains other minor fixes:
- Fix creation k8s clients without workspace id specified;
- Fix tests for KubernetesDeployments;
Reviewers sorry for changes non-related to implemented to feature but I place them here because I discovered them during implementing PR goal feature and didn't create a separate PR not to spend additional time for `ci-build`/`ci-test` jobs.
- Fix typo: 'ingressMeta' -> 'routeMeta'

https://github.com/eclipse/che/issues/12207

### How have you tested this PR?
This PR is tested on the following configurations:
- SingleUser Che Deployed on Minishift
- SingleUser Che Deployed on Minikube

#### Release Notes
Added a periodical job that will stop Runtimes which have no existing pods anymore. 
An example of a situation when such workspaces may appear may be a user stopped it from OpenShift/K8s console, admin terminates pod or infrastructure node goes down.
It is disabled by default. It may be enabled by configuring `CHE_INFRA_KUBERNETES_RUNTIMES__CONSISTENCY__CHECK__PERIOD__MIN` env var for Che Server with value more than `0`. 
Note that to make it working correctly it is needed to configure Che Server to have access token to Kubernetes/OpenShift cluster without user interaction:
- create workspace objects in the same namespace where Che Server is created.
- mount service account with cluster admin permissions to Che Server Pod.
So, and it won't work if OAuth provider is configured.

#### Docs PR
https://github.com/eclipse/che-docs/pull/652

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
@@ -498,6 +498,21 @@ che.infra.kubernetes.tls_enabled=false
 # Ignored by OpenShift infrastructure
 che.infra.kubernetes.tls_secret=
 
+# Defines the period with which runtimes consistency checks will be performed.
+# If runtime has inconsistent state then runtime will be stopped automatically.
+# Value must be more than 0 or `-1`, where `-1` means that checks won't be performed at all.
+#
+# It is disabled by default because there is possible Che Server configuration when Che Server
+# doesn't have an ability to interact with Kubernetes API when operation is not invoked by user.
+#
+# It DOES work on the following configurations:
+# - workspaces objects are created in the same namespace where Che Server is located;
+# - cluster-admin service account token is mount to Che Server pod;
+#
+# It DOES NOT work on the following configurations:
+# - Che Server communicates with Kubernetes API using token from OAuth provider;
+che.infra.kubernetes.runtimes_consistency_check_period_min=-1
+
 ### OpenShift Infra parameters
 #
 # Since OpenShift infrastructure reuse Kubernetes infrastructure components
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/InconsistentRuntimesDetector.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/InconsistentRuntimesDetector.java
@@ -1,0 +1,154 @@
+/*
+ * Copyright (c) 2012-2018 Red Hat, Inc.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.workspace.infrastructure.kubernetes;
+
+import static java.lang.String.format;
+import static java.util.Collections.emptyMap;
+
+import com.google.common.annotations.VisibleForTesting;
+import java.util.concurrent.TimeUnit;
+import javax.inject.Inject;
+import javax.inject.Singleton;
+import org.eclipse.che.api.core.ServerException;
+import org.eclipse.che.api.core.model.workspace.WorkspaceStatus;
+import org.eclipse.che.api.core.model.workspace.runtime.RuntimeIdentity;
+import org.eclipse.che.api.workspace.server.WorkspaceRuntimes;
+import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
+import org.eclipse.che.api.workspace.server.spi.InternalRuntime;
+import org.eclipse.che.commons.schedule.ScheduleDelay;
+import org.eclipse.che.workspace.infrastructure.kubernetes.util.RuntimeEventsPublisher;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Periodically checks runtimes consistency and forcibly stop ones which has inconsistent state.
+ *
+ * @author Sergii Leshchenko
+ */
+@Singleton
+public class InconsistentRuntimesDetector {
+
+  private static final Logger LOG = LoggerFactory.getLogger(InconsistentRuntimesDetector.class);
+
+  private final RuntimeEventsPublisher eventPublisher;
+  private final WorkspaceRuntimes workspaceRuntimes;
+
+  @Inject
+  public InconsistentRuntimesDetector(
+      RuntimeEventsPublisher eventPublisher, WorkspaceRuntimes workspaceRuntimes) {
+    this.eventPublisher = eventPublisher;
+    this.workspaceRuntimes = workspaceRuntimes;
+  }
+
+  @ScheduleDelay(
+      delayParameterName = "che.infra.kubernetes.runtimes_consistency_check_period_min",
+      initialDelayParameterName = "che.infra.kubernetes.runtimes_consistency_check_period_min",
+      unit = TimeUnit.MINUTES)
+  public void check() {
+    LOG.debug("Runtimes consistency check is running");
+    for (String runningWorkspaceId : workspaceRuntimes.getRunning()) {
+      try {
+        checkOne(runningWorkspaceId);
+      } catch (InfrastructureException e) {
+        LOG.error(
+            "Checking consistency of runtime for workspace `{}` is failed. Cause: ",
+            e.getMessage(),
+            e);
+      }
+    }
+  }
+
+  @VisibleForTesting
+  void checkOne(String workspaceId) throws InfrastructureException {
+    LOG.debug("Checking consistency of runtime for workspace `{}`", workspaceId);
+    KubernetesInternalRuntime k8sRuntime = getKubernetesInternalRuntime(workspaceId);
+    RuntimeIdentity runtimeId = k8sRuntime.getContext().getIdentity();
+
+    try {
+      if (k8sRuntime.isConsistent()) {
+        return;
+      }
+    } catch (InfrastructureException e) {
+      throw new InfrastructureException(
+          format(
+              "Error occurred during runtime '%s:%s' consistency checking. Cause: %s",
+              runtimeId.getWorkspaceId(), runtimeId.getOwnerId(), e.getMessage()),
+          e);
+    }
+
+    // check if status is still RUNNING
+    // not to initialize abnormal stop for a runtime that is not RUNNING anymore
+    if (!isRunning(k8sRuntime)) {
+      return;
+    }
+
+    LOG.warn(
+        "Found runtime `{}:{}` with inconsistent state. It's going to be stopped automatically",
+        runtimeId.getWorkspaceId(),
+        runtimeId.getOwnerId());
+
+    stopAbnormally(k8sRuntime);
+    LOG.debug("Checking consistency of runtime for workspace `{}` is finished", workspaceId);
+  }
+
+  private boolean isRunning(KubernetesInternalRuntime k8sRuntime) throws InfrastructureException {
+    try {
+      return k8sRuntime.getStatus() == WorkspaceStatus.RUNNING;
+    } catch (InfrastructureException e) {
+      throw new InfrastructureException(
+          "Error occurred during runtime status fetching during consistency checking. Cause: "
+              + e.getMessage(),
+          e);
+    }
+  }
+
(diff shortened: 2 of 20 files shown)


infrastructures/openshift/pom.xml
```diff
@@ -55,6 +55,10 @@
             <groupId>javax.inject</groupId>
             <artifactId>javax.inject</artifactId>
         </dependency>
+        <dependency>
+            <groupId>javax.ws.rs</groupId>
+            <artifactId>javax.ws.rs-api</artifactId>
```
mkuznyetsov: is this dependency needs to be declared?

sleshchenko: Maven dependency plugin asked me to add it on some development phase. Not it is unused and I've removed it.

sleshchenko:
ci-test

skabashnyuk (approved):
In general looks good to me.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/InconsistentRuntimesDetector.java
```diff
@@ -0,0 +1,112 @@
...
+
+  @ScheduleDelay(
+      delayParameterName = "che.infra.kubernetes.runtimes_concistency_check_period_min",
+      initialDelayParameterName = "che.infra.kubernetes.runtimes_concistency_check_period_min",
+      unit = TimeUnit.MINUTES)
+  public synchronized void check() {
+    LOG.debug("Runtimes consistency checks is running");
+    for (String runningWorkspaceId : workspaceRuntimes.getRunning()) {
+      checkOne(runningWorkspaceId);
+    }
```
mshaposhnik, line 67: maybe it makes sense to debug the end of the checking process too ?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/InconsistentRuntimesDetector.java
```diff
@@ -0,0 +1,112 @@
...
+    this.workspaceRuntimes = workspaceRuntimes;
+  }
+
+  @ScheduleDelay(
+      delayParameterName = "che.infra.kubernetes.runtimes_concistency_check_period_min",
+      initialDelayParameterName = "che.infra.kubernetes.runtimes_concistency_check_period_min",
+      unit = TimeUnit.MINUTES)
+  public synchronized void check() {
+    LOG.debug("Runtimes consistency checks is running");
+    for (String runningWorkspaceId : workspaceRuntimes.getRunning()) {
```
mshaposhnik, line 58: for me looks like a classic example where the parallel stream should be used, no? or we saving threads :-) ?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
```diff
@@ -878,13 +862,33 @@ private String getPodName(String name) throws InfrastructureException {
...
+  /**
+   * Returns the name of a specified Pod given either the actual Pod name or the name of the
+   * DeploymentConfig that controls it. <br>
+   * This is necessary because we are trying to transparently wrap Pods in DeploymentConfigs;
+   * attempting to create a Pod named {@code testPod} will result in a DeploymentConfig with name
+   * {@code testPod}, which will in turn create a Pod named e.g {@code testPod-1-xxxxx}.
+   *
+   * @param name Pod or DeploymentConfig name
+   * @return the name of the intended pod.
+   * @see
```
mshaposhnik: complete or remove the see tag

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/InconsistentRuntimesDetector.java
mshaposhnik: I have feeling, that the longer process is, the less actual data here - some of WS are about to stop etc and when their turn to check it- they may already became stopped etc. This will produce some amount of unclear errors. Maybe im wrong.

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/InconsistentRuntimesDetectorTest.java
```diff
@@ -0,0 +1,27 @@
...
+package org.eclipse.che.workspace.infrastructure.kubernetes;
+
+import org.testng.annotations.Test;
+
+/**
+ * Tests {@link InconsistentRuntimesDetectorTest}.
+ *
+ * @author Sergii Leshchenko
+ */
+public class InconsistentRuntimesDetectorTest {
```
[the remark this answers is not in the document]
sleshchenko: It is already there.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/KubernetesInternalRuntime.java
```diff
@@ -902,6 +903,34 @@ protected void handleUnrecoverableEvent(PodEvent podEvent) {
...
+   *
+   * @throws InfrastructureException if any exception occurs during check performing
+   */
+  public boolean isConsistent() throws InfrastructureException {
+    Set<String> podNames =
+        getInternalMachines()
+            .values()
+            .stream()
+            .map(KubernetesMachineImpl::getPodName)
+            .collect(Collectors.toSet());
```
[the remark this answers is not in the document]
sleshchenko: `KubernetesDeployments#getPodName` throws checked `InfrastructureException` and because of that, I can't implement the proposed way.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/InconsistentRuntimesDetector.java
sleshchenko: Added the following `LOG.debug ` lined in `checkOne` method

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
sleshchenko: Good catch. This line was added some time ago and I just moved it. I have no idea what should be referenced here, so removed.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/InconsistentRuntimesDetector.java
sleshchenko: @mshaposhnik Not sure about a parallel stream. I think it would be better to run threads in the controlled limited executor. Sounds meaningful but I would leave it as is and rework later if we will face issues with unclear errors.

> when their turn to check it- they may already became stopped etc.

Makes sense. I'm adding runtime status checking directly before abnormal stopping of runtime.

sleshchenko:
I refactored `InconsistentRuntimesDetector`. Please take a look one more time at the following diff https://github.com/eclipse/che/pull/12257/commits/0f0eda6740a6b3a57c439429d0fb12c630157a25

sleshchenko:
Marked PR as WIP not to merge it by mistake.
https://github.com/eclipse/che/issues/12282 should be fixed before merging my PR or PR should be reworked to schedule job manually without using `@ScheduleDelay(initialDelayParameterName=..., delayParameterName=...)`. Will do short investigate and will decide what should be done before PR merging.

garagatyi (approved):
Looks good

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java
```diff
@@ -879,13 +862,32 @@ private String getPodName(String name) throws InfrastructureException {
...
+          String.format("Found multiple pods in DeploymentConfig '%s'", name));
     }
 
-    return pods.get(0).getMetadata().getName();
+    return Optional.of(pods.get(0));
+  }
+
+  /**
+   * Returns the name of a specified Pod given either the actual Pod name or the name of the
+   * DeploymentConfig that controls it. <br>
```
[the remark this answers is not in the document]
sleshchenko: I just moved this method. I think someone meant Deployment. So, `DeploymentConfig` changed to `Deployment`

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/InconsistentRuntimesDetector.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    LOG.warn(
+        "Found runtime `{}:{}` with inconsistent state. It's going to be stopped automatically",
+        runtimeId.getWorkspaceId(),
+        runtimeId.getOwnerId());
+
+    // check if status is still RUNNING
+    // not to initialize abnormal stop for a runtime that is not RUNNING anymore
+    if (!isRunning(k8sRuntime)) {
```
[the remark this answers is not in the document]
sleshchenko: Good catch. Fixed

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -219,7 +219,7 @@ public void injectRuntime(WorkspaceImpl workspace) throws ServerException {
     }
   }
 
-  private InternalRuntime<?> getInternalRuntime(String workspaceId)
+  public InternalRuntime<?> getInternalRuntime(String workspaceId)
```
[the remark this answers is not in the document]
sleshchenko: Added

assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
```diff
@@ -498,6 +498,21 @@ che.infra.kubernetes.tls_enabled=false
...
+# It is disabled by default because there is possible Che Server configuration when Che Server
+# doesn't have an ability to interact with Kubernetes API when operation is not invoked by user.
+#
+# It DOES work on the following configurations:
+# - workspaces objects are created in the same namespace where Che Server is located;
+# - cluster-admin service account token is mount to Che Server pod;
+#
+# It DOES NOT work on the following configurations:
+# - Che Server communicates with Kubernetes API using token from OAuth provider;
+che.infra.kubernetes.runtimes_concistency_check_period_min=-1
```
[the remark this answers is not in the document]
sleshchenko: Added

PR #12264, opened 2018-12-22
[WIP DO NOT MERGE] Update license headers

mkuznyetsov:
### What does this PR do?
Adapting to `license-maven-plugin-git` extension, that would be used for checking license years: https://github.com/mycila/license-maven-plugin/tree/master/license-maven-plugin-git

here are the changes to existing files
- license headers with starting and ending years being the same year, would be collapsed into one - e.g. "2018-2018"  would be transformed into "2018"
- there are some files, that have last year in license header as 2018, however git indicates those files being last changed in 2017 (date of commit).
 This causes an odd behaviour of license plugin - mvn license:check cannot pass, because it wants year 2017, but setting it to 2017 and running maven license:check again would fail, because it picks up the last change from git, that was made now (in 2018), so it wants 2018 again, creating a loop situation.

The solution to this is described on `license-maven-plugin-git` page, is to make two commits, one which will make changes to files ending on -2017, and the next commit that will change it back to 2018. This is why it is important to merge this PR without squash, with two commits as is.

This PR should be merged only after respective changes in che-parent and che-dev
https://github.com/eclipse/che-dev/pull/16

### DIFF
--- a/assembly/assembly-workspace-loader-war/pom.xml
+++ b/assembly/assembly-workspace-loader-war/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!--
 
-    Copyright (c) 2018-2018 Red Hat, Inc.
+    Copyright (c) 2018 Red Hat, Inc.
     This program and the accompanying materials are made
     available under the terms of the Eclipse Public License 2.0
     which is available at https://www.eclipse.org/legal/epl-2.0/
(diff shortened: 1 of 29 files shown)


mkuznyetsov:
ci-build

mkuznyetsov:
ci-test

mkuznyetsov:
ci-build

mkuznyetsov:
ci-build

mkuznyetsov:
ci-build

PR #12271, opened 2018-12-23
Fix the output of the GET /activity to just contain the array of ws ids

metlos:
### What does this PR do?
The new `GET /activity` REST endpoint introduced by #12177 returns the full serialized page object as its output instead of just the array of workspace ids as it was meant to return.

This PR fixes that and also adds a test for it.

#### What issues does this PR reference?

#12091


metlos:
ci-test

skabashnyuk (approved):

mshaposhnik (approved):

sleshchenko (commented):
LGTM

PR #12289, opened 2018-12-27
Use 'envFrom' in Che Deployment instead of duplicating of entries from config map

sleshchenko:
### What does this PR do?
I wanted to test my functionality which requires to be set environment variable that is not present in Helm Chart yet. I deployed Che with helm chart and then modified ConfigMap, and it doesn't work since we declare environment variables in Che Deployment one by one. So, Che Deployment and ConfigMap should be updated but it is not really needed.
This PR adds using 'envFrom' in Che Deployment instead of duplicating of entries from config map. Now helm chart doesn't have duplicated environment vars and `if` statements for optional env vars.

Dear reviewers, what your opinion about these changes? Maybe someone knows the bad side of using `envFrom`.

-

#### Release Notes
N/A

#### Docs PR
N/A


eivantsov:
💓 envFrom :)

ghost (approved):

garagatyi (approved):
Amazing stuff!

l0rd (approved):
Love that. What about OpenShift? Do you plan do open a similar PR for OpenShift as well replacing template parameters with configmaps? I mean similar deployments are easier to maintain.

sleshchenko:
@l0rd 

> Do you plan do open a similar PR for OpenShift as well replacing template parameters with configmaps? 

I don't plan to open a similar PR for OpenShift template because there is no config map at all. Che Server configuration is listed as environment variables in Che Deployment directly and there are no copied/pasted pieces. 

We won't be able to remove template parameters by using ConfigMap. One thing that would change - template parameters would be set to config map but not to environment variables for Che Deployment.

It makes sense to consider moving Che Server configuration from Deployment to ConfigMap but that another story =)
@eivantsov WDYT?

eivantsov:
Sure. This is what I do in https://github.com/eclipse/che-operator/blob/initial_pr/pkg/operator/che_cm.go

skabashnyuk (approved):

PR #12305, opened 2018-12-28
Initial devfile to work che in che

skabashnyuk:
### What does this PR do?

Introduce ***.devfile** skeleton to work che in che.
input

### How to test ###
- ```http://{CHE_HOST}/f?url=https://github.com/eclipse/che/tree/devfileche```
- after pr will be merged  ```http://{che-host}/f?url=https://github.com/eclipse/che.git```

n/a

#### Release Notes
n/a

#### Docs PR
n/a

### DIFF
--- a/.che-dev.yaml
+++ b/.che-dev.yaml
@@ -1,0 +1,46 @@
+---
+kind: List
+items:
+ - apiVersion: v1
+   kind: Pod
+   metadata:
+     name: ws
+   spec:
+     containers:
+     - name: dev
+       image: 'eclipse/che-dev:nightly'
+       resources:
+         limits:
+           memory: 5Gi
+       volumeMounts:
+       - mountPath: /projects
+         name: projects
+       - mountPath: /home/user/.m2
+         name: maven
+     volumes:
+     - name: projects
+       persistentVolumeClaim:
+         claimName: projects
+     - name: maven
+       persistentVolumeClaim:
+          claimName: maven
+ - apiVersion: v1
+   kind: PersistentVolumeClaim
+   metadata:
+     name: projects
+   spec:
+     accessModes:
+      - ReadWriteOnce
+     resources:
+       requests:
+         storage: 2Gi
+ - apiVersion: v1
+   kind: PersistentVolumeClaim
+   metadata:
+     name: maven
+   spec:
+     accessModes:
+      - ReadWriteOnce
+     resources:
+       requests:
+         storage: 3Gi
--- a/.devfile
+++ b/.devfile
@@ -1,0 +1,24 @@
+specVersion: 0.0.1
+name: che-in-che
+projects:
+  - name: che
+    source:
+      type: git
+      location: 'https://github.com/eclipse/che.git'
+tools:
+  - name: theia-editor
+    type: cheEditor
+    id: org.eclipse.che.editor.theia:1.0.0
+  - name: exec-plugin
+    type: chePlugin
+    id: che-machine-exec-plugin:0.0.1
+  - name: che-dev
+    type: kubernetes
+    local: .che-dev.yaml
+commands:
+  - name: fastbuild
+    actions:
+      - type: exec
+        tool: che-dev
+        command: cd /projects/che && mvn clean install -Pnative -DskipTests=true
+        workdir: /projects/che


.devfile
```diff
@@ -0,0 +1,21 @@
...
+    id: org.eclipse.che.editor.theia:1.0.0
+  - name: exec-plugin
+    type: chePlugin
+    id: che-machine-exec-plugin:0.0.1
+commands:
+  - name: build
+    actions:
+      - type: exec
+        tool: theia-editor
+        command: ls
```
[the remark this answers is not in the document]
skabashnyuk: you are right. It will not be able to build che. It's just a dummy stub.

.devfile
```diff
@@ -0,0 +1,21 @@
...
+      location: 'https://github.com/eclipse/che.git'
+tools:
+  - name: theia-editor
+    type: cheEditor
+    id: org.eclipse.che.editor.theia:1.0.0
+  - name: exec-plugin
+    type: chePlugin
+    id: che-machine-exec-plugin:0.0.1
+commands:
+  - name: dummy-stub-command
```
sleshchenko: Why do we need `dummy-stub-command` for work Che in Che?

skabashnyuk: Because commands is a mandatory section. I'll create an issue to make them optional. But in general, we need to populate this list with a lot of useful commands. Here we can elaborate on which one we will be able to introduce with the current level of devfile support.

skabashnyuk:
I think the goal of this pr is to start using devfile to work on che in che. There is a lot to do but I think it's a good starting point.

.devfile
sleshchenko: We definitely need to create an issue because commands provided by plugins may be enough for development. If we all agree on that - maybe you can do it in this PR? I guess there is one line to change. @mshaposhnik Can you point what should be changed to make commands optional?

mshaposhnik: yes. remove the https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json#L16 line.

skabashnyuk: https://github.com/eclipse/che/pull/12308

.devfile
```diff
@@ -0,0 +1,14 @@
+specVersion: 0.0.1
+name: che-in-che
+projects:
+  - name: che
+    source:
+      type: git
+      location: 'https://github.com/eclipse/che.git'
+tools:
+  - name: theia-editor
+    type: cheEditor
```
garagatyi, line 10: Can you elaborate on why types differ from types in Che plugin registry?

skabashnyuk: I think the answer for now - because it was originally defined like that in the reference implementation.  https://github.com/redhat-developer/devfile/blob/master/petclinic.yaml#L25
Do you want to initiate a task to syncronize this names?

sunix:
> * Introduce the ability to define an environment variable in cheEditor/CheTool to be able to define Theia extensions
>   Example
> 
> ```
>           "env": {
>             "HOSTED_PLUGIN_HOSTNAME": "0.0.0.0",
>             "THEIA_EXTENSIONS": "@devpodio/theia-vue-extension,@theia/bunyan,@theia/debug,@theia/java,@theia/java-debug"
>           },
> ```

I am not sure we want to promote theia extensions (but theia plugins)

skabashnyuk:
>I am not sure we want to promote theia extensions (but theia plugins)

Ok, that was only my quick thoughts.

@sunix ok. so, for now, users would not be able to select any language specific things in devfile, they will be all prepackaged with Theia editor. Right?

sunix:
> > I am not sure we want to promote theia extensions (but theia plugins)
> 
> Ok, that was only my quick thoughts.
> 
> @sunix ok. so, for now, users would not be able to select any language specific things in devfile, they will be all prepackaged with Theia editor. Right?

yep unless we have some languages available as vscode ext/theia plugin

che-dev.yaml
```diff
@@ -0,0 +1,15 @@
...
+  metadata:
+   name: ws
+  spec:
+   containers:
+    -image: 'eclipse/che-dev:nightly'
+     name: dev
+     mem_limit: 1073741824
+     resources:
+      limits:
+       memory: 1536Mi
```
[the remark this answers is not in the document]
sleshchenko: @mshaposhnik I think for k8s/OS recipe `mem_limit` is not old outdated property but property added by mistake copy/pasted from `docker compose` recipe.

benoitf:
I have comment about its name: as .devfile is hidden, looks like che-dev.yaml should be hidden as well no ?

che-dev.yaml --> .che-dev-yaml or .devfile-che-dev.yaml, or whatever.

skabashnyuk:
we start making devfile hidden after we saw @l0rd 's repository https://github.com/che-incubator/chectl.
If no objections I'll put ```che-dev.yaml --> .che-dev-yaml```

skabashnyuk:
ci-test

che-dev.yaml
```diff
@@ -0,0 +1,32 @@
+---
```
sleshchenko: Looks like this file is unused since `Devfile` references `.che-dev.yaml`

skabashnyuk: it is suppose to be deleted

.che-dev.yaml
```diff
@@ -0,0 +1,32 @@
...
+ - apiVersion: v1
+   kind: PersistentVolumeClaim
+   metadata:
+     name: projects
+   spec:
+     accessModes:
+      - ReadWriteOnce
+     resources:
+       requests:
+         storage: 1Gi
```
sleshchenko: Since binaries are stored in the same folder where sources are stored - I'm not sure whether 1Gi is enough for Che project folder.
My local Che project folder requires 1.8GiB 
screenshot_20190130_151727

Also, are there plans to use PVC for maven repo not to download maven dependencies after each workspace restart?

l0rd:
@skabashnyuk testing with `http://{che-host}/f?url=https://raw.githubusercontent.com/eclipse/che/devfileche/.devfile` doesn't work. 

I had the following error:

```
The local file '.che-dev.yaml' defined in tool 'che-dev' is unreachable or empty.
```

skabashnyuk:
@l0rd try this way ```http://{che-host/f?url=https://github.com/eclipse/che/tree/devfileche``` we need to check if external resources are allowed in factories where you specify a direct link to a devfile.

sleshchenko (approved):
I think it's a good start point.

.devfile
```diff
@@ -0,0 +1,24 @@
...
+    id: org.eclipse.che.editor.theia:1.0.0
+  - name: exec-plugin
+    type: chePlugin
+    id: che-machine-exec-plugin:0.0.1
+  - name: che-dev
+    type: openshift
```
l0rd: This doesn't work on kubernetes. If you set `type: kubernetes` would that work on OpenShift too?

sleshchenko: @l0rd Good catch! `kubernetes` tool works fine on both of Kuberentes/OpenShift infrastructures since `.che-dev.yaml` does not contain OpenShift specific objects like Routes.

.devfile
```diff
@@ -0,0 +1,24 @@
...
+    id: che-machine-exec-plugin:0.0.1
+  - name: che-dev
+    type: kubernetes
+    local: .che-dev.yaml
+commands:
+  - name: build
+    actions:
+      - type: exec
+        tool: che-dev
+        command: cd /projects/che && mvn -Pnative clean install
```
sleshchenko: I'm OK with merging this PR as an initial state. Just FYI command is failed with the following error
screenshot_20190204_171022

skabashnyuk: I'll take a look.

.devfile
```diff
@@ -0,0 +1,24 @@
...
+    id: che-machine-exec-plugin:0.0.1
+  - name: che-dev
+    type: kubernetes
+    local: .che-dev.yaml
+commands:
+  - name: build
+    actions:
+      - type: exec
+        tool: che-dev
+        command: cd /projects/che && mvn clean install -Pnative -DskipTests=true
```
sleshchenko, line 23: I think it makes sense to have two commands - for building with and without tests. Anyway, command name says nothing about skipping tests.

vparfonov (approved):

Issue #11906, opened 2018-11-12
An ability to specify Kubernetes/OpenShift List as tool in Devfile


sleshchenko:
### Description
To start quicker it was decided to start with Devfile that will contain editor and plugins only.
But it is needed to add an ability to specify workspace environment in Devfile.
This issue is about changing Devfile model to be able to handle Kubernetes/OpenShift tool.
<details>
<summary>
# An example of Devfile with Kubernetes/OpenShift tool specified
</summary>

```yaml
---
specVersion: 0.0.1
name: petclinic-dev-environment
projects:
  - name: petclinic
    source:
      type: git
      location: 'git@github.com:spring-projects/spring-petclinic.git'
tools:
  - name: petclinic-app
    type: kubernetes
    local: petclinic-app.yml
  - name: theia-ide
    type: cheEditor
    id: eclipse/theia:0.0.3
  - name: jdt.ls
    type: chePlugin
    id: eclipse/theia-jdtls:0.0.3
commands:
  - name: run
    actions:
      - type: exec
        tool: petclinic-app
        command: `./catalina.sh start`
        workdir: /home/user/app/tomcat
```
</details>

Note that this issue is only about Devfile model changes to be able to handle kubernetes/openshift tool.
`DevfileService#createWorkspaceFromYaml` and `DevfileService#createFromWorkspace` methods won't support `kubernetes`/`openshift` tool since it contains relative link to tool configuration file (at least  for time being).
`kubernetes`/`openshift` tool will be supported by Factory workflow when workspace is created from repository inside of repository, the following issue https://github.com/eclipse/che/issues/12266.

sleshchenko:
The issue is updated to support Kubernetes/OpenShift List as tool in Defile. The separate issue should be created for supporting `dockerimage` tool by Devfile.

PR #12307, opened 2018-12-28
Add kubernetes type tool into devfile API

mshaposhnik:
### What does this PR do?
Add for new `kubernetes` and `openshift` tool types for K8S list  into devfile API to be able to specify workspace environment configuration in Devfile.
<details>
<summary>
# Example of Devfile with Kubernetes/OpenShift tool specified
</summary>

```yaml
---
specVersion: 0.0.1
projects:
  - name: petclinic
    source:
      type: git
      location: 'git@github.com:spring-projects/spring-petclinic.git'
tools:
  - name: petclinic-app
    type: kubernetes
    local: petclinic-app.yml
  - name: theia-ide
    type: cheEditor
    id: eclipse/theia:0.0.3
  - name: jdt.ls
    type: chePlugin
    id: eclipse/theia-jdtls:0.0.3
commands:
  - name: run
    actions:
      - type: exec
        tool: petclinic-app
        command: `./catalina.sh start`
        workdir: /home/user/app/tomcat
```
</details>

https://github.com/eclipse/che/issues/11906

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/Constants.java
+++ b/wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/Constants.java
@@ -23,6 +23,8 @@ public class Constants {
 
   public static final String KUBERNETES_TOOL_TYPE = "kubernetes";
 
+  public static final String OPENSHIFT_TOOL_TYPE = "openshift";
+
   /**
    * Workspace attribute which contains comma-separated list of mappings of tool id to its name
    * Example value:
--- a/wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/Constants.java
+++ b/wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/Constants.java
@@ -21,6 +21,8 @@ public class Constants {
 
   public static final String PLUGIN_TOOL_TYPE = "chePlugin";
 
+  public static final String KUBERNETES_TOOL_TYPE = "kubernetes";
+
   /**
    * Workspace attribute which contains comma-separated list of mappings of tool id to its name
    * Example value:
--- a/wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileConverter.java
+++ b/wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileConverter.java
@@ -125,7 +125,9 @@ public class DevfileConverter {
           break;
         case KUBERNETES_TOOL_TYPE:
         case OPENSHIFT_TOOL_TYPE:
-          continue; // this kind of tools shouldn't be present in mapping
+          // this kind of tool ignored here since it contain only reference to tool configuration
+          // which should be resolved and provisioned into workspace config recipe separately.
+          continue;
         default:
           throw new DevfileFormatException(
               format("Unsupported tool %s type provided: %s", tool.getName(), tool.getType()));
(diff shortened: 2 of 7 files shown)


wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileConverter.java
```diff
@@ -121,6 +122,8 @@ public WorkspaceConfigImpl devFileToWorkspaceConfig(Devfile devfile)
         case PLUGIN_TOOL_TYPE:
           pluginsStringJoiner.add(tool.getId());
           break;
+        case KUBERNETES_TOOL_TYPE:
+          continue; // this kind of tools shouldn't be present in mapping
```
skabashnyuk: throw exception?

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/Constants.java
```diff
@@ -21,6 +21,8 @@
 
   public static final String PLUGIN_TOOL_TYPE = "chePlugin";
 
+  public static final String KUBERNETES_TOOL_TYPE = "kubernetes";
```
sleshchenko, line 24: Please add `openshift` tool type as well.

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileConverter.java
mshaposhnik: No, that's normal situation. We just have nothing to do with those tool type during conversion into WS config (and it should not be present in workspace plugins mapping attribute).

mshaposhnik:
ci-test

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileConverter.java
sleshchenko: Well, I also find this behavior and comment confusing. Maybe it would be better to rewrite this in another way like  the following:
```java
         case KUBERNETES_TOOL_TYPE:
         case OPENSHIFT_TOOL_TYPE:
           // this kind of tool contain only reference to tool configuration
           // it is ignored on converting to workspace config
          // then tool configuration should be provisioned to workspace config recipe
          // after downloading it.
           continue;                         
```

wsmaster/che-core-api-devfile/src/test/resources/devfile.yaml
```diff
@@ -28,6 +28,12 @@ tools:
   - name: jdt.ls
     type: chePlugin
     id: eclipse/theia-jdtls:0.0.3
+  - name: k8s-recipe
```
sleshchenko: We are not going to support more than one tool that is not `cheEditor` nor `chePlugin`. It would be better not to have such example even in test sources

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileConverter.java
sleshchenko: Also, we are not going to support more than one tool that is not cheEditor nor chePlugin. Is there any validation check for it?

wsmaster/che-core-api-devfile/src/test/resources/devfile.yaml
mshaposhnik: ok

sleshchenko (approved):
LGTM

skabashnyuk (approved):
Please add more information in description. With examples and motivational part.

Issue #12266, opened 2018-12-22
Factory flow should support k8s/OS tool in Devfile


sleshchenko:
### Description
The goal is to improve factory flow to support k8s/OS tool with a relative path to a configuration file specified in Devfile.
<details>
<summary>
# An example of Devfile with Kubernetes/OpenShift tool specified
</summary>

```yaml
---
specVersion: 0.0.1
name: petclinic-dev-environment
projects:
  - name: petclinic
    source:
      type: git
      location: 'git@github.com:spring-projects/spring-petclinic.git'
tools:
  - name: petclinic-app
    type: kubernetes
    local: petclinic-app.yml
  - name: theia-ide
    type: cheEditor
    id: eclipse/theia:0.0.3
  - name: jdt.ls
    type: chePlugin
    id: eclipse/theia-jdtls:0.0.3
commands:
  - name: run
    actions:
      - type: exec
        tool: petclinic-app
        command: `./catalina.sh start`
        workdir: /home/user/app/tomcat
```
</details>

An idea is not to embed k8s/OS yaml into Devfile model but just reference file with k8s/OS content that is located in the same repository where Devfile lives.
Then `resolve factory` flow should parse Devfile and fetch k8s/OS tool configuration content by specified relative path in repository. Then the environment should be provisioned into `Factory#getWorkspaceConfig` with fetched k8s/OS configuration as content of recipe with type `kubernetes/openshift`.
<details>
<summary>
# An example OpenShift Tool Configuration
</summary>

```yaml
apiVersion: v1
kind: List
items:
- apiVersion: v1
  kind: Pod
  metadata:
    name: petclinic
    labels:
      app: petclinic
  spec:
    containers:
    - name: server
      image: mariolet/petclinic
      ports:
      - containerPort: 8080
        protocol: TCP
      resources:
        limits:
        memory: 512Mi
    - name: mysql
      image: centos/mysql-57-centos7
      env:
      - name: MYSQL_USER
        value: petclinic
      - name: MYSQL_PASSWORD
        value: petclinic
      - name: MYSQL_ROOT_PASSWORD
        value: petclinic
      - name: MYSQL_DATABASE
        value: petclinic
      ports:
      - containerPort: 3306
        protocol: TCP
      resources:
        limits:
        memory: 512Mi
- kind: Service
  apiVersion: v1
  metadata:
    name: petclinic
    labels:
      app: petclinic
  spec:
    ports:
      - name: web
        port: 8080
        targetPort: 8080
[...]

PR #12353, opened 2019-01-08
Support of kubernetes/openshift tools in Factory

mshaposhnik:
### What does this PR do?
This PR intoduces support of kubernetes/openshift tools in Factories sourced by devfile, based on relative links to k8s/os lists file inside of those tools.  
**Note:** this feature requires that the underlying service provider have to be able to convert relative link into full URL to the recipe content file, so currenlty only GitHub based factories will support such kind of tools.

For the examples of devfile and recipe file, see the  generic issue.

https://github.com/eclipse/che/issues/12266


wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/urlfactory/DevfileEnvironmentProvisioner.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  /**
+   * Finds an recipe-type tool (openshift or kubernetes) in devfile and tries to provision {@link
+   * EnvironmentImpl} from it. If such tools are present in devfile, an file URL composer function
+   * MUST be provided to allow to fetch recipe content.
+   *
+   * @param devfile source devfile
+   * @param fileUrlComposer optional service-specific composer of URL's to the file raw content
+   * @return provisioned enviromnent
```
sleshchenko: typo `enviromnent` -> `environment`

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/urlfactory/DevfileEnvironmentProvisioner.java
```diff
@@ -0,0 +1,89 @@
...
+    String localFileContent = urlFetcher.fetch(fileUrlComposer.apply(recipeTool.getLocal()));
+    if (isNullOrEmpty(localFileContent)) {
+      throw new BadRequestException(
+          "The local file '"
+              + recipeTool.getLocal()
+              + "' defined in tool  '"
+              + recipeTool.getName()
+              + "' is unreachable or empty.");
+    }
+    // TODO: it would be great to check there is real yaml and not binary etc
```
sleshchenko: It would be brilliant :smile_cat: 
Are you going to do it in this PR?

plugins/plugin-github/che-plugin-github-factory-resolver/src/main/java/org/eclipse/che/plugin/github/factory/resolver/GithubUrl.java
```diff
@@ -173,6 +173,21 @@ protected String devfileFileLocation() {
         .toString();
   }
 
+  /**
+   * Provides location to raw content of specified file
+   *
+   * @return location of specified file in a repository
+   */
+  protected String rawFileLocation(String fileName) {
+    return new StringJoiner("/")
```
sleshchenko, line 170: This code duplicates `protected String devfileFileLocation()`. I think it would be better to avoid such duplication by rewriting `devfileFileLocation` to 
```java
protected String devfileFileLocation() {
  return rawFileLocation(devfileFilename);
}
```

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/urlfactory/DevfileEnvironmentProvisioner.java
```diff
@@ -0,0 +1,89 @@
...
+import org.eclipse.che.api.core.BadRequestException;
+import org.eclipse.che.api.devfile.model.Devfile;
+import org.eclipse.che.api.devfile.model.Tool;
+import org.eclipse.che.api.workspace.server.model.impl.EnvironmentImpl;
+import org.eclipse.che.api.workspace.server.model.impl.RecipeImpl;
+import org.eclipse.che.commons.annotation.Nullable;
+
+/** Creates workspace environment from specific tool in devfile if any. */
+@Singleton
+public class DevfileEnvironmentProvisioner {
```
sleshchenko: Java doc says that it `creates` but not `provisions`. Maybe then it is `DevfileEnvironmentFactory`?

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/urlfactory/DevfileEnvironmentProvisioner.java
```diff
@@ -0,0 +1,89 @@
...
+  private final URLFetcher urlFetcher;
+
+  @Inject
+  public DevfileEnvironmentProvisioner(URLFetcher urlFetcher) {
+    this.urlFetcher = urlFetcher;
+  }
+
+  /**
+   * Finds an recipe-type tool (openshift or kubernetes) in devfile and tries to provision {@link
+   * EnvironmentImpl} from it. If such tools are present in devfile, an file URL composer function
```
sleshchenko: > ... If such tools are present in devfile ...

I guess it must be only `ONE` or `0`. Maybe it would be better to throw an exception in case when there are multiply `Kubernetes`/`OpenShift` tools instead of ignoring them?

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/urlfactory/URLFactoryBuilder.java
```diff
@@ -84,20 +91,28 @@ public URLFactoryBuilder(
...
+    if (isNullOrEmpty(devfileYamlContent)) {
       return Optional.empty();
     }
     try {
       Devfile devfile = devfileManager.parse(devfileYamlContent, false);
+      Optional<EnvironmentImpl> environment =
+          devfileEnvironmentProvisioner.tryProvision(devfile, fileUrlComposer);
       WorkspaceConfigImpl wsConfig = devfileManager.createWorkspaceConfig(devfile);
+      if (environment.isPresent()) {
+        wsConfig.setDefaultEnv("default");
```
sleshchenko: I think it would be better to use tool name as environment name instead of using `default` value. WDYT?

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/urlfactory/DevfileEnvironmentProvisioner.java
sleshchenko: Don't be shy and add your authorship :smiley_cat:

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/urlfactory/DevfileEnvironmentProvisioner.java
```diff
@@ -0,0 +1,89 @@
...
+        devfile
+            .getTools()
+            .stream()
+            .filter(
+                tool ->
+                    tool.getType().equals(KUBERNETES_TOOL_TYPE)
+                        || tool.getType().equals(OPENSHIFT_TOOL_TYPE))
+            .findFirst();
+    if (!recipeToolOptional.isPresent()) {
+      return Optional.empty();
```
sleshchenko: As far as I see this case is not covered by unit tests. Please consider adding such test.

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/urlfactory/DevfileEnvironmentProvisioner.java
```diff
@@ -0,0 +1,89 @@
...
+  /**
+   * Finds an recipe-type tool (openshift or kubernetes) in devfile and tries to provision {@link
+   * EnvironmentImpl} from it. If such tools are present in devfile, an file URL composer function
+   * MUST be provided to allow to fetch recipe content.
+   *
+   * @param devfile source devfile
+   * @param fileUrlComposer optional service-specific composer of URL's to the file raw content
+   * @return provisioned enviromnent
+   * @throws BadRequestException when there is recipe-type tool present in devfile but no URL
+   *     composer provided when file specified in local section of the tool is unreachable or empty
```
sleshchenko: I guess `of the tool is unreachable or empty` should be `or the tool is unreachable or empty`

I think it would be better to describe different exception cases in different `@throws` sections:
```java
   * @throws BadRequestException when there is no URL provider for recipe-type tool present in devfile
   * @throws BadRequestException recipe-type tool content is unreachable or empty
```

sleshchenko (commented):
In PR description you described that `K8s/OS_Tool#local` contains a relative link. I guessed that link should be relative to Devfile location, but as far as I understand it is relative to the root of a repository.

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/urlfactory/DevfileEnvironmentProvisioner.java
mshaposhnik: That will require to introduce an yaml parsing framework into factory. I think we can postpone that until we start work on #12361 which will require that anyway.

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/urlfactory/URLFactoryBuilder.java
mshaposhnik: Yes i thought abt this but it is little bit tricky since we have no tool itself here. I will try again.

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/urlfactory/DevfileEnvironmentProvisioner.java
sleshchenko: OK.

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/urlfactory/DevfileEnvironmentProvisioner.java
mshaposhnik: It is covered by validator, but yes maybe it makes sense.

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/urlfactory/DevfileEnvironmentProvisioner.java
sleshchenko: Referenced this conversation in issue.

plugins/plugin-github/che-plugin-github-factory-resolver/src/main/java/org/eclipse/che/plugin/github/factory/resolver/GithubUrl.java
mshaposhnik: Ok

mshaposhnik:
> In PR description you described that K8s/OS_Tool#local contains a relative link. I guessed that link 
> should be relative to Devfile location, but as far as I understand it is relative to the root of a repository.

At the given point of time this is quite the same - because we doesn't support .devfile or factory.json to be located into subfolders now.

skabashnyuk (approved):

mshaposhnik:
ci-test

sleshchenko (approved):
LGTM

PR #12367, opened 2019-01-09
use che.ls.jdt.version = 0.0.3-SNAPSHOT (DO...

nickboldt:
use che.ls.jdt.version = 0.0.3-SNAPSHOT (DO NOT MERGE until https://ci.codenvycorp.com/job/che-ls-jdt-ci-master/ is green w/ all tests passing)

Change-Id: Ifb6e3446160d014da5945f78c9ec3762f42033b2
Signed-off-by: nickboldt <nboldt@redhat.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,7 +52,7 @@
         <che.dashboard.version>6.17.0-SNAPSHOT</che.dashboard.version>
         <che.docs.version>6.17.0-SNAPSHOT</che.docs.version>
         <che.lib.version>6.17.0-SNAPSHOT</che.lib.version>
-        <che.ls.jdt.version>0.0.2</che.ls.jdt.version>
+        <che.ls.jdt.version>0.0.3-SNAPSHOT</che.ls.jdt.version>
         <che.version>6.17.0-SNAPSHOT</che.version>
         <specification.version>1.0-beta2</specification.version>
     </properties>


riuvshin (changes requested):
we should always use TAG of che-ls-jdt, since it is separate project with own release cycle

nickboldt:
well, if we don't use the snapshot version, I can still build Che & CRW using 0.0.3-snapshot version here: https://codeready-workspaces-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/CRW_CI/view/Pipelines/job/crw_master/148/consoleFull (thanks to the magic of sed patching) :D

riuvshin:
@nickboldt why do you need che-ls-jdt SNAPSHOT?
normally CHE should never use snapshot version unless you are developing che-ls-jdt and want to test locally changes.
Think of it as just a 3rd party dependency, like testng
you don't want to use snapshot of testng because it may broke your application, so you always want to use tested tag.

tsmaeder:
@riuvshin because we cannot build che-ls-jdt 0.0.2 anymore, due to the fact that jdt.ls does not have a stable target platform. @nickboldt  rebuilds anything he can in his build process (for his own good reasons), so now his build is borked. Anyway, che now depends on che-ls-jdt 0.0.3-SNAPSHOT, so we can close this one.

Issue #12242, opened 2018-12-20
[Metric] Number of failed workspace startup


ibuziuk:
**Description**
Add Prometheus metrics that will show us a number of failed workspace startup

PR #12368, opened 2019-01-09
Add workspace failure count metric.

metlos:
### What does this PR do?

Reports the number of failed workspaces as a prometheus metric.
The metric is reported separately for each status in which a workspace can fail.

This is an excerpt from the raw output of the `/metric` prometheus endpoint:
```
# HELP che_workspace_failure_total The count of failed workspaces
# TYPE che_workspace_failure_total counter
che_workspace_failure_total{area="workspace",while="STARTING",} 1.0
che_workspace_failure_total{area="workspace",while="RUNNING",} 0.0
che_workspace_failure_total{area="workspace",while="STOPPING",} 0.0
```

the `while` tag signifies the status in which the workspace was when it failed.

#12242 

#### Release Notes
* https://github.com/eclipse/che-docs/pull/658

### DIFF
--- a/wsmaster/che-core-api-metrics/pom.xml
+++ b/wsmaster/che-core-api-metrics/pom.xml
@@ -41,10 +41,6 @@
         </dependency>
         <dependency>
             <groupId>org.eclipse.che.core</groupId>
-            <artifactId>che-core-api-dto</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.eclipse.che.core</groupId>
             <artifactId>che-core-api-model</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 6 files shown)


wsmaster/che-core-api-metrics/pom.xml
```diff
@@ -39,6 +39,10 @@
             <groupId>org.eclipse.che.core</groupId>
             <artifactId>che-core-api-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.che.core</groupId>
+            <artifactId>che-core-api-dto</artifactId>
```
sleshchenko: I guess it should have `test` scope. To check it - please execute `mvn clean install -DskipTests` and maven dependency plugin will test the right scope.

sleshchenko (approved):
LGTM

Please take a look my inline comment

wsmaster/che-core-api-metrics/pom.xml
metlos: Good point, it should be in the test scope! maven dependency plugin doesn't complain though in either case...

mshaposhnik (approved):

skabashnyuk (approved):

PR #12375, opened 2019-01-10
CHE-12329 Rework the way of retrieving editor provider by file-type

### LINKED ISSUE #12329: Regexp for pom.xml not working in LanguageServerConfig
### Description

The issue emerged between 6.15 (it's ok in 6.15) and 6.16.

In Bayesian language server config, the following regexp is used. However, Language Server isn't initialized when pom.xml is opened. LS is inited for package.json and requirements.txt though.

```
      public Map<String, String> getLanguageRegexes() {
        return ImmutableMap.<String, String>builder()
            .put(BayesianLanguageServerModule.TXT_LANGUAGE_ID, ".*requirements\\.txt")
            .put(JsonModule.LANGUAGE_ID, ".*package\\.json")
                // todo revert back to pom.xml when 
            .put("pom", ".*pom\\.xml")
            .build();
      }
```

Interestingly, `".*\\.xml"` works fine, however, in this case LS is initialized for all .xml files.

vinokurig:
### What does this PR do?
Fixes bug when a maven file-type (without linked editor provider) is registered and another file type for `pom.xml` files (with linked editor provider) is registered as well, but when the `pom.xml` file is opened default editor provider is returned. The fix iterates registered editor providers if the the default provider is registered for the found file type.

fixes #12329 

#### Release Notes

N/A

#### Docs PR

N/A

### DIFF
--- a/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/editor/EditorAgentImpl.java
+++ b/ide/che-core-ide-app/src/main/java/org/eclipse/che/ide/editor/EditorAgentImpl.java
@@ -308,9 +308,9 @@ public class EditorAgentImpl
       VirtualFile file, EditorPartStack editorPartStackConsumer, OpenEditorCallback callback) {
 
     addToOpeningFilesList(file.getLocation(), editorPartStackConsumer);
-    Set<FileType> fileTypesByFileName = getFileTypesByFileName(file.getName());
+    Set<FileType> fileTypesByFile = getFileTypesByFile(file);
     Optional<FileType> registeredFileType =
-        fileTypesByFileName
+        fileTypesByFile
             .stream()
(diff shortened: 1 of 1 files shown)


RomanNikitenko (changes requested):
What about recovery editor state after refreshing page for example?
I guess we have the same problem for this case and the corresponding logic is placed in the same file.
So may be we should apply the same fix for this case.
WDYT?

RomanNikitenko (approved):

vinokurig:
@RomanNikitenko Agree, added changes

vinokurig:
ci-test

dkulieshov (approved):

eivantsov:
Are we good to merge this one?

vinokurig:
I think we need to check the selenium tests report @musienko-maxim

vinokurig:
ci-test

vinokurig:
ci-test

musienko-maxim (approved):

Issue #7381, opened 2017-11-15
Implement validation of OpenShift recipe


sleshchenko:
### Description
We parse OpenShift list with fabric8 OpenShift client, but it doesn't perform any validation like checking required fields (metadata.name, pod.spec.containers). It is needed to implement validation of OpenShift recipe.

ibuziuk:
Some UD validation has been implemented as part of the https://github.com/eclipse/che/pull/12203

PR #12383, opened 2019-01-10
Fix NPE on recipe parsing

garagatyi:
### What does this PR do?
Fix NPE on recipe parsing.

### DIFF
--- a/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenShiftEnvironmentFactory.java
+++ b/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenShiftEnvironmentFactory.java
@@ -111,6 +111,7 @@ public class OpenShiftEnvironmentFactory extends InternalEnvironmentFactory<Open
         throw new ValidationException("Supporting of deployment configs is not implemented yet.");
       } else if (object instanceof Pod) {
         Pod pod = (Pod) object;
+        checkNotNull(pod.getMetadata(), "Pod metadata must not be null");
         pods.put(pod.getMetadata().getName(), pod);
       } else if (object instanceof Service) {
         Service service = (Service) object;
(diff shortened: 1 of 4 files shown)


infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenShiftEnvironmentFactory.java
```diff
@@ -111,6 +111,7 @@ protected OpenShiftEnvironment doCreate(
         throw new ValidationException("Supporting of deployment configs is not implemented yet.");
       } else if (object instanceof Pod) {
         Pod pod = (Pod) object;
+        checkNotNull(pod.getMetadata(), "Pod metadata must not be null");
```
sleshchenko, line 116: I think the same should be done for other objects. Also, I see no check that name inside of metadata is not null or empty while it must not be.
Maybe it would be better have somehow util method that will safely retrieve name of `HasMetadata` object?
I mean something like that
```
  public String getName(HasMetadata object) {
    ObjectMeta metadata = object.getMetadata();
    checkNotNull(metadata, String.format("%s metadata must not be null", object.getKind()));
    String name = metadata.getName();
    checkNotNull(metadata, String.format("%s name must not be null", object.getKind()));
    return name;
  }
```
Then we would be able to reuse it for all kinds of K8s/OpenShift objects. WDYT?

skabashnyuk:
@sleshchenko @garagatyi Does it make sense to discuss here similar issue in other places or you want to narrow the scope of this pr to ```OpenShiftEnvironmentFactory.java```

sleshchenko:
@skabashnyuk Good catch. I think that It's good time to implement other validations rules https://github.com/eclipse/che/issues/7381

garagatyi:
@sleshchenko @skabashnyuk I added the same check to k8s recipe parser. Also added the check of name nullness. But the implementation of this is not in my sprint, so I'm not going to expand improvements in this PR. I just wanted to fix NPE we've got in RhChe installation.

skabashnyuk (approved):

sleshchenko (approved):
> I just wanted to fix NPE we've got in RhChe installation.

I know what to do to get the same checks fixed for another objects ;)

Issue #11648, opened 2018-10-17
Wrong behavior in the editor when perform git pull after delete files


artaleks9:
### Description

**Reproduction Steps:**
* Open the  IDE and import a project (e.g. https://github.com/artaleks9/Spoon-Knife.git)
* Expand the project and open in the editor all five files
* Go to GitHub side and delete three files from the _plan-files_ folder
* Go to back the IDE and perform git pull

**Expected behavior:**
* the editor should be updated 
* all tabs of the deleted files should disappear from the editor
 
**Observed behavior:**
* one or two tabs of the deleted files are present in the editor
* after refreshing the browser page the deleted 
tabs disappear, 
but in the browser console there is an error
 
**Che version**: 6.13.0
**OS and version:** Fedora 28
**Che install:** docker/ocp
**Docker version:** 17.09
**API docker version:** 1.32

**Additional information:**
* Problem can be reliably reproduced, doesn't happen randomly: [Yes]
* Related selenium test: _GitPullTest_
* Failed test report: https://ci.codenvycorp.com/view/qa-nightly/job/che-integration-tests-master-ocp/433/Selenium_20tests_20report/failsafe-report.html#org.eclipse.che.selenium.gitGitPullTest 
* See the attachment:

**screenshots:**

**_after git pull_**
![git-pull-test-2](https://user-images.githubusercontent.com/5408906/47089309-3afdbe00-d229-11e8-91c0-8fdb2632cab8.png)

**_after refresh browser page_**
![git-pull-test-1](https://user-images.githubusercontent.com/5408906/47089432-831ce080-d229-11e8-9a25-d980bc34a68b.png)

**_github page_**
![git-pull-test-3](https://user-images.githubusercontent.com/5408906/47089457-8e700c00-d229-11e8-86a6-18f138d8b396.png)

PR #12384, opened 2019-01-10
Check for additional file events in watch event poll

vzhukovs:
### What does this PR do?
This changes proposal adds fix for the problem related to sequential handling file watcher events. The problem is in mechanism of receiving events: https://github.com/eclipse/che/blob/master/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/watcher/server/impl/FileWatcherService.java#L272. We try to get events and continue work with them but we don't check that in this moment we can receive additional events.

Use case to reproduce such problem: use git repository with the following structure:
```
.
├── dir
│   ├── file1
│   ├── file2
│   └── file3
└── other.file
```
clone this repo into che and make outside operation to make the following structure:
```
.
├── dir
└── other.file
```
file watcher detects changes and we call https://github.com/eclipse/che/blob/master/wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/watcher/server/impl/FileWatcherService.java#L272 we receive two events about delete `file1` and `dir`, but in a moment we receive rest two events that is not handled for `file2` and `file3`.

#11648 

#### Release Notes

N/A

#### Docs PR

N/A


vzhukovskii:
ci-test

vzhukovskii:
ci-test

vzhukovskii:
ci-build

artaleks9:
Selenium tests execution on Eclipse Che Multiuser on OCP (https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1408/) doesn't show any regression against this Pull Request.
The bug is not reproduced when checking on local machine.

wsagent/che-core-api-project/src/main/java/org/eclipse/che/api/watcher/server/impl/FileWatcherService.java
```diff
@@ -269,33 +270,35 @@ private void run() {
           continue;
         }
 
-        List<WatchEvent<?>> watchEvents = watchKey.pollEvents();
+        for (List<WatchEvent<?>> watchEvents = watchKey.pollEvents();
```
vparfonov: maybe `while` statement here make code more readable ?

vzhukovs: You mean something like this?:

```
        List<WatchEvent<?>> watchEvents = watchKey.pollEvents();
        while(!watchEvents.isEmpty()) {
          if (suspended.get()) {
            ...
          }

          for (WatchEvent<?> event : watchEvents) {
            ...
          }

          watchEvents = watchKey.pollEvents();
        }
```

dkulieshov (approved):

vzhukovskii:
ci-build

vparfonov (approved):

Issue #11804, opened 2018-11-01
Apply ability filter user containers from among others


AndrienkoAleksandr:
### Description
In the Che 7 Workspace Model, a user is able to define a list of own "containers" and apply Che Plugins "tooling containers". Then all containers(machines) are available in `Workspace#getRuntime#getMachines` but it is not possible to identify which of containers are user-defined.
It makes sense to have an ability to identify where container comes from User Recipe or Che Plugin since this list is used for opening a terminal, executing of command.
A special attribute in [Machine#getAttributes](https://github.com/eclipse/che/blob/master/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/runtime/Machine.java#L24) would be fine.

garagatyi:
Well, sometimes it makes sense to execute command or have a terminal into tolling containers too! For example into Theia container

sleshchenko:
> Well, sometimes it makes sense to execute command or have a terminal into tolling containers too! For example into Theia container

@garagatyi Yeah, I agree. Let me clarify that Client would be able not just filter machines, but like order them (like first user-defined and then tooling) or display some labels near machine names to make it clear where it comes from.

slemeur:
What is your proposal for jwt-proxy then?

AndrienkoAleksandr:
Needed for https://github.com/eclipse/che/issues/11804 too.

PR #12386, opened 2019-01-11
Filter workspace containers like tools and user's

AndrienkoAleksandr:
### Referenced issue
https://github.com/eclipse/che/issues/11804

Add ability filter workspace containers like 'user-container' and 'tool-container'. 

Signed-off-by: Oleksandr Andriienko <oandriie@redhat.com>


wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
```diff
@@ -162,5 +162,7 @@
 
   public static final String NO_ENVIRONMENT_RECIPE_TYPE = "no-environment";
 
+  public static final String CONTAINER_TYPE_ATTRIBUTE = "type";
```
garagatyi: Please, add javadocs

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsplugins/KubernetesPluginsToolingApplier.java
```diff
@@ -194,6 +196,7 @@ private void addSidecar(
             .build();
 
     InternalMachineConfig machineConfig = machineResolver.resolve();
+    machineConfig.getAttributes().put(CONTAINER_TYPE_ATTRIBUTE, "tool-container");
```
garagatyi: It would be better to make value a constant too

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -340,6 +341,11 @@ public WorkspaceStatus getStatus(String workspaceId) {
           createInternalEnvironment(
               environment, workspaceConfig.getAttributes(), workspaceConfig.getCommands());
       RuntimeContext runtimeContext = infrastructure.prepare(runtimeId, internalEnv);
+
+      internalEnv.getMachines()
```
sleshchenko: Please consider moving this logic to [InternalEnvironmentFactory](https://github.com/eclipse/che/blob/aaa8f424cd198e68e128bca939d9ac058b9ea52e/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/environment/InternalEnvironmentFactory.java)

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -340,6 +341,11 @@ public WorkspaceStatus getStatus(String workspaceId) {
...
+              .values()
+              .forEach(internalMachineConf -> internalMachineConf.getAttributes().put(CONTAINER_TYPE_ATTRIBUTE, "user-container"));
```
garagatyi: It would be better to make value a constant too

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
garagatyi: What behavior would be if there was a machine without such an attribute?
What if the container doesn't have a corresponding machine?
Do you cover JWT proxy in this PR?

AndrienkoAleksandr: > What behavior would be if there was a machine without such an attribute?

I think if machine without attribute - that's something that was bringing from our api, it's some internal stuff and we should not display it for end users. It's implementation specific stuff and it should be hidden.

> Do you cover JWT proxy in this PR?

I didn't take a look this case. I will take a look.

l0rd:
The final goal is to filter out containers where a user can open a terminal. I think we should make it explicit and instead of `type: [user-container|tool-container]` we should use `terminal-available: [true|false]`. Does that make sense?

sleshchenko:
@l0rd It makes sense for me. Because plugin (tooling) may provide some CLI software in sidecar, like maven, git. And then it makes sense to propose user open terminal there.
But I also think that `terminal-available: false` should not deny terminal access but just hide such containers from a containers list by default. Terminal access to such containers may be useful while investigation plugins errors. WDYT?

garagatyi:
> The final goal is to filter out containers where a user can open a terminal. I think we should make it explicit and instead of `type: [user-container|tool-container]` we should use `terminal-available: [true|false]`. Does that make sense?

Why should we state that the terminal is NOT available in other containers? I think that there are cases when someone would want to have the terminal in those containers. At least Che/Theia contributors that found a bug while working with the project. We might skip those containers by default, but this doesn't mean we should say that the terminal is not available for them overall.

AndrienkoAleksandr:
@l0rd My changes supposed to be used not only for terminal access but also for an ability for hiding tools containers on  `Theia Containers View` that Stevan asks [[1](https://github.com/eclipse/che-theia/pull/13#issuecomment-453426644)]
[1] https://github.com/eclipse/che-theia/pull/13#issuecomment-453426644
And it could be useful for che-theia tasks too.

l0rd:
@sleshchenko yes it's just to filter them out from theia command palette. Maybe the attribute `visible-in-terminal` or `enable-terminal` would be better better? But anyway I think that we should avoid that common users get confused with containers where they should 99% of the time they will never open a terminal. For debugging purposes advanced contributors will `oc rsh` or `k exec`.

@garagatyi agreed. I see this PR as a first step (add the enable terminal attribute and make it true by default for users containers and false for plugins). A second PR should follow to make it possible to enable terminal for plugin containers. And as a third and final step we should implement the ability to disable terminal for some runtime containers.

@AndrienkoAleksandr I believe that the semantic is different and thus we need 2 distinct attributes. There will be situations where I don't want to enable a terminal but I want to see the container amongst the list and vice-versa. Think about plugins as well: jdt.ls sidecar for instance won't be a user container but the user will probably want to open a terminal there and we want to see it amongst the container list as well. Let's be explicit and use 2 different attributes (e.g. `enable-terminal` + `view-in-container-list`).

wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
AndrienkoAleksandr: Done.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
AndrienkoAleksandr: Thanks. Done.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsplugins/KubernetesPluginsToolingApplier.java
AndrienkoAleksandr: Done.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
AndrienkoAleksandr: Done.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
sleshchenko: @garagatyi Good catch. I think the following attribute should be added to machine of `JwtProxy`. See https://github.com/eclipse/che/blob/a1417ac9c63d40ffbbb5bffee46a55c36788c150/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/server/secure/jwtproxy/JwtProxyProvisioner.java#L256

AndrienkoAleksandr:
@l0rd Thanks for sharing Devfile model. I was not aware of it before.
@skabashnyuk  @sleshchenko and I discussed a bit how it can works with Devfile. Now, Devfile is converted to Workspace Config model. plugins, editor tools are stored in `WorkspaceConfig#attributes` and `kubernetes/openshift` tool is stored in `WorkspaceConfig#recipe`. Then we could auto provision source machine attribute that would indicate where machine comes from. Possible values: `recipe` , `plugin`, `system` (for jwt-proxy machine).
@l0rd WDYT?

l0rd:
Well Devfile is supposed to become the new way to define a workspace and the wording should not change much. WorkspaceConfig attributes instead may change names in next releases. I am not sure how long we will keep the name `recipe` for instance.

sleshchenko:
@l0rd Maybe later when `Devfile` will be the new way to define a workspace we could use another attribute marker like `tool type`, then we would have `plugin`, `editor`, `kubernetes`, `dockerimage`, etc. and client would not show containers of `plugin` and `editor` tools.
Since this attribute marker is not stored anywhere it should not be a big deal to rework Che Server and UI to use new attribute marker later. WDYT?

l0rd (approved):
@AndrienkoAleksandr let's merge this PR in its current status. We don't need to discuss too much about it right now and we can rework if needed in the future.

We can even use the `tool` attribute because we plan to rename `tools` to `components` in the devfile.

wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
```diff
@@ -162,5 +162,14 @@
 
   public static final String NO_ENVIRONMENT_RECIPE_TYPE = "no-environment";
 
+  /** Attribute to mark source of the container. */
+  public static final String CONTAINER_SOURCE_ATTRIBUTE = "source";
+
+  /** Mark containers applied to workspace with help recipe definition. */
+  public static final String RECIPE = "recipe";
```
sleshchenko: I think the constant name should be more self-explained, like `RECIPE_CONTAINER_SOURCE`. The same about `TOOL` constant

sleshchenko (approved):
LGTM

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/environment/InternalEnvironmentFactory.java
```diff
@@ -100,14 +103,17 @@ public T create(@Nullable final Environment sourceEnv)
...
+        InternalMachineConfig internalMachineConfig =
             new InternalMachineConfig(
                 installers,
                 normalizeServers(machineConfig.getServers()),
                 machineConfig.getEnv(),
                 machineConfig.getAttributes(),
-                machineConfig.getVolumes()));
+                machineConfig.getVolumes());
+
+        internalMachineConfig.getAttributes().put(CONTAINER_SOURCE_ATTRIBUTE, RECIPE);
```
sleshchenko: Please adapt `InternalEnvironmentFactoryTest` to check this case.

garagatyi: I think that in this case, you would not add an attribute to machines added by the environment factory implementation.

garagatyi (changes requested):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/environment/InternalEnvironmentFactory.java
sleshchenko: @garagatyi Nice catch!
@AndrienkoAleksandr Please consider reworking it in the following way:
```java
    T internalEnv = doCreate(recipe, machines, warnings);

    internalEnv.getMachines()
        .values()
        .forEach(m -> m.getAttributes().put(CONTAINER_SOURCE_ATTRIBUTE, RECIPE));

    return internalEnv;
```
And consider adding a test case for it.

AndrienkoAleksandr: Fixed

wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
AndrienkoAleksandr: Done

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/server/secure/jwtproxy/JwtProxyProvisioner.java
```diff
@@ -126,7 +128,9 @@ public JwtProxyProvisioner(
             MEMORY_LIMIT_ATTRIBUTE,
             Long.toString(memoryLimitLong),
             MEMORY_REQUEST_ATTRIBUTE,
-            Long.toString(memoryLimitLong));
+            Long.toString(memoryLimitLong),
+            CONTAINER_SOURCE_ATTRIBUTE,
+            TOOL_CONTAINER_SOURCE);
```
sleshchenko, line 133: It would be nice to have the corresponding check in `JwtProxyProvisionerTest`

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/wsplugins/KubernetesPluginsToolingApplier.java
```diff
@@ -194,6 +196,7 @@ private void addSidecar(
             .build();
 
     InternalMachineConfig machineConfig = machineResolver.resolve();
+    machineConfig.getAttributes().put(CONTAINER_SOURCE_ATTRIBUTE, TOOL_CONTAINER_SOURCE);
```
sleshchenko, line 199: It would be nice to have the corresponding test in `KubernetesPluginsToolingApplierTest`

sleshchenko (approved):
LGTM

sleshchenko:
ci-test

garagatyi (approved):

AndrienkoAleksandr:
ci-test

sleshchenko:
@AndrienkoAleksandr `ci-test` job failed because there is file with broken format `/home/codenvy/workspace/che-pullrequests-test-ocp/wsmaster/che-core-api-workspace/src/test/java/org/eclipse/che/api/workspace/server/spi/environment/InternalEnvironmentFactoryTest.java`. Please execute `mvn sortpom:sort license:format fmt:format`

ashumilova:
ci-test

AndrienkoAleksandr:
@eclipse/eclipse-che-qa Please verify the last test-report: https://github.com/eclipse/che/pull/12386#issuecomment-458638296

musienko-maxim (approved):

Issue #12225, opened 2018-12-19
Refresh workspace activity during runtime restore on workspace master startup


skabashnyuk:
### Description
I've noticed the situation when we have a lot of restored workspaces on startup
of workspace master after OOM crash, however number of them are not reduced after default activity timeout.
I suspect that during previous crashes we lost the fact that workspace is active. 
The goal of this task is to tick workspace activity on restore so if no activity happened
from the workspace, during the timeout, we can easily suspend it.

metlos:
Closing as the PR is merged.

PR #12387, opened 2019-01-11
Make activity manager participate in recovering the workspaces during startup.

metlos:
### What does this PR do?

Makes the workspace activity manager participate on the workspace recovery during startup, making sure that the workspace expiration and activity corresponds to the true state of workspaces as detected in the infrastructure.

#12225

#### Release Notes
N/A, not user facing

### DIFF
--- a/wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/InmemoryWorkspaceActivityDao.java
+++ b/wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/InmemoryWorkspaceActivityDao.java
@@ -16,6 +16,7 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 import javax.inject.Singleton;
+import org.eclipse.che.api.core.ConflictException;
 import org.eclipse.che.api.core.Page;
 import org.eclipse.che.api.core.ServerException;
 import org.eclipse.che.api.core.model.workspace.WorkspaceStatus;
@@ -137,6 +138,15 @@ public class InmemoryWorkspaceActivityDao implements WorkspaceActivityDao {
   @Override
   public void removeActivity(String workspaceId) throws ServerException {
     workspaceActivities.remove(workspaceId);
+  }
+
+  @Override
+  public void createActivity(WorkspaceActivity activity) throws ConflictException {
+    if (workspaceActivities.containsKey(activity.getWorkspaceId())) {
+      throw new ConflictException("Already exists.");
+    } else {
+      workspaceActivities.put(activity.getWorkspaceId(), activity);
+    }
   }
 
   private boolean isGreater(Long value, long threshold) {
--- a/wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/WorkspaceActivityChecker.java
+++ b/wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/WorkspaceActivityChecker.java
@@ -33,8 +33,8 @@ import org.slf4j.LoggerFactory;
 
 /**
  * Is in charge of checking the validity of the workspace activity records. The sole important
- * method is {@link #validate()} which is run on a schedule to periodically check the validity
- * of the records and report the potential error conditions.
+ * method is {@link #validate()} which is run on a schedule to periodically check the validity of
+ * the records and report the potential error conditions.
  *
  * @author Lukas Krejci
  */
--- a/wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/WorkspaceActivityChecker.java
+++ b/wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/WorkspaceActivityChecker.java
@@ -86,6 +86,11 @@ public class WorkspaceActivityChecker {
   void validate() {
     try {
       stopAllExpired();
+    } catch (ServerException e) {
+      LOG.error(e.getLocalizedMessage(), e);
+    }
+
+    try {
       checkActivityRecordValidity();
     } catch (ServerException e) {
       LOG.error(e.getLocalizedMessage(), e);
--- a/wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/WorkspaceActivityChecker.java
+++ b/wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/WorkspaceActivityChecker.java
@@ -1,0 +1,412 @@
+/*
+ * Copyright (c) 2012-2018 Red Hat, Inc.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.api.workspace.activity;
+
+import static java.util.Collections.singletonMap;
+import static org.eclipse.che.api.workspace.shared.Constants.WORKSPACE_STOPPED_BY;
+import static org.eclipse.che.api.workspace.shared.Constants.WORKSPACE_STOP_REASON;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+import java.time.Clock;
+import org.eclipse.che.api.core.ConflictException;
+import org.eclipse.che.api.core.NotFoundException;
+import org.eclipse.che.api.core.ServerException;
+import org.eclipse.che.api.core.model.workspace.Workspace;
+import org.eclipse.che.api.core.model.workspace.WorkspaceStatus;
+import org.eclipse.che.api.workspace.server.WorkspaceManager;
+import org.eclipse.che.api.workspace.server.WorkspaceRuntimes;
+import org.eclipse.che.api.workspace.shared.Constants;
+import org.eclipse.che.commons.schedule.ScheduleDelay;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Is in charge of checking the validity of the workspace activity records. The sole important
+ * method is {@link #validate()} which is run on a schedule to periodically check the validity
+ * of the records and report the potential error conditions.
+ *
+ * @author Lukas Krejci
+ */
+@Singleton
+public class WorkspaceActivityChecker {
+
+  private static final Logger LOG = LoggerFactory.getLogger(WorkspaceActivityChecker.class);
+  private static final String ACTIVITY_CHECKER = "activity-checker";
+
+  private final WorkspaceActivityDao activityDao;
+  private final WorkspaceManager workspaceManager;
+  private final WorkspaceRuntimes workspaceRuntimes;
+  private final WorkspaceActivityManager workspaceActivityManager;
+  private final Clock clock;
+
+  @Inject
+  public WorkspaceActivityChecker(
+      WorkspaceActivityDao activityDao,
+      WorkspaceManager workspaceManager,
+      WorkspaceRuntimes workspaceRuntimes,
+      WorkspaceActivityManager workspaceActivityManager) {
+    this(
+        activityDao,
+        workspaceManager,
+        workspaceRuntimes,
+        workspaceActivityManager,
+        Clock.systemDefaultZone());
+  }
+
+  @VisibleForTesting
+  WorkspaceActivityChecker(
+      WorkspaceActivityDao activityDao,
+      WorkspaceManager workspaceManager,
+      WorkspaceRuntimes workspaceRuntimes,
+      WorkspaceActivityManager workspaceActivityManager,
+      Clock clock) {
+
+    this.activityDao = activityDao;
+    this.workspaceManager = workspaceManager;
+    this.workspaceRuntimes = workspaceRuntimes;
+    this.workspaceActivityManager = workspaceActivityManager;
+    this.clock = clock;
+  }
+
+  @ScheduleDelay(
+      initialDelayParameterName = "che.workspace.activity_check_scheduler_delay_s",
+      delayParameterName = "che.workspace.activity_check_scheduler_period_s")
+  @VisibleForTesting
+  void validate() {
+    try {
+      stopAllExpired();
+      checkActivityRecordValidity();
+    } catch (ServerException e) {
+      LOG.error(e.getLocalizedMessage(), e);
+    }
+  }
+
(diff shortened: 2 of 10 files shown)


assembly/assembly-wsmaster-war/src/main/java/org/eclipse/che/api/deploy/WsMasterModule.java
```diff
@@ -292,6 +293,9 @@ protected void configure() {
       install(new org.eclipse.che.core.metrics.MetricsModule());
       install(new WsMasterMetricsModule());
     }
+
+    // now that everything is installed, we can recover the state of the workspaces
+    getProvider(WorkspaceManager.class).get().recover();
```
sleshchenko: Could you please provide a bit context why it is needed?
As far as I see you did not remove recovering on `PostConstruct` as well. So, now recovering it done twice, right?

metlos: Ah, that's correct, the `recover()` in `@PostConstruct` should have been removed.

metlos:
ci-test

wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/WorkspaceActivityManager.java
```diff
@@ -167,12 +171,82 @@ protected long getIdleTimeout(String wsId) {
...
+        if (idleTimeout == 0) {
+          // expiry not used at all...
+          continue;
+        }
+
+        long expectedExpiryTime =
+            valueOrDefault(activity.getLastRunning(), System.currentTimeMillis()) + idleTimeout;
+        long actualExpiryTime = valueOrDefault(activity.getExpiration(), 0);
+
+        if (actualExpiryTime != expectedExpiryTime) {
```
sleshchenko: Sorry, I don't understand this check. 

> expectedExpiryTime = valueOrDefault(activity.getLastRunning(), System.currentTimeMillis()) + idleTimeout;

I think it does not make sense to check  `System.currentTimeMillis() + idleTimeout` if `getLastRunning` is absent. 
What is the probability that expires time was set in the same millisecond?

Also, I think that that `actualExpiryTime` should not equal `activity.getLastRunning() + idleTimeout`. Because `getLastRunning` is timestamp when workspace was started, right?
If yes, then a user may use this workspace and expiration time will be updated to like `activity.getLastRunning() + 2*idleTimeout` and it's a valid case.

I mean this statement will true always when user uses workspace more than idleTimeout. I might be wrong.

wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/WorkspaceActivityManager.java
```diff
@@ -167,12 +171,82 @@ protected long getIdleTimeout(String wsId) {
...
+        } catch (ConflictException e) {
+          LOG.debug(
+              "Activity record created while we were trying to rectify its absence for a"
+                  + " running workspace {}.",
+              runningWsId);
+        }
+      } else {
+        long idleTimeout = getIdleTimeout(runningWsId);
+        if (idleTimeout == 0) {
+          // expiry not used at all...
```
sleshchenko: Why do you use `...`? Is it an incomplete sentence or you want to keep the intrigue?

wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/WorkspaceActivityManager.java
metlos: Well, you're right that I missed the fact that the expiry period is updated remotely from the clients on user activity, that's really stupid of me.

On the other hand, the code as is atm deals with the case where the expiration timestamp is not set when it should have been. For that, we need a value to set the expiration period to, and IMHO current time + the timeout is a reasonable value to use. It will be updated if there is any user activity on the workspace. This will also happen when the schedule exactly coincides with the workspace transitioning to RUNNING, but no harm should be done here - the two expiration times will be almost identical depending on which transaction commits last.

As for the `!=` - I agree that rarely we are going to see that the first time the expiry is set, it is going to be in the exact millisecond. The reason why I used `!=` here is to be able to settle on a value eventually. The silly me wanted to account for some "rogue" user setting the expiry to some distant future value, which is something that actually can't happen.

It is also obviously wrong, because it doesn't account for the updates due to the user activity - I should have used `<` instead - the expiry time should never be less than the time the workspace started + the timeout.

If it is, then the expiry must not have been set when the workspace started, which is wrong.

So IMHO, it is enough to change the code to use `actualExpiryTime < expectedExpiryTime` and it should work correctly. Also I should change the log level of the subsequent message to at least INFO or maybe even warning, because it really is an error condition that we're correcting there.

wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/WorkspaceActivityManager.java
metlos: It's just me - never want to make definitive statements ;)

sleshchenko:
Please consider adding unit tests for added functionality

wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/WorkspaceActivityManager.java
```diff
@@ -165,14 +191,282 @@ protected long getIdleTimeout(String wsId) {
...
+            maxOf(
+                activity.getCreated(),
+                activity.getLastStarting(),
+                activity.getLastRunning(),
+                activity.getLastStopping(),
+                activity.getLastStopped());
+
+        if (noLastRunningTime) {
+          rectifyNoLastRunningTime(runningWsId, activity, now, lastKnownActivity);
+        } else if (lastKnownActivity != null && lastKnownActivity > activity.getLastRunning()) {
```
metlos: This is a wrong check.

metlos: Actually, no, it is correct :smile: We're talking about a workspace that is running - i.e. it is neither starting, stopping, nor stopped according to what the runtime thinks.

wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/InmemoryWorkspaceActivityDao.java
```diff
@@ -139,6 +140,15 @@ public void removeActivity(String workspaceId) throws ServerException {
     workspaceActivities.remove(workspaceId);
   }
 
+  @Override
+  public void createActivity(WorkspaceActivity activity) throws ConflictException {
+    if (workspaceActivities.containsKey(activity.getWorkspaceId())) {
+      throw new ConflictException("Already exists.");
```
sleshchenko, line 146: Could you please add more details in error message, like
`Workspace activity record for workspace with id `%s` already exists`

wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/WorkspaceActivityManager.java
```diff
@@ -165,14 +191,276 @@ protected long getIdleTimeout(String wsId) {
   @ScheduleDelay(
       initialDelayParameterName = "che.workspace.activity_check_scheduler_delay_s",
       delayParameterName = "che.workspace.activity_check_scheduler_period_s")
-  private void invalidate() {
+  @VisibleForTesting
+  void validate() {
     try {
-      activityDao.findExpired(System.currentTimeMillis()).forEach(this::stopExpired);
+      stopAllExpired();
+      checkActivityRecordValidity();
```
sleshchenko: So many lines of code you added in `WorkspaceActivityManager`. Please consider moving it to a dedicated class.

wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/WorkspaceActivityManager.java
```diff
@@ -165,14 +191,276 @@ protected long getIdleTimeout(String wsId) {
...
     } catch (ServerException e) {
       LOG.error(e.getLocalizedMessage(), e);
     }
   }
 
+  private void stopAllExpired() throws ServerException {
+    activityDao.findExpired(clock.millis()).forEach(this::stopExpired);
+  }
+
+  private void checkActivityRecordValidity() throws ServerException {
```
sleshchenko: because of meaningful and useful comments this method is huge (~100 lines). Please consider using separate private methods, then it might looks like the following
```
private void checkActivityRecordValidity() throws ServerException {
  ...
  WorkspaceActivity activity = activityDao.findActivity(runningWsId);
  if (activity == null) {
    createMissingActivityRecord(...);
  } else {
    checkAndFixCreatedTime(activity);
    checkAndFixLastRunningTime(activity);
    checkAndFixExpirationTime(activity);
  }
} 

/**
* Makes sure activity record as at least 1 activity record - if it lacks last_running,
* we look at last_starting and if even that is not there we make sure there is at least
* created_time - the last_* are not mandatory, but the created_time should be present
*/
private void checkAndFixCreatedTime(...) {
   ...
}
```

Sorry, I did not fully analyze `checkActivityRecordValidity` method and I assume that it may not be so easy and nice looking as I described but please try.

wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/InmemoryWorkspaceActivityDao.java
metlos: As far as I looked `InmemoryWorkspaceActivityDao` is actually not used. I just didn't want to remove it as part of this PR ;)

wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/WorkspaceActivityManager.java
metlos: Done (to a large degree).

metlos:
ci-test

skabashnyuk (approved):

wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/WorkspaceActivityChecker.java
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+    String wsId = activity.getWorkspaceId();
+
+    // define the error message upfront to make it easier to follow the actual logic
+    final String noActivityFoundWhileHandlingExpiration =
+        "Found no expiration time on workspace {}. No prior activity was found on the  workspace."
+            + " To restore the normal function, the expiration time has been set to {}.";
+    final String noExpirationWithoutLastRunning =
+        "Found no expiration time on workspace {} and no  record of the last time it started. The"
```
sleshchenko: Double space after `... workspace {} and no`

sleshchenko (approved):
Generally LGTM. Please take a look my inline comments

wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/WorkspaceActivityChecker.java
```diff
@@ -0,0 +1,412 @@
...
+  @ScheduleDelay(
+      initialDelayParameterName = "che.workspace.activity_check_scheduler_delay_s",
+      delayParameterName = "che.workspace.activity_check_scheduler_period_s")
+  @VisibleForTesting
+  void validate() {
+    try {
+      stopAllExpired();
+      checkActivityRecordValidity();
+    } catch (ServerException e) {
+      LOG.error(e.getLocalizedMessage(), e);
```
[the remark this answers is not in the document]
metlos: Well, I suppose we could spit that. I just considered both a part of maintaining the activity records valid.

wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/WorkspaceActivityChecker.java
```diff
@@ -0,0 +1,412 @@
...
+  private void createMissingActivityRecord(String runningWsId, long idleTimeout)
+      throws ServerException {
+    LOG.warn(
+        "Found a running workspace {} without any activity record. This shouldn't really happen"
+            + " but is being rectified by adding a new activity record for it.",
+        runningWsId);
+    try {
+      Workspace workspace = workspaceManager.getWorkspace(runningWsId);
+      long createdTime =
+          Long.parseLong(workspace.getAttributes().get(Constants.CREATED_ATTRIBUTE_NAME));
```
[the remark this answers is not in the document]
metlos: Considering https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java#L494 I thought the created attribute should never be null and always be a long.

wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/WorkspaceActivityChecker.java
```diff
@@ -0,0 +1,412 @@
...
+    }
+
+    // there was a running time before
+    return false;
+  }
+
+  private void rectifyNoLastRunningTime(
+      String runningWsId, WorkspaceActivity activity, long now, Long lastKnownActivity)
+      throws ServerException {
+    // k, so we don't have the information about when the workspace was last started here.
```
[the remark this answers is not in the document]
metlos: "k" is ok, k? Will change it :smile:

wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/WorkspaceActivityChecker.java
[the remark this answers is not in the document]
sleshchenko: Actually, it should not be null but there is no secure mechanism to guarantee that this attribute is not removed from workspace attributes, like any user is able to remove it by purpose or client like dashboard may remove it by mistake.

wsmaster/che-core-api-workspace-activity/src/main/java/org/eclipse/che/api/workspace/activity/WorkspaceActivityChecker.java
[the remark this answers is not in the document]
sleshchenko: k, o problem

metlos:
ci-test

PR #12404, opened 2019-01-12
[CORS] Correctly determine server from which IDE was loaded

### LINKED ISSUE #12403: Gwt ide in sidecar is not working due to CORS configuration on ws-agent
### Description
Ws-agent by default set cors filter configuration to allow request only from che.api. But if gwt ide running as sidecar it loads IDE not from the master but from a different server running in the dedicated container.
We need to change default cors logic to be more similar to workspace loader. https://github.com/eclipse/che/blob/master/workspace-loader/src/index.ts#L349-L361
It should detect ide location from servers configuration or fallback to ide attribute of workspace configuration.

output

### Reproduction Steps
- Enable GWT Editor

- Start workspace
- Try to create project

- In console you can see CORS errors

skabashnyuk:
### What does this PR do?
Correctly determine server from which IDE was loaded for the GWT IDE running as a sidecar
Additional changes:
- Update version of tomcat. Now it uses latest Che tomcat 8.5.x
- ws agent CORS providers moved from core/che-core-api-core wsagent/che-wsagent-core/
- added test-logback.xml to reduce log output

Fixes https://github.com/eclipse/che/issues/12403

#### Release Notes
n/a

#### Docs PR
n/a


skabashnyuk:
ci-test

benoitf (approved):

wsagent/che-wsagent-core/src/main/java/org/eclipse/che/wsagent/server/CheWsAgentCorsAllowedOriginsProvider.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+  }
+
+  @Override
+  public String get() {
+    LOG.debug("allowedOrigins {} ", allowedOrigins);
+    return allowedOrigins;
+  }
+
+  private String getIdeUrl(WorkspaceDto workspaceDto) {
```
mshaposhnik, line 89: Maybe it's  easier to just get the link by [this](https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java#L24 ) rel from WS links ?

skabashnyuk: I'm using this link here https://github.com/eclipse/che/pull/12404/files#diff-11cbb1d2057d9e410a50d3a7c532cc2bR71 in case if there is no server declared with ide attribute aka Che6 mode for GWT Ide

mshaposhnik: So can it be different between server url and link in workspace?

skabashnyuk: yes

wsagent/che-wsagent-core/src/main/java/org/eclipse/che/wsagent/server/CheWsAgentCorsAllowedOriginsProvider.java
```diff
@@ -0,0 +1,99 @@
...
+
+  @Inject
+  public CheWsAgentCorsAllowedOriginsProvider(
+      @Named("che.api") String apiEndpoint,
+      @Nullable @Named("che.wsagent.cors.allowed_origins") String allowedOrigins,
+      HttpJsonRequestFactory httpJsonRequestFactory)
+      throws ApiException, IOException {
+    this.apiEndpoint = apiEndpoint;
+
+    if (allowedOrigins == null) {
```
sleshchenko: such heavy logic for the constructor. What if rewrite it in the following way:
```java
public CheWsAgentCorsAllowedOriginsProvider(@Named("env.CHE_WORKSPACE_ID") String wsId, ....) {
....
  this.allowedOrigins = evaluateOrigins(allowedOrigins, wsId);
}

private String evaluateOrigins(....) {
  if (allowedOrigins != null) {
    return allowedOrigins
  } 

  WorkpsaceDto workpsace = fetchWorkpace(workspaceId);
  Optional<String> ideServerURL = findIdeServerURL(workpsace);

  if (ideServerURL.isPresent()) {
   return UriBuilder.fromUri(ideServerURL.get()).replacePath(null).build().toString();
  } 

  return UriBuilder.fromUri(workspaceDto.getLinks().getOrDefault("ide", apiEndpoint))
               .replacePath(null)
               .build()
               .toString();
}
```
WDYT?

wsagent/che-wsagent-core/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,27 @@
...
+-->
+<configuration>
+
+    <appender name="stdout" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>%-41(%date[%.15thread]) %-45([%-5level] [%.30logger{30} %L]) - %msg%n</pattern>
+        </encoder>
+    </appender>
+
+    <root level="INFO">
```
sleshchenko, line 23: error log level may reduce build time a bit

sleshchenko (commented):
It would be nice to separate updating of tomcat to separate commit and then merge this PR with `merge` option to have different commits for tomcat updating and fix of determining server.

mkuznyetsov (approved):

skabashnyuk:
ci-test

sleshchenko (approved):
Looks better now

skabashnyuk:
Ci-build

PR #12405, opened 2019-01-13
Add @SuppressWarnings("serial").

monaka:
### What does this PR do?

Just reducing warnings. Not critical.

### DIFF
--- a/core/che-core-db/src/main/java/org/eclipse/che/core/db/jpa/DuplicateKeyException.java
+++ b/core/che-core-db/src/main/java/org/eclipse/che/core/db/jpa/DuplicateKeyException.java
@@ -19,6 +19,7 @@ import org.eclipse.che.core.db.DBErrorCode;
  * @author Yevhenii Voevodin
  * @see DBErrorCode#DUPLICATE_KEY
  */
+@SuppressWarnings("serial")
 public class DuplicateKeyException extends DetailedRollbackException {
 
   public DuplicateKeyException(String message, Throwable cause) {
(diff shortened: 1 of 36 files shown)


core/che-core-db/src/main/java/org/eclipse/che/core/db/jpa/DuplicateKeyException.java
```diff
@@ -19,6 +19,7 @@
  * @author Yevhenii Voevodin
  * @see DBErrorCode#DUPLICATE_KEY
  */
+@SuppressWarnings("serial")
```
sleshchenko, line 22: The only warning that I found during maven build in `/core/che-core-db` module is 
```
[INFO] Compiling 27 source files to /home/serg/projects/che/core/che-core-db/target/classes                                                                                                                       
/home/serg/projects/che/core/che-core-db/src/main/java/org/eclipse/che/core/db/jpa/guice/GuiceJpaInitializer.java:33: warning: [MissingOverride] init implements method in JpaInitializer; expected @Override     
  public void init() {                                                                                                                                                                                            
              ^                                                                                                                                                                                                   
    (see http://errorprone.info/bugpattern/MissingOverride)
  Did you mean '@Override public void init() {'?
Note: /home/serg/projects/che/core/che-core-db/src/main/java/org/eclipse/che/core/db/jpa/eclipselink/GuiceEntityListenerInjectionManager.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
1 warning
```
Can you elaborate which warnings you suppress? What the cause of them? Maybe there is an ability to fix cause instead of suppressing warning?

monaka: Warnings are shown on Eclipse/JDT (not Che).
I can suppress that warnings by settings. But I guess many IDE (except Che) will show this warning by default.

The reason why `@SuppressWarnings("serial")` should be added, it will be some helps to clarify whether "the author of the class decided not to care about serialize" or "he/she just forgot to add `serialVersionUID`."

monaka:
ci-build

musienko-maxim:

Issue #12452, opened 2019-01-17
Fail to build theia dockerimage module


sleshchenko:
### Description
The following CI test jobs fail because of error in Theia, TheiaDev modules:
https://github.com/eclipse/che/pull/12335#issuecomment-454810864
https://github.com/eclipse/che/pull/12421#issuecomment-454985674
https://github.com/eclipse/che/pull/12335#issuecomment-455056826

### Reproduction Steps
Execute `mvn clean install -Pintegration` in che project or directly in [dockerfiles/theia](https://github.com/eclipse/che/tree/master/dockerfiles/theia)
**OS and version:**    

**Diagnostics:** 
<details>
<summary>One of the job logs fragment</summary>

```
$ yarn run clean && yarn run build
$ rimraf lib
$ yarn run format-code && yarn run compile && yarn run compileWorker
$ tsfmt -r
$ tsc
src/node/che-task-service.ts(45,13): error TS2322: Type 'CheTask' is not assignable to type 'Task'.
  Types of property 'getRuntimeInfo' are incompatible.
    Type '() => Promise<TaskInfo>' is not assignable to type '() => TaskInfo'.
src/node/che-task-service.ts(84,11): error TS2416: Property 'getRuntimeInfo' in type 'CheTask' is not assignable to the same property in base type 'Task'.
  Type '() => Promise<TaskInfo>' is not assignable to type '() => TaskInfo'.
    Type 'Promise<TaskInfo>' is not assignable to type 'TaskInfo'.
      Property 'taskId' is missing in type 'Promise<TaskInfo>'.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

    at Promise.all.then.arr (/home/theia-dev/theia-source-code/node_modules/lerna/node_modules/execa/index.js:236:11)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:189:7)
  code: 1,
  killed: false,
[...]

PR #12453, opened 2019-01-17
Add patch to fix Theia build

vitaliy-guliy:
Signed-off-by: Vitaliy Gulyy <vgulyy@redhat.com>

### What does this PR do?

Adds a patch file to be applied when building Theia image.

https://github.com/eclipse/che/issues/12452


benoitf:
I would rather use tagged version of che-theia components rather than patching an official tag.

vinokurig (approved):

benoitf:
I still won't approve. I didn't approve the PR that cause the troubles so I won't approve ugly hacks as well.
And yesterday I said you could simply use a tagged version of che-theia as well.
now, up to you.

vitaliy-guliy:
@benoitf it was decided revert the PR which breaks the build

Issue #12454, opened 2019-01-17
Release CHE 6.17.0 failed


riuvshin:
release failed on `Che TypeScript DTO Generate and Publish`
the problem is that it publishing to npm registry on `release:prepare` and `release:perform`
while it should push only on perform.

riuvshin:
fixed

PR #12455, opened 2019-01-17
Check the latest version of @eclipse-che/api

evidolob:
And publish only when it's don't match

### What does this PR do?
Fix publishing npm package with che dto d.ts file, during che release

#12454

### DIFF
--- a/typescript-dto/publish.sh
+++ b/typescript-dto/publish.sh
@@ -1,10 +1,22 @@
 #!/usr/bin/env sh
+#
+# Copyright (c) 2019 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# See: https://sipb.mit.edu/doc/safe-shell/
+
+set -e
+set -u
 
 #get the latest version of @eclipse-che/api
 api_version=$(yarn -s info @eclipse-che/api version)
 
 #publish only if latest version doesn't match current version
-if [ $api_version != $1 ];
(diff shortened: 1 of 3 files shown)


typescript-dto/publish.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env sh
```
benoitf, line 1: missing copyright headers

typescript-dto/publish.sh
```diff
@@ -0,0 +1,10 @@
...
+
+#get the latest version of @eclipse-che/api
+api_version=$(yarn -s info @eclipse-che/api version)
+
+#publish only if latest version doesn't match current version
+if [ $api_version != $1 ];
```
benoitf: use of double quotes (https://www.shellcheck.net/)

typescript-dto/publish.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env sh
+
+#get the latest version of @eclipse-che/api
+api_version=$(yarn -s info @eclipse-che/api version)
```
benoitf, line 16: we should have

```
set -e
set -u
```

 as well

should we first check that yarn tool is installed as well ?

riuvshin (approved):

benoitf (approved):

typescript-dto/publish.sh
evidolob: I added 
```
set -e
set -u
```
No, as that script called from docker which already have `yarn` https://github.com/eclipse/che/blob/b818fccea3496f85579ad1affaf880674bb718f1/typescript-dto/Dockerfile#L25

typescript-dto/publish.sh
evidolob: Done

Issue #11530, opened 2018-10-08
Support Route object in OpenShift recipe


sleshchenko:
### Description
An application on OpenShift usually has at least one Route to make application accessible outside from cluster. 
Now Che Server ignores Routes in the original recipe and requires a user to declare Che Machine for container and Che Server inside it for exposing container port. But it is not really needed and it would be better from UX perspective to respect Routes object from the recipe.

amisevsk:
Support added in https://github.com/eclipse/che/pull/12486

Issue #11508, opened 2018-10-05
Support Secret object in Kubernetes/OpenShift recipe


sleshchenko:
### Description

A using of Secret is a very common way to store application credentials. But Eclipse Che ignores Secrets which are specified in Kubernetes/OpenShift recipe which leads to workspace start failing if there is any Pod/ConfigMap which references it.
It is needed to support `Secret` object in Kubernetes/OpenShift recipe.
The only one issue that should be decided - should we create Secret with an original name, or rename it to a random one (with fixing references in Pods, ConfigMaps) to avoid collisions if two or more workspaces are run in the same namespace/project.

amisevsk:
Added in https://github.com/eclipse/che/pull/12486

PR #12486, opened 2019-01-22
Add recipe support for Kubernetes secrets / OpenShift routes

amisevsk:
### What does this PR do?
Add support for Secrets and Routes in OpenShift/Kubernetes recipes. 

- Routes will only work in OpenShift recipes, as they are not a Kubernetes object. 
- There is some minimal user dashboard validation (metadata + fields on secrets only) since all validation is done through the Kubernetes side (and so it doesn't make sense to mention routes there). Since this is more involved and will likely be harder to read, it should be done in a separate PR / issue.
   - Additionally, more validation for secrets could be beneficial, as e.g. `data` values must be base64 encoded.
- Routes are renamed by the existing mechanism in Che. This should not affect behavior.
- Secrets are left unmodified. Renaming can be added later if we find it is necessary (I don't really see a use case, personally).

This PR is broken up into 3 commits:
- The first commit adds support for routes on the infra side
- The second commit adds support for secrets on the infra side.
- The third commit adds the necessary framework for UD validation.

<details>
<summary>A sample recipe for testing:</summary>

```yaml
kind: List
apiVersion: v1
items:
  - 
    apiVersion: v1
    kind: Pod
    metadata:
      name: ws
    spec:
      containers:
        - 
          image: 'eclipse/che-dev:nightly'
          name: dev
          resources:
            limits:
              memory: 512Mi
          volumeMounts:
          - name: test
            mountPath: "/test"
            readOnly: true
      volumes:
      - name: test
        secret:
          secretName: test-secret
  -
    apiVersion: v1
    kind: Secret
    metadata:
      name: test-secret
    data:
      test1: dGVzdDE=
    stringData:
      test2: test2
  -
    apiVersion: v1
    kind: Route
    metadata:
      name: test-route
    spec:
      host: test.192.168.42.162.nip.io
      port:
        targetPort: http
      to:
        kind: Service
        name: che-host

```
</details>

https://github.com/eclipse/che/issues/11508, https://github.com/eclipse/che/issues/11530, https://github.com/eclipse/che/issues/12522

#### Release Notes
Add support for secrets and routes in Kubernetes/OpenShift recipes

#### Docs PR
TODO


garagatyi (approved):
Looks good! Your commits have leftovers of conflict resolution error. Please, use squash when you merge.

ibuziuk:
@amisevsk could you please provide a recipe for testing?

garagatyi:
@amisevsk there are no leftovers in the final diff. But since you prepared PR with the commit-by-commit approach (thanks for that) I found it in one of the commits and found that it was fixed in the following one

ibuziuk:
ci-test

ibuziuk (approved):
@amisevsk great job! tested on minishift and routes / secrets are created / deleted with workspace just fine

ibuziuk:
@ashumilova could you please take a look at UD part of the PR, before we proceed with merging ?

ibuziuk:
@vkuznyetsov @rhopp could QA provide their approval based on the test report. There are 6 test failures, but I do not think those are related to the PR

infrastructures/openshift/src/test/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenShiftEnvironmentFactoryTest.java
```diff
@@ -158,39 +154,34 @@ public void shouldCreateOpenShiftEnvironmentWithPVCsFromRecipe() throws Exceptio
...
   }
 
-  @Test
-  public void ignoreRoutesWhenRecipeContainsThem() throws Exception {
-    final List<HasMetadata> objects =
-        asList(
-            new RouteBuilder().withNewMetadata().withName("route1").endMetadata().build(),
-            new RouteBuilder().withNewMetadata().withName("route2").endMetadata().build());
-    when(parsedList.getItems()).thenReturn(objects);
+  public void addRoutesWhenRecipeContainsThem() throws Exception {
```
sleshchenko, line 157: Looks like `@Test` is missing and this test method won't be run

sleshchenko (commented):
**1**
During manual testing, I successfully created a stack from Kubernetes recipe with content you provide. But then I failed to create a workspace from such stack with the following error 
screenshot_20190125_101248
1. Can we improve validation on Dashboard in this PR or in another issue?
2. Please improve the validation message in EnvFactories to contain object kind and name instead of ObjectMeta json.

**2** Then I created a stack using OpenShift recipe you provided. I edited config of created workspace and changed `Route` kind to `Route1`, as result I get 500 exception from workspace API
screenshot_20190125_101926
screenshot_20190125_102059
The same question about improving UX. It would be nice to improve dashboard, API to handle such cases

**3**
A workspace created from your provided recipe is started successfully. But what I see here there is no any integrity validation, recipe contains Route that references service that is not present in recipe. Do you think is there use case for such workspace? Or maybe we should add integrity validation for Routes, like we have for PVC (Pods can not reference PVCs which are not present in the recipe). See https://github.com/eclipse/che/blob/master/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/environment/KubernetesEnvironmentPodsValidator.java#L59

**4**
The same question about recipe integrity validation but about secrets. Should we allow using Secret sources volumes that are not defined in the recipe? 
It's just a question since I don't know what the right approach and just wanted to raise discussion.

amisevsk:
@sleshchenko Regarding the big issues in your review: the short of it is that Route validation is a bigger task than this PR (in my opinion). All validation goes through [`kubernetes-machine-recipe-parser.ts`](https://github.com/eclipse/che/pull/12486/files#diff-e7483d1658b9e7a96a50fc5c3b56e68cR147), and Routes are not Kubernetes objects. As a result:

1. While we should block creation of Kube stacks with Routes, [enabling the check](https://github.com/eclipse/che/pull/12486/files#diff-4fdedd239e20275b6cabbaad2a1b3134R90) (which is implemented [here](https://github.com/eclipse/che/pull/12486/files#diff-e7483d1658b9e7a96a50fc5c3b56e68cR18)) would block creation of OpenShift stacks as well
2. I agree that this should be improved -- I'll work on it. Should be relatively easy to implement on the infra side. Note that this will happen for any errors of this sort, not Routes specifically (thrown [here](https://github.com/eclipse/che/blob/40131c332e3fbbb3e540fc466d4d15c3b834c28b/infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenShiftEnvironmentFactory.java#L102)-- WDYT about moving it to a new issue?
3. Will look into this before Monday.
4. I think there's a use case for secrets outside the recipe, because otherwise we require secrets used in workspace to be stored in a trivially decodable format in recipes. There's definitely a case in my mind where a user would want to create a secret more securely and still use it in a workspace.

infrastructures/openshift/src/test/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenShiftEnvironmentFactoryTest.java
amisevsk: Fixed -- good catch.

amisevsk:
Created an issue for point 2. above: https://github.com/eclipse/che/issues/12522

To add: validation on the dashboard side is currently not possible, and requires the more significant refactor to allow different validation paths for OpenShift vs. Kubernetes.

sleshchenko:
> 4. I think there's a use case for secrets outside the recipe, because otherwise we require secrets used in workspace to be stored in a trivially decodable format in recipes. There's definitely a case in my mind where a user would want to create a secret more securely and still use it in a workspace.

Makes sense for me, thanks for the explanation.

amisevsk:
@sleshchenko I've added a commit to fix some of the validation issues:
1. Avoid the internal server error when a recipe cannot be loaded (e.g. with `kind: Route1`) (Point 2 above)
2. Show message like `Found unknown object type in recipe -- name: 'test-route', kind: 'Route'` when unsupported object is in recipe (Point 1 above)
3. Throw validation exception if route refers to service that is not in recipe.

Anything I missed?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/environment/KubernetesEnvironmentFactory.java
```diff
@@ -97,16 +98,28 @@ protected KubernetesEnvironment doCreate(
...
     }
 
-    final KubernetesList list =
-        clientFactory.create().lists().load(new ByteArrayInputStream(content.getBytes())).get();
+    final KubernetesList list;
+    try {
+      list =
+          clientFactory.create().lists().load(new ByteArrayInputStream(content.getBytes())).get();
+    } catch (KubernetesClientException e) {
+      String message = e.getCause() == null ? e.getMessage() : e.getCause().getLocalizedMessage();
```
sleshchenko: `e.getMessage()` but `e.getCause().getLocalizedMessage()` <- Did you make it on the purpose?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/environment/KubernetesEnvironmentFactory.java
```diff
@@ -97,16 +98,28 @@ protected KubernetesEnvironment doCreate(
...
+      if (message.contains("\n")) {
+        // Clean up message if it comes from JsonMappingException. Format is e.g.
+        // `No resource type found for:v1#Route1\n at [...]`
+        message = message.split("\\n")[0];
```
sleshchenko: you use the only first element, so you can limit elements

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenShiftEnvironmentFactory.java
```diff
@@ -98,16 +97,27 @@ protected OpenShiftEnvironment doCreate(
...
+    final KubernetesList list;
+    try {
+      list =
+          clientFactory.create().lists().load(new ByteArrayInputStream(content.getBytes())).get();
+    } catch (KubernetesClientException e) {
+      String message = e.getCause() == null ? e.getMessage() : e.getCause().getLocalizedMessage();
+      if (message.contains("\n")) {
+        // Clean up message if it comes from JsonMappingException. Format is e.g.
+        // `No resource type found for:v1#Route1\n at [...]`
+        message = message.split("\\n")[0];
```
sleshchenko: The same as in `KubernetesEnvironmentFactory`

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenShiftEnvironmentValidator.java
```diff
@@ -0,0 +1,47 @@
+package org.eclipse.che.workspace.infrastructure.openshift.environment;
```
sleshchenko, line 12: please execute `mvn license:format` to add missing license

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenShiftEnvironmentValidator.java
```diff
@@ -0,0 +1,47 @@
...
+  public OpenShiftEnvironmentValidator(KubernetesEnvironmentPodsValidator podsValidator) {
+    super(podsValidator);
+  }
+
+  public void validate(OpenShiftEnvironment env) throws ValidationException {
+    super.validate(env);
+    validateRoutesMatchServices(env);
+  }
+
+  private void validateRoutesMatchServices(OpenShiftEnvironment env) throws ValidationException {
```
sleshchenko, line 41: A check may be a bit improved by checking Route target port.

sleshchenko: Please consider adding unit tests for added test case

sleshchenko (approved):
Great job :+1: 

Please take a look my minor inline comments

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/environment/KubernetesEnvironmentFactory.java
amisevsk: No, was an accident trying to find a decent way to avoid the newlines and messy cause. Fixed.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/environment/KubernetesEnvironmentFactory.java
amisevsk: Done.

infrastructures/openshift/src/main/java/org/eclipse/che/workspace/infrastructure/openshift/environment/OpenShiftEnvironmentFactory.java
amisevsk: Done.

ashumilova (approved):

ibuziuk:
ci-test

amisevsk:
PR Quality Review is a false positive, the unused field is necessary for injection.

amisevsk:
Rebased and squashed fixup commits.

Issue #12361, opened 2019-01-09
Introduce selector field for Kubernetes/OpenShift tools in Devfile


sleshchenko:
### Description
It was proposed to introduce `selector` field for Kubernetes/OpenShfit tools in Devfile.

See the corresponding discussions 
https://github.com/redhat-developer/devfile/pull/2#discussion_r243471108
https://github.com/redhat-developer/devfile/pull/3#discussion_r244739226

`selector` field is `Map<String, String>`. For workspace should be used only those objects from file that have all labels (`ObjectMeta#labels`) that are specified in `selector`.

It is needed to prepare the corresponding changes for Devfile JSON Spec and make factory flow filter objects before provisioning them to `WorkspaceConfig#recipe`.

Also, the following TODO should be fixed in scope of this issue https://github.com/eclipse/che/pull/12353#discussion_r246680494

PR #12488, opened 2019-01-22
Support of selector in devfile for recipe type (K8S/OS) tools

mshaposhnik:
### What does this PR do?
Allows to specify selectors for the recipe type tools in devfile, so only matched objects will appears in environment recipe.
The key changes:
- `DevfileEnvironmentFactory` moved from factory module into devfile;
- Introduced `RecipeFileContentProvider` interface, which implementation instance should be passed to DevFile API in order to fetch content of the local files;
- Parsing and filtering of source list in `DevfileEnvironmentFactory`

Behavior can be tested using the following factory to my repo:
http://<che_host>/f?url=https://github.com/mshaposhnik/spring-petclinic
it should pick-up only mysql items from recipe list

https://github.com/eclipse/che/issues/12361


wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileEnvironmentFactory.java
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.api.devfile.server;
+
+import static com.google.common.base.Strings.isNullOrEmpty;
+import static io.fabric8.kubernetes.client.utils.Serialization.asYaml;
```
skabashnyuk: does this methods worth to have kubernetes-client as a dependency?

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileEnvironmentFactory.java
```diff
@@ -0,0 +1,95 @@
...
+   * EnvironmentImpl} from it (including filtering of list items using selectors, if necessary). An
+   * {@link RecipeFileContentProvider} MUST be provided in order to fetch recipe content.
+   *
+   * @param recipeTool the recipe-type tool
+   * @param recipeFileContentProvider service-specific provider of recipe file content
+   * @return optional of constructed environment from recipe type tool
+   * @throws BadRequestException when there is no content provider for recipe-type tool
+   * @throws BadRequestException when recipe-type tool content is unreachable or empty
+   */
+  public Optional<EnvironmentImpl> createEnvironment(
```
skabashnyuk: Why we use Optional as a return value? As I can see you can get a valid EnvironmentImpl or exception.

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileEnvironmentFactory.java
```diff
@@ -0,0 +1,95 @@
...
+      throw new BadRequestException(
+          format(
+              "The local file '%s' defined in tool  '%s' is unreachable or empty.",
+              recipeTool.getLocal(), recipeTool.getName()));
+    }
+    try {
+      final KubernetesList list = unmarshal(recipeFileContent, KubernetesList.class);
+
+      if (recipeTool.getSelector() != null && !recipeTool.getSelector().isEmpty()) {
+        List<HasMetadata> itemsList = list.getItems();
```
skabashnyuk: maybe it's better to filter stream from ```List<HasMetadata>``` and create new list instead of removing items from existing?

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileConverter.java
```diff
@@ -32,20 +33,30 @@
 import java.util.Optional;
 import java.util.StringJoiner;
 import java.util.stream.Collectors;
+import javax.inject.Inject;
+import org.eclipse.che.api.core.BadRequestException;
```
skabashnyuk: Is this a good idea to use REST exceptions at this level?

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileEnvironmentFactory.java
```diff
@@ -0,0 +1,95 @@
...
+import static org.eclipse.che.api.devfile.server.Constants.KUBERNETES_TOOL_TYPE;
+import static org.eclipse.che.api.devfile.server.Constants.OPENSHIFT_TOOL_TYPE;
+
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.fabric8.kubernetes.api.model.KubernetesList;
+import io.fabric8.kubernetes.client.KubernetesClientException;
+import java.util.List;
+import java.util.Optional;
+import javax.inject.Singleton;
+import org.eclipse.che.api.core.BadRequestException;
```
skabashnyuk: Is this a good idea to use REST exceptions at this level?

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/urlfactory/URLFactoryBuilder.java
```diff
@@ -33,11 +32,9 @@
 import org.eclipse.che.api.devfile.server.DevfileManager;
```
skabashnyuk, line 32: Is this a good idea to use REST exceptions at this level?

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileEnvironmentFactory.java
mshaposhnik: Yes - together with `io.fabric8.kubernetes.client.utils.Serialization.unmarshal` if we don't want to be responsible of parsing to/from `KubernetesList` correctly.

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileEnvironmentFactory.java
mshaposhnik: Yes it's reduntant

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileEnvironmentFactory.java
mshaposhnik: `list.getItems()` returns copy of KubernetesList internal list, so no much difference here (ony filtering method)

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileEnvironmentFactory.java
```diff
@@ -0,0 +1,95 @@
...
+   * @return optional of constructed environment from recipe type tool
+   * @throws BadRequestException when there is no content provider for recipe-type tool
+   * @throws BadRequestException when recipe-type tool content is unreachable or empty
+   */
+  public Optional<EnvironmentImpl> createEnvironment(
+      Tool recipeTool, RecipeFileContentProvider recipeFileContentProvider)
+      throws BadRequestException {
+    final String type = recipeTool.getType();
+    if (!KUBERNETES_TOOL_TYPE.equals(type) && !OPENSHIFT_TOOL_TYPE.equals(type)) {
+      throw new BadRequestException("Environment cannot be created from such type of tool.");
```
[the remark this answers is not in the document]
mshaposhnik: Yes for this given case IllegalArgument looks logical

sleshchenko (commented):
Generally looks good to me.
Please notify when you will address my inline comment and I'll do a final review.

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/RecipeFileContentProvider.java
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Fetches content of file described in local field of recipe-type {@link
+ * org.eclipse.che.api.devfile.model.Tool}
+ *
+ * @author Max Shaposhnyk
+ */
+@FunctionalInterface
+public interface RecipeFileContentProvider {
+
+  String fetchContent(String localFileName);
```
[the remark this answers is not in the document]
mshaposhnik: Never heard that it should be a path. It's a devfile-relative filename IMO.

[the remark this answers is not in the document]
sleshchenko: >  It's a devfile-relative filename IMO.

Do you mean devfile-relative file path?

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileConverter.java
```diff
@@ -125,8 +136,15 @@ public WorkspaceConfigImpl devFileToWorkspaceConfig(Devfile devfile)
...
-          // this kind of tool ignored here since it contain only reference to tool configuration
-          // which should be resolved and provisioned into workspace config recipe separately.
+          try {
+            EnvironmentImpl environment =
+                devfileEnvironmentFactory.createEnvironment(tool, recipeFileContentProvider);
+            final String environmentName = tool.getName();
+            config.setDefaultEnv(environmentName);
+            config.setEnvironments(singletonMap(environmentName, environment));
+          } catch (IllegalArgumentException e) {
+            throw new DevfileFormatException(e.getMessage());
```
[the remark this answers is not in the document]
mshaposhnik: Hard to say certainly. This gonna happen if someone provide devfile with recipe tools to the bare devfile API, which has no content supplier. So in this case format error looks more or less logical for me.

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/RecipeFileContentProvider.java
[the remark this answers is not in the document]
sleshchenko: I may be wrong but for me filename is like `petclinic.yaml`, and `../app/petclinic.yaml` is file path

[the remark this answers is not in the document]
mshaposhnik: No - filename. It can be `petclinic.yaml` but not sure if we should support also to be `/foo/bar/petclinic.yaml` (btw it currelnty may work OK too :) )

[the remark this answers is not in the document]
mshaposhnik: `../app/petclinic.yaml` definitely will not work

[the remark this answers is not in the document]
sleshchenko: As far as I see it is root a repository-relative path. BTW Please add a clear description for this argument.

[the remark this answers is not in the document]
mshaposhnik: Ok expanded description

mshaposhnik:
ci-build

mshaposhnik:
@sleshchenko seems i'm done with all comments, pls review it whenever possible

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileEnvironmentFactory.java
```diff
@@ -0,0 +1,119 @@
...
+   * @throws IllegalArgumentException when wrong type tool is passed
+   * @throws IllegalArgumentException when there is no content provider for recipe-type tool
+   * @throws DevfileRecipeFormatException when recipe-type tool content is unreachable, empty or has
+   *     wrong format
+   */
+  public EnvironmentImpl createEnvironment(
+      Tool recipeTool, RecipeFileContentProvider recipeFileContentProvider)
+      throws DevfileRecipeFormatException {
+    final String type = recipeTool.getType();
+    if (!KUBERNETES_TOOL_TYPE.equals(type) && !OPENSHIFT_TOOL_TYPE.equals(type)) {
```
metlos, line 59: Wouldn't `tool.isKubernetesListSupported()` be cleaner?

mshaposhnik: it's little complex to implement since tool is an autogenerated pojo :)

wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json
```diff
@@ -4,6 +4,11 @@
       "id": "propertyList",
       "type": "object",
       "javaType": "java.util.Map<String, String>"
+    },
+    "selector" : {
+      "id": "selector",
+      "type": "object",
```
metlos, line 13: Do we want to allow a free-form object or rather something along the lines of:

```json
"type": "object",
"additionalProperties": {
  "type": "string"
}
```

i.e. an object with any keys that are only allowed to have string values?

wsmaster/che-core-api-devfile/src/test/java/org/eclipse/che/api/devfile/server/DevfileEnvironmentFactoryTest.java
```diff
@@ -0,0 +1,163 @@
...
+
+  @Test(
+      expectedExceptions = IllegalArgumentException.class,
+      expectedExceptionsMessageRegExp =
+          "Unable to process tool '"
+              + TOOL_NAME
+              + "' of type '"
+              + KUBERNETES_TOOL_TYPE
+              + "' since there is no content provider supplied.")
+  public void shouldThrowExceptionWhenRecipeToolIsPresentAndNoURLComposerGiven() throws Exception {
```
metlos: What's a URLComposer?

mshaposhnik: yep, outdated -- should be content provider now, fixing.

wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json
mshaposhnik: Well, that's interesting way to limit values type. Makes sense to add it. Tks.

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/RecipeFileContentProvider.java
```diff
@@ -0,0 +1,33 @@
...
+  /**
+   * Fetches content of the file specified under 'local' field fo recipe-type tools in {@link
+   * org.eclipse.che.api.devfile.model.Devfile}.
+   *
+   * @param localFileName file name to fetch content. Only devfile-relative files are currently
+   *     supported, so it means file should be localed at the same directory level as devfile (no
+   *     matter in repository or PR or branch etc )
+   * @return content of the specified file or {@code null} if it is impossible to fetch it.
+   */
+  String fetchContent(String localFileName);
```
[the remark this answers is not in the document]
mshaposhnik: with IOException shouldn't be very complex

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileConverter.java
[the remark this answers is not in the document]
mshaposhnik: will add explanation message

mshaposhnik:
ci-test

artaleks9:
Selenium tests execution on Eclipse Che Multiuser on OCP (https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1468/) doesn't show any regression against this Pull Request.

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileEnvironmentFactory.java
```diff
@@ -0,0 +1,129 @@
...
+    } catch (IOException e) {
+      throw new DevfileException(
+          format("Error during recipe content retrieval for tool '%s': ", recipeTool.getName())
+              + e.getMessage(),
+          e);
+    }
+    if (isNullOrEmpty(recipeFileContent)) {
+      throw new DevfileRecipeFormatException(
+          format(
+              "The local file '%s' defined in tool '%s' is unreachable or empty.",
```
sleshchenko, line 85: It is unreachable then `DevfileException` should be thrown and here it means that the referenced file is empty, right?

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileEnvironmentFactory.java
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  private KubernetesList unmarshal(Tool tool, String recipeContent)
+      throws DevfileRecipeFormatException {
+    try {
+      return Serialization.unmarshal(recipeContent, KubernetesList.class);
+    } catch (KubernetesClientException e) {
+      throw new DevfileRecipeFormatException(
+          format(
+              "Error occurred during parsing list from file %s for tool '%s'",
```
sleshchenko: Please consider adding the original error message to the thrown exception

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileEnvironmentFactory.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+  }
+
+  private String asYaml(Tool tool, KubernetesList list) throws DevfileRecipeFormatException {
+    try {
+      return Serialization.asYaml(list);
+    } catch (KubernetesClientException e) {
+      throw new DevfileRecipeFormatException(
+          format(
+              "Unable to deserialize specified local file content for tool '%s'", tool.getName()));
```
sleshchenko: Please consider adding the original error message to the thrown exception

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileManager.java
```diff
@@ -82,17 +82,21 @@ public Devfile parse(String devfileContent, boolean verbose)
    * Creates {@link WorkspaceImpl} from given devfile with available name search
    *
    * @param devfile source devfile
+   * @param recipeFileContentProvider content provider for recipe-type tool
    * @return created {@link WorkspaceImpl} instance
+   * @throws DevfileException when general devfile error occurs
```
sleshchenko: Sorry I don't fully understand what does `when general devfile error occurs` mean.
I think it would be better to rewrite java doc in the following way:
```java
   * @throws DevfileFormatException when devfile integrity validation fail
   * @throws DevfileRecipeFormatException when devfile recipe format is invalid
   * @throws DevfileException when any another devfile related error occurs
```

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileManager.java
```diff
@@ -102,12 +106,17 @@ public WorkspaceImpl createWorkspace(Devfile devfile)
    * Creates {@link WorkspaceConfigImpl} from given devfile with integrity validation
    *
    * @param devfile source devfile
+   * @param recipeFileContentProvider content provider for recipe-type tool
    * @return created {@link WorkspaceConfigImpl} instance
+   * @throws DevfileException when general devfile error occurs
```
sleshchenko: The same as about java doc for `createWorkspace` method

wsmaster/che-core-api-factory/src/test/java/org/eclipse/che/api/factory/server/urlfactory/URLFetcherTest.java
```diff
@@ -43,7 +43,7 @@
   /** Check that when url is null, NPE is thrown */
   @Test(expectedExceptions = NullPointerException.class)
   public void checkNullURL() {
-    URLFetcher.fetch((String) null);
```
sleshchenko, line 46: It would be nice to add at least one tests for `URLFetcher.fetch` method

sleshchenko (approved):

Other looks good to me

skabashnyuk (approved):

Issue #12445, opened 2019-01-16
Workspace files are not removed when Common PVC strategy if configured


sleshchenko:
### Description
Workspace files are not removed when Common PVC strategy is configured.

P.S. Logged error message is not useful at all. It is needed to consider fething Job Pod logs if possible in case of error in this issue or a separate one. 

### Reproduction Steps
1. Deploy Che Server on OpenShift (minishift on my test case) with `common` PVC strategy configured.
2. Create a workspace from `Che 7` stack or any other.
3. Start the workspace.
4. Stop the workspace.
5. Delete the workspace.
6. Che PV folder for common PVC and check Che Server logs.
**Expected:** PV folder does not contain filed of removed workspace anymore. Che Server log does not container error about failing of files removing job.
**Actual:** PV folder still contains files of removed workspaces. Che Server log contains error that occurred during files cleaning up. See `Diagnostics` section.

**OS and version:**    
Che Server 6.18.0-SNAPSHOT
Docker image `docker.io/eclipse/che-server:nightly` (`987c479083cd` 16.01.2018 14:07)

Single-User Che was deployed on minishift
```
serg@sergpad:~/projects/che$ minishift version
minishift v1.26.1+1e20f27
serg@sergpad:~/projects/che$ oc version
oc v3.10.0-rc.0+c20e215
kubernetes v1.10.0+b81c8f8
features: Basic-Auth GSSAPI Kerberos SPNEGO
```

**Diagnostics:** 

<details>
<summary>Che Server Logs</summary>

```
2019-01-16 12:00:12,536[er-ThreadPool-0]  [ERROR] [e.c.w.i.k.n.p.PVCSubPathHelper 148]  - Job 
command '[rm, -rf, /tmp/job_mount/workspace41yc8hv9o2pacs9i]' execution is failed. Status 
'PodStatus(conditions=[PodCondition(lastProbeTime=null, lastTransitionTime=2019-01-16T12:00:10Z, 
message=null, reason=null, status=True, type=Initialized, additionalProperties={}), 
PodCondition(lastProbeTime=null, lastTransitionTime=2019-01-16T12:00:10Z, message=containers with 
unready status: [rm-workspace41yc8hv9o2pacs9i], reason=ContainersNotReady, status=False, 
type=Ready, additionalProperties={}), PodCondition(lastProbeTime=null, lastTransitionTime=null, 
message=containers with unready status: [rm-workspace41yc8hv9o2pacs9i], 
[...]

PR #12513, opened 2019-01-24
Enable precreate subpaths property by default in templates

amisevsk:
### What does this PR do?
Set `CHE_INFRA_KUBERNETES_PVC_PRECREATE__SUBPATHS=true` by default in deploy scripts for Che. This is necessary to work around issue #12445 (see comment on issue for details)

#12445


ghost (approved):

sleshchenko (approved):
LGTM

You mentioned that it is needed for work around issue #12445. Should we create an issue to track solution on Kubernetes side and revert it when no work around will be needed?
Sorry If you described it somewhere in your investigation report but I missed it.

ibuziuk (approved):

garagatyi (approved):

amisevsk:
@sleshchenko There currently isn't an issue on the Kubernetes side for changing this, as far as I can tell. I need to do some more investigation to see if it's worth opening an upstream issue -- my understanding is that kube gives no guarantee for subpaths.

PR #12489, opened 2019-01-22
Handle more fields in plugin meta YAML

mkuznyetsov:
### What does this PR do?
Add more possible fields for plugin yaml (not validated anyhow atm), for use in plugin viewer:
category
publisher
repository
tags
mediaImage
mediaVideo
firstPublicationDate
latestUpdateDate
preview

### What issues does this PR fix or reference?
https://github.com/eclipse/che-plugin-registry/issues/71

PR #12538, opened 2019-01-29
Add @JsonIgnoreProperties into PluginMeta to allow more fields in meta.yaml

mshaposhnik:
### What does this PR do?
There is an PR which adds more fields into plugins `meta.yaml` in plugin-registry (https://github.com/eclipse/che-plugin-registry/pull/79).  
Update of che server is in separate PR https://github.com/eclipse/che/pull/12489
Since those updates may be desynchronized, so there is possibility that old che-server code will reject to work with updated plugins. Ths PR prevents that situation and allows to work with any additional fields in `meta.yaml`

#### Release Notes
N/A

#### Docs PR
N/A


sleshchenko (approved):

skabashnyuk (approved):

mshaposhnik:
ci-test

garagatyi (approved):

PR #12547, opened 2019-01-29
Fix execution id for generate devfile sources task

sleshchenko:
### What does this PR do?
Previously devfile sources were generated on `test` execution but were used in non-tests classes.
This PR fixes execution id for generate devfile sources task.

N/A

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/wsmaster/che-core-api-devfile/pom.xml
+++ b/wsmaster/che-core-api-devfile/pom.xml
@@ -155,7 +155,7 @@
                 <version>1.4</version>
                 <executions>
                     <execution>
-                        <id>test</id>
+                        <id>compile</id>
                         <phase>generate-sources</phase>
                         <goals>
                             <goal>add-source</goal>


sleshchenko:
ci-test

wsmaster/che-core-api-devfile/pom.xml
```diff
@@ -155,7 +155,7 @@
                 <version>1.4</version>
                 <executions>
                     <execution>
-                        <id>test</id>
+                        <id>compile</id>
```
metlos, line 158: Something more descriptive like `add-json-schema-generated-sources`?

sleshchenko: Thanks for pointing this.
Well. I faced an issue that IDE failed to resolve Devfile classes after `mvn clean install -DskipTests` and I thought it references maven phase and it does not generate sources when tests are disabled.
Now I see this id is just identifier of execution and it does not reference maven phase. So, looks like this PR fixes nothing )

sleshchenko: It even works after removing this execution at all

sleshchenko:
Closed this PR since it fixes nothing.
Rename `test` execution id to something more `add-json-schema-generated-sources` will be done as part of any other further PR not to spend time for test-report reviewing.

PR #12548, opened 2019-01-29
Add workspace ID to secret name (for self signed certs)

ghost:
### What does this PR do?

Adds wsID to certificate secret name, so that when workspaces are created in the same namespace, there are no conflicts.

This is a use case with tls mode on and self signed certs


eivantsov:
ci-build

garagatyi (approved):

garagatyi:
Please, sign your commits

sleshchenko (approved):
LGTM

eivantsov:
Since current selenium tests do not cover such a use case, i have manually verified secrets get created as expected:

image

Issue #12524, opened 2019-01-27
Publish Devfile documentation on http://redhat-developer.github.io/devfile


skabashnyuk:
### Description
followup of this comment https://github.com/eclipse/che/issues/12272#issuecomment-457552948
In this issue, we have an experiment with different  https://github.com/eclipse/che/issues/12272 ways to generate documentation from devfile json-spec. The goal of the current issue would be: generate documentation and set up a process of publishing this documentation to http://redhat-developer.github.io/devfile

mshaposhnik:
Here is the first doc version that i build manually after some enhancements in schema:  https://github.com/mshaposhnik/mshaposhnik.github.io/blob/master/devfile.md
There is an idea to host it somewhere, until we establish fully authomated build & deploy,
but need to find right place for it. @l0rd do you think that http://redhat-developer.github.io/devfile is the best place for that? Since it is weakly related with Che and it is may be hard to find it here for users? WDYT?

l0rd:
I think the `devfile` format should be Che agnostic as much as we can. Other projects should use it the same format to define/persist/configure/export their development environment. So it should not be too coupled with eclipse/che.

In eclipse/che we should document how to use a devfile to define a Che workspace. The format description should go in https://redhat-developer.github.io/devfile

mshaposhnik:
Issue is done, first deployments with Ci job is tested and works OK. Next deployments will be performed on each new tag creation.

PR #12552, opened 2019-01-30
Several json schema enhancements to being better documented

mshaposhnik:
### What does this PR do?
Adds several enhancements to json schema, so output of documentation generator looks more consistent.

A doc build from this schema:
https://github.com/mshaposhnik/mshaposhnik.github.io/blob/master/devfile.md

Related  to https://github.com/eclipse/che/issues/12524

#### Release Notes
N/A

#### Docs PR
N/A


wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json
```diff
@@ -1,49 +1,59 @@
...
+      "title": "Devfile Specification Version",
       "examples": [
         "0.0.1"
       ]
     },
     "name": {
       "type": "string",
-      "title": "Devfile Name",
+      "title": "Devfile Name.",
+      "description" : "The name of the devfile. Workspaces created from devfile, will inherit this name",
```
sleshchenko, line 48: Maybe it makes sense to add a few words about using generate name if a workspace with such name already exists.

wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json
```diff
@@ -76,7 +88,7 @@
               },
               "location": {
                 "type": "string",
-                "description": "Project-s source location address. Should be URL for git and github located projects, and file:// for zip.",
+                "description": "Project's source location address. Should be URL for git and github located projects, and file:// for zip.",
```
sleshchenko: I think the following variant is a bit better:
```
Project's source location address. Should be URL for git and github located projects, **or** file:// for zip.
```

wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json
```diff
@@ -137,21 +151,18 @@
...
             "$ref": "#/definitions/selector",
-            "description": "Describes the objects selector for the recipe type tools",
+            "description": "Describes the objects selector for the recipe type tools. Allows to pick-up only selected items from k8s/openshift list",
             "examples": [
-              {
-                "app.kubernetes.io/name": "mysql",
-                "app.kubernetes.io/component": "database",
-                "app.kubernetes.io/part-of": "petclinic"
-              }
+              "{\n   \"app.kubernetes.io/name\" : \"mysql\", \n   \"app.kubernetes.io/component\" : \"database\", \n   \"app.kubernetes.io/part-of\" : \"petclinic\" \n}"
```
sleshchenko, line 156: It was easy to read the previous variant.
Is updated way looks good in auto-generated doc?

wsmaster/che-core-api-devfile/src/test/java/org/eclipse/che/api/devfile/server/DevfileManagerTest.java
```diff
@@ -79,8 +78,8 @@ public void testValidateAndConvert() throws Exception {
   }
 
   @Test(
-      expectedExceptions = JsonProcessingException.class,
-      expectedExceptionsMessageRegExp = "Unrecognized field \"foos\" [\\w\\W]+")
+      expectedExceptions = DevfileFormatException.class,
```
sleshchenko, line 81: You changed test class without changing tested class source code. How is it supposed to work? Maybe I missed something.

sleshchenko (approved):
LGTM

wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json
mshaposhnik: Yes, library is expect examples to be array of strings, not objects. And didn't render it otherwise.
https://github.com/mshaposhnik/mshaposhnik.github.io/blob/master/devfile.md#selector-example

wsmaster/che-core-api-devfile/src/test/java/org/eclipse/che/api/devfile/server/DevfileManagerTest.java
mshaposhnik: It is due to adding `"additionalProperties": false`  in schema. So now the test raises exception during validation (but this doesn't matter for manager itself).

wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json
mshaposhnik: This is too Che specific IMO. Not related to the schema itself...

skabashnyuk (approved):

Issue #12479, opened 2019-01-21
[Devfile] Support commands for Os/K8s tool


skabashnyuk:
### Description
Now we support commands only for the ChePlugin or CheEditor tool.
We want to introduce an ability to define commands for OS/K8s tools
Like
```yaml
specVersion: 0.0.1
name: che-in-che
projects:
  - name: che
    source:
      type: git
      location: 'https://github.com/eclipse/che.git'
tools:
  - name: theia-editor
    type: cheEditor
    id: org.eclipse.che.editor.theia:1.0.0
  - name: exec-plugin
    type: chePlugin
    id: che-machine-exec-plugin:0.0.1
  - name: che-dev
    type: openshift
    local: .che-dev.yaml
commands:
  - name: build
    actions:
      - type: exec
        tool: che-dev
        command: mvn clean install
        workdir: /projects/che
```

skabashnyuk:
One of the issues that exist is: what to do if OS/K8S recipe contains more then one container.
Should we specify some additional information in command definition?
@l0rd  can you comment?

l0rd:
**Option 1** Use an `action#tool-selector` that, associated with the `action#tool-name`, would select the right container.

**Option 2** Use one of the filter that will be put in place by #12386 

**Option 3** Prompt the user when one or more container match the filters/selectors

These options are not mutually exclusive. Ideally we should put in place all of them.

sleshchenko:
> Option 1 Use an action#tool-selector that, associated with the action#tool-name, would select the right container.

@l0rd Could you clarify what you mean `action#tool-selector`? It should be a new field in action to specify `selector` for picking up the right container? How this selector is supposed to work since Kubernetes Pod has labels but not a container.

> Option 2 Use one of the filter that will be put in place by #12386

Could you clarify what does it mean for Devfile format? It is introducing new field in action to specify whether this action is supposed to be run in `recipe`/`tool` sourced container?

sleshchenko:
@l0rd Thanks for clarifications.
```yaml
    actions:
      - type: exec
        tool: 
           name: che
           selector:
               name: postgres
        command: psql
```
I expect from `selector` to be matched with objects labels. But container does not have own labels. If we would match `name` selector field with container name than it would not be so clear format since `Pods` in tool also have names.
[...]

PR #12576, opened 2019-02-01
Devfile fixes

sleshchenko:
### What does this PR do?
This PR contains several fixes and improvements that are in separate commits:
1. Fix lazy initialization of collections in some POJOs. Previously they returned new collections implementation without saving it to field. As result invocation of `workspaceConfig.getAttibutes().put("something", "something")' does not modify `WorkspaceConfig`.

2. Move 'machineName' and 'plugin' attributes definition to Command class. It would prevent typos errors in other parts of code where this attribute is also set.

3. Fix configuring plugin reference for Devfile commands. 
Previously `pluginId` was used but actually Che Server respects only `plugin` attribute.

3. Fix Devfile commands converting
For some reasons, command name was used as `{DevfileCommand#getName}:{DevfileAction#getTool}`.
Not workspace config commands are named in the same was as Devfile declares.

4. Improve validation of `Devfile#Commands#Actions` list.
We don't support multiple actions in `Devfile#commands` because of current commands model on Workspace API level. `Exec/terminal` plugin is not able to handle multiple exec commands in different containers as one Task.
Later it can be implemented along with Commands model changing on Workspace API level.

5. Set empty HashMap for null Map fields in Devfile.
We used to expect from POJO to return empty collections(including Maps) instead of null. `jsonschema2pojo-maven-plugin` returns empty collections only for List, Set types but not for Map. This commit manually initialize Devfile Maps fields after Devfile parsing. 
Maybe later it can be handled on `jsonschema2pojo-maven-plugin`, the corresponding issues is created https://github.com/joelittlejohn/jsonschema2pojo/issues/955.

It is needed for https://github.com/eclipse/che/issues/12479.

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/config/Command.java
+++ b/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/config/Command.java
@@ -28,6 +28,19 @@ public interface Command {
   String WORKING_DIRECTORY_ATTRIBUTE = "workingDir";
 
   /**
+   * {@link Command} attribute which indicates in which machine command must be run. It is optional,
+   * IDE should asks user to choose machine if null.
+   */
+  String MACHINE_NAME_ATTRIBUTE = "machineName";
+
+  /**
+   * {@link Command} attribute which indicates in which plugin command must be run. If specified
+   * plugin has multiple containers then first containers should be used. Attribute value has the
+   * following format: `{PLUGIN_ID}:{PLUGIN_VERSION}`. For example: org.eclipse.sample-plugin:0.0.1
+   */
+  String PLUGIN_ATTRIBUTE = "plugin";
+
+  /**
    * Returns command name (i.e. 'start tomcat') The name should be unique per user in one workspace,
    * which means that user may create only one command with the same name in the same workspace
    */
--- a/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/factory/model/ActionImpl.java
+++ b/ide/che-core-ide-api/src/main/java/org/eclipse/che/ide/api/factory/model/ActionImpl.java
@@ -41,7 +41,7 @@ public class ActionImpl implements Action {
   @Override
   public Map<String, String> getProperties() {
     if (properties == null) {
-      return new HashMap<>();
+      properties = new HashMap<>();
     }
     return properties;
   }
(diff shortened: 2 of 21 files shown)


garagatyi (approved):

mshaposhnik:
`We do not support multiple actions in Devfile#commands`
This is the doubtful statement  you introduced with this PR.
It goes against the approved Devfile format at https://github.com/redhat-developer/devfile/blob/master/Devfile.yaml
and devfile schema at https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json#L185
So if it's a permanent change, we need to change format & schema. If it's an temporary, i would like to state this more clearly somehow.

sleshchenko:
> It goes against the approved Devfile format at https://github.com/redhat-developer/devfile/blob/master/Devfile.yaml

Well, `Devfile.yaml` just shows what we would like to have in devfile format. Neither `runType: sequential` command attribute nor dockerimage tool, not multiple tools which are not cheEditor/chePlugin are not supported yet.

>  If it's an temporary, i would like to state this more clearly somehow.

Makes sense. 
@mshaposhnik Do you have any proposal how to do it? 
Would you like me to revert my commit and create a separate issue to implement it
Or merge this commit and create a separate issue to describe differences bettwen what we would like to have `https://github.com/redhat-developer/devfile/blob/master/Devfile.yaml` and what we have now - Workspace API.

mshaposhnik:
> Would you like me to revert my commit 
Nope. I propose to 
-  add an explanation why we did this into PR desc, and 
- also add comment in the validator why such validation is done manually instead of defining it in the schema and let be validated authomatically.

sleshchenko:
@mshaposhnik I thought you were asking about this documenting limitation in JSON Spec. Will add a detailed explanation in PR/commit description and add a comment to the validator.

mshaposhnik:
That would be double great if you put note about this into schema description of the Action list :-)

sleshchenko:
ci-test

mshaposhnik (approved):

skabashnyuk (approved):

artaleks9:
Selenium tests execution on Eclipse Che Multiuser on OCP (https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1501/) doesn't show any regression against this Pull Request.

PR #12589, opened 2019-02-04
Added machine name estimation for commands in K8s/OS tool

sleshchenko:
### What does this PR do?
The main purpose of this PR is adding machine name estimation for commands in K8s/OS tool.
Note that machine name will be estimated only if k8s/OS tool contains one container otherwise Theia will ask a user to choose a target container from all workspace containers (not only from tool containers). UX should be improved in such a case.

Also, this PR contains:
- Refactoring Devfile API to be able to provision machine attribute into command;
- Small fix: Use machine name command attribute name from Constants;

### How it can be manually tested

The following URL can be used for testing implemented functionality 
`{CHE_HOST}/f?url=https://github.com/eclipse/che/tree/devfileche`
Where CHE_HOST is a host of Che Server which has the made changes.

https://github.com/eclipse/che/issues/12479

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/KubernetesToolApplier.java
+++ b/wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/KubernetesToolApplier.java
@@ -11,9 +11,11 @@
  */
 package org.eclipse.che.api.devfile.server;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Strings.isNullOrEmpty;
 import static java.lang.String.format;
 import static java.util.Collections.emptyMap;
+import static java.util.stream.Collectors.toList;
 import static org.eclipse.che.api.devfile.server.Constants.KUBERNETES_TOOL_TYPE;
 import static org.eclipse.che.api.devfile.server.Constants.OPENSHIFT_TOOL_TYPE;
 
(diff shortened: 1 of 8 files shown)


wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/KubernetesToolApplier.java
```diff
@@ -0,0 +1,210 @@
...
+   *   <li>provisioning environment into {@link WorkspaceConfigImpl} local;
+   *   <li>provisioning machine name attribute to commands that are configured to be run in the
+   *       specified tool. Note that name will be set only if tool contains the only one container;
+   * </ul>
+   *
+   * <p>NOTE: An {@link RecipeFileContentProvider} MUST be provided in order to fetch recipe
+   * content.
+   *
+   * @param recipeTool the recipe-type tool
+   * @param contentProvider service-specific provider of recipe file content
```
mshaposhnik, line 70: params list is incomplete

mshaposhnik (approved):
Generally ok for me ATM, but `DevfileConverter` and `KubernetesToolApplier` classes became huge and with barery readable logic. (e.g upgrading commands while tools processing etc). So need to keep in mind to simpify/separate those classes on further iterations (say, during dockerimage type add). May be add separate processors for each type of toool/command etc.

skabashnyuk (approved):

sleshchenko:
ci-test

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/KubernetesToolApplier.java
```diff
@@ -135,6 +144,48 @@ private String retrieveContent(
...
+    if (pods.size() != 1) {
+      // can not estimate commands machine name
+      return;
+    }
+
+    Pod pod = pods.get(0);
+    List<Container> containers = pod.getSpec().getContainers();
+
+    if (containers.size() != 1) {
+      // can not estimate commands machine name
```
garagatyi: Could you describe why, not what?

garagatyi: Actually would be better to combine both checks because the reason is probably the same

garagatyi (approved):

sleshchenko:
@eclipse/eclipse-che-qa Please review the last test report https://github.com/eclipse/che/pull/12589#issuecomment-460643457
I guess there should not be any regressions since success rate is `98.42%`

artaleks9:
Selenium tests execution on Eclipse Che Multiuser on OCP (https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1514/) doesn't show any regression against this Pull Request.

PR #12602, opened 2019-02-06
Add docker & native build of devfile docs

mshaposhnik:
### What does this PR do?
Adds possibility to build devfile docs from json schema using docker or native using special build profiles.

https://github.com/eclipse/che/issues/12524

#### Release Notes
N/A

#### Docs PR
N/A


wsmaster/che-core-api-devfile/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+# Contributors:
+#   Red Hat, Inc. - initial API and implementation
+
+# This is a Dockerfile allowing to build dashboard by using a docker container.
+# Build step: $ docker build -t eclipse-che-dashboard
```
sleshchenko: Looks like copy/paste error `eclipse-che-dashboard`

wsmaster/che-core-api-devfile/pom.xml
```diff
@@ -191,4 +191,133 @@
             </plugin>
         </plugins>
     </build>
+    <profiles>
+    <profile>
+        <id>devfile-docs-docker</id>
```
sleshchenko: Maybe it would be better to move Devfile docs generation to a separate maven module. WDYT?

mshaposhnik: It anyway will have to contain that profiles (to allow different build types). And they are incactive by default. So we will have module which will do nothing 99% of time.

skabashnyuk:
@mshaposhnik please discuss with @riuvshin do we need maven profilers or not. Maybe the dedicated script is enough. And let us know

PR #12615, opened 2019-02-06
Use plugin ID instead of sidecar container name to override sidecar memory using workspace attributes

### LINKED ISSUE #11600: Workspace attributes should use plugin IDs rather than plugin name
### Description

In Che 7 workspaces plugins memory setting can be set in the workspace attributes using the following syntax: `sidecar.<plugin name>.memory_limit": "300Mi"`.

But `<plugin name>` can have spaces. We should use `<plugin id>` instead.

garagatyi:
### What does this PR do?
Use plugin ID instead of sidecar container name to override sidecar memory using workspace attributes.
When several sidecars are brought by a plugin all of them got the same RAM limit. 
Add docs.

Fixes https://github.com/eclipse/che/issues/11600

### DIFF
--- a/wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
+++ b/wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
@@ -119,6 +119,12 @@ public final class Constants {
    */
   public static final String WORKSPACE_TOOLING_PLUGINS_ATTRIBUTE = "plugins";
 
+  /**
+   * Template for workspace attribute key that sets sidecar limit in a plugin.
+   * %s should be replaced with plugin ID. When plugin provides several sidecars this property sets
+   * the same limit for each sidecar, so is not that useful in such a case.
+   * Value format see {@link KubernetesSize}
+   */
   public static final String SIDECAR_MEMORY_LIMIT_ATTR_TEMPLATE = "sidecar.%s.memory_limit";
 
   /**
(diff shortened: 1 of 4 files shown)


wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
```diff
@@ -119,6 +119,12 @@
    */
   public static final String WORKSPACE_TOOLING_PLUGINS_ATTRIBUTE = "plugins";
 
+  /**
+   * Template for workspace attribute key that sets sidecar limit in a plugin.
+   * %s should be replaced with plugin ID. When plugin provides several sidecars this property sets
+   * the same limit for each sidecar, so is not that useful in such a case.
+   * Value format see {@link KubernetesSize}
```
sleshchenko: Well, format is described on it class java doc but in `toBytes` method java docs. Consider moving format description to class java docs or reference `KubernetesSize#toBytes`  here

sleshchenko (approved):
LGTM

garagatyi:
ci-test

garagatyi:
@l0rd Could you review the PR? Description contains changes of the behavior

l0rd (approved):
LGTM. Is it documented somewhere how to override sidecar memory?

garagatyi:
@l0rd AFAIK no

l0rd:
@garagatyi ok we need to add something about it on Che 7 doc

garagatyi:

SkorikSergey:
Selenium tests execution on Eclipse Che Multiuser on OCP (https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1530//Selenium_20tests_20report/) doesn't show any regression against this Pull Request.

garagatyi:
@slemeur could you hint me in which section of the doc it would be better to add docs about this feature?

Issue #12442, opened 2019-01-16
Investigate how to synchronise Theia global preferences between Che workspaces


evidolob:
By default, if a user has change some preference in one workspace, we should reuse the configuration into other workspaces
If user changes the configuration options - then it would overload the defaults.
In the end we should have description how we can synchronise Theia preferences

mmorhun:
API for work with Che preferences is added into workspace client: https://github.com/eclipse/che-workspace-client/pull/15

mmorhun:
Done in https://github.com/eclipse/che-theia/pull/80

PR #12637, opened 2019-02-08
Allow to change preferences with machine token

mmorhun:
### What does this PR do?
This PR allows to change (update / replace / delete) user preferences with machine token.
This is done to be able to change user preferences from within a workspace container.

https://github.com/eclipse/che/issues/12442


skabashnyuk:
@mmorhun I think we should notify community on email about this change. If nobody will raise any concerns I'm +1.

skabashnyuk (approved):

sleshchenko (approved):
Previously, there were available the only workspace related methods or read operation for other API.
With these changes, it is possible with one machine token to modify other workspaces data.
But it's OK for me since there is no more suitable service for storing data by IDE from workspace container.

mshaposhnik (approved):

evidolob (approved):

Issue #12415, opened 2019-01-14
[Improve che-plugin.yaml] Rename some fileds


skabashnyuk:
### Description
Subtask of  https://github.com/eclipse/che/issues/12249
- Public should become an endpoint attribute
- consider renaming `Devfile#dockerimageTool#endpoints[]#targetPort` field to `port` https://github.com/redhat-developer/devfile/pull/2#discussion_r242894015
- consider renaming `Devfile#dockerimageTool#services` field to `endpoints` to sync `che-plugin.yaml` and Dockerimage tool models https://github.com/redhat-developer/devfile/pull/2/files#diff-238112cd9401c996e39c118cb7008fdcR19

skabashnyuk:
implemented

PR #12642, opened 2019-02-08
Add targetPort attribute to runtime server

### LINKED ISSUE #12611: Missing mapping between Theia plug-in endpoint port and exposed url from workspace information
### Description
when declaring inside a editor plug-in in Che7 workspaces:

```json
endpoints:
 -  name: "theia"
    public: true
    targetPort: 3100
```

at runtime I can grab some info at `/api/workspace/<id-of-workspace>`
which returns
```json
"theia": {
"url": "https://route-fbenoit-1-che.starter-us-east-2.openshiftapps.com",
"attributes": {
"cookiesAuthEnabled": "true",
"internal": "false",
"secure": "true",
"type": "ide"
```

What I would like is a way to grab from API that `theia` endpoint was for the port `3100` 
So I could see that theia is for port `3100` and grab the exposed URL.

It could be useful for clients to grab that information.
could be in the attributes
like:
```json
"targetPort": "3100"
```
 or
```json
"configPort": 3100
```
For Che6 workspace we had that mapping with `workspace.config`

garagatyi:
### What does this PR do?
Add port attribute to runtime server.
This is needed to help clients in figuring out which port led
to creating a server in a workspace.
Also refactor some code.

image

Fixes https://github.com/eclipse/che/issues/12611

### DIFF
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/server/KubernetesServerResolver.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/server/KubernetesServerResolver.java
@@ -111,7 +111,8 @@ public class KubernetesServerResolver {
               String path =
                   buildPath(ingressRule.getHttp().getPaths().get(0).getPath(), config.getPath());
               servers.put(
-                  name, new RuntimeServerBuilder()
+                  name,
+                  new RuntimeServerBuilder()
                       .protocol(config.getProtocol())
                       .host(host)
                       .path(path)
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/server/KubernetesServerResolver.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/server/KubernetesServerResolver.java
@@ -21,8 +21,6 @@ import io.fabric8.kubernetes.api.model.extensions.IngressRule;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import org.eclipse.che.api.core.model.workspace.config.ServerConfig;
-import org.eclipse.che.api.core.model.workspace.runtime.ServerStatus;
 import org.eclipse.che.api.workspace.server.model.impl.ServerImpl;
 import org.eclipse.che.commons.annotation.Nullable;
 import org.eclipse.che.workspace.infrastructure.kubernetes.Annotations;
@@ -87,12 +85,14 @@ public class KubernetesServerResolver {
             (name, config) ->
                 servers.put(
                     name,
-                    newServer(
-                        config.getProtocol(),
-                        service.getMetadata().getName(),
-                        config.getPort(),
-                        config.getPath(),
-                        config.getAttributes())));
+                    new RuntimeServerBuilder()
+                        .protocol(config.getProtocol())
+                        .host(service.getMetadata().getName())
+                        .port(config.getPort())
+                        .path(config.getPath())
+                        .attributes(config.getAttributes())
+                        .targetPort(config.getPort())
+                        .build()));
   }
 
   private void fillIngressServers(Ingress ingress, Map<String, ServerImpl> servers) {
@@ -111,7 +111,13 @@ public class KubernetesServerResolver {
               String path =
                   buildPath(ingressRule.getHttp().getPaths().get(0).getPath(), config.getPath());
               servers.put(
-                  name, newServer(config.getProtocol(), host, null, path, config.getAttributes()));
+                  name, new RuntimeServerBuilder()
+                      .protocol(config.getProtocol())
+                      .host(host)
+                      .path(path)
+                      .attributes(config.getAttributes())
+                      .targetPort(config.getPort())
+                      .build());
             });
   }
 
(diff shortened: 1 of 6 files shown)


benoitf (approved):
wonderful

garagatyi:
ci-test

sleshchenko:
We have a plan to rename `targetPort` to `port` in the following issue https://github.com/eclipse/che/issues/12415. Initially, this topic was raised by @benoitf https://github.com/redhat-developer/devfile/pull/2#discussion_r242894015
@garagatyi So, can we use `port` attribute instead of `targetPort`?

garagatyi:
@sleshchenko makes sense. Will do

garagatyi:
@sleshchenko, On the other hand, it might be confusing whether it means port that needs to be added to the server URL or port that was originally exposed. 
@benoitf @l0rd How do you think what naming would be better?

garagatyi:
ci-test

l0rd:
👍 to use `port`

benoitf:
fine as well for port (it's an attribute so I'm fine)

garagatyi:
@sleshchenko I renamed the attribute

garagatyi:
ci-build

garagatyi:
ci-test

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/server/KubernetesServerResolverTest.java
```diff
@@ -263,4 +265,10 @@ private Ingress createIngress(
         .endStatus()
         .build();
   }
+
+  private Map<String, String> defaultAttributeAnd(String key, String value) {
```
sleshchenko, line 269: It is not so important but still `defaultAttributeAnd` -> `defaultAttributesAnd`

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/server/KubernetesServerResolverTest.java
```diff
@@ -263,4 +265,10 @@ private Ingress createIngress(
...
+    HashMap<String, String> map = new HashMap<>(ATTRIBUTES_MAP);
+    map.put(key, value);
+    return map;
```
sleshchenko: It's a bad manner to name variable after its type

wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
```diff
@@ -174,5 +175,8 @@
   /** The projects volume has a standard name used in a couple of locations. */
   public static final String PROJECTS_VOLUME_NAME = "projects";
 
+  /** Attribute of {@link Server} that specifies exposure of which port created the server */
+  public static final String SERVER_TARGET_PORT_ATTRIBUTE = "port";
```
sleshchenko: Should the constant name be updated as well?

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/server/KubernetesServerResolver.java
```diff
@@ -87,12 +85,14 @@ private void fillServiceServers(Service service, Map<String, ServerImpl> servers
...
-                        config.getPort(),
-                        config.getPath(),
-                        config.getAttributes())));
+                    new RuntimeServerBuilder()
+                        .protocol(config.getProtocol())
+                        .host(service.getMetadata().getName())
+                        .port(config.getPort())
+                        .path(config.getPath())
+                        .attributes(config.getAttributes())
+                        .targetPort(config.getPort())
```
sleshchenko, line 94: Should be `targetPort` be renamed here to `port` as well?

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/server/KubernetesServerResolverTest.java
garagatyi: Since there is only one default attribute I would leave it as is

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/server/KubernetesServerResolverTest.java
garagatyi: ok

garagatyi: I mean I'll rename it

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/server/KubernetesServerResolver.java
garagatyi: I'll rename it

wsmaster/che-core-api-workspace-shared/src/main/java/org/eclipse/che/api/workspace/shared/Constants.java
garagatyi: I'll rename it

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/server/KubernetesServerResolver.java
garagatyi: Actually there is `port` method which is needed to build real URL of the server. I'll leave targetPort here

garagatyi:

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/server/KubernetesServerResolverTest.java
sleshchenko: But you use map as default attributes, later someone may just put new entry in `ATTRIBUTES_MAP` and do you even notice this private method. Up to you since it is a private method of test class.

sleshchenko (approved):
LGTM

Would be nice to see updated PR description to avoid confusion.

SkorikSergey:
Selenium tests execution on Eclipse Che Multiuser on OCP (https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1537//Selenium_20tests_20report/) doesn't show any regression against this Pull Request.

garagatyi:
@sleshchenko done
@SkorikSergey thanks!

PR #12646, opened 2019-02-11
Introduce devfile docs build & deploy script

mshaposhnik:
### What does this PR do?
Introduces devfile docs build & deploy script. 
 - Docs can be build in both native and docker mode (using `--docker` switch).
-  Script requires `DEVFILE_DOCS_GITHUB_TOKEN` to be set to clone/push docs repo.
- Commit message can be specified with `--message` parameter
- Deploy can be skipped with `--no-deploy` parameter

https://github.com/eclipse/che/issues/12524

#### Release Notes
N/A

#### Docs PR
N/A


riuvshin (approved):
Looks good!

skabashnyuk (approved):

wsmaster/che-core-api-devfile/src/main/resources/build_docs.sh
```diff
@@ -0,0 +1,91 @@
...
+  GH_TOKEN="${DEVFILE_DOCS_GITHUB_TOKEN}"
+fi
+
+build_with_docker() {
+    echo "Building docs using docker."
+    check_packages docker tar git
+    DOCKER_IMAGE_NAME="eclipse-che-devfile-docs"
+    docker build -t ${DOCKER_IMAGE_NAME} .
+    mkdir -p ${TMP_DIR} && cd ${TMP_DIR}
+    mkdir -p docs
```
sleshchenko: You can remove it and rewrite the previous line like the following:
```bash
mkdir -p ${TMP_DIR}/docs && cd ${TMP_DIR}
```

wsmaster/che-core-api-devfile/src/main/resources/build_docs.sh
```diff
@@ -0,0 +1,91 @@
...
+}
+
+cleanup
+if [[ "$1" == "--docker" ]];
+then
+    build_with_docker
+else
+    build_native
+fi
+upload
```
sleshchenko: It would be useful to have the ability to build docs without publishing it, like check how docs will look like with local changes in `devfile.json` file. Please consider adding such.

wsmaster/che-core-api-devfile/src/main/resources/build_docs.sh
```diff
@@ -0,0 +1,91 @@
...
+   mv ./docs/devfile.md ./docs/index.md
+   echo "Building docs done."
+}
+
+upload() {
+   rm -rf devfile && git clone https://${GH_TOKEN}@github.com/redhat-developer/devfile.git
+   cp -f docs/* ./devfile/docs
+   cd devfile
+   if [[ `git status --porcelain` ]]; then
+       git commit -am "Update devfile docs"
```
sleshchenko: I understand that since this script is supposed to be run on CI without developer interaction but from a script user perspective it would be nice to have an ability to specify commit message with some meaningful message like `add a description for new field` instead of hardcoded `Update devfile docs`

wsmaster/che-core-api-devfile/src/main/resources/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+RUN apt-get update && \
+    apt-get install -y git \
+    && apt-get -y clean \
+    && rm -rf /var/lib/apt/lists/*
+RUN git clone https://github.com/adobe/jsonschema2md.git
+RUN cd jsonschema2md && npm install && npm link
+
+COPY ./schema /schema
+RUN cd /schema && jsonschema2md -d . -e json -n
```
sleshchenko: I think it may be a bit optimized like the following:
```
RUN cd /schema && \
    jsonschema2md -d . -e json -n && \
    mv devfile.md index.md && \
    tar zcf /tmp/out.tar.gz -C out .
```
AFAIU docker image which are built in this way will have two layers less.

wsmaster/che-core-api-devfile/src/main/resources/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# It builds an archive file that can be used by doing later
+#  $  docker run --rm eclipse-che-devfile-docs | tar -C target/docs/ -xf -
+FROM node:8.10.0
+
+RUN apt-get update && \
+    apt-get install -y git \
+    && apt-get -y clean \
+    && rm -rf /var/lib/apt/lists/*
+RUN git clone https://github.com/adobe/jsonschema2md.git
+RUN cd jsonschema2md && npm install && npm link
```
sleshchenko, line 24: Why should we build a master branch of `jsonschema2md`? Won't it be more safety to depends on a tag?
Then we would be able to download sources of tag as archive and do not clone whole repository (not sure that it is an issue, but archive should be smaller)

sleshchenko: Note that now jsonschema2md repository contains FAILED ci status
screenshot_20190211_145331

sleshchenko: Also you can avoid installing cli tools for `jsonschema2md` since you don't need it because you can use `node cli.js -d examples/schemas -o examples/docs` instead of `jsonschema2md -d examples/schemas -o examples/docs`
P.S. I'm not familiar with `npm link` command. Maybe it's cheap command and then you can just skip my comment =)

mshaposhnik: it is just simpler to undrerstand

mshaposhnik: master have more bugs fixed & some are pending

mshaposhnik: i will fix their build when i will have free time :)

skabashnyuk: yeh. sounds like the important thing to think about.

wsmaster/che-core-api-devfile/src/main/resources/build_docs.sh
skabashnyuk: I doubt we can automate that.

sleshchenko: But we can set default value that can be overridden if needed

wsmaster/che-core-api-devfile/src/main/resources/Dockerfile
skabashnyuk: I'm worried about reproducibility. What if we check out some specific version. @sleshchenko @mshaposhnik wdyt?

sleshchenko: @skabashnyuk I wonder why do we need to clone the whole repository.
But I'm OK with both of checking out tag version or downloading sources by tag as an archive.

mshaposhnik: If we don't want to track tags manually we need to clone whole repo anyway (to find latest tag using git describe).
Or we can just manually change the tag  when necessary @skabashnyuk @sleshchenko WDYT

sleshchenko: > manually change the tag when necessary

+1

mshaposhnik: ok then i will clone using:
`git clone -b '1.1.0' --single-branch git@github.com:adobe/jsonschema2md.git`
that clones only those tag refs

sleshchenko: It still will contain redundant information about commits, like commit dates, authors, full branch history, like removed lines/files.
But I'm ok with using git if you think that downloading source archive is too difficult to implement and support.

wsmaster/che-core-api-devfile/src/main/resources/build_docs.sh
```diff
@@ -0,0 +1,123 @@
...
+
+parse_args() {
+    for i in "${@}"
+    do
+        case $i in
+           --docker)
+               IS_DOCKER=true
+               shift
+           ;;
+           --no-deploy)
```
sleshchenko, line 102: It would be nice to align the parameter name and variable/function names. `upload` or `deploy`

mshaposhnik: ok

sleshchenko (approved):
LGTM

Would be so nice to have `--help` argument which would print short information about script, possible arguments and example of usage.

Issue #12389, opened 2019-01-11
Add support of dockerimage tool in Devfile


sleshchenko:
### Description
Add support of dockerimage tool in Devfile.
dockerimage tool format should allow to specify dockerimage like (eclipse/che-server:nightly) and Che specific fields for defining volumes, servers, env vars, etc. See the following example of possible format
```yaml
---
version: 0.0.1
name: petclinic-dev-environment
projects:
  - name: petclinic
    source:
      type: git
      location: 'git@github.com:spring-projects/spring-petclinic.git'
tools:
  - name: maven
    type: dockerimage
    image: maven:3.5.4-jdk-8
    volumes:
      - name: maven-repo
        containerPath: /root/.m2
    env:
      - name: ENV_VAR
        value: value
    endpoints:
      - name: maven-server
        port: 3101
        attributes: 
          protocol: http
          secure: true
          public: true
          discoverable: false
    memoryLimit: 1500M
    mountSources: true
  - name: theia-ide
    type: cheEditor
    id: eclipse/theia:0.0.3
  - name: maven-sidecar
    type: chePlugin
    id: eclipse/maven-sidecar:0.0.3
commands:
  - name: build
    actions:
      - type: exec
        tool: maven
        command: mvn package
        workdir: /projects/spring-petclinic
```

The format should be sync with `che-plugin.yaml` format. See [Plugin Java Model](https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/wsplugins/model/PluginBase.java), [che-plugin.yaml example](https://github.com/ws-skeleton/che-machine-exec/blob/master/assembly/etc/che-plugin.yaml)

sleshchenko:
@garagatyi @l0rd @skabashnyuk Are you OK with proposed model (see issue description) for dockerimage tool?

garagatyi:
Looks good to me

l0rd:
LGTM too

sleshchenko:
Added `memoryLimit` field in dockerimage tool model

sleshchenko:
Added `mountSources` field to dockerimage tool model

skabashnyuk:
LGTM

PR #12649, opened 2019-02-11
Add support of dockerimage tool in Devfile

sleshchenko:
### What does this PR do?
The main purpose of this PR is adding support of dockerimage tool in Devfile.

But it also contains the following Devfile API improvements:
1. Improve devfile.json:
  * Deny additional properties for Tool;
  * Make tool implementation specific fields required accordingly to type value;
  * Limit actions number in command to 1 item since we don't support more yet;
2. Add testing of Devfile JSON Schema. It allows us to guarantee that no any restrictions are broken because of a typo in `devfile.json` or validator implementation we use (I faced an issue that we declare our devfile.json as JSON Spec 07 Draft but validator that we use don't support `if/then/else` statements and just skip them.
It also allows seeing an error message that is thrown by a validator in the source code, like `instance failed to match exactly one schema (matched 0 out of 3)` when `chePlugin` tool does not have id required field. It is a definitely confusing message and we should consider rewriting `devfile.json`/use another validator/customize current one to make a message more clear.

As an example of `che-in-che` Devfile with dockerimage
<details>
<summary>Che-in-Che Devfile with dockerimage tool used</summary>

```yaml
specVersion: 0.0.1
name: che-in-che
projects:
  - name: che
    source:
      type: git
      location: 'https://github.com/eclipse/che.git'
tools:
  - name: theia-editor
    type: cheEditor
    id: org.eclipse.che.editor.theia:1.0.0
  - name: exec-plugin
    type: chePlugin
    id: che-machine-exec-plugin:0.0.1
  - name: che-dev
    type: dockerimage
    image: eclipse/che-dev:nightly
    env:
      - name: TEST_ENV_VAR
        value: -\_-_-_/-
    endpoints:
      - name: tomcat
        port: 8080
        attributes:
          public: 'true'
    volumes:
      - name: maven
        containerPath: /home/user/.m2
    memoryLimit: 1536M
    mountSources: true
commands:
  - name: build
    actions:
      - type: exec
        tool: che-dev
        command: cd /projects/che && mvn -Pnative clean install
        workdir: /projects/che
```
</details>

To test changes the following link might be used `{CHE_HOST}/f?url=https://github.com/sleshchenko/che/tree/dockerimage-che-in-che`

https://github.com/eclipse/che/issues/12389

#### Release Notes
Added support of `dockerimage` tool in Devfile. It allows defining one container tool much easier than using other tools types. The detailed description of Devfile format see https://redhat-developer.github.io/devfile.

#### Docs PR
I think `devfile.json` contains all the needed changes.

### DIFF
--- a/wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DockerimageToolApplier.java
+++ b/wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DockerimageToolApplier.java
@@ -20,11 +20,15 @@ import static org.eclipse.che.api.workspace.shared.Constants.PROJECTS_VOLUME_NAM
 
 import com.google.common.collect.ImmutableMap;
 import java.util.HashMap;
+import java.util.Map;
+import java.util.Map.Entry;
 import javax.inject.Inject;
 import javax.inject.Named;
 import org.eclipse.che.api.devfile.model.Devfile;
 import org.eclipse.che.api.devfile.model.Endpoint;
+import org.eclipse.che.api.devfile.model.Env;
 import org.eclipse.che.api.devfile.model.Tool;
+import org.eclipse.che.api.devfile.model.Volume;
 import org.eclipse.che.api.workspace.server.model.impl.EnvironmentImpl;
 import org.eclipse.che.api.workspace.server.model.impl.MachineConfigImpl;
 import org.eclipse.che.api.workspace.server.model.impl.RecipeImpl;
@@ -124,5 +128,96 @@ public class DockerimageToolApplier {
         .stream()
         .filter(command -> command.getActions().get(0).getTool().equals(tool.getName()))
         .forEach(c -> c.getAttributes().put(MACHINE_NAME_ATTRIBUTE, machineName));
+  }
+
+  /**
+   * Creates dockerimage tool from the specified environment.
+   *
+   * @param environmentName the name of environment. Will be used for tool name
+   * @param environment environment that contains configuration which will be used for tool
+   * @return created dockerimage tool
+   * @throws IllegalArgumentException if the specified environment or name are null
+   * @throws IllegalArgumentException if the specified environment does not have recipe
+   * @throws IllegalArgumentException if the specified environment has non dockerimage recipe
+   * @throws WorkspaceExportException when the specified environment can not be converted to
+   *     dockerimage tool
+   */
+  public Tool from(String environmentName, EnvironmentImpl environment)
+      throws WorkspaceExportException {
+    checkArgument(environment != null, "The environment must not be null");
+    checkArgument(environmentName != null, "The environment name must not be null");
+    checkArgument(environment.getRecipe() != null, "The environment recipe must not be null");
+    checkArgument(
+        environment.getRecipe().getType().equals(DOCKERIMAGE_RECIPE_TYPE),
+        "The environment recipe must have recipe with 'dockerimage' tool");
+
+    RecipeImpl recipe = environment.getRecipe();
+    Tool dockerimageTool = new Tool();
+    dockerimageTool.setName(environmentName);
+
+    dockerimageTool.setImage(recipe.getContent());
+    dockerimageTool.setType(DOCKERIMAGE_TOOL_TYPE);
+
+    if (environment.getMachines().isEmpty()) {
+      // environment does not have additional configuration
+      return dockerimageTool;
+    }
+
+    if (environment.getMachines().size() > 1) {
+      throw new WorkspaceExportException(
+          "Environment with 'dockerimage' recipe must contain only one machine configuration");
+    }
+
+    MachineConfigImpl machineConfig = environment.getMachines().values().iterator().next();
+
+    for (Entry<String, ServerConfigImpl> serverEntry : machineConfig.getServers().entrySet()) {
+      dockerimageTool.getEndpoints().add(toEndpoint(serverEntry.getKey(), serverEntry.getValue()));
+    }
+
+    for (Entry<String, VolumeImpl> volumeEntry : machineConfig.getVolumes().entrySet()) {
+      if (volumeEntry.getKey().equals(PROJECTS_VOLUME_NAME)) {
+        dockerimageTool.setMountSources(true);
+        continue;
+      }
+
+      dockerimageTool
+          .getVolumes()
+          .add(toDevfileVolume(volumeEntry.getKey(), volumeEntry.getValue()));
+    }
+
+    dockerimageTool.setMemoryLimit(machineConfig.getAttributes().get(MEMORY_LIMIT_ATTRIBUTE));
+
+    machineConfig
+        .getEnv()
+        .entrySet()
+        .stream()
+        .map(e -> new Env().withName(e.getKey()).withValue(e.getValue()))
+        .forEach(e -> dockerimageTool.getEnv().add(e));
+
+    return dockerimageTool;
+  }
+
+  private Volume toDevfileVolume(String name, VolumeImpl volume) {
+    return new Volume().withName(name).withContainerPath(volume.getPath());
+  }
+
+  private Endpoint toEndpoint(String name, ServerConfigImpl config) {
+    String stringPort =
+        config.getPort().split("/")[0]; // cut protocol from string port like 8080/TCP
+
+    Map<String, String> attributes = new HashMap<>(config.getAttributes());
+    putIfNotNull(attributes, "protocol", config.getProtocol());
+    putIfNotNull(attributes, "path", config.getPath());
+
+    return new Endpoint()
+        .withName(name)
+        .withPort(Integer.parseInt(stringPort))
+        .withAttributes(attributes);
+  }
+
+  private void putIfNotNull(Map<String, String> attributes, String key, String value) {
+    if (value != null) {
+      attributes.put(key, value);
+    }
   }
 }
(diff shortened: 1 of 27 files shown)


wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/validator/DevfileIntegrityValidator.java
```diff
@@ -95,6 +96,9 @@ public void validateDevfile(Devfile devfile) throws DevfileFormatException {
           checkFieldNotSet(tool, "id", tool.getId());
           recipeTool = tool;
           break;
+        case DOCKERIMAGE_TOOL_TYPE:
+          // validation should be done via JSON Schema validator
```
mshaposhnik: 2 questions:
1) Can we have more than one dockerimage tool in devfile?  if no, where it is validated?
2) Can we have an K8S/OS tool and DI tool at the same time in devfile? if no, where it is validated?

sleshchenko: Nice catch. Will add the corresponding check

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DockerimageToolApplier.java
```diff
@@ -0,0 +1,127 @@
...
+   *   <li>provisioning machine name attribute to commands that are configured to be run in the
+   *       specified tool;
+   * </ul>
+   *
+   * @param tool dockerimage tool that should be applied to devfile and workspace config
+   * @param devfile devfile that should be changed according to the provided tool
+   * @param workspaceConfig workspace config that should be changed according to the provided tool
+   * @throws IllegalArgumentException when wrong type tool is passed
+   */
+  public void apply(Tool tool, Devfile devfile, WorkspaceConfigImpl workspaceConfig) {
```
mshaposhnik, line 75: To be honest, for me this method signature looks a bit strange: we're passing tool, and then whole devfile containing this tool.  
Since we're just modifying the commands from a devfile, maybe we just can pass commands list? For me it it will look clearer a bit and javadoc can be more specific than (same can be applied to the K8S tool applier).  WDYT ?

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DockerimageToolApplier.java
```diff
@@ -0,0 +1,127 @@
...
+    if (!DOCKERIMAGE_TOOL_TYPE.equals(tool.getType())) {
```
mshaposhnik: would you like to add `checkArgument()` methods as it done in K8S applier ?

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DockerimageToolApplier.java
sleshchenko: Well, I don't think that putting the command list as an argument is a good way to refactor it.
`ToolApplier` is an abstraction that may modify any part of `devfile` and now it's only command list.
As alternative I think we can refactor it in the following way:
```
interface ToolApplier {
    void apply(Devfile devfile, WorkspaceConfigImpl workspaceConfig);
}

class EditorToolApplier ...
class PluginToolApplier ... //
class KubernetesToolApplier ... // handle openshift as well
class DockerimageToolApplier ...
```
Then `DevfileConverter` won't contain any tool implementation specific logic and will only invoke `ToolApplier` implementation that may be bound as `Set`
@mshaposhnik WDYT?

wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json
```diff
@@ -155,6 +189,128 @@
             "examples": [
               "{\n   \"app.kubernetes.io/name\" : \"mysql\", \n   \"app.kubernetes.io/component\" : \"database\", \n   \"app.kubernetes.io/part-of\" : \"petclinic\" \n}"
             ]
+          },
+          "image": {
+            "type": "string",
+            "description": "Described the docker image that should be used for tool",
```
mshaposhnik: `Describes`  (or, even specifies, i would say, for exact values :) )

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DockerimageToolApplier.java
sleshchenko: I think it's OK to leave it as is since it is used once.
And ideally, it would be good to have a util class in somewhere in `commons` that would make `checkArgument` and we won't copy/paste it everywhere

wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json
sleshchenko: Thanks for catching this copy/paste error

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/validator/DevfileIntegrityValidator.java
sleshchenko: The only one tool with type different from plugin no editor is supported. Added needed check and the corresponding unit test

wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json
sleshchenko: Now it's 
```json
"description": "Specifies the docker image that should be used for tool"
```

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DockerimageToolApplier.java
mshaposhnik: Yep looks good. This will allow to concentrate logic in single place instead of spreading it.

sleshchenko: I've already done some part of the needed refactoring but it would be easier to review dockerimage support and refactoring separately. So, a new PR will be created after this PR is merged.

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DockerimageToolApplier.java
sleshchenko: I thought that you mean usage of private method `checkArgument` but you meant method from `Preconditions` class. Good catch. Fixed

sleshchenko:
@mshaposhnik I've fixed/answered your comments and added an ability to export workspace with dockerimage recipe. Please review again.

sleshchenko:
ci-test

sleshchenko:
`Success Rate: 99.369%` :tada: 
@eclipse/eclipse-che-qa Please review https://github.com/eclipse/che/pull/12649#issuecomment-463180912

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DockerimageToolApplier.java
sleshchenko: A separate issue is created to do this refactoring https://github.com/eclipse/che/issues/12659

skabashnyuk (approved):

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DockerimageToolApplier.java
```diff
@@ -0,0 +1,208 @@
...
+   * @throws IllegalArgumentException if the specified environment or name are null
+   * @throws IllegalArgumentException if the specified environment does not have recipe
+   * @throws IllegalArgumentException if the specified environment has non dockerimage recipe
+   * @throws WorkspaceExportException when the specified environment can not be converted to
+   *     dockerimage tool
+   */
+  public Tool from(String environmentName, EnvironmentImpl environment)
+      throws WorkspaceExportException {
+    checkArgument(environment != null, "The environment must not be null");
+    checkArgument(environmentName != null, "The environment must not be null");
```
mshaposhnik: Missed `name` in message

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DockerimageToolApplier.java
```diff
@@ -0,0 +1,208 @@
...
+    checkArgument(environment.getRecipe() != null, "The environment recipe must not be null");
+    checkArgument(
+        environment.getRecipe().getType().equals("dockerimage"),
+        "The environment recipe must not be null");
```
mshaposhnik: There must be probably another message about type

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DockerimageToolApplier.java
```diff
@@ -0,0 +1,208 @@
...
+    dockerimageTool.setId(recipe.getContent());
+
+    if (environment.getMachines().isEmpty()) {
+      // environment does not have additional configuration
+      return dockerimageTool;
+    }
+
+    if (environment.getMachines().size() > 1) {
+      throw new WorkspaceExportException(
+          "Environment with 'dockerimage' recipe must contains one machine configuration");
```
mshaposhnik: Minor: maybe better to say : _only_ one machine configuration

mshaposhnik (approved):

Issue #12533, opened 2019-01-28
[Tracing] Provide more context information about workspace related operation.


skabashnyuk:
### Description

In traces, we would like to see a couple of tags more
- WorkspaceId (workspace name and namespace if possible)
- UserId
- Stack if awailable

Example how it looks now.
![16 12 37](https://user-images.githubusercontent.com/1614429/51841980-893a5700-2318-11e9-90a6-3edca2981045.png)
or

![16 14 12](https://user-images.githubusercontent.com/1614429/51841971-86d7fd00-2318-11e9-8a90-16eb5245ca83.png)

PR #12652, opened 2019-02-12
Adding tracing tags to workspace related operations

mkuznyetsov:
### What does this PR do?
Add user.id and workspace.id tags for tracing of following operations:
- create workspace
- delete workspace
- start workspace
- stop workspace

https://github.com/eclipse/che/issues/12533

### DIFF
--- a/core/commons/che-core-commons-tracing/src/main/java/org/eclipse/che/commons/tracing/TracingTags.java
+++ b/core/commons/che-core-commons-tracing/src/main/java/org/eclipse/che/commons/tracing/TracingTags.java
@@ -29,6 +29,16 @@ public final class TracingTags {
   /** The id of the user the span is related to. */
   public static final AnnotationAwareStringTag USER_ID = new AnnotationAwareStringTag("user.id");
 
+  /** The id of the stack the span is related to. */
+  public static final AnnotationAwareStringTag STACK_ID = new AnnotationAwareStringTag("stack.id");
+
+  /**
+   * The entity that stopped workspace, which can be either user ID, or name of component that
+   * stopped it (e.g. activity checker) .
+   */
+  public static final AnnotationAwareStringTag STOPPED_BY =
+      new AnnotationAwareStringTag("stopped.by");
+
   /**
    * This is the standard {@link Tags#ERROR} "reexported" as an annotation aware tag so that it can
    * be easily set in the {@link org.eclipse.che.commons.annotation.Traced @Traced} methods.
--- a/core/commons/che-core-commons-tracing/src/main/java/org/eclipse/che/commons/tracing/TracingTags.java
+++ b/core/commons/che-core-commons-tracing/src/main/java/org/eclipse/che/commons/tracing/TracingTags.java
@@ -26,6 +26,9 @@ public final class TracingTags {
   public static final AnnotationAwareStringTag MACHINE_NAME =
       new AnnotationAwareStringTag("machine.name");
 
+  /** The id of the user the span is related to. */
+  public static final AnnotationAwareStringTag USER_ID = new AnnotationAwareStringTag("user.id");
+
   /**
    * This is the standard {@link Tags#ERROR} "reexported" as an annotation aware tag so that it can
    * be easily set in the {@link org.eclipse.che.commons.annotation.Traced @Traced} methods.
--- a/core/commons/che-core-commons-tracing/src/main/java/org/eclipse/che/commons/tracing/TracingTags.java
+++ b/core/commons/che-core-commons-tracing/src/main/java/org/eclipse/che/commons/tracing/TracingTags.java
@@ -37,7 +37,7 @@ public final class TracingTags {
    * stopped it (e.g. activity checker) .
    */
   public static final AnnotationAwareStringTag STOPPED_BY =
-      new AnnotationAwareStringTag("stopped.by");
+      new AnnotationAwareStringTag("stopped_by");
 
   /**
    * This is the standard {@link Tags#ERROR} "reexported" as an annotation aware tag so that it can
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
@@ -48,6 +48,7 @@ import org.eclipse.che.commons.annotation.Nullable;
 import org.eclipse.che.commons.annotation.Traced;
 import org.eclipse.che.commons.env.EnvironmentContext;
 import org.eclipse.che.commons.subject.Subject;
+import org.eclipse.che.commons.tracing.TracingTags;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -103,13 +104,18 @@ public class WorkspaceManager {
   public WorkspaceImpl createWorkspace(
       WorkspaceConfig config, String namespace, @Nullable Map<String, String> attributes)
       throws ServerException, NotFoundException, ConflictException, ValidationException {
+    TracingTags.USER_ID.set(() -> EnvironmentContext.getCurrent().getSubject().getUserId());
+
     requireNonNull(config, "Required non-null config");
     requireNonNull(namespace, "Required non-null namespace");
     validator.validateConfig(config);
     if (attributes != null) {
       validator.validateAttributes(attributes);
     }
-    return doCreateWorkspace(config, accountManager.getByName(namespace), attributes, false);
+    WorkspaceImpl workspace =
+        doCreateWorkspace(config, accountManager.getByName(namespace), attributes, false);
+    TracingTags.WORKSPACE_ID.set(workspace.getId());
+    return workspace;
   }
 
   /**
@@ -264,6 +270,9 @@ public class WorkspaceManager {
    */
   @Traced
   public void removeWorkspace(String workspaceId) throws ConflictException, ServerException {
+    TracingTags.WORKSPACE_ID.set(workspaceId);
+    TracingTags.USER_ID.set(() -> EnvironmentContext.getCurrent().getSubject().getUserId());
+
     requireNonNull(workspaceId, "Required non-null workspace id");
     if (runtimes.hasRuntime(workspaceId)) {
       throw new ConflictException(
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
@@ -105,6 +105,7 @@ public class WorkspaceManager {
       WorkspaceConfig config, String namespace, @Nullable Map<String, String> attributes)
       throws ServerException, NotFoundException, ConflictException, ValidationException {
     TracingTags.USER_ID.set(() -> EnvironmentContext.getCurrent().getSubject().getUserId());
+    TracingTags.STACK_ID.set(() -> attributes.getOrDefault("stackId", "no stack"));
 
     requireNonNull(config, "Required non-null config");
     requireNonNull(namespace, "Required non-null namespace");
@@ -272,6 +273,12 @@ public class WorkspaceManager {
   public void removeWorkspace(String workspaceId) throws ConflictException, ServerException {
     TracingTags.WORKSPACE_ID.set(workspaceId);
     TracingTags.USER_ID.set(() -> EnvironmentContext.getCurrent().getSubject().getUserId());
+    try {
+      TracingTags.STACK_ID.set(
+          workspaceDao.get(workspaceId).getAttributes().getOrDefault("stackId", "no stack"));
+    } catch (NotFoundException e) {
+      LOG.warn("Unable to provide stack id for tracing");
+    }
 
     requireNonNull(workspaceId, "Required non-null workspace id");
     if (runtimes.hasRuntime(workspaceId)) {
--- a/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
+++ b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
@@ -26,6 +26,7 @@ import static org.eclipse.che.api.workspace.shared.Constants.UPDATED_ATTRIBUTE_N
 import com.google.inject.Inject;
 import java.util.Collections;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import javax.inject.Singleton;
 import org.eclipse.che.account.api.AccountManager;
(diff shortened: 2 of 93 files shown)


wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -449,6 +452,9 @@ public void run() {
   @Traced
   public CompletableFuture<Void> stopAsync(Workspace workspace, Map<String, String> options)
       throws NotFoundException, ConflictException {
+    TracingTags.WORKSPACE_ID.set(workspace.getId());
+    TracingTags.USER_ID.set(() -> EnvironmentContext.getCurrent().getSubject().getUserId());
```
sleshchenko: If a workspace is stopped by `ActivitManager` or `ConsistencyChecker` then `"0000-00-0000"` will appear here.
Please take a look how stopped by are evaluated https://github.com/eclipse/che/blob/6553ddfbac6e78b8f3857158668f99ac84c2ea2e/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java#L470-L472
Maybe it would be more helpful to see cron job here instead of `"0000-00-0000"`

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -103,13 +104,18 @@ public WorkspaceManager(
   public WorkspaceImpl createWorkspace(
       WorkspaceConfig config, String namespace, @Nullable Map<String, String> attributes)
       throws ServerException, NotFoundException, ConflictException, ValidationException {
+    TracingTags.USER_ID.set(() -> EnvironmentContext.getCurrent().getSubject().getUserId());
```
garagatyi: Can't we do that in the EnvironementInitializationFilter to avoid adding this top all the sites?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -312,6 +313,8 @@ public WorkspaceStatus getStatus(String workspaceId) {
   public CompletableFuture<Void> startAsync(
       Workspace workspace, @Nullable String envName, Map<String, String> options)
       throws ConflictException, NotFoundException, ServerException {
+    TracingTags.WORKSPACE_ID.set(workspace.getId());
```
garagatyi, line 316: We could do that in the WorkspaceManger.startAsync as the first async action. WDYT?

garagatyi:
@sleshchenko @mkuznyetsov please take a look t my comments

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
sleshchenko: `WorkspaceManger.startAsync` is not traced at all. 
Since `WorkspaceManger.startAsync` does not contain any long living task - I think it's OK to keep `WorkspaceRuntimes.startAsync` traced.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
skabashnyuk: +0. Should we have TracingInitializationFilter for that? wdyt @sleshchenko @metlos ?

sleshchenko: Well, it has two sides:
On one hand: It would mean that we will push `USER_ID` tag to all http requests even when it does not contains any traced method.
On another hand: each traced method would contain user id tags if it is invoked via users request. But it would not have any user id if an operation is done by cron job like activity idling or consistency checker.

I'm OK to leave it as is at least for time being.

skabashnyuk: >I'm OK to leave it as is at least for time being.

+1

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
```diff
@@ -449,6 +452,13 @@ public void run() {
   @Traced
   public CompletableFuture<Void> stopAsync(Workspace workspace, Map<String, String> options)
       throws NotFoundException, ConflictException {
+    TracingTags.WORKSPACE_ID.set(workspace.getId());
+    TracingTags.STOPPED_BY.set(
+        () ->
+            firstNonNull(
```
skabashnyuk: This code looks overweight. Maybe we can leave SubjectID as is? @sleshchenko @mshaposhnik @metlos wdyt?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
```diff
@@ -264,6 +271,15 @@ public WorkspaceImpl updateWorkspace(String id, Workspace update)
    */
   @Traced
   public void removeWorkspace(String workspaceId) throws ConflictException, ServerException {
+    TracingTags.WORKSPACE_ID.set(workspaceId);
+    TracingTags.USER_ID.set(() -> EnvironmentContext.getCurrent().getSubject().getUserId());
+    try {
+      TracingTags.STACK_ID.set(
```
skabashnyuk: This```STACK_ID``` code looks overweight. Maybe we can remove it at all? @sleshchenko @mshaposhnik @metlos wdyt?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
sleshchenko: I think that `stopped by` tag could be really useful in traces. Well, maybe you would like code with private method and method reference?
```
public CompletableFuture<Void> stopAsync(Workspace workspace, Map<String, String> options) throws NotFoundException, ConflictException {
    TracingTags.WORKSPACE_ID.set(workspace.getId());
    TracingTags.STOPPED_BY.set(() -> getStoppedBy(workspace));
     ...
}
```

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
sleshchenko: The code could be rewritten here in more elegant way. More important question - can information about stack id of removed workspace be useful? If yes - then we should set it.
@garagatyi @ibuziuk Guys, do you think it can be important for OSIO installation?

garagatyi: Aren't we going to remove stacks in Che 7? Not sure it is that useful to add stack ID

skabashnyuk:
closing. work would be finalized here https://github.com/eclipse/che/pull/12673

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
mkuznyetsov: @garagatyi I can simplify retrieval of stackId by having removeWorkspace method of dao to return the entity of workspace (so no extra call of dao method will be needed)

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
mkuznyetsov: done like suggested

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
mkuznyetsov: reworked

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
metlos: We already trace all HTTP requests, regardless of whether there is a further more detailed tracing through https://github.com/eclipse/che/blob/master/core/che-core-tracing-web/src/main/java/org/eclipse/che/core/tracing/web/TracingWebModule.java so I see no harm done in adding the user info.

metlos: So in another words, I agree with @garagatyi that this should be moved to `EnvironmentInitializationFilter`. We need to make sure this also works in the multi-user mode though (in which the `EnvironmentInitializationFilter` is not applied I think).

metlos (changes requested):
I can't comment on the issues with `STACK_ID` but I think we could add the `USER_ID` generically for all HTTP requests in a similar fashion how we add the user id to the environment context.

metlos (approved):

skabashnyuk (approved):

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceRuntimes.java
mkuznyetsov: then I'll leave it as it is

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
mkuznyetsov: simplified

core/commons/che-core-commons-tracing/src/main/java/org/eclipse/che/commons/tracing/TracingTags.java
```diff
@@ -26,6 +26,19 @@
...
+
+  /** The id of the stack the span is related to. */
+  public static final AnnotationAwareStringTag STACK_ID = new AnnotationAwareStringTag("stack.id");
+
+  /**
+   * The entity that stopped workspace, which can be either user ID, or name of component that
+   * stopped it (e.g. activity checker) .
+   */
+  public static final AnnotationAwareStringTag STOPPED_BY =
+      new AnnotationAwareStringTag("stopped.by");
```
sleshchenko: I'm not aware of annotation names convention. But here is my felling:
`user.id` and `stack.id` looks logic for me. But I prefer `stopped_by` instead of `stopped.by`. Could we do it? @metlos WDYT?

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/WorkspaceDao.java
```diff
@@ -72,7 +73,7 @@ WorkspaceImpl update(WorkspaceImpl update)
    * @throws NullPointerException when {@code id} is null
    * @throws ServerException when any other error occurs during workspace removing
    */
-  void remove(String id) throws ServerException;
+  Optional<WorkspaceImpl> remove(String id) throws ServerException;
```
sleshchenko, line 77: Please update java doc according to made changes.

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/jpa/JpaWorkspaceDao.java
```diff
@@ -87,15 +87,17 @@ public WorkspaceImpl update(WorkspaceImpl update)
...
+    Optional<WorkspaceImpl> workspaceOpt;
     try {
-      Optional<WorkspaceImpl> workspaceOpt = doRemove(id);
+      workspaceOpt = doRemove(id);
       workspaceOpt.ifPresent(
           workspace -> eventService.publish(new WorkspaceRemovedEvent(workspace)));
     } catch (RuntimeException x) {
       throw new ServerException(x.getLocalizedMessage(), x);
     }
+    return workspaceOpt;
```
sleshchenko, line 100: Please add test in tck tests that should test made changes

sleshchenko (approved):
LGTM

core/commons/che-core-commons-tracing/src/main/java/org/eclipse/che/commons/tracing/TracingTags.java
metlos: > I prefer `stopped_by` instead of `stopped.by`. Could we do it? @metlos WDYT?

:+1:

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/jpa/JpaWorkspaceDao.java
mkuznyetsov: added

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/spi/WorkspaceDao.java
mkuznyetsov: updated

core/commons/che-core-commons-tracing/src/main/java/org/eclipse/che/commons/tracing/TracingTags.java
mkuznyetsov: ok, changing to it then

mkuznyetsov:
ci-test

wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/WorkspaceManager.java
mkuznyetsov: Ok, I did it in appropriate filters now

mkuznyetsov:
ci-test

SkorikSergey:
Selenium tests execution on Eclipse Che Multiuser on OCP (https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1574//Selenium_20tests_20report/) doesn't show any regression against this Pull Request.

Issue #12593, opened 2019-02-05
Propagate meaningful message when OpenShift resources quota is reached


sleshchenko:
### Description
I faced `Pod creation timeout. Error: null` error when OpenShift console said that Quota is reached.
It is needed to improve Che Server to propagate meaningful message when OpenShift Quota is reached.

### Reproduction Steps
1. Start a workspace that requires more resources than OpenShift Quota allows. I faced this issue during creation workspace using che-in-che Devfile http://che.openshift.io/f?url=https://github.com/eclipse/che/tree/devfileche

**Expected**: Workspace is successfully created and started.
**Actual**: Workspace is created and failed started.

**OS and version:**    
Reproduced on https://che.openshift.io 6.17.1

**Diagnostics:** 
![screenshot_20190205_160639](https://user-images.githubusercontent.com/5887312/52279632-b31bfb00-2962-11e9-9526-925ecbc738f9.png)
![screenshot_20190205_161438](https://user-images.githubusercontent.com/5887312/52279640-b7481880-2962-11e9-9db1-b43798eb47fe.png)
<details>
<summary>OpenShift Event JSON</summary>

```json
    {
      "metadata": {
        "name": "workspacegnwuoxurqmws63jy.ws-55bb8488d7.15807cd3abefe205",
        "namespace": "sleshche-che",
        "selfLink": "/api/v1/namespaces/sleshche-che/events/workspacegnwuoxurqmws63jy.ws-55bb8488d7.15807cd3abefe205",
        "uid": "12aa476b-294f-11e9-8f28-02e52a0be43d",
        "resourceVersion": "1478260914",
        "creationTimestamp": "2019-02-05T14:05:20Z"
      },
      "involvedObject": {
        "kind": "ReplicaSet",
        "namespace": "sleshche-che",
        "name": "workspacegnwuoxurqmws63jy.ws-55bb8488d7",
        "uid": "12a2ed1e-294f-11e9-8f28-02e52a0be43d",
        "apiVersion": "apps/v1",
        "resourceVersion": "1478260910"
      },
      "reason": "FailedCreate",
      "message": "Error creating: pods \"workspacegnwuoxurqmws63jy.ws-55bb8488d7-7kr66\" is forbidden: [maximum cpu usage per Pod is 6, but limit is 6153m., maximum memory usage per Pod is 3Gi, but limit is 3303899136.]",
      "source": {
        "component": "replicaset-controller"
      },
[...]

PR #12657, opened 2019-02-13
Enable getting workspace events from ReplicaSets and Deployments

amisevsk:
### What does this PR do?
- Changes `KubernetesDeployments#watchEvents` handle events from replica sets and deployments in addition to pods. This is necessary because when a deployment cannot scale a replica set up due to quota limits, the event occurs on the replica set (since no pod exists)
- Fix regression (?) where **no** unrecoverable events were handled in some cases, due to changes in how `KubernetesEnvironment` works.
  - Basic gist: `KubernetesEnvironment.pods.keys()` did not get provisioned names (i.e. it would be `dockerimage` instead of `<workspace-id>.dockerimage`) which broke matching events to workspaces.
- Add `FailedCreate` to unrecoverable events to catch quota issues.

#12593

### Testing
Create a workspace with memory higher than quota. It should fail with a sensible message. A sample quota that works with the default che deployment and a workspace requesting 1.1GB memory is:
```yaml
apiVersion: v1
kind: ResourceQuota
metadata:
  name: compute-resources
spec:
  hard:
    limits.memory: 2Gi
  scopes:
  - NotTerminating
status: {}

```

### Potential improvements
- There's no rate limiting, and all unrecoverable events are logged. In the case of quota issues, this results in >10 identical messages.
- In the case of a pod requesting more memory than is allowed, we still have to wait 1 minute for `POD_CREATION_TIMEOUT` to elapse. This is because we wait [here](https://github.com/eclipse/che/blob/e596a28030f34b94a49926fc5582c6ca0f4ecf99/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/namespace/KubernetesDeployments.java#L187) and this future does not get completed since no pods are changed. Instead startup should fail as soon as the event is received, but I don't see a clean way to improve this.

### DIFF
--- a/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
+++ b/assembly/assembly-wsmaster-war/src/main/webapp/WEB-INF/classes/che/che.properties
@@ -383,7 +383,7 @@ che.infra.kubernetes.ingress_start_timeout_min=5
 
 # If during workspace startup an unrecoverable event defined in the property occurs,
 # terminate workspace immediately instead of waiting until timeout
-che.infra.kubernetes.workspace_unrecoverable_events=FailedMount,FailedScheduling,MountVolume.SetUp failed,Failed to pull image
+che.infra.kubernetes.workspace_unrecoverable_events=FailedMount,FailedScheduling,MountVolume.SetUp failed,Failed to pull image,FailedCreate
 
 che.infra.kubernetes.bootstrapper.binary_url=http://${CHE_HOST}:${CHE_PORT}/agent-binaries/linux_amd64/bootstrapper/bootstrapper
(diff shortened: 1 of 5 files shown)


sleshchenko:
ci-test

sleshchenko:
Wanted to run tests on my PR but this tab was opened ) BTW your PR needs ci-test as well ;)

sleshchenko (approved):
LGTM

ibuziuk (approved):

garagatyi:

garagatyi (approved):

SkorikSergey:
Selenium tests execution on Eclipse Che Multiuser on OCP (https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1539//Selenium_20tests_20report/) doesn't show any regression against this Pull Request.

Issue #12659, opened 2019-02-13
Refactor DevfileConverter


sleshchenko:
### Description
Now DevfileConverter holds all logic of converting Devfile to WorkspaceConfig and vice versa.
It is quite difficult to test/maintenance/extend such an architecture. 
It is needed to refactor `DevfileConverter` and separate logic for converting commands/projects/different tools implementation to different classes.

Initially, this issue was raised in `dockerimage` tool support PR https://github.com/eclipse/che/pull/12649#discussion_r255855054

PR #12682, opened 2019-02-18
Refactored Devfile converting process

sleshchenko:
### What does this PR do?
Previously, `DevfileConverter` held mostly of all logic for different devfile parts converting.
Now, `DevfileConverter` just invoke different components which convert a small part of devfile, to wit, logic related to converting of commands, projects, different types of tool are placed in separate modules.

https://github.com/eclipse/che/issues/12659

#### Release Notes
N/A

#### Docs PR
N/A


wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/convert/tool/EditorToolConverter.java
```diff
@@ -0,0 +1,103 @@
...
+
+import java.util.Optional;
+import org.eclipse.che.api.devfile.model.Devfile;
+import org.eclipse.che.api.devfile.model.Tool;
+import org.eclipse.che.api.devfile.server.FileContentProvider;
+import org.eclipse.che.api.workspace.server.model.impl.WorkspaceConfigImpl;
+import org.eclipse.che.commons.annotation.Nullable;
+
+/**
+ * Helps to convert cheEditor tools to workspace environments and vice verse.
```
mshaposhnik: `versa`

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/convert/tool/KubernetesToolConverter.java
```diff
@@ -29,24 +29,31 @@
...
 import org.eclipse.che.api.workspace.server.model.impl.EnvironmentImpl;
 import org.eclipse.che.api.workspace.server.model.impl.RecipeImpl;
 import org.eclipse.che.api.workspace.server.model.impl.WorkspaceConfigImpl;
+import org.eclipse.che.commons.annotation.Nullable;
 import org.eclipse.che.workspace.infrastructure.kubernetes.Names;
 
 /**
- * Applies Kubernetes tool configuration on provided {@link Devfile} and {@link
- * WorkspaceConfigImpl}.
+ * Helps to convert kubernetes/openshift tools to workspace environments and vice verse.
```
mshaposhnik: `versa`

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/convert/tool/PluginToolConverter.java
```diff
@@ -0,0 +1,140 @@
...
+  @Override
+  public void toWorkspaceConfig(
+      WorkspaceConfigImpl workspaceConfig,
+      Devfile devfile,
+      @Nullable FileContentProvider contentProvider) {
+    Map<String, Tool> pluginsTools =
+        devfile
+            .getTools()
+            .stream()
+            .filter(tool -> EDITOR_TOOL_TYPE.equals(tool.getType()))
```
mshaposhnik: `PLUGIN_TOOL_TYPE` ?

sleshchenko: Good catch!

skabashnyuk (approved):

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/Constants.java
```diff
@@ -27,13 +29,24 @@
...
    * </pre>
    */
-  public static final String ALIASES_WORKSPACE_ATTRIBUTE_NAME = "toolsAliases";
+  public static final String PLUGINS_TOOLS_ALIASES_WORKSPACE_ATTRIBUTE_NAME = "pluginToolsAliases";
+
+  /** Workspace config attribute which contains cheEditor tool name. */
+  public static final String EDITOR_TOOL_ALIAS_WORKSPACE_ATTRIBUTE_NAME = "editorToolName";
+
+  /** Workspace command attributes that indicates with which tool it is associated. */
+  public static final String TOOL_NAME_COMMAND_ATTRIBUTE = "toolName";
```
mshaposhnik, line 51: Isn't `TOOL_NAME_COMMAND_ATTRIBUTE_NAME` looks more aligned ?

sleshchenko: You're right. Not sure that `NAME` suffix is needed there but it looks more aligned.
Will take a look and add `_NAME` suffix here or remove in other constants in this class.

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileManager.java
```diff
@@ -61,6 +63,20 @@ public DevfileManager(
     this.objectMapper = new ObjectMapper(new YAMLFactory());
   }
 
+  @VisibleForTesting
```
mshaposhnik, line 67: So maybe rewrite 1st constructor body as 
`this(schemaValidator, integrityValidator, devfileConverter, new ObjectMapper(new YAMLFactory()))` ?

sleshchenko: Makes sense. Will change

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/convert/CommandConverter.java
```diff
@@ -0,0 +1,96 @@
...
+    devCommand.setAttributes(command.getAttributes());
+    return devCommand;
+  }
+
+  /**
+   * Converts the specified devfile command to workspace command.
+   *
+   * @param devfileCommand devfile command that should be converted
+   * @return created workspace command based on the specified devfile command
+   */
```
mshaposhnik, line 64: missed `@throws` block

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/convert/tool/KubernetesToolConverter.java
```diff
@@ -62,56 +70,85 @@
...
-    String envName = recipeTool.getName();
+    String envName = k8sTool.getName();
     workspaceConfig.getEnvironments().put(envName, new EnvironmentImpl(recipe, emptyMap()));
     workspaceConfig.setDefaultEnv(envName);
   }
 
+  /**
+   * Converts environment with kubernetes/opensfhit recipe to kubernetes/openshift tool and injects
+   * it into the specified {@link Devfile devfile}.
+   *
```
mshaposhnik: `@params` block missed

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/convert/tool/EditorToolConverter.java
```diff
@@ -0,0 +1,101 @@
...
+        .getCommands()
+        .stream()
+        .filter(c -> c.getAttributes().get(TOOL_NAME_COMMAND_ATTRIBUTE).equals(editorToolName))
+        .forEach(c -> c.getAttributes().put(PLUGIN_ATTRIBUTE, editorTool.getId()));
+  }
+
+  /**
+   * Converts workspace editor attribute to cheEditor tool and injects it into the specified {@link
+   * Devfile devfile}.
+   */
```
mshaposhnik: `@params`

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/convert/tool/PluginToolConverter.java
```diff
@@ -0,0 +1,141 @@
...
+   *
+   * <ul>
+   *   <li>provisioning workspace plugins attribute;
+   *   <li>provisioning plugin attribute to commands that are configured to be run in the specified
+   *       tool;
+   * </ul>
+   *
+   * @param workspaceConfig workspace config that should be changed according to the provided tool
+   * @param devfile devfile that may contains dockerimage tool
+   * @throws IllegalArgumentException when wrong type tool is passed
```
mshaposhnik: outdated `@throws`

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/convert/tool/PluginToolConverter.java
```diff
@@ -0,0 +1,139 @@
...
+import org.eclipse.che.api.devfile.server.FileContentProvider;
+import org.eclipse.che.api.workspace.server.model.impl.WorkspaceConfigImpl;
+import org.eclipse.che.commons.annotation.Nullable;
+
+/**
+ * Helps to convert chePlugin tools to workspace environments and vice versa.
+ *
+ * @author Sergii Leshchenko
+ */
+public class PluginToolConverter implements ToolConverter {
```
garagatyi: Looks like all the new `*converter` classes are not converters but rather appliers/provisioners. Would be cool to refactor them to simplify and make less ambiguous.

mshaposhnik (approved):
OK fo me. Minor comments inlined.

garagatyi (commented):
Overall PR improves the code, so :+1:. Please take a look at my comments about overall design of the sub-system. By reviewing this PR I realized that we have the same issues with the code all over the codebase of workspace master. Would be great to clean up the code with every PR.

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/convert/tool/ToolConverter.java
```diff
@@ -0,0 +1,59 @@
...
+   *
+   * @param workspaceConfig workspace config where tool should be injected
+   * @param devfile devfile that contains tools
+   * @param contentProvider optional file content provider that can be used for external tool
+   *     configuration retrieving.
+   * @throws IllegalArgumentException when there is no content provider for recipe-type tool
+   * @throws DevfileRecipeFormatException when recipe-type tool content is empty or has wrong format
+   * @throws DevfileException when general devfile error occurs
+   */
+  void toWorkspaceConfig(
```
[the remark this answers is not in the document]
sleshchenko: @garagatyi Do you have an idea what how it can be refactored?

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileManager.java
```diff
@@ -169,4 +177,23 @@ private WorkspaceConfigImpl findAvailableName(WorkspaceConfigImpl config) throws
     }
     return config;
   }
+
+  private void initializeMaps(Devfile devfile) {
```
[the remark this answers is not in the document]
sleshchenko: Added
```java
 /**
   * POJOs are more convenient to use if their collections (including maps) are initialized by
   * default. It allows to adding additional check on null, and initialization before putting first
   * element. Our DTOs and POJOs in other APIs are implemented in this way. But, unfortunately
   * Devfile uses POJOs generator that can't initialize maps by default, see
   * https://github.com/joelittlejohn/jsonschema2pojo/issues/955. It's why we need to do it
   * manually.
   *
   * @param devfile devfile where all maps (including nested objects) should be initialized
   */
```

wsmaster/che-core-api-devfile/src/test/java/org/eclipse/che/api/devfile/server/convert/CommandConverterTest.java
```diff
@@ -0,0 +1,115 @@
...
+      throws Exception {
+    // given
+    CommandImpl workspaceCommand = new CommandImpl("build", "mvn clean install", "custom");
+
+    // when
+    commandConverter.toDevfileCommand(workspaceCommand);
+  }
+
+  @Test
+  public void shouldCommandsOfDevfileToWorkspaceCommands() throws Exception {
```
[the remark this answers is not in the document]
sleshchenko: Thanks for catching it. It is supposed to be `shouldConvertCommandsOfDevfileToWorkspaceCommands`

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/convert/DevfileConverter.java
```diff
@@ -0,0 +1,137 @@
...
+  public Devfile workspaceToDevFile(WorkspaceConfigImpl wsConfig) throws WorkspaceExportException {
+    if (wsConfig.getEnvironments().size() > 1) {
+      throw new WorkspaceExportException(
+          format(
+              "Workspace '%s' cannot be converted to devfile because it contains multiple environments",
+              wsConfig.getName()));
+    }
+
+    Devfile devfile =
+        new Devfile().withSpecVersion(CURRENT_SPEC_VERSION).withName(wsConfig.getName());
```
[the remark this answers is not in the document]
sleshchenko: It was not, but now it is. Good catch!

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/convert/CommandConverter.java
```diff
@@ -0,0 +1,96 @@
...
+    command.setType(commandAction.getType());
+    command.setCommandLine(commandAction.getCommand());
+
+    if (commandAction.getWorkdir() != null) {
+      command.getAttributes().put(WORKING_DIRECTORY_ATTRIBUTE, commandAction.getWorkdir());
+    }
+
+    command.getAttributes().put(Constants.TOOL_NAME_COMMAND_ATTRIBUTE, commandAction.getTool());
+
+    if (devCommand.getAttributes() != null) {
```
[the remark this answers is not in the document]
sleshchenko: Actually, we should not have this check since `devCommand` is POJO and we suppose that lists/sets/maps are not nullable in POJOs. Will remove this check.

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/convert/CommandConverter.java
```diff
@@ -0,0 +1,96 @@
...
+   * @param command source workspace command
+   * @return created devfile command based on the specified workspace command
+   * @throws WorkspaceExportException if workspace command does not has specified tool name
+   *     attribute where it should be run
+   */
+  public Command toDevfileCommand(CommandImpl command) throws WorkspaceExportException {
+    Command devCommand = new Command().withName(command.getName());
+    Action action = new Action().withCommand(command.getCommandLine()).withType(command.getType());
+    String workingDir = command.getAttributes().remove(WORKING_DIRECTORY_ATTRIBUTE);
+    if (!isNullOrEmpty(workingDir)) {
```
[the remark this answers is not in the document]
sleshchenko: Test when workingDir is not null exists.
Added test when workingDir is null.

This check came from master branch https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileConverter.java#L275
And I'm not sure if it makes sense to check here value on empty. Maybe we should just propagate empty value?
Like, `cd ` will go to the home directory. @garagatyi WDYT?

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/convert/CommandConverter.java
```diff
@@ -0,0 +1,96 @@
...
+    return toWorkspaceCommand(devfileCommand, commandAction);
+  }
+
+  private CommandImpl toWorkspaceCommand(Command devCommand, Action commandAction) {
+    CommandImpl command = new CommandImpl();
+    command.setName(devCommand.getName());
+    command.setType(commandAction.getType());
+    command.setCommandLine(commandAction.getCommand());
+
+    if (commandAction.getWorkdir() != null) {
```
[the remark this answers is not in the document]
sleshchenko: Test when workingDir is not null exists.
Added test when workingDir is null.

This check came from master branch https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileConverter.java#L253 And I'm not sure if it makes sense to check here value on empty. Maybe we should just propagate empty value?
Like, `cd ` will go to the home directory. @garagatyi WDYT?

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/convert/CommandConverter.java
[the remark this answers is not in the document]
sleshchenko: This check came from master branch https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileConverter.java#L276
But actually I don't see any reason to perform such check. WDYT about rewriting it to 
```java
action.setWorkdir(command.getAttributes().remove(WORKING_DIRECTORY_ATTRIBUTE));
```
?

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/convert/tool/ToolConverter.java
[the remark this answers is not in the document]
garagatyi: Yes, here are a couple of ideas:
- move filtering of tool out of the convert method to a separate method dedicated to checking if a tool needs to be converted by this converter. Maybe expose type of the converter as converter contract and let some class decide which converter should convert a particular tool/environment. 
- from config: make converter return tool instead of directly provisioning it in a devfile
- make sure converter doesn't modify entity which it converts (I don't say it does now, just saying it should not have side effects)
- separate conversion of machines, servers, volumes, commands into a separate methods to clean code

They might not be enough to make code really clean, but it is what I see as an improvement at the moment

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/convert/CommandConverter.java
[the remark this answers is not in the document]
sleshchenko: Both cases (with null and non null value for workdir) are covered

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/convert/tool/PluginToolConverter.java
sleshchenko: I believe it is not actual anymore after the latest changes

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/convert/tool/DockerimageToolConverter.java
```diff
@@ -67,15 +74,25 @@ public DockerimageToolApplier(@Named("che.workspace.projects.storage") String pr
...
    */
-  public void apply(Tool tool, Devfile devfile, WorkspaceConfigImpl workspaceConfig) {
-    checkArgument(tool != null, "Tool must not be null");
-    checkArgument(
-        tool.getType().equals(DOCKERIMAGE_TOOL_TYPE), "The tool must have `dockerimage` type");
+  public void toWorkspaceConfig(
+      WorkspaceConfigImpl workspaceConfig,
+      Devfile devfile,
+      @Nullable FileContentProvider fileContentProvider) {
+    Optional<Tool> editorToolOpt =
```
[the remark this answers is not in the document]
sleshchenko: I think it's OK in this particular case to check if something is present and do nothing if there are no objects to process

wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/convert/tool/ToolConverter.java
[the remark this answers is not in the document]
sleshchenko: I believe I have done most of your suggestions in the latest commit.

sleshchenko:
Dear reviewers, thanks all for reviewing!

I've tried to update PR according to @garagatyi suggestions, and I believe it even better now.

@skabashnyuk @mshaposhnik @garagatyi Please review again.
Note that there is one thing that I'm planning to change - fix tests methods names since they were not updated after the latest refactoring.

garagatyi (approved):
Looks good. Indeed looks cleaner now.

sleshchenko:
ci-test

skabashnyuk (approved):
LGTM

sleshchenko:
@eclipse/eclipse-che-qa Could you please review the latest test report https://github.com/eclipse/che/pull/12682#issuecomment-465979727 ?

SkorikSergey:
Selenium tests execution on Eclipse Che Multiuser on OCP (https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1560//Selenium_20tests_20report/) doesn't show any regression against this Pull Request.

PR #12688, opened 2019-02-18
Added -f -S flags for curl commands in installers

sleshchenko:
### What does this PR do?
Added -f -S flags for curl commands in installers.
From `curl --help`:
1. ` -f, --fail          Fail silently (no output at all) on HTTP errors` <- fails if http code is not 2XX
2. `-S, --show-error    Show error even when -s is used`

How it looks like if workspace agent binaries are missing in Che Server packaging
screenshot_20190218_174737
We can see a message that could help to understand the cause of failing
```
curl: (22) The requested URL returned error: 404
```
Previously it was not clear why archive is not valid
screenshot_20190218_175956

It's related to https://github.com/eclipse/che/issues/12677

#### Release Notes
N/A

#### Docs PR
N/A


sleshchenko:
ci-test

skabashnyuk (approved):

garagatyi (approved):

ghost (approved):

artaleks9:
Selenium tests execution on Eclipse Che Multiuser on OCP (https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1546/) doesn't show any regression against this Pull Request.

Issue #12577, opened 2019-02-01
Support reassigning of env variables in che-plugin.yaml


metlos:
### Description

Because we want to support using official images for tooling like Go compiler, etc, we need to have a way of mapping the way we configure the plugin container using Che specific env vars to the env vars that the tool itself expects.

E.g. we should make the following possible in `che-plugin.yaml`:
```yaml
env:
- name: GOPATH
  value: "$(CHE_PROJECTS_ROOT)"
```

where `CHE_PROJECTS_ROOT` is the env var we expose to the plugin container that contains the mount path of the volume with the project files.

PR #12689, opened 2019-02-18
Make variable expansion work for environment variables in k8s

metlos:
### What does this PR do?
K8s does the expansion only if it already knows about the variable being
expanded.

This means we have to sort the environment variable list prior to sending
it to k8s in such a way that vars that reference others always follow the
referenced ones.

This is a reimplementation of #12610 that uses an appropriate algorithm to produce a consistent sorting of the env vars.

#12577

#### Release Notes
* [ ] TODO

#### Docs PR
N/A

### DIFF
--- a/core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/TopologicalSort.java
+++ b/core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/TopologicalSort.java
@@ -1,0 +1,247 @@
+/*
+ * Copyright (c) 2012-2018 Red Hat, Inc.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.commons.lang;
+
+import static java.util.Collections.emptyList;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Comparator;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * This is an implementation of a stable topological sort on a directed graph.
+ *
+ * <p>The sorting does not pose any requirements on the types being sorted. Instead, the
+ * implementation merely requires a function to provide it with a set of predecessors of a certain
+ * "node". The implementation of the function is completely in the hands of the caller.
+ *
+ * <p>Additionally, a function to extract an "ID" from a node is required. The reasoning behind this
+ * is that it usually is easier to work with identifiers when establishing the predecessors than
+ * with the full node instances. That said, nothing prevents the caller from using the actual node
+ * instance as its ID if the caller so wishes. The consequence of this is that, as a side-effect of
+ * the sorting, the duplicates, as determined by the equality of {@code ID} instances, are removed
+ * from the resulting sorted list.
+ *
+ * @param <N> the type of nodes
+ * @param <ID> the type of a node ID
+ */
+public final class TopologicalSort<N, ID> {
+
+  private final Function<N, ID> identityExtractor;
+  private final Function<N, Set<ID>> directPredecessorsExtractor;
+
+  /**
+   * @param identityExtractor a function to extract some kind of value uniquely identifying a node
+   *     amongst the others.
+   * @param directPredecessorsExtractor a function returning a list of ids of direct predecessors of
+   *     a node
+   */
+  public TopologicalSort(
+      Function<N, ID> identityExtractor, Function<N, Set<ID>> directPredecessorsExtractor) {
+    this.identityExtractor = identityExtractor;
+    this.directPredecessorsExtractor = directPredecessorsExtractor;
+  }
+
+  /**
+   * Given the function for determining the predecessors of the nodes, return the list of the nodes
+   * in topological order. I.e. all predecessors will be placed sooner in the list than their
+   * successors. Note that the input collection is assumed to contain no duplicate entries as
+   * determined by the equality of the {@code ID} type. If such duplicates are present in the input
+   * collection, the output list will only contain the first instance of the duplicates from the
+   * input collection.
+   *
+   * <p>The implemented sort algorithm is stable. If there is no relationship between 2 nodes, they
+   * retain the relative position to each other as they had in the provided collection (e.g. if "a"
+   * preceded "b" in the original collection and there is no relationship between them (as
+   * determined by the predecessor function), the "a" will still precede "b" in the resulting list.
+   * Other nodes may be inserted in between them though in the result).
+   *
+   * <p>The cycles in the graph determined by the predecessor function are ignored and nodes in the
+   * cycle are placed into the output list in the source order.
+   *
+   * @param nodes the collection of nodes
+   * @return the list of nodes sorted in topological order
+   */
+  public List<N> sort(Collection<N> nodes) {
+    // the linked hashmap is important to retain the original order of elements unless required
+    // by the dependencies between nodes
+    LinkedHashMap<ID, NodeInfo<ID, N>> nodeInfos = new LinkedHashMap<>(nodes.size());
+    List<NodeInfo<ID, N>> results = new ArrayList<>(nodes.size());
+
+    int pos = 0;
+    boolean needsSorting = false;
+    for (N node : nodes) {
+      ID nodeID = identityExtractor.apply(node);
+      // we need the set to be modifiable, so let's make our own
+      Set<ID> preds = new HashSet<>(directPredecessorsExtractor.apply(node));
+      needsSorting |= !preds.isEmpty();
+
+      NodeInfo<ID, N> nodeInfo = nodeInfos.computeIfAbsent(nodeID, __ -> new NodeInfo<>());
+      nodeInfo.id = nodeID;
+      nodeInfo.predecessors = preds;
+      nodeInfo.sourcePosition = pos++;
+      nodeInfo.node = node;
+
+      for (ID pred : preds) {
+        // note that this means that we're inserting the nodeinfos into the map in an incorrect
+        // order and will have to sort them in the source order before we do the actual topo sort.
+        // We take that cost because we gamble on there being no dependencies in the nodes as a
+        // common case.
+        NodeInfo<ID, N> predNode = nodeInfos.computeIfAbsent(pred, __ -> new NodeInfo<>());
+        if (predNode.successors == null) {
+          predNode.successors = new HashSet<>();
+        }
+        predNode.successors.add(nodeID);
+      }
+    }
+
+    if (needsSorting) {
+      // because of the predecessors, we have put the nodeinfos in the map in an incorrect order.
+      // we need to correct that before we try to sort...
+      TreeSet<NodeInfo<ID, N>> tmp = new TreeSet<>(Comparator.comparingInt(a -> a.sourcePosition));
+      tmp.addAll(nodeInfos.values());
+      nodeInfos.clear();
+      tmp.forEach(ni -> nodeInfos.put(ni.id, ni));
+
+      // now we're ready to produce the results
+      sort(nodeInfos, results);
+    } else {
+      // we don't need to sort, but we need to keep the expected behavior of removing the duplicates
+      results = new ArrayList<>(nodeInfos.values());
+    }
+
+    return results.stream().map(ni -> ni.node).collect(Collectors.toList());
+  }
+
+  private void sort(LinkedHashMap<ID, NodeInfo<ID, N>> nodes, List<NodeInfo<ID, N>> results) {
+
+    while (!nodes.isEmpty()) {
+      NodeInfo<ID, N> curr = removeFirstIndependent(nodes);
+      if (curr != null) {
+        // yay, simple. Just add the found independent node to the results.
+        results.add(curr);
+      } else {
+        // ok, there is a cycle in the graph. Let's remove all the nodes in the first cycle we find
+        // from our predecessors map, add them to the result in their original order and try to
+        // continue normally
+
+        // find the first cycle in the predecessors (in the original list order)
+        Iterator<NodeInfo<ID, N>> nexts = nodes.values().iterator();
+        List<NodeInfo<ID, N>> cycle;
+        do {
+          curr = nexts.next();
+          cycle = findCycle(curr, nodes);
+        } while (cycle.isEmpty() && nexts.hasNext());
+
+        // If we ever find a graph that doesn't have any independent node, yet we fail to find a
+        // cycle in it, the universe must be broken.
+        if (cycle.isEmpty()) {
+          throw new IllegalStateException(
+              String.format(
+                  "Failed to find a cycle in a graph that doesn't seem to  have any independent"
+                      + " node. This should never happen. Please file a bug. Current state of the"
+                      + " sorting is: nodes=%s, results=%s",
+                  nodes.toString(), results.toString()));
+        }
+
+        cycle.sort(Comparator.comparingInt(a -> a.sourcePosition));
+
+        for (NodeInfo<ID, N> n : cycle) {
+          removePredecessorMapping(nodes, n);
+          results.add(n);
+        }
+      }
+    }
+  }
+
+  private void removePredecessorMapping(Map<ID, NodeInfo<ID, N>> nodes, NodeInfo<ID, N> node) {
+    if (node.successors != null) {
+      for (ID succ : node.successors) {
+        NodeInfo<ID, N> succNode = nodes.get(succ);
+        if (succNode != null) {
+          succNode.predecessors.remove(node.id);
+        }
+      }
+    }
+    nodes.remove(node.id);
+  }
+
+  private List<NodeInfo<ID, N>> findCycle(NodeInfo<ID, N> node, Map<ID, NodeInfo<ID, N>> nodes) {
+    // bail out quickly if there are no preds - should be fairly common occurrence hopefully
+    Set<ID> preds = node.predecessors;
+    if (preds == null || preds.isEmpty()) {
+      return emptyList();
+    }
+
+    List<NodeInfo<ID, N>> ret = new ArrayList<>();
+    List<ID> todo = new ArrayList<>(preds);
+
+    Set<ID> visited = new HashSet<>();
+
+    while (!todo.isEmpty()) {
+      ID n = todo.remove(0);
+
+      if (visited.contains(n)) {
+        continue;
+      }
+
+      visited.add(n);
+
+      NodeInfo<ID, N> predNode = nodes.get(n);
+      if (predNode != null) {
+        todo.addAll(predNode.predecessors);
+        ret.add(predNode);
+
+        if (predNode.equals(node)) {
+          // we found the cycle to our original node
+          return ret;
+        }
+      }
+    }
+
+    return emptyList();
+  }
+
+  private NodeInfo<ID, N> removeFirstIndependent(Map<ID, NodeInfo<ID, N>> nodes) {
+    Iterator<Entry<ID, NodeInfo<ID, N>>> it = nodes.entrySet().iterator();
+
+    while (it.hasNext()) {
+      Entry<ID, NodeInfo<ID, N>> e = it.next();
+      if (e.getValue().predecessors.isEmpty()) {
+        it.remove();
+        NodeInfo<ID, N> ret = e.getValue();
+        removePredecessorMapping(nodes, ret);
+        return ret;
+      }
+    }
+
+    return null;
+  }
+
+  private static final class NodeInfo<ID, N> {
+    private ID id;
+    private int sourcePosition;
+    private Set<ID> predecessors;
+    private Set<ID> successors;
+    private N node;
+  }
+}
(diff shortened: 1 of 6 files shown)


metlos:
ci-test

metlos:
ci-build

metlos:
ci-test

skabashnyuk (approved):

metlos:
ci-test

metlos:
ci-test

metlos:
@sleshchenko @garagatyi could you please take a look and review this?

core/commons/che-core-commons-lang/src/test/java/org/eclipse/che/commons/lang/TopologicalSortTest.java
```diff
@@ -0,0 +1,146 @@
...
+                .mapToObj(c -> String.valueOf((char) c))
+                .collect(toSet());
+
+    TopologicalSort<Var, String> sort = new TopologicalSort<>(idFunction, predecessors);
+    List<Var> newVars = sort.sort(vars);
+    vars.clear();
+    vars.addAll(newVars);
+  }
+
+  private static final class Var {
```
sleshchenko, line 123: It needs some time to understand that `name` is one character string, and value is multi-characters strings that actually contains dependency. Would not it be simpler if it would look like the following:
```java
  private static final class Var {

    private final Character name;
    private final Set<Character> dependencies;
    ...
}

```
WDYT?

core/commons/che-core-commons-lang/src/test/java/org/eclipse/che/commons/lang/TopologicalSortTest.java
```diff
@@ -0,0 +1,146 @@
...
+        vars(var("b", "."), var("a", "."), var("c", "."))
+      },
+
+      // sort dependents after dependencies
+      new Object[] {
+        vars(var("a", "b"), var("b", "."), var("c", ".")),
+        vars(var("b", "."), var("a", "b"), var("c", "."))
+      },
+
+      // sort dependents after dependencies, case 2
```
sleshchenko: What is the difference between `sort dependents after dependencies` cases?
It would be better to see descriptions like
`// sort dependents after dependencies. case 1: 2 independent + 1 dependent on one`
`// sort dependents after dependencies. case 1: 2 independent + 1 dependent on two others`
etc.

core/commons/che-core-commons-lang/src/test/java/org/eclipse/che/commons/lang/TopologicalSortTest.java
```diff
@@ -0,0 +1,146 @@
...
+    // when
+    sort(list);
+
+    // then
+    assertEquals(list, expectedSorting);
+  }
+
+  @DataProvider
+  public static Object[][] sortingCases() {
+    return new Object[][] {
```
sleshchenko, line 39: Here is a valid test case
```java
      // removes duplicates
      new Object[] {
          vars(var("a", "."), var("a", "."), var("b", ".")),
          vars(var("a", "."), var("b", "."))
      }
```
it is expected? if yes - please add note in java doc

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/TopologicalSort.java
```diff
@@ -0,0 +1,210 @@
...
+        // nodes that are not part of the cycle can be sorted properly
+        removePredecessorMapping(predecessors, successors, curr);
+        results.add(curr);
+
+        // make sure the cycle doesn't contain our current node so that it doesn't appear in the
+        // results twice
+        cycle.remove(curr);
+
+        for (ID n : cycle) {
+          removePredecessorMapping(predecessors, successors, n);
```
sleshchenko: You can avoid duplicating this code 
```java
removePredecessorMapping(predecessors, successors, curr);
results.add(curr);
```
Like the following
```java
...
        TreeSet<ID> cycle;
...
        // If we ever find a graph that doesn't have any independent node, yet we fail to find a
        // cycle in it, the universe must be broken.
        if (cycle.isEmpty()) {
          throw new IllegalStateException(
              String.format(
                  "Failed to find a cycle in a graph that doesn't seem to  have any independent"
                      + " node. This should never happen. Please file a bug. Current state of the"
                      + " sorting is: predecessors=%s, successors=%s, results=%s",
                  predecessors.toString(), successors.toString(), results.toString()));
        }

        // remove the cycle from the predecessors and update the predecessor map so that dependent
        // nodes that are not part of the cycle can be sorted properly
        // cycle elements are inserted in result in order they were found but not original order
        for(ID id : cycle) {
          removePredecessorMapping(predecessors, successors, id);
          results.add(id);
        }
      }
    }
  }

  private void removePredecessorMapping(
      Map<ID, Set<ID>> predecessors, Map<ID, Set<ID>> successors, ID node) {
    for (ID succ : successors.getOrDefault(node, emptySet())) {
      predecessors.getOrDefault(succ, emptySet()).remove(node);
    }
    predecessors.remove(node);
  }

  private TreeSet<ID> findCycle(ID node, Map<ID, Set<ID>> predecessors) {
    // bail out quickly if there are no preds - should be fairly common occurrence hopefully
    Set<ID> preds = predecessors.get(node);
    if (preds == null || preds.isEmpty()) {
      return new TreeSet<>();
    }

    TreeSet<ID> ret = new TreeSet<>();
    // start node is first item of result
    ret.add(node);
    List<ID> todo = new ArrayList<>(preds);

    Set<ID> visited = new HashSet<>();

    while (!todo.isEmpty()) {
      ID n = todo.remove(0);

      if (visited.contains(n)) {
        continue;
      }

      visited.add(n);

      todo.addAll(predecessors.getOrDefault(n, emptySet()));
      ret.add(n);

      if (n.equals(node)) {
        // we found the cycle to our original node
        return ret;
      }
    }

    return new TreeSet<>();
  }
...
```
WDYT?

sleshchenko (approved):
Good job :+1: 

Really great javadocs and comments that help understand the implementation

core/commons/che-core-commons-lang/src/test/java/org/eclipse/che/commons/lang/TopologicalSortTest.java
```diff
@@ -0,0 +1,146 @@
...
+      },
+
+      // sort dependents after dependencies, check dependee moved after dependencies
+      new Object[] {
+        vars(var('d', "."), var('a', "c"), var('b', "."), var('c', ".")),
+        vars(var('d', "."), var('b', "."), var('c', "."), var('a', "c"))
+      },
+
+      // test the robustness against cycles
+      new Object[] {vars(var('a', "a")), vars(var('a', "a"))},
```
sleshchenko, line 67: Just one more small proposal for cases names
```java
// item depends on itself
// cycle with 2 items
// cycle with 3 items + 1 independent item
```
The last case is quite tricky not sure if I can easily clear formulate description.

metlos:
ci-test

core/commons/che-core-commons-lang/src/test/java/org/eclipse/che/commons/lang/TopologicalSortTest.java
metlos: explanations added, hopefully they're intelligible :)

metlos:
I squashed and rebased on top of latest master to clean up the PR a little

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/provision/env/EnvVarsConverter.java
```diff
@@ -43,13 +54,29 @@ public void provision(KubernetesEnvironment k8sEnv, RuntimeIdentity identity)
...
+                .entrySet()
+                .stream()
+                .map(e -> new EnvVar(e.getKey(), e.getValue(), null))
+                .collect(toMap(EnvVar::getName, identity()));
+
+        // the env vars defined in our machine config take precedence over the ones already defined
+        // in the container, if any
+        container.getEnv().forEach(v -> envVars.putIfAbsent(v.getName(), v));
+
+        // The environment variable interpolation only works if a variable that is referenced
```
garagatyi:
```suggestion
        // The environment variable expansion only works if a variable that is referenced
```

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/provision/env/EnvVarsConverterTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+    // then
+    EnvVar expectedA = new EnvVar(A_VAR, A_VAL, null);
+    EnvVar expectedB = new EnvVar(B_VAR, B_VAL, null);
+    EnvVar expectedC = new EnvVar(C_VAR, C_VAL, null);
+    EnvVar expectedPreExisting = new EnvVar(PRE_EXISTING_VAR, PRE_EXISTING_VAR_NEW_VALUE, null);
+
+    assertEquals(4, testContainer.getEnv().size());
+    assertEquals(expectedB, testContainer.getEnv().get(0));
+    assertEquals(expectedC, testContainer.getEnv().get(1));
```
garagatyi: There is `assertEqualsNoOrder` for such cases. Unfortunately it check only arrays, but you can convert lists to arrays with `list.toArray()`. This would not change much, but the test would be a little bit more concise.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/util/EnvVars.java
```diff
@@ -0,0 +1,58 @@
...
+import io.fabric8.kubernetes.api.model.EnvVar;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/** Utility class for dealing with environment variables */
+public class EnvVars {
+
+  private static final Pattern REFERENCE_PATTERN = Pattern.compile("\\$\\(\\w+\\)");
```
garagatyi, line 23: There are also variable notation with curly brackets, e.g. `export MY_VAR=${ANOTHER_VAR}`
As far as I understand this class won't support it. Is it deliberately made decision to skip support of it?

garagatyi: It seems this pattern supports `$(some)` which is rather command substitution. See http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_03
Can you elaborate why we don't support more common cases of expansion like `$MY_VAR` or `${MY_VAR}`?

garagatyi: And since it is command substitution it should not even work like that for regular shells. So, it is completely custom feature that just looks like what shell do.

garagatyi: Should we implement this feature with a custom notation or use something widely known like shell expansion? 
@metlos @sleshchenko @skabashnyuk @l0rd WDYT?

metlos: This is specifically for variable expansion in kubernetes, which only allows the `$(...)` syntax. See https://github.com/kubernetes/kubernetes/blob/master/third_party/forked/golang/expansion/expand.go

metlos: Or the docs: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.13/#envvar-v1-core

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/provision/env/EnvVarsConverterTest.java
metlos: `assertEqualsNoOrder` seems to not be available in TestNG assert. I made a change to the test code anyway so that the purpose is hopefully slightly more clear.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/util/EnvVars.java
metlos: Since we assume k8s-based infrastructure, I think we should stick to the conventions available there.

garagatyi: Now it is clear. Thanks for the pointer. BTW here is the corresponding user docs https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#using-environment-variables-inside-of-your-config
Could you add a link to a doc to the source code to make it less confusing to those who don't know that k8s uses special notation?

metlos: Good idea

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/provision/env/EnvVarsConverterTest.java
garagatyi: Now it is cleaner, agree. But arrays assertion should be in test ng. See http://static.javadoc.io/org.testng/testng/6.0.1/org/testng/Assert.html#assertEqualsNoOrder(java.lang.Object[],%20java.lang.Object[])
Maybe it will be useful next time

metlos: Huh, I'm blind :wink: Thanks for pointing it me to the method, I didn't know about it before. In any case, it would not be acceptable here anyway, because the test is specifically for the order of the elements, checking that the env vars are topologically sorted.

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/TopologicalSort.java
```diff
@@ -0,0 +1,247 @@
...
+   * <p>The cycles in the graph determined by the predecessor function are ignored and nodes in the
+   * cycle are placed into the output list in the source order.
+   *
+   * @param nodes the collection of nodes
+   * @return the list of nodes sorted in topological order
+   */
+  public List<N> sort(Collection<N> nodes) {
+    // the linked hashmap is important to retain the original order of elements unless required
+    // by the dependencies between nodes
+    LinkedHashMap<ID, NodeInfo<ID, N>> nodeInfos = new LinkedHashMap<>(nodes.size());
```
garagatyi: FYI If the intention to precisely predict the size of the map than just size might not be a good predictor. This is the reason why Google Guava has methods for creation of maps with predicted size. But this should not be a problem with non-highload projects and could be treated as premature optimization.

metlos: In here we know the size **precisely**, though. Notice that for example for the successors I use the default ctor `HashSet<>()` because I have no way of predicting the target size.

IMHO, trying to "guess" an initial/needed size of the collections is not a premature optimization but just a good practice. In https://github.com/rhq-project/rhq we were able to reduce the memory consumption by 30% (!!!) by just thinking about and limiting the initial capacity of the collections involved.

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/TopologicalSort.java
```diff
@@ -0,0 +1,247 @@
...
+  private void removePredecessorMapping(Map<ID, NodeInfo<ID, N>> nodes, NodeInfo<ID, N> node) {
+    if (node.successors != null) {
+      for (ID succ : node.successors) {
+        NodeInfo<ID, N> succNode = nodes.get(succ);
+        if (succNode != null) {
+          succNode.predecessors.remove(node.id);
+        }
+      }
+    }
+    nodes.remove(node.id);
```
garagatyi: I can misunderstand but isn't it gets removed in `removeFirstIndependent`? And as far as I know LinkedHashMap throws ConcurrentModificationException if iterator is used after map was changed aside of the iterator.
But if this is intended behavior for different cases then it would be great to leave a comment about that in the code - that remove won't be performed in the iterator case and exceptional case would not occur.

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/TopologicalSort.java
```diff
@@ -0,0 +1,247 @@
...
+
+    return results.stream().map(ni -> ni.node).collect(Collectors.toList());
+  }
+
+  private void sort(LinkedHashMap<ID, NodeInfo<ID, N>> nodes, List<NodeInfo<ID, N>> results) {
+
+    while (!nodes.isEmpty()) {
+      NodeInfo<ID, N> curr = removeFirstIndependent(nodes);
+      if (curr != null) {
+        // yay, simple. Just add the found independent node to the results.
```
garagatyi, line 141: This comment says what needs to be done and it is done one line below. It could be better to explain why it needs to be done, but not what needs to be done
```suggestion
        // Independent node doesn't need processing
```

metlos:
Squashed commits together before merging.

garagatyi (commented):
It seems I failed to understand the algorithm to the end. Can't approve or requests changes because of that. I left some comments, but rely on other reviewers reviews on whether to merge right away.

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/TopologicalSort.java
garagatyi: In my opinion, usage of a precise size of another map is not reliable to set capacity of the map. The reason that capacity != size. With the default load factor 0.75 capacity is 25% bigger. See https://stackoverflow.com/questions/30220820/difference-between-new-hashmapint-and-guava-maps-newhashmapwithexpectedsizein

Sorry if I'm missing some important point why this work otherwise in your case. And would be happy to understand this maps sizing issue better.

infrastructures/kubernetes/src/test/java/org/eclipse/che/workspace/infrastructure/kubernetes/provision/env/EnvVarsConverterTest.java
garagatyi: Oh, yeah, you are right. I missed that we need to check the order

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/TopologicalSort.java
metlos: Ok, seems like you got me here :)

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/TopologicalSort.java
metlos: Excellent observation! The CME wouldn't occur, because the exact same entry is removed by the iterator in `removeFirstIndependent` prior to calling this method, so the map is never modified by this call. It is just wasteful to call it. I've modified the impl to take this into account.

metlos:
ci-build

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/TopologicalSort.java
metlos: I actually thought that the behavior was the same for `*List(int capacity)` as well as `*Set` and `*Map` cases. Thanks for pointing out the difference to me.

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/TopologicalSort.java
```diff
@@ -0,0 +1,247 @@
...
+    LinkedHashMap<ID, NodeInfo<ID, N>> nodeInfos = new LinkedHashMap<>(nodes.size());
+    List<NodeInfo<ID, N>> results = new ArrayList<>(nodes.size());
+
+    int pos = 0;
+    boolean needsSorting = false;
+    for (N node : nodes) {
+      ID nodeID = identityExtractor.apply(node);
+      // we need the set to be modifiable, so let's make our own
+      Set<ID> preds = new HashSet<>(directPredecessorsExtractor.apply(node));
+      needsSorting |= !preds.isEmpty();
```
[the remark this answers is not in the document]
metlos: This is not bit manipulation since the variable is a boolean, but I have changed it. I agree your suggestion is easier to read.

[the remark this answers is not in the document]
garagatyi: Oh, I didn't know that this is specified in the language spec https://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.22.2
Thanks for sharing!
But again, this might not be very well known stuff. In a case, when XOR is needed I would go with ^.

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/TopologicalSort.java
metlos: IMHO, both my comment and your suggestion add no essential information :wink: The reason I tend to add comments like this is to highlight the difference between the simple/fast and complicated/slow case.

core/commons/che-core-commons-lang/src/main/java/org/eclipse/che/commons/lang/TopologicalSort.java
garagatyi: I didn't mean to get you. This is a tricky part of java that we discussed a lot back when I was in the platform team, so now we can share this knowledge.

Issue #12587, opened 2019-02-04
Study/Devfile: proper object hierarchy for tools in the schema and java model


skabashnyuk:
### Description
We have in mind to support quite a lot of different tools in devfile

```yaml
tools:
  - name: mvn-stack
    type: dockerImage
    image: maven:3.5.4-jdk-8
    entryPoint: sleep infinity
    persistentVolumes:
      - name: maven-repo
        containerPath: /root/.m2
    services:
      - name: spring-boot
        port: 8080
    mountProjectsSources: true
  - name: mysql
    type: kubernetes
    local: petclinic.yaml
    selector:
      app.kubernetes.io/name: mysql
      app.kubernetes.io/component: database
      app.kubernetes.io/part-of: petclinic
  - name: theia-ide
    type: cheEditor
    id: eclipse/theia:0.0.3
  - name: jdt.ls
    type: chePlugin
    id: eclipse/theia-jdtls:0.0.3
 ```
 At this time we support cheEditor, chePlugin, kubernetes tools. But there is no proper class hierarchy. I Neither in json schema neither in Devtool model. It just a set of fields. It can become a problem when we will introduce complex tools such as dockerImage tool https://github.com/eclipse/che/issues/12389

 In this issue we would like to have a study how we can have proper object hierarchy for tools in schema and in model
 To look like this. See CloudAppDefinition.
![devfile](https://user-images.githubusercontent.com/1614429/52202133-504d3580-2876-11e9-850d-f170fa89968a.png)

mshaposhnik:
There was two directions during work on this issue, which are related each other:
1) Check if any on the JSON schema to POJO generators support and can generate hierarchical structure of the tools object;
2) Check is there is an JSON-schema validators that can handle more complex tool definitions and be compliant with 07 schema draft - support "if"/"else"/"then"  constructions etc.

As for the POJO, there is currently no library which can generate classes with any kind of hierarchy,
the only way to achieve this is to reject POJO autogeneration and use manually created objects model.

As for the schema, recommended by schema comitee library https://github.com/everit-org/json-schema could be good replacement for current one, since it supports latest draft features, but has license limitations since it uses org.json api which cannot be used in eclipse projects. See rejected CQ's for explanation:   
- https://dev.eclipse.org/ipzilla/show_bug.cgi?id=13579
- https://dev.eclipse.org/ipzilla/show_bug.cgi?id=12764

mshaposhnik:
[...]

PR #12700, opened 2019-02-19
Improved JSON Schema not to allow fields that are not typical for concrete tool type

sleshchenko:
### What does this PR do?
This PR improves JSON Schema not to allow fields that are not typical for concrete tool type.
Previously it was done by `DevfileIntegrityValidator` but actually, it is not a thing that should be checked by integrity validator and schema is able to handle it.

Also, it contains the following small but important changes:
- Check if messages are equal instead of matching pattern in DevfileSchemaValidatorTest.
It allows avoiding characters escaping for improving test cases readability.

- Format expected messages in DevfileSchemaValidatorTest for improving test cases readability.

This PR DOES NOT RESOLVE but is related to https://github.com/eclipse/che/issues/12587

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json
+++ b/wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json
@@ -122,11 +122,14 @@
                   "cheEditor",
                   "chePlugin"
                 ]
-              }
+              },
+              "name": {},
+              "id": {}
             },
             "required": [
               "id"
-            ]
+            ],
+            "additionalProperties": false
           },
           {
             "properties": {
@@ -135,11 +138,16 @@
                   "kubernetes",
                   "openshift"
                 ]
-              }
+              },
+              "name": {},
+              "local": {},
+              "localContent": {},
+              "selector": {}
             },
             "required": [
               "local"
-            ]
+            ],
+            "additionalProperties": false
           },
           {
             "properties": {
@@ -147,12 +155,20 @@
                 "enum": [
                   "dockerimage"
                 ]
-              }
+              },
+              "name": {},
+              "image": {},
+              "memoryLimit": {},
+              "mountSources": {},
+              "volumes": {},
+              "env": {},
+              "endpoints": {}
             },
             "required": [
               "image",
               "memoryLimit"
-            ]
+            ],
+            "additionalProperties": false
           }
         ],
         "properties": {
--- a/wsmaster/che-core-api-devfile/src/test/java/org/eclipse/che/api/devfile/server/validator/DevfileSchemaValidatorTest.java
+++ b/wsmaster/che-core-api-devfile/src/test/java/org/eclipse/che/api/devfile/server/validator/DevfileSchemaValidatorTest.java
@@ -92,7 +92,13 @@ public class DevfileSchemaValidatorTest {
       },
       {
         "tool/devfile_tool_with_undeclared_field.yaml",
-        "Devfile schema validation failed. Error: /devfile/tools/0 object instance has properties which are not allowed by the schema: [\"unknown\"]"
+        "Devfile schema validation failed. Errors: [/devfile/tools/0 object instance has properties which are not allowed by the schema: [\"unknown\"],"
+            + "instance failed to match exactly one schema (matched 0 out of 3),"
+            + "/devfile/tools/0 object instance has properties which are not allowed by the schema: [\"unknown\"],"
+            + "/devfile/tools/0 object instance has properties which are not allowed by the schema: [\"id\",\"unknown\"],"
+            + "/devfile/tools/0 object has missing required properties ([\"local\"]),"
+            + "/devfile/tools/0 object instance has properties which are not allowed by the schema: [\"id\",\"unknown\"],"
+            + "/devfile/tools/0 object has missing required properties ([\"image\",\"memoryLimit\"])]"
       },
       // Command model testing
       {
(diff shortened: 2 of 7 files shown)


skabashnyuk (dismissed):

mshaposhnik:
I am generally -1 on this. Looks like this is an unspecified trick, and/or you're operate on bug or side effect. This can potentially cause indefinite behavior. In fact you re-define tool properties with an empty object, when condition matches.  

I wold rather propose to make more researches on that, and maybe we find solution using  "not" operators, etc. Several promising example proposals can be found here:
https://stackoverflow.com/questions/30515253/json-schema-valid-if-object-does-not-contain-a-particular-property/30675218

sleshchenko:
> Looks like this is an unspecified trick, and/or you're operate on bug or side effect. This can potentially cause indefinite behavior. In fact you re-define tool properties with an empty object, when condition matches.

It depends on how you understand Devfile Schema. I do not re-define tool properties. Tools properties are defined in root schema, and tool type specific subschema only guarantees that there is no indistinctive fields for this particular tool type.
So, you do know any issues that can be caused by these changes?

I'm ok with finding alternative ways but I don't like defining like 
```json
          {
            "properties": {
              "type": {
                "enum": [
                  "cheEditor",
                  "chePlugin"
                ]
              }
            },
            "required": [
              "id"
            ],
            "not": {
                "required": [
                     //dockerimage
                     "name",
                     "image",
                     "memoryLimit",
                     "mountSources",
                     "volumes",
                     "env",
                     "endpoints",
                    //kubernetes/openshift
                    "local",
                    "selector"
              ]
            }
          }
```
Because then each tool type subschema should define fields of other tool types specific fields.

garagatyi:
@mshaposhnik @sleshchenko I'm not experienced in JSON Schema enough to decide with whom I agree. Could you provide an explanation (maybe links to schema specification section) that shows whether this trick is legal for the schema or not? Right now you both have your own opinion on that but didn't provide any background on why you have this opinion.

garagatyi:
@sleshchenko your PR description would effectively close mentioned issue, see github hint on that. Please change the working to avoid unintended closing of the issue

sleshchenko:
> @sleshchenko your PR description would effectively close mentioned issue, see github hint on that. Please change the working to avoid unintended closing of the issue

Good catch. Fixed.

sleshchenko:
From https://json-schema.org/
screenshot_20190220_102040

I think validators MUST support empty objects for properties definition and no indefinite behavior should be expected.
@mshaposhnik Could you please provide a result of your investigation and playing with defining the same properties in different subschemas?

mshaposhnik:
There is no problem with empty objects at all. Problem that we duplicate definitions.
From my POW adding duplicate definition for the same property (even if this definition is empty) is a not safe trick, and how validator behave is depends on its internal implementation (it may override definitions or merge them or etc). Schema itself has no examples with definition the same prorerty twice (even in different subschemas).  Let's also not forget that our schema is intended to be public,  and such tricks may be confusing.

sleshchenko:
I expect that subschemas are validated separately. And properties are not duplicated at all. Sorry, but I don't understand what the problem if two subschemas define the same properties.

mshaposhnik:
Even if they validated separately, then by logic, the internal emtpy definition `"local": {}` should allow `local` to be any type of object without resctictions (which is fact not true, it is still validated using it's external definition, e.g. you cannot send integer if it's described as string in main defilition).

sleshchenko:
@mshaposhnik And what about how we define `type` property? Do you think it's tricky too? I mean we define in root schema that it's an string https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json#L161
and then in each of `oneOf` subschemas we duplicate definition for `type` property https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json#L115

mshaposhnik:
This is an validation criteria - it defines nothing about `type` object itself (and it is directly described in specification to be such).

mshaposhnik (approved):
Discussed again and agreed that it's ok as a _temporary_ solution until we can migrate to 07 schema version and implement it in other way.
Please check that it's not hurts docs generation much before merge.

sleshchenko:
@mshaposhnik Here is how my changes affect Devfile Docs:
Diff - https://github.com/sleshchenko/devfile/commit/48899a047b1eff6183eea84bfec365de97d6bcf9
index.md - https://github.com/sleshchenko/devfile/blob/checkDocsAfterDevfileChanges/docs/index.md
It looks OK for me. Could you please review it?

mshaposhnik:
LGTM

sleshchenko:
ci-test

skabashnyuk (approved):

wsmaster/che-core-api-devfile/src/test/java/org/eclipse/che/api/devfile/server/validator/DevfileSchemaValidatorTest.java
```diff
@@ -34,7 +34,7 @@ public void setUp() throws Exception {
   @Test(dataProvider = "validDevfiles")
   public void shouldDoNotThrowExceptionOnValidationValidDevfile(String resourceFilePath)
       throws Exception {
-    schemaValidator.validateBySchema(getResource(resourceFilePath), false);
+    schemaValidator.validateBySchema(getResource(resourceFilePath), true);
```
garagatyi: Are you gonna leave true here?

wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json
```diff
@@ -130,24 +133,36 @@
                   "kubernetes",
                   "openshift"
                 ]
-              }
+              },
+              "name": {},
+              "local": {},
```
garagatyi, line 143: localContent is also needed to be added

garagatyi (commented):
Looks good, but needs syncing with last changes in devfile.json I merged recently

wsmaster/che-core-api-devfile/src/main/resources/schema/devfile.json
sleshchenko: Added

wsmaster/che-core-api-devfile/src/test/java/org/eclipse/che/api/devfile/server/validator/DevfileSchemaValidatorTest.java
sleshchenko: is not actual anymore

sleshchenko:
@mshaposhnik @garagatyi Please take a look one more time after rebasing with the master branch. Note that two more commits are added except JSON Schema improvements. Their descriptions are listed in the PR description.

mshaposhnik (approved):

garagatyi (approved):

metlos:
I hate to be late to the discussion, but have you tried defining all the attributes in definitions and referencing them in the individual properties of the various tool types using `$ref`? While using both `oneOf` and `properties` in the tools definition is very clever, it is a little bit hard to get one's head around. At least for me :smile:

sleshchenko:
> have you tried defining all the attributes in definitions and referencing them in the individual properties of the various tool types using `$ref`? While using both `oneOf` and `properties` in the tools definition is very clever, it is a little bit hard to get one's head around. At least for me

I have not tried it. Not sure if it will improve Devfile JSON Schema readability but it would make things more obvious at least until our schema validator/POJO generator does not support if/else statement.

PR #12713, opened 2019-02-20
Add an ability to provide local file content in devfile

### LINKED ISSUE #12572: REST API to create workspaces from a devfile
### Description
Currently there is a way to create a workspace from a devfile
https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/DevfileService.java#L98-L116

We can post a devfile content.

I would like to be able to create workspaces from content like this one: https://github.com/redhat-developer/devfile/blob/master/Devfile.yaml

Let say I'm a client like the dashboard where I could create content of such devfile, then this devfile is referencing other files by doing something like: `local: petclinic.yaml`

How could I submit such files through REST API ?

would it be something like MULTIPART_FORM_DATA / https://github.com/codenvy-legacy/che-core/blob/3.x/platform-api/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/FactoryService.java#L166

or to submit content with a `all-in-one` format
```
---BEGIN DEVFILE---
content
---END DEVFILE---
---BEGIN petclinic.yaml---
content
---END petclinic.yaml---
```

something else ?

somehow related to https://github.com/eclipse/che/issues/11549

garagatyi:
### What does this PR do?
Add field 'localContent' to devfile tool.
It is allowed only when field 'local' is also used.
Added tests and improved tests error message.
Usage of block string makes yaml less somplicated:
```yaml
parent:
  child: |
    I can put anything here
    and it would be inlined
    but I still can still preserve
    structure so it looks good.
    Just remember to keep all the lines intended.
```
I used field `localContent` to keep content of file specified in field `local`. See an example:
```yaml
---
specVersion: 0.0.1
tools:
  - name: mysql
    type: kubernetes
    content: |
      kind: List
      items:
       -
        apiVersion: v1
        kind: Pod
        metadata:
         name: ws
        spec:
         containers:
          -
           image: 'eclipse/che-dev:nightly'
           name: dev
           resources:
            limits:
             memory: 512Mi
    selector:
      app.kubernetes.io/name: mysql
      app.kubernetes.io/component: database
      app.kubernetes.io/part-of: petclinic
```
Here is a demo (but when I was recording it user had to specify either `local` or `content` and now it is both `local` and `localContent` or just `local`): https://youtu.be/gBIRNmQZNpY

Fixes https://github.com/eclipse/che/issues/12572

### DIFF
--- a/wsmaster/che-core-api-devfile/src/test/java/org/eclipse/che/api/devfile/server/validator/DevfileSchemaValidatorTest.java
+++ b/wsmaster/che-core-api-devfile/src/test/java/org/eclipse/che/api/devfile/server/validator/DevfileSchemaValidatorTest.java
@@ -11,6 +11,7 @@
  */
 package org.eclipse.che.api.devfile.server.validator;
 
+import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.fail;
 
 import java.io.IOException;
@@ -27,12 +28,12 @@ public class DevfileSchemaValidatorTest {
   private DevfileSchemaValidator schemaValidator;
 
   @BeforeClass
-  public void setUp() throws Exception {
+  public void setUp() {
     schemaValidator = new DevfileSchemaValidator(new DevfileSchemaProvider());
   }
 
   @Test(dataProvider = "validDevfiles")
-  public void shouldDoNotThrowExceptionOnValidationValidDevfile(String resourceFilePath)
+  public void shouldNotThrowExceptionOnValidationValidDevfile(String resourceFilePath)
       throws Exception {
     schemaValidator.validateBySchema(getResource(resourceFilePath), false);
   }
@@ -41,7 +42,11 @@ public class DevfileSchemaValidatorTest {
   public Object[][] validDevfiles() {
     return new Object[][] {
       {"editor_plugin_tool/devfile_editor_plugins.yaml"},
+      {"kubernetes_openshift_tool/devfile_kubernetes_tool_local.yaml"},
+      {"kubernetes_openshift_tool/devfile_kubernetes_tool_local_and_content_as_block.yaml"},
       {"kubernetes_openshift_tool/devfile_openshift_tool.yaml"},
+      {"kubernetes_openshift_tool/devfile_openshift_tool_local_and_content.yaml"},
+      {"kubernetes_openshift_tool/devfile_openshift_tool_local_and_content_as_block.yaml"},
       {"dockerimage_tool/devfile_dockerimage_tool.yaml"}
     };
   }
@@ -53,7 +58,12 @@ public class DevfileSchemaValidatorTest {
       schemaValidator.validateBySchema(getResource(resourceFilePath), false);
     } catch (DevfileFormatException e) {
       if (!Pattern.matches(expectedMessageRegexp, e.getMessage())) {
-        fail("DevfileFormatException with unexpected message is thrown: " + e.getMessage());
+        // we don't need assertion here,
+        // but we use that to show the difference to simplify tests fixes
+        assertEquals(
+            e.getMessage(),
+            expectedMessageRegexp,
+            "DevfileFormatException thrown with message that doesn't match expected pattern:");
       }
       return;
     }
@@ -111,6 +121,14 @@ public class DevfileSchemaValidatorTest {
       {
         "kubernetes_openshift_tool/devfile_openshift_tool_with_missing_local.yaml",
         "Devfile schema validation failed\\. Errors: \\[instance failed to match exactly one schema \\(matched 0 out of 3\\)\\]"
+      },
+      {
+        "kubernetes_openshift_tool/devfile_openshift_tool_content_without_local.yaml",
+        "Devfile schema validation failed. Errors: \\[property \"localContent\" of object has missing property dependencies \\(schema requires \\[\"local\"\\]; missing: \\[\"local\"\\]\\), instance failed to match exactly one schema \\(matched 0 out of 3\\)\\]"
+      },
+      {
+        "kubernetes_openshift_tool/devfile_kubernetes_tool_content_without_local.yaml",
+        "Devfile schema validation failed. Errors: \\[property \"localContent\" of object has missing property dependencies \\(schema requires \\[\"local\"\\]; missing: \\[\"local\"\\]\\), instance failed to match exactly one schema \\(matched 0 out of 3\\)\\]"
       },
       // Dockerimage tool model testing
       {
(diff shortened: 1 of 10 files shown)


wsmaster/che-core-api-devfile/src/main/java/org/eclipse/che/api/devfile/server/RecipeFileContentProvider.java
```diff
@@ -20,16 +21,18 @@
...
   /**
    * Fetches content of the file specified under 'local' field fo recipe-type tools in {@link
    * org.eclipse.che.api.devfile.model.Devfile}.
    *
-   * @param localFileName file name to fetch content. Only devfile-relative files are currently
+   * @param localFileName file name to fetch content. Only devfile-related files are currently
    *     supported, so it means file should be localed at the same directory level as devfile (no
    *     matter in repository or PR or branch etc )
-   * @return content of the specified file
+   * @return content of the specified file or null if content is not found
```
sleshchenko: Not sure if we need these changes in `interface` since it already defines that it throws an exception if a content retrieving issue happens.
Also `NullProvider` is used only in one Test class.

Could you please elaborate more, why we need it?

garagatyi: Actually, we have code that returns null if there is no file requested. See https://github.com/eclipse/che/blob/master/wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/urlfactory/URLFactoryBuilder.java#L110
So, it is not true that exception would be thrown in such a case. That's why I added nullable annotation and changed javadocs.

As to the NullProvider it is pure syntax sugar which is indeed used just in one place. If you think that we don't need it I can easily remove NullProvider

sleshchenko: I think it's a bug that `URLFactoryBuilder` return null, because user will see a confusing message like `content is empty` while it is internal server exception that content provider is missing.

garagatyi: Fixed

garagatyi:
@sleshchenko I addressed your concerns, please, take a look

garagatyi:
ci-build

wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/urlfactory/URLFactoryBuilder.java
```diff
@@ -119,6 +116,24 @@ public URLFactoryBuilder(
...
+      this.fileUrlProvider = fileUrlProvider;
+    }
+
+    @Override
+    public String fetchContent(String localFileName) throws IOException {
+      if (fileUrlProvider != null) {
+        return urlFetcher.fetch(fileUrlProvider.apply(localFileName));
+      } else {
+        throw new IOException(
+            format("File '%s' is not found in context of devfile", localFileName));
```
sleshchenko: As far as I see it does not mean that file is not found BUT fileUrlProvider is null.
Now it is not null of GithubParameterResolver is used
And it IS NULL if DefaultParameterResolver is used (like factory to git file content https://github.com/eclipse/che/blob/c44e1bff66c9c0eba430f8207b373fb414fb1794/wsmaster/che-core-api-factory/src/main/java/org/eclipse/che/api/factory/server/DefaultFactoryParameterResolver.java#L55)

And actually, in this place, it can not be null since it NPE will be thrown here is such case https://github.com/eclipse/che/pull/12713/files#diff-21e725624d4fb10931c9a714cfbfbbfaR100

garagatyi: I removed this contradictory fix. You can find a proper solution with Max

wsmaster/che-core-api-devfile/src/test/java/org/eclipse/che/api/devfile/server/validator/DevfileSchemaValidatorTest.java
```diff
@@ -53,7 +58,10 @@ public void shouldThrowExceptionOnValidationNonValidDevfile(
       schemaValidator.validateBySchema(getResource(resourceFilePath), false);
     } catch (DevfileFormatException e) {
       if (!Pattern.matches(expectedMessageRegexp, e.getMessage())) {
-        fail("DevfileFormatException with unexpected message is thrown: " + e.getMessage());
+        assertEquals(
+            e.getMessage(),
+            expectedMessageRegexp,
```
sleshchenko, line 65: We already have a check 
` if (!Pattern.matches(expectedMessageRegexp, e.getMessage())) {`
Why should we use `assertEquals` here? Just to print expected, actual?

sleshchenko (approved):
LGTM

wsmaster/che-core-api-devfile/src/test/java/org/eclipse/che/api/devfile/server/validator/DevfileSchemaValidatorTest.java
garagatyi: We need that because with just `fail` dev would have a message that error in different but would not have any clue what is the difference. With my changes this assert is not for real assertion but just to leverage built-in TestNG/IDE functionality to show what is the difference and do side-by-side compare. This was very handful for me while I was working on this PR.

garagatyi:
ci-test

mshaposhnik (approved):

sleshchenko:
@garagatyi Please update your PR to master state and resolve conflicts after merging of refactoring of Devfile converting. Sorry about that.

garagatyi:
@skabashnyuk Could you review this PR?

garagatyi:
ci-test

ibuziuk:
@eclipse/eclipse-che-qa could you please provide your approval based on the e2e test results ^

SkorikSergey:
Selenium tests execution on Eclipse Che Multiuser on OCP (https://ci.codenvycorp.com/job/che-pullrequests-test-ocp/1567//Selenium_20tests_20report/) doesn't show any regression against this Pull Request.

sleshchenko:
After merging this PR it is possible to implement exporting Devfile of a workspace with Kubernetes/OpenShift recipe.
@l0rd Do you think it is an important functionality? Should I create an issue for it?

l0rd:
@sleshchenko you mean a wsmaster API to export a workspace as a devfile? I don't think we need it in the short term. As of today we have factories and these should be enough to satisfy exporting use cases.

Issue #12668, opened 2019-02-14
Add an ability to specify entrypoint for dockerimage tool in Devfile


sleshchenko:
### Description
There are cases when a user wants to use some official docker image in `dockerimage` tool but default entrypoint is not suitable for his needs. In this case, it is needed to have an ability to specify entrypoint and arguments for `dockerimage` tool in Devfile.
The proposal is to add two optional fields `command` and `arguments` which type is `string array`. It's the same as Kubernetes Container model has https://docs.openshift.com/container-platform/3.5/rest_api/kubernetes_v1.html#v1-container.
It would look like the following
```yaml
  - name: nodejs
    type: dockerimage
    image: node:0.10.40
    command: ['/bin/sh', '-c']
    args: ['tail -f /dev/null']
```

sleshchenko:
@l0rd Do you think the proposed format is good, or it would be better to have something like current devfile draft has `entryPoint: sleep infinity` https://github.com/redhat-developer/devfile/blob/master/Devfile.yaml#L13 ?

l0rd:
@sleshchenko command and args is perfect.

sleshchenko:
Here is a technical issue in implementing it. We convert dockerimage tool to an environment with dockerimage recipe. For example:
```yaml
  - name: che-dev
    type: dockerimage
    image: eclipse/che-dev:nightly
    env:
      - name: CHE_LOGS_DIR
        value: /home/user/che-logs
    endpoints:
      - name: tomcat
        port: 8080
        attributes:
          public: 'true'
    volumes:
      - name: maven
        containerPath: /home/user/.m2
    memoryLimit: 1.5G
    mountSources: true
```
Will be converted to 
```json
{
    "che-dev": {
        "machines": {
            "che-dev": {
                "attributes": {
                    "memoryLimitBytes": "1500000000"
                },
                "servers": {
                    "tomcat": {
                        "attributes": {
                            "public": "true"
                        },
                        "port": "8080",
                        "protocol": "http"
                    }
                },
                "volumes": {
                    "projects": {
                        "path": "/projects"
                    },
                    "maven": {
                        "path": "/home/user/.m2"
                    }
                },
                "installers": [],
                "env": {
                    "CHE_LOGS_DIR": "/home/user/che-logs"
                }
            }
        },
        "recipe": {
            "type": "dockerimage",
            "content": "eclipse/che-dev:nightly"
        }
    }
}
```
[...]

PR #12716, opened 2019-02-20
Support entrypoint definition in dockerimage recipe based workspaces on k8s and os

metlos:
### What does this PR do?
`$TITLE`.

More specifically, this makes sure that at most 1 machine is defined on dockerimage based workspace and if that machine contains `containerCommand` attribute, it is parsed as a YAML list and supplied as the entrypoint command of the container. Additionally if `containerArgs` is present, it is also parsed as a YAML list and supplied as custom arguments to the entrypoint command.

#12668

#### Release Notes
* [ ] TODO

#### Docs PR
* [ ] TODO

### DIFF
--- a/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/config/MachineConfig.java
+++ b/core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/config/MachineConfig.java
@@ -38,6 +38,25 @@ public interface MachineConfig {
   String MEMORY_REQUEST_ATTRIBUTE = "memoryRequestBytes";
 
   /**
+   * Name of the attribute from {@link #getAttributes()} which, if present, defines the entrypoint
+   * command to be executed in the machine/container.
+   *
+   * <p>The format is a YAML list of strings, e.g. {@code ['/bin/sh', '-c']}
+   */
+  String CONTAINER_COMMAND_ATTRIBUTE = "containerCommand";
+
+  /**
+   * Name of the attribute from {@link #getAttributes()} which, if present, defines the command line
+   * arguments of the entrypoint command specified using the {@link #CONTAINER_COMMAND_ATTRIBUTE}.
+   *
+   * <p>If {@link #CONTAINER_COMMAND_ATTRIBUTE} is not present, the default command defined in the
+   * image is used and the arguments are provided to it.
+   *
+   * <p>The format is a YAML list of strings, e.g. {@code ['-f', '--yes']}
+   */
+  String CONTAINER_ARGS_ATTRIBUTE = "containerArgs";
+
+  /**
    * Returns configured installers.
    *
    * <p>Values can be installer id with version separated with ':' symbol or just id then latest
--- a/infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/dockerimage/DockerImageEnvironmentFactory.java
+++ b/infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/dockerimage/DockerImageEnvironmentFactory.java
@@ -64,7 +64,7 @@ public class DockerImageEnvironmentFactory
       @Nullable InternalRecipe recipe,
       Map<String, InternalMachineConfig> machines,
       List<Warning> warnings)
-      throws InfrastructureException, ValidationException {
+      throws ValidationException {
     checkNotNull(recipe, "Null recipe is not supported by docker image environment factory");
     if (!DockerImageEnvironment.TYPE.equals(recipe.getType())) {
       throw new ValidationException(
@@ -77,9 +77,28 @@ public class DockerImageEnvironmentFactory
 
     checkArgument(dockerImage != null, "Docker image should not be null.");
 
+    ensureSingleMachine(machines);
+
     addRamAttributes(machines);
 
     return new DockerImageEnvironment(dockerImage, recipe, machines, warnings);
+  }
+
+  private void ensureSingleMachine(Map<String, InternalMachineConfig> machines)
+      throws ValidationException {
+    int nofMachines = machines.size();
+    if (nofMachines == 0) {
+      // we create a "fake" machine definition where the rest of the code can put additional
+      // definitions, if needed.
+      InternalMachineConfig emptyConfig = new InternalMachineConfig();
+      // let's just call the machine after the type. The name doesn't matter that much anyway.
+      machines.put(DockerImageEnvironment.TYPE, emptyConfig);
+    } else if (nofMachines > 1) {
+      throw new ValidationException(
+          format(
+              "Docker image environment only supports a single machine definition but found %d.",
+              nofMachines));
+    }
   }
 
   private void addRamAttributes(Map<String, InternalMachineConfig> machines) {
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/environment/convert/DockerImageEnvironmentConverter.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/environment/convert/DockerImageEnvironmentConverter.java
@@ -19,13 +19,14 @@ import io.fabric8.kubernetes.api.model.ContainerBuilder;
 import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.api.model.PodBuilder;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.Map;
 import javax.inject.Singleton;
 import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
-import org.eclipse.che.api.workspace.server.spi.InternalInfrastructureException;
+import org.eclipse.che.api.workspace.server.spi.environment.InternalMachineConfig;
 import org.eclipse.che.workspace.infrastructure.docker.environment.dockerimage.DockerImageEnvironment;
 import org.eclipse.che.workspace.infrastructure.kubernetes.environment.KubernetesEnvironment;
+import org.eclipse.che.workspace.infrastructure.kubernetes.environment.util.EntryPoint;
+import org.eclipse.che.workspace.infrastructure.kubernetes.environment.util.EntryPointParser;
 
 /**
  * Converts {@link DockerImageEnvironment} to {@link KubernetesEnvironment}.
@@ -39,15 +40,27 @@ public class DockerImageEnvironmentConverter {
   static final String POD_NAME = "dockerimage";
   static final String CONTAINER_NAME = "container";
 
+  private EntryPointParser entryPointParser = new EntryPointParser();
+
   public KubernetesEnvironment convert(DockerImageEnvironment environment)
       throws InfrastructureException {
-    final Iterator<String> iterator = environment.getMachines().keySet().iterator();
-    if (!iterator.hasNext()) {
-      throw new InternalInfrastructureException(
-          "DockerImage environment must contain at least one machine configuration");
-    }
-    final String machineName = iterator.next();
+
     final String dockerImage = environment.getRecipe().getContent();
+
+    Map.Entry<String, InternalMachineConfig> e =
+        environment.getMachines().entrySet().iterator().next();
+
+    InternalMachineConfig machine = e.getValue();
+    String machineName = e.getKey();
+
+    ContainerBuilder container =
+        new ContainerBuilder()
+            .withImage(dockerImage)
+            .withName(CONTAINER_NAME)
+            .withImagePullPolicy("Always");
+
+    applyEntryPoint(machine, container);
+
     final Map<String, String> annotations = new HashMap<>();
     annotations.put(format(MACHINE_NAME_ANNOTATION_FMT, CONTAINER_NAME), machineName);
 
@@ -58,18 +71,22 @@ public class DockerImageEnvironmentConverter {
             .withAnnotations(annotations)
             .endMetadata()
             .withNewSpec()
-            .withContainers(
-                new ContainerBuilder()
-                    .withImage(dockerImage)
-                    .withName(CONTAINER_NAME)
-                    .withImagePullPolicy("Always")
-                    .build())
+            .withContainers(container.build())
             .endSpec()
             .build();
+
     return KubernetesEnvironment.builder(environment)
         .setMachines(environment.getMachines())
         .setInternalRecipe(environment.getRecipe())
-        .setPods(ImmutableMap.of(POD_NAME, pod))
+        .setPods(ImmutableMap.of(machineName, pod))
         .build();
+  }
+
+  private void applyEntryPoint(InternalMachineConfig machineConfig, ContainerBuilder bld)
+      throws InfrastructureException {
+    EntryPoint ep = entryPointParser.parse(machineConfig.getAttributes());
+
+    bld.withCommand(ep.getCommand());
+    bld.withArgs(ep.getArguments());
   }
 }
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/environment/util/EntryPoint.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/environment/util/EntryPoint.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2012-2018 Red Hat, Inc.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *   Red Hat, Inc. - initial API and implementation
+ */
+package org.eclipse.che.workspace.infrastructure.kubernetes.environment.util;
+
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+import java.util.Objects;
+
+/** Represents an entry-point definition parsed from a string using the {@link EntryPointParser}. */
+public final class EntryPoint {
+
+  private final List<String> command;
+  private final List<String> arguments;
+
+  public EntryPoint(List<String> command, List<String> arguments) {
+    this.command = ImmutableList.copyOf(command);
+    this.arguments = ImmutableList.copyOf(arguments);
+  }
+
+  /** @return unmodifiable list representing the command of the entrypoint */
+  public List<String> getCommand() {
+    return command;
+  }
+
+  /** @return unmodifiable list representing the arguments of the entrypoint */
+  public List<String> getArguments() {
+    return arguments;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    EntryPoint that = (EntryPoint) o;
+    return Objects.equals(command, that.command) && Objects.equals(arguments, that.arguments);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(command, arguments);
+  }
+
+  @Override
+  public String toString() {
+    return "EntryPoint{" + "command=" + command + ", arguments=" + arguments + '}';
+  }
+}
--- a/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/environment/util/EntryPointParser.java
+++ b/infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/environment/util/EntryPointParser.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright (c) 2012-2018 Red Hat, Inc.
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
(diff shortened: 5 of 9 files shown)


core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/config/MachineConfig.java
```diff
@@ -37,6 +37,20 @@
...
+   * Name of the attribute from {@link #getAttributes()} which, if present, defines the entrypoint
+   * command to be executed in the machine/container.
+   */
+  String CONTAINER_COMMAND_ATTRIBUTE = "containerCommand";
+
+  /**
+   * Name of the attribute from {@link #getAttributes()} which, if present, defines the command line
+   * arguments of the entrypoint command specified using the {@link #CONTAINER_COMMAND_ATTRIBUTE}.
+   *
+   * <p>If {@link #CONTAINER_COMMAND_ATTRIBUTE} is not present, this attribute is ignored.
```
sleshchenko: I'm not sure that it's good behavior. As far as I know, it's a normal scenario to specify only arguments for entry point from a container. It's like ENTRYPOINT and CMD in Dockerfile.
Please check if Kubernetes allows specifying arguments only. And let's do in the same way.

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/dockerimage/EntryPointParser.java
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    try {
+      List<String> commandList = parseAsList(command);
+      List<String> argList = args == null ? emptyList() : parseAsList(args);
+
+      return new EntryPoint(commandList, argList);
+    } catch (IOException e) {
+      throw new InfrastructureException(
+          "Failed to apply entry point configuration for container.", e);
```
sleshchenko: Looks like a different message should be here

sleshchenko (approved):
Draft LGTM.

Let's move on PR ready to review
It would be nice to see different commits for entrypoint and additional check for dockerimage 
environment.

sleshchenko:
@garagatyi Please take a look too

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/dockerimage/EntryPointParser.java
metlos: Huh, yes, remnant of the original location of the code :man_facepalming:

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/config/MachineConfig.java
metlos: :+1:  good point, will check

garagatyi (commented):
To be honest I think that this functionality should be implemented with a different approach. 
Current solution brings a lot of code that not really needed. What if we just convert Dokerimage environment to K8s  environment using existing converter and add entrypoint, command arguments, etc?
This should eliminate a need for all this code.
@sleshchenko @metlos WDYT?

sleshchenko:
@garagatyi 
A lot of code that contains this PR is related to converting workspace config attribute to yaml list (that may be a single string, strings list). And we can avoid having such code by declaring that this list MUST contain comma-separated strings.

I do not see how your proposal may reduce the amount of code. Can you elaborate more?

metlos:
> To be honest I think that this functionality should be implemented with a different approach.
> Current solution brings a lot of code that not really needed. What if we just convert Dokerimage environment to K8s environment using existing converter and add entrypoint, command arguments, etc?
> This should eliminate a need for all this code.
> @sleshchenko @metlos WDYT?

Our dockerimage environment doesn't understand entrypoints, AFAIK, and enhancing docker support when we're about to drop it is IMHO not the way to go. That's why I only added the support for it to k8s env.

As @sleshchenko, I don't see where would save a lot of code here though if we went down the route of adding support for entrypoint to dockerimage and added support for the also to the converter.

metlos:
> A lot of code that contains this PR is related to converting workspace config attribute to yaml list (that may be a single string, strings list). And we can avoid having such code by declaring that this list MUST contain comma-separated strings.

In this case, I think it would be counterproductive if we wanted to use 2 different formats for encoding entrypoints in workspace config and in the devfile. If we wanted to have a custom "comma separated list of commands", we'd need a parser (in workspace config) and serializer (in devfile) for that. I the approach I chose, we can use just the `YAMLMapper` for both (the serializer is not part of this PR but I added it in https://github.com/eclipse/che/pull/12761).

garagatyi:
@sleshchenko If we avoid direct implementation of entrypoint and command support in dockerimage workspace recipe (Che 6 stuff) we would do:
1. In `DockerimageToolToWorkspaceApplier` convert devfile dockerimage tool into k8s environment directly. Using existing dockerimage to k8s env converter or even directly tool to k8s.
1. Add entrypoint/args into k8s env

With this approach we would not need anything from this PR at all.  

> Our dockerimage environment doesn't understand entrypoints, AFAIK, and enhancing docker support when we're about to drop it is IMHO not the way to go. That's why I only added the support for it to k8s env.
As @sleshchenko, I don't see where would save a lot of code here though if we went down the route of adding support for entrypoint to dockerimage and added support for the also to the converter.

Well, I'm suggesting not to improve dockerimage recipe at all since it brings new attributes, parsing, etc. And this PR adds entrypoint/args case as machine config attribute which would not be needed when we replace workspace config with devfile usage, eventually.

garagatyi:
@metlos I think improving outdated format with additional attributes is not productive.

metlos:
> @metlos I think improving outdated format with additional attributes is not productive.

ATM devfile "appliers" convert devfile into workspace config. If we were to take a different approach here, we would have to reimplement the whole of devfile conversion, which I think is not practical at this point in time. The amount of code added here is not that huge to justify rewriting everything devfile we have.

garagatyi:
After discussion with @sleshchenko and @metlos we agreed that:
- changing this solution to the one I suggested would simplify the code, but might lead to a very confusing UX for end users when they specify dockerimage tool and endup with kubernetes recipe in the workspace created from the dockerimage devfile tool. Presence of user facing workspace config here leads to this situation. We think that direct usage of devfile would be way better than continuing usage of workspace config 
- this solution is a hack, but because of significant limitations of workspace config we can't avoid it
- we should not document new abilities for Che 6 style workspace config and only promote usage of devfile for setting entrypoint and args

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/dockerimage/EntryPointParser.java
```diff
@@ -0,0 +1,112 @@
...
+  /**
+   * Parses a single entry in the format expected by the entry point attribute, i.e. a YAML list of
+   * strings or a YAML string.
+   *
+   * @param data the string data to parse
+   * @return a parsed list or null if the data is not a YAML list or YAML string
+   * @throws IllegalArgumentException if the parsing of the {@literal entry} failed for other
+   *     reasons
+   */
+  public @Nullable List<String> parseEntry(String data) throws IllegalArgumentException {
```
sleshchenko: Why it should be public?

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/dockerimage/EntryPointParser.java
```diff
@@ -0,0 +1,112 @@
...
+
+  /**
+   * Parses the attributes contained in the provided machine config and produces an entry point
+   * definition.
+   *
+   * @param machineAttributes the attributes of a machine to extract the entry point info from
+   * @return an entry point definition, never null
+   * @throws InfrastructureException on failure to parse the command or arguments
+   */
+  public EntryPoint parse(Map<String, String> machineAttributes) throws InfrastructureException {
```
sleshchenko: It is used only in `kubernetes-infrastructure` module. Why did you put `EntryPointParser` and `EntryPoint` into `docker-environment` module? I think it would be better to put it to `kubernetes-infrastructure` since very soon it will be used for Kubernetes/OpenShift Environments.
WDYT?

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/dockerimage/EntryPointParser.java
```diff
@@ -0,0 +1,112 @@
...
+    String command = machineAttributes.get(MachineConfig.CONTAINER_COMMAND_ATTRIBUTE);
+    String args = machineAttributes.get(MachineConfig.CONTAINER_ARGS_ATTRIBUTE);
+
+    List<String> commandList =
+        command == null
+            ? emptyList()
```
sleshchenko: Are you sure that we should use the empty list here? Won't we override a default command/arguments by specifying an empty list for pod container?

sleshchenko (approved):

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/dockerimage/EntryPointParser.java
metlos: Well, it it not used anywhere but the thinking was to expose the raw parsing of the single piece of data, if needed. But you are right, technically, this could be private.

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/dockerimage/EntryPointParser.java
metlos: That is mostly my initial confusion about the roles of the individual modules. I will move it into kubernetes infrastructure.

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/dockerimage/EntryPointParser.java
metlos: I checked this and nope, that shouldn't be the case. The k8s client interprets empty list as a "undefined". E.g. if you pass null, the `ContainerBuilder` internally creates an empty list anyway.

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/dockerimage/EntryPointParser.java
```diff
@@ -0,0 +1,112 @@
...
+            if (n.getNodeType() != STRING) {
+              return null;
+            }
+            ret.add(n.textValue());
+          }
+          return ret;
+        case STRING:
+          return singletonList(node.textValue());
+        default:
+          return null;
```
sleshchenko: Maybe it would be better to throw an exception here instead of returning null?

metlos: Yeah, I admit to being lazy here :wink:

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/dockerimage/EntryPointParser.java
```diff
@@ -0,0 +1,112 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import org.eclipse.che.api.core.model.workspace.config.MachineConfig;
+import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
+import org.eclipse.che.commons.annotation.Nullable;
+
+/**
+ * Can be used to parse container entry-point definition specified as a YAML list of strings or a
```
sleshchenko: See https://github.com/eclipse/che/pull/12761#discussion_r260754787. 
maybe `list of strings` would be enough here.

metlos: You mean to not allow a simple string or have a custom format of "list of strings" not based on YAML?

sleshchenko: I mean allowing only list of strings, but not string type.
See fabric8 client kubernetes JSON Schema https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-model/kubernetes-model/src/main/resources/schema/kube-schema.json#L5198

I understand benefit of supporting YAML Strings (single string or list of strings) but we follow Kubernetes model here, and looks like they do not support it. For us, it would be much simpler too.

metlos: :+1:

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/environment/util/EntryPointParser.java
```diff
@@ -0,0 +1,114 @@
...
+   * <p>The {@code null} return value is meant to be reinterpreted as some kind of exception with a
+   * more useful error message than the one we could provide here.
+   *
+   * @param data the string data to parse
+   * @return a parsed list or null if the data is not a YAML list
+   * @throws IllegalArgumentException if the parsing of the {@code entry} failed for other reasons
+   */
+  @VisibleForTesting
+  @Nullable
+  List<String> parseEntry(String data) throws IllegalArgumentException {
```
sleshchenko: Do we still need this method?
Can not we rewrite `parseAsList` in the following way:
```java
  private List<String> parseAsList(String data, String attributeName)
    try {
      return mapper.readValue(data, new TypeReference<ArrayList<String>>() {});
    } catch (IllegalArgumentException | IOException e) {
      throw new InfrastructureException(
          format(
              "Failed to parse the attribute %s as a YAML list. The value was: %s. Error: %s",
              attributeName, data, e.getMessage()), e);
    }
  }
```

core/che-core-api-model/src/main/java/org/eclipse/che/api/core/model/workspace/config/MachineConfig.java
```diff
@@ -37,6 +37,21 @@
    */
   String MEMORY_REQUEST_ATTRIBUTE = "memoryRequestBytes";
 
+  /**
+   * Name of the attribute from {@link #getAttributes()} which, if present, defines the entrypoint
+   * command to be executed in the machine/container.
```
sleshchenko, line 42: Please add info about its format: Array of strings. An example would be useful as well, `["sh","-c"]`

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/environment/util/EntryPointParser.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import org.eclipse.che.api.core.model.workspace.config.MachineConfig;
+import org.eclipse.che.api.workspace.server.spi.InfrastructureException;
+import org.eclipse.che.commons.annotation.Nullable;
+
+/**
+ * Can be used to parse container entry-point definition specified as a YAML list of strings or a
+ * single YAML string.
```
sleshchenko: `single YAML string` is outdated info here

metlos:
ci-test

amisevsk (commented):
A few comments

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/environment/convert/DockerImageEnvironmentConverter.java
```diff
@@ -39,31 +40,49 @@
...
+      throw new InfrastructureException(
+          format(
+              "Environment of type '%s' doesn't support multiple machines, but contains machines: %s",
+              environment.getType(), String.join(", ", environment.getMachines().keySet())));
+    }
+
+    InternalMachineConfig machine = null;
+    String machineName = POD_NAME;
+
+    if (!environment.getMachines().isEmpty()) {
```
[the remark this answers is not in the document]
sleshchenko: Maybe it would be even simpler if `DockerImageEnvironmentFactory` provisions empty machine config if it is not present?
It could fix one small bug: machine memory attribute is not propagated if `dockerimage`  recipe does not have machine config specified. (I can be wrong but it's what I see from code 
https://github.com/eclipse/che/blob/aaa8f424cd198e68e128bca939d9ac058b9ea52e/infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/dockerimage/DockerImageEnvironmentFactory.java#L85-L89

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/environment/util/EntryPoint.java
```diff
@@ -0,0 +1,57 @@
...
+  private final List<String> command;
+  private final List<String> arguments;
+
+  public EntryPoint(List<String> command, List<String> arguments) {
+    this.command = command;
+    this.arguments = arguments;
+  }
+
+  public List<String> getCommand() {
+    return command;
```
sleshchenko, line 31: It will be safer if this POJO has getters with null checking, as we have in other POJOs https://github.com/eclipse/che/blob/5907d1cad5c621e7d726cbeb27c89127fcc4d28b/wsmaster/che-core-api-workspace/src/main/java/org/eclipse/che/api/workspace/server/model/impl/WorkspaceConfigImpl.java#L172-L178

sleshchenko (approved):
LGTM

metlos:
ci-build

amisevsk (approved):

metlos:
ci-test

metlos:
ci-test

garagatyi (approved):

metlos:
ci-test, pretty please?

metlos:
ci-test

metlos:
ci-test

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/environment/util/EntryPoint.java
```diff
@@ -0,0 +1,62 @@
...
+    this.command = unmodifiableList(requireNonNull(command));
+    this.arguments = unmodifiableList(requireNonNull(arguments));
+  }
+
+  /** @return unmodifiable list representing the command of the entrypoint */
+  public List<String> getCommand() {
+    return command;
+  }
+
+  /** @return unmodifiable list representing the arguments of the entrypoint */
```
amisevsk, line 34: "Unmodifiable" is deceptive here, since the list elements can be changed. Consider something like `ImmutableList.copyOf(arguments)` and similar for `getCommand()`

infrastructures/docker/environment/src/main/java/org/eclipse/che/workspace/infrastructure/docker/environment/dockerimage/DockerImageEnvironmentFactory.java
```diff
@@ -77,11 +77,30 @@ protected DockerImageEnvironment doCreate(
...
+    ensureSingleMachine(machines);
+
     addRamAttributes(machines);
 
     return new DockerImageEnvironment(dockerImage, recipe, machines, warnings);
   }
 
+  private void ensureSingleMachine(Map<String, InternalMachineConfig> machines)
+      throws ValidationException {
+    int nofMachines = machines.size();
```
amisevsk, line 89: I'd suggest `numberOfMachines` but it's not terribly important.

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/environment/util/EntryPoint.java
metlos: Not sure if we're not overdoing it. The list cannot be changed through the returned reference, but of course it can change if something changes the list through the original reference passed to the ctor.

If that really is something we want to be covered for then I'd rather make the ctor package private and be only invoked from the parser, by which we can control the only code path that would hold such mutable reference.

metlos: I mean your observation is of course 100% valid, but I'd be worried about it in some highly reusable piece of code where we could not assume the way it'd be used. IMHO, this is not such case.

metlos:
rebased and squashed on top of latest master before merge

infrastructures/kubernetes/src/main/java/org/eclipse/che/workspace/infrastructure/kubernetes/environment/util/EntryPoint.java
amisevsk: Ah sorry you are right, I had missed the `ImmutableList.copyOf(arguments);` above

Issue #12337, opened 2019-01-04
[Che 7] Cannot add new env variable, server or volume to Che 7 stack


SkorikSergey:
### Description
This issue reproduces only when all 3 machines("ws/theia-ide", "ws/che-machine-exec" and "ws/dev") exist in **Machines** list.

**Reproduction Steps:**
1. Create and start workspace from "**Che 7**" stack.
2. Open this workspace details page.
3. Click on **Env Variables** tab.
4. Select **ws/dev** machine from machines list.
5. Click on **Add Evn Variable** button.
6. Create a new variable in opened "**Add a new environment variable**" form and click on **Add** button.

**Expected behavior:**
"**Add a new environment variable**" form closed and the new variable added to variables list. **Save** changes button in visible and enabled.

**Observed behavior:**
"**Add a new environment variable**" form is still opened. New variable is visible in variables list but not added to workspace config.

**OS and version:**
CHE 6.17.0-SNAPSHOT, Fedora 28.0

**Additional information:**
**Screencast:**
![screencast-che-eclipse-che 172 19 20 234 nip io-2019 01 04-17-15-41](https://user-images.githubusercontent.com/7760565/50695152-8499c000-1044-11e9-8619-897cddb872eb.gif)

AndrienkoAleksandr:
@ashumilova I guess it could be closed?

PR #12725, opened 2019-02-21
Fix editing running workspace config with tools containers and multiuser

ashumilova:

### What does this PR do?
The issue is reproduced on multi-user Che, but existed in single user too. The idea is that when user edits containers in running workspace and there are tooling containers in runtime - those tooling containers are added to environment (but should not) and thus the problem occurs. In single user such environment can be saved, in multi user we have error, that environment contains machines that are not defined in recipe.

Issue was discovered, when during work on this one - https://github.com/eclipse/che/issues/12337


garagatyi:
@ashumilova I believe that it is a bug that such a configuration can be saved, please create a ticket for that.

ashumilova:
@garagatyi, catched, forgot I promised that. 5 minutes.

ashumilova:
ci-build

ashumilova:
ci-test

benoitf (approved):

olexii4 (approved):

PR #12744, opened 2019-02-22
Different improvements of script that helps to build Devfile Docs

sleshchenko:
### What does this PR do?
This PR contains different improvements of the script that helps to build Devfile Docs:
- Fix build in dockerimage by using https instead of ssh to avoid generation of ssh keys
- Fix docs generation without deploying
- Add --folder argument to build_docs.sh script that allows to save built docs file
- Do not clean up working directory to make script results accessible
- Improve error message when Github token is missing
- Fix an ability to override commit message while publishing

N/A

#### Release Notes
N/A

#### Docs PR
N/A

### DIFF
--- a/wsmaster/che-core-api-devfile/src/main/resources/build_docs.sh
+++ b/wsmaster/che-core-api-devfile/src/main/resources/build_docs.sh
@@ -44,7 +44,7 @@ build_native() {
    mkdir -p ${TMP_DIR}/schema
    cp -f schema/* ${TMP_DIR}/schema
    cd ${TMP_DIR}
-   git clone -b '1.1.0' --single-branch git@github.com:adobe/jsonschema2md.git
+   git clone -b '1.1.0' --single-branch https://github.com/adobe/jsonschema2md.git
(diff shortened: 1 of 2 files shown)


wsmaster/che-core-api-devfile/src/main/resources/build_docs.sh
```diff
@@ -103,33 +109,73 @@ parse_args() {
...
 if [[ "$IS_DEPLOY" == "true" ]]; then
     deploy
 fi
-cleanup
+
+if [[ "$FOLDER" ]]; then
+    copyDocs
+fi
+
+if [[ "$IS_DEPLOY" == "true" ]]; then
```
mshaposhnik: check this block again, something looks wrong

sleshchenko: good catch. Thanks!

mshaposhnik (approved):

skabashnyuk (approved):

